Issue #160, opened 2014-02-11
Speed of update for stack of images


pburdet:
When plotting a stack of images, the speed of update when changing "stack elements" is rather slow. We should implement a "fast" display suppressing all options that might slow down, such as color scale, axes, title, index, same contrast for all images (calculate once). In fact just a display of the image. I don't know what will be the gain, but it should help. Should we call it, plot_simple() plot_fast() or should it an option in plot (only for image)?

vidartf:
As far as I know, the fastest way to achieve this would be to use 
```python
# ax is MLP axes
ims = ax.get_images()[-1]    # Select top most image if there are several in axes
ims.set_data(my_2d_array)
```
I'm unsure whether MPL runs the array through normalization/colormap first though, so that might have to be added.

francisco-dlp:
@vidartf, isn't the method to update images that you propose equivalent to what is [already implemented](https://github.com/hyperspy/hyperspy/blob/master/hyperspy/drawing/image.py#L305) ?

Isn't this fixed by #452?

vidartf:
@francisco-dlp I didn't check the existing code before making the comment, but as it is already like that, and considering the rest of the code that is there, I don't really see any way to speed up the drawing in `update()` (just make sure `auto_contrast` is passed as `False`).

francisco-dlp:
Fixed in #452.

vidartf:
If anyone wants to look more at this, I suggest running a profiler and trying to determine which parts of the code is getting called the most during stack navigation.

francisco-dlp:
Ok, I'll leave this open then in case that someone finds other ways to speed things up, but I'll change it to the Wish List milestone.

vidartf:
I ran some tests on stacks of 2k x 2k images, and from my profiler output, this is the main areas where time is spent:
* Around half the time was spent in ```hyperspy.drawing.image.optimize_contrast()```. Making sure that ```auto_contrast``` is False on navigation removed that time.
* Close to the other half was spent in ```matplotlib.image._get_unsampled_image()```. Detailed profiling shows most of the time is spent with converting the image into a RGBA byte array via the colormap amongst others. I don't think there is a reasonable way to speed this bit up.

vidartf:
[...]

PR #413, opened 2014-12-12
ENH: Disconnected Signal(.data) iterators

vidartf:
Implements two iterators for iterating through the navigation axes of a Signal, and returning either a sub-Signal, or a slice of the data array. As it works with a copy of the AxesManager, it does not trigger any indexing connections on the original one, which can improve speed for those applications where those triggers are unnecessary. A good example for this is when the iteration triggers replots for each step, as this will many-double the time of processing.

PR #454, opened 2015-01-29
NEW: Implementation of plot_images utility function

### LINKED ISSUE #452: plot_images
The `utils.plot.plot_spectra` function contributed by @magnunor has proven to be so useful that now many of us miss a  `utils.plot.plot_images` equivalent function.

jat255:
Resolves #452.

This PR can be used as a starting point for the `plot_images` function. The majority of the code has been lifted from the `plot_decomposition_loadings()` method, and adapted so I could plot whichever images I wanted.

HOLDS:
- [x] Waiting on #160 before merging

FOR DEBATE:
- [ ] debate merits of `_deepcopy`
- [x] ~~change default `tight_layout` ?~~

TODO:
- [ ] add testing notebook to hyperspy demos
- [x] ~~let x and y axes be undefined and remove if the case~~
- [x] ~~look into `perc` with `vmin/vmax` in `image.plot()`~~
- [x] ~~replace example of EDS data with data from pburdet~~
- [x] ~~remove print statements from `image.plot()` code~~
- [x] ~~implement (or not) for 3D RGB images~~
- [x] ~~docstring like numpy style~~
- [x] ~~raise `ValueError` for incorrect scalebar input~~
- [x] ~~investigate reshaping~~
- [x] ~~pep8 compliance~~
- [x] ~~move imports to beginning of file~~
- [x] ~~Documentation in user guide~~
- [x] ~~improve labeling of similar signals~~
- [x] ~~Prevent `plot_images()` from causing iteration of prior calls to `image.plot()`~~ 
- [x] ~~If images have different scales, they get squeezed. The aspect of plt.imshow should be optimized as it is in image.plot()~~
- [x] ~~refactor `plot_scalebar` as `scalebar` (`None, 'all', or list of int choosing which plots should have the scalebar`)~~
- [x] ~~one option for `axes` (`None, 'default' (which is ticks and labels), 'ticks' (no labels)`)~~
- [x] ~~Handling of input if a `Signal` instance is supplied instead of a list (like the output of `get_decomposition_loadings()`~~ This actually already worked because Signal is iterable (I think), so I just added some checks to make sure it's a Signal if not a list. I'd be happy for suggestions of improvement on this part.
- [x] ~~Adding option of fixing the colormap so it is shared between all images, rather than having independent ones for all the individual images~~
- [x] ~~Plot scale bar option~~
- [x] ~~Plot axes option~~ 
- [x] ~~Pass extra args and kwargs to `imshow`~~
- [x] ~~Option to disable titles~~
- [x] ~~Add figure keyword to pass to existing MPL figure~~
- [x] ~~Set default figsize to `k * (per_row, rows)` where `k = np.max(plt.rcParams['figure.figsize']) / np.max(per_row, rows)`~~
- [x] ~~change default `label_list` to list of signal titles (`metadata.General.title`)~~
- [x] ~~enable plotting when `label_list` is not the same length, printing useful output to user~~
- [x] ~~implement single option `label`, which can be a `str` or a `list`~~
- [x] ~~add support for RGB images~~
- [x] ~~refactor `signals` to `images`~~
- [x] ~~implement padding option to control spacing between images~~
- [x] ~~Investigate switching to GridSpec for layout?~~
    - *I don't think this is necessary, since using `rect` in `plt.tight_layout` solves the issue of overlapping colorbar*
- [x] ~~implement single option for `colorbar` (`None, 'default', or 'single'`)~~
- [x] ~~Fix colorbar placement~~
- [x] ~~default cmap could be the one set generally with pyplot.set_cmap()~~
- [x] ~~The label of the axis should be the same as image.plot(): "name (units)"~~
- [x] ~~Should this function return a list of axes as plot_spectra?~~


francisco-dlp:
Nice! Of course this needs to be documented in the User Guide to be ready to merge.

What do you thing about the following suggestions:
- Add plot scale bar option.
- Add plot_axes option.
- Pass extra args and kwargs to `imshow`.
- Add a figure keyword to pass an existing MPL figure.
- Se the default figsize to:  `k * (per_row, rows)` where `k = np.max(plt.rcParams['figure.figsize']) / np.max(per_row, rows)`

jat255:
I think all of those are good ideas. I've added them to the todo list. I'm working on the shared colorbar option currently.

jat255:
I've implemented the scalebar code using the Scale_Bar widget class. Is this the recommended way? It seems to work okay for me, but it takes what seems like a long time for the bars to appear (it's almost instantaneous, but I can watch them showing up, which seems like it is slow).

I suppose people won't be plotting enough images to make this an issue, but I was curious about the slowdown.

jat255:
The default figsize formula needed some tweaks, but I think I've got it working the way you intended. Care to comment on if it looks alright?

hyperspy/drawing/utils.py
```diff
@@ -333,6 +334,188 @@ def _set_spectrum_xlabel(spectrum, ax):
...
+            # If using independent colorbars, add them
+            if plot_colorbar and not single_colorbar:
+                div = make_axes_locatable(ax)
+                cax = div.append_axes("right", size="5%", pad=0.05)
+                plt.colorbar(im, cax=cax)
+
+    # If using a single colorbar, add it
+    if plot_colorbar and single_colorbar:
+        f.subplots_adjust(right=0.8)
+        cbar_ax = f.add_axes([0.9, 0.1, 0.03, 0.8])
```
jat255: This is not very flexible, and leads to the colorbar overlapping some of the images if the `per_row` value is high enough (see figure). There must be a better way to do this that makes it more flexible?

figure_1

pburdet:
Nice function. I had a go with a list of images exported by `get_lines_intensity` and it works well. I have few suggestion.

```python
utils.plot.plot_images(s.get_lines_intensity())
```


- The default label_list could be the `metadata.General.title` of each images
- `single_colorbar` overlaps the images as you spotted. The should be a way to place it on the right of the plots.

figure_8

- I got a problem with `per_row = 2` and `per_row = 1`

capture

figure_2

- When all options disable, the images are too far apart for my taste. Could we have a "padding" option?

figure_10

- You should be able to plot if the length of the `lable_list` is different from the length of the images list. If it is too long take only the first one, if it is two short repeat the label_list.

- The difference between `label_list` and `signal_label` is not easy too understand. I suggest one single option call `labels`. If it is a `str` then it works as `signal_label`, if it is a list of `str` then it works as `label_list`.

- I didn't manage to make `scalebar_color` to work.

pburdet:
Following...

- One option for `label`. It could be None, 'default' (the title of the images), str or a list of str.

- One option for `colorbar`. It could be None, 'default' or 'single'

- `plot_scalebar` should be `scalebar` for consistency. It could be None, 'default', 'single' (could be interesting to have this option)

- One option for `axes`. None, 'default', 'ticks' (only the ticks)

jat255:
Thanks for the comments, @pburdet. I'll work on implementing these suggestions.

As for the `per_row` issue, I cannot reproduce it on my system. Using the call
```
utils.plot.plot_images(si_EDS.get_lines_intensity(), per_row=<2 OR 1>)
``` 
I get the following figures, which look okay (aside from the complete lack of signal, but that's a different problem :smirk:) 

perrow1
perrow2

francisco-dlp:
It just came to my mind that  `plot_images` needs to support RGB images. It is not difficult to implement, just add:

```
from hyperspy.misc import rgb_tools

 if rgb_tools.is_rgbx(data):
    plot_colorbar = False 
    data = rgb_tools.rgbx2regular_array(data, plot_friendly=True)

```

francisco-dlp:
By the way, wouldn't it be better to rename "signals" to "images"?

jat255:
Probably :)

jat255:
I believe RGB plotting should now be fully implemented. I tried to make it as signal-agnostic as possible, so it is possible to plot a mix of RGB images and regular signals. Here is example output of three RGB images with 2 line intensity Signals:
```
utils.plot.plot_images([rgbIm1, rgbIm2, si_EDS.get_lines_intensity()[1], rgbIm3, si_EDS.get_lines_intensity()[2]], plot_colorbar=True, per_row=3)
```
mixedsignals

Please feel free to test this and make sure I didn't leave anything out.

pburdet:
Have you try more than 9 images?

pburdet:
I really like this plot_images function. I merge it in my working branch and use it all the time. A few more suggestions:

- The default cmap could be the one set generally with `pyplot.set_cmap()`
- The label of the axis should be the same as `image.plot()`: name (units)
- If  images have different scales, they get squeezed. The `aspect` of `plt.imshow` should be optimized as it is in `image.plot()`

jat255:
Sorry for the long delay on getting back working on this. I'm having some trouble using tight_layout. One problem seems to be that if the titles for each image are long, and overlap significantly, tight_layout bombs the whole figure and doesn't give an error as to what happened. (I think this is what was happening in @pburdet https://github.com/hyperspy/hyperspy/pull/454#issuecomment-72194135) As such, I can't figure out how to fix it. 

I guess a bad (but perhaps necessary) solution would be to warn the user if titles are long/overlapping, either in the documentation or with a message at runtime. Does anyone have experience with this issue?

francisco-dlp:
One way around it could be to wrap the title text, that's what with do with [the title of the figures](https://github.com/hyperspy/hyperspy/blob/master/hyperspy/drawing/figure.py#L34). Would that work?

jat255:
@francisco-dlp, thanks for the tip. That seems to solve (or at least workaround) the problem. Ultimately, I'm still not sure what's failing, but if the label wraps, it seems to work okay. I've put the default in as 30 characters, but the parameter is configurable as `labelwrap`

francisco-dlp:
Excellent! Did you have a look at the pull request that I sent you some time ago?

jat255:
Yes, I did, thank you! I'm working on incorporating it now. I got caught up with a few other things, but I'm trying to get this function pretty much wrapped up today so it can be included with the 0.8 release.

jat255:
So the multi-dimensional image should be working now. I left the default as the Signal title, rather than just enumerating a list. This option is still available if you provide a single string for the `label` parameter, but I think the titles make more sense as the default (although I'm certainly open to discussion about this).

I tried to incorporate @francisco-dlp's comments with what I already had, and here's an example of it in action:

```python
# rgbIm1 is a preloaded RGB image for testing
# si_EDS is a test signal for plotting bss_loadings
im0 = signals.Image(np.random.random((64,64)))
im0.metadata.General.title = 'im0'
im1_3 = signals.Image(np.random.random((2, 3,64,64)))
im1_3.metadata.General.title = 'multi-image'
im4 = signals.Image(np.random.random((64,64)))
im4.metadata.General.title = 'im4'

utils.plot.plot_images([im0, im1_3, im4, rgbIm1,si_EDS.get_bss_loadings()[1]],colorbar='single')
```

plot_images

jat255:
I've also added a way to shut off the `tight_layout` command, since it causes all sorts of issues:
Notably: if you have too many images, if the labels are too long, if the per_row value is too short (or long, in some cases)
In short, it works sometimes, but causes a lot of problems in fringe cases, in which I think it is not too much to let the user figure it out...

jat255:
@pburdet, you mentioned that "If images have different scales, they get squeezed. The aspect of `plt.imshow` should be optimized as it is in `image.plot()`"

I'm not overly familiar with how the aspect ratio is set in `image.plot()`. It looks like it is set [here](https://github.com/hyperspy/hyperspy/blob/master/hyperspy/drawing/image.py#L116-L128) and used [here](https://github.com/hyperspy/hyperspy/blob/master/hyperspy/drawing/image.py#L143-L159). This looks a little complex for this little helper function, but we can have a go at it if you think it's best.

Just supplying `aspect='auto'` to `imshow()` appears to work decently (in combination with `extent`) to make the images appropriate for display. Using two images with different scales (1st has equal pixel size of 0.5nm, 2nd has unequal pixel sizes  of 1 and 3.4 nm), I get the following output:

plot_images_aspect

Is this suitable?

jat255:
I've implemented control of the axis decorations. Four options are available, and described in the documentation of the function. They appear as such (option used is in the title of each window):

plot_image-axis_decorations

pburdet:
Really good job. Very nice. First I tried the default I obtained that

figure_5

Then with some tuning:

```python

utils.plot.plot_images(im, per_row=5, tight_layout=False, axes_decor='off',
                       colorbar='single', label=[str(i) for i in im.axes_manager[0].axis])

```

figure_19

Base on this example, we might have some fine tuning. 

- I was trying to add the "°" character in this example (this is a tilt series), doesn't work. Is it possible to support non-ascii character? and to support latex format (e.g. "{\mu m}") as in axes_manager?
- In this example, the label could be the value of the axes manager.
- If all label share the same beginning, couldn't we have that part as the figure title and what changes in each label? It will prevent in part the problem of too long label and improve the default display.

About the other fixes:
- Multi-dimension works nicely, great feature. Have you tried im.plot() and then plot_image(im)? im.plot() is moving as the index are accessed. Is there any way to avoid it? It slows down the plotting. It would be nice at least to reset the index at their previous values.
- Axis decoration works nicely, perfect.
- I'm not sure how to use the "padding" option. Do we need it? Disabling the tight_layout do the job for me. But it might sound a bit counter intuitive for the user. Shouldn't we disable it all the time? What does tight_layout brings?
- aspect = "equal" works, but I didn't manage to disable it. I need it for a figure with signals.Image(np.random.random([15,80])). But for most of my images, I want the true aspect ratio. This is not straight forward and I found that im.plot() works nicely. I think plot_image should use the same function.

jat255:
I've changed the implementation of the scalebar code. There is now an option to supply individual scalebars for each image. Currently, a multi-dimensional image is all or nothing, but if this is really necessary, I guess it could be implemented by providing a `tuple` in place of an `int`, specifying which ones get a scalebar. This could get messy pretty quickly though, so I think the way it is is okay.

Here is an image of it in action, using the following code (again, window titles show which option was used):

```python
plt.close('all')
for i in [None, False, 'all', True, [0], [1], [0,1]]:
    utils.plot.plot_images([si_EDS.get_bss_loadings(),si_EDS2.get_bss_loadings()[1]],
                               axes_decor='ticks', tight_layout=True, label=['0','1'],colorbar='single',labelwrap=20,per_row=4,
                               padding={'hspace':.2,'wspace':0.2},scalebar=i)
    fig=gcf()
    fig.canvas.set_window_title(repr(i))
```

plot_image-scalebar

jat255:
@pburdet, thanks for the comments. I think a couple of them are already implemented as is:
 * Non-ascii characters can be input in latex format using the typical matplotlib way of enclosing the text in `$$`:
  ```python
utils.plot.plot_images(im, per_row=5, tight_layout=False, axes_decor='off',
colorbar='single', label=[str(i) + '$\mu m ^\degree$' for i in im.axes_manager[0].axis], cmap='RdYlBu_r')
```
 plot_image-latex2

   * So far, this has been suitable for my needs, but if there is a different way it is implemented in `axes_manager`, I'd be open to changing it
 * I agree that titling the whole figure (if all the images share the same base-title) could save space and improve the default appearance. How would you recommend implementing this? I could picture something like checking if `metadata.General.title` is the same for all the images provided, and then having a special labeling case if this is true, but there's probably a better way.
 * You can thank @francisco-dlp for the mult-dimension images. It does work very well, and makes this function really useful for quickly visualizing a whole stack of images. 
  * It is interesting that the im.plot() changes each time; I hadn't noticed that behavior before, but it makes sense why it happens. I would propose that when plotting each image (around [here](https://github.com/jat255/hyperspy/blob/ENH_plot_images_function/hyperspy/drawing/utils.py#L535-L538)), we check if `ims` is an instance of Signal, and if so, replace `ims` with a `deepcopy()` of itself. This way, the images in `plot_images` would be disconnected from its source, at the expense of some additional memory... What do you think of this solution?
 * `tight_layout` helps with some spacing options, and works well if you have nicely proportioned plots and labels, preventing things from overlapping. It is a bit brittle though, and can break some things, as we have witnessed. The `padding` option will be familiar to anyone who has used `plt.subplots_adjust()`, and basically just provides a way to access those controls within `plot_images`. I've updated the documentation to make this more clear. Is that sufficient, or do you think another implementation would be better? Since this functionality already exists within pyplot, I figured it better to use that then to re-invent the wheel, so to speak.
  * *Note, these are the same options that are provided when you click this button in the GUI:*
plot_image-subplots
 * I agree that the current implementation of the aspect ratio control is not sufficient, and I'm working now on improving it to be more like `im.plot()`

**Comment by jat255** [comment]
After playing around with this some more, I believe I've reached a better implementation of the aspect ratio handling. There are now 4 possible ways to use the aspect parameter: `'auto', 'square', 'equal'`, or `<fixed value>`. 
 * `'auto'` works as expected, adjusting the aspect ratio according to `min_asp` as necessary (like how it works in `image.plot()`)
 * `'square'` is how the images were plotted as of the last commit. *i.e.* the display is forced to be square, and the aspect ratio is adjusted accordingly 
 * `'equal'` is just a shortcut for enforcing an aspect ratio of 1, regardless of image size
 * `<fixed values>` takes a number as input, and sets the aspect ratio to be fixed as that value

Here are some examples that were generated using the following code. Please let me know if you have any suggested improvements. I think this function is nearing completion. If you all agree, I will begin work on documenting it in the user guide, as I wanted to wait until it was feature-complete.

```python
for opt in ['auto','equal','square',2]:
    for zx in [2, 5, 15, 30, 100, 200, 300]:
        im2 = signals.Image(np.random.random([2,15,zx]))
        utils.plot.plot_images(im2.deepcopy(), per_row=5, tight_layout=False, axes_decor='ticks',
                           colorbar='single', label=None,cmap='RdYlBu_r', min_asp=0.3, aspect=opt,interp='nearest')
        fig = gcf()
        fig.canvas.set_window_title("aspect: " + repr(opt) + "; " + repr(zx) + " x 15")
```
* `'auto'`:
  * plot_images_aspect-1-auto
* `'equal'`: 
  * plot_images_aspect-2-equalpng
* `'square'`: 
 * plot_images_aspect-3-square
* `<fixed value = 2>`: 
 * plot_images_aspect-4-fixed2

vidartf:
Regarding the ```im.plot(); plot_images``` replotting problem, this is related to PR #413. The solution I proposed there uses an AxesManager copy to iterate, which will then not be connected to any of the navigation callbacks of the original. If you don't want to use the code in the PR (and I don't think you should for now), I agree that you should use the ```ims._deepcopy_with_new_data(ims.data)``` trick.

pburdet:
- It looks like a nice solution @vidartf. I'm not for deepcopy for an optimize memory control. Let's keep it as it is waiting for vidartf's solution.
- Concerning the aspect ratio, I like the new solution
- The `$$` solution is perfect (I think axes_manager is using the same)
- For `tight_layout` and `padding`, let's keep it as it is. It gives lot of flexibility to the user. I didn't know the gui solution, I discover something great, Thanks.
- For a title containing the common information, I would put all the labels in a list and make a test if the first part of each label is the same.

jat255:
Sorry for the reverted commits, but it seems like the most recent implementation works in all of my testing for preventing the `plot_images()` code from iterating any prior `image.plot()` instances.

jat255:
Ok, after trying out a few different methods, I think I have a good working method of shortening redundant labels. It is activated by using the `labels='auto'` parameter (which I have made the default).

The gist of the method is this:
 1. Get the titles of each image to be plotted
 2. Run through title list, character by character, determining if beginning of title matches the other titles
 3. Store the shortest common string between all the titles
   * if the shortest common string is less than 1/2 of the total length, assume it was just a coincidence and fall back to using each image's title as its label
  * if it is > 1/2 the length, assume that all the labels are part of a series, and save that string as a 'basename'
 4. If there is a match, set the base name to be the figure's suptitle, and then use the remaining substring in the title as each image's label

This appears to work pretty well. It is probably easiest to show a few examples, as I've done below. I'm sure there will be cases where it will not work as expected, but it seems to be doing alright in my tests. Please let me know if you find some cases where the result is not as expected.

 * Testing a single multi-dimensional hyperimage:

  ```python
  im1_3 = signals.Image(np.random.random((2, 3,64,64)))
  im1_3.metadata.General.title = 'multi-dimensional image'
  utils.plot.plot_images(im1_3, axes_decor='ticks', tight_layout=False,
                         label='auto',colorbar='single',labelwrap=30,per_row=3)
  ```
 plot_image-shared_titles-im1_3

 * Testing the results of `Signal.get_bss_loadings()`:

 ```python
im = si_EDS.get_bss_loadings()
utils.plot.plot_images(im, per_row=5, tight_layout=True, axes_decor='off',
                           colorbar='single', label='auto',cmap='RdYlBu_r')
```
  plot_image-shared_titles-auto with bss loadings

 * Testing the results of `Signal.get_lines_intensity()` (note, this does not activate the `auto` code, because there is not enough overlap between the titles):

 ```python
im = si_EDS.get_lines_intensity()
utils.plot.plot_images(im, per_row=5, tight_layout=False, axes_decor='off',
                           colorbar='single', suptitle='EDS Line intensity', label='auto',cmap='RdYlBu_r')
```
 plot_image-shared_titles-auto with no overlap

 * With the same data, combining `suptitle` and a custom `label` list allows for a pretty good display:

 ```python
im = si_EDS.get_lines_intensity()
utils.plot.plot_images(im, per_row=5, tight_layout=False, axes_decor='off', 
                           suptitle_fontsize=20, colorbar='single', suptitle='EDS Line intensity', 
                           label=['La M$\\alpha$', 'Mn K$\\alpha$', 'Zr L$\\alpha$'],cmap='RdYlBu_r')
```
 plot_image-shared_titles-manual

**Comment by francisco-dlp** [comment]
Awesome!

**Comment by pburdet** [comment]
This is great! Nice examples.

We might decide to give a certain structure to automatically generated title. From the more general information to the more specific. Tilte from `Signal.get_lines_intensity()` will become: "X-ray line intensity of EDX spectrum: Al Ka at 1.7 keV". And it will look nice. 

Small bug 

  ```python
utils.plot.plot_images(im, per_row=5,
                       label=[str(i) + '$^\degree$'  for i in im.axes_manager[0].axis])

    712 
    713             if label:
--> 714                 if all_match:
    715                     title = ''
    716                 else:

UnboundLocalError: local variable 'all_match' referenced before assignment
  ```

**Comment by jat255** [comment]
@pburdet, thanks for the bug report. It should be fixed now.

As for your suggestion, that is exactly what I was thinking. If throughout hyperspy the labels are structured to be general first and specific later, this code will work particularly well. I'll open up an issue for that one as well. 

Do you mind posting a picture of your completed tilt series? It's a nice figure to use for the documentation, I think.

**Comment by jat255** [comment]
Ok, I have added some documentation for this function as well in the [user guide](https://github.com/jat255/hyperspy/blob/ENH_plot_images_function/doc/user_guide/visualisation.rst#plotting-several-images).

I think I've covered the basic use cases without being overly verbose. 

I'm wondering: is there a way to build the whole api locally? I've been running `make html` from the `doc` directory, which generates the user guide (thankfully), but does not appear to generate the api for the drawing functions, so I cannot check to make sure that the docstring looks okay in the html documentation. Is there another command I should be running?

I think this function is getting pretty close to ready for merging. Finally! :)

**Comment by francisco-dlp** [comment]
`./generate_api_doc.sh` should do the trick.

I think that It'll be a good idea to add the examples that you have used to demo `plot_images` to an IPython notebook and place it in `hyperspy/tests/drawing`.

**Comment by jat255** [comment]
@francisco-dlp, great, thanks for the tip!

I cleaned up the documentation some and added a testing notebook as well. Should it be linked in the user guide?

**Comment by pburdet** [comment]
The tilt series is from a paper about to be submitted. I'll ask around if it's fine to use it as an example.

For an example of X-ray lines intensity, I have put some EDX data online if you want. Here is the notebook (Hyperspy training) to download them and to configure them. (in hyperspy.org in demos)

http://nbviewer.ipython.org/github/hyperspy/hyperspy-demos/blob/master/electron_microscopy/EDS/Hyperpsy_EDS_SEM_tutorial_IMC_2014.ipynb

**Comment by pburdet** [comment]
Almost finish! Before merging it would be great if you can modify the code to respect PEP8 standard (especially lines with a length shorter than 79 character) and to have the dosctring respecting the numpy style.

https://www.python.org/dev/peps/pep-0008/

https://github.com/numpy/numpy/blob/master/doc/HOWTO_DOCUMENT.rst.txt

**Comment by francisco-dlp** [comment]
@jat255, we don't link to notebooks in the User Guide. However, you could submit the notebook to https://github.com/hyperspy/hyperspy-demos .

File: hyperspy/drawing/utils.py [hunk 2]
```diff
@@ -343,6 +343,466 @@ def _set_spectrum_xlabel(spectrum, ax):
...
+
+    """
+    from mpl_toolkits.axes_grid1 import make_axes_locatable
+    from hyperspy.drawing import widgets
+    from hyperspy.misc import rgb_tools
+    from hyperspy.signal import Signal
+    from traits import trait_base
+    from itertools import compress
+    import textwrap
+
```
pburdet: Isn't it better to have the import with all the others import at the beginning of the file?

francisco-dlp:
@jat255, to make it easier to comply with PEP8 you could use autopep8. Actually, you could run it as a post-commit hook as explained in #181.

hyperspy/drawing/utils.py
```diff
@@ -343,6 +343,466 @@ def _set_spectrum_xlabel(spectrum, ax):
...
+    if isinstance(images, Signal) and len(images) is 1:
+        print "Single image provided, using Signal.plot() instead."
+        print "Use Signal.plot() directly to supply more options."
+        images.plot()
```
pburdet: Good idea! But then the keyword arguments are not working. `plot()` should accept the important keyword of this function (see #160). So leave it as it is and remove the print.

hyperspy/drawing/utils.py
```diff
@@ -343,6 +343,466 @@ def _set_spectrum_xlabel(spectrum, ax):
...
+
+    # Get default colormap from pyplot:
+    if cmap is None:
+        cmap = plt.get_cmap()
+
+    # If input is >= 1D signal (e.g. for multi-dimensional plotting),
+    # copy it and put it in a list so labeling works out as (x,y) when plotting
+    if isinstance(images,Signal) and images.axes_manager.navigation_dimension > 0:
+        images = [images._deepcopy_with_new_data(images.data)]
+
```
pburdet: I think we shouldn't deepcopy the signal to avoid memory problem

hyperspy/drawing/utils.py
francisco-dlp: In this case, we should implement #160 and merge it here before merging `plot_images` into master.

hyperspy/drawing/utils.py
pburdet: It doesn't work for 3D RGB images. If too complex to implement, raise an error

hyperspy/drawing/utils.py
```diff
@@ -343,6 +343,466 @@ def _set_spectrum_xlabel(spectrum, ax):
...
+    # and tag them using the isrgb list
+    for i, img in enumerate(images):
+        if rgb_tools.is_rgbx(img.data):
+            isrgb[i] = True
+
+    # Find global min and max values of all the non-rgb images for use with 'single' scalebar
+    if colorbar is 'single':
+        gl_max, gl_min = max([i.data.max() for i in list(compress(images, [not j for j in isrgb]))]), \
+                         min([i.data.min() for i in list(compress(images, [not j for j in isrgb]))])
+
```
pburdet: Are you aware of the way `plot()` set the vmin and vmax with `perc` arguments? It would be nice to have a similar method.

hyperspy/drawing/utils.py
```diff
@@ -343,6 +343,466 @@ def _set_spectrum_xlabel(spectrum, ax):
...
+    idx = 0
+    # Loop through each image, adding subplot for each one
+    for i, ims in enumerate(images):
+        # Get handles for the signal axes and axes_manager
+        axes_manager = ims.axes_manager
+        axes = axes_manager.signal_axes
+        if axes_manager.navigation_dimension > 0:
+            ims = ims._deepcopy_with_new_data(ims.data)
+        for j, im in enumerate(ims):
```
pburdet: Same comment about deepcopy

hyperspy/drawing/utils.py
```diff
@@ -343,6 +343,466 @@ def _set_spectrum_xlabel(spectrum, ax):
...
+            axes = axes_manager.signal_axes
+
+            if axes_manager.signal_dimension == 2:
+                # get calibration from a passed axes_manager
+                shape = axes_manager._signal_shape_in_array
+
+                # Reshape the data for input into imshow (if not rgb)
+                if not isrgb[i]:
+                    data = data.reshape(shape)
+
```
pburdet: Do you need to reshape?

hyperspy/drawing/utils.py
```diff
@@ -343,6 +343,466 @@ def _set_spectrum_xlabel(spectrum, ax):
...
+            # Label the axes
+            if isinstance(axes[0].units, trait_base._Undefined):
+                axes[0].units = 'pixels'
+            if isinstance(axes[1].units, trait_base._Undefined):
+                axes[1].units = 'pixels'
+            if isinstance(axes[0].name, trait_base._Undefined):
+                axes[0].name = 'x'
+            if isinstance(axes[1].name, trait_base._Undefined):
+                axes[1].name = 'y'
```
pburdet: Shouldn't you let the axis undefined if they are undefined?

hyperspy/drawing/utils.py
francisco-dlp: This brings to my mind that, now that we'll start depending on scikit-image, it'll be nice to use [their histogram equalization tools](http://scikit-image.org/docs/dev/auto_examples/plot_equalize.html). I've just opened an issue (#486) about it.

hyperspy/drawing/utils.py
```diff
@@ -343,6 +343,466 @@ def _set_spectrum_xlabel(spectrum, ax):
...
+    elif scalebar is 'all':
+        # scalebars were taken care of in the plotting loop
+        pass
+    elif scalelist:
+        # scalebars were taken care of in the plotting loop
+        pass
+    else:
+        print 'Did not understand scalebar input. Please use None, \'all\', or list of ints.'
+        print 'No scalebars will be displayed.'
+
```
pburdet: Raise a warning here

pburdet:
I have found a bug with the following image:

```python

>>> im 
<Image, title: SE images, dimensions: (11|896, 668)>
>>> utils.plot.plot_images(im)

```

It doesn't plot anything if tight_layout is True

pburdet:
Nice documentation. Random images are not great. You might use the images from

http://scikit-image.org/docs/dev/api/skimage.data.html

as done in the hyperspy documentation for `map` in signals tool

hyperspy/drawing/utils.py
```diff
@@ -18,6 +18,9 @@
 
 import copy
 import itertools
+import textwrap
+from traits import trait_base
+from mpl_toolkits.axes_grid1 import make_axes_locatable
```
jat255: Moving the hyperspy imports to here caused errors when importing `hspy`, so I've moved what I could up here, and left the others in the function.

hyperspy/drawing/utils.py
jat255: @pburdet, do you have a 3D RGB image that I can test with (or how to generate one)? I'd like to get it to work, if not too tricky.

hyperspy/drawing/utils.py
jat255: Hmm, I suppose you're right. It works fine without it. I'm not sure how/why that was in there, actually... I may have just copied it from the `signal._plot_loadings()` function at the beginning.

jat255:
> I have found a bug with the following image:
> 
> >>> im 
> <Image, title: SE images, dimensions: (11|896, 668)>
> >>> utils.plot.plot_images(im)
> It doesn't plot anything if tight_layout is True

I suspect this is due to `tight_layout()`'s annoying habit of breaking anything it touches... Would it be better to disable it on default? It does help things look better, but I guess it's better to return something (even if ugly) to the user than nothing at all with no error message, no?

pburdet:
Here is a 3D RGB images.

```python

n = 5
a = range(0,255,n)
b = range(0,255,n)
c = range(0,255,n)
d = []
for i in a:
    d.append([])
    for j in b:
        d[-1].append([])
        for k in c:
            d[-1][-1].append([i,j,k])
im = signals.Image(d)
im = im.as_spectrum(2)
im.change_dtype('uint8')
im.change_dtype('rgb8')
for i in range(3):
    ax = im.axes_manager[i]
    ax.name = ['r','b','g'][i]
    ax.scale = 1./ax.size

```

jat255:
> to have the dosctring respecting the numpy style.

@pburdet, I believe after the latest commit that the docstring matches the numpy styling (at least as I read the guide). Please let me know if there's anything else that needs editing on this point.

jat255:
> Here is a 3D RGB images.

Thanks. Upon first glance, nothing about the actual plotting code seems to be preventing this code from working. It does appear to fail due to the auto-labeling and the `tight_layout`.

```python
utils.plot.plot_images(im[:-1], label=None, per_row=5, tight_layout=False, axes_decor=None)
```
plot_images_rgb

pburdet:
Good improvement in term of style. I'm using spyder and it's a big help for pep8. There is few lines (443 and 464) too long in the doctring. Line 697 you define shape but you never use it. Same for cbar line 820. You should remove the spacing inbetween each keyword in the docstring

pburdet:
Great for the colors imaging! tight_layout is causing lot of problems it seems. I'm for disabling it by default, as it is not obvious that it is causing bugs.

pburdet:
You can use the tilt series in the doc if you want:)

francisco-dlp:
It might be worth reporting the `tight_layout` issues to the matplotlib team.

jat255:
> Good improvement in term of style. I'm using spyder and it's a big help for pep8. There is few lines (443 and 464) too long in the doctring. Line 697 you define shape but you never use it. Same for cbar line 820. You should remove the spacing inbetween each keyword in the docstring

Thanks. I've been using [PyCharm](https://www.jetbrains.com/pycharm/), which has some good inspection features, but I didn't have it set to the correct settings (it was ignoring unused variables b/c of I disabled it in another project). I think I've got it configured correctly now.

Also, I corrected some typos once I had the inspections running :smile:

jat255:
Ok, here's a working example for the 3D RGB images:

 ```python
utils.plot.plot_images(im[:-1], per_row=10, axes_decor=None,tight_layout=True)
```

 
plot_images_rgb2

File: hyperspy/drawing/utils.py (hunk 8 above)
Comment (jat255) [inline hyperspy/drawing/utils.py]: > Shouldn't you let the axis undefined if they are undefined?

If this is the case, should we just hide the axes, then? It is really ugly (in my opinion) to have big "<undefined>" labels around the plot

**Comment by jat255** [comment]
As for the `deepcopy()` question, I think we have to choose a lesser of two evils, no? (Until the new slicing method is implemented) Either an existing `image.plot()` will update when calling `plot_images()`, or we will have to make a copy of the data.

After quickly testing, with the `deepcopy()` method, it does appear that the data remains in memory after the figure is closed, persisting until the ipython kernel is restarted. Is there any way to force python to clear that data when the figure is closed? That could be a good solution.

**Comment by francisco-dlp** [comment]
This is an ugly (but hopefully less evil than the alternatives) workaround:
```
amc = im.axes_manager.deepcopy()
ambk = im.axes_manager
im.axes_manager = amc
for image in im:
    # do something
image.axes_manager = ambk
del amc
```
If we decide to use it, we should leave a note in the code indicating that this should be replaced with the proper alternative once it is available.

File: hyperspy/drawing/utils.py (hunk 8 above)
Comment (pburdet) [inline hyperspy/drawing/utils.py]: Ok, let's hide them

**Comment by pburdet** [comment]
Should we use a lower value for `namefrac`? Here is an example where 0.5 doesn't give a good result

figure_1

File: hyperspy/drawing/utils.py [hunk 11]
```diff
@@ -343,6 +346,493 @@ def _set_spectrum_xlabel(spectrum, ax):
...
+                no_nans=False,
+                per_row=3,
+                label='auto',
+                labelwrap=30,
+                suptitle=None,
+                suptitle_fontsize=18,
+                colorbar='multi',
+                scalebar=None,
+                scalebar_color='white',
+                interp='nearest',
```
francisco-dlp: Is the `interp` argument necessary given that **kwargs are passed to `imshow`?

francisco-dlp:
#489 should fix #160 that is currently holding this PR.

hyperspy/drawing/utils.py
jat255: Yes, you're correct that `interpolation` will work as a keyword. If this is done however, then the interpolation will default to `matplotlib`'s `rcParam` value. If this is okay, then I have no problem having interpolation as a keyword. Perhaps we could temporarily override the default to `'nearest'`, as shown [here](https://github.com/matplotlib/matplotlib/blob/master/doc/users/whats_new.rst#new-rc-parameter-functionality)?

jat255:
> Should we use a lower value for namefrac? Here is an example where 0.5 doesn't give a good result

Probably. 0.5 was a guess based off of what I was working with at the time, but I'm sure different situations will call for different thresholds. What about if we make it adjustable and use 0.4 as a default?

jat255:
Here's an example of the new `namefrac_thresh` parameter:

```python
si_EDS = load("plot_images_eds.hdf5")
im = si_EDS.get_lines_intensity()
for i in [0.6,0.8]:
    utils.plot.plot_images(im, per_row=2, tight_layout=True, axes_decor='off', 
                           suptitle_fontsize=20, colorbar='single', 
                           scalebar='all', scalebar_color='white',namefrac_thresh=i,
                           padding={'top':0.85,'bottom':0.10,'left':0.05,'right':0.85,'wspace':0.10,'hspace':0.20})
    fig = gcf()
    fig.canvas.set_window_title('threshold = ' + repr(i))
```

plot_image-namefrac_thresh

jat255:
The `perc` option for image contrast is now implemented. It works (in my testing) when using both single and multi colorbars. I basically copied the `optimize_contrast` method from [`drawing/image.py`](https://github.com/hyperspy/hyperspy/blob/master/hyperspy/drawing/image.py#L131-L142) as a private function.

Here's it is action:

**Multi colorbars:**
```python
im1_3 = signals.Image(np.random.random((2, 3, 64, 64)))
im1_3.metadata.General.title = 'multi-dimensional image'
plt.set_cmap('RdYlBu')

for i in [0, 0.25, 0.5]:
    utils.plot.plot_images(im1_3,namefrac_thresh=0, perc=i,per_row=2)
    plt.tight_layout(rect=[0, 0, 1, 0.9])
    gcf().canvas.set_window_title('perc = ' + repr(i))
```

plot_image-perc_multi

**Single colorbars:**

```python
im1_3 = signals.Image(np.random.random((2, 3, 64, 64)))
im1_3.metadata.General.title = 'multi-dimensional image'
plt.set_cmap('RdYlBu')

for i in [0, 0.25, 0.49]:
    utils.plot.plot_images(im1_3,namefrac_thresh=0, perc=i,per_row=2,colorbar='single')
    plt.tight_layout(rect=[0, 0, 0.9, 0.9])
    gcf().canvas.set_window_title('perc = ' + repr(i))
```

plot_image-perc_single

jat255:
Undefined axes traits now will cause the image axes to not be labeled, rather than changing to a default x/y and pixels, per the discussion on [03/12/15](https://github.com/hyperspy/hyperspy/pull/454#discussion-diff-26289315). See below for example in action.

**Undefined axis name:**
```python
im1_3 = signals.Image(np.random.random((2, 3, 64, 64)))
im1_3.metadata.General.title = 'multi-dimensional image'
im1_3.axes_manager.signal_axes[0].units = 'nm'
im1_3.axes_manager.signal_axes[1].units = 'nm'
im1_3.axes_manager.signal_axes[0].name = 'x'
# im1_3.axes_manager.signal_axes[1].name = 'y'
utils.plot.plot_images(im1_3,namefrac_thresh=0, perc=0, axes_decor='all',tight_layout=True)
gcf().canvas.set_window_title('undefined y-axis name')
```
plot_image-axes-undefined


**All axes traits defined:**
```python
im1_3 = signals.Image(np.random.random((2, 3, 64, 64)))
im1_3.metadata.General.title = 'multi-dimensional image'
im1_3.axes_manager.signal_axes[0].units = 'nm'
im1_3.axes_manager.signal_axes[1].units = 'nm'
im1_3.axes_manager.signal_axes[0].name = 'x'
im1_3.axes_manager.signal_axes[1].name = 'y'
utils.plot.plot_images(im1_3,namefrac_thresh=0, perc=0, axes_decor='all',tight_layout=True)
gcf().canvas.set_window_title('all axes defined')
```
plot_image-axes-defined

jat255:
I've tried to play around some with the memory issue by explicitly deleting the things that are `deepcopy`ed (like @francisco-dlp [suggested](https://github.com/hyperspy/hyperspy/pull/454#issuecomment-78538327)), but I cannot get it to work. Even if I `del` everything that is copied (`ims` and `images`), the memory continues to grow upon subsequent calls to `plot_images`. I've also tried using the `gc` garbage collector interface by using `gc.collect()` after deleting the , but this does not appear to work either...

I'm not sure at this point what is actually being copied vs. not, and why things won't clear from memory. I guess somewhere (even after deleting and closing figures) there are some pointers that are left over to the copies, which is why Python won't delete them fully?

pburdet:
Thanks a lot.

As a big data user, I'm for trying to avoid as much as possible any deepcopy. I don't thing we need it, do we?
 I don't think we should store the test folder of hyperspy. Have a look in #497, I solved this problem by letting the user download the data from a server (as in the EDS tutorial). You can do the same in the documentation.

vidartf:
@jat255 I've had some similar experiences while developing the UI. Does it leak for other MPL backends than Qt? The MPL Qt backend is a bit lazy in cleaning up its references, and as Qt is a C++ library, that can throw the GC off a bit.

pburdet:
Thanks a lot @jat255. Let's merge it as it is.

PR #459, opened 2015-01-30
ENH: PCA with nav dim = 1

vidartf:
I can see no reason why PCA decomposition should not be possible for navigation dimensions equal to 1. The navigation dimensions are unfolded to one dimension either way. Maybe the check should be that the total number of dimensions is greater than one?


francisco-dlp:
Actually there is a typo in the error message: the problem is not that navigation_dimension <=1—that is fine—what is required is that navigation_size > 1 because it is not possible to decompose one single signal (i.e. one single spectrum, image etc). The code checks the correct attribute (`navigation_dimension`) but the error message is wrong.

vidartf:
Ok, I'll pull another one with correction to message.

PR #463, opened 2015-02-06
New extend dimension

francisco-dlp:
New `Signal.extend_dimension` method.

Example
-----------

```python

>>> s = signals.Signal(np.ones((2,2)))
>>> s.data
array([[ 1., 1.],
[ 1., 1.]])
>>> s.extend_dimension(increment=-2, axis=1, fill_value=10).data
array([[ 10., 10., 1., 1.],
[ 10., 10., 1., 1.]])
>>> s.extend_dimension(increment=-2, axis=0, fill_value=10).data
array([[ 10., 10.],
[ 10., 10.],
[ 1., 1.],
[ 1., 1.]])

```

TO DO
--------

- [ ] Re-implement using np.pad as suggested by @vidartf


vidartf:
Seems useful, but have you consider basing this around numpy.pad? It has quite a useful set of features w.r.t. fill values.

francisco-dlp:
You're right, that would be a better approach, thanks! (I should have made a pull request to discuss the feature and the implementation before writing the tests and the docs...)

francisco-dlp:

PR #474, opened 2015-03-02
New simulate simple eds model

pburdet:
Simulate a simple model of the X-ray lines (no background) peak, based on one Gaussian per peak. This function is intended mainly for test purpose. It might be the base of a more complex simulation adding background, taking into account the cross-section of the different lines and the absorption in the sample and in the detector.

Example
-----------

```python

    >>> s = utils.eds.xray_lines_model(['Cu', 'Fe'], beam_energy=30)
    >>> s.plot(True)

```


dnjohnstone:
@k8macarthur I'm currently having an EDS tidy up to help make things go more smoothly when you start to contribute your code... (and some selfish reasons!). What do you think of this potential contribution - it doesn't have proper cross sections in it at the moment, but we could add that and maybe this could be somewhere to start with your things?

dnjohnstone:
Does anyone have an opinion on the value of this? Currently it doesn't do a great deal, just finds the X-ray line positions from the database and puts a Gaussian at each to make something that looks like an EDX spectrum.

@pburdet seemed to have intentions to use this as a platform for much more realistic simulation, which is in principle good, but I'm not sure who's going to take that forward at the moment.

Do we want to polish this up in its current form and merge it to leave the platform there (which is relatively easy) or do we close this and wait for someone, or just leave it?

k8macarthur:
I can see that simulating spectra could be something useful. However, without information on relative heights for particular composition etc. I’m not sure how helpful it is at this stage. I certainly don’t think I have the expertise to take it further...

francisco-dlp:
If I remember well @pburdet intention was implementing a simple way to generate synthetic spectra resembling EDX spectra to test data analysis methods on them. As long a it is made crystal clear in the documentation (and possibly in the function name) that this doesn't generate proper EDX spectra, then I guess that it's fine to add this feature. What do others think?

dnjohnstone:
I think I agree that the functionality to just make something that looks a bit like a spectrum to play with developmental tools is a potentially useful thing to be able to do... I wonder if we should put it in EDS though or rather put it in general utils with a bit of rewording? In principle, I think you could make sensible looking things this way for many spectra...

k8macarthur:
I agree it’s helpful and should be included. Maybe avoid use synthetic or model rather than simulated spectrum as the title to try and avoid confusion?

francisco-dlp:
Is anybody taking care of this?

francisco-dlp:

ericpre:
There is now artificial or example datasets (for EDS #497) available in hyperspy.

PR #477, opened 2015-03-03
Fetch spectra from the new EELS Database website

francisco-dlp:
Example
-----------

```python
hs.plot.plot_spectra(hs.datasets.eelsdb(
        element=["Fe", "O"],
        spectrum_type="coreloss"))

```
index

TODO
--------

* [x] Document in User Guide.
* [x] Store json in original_metadata
* [x] raise warning for invalid files
* [x] Parse the json file to gather more info about the spectra.
* [x] Suggest to upload spectrum if query returns no spectra.
* [x] Unittests (requires testing spectra on the EELSdb).??
* [x] Fix monochromated
* [x] Catch errors
* [x] Change to hs.datasets.eelsdb


ewels:
I'm really not sure you guys want to merge this into the core code base yet - the EELS db website is still in beta and the API isn't mature or stable yet, so the assumptions made in `eelsdb.py` may not be true in a month's time..

francisco-dlp:
This branch is now ready for review.

doc/user_guide/getting_started.rst
```diff
@@ -295,6 +295,32 @@ to the constructor e.g.
...
+.. code-block:: python
+
+    >>> hs.datasets.example_signals.EDS_TEM_Spectrum().plot()
+
+.. versionadded:: 0.9
+
+    The :py:func:`~.misc.eels.eelsdb.eelsdb` function in `hs.datasets` can
+    directly load spectra from `The EELS Database <http://eelsdb.eu>`_. For
+    example, the following loads all the boron trioxide spectra currently
+    available in the database:
```
to266: Github does not render this paragraph (and the `versionadded`). Is it just github? I guess generally the paragraph should not be indented, as the `versionadded` should only be used for short descriptions, with more info most likely following in a normal text.

hyperspy/misc/eels/eelsdb.py
```diff
@@ -0,0 +1,273 @@
...
+                    try:
+                        s.add_elements(json_md.elements)
+                    except ValueError:
+                        messages.warning(
+                            "The following spectrum contains invalid chemical "
+                            "element information: "
+                            "Title: %s id: %s. Elements: %s. "
+                            "Please report this error to http://eelsdb.eu/about" %
+                            (json_md.title, json_md.id, json_md.elements))
+                alpha, beta, beamenergy = None, None, None
```
to266: Line not required

hyperspy/tests/datasets/test_eelsdb.py
```diff
@@ -0,0 +1,51 @@
...
+        order_direction='DESC',
+        monochromated=False, )
+    nt.assert_equal(len(ss), 2)
+    md = ss[0].metadata
+    nt.assert_equal(md.General.author, "Odile Stephan")
+    nt.assert_equal(
+        md.Acquisition_instrument.TEM.Detector.EELS.collection_angle, 24)
+    nt.assert_equal(md.Acquisition_instrument.TEM.convergence_angle, 15)
+    nt.assert_equal(md.Acquisition_instrument.TEM.beam_energy, 100)
+    nt.assert_equal(md.Acquisition_instrument.TEM.beam_energy, 100)
```
to266: line repeated

to266:
Also, I personally would prefer if the warnings were not just one-liners. It would make them much more readable, given they also give the user suggestions in this case, however it's up to you

francisco-dlp:
All should be fixed now, thanks!

to266:
Just fixing https://github.com/hyperspy/hyperspy/pull/747 (as it is merged here as well)

to266:
While using the example you have in the title, it looks like if any json has a list as it's value, the dictionaries are not formatted nicely. While that's always been the case with DTB, I think in this case it might be nice to have it format properly, and assign either `title` or `name` to the entry. Currently it is like this

```python
In [13]: s.original_metadata.json.associated_spectra
Out[13]: 
[{u'link': u'https://eelsdb.eu/spectra/iron-oxide/',
  u'name': u'Iron Oxide',
  u'type': u'Low Loss'}]
In [14]: s.original_metadata.json.other_links
Out[14]: 
[{u'title': u'Old EELS DB',
  u'url': u'http://pc-web.cemes.fr/eelsdb/index.php?page=displayspec.php&id=141'}]

```

What do you think?

francisco-dlp:
I agree that the formatting is not ideal in this case. However, trying to fix it might be dangerous. e.g. if there were more than one associated spectra it's likely that they'll all have the same `name`. Personally I would rather leave it as it is.

to266:
After further though, I agree. I think it would be much better to add nice printing of lists in DTB instead... I think I will attempt it soon.

Also, the `monochromated` keyword seems not to do anything, as both lines return identical results
```python
hs.datasets.eelsdb(element="C", monochromated=True)
hs.datasets.eelsdb(element="C", monochromated=False)
```

Is there a place (in metadata?) where we can denote information like that? I think if you can search with such a keyword, that should be saved somewhere in the results

francisco-dlp:
Yes, that's probably a bug in the EELSdb API, see the lastest @ewels comment in #475.

We don't currently have metadata for "monochromated". In principle HyperSpy's metadata only concerns information that is relevant for analysis and I'm not sure if "monochromated" qualifies as it is the outcome (the actual energy resolution) what matters.

to266:
https://github.com/hyperspy/hyperspy/pull/751 for the printing

francisco-dlp:
@to266, I've fixed the issue with the `monochromated` key. Could you consider merging this?

PR #478, opened 2015-03-03
NEW: tomographic reconstruction

pburdet:
A connection to skimage and their tomographic reconstruction.

```python

        >>> print tilt_series
        >>> rec = tilt_series.tomographic_reconstruction(
        >>>     'SART', iteration=5, relaxation=0.15)
        >>> print rec
        <Image, title: Tomogram, dimensions: (14|80, 92)>
        <Image, title: Reconstruction from Tomogram, dimensions: (80|80, 92)>

```

TO DO
--------

- [ ] test
- [ ] doc


vidartf:
Should probably make a note in the doc that extra keywords (`kwarg`) is passed to `skimage.transform.iradon` or `skimage.transform.iradon_sart`, so that people know they can use it, and know where to find the docs for that (already mentioned in notes).

pburdet:
Thanks, done.

dnjohnstone:
Is this something that we still want to include - given the existence of TomoPy/ASTRA/This in skimage? If so - it looks like it's just waiting for tests and docs? 

I need to do some tomography development so finishing this off might be useful for me anyway.

francisco-dlp:
I think that it makes sense to add wrappers for external tomographic reconstruction algorithms because HyperSpy can take care of keeping the calibration and of parallelizing the reconstructions if needed. However, probably we need a more general approach. What about:

```python
reconstruction = tilt_series.tomographic_reconstruction(reconstruction_algorithm, kw1, kw2, kw3)

```

`tomographic_reconstruction` could call `Signal.map` if parallelization is needed and do the right thing with the output. Otherwise it could simply call the function, getting the angles from the first axis and then return an `Image` with the right calibration.

jat255:
I'd add that I saw a talk about [tomviz](https://github.com/openchemistry/tomviz) at M&M a few months back, and they include a good deal of tomographic stuff. I haven't taken too deep a look at it, but perhaps some of what they do could be useful in this regard.

jeinsle:
I was less than happy with the tomoviz environment when I downloaded it. Especially as it is more of a wrapper around paraview, but with a couple of tomo tools added.

I was at the talk, but after downloading I made above conclusion.

J

dnjohnstone:
Just to bump this again - I have some time to tie this off, are we happy with the following approach:

1. generalise the function to enable the use of external algorithms
2. tests
3. documentation

francisco-dlp:
+1

francisco-dlp:
maybe it could be split in two:`tomographic_reconstruction` and `tomographic_reconstruction_SART`?

francisco-dlp:
Also, is there any reason why this should belong solely to the `Image` class? One may like to reconstruction a line profile tilt-series...

dnjohnstone:
I don't think it needs to be Image class only.... will have a more detailed think though

dnjohnstone:
On a related note, although perhaps better discussed elsewhere are we worried about the ballooning of signal.py? It's getting to be very long.

francisco-dlp:
That's a good point. However, I wouldn't address it here indeed. Maybe open a new issue to see what people have to say about it?

francisco-dlp:

PR #479, opened 2015-03-04
Ehn map multiprocessing

pburdet:
Automatic parallelization of `map` with IPython clusters

Example
-----------

```python
    >>> parallel = 8
    >>> from skimage.restoration import denoise_tv_chambolle
    >>> self = signals.Image(np.random.random((50, 100, 100)))

    >>> %%time
    >>> self.map(denoise_tv_chambolle, eps=2e-5, n_iter_max=2000)
    Wall time: 25.8 s

    >>> %%time
    >>> self.map(denoise_tv_chambolle, parallel=8, eps=2e-5, n_iter_max=2000)
    Wall time: 6.61 s 

```


dnjohnstone:
@to266 Shall we try to get this incorporated with extending functionality of map in #737?

to266:
Oh, that's an old one! Unfortunately the whole framework changed slightly with the ipython splitting into a bunch of individual projects. I guess creating a new PR while looking at this one might be the best way to go forward. 

I'll have a look if there is anything else in ipyparallel worth incorporating....

francisco-dlp:
I've just have a quick look at this one and it seems that, unlike the standard map function, this one doubles the memory footprint. It should be possible to keep track of the indices of the signal and put the result back to where it belongs once it is ready instead, isn't it?

to266:
Well in python in general any kind of multiprocessing involves having multiple copies of the thing being processed - one for each processes. So even if we return the result immediately and store it where intended, we still hope that the `gc` does the job well enough. 
The other option is then to free the memory in the main process after sending the bits and pieces to other processes, and only assemble the whole thing back after computations are done.

francisco-dlp:
I'm assuming that we can trust the garbage collector.

Of course, what I wrote is only relevant for the simple case in which the data shape stays the same. Then, there is nothing to reassemble: the job is simply to store the result at the right place in `self.data` and delete it, isn't it?

vidartf:
> Well in python in general any kind of multiprocessing involves having multiple copies of the thing being processed

Only partly true. Shared memory can be used to help some of it, and data can be split into independent chunks: Say you want to process 200 images with 4 worker threads. Then the extra memory use should only be the equivalent of 4-16 images (depending on how you pipe the data between the processes) as each worker process the data serially.

francisco-dlp:
Does anybody have any thoughts on this? e.g. @to266?

to266:
I'm currently working on lazy signals, which slightly changes (i.e. generalises) the map function.

Separately, samfire uses a sort of parallel pool (which may be either multiprocessing or ipyparallel pool and by extension an EC2-backed pool). It can be adapted without much difficulty to be used in `map`, but from the few tests I performed, it was not worth it. I will have a look eventually, but it's low priority for me now.

francisco-dlp:

ericpre:
Done in #1295.

Issue #487, opened 2015-03-12
Routine for fitting EELSCLEdge L-edges


magnunor:
Hey,

as discussed in https://groups.google.com/forum/?utm_source=digest&utm_medium=email#!topic/hyperspy-users/7Z9DybWoU6s it would be nice to have a way to fit EELSCLEdge components to the continuum background in a robust and semi-autonomous manner. Since this would allow more quantitative analyses of the finestructure, for example: 
![til23_model](https://cloud.githubusercontent.com/assets/1690979/6615652/0cca0a92-c8a6-11e4-9bd0-9d043ef2c3bf.png)
  
I have implemented this in my own code, but haven't had time to properly integrate this into HyperSpy. It works in the following fashion:
- Find the lowest point right before the ionization edge
- Find the highest point in the fine structure (for example in the Manganese edge in LSMO this would be the L3 top point)
- some_good_fraction*(highest_value-lowest_value)+lowest_value = intensity_value, where some_good_fraction is in the 0.05 - 0.2 range
- Find the energy(intensity_value) which is between energy(lowest_value) and energy(highest_value)
- Energy(intensity_value) is then your onset energy

The find-onset-energy is implemented in this code: https://gist.github.com/magnunor/6b7d664faff1362e924d 

Combining this with a function which creates a model from a coreloss spectrum and some input-parameters, one can automate this: https://gist.github.com/magnunor/47c8a8b3b69f4370229e

I'm fairly busy right now, but should have time to properly implement this into HyperSpy in a couple of weeks. So probably too late for the 0.8 release.

jat255:
Hi Magnus,

Thanks for uploading these. I'm working on implementing them into my code (currently under `misc.eels.tools`) but having some trouble figuring out how to use `model_parameter_dict`. Is this a custom dictionary structure that you are supplying to the code for this purpose, or does it exist somewhere else that I cannot find?

jat255:
Once I have the setup working on my system within Hyperspy, I'll start a PR (unless you'd prefer to do it) so we can track how to implement these methods.

francisco-dlp:
I think that a `get_edge_onset` function for EELS would be really useful. The one proposed by @magnunor does a nice job with the spectrum that he uses to demo it, but I'm not sure if it is general enough to work for all edges (e.g. delayed maximum ones) and conditions (e.g. spectra with lower SNR).

[...]

PR #490, opened 2015-03-29
New: set EELSCLEdge onset

magnunor:
As discussed in https://github.com/hyperspy/hyperspy/issues/487 there should be some better way to set the onset of EELSCLEdges.

This implementation uses the routine described in https://github.com/hyperspy/hyperspy/issues/487 and seems to work nicely for the L-edges for the transition metals like Mn, Fe and Ti.

Routine for fitting a Manganese L23-edge, see the image below for the different energy ranges:

set_edge_onset

```python
s = load("some_spectrum.hdf5")
s.add_elements(['Mn',])
s.set_microscope_parameters(convergence_angle=24., collection_angle=26.)
ll = load("lowloss.hdf5")
m = create_model(s, ll=ll)
m.set_coreloss_edge_onset(Mn_L3) # (1) Just before the L3 onset and a little past the peak 
m.set_coreloss_edge_onset(Mn_L2) # (2) Just before the L2 onset and a little past the peak
m.set_signal_range() # (3) About 30 eV before the onset to just in front of the onset
m.add_signal_range() # (4) About 3 eV right after the L2, at the lowest point
m.multifit()
m.reset_signal_range
```

Result: set_edge_onset_result

I can probably automate this even more, is I've done here: https://gist.github.com/magnunor/47c8a8b3b69f4370229e . But I think having a separate function like the one in this PR would be useful.

When using only_current=False with the interactive mode, the process is very slow due to the plot windows updating for every change in position. Is there some way to suppress this?

Todo:
- [x] documentation
- [x] tests


vidartf:
> When using only_current=False with the interactive mode, the process is very slow due to the plot windows updating for every change in position. Is there some way to suppress this?

There is an ongoing PR/Issue for this, which will probably be sorted by events and events suppression. For now the most common workaround is simply to do the calculation on a deepcopy with the same data (`s._deepcopy_with_new_data(s.data)`). This deepcopy will then have a new `AxesManager` that is not connected to the plot. However, for your case, I guess that means that the results of the fit will sit on the wrong `Signal`. Maybe you can temporarily swap the `AxesManager` on your signal (with a copy), do the fit, then swap back?

magnunor:
I added some tests and documentation, but I'm not certain if the explanation in the docstring is really understandable. 

In any case the function itself seems to work nicely, except for the GUI iterating over every position when using the GUI version of the function. Has the event suppression mentioned by vidartf been merged?

francisco-dlp:
Very nice. It'll be good to test this against the spectra in the EELSdb.

2015-05-08 22:23 GMT+01:00 magnunor <notifications@github.com>:

magnunor:
I merged the current master into the branch, and everything seems to be working nicely.

Should be ready for review.

Issue #466, opened 2015-02-08
BUG: polynomial estimate_parameters not working


magnunor:
Using the current master the following does not work:

    s = signals.Spectrum(np.arange(1000).reshape(10,100))
    s.remove_background(signal_range=(0,100), background_type='Polynomial')

Fails with the error https://gist.github.com/magnunor/272a79a54e1735e65ac1

I tried to fix the bug in https://github.com/magnunor/hyperspy/commit/2a1c78e049124ec53c52f8222e48eab47b9b2d71 but it only removes the error for 1D datasets, and the fitting is very bad. And for 2D data it gets some other error.

In addition I started writing a test for `remove_background` with polynomial background estimator: https://github.com/magnunor/hyperspy/commit/6e459d87c002d834d884f87993a4b3fcc830b541 which should probably be expanded to include a lot more tests for all the possible background estimators and different shapes of spectrums.

PR #491, opened 2015-03-30
ENH: s.remove_background, flag to fit the background to the signal

magnunor:
Add possibility to remove the background using the standard model fitter, instead of the (possibly) less accurate component.estimate_parameters. 

Also a workaround for https://github.com/hyperspy/hyperspy/issues/466 , since the full fitting should be more robust compared to component.estimate_parameters.

- [x] Documentation
- [x] Test


francisco-dlp:
Nice!

magnunor:
Add test and some documentation, should be ready to merge if it looks ok

PR #489, opened 2015-03-28
Add kargs to Signal.plot to control image plotting options.

francisco-dlp:
Example of usage
-----------------------

```python

im = signals.Image(np.random.random((64, 64)))
im.data[:10, :10] = np.nan
im.plot(colorbar=False, scalebar=False, axes_ticks=False, no_nans=True)
```
figure

TODO
--------
* [ ] Document in User Guide

Fixes #160

PR #498, opened 2015-04-02
Francisco dlp clone enh plot arguments

to266:
Starting point (and replaces) #489 

@jat255 , I think I fixed all the bugs you mentioned, but let me know if anything else is missing.
@pburdet , could you review the code changes and merge once the documentation is added?

The most notable change from #489 and previous versions, is that the percentile is now calculated sanely - it now gives the part  of datapoints that will __not__ be saturated (i.e. percentile of 0.9 leaves 5% of points black and 5% white (i.e. saturated) ). It does not tak into account the values of the points, only the number of them.

I believe before a bug was left, and since there was no way to specify values of percentiles and the default was sane, noone noticed.


pburdet:
Nice! 

I hit a bug when I'm doing that:

```python
im = signals.Image(np.random.random((10,64, 64)))
im.plot(auto_contrast=False)
```

I'm expecting that `vmin` and `vmax`  are calculated on the current image and keept for the others images.

Also If `vmin` or `vmax` are provided, the `userWarning` is raised the first time but not afterwards. 

```python
im = signals.Image(np.random.random((10,64, 64)))
im.plot(vmin=0, vmax=1.)
```

Could `auto_contrast` be `False` if `vmin` or/and `vmax` are provided?

to266:
Done

vidartf:
> Also If vmin or vmax are provided, the userWarning is raised the first time but not afterwards.

That is standard behavior for python warnings as far as I can tell. Not sure exactly how to change this behavior.

to266:
if interested:
http://stackoverflow.com/a/22661838/3589924

anyway we decided to keep it as it is

pburdet:
That's true, thanks

vidartf:
Thanks for the link. I agree with keeping it as it is, as changing warning behavior should be done by the end user.

PR #500, opened 2015-04-05
BUG: EDSSpetrum.sum vs live_time

vidartf:
EDSSpectrum.sum() kept increasing the live_time metadata exponentially.


to266:
Please add a test for that, and should be good to go :)

vidartf:
Tests added, which also made me realize the fix introduced another bug :)

PR #501, opened 2015-04-05
ENH: Utility function to remove x-ray markers

vidartf:
For interactive use, it is a lot speedier to simply remove individual
lines compared to replotting, so a util function like this can be very
useful.


pburdet:
And it makes things more tidy. Thanks

PR #502, opened 2015-04-05
ENH: Consistent format for 'only_lines'

vidartf:
This was probably in the works somewhere else, but I needed it now :)


pburdet:
It looks good. Can you fast-forward it? `_parse_only_lines` can be used in `plot` also

vidartf:
Should be OK now?

pburdet:
Perfect

PR #503, opened 2015-04-06
Ehn plot images doc

pburdet:
Change the documentation to link it to the rest. Fix a bug (with rgb image). `Perc` to `percentile` as with `plot`. `interpolation` = "nearest" by default. Removing all data from the documentation and the test.


jat255:
These changes look good to me. Thanks for taking care of them; I was going to get to these last few today, but it looks like you've got it covered.

pburdet:
Sorry to have overtake your branch like that. We're running out of time for the release.

PR #504, opened 2015-04-06
Ehn whats new eds

pburdet:
What's new in 0.8 plotting and EDS


to266:
Thanks for these. The tag in changes should probably be 0.8 (not 0.8.0), but I will fix it when adding other parts.

PR #507, opened 2015-04-06
Contrast stretching

francisco-dlp:
Add `contrast_stretching` function that replaces similar code used in `drawing.image.py` and `drawing.utils.plot_images`. Also renames `percentile` to `saturated_pixels`.


pburdet:
More tidy and clearer for the users. Thanks

Issue #517, opened 2015-04-08
Problem with 0D signal


pburdet:
The following code fails:

``` Python

>>> s = signals.Spectrum(range(10))
>>> s.integrate1D(0).save('test')


>>> signals.Spectrum(1).save('test')
```

Should we make hyperspy able to save 0D signal? Or should 0D signal should not be generated/supported?

to266:
adding `np.atleast_1d` in the hdf5.py fixes it:

```
diff --git a/hyperspy/io_plugins/hdf5.py b/hyperspy/io_plugins/hdf5.py
index 50c3ef3..7f32613 100644
--- a/hyperspy/io_plugins/hdf5.py
+++ b/hyperspy/io_plugins/hdf5.py
@@ -329,7 +329,7 @@ def dict2hdfgroup(dictionary, group, compression=None):
                 write_signal(value, group.create_group('_sig_' + key))
         elif isinstance(value, np.ndarray):
             group.create_dataset(key,
-                                 data=value,
+                                 data=np.atleast_1d(value),
                                  compression=compression)
         elif value is None:
             group.attrs[key] = '_None_'
@@ -444,7 +444,7 @@ def write_signal(signal, group, compression='gzip'):
         original_metadata = "original_metadata"
 
     group.create_dataset('data',
-                         data=signal.data,
+                         data=np.atleast_1d(signal.data),
                          compression=compression)
     for axis in signal.axes_manager._axes:
         axis_dict = axis.get_axis_dictionary()
```

however then the shape of the signal is not preserved when loading, that would require creating a new flag for 0-d datasets.... Which anyway can be saved (as just scalars) in a different (and probably more compact) way.

francisco-dlp:
Good catch. I think that we shouldn't support 0D because they cannot be indexed. If we don't support them then the bug is not that they cannot be saved but rather that we can create them.

pburdet:
It sounds good.

PR #518, opened 2015-04-08
Fixes the issue #517, but does not preserve the shape of signal when loading

to266:
Fixes the issue #517.

Still need to decide if we want to support 0-d signals and (if so) what flag to use for them in order to preserve the shape when loading


francisco-dlp:
#622 is another attempt at fixing this.

to266:
deprecated by #622

PR #521, opened 2015-04-17
ENH: Line segment marker starts correct

vidartf:
Previously, when creating a line segment marker, a vline was added that screwed up the axes limits (not sure why). Either way, the easiest way to get a line segment in MPL is via the plot function.


francisco-dlp:
Nice. Given that this fixing a bug, could you send another PR with this to hyperspy:0.8.x instead of master? Thanks!

vidartf:
Available in #525.

PR #522, opened 2015-04-17
ENH: Load multi-stacks from multi-signal files

vidartf:
When trying to load a stack, previously files that contained multiple
signals were not allowed. This change allows that as long as each file
has the same number of signals. If so, one stack will be created for
each such sub-signal in the files.

Example:
Files 1-10 each contain 3 signals: A, B and C. With the new changes,
loading files 1-10 with stack=True, will return 3 signals of 10 signals
each (A from 1-10, B from 1-10, C from 1-10). The signals are grouped
simply by order loaded, but possibly some logic could be added that
checks for names etc.

Practical use case:
Building stacks from EMI files.

TODO
--------

* [ ] Unittests
* [x] Docstring documentation
* [x] User Guide documentation


hyperspy/io.py
```diff
@@ -178,24 +178,42 @@ def load(filenames=None,
...
-            signal.metadata.General.title = \
-                os.path.split(
+                if i == 0:
+                    if isinstance(obj, (list, tuple)):
+                        n = len(obj)
+                    else:
+                        n = 1
+                    signals = [[] for j in xrange(n)]
+                else:
+                    if isinstance(obj, (list, tuple)):
```
francisco-dlp: What's the point of the next four lines? If the code above is correct the ValueError exception will never be raised, and if it is incorrect the exception will confuse the users as it is not their fault that it doesn't work...

francisco-dlp:
Cool, I see how this can be useful to  build stacks from emi files. However this works only if the emi files are "stackable" in this way. Therefore, before merging into master this feature needs to be well documented in the docstring and the User Guide, possibly explaining how to acquire emi files that are stackable. It'll also be a good idea to provide unittest if possible, which requires contributing a little set of tiny stackable emi files.

hyperspy/io.py
vidartf: This `ValueError` will be raised if the different files do not contain the same number of signals. `n` is set from the first file loaded, and then is checked for each subsequent file. To make it less confusing, I think to add an exception message along the lines of "Files do not have the same number of sub-signals, so they cannot be stacked", and then throw in offending file names and numbers for details.

vidartf:
> before merging into master this feature needs to ...

Sure, just checking that the feature is wanted and making sure that the implementation details are agreeable first. I'll make the needed changes.

francisco-dlp:
Great!

vidartf:
By the way, do hyperspy support writing emi files? If not, I will have to rely on someone with access to an FEI system to supply stackable emi files for unittests.

francisco-dlp:
No it doesn't. We have some computers with TIA, so I could try to produce a set, but I don't know if I'll manage (I've never used this software for anything but acquisition...).

magnunor:
I have access to TIA 4.2 if you want to generate some test files.

vidartf:
Well, I don't much about it myself, but the files I have are saved with simultaneously aquired signals (BF and HAADF) as sub-signals of the EMI. Not sure how easy this is to setup in the software without hardware, but if there's a microscope attached I'd guess it only a matter of acquiring and saving a few small images (5x10 pixels?) with at least two signals. No sample (or beam) needed unless we also want to use this test data for something else.

vidartf:
All issues should now be addressed, except the unit test as I'm still lacking test files to use.

vidartf:
Is there any other formats except `.emi` that supports several signals? Where perhaps it is easier to get some test files?

francisco-dlp:
No, that's the only one. #576 implements it for hdf5.

francisco-dlp:
Couldn't this features be added to stack instead? This would have the advantage of making it easier to write unittests for it.

vidartf:
> Couldn't this features be added to stack instead?

I'm not sure if I follow what you mean.

francisco-dlp:
`hs.stack([[haadf1, si1], [haadf2, si2]])`

vidartf:
But wouldn't `io.load(<list of emi files>, stack=True)` still be broken?

vidartf:
Oh, no wait, no I see what you mean (rather long time since I looked at this code last): The code should probably be moved to stack as you suggest, as that offers extended use. However, I want to retain the "fail on first bad signal" so that you don't have to load all the files into memory before it raises an exception. This can be done by generator expressions, but require that `hs.stack` only treats it as an iterable collection without indexing (not sure what it does now).

vidartf:
PS: Having stack creation via a generator will probably benefit all types of stack creation, not just muti-signal ones.

francisco-dlp:
I think that @to266 already contributed stack creation via generators somewhere, so that shouldn't be a problem.

vidartf:
I started looking on this, and `hs.misc.utils.stack()` does not currently support creation by generator expressions. As such, moving this functionality into `stack()` is way more complex a task than I have time for right now.

francisco-dlp:
@to266, am I wrong in thinking that you've implemented support for generators in `stack` in one of your branches?

to266:
here https://github.com/hyperspy/hyperspy/pull/736

to266:
It still stacked everything into one signal though

to266:
https://github.com/hyperspy/hyperspy/pull/825 may be slightly more up-to-date.... not sure though, it's been a while

vidartf:
The whole logic flow of `stack()` could use a revamp, both for what @to266 is doing, and what this PR suggests to do. Either way it would take more time than I have available.

francisco-dlp:
If you explain what you have in mind in a proposal issue maybe someone will find the time for it.

vidartf:
The main issues it there is basically two to three more or less independent behaviors merged into one function (only common thing is that they loop over the input). Maybe the functionality should be moved into sub-functions (e.g. `_stack_memory()`, `_stack_mmap()`)? The main function (`stack()`) would then serve as a common entry point that calls the correct function based on the input. If there are common tasks in both of the sub-functions, that logic should be moved out into another function. Finally, the code should be commented so that other people that need to modify the code later can follow the logic.

vidartf:
Merged master and updated to Python3. Also added two simple tests (one success, one expected failure) based on stacking two identical signals.

francisco-dlp:
Needs updating with current master.

sem-geologist:
I am curious about this stuff, bcf returns few signals together with edx hypermaps. as I understand this would allow to load whole bunch of bcf files (in my situation) and i.e. it would stack and produce few cubes:
on for BSE, one for SEI, one hypercube for EDX?

vidartf:
@sem-geologist: Yes, that is the purpose, although it is currently only tested with FEI files. I think the only requirement is that the signals (BSE/SEI/EDX) are returned in a consistent order for each file (e.g. BSE first, then SEI and finally EDX).

francisco-dlp:

PR #523, opened 2015-04-17
ENH: Order of FEI data

vidartf:
I don't have a lot of FEI data files to test this on, but for me the FEI files where loaded with two out of three axes inverted. This fixes the problem for me, but test should be run with other files.


francisco-dlp:
Could you send a new PR to the maintenance branch (0.8.x) making sure that the parent is also 0.8.x?

vidartf:
#545

PR #524, opened 2015-04-20
ENH: Faster image stack navigation

vidartf:
This is a suggestion for a way to speed up image stack navigation (see #160). What it does is to cache the entire stack in display compatible format (i.e. in MPL's `AxesImage`). This is then a trade off between speed and memory. To use it, set the property `cache_stack` on the `ImagePlot` (typically via `s._plot.signal_plot.cache_stack = True`). This will then process the entire stack (which takes a while, and should therefore maybe use the progress indicator). I decided to use a custom version of the animation classes in MPL as I thought it would keep the main class clean and keep all the bookkeeping somewhere else, but we could do without it (and it might actually be slightly faster on our own).

The performance numbers are as follow (test routine below):
```
Without cached stack (i.e. like before):
  setup():  5.704 s
  iterate():  28.382 s

With cached stack:
  setup():  38.182 s
  iterate():  4.891 s

Cache memory use:
  Approximately: Number data elements (in both signal and nav) X 8 bytes
```

Of course, once the cache has been set up, any subsequent navigation will stay fast for any number of iterations (normally the first iteration through is 0.8 s slower actually).

The cache setup does unfortunately not scale linearly with the number of elements, as it slows down as more and more cache gets allocated. This is due to the way the cache is distributed across several `AxesImage` instances. However, this means the cached setup time should approach the uncached iteration time as the data set becomes smaller.

Test code (profiled using cProfile plugin in my IDE):
```python
from matplotlib import use
import numpy as np
use('Qt4Agg')
import matplotlib.pyplot
matplotlib.pyplot.ion()
import hyperspy.hspy as hspy

def setup():
    signal = hspy.signals.Signal(np.random.rand(30, 2000, 2000))
    signal.axes_manager.set_signal_dimension(2)
    signal.plot()
    return signal

def iterate(signal):
    for idx in signal.axes_manager:
        pass

s = setup()
iterate(s)
```

### TODO:
- [ ] Use progress indicator
- [ ] Check for memory availability at start, and raise `MemoryError` if not enough (saves user time + assures plot in valid state). If we have to use the swap disk to cache we lose most of the advantage.
- [ ] Put property on `Signal` so that the behavior is sustained across replots?
- [ ] Try to store `kwargs` passed to plot, so that it can be reused when replotting?
- [ ] Get input on whether to use the MPL animation classes.
- [ ] Test using different backends.

### DIFF
--- a/hyperspy/drawing/image.py
+++ b/hyperspy/drawing/image.py
@@ -136,6 +136,7 @@ class ImagePlot(BlittedFigure):
         self._use_cache = True
         if not value:
             self._cached_stack = None
+            self._stack_colorbars = []
         self.plot()
 
     def configure(self):
@@ -248,6 +249,8 @@ class ImagePlot(BlittedFigure):
                 animated=True)
         if self._colorbar is not None:
             self._colorbar.remove()
+        if self.auto_contrast:
+            self.vmin = self.vmax = None
 
         stack_artists = []
         for index in stack_iterable:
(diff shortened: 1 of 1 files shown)


vidartf:
Overly complex compared to the benefits added.

PR #525, opened 2015-04-20
BUG: Line segment marker starts correct

vidartf:
Previously, when creating a line segment marker, a vline was added that
screwed up the axes limits (not sure why). Either way, the easiest way
to get a line segment in MPL is via the plot function.


francisco-dlp:
It seems that this breaks something when the line segment marker is multidimensional. The following stops working:

```python

# Create test image 100x100 pixels:
img = signals.Image(np.zeros((10, 100,100)))

# Add four line markers:
m1 = utils.plot.markers.line_segment(x1=np.zeros((10, )),
                                     y1=np.zeros((10, )),
                                     x2=np.arange(10, 110, 10),
                                     y2=np.arange(10, 110, 10),
                                     color='red',linewidth=3)

# Plot image and add markers to img:
img.plot()
img.add_marker(m1)

```

vidartf:
Ah, missed the change in underlying MPL type, so the update function need to be fixed. Should be alright now.

PR #438, opened 2015-01-22
Add support for non linear axes

francisco-dlp:
This adds support for non-linear axes. Although it is still work in progress, I think that the main functionality is already here, including slicing support. Please let me know what you think.

Usage
--------

```python

s = signals.Spectrum(np.random.random((64, 1000)))
s.axes_manager[0].__init__(axis=np.arange(64)**10, navigate=True)
s.plot(navigator="spectrum")

````

index
index

TO DO
---------

- [ ] unittests
- [ ] Fix image plotting that relies on DataAxis.scale
- [ ] Search the code and fix all functions that rely on DataAxis.{scale, offset}
- [ ] Increase the hdf5 version number when saving non-linear axes. HyperSpy < 0.8 won't be able to read these files.
- [ ] User Guide documentation
- [ ] Improve syntax
- [ ] Fix performance bug (probably a trigger infinite loop)
- [ ] Raise exception when axes is unsorted

PR #528, opened 2015-04-24
FIX: dm-files with incomplete metadata would not load correctly

magnunor:
For example the StackBuilder plugin by Bernhard Schaffer creates
a 3D-dataset, but the scale, offset and units metadata will only be for
2 dimensions. This would lead to the signal.data being 3D, but with
only 2 axes in the axes_manager. This commit fixes this issue by
making an axis with default values if the data dimensions is larger
then the metadata.

Acknowledgement to Vidar for helping with this :)


francisco-dlp:
Nice. More than fixing a bug, this is enabling loading of dm3 files hacked by StackBuilder, isn't it? Could you add some comments to the code to explain what the modifications does and why? Also, could you add a test to the dm tests to test this feature using the smallest test file that you can create?

magnunor:
Yes, the datasets from the plugin loads nicely now. And if acquiring atomic resolution STEM images s.align2D(reference='cascade') does a pretty good job at correcting for drift :)

I'll make a dm-file to run test, add some comments to the code, and perhaps make the old code a bit more readable in general? Currently the old code is a bit hard to "read".

I'm also sending an email to Bernhard, in case he wants fix it in the plugin.

francisco-dlp:
> make the old code a bit more readable in general?

You're more than welcome to do that in a separate branch.

> I'm also sending an email to Bernhard, in case he wants fix it in the plugin

That would be the best solution indeed.

magnunor:
After discussing this with Bernhard Schaffer it seems this is caused by a bug in GMS: if a 2D image is "resized" into a 3D image (a stack), the calibration will only be for 2 dimensions, while the data will be in 3 dimensions. This can be fixed by opening the file, and resaving it in GMS.

Still, I think it would be useful for HyperSpy to be able to handle files with incomplete calibration data.

I'll add some comments in the code and try to generate a small file for making a test.

Incidentally the StackBuilder plugin also stores a time stamp for every image in the stack, which could be used for calibrating the "Z" direction as a time series.

francisco-dlp:
Great! I agree in that, even if the bug in on DM side, if there are dm files around that won't open nicely in HyperSpy because of this, it'll be nice to have this implemented.

vidartf:
> Incidentally the StackBuilder plugin also stores a time stamp for every image in the stack, which could be used for calibrating the "Z" direction as a time series.

For the moment that would assume equally spaced images in the time domain though, at least until #438 is working.

francisco-dlp:
I'll add #438 to the 0.9 milestone because it is currently blocking several other PR e.g. #478.

magnunor:
I added some comments in the code and a test for loading a dm-file made using the stackbuilder plugin, which is missing one of the dimensions in the metadata.

hyperspy/tests/io/test_dm_stackbuilder_plugin.py
```diff
@@ -0,0 +1,33 @@
...
+from nose.tools import assert_equal
+from hyperspy.io import load
+
+my_path = os.path.dirname(__file__)
+
+class TestStackBuilder():
+    
+    def test_load_stackbuilder_imagestack(self):
+        image_stack = load(my_path + "/dm_stackbuilder_plugin/test_stackbuilder_imagestack.dm3")
+        data_dimensions = len(image_stack.data.shape)
```
francisco-dlp: I suggest the following equivalent syntax:

```python

data_dimensions = image_stack.data.ndim
am = image_stack.axes_manager
axes_dimensions = am.signal_dimension + am.navigation_dimension

```

francisco-dlp:
Nice, I think that this is almost ready to merge except for two little details:

* Could you use autopep8 to correct some of the little style issues? I case that you've missed it, you can configure git to run autopep8 after every commit, see #181.
* Could you resend the PR to 0.8.x, making sure that the base is 0.8.x an not master?

magnunor:
Sent to 0.8.x instead: https://github.com/hyperspy/hyperspy/pull/616

PR #530, opened 2015-04-24
Code clean-up: increase readability of io_plugins.digital_micrograph

magnunor:
I tried to make the code a bit more readable by splitting up some one-codeliner list comprehensions. I don't know if I really succeeded, but to my eyes to code is a little less "opaque" now.

It doesn't break any of the unit tests, and loading my spectra seems to work fine.

If this seem ok and ready to be merged, I'll implement the "fixes" discussed in https://github.com/hyperspy/hyperspy/pull/528 after this one is merged. Seems smarter to separate them, in case there are some regressions due to one or the other.

### DIFF
--- a/hyperspy/io_plugins/digital_micrograph.py
+++ b/hyperspy/io_plugins/digital_micrograph.py
@@ -525,14 +525,22 @@ class ImageObject(object):
     @property
     def scales(self):
         dimensions = self.imdict.ImageData.Calibrations.Dimension
-        return np.array([dimension[1].Scale for dimension in dimensions])[::-1]
+        scale_list = []
+        for dimension in dimensions:
+            scale_list.append(dimension[1].Scale)
+        scale_array = np.array(scale_list)[::-1]
+        return scale_array
 
     @property
     def units(self):
         dimensions = self.imdict.ImageData.Calibrations.Dimension
-        return tuple([dimension[1].Units
-                      if dimension[1].Units else ""
-                      for dimension in dimensions])[::-1]
+        units_tuple = ()
+        for dimension in dimensions:
+            if dimension[1].Units:
+                units_tuple += (dimension[1].Units,)
+            else:
+                units_tuple += ("",)
+        return units_tuple[::-1]
 
     @property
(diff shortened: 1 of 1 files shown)


stefano-m:
Hi,

thank you for your work on Hyperspy.

I understand that list comprehensions are a bit hard to read. However, they are a common Python idiom and I don't feel that this refactoring adds much readability. List comprehensions are also more efficient than for loops, even though performance might not be relevant in this specific case.

(Note that even though I am the original author of this bit of code, I have not been working on it for a long time. In fact, I don't remember having added the code that you refactored).

magnunor:
It might just be my personal taste, since I don't use list comprehensions so much. But I still think functions like the old `get_axes_dict` is a bit too hard to read:

``` python
def get_axes_dict(self):
    return [{'name': name,
                 'size': size,
                 'index_in_array': i,
                 'scale': scale,
                 'offset': offset,
                 'units': unicode(units), }
                for i, (name, size, scale, offset, units) in enumerate(
                    zip(self.names, self.shape, self.scales, self.offsets,
                        self.units))]
```

due to so many things happening in essentially the same code line, especially the nested [{(.

stefano-m:
Hi,

thanks for your reply.

Having had to do with [nested list comprehensions](https://docs.python.org/2/tutorial/datastructures.html#nested-list-comprehensions), I don't find that hard to read, but I guess it's a matter of habit.

In the end, it's your call. Just be aware that if you are iterating over very long lists there may be some performance loss.

Maybe you can ask the opinion of other contributors too.

francisco-dlp:
I agree with @stefano-m on this. Actually, using list comprehensions in combination with `plot_spectra` and `plot_images` it is  possible to create awesome plots with a very compact and readable syntax, so, although it is a mildly advance feature of the Python language, I think that it's worth to promote it in the code and possibly also the documentation. Any more opinions on this?

magnunor:
Ok, closing pull request then. 

I'll try to get https://github.com/hyperspy/hyperspy/pull/528 ready for merging.

Issue #531, opened 2015-04-27
BUG: model.as_signal does not know how to deal with active_is_multidimensional =True components


to266:
if a component list is passed, it switches off the not required components using `c.active` field. However if `active_is_multidimensional` is `True` for any component, it is correctly switched on(or off) only in the first pixel.

PR #532, opened 2015-04-27
Bugfix model as signal multi dim comps

to266:
fix for #531

### DIFF
--- a/hyperspy/component.py
+++ b/hyperspy/component.py
@@ -561,6 +561,28 @@ class Component(t.HasTraits):
         if f in self.connected_functions:
             self.connected_functions.remove(f)
 
+    def _toggle_connect_active_array(self, if_on):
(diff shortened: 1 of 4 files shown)


hyperspy/component.py
```diff
@@ -561,6 +564,32 @@ def disconnect(self, f):
         if f in self.connected_functions:
             self.connected_functions.remove(f)
 
+    def _toggle_connect_active_array(self, if_on):
+        # nothing to do (was never multidimensional)
+        if self._active_array is None:
+            return
+        # as it should be (both True)
+        if self.active_is_multidimensional and if_on:
```
vidartf: This and the next clause can be combined as `if self.active_is_multidimensional == if_on:`

to266: True! Thanks, I'll fix it tomorrow

PR #536, opened 2015-05-06
New_Model2D

dnjohnstone:
Developing 2D model fitting in Hyperspy for use with the image class. It's a work in progress but as a first attempt this commit works with the simple test case below fitting a single 2D Gaussian. Many thanks to Francisco for doing most of it...

Next steps will be fitting multiple components in selected regions.

```python

import numpy as np
import hyperspy.hspy as hs
from hyperspy.model import Model2D

g = hs.components2d.Gaussian2D(
    centre_x=-5,
    centre_y=-5.,
    sigma_x=1.,
    sigma_y=2.)
x = np.arange(-10, 10, 0.01)
y = np.arange(-10, 10, 0.01)
X, Y = np.meshgrid(x, y)
im = hs.signals.Image(g.function(X, Y))
im.axes_manager[0].scale = 0.01
im.axes_manager[0].offset = -10
im.axes_manager[1].scale = 0.02
im.axes_manager[1].offset = -10
im = hs.utils.stack([im] * 2)
              
m = Model2D(im)
m.append(hs.components2d.Gaussian2D())
m.fit()

```


hyperspy/component.py
```diff
@@ -756,6 +756,9 @@ def summary(self):
                                                parameter.std,
                                                parameter.units)
 
+    def __tempcall2d__(self, p, x, y, onlyfree=True):
+        self.fetch_values_from_array(p, onlyfree=onlyfree)
+        return self.function(x, y)
```
vidartf: I think we could get away with not having to define a 2D version of this if we simply replace `x` in the original with `position`, and pass a tuple `(x, y)` for anything 2D. `Gaussian2D.function()` would of course have to be modified to take a tuple instead of two separate arguments, but that's no problem. Should give better code reusability, meaning any changes to `__tempcall__` won't have to be mirrored in 2D version, and it will also be able to support 3D (or higher) versions later on at no extra cost.

to266: Actually the modification of the component would be somewhat problematic - all of the boundaries and other individual properties of a parameter would have to be rewritten to be able to use tuples. While it may be useful for support of even higher dimensions, I think it would drastically reduce readability and usability of models in general (of we would have to "unwrap" the tuples for users anyway, which sort of defeats the purpose).

I also had a look and it's possible to get rid of the coordinates argument for the `__tempcall__` altogether, however that makes it more complex (i.e. checking what kind of model the component is in), and the maintainability of the 1D model is reduced dramatically - namely convolved models now have part of the code in components, and it is not obvious at all that anything changes from convolved to non-convolved cases unless you actually look at the code.

Then again, maybe there can be more structural changes that make it clearer again, but I doubt it's worth it for now.

Besides, while 3d models do sound fancy and exciting, I think it makes little sense to support them yet - not only very few use-cases exist IMO, but the parameter space gets enormous with higher dimensionality, so likely completely new algorithms have to be employed. Even the 2D gaussian fit (which is/should be an extremely trivial example to test that everything works) given by Duncan takes surprisingly long on my machine.

vidartf: I'm convinced: I agree that readability and usability trumps the code reuse in this case. :)

to266: ...though I just may be avoiding rewriting the parameters to be truly multi-dimensional. It's (apparently) the oldest part of the HyperSpy, it may be time to rewrite / restructure the code. It's a big task, though. I'd love to do it, but don't think I can afford to atm :)

magnunor:
I added a rotation parameter to the 2D Gaussian: https://github.com/dnjohnstone/hyperspy/pull/1

dnjohnstone:
I think this is getting quite close except for tests (which I'm doing now) but I just wanted to clarify a few things that I noticed might cause confusion while I was updating docs...

1. As it is this pull request includes a method "create_model" for both a Spectrum and an Image - the method calls Model1D in the former case and Model2D in the latter case. Is there a convention on having slightly different methods with identical names? I think it's good because it makes the syntax essentially identical for both kinds of model...but?

2. This pull request has 3 classes currently called "BaseModel", "Model1D" and "Model2D" - I think it might be better to simply rename "BaseModel" --> "Model" it would at least make the user-guide read more nicely in English and I think makes sense anyway. (may also minimise bugs with people still writing .Model which will usually be the correct thing. Opinions?

3. There are a number of functions that exist for Model1D but not for Model2D (essentially the plotting ones) - since they simply don't exist it's a little unclear to me how this should be in terms of error raising... Is it ok (in the interest of getting the restructuring merged) if I simply note in the user-guide and Model2D docstring that plotting these models is not yet provided?

4. Is there anything not mentioned above that needs doing to get this merged up?

to266:
1. That is actually a very good thing and done intentionally (i.e. each signal creates only appropriate model)
 2. Essentially the `BaseModel` is just that - the bare-bones version of the model that is not usable on its own. I may be wrong, but I think there is no way to even create it? I think for these reasons the name fits. It should not be changed in the user-guide, though (i.e. it should stay as "Model"). That should be clear enough, as "Model" will refer to both `Model1D` and `Model2D`, that will both have all the methods defined in `BaseModel`.
 3. Use a warning like this (https://github.com/hyperspy/hyperspy/blob/master/hyperspy/model.py#L369 ) for now, and we'll gradually implement those functions.

dnjohnstone:
@francisco-dlp this is now tidied up and passes Travis including the new tests for model and components, which is good. There is some significant functionality missing for Model2D i.e. plotting and gradient based methods (because _jacobian) needs to be done a different way... I have included all the methods that are available for Model1D into the Model2D class though and just set them to raise a not implemented error. I don't really have time to do these things quickly (although I'd like to and would be happy to in the longer term) and I think we're all in agreement that getting the split of Model into the master as soon as possible, even without much functionality is a good thing for everyone... I'll finish tests for the functions that are there tonight, which should make it pass coveralls - will that be considered enough? (assuming normal review)

@vidartf I think you might have done a lot of the plotting stuff already? Do you think replicating what's available for Model1D in the case of Model2D could be quick/easy?

vidartf:
@dnjohnstone : Not sure how you imagine the plotting is done best:
 1. Plot signal and model as separate images
 2. Plot signal and model overlaying in an x-y-z plot
 3. Other ways?

vidartf:
PS: I'd argue that model.py should be split into several files as it is becoming rather unmanageable, and splitting it across 1D/2D seems logical.

dnjohnstone:
@vidartf Thanks for quick reply - I think that option 1 i.e. plot the signal and model next to each other is the easiest to implement and actually perhaps the most useful to the user (certainly I'd find it easier to visualise that way).

As regards splitting model.py into multiple files --> I guess you mean it becomes 3 files i.e. basemodel.py, model1D.py and model2D.py

I can see that being easier in the long run as we will hopefully develop more specialised tools for each type of model so they may well grow... If others agree, that split is pretty easy so just say?

vidartf:
@dnjohnstone : Well, in that case the signal plot and model plot can exist independently from each other (but with shared navigation). This should mean that you don't have to custom-wire the plot as much, and will therefore be less dependent on the internals of the plotting. However, it does make it difficult to have a plot where the individual components are separated (current `Model.enable_plot_components()`). Color coding could work, but that limits you to a three component highlighting.

If you go with option 1, I think you should be able to do the main plotting by creating a new `ImagePlot` instance. This can be initialized in a similar way as what is done in `MPL_HyperImage_Explorer.plot_signal()`, but instead of connecting to its `navigator_plot`, you'd connect to the signal's plot. Then, most of the rest of the plotting code should follow naturally (just a matter of wiring the right bits together). Without color coding you can simply wire the `data_function` to `Model.__call__()`, if you want color coding you need to create a wrapper to add the coloring.

Regarding the `set_signal_range` functions and similar, I think it would be best to tie that up to the ROI PR. Then we won't have duplicate work for 2D widgets.

francisco-dlp:
I agree with both in that option 1 for plotting is the most sensible
approach. We can then add widgets for certain components later on e.g. a
point at the centre for a gaussian etc.

+1 also on splitting model. I suggest to keep `BaseModel` in `
hyperspy.model.py` and to create the two new files in `hyperspy.models`
i.e. `hyperspy.models.model1d.py` and `hyperspy.models.model2d.py`. In this
way we may ease the merging of old PRs.

Regarding easing the merging task, we really should aim at merging this
asap even if some functionality is not yet there, because rebasing this
branch everytime that `model.py` gets some care (which is very often
lately) must not be fun...

2015-12-06 12:32 GMT+00:00 Vidar Tonaas Fauske <notifications@github.com>:

dnjohnstone:
Ok - I'll split the model files.

Plotting-wise -- I guess the right thing to do is probably to implement what @vidartf described as it's very natural that someone using this is going to want to plot the results (in fact I can't really imagine someone using it and not wanting that). It doesn't sound, or look, too bad so I'll have a go at that too. I must say I haven't really looked at the internals of HyperSpy plotting before though - so this may take a little while... we'll see.

dnjohnstone:
Ok - this should now be ready for review... hopefully there are not too many slips remaining!

In the interest of keeping things manageable and getting the main split sorted quickly the following functionality is not yet implemented for 2D-models: 

1. Plotting 2D models
2. Gradient based fitting methods (requires a rewrite of _jacobian)
3. Many components - the most obvious next additions would be Expression2D and ScalableFixedPattern2D by analogy to those available in the 1D case.

I'll open issues to this effect so they are not forgotten and will probably add ScalableFixedPattern2D myself in the next couple of weeks.

to266:
I hope to review this branch today. Most likely will leave a bunch of small comments where I think necessary, feel free to update as I go or just wait until I went through the diff once.

Oh and currently there are merge conflicts ;)

doc/user_guide/model.rst
```diff
@@ -13,26 +26,38 @@ bounds and weights.
...
-    >>> s = hs.load('YourDataFilenameHere') # Load the data from a file
-    >>> m = s.create_model() # Create the model and asign it to the variable m
+    >>> im = hs.signals.Image('SomeDataHere') # Load the data from a file
+    >>> mod = im.create_model() # Create the 2D-Model and asign it to the variable mod
 
-At this point you may be prompted to provide any necessary information not
-already included in the datafile, e.g.if s is EELS data, you may be asked for
-the accelerating voltage, convergence and collection angles etc.
+The syntax for creating both one-dimensional and two-dimensional models is thus
+identical for the user in practice.When a model is created  you may be prompted
```
to266: Space before `When`

doc/user_guide/model.rst
```diff
@@ -247,7 +277,7 @@ index in the model.
 .. versionadded:: 0.8.1 :py:attr:`components` attribute
 
 In addition, the components can be accessed in the
-:py:attr:`~.model.Model.components` `Model` attribute. This is specially
+:py:attr:`~.models.model1D.Model1D.components` `Model` attribute. This is specially
```
to266: Any reason why the `components` module is not defined for the `BaseModel`? Of course if just copy-pasting does not work, let me know and we'll add that to the list of things not yet supported for `Model2D`

doc/user_guide/model.rst
```diff
@@ -716,9 +746,9 @@ by hand.
     :py:meth:`~.model.Model.enable_adjust_position` and
     :py:meth:`~.model.Model.disable_adjust_position`
 
-Also, :py:meth:`~.model.Model.enable_adjust_position` provides an interactive
+Also, :py:meth:`~.models.model1D.Model1D.enable_adjust_position` provides an interactive
 way of setting the position of the components with a well define position.
-:py:meth:`~.model.Model.disable_adjust_position` disables the tool.
+:py:meth:`~.models.model1D.Model1D.disable_adjust_position` disables the tool.
```
to266: These methods are in `Model2D` as well, just raising a `NotImplementedError`, just like `suspend_update` and others. For consistency I would use `BaseModel` eveywhere

doc/user_guide/model.rst
```diff
@@ -733,37 +763,37 @@ way of setting the position of the components with a well define position.
...
-:py:meth:`~.model.Model.fetch_stored_values`.
+:py:meth:`~.model.BaseModel.store_current_values` and
+:py:meth:`~.model.BaseModel.fetch_stored_values`.
 
 
 Visualising the result of the fit
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
-The :py:class:`~.model.Model` :py:meth:`~.model.Model.plot_results`,
+The :py:class:`~.models.model1D.Model1D` :py:meth:`~.models.model1D.Model1D.plot_results`,
```
to266: as previous

hyperspy/_components/gaussian2d.py
```diff
@@ -0,0 +1,92 @@
...
+    def fwhm_x(self, value):
+        self.sigma_x.value = value/sigma2fwhm
+
+    @property
+    def fwhm_y(self):
+        return self.sigma_y.value*sigma2fwhm
+
+    @fwhm_y.setter
+    def fwhm_y(self, value):
+        self.sigma_y.value = value/sigma2fwhm
```
to266: Gradients for the parameters should be trivial (however not essential) to add. I'm just leaving a reminder for someone to eventually get to it

hyperspy/model.py
```diff
@@ -499,133 +378,6 @@ def _touch(self):
...
-        if self._model_line is None:
-            return
-        for component in self:
-            component.disconnect(self._model_line.update)
-            for parameter in component.parameters:
-                parameter.disconnect(self._model_line.update)
-        if self._plot_components is True:
-            self._disconnect_component_lines()
-
-    def as_signal(self, component_list=None, out_of_range_to_nan=True,
```
to266: Should be model dimensionality-ambiguous - i.e. in `BaseModel`. I guess the bits that get in the way are `channel_switches`?

hyperspy/model.py
```diff
@@ -822,231 +505,9 @@ def _model2plot(self, axes_manager, out_of_range2nans=True):
...
-
-        """
-        i1, i2 = self.axis.value_range_to_indices(x1, x2)
-        self._add_signal_range_in_pixels(i1, i2)
-
-    def reset_the_signal_range(self):
-        self.channel_switches[:] = True
-        self.update_plot()
-
-    def _model_function(self, param):
```
to266: Should be left in `BaseModel`. Other models will define `__call__` method, and the rest should be fine

hyperspy/model.py
```diff
@@ -822,231 +505,9 @@ def _model2plot(self, axes_manager, out_of_range2nans=True):
...
-                                np.add(par_grad, par.grad(
-                                    axis), par_grad)
-                        grad = np.vstack((grad, par_grad))
-                    counter += component._nfree_param
-            to_return = grad[1:, :] * weights
-        if self.spectrum.metadata.Signal.binned is True:
-            to_return *= self.spectrum.axes_manager[-1].scale
-        return to_return
-
     def _function4odr(self, param, x):
```
to266: Since it takes `x` as a parameter, should it not be moved to `Model1D`?

hyperspy/model.py
```diff
@@ -822,231 +505,9 @@ def _model2plot(self, axes_manager, out_of_range2nans=True):
...
         return self._model_function(param)
 
-    def _jacobian4odr(self, param, x):
-        return self._jacobian(param, x)
-
     def _poisson_likelihood_function(self, param, y, weights=None):
```
to266: Same as above, is it really dimension-ambiguous?

hyperspy/model.py
```diff
@@ -1087,27 +532,27 @@ def _errfunc4mpfit(self, p, fjac=None, x=None, y=None,
...
 
-            variance = self.spectrum.metadata.Signal.Noise_properties.variance
+            variance = self.signal.metadata.Signal.Noise_properties.variance
             if isinstance(variance, Signal):
                 variance = variance.data.__getitem__(
                     self.axes_manager._getitem_tuple)[self.channel_switches]
         else:
             variance = 1.0
-        d = self(onlyactive=True) - self.spectrum()[self.channel_switches]
+        d = self(onlyactive=True) - self.signal()[self.channel_switches]
```
to266: `self(onlyactive=True).ravel()` is required, as `[self.channel_switches]` will return a flat array.

hyperspy/model.py
```diff
@@ -1087,27 +532,27 @@ def _errfunc4mpfit(self, p, fjac=None, x=None, y=None,
...
 
     def _set_current_degrees_of_freedom(self):
-        self.dof.data[self.spectrum.axes_manager.indices[::-1]] = len(self.p0)
+        self.dof.data[self.signal.axes_manager.indices[::-1]] = len(self.p0)
 
     @property
     def red_chisq(self):
         """Reduced chi-squared. Calculated from self.chisq and self.dof
         """
         tmp = self.chisq / (- self.dof + sum(self.channel_switches) - 1)
```
to266: Replace `sum(self.channel_switches)` with `self.channel_switches.sum()`. Of course neither works when `self.channel_switches = None`

hyperspy/tests/model/test_components2D.py
```diff
@@ -0,0 +1,37 @@
+import numpy as np
+import nose.tools as nt
+
+import hyperspy.api as hs
+
+
+class TestGaussian2D:
```
to266: This class should be renamed and/or moved to a different file. At the moment it claims to test `Gaussian2D`, however tests `Model2D` fitting using identical gaussian, which can in principle be a big number of functions and not necessarily a gaussian.

The test itself is useful, to test fitting, however an additional test is required to compare the gaussian function output with some defined array to check that it actually works

doc/user_guide/model.rst
vidartf: Agree. The tilde removes everything but the last bit from what the users see anyway. The first bits are only important for where the link in the doc leads.

dnjohnstone: Sure - I think that was a relic from before I also included these methods in Model2D with the NotImplementedError - will change.

hyperspy/model.py
dnjohnstone: Yes - I agree in principle. I moved these to Model1D and 2D because the disconnecting-connecting methods are not implemented for 2D but on reflection it doesn't really matter because it will just raise a NotImplementedError for Model 2D anyway -- I'll move the above bits to BaseModel

dnjohnstone: Just to note though the as_signal method is a bit more problematic because it uses the spectrum class and has options for convolution... I think the best is probably to move it to Model1D as a method as_spectrum and then eventually have an equivalent as_image for Model2D

hyperspy/model.py
dnjohnstone: ok

hyperspy/tests/model/test_components2D.py
dnjohnstone: Thanks for that - I'll move this to test_model and add a new test that does what you describe

hyperspy/model.py
dnjohnstone: Done

hyperspy/model.py
dnjohnstone: Done

dnjohnstone: Actually - I'm not sure that's correct... Implementing what you suggest breaks the fitting test that I wrote as in the screenshot below. Is this indicative of a more fundamental error in something I've done, or is it just a slip?
screenshot from 2015-12-12 11-32-19

hyperspy/model.py
dnjohnstone: Mmm I agree in principle but this is causing many issues with nosetests that I'm not sure how to resolve. The primary reason is because different arguments need to be passed to __call__ for model1D and model2D, or rather it doesn't make sense to pass non_convolved to model2D

It seems the main thing that _model_function does is to set onlyactive=True but if you do that and just don't specify non_convolved it fails a test for GetExplainedVarianceRatio - can't really work out why... Not sure what the best solution is, but the simplest is to specify _model_function for both Model1D and Model2D

vidartf: Strictly, everything with low-loss should be moved to `EELSModel`. If convolution stuff is left in model, it should be made general. For 2D, convolution should typically be with the point spread function then.

dnjohnstone: Ok sold - on the basis that 2D convolution could well make sense in the future I've added the non_convolved argument to __call__ for Model2D and then followed the suggestion to keep this in BaseModel

dnjohnstone:
Thanks @to266 and @vidartf for your comments - I think I've responded to all points now.

Let me know if more is required.

hyperspy/model.py
to266: Yes, that's something I forgot to mention - in `Model2D` you set the default value for `channel_switches` as `None`. In numpy that adds a new dimension, which is not what I suspect you want.  Instead, if you want just a generic thing that does not do anything, you should use `slice(None)`:
```python
In [8]: a.shape
Out[8]: (5, 7)

In [9]: a[None].shape
Out[9]: (1, 5, 7)

In [10]: a[slice(None)].shape
Out[10]: (5, 7)
```

Ideally (i.e. that is something we will have in the end, I have no doubts) `channel_switches` for 2d model should be initialized as `self.channel_switches = np.ones(self.xaxis.shape, dtype=bool)` - i.e a 2D array of the "signal" shape

hyperspy/models/model2D.py
```diff
@@ -0,0 +1,284 @@
...
+    def __init__(self, image, dictionary=None):
+        self.image = image
+        self.signal = self.image
+        self.axes_manager = self.signal.axes_manager
+        self.xaxis, self.yaxis = np.meshgrid(
+            self.axes_manager.signal_axes[0].axis,
+            self.axes_manager.signal_axes[1].axis)
+        self.axes_manager.connect(self.fetch_stored_values)
+        self.free_parameters_boundaries = None
+        self.channel_switches = None
```
to266: The definition I just mentioned in another comment... Sorry for forgetting it.

hyperspy/model.py
to266: I debated suggesting moving all convolutions to `EELSModel` but that may involve too much work, and:
 1. We want to merge this asap
 2. Convolutions in hyperspy in general will have to be redone to be more general (i.e. to be able to convolve nd-signals and even two components)

dnjohnstone:
Ok @to266 thanks for that -- I've made the channel_switches changes you pointed out (thanks for that, makes much more sense now!).... So everything you said should be done now. Let me know if there's anything more

hyperspy/model.py
```diff
@@ -1257,11 +790,10 @@ def fit(self, fitter=None, method='ls', grad=False,
         elif fitter == "odr":
             modelo = odr.Model(fcn=self._function4odr,
                                fjacb=odr_jacobian)
-            mydata = odr.RealData(
-                self.axis.axis[
-                    self.channel_switches], self.spectrum()[
-                    self.channel_switches], sx=None, sy=(
-                    1 / weights if weights is not None else None))
+            mydata = odr.RealData(self.axis.axis[self.channel_switches],
```
to266: `Model2D` does not have `self.axis` or `self._function4odr` for the moment. I'm not sure if the fitter supports 2d data, or we'll have to trick it.. Any way, another note for the "cleanup" once this is merged - either we'll have to add all the missing functions and change this bit to actually work, or raise an error before it gets to here.

hyperspy/models/model2D.py
```diff
@@ -0,0 +1,268 @@
...
+                self.chisq.data,
+                dtype='int'))
+        self.dof.metadata.General.title = self.signal.metadata.General.title + \
+            ' degrees of freedom'
+        # self._suspend_update = False
+        self._adjust_position_all = None
+        self._plot_components = False
+        self.components = ModelComponents(self)
+        if dictionary is not None:
+            self._load_dictionary(dictionary)
```
to266: `_whitelist` and `_slicing_whitelist` are missing from `Model2D`. Should be very similar though, just with irrelevant bits removed (i.e. `low_loss`)

dnjohnstone: ok -- I think I removed that because I thought it was only used by ModelSpecialSlicers -- which I put in Model1D but not Model2D.

Looking at the code for ModelSpecialSlicers though it is probably dimension ambiguous if defined for .signal rather than .spectrum - do you think that's right? If so shall I put that in BaseModel then?

hyperspy/model.py
dnjohnstone: it also doesn't have self._jacobian so I don't think the odr fitter will work for many reasons at the moment... My thinking at the moment was lets separate out as much as we can that can be general compared to that which can't and then go back through and add in functionality for Model2D bit-by-bit. Mostly because if we tried to do everything in one go the diff would be massive and it would be hard to track down errors.

I actually want to use the gradient based fitting methods in 2D for something I'm working on so this will be addressed before long.

dnjohnstone:
@to266 many thanks for the whitelist comment -- I've added it for Model2D (as well as having a general tidy up of __init__ so that they should equate nicely between the two models. Also made clear that ModelSpecialSlicers is properly dimension ambiguous so I moved that to BaseModel and relabelled for .signal rather than .spectrum.

to266:
Sorry I'm away and can't do that myself, but could you check that `model2d.as_dictionary` and slicing work? They should mostly work, but may require additional lines to point to `components2d` and similar. Though I suppose I'll just make them work after this is merged, but I'm still curious. Once you've checked, if @francisco-dlp is happy, I suggest merging, although I'm away today, and can't actually spend enough time myself to check again.

hyperspy/model.py
```diff
@@ -609,12 +456,12 @@ def as_signal(self, component_list=None, out_of_range_to_nan=True,
         pbar.finish()
```
to266: Here slightly higher (can't comment there for some reason) it should read 
```python
data[self.axes_manager._getitem_tuple][
    self.channel_switches] = self.__call__(
    non_convolved=not self.convolved, onlyactive=True).ravel()
```
(i.e. add `.ravel()` to the end)

to266:
Also in signal.py `store` definition, change (around like 173 I looks like)
`if model.spectrum is self._signal`

to `if model.signal is self._signal`

hyperspy/model.py
```diff
@@ -602,19 +449,19 @@ def as_signal(self, component_list=None, out_of_range_to_nan=True,
...
             i += 1
             if maxval > 0:
                 pbar.update(i)
         pbar.finish()
         if out_of_range_to_nan is True:
             self.channel_switches[:] = channel_switches_backup
-        spectrum = self.spectrum.__class__(
+        signal = self.signal.__class__(
             data,
             axes=self.spectrum.axes_manager._get_axes_dicts())
```
to266: Sorry, one more - in this line change `spectrum` to `signal`

Issue #533, opened 2015-04-28
BUG: incomplete copying when using spectrum slicing and copy


magnunor:
When trying to align a low loss spectrum using a slice of itself, I noticed some weird behavior. The zero loss peak was not properly aligned. This can be replicated using the code: 

```
import hyperspy.hspy as hspy
import numpy as np
import copy

g = hspy.components.Gaussian(A=100., sigma=10., centre=20)
data = []
for i in range(10):
    g.centre.value = i*10+120
    data.append(g.function(np.arange(1000)))
s = hspy.signals.EELSSpectrum(data)

s1 = copy.deepcopy(s)
s2 = copy.deepcopy(s)
s3 = copy.deepcopy(s)

s1.copy()[...,80:240].align_zero_loss_peak(also_align=[s1])
s2[...,80:240].align_zero_loss_peak(also_align=[s2])
copy.deepcopy(s3)[...,80:240].align_zero_loss_peak(also_align=[s3])
```
Original spectrum (s):
![spectrum_original](https://cloud.githubusercontent.com/assets/1690979/7378660/5834fe42-edef-11e4-8618-244d6e7a5813.png)

Spectrum (s1) using s.copy:
![spectrum_copy](https://cloud.githubusercontent.com/assets/1690979/7378667/67d4aca8-edef-11e4-94a3-6da2e582e4f2.png)

Spectrum (s2):
![spectrum_align](https://cloud.githubusercontent.com/assets/1690979/7378670/70968c4e-edef-11e4-842a-1c9bee65d82a.png)

Spectrum copied using deepcopy (s3):
![spectrum_deepcopy](https://cloud.githubusercontent.com/assets/1690979/7378674/790f677e-edef-11e4-8f61-193fcd8e73a9.png)

Only the last spectrum using deepcopy is as expected. 

So I'm guessing there is some kind of bug in the slicing code, leaving some part of the new spectrum linked to the old one.

francisco-dlp:
Indeed, when using `copy` or slicing the data of the new `Signal` instances are a view of the original data but this is supposed to be a feature not a bug ;) Therefore, I would expect only the `deepcopy` approach to work in this case.

Some comments:
* `s1.copy()[...,80:240]` and `s2[...,80:240]` are equivalent.
* You could get a deepcopy more conveniently using `s3.deepcopy()` instead of `copy.deepcopy(s3)`

[...]

Issue #388, opened 2014-10-26
Remove Signal fancy indexing feature


francisco-dlp:
Currently we have two ways to index a `Signal`: i) directly using the [] syntax and ii) using the special slicers `.isig` and `.inav`.  Although ii) is more verbose than i), it is also more precise and explicit. For example, for a signal with dimension (2, 3, 4| 5) `s[:,:,2]` is equivalent to `s.inav[:,:,2]`. However, if we rearrange the axes so that it becomes (2, 3 | 4, 5) `s[:, :, 2]` returns the first signal axis, while `s.inav[:, :, :2]` raises an exception. The latter, in my opinion, is more correct, as it can help avoiding difficult to debug bugs.  More importantly, from [The Zen of Python](http://legacy.python.org/dev/peps/pep-0020/):

    There should be one-- and preferably only one --obvious way to do it.

So, I propose to remove the possibility to directly slice `Signal`. What do you think?

PR #544, opened 2015-05-09
Enh align zero loss peak signal range

magnunor:
I added a `signal_range` argument for `eels_spectrum.align_zero_loss_peak`. As discussed in https://github.com/hyperspy/hyperspy/issues/533. 

Can potentially reduce the risk of random spikes being mistaken as the ZLP.

### DIFF
--- a/hyperspy/_signals/eels.py
+++ b/hyperspy/_signals/eels.py
@@ -215,7 +215,7 @@ class EELSSpectrum(Spectrum):
             Will only search for the ZLP within the signal_range. If given
             in integers, the range will be in index values. If given floats,
             the range will be in spectrum values. Useful if there are features
-            in the spectrum which are more intensite than the ZLP.
+            in the spectrum which are more intense than the ZLP.
             Default is searching in the whole signal.
 
         Examples
@@ -279,7 +279,7 @@ class EELSSpectrum(Spectrum):
         right = (right if right < self.axes_manager[-1].axis[-1]
                  else self.axes_manager[-1].axis[-1])
         self.align1D(left, right, also_align=also_align, **kwargs)
-        zlpc = estimate_zero_loss_peak_centre(self, mask, signal_range)
+        zlpc = estimate_zero_loss_peak_centre(self, mask, (-3., 3.))
         if calibrate is True:
             substract_from_offset(without_nans(zlpc.data).mean(),
                                   also_align + [self])
--- a/hyperspy/_signals/eels.py
+++ b/hyperspy/_signals/eels.py
@@ -243,41 +243,44 @@ class EELSSpectrum(Spectrum):
                 signal.axes_manager[-1].offset -= value
         
         if signal_range:
-            self_signal = self[...,signal_range[0]:signal_range[1]].deepcopy()
-            also_align.append(self)
+            zlpc = self[...,signal_range[0]:signal_range[1]].estimate_zero_loss_peak_centre(mask=mask)
         else:
-            self_signal = self
-
-        zlpc = self_signal.estimate_zero_loss_peak_centre(mask=mask)
+            zlpc = self.estimate_zero_loss_peak_centre(mask=mask)
         mean_ = without_nans(zlpc.data).mean()
         if print_stats is True:
             print
             print(underline("Initial ZLP position statistics"))
             zlpc.print_summary_statistics()
 
-        for signal in also_align + [self_signal]:
+        for signal in also_align + [self]:
             signal.shift1D(-zlpc.data + mean_)
 
         if calibrate is True:
-            zlpc = self_signal.estimate_zero_loss_peak_centre(mask=mask)
+            if signal_range:
+                zlpc = self[...,signal_range[0]:signal_range[1]].estimate_zero_loss_peak_centre(mask=mask)
+            else:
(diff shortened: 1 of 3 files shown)


francisco-dlp:
Nice. However, wouldn't it be possible to avoid deepcopying? Actually, I think that it might be enough to restrict the signal range when calling `estimate_zero_loss_peak_centre` because, by the time that `align_zero_loss_peak` call `align1D` it has already restricted the signal range internally to an interval of 3eV around the ZLP. What do you think?

magnunor:
I removed the deepcopy. 

Could maybe make it even better by implementing a `signal_range` argument in `estimate_zero_loss_peak_centre`, and passing the argument from `align_zero_loss_peak` there?

hyperspy/_signals/eels.py
```diff
@@ -210,6 +211,22 @@ def align_zero_loss_peak(
             It must have signal_dimension = 0 and navigation_shape equal to the
             current signal. Where mask is True the shift is not computed
             and set to nan.
+        signal_range : tuple of integers, tuple of floats. Optional
+            Will only search for the ZLP within the signal_range. If given
+            in integers, the range will be in index values. If given floats,
+            the range will be in spectrum values. Useful if there are features
+            in the spectrum which are more intensity than the ZLP.
```
francisco-dlp: Small typo: intensity  should probably be intense.

hyperspy/_signals/eels.py
```diff
@@ -224,8 +241,11 @@ def align_zero_loss_peak(
         def substract_from_offset(value, signals):
             for signal in signals:
                 signal.axes_manager[-1].offset -= value
-
-        zlpc = self.estimate_zero_loss_peak_centre(mask=mask)
+        
+        if signal_range:
+            zlpc = self[...,signal_range[0]:signal_range[1]].estimate_zero_loss_peak_centre(mask=mask)
```
francisco-dlp: I propose the following equivalent but more readable syntax (see #388 for a discussion):

```python

zlpc = self.isig[signal_range[0]:signal_range[1]].estimate_zero_loss_peak_centre(mask=mask)

```

hyperspy/_signals/eels.py
```diff
@@ -236,7 +256,10 @@ def substract_from_offset(value, signals):
             signal.shift1D(-zlpc.data + mean_)
 
         if calibrate is True:
-            zlpc = self.estimate_zero_loss_peak_centre(mask=mask)
+            if signal_range:
+                zlpc = self[...,signal_range[0]:signal_range[1]].estimate_zero_loss_peak_centre(mask=mask)
```
francisco-dlp: The same goes here.

hyperspy/_signals/eels.py
```diff
@@ -254,7 +277,10 @@ def substract_from_offset(value, signals):
         right = (right if right < self.axes_manager[-1].axis[-1]
                  else self.axes_manager[-1].axis[-1])
         self.align1D(left, right, also_align=also_align, **kwargs)
-        zlpc = self.estimate_zero_loss_peak_centre(mask=mask)
+        if signal_range:
+            zlpc = self[...,-3.:3.].estimate_zero_loss_peak_centre(mask=mask)
```
francisco-dlp: And here.

hyperspy/tests/signal/test_eels.py
```diff
@@ -132,6 +132,18 @@ def test_also_aligns(self):
...
         nose.tools.assert_equal(zlpc.data.std(), 0)
 
+    def test_align_zero_loss_peak_with_spike_signal_range(self):
+        s = self.spectrum
+        spike = np.zeros((10,100))
+        spike_amplitude = 20
+        spike[:,75] = spike_amplitude
+        s.data += spike
+        s.align_zero_loss_peak(print_stats=False, subpixel=False, signal_range=(98.,102.))
+        zlp_max = s[...,-0.5:0.5].max(-1).data
```
francisco-dlp: And here.

hyperspy/_signals/eels.py
francisco-dlp: Shouldn't this be:

```python

 zlpc = self.isig[signal_range[0]:signal_range[1]].estimate_zero_loss_peak_centre(mask=mask)

```

hyperspy/_signals/eels.py
```diff
@@ -224,8 +241,11 @@ def align_zero_loss_peak(
         def substract_from_offset(value, signals):
             for signal in signals:
                 signal.axes_manager[-1].offset -= value
-
-        zlpc = self.estimate_zero_loss_peak_centre(mask=mask)
+        
+        if signal_range:
```
francisco-dlp: Actually, given that the following 4 lines are repeated several times in the code it might be worth it to turn them into a function and call the function instead.

```python

def estimate_zero_loss_peak_centre(s, mask, signal_in_range):
    if signal_range:
        zlpc = s.isig[signal_range[0]:signal_range[1]].estimate_zero_loss_peak_centre(mask=mask)
    else:
        zlpc = s.estimate_zero_loss_peak_centre(mask=mask)

   return zlpc

```

magnunor:
I implemented the changes you pointed out @francisco-dlp, and did the PEP8 style changes.

So it should hopefully be good to merge now :)

PR #545, opened 2015-05-10
ENH: Order of FEI data

vidartf:
I don't have a lot of FEI data files to test this on, but for me the FEI
files where loaded with two out of three axes inverted. This fixes the
problem for me, but test should be run with other files.

### TODO
 - [ ] Check if this issue is pervasive for all FEI files, or only certain ones.
 - [ ] Possibly acquire small 3D file to create test

### DIFF
--- a/hyperspy/io_plugins/fei.py
+++ b/hyperspy/io_plugins/fei.py
@@ -425,7 +425,7 @@ def ser_reader(filename, objects=None, verbose=False, *args, **kwds):
 
     dc = dc.reshape(array_shape)
     if record_by == 'image':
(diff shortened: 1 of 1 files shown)


vidartf:
Note: Previous code reversed all axes, new code only reverses 'x'. Again, I do not know whether FEI has changed how it saves data, only that this gives correct order for my data (taken on the Osiris at Cambridge).

francisco-dlp:
I'll merge this (because it's fixing a bug) but I'll open a new issue about FEI reader requiring test.

PR #547, opened 2015-05-27
New util: xray near energy

vidartf:
Added a utility function to lookup x-ray lines near a given energy. Useful for matching unknown EDS peaks.

### TODO
- [x] Tests
- [x] Documentation


hyperspy/misc/eds/utils.py
```diff
@@ -13,11 +13,44 @@ def _get_element_and_line(Xray_line):
...
+    List of xray-lines sorted by energy difference to given energy.
+    """
+    valid_lines = []
+    E_min, E_max = energy - spread, energy + spread
+    for element, el_props in elements_db.iteritems():
+        # Not all elements in the DB have the keys, so catch KeyErrors
+        try:
+            lines = el_props['Atomic_properties']['Xray_lines']
+        except KeyError:
+            continue
```
pburdet: A question of style, something like that works as well

```python
    
if 'Xray_lines' in el_props['Atomic_properties']:
    lines = el_props['Atomic_properties']['Xray_lines']
```

hyperspy/misc/eds/utils.py
```diff
@@ -13,11 +13,44 @@ def _get_element_and_line(Xray_line):
 
 
 def _get_energy_xray_line(xray_line):
-    energy, line = _get_element_and_line(xray_line)
-    return elements_db[energy]['Atomic_properties']['Xray_lines'][
+    element, line = _get_element_and_line(xray_line)
+    return elements_db[element]['Atomic_properties']['Xray_lines'][
         line]['energy (keV)']
```
pburdet: Thanks for the correction

hyperspy/misc/eds/utils.py
vidartf: Yes, as long as they all have `Atomic_properties`. If they do (and always will), I'll make the change.

pburdet:
Tests and documentation are needed

hyperspy/misc/eds/utils.py
pburdet: The database has been so far build for EELS and EDS purpose. Elements always have atomic properties. Anyway, minor issue. Your version is more stable

vidartf:
I'm struggling a bit to figure out where is the best spot to put the docs: `utils.rst` or `eds.rst`? Also, I'm unsure if the function should be added to `utils/eds.py` or not.

vidartf:
Moved `_signals.eds._parse_only_lines()` to `misc.eds.utils`, as it was independent of any instance, and it was also useful elsewhere. As it is an internal function that is quite new, the move should have no other repercussions at the moment. Also added relevant tests.

pburdet:
I would put the documentation in `eds.rst`. Ithink it's more logic.

I'm all for adding it in `utils.eds`, this is where all utility functions for eds should be.

vidartf:
Docs added. Looks OK?

hyperspy/misc/eds/utils.py
```diff
@@ -13,11 +13,62 @@ def _get_element_and_line(Xray_line):
...
+    only_lines = list(only_lines)
+    for only_line in only_lines:
+        if only_line == 'a':
+            only_lines.extend(['Ka', 'La', 'Ma'])
+        elif only_line == 'b':
+            only_lines.extend(['Kb', 'Lb1', 'Mb'])
+    return only_lines
+
+
+def get_xray_lines_near_energy(energy, spread=0.1, only_lines=None):
```
francisco-dlp: Wouldn't an 'radius' be more explicit and adequate than `spread`?

`width`, `size` or `length` could also be good alternatives, but would require

    E_min, E_max = energy - spread/2., energy + spread/2.

vidartf: Makes sense. I struggled to come up with a good argument name, so good input. I think `width` makes the most sense to me, or maybe even better: `window_width`?

francisco-dlp: `width`, `window_width` or `interval_width` all make sense to me. `width` has the advantage of being shorter.

vidartf: Ended up using `width`, while referring to it as a window width in the docstrings, that was also updated.

vidartf:
PS: I don't know how 996c98d is marked as passing the test before my test fix. If you click on the log, it actually says it has one failure...

PR #548, opened 2015-05-27
ENH: Split off x-ray line plotting

vidartf:
Split the EDSSpectrum.plot, such that the plotting of the xray lines are performed by a separate function. This changes nothing in the default behaviour, but allows for plotting xray lines without creating a fresh plot.


pburdet:
It might be usefull. Thanks

PR #554, opened 2015-06-02
Add pypi badge

francisco-dlp:
Just click in "View" on the README.rst file to see it in action.


francisco-dlp:
Any opinions for or against including the pypi version and downloads badges in the README file?

to266:
I see no reason not to include the information.

As a side note, automatically updating pypi version with new releases would make sense as well (if I recall, at the moment it has to be done by hand).

PR #556, opened 2015-06-02
BUG: Replaced type checks of str with basestring

vidartf:
Several places in code compare type to str with isinstance. In many cases, unicode strings should also be caught by these expressions, so I've replaced str iwth basestring. No rigorous test have been performed, but from the context of the changes, it is not apparent that the test should not include unicode strings (there are some clear examples around, which I excluded). However, it passes the nose tests.


francisco-dlp:
Could you send this PR to 0.8.x instead? Thanks!

vidartf:
Resubmitted as #557

PR #557, opened 2015-06-02
BUG: Replaced type checks to str with basestring

vidartf:
Several places in code compare type to str with isinstance. In many cases, unicode strings should also be caught by these expressions, so I've replaced str iwth basestring. No rigorous test have been performed, but from the context of the changes, it is not apparent that the test should not include unicode strings (there are some clear examples around, which I excluded). However, it passes the nose tests.

### DIFF
--- a/hyperspy/_signals/eds.py
+++ b/hyperspy/_signals/eds.py
@@ -298,9 +298,9 @@ class EDSSpectrum(Spectrum):
 
     def _parse_only_lines(self, only_lines):
         if hasattr(only_lines, '__iter__'):
-            if isinstance(only_lines[0], str) is False:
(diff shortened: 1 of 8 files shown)


hyperspy/Release.py
```diff
@@ -26,7 +26,7 @@
 # When running setup.py the "+dev" string will be replaced (if possible)
 # by the output of "git describe" if git is available or the git
 # hash if .git is present.
-version = "0.8.x+dev"
```
francisco-dlp: Why are you changing the version string in this PR?

vidartf:
Forgot to rebase. Give me a minute.

PR #558, opened 2015-06-02
Update master version.

francisco-dlp:
Currently the master version is `0.8+dev` and the maintenance branch `0.8.x+dev`. I suggest to change the convention to the more standard: `0.9.dev` and `0.8.x`. What do you think? (This PR does just that.)


francisco-dlp:
Any comments on this?

to266:
I agree, makes more sense

PR #563, opened 2015-06-04
New: find_peaks2D

dnjohnstone:
A "working" (at least for my data which has a rather high signal : noise) though ugly hack at getting out a structured array of peak positions from a stack of images - see attached screen grabs...
image_peaks
peak_positions

So far - two very basic peak finding approaches are implemented along with a little code for generating the structured array and  allowing the choice between approaches. 

The two methods output the peak positions in different formats which is obviously not satisfactory in the long run. At this stage though I'm unsure as to what structure will be the most generally useful output for those positions. 

I'd appreciate comments on:

1. What the preferred structure for peak position output is in general applications?
2. Any neat ideas or implementations for good peak finders that will be more suited to more general data that will most likely be much more noisy that mine?


vidartf:
Neat! Regarding your question no 1, my initial thought would be a list of indices compatible to what you would get with `AxesManager`. I.e. for a signal with nav dim of 2 `[(<x point 1>, <y point 1), (<x point 2>, <y point 2>), etc.]` (or of course numpy equivalents). You could then use them to navigate a signal, or slice with `signal.inav` (`isig`?) as well.

vidartf:
PS: The best way to plot them might be using markers (as they are navigation friendly), in which case it would be nice if the output was compatible with that (pass in x and y coords as separate arrays I think).

magnunor:
Very nice!

I got it working with the "basic" method on some of my SrTiO3 datasets:
sto_adf_basic_method

This could be very useful in conjunction with the https://github.com/hyperspy/hyperspy/pull/536 for getting very precise atomic positions. Which I guess is your final goal, @dnjohnstone.

Some comments: 
The skimage included in Ubuntu 14.10 did not work, due the `resize(preserve_range)` argument not being implemented there.
The threshold method did not return any peaks at all, but I'm guessing I need to play around with the various arguments in `s.find_peaks2D`.

dnjohnstone:
Yes, the plan is to find the peaks using this method and then refine the positions using #563, in principle that shouldn't be too hard I just need to find a minute to do it.

The threshold method has a bit of a problem in terms of setting a good default for the threshold value, it's quite possible that everything in your data could have been below the threshold that I set.

In my case another thing to do is to use the peak positions to fit a lattice to the data - I guess this may be also useful for high-resolution images. I've made some steps in that direction but I think the best way might be to treat the lattice as a model, guess its initial basis vectors and then find the optimal transformation to bring the lattice points into best agreement with the peak positions.

magnunor:
One way to set the threshold automatically could be using a histogram to find the most common intensities.

How do you plan on fitting many atoms with Gaussians? I'm guessing that would be quite slow fitting them all at the same time. One way could be making Voronoi diagrams and just fitting one atom at a time, having the signal range within the Voronoi diagram.

dnjohnstone:
Decided in conversation with @francisco-dlp that this is deprecated by #606 as the proposed method specified by find_peaks2D is really just a way of mapping the peak_local_max method of skimage on a multi-dimensional dataset.

Issue #541, opened 2015-05-07
hyperspy starting script doesn't work with IPython 3 notebook


francisco-dlp:
The main purpose of the `hyperspy` script is to setup IPython and traitsui to run properly with the chosen toolkit. It relies on the `--pylab` flag that was available for all IPython interfaces until IPython 3.0. However, the `--pylab` flag has been removed for the notebook in IPython 3.0 and, although it's still available for the other interfaces, [its usage is discouraged](http://nbviewer.ipython.org/github/Carreau/carreau.github.io/blob/master/posts/10-No-PyLab-Thanks.ipynb).

Currently this is what happens when executing `hyperspy notebook` if IPython > 3 is installed:

```
$ ana-hyperspy notebook
[E 16:07:15.695 NotebookApp] Support for specifying --pylab on the command line has been removed.
[E 16:07:15.695 NotebookApp] Please use `%pylab qt` or `%matplotlib qt` in the notebook itself.

```

Worst, when starting the "HyperSpy Notebook" entry in the Window file explorer context menu, the notebook doesn't start but there is no warning.

We could fix the issue by setting up the environment in an IPython profile, however I think that the `--pylab` flag is gone for good reasons and for the same reasons we should consider not providing the feature anymore when using the IPython notebook. With this is mind, this is the "fix" that I propose:

* Replace the "Start HyperSpy Notebook here" windows by "Start IPython Notebook here" and add "Start IPython QtConsole here" without replacing "Start HyperSpy QtConsole here".
* When using IPython > 3, simply start the IPython notebook without any flag and print a warning explaining the "hyperspy notebook" will be deprecated in the next release.
* Modify the documentation so that all the imports are explicit e.g.:
 ```python
 import numpy as np
 import matplotlib.pyplot as plt
 import hyperspy.hspy as hs

 ```

What do you think?

to266:
I agree.

As a minor thing, many warnings should probably be given for the users less familiar with python and/or hyperspy structure - by that I mean that with your given explicit imports now suddenly they will have to rewrite all their scripts to use the hyperspy and/or will not know how to create signals (i.e. now suddenly `signals.Spectrum()` does not work, but you have to write `hs.signals.Spectrum()` instead)

Of course the alternative is to do what we did all the time (i.e. `from hyperspy.hspy import *` ), however I am aware that is very much so discouraged. 

Nevertheless, I think the change will be for the better and quite like your 'fix'

pburdet:
[...]

PR #564, opened 2015-06-04
Enh ipython3

to266:
The changes discussed in https://github.com/hyperspy/hyperspy/issues/541 

This changes the profiles, so even with old ipython versions the hyperspy has to be used explicitly (i.e. `hs.load()`, etc.).
Should we separate these? With ipython<3.0 we could keep the old behaviour still. I don't think I am a fan of it, but should be relatively easy to do by just having two copies of the profiles..

Let me know if there are any features you think are reasonable to have by default (in #541 people seem to want to change figuresize... however I never needed / used this function)

TODO
--------
* [x] add `-r` flag to replace the cell
* [x] Document the hyperspy magic in the User Guide.
* [x] Use new syntax in:
    * [x] Docstrings
    * [x] User guide
    * [x] Examples


bin/hyperspy
```diff
@@ -145,7 +145,13 @@ for f in to_copy:
             destination, os.path.split(f)[-1])) or \
             args.overwrite_profile:
         shutil.copy(f, destination)
-
```
francisco-dlp: I think that the `startup` folder feature was introduced in IPython 1.0, couldn't we then use this nicer approach with IPython >= 1.0?

bin/hyperspy
```diff
@@ -208,4 +214,10 @@ else:
         sys.argv.append('--pylab=inline')
     elif add_pylab is True:
         sys.argv.append('--pylab=%s' % toolkit.replace("qt4", "qt"))
+
+    if ipy_version >= LooseVersion('3.0.0') and args.environment == "notebook":
```
francisco-dlp: This will launch the notebook with everything but matplotlib correctly loaded which may not be what we want given that we seem to agree in that in the notebook all the imports should be explicit. Wouldn't it be better not to copy the `startup` folder if IPython >= 3.0 and to actually delete `default_imports.py` if present (e.g. the user might have upgraded)? Also I would suggest to rename `default_imports.py` to `default_hspy_imports.py`. In this case, I think that the users would also appreciate to get a warning in the notebook explaining that they have to import manually.

to266:
I think this new way should be better - I added a (cube-like) button in the notebook, that, when pressed, explicitly adds the hyperspy imports in a new cell, created above the current one, and executes it.

I think it's both convenient and explicit..

to266:
the last couple of commits introduce hyperspy magic
```python
%hyperspy [toolkit]
```

That, when executed, replaces the cell with the following:
```python
# %hyperspy {toolkit}
import numpy as np
import hyperspy.hspy as hs
%matplotlib {toolkit}
```
and effectively executes it (inside the magic, it's not quite as obvious, nevertheless equivalent)

Here `[toolkit]` is optional user input, and `{toolkit}` is either [toolkit], or the default from parameters (note that if [toolkit] is not supplied, the replaced text actually includes the default)

to266:
as per https://stackoverflow.com/questions/30705358/ipython-magic-that-replaces-some-text-and-runs-the-cell-afterwards, it seems there is no way to tell the ipython kernel to execute a frontent cell.. So I guess the current version is the best in terms of being explicit and convenient enough (I've used it multiple times now in console environment, and it was much less annoying than I expected at first)

It still replaces the other content in the cell, though..

hyperspy/ipython_profile/hyperspy_magics.py
```diff
@@ -0,0 +1,30 @@
...
+        else:
+            toolkit = line
+
+        sh.enable_matplotlib(toolkit)
+
+        second_import_part = "import matplotlib.pyplot as plt\n"
+        exec(second_import_part, sh.user_ns)
+
+        sh.set_next_input("# %hyperspy " + toolkit + "\n" + first_import_part +
+                          "%matplitlib " + toolkit + "\n" + second_import_part,
```
jat255: There's a typo here. `matplitlib` should be `matplotlib`

to266: Ouch, sorry for that - I forgot to push the commit. 

Let me know if the usage is annoying for you (as far as I know, I'm the only one using the magic + button combo). 

The button and magic work for the notebooks, however you have to be careful not to use magic with other code in the cell - it just replaces it. You can, of course, undo it (CTRL+z), but that's not very convenient.

The magic also works in consoles, but, as already mentioned in previous comments, leaves the "replaced" code in the not-run state... Which (at least for me) requires pressing return twice more. At the very least it's very verbose :)

jat255:
I've been using this a little bit in a separate winpython installation for testing ipython3. It seems to be working okay. Having to change old code to include the `hs.` prefix is a bit of a hassle, but I suppose it is worth it to have everything done explicitly like was discussed.

As an alternative to the magic command, is there some way with an ipython profile to pre-populate a cell when creating a new notebook or console session? That would be very explicit to the user as "these are the commands you have to run first in order to use hyperspy." I'm not sure if this is desirable (or even possible), but I thought I could add it as a suggestion.

pburdet:
One way to pre-populate a cell is to change the ipython profile. Along the doc of the magic, we could explain how to do it, for user that are frequent user

to266:
@jat255 , I think there is a way to create notebook templates, and then when creating a new one, you can select it.. I haven't tried it out myself, but @francisco-dlp apparently used it for some time. I just find the button more convenient - just a single click, and you can insert the standardized imports to old notebooks as well.

As for the conversion, I think a tool to do that might be in order... I'll try to cook something up today when I find the time. In the worst case, you can manually practice the "bad old" way :)

to266:
@jat255 , this seems to work for me. I haven't tested it (almost at all), but it should be very easy to fix if anything's amiss 

```python
filename = 'something.ipynb'
import json
with open(filename, 'r') as f:
    dictionary = json.load(f)
    
for cell in dictionary[u'worksheets'][0][u'cells']:
    if cell[u'cell_type'] == u'code':
        for num, line in enumerate(cell[u'input']):
            new_line = line.replace('signals', 'hs.signals')
            new_line = new_line.replace('create_model', 'hs.create_model')
            new_line = new_line.replace('components', 'hs.components')
            new_line = new_line.replace('load', 'hs.load').replace('np.hs.load', 'np.load')
            new_line = new_line.replace('utils', 'hs.utils')
            cell[u'input'][num] = new_line
with open(filename, 'w') as f:
    json.dump(dictionary, f)
```

to266:
I'd like to ask peoples opinion:

Which one would you rather have:
 - a magic that injects user name space and __prints__ exactly what it does to the "output"/ something similar (i.e. somewhat more explicit that the `%pylab`, as it only warns about potential errors)
 - a magic that injects user name space  __replaces the cell with explicit code of what it did__, so that next time you don't have to run the magic at all, and rather execute the explicit code (i.e. what the current implementation of hyperspy magic does)

@francisco-dlp , @vidartf , @jat255 , @pburdet are all very welcome to express their opinions on the matter :) 

If you can come up with any unforeseen difficulties experienced and new users might have with either approach, please let me know.

At the moment the main drawback of the "replace" approach is that when run in a console, you have to press "enter" multiple times (that can be reduced to just a single additional button press, but still not obvious). It also tries to populate the namespace twice (once when the magic is run, and the second time when the replaced text is run).

Finally, with the "replace" approach, if you run the magic not in an empty cell, it overwrites all its content. 


The "print" approach is just less explicit, and if the output is suppressed/cleared, the user does not know what happened.

francisco-dlp:
I'm +0.000001 for the print solution.

By the way, @magnunor and of course anybody else are more than welcome to join the discussion.

jat255:
I would say that I prefer the print solution as well, as long as it is running the commands behind the scene. At the same time, the idea of running things without explicit "permission" (or notification) could potentially confuse new users. When first starting (and honestly, still, to some degree), it was unclear to me what exactly is loaded when you start `hyperspy notebook` vs `hyperspy qtconsole`, etc. 

I think the replacing the current cell problem is probably a fringe case, as most people will be running it the first thing in their session. But we could solve this by enforcing the loading of hyperspy (sort of like how it is done now). Is it possible to auto-populate or run a cell when first creating a notebook or console session? That way you could set things up easily for the user, and still let them know what happened behind the scenes by printing some output like (for example...):

```
This is HyperSpy v. XX on iPython XXXX
The following commands have already been run, 
and Hyperspy has been loaded into the session:

    %matplotlib qt4
    import hyperspy.hspy as hs
    ...

Hyperspy command can be accessed through the hs module
``` 

This would also prevent the user from having to remember "oh yeah, I need to type the magic in before I can use anything". I'm not familiar enough with ipython to know what's possible with the profiles, but I might propose this as another option.

jat255:
Also, I don't know if it's possible, but would it be difficult/possible to use a different profile name for ipython3? I have been switching back and forth between them for testing, and even though I'm using a different Python environment, they use the same profile, so when I use ipython3, it breaks some extensions and custom stuff I have in my `profile-hyperspy` directory. Currently, I've just been backing up my profile before launching a new ipython3 notebook session, but having a unique profile name (like `profile-hyperspy_ip3` or something) could perhaps ease a number of "growing pains" as users upgrade. Thoughts?

jat255:
> @jat255 , this seems to work for me. I haven't tested it (almost at all), but it should be very easy to fix if anything's amiss

@to266 Sorry that I haven't had a chance to test this yet, but I'll get to it in the next few days!

pburdet:
Print solution sounds good. It would be a plus if you can copy/paste it in another cell and obtain the same result.

francisco-dlp:
Any more opinions on this? @magnunor? @vidartf? Users?

francisco-dlp:
What do you think about having a flags and/or argument to import everything into the user namespace in order? e.g. `hyperspy qtconsole -a` imports all. This may ease the transition for users with a lot of scripts using the—now discouraged—`from hyperspy.hspy import *`.

vidartf:
I don't use notebook, so I don't think I can contribute constructively there. For console, printing something as suggested as the banner sounds reasonable to me.

to266:
Maybe we want to change hyperspy to an extension? I haven't looked at it in detail, but seems to be similar to what we want in the end:
http://ipython.org/ipython-doc/stable/config/extensions/index.html#writing-extensions

francisco-dlp:
@to266, if I understand well your point, the only way to make this feature work in any IPython profile is to turn it into an extension? If that's what you mean, then I'm -1 on this.

Then, if we need to use the hyperspy profile in order to get the magic, I change my vote on the previous poll: for the sake of reproducibility the magic must replace the content of the cell with the code i.e. just printing is not enough. In this way a notebook that uses that magic when running using the hyperspy IPython probably can be run without modification when running on any other profile. Does this make sense or did I get everything wrong?

to266:
No, you misunderstood me. At the moment the magic just prints a greeting and what exactly it imported. I suggested looking into extensions in order to get rid of the ipython profile for hyperspy. 

For the magics, you need some way to define (or import) them first, and then run the following:

```python
# In order to actually use these magics, you must register them with a
# running IPython.  This code must be placed in a file that is loaded once
# IPython is up and running:
ip = get_ipython()
# You can register the class itself without instantiating it.  IPython will
# call the default constructor on it.
ip.register_magics(MyMagics)
```

As a side note, it looks like travis here has a billion of errors, but shows the build as passing, and as a result the coveralls fail, since a big part of the tests did not run at all.

vidartf:
> As a side note, it looks like travis here has a billion of errors

A lot of these errors are about `scipy.lib.blas`. As far as I know this is deprecated, and `scipy.linalg.blas` should be used instead (found it out thanks to playing around with warning filters).

to266:
Having tried using the magic as a print solution, I am in favour of it. I suppose it's ready to merge from my side.

to266:
@francisco-dlp , please have a look, should be ready for review.

hyperspy/ipython_profile/hyperspy_magics.py
```diff
@@ -0,0 +1,57 @@
+from IPython.core.magic import Magics, magics_class, line_magic
+
+
+@magics_class
+class HyperspyMagics(Magics):
+
+    @line_magic
+    def hyperspy(self, line):
```
francisco-dlp: The magic does not have docstring.

to266: added

bin/hyperspy
```diff
@@ -145,6 +145,13 @@ for f in to_copy:
             destination, os.path.split(f)[-1])) or \
             args.overwrite_profile:
         shutil.copy(f, destination)
+if ipy_version >= LooseVersion('3.0.0'):
+    additional_dest = os.path.join(destination, 'startup')
+    if not os.path.isdir(additional_dest):
+        os.makedirs(additional_dest)
+    if not os.path.isfile(os.path.join(additional_dest, 'hyperspy_magics.py')) or \
```
francisco-dlp: If in  a new release we change `hyperspy_magics.py`, the copy in the user's profile won't be updated unless he asks to overwrite the profile. Couldn't the content for `hyperspy_magics.py` be actually imported from the `hyperspy` module itself in order to workaround this?

to266: Good point, done

to266:
Should be ready for merge. The last couple of commits:

 - dropped support for < IPython 2.0
 - removed unnecessary files from hyperspy profile
 - script deletes the unnecessary files if present
 - the magic is defined not in the profile, but hyperspy itself, hence simple to maintain (no need to overwrite files / check for changes)
 - the hyperspy (binary) and the main ipython config files now have versions. If the config version is lower than the binary, ipython profile will be overwritten.

Issue #494, opened 2015-03-31
Making the most of auto-completion


pburdet:
(Original proposal by @pburdet is just below)

This is being implemented in #566, see #566 for an up-to-date progress status and final implementation details.

pburdet:
#1 moving to utils

To make the most of auto completion and to helps user to find/remember the function of hyperspy, I suggest to move a maximum of method under utils.

- preferences -> utils.preferences
- __version__ -> utils.__version__
- get_configuration_directory_path -> utils.get_configuration_directory_path
- model
   - components -> utils.model.components
   - create_model -> utils.model.model
- load -> utils.load (I'm not sure about this one)
- signals -> utils.signals (I'm not sure about this one)

#2 Simplifying utils

Mainly for a nicer looking

- utils.plot.plot_spectra -> utils.plot.spectra
- utils.plot.plot_images -> utils.plot.images
- utils.plot.plot_signals -> utils.plot.signals
- utils.plot.plot_histograms -> utils.plot.histograms

#3 Package other than hyperspy

For the user to be aware when he is using hyperspy and to help him to understand python, I suggest to stop importing the full package. I think it's a best practice and it will help attracting people from the python community

- from numpy import * -> numpy as np
- from matplotlib.pyplot import * -> import matplotlib.pyplot as plt

to266:
Moving things around too much would break a lot! Or we'd have to write many new functions for backwards - compatibility. One way to mitigate it would be to only have "links" in the utils (with some sort of explanation).. However most of the functions are not utils in the technical sense.

Maybe then some sort of `utils.helper`? It could be either links to the actual methods / objects, etc, or just functions printing out documentation of the function/structure, etc. Either way the auto-completion should be used for the helper.

--------------

I'm all for the simplification, I think it makes sense


-----------------

That would make a lot of sense too, especially considering that is the direction the IPython is going forward to and encouraging. However, I would still like to not have to remember to include all the things myself! So either (somehow) adding the include lines at the top when creating a new notebook or (in case of console/QTconsole) just printing what was done would be useful.

On the other hand, then there is less sense in providing the "bundle" - the hyperspy is then used exactly as any other imported package..

francisco-dlp:
I fully agree with @pburdet, we have to undertake a major reorganization of the placement of the public interface. As it seems that we all agree in that we should stop injection all `hspy` into the user namespace, then, most of @pburdet suggestions are already in-place:


* [x] `preferences` -> `hs.preferences`
[...]

PR #566, opened 2015-06-05
New api module set to replace the hspy module

francisco-dlp:
This PR intends to implement the changes proposed in #494. It is still work in progress but ready for discussion.

TODO
--------

* [ ] Promote the usage of `import hyperspy.api as hs` in
    * [ ] Docstrings
    * [ ] User Guide
    * [ ] Demos
* Rename/move:
    * [x] `hspy.components` -> `api.model_components` 
    * [x] `hspy.utils.*`-> `api.*`
    * [ ] `hspy.utils.markers` -> `api.plot.markers`
    * [ ] Write automatic translator to run on the remaining code, User Guide, docstrings and demos and make it available to the users.

### DIFF
--- a/hyperspy/__init__.py
+++ b/hyperspy/__init__.py
@@ -1,36 +1,23 @@
-"""
+# -*- coding: utf-8 -*-
+
+from hyperspy import docstrings
+
+__doc__ = """
 HyperSpy: a multi-dimensional data analysis package for Python
 ==============================================================
 
 Documentation is available in the docstrings and online at
 http://hyperspy.org/hyperspy-doc/current/index.html.
 
-All public packages, functions and classes are in :mod:`~hyperspy.hspy`. All
-other packages are for internal consumption.
+All public packages, functions and classes are in :mod:`~hyperspy.api`. All
+other packages and modules are for internal consumption and should not be
+needed for data analysis.
 
-When starting HyperSpy using the starting script e.g. by typing ``hyperspy`` in
-a console, using the context menu entries or using the links in the
-``Start Menu``, the :mod:`~hyperspy.hspy` package is imported in the user
-namespace. When using HyperSpy as a library, it is reccommended to import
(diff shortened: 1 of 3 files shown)


francisco-dlp:
Note that I have just added a TODO list which is slightly different to the one found in #494.

hyperspy/__init__.py
```diff
@@ -39,7 +26,6 @@
 
 import Release
 
+__all__ = ["api"]
```
vidartf: I'm unsure whether this is necessary or not. Don't we risk hiding functionality for those that want to explore? What are the pros/cons?

francisco-dlp: Maybe, but this is actually an improvement over the current situation in which "from hyperspy import *" only imports the modules explicitly imported in `hyperspy/__init__.py` i.e. `Release`.

Actually, given that the API will be in `hyperspy.api` it doesn't make much sense to use "from hyperspy import *", so one wonders indeed if the current empty `__all__` was actually better. (Any opinions on this?)

Regarding hiding functionality, if someone wants to access the internal modules they can simply import them explicitly but the point is that there should be no need to import anything outside `hyperspy.api` for data analysis.

vidartf: The explanation you give makes sense, and I see why you'd want to add `api`. My only question is if this will hide the other packages from autocomplete in editors/consoles. If not, then I have no issue, if they do, it will be an inconvenience.

francisco-dlp: Sorry for misunderstanding your question. IPython's autocompletion works fine, this should only impact what is imported by `from hyperspy import *`.

to266:
Either the milestone is wrong, or it's PRd to the wrong branch. @francisco-dlp ?

francisco-dlp:
I have pull requested it to the wrong branch. I have also branched it from master instead of 0.8.x but this happened before they've diverged. I close this PR and send a new one.

PR #567, opened 2015-06-05
Bugfix signal rebin variance

to266:
The variance was not binned when binning the signal, and now is. The rebinning now works as expected


francisco-dlp:
Good! Could you add a little description of the bug in header before merging it?

to266:
Done

PR #573, opened 2015-06-12
New `Expression` component

### LINKED ISSUE #488: Speed up fitting in HyperSpy
Fitting in HyperSpy is slow, or at least it is not as fast as it could be. For multidimensional datasets parallesation can help in some case and is in the works (see #242). Another complementary approach is to speed up function evaluation. One way to do this would be to use [numexpr](https://github.com/pydata/numexpr). Any other ideas?

francisco-dlp:
New `Expression` component that takes a expression as a string and creates a HyperSpy component from it. The magic is done by SymPy. As a bonus it can generate the function using numexpr which should boost performance in some cases.

Example of usage:

```python

import hyperspy.hspy as hs
import numpy as np
%matplotlib inline

g_np = hs.components.Expression(
    expression="height * exp(-(x - centre) ** 2 * 4 * log(2)/ fwhm ** 2)",
    name="Gaussian",
    position="centre",
    height=1,
    fwhm=1,
    centre=0,
    module="numexpr")


x = np.linspace(-2,2,1000)
plt.plot(x, g_np.function(x), label="f")
plt.plot(x, g_np.grad_centre(x),
         label=r"$\frac{\delta f}{\delta \mathrm{centre}}$")
plt.plot(x, g_np.grad_height(x), "--", linewidth=2,
         label=r"$\frac{\delta f}{\delta \mathrm{height}}$")
plt.plot(x, g_np.grad_fwhm(x),
         label=r"$\frac{\delta f}{\delta \mathrm{FWHM}}$")
plt.legend()

```
example_expression

In my computer the gaussian component generated as above is twice as fast as the one that HyperSpy ships, so I think that this Fixes #488 .

TODO
--------

* [x] Test
* [x] Docs 
* [ ] Make it work with non-differentiable everywhere functions
* [x] Solve security concerns raised by @stefano-m (see discussion below)


francisco-dlp:
I have tentatively added this to the 0.8.1 milestone because, although it is a new feature, being just a new component it most likely doesn't break anything. However this is open for debate.

hyperspy/_components/expression.py
```diff
@@ -0,0 +1,126 @@
...
+        # Extract x
+        x, = [symbol for symbol in expr.free_symbols if symbol.name == "x"]
+        # Create compiled function
+        self._f = lambdify([x] + parameters, eval_expr,
+                           modules=module, dummify=False)
+        parnames = [symbol.name for symbol in parameters]
+        eval_str = "self._f(x, %s)" % ", ".join(
+            ["self.%s.value" % par for par in parnames])
+        # Generate string to be evaluated by self.function
+        self._f_eval_str = compiler.compile(eval_str, '<string>', 'eval')
```
stefano-m: Isn't using `eval` a bit unsafe? It can be leveraged to run arbitrary code.

For example, see [this stackoverflow answer](https://stackoverflow.com/questions/3949727/why-i-can-call-print-from-eval/3949812#3949812).

```python
src = """
import os
nasty_command = "echo I WILL DESTROY YOUR HARD DRIVE MWHAHAHA"
os.system(nasty_command)
"""

obj = compile(src, '', 'exec')
eval(obj, {'__builtins__': {'__import__':__builtins__.__import__}})
```

Also, the `compiler` package has been removed in Python3 (the builtin `compile` functions is still there though).

francisco-dlp: Two good points, thanks! I've fixed the compiler bit.

Regarding `eval`, I  agree that it can be dangerous. I would be worried if we were saving the strings that we pass to `eval` in the hdf5 file, in which case it would indeed be possible to do something nasty just by loading the file using HyperSpy. However, only the expression in string format as passed to the constructor will be stored in the hdf5 file, so one may think that there is not much room for evilness here (at least there is no increased risk over running any other python script). However, one will be very wrong! `sympy.sympify` uses `eval` internally (!) I don't see any way out. We could require `expression` to be SymPy expression instead of a string, but we cannot save SymPy expressions in hdf5, so we have to turn it into a string at some point... An inconvenient workaround could be to print the string before executing sympyfy and ask the user to confirm the action. What do you think?

francisco-dlp: Your previous comment and my reply seem to have vanished, any clue of
where they are? Did you get my reply to your first post?

Francisco

stefano-m: I can see the comments now. But I did not notice them when I posted my first comment on your latest commit 46032d9

stefano-m: If the expressions have some predictable form, the strings could be verified or sanitized.
 
Another possibility would be to serialize the SymPy expression and save it in hdf5 as bytes. Of course, there's nothing to be done about SymPy using `eval` internally but talking to the devs.

As far as the input consumed by this code is not provided by external users (e.g. via a web service), `eval` should be fine. Still, one could craft a malicious hdf5 file and trick a user into opening it...

It all boils down to how much risky all this stuff is and how much effort must be taken to avoid it.

to266:
I PR-ed an eval-free implementation of the component, with somewhat more thorough tests.
As previously mentioned, the `eval` insecurity from the `sympy` side is still there, however when looking at the source code, I could not find it. So either the warning is out-dated, I didn't look hard enough, or it's just for the precautionary reasons...

doc/user_guide/model.rst
```diff
@@ -54,11 +54,99 @@ These are some of the components which are currently available:
...
-component, it should be easy to write your own component. If you need help for
+However, this doesn't mean that you have to limit yourself to this meagre list of function.
+
+
+
+.. versionadded:: 0.8.1
+
+The easiest way to turn a mathematical expression into a component is using the
+:py:class:`~._components.expression.Expression` component. For example, the
+following is all you need to create a`Gaussian` component  with more sensible
```
to266: a space missing before `Gaussian`

hyperspy/_components/expression.py
```diff
@@ -0,0 +1,136 @@
...
+            Name of the component.
+        position: str, optional
+            The parameter name that defines the position of the component if
+            applicable. It enables adjusting the position of the component
+            interactively in a model.
+        module: {"numpy", "numexpr"}, default "numpy"
+            Module used to evaluate the function. numexpr is often faster but
+            it supports less functions.
+
+        **kwarfs
```
to266: kwargs

francisco-dlp:
It should be fine now, thanks @to266

vidartf:
This did not work (didn't pass tests) with sympy 0.7.4.1 installed. Upgraded to 0.7.6 and everything was OK. Consider updating package requirements!

PR #577, opened 2015-06-25
BUG: Fix for #562

### LINKED ISSUE #562: Auto-scaling issue when interactively removing EELS background
When trying to remove the background with the GUI, the scale of the plot will sometimes change drastically such that I will not be able to see the data at all. I can either close and restart the process, or drag the data into view with the window controls. I've figured out that this behavior is dependent on where you click in the spectrum. If you click right on the line, it will work fine. If you click far from the line (vertically), then the scale will get all out of sorts. See this video for example, where the first time I click near the line, and then the next time I click far from it:

remove_background_scale

vidartf:
Fixes #562.

### DIFF
--- a/hyperspy/gui/egerton_quantification.py
+++ b/hyperspy/gui/egerton_quantification.py
@@ -104,8 +104,8 @@ class BackgroundRemoval(SpanSelectorInSpectrum):
         self.bg_line.set_line_properties(
             color='blue',
             type='line')
-        self.signal._plot.signal_plot.add_line(self.bg_line)
         self.bg_line.autoscale = False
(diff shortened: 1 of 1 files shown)


vidartf:
Ah, I had a false positive since I tested it on a branch that had some other changes to the plotting code. Your change makes sense if `autoscale` is only supposed to be applied on updates. If not, I'll change it so that `autoscale` is applied during `SpectrumLine.plot()`, i.e. basically applying `scaley=False` there instead.

francisco-dlp:
I agree with your interpretation of the way `autocale` and `scaley` works.

PR #579, opened 2015-07-03
Enh hist tools numerical optimisation

to266:
numerically identical, but each bit runs at ~ 0.3 - 0.7 of the previous time. Overall probably negligible


francisco-dlp:
Good! This is not our code though, we're borrowing it from AstroML so it'll be nice to submit your changes there too.

Issue #527, opened 2015-04-23
BUG: trait / type errors when iterating in spikes_removal_tool


to266:
With a particular type of files (I haven't found what's so peculiar about them and managed to reproduce with a synthetic example yet) `spikes_removal_tool` (the GUI version) outputs one exception per iteration (i.e. per navigation pixel):

```python
ERROR:traits:Exception occurred in traits notification handler for object: y axis, trait: index, old value: 49, new value: 0
Traceback (most recent call last):
  File "/home/to266/anaconda/lib/python2.7/site-packages/traits/trait_notifiers.py", line 520, in _dispatch_change_event
    self.dispatch( handler, *args )
  File "/home/to266/anaconda/lib/python2.7/site-packages/traits/trait_notifiers.py", line 483, in dispatch
    handler( *args )
  File "/home/to266/hyperspy/hyperspy/drawing/spectrum.py", line 146, in update
    line.update()
  File "/home/to266/hyperspy/hyperspy/drawing/spectrum.py", line 312, in update
    ydata = self.data_function(axes_manager=self.axes_manager).real
  File "/home/to266/hyperspy/hyperspy/gui/egerton_quantification.py", line 407, in get_interpolated_spectrum
    left, right = self.get_interpolation_range()
  File "/home/to266/hyperspy/hyperspy/gui/egerton_quantification.py", line 391, in get_interpolation_range
    left = self.argmax - self.default_spike_width
TypeError: unsupported operand type(s) for -: 'NoneType' and 'int'
```

Strangely the function actually does its job, the only problem is then the actual error reports, that (if run in a notebook) just jam the pipeline and crashe both the browser (chromium / firefox in my case) and the ipython kernel. It takes only about 30seconds when run in a qtconsole (as the output just scrolls and is probably stored differently than the notebook), so for now I use that.

vidartf:
The two PRs that together were hoped to fix this have been merged. Is this still an issue?

francisco-dlp:
AFAIK yes, thanks for spotting it.

PR #423, opened 2015-01-14
General events implementation

francisco-dlp:
At present some HyperSpy objects have a very primitive events implementation that is inconsistent and limiting. Given the increasing interest in implementing interactive features in HyperSpy, the time has come to get a more general event system for it.

I have considered using a third party solution such as pydispatcher. However, because HyperSpy's main interface is IPython, I think that it'll better to implement our own IPython friendly events. Any opinions on this?

Usage
--------

```python

>>> from hyperspy.events import Event, Events
>>> events = Events()
>>> events.speak = Event()
>>> def say_hi(n):
...         print(" Hi." * n)
...     
>>> def say_bye(n):
...         print(" Bye." * n)
...     
>>> events.speak.connect(say_hi)
>>> events.speak.connect(say_bye)
>>> events.speak.connected
set([<function say_hi at 0x7f062c065758>, <function say_bye at 0x7f062c0656e0>])
>>> events.speak.trigger(3)
 Hi. Hi. Hi.
 Bye. Bye. Bye.
>>> events.speak.disconnect(say_hi)
>>> events.speak.trigger(1)
 Bye.

```

TODO
--------
- [x] unittests
- [x] docstrings
- [x] User guide
- [x] Reimplement old events using the new infraestructure.
- [x] Implement more explicit connection syntax:
```python
index_changed.connect(f, ["axis"]) # pass just the axis kwarg
index_changed.connect(f) # pass all events
index_changed.connect(f, {"axis" : "ax", "index": "idx"}) # Map the event kwargs to match those in the function
index_changed.connect(f, []) # don't pass any event
index_changed.connect(f, "auto")

```

```

PR #580, opened 2015-07-08
BUG_FIX: closing a SpectrumLine was not removing it from SpectrumFigure

francisco-dlp:
This fixes half of #527 and possibly other unreported issues.

### DIFF
--- a/hyperspy/drawing/spectrum.py
+++ b/hyperspy/drawing/spectrum.py
@@ -84,7 +84,7 @@ class SpectrumFigure(BlittedFigure):
         elif ax == 'right':
             line.ax = self.right_ax
             self.right_ax_lines.append(line)
-            line.sf_lines = self.ax_lines
+            line.sf_lines = self.right_ax_lines
             if line.axes_manager is None:
                 line.axes_manager = self.right_axes_manager
         line.axis = self.axis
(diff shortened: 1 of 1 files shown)


hyperspy/drawing/spectrum.py
```diff
@@ -339,6 +341,8 @@ def close(self):
         if self.text and self.text in self.ax.texts:
             self.ax.texts.remove(self.text)
         self.axes_manager.disconnect(self.update)
+        if self.spectrum_figure:
+            self.spectrum_figure.ax_lines.remove(self)
```
vidartf: Do we need to check/remove from `self.right_ax_line` as well?

francisco-dlp: Thank you for reporting the issue. I have just committed a fix.

I know that the fix is ugly. Hopefully we'll refactor the drawing code once #423 is merged.

hyperspy/drawing/spectrum.py
```diff
@@ -81,9 +80,11 @@ def add_line(self, line, ax='left'):
             if line.axes_manager is None:
                 line.axes_manager = self.axes_manager
             self.ax_lines.append(line)
+            line.sf_lines = self.ax_lines
         elif ax == 'right':
             line.ax = self.right_ax
             self.right_ax_lines.append(line)
+            line.sf_lines = self.ax_lines
```
vidartf: This should be `line.sf_lines = self.right_ax_lines`? :)

francisco-dlp:
With a bit of luck this should be ready now, is it @vidartf?

vidartf:
@francisco-dlp All good :)

PR #581, opened 2015-07-08
Fix out of range error in `SpanSelectorInSpectrum`

francisco-dlp:
The problem was that expanding the slider triggered the code to plot the
interpolated line before the right edge of the slider was defined.

This hopefully fixes the other half of #527.

TODO
--------

* [x] Check that this doesn't break other parts of the code that inherits from SpanSelectorInSpectrum


hyperspy/gui/egerton_quantification.py
```diff
@@ -369,10 +369,12 @@ def _interpolator_kind_changed(self, old, new):
         self.span_selector_changed()
 
     def _ss_left_value_changed(self, old, new):
-        self.span_selector_changed()
+        if not np.isnan(self.ss_right_value):
+            self.span_selector_changed()
 
     def _ss_right_value_changed(self, old, new):
-        self.span_selector_changed()
+        if not np.isnan(self.ss_right_value):
```
vidartf: Should this check `self.ss_left_value`? If not, maybe it should check `new` instead?

francisco-dlp: I think that this is correct: the idea is that, when just starting dragging the span selector, `ss_left_value` and `ss_right_value` will be updated with the same value (width=0) and we want to avoid triggering `span_selector_changed` in this situation. (This situation is what was causing the bug the this PR intends to fix). Do you disagree? If not, I'll add comments to explain this.

francisco-dlp: Sorry, I wrote too fast: I'll add comments to the code if you agree, otherwise I can fix it if you think that this is broken.

vidartf: Ah, so this code basically does two things then:
 - If the values were already NaN, and left is being set, it does not trigger the event since right is still NaN. When right is then set, it triggers the event. This point presupposes that the left value is always set before the right value.
 - If being reset, it should prevent the event from triggering as well. However, as far as I can tell, this should presuppose that the right is being set first, otherwise the event will trigger when the left value is being set to NaN, and currently the reset function sets the left value first.

A more robust way of checking this would perhaps be to check in both left and right `value_changed` functions whether _either_ of the two values are NaN?

francisco-dlp: Sorry, you were right from the beginning, this was a typo—sorry for the noise...

francisco-dlp: (I've just fixed it)

vidartf: I still think you can get in problems when the values are being reset (`reset_span_selector()`): When the first of the values is being set to NaN, the other still has a valid value, so the event is triggered.

I think it is best to check like this (as far as I can tell, that shouldn't break anything that isn't already broken?):
```python
if not (np.isnan(self.ss_left_value) or np.isnan(self.ss_right_value)):
```

francisco-dlp: This doesn't cause troubles because this better check is performed
[here](https://github.com/hyperspy/hyperspy/pull/581/files#r35655194).
However, I fully agree that it is better to do it properly here.

francisco-dlp: Sorry for the noise again: this doesn't cause any troubles in the
present case because we don't actually reset the span selector. But
again, it's better to check as you propose. (If I'm still getting
something wrong please let me know.)

vidartf: Did you want to push those tests here, or should I merge as it is?

francisco-dlp: It should be fine now, thanks.

hyperspy/gui/egerton_quantification.py
```diff
@@ -391,7 +395,7 @@ def create_interpolation_line(self):
 
     def get_interpolation_range(self):
         axis = self.signal.axes_manager.signal_axes[0]
-        if self.ss_left_value == self.ss_right_value:
+        if np.isnan(self.ss_left_value) or np.isnan(self.ss_right_value):
```
vidartf: If left == right, but both are valid numbers, should this branch be entered?

francisco-dlp: I've just tested a zero-width span selector and it doesn't upset HyperSpy. Hence, I suggest to leave it as it is because, otherwise, the interpolation line will expand to the default size when the span selector width becomes 0 which is counterintuitive.

francisco-dlp:

PR #582, opened 2015-07-08
ENH: rgbx2regular_array keeps masks

vidartf:
Currently, `misc.rgb_tools.rgbx2regular_array()` checks if the data is C contiguous, if not it converts it with  `np.ascontiguousarray()`. However, if the data is a masked array the mask will dissappear in this operation. This PR allows the data to retain its mask, while still ensuring C order. It is possible that the entire code could simply be replaced with `data = data.copy(order='C')` as it is the same for numpy arrays and masked arrays, however, `np.ascontiguousarray()` might be more efficient (I don't know).

TODO
--------
* [x] Unittest

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -1,0 +1,14 @@
+[run]
+branch = True
+source = hyperspy
+include = */hyperspy/*
+omit =
+    */hyperspy/misc/borrowed/*
+    */hyperspy/ipython_profile/*
+    */hyperspy/drawing/*
+    */hyperspy/mpfit/*
+    */hyperspy/gui/*
+    */setup.py
+    */examples/*
+    */bin/*
+    */hyperspy/misc/io/tifffile.py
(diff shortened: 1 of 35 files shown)


francisco-dlp:
Good catch and fix. Only tests are missing.

vidartf:
Added same functionality to `regular_array2rgbx()`, as well as tests. Tests do not cover other things than data order + masked type.

francisco-dlp:
Each test method contains multiple assertions. I think that each assertion should be a separate test because if one fails all the tests below the ones that failed won't run. Also, having a method per test helps making the code (and the error message) more readable.

vidartf:
I agree that it makes sense to split up the tests. Some of the tests still have several asserts, however, these logically belong together: I would normally use `and` between theses statements, but I split them up into different asserts to make it more readable. They are ordered such that if one of them fails, the outcome of those following is non-essential.

PR #583, opened 2015-07-10
Fix MVA.normalize factors

francisco-dlp:
`MVA.normalize_factors` was not rescaling properly the loadings when `on_loadings=True` (which was anyway broken, see #587).

This PR also deprecates `normalize_factors` in favor of the new `normalize_decomposition_components` and `normalize_bss_components`.


hyperspy/learn/mva.py
```diff
@@ -560,13 +572,64 @@ def normalize_factors(self, which='bss', by='area', sort=True):
         else:
             raise ValueError("by must be max or mean")
 
-        factors /= by(factors, 0)
-        loadings *= by(factors, 0)
+        coeff = by(factors, 0)
+        factors /= coeff
+        loadings *= coeff
```
to266: Why not use the `_normalize_components`?
I guess 
```python
_normalize_components(factors, loadings, by)
```
should do

to266:
also needs solving the merge conflicts.. otherwise looks good

hyperspy/learn/mva.py
francisco-dlp: I prefer it in this way because it is more explicit and not much longer.

francisco-dlp: Sorry for replying in a rush, you're right, I'll fix that.

francisco-dlp:
It should be fine now, thanks for the review.

Issue #561, opened 2015-06-03
remove_background GUI cannot be closed.


jat255:
I don't think this is related to #560, but I have also noticed some undesired behavior when using the `remove_background()` function to start the GUI. I assume this has something to do with my setup, but I am using all the latest packages for dependencies, and the 0.8 release version of Hyperspy.

When I do an interactive background removal, the window that controls the process and allows me to choose a background model remains behind after clicking 'OK'. I cannot get rid of it without restarting the kernel. See the following video for an example:

![remove_background_persist](https://cloud.githubusercontent.com/assets/1278301/7966515/9b749874-09f2-11e5-86a5-5f514926b26d.gif)

francisco-dlp:
Great bug report! However, here we have two bugs in one single issue, could you submit separate issues for each of them?

jat255:
Sorry, sure thing! I'll open another one for the background scaling issue.

francisco-dlp:
I cannot reproduce this in Linux using the latest packages is Anaconda.

vidartf:
Cannot reproduce on Windows 7 (64-bit) using WinPython 64-bit. Does this happen no matter what console you use (notebook/qtconsole) ?

jat255:
Sorry for the delay on getting back to this. I can confirm it doesn't happen on the `wx` backend, so it appears to be some sort of bug with my PyQt setup, it seems.

vidartf:
Have you checked if it happens for both PyQt and PySide?

jat255:
Is there a simple way to choose whether PyQt or PySide is used? I always thought you could only choose `qt4` as the backend.

vidartf:
Possible solutions from google:
```python 
matplotlib.use('Qt4Agg')
matplotlib.rcParams['backend.qt4']='PySide'
```

jat255:
Hmm, I'm not sure if that command had any impact. When I ran from a notebook, the problem persisted. When running from a console (just `hyperspy --toolkit qt4` from the command line), the background selection dialog looks different and runs as expected (closing when it should).

FYI, I have PySide (v1.2.2) and PyQt4 (v4.11.3) installed.

I'm guessing something is being loaded in the notebook script that's causing the issue?

jat255:
Also, I tried running in a qtconsole, but I got this error (which may be unrelated):
```python
ImportError:
    Could not load requested Qt binding. Please ensure that
[...]

PR #584, opened 2015-07-11
Fix: component.py's import of traitsui

vidartf:
There where quite a few issues with importing traitsui directly in component.py (introduced in #408). The need for this was because of a bug in traitsui's handling of a default editor for t.Either(t.CFloat(), Array()). This change implements a workaround that fixes both problems, but it accomplishes this by overriding a function on the trait. As such, it quite a bit of a hack.

Should fix half of #561, but I'll need @jat255 to confirm!

### DIFF
--- a/hyperspy/component.py
+++ b/hyperspy/component.py
@@ -22,7 +22,6 @@ import numpy as np
 import warnings
 
 import traits.api as t
-import traitsui.api as tu
 from traits.trait_numeric import Array
 
 from hyperspy.defaults_parser import preferences
@@ -42,6 +41,13 @@ class NoneFloat(t.CFloat):   # Lazy solution, but usable
             super(NoneFloat, self).validate(object, name, 0)
             return None
         return super(NoneFloat, self).validate(object, name, value)
+
+
+def _editor_wrapper(self):
+    from traitsui.api import TextEditor
+    if not isinstance(self.editor, TextEditor()):
+        self.editor = TextEditor()
+    return self.editor
 
 
 class Parameter(t.HasTraits):
@@ -101,9 +107,12 @@ class Parameter(t.HasTraits):
 
     # traitsui bugs out trying to make an editor for this, so always specify!
     # (it bugs out, because both editor shares the object, and Array editors
-    # don't like non-sequence objects). TextEditor() works well.
+    # don't like non-sequence objects). TextEditor() works well. Use a wrapper
+    # to avoid importing traitsui prematurely
     value = t.Property(
-        t.Either([t.CFloat(0), Array()]), editor=tu.TextEditor())
+        t.Either([t.CFloat(0), Array()]))
+    value.get_editor = _editor_wrapper
+
     units = t.Str('')
     free = t.Property(t.CBool(True))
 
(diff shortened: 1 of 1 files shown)


francisco-dlp:
I agree that this fixes it and also that this is quite a bit of  a hack.

What do you think about the following alternative?

1. Create a new file hyperspy/gui/traitsui_views.py that contain a view for `Parameter` where it is possible to define the editor.
2. Add a `gui` method to `Parameter` and `Component` that lazy imports the view from `hyperspy.gui.traitsui_views` and launches the gui code.

vidartf:
The advantage with the hack is that it works around the fault of traitsui, namely that of a broken default editor. If we give up on a full workaround, I think the best alternative would be to implement [default_traits_view](http://docs.enthought.com/traitsui/traitsui_user_manual/advanced_view.html#defining-a-default-view), as it is tasked with creating a default view for the HasTraits. That implementation can then lazy import what is needed. This is basically identical to the `gui` method you suggested (and that is used elsewhere in hyperspy), but integrates with traitsui. What do you prefer?

All proposed solutions will still fail if the trait is put into a custom view without specifying an editor, so the code should be clearly marked at the traits definition if we go for any of those.

francisco-dlp:
I agree, setting the default traits view sounds like a better approach as long as it can be done lazily. Also, if you believe that this is a traitsui bug, it might be worth reporting it to the traitsui team.

vidartf:
This should now be fixed. @jat255 Can you confirm whether this fixes half of #561 ?

hyperspy/component.py
```diff
@@ -470,6 +471,19 @@ def export(self, folder=None, name=None, format=None,
             self.as_signal(field='std').save(append2pathname(
                 filename, '_std'))
 
+    def default_traits_view(self):
```
francisco-dlp: Nice solution. Could you just add some comments to the code to explain why you need to set the default view.

hyperspy/component.py
```diff
@@ -470,6 +471,19 @@ def export(self, folder=None, name=None, format=None,
...
+        from traitsui.api import RangeEditor, View, Item
+        et = self.editable_traits()
+        whitelist = ['bmax', 'bmin',  'free', 'name', 'std', 'units', 'value']
+        et = [v for v in et if v in whitelist]
```
francisco-dlp: Wouldn't the following be more readable?

```python

editable_traits = [trait for trait in self.editable_traits() if trait in whitelist]

```

vidartf:
This should do it. Do you agree to have the whitelist, or is that superfluous?

francisco-dlp:
I'm +1 on the whitelist: I think that it is nice to take the opportunity to customize the view a little.

jat255:
> This should now be fixed. @jat255 Can you confirm whether this fixes half of #561 ?

I just tested this branch out. It does indeed fix the issue of a `wx` window getting loaded when using the `qt4` backend. I confirmed that `ETSConfig.toolkit` was `qt4`, as expected.

Unfortunately, the window disappearing silently still is an issue when using the qt backend...

vidartf:
Alright, then this should be ready for merge.

PR #585, opened 2015-07-11
BUG: O'haver maxpeakn & dtype order

vidartf:
Fixed two bugs with `find_peaks_ohaver()`:
* Order of dtype `height` and `width` were wrong compared to the order of the data.
* Now apply the `maxpeakn` argument, sorted by peak height. Can possibly be improved by sorting by a more advanced metric for peak strength.

Also cleaned up some formatting.

TODO
--------

* [x] Unittests.


francisco-dlp:
Nice. To be ready to merge it needs tests for the two issues that this is fixing: `maxpeakn` and the order `width` and `height`. There are already some tests [here](https://github.com/hyperspy/hyperspy/blob/master/hyperspy/tests/signal/test_1D_tools.py#L128).

vidartf:
Tests added, should be ready to go!

PR #588, opened 2015-07-13
FIX `MVA.blind_source_separation` broken when `on_loadings=True` broken #587

### LINKED ISSUE #587: `MVA.blind_source_separation` broken when `on_loadings=True` broken
The result of executing `MVA.blind_source_separation` is wrong when `on_loadings=True`. This bug is as old as the feature but it has not been noticed until now because the coverage of this part of the code is low and the feature is probably rarely used (the default is `on_loadings=False`).

francisco-dlp:
Fixes #587 and #602 and adds a tests.

Note that it includes #600, #591 and #589.


hyperspy/learn/mva.py
```diff
@@ -435,97 +470,196 @@ def blind_source_separation(self,
...
+                unmixing_matrix = lr.bss_node.components_
+        else:
+            if mdp_installed is False:
+                raise ImportError(
+                    'MDP is not installed. Nothing done')
+            to_exec = 'lr.bss_node=mdp.nodes.%sNode(' % algorithm
+            for key, value in kwargs.iteritems():
+                to_exec += '%s=%s,' % (key, value)
+            to_exec += ')'
+            exec(to_exec)
```
to266: how about:
```python
temp_function = getattr(mdp.nodes, algorithm+"Node")
lr.bss_node = temp_function(**kwargs)
```

hyperspy/learn/mva.py
```diff
@@ -435,97 +470,196 @@ def blind_source_separation(self,
...
+        # purpose as an user may like to apply pretreaments that change their
+        # dimensionality.
+
+        # The diff_axes are given for the main signal. We need to compute
+        # the correct diff_axes for the factors.
+        # Get diff_axes index in axes manager
+        if diff_axes is not None:
+            diff_axes = [1 + axis.index_in_axes_manager for axis in
+                         [self.axes_manager[axis] for axis in diff_axes]]
+            if on_loadings is False:
```
to266: I think the more pythonic way would be 
```python
if not on_loadings:
#    pass
```

hyperspy/learn/mva.py
```diff
@@ -435,97 +470,196 @@ def blind_source_separation(self,
...
+                if on_loadings
+                else
+                (self.axes_manager.signal_shape, "signal"))
+            if isinstance(mask, np.ndarray):
+                ref_shape = ref_shape[::-1]
+                if mask.shape != ref_shape:
+                    raise ValueError(
+                        "The `mask` shape is not equal to the %s shape in the"
+                        "array. Mask shape: %s\tSignal shape in array: %s" %
+                        (space, str(mask.shape), str(ref_shape)))
```
to266: I would change the last "Signal" to another `space` - it can be confusing with an error that states

"The `mask` shape is not equal to the __navigation__ shape in the array. Mask shape: (1,2,3)   __Signal__ shape in array: (3,2,1)"

hyperspy/tests/mva/test_bss.py
```diff
@@ -0,0 +1,197 @@
...
+        self.s.blind_source_separation(3, diff_order=0, mask=mask,
+                                       on_loadings=True)
+
+    def test_mask_diff_order_1_on_loadings(self):
+        mask = self.s._get_navigation_signal(dtype="bool")
+        mask[5] = True
+        self.s.learning_results.loadings[5, :] = np.nan
+        self.s.blind_source_separation(3, diff_order=1, mask=mask,
+                                       on_loadings=True)
+
```
to266: I think all mask_diff_* tests above don't actually test anything... At least no `nose.tools` are used, which is somewhat odd

hyperspy/tests/mva/test_bss.py
```diff
@@ -0,0 +1,197 @@
...
+            axis="x", order=1)
+        self.s.blind_source_separation(
+            3, diff_order=0, fun="exp", on_loadings=False, factors=factors)
+        matrix = self.s.learning_results.unmixing_matrix.copy()
+        print matrix
+        self.s.blind_source_separation(
+            3, diff_order=1, fun="exp", on_loadings=False,
+            diff_axes=["x"],
+        )
+        matrix = self.s.learning_results.unmixing_matrix.copy()
```
to266: This line should be deleted, as it overwrites the results of previous calculations, hence in the end the test checks that  a copy is equal to the original object

hyperspy/tests/mva/test_bss.py
```diff
@@ -0,0 +1,197 @@
...
+        ics = np.random.laplace(size=(3, 1024))
+        np.random.seed(1)
+        mixing_matrix = np.random.random((100, 3))
+        s = Image(np.dot(mixing_matrix, ics).reshape((100, 32, 32)))
+        for (axis, name) in zip(s.axes_manager._axes, ("z", "y", "x")):
+            axis.name = name
+        s.decomposition()
+        self.s = s
+
+    def test_diff_axes_string_with_mask(self):
```
to266: mask is not used in this test

hyperspy/tests/mva/test_bss.py
```diff
@@ -0,0 +1,197 @@
...
+    def test_mask_diff_order_1_on_loadings_diff_axes(self):
+        s = self.s.to_spectrum()
+        s.decomposition()
+        mask = s._get_navigation_signal(dtype="bool")
+        mask.unfold()
+        mask[5] = True
+        mask.fold()
+        s.learning_results.loadings[5, :] = np.nan
+        s.blind_source_separation(3, diff_order=1, mask=mask,
+                                  on_loadings=True, diff_axes=["x"])
```
to266: Again, all masked tests above don't test anything, just perform some calculations

francisco-dlp:
@to266, thank you for the thorough review, all the issues that you've mention should be fixed now.

francisco-dlp:
@to266, is this ok to merge now?

PR #589, opened 2015-07-14
Add `data` argument to `_get_*_signal` methods

francisco-dlp:
Add  `data` argument to `Signal._get_signal_signal` and `Signal._get_navigation_signal`.

This PR also includes changes to the code where using the new `data` argument can avoid memory duplication.

### DIFF
--- a/hyperspy/misc/eels/tools.py
+++ b/hyperspy/misc/eels/tools.py
@@ -293,7 +293,7 @@ def eels_constant(s, zlp, t):
     # Kinetic definitions
     ke = e0 * (1 + e0 / 2. / me) / (1 + e0 / me) ** 2
     tgt = e0 * (2 * me + e0) / (me + e0)
-    k = s._get_navigation_signal()
-    k.data = (t * i0 / (332.5 * ke)) * np.log(1 + (beta * tgt / eaxis) ** 2)
+    k = s.__class__(data=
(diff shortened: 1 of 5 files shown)


hyperspy/misc/eels/tools.py
```diff
@@ -293,7 +293,7 @@ def eels_constant(s, zlp, t):
     # Kinetic definitions
     ke = e0 * (1 + e0 / 2. / me) / (1 + e0 / me) ** 2
     tgt = e0 * (2 * me + e0) / (me + e0)
-    k = s._get_navigation_signal()
-    k.data = (t * i0 / (332.5 * ke)) * np.log(1 + (beta * tgt / eaxis) ** 2)
+    k = s.__class__(data=
+        (t * i0 / (332.5 * ke)) * np.log(1 + (beta * tgt / eaxis) ** 2))
```
vidartf: Was this a bug previously?

francisco-dlp: It was a bug but, AFAIK, it didn't have any consequences for the user as only the `k.data` is used somewhere else in the code. Hence, I thought that it wasn't worth to report it as a separate issue—do you think otherwise?

vidartf: No, I was just curious for the reason of the change, making sure it was intended :)

vidartf:
Looks OK to me, except docstring for `dtype` argument is missing. Some merge conflicts as well.

francisco-dlp:
This should be ok now.

PR #591, opened 2015-07-15
Fix unfold

### LINKED ISSUE #590: `Signal.unfold` does nothing when `Signal.data.ndim` is 2.
The following reproduces the bug.

```python

>>> import numpy as np
>>> im = hs.signals.Image(np.empty(( 2, 2)))
>>> im = hs.signals.Image(np.empty(( 2, 2)))
>>> im
<Image, title: , dimensions: (|2, 2)>
>>> im.unfold()
>>> im
<Image, title: , dimensions: (|2, 2)>

```

francisco-dlp:
Fixes #590

### DIFF
--- a/hyperspy/signal.py
+++ b/hyperspy/signal.py
@@ -3513,7 +3513,7 @@ class Signal(MVA,
             unfolded_axis = (
                 self.axes_manager.navigation_axes[0].index_in_array)
             self._unfold(steady_axes, unfolded_axis)
-            return needed_unfolding
+        return needed_unfolding
 
     def unfold_signal_space(self):
         """Modify the shape of the data to obtain a signal space of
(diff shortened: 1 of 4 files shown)


hyperspy/signal.py
```diff
@@ -3485,35 +3482,58 @@ def unfold(self):
...
+            needed_unfolding = False
+        else:
+            needed_unfolding = True
+            steady_axes = [
+                axis.index_in_array for axis in
+                self.axes_manager.signal_axes]
+            unfolded_axis = (
+                self.axes_manager.navigation_axes[0].index_in_array)
+            self._unfold(steady_axes, unfolded_axis)
+            return needed_unfolding
```
vidartf: This `return` line should be unindented.

hyperspy/signal.py
```diff
@@ -3485,35 +3482,58 @@ def unfold(self):
...
-        self._unfold(steady_axes, unfolded_axis)
+            needed_unfolding = False
+        else:
+            needed_unfolding = True
+            steady_axes = [
+                axis.index_in_array for axis in
+                self.axes_manager.navigation_axes]
+            unfolded_axis = self.axes_manager.signal_axes[0].index_in_array
+            self._unfold(steady_axes, unfolded_axis)
+            return needed_unfolding
```
vidartf: This one too.

francisco-dlp: Thanks, it should be fine now.

Issue #599, opened 2015-07-20
`Signal.diff` sets incorrect axis offset when `order > 1`


francisco-dlp:
The following reproduces the bug:

```python
>>> import hyperspy.hspy as hs
>>> s = hs.signals.Spectrum(np.random.random((10, 20, 30)))
>>> s.axes_manager[-1].offset
0.0
>>> s.diff(axis=-1, order=1).axes_manager[-1].offset
0.5
>>> s.diff(axis=-1, order=2).axes_manager[-1].offset
0.5
>>> s.diff(axis=-1, order=1).diff(axis=-1, order=1).axes_manager[-1].offset
1.0

```

PR #600, opened 2015-07-20
Fix diff

### LINKED ISSUE #598: `Signal.diff` axis argument only takes integers
All `axis` arguments in HyperSpy should take string (name of axis) and integer arguments. According to its docstring `Signal.diff` should be no exception, however:

```python

>>> s = signals.Spectrum(np.random.random((10, 20, 30)))
>>> s.axes_manager[1].name = "y"
>>> s.diff(axis="y")
  File "<ipython-input-4-1b0dccc8eb43>", line 1, in <module>
    s.diff(axis="y")
  File "/home/fjd29/Python/hyperspy/hyperspy/signal.py", line 3794, in diff
    np.diff(self.data, order, axis))
  File "/home/fjd29/Anaconda/anaconda2/lib/python2.7/site-packages/numpy/lib/function_base.py", line 1107, in diff
    slice1[axis] = slice(1, None)
TypeError: list indices must be integers, not str

```

francisco-dlp:
Fixes #598 and #599 and adds tests.

### DIFF
--- a/hyperspy/signal.py
+++ b/hyperspy/signal.py
@@ -3767,18 +3767,15 @@ class Signal(MVA,
     def diff(self, axis, order=1):
         """Returns a signal with the n-th order discrete difference along
         given axis.
-
         Parameters
         ----------
         axis : {int | string}
            The axis can be specified using the index of the axis in
            `axes_manager` or the axis name.
         order: the order of the derivative
-
         See also
         --------
         mean, sum
-
         Examples
         --------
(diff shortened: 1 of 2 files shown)


to266:
Looks good. The only question is if we should let users know that the procedure is recursive, hence if diff order is passed, it's rounded up..? (e.g. `np.diff(anything, n=1.3)` is actually equal to `np.diff(anything, n=2)`)

PR #603, opened 2015-07-21
Update external tifffile to lastest version and prefer importing it from skimage

francisco-dlp:
In order to read scientific tiff file formats HyperSpy includes the external library [tifffile](http://www.lfd.uci.edu/~gohlke/code/tifffile.py.html). This PR upgrades it to its latest version (2014.08.24). Also it tries to import it from scikit image first as, if installed, it comes with the optional C module to speed some things up.


hyperspy/io_plugins/tiff.py
```diff
@@ -20,10 +20,18 @@
 import warnings
 
 import traits.api as t
-with warnings.catch_warnings():
-    warnings.simplefilter("ignore")
-    from hyperspy.misc.io.tifffile import imsave, TiffFile
 from hyperspy.misc import rgb_tools
+try:
+    from skimage.external.tifffile import imsave, TiffFile
+    raise ImportError
```
vidartf: Is this supposed to be here?

francisco-dlp: At least [it seems so](https://github.com/scikit-image/scikit-image/blob/master/skimage/external/tifffile/tifffile.py).

vidartf: I'm thinking about the `raise ImportError` that will always trigger.

francisco-dlp: As you may have guessed, that was a debugging import. It should be fixed now, thanks!

PR #605, opened 2015-07-21
Tidy external modules

francisco-dlp:
HyperSpy ships with some few external modules that, at present, are somehow randomly scattered in the package. This PR place them all in a new `hyperspy.external` submodule as seems to be the convention in other packages.


vidartf:
Partly related: I see that for mpfit we're actually embedding an SVN repository. Is this intentional?

francisco-dlp:
No, it's not. I've just deleted it, thanks!

vidartf:
Docs still need updating for this one:
 - hyperspy.misc.rst: References to progressbar and borrowed
 - hyperspy.misc.mpfit.rst
 - hyperspy.misc.mpfit.tests.rst

Also, old folders need to be removed (hyperspy/misc/mpfit, borrowed, hio?).

francisco-dlp:
Good point about the docs: the api docs are automatically generated but the old files are not automatically deleted. Regarding the folders, if they're empty they won't appear in a freshly cloned repository. To clean an old one in [stackoverflow](http://stackoverflow.com/questions/10074582/git-how-to-remove-empty-folder-and-push-that-change) they suggest to use `git clean -fdn` and if it looks fine `git clean -fd`

vidartf:
If your saying the docs are as they should, then all that is missing is to solve merge conflicts.

francisco-dlp:
It should be fine now. Notice that I've merged #635 here in order to fix the api docs issue that you've spotted.

Issue #496, opened 2015-03-31
EHN: `map` more flexible


pburdet:
In the following example, `map` doesn't work because the function returns data with a different shape.

`map` should support returning any data shape

``` python

from skimage.feature import peak_local_max
import scipy.misc
ims = signals.Signal(scipy.misc.face()).as_image([0,1])
index = im.map(peak_local_max, min_distance=100, num_peaks=4)
for i in range(4):
     m = utils.plot.markers.point(x=index[:, i, 1],
                                  y=index[:, i, 0], color='red')
     ims.add_marker(m)

```

![plot_markers_im](https://cloud.githubusercontent.com/assets/3855190/6921339/2c7bdf44-d7bc-11e4-8a6f-12acb72eb907.png)

dnjohnstone:
I think I faced this issue recently trying to find peaks and extract lattice vectors in images. I dealt with this as in #563 but not sure if a similar solution could be used for the map function - or perhaps I misunderstood the issue.

A related issue is what to do when you essentially have a variable signal range e.g. 4 peaks for the first image, 5 peaks in the second image... Apparently this is discussed in some places as dealing with "ragged arrays" but not sure that there is a standard/good solution.

francisco-dlp:
[dynd](https://github.com/libdynd/libdynd) implements ragged arrays.

pburdet:
Thanks for the answer, my problem here was not ragged array, but when the array returns by ```map``` is a regular array of different shape than the input. In this case, returning the result as a signal would be easy to implement. It would makes ```map``  of a broader use (similarly to the python map)

francisco-dlp:
Once again, this is a nice feature request. However, in order to keep it in for HyperSpy 0.9 someone should be assigned to it (any volunteer?), otherwise we should move it to the wish list.

dnjohnstone:
I'll have a go at adding this functionality ideally in such a way that other analyses within HyperSpy can call the mapping functionality because in trying to find peaks in an image and then convert this to a pair of lattice vectors I have essentially been repeatedly writing such functions. From a brief chat with Francisco we think that the map function should be able to operate in at least 3 modes and below I list the 5 that I think would make it fully general:

Mode 1: output dimension equal to input dimension (currently optimised for speed and memory)
Mode 2: output dimension regular but not equal to input dimension
Mode 3: output dimension ragged and not equal to input dimension
Mode 4: output dimension regular but input dimension ragged
[...]

PR #606, opened 2015-07-21
ENH: Generalising 'map'

dnjohnstone:
Working towards a more generalised 'map' function as discussed in #496.

At present 'map' overwrites the original signal that is being acted on and can only accept functions for which the output dimension is the same as the original signal dimension. This is good in terms of memory usage and speed but not so good if you want to apply a function which returns some other dimension or if you want to tweak the parameters of that function.

Here, I have implemented a second approach based on first generating a structured array with dimensions equal to the navigation dimensions and iterating over that. This approach is used elsewhere in the source code and is potentially very general as the output of the function to be mapped can now have any output dimension and furthermore may be ragged. The downside of the structured array is that it's not very compatible with indexation within HyperSpy (though it is fully indexable).

Next step: Enable the output to be returned as a HyperSpy signal when the output dimensions are not ragged.

### DIFF
--- a/hyperspy/signal.py
+++ b/hyperspy/signal.py
@@ -4001,8 +4001,7 @@ class Signal(MVA,
         hist_spec.metadata.Signal.binned = True
         return hist_spec
 
-    def map(self, function,
-            show_progressbar=None, **kwargs):
+    def map(self, function, overwrite=False, show_progressbar=None, **kwargs):
         """Apply a function to the signal data at all the coordinates.
 
         The function must operate on numpy arrays and the output *must have the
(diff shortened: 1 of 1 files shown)


to266:
https://github.com/hyperspy/hyperspy/pull/737

Will update in a second to be up-to-date. 

Sorry for creating a very similar branch, did not know about this one at the time. Have you actually user yours in any task? If so, could you see if mine (once updated) is good enough? Looks like there is slightly more done in #737, so fewer things to do next

dnjohnstone:
@to266 Happy to work from #737  - I agree that you've got more done there! As you know, I'm currently going through the obligatory baptism of fire regarding what's actually required to consider something finished for inclusion in HyperSpy...

If we're agreed - I'll close this pull request and then (if it looks like it's worth it) will try and include the idea I had here into #737... Sound ok?

to266:
Yup, go for it. I'm happy to add things if something is lacking any way

dnjohnstone:
Closed in favour of #737

PR #607, opened 2015-07-21
ENH: scree plot improvements

jat255:
**Todo**
- [ ] Highlighting by using different symbols (for gray figures).
- [ ] Ability to pass the matplotlib axes (to plot several scree plots in the same axes)
- [ ] Ability to pass the matplotlib figure (e.g. to customize the size).
- [ ] Implement dictionary parsing for options
- [ ] Add options for horizontal line
- [ ] Add options for axes titles

**Explanation**
I wrote this function a while ago, and forgot to see if anyone would want to see it included.

Basically, it is a series of improvements to the "explained variance ratio" or Scree plot that should make for  better display of this information. I didn't want to step on any toes, so I wrote it as a different function, but if people like it, I guess we could use it as the default `plot_explained_variance_ratio`.

Primary improvements are:
 * Easy ability to change color of plotted components
 * Capability to highlight first `n` components (to easily show what was included in decomposition, for example)
 * Easy ability to control size of markers from within function

I'm not sure if anyone else would appreciate this functionality, but I thought I'd throw it out there in case anyone is interested.

For reference, here's a comparison of the default function, and what this version produces:

```python
s.plot_explained_variance_ratio(n=20)
```
plot_evr

```python
s.scree_plot(n=20, highlight=5)
```
plot_scree


francisco-dlp:
Very nice. I think that this is indeed very useful as most people using PCA/BSS will want to include an scree plot figure in their papers. I'm +1 on replacing the current version with (an improved version of) this one.

Then, if the purpose of this function is to have a publication ready figure I think that this function should also include:

1. Highlighting by using different symbols (for gray figures).
2. Ability to pass the matplotlib axes (to plot several scree plots in the same axes)
3. Ability to pass the matplotlib figure (e.g. to customize the size).


By the way, wouldn't it be more powerful to have two keywords that takes dictionaries with the keywords to pass to each set of points e.g. `signal_kwds` and `noise_kwds`? In this case 1 could be accomplished just by adding an example showing how to do it.

jat255:
> Then, if the purpose of this function is to have a publication ready figure I think that this function should also include:
> 
> Highlighting by using different symbols (for gray figures).
> Ability to pass the matplotlib axes (to plot several scree plots in the same axes)
> Ability to pass the matplotlib figure (e.g. to customize the size).

All of these are great suggestions. I'll get to working on them when I can find a few free moments :smile: 


> By the way, wouldn't it be more powerful to have two keywords that takes dictionaries with the keywords to pass to each set of points e.g. signal_kwds and noise_kwds? In this case 1 could be accomplished just by adding an example showing how to do it.

This is a smart way to accomplish this, and will allow us to keep it close to the matplotlib syntax, hopefully.

Would it be best to create a new branch that doesn't have all those spurious merge commits?

francisco-dlp:
> Would it be best to create a new branch that doesn't have all those spurious merge commits?

It is up to you, creating a new branch is cleaner but keeping the current one is more historically accurate and I see nothing wrong with it.

dnjohnstone:
I think this is a, two suggestions:

1. It would be good to have the option to add a horizontal line (perhaps dashed) at a user specified variance level to indicate a threshold for component selection. This serves much the same purpose as different colours but I think it's nice to leave that choice to users.

2. Axes titles... some people have complained at me regarding the use of the word "explained" in the y-axis title and prefer me to omit it (i.e. just "Variance Ratio") when producing figures for them. I tend to agree that the word explained is a slightly strange choice of English, though one can construct an argument for it. Similarly a part  of me thinks that the x-axis should read simply "Principal Component" or perhaps "Principal Component Ranking".

Point 2 may seem menial but default values for such things are important in setting standard terminology amongst users. I think we should set out why our terminology is correct.

dnjohnstone:
@jat255 did this get any further? It would be nice to use some of this if it has :)

jat255:
@dnjohnstone no, unfortunately I haven't added those suggestions yet. I should hopefully be able to get to it in the next few days!

thomasaarholt:
Just basically copying @dnjohnstone's comment :) looks very nice!

jat255:
Ha. "Few days." Look at me being optimistic :smile:  Thesis days are apparently longer than non-thesis days. 

Honestly, I probably wouldn't be able to finish this up until after my defense in July. If someone wants to fork it and push it through, that's fine by me, or I can do it later this summer...

francisco-dlp:

PR #610, opened 2015-07-21
Proposed fix for #466: polynomial estimate_parameters

vidartf:
Proposed fix for #466. A little unsure about if these lines are correct when unfolding:
```python
if axis.index_in_array > 0:
    dc = np.rollaxis(dc, 1, 0)
...
if axis.index_in_array > 0:
    cmaps = np.rollaxis(cmaps, 0, 2)
```
@francisco-dlp you seem quite steady on the (un)folding code now. What do you think?

### TODO
 - [x] Confirm `rollaxis`
 - [x] Tests


hyperspy/_components/polynomial.py
```diff
@@ -104,19 +104,23 @@ def estimate_parameters(self, signal, x1, x2, only_current=False):
...
-            if binned is True:
-                self.coefficients.map["values"] /= axis.scale
-            self.coefficients.map['is_set'][:] = True
-            signal.fold()
+            unfolded = signal.unfold()
+            try:
+                dc = signal.data
+                # For polyfit the spectrum goes in the first axis
+                if axis.index_in_array > 0:
+                    dc = np.rollaxis(dc, 1, 0)  # Unfolded, so use 1
```
francisco-dlp: > # Unfolded, so use 1

Good point, then why not simply `dc = dc.T`

hyperspy/_components/polynomial.py
```diff
@@ -104,19 +104,23 @@ def estimate_parameters(self, signal, x1, x2, only_current=False):
...
+                cmaps = np.polyfit(axis.axis[i1:i2], dc[i1:i2, :],
+                                   self.get_polynomial_order())
+                if axis.index_in_array > 0:
+                    cmaps = np.rollaxis(cmaps, 0, 2)
```
francisco-dlp: Same here  `cmaps = cmaps.T` should be equivalent but more readable.

hyperspy/_components/polynomial.py
vidartf: Missed that opportunity, thanks!

vidartf:
Tests added, so if they look OK this should be ready for pull!

PR #611, opened 2015-07-21
DOCENH: Add examples for using colormap in plot_spectra

jat255:
I just needed this functionality, I think it is a good example to show users how they can use colormaps to get a sequential list of colors, which can then be fed into `plot_spectra`'s `color` option.

For easy reference, I generated the docs locally, and here's what the section looks like:
docenh_plot_spectra


francisco-dlp:
That's a nice example, thanks!

francisco-dlp:
Could you pull request it to the 0.8.x branch?

jat255:
Closed. See PR #614 for PR onto 0.8.x branch

PR #612, opened 2015-07-21
Add numerical derivative to Signal

francisco-dlp:
Example of usage:

```python

>>> import numpy as np
>>> offset = 3
>>> scale = 0.01
>>> x = np.arange(-offset, offset, scale)
>>> s = signals.Spectrum(np.sin(x))
>>> s.axes_manager[0].scale = scale
>>> s.axes_manager[0].offset = -offset
>>> utils.plot.plot_spectra([s.derivative(0, i) for i in range(5)])
<matplotlib.axes._subplots.AxesSubplot object at 0x7f09d0d0acd0>


```
derivative


francisco-dlp:
Note that, although this is a new feature, I'm pull requesting it to 0.8.x because it is required for #588.

hyperspy/signal.py
```diff
@@ -3810,16 +3807,46 @@ def diff(self, axis, order=1):
...
-        axis.offset += (axis.scale / 2)
+            np.diff(self.data,
+                    n=order,
+                    axis=self.axes_manager[axis].index_in_array))
+        axis = s.axes_manager[axis]
+        axis.offset += (order * axis.scale / 2)
         s.get_dimensions_from_data()
         return s
 
+    def derivative(self, axis, order=1, central_differences=False):
```
to266: `central_differences` is not used (and not clear what it does as well)

francisco-dlp: Fixed, thanks!

hyperspy/tests/signal/test_tools.py
```diff
@@ -318,3 +331,20 @@ def test_signal_iterator():
...
+        offset = 3
+        scale = 0.1
+        x = np.arange(-offset, offset, scale)
+        s = signals.Spectrum(np.sin(x))
+        s.axes_manager[0].offset = x[0]
+        s.axes_manager[0].scale = scale
+        self.s = s
+
+    def test_derivative_data(self):
+        self.s.derivative(axis=0, order=4)
```
to266: Here you don't save the derivative signal, hence don't test it in any way...

hyperspy/signal.py
```diff
@@ -3808,16 +3805,46 @@ def diff(self, axis, order=1):
...
+        order: int
+            The order of the derivative
+
+        Returns
+        -------
+        def : Signal
+            The size of the
+
+        See also
+        --------
```
to266: I would have a look at the documentation again, to make sure that it states correct things and remove any unfinished sentences.

francisco-dlp: It should be fixed now, thanks again!

PR #613, opened 2015-07-21
ENH: PEP8

vidartf:
While most PRs now follow PEP8, the existing code base does not. This leads to unfocused PR diffs for PRs where autopep8 has been used, or where PEP8 cleanup has been manually added. I tried to clean up most of the code base in this PR, and several of the changes are quite simple autopep8 rearrangements, but I also had to tune quite a bit manually. Especially the maths for the components. For this reason it should be reviewed before merge. Note that the "severity" of the changes has been graded in the commits, so quite a bit of filtering can be done based on them.


francisco-dlp:
I have gone through it all and I didn't find anything to object to it so, even though I know I will repent the next time that I have to merge something, here this goes into 0.8.x. Thanks @vidartf! By the way, how can this increase coverage by 0.004%?

PR #617, opened 2015-07-24
Enable travis docker infraestructure (faster)

francisco-dlp:
This is just to migrate to the container based infrastructure. Details [here](http://docs.travis-ci.com/user/migrating-from-legacy/?utm_source=legacy-notice&utm_medium=banner&utm_campaign=legacy-upgrade).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,7 @@
 language: python
 
+sudo: false
+
 python:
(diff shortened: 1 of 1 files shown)


francisco-dlp:
In a way, travis itself has reviewed this one so it's safe to merge.

PR #618, opened 2015-07-24
API change: create model

### LINKED ISSUE #568: Deprecate `create_model`
I propose to turn the `hspy.create_model` function into a `Signal` method that should allow us to write:

```python
 m = s.create_mode()

```

Instead of 

```python

m = create_model(s)

```

The advantage is that, given that there are options which are specific for each `Signal` type, the docstring and argument list will be less confusing. What do you think?

francisco-dlp:
Fixes #568


hyperspy/_signals/eels.py
```diff
@@ -1184,3 +1184,43 @@ def kramers_kronig_analysis(self,
...
             return eps, output
+
+    def create_model(self, ll=None, auto_background=True, auto_add_edges=True,
+                     GOS=None):
+        """Create a model for the current EELS data.
+
+        Parameters
+        ----------
+        ll : EELSSpectrum, optional
+            If an EELSSPectrum is provided, it will be assumed that it is
```
vidartf: EELSSPectrum -> EELSSpectrum :)

vidartf:
I see that #491 is merged here, is that on purpose?

vidartf:
I see you merged #491 here. Did you want to get that into 0.8.x as well?

francisco-dlp:
No, that wasn't intended. I'll close this and send a new PR to fix it. Thanks!

PR #620, opened 2015-07-24
New syntax to acess model components

### LINKED ISSUE #569: Add `Model.components.<component name>` attribute as an alternative to Mode["<component name>"]
Currently there is `__getitem__` so that one can do the following:

```python

>>> s = signals.Spectrum(np.empty((10, 20, 30)))
>>> m = create_model(s)
>>> m.append(components.Gaussian())
>>> m["Gaussian"].A.value
1.0
```

What I propose is to implement

```python
>>> m.components.Gaussian.A.value
1.0
```

The advantage is that with the latter we can enjoy IPython's tab completion. If this gets implemented then we can get rid of the ugly functionality that add the EELSCLEdge component variables to the user namespace automatically.

What do you think?

francisco-dlp:
Fixes #569 

TODO
--------
- [x] Change docs (new model and components printing)
- [x] Create issue describing the changes to __repr__ (@to266)
- [x] Fix bug: mysterious duplicate call to __delattr__
- [x] Fix warnings issue


doc/user_guide/model.rst
```diff
@@ -141,6 +141,24 @@ index in the model.
...
+
+    in interactive data analysis with IPython because it enables tab completion.
+
+    .. code-block:: python
+
+        >>> m
+        [<Carbon (Gaussian component)>,
+         <Hydrogen (Gaussian component)>,
+         <Nitrogen (Gaussian component)>]
+        >>> m.components.Gaussian
```
to266: `Carbon` instead of `Gaussian`?

doc/user_guide/eels.rst
```diff
@@ -159,6 +159,19 @@ Furthermore, the components are available in the user namespace
...
+   the automatically created component use the following equivalent syntax:
+
+    .. code-block:: python
+
+        >>> m.components.N_K
+        <N_K (EELSCLEdge component)>
+        >>> m.components.B_K
+        <B_K (EELSCLEdge component)>
+        >>> m.background
+        <background (PowerLaw component)>
```
to266: `m.components.background` ? 

I think it would be strange to single out this one particular component..

hyperspy/model.py
```diff
@@ -49,6 +49,17 @@
...
+
+
+class ModelComponents(object):
+
+    """Container for model components.
+
+    Useful to provide tab completion when running in IPython.
+
+    """
+    pass
```
to266: How about adding `__repr__` method, where each component is printed in a new line?
I personally find the current printing confusing, as all the extra characters (while useful to determine actual class of the component) clutter the print statement and it takes extra time to figure out where components start and end

doc/user_guide/eels.rst
francisco-dlp: I agree, although this may break a lot of scripts out there, I'll deprecate it.

hyperspy/io_plugins/tiff.py
```diff
@@ -27,6 +27,7 @@
     with warnings.catch_warnings():
         warnings.simplefilter("ignore")
         from hyperspy.misc.borrowed.tifffile import imsave, TiffFile
+        warnings.simplefilter("default")
```
vidartf: This line doesn't actually do anything (filters are reset after the `with` block resolves). There is however an interesting bug/feature for python 2 that was resolved for python 3.4: https://bugs.python.org/issue4180

hyperspy/misc/machine_learning/import_sklearn.py
```diff
@@ -27,3 +27,5 @@
         sklearn_installed = True
 except ImportError:
     sklearn_installed = False
+finally:
+    warnings.simplefilter("default")
```
vidartf: This should also be redundant as context managers get released even if an exception happens in the `with` block.

francisco-dlp: That's what I thought, but adding it fixed the issue. However this may be because, as you pointed out somewhere else, setting it to "default" has the counterintuitive consequence of printing the DeprecationWarnings which are not printed by default...

francisco-dlp:
This one should be ready now (could somebody check it once more?)

Thank you @to266 for the contributions and @vidartf for the review.

Issue #615, opened 2015-07-23
Context manager for unfold/fold


vidartf:
Several places in the code we use the pattern
```python
unfolded = signal.unfold()
... Do what is needed ...
if unfolded:
    signal.fold()
```

I'm proposing to add a context manager to ease this pattern, with the added benefit of making the code more robust to exceptions in the middle part:
```python
with signal.unfolded():
    ... Do what is needed ...
```
where the signal would automatically refold (if needed) when exiting the code block.

francisco-dlp:
I'm +1 on this.

PR #621, opened 2015-07-24
NEW: Signal.unfolded context manager

vidartf:
Suggested implementation of #615.

### TODO
 - [x] Tests
 - [ ] Use in appropriate places in existing code
 - [ ] Implement `unfolded_signal()` / `unfolded_navigation()` ?


francisco-dlp:
Very nice.I didn't find anything to object in the code. Regarding `unfolded_signal()` / `unfolded_navigation()`, probably this could be implemented using arguments instead of new methods. e.g. `unfolded(signal=True, navigation=False)`. The same argument goes for `unfold` itself of course.

Given that the context manager helps preventing bugs, I'm +1 on this going into 0.8.1.

PR #622, opened 2015-07-24
Fix zero dim data

### LINKED ISSUE #517: Problem with 0D signal

francisco-dlp:
Fixes #517

### DIFF
--- a/hyperspy/signal.py
+++ b/hyperspy/signal.py
@@ -2809,7 +2809,7 @@ class Signal(MVA,
 
         """
 
-        self.data = np.asanyarray(file_data_dict['data'])
+        self.data = np.atleast_1d(np.asanyarray(file_data_dict['data']))
         if 'axes' not in file_data_dict:
             file_data_dict['axes'] = self._get_undefined_axes_list()
         self.axes_manager = AxesManager(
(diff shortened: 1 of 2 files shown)


hyperspy/signal.py
```diff
@@ -3585,19 +3585,29 @@ def _iterate_signal(self):
...
+                self.metadata.Signal.record_by = self._record_by
+                self._assign_subclass()
+        else:
+            # Don't remove axis because it is the only one and HyperSpy does not
+            # support 0 dimensions
+            axis.size = 1
+            axis.scale = 1
+            axis.offset = 0
+            axis.name = "scalar"
+            axis.navigate = False
```
vidartf: Should we maybe still remove the axis and replace it with a new one like this? Then at least the old instance will not be in the collection anymore (and the new axis won't retain any of its bindings).

francisco-dlp: Good point. Actually, it might be worth to think about this a little bit more. Should we have a different class for the dummy scalar axes .e.g. `ScalarAxis`, that does not have `scale` and `offset` attributes? Also, given that many `Signal` methods will fail on 1d signals, should we have a special `Signal` too, e.g. `ScalarSignal`?

vidartf: I don't have the overview to see all the pros and cons with these suggestions, so I can't really have an informed opinion in terms of the consequences. For now I would maybe just do a new axis here, and then (if wanted) put these suggestions as more comprehensive issues for 0.9.

to266: I agree with @vidartf . For now lets just merge the current solution so that the bug is fixed

PR #624, opened 2015-07-24
Api create model method

### LINKED ISSUE #568: Deprecate `create_model`

francisco-dlp:
Fixes #568

### DIFF
--- a/doc/user_guide/eels.rst
+++ b/doc/user_guide/eels.rst
@@ -13,7 +13,7 @@ The functions described in this chapter are only available for the
 :py:class:`~._signals.eels.EELSSpectrum`:
 
 .. code-block:: python
-       
+
     >>> s.set_signal_type("EELS")
 
 Note these chapter discusses features that are available only for
@@ -104,13 +104,13 @@ EELS curve fitting
 HyperSpy makes it really easy to quantify EELS core-loss spectra by curve
 fitting as it is shown in the next example of quantification of a boron nitride
 EELS spectrum from the `The EELS Data Base
-<http://pc-web.cemes.fr/eelsdb/index.php?page=home.php>`_. 
+<http://pc-web.cemes.fr/eelsdb/index.php?page=home.php>`_.
 
 Load the core-loss and low-loss spectra
 
 
 .. code-block:: python
-       
+
     >>> s = load("BN_(hex)_B_K_Giovanni_Bertoni_100.msa")
     >>> ll = load("BN_(hex)_LowLoss_Giovanni_Bertoni_96.msa")
 
@@ -119,29 +119,29 @@ Set some important experimental information that is missing from the original
 core-loss file
 
 .. code-block:: python
-       
+
     >>> s.set_microscope_parameters(beam_energy=100, convergence_angle=0.2, collection_angle=2.55)
-    
-    
+
+
 Define the chemical composition of the sample
 
 .. code-block:: python
-       
+
     >>> s.add_elements(('B', 'N'))
-    
-    
+
+
 We pass the low-loss spectrum to :py:func:`~.hspy.create_model` to include the
 effect of multiple scattering by Fourier-ratio convolution.
 
 .. code-block:: python
-       
-    >>> m = create_model(s, ll=ll)
+
+    >>> m = s.create_model(ll=ll)
 
 
 HyperSpy has created the model and configured it automatically:
 
 .. code-block:: python
-       
+
     >>> m
     [<background (PowerLaw component)>,
     <N_K (EELSCLEdge component)>,
@@ -166,7 +166,7 @@ of them at once. Although in this example the list contains just one component
 this is not generally the case.
 
 .. code-block:: python
-       
+
     >>> N
     [<N_K (EELSCLEdge component)>]
 
@@ -176,7 +176,7 @@ can be configured (see :ref:`configuring-hyperspy-label`). We must enable them
 to accurately fit this spectrum.
 
 .. code-block:: python
-       
+
     >>> m.enable_fine_structure()
 
 
@@ -184,7 +184,7 @@ We use smart_fit instead of standard fit method because smart_fit is optimized
 to fit EELS core-loss spectra
 
 .. code-block:: python
-       
+
     >>> m.smart_fit()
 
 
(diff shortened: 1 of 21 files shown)


hyperspy/tests/model/test_eelsmodel.py
```diff
@@ -4,6 +4,53 @@
...
+    def test_auto_add_edges_true(self):
+        m = self.s.create_model(auto_add_edges=True)
+        cnames = [component.name for component in m]
+        nose.tools.assert_true("B_K" in cnames and "C_K" in cnames)
+
+    def test_gos(self):
+        m = self.s.create_model(auto_add_edges=True, GOS="hydrogenic")
+        nose.tools.assert_true(m["B_K"].GOS._name == "hydrogenic")
+
+    def test_auto_add_edges_true(self):
```
vidartf: Should be renamed to `test_auto_add_background_true`.

francisco-dlp: Fixed, thanks!

doc/user_guide/eels.rst
```diff
@@ -119,29 +119,29 @@ Set some important experimental information that is missing from the original
...
 
 .. code-block:: python
-       
+
     >>> s.add_elements(('B', 'N'))
-    
-    
+
+
 We pass the low-loss spectrum to :py:func:`~.hspy.create_model` to include the
```
vidartf: This one should be updated. (:py:func:`~.hspy.create_model`)

hyperspy/hspy.py
```diff
@@ -97,11 +97,9 @@ def create_model(signal, *args, **kwargs):
     A Model class
```
vidartf: Docstring of hspy.py still has docs for `create_model`, should be removed.

hyperspy/tests/model/test_chi_squared.py
```diff
@@ -28,7 +28,7 @@ class TestChiSquared:
```
vidartf: Import `from hyperspy.hspy import create_model` unused in this file.

hyperspy/tests/model/test_fit_component.py
```diff
@@ -34,7 +34,7 @@ def setUp(self):
         g.sigma.value = 500.0
```
vidartf: Import `from hyperspy.hspy import create_model` unused in this file.

hyperspy/tests/model/test_set_parameter_state.py
```diff
@@ -31,7 +31,7 @@ def setUp(self):
         g2 = Gaussian()
```
vidartf: Import `from hyperspy.hspy import create_model` unused in this file.

hyperspy/tests/model/test_set_parameter_value.py
```diff
@@ -31,7 +31,7 @@ def setUp(self):
         g2 = Gaussian()
```
vidartf: Import `from hyperspy.hspy import create_model` unused in this file.

hyperspy/hspy.py
```diff
@@ -97,11 +97,9 @@ def create_model(signal, *args, **kwargs):
...
-    from hyperspy.model import Model
-    if isinstance(signal, EELSSpectrum):
-        return EELSModel(signal, *args, **kwargs)
-    else:
-        return Model(signal, *args, **kwargs)
+    import warnings
+    warnings.warn(
+        "This function is deprecated and will be removed in HyperSpy 0.9. "
+        "Please use the equivalent `Signal.create_model` method "
+        "instead.", DeprecationWarning)
```
vidartf: I principle I strictly agree with using `DeprecationWarning`, however, it is currently ignored by default (i.e. not printed to console). Are we OK with that?

francisco-dlp: This issue (warnings not being printed) has been fixed in #620 (9a4c307a4b103803c4099c92c47882cbe4cc555e). I thought that the issue was small enough not to be noticed but I was wrong: I should have put that into a separate branch...

francisco-dlp:
@vidartf, thanks for the thorough review: I've just uploaded fixes for all the issues that you have raised.

hyperspy/hspy.py
vidartf: OK, so `warnings.simplefilter('default')` fixes this, but shouldn't it be put somewhere more central than those files?

francisco-dlp: I'm not sure if I get your point. I think that the issue was caused by the fact that [here](https://github.com/hyperspy/hyperspy/commit/9a4c307a4b103803c4099c92c47882cbe4cc555e#diff-e27c75df2dece36803d7c8ccc9166673L11) the filter default was changed to "ignore" and never changed back to "default".

vidartf: Case in point: When I was testing out this PR, I never got a warning while using `hspy.create_model`

vidartf: I think I see the problem: DeprecationWarnings are normally ignored, however when setting the filter to `default`: "This enables default handling for all warnings, including those that are ignored by default". This is quite counter intuitive, but...

vidartf: Ref: https://docs.python.org/2/library/warnings.html#updating-code-for-new-versions-of-python

vidartf: Proposed solution: Run `warnings.simplefilter('default')` somewhere (e.g. hyperspy.hspy).

vidartf: I think I'll make this a separate issue, and merge this. Agreed?

francisco-dlp: Yes, I think that the issue is more general, so I agree that it'll be
better to let this in and fix the general issue in a different PR.
An alternative would be to use numpy.VisibleDeprecationWarning.

vidartf:
PS: Strange that the landscape bot didn't pick up on the unused imports. Is that not something it checks for?

francisco-dlp:
Yes, that's strange: maybe the `create_model` method calls confused it. Anyway, with you who needs landscape ;) ?

vidartf:
> with you who needs landscape ;) ?

I'd recommend everyone to use an IDE with integrated `PEP8`, `pyflakes`, and possibly `pylint`.

PR #626, opened 2015-07-26
BUGFIX: Edited spikes_diagnosis to be able to run on single spectra

### LINKED ISSUE #625: BUG: Spikes removal tool crashes when run on a single spectrum
When running on just a single spectrum, the spikes removal tool will crash when generating the derivative histogram. This is due to an overzealous flattening of the data array, which results in trying to plot a single-valued array as a histogram.

jat255:
Resolves #625.

Added little check to see if signal is one dimensional. If so, plot a histogram of the data without flattening.

I tested this out some and it didn't seem to break anything else, but I'd be happy to incorporate any other suggestions.


hyperspy/_signals/spectrum.py
```diff
@@ -84,7 +84,10 @@ def _spikes_diagnosis(self, signal_mask=None,
             dc = dc[~navigation_mask, :]
         der = np.abs(np.diff(dc, 1, -1))
         plt.figure()
-        plt.hist(np.ravel(der.max(-1)), 100)
+        if len(dc.shape) == 1:
```
francisco-dlp: Good. Actually, taking the max is there in order to decrease the burden on the `hist` function when using the tool on a large number of spectra. However, it's probably not needed (and actually counterproductive) when the number of spectra (n) is below 1000. Here you fix the corner case n =1, but one can argue that this is also broken (although it doesn't raise an exception) for n < 100. So what about:

```python

n = (~navigation_mask).sum() if navigation_mask else self.axes_manager.navigation_size
if n < 1000:
    plt.hist(der, 100)
else:
    plt.hist(np.ravel(der.max(-1)), 100)

```

Also, now that we're at this, it might be worth to replace matplotlib's `hist` with hyperspy's `get_histogram` which actually uses astroML code to automatically determine the number of bins. By default it uses the freedman rule (which is probably a good choice for the purpose), but there is also the option of using non-uniform bin sizes.

to266:
@jat255 , if you could PR this to the 0.8.x branch for tomorrow (including any fixes that you agree with that @francisco-dlp made), that would be great. If you don't have the time to do it, I suggest changing the milestone :)

hyperspy/_signals/spectrum.py
jat255: > So what about:

I like this implementation. I'll be committing a change shortly to match it. I did find in my testing however, that even with a threshold of `n<200`, plotting the bare histogram is not very useful. 

Here are some plots from an EELS spectrum image with `n = 164`. Are these more useful than the "max" histogram? It's hard for me to say.

histogram with bins blocks
histogram with bins freedman
histogram with bins scotts
histogram with bins knuth
histogram with bins 500
histogram with bins 100

francisco-dlp: Could you send the same figures with the max histogram so that we can compare? Also, are there spikes in the data?

jat255: For reference, here's the same data with the threshold changed to `n < 100`. I almost think this is more useful.

Here's a link so I'm not spamming the PR with tons of images:
https://imgur.com/a/SFrTl

francisco-dlp: Are there spikes in your data?

jat255: That's a good point... Not really! I added some:
spike1
spike2

Tests incoming...

jat255: Ok, new histograms: https://imgur.com/a/AfpK6

You can see that since there are just a few spikes, the overall scale of the plot is dominated by the "no spikes" region that is so tall, making it hard to see the spiked points. They are slightly visible in the current implementation (first image), and in a lot of the `get_histogram` methods, they are not visible at all due to the binning.

to266: Change the `counts` scale to log? Should make it considerably more obvious.

francisco-dlp: That's a good point.  I have just realised that, when using the tool I
always zoom in to look for spikes so, why not doing it automatically?
Xould using log y-axis or setting upper y limit to the 10th percentile
help with this?

jat255: Here's how it looks with the semilog scale on Y:
https://imgur.com/a/49XnJ#2

I'll set the ylimit and upload those as well

jat255: The 10th percentile didn't seem to work (on this data set, at least). I'm thinking it could be made more robust if the y limit is set as some function of how many spectra are in the image?

I have to run to a meeting, but here is the code that I've been testing with:

```python
for i in ['scotts', 'freedman', 100]:
    sig = s
    navigation_mask = None
    dc = sig.data
    der = np.abs(np.diff(dc, 1, -1))
    n = ((~navigation_mask).sum() if navigation_mask else
                 sig.axes_manager.navigation_size)
    if n < 200:
        tmp = signals.Signal(der)
        tmph = tmp.get_histogram(bins=i, range_bins=(0.9 * tmp.data.min(), 1.1 * tmp.data.max()))
        ylim = np.max(tmph.data) * 0.1
    else:
        tmp=signals.Signal(np.ravel(der.max(-1)))
        tmph = tmp.get_histogram(bins=i)
        ylim = np.max(tmph.data) * 0.1
    tmph.metadata.General.title = "LogY histogram with bins = " + str(i)
    tmph.plot()
    plt.xlabel('Threshold')
    plt.ylabel('Counts')
    plt.gca().set_title(tmph.metadata.General.title)
    plt.gca().set_ylim(0,(sig.axes_manager.navigation_size * sig.axes_manager.signal_size) * 0.000001)
   # plt.gca().set_yscale('log')
  #  plt.gcf().savefig('C:\\tmp\\2-reg_log_' + tmph.metadata.General.title + '.png')

```

to266: I think by just changing the `ylim` to -1 it should be good enough. Remember this is in log scale still, so leaving it as 0 just cut off any bins with just one count (you can _barely_ see them in the figures you attached... in some cases at least).

francisco-dlp: Thinking about this a bit more, most spikes will show as low count 
outliers (rarely anything about 5 at the current number of bins) at the 
right of the histogram. The problem is to set the default y range so 
that the user doesn't need to zoom in to see them and estimate the 
suitable threshold. So, what about a simple solution like setting the 
top y limit to something like 30 if the highest bin is above 30?

jat255:
I liked the logarithmic approach and as @to266 suggested, setting the minimum to `10 ** -1` makes it so you can see all the spikes fairly clearly. This code uses the `get_histogram()` method with the default settings to get the data to plot. I also changed the labels on the plot a little and the styling. Without any styling, the histogram looks like any other plot, so I thought some distinction might be in order. If there is any opposition, I can change it to be just a regular-looking plot though, no problem.

Here's what it produces on my test data from before:

Single spectrum:
hist_single-spectrum

Area spectrum:
hist_area-spectrum

Area spectrum with no spikes:
hist_area-spectrum_after removal

jat255:
If this looks okay, I'll go ahead and close this PR and open a new one against `0.8.x`

jat255:
Closing to reopen against 0.8.x

PR #629, opened 2015-07-27
Enh sort find peaks1 d ohaver results

magnunor:
Previously they were mostly returned sorted as a function of
position, however in certain cases they would be sorted in
a semi-random fashion. This could potentially lead to unexpected
behaviours.


to266:
Looks good, if you could PR to the 0.8.x branch instead, I'll merge it

magnunor:
https://github.com/hyperspy/hyperspy/pull/634

Issue #628, opened 2015-07-27
spectrum.find_peaks1D_ohaver() when a datapoint is zero


magnunor:
The following code with [this](http://folk.ntnu.no/magnunor/r4535h4t35345nh43n5h43tn5h34tn5h34/dataset.hdf5) dataset does not work:

> s = load("dataset.hdf5")
> s.find_peaks1D_ohaver()

It causes a `hyperspy/misc/spectrum_tools.py:133: RuntimeWarning: divide by zero encountered in log10` and the terminal becomes unresponsive. And HyperSpy uses 100% of one of the CPU cores.

The following works:

> s = load("dataset.hdf5")
> s.data = s.data + 0.0000001
> s.find_peaks1D_ohaver()

francisco-dlp:
Thank you for reporting this bug. Is anybody available to work on this before the 0.8.1 release?

vidartf:
I had a look at the offending code, and have to say I'm not able to follow it. From the docstring this code seems like it has been lifted from somewhere else, perhaps we should look around and check whether there is an existing fix at the source?

francisco-dlp:
@vidartf, I think that the original author of the code, @msarahan, translated the code from the MATLAB code that can be found [here](http://terpconnect.umd.edu/~toh/spectrum/PeakFindingandMeasurement.htm). Those were HyperSpy's early days. Today I would inclined to think that a peak finding routine does not belong to HyperSpy. Of course the same can be said of many parts of the code and, indeed, if we're serious about seeking adoption from beyond the EM community, we should start considering splitting HyperSpy into separate projects. For example: io plugins to a Python microscopy file format repository, EELS code to an EELS repository etc...

magnunor:
This fixes it for my dataset at least: https://github.com/magnunor/hyperspy/commit/3957ea6d907e03daa4714baa965ba55e0669ddf9

But I think we should use something from scipy/skimage/... instead for future versions.

PR #630, opened 2015-07-29
Bugfix for find_peaks1D_ohaver() when a datapoint is zero

magnunor:
Bugfix for https://github.com/hyperspy/hyperspy/issues/628

Send it to 0.8.x?

- [x] Unit test


magnunor:
I added a test, but if the bug occurs the nosetest will basically "hang". I tried this: http://stackoverflow.com/questions/6548496/making-nose-fail-slow-tests but that didn't work either

francisco-dlp:
Moved this to 0.8.2.

francisco-dlp:
@magnunor, could you base this on 0.8.x and resubmit the PR there in time for the 0.8.3 release?

Beside, wouldn't masking the zero values for the fitting step be a more robust fix to the issue, for example by editing line 139 as follows:

```python
yynz = yy != 0
coef = np.polyfit(
                          xxf[yynz], np.log10(np.abs(yy[yynz])), 2)

```

magnunor:
I'll fix this later today.

magnunor:
I rebased this to the current master, and changed the mechanism to your suggestion @francisco-dlp, which seems to be working.

For the future we should think about replacing this function, as discussed here: https://github.com/hyperspy/hyperspy/issues/711

vidartf:
@magnunor: Could you rebase to the 0.8.x branch and resubmit there?

vidartf:
Also, the travis build fails since the test code uses `from hyperspy.hspy import load` instead of `hyperspy.api`.

magnunor:
https://github.com/hyperspy/hyperspy/pull/870

PR #631, opened 2015-07-29
ENH: alignXD() expand keyword

vidartf:
Previously `Signal.align2D()` and `Signal.align1D()` had the options to either crop the image, or leave it untouched. This PR adds a keyword `expand`, which causes the data to be expanded to fill the aligned data. In this way, none of the original data is lost, but a lot of the data set will now be filled with the fill value.

### TODO:
 - [x] Implementation
 - [x] Docstrings
 - [x] Tests


vidartf:
Not sure why coveralls is so moody, but if anything the coverage should go up for this PR as I've added tests for `Signal.align2D()` which was previously missing.

to266:
I have change the milestone to 0.9 for now. If you want it included in the tomorrow release, please PR to the 0.8.x branch and change the milestone. I'll review the code shortly

hyperspy/tests/signal/test_1D_tools.py
```diff
@@ -48,56 +48,71 @@ def setUp(self):
         self.spectrum = s
 
     def test_estimate_shift(self):
-        s = self.spectrum
+        s = self.spectrum.deepcopy()
```
to266: I don't thing the `deepcopy` here (and a couple more times later) is required. The `setUp` is run before every test anyway, to ensure that identical starting conditions are created.

hyperspy/tests/signal/test_1D_tools.py
```diff
@@ -48,56 +48,71 @@ def setUp(self):
...
-        nose.tools.assert_equal(s.axes_manager._axes[1].scale, self.scale)
+        nt.assert_equal(s.axes_manager._axes[1].scale, self.scale)
+
+    def test_align_expand(self):
+        s = self.spectrum.deepcopy()
+        s.align1D(expand=True)
+
+        # Check the numbers of NaNs to make sure expansion happened properly
+        Nnan = self.ishifts.max() - self.ishifts.min()
+        Nnan_data = np.sum(1*np.isnan(s.data), axis=1)
```
to266: I don't think the multiplication is required here

vidartf:
Don't want this for 0.8.1 since it's so close to release, although we might consider it for 0.8.2 if we're doing that.

to266:
Sure, whatever you feel like. Though the two minor comments and the wrong branch are the only reasons why I did not merge.

hyperspy/tests/signal/test_1D_tools.py
vidartf: Strictly not necessary, but done for increased readability. Keep or remove?

hyperspy/tests/signal/test_1D_tools.py
vidartf: Should be fixed.

hyperspy/tests/signal/test_1D_tools.py
to266: I would remove. Your call

vidartf:
Should be ready to merge.

PR #410, opened 2014-12-05
ENH: Added only_current checkbox to ComponentFit

vidartf:
Added a only_current option for the ComponentFit traitsui view.
Implemented it using the CheckListEditor as I could make the label come
on the right of the tickbox, instead of having the label screw up the
layout.

Useful shortcut if you're fitting interactively.

PR #638, opened 2015-07-31
ENH: Added only_current option to ComponentFit

vidartf:
Added a only_current option for the ComponentFit traitsui view.
Implemented it using the CheckListEditor as I could make the label come on the right of the tickbox, instead of having the label screw up the layout.

Useful shortcut if you're fitting interactively (although it could greatly benefit from not replotting during multifit).

Resubmit of #410


to266:
Looks good to me. 

The only strange behaviour I noticed is that "OK" button is essentially "Fit + Close". So if you have a large dataset that takes a minute to fit, run the fit on all pixels first (i.e. `only_current` unchecked) and then want to close the thing, it fits everything again. 

I would prefer to instead have a "close" button at the bottom (essentially just renaming the "cancel") and remove "OK". That might be too many changes at once and with such short time-frame, though. 

if @francisco-dlp agrees, I suggest merging

francisco-dlp:
I agree with @to266 comments. I'll move this to 0.9 and wait for @vidartf to reply.

francisco-dlp:
Sorry, I meant 0.8.2

vidartf:
> I noticed is that "OK" button is essentially "Fit + Close"

This has always been the case, it is simply more apparent when Fit action takes longer. However, I agree that the proposed alternative is better, so I'll push some fixes for that shortly.

vidartf:
Fix pushed, ready for review.

to266:
Looks good, however I guess we'll merge after the 0.8.1 release (unless you want it included after all).

I've never used the `fit_component` myself, what is the usual use-case for it? My first impression is that it would also be useful to enable/disable adjusting position of the component when fitting (in addition to being able to select the range of energies the fit is performed on).

There may be additional functions that make sense to include into this tool, I'm not sure...

Good job anyway

vidartf:
> what is the usual use-case for it?

I'm sure there are several uses, but it basically fits the component in isolation, and for a given/interactive signal range. An example is fast initial fitting of background of EELS data.

Another useful example is if you have a `Spectrum` with several shifting peaks, then if you simply add a `Gaussian` and run `fit()`/`multifit()` there's no guarantee for which peak it will go to. If you then do an initial fit with `fit_component()` you can initialize the component parameters to the peak you want (assuming shifts are gradual), basically trapping it in the local fitting minimum of that peak. I you then run `multifit` it is then much more likely to track the right peak. You might also then want to lock FWHM to help lock-in a given peak.

vidartf:
> I guess we'll merge after the 0.8.1 release 

No worries, then I'll change the milestone of #410 as well.

francisco-dlp:
Given that this is an improvement over the previous situation it makes sense to include it in 0.8.1. I'll merge it if nobody is against it.

PR #639, opened 2015-07-31
ENH: improve usability of spikes tool

jat255:
I noticed a few areas where the spikes tool could be improved, so I made some changes. I think they shouldn't be too disagreeable...

**First:**
On my systems (both windows and linux), the window display would get "squished" when switching to spline mode. This was because the traits window was not re-sizable. See below:
spikes_squished_window
Making it re-sizable fixed this, as the window was allowed to expand, but it wouldn't shrink back when switching the interpolation back to linear. To solve this, I just made the spline controls disabled, as opposed to invisible, so the size of the window does not need to change.

**Second:**
There is poor documentation in the user guide for this tool (it should really be added), so I added some instructions to the tool itself. This was done two ways. First, some parameters were given a more descriptive name. Also, each parameter was given a tooltip that explains the parameter in more detail when hovering over with the mouse. Lastly, I added another button that creates a message window that spells out the specific instructions for running the tool.

**_Tooltip example:_**
spikes_tooltip

**_Instructions window:_**
spikes_instructions

Note, I've never really worked with traits before, so I was just sort of hacking my way through this. In my testing, I didn't break anything, but the Instructions window really doesn't look as nice as I think it should, but I couldn't figure out how to get padding around the text or anything. Any suggestions would be appreciated :smile:

### DIFF
--- a/hyperspy/gui/egerton_quantification.py
+++ b/hyperspy/gui/egerton_quantification.py
@@ -22,6 +22,7 @@ import matplotlib.text as mpl_text
 import traits.api as t
 import traitsui.api as tu
 from traitsui.menu import OKButton, CancelButton
+from pyface.message_dialog import information
 
 from hyperspy import components
 from hyperspy.component import Component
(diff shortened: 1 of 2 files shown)


to266:
I would include two things somewhere in the instructions:
 - that you can (not have to though!) use mouse to change the spike position and width.
 - that "Remove spike" button also finds the next spike in the dataset (so no need to press "find next")

jat255:
I've changed how the instructions are displayed. They now use a `pyface` window rather than a `traitsui` one, which handles the text a little better. The only annoyance here is that the qt backend activates the system bell when showing this window (wx doesn't for what it's worth). 

I would still prefer that the list of instructions is more nicely formatted (_e.g._ hanging indents after each number in the list) but I couldn't figure out how to get that done. I looked into a simple html display using pyface, but couldn't find any examples of how to do that, so I suppose it will stand as is, unless anyone has any other suggestions.

spikes_instructions2

jat255:
> that you can (not have to though!) use mouse to change the spike position and width.

This is a perfect example of why I wrote this. I had no idea you could do that!!

francisco-dlp:
Excellent job @jat255!

ricounet67:
I have an issue with this tool: after performed s.spikes_removal_tool() and displayed s again, the spikes seems not to be removed.

PR #640, opened 2015-07-31
BUGFIX: spike removal histogram improvements

### LINKED ISSUE #625: BUG: Spikes removal tool crashes when run on a single spectrum

jat255:
Continuation of #626, resolves #625.

Reopened against branch `0.8.x`


to266:
I personally like the changes. If @francisco-dlp is in favour, we can merge

Issue #642, opened 2015-07-31
saving lists in hdf5 is slow


to266:
as @francisco-dlp  wrote:
"..writing lists to hdf5 is currently slow and this is an issue because when saving files coming e.g. from the Osiris, the metadata always contain long lists that take several minutes to write even if the data itself is tiny"

to266:
I've looked into this, and the problem is that HDF5 files do not have lists or tuples. 

As the file format supports (essentially) either arrays or dictionaries, at the moment we work around that by creating a dictionary with every element index as the key( `[0, 13., 'asd']` -> `{'0': 0, '1': 13, '3':'asd'}`)
The problem is then that creating dictionaries with so many key:value pairs is inefficient, when compared to saving (numpy) arrays.

The solution I've come up so far is to try to cast any lists or tuples as numpy arrays, and use the inefficient (but faithful) saving scheme only if the dtype is not 'Object'. If it is, instead save the result as an array (which can be done very efficiently and quickly).

The plus side is that it does not break old saved files and numpy-able structures are saved quickly and efficiently. 

The main drawback is that types are all homogenized if possible:

`[('tuple', 1), ('tuple', 2), ['list', 3]] ` -> `array([['tuple', '1'], ['tuple', '2'], ['list', '3']], dtype='|S5')`

but 

`[('tuple', 1), ('tuple', 2), ['list', 3, -30]] ` will stay the same (i.e. numbers are still numbers, tuples - tuples) due to numpy not being able to cast it as [3,2] array.

This makes things confusing. 
We could only use arrays when saving individual lists (i.e ndim=1), however as regular python lists can have mixed types and numpy (and hdf5) do not, there is no way to save those quickly and ensure that all objects are identical and of the original type..

__TLDR:__ Can either save lists and tuples faithfully and slow (what we do now), or "homogenize" them into the most general dtype in the list (i.e. all numbers get turned into strings if there is a string in the list) and save them quickly. 

Thoughts?

to266:
Also, according to [this](http://docs.scipy.org/doc/numpy/reference/generated/numpy.can_cast.html#numpy.can_cast) if numpy always casts according to "safe" rules by default, even changing the datatype actually preserves full precision. So maybe not that terrible, just inconvenient..

vidartf:
Do you have a file I can test with? I have an EDS map from the Osiris in Cambridge myself, but I'm not seeing any performance hit.

to266:
[...]

PR #646, opened 2015-08-02
Enh save lists quick

to266:
The numpy solution, suggested in #642.

Currently will fail for 0d data, as it has not been fixed in the upstream, however two possible fixes are:
#644, #645


francisco-dlp:
On loading lists become arrays:

```python
>>> s = hs.signals.Signal([0])
>>> s.metadata.test = [1, 2, 3]
>>> s.save("test")
The test.hdf5 file was created
>>> load("test.hdf5").metadata.test
array([1, 2, 3])

```

Is this intended?

hyperspy/io_plugins/hdf5.py
```diff
@@ -339,20 +339,40 @@ def dict2hdfgroup(dictionary, group, compression=None):
             group.attrs["_datetime_" + key] = repr(value)
         elif isinstance(value, list):
```
francisco-dlp: It seems that a good fraction of the code for saving lists and tuples is the same, wouldn't it be possible to combine the two?

to266:
Yes, as discussed in https://github.com/hyperspy/hyperspy/issues/642.

It's essentially the price we pay if we want to save them quickly

francisco-dlp:
But, wouldn't it be possible to convert them back to list on loading?

hyperspy/io_plugins/hdf5.py
to266: It is possible, I was just lazy, as we have not decided we are going this way. So.. a PR with the example behaviour, but not quite ready for merge?

hyperspy/io_plugins/hdf5.py
```diff
@@ -339,20 +339,40 @@ def dict2hdfgroup(dictionary, group, compression=None):
...
-                dict2hdfgroup(dict(zip(
-                    [unicode(i) for i in xrange(len(value))], value)),
-                    group.create_group(
-                    '_list_' + str(len(value)) + '_' + key),
-                    compression=compression)
+                try:
+                    tmp = np.array(value)
+                except ValueError:
+                    tmp = np.array([[0]])
+                if tmp.dtype is np.dtype('O') or tmp.ndim is not 1:
```
francisco-dlp: What's wrong with ndim > 1?

to266:
It is possible, sure.
However I think there is no point in converting them back - most of the time you'll want to perform array operations any way. 

Also, if we leave the structures as numpy arrays, it may be somewhat clearer for the user why all floats became strings and similar. If instead we converted them back to regular lists and tuples, then it gets hidden.

hyperspy/io_plugins/hdf5.py
to266: For consistency. If no such exception is made, then [(1,2),(3,4)] will get turned into one array of ndim=2, but [(1,2), (3,4,5)] - into a list of two arrays, each of ndim=1. So here I make sure it's always one-dimensional. Effectively preventing numpy from "eating up" not only the attribute types (float, string), but also outer layer type (tuple, list)

francisco-dlp:
I see your point. However, I think that most people would expect to get back exactly what they've saved. Otherwise, one can write a script that works on some data, save the data in hdf5, load it and find that the script is broken when operating on the newly loaded data.

hyperspy/io_plugins/hdf5.py
francisco-dlp: Very good point. Could you add a little comment to the code to explain this?

to266:
But it will not be the same any way - as I've showed in https://github.com/hyperspy/hyperspy/issues/642, floats will be turned to strings, etc

francisco-dlp:
Ok, but fixing that other (possibly rare) issue is not straightforward, while fixing the (far more common) list/tuple becomes array issue should be beyond reach.

By the way, we should not forget to warn the users in the docs about numpy's homogenizing efforts and the consequences for their data.

to266:
Ok, a fair point. So do we look for other solutions for #642, or go this way? If we do, I'll add the missing bits and clean up the code somewhat, as it's not ready for merge now.

francisco-dlp:
I'm +1 on this solution.

PR #650, opened 2015-08-05
Enh doc tidyup

to266:
- pycharm inspect fixes
 - updated copyright to 2015
 - changed all "Hyperspy" to "HyperSpy"

### DIFF
--- a/hyperspy/_signals/eds.py
+++ b/hyperspy/_signals/eds.py
@@ -1,5 +1,5 @@
 # -*- coding: utf-8 -*-
-# Copyright 2007-2011 The HyperSpy developers
+# Copyright 2007-2015 The HyperSpy developers
 #
 # This file is part of  HyperSpy.
 #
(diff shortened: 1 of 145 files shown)


hyperspy/_signals/eds.py
```diff
@@ -122,7 +122,7 @@ def _get_beam_energy(self):
         units_name = self.axes_manager.signal_axes[0].units
 
         if units_name == 'eV':
-            beam_energy = beam_energy * 1000
+            beam_energy *= 1000
         return beam_energy
```
vidartf: Is it possible for `beam_energy` to be a reference (like an `ndarray`)? If so, this change might mean that you will change the value stored in the metadata: Previously the value of `beam_energy` was copied and reassigned, now it modifies the value "in place".

vidartf:
I went through the diff, and except my comment about the compound operator (might be valid concern for other such replacements as well), it looks OK to me.

hyperspy/_signals/eds.py
to266: That's true. In this particular case it is not relevant, but I will go through and check

to266:
@vidartf thanks for the review. Apparently the `beam_energy` is the most problematic - if it was indeed an array, I'd have to change it back. However it looks like it's always a number, so it's fine

francisco-dlp:
The landscape report points at some genuine remaining problem.

to266:
Attempted to fix the landscape problems, ready for review

PR #526, opened 2015-04-22
set default encoding to utf8

to266:
as per http://stackoverflow.com/questions/21129020/how-to-fix-unicodedecodeerror-ascii-codec-cant-decode-byte general suggestion.

PR #651, opened 2015-08-06
Bugfix polynomial repr

### LINKED ISSUE #472: BUG: polynomial name
when changing the name of a polynomial component the name is not printed

to266:
fixes #472

### DIFF
--- a/hyperspy/_components/polynomial.py
+++ b/hyperspy/_components/polynomial.py
@@ -64,8 +64,10 @@ class Polynomial(Component):
             return x * 0
 
     def __repr__(self):
-        return u'Component <%s order polynomial>'.encode('utf8') % (
-            ordinal(self.get_polynomial_order()))
+        text = u"%s order Polynomial component" % ordinal(self.get_polynomial_order())
+        if self.name:
+            text = u"%s (%s)" % (self.name, text)
+        return u"<%s>".encode('utf8') % text
 
     def estimate_parameters(self, signal, x1, x2, only_current=False):
         """Estimate the parameters by the two area method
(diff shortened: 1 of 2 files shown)


hyperspy/_components/polynomial.py
```diff
@@ -64,8 +64,11 @@ def grad_coefficients(self, x):
             return x * 0
 
     def __repr__(self):
-        return u'Component <%s order polynomial>'.encode('utf8') % (
-            ordinal(self.get_polynomial_order()))
+        text = u"%s order Polynomial component" % ordinal(
+            self.get_polynomial_order())
+        if self.name:
+            text = u"%s (%s)" % (self.name, text)
+        return u"<%s>".encode('utf8') % text
```
francisco-dlp: Why are you using unicode instead of string and why do you need ``encode``?

to266: The unicode - just out of habit.

At some point we some problems with strings (I forget what exactly), and I thought we encoded all `__repr__` as utf8?

francisco-dlp: Unless you can provide an example of where the extra complexity of the code helps here I would suggest to remove the ``encode`` and ``u`` bits.

to266: I'll remove the `u`, but regarding encode, please see here:

https://github.com/hyperspy/hyperspy/pull/526

francisco-dlp: I see, but could you provide an example where ``encode`` fixes
something?

to266: Huh, good point - it will never have to deal with any strings from "outside", hence should be fine. I removed it

Issue #647, opened 2015-08-03
"Remove spike" button in contrast adjustment tool


to266:
Instead of the usual "apply", there is "remove spike" button in contrast adjustment tool (created by pressing "h"  when anything with colorbar is plotted by hyperspy). The behaviour still seems to be the same, however the button name does not make sense

Most likely introduced by https://github.com/hyperspy/hyperspy/pull/639

PR #656, opened 2015-08-06
Bugfix apply button

to266:
#647
spikes_removal_tool has its own buttons


jat255:
Thanks for fixing this. I had meant to get to it this week, but it's tough at a conference... Looks good!

PR #671, opened 2015-08-07
Set default colormap back to gray

### LINKED ISSUE #660: default colormap is jet
Somewhere along the various merges (I guess) the matplotlib preferences were not updated as required

francisco-dlp:
Fixes #660


francisco-dlp:
It doesn't use the `Image.plot(cmap="gray") approach though because that wouldn't work with the navigator.

PR #678, opened 2015-08-10
Centre colormap 

francisco-dlp:
When using divergent color maps one often want the zero at the centre of the cmap. This PR does just that by adding a `centre_colormap` argument to ``Image.plot`` and ``plot_images``.

Example (from the User Guide):

```python

>>> x = np.linspace(-2 * np.pi, 2 * np.pi, 128)
>>> xx, yy = np.meshgrid(x, x)
>>> data1 = np.sin(xx * yy)
>>> data2 = data.copy()
>>> data2[data2 < 0] /= 4
>>> im = hs.signals.Image([data1, data])
>>> hs.plot.plot_images(im, cmap="RdBu", tight_layout=True)

```

divergent_cmap

```python

>>> hs.plot.plot_images(im, cmap="RdBu", tight_layout=True)

```

divergent_cmap_no_centre


hyperspy/drawing/image.py
```diff
@@ -303,9 +320,14 @@ def format_coord(x, y):
             self._text.set_text(self.axes_manager.indices)
         if self.no_nans:
             data = np.nan_to_num(data)
+        if self.centre_colormap is True:
+            vmin, vmax = centre_colormap_values(self.vmin, self.vmax)
+            print vmin, vmax
```
to266: i'd write
```python
if self.centre_colormap:
    #do things
```
Also, why printing?

hyperspy/misc/image_tools.py
```diff
@@ -218,3 +218,38 @@ def contrast_stretching(data, saturated_pixels):
...
+    Returns
+    -------
+    cvmin, cvmax : scalar
+        The values to obtain a centre colormap.
+
+    """
+
+    absmax = max(abs(vmin), abs(vmax))
+    cvmin, cvmax = -absmax, absmax
+    return cvmin, cvmax
```
to266: The second line is not needed, can return it straight away:
```python
return -absmax, absmax
```

francisco-dlp: Fixed.

hyperspy/drawing/image.py
francisco-dlp: Fixes, thanks!

hyperspy/drawing/utils.py
```diff
@@ -500,7 +507,15 @@ def plot_images(images,
...
-        cmap = plt.get_cmap()
+        cmap = plt.get_cmap().name
+    elif isinstance(cmap, mpl.colors.Colormap):
+        cmap = cmap.name
+    if centre_colormap == "auto":
+        if cmap in MPL_DIVERGING_COLORMAPS:
+            centre_colormap = True
+        else:
+            centre_colormap = False
+        print centre_colormap
```
to266: This whole dance with "auto" gets done twice in the code... Any way to combine the two? 

Also `print` here as well

hyperspy/misc/image_tools.py
to266: ...almost. The line above return statement is now not required at all :)
(this one: `cvmin, cvmax = -absmax, absmax` )

francisco-dlp: Sorry, fixed.

hyperspy/drawing/utils.py
francisco-dlp: We could turn the code into a function but, although this is a good
idea in general, in this case I feel that it can obscure the code
because it changes the value of two local variables, `cmap` and
`centre_colormap`.

hyperspy/misc/image_tools.py
```diff
@@ -218,3 +218,38 @@ def contrast_stretching(data, saturated_pixels):
...
+    "bwr",
+    "coolwarm",
+    "PiYG",
+    "PRGn",
+    "PuOr",
+    "RdBu",
+    "RdGy",
+    "RdYIBu",
+    "RdYIGn",
+    "seismic", ]
```
to266: according to [this](http://matplotlib.org/examples/color/colormaps_reference.html), `Spectral` is missing... Though it's possible you skipped it intentionally?

hyperspy/drawing/utils.py
to266: Yeah, after writing the comment I looked more closely, and the code is different enough that combining the two cases would be very confusing. I agree

to266:
Merge conflicts need to be resolved, otherwise LGTM

francisco-dlp:
Done!

PR #683, opened 2015-08-11
Resurrect unfold_if_multidim for 0.8.1

francisco-dlp:
This gives back the functionality to the deprecated `Signal.unfold_if_multidim`. Thanks @vidartf for reporting this issue.


vidartf:
I see this was already merged, but please remember that there were two `unfold_if_multidim` functions that were both deprecated.

to266:
That's a good point, however the other one (in `misc.utils`) is not supposed to be visible by a user, so I think it's fine to leave it as it is...

PR #684, opened 2015-08-11
Bugfix backends

### LINKED ISSUE #681: windows bundle install notebook traits toolkit not set
When installed with hyperspy bundle ([64bit exe here](https://mega.co.nz/#!Yw9wlJjS!7JvnpFfadAQVTyvPEG84PpFReCLuytXcU9cI09QLd84)) (both 32bit and 64bit, and with both traitsui 4.4 and traitsui 4.5), starting hyperspy notebook does not initialize traits toolkit correctly (i.e. it's set to `none`)

matplotlib backend gets initialized fine in all cases.

starting qtconsole configures the toolkit correctly though. 

Importing hyperspy as a library to a ipython session from default profile still sets the toolkit to `none` (so is identicat to starting hyperspy with right-click shortcut)

Maybe related to #435, #584 and similar?

to266:
- no "always on" importing of hyperspy whatever the frontend
 - no `--pylab_inline` or `--toolkit` arguments when starting hyperspy
 - updated preferences screenshot in the user guide
 - fixes #681


bin/hyperspy
```diff
@@ -154,41 +125,10 @@ sys.argv = ipy_argv
...
-for arg in sys.argv:
-    if 'pylab' in arg:
-        add_pylab = False
-if pylab_inline is True and args.environment in ("notebook",
-                                                 "qtconsole"):
-    sys.argv.append('--pylab=inline')
-elif add_pylab is True:
-    sys.argv.append('--pylab=%s' % toolkit.replace("qt4", "qt"))
-
 if ipy_version >= LooseVersion('3.0.0') and args.environment == "notebook":
```
francisco-dlp: Do we need this now that the script doesn't add pylab?

to266: A user might still pass pylab to ipython arguments... So either we leave it as it is (which fixes users mistake), or let the user deal with wrong call himself. Your call :)

francisco-dlp: I think that it is better to leave the user know about their "mistake".

to266: Ok, will remove that. I'll wait for all your comments before pushing though

hyperspy/misc/hyperspy_magics.py
```diff
@@ -28,8 +39,7 @@ def hyperspy(self, line):
         ---------------------
             toolkit
```
francisco-dlp: It would be good to mention that supported toolkits e.g. {"wx", "qt4", "None"}

francisco-dlp: If the user passes an unsupported toolkit the magic should raise a ValueError

to266:
Thanks for the comments @francisco-dlp, should be fixed

hyperspy/misc/hyperspy_magics.py
```diff
@@ -26,10 +37,9 @@ def hyperspy(self, line):
 
         Positional arguments:
         ---------------------
-            toolkit
+            toolkit : {qt4, gtk, wx, tk, None}
```
francisco-dlp: Shouldn't this be: {"qt4", "gtk", "wx", "tk", "None"}

It would also be good to remark that HyperSpy only fully supports qt4 and wx. "gtk" and "tk" do not have gui elements.

francisco-dlp: Passing a wrong flag should also complain with an UsageError.

hyperspy/misc/hyperspy_magics.py
```diff
@@ -14,11 +14,22 @@ def hyperspy(self, line):
...
+        >>> # if toolkit is not "None"
+        >>> %matplotlib [toolkit]
+
+        >>> # run in all cases
         >>> import numpy as np
         >>> import hyperspy.api as hs
-        >>> %matplotlib [toolkit]
         >>> import matplotlib.pyplot as plt
 
         If you pass `-r`, the current input cell will be overwritten with the above specified commands. As a
```
francisco-dlp: Probably this is not needed as it is documented in optional arguments.

hyperspy/misc/hyperspy_magics.py
```diff
@@ -62,29 +69,47 @@ def hyperspy(self, line):
         else:
             toolkit = preferences.General.default_toolkit
 
-        if toolkit not in ["None", "none"]:
+        if toolkit not in ['qt4', 'gtk', 'wx', 'tk', 'None', 'none']:
+            raise ValueError("The '%s' toolkit is not supported.\n" % toolkit +
+                             "Supported toolkits: {qt4, gtk, wx, tk, None}")
+
+        mpl_code = ""
+        if toolkit in ["None", "none"]:
```
francisco-dlp: Shouldn't this be just "None"?

hyperspy/misc/hyperspy_magics.py
to266: No, we should not add the quotation marks - everything after the magic is passed as a string, and if user actually writes `"qt4"` it will fail, as it's supposed to be.

Also, the only recognised flat is `-r` - if anything else is passed, it just stops at toolkit check (where we check if the passed toolkit is supported).

Finally, can we really say that we allow gtk and tk? Or just "it won't crash, but won't work either"? If they don't give full functionality, I see no point allowing them

hyperspy/misc/hyperspy_magics.py
to266: Done by `pylab` example... They have their `--no-import-all` described in both places

hyperspy/misc/hyperspy_magics.py
to266: I'd prefer the freedom to type it all lower-case, but of course can be changed

francisco-dlp: Everything else is case sensitive...

hyperspy/misc/hyperspy_magics.py
francisco-dlp: > Also, the only recognised flat is -r - if anything else is passed, it just stops at toolkit check (where we check if the passed toolkit is supported).

Yes, but the error is not very friendly for users. For comparison this is how the matplotlib magic fails:

```python
In [1]: %matplotlib -a
UsageError: unrecognized arguments: -a

```

> Finally, can we really say that we allow gtk and tk? Or just "it won't crash, but won't work either"? If they don't give full functionality, I see no point allowing them

I think that it is good to keep them because HyperSpy is useful without the GUI elements and someone may want to use it e.g. with the GTK backend for some reason...

hyperspy/misc/hyperspy_magics.py
to266: True. I'll change that, as I said.

I wrote it this way at first because that was the only option that was not passed to the `%matplotlib` magic, hence was not handled by matplotlib (however gracefully)

PR #687, opened 2015-08-11
Fix read the docs build

francisco-dlp:
The HyperSpy docs were not building in Read the Docs because of lack of
pip-req.txt to install the extra dependencies. This also removes no
longer needed sphinx extension package dependencies.


to266:
```python
from . import Release
``` 
in `hyperspy.__init__.py`

and anoter conf.py in `user_guide`

francisco-dlp:
Fixed, thanks!

PR #689, opened 2015-08-11
Fix ETSConfig import

francisco-dlp:
In `hyperpy.gui.__init__.py` ETSConfig was imported from `traits.etsconfig.etsconfig` instead of `traits.etsconfig.api`. It was working but better we get it from the API just in case it gets broken with a future release of traits.


to266:
I guess it has to be merged after #684 (it nothing is broken), as it contains it

Issue #695, opened 2015-08-12
BUG: save broken


pburdet:
```save``` does not work with old .hdf5 files. Due to the type of string, I guess.

```python 

s = hs.datasets.example_signals.EDS_SEM_Spectrum()
s.save('test')

```

```python 

s = hs.signals.Signal(range(10))
s.metadata.set_item('result',[u'Al', u'Cu'])
s.save('a.hdf5', overwrite=True)

```

pburdet:
Also ```save``` becomes way slower saving list of signals.

With 0.8

```python 

>>> %%time
>>> s = hs.utils.stack([hs.utils.example_signals.EDS_SEM_Spectrum()]*100)
>>> s.add_elements(['Al', 'C', 'Cu', 'Mn', 'Zr'])
>>> s.metadata.set_item('result', s.get_lines_intensity())
>>> s.save('test', overwrite=True)
The test.hdf5 file was created
Wall time: 3.05 s

```


With 0.8.1

```python 

>>> %%time
>>> s = hs.stack([hs.datasets.example_signals.EDS_SEM_Spectrum()]*100)
>>> s.add_elements(['Al', 'C', 'Cu', 'Mn', 'Zr'])
>>> s.metadata.set_item('result', s.get_lines_intensity())
>>> s.save('test', overwrite=True)
The test.hdf5 file was created
Wall time: 40.9 s

```

to266:
Regarding the first bug, would you rather keep the strings as unicode and (if you have a very long list of them) save them very very slowly, or convert it to `str` and save it much quicker, but have to deal with the unicode->str conversion at some point if you have any fancy characters?

to266:
I'll fix the slow thing as well, should be just as quick as previous

pburdet:
I'm fine with the second solution. But a warning would be very welcome. I don't understand why at some point hyperspy was saving as unicode characters that could have been saved as as ```str```.

to266:
Previously we saved a list "artificially", but that's very slow with long lists. Now we try to use numpy (i.e. we change lists to numpy arrays) to save much much quicker, but then saving the numpy arrays constrains the dtypes that hdf5 can save. The slow saving was due to
```python
np.array([s,s,s])
```
not returning any errors and instead creating something horrible... And that took a long time with many signals in a list. So now (i.e. in the fix I'll push in 1hour or so, after I've had some food in me) we first check if there are any signals in a list, and only try to "numpy" things if there aren't any.

[...]

PR #696, opened 2015-08-12
Bugfix saving unicode

to266:
solves https://github.com/hyperspy/hyperspy/issues/695

### DIFF
--- a/hyperspy/io_plugins/hdf5.py
+++ b/hyperspy/io_plugins/hdf5.py
@@ -310,10 +310,10 @@ def dict2hdfgroup(dictionary, group, compression=None):
                 compression=compression)
         elif tmp.dtype.type is np.unicode_:
             group.create_dataset(_type + key,
-                    tmp.shape,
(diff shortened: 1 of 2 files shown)


francisco-dlp:
This deserves unittests...

to266:
Added

francisco-dlp:
When there are a mixture of strings and unicode in the list everything is casted to unicode on loading:

```python

>>> s = hs.signals.Signal(range(10))
>>> s.metadata.test = [u"Al", "B"]
>>> s.save('test')
The test.hdf5 file was created
>>> load("test.hdf5").metadata
├── General
│   └── title = u''
├── Signal
│   ├── binned = False
│   ├── record_by = spectrum
│   ├── signal_origin = u''
│   └── signal_type = u''
└── test = [u'Al', u'B']

```

I just wanted to leave this here for the record, I don't think that this is harmful and possibly there is no reasonable way to fix it anyway...

vidartf:
Just wanted to note that this behavior is what would be expected for python3 either way.

PR #708, opened 2015-08-19
 ENH: Add sub-pixel accuracy to `Signal.estimate_shift2D`

vidartf:
Added an argument `sub_pixel_factor` to all functions that estimate shifts for 2D signals. This allows reasonably fast sub-pixel accuracy registration of images.

### DIFF
--- a/hyperspy/_signals/signal2d.py
+++ b/hyperspy/_signals/signal2d.py
@@ -22,13 +22,15 @@ import numpy.ma as ma
 import scipy as sp
 import logging
 from scipy.fftpack import fftn, ifftn
+from skimage.feature.register_translation import _upsampled_dft
 
 from hyperspy.defaults_parser import preferences
 from hyperspy.external.progressbar import progressbar
 from hyperspy.misc.math_tools import symmetrize, antisymmetrize
 from hyperspy.signal import BaseSignal
 from hyperspy._signals.common_signal2d import CommonSignal2D
-from hyperspy.docstrings.plot import BASE_PLOT_DOCSTRING, PLOT2D_DOCSTRING, KWARGS_DOCSTRING
+from hyperspy.docstrings.plot import (
+    BASE_PLOT_DOCSTRING, PLOT2D_DOCSTRING, KWARGS_DOCSTRING)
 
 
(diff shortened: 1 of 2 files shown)


francisco-dlp:
I have only gone quickly through this (I'm very busy with other urgent things and it won't actually have time for a more thorough review until September). The following thoughts came to my mind:

* If now skimage offers xc registration we should consider replacing as much of our feature as possible with theirs as image registration certainly belongs to skimage.
* Although I couldn't verify this in detail, I think that the skimage function uses cross correlation while HyperSpy's implementation has the option to use phase correlation (`normalize_corr=True`). This normalization step actually has an important impact on the outcome so, we should consider adding phase correlation to the upscaled xc or raising an exception if `normalize_corr=True` and `subpixel_factor != 1`.
* skimage should be listed among the dependencies in `setup.py`, user guide's installing hyperspy and `stdeb.cfg`

vidartf:
Strange... I'm trying to add `scikit-image` dependency, but the `distutils` field`reuires` doesn't support dashes, and `install_requires` doesn't seem to go over well with the CI... ?

vidartf:
While some of the issues @francisco-dlp outline earlier still remains, this is now merge with master, and tests successfully. Note that unit tests for the new feature is still missing.

CodeMonkeyJan:
I'm currently on it, I just have to find out how to properly split my current PR #1058 ;-)...

dnjohnstone:
@CodeMonkeyJan that's an easy one :)

in the terminal do the following.... (making sure your master is up to date)

git checkout master
git checkout -b NEW_scikit_image_dependency

then, to see a list of all your branches

git branch

make your changes in the correct branch and push to yourself e.g.

git push origin NEW_scikit_image_dependency

make a PR from that branch to the master

dnjohnstone:
hope that makes some sense

CodeMonkeyJan:
I hope this works and I did not miss anything during the split: #1074

vidartf:
Rebased to be compatible with split, so any code derived from this branch will need to rebase as well!

francisco-dlp:

PR #710, opened 2015-08-23
NEW: add very basic support for FEI Velox software EMD format

magnunor:
Very early support for the FEI Velox EMD file format.

Currently it only works with image files, not spectroscopy. 

It shouldn't be too hard getting the spectroscopy parts working as well, just needs to use the "edx" key as opposed to "images".

I'll see if I can add support for the spectroscopy part during the week.

For testing it:
```python
s = load("vac.emd")
s[0].plot()
```

TODO:
- [ ] Support for EDX/EELS
- [ ] Extract all the useful metadata
- [ ] Check metadata axes dim = data dim
- [ ] Tests
- [ ] Get some minimal test files

### DIFF
--- a/hyperspy/io_plugins/emd_fei.py
+++ b/hyperspy/io_plugins/emd_fei.py
@@ -53,7 +53,7 @@ def file_reader(filename, mode='r', driver='core',
                 exp_dict_list.extend(exp_dict)
         return exp_dict_list
 
-def _get_emd_metadata(metadata_string):
+def _get_emd_metadata_from_dict(metadata_string):
     original_metadata_dict = json.loads(metadata_string)
     axes_offset = original_metadata_dict['_values']['Scale/Spatial/Start']
     axes_scale = original_metadata_dict['_values']['Scale/Spatial/Step']
@@ -92,26 +92,42 @@ def _make_axes(data_shape, axes_dict):
         axes_list.append(temp_axes)
     return(axes_list)
     
-
 def emdimagegroup2signaldict(experiment_key, group):
     image_data_list = []
-    for image_group in group['images'].itervalues():
-        temp_image_data = {'data':image_group['image'].value}
+    if 'images' in group.keys():
+        for image_group in group['images'].itervalues():
+            if 'image' in image_group.keys():
+                temp_image_data = {'data':image_group['image'].value}
 
-        data_shape = temp_image_data['data'].shape
+                metadata_tag_string = image_group.attrs['links']
+                metadata_tag = json.loads(metadata_tag_string)['meta_d']
+                original_metadata = group['metadata'][metadata_tag].attrs['metadata']
 
-        metadata_tag_string = image_group.attrs['links']
-        metadata_tag = json.loads(metadata_tag_string)['meta_d']
+                data_shape = temp_image_data['data'].shape
+                metadata_dict = _get_emd_metadata_from_dict(original_metadata)
 
-        original_metadata = group['metadata'][metadata_tag].attrs['metadata']
-        metadata_dict = _get_emd_metadata(original_metadata)
+                axes_list = _make_axes(data_shape, metadata_dict['axes'])
 
-        axes_list = _make_axes(data_shape, metadata_dict['axes'])
+                temp_image_data['axes'] = axes_list
 
-        temp_image_data['axes'] = axes_list
+                temp_image_data['metadata'] = {}
 
-        temp_image_data['metadata'] = {}
+                image_data_list.append(temp_image_data)
 
-        image_data_list.append(temp_image_data)
+    if 'edx' in group.keys():
+        temp_data = {}
+        temp_spectrum_data = []
+        for spectrum in group['edx']['spectrum'].itervalues():
+            if 'data' in spectrum.keys():
+                temp_spectrum_data.append(spectrum['data'].value)
+            else:
+                if 'compressed_cube' in spectrum.keys():
+                    temp_spectrum_data.append(
+                        spectrum['compressed_cube'].value)
+
+        temp_data['metadata'] = {}
+        temp_data['data'] = np.array(temp_spectrum_data)
+
+        image_data_list.append(temp_data)
 
     return(image_data_list)
--- a/hyperspy/io_plugins/emd_fei.py
+++ b/hyperspy/io_plugins/emd_fei.py
@@ -1,0 +1,117 @@
+# -*- coding: utf-8 -*-
+# Copyright 2007-2015 The HyperSpy developers
+#
+# This file is part of  HyperSpy.
+#
+#  HyperSpy is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+#  HyperSpy is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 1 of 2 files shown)


hyperspy/io_plugins/emd_fei.py
```diff
@@ -0,0 +1,117 @@
...
+        experiments = []
+        exp_dict_list = []
+        for experiment_key, experiment_group in f.iteritems():
+            if isinstance(experiment_group, h5py.Group):
+                exp_dict = emdimagegroup2signaldict(experiment_key, experiment_group)
+                exp_dict_list.extend(exp_dict)
+        return exp_dict_list
+
+def _get_emd_metadata(metadata_string):
+    original_metadata_dict = json.loads(metadata_string)
```
francisco-dlp: This function assumes that the dictionary contains of these keys, what if it doesn't?

magnunor: I'll add a check for that later, and check if the number of axes dimensions in the metadata is equal to the data dimensions. If not, it will generate some "default" axes with scale=1, offset=1, units=''.

dnjohnstone:
Hi @magnunor did you (or anyone else) make any further progress with this? 

I've got a load of .emd files from Velox that I can't open at all (since I don't have Velox myself) so will need to resolve this... I also know a fair number of people who are interested in this.

dnjohnstone:
PS... In particular I'm bothered about EDS spectrum images.

magnunor:
@dnjohnstone the code in this pull request is unfortunately as far as I got. There is basic support for EDS, but I can't remember exactly how well it worked. But you (or anyone) are welcome to fork the code and implement the missing functionality.

If noone else picks this up, I'll probably finish it in 2-3 months time. (Which is true for all my semi-finished issues/pull requests.)

dnjohnstone:
For more info - there seems to be a massive compression of the spectrum images I have compared to what I would expect... Not really sure how the actual data is to be extracted from that - if anyone has any hints it would be appreciated

vidartf:
If FEI runs a compression algorithm on their files, my guess is that they didn't write the compression code themselves. Try sniffing around licenses/DLLs in the FEI software to see which compression library they use. Maybe zlib? :)

MartialD:
I'm not fully sure but I expect the FEI uses the same compression as Bruker does (through the bcf file format). BCF has been implemented recently on hyperspy, I'm looking at that and give you more information soon:  Bcf implementation #712

dnjohnstone:
My suspicion is that you're right about that @MartialD

sem-geologist:
The Bcf is still not implemented... The part of code I uploaded near half year ago dealt just with sfs container (which different kind of bcf uses). I have pure python working prototype, the way I was planing to do at the beging was very slow. I Am working on optimisation, I should update code soon. I spend "ages" hexing throught bcf file, could you upload or share somehow the sample of your format?

MartialD:
Hi @sem-geologist , you can download a bcf file taken with ESPRIT bruker imaging software, version 1.8 or 1.9 I have to check:
The bcf file: http://duchamp.synology.me:5000/fbsharing/D2SrjV2t

The size/type of the raw data (rpl file) are:
key        	 value
width      	 744
height     	 561
depth      	 2048
offset     	 0
data-Length	 1
data-type  	 unsigned
byte-order 	 dont-care
record-by  	 vector

Tell me if you need more information about the file.

sem-geologist:
I have tons of bruker esprit files, I even have some samples from new version 2 Esprit. I just wanted to see what is the difference between your FEI saved and real Bruker saved bcf. The problem about BCF file format/ libraries is few fold:
1. Esprit itself uses bcf file ending to save different kind of containered data, for example not only elemental hypermaps, but also EBSD (electron back scattered diftraction) which tends to be up to hundred GB.
2. under the hud the format is SFS - aidaim softwareTM single file system.
3. i.e. Esprit uses same SFS technology for saving SEM particle analyses, where it gives file ending as *.pan.
--all of such files have the same file signature, which have nothing to do with bcf, but with SFS.

Bruker Hypermaps inside such SFS container are divided to two (version<2) or more (>v2) files: header (XML based file with all SEM/TEM params and basic imagery) and spectral DELPHI / Bruker **packed(!!!)**(_I am not sure who thought up such packing scheme, my knowledge about delphi/pascal is very limited_)  multidimentional array + (for version >=2) simple binary files with pixel pointer table (looks Bruker started to think when doing that ESPRIT 2). Now this packed array is the real pain. And that slowed me down from proper implimentation.
**rant_on** Actually technologically the used array packing is such a nonsense: I just tried to extract such (additionally compressed with zlib 1) bruker bcf (about 300Mb, after extraction should take few GB) and saved to hdf with compression on and it was smaller. More over, hdf is superior, as I can retrieve any chunk from such multi-dimentional array (chunked hdf) any portion much effieciently, while in bruker implimentation it needs to read whole array(<2) or whole spectras (=>2) for selected pixels just for reaching wanted roi, so the reading and doing anything with data is much better in hdf5. Why they (Bruker) would not use better and checked technology?
**_rant_off** 

What I wanted in previous message is simply more details about FEI emd file format, which as I understood from this post, is same/similar to bcf. How do you know it is as BCF? maybe it is just SFS container which looks similar to BCF (the same signature of file). Does it contains some additional data?

sem-geologist:
ok... I found in mailing list those edm velox formats, it is nothing like bcf.... so I cant understand what it have to do with BCF... actually I find that velox stuff very terribly formated, I thought bruker lacked in doing it good. Gosh... after seeing velox I should never again complain about bruker format period....

MartialD:
The emd format from Velox is indeed hdf5 (which is good, but agree the organization inside the file is far from optimal, the medata are condensed in a give string, the subfolders have no readable names, no information on the way edx data are compressed...). By opening such emd file with a hdf5 viewer, you will find an image folder from which you can extract the different images acquired quite easily. If you have edx data, you will also find an edx folder (I'm not sure if the edx folder is always present). Inside edx/spectrum they are 2 subfolders with compressed data (why 2, no idea). I know nothing about the compression of the edx data they have used, but the compression factor is similar to the bcf format and I won't be surprised FEI does not bother and just used the compressed data obtained from the Esprit software, but I can be completely wrong. 

If some people from FEI which know about could give us some advices, it would be appreciated

sem-geologist:
I cant not be sure about what  do you mean "the bcf compression". Bruker uses zlib for compression and bit packing for multidimentional arrays (which I suspect is pascal/delphi feature).
It is really hard to talk about details without proper file sample... Now from the files in some mailing list, exactly file "SI_DF4-HAADF-SuperX_Velox_Data_0000.emd" contain compressed data in edx subfolder. From what I see opening ViTables and looking through attributes, THE_NODE_TYPE=file, but CLASS=earray. Now this is interesting... I know what pytables uses earrays which stands for extensible in one dimension arrays. Maybe Velox uses pytables and not pure hdf5?

sem-geologist:
actually I cant see how this can be called the compression, the file with less than 1000bytes of information with more than 4million 0 bytes, this is nonsense... I could help in searching what kind of compression is in the those velox edm, but I need edm file with known array.

dnjohnstone:
So everyone is aware I've been having a conversation with the lead developer at FEI about this over the past couple of months. It seems they're happy to help and as I understand it he will give me the required information to unwrap the file format in the coming week or so... Then I'll get to it with producing a fully functioning reader. Watch this space.

francisco-dlp:

ericpre:
Superseded by #1662.

PR #712, opened 2015-08-27
NEW: bruker's composite file (bcf) hypermap io (read-only) library

sem-geologist:
The library with reader of bruker composite file (BCF) hypermaps and images.

TO DO:

- [x] header parser, SEM imagery, sum EDS spectra to numpy
- [x] pure python parser functions of elemental mapping to numpy (fuly functioning, but memory inefficient)
- [x] pure python parser with dynamic loading of file and optional dynamic zlib decompression (required to enable cython based parser)
- [x] the hyperspy abstract/maped functions of library (file_reader)
- [x] cython based implementation of Delphi/Bruker packed array fast parsing
- [x] modify the setup.py, to compile cython code
- [x] some minimal loading tests
- [x] integration to hyperspy tests


sem-geologist:
Why checking code quality takes so long?

francisco-dlp:
This is a recurrent issue with landscapeio, but it is unfortunately out
of our control.

sem-geologist:
The python code will not work on big endian machine... excatly between 563 to 581 lines, in function bin_to_numpy. Propositions how to fix it for big endian machines is highly apreciated. Cython code is not going to have this problem...

sem-geologist:
I remember I did asked half year ago about how to return both imagery and hyperspectral data with one file_reader function.... Did something changed in that field? I just want to remind, that bcf contains both sem imagery and hyperspectral data.

francisco-dlp:
The emi file reader also returns several signals as a list. See https://github.com/hyperspy/hyperspy/blob/master/hyperspy/io_plugins/fei.py#L231

sem-geologist:
So in the end, now I have moved everything to classes and saw no slow down of excecution, contrary, I got rid of some parts and saw 1-2% improvement over previously heavy parsing functions siting outside of class. I think lanscape will complain about that function(now method of the BCF_parser class) like a hell for too many variables and complexity. The price of dividing it would be additional slowness (that was my first aprouch, but calling function/method per pixel is expensive (not only calling, but also copying data)). I am not creating something new, the method have to parse binary data with so many quirks and whistles that to make it fast in python is near impossible... actually python is far from best language for doing that and doing it with raw pointers would be much easier and faster (whats why I have the goal to put heavy parsing functions to cython). My profiling showed that the highest time spend is in numpy. Numpy is efficient in reading, saving, doing vectorised math on whole array, but is very slow in doing it element by element, and in this data type that is the case.
My plan is to create the final numpy array, then pass it to cython as memoryview, where accesing and changing the array elements is orders of magnitude faster (what my experimental code do). I really would apreciate if somebody would look through my code and would suggest if it should be changed somehow. I tested it with single bcf. The largest map I have (where non compressed saved as ripple is ~4GB) is parsed in ~6-7minutes (on my overheating amd Turion II 2100Mhz laptop processor). I expect better results on more powerful machines. However as hyperspy main aim is FIB based slicing hypermaps, it tends to be small and have low counts. I specifically thought about it and added downsampling, and energy cuttof arguments for parser. In that way high resolution imagery (Bruker beam controlling board and Esprit software is capable of doing 4096x4096 pixels) and hypermapping can be done simultaniously, then imagery read in full resolution, and hypermap in sensible downsampled resolution. Downsampling as I tested gives no (~1%) overhead coming from the slow numpy inplace addition on selection:
```python
numpy_array[i] += value  # this is 3 times slower
numpy_array[i] = value  # this is slow
```
so in example on downsample ratio of 2, the 2x2 pixels will be squeezed into 1. So, slow assignemet or "inplace" addition to numpy slice is done 4 time per pixel, but there is 4 time less pixels in downsampled array so this counts out. Downsampled hypermaps takes just 2 times less of memory, instead of 4 times. The problem is that python integer cant be added "inplace" to numpy slice with unsinged type. So the final arrays are signed (instead of uint8-> int16, uint16->int32....).
The cut_off by energy I think could be also very valuable. The Brukers Esprit software allows seting the energy range to these: 10, 20, 40, 80kV. I am not very experienced on TEM ( and if somebody have Bruker on TEM, the testing is highly appreciated of this library), but on SEM the most commonly 0-15kV range is most useful, as it lets resolve some overlaps (Y,Nb,Zr,P). The 15-20kV for us have little value as in case of our SEM acceleration voltage limits, or lack of elements in geological speciments for that region. In hypermaps with 20kV range that region normally for us contains nothing except the noise. Another situation I come up with is FIB slicing, where on some SEM the imaging is wished to be done on low voltages (3-7kV) (InLens Detector and ESB (energy selective BSE) mounted in the beam column). Cut off value also will be useful for stacking the hypermaps, where it can be set same for all hypermap files imported, so instead of setting all slices to one of the Esprit setted value (10, 20...) of channels, it can be set to some sane value saving 10-40% of memory.

There is some stuff which makes me woried. If I test my library in ipython:
1.create the instance of the BCF_parser class and assign it to variable
2. del instance, or do variable (which pointed to instance) = 0

the memory is not released and garbage collected... if I import gc, and do gc.collect() it gets then released. Probably I will have to use gc in the file_reader function to cope with this...  It probably points to some badly designed references by me :/ which I can't find, and default gc behaviour do not clean the memory automaticaly.

hyperspy/io_plugins/unbcf_fast.pyx
```diff
@@ -0,0 +1,699 @@
...
+    uint32_t val
+
+cdef struct Gain64:
+    uint64_t val
+
+cdef struct Val16:
+    uint16_t val
+
+cdef struct Val32:
+    uint32_t val
```
vidartf: Can you not use `ctypedef uint32_t Val32`?

sem-geologist: I will try and see if that will work... I noticed some bugs in some not essential code in python code (night work...) and just updated... 
I used structs as I find it much easier  and cleaner to cast pointer to struct than to ctypedef'ed variables... those still will need to be be accesed by Val32[0] (probably, I just got rid of those stucts and '.val', and it do not compile anymore) notation instead of Val32.val, and casting getts complicated... I am very inexperienced with c and cython, and the syntax of python is quite mixing up...  I will see how this will go

vidartf:
I see quite often repeated the following pattern:

```python
if (offset + 4) > size:
    size = size_chnk + size - offset
    buffer1 = b''
    raw_bytes = raw_bytes[offset:] + next(blocks)
    offset = 0
    buffer1 = <bytes>raw_bytes
```

This should probably be extracted out to a separate function. First of all, it should help make it obvious what is happening (if the function is named in a helpful way). Secondly it will make the code easier to maintain.

vidartf:
If I follow the flow of the code correctly, I would consider creating a "stream object" out of this: A simple class/struct with a util function `read_bytes(nbytes)` which is responsible for maintaining the input (`raw_bytes`, `offset`, `buffer1`, blocks/chunks), and just produce raw bytes, then you could separate input fetching from input processing more.

sem-geologist:
well, because it does next(blocks) which is pure python thing, such function would have slow call time overhead. And that is what I want to get away from. 
```python
raw_bytes = raw_bytes[offset:] + next(blocks)
```
is pure python byte string operation. Actually the only python bits in this code, if it would be in c, I for sure would move it to other function (I had so at the begining when was using fread, fopen from c)

sem-geologist:
yeah, just read last one... the same applies here, such separate function could not be optimised due to python function call time. Or would it?

sem-geologist:
If not the optional zlib compression, I would had done exactly that. However because of compressed data  I would have to use external zlib.h, which on different OS would be or would not be installed... using python built in decompression simplifies it for that point.

I cant get cython code working without those structs.... :/ I need to figure out how to use ctypedef with pointers...

vidartf:
@sem-geologist : What I meant would be to have a class that has `blocks` as a member. Then you just access it when you need it. Either way, passing it around shouldn't be a problem since it is just a reference to an iterator?

sem-geologist:
Ok, I will look into those issues... looks quite complicated....

sem-geologist:
I need to put it clear here that bruker packed data is highly unaligned, and reading uint32 or uint16 in many many cases is at addresses not dividable by size. I just realised that whole this unbcf_fast.pyx library would fail on risc (SPARC, ARM, ppc) but not on amd64, x86, so even if I added big-endian support without solving alignenment it is pointless. I have thought about making the separation for input and parsing by of doing the "datastream" class for handling the data, additionally I will add more use of memcpy where it relevant, to make it safe on missalignment unforgiving machines...

sem-geologist:
I just updated the bruker composite file cython parsing code:
I moved the repetative logic into DataStream cython class. I got rid of most of structs, instead I used custom functions to read byte by byte for 16, 32, 64 bits. Theoretically this should make it work on big-endian machines and on machines which are unhappy with missaligned memory. I have no computer to test that. Code got shorter (like ~20%), but my benchmarking shows no significant slowdown. However I left the passing buffer pointer to 6 (2x3: instructed/12bitpacked x 8/16/32bit arrays) functions, as this is more convinient. The triplication of code (for 8,16 and 32bit arrays) is caused due to lack of possiblity cdef'ing in loops. I saw that is quite common practice in other libraries.

 I see the milestone is reset from 0.9.x to 1.0.0...
Things what left to do:
* add the code to the setup.py to compile the unbcf_fast.c during install 
* finish the file_reader function (sem imagery is ready, add the hyperspectral part)

hyperspy/io_plugins/unbcf_fast.pyx
```diff
@@ -0,0 +1,544 @@
...
+            self.load_next_block()
+        self.offset += 8
+        return read_64(&self.buffer2[self.offset-8])
+
+    cdef unsigned char *ptr_to(self, int length):
+        """get the pointer to the raw buffer,
+        making sure the array have the required length
+        counting from the offset, increase the internal offset
+        by given length"""
+        if (self.offset + length) > self.size:
```
vidartf: Can `length` be larger than the size of a single block? If so, you should change this to:
```python
while (self.offset + length) > self.size:
```

vidartf:
> The triplication of code (for 8,16 and 32bit arrays) is caused due to lack of possiblity cdef'ing in loops. I saw that is quite common practice in other libraries.

That is OK, but if you want to address this issue, you could do it with [fused types](http://docs.cython.org/src/userguide/fusedtypes.html) together with using `sizeof()` (to determine byte size of type) and `typeof()`. Something like this?

```cython
cdef element_t fused:
    uint8_t
    uint16_t
    uint32_t

cdef bin_to_numpy(DataStream data_stream,
                      element_t[:, :, :] hypermap,
                      int max_chan,
                      int downsample):

   ... calls unpack12bit(hypermap)]( ... ) ...

cdef void unpack12bit(element_t[:, :, :] dest, int x, int y,
                           unsigned char * src,
                           uint16_t no_of_pulses,
                           int cutoff):
```

Then you would call `bin_to_numpy[uint8_t](...)` or `bin_to_numpy[uint16_t](...)`.

vidartf:
Actually, I think you can simply pass `hypermap` as a type `np.ndarray` ?

hyperspy/io_plugins/unbcf_fast.pyx
sem-geologist: for pixel size (lenght) to be larger than single block it is hardly possible in standard bruker bcf files, as it have hardware limitation up to 4096 channels... but in mapping mode it is reduced even down to 2048 channels (at least in version 1.x bcf) and agressive brukers packing succesfully allows the size of one pixel to be reduced much more bellow that even in intensive mappings. The bcf v1.x have blocks of 4064 bytes if uncompressed, and in bcf v2.x the block size is 0x10000 bytes. If bcf is compressed the compression blocks by bruker is around ~0.5MB.
So I didn't used "while" as there is no real world scenario for that, and I think if can be safely 'if' used here.

sem-geologist:
hmm, fused types, I forgot to research that possibility. Looks promissing :)

np.ndarray? this looks as cython numpy -- the old type of interface of numpy to cython. That is slower than memoryviews, I willl rather check out fused types. Thank you for the review!

sem-geologist:
@vidartf , your suggestion with fused types worked like a charm. I updated the implementation. The only worry is that in cython documentation fused types are marked as experimental.
Not only this cut the size of pyx file in half, made it more readable, but I noticed like few percent improvement in speed (got back to 3.9 sec range for P50_situation.bcf) which is good.
C file however didn't cut at all in size.

That fused types are really cool, I couldn't figure out how to use it exactly before, it makes cython much more useful for me. Thank you again!

vidartf:
Github is complaining about merge issues. If you have Python 3 installed, would you mind merging the current master into this PR? Then you also get to check if everything works on Python 3. If not, I suspect it is only `hyperspy/io.py` that needs to be brought up to date (all the other files are new?), so you could possibly get away with only updating that one.

sem-geologist:
I am using python 3 (python 3.4 on debian exactly) and code are tested on python3 all the time. Sorry for noob question: Is the "PR" the branch? and how to merge the current master into it?

sem-geologist:
probably I am missing something obvious, but hyperspy after instalation do not want to open bcf, they do not recognise that it is bruker bcf, and tries and fails trying to open it as image... I looked throught io.py, I saw nothing there to add the type of files..

vidartf:
In order for the continuous integration to work (the automated checks/tests), you will need to add the `lxml` dependency. Off the top of my head, this needs to be added to `setup.py`, `.travis.yml`, `appveyor.yml` and possibly a debian dependency to `stdeb.cfg`. If you are unsure how to add this, I'm sure somebody can help you with the specifics, but I don't have the opportunity right now. My best bet would be to add it wherever e.g. `natsort` is added?

sem-geologist:
whew... at last it is not complaining about lxml :)

vidartf:
Nice! :) 

Final step would be to add one or more (small) test files and add some test routines. This will likely help you discover any final bugs in the code. If you are unfamiliar with the tests, I would check out the test code for other IO plugins (hyperspy/tests/io/). At the very least run tests to check that it can read successfully files of different types. Possibly, also see what happens if you feed it a corrupt file (e.g. last part of file is missing? Header data does not match actual data?). This is not critical, but can help make the code better.

Finally, since you are still new to many aspects of developing for Hyperspy, any and all input you have to the [developer guide](http://hyperspy.readthedocs.org/en/latest/dev_guide.html#write-tests-documentation) would be very helpful!

vidartf:
PS: CI failing is caused by this issue #798.

sem-geologist:
@vidartf , I have few problems:
1. It seems that travis ci compiles and places code correctly (it seem so looking at travis logs), however I tried to write test using it, but it fails, as it wouldn't be there. For now I commented out the test trying to import unbcf_fast cythonised library. The coverage is quite good. I could add one more bcf file which would increase the coverage more, but unfortunately I don't know what is its license (the Bruker Sample.bcf bundled with they pascal demo library)
2. maybe you noticed that I have some verbose=True set in bcf.py header and later I check it if to print different kind of info. I saw there is new logger thing in hyperspy, I wish either to wrap the verbose, or rewrite printing properly( I saw there are different levels of logging, and my verbose uses just one). Could you give a minimal code example how to use the logger?

despite of those, I think it is really getting to be ready. I prepared documentation and made pull request #934.

to266:
@sem-geologist , as you can see in almost all cases where we use logging, the usage is quite simple:
```python
import logging
_logger = logging.getLogger(__name__)

_logger.debug('Some debugging information that a normal user (i.e. one that does not debug) should NEVER need')
_logger.info('Some information logging. Used to tell non-critical things that might be useful - for example an intermediate result')
_logger.warn('a warning that something happened that we did not expect, but we can continue. The first level that is by default visible in HyperSpy, I believe')
_logger.error('Something went really wrong, and could not perform a function (or similar)')
_logger.critical('very bad, we crashed')
```
More serious examples [here](https://docs.python.org/3.5/howto/logging.html)

vidartf:
@sem-geologist : 
 1) I have no idea why the CI fail to import your library, but I don't have a lot of experience with that. My biggest concern is that if this fails on CI builds, it will likely also fail on user installs, so it is worth figuring out.
 2) Basic use of logging:
```python
# At start of file, get logger instance
import logging
_logger = logging.getLogger(__name__)  # This convention is very helpful

# Where appropriate, use log statements instead of print()
_logger.debug("My log entry!")
a = dict(a=5, b='test')
_logger.info("I have some information: %s", a)
```

E.g. when failing to import extension you can use `_logger.warning()`, and if it succeeds you can use `_logger.debug()` (since user don't care except when something goes wrong).

To get the log output when using hyperspy, do:
```python
import hyperspy.api as hs
hs.set_log_level('DEBUG')
```

to266:
By the way @sem-geologist, the cython-reliant test fails if I just checkout your branch. I tried to compile it manually (`cython unbcf_fast.pyx`), but it did not help - I could not import it. 

To be fair I've never used Cython before, so I may be compiling it wrong, but @vidartf first point is definitely important.

sem-geologist:
@to266  thank You for your trial.
cython unbcf_fast.pyx just converts the code to C.
To compile it you need to use setup.py.
the ```python3 setup.py install``` normally takes care of compiling c code and making *.so (if on *nx) or *.pyd (if on windows) and copying it to right places. I added those into setup.py to do that:
```python
from distutils.extension import Extension  # this is added

with update_version_when_dev() as version:  #this was here
    setup(some_other_stuff,
              ext_modules=[Extension("hyperspy.io_plugins.unbcf_fast",
                               ['hyperspy/io_plugins/unbcf_fast.c'])]
)

```

as you see it even does not check the cython code (it is possible to impliment that it would do that) and cython is not required. I compiles and builds straight from c which is universal.

if you want just to try this library you can (on linux bash):
```bash
$ python3 setup.py build_ext --inplace
```

if without --inplace option, it will not copy compiled library (*.so or *pyd) to the hyperspy/io_plugins.


However looking to logs of appveyroy and travis ci

sem-geologist:
@to266 I disabled cython test for now, try installing it. You can also install thing at user folder with 
python3 setup.py install --user

to266:
You did not have to disable the cython test, as I can comment it out myself.

I also use anaconda and not system python

Just for reference, the command to install hyperspy using pip (including compiling the cython) + keep the ability to modify the .py files is as follows:
```bash
pip install --no-deps -U --force-reinstall -e <hyperspy directory with setup.py>
```

The `--no-deps` is required so that pip does not reinstall all the packages required in setup.py that may or may not be managed by pip to begin with. In my case they are mostly managed by conda, and I want to keep it that way.

At any rate, the current version of tests (only the bcf) segfaults for me. However, I can load the included test files manually fine, and it does use the cython

hyperspy/tests/io/test_bcf.py
```diff
@@ -0,0 +1,29 @@
...
+def test_load():
+    # test bcf from hyperspy load function level
+    # some of functions can be not covered
+    # it cant use cython parsing implementation, as it is not compiled
+    for thingy in test_files:
+        my_path = os.path.dirname(__file__)
+        filename = os.path.join(my_path, 'bcf_data', thingy)
+        print('testing bcf loading from hyperspy level...')
+        s = load(filename, downsample=2, cutoff_at_kV=10)
+        s = load(filename)
```
to266: Why load the same thing twice, but with different options? You also do not check that the loading result is correct - just that it completed and that the dtype is correct. 

I would pick a pixel and record it's non-zero elements here (in a numpy array) for reference, so that you can check if the correct values are loaded.

If not this, you could at least check that the shape of the loaded files is correct

sem-geologist: @to266 , do you mean I should save few pixels on my machine to numpy file, so I could test if it is the same?... hm that means it can be tested if everything works fine on big-endian machines. 

Yes it is loaded twice, because the options or lack of it causes code to go a bit different ways.
In this way test covers larger part of code. I probably should separate it to two test functions, especially if I am going to test some pixels.

to266: Yes, separate into more functions, to explicitly test that the downsampling and cutoff works (both independently).

What I mean is, for example, this:
```python
In [13]: s = hs.load('tests/io/bcf_data/P45_the_default_job.bcf')
spectrum have no dead time records...

In [14]: s
Out[14]: 
[<Image, title: BSE, dimensions: (|100, 75)>,
 <Image, title: SE, dimensions: (|100, 75)>,
 <EDSSEMSpectrum, title: EDX, dimensions: (100, 75|2048)>]

In [15]: np.where(s[-1].inav[37,37].data)                           
Out[15]: 
(array([  45,   46,   48,   49,   50,   52,  100,  102,  103,  155,  183,
         193,  218,  220,  223,  227, 1321]),)

In [16]: locations = np.where(s[-1].inav[37,37].data)[0]

In [17]: s[-1].inav[37,37].data[locations]
Out[17]: array([1, 1, 1, 1, 1, 1, 2, 1, 1, 1, 1, 1, 3, 1, 1, 1, 1], dtype=uint8)

In [18]: values = s[-1].inav[37,37].data[locations]
```

I just randomly picked the (37,37) pixel, obviously.

You can then define the locations and values manually and check that they match (and that all other elements are zero!)

```python
values = np.array([1, 1, 1, 1, 1, 1, 2, 1, 1, 1, 1, 1, 3, 1, 1, 1, 1], dtype=uint8)
locations = np.array([  45,   46,   48,   49,   50,   52,  100,  102,  103,  155,  183, 193,  218,  220,  223,  227, 1321])

```

So no new numpy files are necessary.

sem-geologist:
@to266 , you said that you had segfaults, is your machine big-endian?

is it not simplier to deal with data at numpy level:
```python
bse, sei, hype = s  #unpacking
assert_true(np.all(hype.data[12:22,12:22,212:222] == np.load('the_sameselction.npy')))
```

I dont know what is .inav

to266:
No, it's little-endian.

It is simpler, but I tend to avoid having too many binary files - especially when the result is only a couple of numbers. It's of course up to you.

In the example you also user `assert_true`, but I would use `np.testing.assert_arrays_equal`, or something similar - gives a much cleaner error message when it fails.

Also, [this](http://hyperspy.org/hyperspy-doc/dev/user_guide/tools.html#indexing) is what `.inav` does. Though apparently the documentation for this particular topic (indexing) is massively out of date - now we only support `.inav` and `.isig` indexing, and just calling `s[0]` should result in an error

sem-geologist:
I don't know if you can call dozen of GB of data a couple of numbers. The cython library does very big difference. Try renaming it to switch it off, then you will see how slow is python version. Of course it makes no big difference to open one bcf... (unless you are opening bcf with sizes of >500GB) but to open few hundred bcf... or few thousands. It is difference if job can be done in less then half hour, or takes two days (especially it would be ridiculous if opening the file would take significantly longer than SEM data acquisition). 

I am very confused why cython test fails.. but You told you opened the files and it worked. 

I am confused.... If I try running test function straight in the fresh jupyter notebook -- it crashes, unless I import hyperspy.api as hs to notebook before. This is very very weird.

sem-geologist:
Looks I am struggling with this hard... :/
It was good to do the tests, I found few important bugs (with downsampling, without it everything works fine, I was not aware of those bugs)
1. shape of cython implementation and python implementation was different (there were missing ceil in calculating the shape in cython, preventing to do right shape if dimention would not be divisible by downsample ratio (ie 100 [pixels]/ 3 [downsample ratio]) 
2. python pure parser downsample implementation results in right result only if shape dimentions can be divided by the downsample ratio (i. e. if width is 100 then 2,4,5,10.. will result in correct comparable results with cython), else it is corrupt. Cython version works correctly with any downsample ratio. Cython downsampling was much easier to implement, as in cython we are iterating through array at x,y. Python implementation however generates firstly a flat/1D array, and then it reshapes it to 3D. 

I have few options:
    1. disable downsample completely in pure python implementation (easiest), or get rid of python parsing implementation. Practically I don't see the situation where pure python parser function would be used: the hyperspy dependencies requires C compiler to be present anyway during hyperspy installation. The python based parser function is ugly(landscape complains about it), long and and complicated (due to practical speed reasons). Cython parser is much more readable, and more appropriately subdivided. Removing it completely would allow to shave off also some other code.
    2. inject some filter for downsample keyword to raise error if downsample is not divisible by dimensions (can be user annoying, as user can get to know what is the dimentions of file only after loading the file). This however would be only then unbcf_fast compiled cython library would be missing. The warning could be add during the check if fast library is present.)
    3. find out the way to do it properly (that is quite hard stuff,and I have less and less time for that)

to266:
@sem-geologist, by a couple of numbers I meant the exact example I gave - we don't need to check the full dataset, just one particular (previously probably randomly picked, like in my example [37, 37]) pixel has to be checked. If it's correct, it's much more likely that the rest of the data is correct.
On top of that I would also check the shapes of the loaded signals in the tests.

As for the downsampling, how exactly does it work? Do you average / sum the pixels that you "skip", or just ignore them? Could you point to the particular line that you thing currently breaks in Python implementation? I may have some not too difficult to test ideas.

sem-geologist:
@to266 
I found the bug, and fixed the python pure parser, it was ceil missing (something like ```-(-x_or_y/downsample)```) at the offset calculation. Actually testing on few pixels would made me miss some very nasty bugs. At local repository for testing I use something like 10 different kind of files, and I had moment when 9/10 would go ok, but tenth not, I think the more different kind of files I test the better it will be.
The main test which allowed to troubleshoot the downsample issue were generating the array by cython mode, and pure python, then using ```np.testing.assert_array_equal```. If it was failing I used to plot the difference of such arrays to find out where error is introduced. Looks testing on fixed options on small subset could made me leave those bugs, so I constructed and used something like this on my local machine:

```python
from hyperspy.io_plugins import bcf
from glob import glob
import numpy as np

bcf_list = glob('tests/*.bcf')                     # folder with many bcf's

for i in bcf_list:
    print(i)
    bcf.fast_unbcf = True                          # manually enabling fast parsing
    thingy = bcf.BCF_reader(i)
    for j in range(2, 8, 1):
        print('downsampling:', j)
        hmap1 = thingy.parse_hypermap(downsample=j)    # using cython
        bcf.fast_unbcf = False                         # manually disabling fast parsing
        hmap2 = thingy.parse_hypermap(downsample=j)    #py implementation
        np.testing.assert_array_equal(hmap1, hmap2)

```

and so I find out that with downsampling 2 everything worked fine, with 3 - not, with 4 ok, with 5 ok, with rest - not, then I plotted hmap2-hmap1 and saw that first row was 0, but other was bad, so knew what is causing the trouble.
also I got the shape missmatch reported. So the trouble was that cython was creating wrong shape, but filling ok, segfaulting if width was not divisable by downsample.
Python was creating good shape, but filling values bad. Now those are fixed.

Another thing, what I could get just with one file was dtype overflow. (to add python values inplace to numpy the values have to be signed, else numpy is unhappy). Thing is it was overflowing, but I fixed that by converting numpy array in python parser to unsigned. So now the python and cython functions returns exactly the same values and shapes and dtypes.

Unfortunately if travis ci is not able to import cython library, I can't write such routine test. (Also such test would take too much time).Present tests I think is enought to catch if something is going wrong in the bcf library and that is enought.


downsampling works by sum (not a sum function, but stacking single pixels one by one during the parsing, so it do not use any additional memory, differently than ```block_reduce``` from ```scipy```). Averaging makes no sense, as in EDS we want as much as possible counts.

So I would want to freeze the library, I think now it is really finished (travis is not happy at last commit because of some other test, not mine). The coverage of bcf is shown to be 88% (I cant get more if unbcf_fast can't be imported in travis tests). DOC is also ready (pull requested).

to266:
Well obviously when developing you should use full tests as you just described you did. 

However, once that is done, there should be some way for us to quickly (= not necessarily extensively, since we can't create whatever file we want to) check that the reader is not broken, just like your current tests (most of them at least) do.

The last one in particular (`test_load_8bit`) is strange - it just loads the data and that's it. You could check that the loaded data is correct (e.g. in the same way with the numpy arrays) or at least the shape.

You also only check the shapes / etc of the `hype` signal, and never `bse` or `sei`. You also never check the dtype of the resulting signal, however I'm not sure how relevant that is.

The final check that I can (currently) see missing is the axes - test that all the resulting signals (so including `bse` and `sei`) have the correct axes scales, offsets, names, units, whatever else you can actually read from the file and return for HyperSpy. I guess that is especially important with the down-sampled examples.

sem-geologist:
dtype is irelevant, 8bit arrays are much simpler, they are not effected by endianes, so basically if 16bit works ok, it will also work. It however covers some code which is not covered by 16bit parsing, so if something would deprecate (i.e. numpy as I already experienced) in the future, simple loading will give to know. I dont know what wrong can go with sei, bse or any other sem image shapes, that is simple np.fromstring(decoded base64 string) extremely uncomplicated stuff. The shape is taken straight from the xml header, if xml would be corrupt, nothing would load at all as this is first thing what gets parsed by lxml library.
I will look in testing axes....

sem-geologist:
ok... I don't know how to check the axis? The only place I can find is ```the_example_hyper_signal.axes_manager.as_dictionary()```. It is only place where I can check things?

to266:
Well the `axes_manager` is the object that contains the axes. You can either check by either doing that directly or exporting as the dictionary.

an example would be as follows:
```python

In [2]: s = hs.datasets.example_signals.EDS_TEM_Spectrum()

In [3]: s.axes_manager
Out[3]: <Axes manager, axes: (|<Energy axis, size: 992>)>

In [4]: s.axes_manager[0].scale
Out[4]: 0.020028000000000001

In [5]: s.axes_manager[0].offset
Out[5]: 0.16931499999999999

In [6]: s.axes_manager[0].name  
Out[6]: 'Energy'

In [7]: s.axes_manager[0].units
Out[7]: 'keV'

In [8]: s.axes_manager.as_dictionary()
Out[8]: 
{'axis-0': {'name': 'Energy',
  'navigate': False,
  'offset': 0.16931499999999999,
  'scale': 0.020028000000000001,
  'size': 992,
  'units': 'keV'}}

```
You should not check the `navigate` and `size` - one can be changed and everything would still be valid, and the other is just calculated from the shape of the data.

Also obviously in this case the example signal only has the energy axis, and in your case there would be 2 more - spatial, or as we call in HyperSpy, "navigation" axes.

sem-geologist:
I added the suggested tests. I think it is finished. I downloaded and tried to install this branch on windows7 64, anaconda. It works and uses unbcf_fast.  I cant find what is the reason that tests do not want to import unbcf_fast. Maybe it is some kind of configuration of testing suite ( I have no idea how does it works). However it is visible in the logs of travis ci, and appveyor that it succesfully builds and installs the library with unbcf_fast cython code.
What next?

to266:
could wait for https://github.com/hyperspy/hyperspy/pull/944, then merge master again (so that this is up to date)?

I'll then have a look at this PR once again and either merge it or let you know what to change. Thanks!

sem-geologist:
I merged with upstream so it is  up to date.

sem-geologist:
this is waiting for #949 to be merged, so then it will be possible to get rid of huge c file, and some stuff will need minimal tweak (all those travis ci, appveyor, setup.py...) (not the main code which can be reviewed)

francisco-dlp:
Requires updating with new master

francisco-dlp:
This still requires updating with master.

sem-geologist:
this is still waiting for #949 to be review and merged

k8macarthur:
I tried to use this branch today to import my TEM .bcf files as @francisco-dlp suggested. I come up with two errors: 

1. The first is that it can't find the 'fast' library
`WARNING:hyperspy.io_plugins.bcf:unbcf_fast library is not present...
Falling back to slow python only backend.
2016-07-06 15:31:14.742 python[4477:1952740] modalSession has been exited prematurely - check for a reentrant call to endModalSession:
WARNING:hyperspy.io_plugins.bcf:spectrum have no dead time records...
WARNING:hyperspy.io_plugins.bcf:using slow python parser,
this is going to take a while... please wait`
However I'm willing to believe that this could be a problem with what exactly I've installed.

2. The bcf seems to import as a list rather than a data cube:

`s = hs.load()
In [4]:

s.set_signal_type('EDS_TEM')
s.metadata.General.title = 'Original Data Cube'
s
<ipython-input-4-a1b7fc94e44f> in <module>()
----> 1 s.set_signal_type('EDS_TEM')
      2 s.metadata.General.title = 'Original Data Cube'
      3 s

AttributeError: 'list' object has no attribute 'set_signal_type'`

francisco-dlp:
You can ignore the first warning for now (to get the fast version running you need to compile the Cython code).

Regarding the second issue, could you verify if the HyperSpy are contained in the list? `load` returns a list when there is more than 1 signal in the file.

sem-geologist:
@k8macarthur , 1. this is not an error, but warning for situations like this when library is not fast bcf library is not cythonized for accident or explicit reasons. The whole automatic cythonisation process is waiting to be implemented in #949, and is out of scope of this plugin. For now there is work around: :)
if you can get working cython on your machine then it is duable. Else there is that slow implementation (I guess for TEM it does not so bad as it crunches few MB sized bcf in less than minute :P)

2. bcf files contains everything what bruker Esprit software is recording during hypermapping. If  2-4 or more detectors are enabled in Esprit, they will be saved in bcf together with EDX datacube. Thats why by default the library returns a list!
If you want to load only the hyperspectral cubes without sem images there is flag `record_by` which by deafult is `None`, so it fetches everything from the bcf.
to get spectras only do this:
``` python
s = hs.load(your_filepath_to_bcf*, record_by='spectrum')
```

The reverse can be achieved by using `record_by='image'`.
If you are using new Esprit (>=v2.0.0), and there are few slices inside one bcf, use keyword `index` to reach the slice (default is 0, as in <v2.0.0 Esprit saved everything quite flat).

I am happy the library can be at last useful not only for myself :)

k8macarthur:
`
s = hs.load()
s
Out[3]:
[<Image, title: HAADF, dimensions: (|1024, 1024)>,
 <EDSSEMSpectrum, title: EDX, dimensions: (1024, 1024|1024)>]
`

Sorry, it would help if I printed the list myself first. Oops! Separating out the images and the spectrum lets me open them both separately and they look very sensible on a first glance. For some reason the uploaded file has a beam_current of zero, which might be worrying in terms of compatibility for the quantification function. Part apart from that it looks good.

k8macarthur:
Also, the delay in importing the file doesn't seem to be any longer than importing to dm3 to me.

sem-geologist:
@k8macarthur , beam current is 0, as I saw no Esprit system ask from microscope or record that value into bcf. At least we have no such at zeiss and jeol SEM attached to the bruker EDS system and Esprit demo. And I never saw it recorded into bruker eds files too on those machines, demo files, and even on the Cameca microprobe (where is precise measurement of current present and key hardware piece). It could have change with Esprit 2, but I am not sure as I have none such newer systems in possession. Can you tell which Esprit version you are using? If you want to be sure the beam current is not recorded (or is recorded but not read by my library), would you mind donating bcf sample (can be anything with no scientific value, even the mapped empty void) to the https://github.com/sem-geologist/hyperspy-bcf-testing-samples. I have no bcf samples from any TEM, this would help me to reasure that bruker does/doesn't writes some  metadata to bcf.

k8macarthur:
@sem-geologist I'm not sure what version of Espirit we have but I am very certain we have no way of the beam current being automatically measured. I was asking if the beam_current with a value of 0 will always be in the metadata for imported .bcf files? If so an adjustment for my quantification function will need to be made. I'm sure it will be relatively easy to and a check whether or not beam_current is zero and rule that out as well. I was just curious why it's appeared in the metadata if there is no measure in the original data. However, if some versions of Espirit actually do record the beam current some how that makes more sense.

Happy to donate a .bcf sample if you think it will be of help. I have plenty of maps of single element FIB wedges which are my EDX calibration samples so have no scientific value on their own. They will also have a clean ramping intensity from left to right which should be easy to carry out sanity checks with.

hyperspy/io_plugins/bcf.py
```diff
@@ -0,0 +1,1099 @@
...
+                     'size': obj_bcf.hypermap[index].hypermap.shape[2],
+                     'offset': obj_bcf.hypermap[index].calib_abs,
+                     'scale': obj_bcf.hypermap[index].calib_lin,
+                     'units': 'keV'}],
+           'metadata':
+             # where is no way to determine what kind of instrument was used:
+             # TEM or SEM
+             {'Acquisition_instrument': {
+                  'SEM': {
+                     'beam_current': 0.0,  # There is no technical
```
sem-geologist: TODO: why when I hard coded it?

sem-geologist:
@k8macarthur , yes, the 0 beam current is hard codded, I just cant remember why I set this that way, there is more metadata missing than  I remeber having some problems with tests without it..., I will try to remove it and see how it works as soon the #949 will be merged.

k8macarthur:
Ok brilliant. Let me know if you got the example file ok. It should look a little odd when you import it because I only scan a small line profile region of the whole image, with a scan areas of 64 x 1024. I already know that when Espirit saves such maps they save the whole area with lots of blank space so feel free to crop to make the file smaller. I just haven't gotten to that yet.

sem-geologist:
hmm... sounds a bit weird, cause in my experience it save exactly only the area if selected unless it is from version2, which saves then an overview image of full size, which my library does not read, as I had not enought samples of that new format.... sounds interesting to investigate.

sem-geologist:
for the future, you can use Esprit compression, my library is able to decompress it (tick it when saving bcf in Esprit). It would reduce the space a lot as there is lots of images with 0, and images are saved in internal xml, while only spectral datacube is tightly packed. In example in spectral cube every empty line takes exactly 4 bytes :), while in XML one line few kB. I confirm that bcf you sent is quite weird, it saves 1024x1024 pixels, while as I said on spectral cube have no influence, it have a lot for xml, making it quite too big. It is also is saving 4 images: BF, DF, DF4 and HAADF, where 3 first are completely "empty" (few MB of zeros). Probably those were enabled in Esprit image settings... I have seen such behaviour on our SEM, where we had some Bruker Argus detectors (detectors bundled together with EBSD) that is very much depending from microscope configuration and it does not cause much problem. Thats why I in my library made to not return empty images. I am not sure if this should be counted as feature or a bug. 

Now, there is no beam current or any mention of instrument as TEM, as I expected. There is also no stage information. but probably it is also normal on TEM.

Did you selected the region for 64x1024 in Esprit, or that was region scanned on microscope native software level?

francisco-dlp:
We are planning to release HyperSpy 1.0 and it'll be cool if we could ship it with this.

Although I know that this was discussed above, I am now a bit hesitant to depend on the non-pure-python lxml. Would it be very hard to replace it with a pure python alternative, ideally from the std library?

Here you make an unorthodox use of the `record_by` keyword argument. In all other readers `record_by` sets the signal dimension. However, here it is used to filter what's read from the file. Could you simply rename it to something else? Most likely we'll deprecate `record_by` for `1.0` anyway.

Finally, could you merge the current master into your branch?

Thanks!

sem-geologist:
I don't see what is the problem with lxml, it is same as numpy, scipy or other used libraries underneath the C. problem with other parsers is that I tried them before, but bruker xml is such a overhelmingly deep mess and sometimes it is even randomly broken xml (yeah, I dont know if that is the problem of dephi they use (at least allowing to save in such shitty shape), or programers mistake, but sometimes tags begins with a number! and contains all other crap), and lxml can handle that without a problem. Also bruker xml tends to be overhelmingly pooped with some worthless repetitive stuff which makes it very large. lxml works ok, while with built-in sax, or etree I always had more problems.

francisco-dlp:
It makes sense, thanks for the explanation. However, I would still like to minimise the required libraries that are not pure python. What about lazy importing it and raising a meaningful error when loading a bcf file if the library is not installed? Of course we will ship the windows bundle with lxml and we'll keep it in the anaconda environment but not in `setup.py`.

sem-geologist:
Fine I will add lazy import and take out lxml dependency from setup.py.
But I still don't get it, as numpy and scipy is not pure python too. shouldn't numpy and scipy dependencies also be taken out from setup.py? And by the way something like traitsui, which is in dependency, is impossible to install on ubuntu distros using only package manager while evading gcc and pip (which depends on build-essentials) installation.

francisco-dlp:
It is just a matter of keeping the C python libraries to the bare minimum. HyperSpy can't possibly operate without numpy, so that one is a strict requirement and it'll always be. Actually, ideally we would like to reduce the required libraries which are not pure Python to just numpy. Unfortunately, early in the development of HyperSpy we started depending on traits and traitsui. In hindsight, this was a poor choice and I dream with the day in which we finally get rid of them. We've learnt from our mistake and now we only add C libraries as strict dependencies when it is strictly necessary. In the case of lxml, given that it is used just to read bcf, in my opinion, it is not justified to add an extra C library to the strict requirements as only users of brucker instruments will benefit from it.

sem-geologist:
this fails because it is waiting for #949 to add last missing parts of code (to cythonize and compile)

francisco-dlp:
Could you update this with the current master?

francisco-dlp:
Let us know when this is ready to merge.

sem-geologist:
@francisco-dlp yesterday I looked throught `bye bye record_by`. While I renamed `record_by` which was used not as meant in the kwargs, it is used inside dictionary, which is passed to hyperspy loader. Can you confirm that it can stay in the dict, or should I change it to (is there some alternative/new api to tell the loader that this array is image and this array is eds signal raster/whatever)

francisco-dlp:
If your reader doesn't read anything signal_dimension > 2 or signal_dimension == 0 it can carry on using record_by internally. `record_by` is removed from metadata afterwards. [This is the code](https://github.com/hyperspy/hyperspy/pull/1127/files#diff-d5709478406f468e29ddb62fcc41f709R332). I took this approach because, although `record_by` is not general enough for hyperspy, it is general enough for most file formats. In this was we avoid the danger of introducing bugs by touching all the io readers...

sem-geologist:
I dont get how much dimentions is what: image (BSE, SEI whatever, 1 channel) 2 dimentions (x, y)?
eds hypermap 3 dimmentions (x, y, Energy)? Aslo I can think 2 dimentions as  crossection (x, Energy), one dimention - eds from one point (Energy). What is 0 dimentions? If I get this clear then my hypermaps is >2 and I am in the trouble... how to tell the loader what is what?

francisco-dlp:
You got it right. 0 signal dimension would be an scalar e.g. mapped across a 2D space:

```python

>>> scalar = hs.signals.BaseSignal(np.random.random((32, 32)))
>>> scalar.axes_manager.set_signal_dimension(0)
>>> scalar
<BaseSignal, title: , dimensions: (32, 32|)>

```

sem-geologist:
@francisco-dlp  I am very confused, tests cant pass throught on bcf and it complains about record_by. I got reid of it completely (also in dictionary which is passed to loader) and the test is still failing, What did changed in api and what do I need to change for it to work back?

francisco-dlp:
Notice that #1127 has not been merged yet, that's probably the issue: you can't get rid of `record_by` without #1127.

sem-geologist:
@francisco-dlp I see, it is not merged, however I tried to merge bcf implementation with #1127 on separate branch: the previously failing tests on `record_by` are gone... BUT it apears scale it introduces scale to be broken!
```
AssertionError: 0.010001 != 4.368392531e-05 within 12 places
```
which looks for me that axes got mixed again (instead of x or y scale it took the energy scale) (I had problem with this at the step of finalizing this io_plugin before)... :/ (maybe because it does not know what is what? as we don't tell it anymore...)

francisco-dlp:
That makes sense. I think that for your reader it is ok to keep record_by as it doesn't need the more general approach. I would simply revert the latest commit.

francisco-dlp:
This is still failing. We'll probably release HyperSpy 1.0 today or early tomorrow, any chance of fixing this before the release? If not, no worries, HyperSpy 1.1 will probably be released in a few weeks anyway given the number of almost ready PRs in the pipeline.

sem-geologist:
does #1127 got merged?I tested this with it merged on separate branch (fix_bcf) and it works. if `remove record_by`  will be merged on now then after i can udate this to master and it should stop throwing errors without any more tweeking. basicaly it is ready for review, as nothing is gona change

francisco-dlp:
I just merged #1127.

francisco-dlp:
Once Travis and Appveyor are satisfied ,I'll happily merge this.

sem-geologist:
@francisco-dlp , the appveoyr and travis just passed :) . The `landscape` probabably is hanged.
Do not forget documentation #934, which I updated and merged with master today.

francisco-dlp:
Reverse engineering this format is an amazing achievement. Thanks a lot for contributing your work to HyperSpy.

sem-geologist:
Indeed, it gives a lot of satisfaction. I look now to any binary format with a bit of hope :D that one day... I will rule it also.

Issue #713, opened 2015-08-27
HyperSpy doesn't start in IPython 4.0


to266:
IPython 4.0 involves the "Big Split", which means `notebook` and `qtconsole` are no longer ipython, but rather Jupyter parts.

As a result, the hyperspy starting scripts no longer work (as they rely on starting `ipython [notebook/qtconsole] --profile=hyperspy`), as not only the objects have been moved, but also jupyter does not have profiles.

to266:
I will mention that yet another way to solve the issue (not the #714 ) would be according to [this](https://groups.google.com/d/msg/jupyter/7q02jjksvFU/MLE3Xw1ADQAJ), which would effectively replace the profile when jupyter has to be run....

I personally think the extension solution is much more elegant / consistent, but I may be biased having had to deal with all the profile files when transitioning to IPython 3.0 :)

PR #714, opened 2015-08-27
New IPython extension to start HyperSpy (works with Jupyter/IPython 4)

to266:
Allows to use hyperspy as an ipython extension. 

It's really one of the ways how to transition to IPython 4, where notebook and qtconsole are moved to Jupyter, which does not have profiles (i.e. one of possible ways to solve https://github.com/hyperspy/hyperspy/issues/713)

PROS:
 - HyperSpy no longer requires starting script and profile, hence instead of `hyperspy notebook` the user will have to use the much more general (and I assume preferred) `jupyter notebook`, and have the full access to the rest of its functions
 - Very easy to maintain (we are no longer required to maintain the profile)

CONS:
 - Additional line has to be executed every time you want to access HyperSpy:
```python
%load_ext hyperspy
%hyperspy
```

It's yet another issue, where it would be useful if people stated their preferences, hence:
@hyperspy/developers, @jat255

### DIFF
--- a/doc/user_guide/getting_started.rst
+++ b/doc/user_guide/getting_started.rst
@@ -47,8 +47,6 @@ can be started as follows:
 
     $ hyperspy notebook
 
-
-
 .. versionchanged:: 0.8.3
 
 Once a new Jupyter notebook is started in a browser window, HyperSpy needs to
@@ -58,6 +56,18 @@ be loaded and started using the :ref:`hyperspy magic <magic-label>`.
 
     >>> %load_ext hyperspy
     >>> %hyperspy
+
+.. note::
+
+    As per `IPython extension documentation
+    <https://ipython.readthedocs.org/en/stable/config/extensions/index.html#using-extensions>`_,
+    the first line, loading HyperSpy extension, can be avoided typing every time by adding it to your default extensions:
+
+    .. code-block:: python
+
+        >>> c.InteractiveShellApp.extensions = [ #some other default extensions
+                'hyperspy',
+                ]
 
 
 Alternatively, hyperspy can be run in two console modes: the terminal you run
--- a/doc/user_guide/getting_started.rst
+++ b/doc/user_guide/getting_started.rst
@@ -49,13 +49,14 @@ can be started as follows:
 
 
 
-.. versionadded:: 0.8.1
+.. versionchanged:: 0.8.3
 
 Once a new Jupyter notebook is started in a browser window, HyperSpy needs to
-be imported using the :ref:`hyperspy magic <magic-label>`.
+be loaded and started using the :ref:`hyperspy magic <magic-label>`.
 
 .. code-block:: python
 
+    >>> %load_ext hyperspy
     >>> %hyperspy
 
 
@@ -160,7 +161,7 @@ menu.
 HyperSpy IPython magic
 ^^^^^^^^^^^^^^^^^^^^^^
 
-.. versionadded:: 0.8.1
+.. versionchanged:: 0.8.3
 
 HyperSpy provides an IPython magic in order to ease the task of setting up the
 typical interactive data analysis environment. The magic imports the HyperSpy API,
@@ -170,10 +171,12 @@ automatically.  However, starting with IPython version 3.0, auto-importing and
 auto-configuring is discouraged (for very good reasons). In order to comply
 with the new recommendations without loosing convenience we supply the
 ``hyperspy`` IPython magic that performs the same operations in a more explicit
-and transparent way.
+and transparent way. As of IPython version 4.0 and HyperSpy 0.8.3, HyperSpy has
+to be loaded before the magic becomes available:
 
 .. code-block:: python
 
+    >>> %load_ext hyperspy
     >>> %hyperspy [-r] [toolkit]
 
     HyperSpy imported!
@@ -202,11 +205,7 @@ without any magic commands (e.g. included in a script).
 
 .. note::
 
-        The magic is only available when using HyperSpy's IPython profile. The :ref:`starting script <starting_hyperspy_label>` automatically starts IPython using the HyperSpy profile. If you prefer you can do it manually as follows:
-
-   .. code-block:: bash
-
-      $ ipython --profile=hyperspy
+        The magic is available when either using HyperSpy's IPython profile (:ref:`starting script <starting_hyperspy_label>` automatically starts IPython using the HyperSpy profile, however is deprecated starting IPython 4.0) or having loaded the HyperSpy as an IPython extension
 
 
 Getting help
--- a/hyperspy/__init__.py
+++ b/hyperspy/__init__.py
@@ -28,3 +28,8 @@ from . import Release
 
 __all__ = ["api"]
 __version__ = Release.version
+
+
+def load_ipython_extension(ip):
+    from hyperspy.misc.hyperspy_magics import HyperspyMagics
+    ip.register_magics(HyperspyMagics)


jat255:
I would agree that moving things to be compliant with Jupyter standards is a good idea. I've seen more and more people using Jupyter as a computing environment, so I think going forward more people will be familiar with it and how you are supposed to use it. 

I still find it useful to fire up a qtconsole every now and again for quick looks at things (while the notebook I reserve for more in-depth investigations). This will still be possible, correct? It is just a matter of loading the extension in both? 

Is it possible to have a little script that could be supplied that would load the hyperspy extension automatically? I'm just thinking out loud, but this could be provided as a way to keep the experience more or less consistent for people that are used to the older behavior. I also see the benefit of forcing people to move to the new method, though.

to266:
Yes, both `qtconsole` and `notebook` will still be available, they're just moved to Jupyter, which is the "programming-language-agnostic" framework. Hence from the `notebook` you will be able to launch any kernels you might have installed (for me it's Python2, Python3 and Julia).

Strangely enough the default `jupyter qtconsole` always starts the system default (python3.4) qtconsole.. However I tend to use just normal ipython console (which, confusingly, can still be started as previously, as it is separate from the Jupyter) for quick investigations, so I only noticed it now

According to https://ipython.org/ipython-doc/dev/config/extensions/index.html#using-extensions, the extension can be added to be loaded with all ipython kernels, however I would leave that for the user to do himself/herself... I.e. I would prefer such convenience functions not be "automatic", and instead just supply explanations what and where to add.

I also just now realised that with Jupyter and IPython config files becoming separate, it's not clear at all how to autoload extensions in a notebook...

jat255:
Hi Tomas, when I test this out on my system, I get:

```python
%load_ext hyperspy

Exception occurred in traits notification handler for object: <hyperspy.defaults_parser.GeneralConfig object at 0x7f38ea1a18f0>, trait: logger_on, old value: False, new value: True
  File "/usr/lib/python2.7/site-packages/traits/trait_notifiers.py", line 340, in __call__
    self.handler( *args )
  File "/home/josh/git_repos/hyperspy/hyperspy/defaults_parser.py", line 126, in _logger_on_changed
    turn_logging_on()
  File "/home/josh/git_repos/hyperspy/hyperspy/misc/ipython_tools.py", line 85, in turn_logging_on
    "# ============================\n")
  File "/usr/lib/python2.7/site-packages/IPython/core/logger.py", line 199, in log_write
    write(data)
TypeError: must be unicode, not str
ERROR:traits:Exception occurred in traits notification handler for object: <hyperspy.defaults_parser.GeneralConfig object at 0x7f38ea1a18f0>, trait: logger_on, old value: False, new value: True
  File "/usr/lib/python2.7/site-packages/traits/trait_notifiers.py", line 340, in __call__
    self.handler( *args )
  File "/home/josh/git_repos/hyperspy/hyperspy/defaults_parser.py", line 126, in _logger_on_changed
    turn_logging_on()
  File "/home/josh/git_repos/hyperspy/hyperspy/misc/ipython_tools.py", line 85, in turn_logging_on
    "# ============================\n")
  File "/usr/lib/python2.7/site-packages/IPython/core/logger.py", line 199, in log_write
    write(data)
TypeError: must be unicode, not str
```

My installation is currently:
```
hyperspy -> up to date with to266/ENH_extension
jupyter -> 4.0.4-2
python2-traits -> 4.5.0-1
python2-traitsui -> 4.5.1-1
```

to266:
Hi Josh,

Please check that you start the correct ipython kernel when using jupyter. I had them mixed up at first (as it usually starts with the system default python), and had to add (in my case) anaconda kernel in order to use the jupyter.

As for the error, it's not one of the things I've touched in this PR. My guess is that's the first bit in hyperspy that is not compatible with Python3.*, and it starts by default. Alternatively, you could try to start ipython console directly (without jupyter), and change the settings to check that the logging is off, and try again...

Sorry for the vague answer, just wanted to quickly list my thoughts, and I'll have a better look tomorrow :)

jat255:
Hmm, I just use the system installation, so it appears to be finding it correctly, as it's running python2 and not 3. I seem to be having other issues with Jupyter though, so I would disregard this, as I'm guessing it's a local install problem.

thomasaarholt:
Updating to the latest Jupyter (and running either 
`jupyter notebook` or `python notebook`
and then running 

```
%load_ext hyperspy
%hyperspy
```
gives the error 

**The hyperspy module is not an IPython extension.
ERROR: Line magic function `%hyperspy` not found.**

I attempted to reinstall hyperspy via `pip`without success.

I also tried installing it by `%install_ext https://raw.githubusercontent.com/hyperspy/hyperspy/a46e3f12de78e04378378fe8880e12775490c736/hyperspy/misc/hyperspy_magics.py` but that didn't work either.

to266:
This branch has not been merged, as you can see.

To enable the hyperspy as an extension, you have to either merge it in your own local repository, or just manually add the three lines.

Also, as I've already mentioned somewhere, please check if the python kernel that starts is the correct one (if you use anaconda). To do that, see if you can run 
```python
from hyperspy.misc.hyperspy_magics import HyperspyMagics
ip.register_magics(HyperspyMagics)
```
If that works (and hence afterwards you should be able to use the `%hyperspy` magic), you probably did not merge this branch locally. 

If it does not, most likely by default you start a different kernel
Let me know either way :)

dnjohnstone:
I think this is a very good idea and should be merged ASAP...since  #713 is incredibly annoying for people who don't want to worry about wider Python developments.

This solution seems very simple, and is similar to what people will have been getting used to with starting HyperSpy via %hyperspy. I pulled this branch to my local machine and it ran exactly as expected (see screen shot below, where I first demonstrated that %hyperspy fails and then that this works).

load_ext

Does this need docs / tests? Is that something I can help with or will you do it @to266?

to266:
...no, I think it was just a question if that's the way we want to start hyperspy. I just merge this branch into everything I work with as well, so I'm keen to stop doing it as well...

@francisco-dlp , any reason why we should delay it?

francisco-dlp:
Although it might be a minor drawback, I dislike having to add the extra line to get this working. With the old ipython just executing the script was enough to start working. With 3.0 we needed the magic, and now with 4.0 an extra line to load an extension that all it does is enabling the magic.

I think that there must be a better way. For example, what about:

```python
import hyperspy.api_ipyqt as hs

```

This has the advantage of being one single line that contains just a standard import, something that most users are familiar with.

to266:
To be fair, with old hyperspy we __had__ to have a script. With 3.0 we still needed the script __and__ had to add the magic. Now with 4.0, we no longer need the script or, for that matter, hyperspy profile - just the extension loading and magic is enough. The one place where it was particularly convenient - when running parallel ipython kernels. I did not have to bother with profiles, as just running these two likes was enough.

The single-line imports, I guess, can be done, however as a user I would prefer the magics. The auto-completion would help, but the long import still seems clunky.

dnjohnstone:
I agree that one line is better than two but I do wonder (as ridiculous as this will sound to some) if we can change things so that it's something a little cleaner than .api_ipyqt which could read as a string of random letters to some users.

vidartf:
Can we not have hyperspy install a permanent extension into IPython during the install of hyperspy? I'm a bit hazy on the details, but from what I can gather from a quick browse, it should be possible. Then you'd only have to run `%hyperspy` when you wanted to use it in IPython. Citing [the IPython docs](http://ipython.readthedocs.org/en/stable/config/extensions/index.html):
> To load it each time IPython starts, list it in your configuration file:
> ```python
c.InteractiveShellApp.extensions = [
    'myextension'
]
```

**Comment by to266** [comment]
Yes you can, exactly as you described. I personally don't like that, mostly because I did not find a simple way to add an extension to the user profile - you just have to find and edit the file manually. So if the user has the whitespaces mixed up or something we did not foresee, things get broken in a very inconvenient way.

**Comment by vidartf** [comment]
Well, then I think the solution is to leave it as a two-line load, where the end-user has the option to change their config in order to make it a one-line load.

**Comment by francisco-dlp** [comment]
Ok then. Once this gets documented in the User Guide it'll be ready to merge.

**Comment by to266** [comment]
Documentation added. As always, suggestions, corrections and comments welcome

File: doc/user_guide/getting_started.rst [hunk 1]
```diff
@@ -49,13 +49,14 @@ can be started as follows:
 
 
 
-.. versionadded:: 0.8.1
+.. versionchanged:: 0.8.3
```
vidartf: I'm trying to learn Sphinx here, so I'm trying to figure out what this does. I get why you'd want to add the `versionchanged`, but why remove the `versionadded`? What is the end-result of these tags anyway?

to266: Never looked into that too much myself. I suppose the way we actually use it is just having the text there. Also, if we ever end up using it, [this](http://sphinx-doc.org/builders.html?highlight=changes#sphinx.builders.changes.ChangesBuilder)

vidartf:
@to266 : Do you want to add to the documentation that `load_ext hyperspy` can be avoided by modifying the config to have:

```python
c.InteractiveShellApp.extensions = [
    'hyperspy'
]
```
?

to266:
Done, sorry for the delay

francisco-dlp:
This was deprecated before release by #819

PR #718, opened 2015-09-04
ENH: store model (hence enable saving models)

to266:
Adds ability to store and save models in signal

new `Model` methods:
   - `store(name)` - stores the current state of the model to signal
   - `save(filename, name)` - stores the model and immediately saves the signal

new things in `signal`:
   - `signal.models` - the new object that hold all stored models. 
   - `signal.models.remove(name)` - removes the stored model
   - `signal.models.restore(name)` - returns a new recreated model
   - `signal.models.<name>.remove()` and `signal.models.<name>.restore()` are also available

General things:
   - When storing, the model names are either given, or generated using letters (a, ..., z, aa, .., az, ba, ..., bz,...).
   - `signal.models` is not copied on deepcopy
   - as a consequence, `signal.models` are not sliceable - I think only actual objects should be "sliceable", not their stored copies. If a model needs to be sliced, it has to be done on "live" copy. 

__________
Usage example:
---

```python

In [2]: s = hs.signals.Spectrum(range(100))
In [3]: s.models
Out[3]:
In [4]: m = s.create_model()
In [5]: s.models
Out[5]:
In [6]: m.append(hs.model.components.Lorentzian())
In [8]: m.store()
In [9]: s.models
Out[9]:
└── a
    ├── components
    │   └── Lorentzian
    ├── date = 2015-11-22 18:32:25
    └── dimensions = (|100)
In [10]: m1 = s.models.restore('a')
/home/to266/dev/hyperspy/hyperspy/component.py:612: RuntimeWarning: `active_is_multidimensional` already False for
 Lorentzian
  (str(value), self.name), RuntimeWarning)
In [11]: m1.components
Out[11]:
   # |            Attribute Name |            Component Name |            Component Type
---- | ------------------------- | ------------------------- | -------------------------
   0 |                Lorentzian |                Lorentzian |                Lorentzian
In [12]: m2 = s.models.a.restore()
In [13]: m2.components
Out[13]:
   # |            Attribute Name |            Component Name |            Component Type
---- | ------------------------- | ------------------------- | -------------------------
   0 |                Lorentzian |                Lorentzian |                Lorentzian
In [14]: m2 is m1
Out[14]: False
In [18]: m2[0].A.map
Out[18]:
array([(0.0, nan, False)],
      dtype=[('values', '<f8'), ('std', '<f8'), ('is_set', '?')])
In [19]: m1[0].A.map
Out[19]:
array([(0.0, nan, False)],
      dtype=[('values', '<f8'), ('std', '<f8'), ('is_set', '?')])
In [20]: m1.spectrum is m.spectrum
Out[20]: True
In [21]: m1.store()
In [22]: s.models
Out[22]:
├── a
│   ├── components
│   │   └── Lorentzian
│   ├── date = 2015-11-22 18:32:25
│   └── dimensions = (|100)
└── b
    ├── components
    │   └── Lorentzian
    ├── date = 2015-11-22 18:34:27
    └── dimensions = (|100)
In [23]: s1 = s.isig[::2]
In [24]: s1.models
Out[24]:
In [25]: m.save('my_file')
The my_file.hdf5 file was created
In [26]: l = hs.load('my_file.hdf5')
In [27]: l.models
Out[27]:
├── a
│   ├── components
│   │   └── Lorentzian
│   ├── date = 2015-11-22 18:32:25
│   └── dimensions = (|100)
├── b
│   ├── components
│   │   └── Lorentzian
│   ├── date = 2015-11-22 18:34:27
│   └── dimensions = (|100)
└── c
    ├── components
    │   └── Lorentzian
    ├── date = 2015-11-22 18:35:02
    └── dimensions = (|100)
In [28]: l.models.c.remove()
In [29]: l.models.b.remove()
In [30]: m3 = l.models.a.restore()
In [31]: m3[0].A.map
Out[31]:
array([(0.0, nan, False)],
      dtype=[('values', '<f8'), ('std', '<f8'), ('is_set', '?')])
In [32]: m3.components
Out[32]:
   # |            Attribute Name |            Component Name |            Component Type
---- | ------------------------- | ------------------------- | -------------------------
   0 |                Lorentzian |                Lorentzian |                Lorentzian

```
__________

TODO:
---
 - [x] add tests
 - [x] add the rest of documentation
 - [x] Automatically save current model when loading other stash? Not sure if needed, but would prevent data loss with accidental stash loading..


vidartf:
Nice! But does `pop()` apply the removed model as well? If so, I'm not sure if I see a usage scenario, and would argue that a `remove()` function might be better.

to266:
Yes it does... 
I was essentially following git stash operations.

I agree that `remove` is necessary (I'll add that, thanks for suggestion), but I'll leave the pop for the moment, as sometimes a user would not want to keep two copies of the same thing

to266:
Ready for merge.
It's not quite compatible with https://github.com/hyperspy/hyperspy/pull/720, but I will fix it when either one is merged

francisco-dlp:
I'll wait until the conflicts are fixed to review it.

to266:
The conflicts are resolved, however there are a couple of questions that we've discussed a bit already, so I will just write it out the best I can, and hopefully others will have something to contribute too:

1. __stash should not be saved in metadata__, as it gets copied whenever we use deepcopy. Instead, create a `DictionaryTreeBrowser` named `Analysis` and put it in the signal. That way it's still just as easy to reach, does not get mixed up with metadata, and does not get copied

2. __Saving a model should follow the same basic pattern (from the user's POV) as saving a signal__ - the thing that will be saved has a method `save` that creates a file. Once I load the file, I get identical thing to the one saved. Behind the scene it would probably still do the same (i.e. stash the model, save it to metadata, save signal, and on loading instead of returning signal, check if it has any stashed models and return it if so). 

3. Given second point above, there are complications: __do we save all stashes, or just the current model?__ Do we let user to save only some of the stashes? (we could use arguments when saving or flags in the stash itself). We could also (since stash has a history) just ask user how many most recent stashes he wants to save...

4. __Slicing__. A somewhat ambiguous problem, as technically it's possible to slice the stash whenever you slice the model, however that would be very expensive. We can also get rid of the stash when slicing altogether (as the resulting signal is different, so the previous analysis does not quite belong there..). Another approach - save the slices (and learn how to combine them if there are more than one), and when restoring sliced model from unsliced stash, just pick the right numbers on-the-fly. It still involves having the full copy of the stash even for the slice, so that's a drawback. Some sort of soft-linking may be the solution?

vidartf:
@to266 My two cents on some of these issues:
What I think the overall goal of this proposal is to be able to save `Model`s together with a `Signal`. It does this by:
 a. Creating (dictionary reduced) copies of wanted models, and storing them on `Signal` somehow.
 b. Still allow other `Model`s to behave as consumers of `Signal`s, not as children.

In light of this, here's my take on the points above:
 1. Here I think a `Model` is similar to `AxesManager`, the reduced description is stored in metadata, but for actually working with it, you have objects. From this understanding, it makes sense to keep the stash as part of the metadata, but we can of course make it more accessible by a manager object (`StashManager` similar to `AxesManager`). For copying, what I would expect is that all the stored, reduced models are copied, and when they are recalled (`pop()` or `apply()`), the `Model.spectrum` reference is set to the `Signal` instance in which they are stored. If the models are stored un-reduced, I expect the signal references to be updated on deepcopy (same as for `AxesManager`?).
 2. Here, it get's a bit tricky. I guess we could support both paths? One function that simply stores it to `Signal`, and one that stores it, and then saves the file. If you are stashing several models, you definitely do not want it to write to file for every model stashed, but for single model use you'd expect it. Maybe call it `commit()` and `save()`?

To me, I here feel I disagree with calling it a "stash". It makes me think its some sort of temporary storage, while what we are intending to do is actually to store it permanently. I also don't understand why the stash manager is put on `Model`? They are all stored on `Signal`, so that is where it makes sense to have the manager. Then each Model instance would only have the utility functions `save()` and `commit()` (which would be identical to `Signal<.analysis>.models.save(model_instance)` etc.). And recalling a stored model would create a new `Model` instance, instead of transmuting an existing one. `apply()` could therefore better be named `recall()`, `load()` or similar.

 - 3. `Signal.save()` and `Model.save()` should save the entire stash. The user is responsible for the state of their data, and we should not drop data when some would expect it to be retained.
 - 4. I'm not entirely sure what happens when you slice a model...

to266:
Thanks for the response, awesome points!

One possible misconception you have (I may be wrong though) is that each stash has a copy of signal - it does not. Your (b) point is actually something we've though about before, however not done anything except moving `hs.create_model(signal)` to `signal.create_model()` - at least I am tempted to think of a model as a _child_ of a signal. In my limited experience it rarely makes sense to apply the same model to a different signal (the big obvious exception being raw and denoised signals). Do you think otherwise purely due to how it's structured in the code, or do you think it should be that way conceptually? I'm actually really curious about it!

Further, the current stash implementation essentially has a stash manager. And I guess moving it to signal instead of a model indeed makes sense.

Now the "reference swapping" is tricky - the current stash is written to behave as checkpoints - once saved, you can always revert to them and have exactly the same thing. If we pass references instead, this behavior disappears. 

> If you are stashing several models, you definitely do not want it to write to file for every model stashed, but for single model use you'd expect it.

Could you explain that some more? I am not sure I got the point...

As for the final (4) point, when slicing a model you create a completely new model filled with identical components, where all parameters `map`s (i.e. the arrays that store actual values) point to appropriate references. 

Now that I think about it, it may not be that difficult to actually slice stash as well (by creating almost-deepcopy, where the only difference is that the appropriate maps point to appropriate regions). Probably just got used to large arrays and out-of-memory operations, hence tend to avoid things like that....

vidartf:
Regarding `Model` as child thing: What I meant to say is that currently models keep a reference to Signal, but `Signal` does not have a collection of all its child `Model`s (the creation is fire-and-forget from `Signal`'s point of view). I'm not saying I agree/disagree with that, just stating what the current reality is. Having `Signal` keep track of all its models would make the bookkeeping easier, but make garbage collection less robust since we're introducing a circular reference. If `Signal` did keep track of all the models that refer back to it, it should either way not be stored with the Signal on save, so not that relevant for this discussion.

I get what you mean by "checkpoints", but as far as I can see, recalling a stored Model would completely overwrite the Model (or is something retained?). I think maybe it is better by example:
```python
m1 = s.create_model()
m2 = s.create_model()

 ... Setup m1 and m2 differently

m1.stash('Model1a')
m2.stash('Model2a')

... Modify m1 and m2 again

m1.stash('Model1b')
m2.stash('Model2b')

# In current implementation, all of these are valid:

# Using checkpoints:
m1.apply('Model1a')
m1.apply('Model1b')
m2.apply('Model2a')

# Exchanging models
m1.apply('Model2b')
m2.apply('Model1b')
```

Since the models are fully interchangeable, there is no reason for keeping using the same instances. By simply returning new instances, this disconnect is made explicit.

vidartf:
> >If you are stashing several models, you definitely do not want it to write to file for every model stashed, but for single model use you'd expect it.

> Could you explain that some more? I am not sure I got the point...

Say you have model of a signal, and you want to apply three operations on it, and save a checkpoint after each one. Then, if you write out the signal for each checkpoint, you are writing a large Signal to disk 3 times. If instead each checkpoint was only stored on the signal for each checkpoint, and then saved at the end, you only get one write, saving time.

vidartf:
Regarding slicing I now understand the question. If it truly was expensive, then I'd be in favor of lazy slicing (slice on access), but I'd prefer to just slice it immediately to make the bookkeeping simpler.

to266:
Makes sense, I agree. 

My only gripe is that metadata gets bloated, somewhat following DM style. Not sure I agree with it, but then we'd have to save heavy metadata parts separately in a separate file, which complicates things as well.

to266:
As for the slicing, having the `Analysis` in metadata is part of the problem - all of metadata just gets blindly deepcopied atm, hence creating a large deepcopy of all the stashes.

So would have to either create flags in metadata (or `DictionaryTreeBrowser`) that prevent deepcopy of certain branches, or just move the `Analysis` to a different DTB that just gets referenced in metadata when saving (in order to use existing IO methods without a change)

Any other ways around it? I'd be in favor of the second one (moving out of metadata except when saving)

to266:
@francisco-dlp what's your take?

francisco-dlp:
Excellent and important discussion! Here are my thoughts:

* +1 on having a separate container for analysis results that doesn't get copied when copying or slicing but gets saved to this. We could move `learning_results` here too. I support this because:

  1. It would be misleading to store analysis results in a signal that is not the one that was used to obtain those analysis results.
  2.  Analysis results can be big, potentially causing unnecessary overhead when e.g. slicing.
  3.  It is not always clear how to slice analysis results.

* However, there are cases when one would like to copy the analysis with the `Signal`. We could enable this by adding a keyword argument to `deepcopy` e.g. `copy_analysis`.

* +1 on implementing `m.save`. However, I think that it should serve a different purpose. Imagine that you're working with a big file that is not in hdf5. Imagine that it is so big that you don't actually want to save it in a different format, you just want to analyze it by curve fitting and save the results of the analysis, not the data. `m.save` could serve that purpose: store only the model without the data in hdf5 format. To load it back we could implement e.g. `Signal.load_model`. Of course things can get broken if the model is loaded into the wrong signal, but there is not much that we can do about it.

* Although currently a `Model` requires a `Signal` to function, actually it wouldn't be difficult to emancipate it from `Signal`. For example, one may want to load a `Model` without a `Signal` if one is only interested in the results. Also, `Model` can be used to simulate data and for this purpose there is not need to associate it with a `Signal`. For this, one could load a model saved with `Model.save` also using the `load` function e.g. `m = hs.load(<model filename>)`.

* Still it makes a lot of sense to save the model and data together.

* When working with big hdf5 file we could face the same sort of  issues. It would be really cool if we could update an hdf5 to add analysis information when storing the model using `Model.stash`.

* +1 on renaming the "stashing" functions, to create new instances instead of modifying the current model and to move some of the methods from `Model` to `Signal`. What about:

  * `Model.store` or even `Model.store_in_signal`
  * `Signal.create_model("<name of stored model>")
  * `Signal.remove_model("<name of stored model>")

* +1 on saving all the stored models when saving in hdf5 format.
* We probably should also raise a warning when saving in any other format in there are analysis results will be lost if not saving in hdf5.

to266:
Ok, so if I understand correctly, we want:

1. __A separate way to store analysis in hdf5__. Then model (learning_results?) and signal saving will be completely decoupled, but still possible to store different numbers of each in one file.

2. Some sort of __model management tool__ that is present both on its own (e.g. creating an empty model without signal), in a model (to save a model to a file) and in a signal (to create a model with the signal, current behavior). No idea how to create it / what it should be "attached to" (the hdf5 file? Signal?) or even how to call it, so ideas are welcome. The actual implementation should be easy enough.

Anything else?

francisco-dlp:
I think that you have already implemented almost everything required. Saving the model on its own could simply save to hdf5 exactly in the same that it is currently stored in metadata when using `Model.stash`, except for the reference  to the signal. storing the model in the Signal would be exactly the same as the current `Model.stash`. The only difference is that, as @vidartf proposes, "loading" the model stored in `Signal` could return a new instance of `Model` instead of modifying an existing one.

Regarding how to create a `Model` instance, I think that also should require very little modification. It'll be a matter of simply making `signal` optional in the model constructor and modify the few functions that require `Model.signal` and that are not related to fitting. This probably should be implemented in a separate pull request.

I suggest the following way forward:

* Make sure that we all agree with the implementation (@vidartf ?)
* Implement in this PR just the storing inside `Signal` part.
* Emancipate `Model` in a separate PR. In order to avoid merging nightmares I suggest to wait until we merge #536 before starting working on this.
* Afterwards implement saving the standalone model in another PR.

to266:
I realise that almost everything is actually done and only requires minor shuffling, however I would rather store the model in the correct place straight away from this PR, as changing that later would be cumbersome / in the mean time data loss would be possible.

Thinking of only hdf5, currently signals are stored in `/Experiments/<signal title>`. How about saving the models in `/Analysis/Models/<stash name>`? The idea being that it's best to decide where to put things in files and not change it if possible.

francisco-dlp:
Sounds good to me.

vidartf:
I think the current plan sounds good, let me know when all the discussed changes are committed and I'll see if I can break it >D.

I did a global search for deepcopy in hyperspy, and from the usage that is there, I agree that we can drop analysis results on deepcopy.

to266:
Should be "ready for breaking". I have not played with the code myself, let me know what you think.

Also, I did not put the `models` inside `Analysis` - at the moment I could not come up with a satisfactory  option for the `Analysis` object - DTB results in awkward printing, and there is nothing else there any way.... I suppose I'll have to change it to something, but I need good ideas for that :)

hyperspy/signal.py
```diff
@@ -76,6 +76,128 @@
...
+        for n in names:
+            node.add_node('components.' + n)
+
+    def _save(self, name, dictionary):
+
+        from itertools import product
+        _abc = 'abcdefghijklmnopqrstuvwxyz'
+
+        def get_letter(models):
+            l = len(models)
```
vidartf: Bad convention to use `l` for variable name, since it is so similar to 1. Syntax highlighting fixes this, but it is not always available. Simple fix.

hyperspy/signal.py
```diff
@@ -76,6 +76,128 @@
...
+            raise KeyError('Name has to be a string')
+        if name.startswith('_'):
+            raise KeyError('Name cannot start with "_" symbol')
+        if '.' in name:
+            raise KeyError('Name cannot contain dots (".")')
+        name = slugify(name, True)
+        if existing:
+            if name not in self._models:
+                raise KeyError(
+                    "Model named '%s' is not currently saved" %
```
vidartf: 'saved' -> 'stored'?

hyperspy/signal.py
```diff
@@ -76,6 +76,128 @@
...
+
+    """Container for models
+    """
+
+    class ModelStub(object):
+
+        def __init__(self, mm, name):
+            self._name = name
+            self.restore = lambda: mm.restore(self._name)
+            self.remove = lambda: mm.remove(self._name)
```
vidartf: These guys have no docstring, add by ```self.restore.__doc__ = "Docstring goes here"``` ?

hyperspy/io_plugins/hdf5.py
```diff
@@ -116,15 +139,29 @@ def file_reader(filename, record_by, mode='r', driver='core',
...
+            _to_remove = []
+            for i, (key, _dict) in enumerate(models_with_signals):
+                if key == exg.name:
+                    _tmp.update(_dict)
+                    _to_remove.append(i)
+            exp['models'] = _tmp
+            _to_remove.reverse()
+            for i in _to_remove:
+                models_with_signals.pop(i)
+
```
vidartf: I think this should be equivalent:
```python
_tmp = {}
for (key, _dict) in reversed(models_with_signals):
    if key == exg.name:
        tmp.update(_dict)
        models_with_signals.remove((key, _dict))
exp['models'] = _tmp
```

vidartf:
A few suggestions:
 * Add a method `ModelManager.__getitem__(self, name)` to make access interface consistent with that of `DictionaryTreeBrowser`.
 * Add a method `ModelManager.store(Model, name=None)`. Have it check if `Model.spectrum is ModelManager._signal`, and make `Model.store()` simply wrap this. This centralizes the behavior definition of `ModelManager`, and prevents `Model.store()` from accessing the protected method `ModelManager._save()` (code smell). Consider doing the same for `Model.save()`, but it's not as important for this one.

vidartf:
Playing around with it I find myself missing a method `ModelManager.pop(name)`. :)

Otherwise, I was not able to break anything outright, but I discovered this, which is basically an independent issue that is highlighted by the model store:

```python
s = hs.signals.Spectrum(np.random.rand(10,10,100))
m = s.create_model()
m.store()
s.crop(0+2j, 0, 50)
m.plot()    # This fails even before this PR
s.models.a.restore().plot()    # This will now fail
```

The obvious fix would be to have `Signal.crop()` fix its dependent models. As signals currently are unaware of their dependents, the current behavior might be expected, but it is definitely away of its stored stubs. `isig`/`inav` sidestep this issue by making a new `Signal` instance, but `crop` and other functions that modify the signal in-place break their dependents. As far as I can see, the only other functions that modify the shape of the signal in-place are the `align_?D` functions if the arguments `crop` or `expand` are true.

to266:
Good points, thanks.

As for the `crop`, I never use it myself, encourage others to never use it, and treat it as a deprecated feature that will be removed ASAP. 

On the other hand, I'm not even sure if it is the case. @francisco-dlp ?

francisco-dlp:
Good points @vidartf. In addition to `align2D` and `crop`, `map` also modifies the signal in-place which, although it won't raise . Changing the axes would also break a stored model. Also performing operations such as s *= 10.

I think that there is not much that we can do about this other than document it very clearly in the User Guide. What do you think?

to266:
I agree. The `crop` and similar things should be rewritten anyway, not related to this PR.

Anything else needs doing? I'm done from my side

francisco-dlp:
What about adding the warning box to the User Guide as discussed above to let users know that things can get broken when using e.g. `crop`, `align*` and `map` of changing the axes?

Other than that, for me it is ready to merge, but I leave last word to @vidartf.

to266:
https://github.com/hyperspy/hyperspy/pull/718/files#diff-4717924c9d836c8feccb1246064a38fbR789

Something different?

vidartf:
As soon as the doc warning is agreed, I say this is ready for merge :)

to266:
Done

francisco-dlp:
It's your privilege to merge it @vidartf ;)

vidartf:
Currently it says:
> Modifying a signal in-place (e.g. `signal.map`, `signal.crop`, `signal.align1D`, `signal.align2D`, `s += s_new` and similar) will remove all stored models. This is done intentionally.

As far as I see from the current code, the models are not removed here, they simply become invalid? Either the warning should be different, or the functions should as specified.

to266:
Changed the warning and removed the `s += s_new` example, as it does not belong here (it actually just overwrites the reference `s` with a new signal, not modifies it)

PR #167, opened 2014-02-13
Model as dict

to266:
as per https://github.com/hyperspy/hyperspy/wiki/HSPEP1%3A-Save-models-in-the-same-hdf5-file-as-the-data

Still lacks user guide entries and tests (any ideas for them are welcome, as none similar are present at the moment)

PR #720, opened 2015-09-07
ENH: Fancy general indexing

to266:
contains #167

Makes the signal slicing more modular / general, adds model slicing.
This branch does not contain `model.stash`, so no attempt to slice these is made


hyperspy/_components/eels_cl_edge.py
```diff
@@ -93,6 +93,9 @@ def __init__(self, element_subshell, GOS=None):
                             'fine_structure_coeff',
                             'effective_angle',
                             'onset_energy'])
+
+        self._whitelist['element_subshell'] = ('init', element_subshell)
+        self._whitelist['GOS'] = ('init', GOS)
```
francisco-dlp: Shouldn't `fine_structure_smoothing` and `fine_structure_active` be whitelisted too?

hyperspy/_signals/eels.py
```diff
@@ -1222,13 +1222,13 @@ def kramers_kronig_analysis(self,
         else:
             return eps, output
 
-    def create_model(self, ll=None, auto_background=True, auto_add_edges=True,
+    def create_model(self, low_loss=None, auto_background=True, auto_add_edges=True,
```
francisco-dlp: I agree that `low_loss` is better than `ll`, but this renaming requires its own branch, deprecating `ll` and keeping it until the next major release...

hyperspy/_components/eels_cl_edge.py
to266: It looks like it should. `fine_structure_width` as well, don't know why I did not add them. Will add to https://github.com/hyperspy/hyperspy/pull/167 , where it belongs. By the way, should I PR that again (to get rid of the old discussions), or is it fine like that?

hyperspy/component.py
```diff
@@ -470,6 +513,32 @@ def export(self, folder=None, name=None, format=None,
...
                 filename, '_std'))
 
+    def as_dictionary(self, picklable=False):
+        """Returns parameter as a dictionary, saving all attributes from self._whitelist.keys()
+                For more information see :meth:`hyperspy.misc.export_dictionary.export_to_dictionary`
+
+        Parameters
+        ----------
+        picklable : Bool (optional, False)
+            If any found functions will be pickled, and signals converted to dictionaries
```
francisco-dlp: No clear for the user what `pickable` does. Most hyperspy users won't know what pickling is and won't be aware of the consequences of their choice here.

hyperspy/_components/eels_cl_edge.py
francisco-dlp: If you don't mind it I'll keep on commenting on this ones, but please
fix the issues that belong to #167  in #167 and we'll make sure that we
merge that one first, or is it not a good idea?

hyperspy/component.py
```diff
@@ -119,8 +120,50 @@ def __init__(self):
         self.component = None
         self.grad = None
         self.name = ''
+        self.units = ''
```
francisco-dlp: Adding `units` to `Parameter` is probably also a good idea but one that requires its own PR...

hyperspy/misc/export_dictionary.py
```diff
@@ -0,0 +1,236 @@
...
+    return flags.replace(' ', '').split(',')
+
+def export_to_dictionary(target, whitelist, dic, picklable=False):
+    """ Exports attributes of target from whitelist.keys() to dictionary dic
+        All values are references only by default.
+        If picklable=True, the functions are copies (hence allows arbitrary closure)
+
+        Parameters
+        ----------
+            target : object
```
francisco-dlp: Too much indentation here

hyperspy/_components/eels_cl_edge.py
to266: It will be fine :)

hyperspy/misc/export_dictionary.py
```diff
@@ -0,0 +1,236 @@
...
+                    the targeted attribute is a function, and may be pickled (preferably with dill package).
+                    A tuple of (thing, value) will be exported to the dictionary, where thing is None if
+                    function is passed as-is, and bool if dill package is used to pickle the function, and
+                    value is the result.
+                * 'id':
+                    the id of the targeted attribute is exported (e.g. id(target.name) )
+                * 'sig':
+                    The targeted attribute is a signal, and will be converted to a dictionary if
+                    picklable=True
+                * 'inav':
```
francisco-dlp: That's cool, I didn't know that you've implemented this feature in a general way. If you have, again this should go to its own PR and be documented in the User Guide and listed in the highlights.

to266: This exporting to dictionary essentially __is__ https://github.com/hyperspy/hyperspy/pull/167 , so I think no need for another PR?

hyperspy/component.py
to266: I think I will remove it from here. Now that I look at it, no idea why I added it in the first place.. I guess I thought that we had it to begin with

hyperspy/misc/export_dictionary.py
francisco-dlp: Sorry, I meant these feauture:

    "The targeted attribute should be sliced when slicing navigation dimension of the original object"

hyperspy/misc/slicing.py
```diff
@@ -0,0 +1,162 @@
+from operator import attrgetter
+from hyperspy.misc.utils import attrsetter
+from hyperspy.misc.export_dictionary import parse_flag_string
+
+import numpy as np
+
+
+def slice_thing(thing, dims, both_slices=None, slice_nav=None, issignal=False):
```
francisco-dlp: Docstring missing.

hyperspy/component.py
to266: And actually we do, we just did not initialize it, as it is a `traits.Str("")`. This line is essentially only for clarity of the coder - that the default units are empty

francisco-dlp: You're right!

hyperspy/models/eelsmodel.py
```diff
@@ -55,7 +55,7 @@ class EELSModel(Model):
         Spectrum instance. Adding a new element to the spectrum using
         the components.EELSSpectrum.add_elements method automatically
         add the corresponding ionisation edges to the model.
-    ll : {None, EELSSpectrum}
+    low_loss : {None, EELSSpectrum}
```
francisco-dlp: Same renaming here "ll" -> "low_loss"

hyperspy/model.py
```diff
@@ -2107,3 +2217,80 @@ def __getitem__(self, value):
...
+
+        dims = self.model.axes_manager.navigation_dimension, self.model.axes_manager.signal_dimension
+        if self.isNavigation:
+            _model.channel_switches[:] = self.model.channel_switches
+        else:
+            _model.channel_switches[:] = \
+                np.atleast_1d(
+                    self.model.channel_switches[tuple(array_slices[-dims[1]:])])
+        from hyperspy import components
+        for _ in xrange(len(_model)):
```
francisco-dlp: Why is this necessary?

If it is indeed necessary, could you add a comment explaining why?

to266: It's a leftover from older versions, there to remove any `auto_background` and/or any `auto_add_edges` components - in case I did not catch it previously. Should I leave this fail-safe and add a description, or remove it?

francisco-dlp: I suggest to remove it.

hyperspy/component.py
to266: How about `saveable`? Not even sure it that's a word, but at least it should be relatively easy to understand what it means?

francisco-dlp: Why is it `False` by default and why giving the choice at all?

hyperspy/model.py
```diff
@@ -2107,3 +2217,80 @@ def __getitem__(self, value):
...
+                                          self.isNavigation)
+                twin_dict[id(po)] = ([id(i) for i in list(po._twins)], pn)
+
+        for k in twin_dict.keys():
+            for tw_id in twin_dict[k][0]:
+                twin_dict[tw_id][1].twin = twin_dict[k][1]
+
+        _model.chisq.data = _model.chisq.data.copy()
+        _model.dof.data = _model.dof.data.copy()
+        if not self.isNavigation:
```
francisco-dlp: I couldn't reproduce the bug after commenting out these lines. Could you check if this hack is still necessary? If it is, we should rather find the source of the problem...

to266: Looks like it is still necessary (fails tests without it). Currently trying to figure out why I need it to begin with....

to266: Fixed. Figured it out and added a comment what happens

to266:
Almost done, only one question - (where) should I add anything in the documentation?

this PR only extends previous behaviour and implements a feature that I felt should have been there from the start (i.e. as a user I'd just assume that the same slicing works for both models and signals, hence I would not look for a different set of instructions)

francisco-dlp:
I would suggest to add a section the the model documentation explaining that slicing works for model but referring to signal slicing for details. It'll be nice if you could also add a little example.

to266:
Done

Issue #722, opened 2015-09-09
BUG: adding model components after disable_adjust_position


to266:
I believe introduced by https://github.com/hyperspy/hyperspy/commit/8783c995a2197f6497a19cd1b6f6a3fe450e39bf . 

Once `disable_adjust_position` is run, the `self._adjust_position_all` is set to `False` and not `None`, hence after that any appends to the model return an error.

PR #723, opened 2015-09-09
BUGFIX for enable_adjust_position of the model

to266:
https://github.com/hyperspy/hyperspy/issues/722 Fix


dnjohnstone:
As far as I can see this should be fine to merge... very simple and does what it says on the tin. Any reason it hasn't been?

to266:
No one reviewed it :)

PR #730, opened 2015-09-23
`shift_image` use linear interpolation instead of spline interpolation.

francisco-dlp:
`shift_image` currently uses spline interpolation which is a bad idea for registration purposes. 

TODO
--------

* [ ] Write unittests for align2D


francisco-dlp:
#631 adds `align2D` unittests (among other things), so I propose to hold this one until #631 is merged.

dnjohnstone:
This seems fine to me with the unittests now implemented from #631

vidartf:
Tests from #631 do not use sub-pixel shifts, so that should be added if we want to test the fractional shifts (?). Other than that the branch needs to be made merge-compatible.

dnjohnstone:
Ah ok that's interesting.... I'm happy to make this merge compatible and to update the test_align / add a new test to cover sub-pixel shifts.

On a related note it seems (at least to me) that there is some confusion / duplication for testing the methods in misc/image_tools.py

First there are tests in: tests/misc/test_image_tools.py

Second there are tests in tests/signal/test_2D_tools.py

If I'm not wrong, I think these two files should be just one and I would probably go with the test_image_tools.py version because that's in a more logical place to me? Shall I do that too, or did I miss something/get something wrong?

francisco-dlp:
This is now up to date, but the tests are still missing. @dnjohnstone, let us know if you're going to work on them.

dnjohnstone:
@francisco-dlp sure, I can try and put something together... did you have an opinion on my earlier comment though about the correct file to contain these tests and whether the aforementioned should be combined?

francisco-dlp:
I think that they're fine as they are: one is for Signal2D and the
other one for the image tools in `hyperspy.misc`.

dnjohnstone:
Ah yes - whoops... On another point, this pull request now contains a load of style corrections for various bit of the code (mostly the new file formats putting spaces around math operations) do you want to merge those from this branch or should they be separate?

francisco-dlp:
It should be fine to merge them from this branch. Sometime ago we passed all of the code through autopep8 and some of us started using the git hook  in #181 to try to keep the style in good shape automatically. However, not everybody use it, and when someone does (me in this case) it causes this long diffs...

dnjohnstone:
Pull request made to @francisco-dlp with unit test and up to date with current master.

vidartf:
francisco-dlp/hyperspy#25 ;)

francisco-dlp:
@dnjohnstone will submit another fix for this, so closing.

PR #734, opened 2015-09-23
BUG: Fixed single colorbar vmin/vmax

### LINKED ISSUE #732: `plot_images` not working as expected when `colorbar="single"`
```python

im = hs.signals.Image((np.arange(32*32).reshape((32,32))))
hs.plot.plot_images([im, -im])

```
plot_image_colobar_multi

```python

hs.plot.plot_images([im, -im], colorbar="single")

```

plot_image_colobar_single

vidartf:
Fixes #732.

### DIFF
--- a/hyperspy/drawing/utils.py
+++ b/hyperspy/drawing/utils.py
@@ -663,9 +663,9 @@ def plot_images(images,
     # Find global min and max values of all the non-rgb images for use with
     # 'single' scalebar
     if colorbar is 'single':
-        global_max = max([i.data.max() for i in non_rgb])
-        global_min = min([i.data.min() for i in non_rgb])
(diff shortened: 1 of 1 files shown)


francisco-dlp:
Although this certainly makes things better, I think that it doesn't fully fix the issue. Fixing the issue would involve running `constrast_stretching` on all the images stacked, isn't it? Nevertheless, unless someone comes with a clever idea on how to fix this without actually stacking the arrays, I'm +1 on merging this as long as the subtle difference in behavior is explained in the docstring.

francisco-dlp:
Also, this should be sent to the maintenance branch.

vidartf:
Yeah, I figured out it should probably be stretched for the combined data after submission. I'll resub to maintenance with the combined data later.

Issue #592, opened 2015-07-15
EHN: Loading heavy data set


pburdet:
@to266 edit, 2015-09-25:

There is an increasingly grave need to be able to load datasets bigger than the available (RAM) memory. Unfortunately since various different formats have different functionalities, it has to be done on a format-by-format basis, and I guess we all agree that the first target should be the HyperSpy native HDF5.

I suggest using [`dask`](https://github.com/blaze/dask) for the numerical routines (it looks to be the perfect choice by supporting multiple formats).
### The landmarks for HDF5:
- [x] Read-only of big datasets (i.e. everyone can open and plot (with limited navigators) big signals) (#735)
- [x] Write big datasets (in addition to opening and plotting, you can stack various other signals (that individually fit in memory) and save the result as one big HDF5) (#736)
- [x] functions for big data (most notably [`dask`](https://github.com/blaze/dask) ) (used to think `signal.map`, but probably not worth it). `dask` has to catch up or we have to implement a bunch of things ourselves
- [x] enable batchPCA and batchICA (could not find the routines quickly, should be in either `scipy` or `scikit-learn` hopefully
- [ ] [optionally] rewrite (now hopefully deprecated) convenience functions (such as `fold`) that were mainly required for PCA / other machine learning
# Original @pburdet message:

I'm trying to load 2500 images as one 3D image (one image can be found under this link)

https://app.box.com/s/h73aoypja66n1fe0q2bl

The total size on the disk is 12.5GB. I'm able to open them as one stack with Imagej, but not with hyperspy. I'm running out of RAM even with 32GB. I tried `mmap=True` and `stack=False`. Nothing work.

Imagej load them as 8-bits images, Hyperspy as 16-bits with 3 channels (as if RGB). Can we control the way hyperspy load the data? Does the mmap is working correctly?
## TODO
- [ ] Enhance `mmap` loading of stacks: currently it loads all the files in memory before stacking them in disk (!). This is probably easy to fix and would solve the main problem pointed out by @pburdet.
- [ ] When loading multiple files at once, add the ability to change the dtype after loading each individual file.
- [ ] Implement `mmap` for all readers that can support it.
- [ ] Option of not saving the orginial_metadata (usefull when many files are loaded)
- [ ] Improve docstring regarding `load` `mmap` argument.

francisco-dlp:
[...]

PR #735, opened 2015-09-25
Enh hdf5 read out-of-memory

to266:
Pass `load_to_memory=False` when loading to enable.

Afterwards the file can be plotted if the navigator does not require summing (i.e. if the navigator='slider' is passed), but the file is read-only so far

First milestone for hdf5 in https://github.com/hyperspy/hyperspy/issues/592

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,7 +6,7 @@ python:
   - 2.7
 
 env:
-  - DEPS="nose numpy scipy matplotlib traits traitsui ipython h5py sympy scikit-learn netcdf4"
+  - DEPS="nose numpy scipy matplotlib traits traitsui ipython h5py sympy scikit-learn"
 
 install:
   - conda create -n testenv --yes python=$TRAVIS_PYTHON_VERSION
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,7 +6,7 @@ python:
   - 2.7
 
 env:
-  - DEPS="nose numpy scipy matplotlib traits traitsui ipython h5py sympy scikit-learn"
+  - DEPS="nose numpy scipy matplotlib traits traitsui ipython h5py sympy scikit-learn netcdf4"
 
 install:
   - conda create -n testenv --yes python=$TRAVIS_PYTHON_VERSION
--- a/hyperspy/io.py
+++ b/hyperspy/io.py
@@ -124,6 +124,10 @@ def load(filenames=None,
         mapped file will be created in the given directory,
         otherwise the default directory is used.
 
+    load_to_memory: bool
+        for HDF5 files, if True (default) loads all data to memory. If False,
+        enables only loading the data upon request
+
     Returns
     -------
     Signal instance or list of signal instances
--- a/hyperspy/io.py
+++ b/hyperspy/io.py
@@ -353,7 +353,7 @@ def dict2signal(signal_dict):
         if "Signal" in mp and "signal_origin" in mp["Signal"]:
             signal_origin = mp["Signal"]['signal_origin']
     if (not record_by and 'data' in signal_dict and
-            signal_dict['data'].ndim < 2):
+            len(signal_dict['data'].shape) < 2):
         record_by = "spectrum"
 
     signal = assign_signal_subclass(record_by=record_by,
--- a/hyperspy/io_plugins/hdf5.py
+++ b/hyperspy/io_plugins/hdf5.py
@@ -19,7 +19,6 @@
 from distutils.version import StrictVersion
 import warnings
 import datetime
-from dask import array as da
 
 import h5py
 import numpy as np
@@ -389,13 +388,6 @@ def dict2hdfgroup(dictionary, group, compression=None):
                     write_signal(value, group.create_group(key))
             else:
                 write_signal(value, group.create_group('_sig_' + key))
-        elif isinstance(value, da.Array):
-            dset = group.create_dataset(key,
-                                        shape=value.shape,
-                                        chunks=value.chunks,
-                                        compression=compression,
-                                        dtype=str(value.dtype))
-            da.store(value, dset)
         elif isinstance(value, np.ndarray):
             group.create_dataset(key,
                                  data=value,
--- a/hyperspy/io_plugins/hdf5.py
+++ b/hyperspy/io_plugins/hdf5.py
@@ -41,7 +41,7 @@ default_extension = 4
 
 # Writing capabilities
 writes = True
-version = "1.4"
+version = "1.3"
 
 # -----------------------
 # File format description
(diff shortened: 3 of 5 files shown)


hyperspy/tests/io/test_hdf5.py
```diff
@@ -229,3 +230,33 @@ def test_rgba16():
...
+        s = f['Experiments/__unnamed__']
+        del s['data']
+        s.create_dataset(
+            'data',
+            shape=self.shape,
+            dtype='float64',
+            chunks=True)
+        f.close()
+
+    @nt.raises(MemoryError)
```
francisco-dlp: This depends on the memory available in the execution machine.

hyperspy/tests/io/test_hdf5.py
```diff
@@ -229,3 +230,33 @@ def test_rgba16():
...
         "test_rgba16.npy"))
     nt.assert_true((s.data == data).all())
+
+
+class TestLoadingOOMReadOnly:
+
+    def setUp(self):
+        s = Signal(np.empty((5, 5, 5)))
+        s.save('tmp.hdf5')
+        self.shape = (10000, 10000, 100)
```
francisco-dlp: I wouldn't like to execute a test that writes such a big file in my harddrive.

hyperspy/io_plugins/hdf5.py
```diff
@@ -342,7 +391,8 @@ def parse_structure(key, group, value, _type, compression):
         elif isinstance(value, np.ndarray):
             group.create_dataset(key,
                                  data=value,
-                                 compression=compression)
+                                 compression=compression,
+                                 chunks=True)
```
francisco-dlp: Why is `chunks=True` required?

hyperspy/tests/io/test_hdf5.py
to266: It does not - it just creates an empty array that would take much more space than commonly available. Since there is no real data there, the actual file size is ~19KB

.travis.yml
```diff
@@ -6,7 +6,7 @@ python:
   - 2.7
 
 env:
-  - DEPS="nose numpy scipy matplotlib traits traitsui ipython h5py sympy scikit-learn"
+  - DEPS="nose numpy scipy matplotlib traits traitsui ipython h5py sympy scikit-learn netcdf4"
```
francisco-dlp: Why is netcdf4 required?

hyperspy/io_plugins/hdf5.py
```diff
@@ -94,44 +94,86 @@ def get_hspy_format_version(f):
...
+                      'You can still load the data using a hdf5 reader, '
+                      'e.g. h5py, and manually create a Signal. '
+                      'Please, refer to the User Guide for details')
+    if load_to_memory:
+        f.close()
+    return exp_dict_list
+
+
+# not used at the moment, but might be useful for the future
+def get_signal_chunks(data, metadata=None):
```
francisco-dlp: If this chunk of code is not used currently I would suggest not to include it until it is actually used for something.

hyperspy/tests/io/test_hdf5.py
to266: ...I think it's quite safe to assume that very few machines have 74.5GB of RAM memory free

hyperspy/tests/io/test_hdf5.py
francisco-dlp: Cool.

hyperspy/tests/io/test_hdf5.py
francisco-dlp: As of 2015 you mean...

.travis.yml
to266: Just trying to fix travis errors... For some reason all hdf5 tests on travis fail, while I cannot reproduce it myself. Hence "test" in the commit message

hyperspy/io_plugins/hdf5.py
to266: My bad, I mixed up the code. I'll remove it, as this branch only deals with reading, and that is writing option/routine

hyperspy/io_plugins/hdf5.py
to266: Ok

hyperspy/tests/io/test_hdf5.py
to266: Would you like me to add a few more zeroes, so that we're safe for some time?

to266:
Ok, travis does not make sense for me, but the branch seems to be ready

hyperspy/tests/io/test_hdf5.py
francisco-dlp: That would only buy us a limited amount of time. Why not fixing the issue for eternity by using a prediction of the average installed ram vs year based in current trends?

graph-windows-64bit-p93

(Yes, this is a joke.)

hyperspy/tests/io/test_hdf5.py
```diff
@@ -229,3 +230,33 @@ def test_rgba16():
...
+        del s
+        f = h5py.File('tmp.hdf5', model='r+')
+        s = f['Experiments/__unnamed__']
+        del s['data']
+        s.create_dataset(
+            'data',
+            shape=self.shape,
+            dtype='float64',
+            chunks=True)
```
francisco-dlp: Why do you need `chunks=True`?

hyperspy/io_plugins/hdf5.py
```diff
@@ -41,7 +41,7 @@
 
 # Writing capabilities
 writes = True
-version = "1.3"
+version = "1.4"
```
francisco-dlp: Why increasing the version? (Note that this is the format version.)

hyperspy/tests/io/test_hdf5.py
to266: ...I __could__ actually just check what's the available memory and then create the data somewhat bigger (`psutil` is wonderful, in case you did not know ;) ).. would probably only be marginally slower and we'd be safe for eternity :)

hyperspy/io_plugins/hdf5.py
to266: ...Yet another leftover, and an unnecessary one too.

hyperspy/tests/io/test_hdf5.py
to266: I don't. Just a habit from personal testing. Should do no harm (as we don't actually read the data in these tests). And if it does, at least we'll notice that HDF5 chunking is broken

hyperspy/io_plugins/hdf5.py
```diff
@@ -94,44 +94,47 @@ def get_hspy_format_version(f):
 
 
 def file_reader(filename, record_by, mode='r', driver='core',
```
francisco-dlp: Should we change `driver` to None?

hyperspy/io.py
```diff
@@ -124,6 +124,10 @@ def load(filenames=None,
         mapped file will be created in the given directory,
         otherwise the default directory is used.
 
+    load_to_memory: bool
```
francisco-dlp: All readers could support `load_to_memory=False` by just reading the array as a `numpy.memmap`. Currently this is the default for `mrc` files.

hyperspy/io_plugins/hdf5.py
to266: ...maybe, but in some other PR? I personally have no clue what the `driver` does

hyperspy/io.py
to266: probably, but, again, this PR only adds hdf5 oom reading

PR #736, opened 2015-09-29
Temporary hdf5 files for bigger than memory signals, stacking and deepcopy

to266:
`:%s/oom/out of memory/g`

Second milestone for hdf5 in #592 

Implements a number of things:
 - safe temporary files that can be easily tracked and are deleted upon exit. The one drawback is that when deleting the file on garbage collection, all exceptions (that always occur due to `h5py` reference counting) are printed to `stderr` that get flushed to the terminal, so it just does not look clean...
 - `signal.deepcopy` and `signal._deepcopy_with_new_data` work for oom signals.
 - when loading multiple signals (e.g. wildcard) and `load_to_memory=False`, the signal list is now a generator, not requiring loading all of them to the memory at once. Works even with non-hdf5 files, as long as the shapes are compatible (just like it used to be).
 - some internal hdf5 writing routines got split to enable writing "skeletons" of signals that can be populated later. Also the writing routines now should overwrite parts of already open the file when required (before it would just nuke the file and start from scratch)
 - `driver="core"` got removed, as it is not recommended (best to leave default).
 - `deepcopy2hdf5` got added - effectively writes a dictionary to hdf5 group, and then reconstructs it from the same group. The end effect is that all `np.array` members get "mapped" to `h5py.Dataset` in that particular file
 - ... maybe something else that I missed?

Comments (especially on the tempfile creation and destruction and interplay with the h5py) and advices are very welcome, as I might have missed something trivial / fundamental.

## TODO:
_______________
 - [x] `signal.deepcopy` tests
 - [x] `signal._deepcopy_with_new_data` tests
 - [x] `deepcopy2hdf5` tests
 - [x] all documentation
 - [x] fix docstring lines where `load_to_memory=None` and/or finally figure out how to deal with it. If not None is left as default and passed to other readers, they crash
 - [x] check that the iterator approach actually prevents loading everything at once and only loads one file at a time. It should also work with other files (dm3, etc)
 - [x] extend `stack` tests

### DIFF
--- a/hyperspy/misc/utils.py
+++ b/hyperspy/misc/utils.py
@@ -811,17 +811,36 @@ def stack(signal_list, axis=None, new_axis_name='stack_element',
 
     """
 
+    import h5py
+    from hyperspy.io_plugins.hdf5 import write_empty_signal, write_signal
+    import dask.array as da
     axis_input = copy.deepcopy(axis)
 
     for i, obj in enumerate(signal_list):
         if i == 0:
             if axis is None:
                 original_shape = obj.data.shape
-                stack_shape = tuple([len(signal_list), ]) + original_shape
-                tempf = None
+                stack_shape = (len(signal_list),) + original_shape
+
+                # TODO: figure out what happens when deepcopy h5py.Dataset !!!
+
+                new_metadata = copy.deepcopy(obj.metadata)
+                # Get the title from 1st object
+                new_metadata.General.title = "Stack of " + \
+                    new_metadata.General.title
+
                 if mmap is False:
-                    data = np.empty(stack_shape,
-                                    dtype=obj.data.dtype)
+                    if isinstance(obj.data, h5py.Dataset):
+                        tempfname = tempfile.NamedTemporaryFile(
+                            dir=mmap_dir).name
+                        tempf = h5py.File(name=tempfname, mode='w')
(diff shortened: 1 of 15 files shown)


hyperspy/misc/utils.py
```diff
@@ -793,6 +793,10 @@ def stack(signal_list, axis=None, new_axis_name='stack_element',
         If mmap_dir is not None, and stack and mmap are True, the memory
         mapped file will be created in the given directory,
         otherwise the default directory is used.
+    load_to_memory : bool
+        If True loads all data to memory.
+        If False only loads the data upon request.
+        If None loads to memory
```
vidartf: As far as I can tell, this keyword is very similar to `mmap`, but is only used for h5py datasets?

to266: yeah

francisco-dlp:
Requires updating.

to266:
I now actually think it's a bad idea to create the temp files, so I'll close the PR.
Instead, I'm currently working on lazy signals that lazily delay the computations until actually required. The temp files did not work on windows even with (more than I'd prefer) hacking, and the idea is bad to begin with - you don't want creating a new multi-gigabyte file for every operation. Instead you want to either compute only the required bits, or do the computation only when explicitly saving to file.

PR #737, opened 2015-10-07
`signal.map` with output

to266:
as a part of https://github.com/hyperspy/hyperspy/issues/592, required for changing signal shape with `signal.map`.

---------
### TODO:
 - [x] Documentation
 - [x] Tests
 - [x] User guide


to266:
Updated and ready for review / requests

hyperspy/signal.py
```diff
@@ -4460,9 +4466,36 @@ def map(self, function,
...
+            return function(d, **_kwargs)
+
         if show_progressbar is None:
             show_progressbar = preferences.General.show_progressbar
+        need_output = False
+        if out is True:
+            out = self._deepcopy_with_new_data()
+            need_output = True
+        if not isinstance(out, Signal) and out is not False:
+            warnings.warn("Can only pass bool or Signal to 'out'")
```
francisco-dlp: This should raise a `ValueError` instead.

hyperspy/signal.py
```diff
@@ -4491,32 +4524,63 @@ def map(self, function,
...
+            res_data = []
             for data in zip(*iterators):
                 for (key, value), datum in zip(ndkwargs, data[1:]):
                     kwargs[key] = datum[0]
-                data[0][:] = function(data[0], **kwargs)
+                if out:
+                    ans = run_function(data[0], **kwargs)
+                    # res_data.append(run_function(data[0], **kwargs))
+                    res_data.append(ans)
+                    shapes.add(ans.shape)
```
francisco-dlp: What if the output doesn't have a shape?

to266:
Fixed the error and `out` shape in corner cases.

Also had a look at creating a numpy array of objects full of signals (so that, as you suggested, we keep the navigation space even when the outputs are of different shapes), however the usual bug with numpy arrays occurs:

```python
In [3]: s
Out[3]: <Spectrum, title:  histogram, dimensions: (|60)>

In [4]: np.array([s, s], dtype='O')
/home/to266/dev/hyperspy/hyperspy/signal.py:2859: VisibleDeprecationWarning: Indexing the `Signal` class is deprecated and will be removed in HyperSpy 0.9. Please use `.isig` and/or `.inav` instead.
  VisibleDeprecationWarning)
Out[4]: 
array([[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[<Spectrum, title:  histogram, dimensions: (|1)>]]]]]]]]]]]]]]]]]]]]]]]]]]]]]]],






























       [[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[<Spectrum, title:  histogram, dimensions: (|1)>]]]]]]]]]]]]]]]]]]]]]]]]]]]]]]]], dtype=object)
```

dnjohnstone:
@to266 I just got chance to look at this - it works fine, but the output returned at least in the case that I tried is not very user friendly. Unfotunately I can't upload the data, because it's huge, but it was a stack of diffraction patterns, but the screen shots attached demonstrate the issue...

I used skimage find_local_max as a peak finder but the output 'peaks' is a stack of signals in the Image class where what I would ideally want is a structured array with the same dimensions as the original data as I implemented in #606.

Really it should be the case that the output of such a function can be plotted easily using markers as discussed in #496 

Not sure what the best way to do this whole thing is but I think we need a few more options - what do you think?

top
bottom

to266:
Oooo, that's a nice and obvious way to solve it that I missed! Will be done today.
Hopefully markers will also work out-of-the-box

Thanks @dnjohnstone

to266:
It's done - now the returned array is a signal of signals.

I also added a special case for indexing such a beast, where if just one signal is in the result, no "wrapper" is returned - just the actual signal in question... I.e.

```python
In [2]: im = hs.signals.Image(np.arange(0., 18).reshape((2,3,3,)))

In [3]: from scipy.ndimage import rotate

In [4]: angles = hs.signals.Signal(np.array([0, 45, 90, 135]))

In [5]: angles.axes_manager.set_signal_dimension(0)

In [6]: res = im.map(rotate, angle=angles, reshape=True, out=im)
calculating 100% |#############################################| ETA:  00:00:00 


In [7]: res
Out[7]: <Signal, title: , dimensions: (2|)>

In [8]: res.data
Out[8]: 
array([<Image, title: , dimensions: (|3, 3)>,
       <Image, title: , dimensions: (|4, 4)>], dtype=object)

In [9]: res.data[0]
Out[9]: <Image, title: , dimensions: (|3, 3)>

In [10]: res.inav[0]
Out[10]: <Image, title: , dimensions: (|3, 3)>

In [11]: res.inav[0] is res.data[0]
Out[11]: True

```

dnjohnstone:
I've played with this a bit and two points come to mind:

1. this now returns a signal of images if you start with an image -- that makes sense in the case you tried where you transform an image to an image, but it doesn't make so much sense if you say find peaks and therefore the thing you get out is a list of [x,y]. Perhaps the simplest way to do this would be to make it a signal of signals with the option to specify it should keep the same as the input?

2. In the case where you get e.g. a list of peak positions it's quite natural to want to plot them with multidimensional markers -- this is tricky because the output now has different nav dimensions compared to the data. I think that needs markers to be changed rather than map though so is probably beyond the scope of this PR. That said if anyone can think of an output format (I can't) that would make this much easier than what you've got here that might be better done at this point.

dnjohnstone:
I was just thinking about this and wondered how we'd feel about having an additional "output" argument where the user might specify "structured array" or "signal of signals"? In an ideal world we would maintain the nav-dimensions throughout -- but that is a bit beyond me right now.

dnjohnstone:
Is there a plan for this?

to266:
I can obviously update it, but most likely some time next week. Once that's done it will probably need a review

francisco-dlp:
Needs updating with master

to266:
Again, lazy signals will change the `map` and how it's used slightly. I will add this afterwards, will probably be better.

francisco-dlp:

Issue #698, opened 2015-08-13
Add support for FEI emd file format


francisco-dlp:
The file format turns out to be standard HDF5. A file containing 4 images in this format can be downloaded from the link in [this post](https://groups.google.com/d/msg/hyperspy-users/voA3eutSPG4/-ismMx0SBwAJ)  in the HyperSpy users mailing list.

Thank you Mingjian Wu for drawing our attention to this and providing the file.

magnunor:
Nice! I've actually been in contact with Colin Ophus before, discussing the EMD-format. They've got a project here: http://emdatasets.lbl.gov/view/

This is a very nice development, which avoids the tedious reverse-engineering usually associated with data from these kinds instruments. Hoping the other manufactures follow suit :)

magnunor:
Do we know who is developing this at (or for) FEI? Would be nice to directly communicate with them.

francisco-dlp:
[This suggests](https://www.enthought.com/company/news/nthought-canopy-helps-fei-revolutionize-transmission-electron-microscopy) that Enthought has been involved in the development.

magnunor:
There is also an example file from Colin here: http://emdatasets.lbl.gov/category/emd-examples/ , but that is probably slightly different from the FEI one, according to Mingjian Wu.

francisco-dlp:
Given the similarities between Colin's hdf5 format and HyperSpy's one, it should be simple to support that one too. @magnunor, would you like to open an issue about this?

magnunor:
I opened an issue about it: https://github.com/hyperspy/hyperspy/issues/699

francisco-dlp:
New example files kindly provided by Mingjian Wu can be downloaded from the link in [his post in the mailing list](https://groups.google.com/d/msg/hyperspy-users/voA3eutSPG4/xrJ_-teVIgAJ).

francisco-dlp:
By the way, if somebody plans to start working on this, please mention it here before starting so that we don't get several of us working on it at the same time.

I can't commit to this at the moment but I'll be happy to help anybody who wishes to give it a try.

magnunor:
I can try to implement a reader for this and the NCEM one https://github.com/hyperspy/hyperspy/issues/699 . Unless someone else wants to :)

Timeframe: about a week for basic functionality

francisco-dlp:
Great. Once there is a PR, if everybody agree, we can set its milestone 
to 0.8.3.

magnunor:
[...]

Issue #699, opened 2015-08-13
Add support for EMD-file format from NCEM


magnunor:
EMD (Electron Microscopy Dataset) is a file format based on HDF5, developed at National Center for Electron Microscopy: http://emdatasets.lbl.gov/spec/

It should be fairly easy to implement a reader for it, since it is a standard HDF5-file.

FEI is also making a EMD-format: https://github.com/hyperspy/hyperspy/issues/698 . So the readers will have to distinguish between the two.

magnunor:
I've been in touch with the maintainer of http://emdatasets.lbl.gov/ about getting some test-files for implementing this, since currently there is only one outdated file: http://emdatasets.lbl.gov/category/emd-examples/

magnunor:
FDES (https://github.com/woutervandenbroek/FDES) a TEM/STEM simulation software also uses this kind of EMD-files as output format. I'll see if can get some output-files so we can make sure they work nicely in HyperSpy.

francisco-dlp:
It seems to me that the hspy hdf5 format is a valid EMD format except for some details e.g. the versioning. Am I wrong?

Wouldn't it be make sense to have a general EMD reader which simply reads all the datasets in the file and sets the `original_parameters` attribute properly? Then, adding "full support" would involve, in simple cases (the hspy format not falling in this category), just mapping from `original_parameters` to `mapped_parameters`.

magnunor:
I think you're right, the only thing which is needed to have a valid EMD-file is having the right attribute tags + a dataset called "data"

That should work, but I think we should separate the FEI and NCEM version in two different readers since they look quite different.

In addition there are two different versions of the NCEM, 0.1 and 0.2. The 0.2 version is a little less strict, so waiting for Colin to generate some sample files.

magnunor:
Firstly, a big thanks to @CodeMonkeyJan for implementing support in the https://github.com/hyperspy/hyperspy/pull/813 pull request.

A small on update on the EMD-project:

They've open sourced their EMD viewer, which is implemented in C++ and Qt: https://github.com/pophus/emdviewer. Also a reader/writer for EMD, written in C++: https://github.com/pophus/emdlib. 

The project webpage at http://emdatasets.lbl.gov/ is down at the moment, but should be back up soonish.

to266:
FYI #953 is merged now.

CodeMonkeyJan:
Thanks a lot :-)!

magnunor:
Very nice!

[...]

PR #738, opened 2015-10-15
IO-Plugins for EMD and SEMPER formats

CodeMonkeyJan:
Added two io_plugins for the EMD format, a hdf5 standard proposed by Colin Ophus from Lawrence Berkeley National Lab (see http://emdatasets.lbl.gov/ not to be confused with the FEI EMD format which was developed later and is mentioned here: https://github.com/hyperspy/hyperspy/issues/698)
and a format for the image processing system Semper, which was developed by Owen Saxton. It has the ending '.unf' and is a binary file with an extensive header.

### DIFF
--- a/hyperspy/io_plugins/emd.py
+++ b/hyperspy/io_plugins/emd.py
@@ -1,0 +1,374 @@
+# -*- coding: utf-8 -*-
+# Copyright 2007-2015 The HyperSpy developers
+#
+# This file is part of  HyperSpy.
+#
+#  HyperSpy is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+#  HyperSpy is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with  HyperSpy.  If not, see <http://www.gnu.org/licenses/>.
+
+# INFORMATION ABOUT THE EMD FORMAT CAN BE FOUND HERE:
+# http://emdatasets.lbl.gov/
+
+
+import re
+import h5py
+
+import logging
+
+
+# Plugin characteristics
+# ----------------------
+format_name = 'Electron Microscopy Data (EMD)'
+description = 'Read data from Berkleys EMD files.'
+full_support = True  # Hopefully?
+# Recognised file extension
+file_extensions = ('emd', 'EMD')
+default_extension = 0
+# Writing features
+writes = True
+EMD_VERSION = '0.2'
+# ----------------------
+
+
+class EMD(object):
+
+    '''Class for storing electron microscopy datasets.
+
+    The :class:`~.EMD` class can hold an arbitrary amount of datasets in the `signals` dictionary.
+    These are saved as HyperSpy :class:`~hyperspy.signal.Signal` instances. Global metadata
+    are saved in four dictionaries (`user`, `microscope`, `sample`, `comments`). To print
+    relevant information about the EMD instance use the :func:`~.print_info` function. EMD
+    instances can be loaded from and saved to emd-files, an hdf5 standard developed at Lawrence
+    Berkeley National Lab (http://emdatasets.lbl.gov/).
+
+    Attributes
+    ----------
+    signals: dictionary
+        Dictionary which contains all datasets as :class:`~hyperspy.signal.Signal` instances.
+    user: dictionary
+        Dictionary which contains user related metadata.
+    microscope: dictionary
+        Dictionary which contains microscope related metadata.
+    sample: dictionary
+        Dictionary which contains sample related metadata.
+    comments: dictionary
+        Dictionary which contains additional commentary metadata.
+
+    '''
+
+    _log = logging.getLogger(__name__)
+
+    def __init__(self, signals={}, user={}, microscope={}, sample={}, comments={}):
+        self._log.debug('Calling __init__')
+        # Make sure some default keys are present in user:
+        for key in ['name', 'institution', 'department', 'email']:
+            if key not in user:
+                user[key] = ''
+        self.user = user
+        # Make sure some default keys are present in microscope:
+        for key in ['name', 'voltage']:
+            if key not in microscope:
+                microscope[key] = ''
+        self.microscope = microscope
+        # Make sure some default keys are present in sample:
+        for key in ['material', 'preparation']:
+            if key not in sample:
+                sample[key] = ''
+        self.sample = sample
+        # Add comments:
+        self.comments = comments
+        # Make sure the signals are added properly to signals:
+        self.signals = {}
+        for name, signal in signals.iteritems():
+            self.add_signal(signal, name)
+
+    def __getitem__(self, key):
+        # This is for accessing the raw data easily. For the signals use emd.signals[key]!
+        return self.signals[key].data
+
+    def _write_signal_to_group(self, data_group, signal):
+        self._log.debug('Calling _write_signal_to_group')
+        # Save data:
+
+        dataset = data_group.create_group(signal.metadata.General.title)
+        dataset['data'] = signal.data
+        # Iterate over all dimensions:
+        for i in range(len(signal.data.shape)):
+            key = 'dim{}'.format(i+1)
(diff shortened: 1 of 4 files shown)


to266:
Hi there, thanks for the contribution!

As always for all PRs, you will need to add tests and documentation + user guide changes. 

Furthermore, a couple of things related to hdf5 that I'm trying to work on at the moment, and if you included in your PR, that would be awesome:
 - for datasets (with actual data) `maxshape = tuple(None for _ in data.shape)` when creating them so that we can reshape it to whatever.
 - chunked storage (is actually a prerequisite of the previous point). The chunk estimation, I suppose, can be left default (i.e. `chunks=True`), as the reader should be smart enough to keep what's needed. In case you feel you can do a better job (i.e. making sure each chunk is at least one full "signal" (image / spectrum / whatever it might be)), you are welcome to give it a shot! This last request is really optional, as it is significantly more complex than just adding two copy-pasted bits of code into current dataset creation routines..

Finally, I will add a couple of general code-style comments in the code itself. It would be great if you could change it accordingly, should not take too long.

Having said all that, I haven't had a chance to actually have a look what the code does, but it's awesome that you contributed it, thank you very much!

hyperspy/io_plugins/emd.py
```diff
@@ -0,0 +1,374 @@
...
+    sample: dictionary
+        Dictionary which contains sample related metadata.
+    comments: dictionary
+        Dictionary which contains additional commentary metadata.
+
+    '''
+
+    _log = logging.getLogger(__name__)
+
+    def __init__(self, signals={}, user={}, microscope={}, sample={}, comments={}):
```
to266: We encourage not to use new/empty dicts as default values. Instead you should write, e.g.

```python
def __init__(self, signals=None, ): # same for others if appropriate
    if signals is None:
        signals = {}
    # other default checks / code follows
```

hyperspy/io_plugins/emd.py
```diff
@@ -0,0 +1,374 @@
...
+    user: dictionary
+        Dictionary which contains user related metadata.
+    microscope: dictionary
+        Dictionary which contains microscope related metadata.
+    sample: dictionary
+        Dictionary which contains sample related metadata.
+    comments: dictionary
+        Dictionary which contains additional commentary metadata.
+
+    '''
```
to266: Could you change `'''` to `"""` for the documentation definitions?

hyperspy/io_plugins/emd.py
```diff
@@ -0,0 +1,374 @@
...
+        # Write metadata:
+        dataset.attrs['emd_group_type'] = 1
+        for key, value in signal.metadata.Signal:
+            dataset.attrs[key] = value
+
+    def _read_signal_from_group(self, name, group):
+        self._log.debug('Calling _read_signal_from_group')
+        import hyperspy.api as hp
+        # Extract essential data:
+        data = group.get('data')[...]
```
to266: You can just pass the `h5py.Dataset` as-is. It should ideally avoid loading it into memory unless we really need to. It's easy to get the numpy array later anyway

hyperspy/io_plugins/semper_unf.py
```diff
@@ -0,0 +1,516 @@
...
+        arg_dict['ICLASS'] = iclass
+        arg_dict['IFORM'] = iform
+        arg_dict['IVERSN'] = iversn
+        arg_dict['ILABEL'] = ilabel
+        arg_dict['IFORMAT'] = iformat
+        arg_dict['IWP'] = iwp
+        arg_dict['IPLTYP'] = ipltyp
+        arg_dict['ICCOLN'] = iccoln
+        arg_dict['ICROWN'] = icrown
+        arg_dict['ICLAYN'] = iclayn
```
to266: Since it's just assignment, it's more efficient to create the dictionary already full, so that it reads
```python
arg_dict = { 'data': data,
             'title': title,
             'offsets': (x0, y0, z0),
             'scales': (dx, dy, dz),
             'units': (ux, uy, uz),
             'date': date,
             'ICLASS': iclass,
             'IFORM': iform,
             'IVERSN': iversn,
             'ILABEL': ilabel,
             'IFORMAT': iformat,
             'IWP': iwp,
             'IPLTYP': ipltyp,
             'ICCOLN': iccoln,
             'ICROWN': icrown, 
             'ICLAYN': iclayn
            }
```
instead. do check for spelling / similar errors if copy-pasting

hyperspy/io_plugins/semper_unf.py
```diff
@@ -0,0 +1,516 @@
...
+        arg_dict['ICLASS'] = iclass
+        arg_dict['IFORM'] = iform
+        arg_dict['IVERSN'] = 2  # current standard
+        arg_dict['ILABEL'] = 1  # True
+        arg_dict['IFORMAT'] = None  # not needed
+        arg_dict['IWP'] = 0  # seems standard
+        arg_dict['IPLTYP'] = 248  # seems standard
+        arg_dict['ICCOLN'] = data.shape[2]//2 + 1
+        arg_dict['ICROWN'] = data.shape[1]//2 + 1
+        arg_dict['ICLAYN'] = data.shape[0]//2 + 1
```
to266: same here with dict creation

to266:
Finally, I think if you could split the two new formats into two separate pull-requests, it would be better as well

magnunor:
Very nice :)

I'm guessing the EMD-reader will only work with the newest version of EMD (0.2)? 

As mentioned in https://github.com/hyperspy/hyperspy/issues/699, it would be nice to make sure this works nicely with EMD-files made from different programs (excluding the FEI-EMD which probably needs its own reader plugin).

CodeMonkeyJan:
Hello everyone and sorry for the delayed reply!

Maybe I should have included a bit more information instead of just requesting a pull right out of the blue :-P, so I’ll catch up on that now:
I’m a PhD student at the Ernst Ruska-Centre for electron microscopy at the research centre in Jülich (Germany) and now have roughly three years of experience with Python. Recently, we addressed the issue of handling a vast array of different file formats and analytical tools which are present in our institute and the problems of exchanging data which stem from that. Our first idea was to rewrite our existing code in an open-source multiplatform language (obviously leading to our choice of Python), including read and write routines for the different formats. After checking out HyperSpy, we decided that we do not have to “reinvent the wheel” and instead maybe could contribute our efforts to it. The two io_plugins I wrote are a first step in this direction and would be useful for people (hopefully not only in our institute) who’d like to make the transition to Python.

First of all a big thank you for your input, I’m always eager to improve my coding style and learning from more experienced people is one of the best ways in my opinion. I mainly try to be compliant with PEP8, although I increased the max character number to 99 in my code (I hope that’s okay).
I included the `maxshape` and `chunksize` arguments during dataset creation (it now should also be passed directly, avoiding the loading into memory), removed the empty dictionaries as defaults and changed the documentation definitions from `'''` to `"""`. For the last one, is this purely an aesthetic choice or has this an actual influence on the documentation via sphinx?

I’m currently looking into splitting the pull request into two smaller requests. Up till now I only used versioning software for my PhD thesis to have a file history, handling larger structures with multiple branches and contributors is new for me. Would the best strategy be to split my fork of Hyperspy into two branches (each just adding one plugin) and then submitting two pull requests?
If yes I would do this (after adding documentation and tests for both plugins).

Thanks again :-),

Jan

to266:
Hi Jan,

Yes, splitting your fork into two branches and submitting two separate pull-requests is the preferred way to do it.

One other thing that we don't strictly require but is a nice practice when using git is to rebase your branch (the one you want to merge) on hyperspy master once it is ready for merge.

That most often requires forced pushing to your github, but in the end it keeps the history "linear", which helps to make sense of it.

CodeMonkeyJan:
Rebasing also means that I can only request a pull for one feature at a time, right? Rebasing both io_plugins would mean that both would be in the master branch at the same time, so I would simply submit the second one after the first is accepted? Sorry if the question is stupid, as I said, I'm relatively new to git :-).
Also a small question regarding the documentation, is it enough to modify doc/api/hyperspy.io_plugins.rst to include the new plugin and let sphinx autotdoc take care of the rest or should I include something else?
Thanks!

to266:
Yes, that is correct. Though you can PR both at the same time, and just rebase whichever is the second one after the first is merged - that way there would be more time for us to review the code in both branches :)

And yes, the `.rst` edit should be enough. You could try to compile the documentation yourself if you're worried that something does not work, of course.

Finally, really do not worry to PR not finished branches - it's a nice way for other people to follow the progress and advice and discuss if they spot something they are interested in, or if you have a question yourself.

CodeMonkeyJan:
Thanks for the help, I'm going to open a new pull request for the first plugin now :-)!

PR #741, opened 2015-10-27
IO-Plugins for the SEMPER unf format

CodeMonkeyJan:
Added io_plugins a format for the image processing system Semper,
which was developed by Owen Saxton. It has the ending '.unf' and is a binary
file with an extensive header.
This time should include testing and documentation.


to266:
Hi, to begin with, have a look at this [mrc reader we have](https://github.com/hyperspy/hyperspy/blob/master/hyperspy/io_plugins/mrc.py) - it's also a binary file reader, however it uses numpy binary reader extensively. My guess is that the code would be more readable and probably faster if you used it instead of the "manual" reading you have at the moment.

Furher comments:
```python
# some stuff....
@classmethod
def load_from_unf(self, filename):
    #some other stuff
    return SemperFormat(arg_dict)
```
is somewhat bad practice not only because you use `self` instead of `cls` (or something similar), which might trick code maintainers that it's actually instance method, but also at the end (when you actually want to use the `cls` (or `self` in this case)), you just refer to the `SemperFormat` class from "outside". 
My understanding is that is the point of the `classmethod` - in the end you can write `cls.__init__(arg_dict)` to properly initiate the class after some pre-processing of input.

You also do a similar thing at the return of `from_signal` (although there you correctly use `cls` in the beginning)

You also use many `assert` statements, which would just crash if something went wrong. I think it's much more graceful to instead use `try/except/raise/finally`. It may not be better in all cases, but definitely at any point where user might have entered something wrong (e.g. `from_signal` has `assert` at the beginning checking dimensionality of the data. That should definitely raise an exception with a helpful error message instead).

Next, testing:

Currently we only use `nosetests`, so `unittests` are not required. That will probably get rid of the `if __name__ == "__main__": #...` part.

Next, only the last test (`TestCaseSaveAndRead`) may actually require a class, and only to clean up afterwards. All other tests can just be run as test functions, as an example:

```python
def test_with_title_loading():
    signal = load(os.path.join(my_path, 'unf_files', 'example_with_title.unf'))
    assert_equal(signal.data, data_image)
    assert_equal(signal.original_metadata.IFORM, 2)  # float
    assert_equal(signal.metadata.General.title, 'This is a test!')
    assert isinstance(signal, Image)
```

A couple more caveats:
 - In this example I left `assert` as you did in the code. Instead you should use `nt.assert_is_instance(signal, Image)`
 - Since you already have `title = 'This is a test!'` defined at the top of the file, you should use it in the test instead.
 - Once (if?) you implement exceptions when reading / writing files to semper format, you should test for them as well - e.g. if a user tries to save more than 3D file, that a suitable exception is raised.


I'll leave a couple of comments in the code itself too.
As always, thanks for contributing!

hyperspy/io_plugins/semper_unf.py
```diff
@@ -0,0 +1,520 @@
...
+                    z0 = struct.unpack('<f', ''.join([chr(x) for x in label[79:83]]))[0]
+                    dy = struct.unpack('<f', ''.join([chr(x) for x in label[83:87]]))[0]
+                    y0 = struct.unpack('<f', ''.join([chr(x) for x in label[87:91]]))[0]
+                    dx = struct.unpack('<f', ''.join([chr(x) for x in label[91:95]]))[0]
+                    x0 = struct.unpack('<f', ''.join([chr(x) for x in label[95:99]]))[0]
+                assert ''.join([str(unichr(l)) for l in label[100:100+ntitle]]) == title
+                ux = ''.join([chr(l) for l in label[244:248]]).replace('\x00', '')
+                uy = ''.join([chr(l) for l in label[248:252]]).replace('\x00', '')
+                uz = ''.join([chr(l) for l in label[252:256]]).replace('\x00', '')
+                assert np.frombuffer(f.read(4), dtype=np.int32)[0] == rec_length
```
to266: I would move all the "label extraction" to a separate (private) function to increase code readability. That could also include a nice checking that the it is actually formatted properly (i.e. all the `assert`s), and if not, just return an empty label and a warning (so that you can still load the data and only label is lost)

hyperspy/io_plugins/semper_unf.py
```diff
@@ -0,0 +1,520 @@
...
+                    'IFORM': iform,
+                    'IVERSN': iversn,
+                    'ILABEL': ilabel,
+                    'IFORMAT': iformat,
+                    'IWP': iwp,
+                    'IPLTYP': ipltyp,
+                    'ICCOLN': iccoln,
+                    'ICROWN': icrown,
+                    'ICLAYN': iclayn}
+        return SemperFormat(arg_dict)
```
to266: As mentioned in previous comment regarding `classmethod`

hyperspy/io_plugins/semper_unf.py
```diff
@@ -0,0 +1,520 @@
...
+                    label[95:99] = [ord(c) for c in struct.pack('<f', self.offsets[0])]  # X0
+                    label[100:100+len(self.title)] = [ord(s) for s in self.title]
+                    label[244:248] = [ord(c) for c in self.units[0]] + [0]*(4-len(self.units[0]))
+                    label[248:252] = [ord(c) for c in self.units[1]] + [0]*(4-len(self.units[1]))
+                    label[252:256] = [ord(c) for c in self.units[2]] + [0]*(4-len(self.units[2]))
+                    # Write label:
+                    f.write(struct.pack('I', 2*256))  # record length, 4 byte format!
+                    for element in label:
+                        f.write(struct.pack('h', element))  # 2 byte format!
+                    f.write(struct.pack('I', 2*256))  # record length!
```
to266: Again, most of this `if self.ilabel: #...` can be moved to a private function that actually creates it (i.e. lines 310-340)

hyperspy/io_plugins/semper_unf.py
```diff
@@ -0,0 +1,520 @@
...
+        assert len(data.shape) <= 3, 'Only up to 3-dimensional datasets can be handled!'
+        scales, offsets, units = [1., 1., 1.], [0., 0., 0.], ['', '', '']
+        for i in range(len(data.shape)):
+            scales[i] = signal.axes_manager[i].scale
+            offsets[i] = signal.axes_manager[i].offset
+            units[i] = signal.axes_manager[i].units
+            if str(units[i]) == '<undefined>':
+                units[i] = ''   # Traits can not be written!
+        for i in range(3 - len(data.shape)):  # Make sure data is 3D!
+            data = np.expand_dims(data, axis=0)
```
to266: I would instead write [(newaxis)](https://docs.scipy.org/doc/numpy/reference/arrays.indexing.html#numpy.newaxis)
```python
data = data[tuple(None for _ in xrange(3-len(data.shape)))]
```
alternatively `np.atleast_3d ` may also be used, but I guess you want to extend the dimensions only in particular direction?

hyperspy/io_plugins/semper_unf.py
```diff
@@ -0,0 +1,520 @@
...
+            print 'ICROWN: ', self.icrown
+            print 'ICLAYN: ', self.iclayn
+        print '------------------------------------------------------\n'
+
+
+def file_reader(filename, print_info=False, **kwds):
+    semper = SemperFormat.load_from_unf(filename)
+    if print_info:
+        semper.print_info()
+    signal = semper.to_signal()
```
to266: to construct the dictionary you can just call
```python
dictionary = signal._to_dictionary()
```

hyperspy/tests/io/test_semper_unf.py
```diff
@@ -0,0 +1,140 @@
+# -*- coding: utf-8 -*-
+# Copyright 2015 by Forschungszentrum Juelich GmbH
+# Author: Jan Caron
+#
+
+import os.path
+from os import remove
+import datetime
+import h5py
```
to266: `h5py` and `datetime` are not used

to266:
Well done!

Any reasons why you prefer `OrderedDict`?

CodeMonkeyJan:
Thanks for the input and sorry for the long delay, I just got back to working on my Hyperspy fork.

To adress a few things:
- I used `OrderedDict` because I saw it in the FEI plugin and I liked the fact that it keeps the order I defined in the structured arrays (especially for the `print_info`-function). It's just for aesthetics and is not necessary for the program to work.
- I use the `__name__ == "__main__":` part in the test_semper_unf file to be able to simply execute it as a script to run the tests (which made writing it a lot easier). If this is bad practice I will remove this line. By the way, I really like `nose`, it seems a lot easier than unittest, thanks for pointing me in that direction.
- I just saw that I forgot your suggestion to use `data = data[tuple(None for _ in xrange(3-len(data.shape)))]`, I will include that in the next commit! `np.atleast3d` sadly does not work because (as you suspected), the dimensions have to be added at the beginning (`np.atleast3d` will put a 1D-array onto the second axis for some reason).

Thanks again!

hyperspy/io_plugins/semper_unf.py
```diff
@@ -0,0 +1,563 @@
...
+    @classmethod
+    def _read_label(cls, unf_file):
+        rec_length = np.fromfile(unf_file, dtype='<i4', count=1)[0]  # length of label
+        label = sarray2dict(np.fromfile(unf_file, dtype=cls.LABEL_DTYPES, count=1))
+        label.update({'SEMPER': ''.join([str(unichr(l)) for l in label['SEMPER']])})
+        assert label['SEMPER'] == 'Semper'
+        # Process dimensions:
+        for key in ['NCOL', 'NROW', 'NLAY', 'ICCOLN', 'ICROWN', 'ICLAYN']:
+            value = 256**2*label.pop(key+'H') + 256*label[key][0] + label[key][1]
+            label.update({key: value})
```
to266: Why not just `label[key] = value`? I suppose it's the same, but just seems a little strange. It would be slightly quicker as well:
```python
In [3]: %timeit {}.update({'a':1})
The slowest run took 6.20 times longer than the fastest. This could mean that an intermediate result is being cached 
1000000 loops, best of 3: 308 ns per loop

In [4]: %timeit {}['a'] = 1
10000000 loops, best of 3: 71.7 ns per loop
```

CodeMonkeyJan: It seems I got a little carried away with the update function...
It makes sense in a few places where I really have a big dictionary to add to the metadata. I replaced it in all instances where just one key is updated!

hyperspy/io_plugins/semper_unf.py
```diff
@@ -0,0 +1,563 @@
...
+        else:
+            range_string = ''.join([str(unichr(l)) for l in label['RANGE'][:label['NCRANG']]])
+        label.update({'RANGE': range_string})
+        # Process real coords:
+        x0 = struct.unpack('<f', ''.join([chr(x) for x in label.pop('X0V0')]))[0]
+        dx = struct.unpack('<f', ''.join([chr(x) for x in label.pop('DXV1')]))[0]
+        y0 = struct.unpack('<f', ''.join([chr(x) for x in label.pop('Y0V2')]))[0]
+        dy = struct.unpack('<f', ''.join([chr(x) for x in label.pop('DYV3')]))[0]
+        z0 = struct.unpack('<f', ''.join([chr(x) for x in label.pop('Z0V4')]))[0]
+        dz = struct.unpack('<f', ''.join([chr(x) for x in label.pop('DZV5')]))[0]
```
to266: I would also prefer if you moved all of these unpacking operations into a small (possibly not even class-level) function, e.g
```python
def unpack_label(fmt, label):
   return struct.unpack(fmt, ''.join([chr(x) for x in label]))
# or, I think it's marginally faster on my system:
   return struct.unpack(fmt, ''.join(map(chr, label)))
```
two reasons for that:
 1. It's repeated many times, so if anything changes we would only need to change the code once. Additionally you could even use `functools.partial` to only define the `fmt` once and not hard-code it into the function
 2. We will migrate to python3 in the beginning of next year, so it's the function that will definitely change, hence first the point is even more important.

hyperspy/io_plugins/semper_unf.py
```diff
@@ -0,0 +1,563 @@
...
+        if label['REALCO'] == 1:
+            label.update({'X0V0': x0})
+            label.update({'DXV1': dx})
+            label.update({'Y0V2': y0})
+            label.update({'DYV3': dy})
+            label.update({'Z0V4': z0})
+            label.update({'DZV5': dz})
```
to266: Either update with all values at once (1), or just set each value individually (2). I guess (1) is more appropriate here.
```python
# (1)
label.update({'X0V0': x0, 'DXV1': dx, <other key:value pairs>})
# (2)
label['X0V0'] = x0
label['DXV1'] = dx
<other updates>
```

to266:
That was quick :) once you add the unpacking function, I think I'll merge if no one else comments

hyperspy/io_plugins/semper_unf.py
```diff
@@ -0,0 +1,563 @@
...
+        dy = struct.unpack('<f', ''.join([chr(x) for x in label.pop('DYV3')]))[0]
+        z0 = struct.unpack('<f', ''.join([chr(x) for x in label.pop('Z0V4')]))[0]
+        dz = struct.unpack('<f', ''.join([chr(x) for x in label.pop('DZV5')]))[0]
+        if label['REALCO'] == 1:
+            label['X0V0'] = x0
+            label['DXV1'] = dx
+            label['Y0V2'] = y0
+            label['DYV3'] = dy
+            label['Z0V4'] = z0
+            label['DZV5'] = dz
```
to266: Actually `update` would work here, if you used just one call. Although I don't think it matters and it's good as it is right now

hyperspy/io_plugins/semper_unf.py
CodeMonkeyJan: Good idea, I will do the same with the opposite direction (`struct.pack`), because it's also repeated several times.
Where would you put these small functions? Define them inside the `_read_label`-function or directly on the module level?

to266: I think for now you can put them in the module itself, but keep it general (i.e. with `fmt` as a parameter) and define the format using `partial` in each function that needs it?
If someone else ends up using the function elsewhere in the code, it's trivial to move it to a `utils` module anyway

CodeMonkeyJan: Thanks, I implemented on a modular level now and used partial as suggested.

hyperspy/io_plugins/semper_unf.py
CodeMonkeyJan: After your first comment I've implemented it with update (one call) now, it looks a bit cleaner in my opinion :-).

CodeMonkeyJan:
That was fast as well, thanks :-)!
I closed the pull request #738 (where I had both plugins for EMD and the Semper format) as it is redundant now. I will open a new pull request as soon as I worked on the EMD plugin a bit more.

to266:
@CodeMonkeyJan as mentioned in #771, it looks like the default line endings in your machine are not setup to LF
Could you set up `core.autocrlf = true` on your side? [This](http://adaptivepatchwork.com/2012/03/01/mind-the-end-of-your-line/) is a good write-up about that

CodeMonkeyJan:
On my PC, I am using Mercurial Hg to do versioning (using hggit to push to github), so I had to look up how to change the line ending behaviour. I found a plugin which let's me set the native EOL to LF. After  setting up a new clone of the hyperspy repository this should (!) now use the correct line endings. Furthermore I fixed a very specific bug which I just found out about: Reading and writing byte data (int8) is a bit strange in Semper, because it (for some reason) only can save even numbers of bytes (which is only a problem in the very specific case when the dimensions of an int8 array are odd). The issue is fixed now and appropriate testing is added.
I will also add documentation as requested by #777.
After that, how should I proceed? Can I just open a new PR? Or should this somehow be added here?
Thanks in advance!

to266:
A new PR will do the job. Thanks for the promptness

francisco-dlp:
Opening a new PR and mentioning this issue there should be fine.

Another issue that I have noticed when writing to UNF is that, when the dtype is not supported I get e.g. the following warning:

```TypeError: Data type not understood (float64))!```

Would it be better to raise a more informative `IOError` with a message such as:

```IOError: The Semper file format doesn't support float64 data type. Supported data types are: int8, int16, int32, float32, complex64```?

CodeMonkeyJan:
Thanks, I will try to have it ready by tomorrow.
You're right, a more informative `IOError` would be much better, I'll change that too.

francisco-dlp:
@CodeMonkeyJan, because you have contributed to HyperSpy your name will be among the HyperSpy authors in [Zenodo](https://zenodo.org/record/46874#.VtgpZM1amV6). Unfortunately, I failed to warn you about the fact that zenodo takes your name from your github profile, and there you are simply Jan. If you would like your contribution to be credited to your full name for the next release, could you complete your name and affiliation in your github profile? Thanks!

CodeMonkeyJan:
Thanks for letting me know, I completed the profile, so it should be ready for the next release.

PR #746, opened 2015-11-02
Doc Fix python prompts hiding feature

francisco-dlp:
When changing to the RTD theme we lost the ability to hide the Python prompts in the docs. This PR restores it using the solution suggested by @matthew-brett jere snide/sphinx_rtd_theme#167 .

### DIFF
--- a/doc/_static/jquery.js
+++ b/doc/_static/jquery.js
@@ -1,0 +1,9404 @@
+/*!
+ * jQuery JavaScript Library v1.7.2
+ * http://jquery.com/
+ *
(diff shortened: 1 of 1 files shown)


to266:
...That's over 9000 lines of code in that one file, and only for the prompt hiding. I doubt anyone will actually review the thing... I suggest just merging?

francisco-dlp:
I've tested it in my local machine and it works. Of course I haven't gone through the code myself either nor I intend to do it. :+1:  for merging it.

PR #748, opened 2015-11-03
Remove printing elements

francisco-dlp:
When adding e.g. the Si element to an `EELSSpectrum` HyperSpy prints:

```
Adding Si_L2 subshell
Adding Si_L3 subshell
Adding Si_L1 subshell
Adding Si_L2 subshell
Adding Si_L3 subshell
```

This is also printed every time that the `EELSSpectrum` is deepcopied, what can become annoying.
This PR removes this "feature".


vidartf:
This is OK to merge by me, but we should try to add them back as logging lines in implementation of #749.

Issue #739, opened 2015-10-15
Align 2D should apply Sobel filter before Hanning filter


vidartf:
Currently, `hyperspy.misc.image_tools.estimate_image_shift` applies the hanning filter before the sobel filter. This means that a lot of spurious edges will be made, as the gradient of the hanning creates concentric circles in the intensity, making the shift estimation error prone. If instead the hanning filter is applied last, the registration becomes much more accurate.

[Relevant lines in code](https://github.com/hyperspy/hyperspy/blob/master/hyperspy/misc/image_tools.py#L151-L156):
```python
if hanning is True:
   im *= hanning2d(*im.shape)
if medfilter is True:
    im[:] = sp.signal.medfilt(im)
if sobel is True:
    im[:] = sobel_filter(im)
```
Switch to:
```python
if medfilter is True:
    im[:] = sp.signal.medfilt(im)
if sobel is True:
    im[:] = sobel_filter(im)
if hanning is True:
   im *= hanning2d(*im.shape)
```
I think that this will also improve the median filter, but I'm not exactly sure how that works, so the effect on that needs to be checked.

francisco-dlp:
It makes sense, would you like to send a PR?

vidartf:
PR'd as #752.

PR #752, opened 2015-11-06
ENH: Reorder filters in `estimate_image_shift()`

vidartf:
Reordered filters in `misc.image_tools.estimate_image_shift()` as discussed in #739.

Also fixed a rather strange default option for the sobel filter's `mode` argument: This setting is really only important at image edges. The default value (suggested change) assures that the sobel is always zero at the edges. The previous setting of `'constant'` is only significant when not using a hanning filter, but if a hanning is not used it basically introduces a strong edge along the outside of the image for any non-zero edges, basically locking the image in place.

### DIFF
--- a/hyperspy/misc/image_tools.py
+++ b/hyperspy/misc/image_tools.py
@@ -148,12 +148,12 @@ def estimate_image_shift(ref, image, roi=None, sobel=True,
 
     # Apply filters
     for im in (ref, image):
-        if hanning is True:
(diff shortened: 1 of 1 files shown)


vidartf:
I'm guessing the strange `mode` used previously was due to a confusion about the time of application: The expansion determined by `mode` is applied to the input, not the output. If it was applied at the output, `'constant'` would be the obvious choice.

PR #753, opened 2015-11-06
ENH: Updating plot in `estimate_image_shift`

vidartf:
Previously for `Signal.estimate_shift2D()`, if the argument `plot` was `True`, a figure would be created for each alignment step. This PR gives the option to pass a value of `'reuse'` to `plot`, which well then make it reuse a single plot for each frame. This effectively allows the live monitoring of the alignment process, but without creating more than one figure (i.e. only the final frame is accessible after alignment finishes). This is useful when you have a large alignment job that needs to be monitored (e.g. to watch for divergence), but where its unfeasible with hundreds or thousands of figures.

As an alternative, you can also pass a `matplotlib.figure.Figure` as `plot`, which will then be used, but I'm not sure if this has any advantages. It just simply follows as a "free" alternative because of how the code is structured.


hyperspy/misc/image_tools.py
```diff
@@ -170,15 +170,34 @@ def estimate_image_shift(ref, image, roi=None, sobel=True,
...
+        if full_plot:
+            axarr[0].set_title('Reference')
+            axarr[1].set_title('Image')
+            axarr[2].set_title('Phase correlation')
+            axarr[0].imshow(ref)
+            axarr[1].imshow(image)
+            d = (np.array(phase_correlation.shape) - 1) // 2
+            extent = [-d[1], d[1], -d[0], d[0]]
+            axarr[2].imshow(np.fft.fftshift(phase_correlation),
+                            cmap=plt.cm.jet, extent=extent)
```
to266: Why `jet`? I think anything would be better. You could use [`viridis`](http://matplotlib.org/1.5.0/examples/color/colormaps_reference.html), as it will be the new default anyway. Also strange that the colormap is set explicitly, but you can't pass your own value...

francisco-dlp: `viridis` is only available in the very latest mpl release. I don't think that it is justified to restrict our compatibility to mpl > 1.5 just because of this.

+1000 on anything-but-jet, and +1 to using the default cmap that will end up being viridis for mpl >= 1.5

francisco-dlp:
A drawback of "reuse" is that, when the operation time is short, the figure updates too fast for inspection. What about storing the images in `Signal` instances and returning them. The could also be plotted and updated during the iteration loop. This would have the extra advantage of using the blitting capatilities of the toolkit and hence updating faster.

vidartf:
> A drawback of "reuse" is that, when the operation time is short, the figure updates too fast for inspection.

If the operation is fast, it is less likely to need continuous feedback, but not completely unlikely; so yes, it would definitely be nice to plot it better/faster (does it flicker for you?). Doing it by blit is of course the best way.

> What about storing the images in Signal instances and returning them.

That could be an option, but the reference/current is not needed as a new Signal since they're basically from the input signal (so it would be created only for the purpose of plotting). Further, an argument against this is the increased memory use. For large datasets, which is when the continuous monitoring would be the most useful, the increase in memory use would be significant. As such, it could be added as a future option, but I don't think it should replace the solution proposed.

hyperspy/misc/image_tools.py
vidartf: Used jet initially to get it to feel like DM's drift correction, and then never thought to change it. Putting it as an optional kwarg is of course doable, but would expand the already very large function signature of `Singal.estimate_shift2d()`. Leaving it to default means it gets plotted in the same cmap as as the two other frames, and it would be nice to differentiate it. We could of course try to plot it as an outline plot? :) Something with reasonable contrast in the high-intensity regime would be at least be useful, as it can help differentiate local maxima that are close to the global max.

francisco-dlp:
This PR doesn't do it claims to do:

> This PR gives the option to pass a value of 'reuse' to plot, which well then make it reuse a single plot for each frame.

Actually, it replaces the old behavior, which is fine for me as it was horrible :) However, the header must be amended to explain exactly what it does.

I agree with @to266 in the "jet" issue. Making HyperSpy feel like DM it is not our aim. Also, I don't see why the correlation plot must use a different cmap. +1 on using the default cmap or to add a `cmap` argument that affect all plots.

vidartf:
> This PR doesn't do it claims to do:

>>    This PR gives the option to pass a value of 'reuse' to plot, which well then make it reuse a single plot for each frame.

> Actually, it replaces the old behavior [...]

Strange, for me it works like I'd expect. Can you detail what you do, what you expect should happen, and what actually happens? There seem to be some effect of backend/interactivity, so if you could list those settings as well, it would help out a lot.

hyperspy/misc/image_tools.py
vidartf: How about putting this as a preference? It's not something where I would imagine people would change cmap from time to time, but they might have a general preference. However, I'm not sure what the guidelines are for making something a setting (e.g. if this is too trivial).

dnjohnstone:
I'd quite like to get this one wrapped up, along with #708 - @vidartf are you still interested in this?

vidartf:
I nuked and redid on moved files. Needs review.

hyperspy/_signals/signal2d.py
```diff
@@ -176,15 +180,34 @@ def estimate_image_shift(ref, image, roi=None, sobel=True,
...
+                for i in range(3):
+                    f.add_subplot(1, 3, i)
+                axarr = plot.axes
+        else:
+            f, axarr = plt.subplots(1, 3)
+        full_plot = len(axarr[0].images) == 0
+        if full_plot:
+            axarr[0].set_title('Reference')
+            axarr[1].set_title('Image')
+            axarr[2].set_title('Phase correlation')
```
francisco-dlp: The title should be "correlation" when `normalize_corr` is False.

francisco-dlp:
I still think that the proper way to fix this is by storing or the correlations in a signal and returning it, maybe with the option to plot it as the calculation progresses. What do you think?

hyperspy/_signals/signal2d.py
vidartf: This is simply the existing behavior, and as such not really part of this PR :)

vidartf:
@francisco-dlp I think what you are proposing is a different feature than what this PR is attempting to do, and that is can be implemented independently from this. Building up a correlation signal will double the memory use, and as such, plotting without creating this signal would still be useful.

dnjohnstone:
if storing the correlation signal does become a thing, it needs to be optional because of the memory issue - in any case these are certainly separate issues.

francisco-dlp:
I agree. However, it'll be still good to use HyperSpy's image plotting features that automatically use the blitting feature to speep up the update of the image. This is as simple as putting the image data into a `Signal2D` instances and triggering the `data_changed` when updating the data.

francisco-dlp:

PR #754, opened 2015-11-06
restructured signal slicing code to be more general

to266:
..essentially I had it in https://github.com/hyperspy/hyperspy/pull/720, but it includes model slicing and other things, and I just need it quicker :)


to266:
deprecated by https://github.com/hyperspy/hyperspy/pull/720

PR #758, opened 2015-11-09
Fix travis

francisco-dlp:
As reported by @vidartf in #547, travis was not complaining when tests failed. This fixes it.

### DIFF
--- a/continuous_integration/nosetest.py
+++ b/continuous_integration/nosetest.py
@@ -1,10 +1,10 @@
-"""Run nosetests after setting ETS toolkit to "null
-"
-"""
 #!/usr/bin/python
-import nose
-from traits.etsconfig.api import ETSConfig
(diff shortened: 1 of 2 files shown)


to266:
...we should definitely find a way to merge into both master and 0.8.x at the same time

francisco-dlp:
I've just done it  manually once more. Maybe this is what we need: https://github.com/liveperson/Auto-Merger?

to266:
Yes, it provides the functions we need, however for svn, not git - so to use we'd probably need to not only rewrite the thing, but also somehow install it on github... Which I'm not sure is possible.

Also, my general impression is that we are working not in the usual way in general - I thought usually people use `master` as "bugfix" branch, and `v0.9` branch for fancy new things, that eventually get merged into `master` upon release..?

PR #766, opened 2015-11-23
Blockfile reader/writer

vidartf:
Adds support for reading/writing blockfiles created by the scanning precession electron diffraction software ASTAR.

TODO
--------
* [x] Figure out metadata
* [x] Add writing capability
* [x] Unittests
    * [x] dict2sarray
    * [x] saving from new signal
* [x] User Guide
 * [x] Document
 * [x] Add proprietary format warning
* [x] Update `load` docstring


francisco-dlp:
Nice. Wouldn't it be possible to ask the manufacture about the metadata?

Anyway, once unittests are ready and the new format is documented in the User Guide, I think that this will be ready to merge.

vidartf:
Should be ready!

hyperspy/io_plugins/blockfile.py
```diff
@@ -0,0 +1,241 @@
...
+    # Every frame is preceeded by a 6 byte sequence (AA 55, and then a 4 byte
+    # integer specifying frame number)
+    data = data[:, :, 6:]
+    data = data.reshape((NY, NX, DP_SZ, DP_SZ), order='C')
+
+    units = ['nm', 'cm', 'cm', 'nm']
+    names = ['x', 'dy', 'dx', 'y']
+    scales = [header['SX'], SDP, SDP, header['SY']]
+    metadata = {'General': {'original_filename': os.path.split(filename)[1]},
+                "Signal": {'signal_type': "",
```
francisco-dlp: Is this format used for anything other than diffraction data? If not, "signal_type" should be "diffraction". Currently we don't have a `DiffractionImage` signal, but I think that there is one on the works.

hyperspy/misc/array_tools.py
```diff
@@ -125,3 +125,33 @@ def sarray2dict(sarray, dictionary=None):
...
+
+    Returns
+    -------
+    Structure array
+
+    """
+    if sarray is None:
+        sarray = np.zeros((1,), dtype=dtype)
+    for name in set(sarray.dtype.names).intersection(set(dictionary.keys())):
+        if len(sarray[name]) == 1:
```
francisco-dlp: Why do you need to do it differently for `len(sarray[name]) == 1`?

hyperspy/misc/array_tools.py
```diff
@@ -125,3 +125,33 @@ def sarray2dict(sarray, dictionary=None):
         dictionary[name] = sarray[name][0] if len(sarray[name]) == 1 \
             else sarray[name]
     return dictionary
+
+
+def dict2sarray(dictionary, sarray=None, dtype=None):
```
francisco-dlp: This function needs a test.

hyperspy/tests/io/test_blockfile.py
```diff
@@ -0,0 +1,128 @@
...
+
+
+def test_load2():
+    s = hs.load(file2)
+    nt.assert_equal(s.data.shape, (2, 3, 5, 5))
+    np.testing.assert_equal(s.axes_manager.as_dictionary(), axes2)
+    np.testing.assert_allclose(s.data, ref_data2)
+
+
+def test_save_load_cycle():
```
francisco-dlp: You should also test saving a signal that was not previously loaded from a blo file.

hyperspy/io_plugins/blockfile.py
```diff
@@ -0,0 +1,241 @@
...
+    dictionary = {'data': data,
+                  'axes': axes,
+                  'metadata': metadata,
+                  'original_metadata': original_metadata,
+                  'mapping': mapping, }
+
+    return [dictionary, ]
+
+
+def file_writer(filename, signal, **kwds):
```
francisco-dlp: So far we haven't supported writing to proprietary formats because we can't guarantee that the proprietary software will be able to read the file.  However, I see that it can be useful in this and other cases. What about printing a warning so that the user is aware that the file is proprietary and we can't guarantee interoperability?

francisco-dlp:
As far as I can tell, User Guide documentation is still missing. This deserves its entry in the "Loading and saving data" chapter. Also, the `load` docstring needs updating.

hyperspy/io_plugins/blockfile.py
vidartf: My suspicion is that the two bytes at offset 6-7 (`header['MAGIC']`) actually specify format version, because it is quite a common convention to detect filetypes by header signature patterns. If so, we can simply specify which versions are supported and throw a specific warning if not. I'll try to confirm this with the manufacturer. Either way, I think the warning belongs in the user guide, and/or possibly at a logging level below warning (information might be the most appropriate).

francisco-dlp: Yes, thinking about it, getting those warnings that I was proposing can
easily become annoying. I agree that adding a warning box in the User Guide
and in `Signal.save` is the best approach.



2015-12-05 1:02 GMT+00:00 Vidar Tonaas Fauske <notifications@github.com>:

hyperspy/misc/array_tools.py
vidartf: If the length is 1, we assign the value (scalar/object) to the dict, if not, we assign the entire ndarray. Simply the reverse of what happens in sarray2dict.

vidartf:
Think everything should be in order now. However, I wasn't able to get sphinx working locally, so I couldn't verify that the RST is correct.

doc/user_guide/io.rst
```diff
@@ -365,4 +367,43 @@ Therefore strongly reccommend to load using the ``.emi`` file instead.
...
+Blockfile
+---------
+
+HyperSpy can read and write the blockfile format from NanoMegas ASTAR software.
+It is used to store a series of diffraction patterns from scanning precession
+electron difraction (SPED) measurements, with a limited set of metadata. The
+header of the blockfile contains information about centering and distortions
+of the diffraction patterns, but is not applied to the signal during reading.
+
+.. warning::
```
francisco-dlp: I would also warn the user about:

* This format only supporting uint8
* We provide no guarantee that they'll be able to open the file with the ASTRA software.

doc/user_guide/io.rst
```diff
@@ -365,4 +367,43 @@ Therefore strongly reccommend to load using the ``.emi`` file instead.
...
+
+   While Blockfiles are supported, it is an proprietary format, and future
```
francisco-dlp: Typo: "[...] a proprietary format [...]"

doc/user_guide/io.rst
```diff
@@ -365,4 +367,43 @@ Therefore strongly reccommend to load using the ``.emi`` file instead.
...
+    >>> hs.load('file.blo', load_to_memory=True)    # Load directly to memory
+    >>> # Default memmap loading:
+    >>> hs.load('file.blo', load_to_memory=False, mmap_mode='c') 
+
+    >>> # Loads data read only:
+    >>> hs.load('file.blo', load_to_memory=False, mmap_mode='r')
+    >>> # Loads data read/write:
+    >>> hs.load('file.blo', load_to_memory=False, mmap_mode='r+')
 
+By loading the data read/write, any changes to the orignal data array will be
```
francisco-dlp: Typo: original

doc/user_guide/io.rst
```diff
@@ -365,4 +367,43 @@ Therefore strongly reccommend to load using the ``.emi`` file instead.
...
+    >>> # Default memmap loading:
+    >>> hs.load('file.blo', load_to_memory=False, mmap_mode='c') 
+
+    >>> # Loads data read only:
+    >>> hs.load('file.blo', load_to_memory=False, mmap_mode='r')
+    >>> # Loads data read/write:
+    >>> hs.load('file.blo', load_to_memory=False, mmap_mode='r+')
 
+By loading the data read/write, any changes to the orignal data array will be 
+written to disk. The data is written when the orignal data array is deleted, or
```
francisco-dlp: Typo: original

hyperspy/io_plugins/blockfile.py
```diff
@@ -0,0 +1,273 @@
...
+format_name = 'Blockfile'
+description = 'Read/write support for ASTAR blockfiles'
+full_support = False
+# Recognised file extension
+file_extensions = ['blo', 'BLO']
+default_extension = 0
+
+# Writing capabilities:
+# writes = False
+writes = [(2, 2)]
```
francisco-dlp: Doesn't the format support line scans?

vidartf: Not as a different type, but I guess we can upconvert (1,Z,Z) and (X,Z,Z) to simply having NX and/or NY as 1.

hyperspy/misc/array_tools.py
francisco-dlp: Not sure about this. The purpose of the similar code in `sarray2dict`
was to avoid using 1d np.array to store metadata. However, in this case
we are actually saving in the array, so, probably there is not point at
doing the same. Also, given that `sarray = np.zeros((1,),
dtype=dtype)`, `len(sarray[<any field>]` should alway be 1 anyway.
Wouldn't `sarray[name] = dictionary[name]` suffice in this case?

hyperspy/io_plugins/blockfile.py
francisco-dlp: That sounds good.

doc/user_guide/io.rst
vidartf: I agree on the datatype warning. Regarding limitations of guarantees, I'm not sure I see a scenario where this would be relevant. If somebody overwrites their data using this, and then blames us, I doubt having a notice in the docs will make a difference. If its just to limit liability, there's also this in the license:
> HyperSpy is distributed in the hope that it will be useful, but WITHOUT ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.

:P
I would agree if this write-support was somehow experimental or incomplete, but at the moment there's basically only one header field that is partially unknown, and that we treat in a quite safe manner.

vidartf: Well, I said that, but I ended up adding one :)

vidartf:
Latest fixes pushed and passing.

francisco-dlp:
Mmm, it seems that I was too quick to merge this, travis is now failing in master...

vidartf:
I've spotted the cause: `Aquistion time` field. Will pass if it runs fast enough :-/

francisco-dlp:
:)

Issue #765, opened 2015-11-23
Cannot import hyperspy magic with pyface v5.0.0


jat255:
The newest versions of `pyface` and `traits-ui` just landed in Arch Linux, and `pyface` borks the importing done by the hyperspy magic. The traceback is below. It appears to be a complaint about `QDate`, which has been already set to a value that conflicts with `pyface`.

```python
>>> %hyperspy

---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
<ipython-input-1-d79832fa989d> in <module>()
----> 1 get_ipython().magic(u'hyperspy')

/usr/lib/python2.7/site-packages/IPython/core/interactiveshell.py in magic(self, arg_s)
   2305         magic_name, _, magic_arg_s = arg_s.partition(' ')
   2306         magic_name = magic_name.lstrip(prefilter.ESC_MAGIC)
-> 2307         return self.run_line_magic(magic_name, magic_arg_s)
   2308 
   2309     #-------------------------------------------------------------------------

/usr/lib/python2.7/site-packages/IPython/core/interactiveshell.py in run_line_magic(self, magic_name, line)
   2226                 kwargs['local_ns'] = sys._getframe(stack_depth).f_locals
   2227             with self.builtin_trap:
-> 2228                 result = fn(*args,**kwargs)
   2229             return result
   2230 

/home/josh/git_repos/hyperspy/hyperspy/misc/hyperspy_magics.pyc in hyperspy(self, line)

/usr/lib/python2.7/site-packages/IPython/core/magic.py in <lambda>(f, *a, **k)
    191     # but it's overkill for just that one bit of state.
    192     def magic_deco(arg):
--> 193         call = lambda f, *a, **k: f(*a, **k)
    194 
    195         if callable(arg):

/home/josh/git_repos/hyperspy/hyperspy/misc/hyperspy_magics.pyc in hyperspy(self, line)
     74                              "import hyperspy.api as hs\n"
     75                              "import matplotlib.pyplot as plt\n")
---> 76         exec(first_import_part, sh.user_ns)
     77 
     78         if preferences.General.import_hspy:

<string> in <module>()

/home/josh/git_repos/hyperspy/hyperspy/api.py in <module>()
     57 
     58 from hyperspy.Release import version as __version__
---> 59 from hyperspy import signals
     60 from hyperspy.io import load
     61 from hyperspy.defaults_parser import preferences

[...]

Issue #455, opened 2015-01-29
Getting HyperSpy in Debian repository


magnunor:
It would be very nice having HyperSpy releases in the Debian repository, and since Ubuntu basically use the Debian repository it would be available there as well.

I'm guessing it won't be too complicated, since we already have a working Debian package. Someone just have to jump through the various Debian http://www.debian.org/doc/manuals/maint-guide/ hoops.

I could probably have a look at it, and see how it goes :)

francisco-dlp:
This is much needed. We create the debian package using stdeb, which most likely doesn't produce a Debian compliant package. In addition to  the document that you mention it might be worth having a look at the  [Debian Python packaging guide](https://wiki.debian.org/Python/Packaging).

magnunor:
I've had a little look into this, and Debian is planning on phasing out Python 2 in favour of Python 3.
What is the status for getting HyperSpy to run in Python 3?

to266:
Basically `traitsui` is the only thing holding us back. We can either get rid of it (`traitsui` only) and replace it with Qt, or replace both `traits` and `traitsui` with `traitlets` and (if I understood that correctly) Qt respectively.

I don't think we started any of those steps, but I would be willing to help after the 0.9 release (which will have a major contribution from my side, so probably I will have enough things to work on)

vidartf:
I could probably help out if we're porting to use Qt. It would be good to do that at the same time as we implement the proposed event structures (#423).

to266:
I'm all for it. Starting next week though ;)

magnunor:
Reviving this again.

Anything blocking this? Any dependencies which are not in the Debian repositories?

magnunor:
I've started looking into this (again). Seeing if I can get it to build locally on my computer using the Debian Python build system.

magnunor:
Also, getting HyperSpyUI into the Debian repositories would be nice. I'll have a look at that as well when (or if) we get HyperSpy included.

francisco-dlp:
Did  you have success with your debian building experiment?

I think that all dependencies are in Debian. The only one that is not is dask (required for #1219) but they're currently [packaging it](https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=817777).

magnunor:
I got the build system running yesterday, but was missing some packages so the building failed. I'll look into it (probably) during the weekend.

magnunor:
[...]

PR #767, opened 2015-11-23
Fix windows traitui5

francisco-dlp:
Workaround/Fix? #765

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,8 +16,6 @@ install:
   - pip install coverage coveralls
   # The following two lines set hyperspy to run headless. This is a temporary
   # workaround until we fix hyperspy so that this is not required.
-  - mkdir ~/.hyperspy
-  - printf "[General]\ndefault_toolkit = None" > ~/.hyperspy/hyperspyrc
   - python setup.py install
 
 before_install:
--- a/continuous_integration/nosetest.py
+++ b/continuous_integration/nosetest.py
@@ -7,4 +7,6 @@ if __name__ == '__main__':
     from traits.etsconfig.api import ETSConfig
 
     ETSConfig.toolkit = "null"
+    import matplotlib
+    matplotlib.use("Agg")
     sys.exit(run_exit())
--- a/hyperspy/__init__.py
+++ b/hyperspy/__init__.py
@@ -1,13 +1,15 @@
 # -*- coding: utf-8 -*-
-
-import sip
-sip.setapi('QVariant', 2)
-sip.setapi('QDate', 2)
-sip.setapi('QDateTime', 2)
-sip.setapi('QTextStream', 2)
-sip.setapi('QTime', 2)
-sip.setapi('QUrl', 2)
-del sip
+try:
+    import sip
+    sip.setapi('QVariant', 2)
+    sip.setapi('QDate', 2)
+    sip.setapi('QDateTime', 2)
+    sip.setapi('QTextStream', 2)
+    sip.setapi('QTime', 2)
+    sip.setapi('QUrl', 2)
+    del sip
+except ImportError:
+    pass
 
 from hyperspy import docstrings
 
--- a/hyperspy/__init__.py
+++ b/hyperspy/__init__.py
@@ -1,5 +1,14 @@
 # -*- coding: utf-8 -*-
 
+import sip
+sip.setapi('QVariant', 2)
+sip.setapi('QDate', 2)
+sip.setapi('QDateTime', 2)
+sip.setapi('QTextStream', 2)
+sip.setapi('QTime', 2)
+sip.setapi('QUrl', 2)
+del sip
+
 from hyperspy import docstrings
 
 __doc__ = """
--- a/hyperspy/api.py
+++ b/hyperspy/api.py
@@ -1,5 +1,16 @@
 # -*- coding: utf-8 -*-
 
+# Set the PyQt API to 2 to avoid incompatibilities between matplotlib
+# traitsui
+import sip
+sip.setapi('QVariant', 2)
+sip.setapi('QDate', 2)
+sip.setapi('QDateTime', 2)
+sip.setapi('QTextStream', 2)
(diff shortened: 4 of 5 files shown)


jat255:
Works for me on Arch Linux using the following packages:

```
python2-pyface 5.0.0-1
python2-traits 4.5.0-1
python2-traitsui 5.0.0-1
```

Importing the hyperspy magic gives me the following, and everything seems to work alright.
```python
>>> %hyperspy

HyperSpy imported!
The following commands were just executed:
---------------
import os
os.environ['QT_API'] = 'pyqt'
import sip
sip.setapi('QDate', 2)
sip.setapi('QDateTime', 2)
sip.setapi('QTextStream', 2)
sip.setapi('QTime', 2)
sip.setapi('QUrl', 2)
%matplotlib qt4
import numpy as np
import hyperspy.api as hs
from hyperspy.hspy import *
```

to266:
...it's getting ridiculous. We should get rid of traits asap. @vidartf, would switching from `traits` and `traitsui` to `traitlets` and some kind of simple UI (Qt?) be terrible for hyperspyui?

vidartf:
@to266 If it switches to something that is compatible with Qt, it is no problem. But almost everything that supports PyQt on python v2 has to deal with `sip.setapi` at some point, to my experience. See http://wiki.qt.io/Differences_Between_PySide_and_PyQt.

I think for this PR, it should be sufficient to call this:
```python
sip.setapi('QString', 2)
sip.setapi('QVariant', 2)
```
If so, it would at least simplify the code a bit.

vidartf:
PS: Even IPython's qtconsole requires PyQt's v2 API for interoperability, so order is important even then, c.f. https://github.com/ipython/ipython/issues/276.

to266:
`traitsui` is the reason why we are not in python 3.x. Hopefully we won't have to care about the problems once we actually rewrite. 

The problem is that the task would be significantly big, however in the end the undertaker would receive no or very little reward

vidartf:
> `traitsui` is the reason why we are not in python 3.x.

Version 5.0 of `traitsui` builds on 3.4 (https://travis-ci.org/enthought/traitsui/builds/87771060), so as far as I can see that is no longer true. I did some tests previously against the dev branch of `traitsui` to see how compatible hyperspy was, and there is quite a bit more than just running the `2to3` script to get python 3 support.

vidartf:
From the 5.0 changelog:
> What's new in Traits UI 5.0.0
> This release features experimental support for Python 3 with the Qt toolkit!

to266:
Good to know, thanks. I think I just may give it a go....

francisco-dlp:
Thank you for the news. This being the case, there is nothing that stops us from supporting Python 3 anymore. Let's do it!

magnunor:
Nice! Getting HyperSpy over to Python 3 would also make it easier getting it into the Debian repository (https://github.com/hyperspy/hyperspy/issues/455)

vidartf:
Coming back on-topic for this PR:
For now, I'd suggest switching the PR to use
```python
sip.setapi('QString', 2)
sip.setapi('QVariant', 2)
```
as long as that works as I think it will (i.e. that other types inherit API version from QVariant). Then this should go into 0.8.x. If we get support for python3 in 0.9, this can then be removed (but should be kept in the python2 version).

to266:
@vidartf , @francisco-dlp mentioned that only adding the two lines that you suggest is not actually enough - he tried that before.

vidartf:
@to266 Ok, then it might have changed from before, or my memory was simply wrong ;)

francisco-dlp:
@vidartf, most likely you're right but I have to test it in the windows
machine where I faced the problem as enthought 5s didn't yet landed in
Debian and anaconda. I kept on adding `setapi` entries until it stopped
complaining, but I didn't try QVariant. Thanks for the trick.

Francisco

2015-11-25 21:57 GMT+00:00 Vidar Tonaas Fauske <notifications@github.com>:

vidartf:
@francisco-dlp : Update PR to `QVariant`, and then merge?

francisco-dlp:
I've tested what you propose but it doesn't work in my system, should we merge it as it is?

vidartf:
As you can see from the latest CI, moving the code to `hyperspy.api` breaks the build.

francisco-dlp:
It should be fixed now, I''ve moved it to `hyperspy/init.py"

2015-12-18 15:05 GMT+00:00 Vidar Tonaas Fauske <notifications@github.com>:

francisco-dlp:
@vidartf, is this ready to merge?

vidartf:
@francisco-dlp : From my point of view, yes. But #805 confused me a bit. Does #805 supersede this? Undo this? Extended it? I kind of feel I've lost the overview here.

francisco-dlp:
No, none of them. One of the reasons why we used to provide a start script and later a magic was to set up the toolkits properly. With old IPython versions a start script was necessary, but this is not the case any more. We introduced the magic to replace the start script, but, by moving `setapi`  from the magic to `hyperspy.__init__`, we no longer require the magic either. That's the point of the discussion in #805. Regardless of the outcome of that discussion, in my opinion it is right to move `setapi` to `hyperspy.__init__`. If you agree, could you consider merging this branch?

vidartf:
Do we have to add a dependency on `sip` in `setup.py` etc.?

francisco-dlp:
Good point. I would suggest to place the code in a  `try` statement
instead. Do you agree?

2015-12-29 16:20 GMT+00:00 Vidar Tonaas Fauske <notifications@github.com>:

vidartf:
Sure, a `try`/`except ImportError` bracket should fix any issues. I agree that it is a better fix, since `sip` is linked to PyQt only, which is not a requirement.

francisco-dlp:
Done, thanks!

2015-12-29 17:05 GMT+00:00 Vidar Tonaas Fauske <notifications@github.com>:

Issue #768, opened 2015-11-24
Migration to Python 3


francisco-dlp:
Now that all of our dependencies has finally been ported to Python 3 (thanks @vidartf  for poiting this out) we should start considering porting HyperSpy too.

I'm opening this issue to discuss how to proceed.

A number of core packages support both Py2 and 3. Many of them using the [six library](https://pythonhosted.org/six/).  Personally I would prefer to port HyperSpy to Python 3 for good without relying in any extra library dependency to maintain support with Py3. Then we could run 3to2 or [these other tools](http://python-future.org/) to create the Py2 versions for the releases until we drop Py2 support. What do you think?

TODO
--------
(This is a draft, all is open for discussion)

* [ ] Merge as much as possible for HyperSpy 0.9
* [ ] Run 2to3 + manual polishing
* [ ] Run to 3to2 or similiar to support Py2 just for release.

francisco-dlp:
Regarding to timing, I think that this could be the last task before releasing 0.9. We should aim at merging most pending branches before running 2to3...

vidartf:
Note that `traitsui`'s support for python3 is currently listed as "experimental", and does not support the wx backend. We should therefore test the port (#769) to see if there are anything that gets broken, and if so, submit bug reports to `traitsui`'s repo. If there's any showstoppers, I would be -1 on switching master to pure python3 until those are fixed.

Regarding the use of `six` vs using pure python3 with back-porting, I'm not sure. If we go pure python3, then the dev versions become incompatible with other packages that are only on python2. Not sure if that is an issue. Other than that it depends on how much manual tweaking is necessary for backporting (I guess #769 should illuminate that).

jrminter:
This is timely. I just noticed that Ubuntu is dropping python 2.7 from 16.04LTS (https://wiki.ubuntu.com/Python/FoundationsXPythonVersions)

> Feature freeze is the drop dead date for dropping Python 2.7 from the images. We will be hard-nosed about this! Once python2.7 and libpython2.7 are removed from a seed, we'll blacklist them so they can't be accidentally reseeded.

Perhaps this will spur some of the package maintainers to accelerate their efforts.

magnunor:
To clarify: Ubuntu is dropping Python 2.7 from the default install, so the install ISO will only contain Python 3. However, Python 2.7 will still be in the repository for a very long time, probably as long as Debian has it.

But we should still get HyperSpy over to Python 3 as soon as possible :)

[...]

Issue #749, opened 2015-11-04
Implement logging


vidartf:
Depending on your situation, you might want varying degrees of output from hyperspy (and its dependencies). Currently, the code includes the sporadic `print` statements, but its use is inconsistent, as highlighted in PR #748. A better alternative would be to use the [standard logging package](https://docs.python.org/2/library/logging.html), as that allows for a grading of importance of messages, which can then be filtered before being displayed to the user. It would also ease development, as a detailed/verbose output can help when debugging and/or trying to understand exactly what something does.

In the short term, this would only affect how `print` statements are removed (i.e. replaced by logging calls), but in the long term this would hopefully also include new logging statements in existing code. The most important logging level to include would be `information`, as that would serve best to replace today's print statement. Also, `critical`/`error`/`warning` is already handled by exceptions/warnings. Then `debug` could be added slowly over time (unless someone volunteers for performing an overhaul).

Specifically, I'd suggest using named loggers:
> A good convention to use when naming loggers is to use a module-level logger, in each module which uses logging, named as follows:
> `logger = logging.getLogger(__name__)`

francisco-dlp:
Good proposal, I fully agree.

to266:
Agreed

vidartf:
While I think most current `print` statements can fit well with the standard logging levels, we should consider if there are any custom log levels that we want to add (especially if the current `print` statements would fit better into these levels).

I see that some people like to add full tracing (function entry/exit) on the debug level. If we do something similar, I think it should be added below the `DEBUG` level (e.g. a custom level `TRACE`). We might also consider another level between `INFORMATION` and `DEBUG` which can give user relevant information that would normally flood the log (e.g. a custom `VERBOSE` level).

If such custom levels are not appropriate for the current `print` statements, then changing the current code is simple (search and replace with `INFORMATION` level, possibly with `WARNING` level for a few). Either way I think we should establish a convention/guideline for future additions early on to keep things consistent.

vidartf:
This should supersede `hyperspy.messages`.

dnjohnstone:
@vidartf Is this one done now? I thought #877 would close this, or did I miss something that remains to be done?

PR #769, opened 2015-11-24
ENH python3

to266:
Read https://github.com/hyperspy/hyperspy/issues/768 too late, so here is a ~~starting point~~ finished PR.

## Stuff that will be done / is done

- [x] 2to3
- [x] manual checking of the 2to3 changes
- [x] remove non-descriptive signal slicing (https://github.com/hyperspy/hyperspy/issues/388) (as it conflicts for some reason, and we are going to remove it any way - so might as well not fix the problem that we don't have to have)
- [x] stop using `.encode('utf8')` everywhere, so that the default string we use is unicode (i.e. `str` in python3). Notable exception / problem is `slugify` - at the moment we use `bytes.translate(None, deletechars)` to get rid of the characters we do not want - such function is not straightforwardly available for unicode (has the `translate`, but not `deletechars`), so at the moment I actually encode the string, remove unnecessary characters, and then decode back
- [x] remove `misc.natsort.py` in favour of using actual package that does it and supports both pythons (our implementation did not support python3)
- [x] move `traits` and `traitsui` from `conda install` to `pip install` in travis, because conda's traits are not new enough to support python3.5
- [x] changed one of the `.hdf5` example files, because it contained a binary string that was created using `marshal` - apparently they are not guaranteed to be backwards-compatible, and in this case they are not. Replaced with a `dill` - generated binary string. In general should probably remove support for `marshal` when saving dictionaries and make the `dill` mandatory for hyperspy, but that's for another PR
- [x] removed tests that are not required due to removing `signal[0]` slicing support


vidartf:
Note that there is likely to be many overlapping changes here with #749. :)

to266:
@vidartf have you actually started the logging? Though I suppose it does not matter that much - the printing is trivial to track and find.

vidartf:
@to266 : No, I simply wanted to report it while I remembered it. There is another point though, that I'll add on #749.

to266:
....and actually I'm done! :) You are welcome to review things if you'd like ;) 

The most important changes are in the OP

hyperspy/io_plugins/hdf5.py
```diff
@@ -317,14 +318,14 @@ def parse_structure(key, group, value, _type, compression):
         except ValueError:
             tmp = np.array([[0]])
         if tmp.dtype is np.dtype('O') or tmp.ndim is not 1:
-            dict2hdfgroup(dict(zip(
-                [unicode(i) for i in xrange(len(value))], value)),
+            dict2hdfgroup(dict(list(zip(
+                [str(i) for i in range(len(value))], value))),
```
vidartf: I think we can skip the `list` wrap here? dict should be able to work with iterators, right?

hyperspy/misc/material.py
```diff
@@ -33,7 +33,7 @@ def _weight_to_atomic(weight_percent, elements):
     atomic_weights = np.array(
         [elements_db[element]['General_properties']['atomic_weight']
             for element in elements])
-    atomic_percent = np.array(map(np.divide, weight_percent, atomic_weights))
+    atomic_percent = np.array(list(map(np.divide, weight_percent, atomic_weights)))
```
vidartf: Necessary?

hyperspy/signal.py
```diff
@@ -555,10 +555,10 @@ def interpolate_in_between(self, start, end, delta=3,
             disabled=not show_progressbar)
         for i, dat in enumerate(self._iterate_signal()):
             dat_int = sp.interpolate.interp1d(
-                range(i0, i1) + range(i2, i3),
+                list(range(i0, i1)) + list(range(i2, i3)),
```
vidartf: Necessary?

vidartf:
Did a quick skim through the diffs, and only pointed out some places where I think we can improve a bit by switching to iterators instead of lists. This is strictly not necessary, and changes behavior, but since they are already nicely highlighted I figured why not? :)

vidartf:
PS: Did you test any `traitsui` things?

hyperspy/signal.py
to266: Yup
```python
In [11]: range(10) + range(3)
<ipython-input-11-d08adf305200> in <module>()
----> 1 range(10) + range(3)

TypeError: unsupported operand type(s) for +: 'range' and 'range'
```

hyperspy/misc/material.py
to266: Yup:
```python
In [8]: np.array(map(np.divide, wp, aw))
Out[8]: array(<map object at 0x7fe8ac8e6748>, dtype=object)
```

hyperspy/io_plugins/hdf5.py
to266: Did not even notice this one. You're correct, `list` is unnecessary

to266:
I've tested `model.fit_component`, `hs.preferences.gui()` and `s.spikes_removal_tool()`, and all seem to behave nicely.

Any other things I should look for? Thanks for the quick glance, by the way. For some reason I'm very excited for this PR, hopefully we'll merge soon!

vidartf:
> Any other things I should look for?

For 2to3 conversion, not that I know.

In terms of `traitsui`, we're mostly using simple editors, so I think we should be safe. Only other thing I can think of is to test editing an array (e.g. fine structure residuals?).

> For some reason I'm very excited for this PR.

:D

vidartf:
As far as I can see this should be ready for merging. However, I'll leave the timing/final review of it up to @francisco-dlp.

francisco-dlp:
Let's discuss the timing in #768

to266:
Updated

francisco-dlp:
Finally, mayavi for py3: https://github.com/enthought/mayavi/pull/250

to266:
So.... Do we go for it? Let me know and I will update it to be compatible with master. Until then, it's pointless, as almost any change would require me to merge again and again..

francisco-dlp:
We should release 0.8.3 soon. What about having a Py3 0.8.3 branch and
aiming at releasing 0.8.3 for both Py2 and Py3 asap?

master is indeed very active now and there are a number of big branches
that are going to be merged soon, so, what about waiting until the end of
this month to merge this branch into master?

2016-01-21 12:02 GMT+00:00 Tomas Ostasevicius <notifications@github.com>:

vidartf:
Some relevant tips:
 - [ ] In Python 3, strings have `__iter__` attributes, which causes some hard to trace errors. See 7bf06fdcec42e164d5c2ab27e3e49ef7cf474207 for fixes.
 - [ ] Function signatures and how to use result from `exec` changed for code in `Events`, see relevant parts of 9f232d5ccdac325d5d372d3a200275bf13d80784 for fix.
 - [ ] I got some errors related to `DataAxis` not being sortable (related to `unique` calls I think). Could be fixed by adding a `__lt__()` function, as in 9f232d5ccdac325d5d372d3a200275bf13d80784.

vidartf:
- [ ] FEI reader can be fixed like this: 7020d958a98686e8f37733994b5d8d20b4dfcec8

vidartf:
There's one issue for python 3 that might warrant some discussion:
In Python 2, the behavior of `round()` for the edge case of 0.5, 1.5, 2.5 etc. was to round up. For Python 3, it rounds to the nearest multiple of two (0.5 -> 0, 1.5 -> 2, 2.5 -> 2). The Python 3 behavior is also the behavior of numpy's `round()` function on both versions. There are good arguments for this, but it might be an issue for ROIs/widgets or other thing that directly code in .5 values.

For example, a naive ROI with a width of `3 * scale`, and an offset of `0.5 * scale` will slice as `[0:4]`, while the same ROI with an offset of `1.5 * scale` will slice as `[2:4]`. For the Circle widget I solved this by using `(0.5 + 1e-8) * scale` instead of 0.5, but it's a bit cumbersome.

Not sure what to do elsewhere, just thought it should be mentioned.

to266:
Thanks.

I'll update this branch once all the ROI / interactivity code is merged, should not be too long

vidartf:
Sure. The goal for Circle ROIs would always be to have them an odd number of pixels in diameter by default, so that the mask can be symmetric. Feel free to take the opportunity to adjust rounding for the masking at that point as well :)

to266:
Updated: merged both the current master and 0.8.x, so ready for review. Also passes all tests in my machine, but travis seems to be broken (no our tests, travis in general) right now, so only appveyor is available

vidartf:
Note that the current master does not complete CI due to unhandled warnings, so you might have to mix that as well.

vidartf:
I made a PR to the branch in question: to266/hyperspy#15.

to266:
Thanks for the help @vidartf

vidartf:
No problem. I'm rather curios about what is causing the travis issue, but not something I can track down now :)

vidartf:
A possible check is to do a `git checkout <branch with working travis> -- hyperspy/Release.py` to see if there is any issue with the file itself (line endings, encoding, etc).

to266:
Lets see.... Ugh, I should have probably done this in the correct order, but oh well

to266:
Ok so this works and is ready to be merged

vidartf:
I'm all for it. The longer dev is on Python3, the longer we have to root out any remaining issues. Any existing PRs would have to be checked for Python3 compatibility anyways, so easiest if this is also double checked during PR review.

PR #770, opened 2015-11-27
More tests for model, component and parameter

to266:
Things that can already be reviewed:
 - [x] New parameter tests
 - [x] New component tests
 - [x] New model tests

### DIFF
--- a/hyperspy/component.py
+++ b/hyperspy/component.py
@@ -886,10 +886,6 @@ class Component(t.HasTraits):
                                                parameter.std,
                                                parameter.units)
 
-    def __tempcall__(self, p, x, onlyfree=True):
-        self.fetch_values_from_array(p, onlyfree=onlyfree)
-        return self.function(x)
-
     def __call__(self):
         """Returns the corresponding model for the current coordinates
 
--- a/hyperspy/component.py
+++ b/hyperspy/component.py
@@ -752,7 +752,8 @@ class Component(t.HasTraits):
         return text
 
     def _update_free_parameters(self):
-        self.free_parameters = {par for par in self.parameters if par.free}
+        self.free_parameters = sorted([par for par in self.parameters if
+                                       par.free], key=lambda x: x.name)
         self._nfree_param = sum([par._number_of_elements for par in
                                  self.free_parameters])
 
@@ -769,7 +770,7 @@ class Component(t.HasTraits):
         else:
             parameters = self.parameters
         i = 0
-        for parameter in parameters:
+        for parameter in sorted(parameters, key=lambda x: x.name):
             length = parameter._number_of_elements
             parameter.value = (p[i] if length == 1 else p[i:i + length])
             if p_std is not None:
--- a/hyperspy/component.py
+++ b/hyperspy/component.py
@@ -466,8 +466,10 @@ class Parameter(t.HasTraits):
 
         s = Signal(data=self.map[field],
                    axes=self._axes_manager._get_navigation_axes_dicts())
-        if self.component.active_is_multidimensional:
+        if self.component is not None and \
+                self.component.active_is_multidimensional:
             s.data[np.logical_not(self.component._active_array)] = np.nan
+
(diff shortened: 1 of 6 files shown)


hyperspy/component.py
```diff
@@ -910,20 +907,18 @@ def _component2plot(self, axes_manager, out_of_range2nans=True):
         if axes_manager is not self.model.axes_manager:
             old_axes_manager = self.model.axes_manager
             self.model.axes_manager = axes_manager
-            self.charge()
```
francisco-dlp: This was actually a bug: `charge` was renamed to `fetch_stored_values`, but not here. Removing it causes another issue. I would suggest to simply replace `charge` with `fetch_stored_values`.

hyperspy/component.py
```diff
@@ -910,20 +907,18 @@ def _component2plot(self, axes_manager, out_of_range2nans=True):
...
         s = self.__call__()
         if not self.active:
             s.fill(np.nan)
         if self.model.spectrum.metadata.Signal.binned is True:
             s *= self.model.spectrum.axes_manager.signal_axes[0].scale
-        if old_axes_manager is not None:
-            self.model.axes_manager = old_axes_manager
-            self.charge()
```
francisco-dlp: Same as above.

hyperspy/model.py
```diff
@@ -456,7 +451,7 @@ def remove(self, thing, touch=True):
             idx = self.index(thing)
             self.spectrum._plot.signal_plot.ax_lines.remove(
                 self.spectrum._plot.signal_plot.ax_lines[2 + idx])
-        list.remove(self, thing)
+        list.__delitem__(self, self.index(thing))
```
francisco-dlp: why is this better?

to266: Hm... don't remember why I changed it. Probably not better

hyperspy/component.py
to266: I guess it's one of the plotting bugs? Now I can see what might be a problem, I'll add the lines shortly

to266:
All done

hyperspy/model.py
```diff
@@ -960,60 +946,20 @@ def reset_the_signal_range(self):
...
-                                counter:counter +
-                                component._nfree_param],
-                            axis)
-                    counter += component._nfree_param
-            to_return = sum
-
-        if self.spectrum.metadata.Signal.binned is True:
-            to_return *= self.spectrum.axes_manager[-1].scale
+        old_p0 = self.p0
+        self._set_p0()
```
francisco-dlp: Isn't setting p0 unnecessary?

hyperspy/model.py
```diff
@@ -960,60 +946,20 @@ def reset_the_signal_range(self):
...
+        old_values = self.p0
```
francisco-dlp: Backing up p0 should also be unnecessary.

hyperspy/model.py
```diff
@@ -960,60 +946,20 @@ def reset_the_signal_range(self):
...
+        self.p0 = param
+        self._fetch_values_from_p0()
+        to_return = self.__call__(non_convolved=False, onlyactive=True)
+        self.p0 = old_values
+        self._fetch_values_from_p0()
```
francisco-dlp: If the answer to the above comment is yes, then this line can also go.

francisco-dlp: Now we can get rid of `Component.__tempcall__` :)

to266: All this play was only to, in the end of the function call, have the same `self.p0` and all `values`. If it's actually not necessary (i'm not sure if we break anything), then yes, only fetching the `param` values should be enough.

I'll try to make it so and see if anything breaks

to266: And it wasn't :) done

hyperspy/component.py
francisco-dlp: This `charge` has to come back too.

francisco-dlp: No it doesn't, sorry.

Issue #733, opened 2015-09-23
Make our VisibleDeprecationWarnings fail tests


vidartf:
We added our own VisibleDeprecationWarning in #657, but there still exists several hyperspy functions that rely on the deprecated functionality. This kind of defeats the purpose of making the warning visible, as users get the warning for calling normal, non-deprecated hyperspy functions.

The easiest solution to root out most of these cases is to make any of our own deprecation warnings fail the test. We should also consider whether we should simply have all deprecation warnings fail the unit tests.

PR #776, opened 2015-12-06
Make CI fail on warnings

vidartf:
In order to make the repo more polished, this PR proposes to make the continuous integration fail on warnings. Any acceptable warnings (e.g. deprecation warnings) should then be manually allowed in the tests (possibly even asserting that a warning is given as expected).

Main motivation: Some features in master currently rely on deprecated functionality internally, meaning users get deprecation warnings for what should be non-deprecated functions from their point of view. This could possibly lead to users ignoring deprecation warnings, either mentally or by filtering. By enforcing strict CI, these should be weeded out. There are also some warnings that have been there for as long as I can remember, so those should get some extra attention to see if the warnings are really needed or cleaned up somehow.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,6 +19,7 @@ install:
   - mkdir ~/.hyperspy
   - printf "[General]\ndefault_toolkit = None" > ~/.hyperspy/hyperspyrc
   - python setup.py install
+  - export PYTHONWARNINGS=error
 
 before_install:
   # then install python version to test
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,8 +10,8 @@ env:
   - export FAIL_ON_EXTERNAL_DEPRECATION='True'
 
 matrix:
-    allow_failures:
-        env: export FAIL_ON_EXTERNAL_DEPRECATION='True'
+  allow_failures:
+    - env: export FAIL_ON_EXTERNAL_DEPRECATION='True'
     
 
 install:
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,8 +10,8 @@ env:
   - export FAIL_ON_EXTERNAL_DEPRECATION='True'
 
 matrix:
-+  allow_failures:
-+      env: FAIL_ON_EXTERNAL_DEPRECATION='True'
+    allow_failures:
+        env: FAIL_ON_EXTERNAL_DEPRECATION='True'
     
 
 install:
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,9 +6,12 @@ python:
   - 2.7
 
 env:
-  - DEPS="nose numpy scipy matplotlib traits traitsui ipython h5py sympy scikit-learn dill mock"
+  - PYTHONWARNINGS="error,default:DeprecationWarning::(!hyperspy),ignore:UserWarning:Failed to import the optional scikit image package"
+  - PYTHONWARNINGS="error,ignore:DeprecationWarning:BaseException\.message has been deprecated as of Python 2\.6,ignore:UserWarning:Failed to import the optional scikit image package"
+    
 
 install:
+  - export DEPS="nose numpy scipy matplotlib traits traitsui ipython h5py sympy scikit-learn dill mock"
   - conda create -n testenv --yes python=$TRAVIS_PYTHON_VERSION
   - source activate testenv
   - conda install --yes $DEPS
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,7 +11,7 @@ env:
 
 matrix:
     allow_failures:
-        env: FAIL_ON_EXTERNAL_DEPRECATION='True'
+        env: export FAIL_ON_EXTERNAL_DEPRECATION='True'
     
 
 install:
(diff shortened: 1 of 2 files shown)


vidartf:
Shouldn't this fail on checks? It fails as expected when I run `python continuous_integration/nosetest.py --with-coverage hyperspy` locally ...

vidartf:
There we go. Now all the relevant warnings make CI fail.

to266:
Nice, I agree with the proposal

There are warnings, though, that we used only to avoid print statements and that are actually only there to notify the user - e.g. when `active_is_multidimensional` is already `True` for a component. I suppose a better method should have been picked at the time, but there you go.

I suppose the logging should be implemented first / in parallel, in order to remove the just-an-FYI-warning instances?

vidartf:
Sure, I understand why the warnings are there. The goal of this PR is simply to force a classification:
 1. Warnings that should not appear during tests (i.e. tests need fixing)
 2. Warnings that are expected during tests (tests should then either actively ignore, or check for warnings and fail if not given)
 3. Warnings that are simply information to the user (should be fixed by logging PR)

I'm not sure what the `active_is_multidimensional` warning is for, or how much of a problem it is to set it to `True` if it's already set. Depending on this, the warning should either go in category 1, 2 or 3.

francisco-dlp:
@vidartf, I didn't get your point 3, why warnings designed to inform the user should be turned into logging messages? Only those who check the logs will see those warnings then...

I am in favour of fixing our warnings problem. However, having to maintain (i.e. create and keep up to date) a list of warnings to be ignored... not fully sure about it. Notice that we're not the only ones emitting warnings. Say numpy starts emitting a deprecation warning that is triggered by one of our calls to a scipy function. There is nothing to be done on our side, but our tests will fail until we filter the warning. Say we add the filter, then, once the scipy code gets updated, filtering the warning is no longer necessary but we won't be notified about that. Hence, our warning filter list will grow and, chances area that at some point we will be filtering new warnings that we shouldn't filter...

What about keeping this PR always open and use it to help us identify the problems (e.g. before a release), but without actually ever merging it into master?

to266:
Then better create a branch with this PR, as that would not involve @vidartf always keeping this PR up to date
On 7 Dec 2015 12:46 pm, Francisco de la Peña <notifications@github.com> wrote:@vidartf, I didn't get your point 3, why warnings designed to inform the user should be turned into logging messages? Only those who check the logs will see those warnings then...

I am in favour of fixing our warnings problem. However, having to maintain (i.e. create and keep up to date) a list of warnings to be ignored... not fully sure about it. Notice that we're not the only ones emitting warnings. Say numpy starts emitting a deprecation warning that is triggered by one of our calls to a scipy function. There is nothing to be done on our side, but our tests will fail until we filter the warning. Say we add the filter, then, once the scipy code gets updated, filtering the warning is no longer necessary but we won't be notified about that. Hence, our warning filter list will grow and, chances area that at some point we will be filtering new warnings that we shouldn't filter...

What about keeping this PR always open and use it to help us identify the problems (e.g. before a release), but without actually ever merging it into master?

vidartf:
> I didn't get your point 3, why warnings designed to inform the user should be turned into logging messages?

I'm not sure if any such cases exist currently, but if there are any warnings that were simply used as a filterable debug output, that should be done by logging instead.

> Say numpy starts emitting a deprecation warning

The default filter can be set to print but not fail on all deprecation warnings not from our package, see https://docs.python.org/2/library/warnings.html#the-warnings-filter. Or we can use two build profiles: One with deprecations above hyperspy allowed, and one with it disallowed (whose build could then be allowed to fail). By one of these ways, we'll still be kept up to date on deprecation warnings in higher packages that might soon start to cause errors in hyperspy.

vidartf:
There, now this PR is set up with two builds that separate hyperspy warnings and DeprecationWarnings in other modules/packages. The latter is allowed to fail, but can still easily be checked for status.

francisco-dlp:
Nice. Now we only have to do something about all of those warning before merging this...

vidartf:
My suggestion would be to create another PR based on this. When that PR is merged, this one will (should) automatically be set as merged.

to266:
A number of those were (are?) fixed in the python3 PR, where I removed the old deprecated signal slicing and any such usages in tests. Someone can use it as a reference / starting point, if you can ignore wrong python ;)

francisco-dlp:
Sure, I was not suggesting that you take care of that alone :)

I like @to266 suggestion: create a branch in the main repository with this so that you don't have to merge a PR every time that someone rebases it with master to monitor how the warnings are doing. What do you think?

vidartf:
@francisco-dlp Not sure I follow. The point of the split build was to be able to monitor warnings easily without failing the build. As such, it should be able to go into master. I suggested that whoever that wants to fix he warnings should base their code of this PR, because it will make their job easier (tests will work as they should), and once a PR from that branch is merged, this PR will have been merged as well.

vidartf:
This will fix #733.

vidartf:
Since @to266 pointed out that several of the warnings are fixed in #769, this should wait for that to be merged in order to minimize merge conflicts.

PR #778, opened 2015-12-07
Quickfix for #766

vidartf:
#766 had a timing dependent test bug. This should fix it, but please review carefully :)

### DIFF
--- a/hyperspy/io_plugins/blockfile.py
+++ b/hyperspy/io_plugins/blockfile.py
@@ -109,7 +109,9 @@ def get_default_header(endianess='<'):
     header['Data_offset_1'][0] = 0x1000     # Always this value observed
     header['UNKNOWN1'][0] = 131141          # Very typical value (always?)
     header['Acquisition_time'][0] = _to_serial_date(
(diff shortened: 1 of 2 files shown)


vidartf:
Note that while `test_save_load_cycle` compares the `Aquisiton_time` field as well, this should still be correct, as the value from the original file will be kept.

francisco-dlp:
Actually, I'm not sure that setting the Acquisition_time to now is a
good idea as it may be misleading. What about setting it to a default
clearly wrong value instead if the field is required?
Also, now that you're at it, shouldn't `Aquisiton_time field` be
`Acquisition_time`?

vidartf:
Defaulting to UNIX epoch (1970) as default when missing is definitely possible, or the field could be renamed to something like `creation_time`? Not sure what you refer to by `Aquisiton_time field`.

francisco-dlp:
UNIX epoch should be fine.

I was just pointing at a typo: you wrote Aquisition, should be Acquisition. The typo is also present in the code.

vidartf:
Erm, currently `hyperspy.tests.mva.test_decomposition.TestGetExplainedVarinaceRation.test_data` is failing... Other than that I think this is OK now.

PR #780, opened 2015-12-08
Correct .gitignore that was ignoring the API docs and /bin files

francisco-dlp:
As it was, the API rst files were ignored by git. Hence, our docs in readthedocs were partly broken. Also stop ignoring `hyperspy/bin`, except for the `.bat` files.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,7 +2,6 @@ syntax: glob
 .spyderworkspace
 .coverage
 /MANIFEST
-/*.rst
 *.pyc
 *.pyd
 *.swp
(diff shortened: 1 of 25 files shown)


vidartf:
I kinda figured that the api/*.rst files where compiled files?

francisco-dlp:
Yes, they're automatically generated using `sphinx-apidoc`. However, readthedocs doesn't by default execute apidoc, so currently our docs there are missing the API. There might be a way to instruct it to run apidoc, but I couldn't find how...

francisco-dlp:
@vidartf, is this ok then?

vidartf:
I don't really know. I'm not really experienced with sphinx enough to know how it should work :-/

francisco-dlp:
In that case, at least anybody has something against it, I'll merge it to fix the issue with ReadTheDocs. If we find a better solution (i.e. a way to execute sphinx-apidoc in RtD) then it'll be easy to undo it.

PR #473, opened 2015-02-27
New eds model

pburdet:
New `eds_model` as proposed in #390

Example
-----------

```python

    >>> m = create_model(st)
    >>> m.fit()
    >>> m.fit_background()
    >>> m.calibrate_energy_axis('resolution')
    >>> m.calibrate_xray_lines('energy',['Au_Ma'])
    >>> m.calibrate_xray_lines('sub_weight',['Mn_La'],bound=10)

```

TO DO
--------

- [x] DOCstring
- [x] test
- [x] doc
- [x] Add 'scale' and 'offset' options to calibrate_energy_axis
- [x] Add 'width' options to calibrate_xray_lines
- [x] Get line intensity method

Issue #390, opened 2014-10-27
NEW: EDS model


pburdet:
Given the specificity of EDS spectrum, it is important to have an EDSmodel. I'm working on it currently, see branch https://github.com/pburdet/hyperspy/tree/NEW_edsModel

![edsmodel](https://cloud.githubusercontent.com/assets/3855190/4789872/7a4d145c-5dca-11e4-8f63-bb7079a25b3f.PNG)
![edsmodel2](https://cloud.githubusercontent.com/assets/3855190/4789897/d52b2ee0-5dca-11e4-9fe5-fbca17fba37b.PNG)

dnjohnstone:
This is now resolved through #786

PR #786, opened 2015-12-09
NEW: EDSModel

dnjohnstone:
Deprecates #473 and is written to be implemented after #536  so although it looks ridiculously big it won't be very big at all when someone gets round to reviewing #536.

Otherwise I think this should be ready for review.


vidartf:
I trust you meant #536, and not #563 :)

dnjohnstone:
Whoops, yes sorry, corrected that.

dnjohnstone:
With the merge of #536 done this is now updated and ready for review.

hyperspy/hspy.py
```diff
@@ -53,6 +53,26 @@ def create_model(signal, *args, **kwargs):
...
+
+    Examples
+    --------
+
+    With s an EDS instance
+    >>> m = create_model(s)
+    >>> m.mulifit()
+    >>> m.plot()
+    >>> m.get_lines_intensity(plot_result=True)
+
```
vidartf: Not really needed, since this is deprecated for 0.9, but OK :)

hyperspy/_signals/eds_tem.py
```diff
@@ -452,3 +452,30 @@ def decomposition(self,
...
+            If True, automatically add Gaussians for all X-rays generated in the
+            energy range by an element using the edsmodel.add_family_lines
+            method.
+
+        Returns
+        -------
+
+        model : `EDSTEMModel` instance.
+
+        """
```
vidartf: As this replaces the docstring of the underlying method, the parameters of the general model should also be listed here. I don't know if there is an automated way to add that (doubt it), so double typing is probably necessary (copy-paste from Model docstring / deprecated `hspy.create_model()`)

hyperspy/misc/eds/utils.py
```diff
@@ -18,6 +18,10 @@ def _get_energy_xray_line(xray_line):
         line]['energy (keV)']
 
 
+def _get_xray_lines_family(xray_line):
+    return xray_line[:xray_line.find('_') + 2]
```
vidartf: If you know what this does, can you add a docstring? It's not needed for the documentation, but still useful for us devs :)

hyperspy/misc/eds/utils.py
```diff
@@ -243,6 +247,72 @@ def take_off_angle(tilt_stage,
                                   + math.sin(a) * math.sin(c)))
 
 
+def xray_lines_model(elements=['Al', 'Zn'],
```
vidartf: Why does `elements` default to `['Al', 'Zn']` ? `weight_percents` should possibly default to `None`, but can be autofilled to an even distribution depending on the number of elements passed.

hyperspy/misc/eds/utils.py
```diff
@@ -243,6 +247,72 @@ def take_off_angle(tilt_stage,
...
+    from hyperspy.model import components
+    s = EDSTEMSpectrum(np.zeros(energy_axis['size']), axes=[energy_axis])
+    s.set_microscope_parameters(
+        beam_energy=beam_energy,
+        energy_resolution_MnKa=energy_resolution_MnKa)
+    s.add_elements(elements)
+    counts_rate = 1.
+    live_time = 1.
+    if weight_percents is None:
+        weight_percents = [100] * len(elements)
```
vidartf: Should be division?

hyperspy/models/edsmodel.py
```diff
@@ -0,0 +1,822 @@
...
+        weight_line = elements_db[
+            element]['Atomic_properties']['Xray_lines'][line]['weight']
+    return lambda x: x * weight_line
+
+
+def _get_iweight(element, line, weight_line=None):
+    if weight_line is None:
+        weight_line = elements_db[
+            element]['Atomic_properties']['Xray_lines'][line]['weight']
+    return lambda x: x / weight_line
```
vidartf: This utils can possibly be moved to `utils_eds` without the leading underscore.

hyperspy/models/edsmodel.py
```diff
@@ -0,0 +1,822 @@
...
+    return lambda sig_ref: math.sqrt(abs(
+        4.5077 * 1e-4 * (E - E_ref) * units_factor + np.power(sig_ref, 2)))
+
+
+def _get_offset(E, diff):
+    return lambda E: E + diff
+
+
+def _get_scale(E1, E_ref1, fact):
+    return lambda E: E1 + fact * (E - E_ref1)
```
vidartf: Not sure if these utils should be moved as well, but make the consideration.

hyperspy/models/edsmodel.py
```diff
@@ -0,0 +1,822 @@
...
+
+    @spectrum.setter
+    def spectrum(self, value):
+        if isinstance(value, EDSSpectrum):
+            self._spectrum = value
+        else:
+            raise ValueError(
+                "This attribute can only contain an EDSSpectrum "
+                "but an object of type %s was provided" %
+                str(type(value)))
```
vidartf: Wait, should this be `spectrum` or `signal` now? I'm confused :)

hyperspy/models/edstemmodel.py
```diff
@@ -0,0 +1,48 @@
...
+
+    """
+
+    def __init__(self, spectrum, auto_background=True,
+                 auto_add_lines=True,
+                 *args, **kwargs):
+        EDSModel.__init__(self, spectrum, auto_add_lines, *args, **kwargs)
+        self.background_components = list()
+        if auto_background is True:
+            self.add_polynomial_background()
```
vidartf: The `__init__` function can actually be moved to `EDSModel`, only the class definition is currently needed. Same issue for the class docstring here as for `EDSModel`.

vidartf:
Did a quick initial review of the layout/design, sill need to review all the inner workings of the larger functions.

hyperspy/models/edsmodel.py
```diff
@@ -0,0 +1,822 @@
...
+        Parameters
+        ----------
+        start_energy : {float, None}
+            If float, limit the range of energies from the left to the
+            given value.
+        end_energy : {float, None}
+            If float, limit the range of energies from the right to the
+            given value.
+        windows_sigma: list of two float
+            The uppet and lower bounds around each X-ray lines to define
```
siriagus: Typing error in docstring: "The **upper** and lower bounds (...)"

hyperspy/models/edsmodel.py
dnjohnstone: Since it EDSModel inherits from Model1D I think spectrum is correct here... Essentially the structure is that anything in BaseModel is .signal anything in Model1D (and by extension here) is .spectrum and anything in Model2D is .image .... the __init__ for Model1D then days self.signal = self.spectrum for 1D and vice-versa. Does that make sense?

hyperspy/hspy.py
dnjohnstone: Yeah it was just there from Pierre's earlier stuff so I just left it, but I'll delete it so that the diff is smaller since it's about to make no difference...

hyperspy/_signals/eds_tem.py
dnjohnstone: Ok -- I can buy that, but in that case the docstring for def create_model in eels.py has the same issue? (that was what I modeled this on) So shall I correct that too?

vidartf: Ah, I went a bit quickly, and didn't spot that most of the args in `hspy.create_model` are for *either* EDS or EELS models. The only bit missing is this line I guess: `Any extra argument is passed to the Model constructor.`

hyperspy/misc/eds/utils.py
```diff
@@ -243,6 +247,72 @@ def take_off_angle(tilt_stage,
...
+    m = s.create_model()
+    for i, (element, weight_percent) in enumerate(zip(
+            elements, weight_percents)):
```
vidartf: `zip` crops to the output to the shortest list. Maybe we want to check that the lengths correspond first, and raise a `ValueError` if they disagree?

hyperspy/misc/eds/utils.py
```diff
@@ -243,6 +247,72 @@ def take_off_angle(tilt_stage,
...
+        for line in elements_db[
+                element]['Atomic_properties']['Xray_lines'].keys():
+            line_energy = elements_db[element]['Atomic_properties'][
+                'Xray_lines'][line]['energy (keV)']
+            ratio_line = elements_db[element]['Atomic_properties'][
+                'Xray_lines'][line]['weight']
```
vidartf: These lines should read better if replaced as:
```python
for line, properties in elements_db[
        element]['Atomic_properties']['Xray_lines'].iteritems():
    line_energy = properties['energy (keV)']
    ratio_line = properties['weight']
```

hyperspy/models/edsmodel.py
```diff
@@ -0,0 +1,822 @@
...
+
+
+def _get_sigma(E, E_ref, units_factor):
+    # 2.5 from Goldstein, / 1000 eV->keV, / 2.355^2 for FWHM -> sigma
+    return lambda sig_ref: math.sqrt(abs(
+        4.5077 * 1e-4 * (E - E_ref) * units_factor + np.power(sig_ref, 2)))
+
+
+def _get_offset(E, diff):
+    return lambda E: E + diff
```
vidartf: `_get_offset` doesn't need argument `E`, as that is overridden in the lambda function.

hyperspy/models/edsmodel.py
```diff
@@ -0,0 +1,822 @@
...
+                 auto_add_lines=True,
+                 *args, **kwargs):
+        Model1D.__init__(self, spectrum, *args, **kwargs)
+        self.xray_lines = list()
+        self.background_components = list()
+        end_energy = self.axes_manager.signal_axes[0].high_value
+        if self.spectrum._get_beam_energy() < end_energy:
+            self.end_energy = self.spectrum._get_beam_energy()
+        else:
+            self.end_energy = end_energy
```
vidartf:
```python
end_energy = min(end_energy, self.spectrum._get_beam_energy())
```

hyperspy/_signals/eds_tem.py
dnjohnstone: Ah - that identifies a potential slip -- in the create_model method for eels *args, *kwargs are not included... they're here because I based this on the method in hspy.create_model. Do you think we want them (and then your suggested docstring line) or just remove them?

hyperspy/models/edsmodel.py
```diff
@@ -0,0 +1,822 @@
...
+        self.start_energy = self.axes_manager.signal_axes[0].low_value
+        units_name = self.axes_manager.signal_axes[0].units
+        if units_name == 'eV':
+            self.units_factor = 1000.
+        elif units_name == 'keV':
+            self.units_factor = 1.
+        else:
+            raise ValueError("Energy units, %s, not supported" %
+                             str(units_name))
```
vidartf: This is OK, but I'd much rather prefer `start_energy`, `end_energy` and `units_factor` to be (read-only) properties that lazily gets this from the `axes_manager`. Unless there is an added value to being able to overwrite these with custom values?

hyperspy/models/edsmodel.py
```diff
@@ -0,0 +1,822 @@
...
+            else:
+                raise ValueError(
+                    "No elements defined, set them with `add_elements`")
+
+        components_names = [xr.name for xr in self.xray_lines]
+        xray_lines = filter(lambda x: x not in components_names, xray_lines)
+        xray_lines, xray_not_here = self.spectrum.\
+            _get_xray_lines_in_spectral_range(xray_lines)
+        for xray in xray_not_here:
+            print("Warning: %s is not in the data energy range." % (xray))
```
vidartf: Replace with `warnings.warn()`, as we are moving away from print statements.

hyperspy/models/edsmodel.py
```diff
@@ -0,0 +1,822 @@
...
+
+        for i, xray_line in enumerate(xray_lines):
```
vidartf: I can't see that `i` is used, so remove it and `enumerate`.

hyperspy/misc/eds/utils.py
dnjohnstone: I have absolutely no idea why these were the defaults -- I'm going to set elements=None and weight_percents=None.

The only more intelligent thing I can think of is to read the elements from the metadata i.e. s.metadata.Sample.elements - but probably many people won't have set that and it will just cause an error -- on which point do you think I should add 

if elements is None:
raise Warning (no elements specified - please specify some)

hyperspy/models/edsmodel.py
```diff
@@ -0,0 +1,822 @@
...
+            element, line = utils_eds._get_element_and_line(xray_line)
+            line_energy, line_FWHM = self.spectrum._get_line_energy(
+                xray_line,
+                FWHM_MnKa='auto')
+            component = create_component.Gaussian()
+            component.centre.value = line_energy
+            component.sigma.value = line_FWHM / 2.355
```
vidartf: Magic numbers are Bad. Either way, we can simply use:
```python
component.fwhm = line_FWHM
```

hyperspy/models/edsmodel.py
```diff
@@ -0,0 +1,822 @@
...
+                    if self.spectrum.\
+                            _get_xray_lines_in_spectral_range(
+                                [xray_sub])[0] != []:
+                        line_energy, line_FWHM = self.spectrum.\
+                            _get_line_energy(
+                                xray_sub, FWHM_MnKa='auto')
+                        component_sub = create_component.Gaussian()
+                        component_sub.centre.value = line_energy
+                        component_sub.name = xray_sub
+                        component_sub.sigma.value = line_FWHM / 2.355
```
vidartf: Same as above.

hyperspy/models/edssemmodel.py
```diff
@@ -0,0 +1,72 @@
...
+# To do: weight_fraction different for different pixe. (so basckground)
+# Calibrate on standard and transfer dictionnary
+# k-ratios
+
+# import copy
+# import numpy as np
+# import math
+
+from hyperspy.models.edsmodel import EDSModel
+# import hyperspy.components as create_component
```
vidartf: Remove commented code such as these imports throughout the new code.

hyperspy/misc/eds/utils.py
dnjohnstone: Hmmm this is actually a weird bit of code in some ways because [100] is a list object so:

[100] * 2 = [100, 100]

[100] / 2 = Not possible

so it seems this bit exists to make weight_percents have the correct number of numbers in the list. I don't know if there is a sanity check anywhere to make sure that the sum of values in weight_percents = 100 -- but I would suspect that might have been avoided because so many attempted quantifications would fail that test without giving useful information to the user that might help them see what's gone wrong.

hyperspy/models/edstemmodel.py
dnjohnstone: That causes and issue with auto_background, which is not an argument for EDSModel - I think it's better left separate here as we might not want that for SEM data by default and it leaves a starting point for more specific TEM methods

vidartf: You can leave the `__init__` methods empty except for a call to the super, i.e. move
```python 
self.background_components = list()
if auto_background is True:
    self.add_polynomial_background()
```
to `EDSModel`. Also remove `auto_background=True` from signature here and in `EDSSEMModel`, and add it to `EDSModel`.

hyperspy/misc/eds/utils.py
dnjohnstone: Agreed and done.

hyperspy/misc/eds/utils.py
dnjohnstone: yes - done this one and a few of the others above it.

hyperspy/_signals/eds_tem.py
vidartf: Hmm... Let's try again: The current signature of `Spectrum.create_model()` takes only one optional argument: `dictionary=None`. As such, the descendant implementations of `create_model` should include the docstring of that:
```python
        dictionary : {None, dict}, optional
            A dictionary to be used to recreate a model. Usually generated using
            :meth:`hyperspy.model.as_dictionary`
```

hyperspy/models/edsmodel.py
dnjohnstone: I think I agree for units_factor -- so I'll do that.

For start_energy/end_energy isn't it reasonable that someone might want to fit only part of the energy range? Or is that not what setting custom values allows...

hyperspy/models/edsmodel.py
```diff
@@ -0,0 +1,822 @@
...
+    if weight_line is None:
+        weight_line = elements_db[
+            element]['Atomic_properties']['Xray_lines'][line]['weight']
+    return lambda x: x / weight_line
+
+
+def _get_sigma(E, E_ref, units_factor):
+    # 2.5 from Goldstein, / 1000 eV->keV, / 2.355^2 for FWHM -> sigma
+    return lambda sig_ref: math.sqrt(abs(
+        4.5077 * 1e-4 * (E - E_ref) * units_factor + np.power(sig_ref, 2)))
```
vidartf: Magic numbers are Bad, but at least it is documented in the comment somehow (Goldstein? What is order of division? Is eV->keV being handled correctly together with `units_factor`? This can definitely be made more readable / commented more).

hyperspy/models/edsmodel.py
```diff
@@ -0,0 +1,822 @@
...
+        energy_Mn_Ka, FWHM_MnKa_old = self.spectrum._get_line_energy('Mn_Ka',
+                                                                     'auto')
+        FWHM_MnKa_old *= 1000. / self.units_factor
+        get_sigma_Mn_Ka = _get_sigma(
+            energy_Mn_Ka, self[xray_lines[0]].centre.value, self.units_factor)
+        FWHM_MnKa = get_sigma_Mn_Ka(self[xray_lines[0]].sigma.value
+                                    ) * 1000. / self.units_factor * 2.355
+        if FWHM_MnKa < 110:
+            print "FWHM_MnKa of " + str(FWHM_MnKa) + " smaller than " + \
+                "physically possible"
```
vidartf: `warnings.warn()`

hyperspy/models/edsmodel.py
```diff
@@ -0,0 +1,822 @@
...
+        """
+        Adjust the width of all lines and set the fitted energy resolution
+        to the spectrum
+
+        Parameters
+        ----------
+        xray_lines: list of str or 'all_alpha'
+            The Xray lines. If 'all_alpha', fit all using all alpha lines
+        ref: None
+            dummy args, to work like other set_..._energy
```
vidartf: Use `*args, **kwargs` instead of dummy arg `ref`. Then the docstring entry for `ref` can also be removed.

hyperspy/models/edsmodel.py
```diff
@@ -0,0 +1,822 @@
...
+            component = self[xray_line]
+            if i == 0:
+                component_ref = component
+                component_ref.centre.free = True
+                E_ref = component_ref.centre.value
+                ref.append(E_ref)
+            else:
+                component.centre.free = True
+                E = component.centre.value
+                fact = float(ax.value2index(E)) / ax.value2index(E_ref)
```
vidartf: Why value2index? Should be commented. Also, what happens if there is both an offset and scaling issue? Which order should they go in? Needs to be put in (missing) documentation, and should be tested for.

vidartf: Nevermind, it makes sense why, but I'm still curios about a combination of offset and scaling error (and think it should be documented).

hyperspy/models/edsmodel.py
```diff
@@ -0,0 +1,822 @@
...
+                    component_sub.A.twin_function = _get_weight(
+                        element, li, weight_line)
+                    component_sub.A.twin_inverse_function = _get_iweight(
+                        element, li, weight_line)
+                    component_sub.A.twin = component.A
+        for component in self.xray_lines:
+            if xray_lines == 'all':
+                fix_twin()
+            elif component.name in xray_lines:
+                fix_twin()
```
vidartf: Simpler:
```python
if xray_lines == 'all' or component.name in xray_lines:
    fix_twin()
```

hyperspy/models/edsmodel.py
```diff
@@ -0,0 +1,822 @@
...
+            self.xray_lines.append(component)
+            init = True
+            if init:
+                self[xray_line].A.map[
+                    'values'] = self.spectrum[..., line_energy].data * \
+                    line_FWHM / self.spectrum.axes_manager[-1].scale
+                self[xray_line].A.map['is_set'] = (
+                    np.ones(self.spectrum[..., line_energy].data.shape) == 1)
+
+            component.A.ext_force_positive = True
```
vidartf: Cannot the following region simply be replaced by a call to `self.fix_sub_xray_lines_weight(xray_lines)` at the end (outside the loop)? Avoids code duplication.

hyperspy/models/edsmodel.py
```diff
@@ -0,0 +1,822 @@
...
+        """
+        Calibrate the resolution, the scale or the offset of the energy axis
+        by fitting.
+
+        Parameters
+        ----------
+        calibrate: 'resolution' or 'scale' or 'offset'
+            If 'resolution', calibrate the width of all Gaussian. The width is
+            given by a model of the detector resolution, obtained by
+            extrapolation the `energy_resolution_MnKa` in `metadata`
```
vidartf: Is this really accurate? Either way the docstring should specify that this will update the value of `energy_resolution_MnKa`.

hyperspy/models/edsmodel.py
dnjohnstone: Ok this one's a little bit tricky without @pburdet --- not at all clear where this is all from.... My best guess at what Goldstein means is a reference to the book "Scanning Electron Microscopy & X-ray Microanalysis" - but that's quite a big book and it doesn't have a section 2.5.... sooo I'll just have to bash my head against this one.

If anyone knows an EDS expert who might recognise the formula that would be great...

hyperspy/models/edssemmodel.py
dnjohnstone: Done

hyperspy/models/edsmodel.py
dnjohnstone: Done

hyperspy/models/edsmodel.py
dnjohnstone: Done -- note making this change in isolation decreases the accuracy of the quantification probably because 2.355 is used somewhere else and there was a conversion cancellation before... should check that the test will pass with atol 10-6 as it did at the beginning when all is done.

hyperspy/models/edsmodel.py
dnjohnstone: Done

hyperspy/models/edsmodel.py
dnjohnstone: Done

hyperspy/models/edsmodel.py
dnjohnstone: Done

hyperspy/misc/eds/utils.py
dnjohnstone: Thanks for that! That's a neat one I wouldn't have figured out myself :)

hyperspy/models/edsmodel.py
dnjohnstone: done

hyperspy/models/edsmodel.py
dnjohnstone: done

hyperspy/models/edsmodel.py
dnjohnstone: done

hyperspy/models/edsmodel.py
```diff
@@ -0,0 +1,822 @@
...
+        kind : {'single', 'multi'}
+            If 'single' fit only the current location. If 'multi'
+            use multifit.
+        **kwargs : extra key word arguments
+            All extra key word arguments are passed to fit or
+            multifit, depending on the value of kind.
+        """
+
+        if calibrate == 'energy':
+            bound = bound / 1000. * self.units_factor
```
vidartf: Parentheses for explicit operator order?

hyperspy/models/edsmodel.py
dnjohnstone: done

hyperspy/models/edsmodel.py
```diff
@@ -0,0 +1,822 @@
...
+                            **kwargs):
+        """
+        Return the fitted intensity of the X-ray lines.
+
+        Return the area under the gaussian corresping to the X-ray lines
+
+        Parameters
+        ----------
+        xray_lines: list of str or None or 'from_metadata'
+            If None, all main X-ray lines (alpha)
```
vidartf: Maybe add a way to return element maps where all the lines for one element are integrated together? Whether this should replace `None` behavior, or be a new one, or a separate function, I don't know.

hyperspy/models/edstemmodel.py
dnjohnstone: Ok sold - done.

vidartf:
I've completed the more in-depth review, at least for now. Lots of small comments, but many of them should be quick to fix. I appreciate that this is not you code originally @dnjohnstone, but hopefully you get what I meant by most of them :)

hyperspy/_signals/eds_tem.py
dnjohnstone: done

hyperspy/misc/eds/utils.py
vidartf:
```python
if elements is None:
raise Warning (no elements specified - please specify some)
```
I would rather just not have a default value for `elements`, i.e. make it non-optional. As this is a utility function where you never gain access to the intermediate model, calling it without `elements` just doesn't make sense...

hyperspy/misc/eds/utils.py
vidartf:
```python
weight_percents = [100/len(elements)] * len(elements)
```
This should work?

hyperspy/models/edsmodel.py
dnjohnstone: done

hyperspy/models/edsmodel.py
dnjohnstone: I agree that this is somewhat optimistic wording of what's really going on -- I'll update it to something that I find a little more clear...
Done

hyperspy/models/edsmodel.py
dnjohnstone: Can we leave that for another time? Perhaps it's something that would fit with @siriagus work moving forward?

hyperspy/models/edsmodel.py
vidartf: I see your point. It is basically only used in `fit_background()` (`set_signal_range` + reset at end), so keep them writable.

hyperspy/misc/eds/utils.py
dnjohnstone: yep that's a better idea.

hyperspy/models/edsmodel.py
vidartf: There are tons of magic numbers in this code. I'd replace all of them if possible. If something is calculated a lot of times, then put it as a global constant (c.f. `gaussian.py`'s sigma2fwhm constant).

hyperspy/misc/eds/utils.py
dnjohnstone: ah yes same end goal but that's much cleaner.

hyperspy/models/edssemmodel.py
```diff
@@ -0,0 +1,72 @@
...
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with  HyperSpy.  If not, see <http://www.gnu.org/licenses/>.
+
+# To do: weight_fraction different for different pixe. (so basckground)
+# Calibrate on standard and transfer dictionnary
+# k-ratios
```
vidartf: Either make these a separate issue after merge, or fix before.

hyperspy/models/edsmodel.py
dnjohnstone: probably you're right - not 100% sure on this though, it's a little cryptic to me, which is probably enough to say this is needed - I'll come back to this I think....

vidartf: I think the best way to figure this out is to create a test that adds both a scaling and an offset, and see if order is important or not, then keep the test(s) for the order that works, and document accordingly.

hyperspy/models/edssemmodel.py
dnjohnstone: we'll go for separate issues -- there are a number of other EDS related branches such as absorption corrections that are more pressing I think because there is other potential functionality that depends on that.

hyperspy/models/edsmodel.py
dnjohnstone: I would guess that you're correct because you're better at this stuff than me - but I can't see how to do it immediately... I'll come back to it if we want this done before merging.

hyperspy/models/edsmodel.py
vidartf: Sure, I just thought it would be quick to add:

First, separate out the loop body in `get_lines_intensity` (after `for xray_line in xray_lines:`) as a new function (e.g. `_get_single_line_intensity(self, xray_line)`), then add the following method:
```python
def get_element_intensities(self, elements=None, **kwargs):
    if elements is None:
        elements = self.spectrum.metadata.Sample.elements
    xray_lines = self.spectrum._get_lines_from_elements(elements, only_lines=None)
    intensities = {}
    for xray_line in xray_lines:
        element, line = utils_eds._get_element_and_line(xray_line)
        i = self._get_single_line_intensity(xray_line)
        if element in intensities:
            intensities[element] += i
        else:
            intensities[element] = i
    return intensities
```

If you don't want it in, then I can split it off as a new PR.

hyperspy/models/edsmodel.py
vidartf: I'll do a PR to you.

dnjohnstone: give me 2 seconds and i'll push the newest version which corrects nearly all of this to save merge issues!

vidartf: Actually, the advantages gained are minimal, as a lot of the code needs to be in both places no matter what. Consider this ok as it is :)

hyperspy/models/edsmodel.py
vidartf: I'm guessing 2.5 is a factor that is found/recommended in Goldstein. If so, the magic number is:
```python
2.5 / (1000 * 8 * math.log(2))
```
or in a more self-documenting way:
```python
energy2sigma_factor * eV2keV / (sigma2fwhm**2)
```

hyperspy/models/edsmodel.py
dnjohnstone: Lets split this off as a new PR either from yourself or we can leave it as low hanging fruit for people who I'm hoping are going to start contributing to EDS tools in HyperSpy soon...

hyperspy/models/edsmodel.py
dnjohnstone: Ok - I'll rewrite this along the lines of this more self-documenting style -- I've had a look around though and I can't figure out exactly where this 2.5 comes from though... I think I'll send @pburdet an e-mail and see if I can get it out of him as I'm pretty confident it will come round to have some sensible reasoning behind it knowing him.

hyperspy/models/edsmodel.py
```diff
@@ -0,0 +1,860 @@
...
+        weight_line = elements_db[
+            element]['Atomic_properties']['Xray_lines'][line]['weight']
+    return lambda x: x / weight_line
+
+
+def _get_sigma(E, E_ref, units_factor):
+    """
+    Gets sigma value for particular energy.
+
+    energy2sigma_factor : float
```
vidartf: This is an internal factor, so no need to document it. I guess the docstring could read along the lines of:
```python
"""
Gets sigma value for particular energy, according to the equation(s) derived in Goldstein (YYYY, ch. ???).
"""
```

hyperspy/models/edsmodel.py
```diff
@@ -0,0 +1,860 @@
...
+
+    This is a factor that ???? and is set to 2.5 here with reference to
+    Goldstein ???
+
+    """
+    goldstein_factor = 2.5
+    eV2keV = 1000.
+    sigma2fwhm = 2 * math.sqrt(2 * math.log(2))
```
vidartf: Put these factors in root of the file (i.e. outside the function), at least `eV2keV` and `sigma2fwhm`, and reuse them throughout the file instead of magic numbers.

hyperspy/models/edsmodel.py
dnjohnstone: Scratch that I found it... He documented it in the docstring for get_FWHM_at_Energy in utils... I'll update the docstring accordingly :)

hyperspy/models/edsmodel.py
dnjohnstone: already done it just not pushed ;)

hyperspy/models/edsmodel.py
dnjohnstone: Got a proper documentation coming for this now it's an empirical value for detector broadening :)

dnjohnstone:
Ok @vidartf many thanks for reviewing this work so quickly and thoroughly!

I think I've sorted all of the important stuff now including adding some explicit documentation for places where "magic numbers" pop up -- these turn out to be values lifted directly from known parameterizations, but since that caused so much concern/confusion I've explicitly said so in docstrings where I noticed they were present in methods.

I'll open an issue for facets of EDS quantification that could be added to HyperSpy by keen parties - there is a lot of low hanging fruit here for someone wanting to start out contributing!

So should be good to go if I'm not wrong :)

dnjohnstone:
Note if/when this is merged we can also close #473 and #390

hyperspy/misc/eds/utils.py
```diff
@@ -89,22 +115,30 @@ def get_FWHM_at_Energy(energy_resolution_MnKa, E):
...
 
 
 def xray_range(xray_line, beam_energy, density='auto'):
-    """Return the Anderson-Hasler X-ray range.
+    """Return the maximum range of X-ray generation according to the
+    Anderson-Hasler parameterization.
 
-    Return the maximum range of X-ray generation in a pure bulk material.
+    Note that the magic numbers appearing in the equation are the values from
+    the parameterization mentioned above.
```
vidartf: In general, docstrings are how to use the function. I feel this belongs as a comment inside the function body, e.g. next to the first magic number. Same for `electron_range`.

hyperspy/misc/eds/utils.py
```diff
@@ -243,6 +279,74 @@ def take_off_angle(tilt_stage,
...
 
+def xray_lines_model(elements,
+                     beam_energy=200,
+                     weight_percents=None,
+                     energy_resolution_MnKa=130,
+                     energy_axis={'name': 'E', 'scale': 0.01, 'units': 'keV',
+                                  'offset': -0.1, 'size': 1024}
+                     ):
+    """
+    Generate a model of X-ray lines using a Gaussian epr x-ray lines.
```
vidartf: "epr"?

vidartf:
Looks good to me, I'll merge tomorrow unless anybody else has some feedback. @to266 : Are the use of components/parameters correct? I feel you have a better overview there than me.

@dnjohnstone : In the mean time, please consider going over the [coverage report for the new file](https://coveralls.io/builds/4467126/source?filename=hyperspy%2Fmodels%2Fedsmodel.py) to see if there are any further tests you feel should be included. You might consider testing for expected exceptions, i.e. use `nose.tools.raises` while feeding it input that should always fail.

dnjohnstone:
Good stuff - thanks @vidartf. I've made corrections for the two final docstring points that you made and I'll have a look at the coverage report if I get a moment.

hyperspy/models/edsmodel.py
```diff
@@ -0,0 +1,879 @@
...
+        Fix the X-ray line energy (shift or centre of the Gaussian)
+
+        Parameters
+        ----------
+        xray_lines: list of str or 'all'
+            The Xray lines. If 'all', fit all lines
+        bound: float
+            the bound around the actual energy, in keV or eV
+        """
+        if xray_lines == 'all_alpha':
```
vidartf: `'all_alpha'` option is not specified in docstring. Same in several functions

vidartf:
PS: I almost forgot! There should of course be a write up in the docs! I think we should simply follow the patter for [EELSModel](http://hyperspy.org/hyperspy-doc/current/user_guide/eels.html#eels-curve-fitting).

dnjohnstone:
That was strange - not sure where I lost the docs (Pierre had done those) anyway added them back now so that's all done :)

dnjohnstone:
@vidartf are we good to go? or are you waiting for something from me. would be great to get this tied off before the end of the week :)

vidartf:
@dnjohnstone Sorry, I've been having some technical difficulties. Will merge it as soon as you fix the final nitpick on the last commit :)

dnjohnstone:
Ok @vidartf done then.

to266:
I'll have a quick look at the components / parameters in the next hour

hyperspy/_signals/eds_tem.py
to266: You should actually add the `dictionary` argument and pass it to the required model as well, not just add a docstring :)

hyperspy/misc/eds/utils.py
```diff
@@ -243,6 +275,74 @@ def take_off_angle(tilt_stage,
                                   + math.sin(a) * math.sin(c)))
 
 
+def xray_lines_model(elements,
+                     beam_energy=200,
+                     weight_percents=None,
+                     energy_resolution_MnKa=130,
+                     energy_axis={'name': 'E', 'scale': 0.01, 'units': 'keV',
+                                  'offset': -0.1, 'size': 1024}
```
to266: Should not use mutable types as default values.
Instead leave default value as `None`, and if a better thing is required, just create the dict in the first couple of lines of the actual code:
```python
if energy_axis is None:
    energy_axis = {'name': 'E', 'scale': 0.01, 'units': 'keV',
                   'offset': -0.1, 'size': 1024}
```

hyperspy/misc/eds/utils.py
```diff
@@ -243,6 +275,74 @@ def take_off_angle(tilt_stage,
...
+    from hyperspy.model import components
+    s = EDSTEMSpectrum(np.zeros(energy_axis['size']), axes=[energy_axis])
+    s.set_microscope_parameters(
+        beam_energy=beam_energy,
+        energy_resolution_MnKa=energy_resolution_MnKa)
+    s.add_elements(elements)
+    counts_rate = 1.
+    live_time = 1.
+    if weight_percents is None:
+        weight_percents = [100 / len(elements)] * len(elements)
```
to266: Here I would use a float 100 instead of int.

hyperspy/misc/eds/utils.py
```diff
@@ -243,6 +275,74 @@ def take_off_angle(tilt_stage,
...
+    """
+    from hyperspy._signals.eds_tem import EDSTEMSpectrum
+    from hyperspy.model import components
+    s = EDSTEMSpectrum(np.zeros(energy_axis['size']), axes=[energy_axis])
+    s.set_microscope_parameters(
+        beam_energy=beam_energy,
+        energy_resolution_MnKa=energy_resolution_MnKa)
+    s.add_elements(elements)
+    counts_rate = 1.
+    live_time = 1.
```
to266: What's the purpose of `counts_rate` and `live_time`? They are always 1, and only get multiplied in one place.. If it was any other number then it's file (i.e. no magic numbers), but with 1.....

doc/user_guide/eds.rst
```diff
@@ -545,3 +545,125 @@ either with :py:func:`~.misc.material.weight_to_atomic`. The reverse method is :
...
+
+    >>> s = load('bam.msa')
+    >>> s.add_elements(['Al', 'Ar', 'C', 'Cu', 'Mn', 'Zr'])
+    >>> s.set_microscope_parameters(beam_energy=10)
+
+The model is created with :py:func:`~._signals.eds_sem.create_model`. One gaussian is automatically created per X-ray line and a polynomial for the backgronud.
+
+.. code-block:: python
+
+    >>> m = create_model(s)
```
to266: Old documentation - should read `s.create_model()`. Same for `hs.load('bam.msa')` higher up

hyperspy/models/edsmodel.py
```diff
@@ -0,0 +1,881 @@
...
+
+    auto_background : boolean
+        If True, adds automatically a polynomial order 6 to the model,
+        using the edsmodel.add_polynomial_background method.
+
+    Any extra arguments are passed to the Model creator.
+
+    Example
+    -------
+    >>> m = create_model(s)
```
to266: old docs again

hyperspy/models/edsmodel.py
```diff
@@ -0,0 +1,881 @@
...
+            component.fwhm = line_FWHM
+            component.centre.free = False
+            component.sigma.free = False
+            component.name = xray_line
+            self.append(component)
+            self.xray_lines.append(component)
+            init = True
+            if init:
+                self[xray_line].A.map[
+                    'values'] = self.spectrum[..., line_energy].data * \
```
to266: Should replace the old slicing / indexing convention with the new one
`self.spectrum[..., line_energy].data` -> `self.spectrum.isig[line_energy].data`.

Not only here, but other places as well

hyperspy/models/edsmodel.py
```diff
@@ -0,0 +1,881 @@
...
+            component = create_component.Gaussian()
+            component.centre.value = line_energy
+            component.fwhm = line_FWHM
+            component.centre.free = False
+            component.sigma.free = False
+            component.name = xray_line
+            self.append(component)
+            self.xray_lines.append(component)
+            init = True
+            if init:
```
to266: Is this `init` required? I can't find it changed to `False` or checked at any other point. If it's true, I would just remove it and move the `if` clause to a normal identation

hyperspy/models/edsmodel.py
```diff
@@ -0,0 +1,881 @@
...
+        """Disable the X-ray lines components.
+
+        """
+        for component in self._active_xray_lines:
+            component.active = False
+
+    def fit_background(self,
+                       start_energy=None,
+                       end_energy=None,
+                       windows_sigma=[4., 3.],
```
to266: Again a mutable type in default. Simplest would be to change it to a tuple (with doc change as well)

hyperspy/misc/eds/utils.py
vidartf: I think there is a field in metadata for live time, possibly also counts rate? If so, we could check those.

hyperspy/models/edsmodel.py
```diff
@@ -0,0 +1,881 @@
...
+            else:
+                component.sigma.free = True
+                E = component.centre.value
+                component.sigma.twin_function = _get_sigma(
+                    E, E_ref, self.units_factor)
+                component.sigma.twin_inverse_function = _get_sigma(
+                    E_ref, E, self.units_factor)
+                component.sigma.twin = component_ref.sigma
+
+    def _set_energy_resolution(self, xray_lines, *args, **kwargs):
```
to266: why do we need `*args, **kwargs`?

hyperspy/models/edsmodel.py
```diff
@@ -0,0 +1,881 @@
...
+            xray_lines = [compo.name for compo in self.xray_lines]
+        energy_Mn_Ka, FWHM_MnKa_old = self.spectrum._get_line_energy('Mn_Ka',
+                                                                     'auto')
+        FWHM_MnKa_old *= eV2keV / self.units_factor
+        get_sigma_Mn_Ka = _get_sigma(
+            energy_Mn_Ka, self[xray_lines[0]].centre.value, self.units_factor)
+        FWHM_MnKa = get_sigma_Mn_Ka(self[xray_lines[0]].sigma.value
+                                    ) * eV2keV / self.units_factor * sigma2fwhm
+        if FWHM_MnKa < 110:
+            warnings.warn("FWHM_MnKa of " + str(FWHM_MnKa) + " smaller than" +
```
to266: I would use `ValueError` instead of a warning. The execution should stop here any way

hyperspy/models/edsmodel.py
```diff
@@ -0,0 +1,881 @@
...
+        Parameters
+        ----------
+        start_energy : {float, None}
+            If float, limit the range of energies from the left to the
+            given value.
+        end_energy : {float, None}
+            If float, limit the range of energies from the right to the
+            given value.
+        windows_sigma: list of two float
+            The upper and lower bounds around each X-ray lines to define
```
vidartf: Strictly, I guess this should be "lower and upper", as that is how it is used, and since that is the order that makes the most sense.

hyperspy/models/edsmodel.py
```diff
@@ -0,0 +1,881 @@
...
+        """
+        Adjust the width of all lines and set the fitted energy resolution
+        to the spectrum
+
+        Parameters
+        ----------
+        xray_lines: list of str or 'all_alpha'
+            The Xray lines. If 'all_alpha', fit all using all alpha lines
+        ref: float
+            the centre of the first line before the fit
```
to266: Documentation is wrong - in the code `ref` is not just one number, but as long as the `xray_lines` are (and of 'all_alpha', then as many components the model has)

hyperspy/models/edsmodel.py
```diff
@@ -0,0 +1,881 @@
...
```
to266: Again same thing as with scale changing

hyperspy/models/edsmodel.py
```diff
@@ -0,0 +1,881 @@
...
+
+        Parameters
+        ----------
+        xray_lines: list of str or 'all'
+            The Xray lines. If 'all', fit all lines
+        bounds: float
+            Bound the height of the peak to a fraction of
+            its height
+        """
+        def free_twin():
```
to266: It's not strictly necessary, but I would add an argument `component` to this function. Otherwise the usage makes me cringe

hyperspy/models/edsmodel.py
```diff
@@ -0,0 +1,881 @@
...
+        Establish the twin on the height of sub-Xray lines (non alpha)
+        """
+        def fix_twin():
+            component.A.bmin = 0.0
+            component.A.bmax = None
+            element, line = utils_eds._get_element_and_line(component.name)
+            for li in elements_db[element]['Atomic_properties']['Xray_lines']:
+                if line[0] in li and line != li:
+                    xray_sub = element + '_' + li
+                    component_sub = self[xray_sub]
```
to266: Should use `try` in case there is no such component

hyperspy/models/edsmodel.py
```diff
@@ -0,0 +1,881 @@
...
+        ----------
+        xray_lines: list of str or 'all'
+            The Xray lines. If 'all', fit all lines
+        bound: float
+            the bound around the actual energy, in keV or eV
+        """
+
+        for component in self:
+            if component.isbackground is False:
+                if xray_lines == 'all':
```
to266: combine the two `if`s:

`if xray_lines == 'all' or component.name in xray_lines:`

hyperspy/models/edsmodel.py
```diff
@@ -0,0 +1,881 @@
...
+            The Xray lines. If 'all', fit all lines. If 'all_alpha' fit all
+            using all alpha lines.
+        bound: float
+            the bound around the actual energy, in keV or eV
+        """
+        if xray_lines == 'all_alpha':
+            xray_lines = [compo.name for compo in self.xray_lines]
+        for component in self:
+            if component.isbackground is False:
+                if xray_lines == 'all':
```
to266: again combine

hyperspy/models/edsmodel.py
```diff
@@ -0,0 +1,881 @@
...
+
+        Examples
+        --------
+        >>> m.multifit()
+        >>> m.get_lines_intensity(["C_Ka", "Ta_Ma"])
+        """
+        from hyperspy import utils
+        intensities = []
+        if xray_lines is None:
+            xray_lines = []
```
to266: `xray_lines = [component.name for component in self.xray_lines]`

Will be quicker and more pythonic

hyperspy/models/edsmodel.py
```diff
@@ -0,0 +1,881 @@
...
+            for component in self.xray_lines:
+                xray_lines.append(component.name)
+        else:
+            if xray_lines == 'from_metadata':
+                xray_lines = self.spectrum.metadata.Sample.xray_lines
+            xray_lines = filter(lambda x: x in [a.name for a in
+                                self], xray_lines)
+        if xray_lines == []:
```
to266: I would replace that with `if not xray_lines:`

hyperspy/models/edsmodel.py
```diff
@@ -0,0 +1,881 @@
...
+            raise ValueError("These X-ray lines are not part of the model.")
+        for xray_line in xray_lines:
+            element, line = utils_eds._get_element_and_line(xray_line)
+            line_energy = self.spectrum._get_line_energy(xray_line)
+            data_res = self[xray_line].A.map['values']
+            if self.axes_manager.navigation_dimension == 0:
+                data_res = data_res[0]
+            img = self.spectrum.isig[0:1].integrate1D(-1)
```
to266: ...While that is true, I feel we unnecessary integrate and the purpose of the code is not obvious from the first glance. Instead I would write 
```python
img = self._get_navigation_signal(data_res)
img.axes_manager.set_signal_dimension(0)
```

doc/user_guide/eds.rst
dnjohnstone: done

hyperspy/_signals/eds_tem.py
dnjohnstone: done

hyperspy/misc/eds/utils.py
dnjohnstone: done

hyperspy/misc/eds/utils.py
dnjohnstone: done

hyperspy/models/edsmodel.py
dnjohnstone: done

hyperspy/models/edsmodel.py
dnjohnstone: done

hyperspy/models/edsmodel.py
dnjohnstone: done

hyperspy/models/edsmodel.py
dnjohnstone: as far as I can see we don't so long as ref is added as an argument - so I've done that.

hyperspy/models/edsmodel.py
dnjohnstone: done

hyperspy/models/edsmodel.py
dnjohnstone: done

hyperspy/models/edsmodel.py
dnjohnstone: done

hyperspy/models/edsmodel.py
dnjohnstone: done

hyperspy/models/edsmodel.py
dnjohnstone: done

hyperspy/models/edsmodel.py
dnjohnstone: done

hyperspy/models/edsmodel.py
vidartf: Previously this said `dummy`, since `ref` would be unused. I suggested to use `*args, **kwargs` to handle unwanted arguments.

hyperspy/models/edsmodel.py
dnjohnstone: done

hyperspy/models/edsmodel.py
dnjohnstone: done

hyperspy/models/edsmodel.py
dnjohnstone: done

hyperspy/models/edsmodel.py
dnjohnstone: I asked @francisco-dlp about this (because I wasn't 100% sure what to put in the try) he suggested instead using a check of whether the component is in the model with an if statement so that when it breaks you really know what's gone wrong where a try function may cause errors that are more tricky to trace - so that's what I've implemented.

hyperspy/models/edsmodel.py
vidartf: This one should probably still be done.

hyperspy/models/edsmodel.py
dnjohnstone: I can't get that to work:

probably should be -->  img = self.spectrum._get_navigation_signal(data_res)

but still can't make it work

dnjohnstone:
Ok - I think I covered everything... Thanks for the lesson in Python @to266 

Let me know if more is required.

hyperspy/misc/eds/utils.py
dnjohnstone: I think this should stay - I think the main point is to initialise the model here whereas in the metadata we might want to use those values for more sophisticated treatements later.

dnjohnstone: Also - the eqn which uses them is much more self-documenting if we decided to change it later when it's done this way.

vidartf: To get metadata `live_time` safely:
```python
if "Acquisition_instrument.SEM" in s.metadata:
    mp = s.metadata.Acquisition_instrument.SEM
else:
    mp = s.metadata.Acquisition_instrument.TEM
if mp.has_item('Detector.EDS.live_time'):
    live_time = mp.Detector.EDS.live_time
```

dnjohnstone: Yes but I think the point is not to initialise the model with the actual value for the time being - shouldn't really matter too much?

vidartf: I'm not sure what `counts_rate` is supposed to be. Maybe `mp.Detector.EDS.real_time` ?

vidartf: Sorry for the mixed in comments, I wasn't refreshing properly.

Regarding which value it should have: Looking at the full context of the code, these values (`live_time` and `count_rate`) should either be removed, or taken as arguments, as they currently have no function. If they are taken as arguments, they are currently used correctly to create the signal, but they should then be saved to the metadata, so that it will operate correctly afterwards. No?

vidartf: Looking through the current code-base, I'm in favor of removing them completely, and leaving these issues to the user if for some reason they need this capability.

hyperspy/models/edsmodel.py
to266: Ok, I've figured out what does not work here - the code is old and essentially works with 0-dimensional data (i.e. scalars)... And we no longer support such signals.

So the problem is that @pburdet used to rely on such 0-dim signals to tell when he should just print scalars, and when to plot things (among other things). The proper solution would be to fix this behaviour. In order to just make it work without fixing, you can use the following
```python
img = self.spectrum._get_navigation_signal()
img.axes_manager.set_signal._dimension(0)
img.data = data_res
```

If you do end up just "patching" - please raise a new issue so that we don't forget about it

vidartf:
Looking over this again, I think any other outstanding issues can come as separate enhancement PRs, as the remaining requests simply add more functionality to this code. For now, all the code here seems to do its job properly.

PR #791, opened 2015-12-11
added first iteration of anaconda hyperspy environment

to266:
Seems to run fine. 

run like so:
```bash
conda env create -f anaconda_hyperspy_environment.yml
source activate hyperspy
```

It has the drawback that every time before using hyperspy the user has to activate the environment, but at least it stays consistent.... And the user can use it as root, env, if s(he) chooses to


to266:
...Wrong target to merge into, closing

PR #792, opened 2015-12-11
added first iteration of anaconda hyperspy environment (again)

to266:
Seems to run fine. 

run like so:
```bash
conda env create -f anaconda_hyperspy_environment.yml
source activate hyperspy
```

It has the drawback that every time before using hyperspy the user has to activate the environment, but at least it stays consistent.... And the user can use it as root, env, if s(he) chooses to


jrminter:
Thanks. Worked nicely for me.

francisco-dlp:
Couldn't we limit the dependencies to the actual dependencies? i.e. no mayavi, pandas etc...

to266:
Done, I think

francisco-dlp:
Just a little bit of documentation in the User Guide is missing.

to266:
Added

francisco-dlp:
The download link should not point to the master branch as the dependencies might be different from those of the latests pypi release.

to266:
Done, points to 0.8.x

PR #797, opened 2015-12-15
X-ray Absorption Coefficient Database

dnjohnstone:
This deprecates #471 from @pburdet.

Based on this: www.nist.gov/srd/upload/jpcrd488.pdf
http://www.nist.gov/pml/data/ffast/

Code wise this is merged up and so should be ready for review - EDS wise it would be great if there were a proper EDS expert out there to help...


dnjohnstone:
I guess this may need an additional test or two but other than that are there any issues with this?

dnjohnstone:
Ah, I see what coveralls is complaining about now - in fact the structure in test_material.py was wrong such that the tests weren't really doing anything. Having corrected that there is an issue somewhere down the line on the database lookup for absorption in mixtures.... It looks to be a bit buried to me so bear with my while I figure out what's going on there...

dnjohnstone:
Ok - I think this is ready for review.

hyperspy/misc/material.py
```diff
@@ -253,6 +256,209 @@ def density_of_mixture_of_pure_elements(weight_percent,
...
+            mac_res += np.array([weight * ma for ma in mac_re])
+        return mac_res
+    else:
+        mac_res = np.array([mass_absorption_coefficient(
+            el, energies) for el in elements])
+        mac_res = np.dot(weight_percent, mac_res) / np.sum(weight_percent, 0)
+        return mac_res
+
+
+def mass_absorption_coefficient_of_mixture_of_pure_elements(weight_percent,
```
vidartf: I'm all for self-documenting code, but with a style guide saying a maximum line length of 80 characters, this function name is way too long. How about `mass_absorption_mixture`?

dnjohnstone: Fair point - will change that.

dnjohnstone:
@vidartf function names involving mixtures are now shortened and merged up with current master

dnjohnstone:
I'm a little confused as to what the issue is here - at first I thought it was that something in the diff was not python 3 compatible, but it seems to me that the errors that are causing nosetest to fail are in parts of the code that I haven't touched? @vidartf @to266 anyone know what's going on here?

vidartf:
@dnjohnstone: You are correct. Seems the new FEI code raises a warning that isn't dealt with in the tests. This only became an issue when 0.8.x was merged into master, since only master fails on warnings. Should be a quick fix (ignoring/expecting a warning in test).

dnjohnstone:
@vidartf this is now up to date with the current master and incorporates previously discussed changes. Is this now ready to merge?

doc/user_guide/eds.rst
```diff
@@ -360,6 +360,24 @@ database:
 The lines are returned in order of closeness to the specified energy, and can
 be limited by additional, optional arguments.
 
+Mass absorption coefficient database
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+A mass absorption coefficient database (Chantler2005 see http://physics.nist.gov/ffast) is available
```
k8macarthur: Maybe add the reference to the bibliography for consistency?

tjof2: Worth noting I've rewritten the EDS documentation in #965, so yes, the bibliography is the right way to go.

dnjohnstone:
@siriagus have you tried using this at all?

siriagus:
@dnjohnstone Hey, I have not. I will today if I get time or tomorrow. Thanks for updating it with the current master.

dnjohnstone:
@siriagus thanks for that -- just updated again so all as up-to-date as possible

k8macarthur:
Question: What units are the absorption coefficients going to be in?

hyperspy/misc/material.py
```diff
@@ -236,21 +237,220 @@ def density_of_mixture_of_pure_elements(weight_percent,
...
+    ----------
+    element: str
+        The element symbol of the absorber, e.g. 'Al'.
+    energies: float or list of float or str or list of str
+        The energy or energies of the X-ray in keV, or the name of the X-rays,
+        e.g. 'Al_Ka'.
+
+    Return
+    ------
+    mass absorption coefficient(s) in cm^2/g
```
dnjohnstone: @k8macarthur these are the units... Not sure if that's the best...

k8macarthur: Yeah I saw this but I thought mu was in units of barns/atom and rho was in units of g/cm^3 so mu/rho should be cm/g surely. Not necessarily needed to be understood for this pull to be merged maybe more of an issue for #962 .

dnjohnstone: Mmmm I think the answer is that mu isn't in barns/atom (that would be sigma) at least according to the convention followed by NIST, which is where this comes from. This is also the convention followed by Goldstein et al in "Scanning Electron Microscopy & Microanalysis" so it's the one we should follow. In any case, we don't have much choice as these are the units of the NIST database.

http://physics.nist.gov/cgi-bin/ffast/ffast.pl?Z=31&Formula=&gtype=3&lower=&upper=&density=

k8macarthur: Ok thanks I'll take another look and see if I can get to the bottom of it. I'll post any more comments on the correct issue.

hyperspy/misc/material.py
```diff
@@ -236,21 +237,220 @@ def density_of_mixture_of_pure_elements(weight_percent,
     Examples
     --------
     Calculate the density of modern bronze given its weight percent:
-    >>> hs.material.density_of_mixture_of_pure_elements(
-            (88, 12),("Cu", "Sn"))
+    >>> hs.material.density_of_mixture((88, 12),("Cu", "Sn"))
```
siriagus: Docstring: Use list instead of tuples and use single quotation marks to be consistent with other documentation. i.e. [88,12], ['Cu', 'Sn'].

hyperspy/misc/material.py
```diff
@@ -179,8 +181,7 @@ def _density_of_mixture_of_pure_elements(weight_percent,
...
```
siriagus: Docstring: Use list instead of tuples and use single quotation marks to be consistent with other documentation. i.e. [88,12], ['Cu', 'Sn'].
(Same as below)

hyperspy/misc/material.py
```diff
@@ -236,21 +237,220 @@ def density_of_mixture_of_pure_elements(weight_percent,
...
+    ----
+    See http://physics.nist.gov/ffast
+    Chantler, C.T., Olsen, K., Dragoset, R.A., Kishore, A.R., Kotochigova,
+    S.A., and Zucker, D.S. (2005), X-Ray Form Factor, Attenuation and
+    Scattering Tables (version 2.1).
+    """
+
+    energies_db = np.array(ffast_mac[element].energies_keV)
+    macs = np.array(ffast_mac[element].mass_absorption_coefficient_cm2g)
+    energies = copy.copy(energies)
```
siriagus: Does this line serve any purpose? It doesn't make any sense to me...

hyperspy/misc/material.py
```diff
@@ -236,21 +237,220 @@ def density_of_mixture_of_pure_elements(weight_percent,
...
+    if isinstance(energies, str):
+        energies = utils_eds._get_energy_xray_line(energies)
+    elif hasattr(energies, '__iter__'):
+        for i, energy in enumerate(energies):
+            if isinstance(energy, str):
+                energies[i] = utils_eds._get_energy_xray_line(energy)
+    index = np.searchsorted(energies_db, energies)
+    mac_res = np.exp(np.log(macs[index - 1]) +
```
siriagus: Quick note: I tested this interpolation,it gives the correct result :)

siriagus:
@dnjohnstone I have used this for a while now. I went through line for line now and could not spot any errors. I also tested the functions + parts of functions manually to make sure that it was implemented correct. I say this is ready to merge. Sorry for delaying for so long, guess I had to see you in person to finally get through it.

dnjohnstone:
This is now updated with respect to the current master and I have made the minor style adjustments suggested by @siriagus

siriagus:
@dnjohnstone Sorry for bringing this up right after it merged, but I am reading up more on theory about this. According to the NIST webpage, the FFAST contains mass attenuation coefficients and NOT mass absorption coefficients. Due to being new to the field, I assumed that mass attenuation coefficients and mass absorption coefficients must be the same. Especially since @pburdet initialised this I assumed that the imported data was correct. All literature regarding absorption correction I have encountered refers to u/p as mass absorption coefficients. I believe the coefficents are similar, however, mass absorption coefficients are lower as attenuation coefficients also includes scattering effects. Since HyperSpy 1.0.0 is planned to release next month, I think this should be correct. So, if the database actually contain mass attenuation coefficients, the code should reflect this. I understand @dnjohnstone is not an EDS person either, but took initative to finish what @pburdet started (which is awesome!). So, unless someone can confirm that these are in fact mass absorption coefficients, I think every occurance of _absorption_ should be replace by _attenuation_. I know it is really annoying that this comes up 16 hours after it just merged, but I was just reading up on the theory. It might also be that the terms are used interchangeably in the literature, however, if we are not 100% sure ourselves, we should change it to attenuation all over. I don't mind doing this, but wanted to ask if this is something you @dnjohnstone have considered or if you just implemted what @pburdet had made assuming like me that the values were the same. Again, really sorry that I didn't catch this in the initial review as I assumed the database itself had the correct values.

francisco-dlp:
@siriagus, thank you for bringing this up. In prevention, I'll revert the merge until you reach consensus on this.

dnjohnstone:
I think the answer to this is that the database contains two things, which are illustrated below for Ga:

"photoabsorption coefficients" - http://physics.nist.gov/cgi-bin/ffast/ffast.pl?Z=31&Formula=&gtype=2&lower=&upper=&density=

"attenuation coefficients" - http://physics.nist.gov/cgi-bin/ffast/ffast.pl?Z=31&Formula=&gtype=3&lower=&upper=&density=

You will notice that both are mu/rho and in-fact both are very similar. The highest energy X-rays, in this case 4.329451E+02  eV show the difference most clearly (because the coefficients are smallest) where we have in the above case 8.9633E-02 for attenuation and 4.0914E-03 for photoabsorption.

As @siriagus says you have a smaller value for absorption only.

If you compare a few values in ffast_mac with the two database options - you will see that @pburdet has taken the "photoabsorption coefficients" not the "attenuation coefficients".

The upshot is that I think everything is consistent with @siriagus's potential concerns and indeed this draws on absorption coefficients not attenuation coefficients.

siriagus:
@dnjohnstone Thank you for the clarification. I raised this concern as it could be a potential issue. At least this is a really good example of why #1086 is important!

@francisco-dlp I don't know what the easiest fix now is, reverting the revert (#1083) is probably the safest option though.

dnjohnstone:
@siriagus it's only a good thing you said something!! I hadn't checked that until this point so please carry on saying - I think we're ok in this case, but that might not have been the case.

francisco-dlp:
Remerged in #1090 :)

PR #806, opened 2015-12-20
FIX: Small EDSModel test fixups and cleanups

vidartf:
Fixing some small issues that remained in #786 + did some cleanup + added a test.


hyperspy/models/edsmodel.py
```diff
@@ -554,7 +556,8 @@ def calibrate_energy_axis(self,
         calibrate: 'resolution' or 'scale' or 'offset'
             If 'resolution', fits the width of Gaussians place at all x-ray
             lines. The width is given by a model of the detector resolution,
-            obtained by extrapolation the `energy_resolution_MnKa` in `metadata`
+            obtained by extrapolation the `energy_resolution_MnKa` in
```
dnjohnstone: should probably be "extrapolating" rather than "extrapolation"

dnjohnstone:
Thanks for that @vidartf - it looks good to me aside from the very small change in English I'd suggest.

vidartf:
Typo fixed as suggested.

Detailed breakdown of changes:
 * Makes sure `EDSTEMModel` passes `*args, **kwargs` to ancestor in init.
 * Makes sure `EDSSEMSpectrum.create_model()` passes `*args, **kwargs` to models.
 * Adds a test routine to test `EDSModel.get_lines_intensity()`.
 * Minor tidying up.

hyperspy/models/edstemmodel.py
```diff
@@ -42,4 +42,5 @@ def __init__(self, spectrum,
                  auto_background=True,
                  auto_add_lines=True,
                  *args, **kwargs):
-        EDSModel.__init__(self, spectrum, auto_add_lines, *args, **kwargs)
+        EDSModel.__init__(self, spectrum, auto_background, auto_add_lines,
```
francisco-dlp: Here you're actually fixing a bug that didn't cause any failure in the tests. It would be nice to add a little test for the `EDSModel` constructor.

vidartf:
I've now added tests, and uncovered some more issues. Some of them I fixed, but I need help (from @to266 ?) in order to fix the restoration a stashed model:

Two attributes of `EDSModel` needs to be stored and restored, which are both lists of references to components. I'm not sure how to whitelist these for export/restore, as `'id'` stores the ID of the list itself, not the components in it. Can this be done?

to266:
At present, no. You'll have to store the `id` of each of the components "by hand" (like [here](https://github.com/hyperspy/hyperspy/blob/master/hyperspy/component.py#L546)).

vidartf:
Thanks. I already made a manual workaround using the `name` attribute in 9611c1d, would you suggest to switch to `id`?

to266:
No need to, using names is even better, as you don't have to keep track of the recoded `id` of each component for reference.

vidartf:
If so, this PR should be ready for review.

hyperspy/tests/model/test_edsmodel.py
```diff
@@ -96,5 +143,56 @@ def test_calibrate_xray_width(self):
...
+        m.fit()
+        w1 = np.array([0.3, 0.7])
+        w2 = np.array([0.5, 0.5])
+        w = np.zeros((4,) + self.mix.shape)
+        for x in xrange(self.mix.shape[0]):
+            for y in xrange(self.mix.shape[1]):
+                w[0:2, x, y] = w1 * self.mix[x, y]
+                w[2:4, x, y] = w2 * (1-self.mix[x, y])
+        xray_lines = s._get_lines_from_elements(
+            s.metadata.Sample.elements, only_lines=('Ka',))
```
to266: In my system this returns the elements in alphabetical order (i.e. it's `['As_Ka', 'Cr_Ka', 'Fe_Ka', 'Ga_Ka']`). The `w1`, `w2` and `w` definition is such that you expect the elements to be ordered `Fe, Cr, Ga, As`.

The test does not fail because you misspelled: `atol=10-4` (not `10e-4`)

vidartf:
Fixed the issues with the test, and should now be OK.

vidartf:
Master merged to solve conflicts. Consider merging while fresh :)

Issue #775, opened 2015-12-03
AppVeyor build tests?


vidartf:
Currently, commits are not tested on windows platform, which means every once in a while certain platform-specific issues like that in #774 slips through. AppVeyor is similar to Travis, but runs on Windows hosts, so it should pick up on these.

For a guide on setting up AppVeyor, see http://python-packaging-user-guide.readthedocs.org/en/latest/appveyor/ . We get to skip all the compiler bits, since we have no extensions that need building, but can still run testing.

PR #807, opened 2015-12-23
New: AppVeyor CI + bundle installer deployment

vidartf:
Implements #775. Also adds automated bundle installer creation for windows when a tag is committed, which is deployed as a draft release. Access for AppVeyor of course needs to be set up, and the secured key needs to be replaced. There is also some 3.4 vs 3.5 issues that will need to be resolved once we make the transition to python3 ( #769), but we can take that decision once one of the branches has been merged.


vidartf:
There's also several minor issues that I had to fix first that are separate commits here, which I will submit separately after Christmas :)

vidartf:
PS: I'm guessing it would be more appropriate to PR this to 0.8.x? If others agree, I'll resubmit.

francisco-dlp:
+1

2015-12-30 10:26 GMT+00:00 Vidar Tonaas Fauske <notifications@github.com>:

PR #808, opened 2015-12-30
New: AppVeyor CI + bundle installer deployment

vidartf:
Implements #775, resubmission of #807. Implements testing on Windows machines.

Also adds automated bundle installer creation for windows when a tag is committed, which is deployed as a draft release. Access for AppVeyor of course needs to be set up, and the secured key in `appveyor.yml` needs to be replaced. There is also some 3.4 vs 3.5 issues that will need to be resolved once we make the transition to python3 ( #769), but we can take that decision once one of the branches has been merged.

An alternative option to what is proposed here is to PR most of the bundle creation code back to [hyperspy/hyperspy-bundle](https://github.com/hyperspy/hyperspy-bundle), and then fetch that code in the deployment step.


to266:
I'm for this solution. A couple of things:

1. We also include `mayavi` package in the bundle (from [here](http://www.lfd.uci.edu/~gohlke/pythonlibs/) ).
2. I don't really like that there is no way to automatically look up latest winpython `.exe`... I'm aware it's not a problem that we can solve, just a general thing I'm not exactly happy about
3. Is the default included numpy compiled with `mkl`? I guess so, but it's worth checking.
4. As for the secure key (I should have enough access to the hyperspy repo), you can try putting this `ptV5Dkz3pSVdjD0qRDpxJgjVlddFtleZ+B+c2X1Fg67P8OX3bHWVktRmlj6hfLhM`

I've created the appveyor account and hyperspy project, however I suppose no additional settings (for build / test / etc) need to be changed, as everything should be specified in the `.yml`, right? 

Thanks for this, by the way, beat me to it - I was thinking of doing exactly this once I was back. :)

vidartf:
> I don't really like that there is no way to automatically look up latest winpython.exe

We can, but currently that points to the v.3.4 exe, so until we go to python 3 we have to keep it hard-coded.

> Is the default included numpy compiled with mkl? I guess so, but it's worth checking.

The WinPython default is MKL, as confirmed on my machine and [here](https://github.com/winpython/winpython/issues/194).

vidartf:
PS: By hard-coding URLs, we can also hard-code checksums, which I should still add.

vidartf:
> We also include mayavi package in the bundle (from here ).

The page you link to really don't want people to automatically download files from that page. Suggestions?

vidartf:
> I've created the appveyor account and hyperspy project,

Yes, I see that the PR stuff currently link to `https://ci.appveyor.com/project/to266/hyperspy/...`. Would @francisco-dlp be able to set it up for `https://ci.appveyor.com/project/hyperspy/hyperspy` in the same way as for Travis?

to266:
Is a link like [this](http://www.lfd.uci.edu/~gohlke/pythonlibs/zpx9vnfu/mayavi-4.4.3+vtk620-cp27-none-win_amd64.whl) insufficient? I guess the random bit may change from IP to IP though.. let me know if it does not work. In case it works fine, I got the link by starting to download (in chrome), then cancelling ant looking at the source, so to say.

I was hoping that the appveyor would understand what I want and create hyperspy-specific build, but I guess we'll have to wait for Francisco. I can keep the current behaviour for now, or disable it. Up to you

to266:
http://help.appveyor.com/discussions/questions/1154-appveyor-account-for-github-organizations

So @francisco-dlp will have to create an email / password to create an account (http://www.appveyor.com/docs/team-setup) specifically for hyperspy

vidartf:
> Is a link like this insufficient? 

The link works, but I'm not sure how stable it is, and the page does say "Please only download files manually as needed" and has all these javascript things. If the link needs to be updated for every release it kind of defeats the purpose of having automated deployment (or at least makes it much more complicated).

vidartf:
PS: For the long term, please consider upvoting this: http://vtk.uservoice.com/forums/31508-general/suggestions/11313462-distribute-python-binary-wheel-on-pypi

francisco-dlp:
I have set up an AppVeyor account for the HyperSpy. Do I need to do something else?

@vidartf, wouldn't it be possible/desirable to PR the changes to NSIS installer to the `hyperspy-bundle` repository? The `hyperspy-bundle` package can then be installed as any other package to create the NSIS installer.

vidartf:
@francisco-dlp : Yes, that would be possible, as I wrote in the initial post. If you'd prefer that, I'll split that code off this PR.

francisco-dlp:
I think that it is good to keep hyperspy and the NSIS installer separated, why did you think otherwise?

vidartf:
I've created a PR to hyperspy-bundle with the relevant code ( hyperspy/hyperspy-bundle#3 ). It will need to merged before this PR can be updated to use the code from that repo.

vidartf:
> I have set up an AppVeyor account for the HyperSpy. Do I need to do something else?

@francisco-dlp: Did you also add the hyperspy repository to the projects of the AppVeyor user? I cannot access https://ci.appveyor.com/project/hyperspy/hyperspy, which is where I would expect it to show up.

francisco-dlp:
@vidartf, I have now updated the account name, so now you should be able to access  https://ci.appveyor.com/project/hyperspy/hyperspy. I have also granted all developers admin permissions to the account.

vidartf:
> I have also granted all developers admin permissions to the account.

Thanks! Just a heads up for others, and note for the future, the AppVeyor docs says this:

> To switch between accounts user must first sign out. While signing in again user will be presented with a dropdown listing all accounts he/she is assigned:

vidartf:
As far as I can tell, this should be ready for review now. After passed review, I'll switch to the secure key of @to266, as that should be relevant longer than mine. I'll keep my key in place for review-process though, in case I need to run another deployment test.

For reviewing, this is the latest deployment results from this PR's branch: https://github.com/vidartf/hyperspy/releases/tag/v0.8.x%2Breview_test

I'm currently lacking access to a machine without a python installation where I also have admin rights, so that would be an ideal target for testing the installers/uninstallers.

to266:
I've just tested the 32bit version, and while the installer runs fine, `traits` and `traitsui` are not installed, hence the hyperspy does not start.
Not sure if any other packages are missing as well..

vidartf:
Yes, I noticed myself as well, and fixed in the latest commit, but I haven't redeployed yet. When I ran `pip install traitsui` on the fresh install everything worked as it should, except I had similar issues with `qtconsole` as what @thomasaarholt mentioned on gitter yesterday, so I got distracted by that.

vidartf:
@to266: Since we've set up a separate account for this, would you mind turning off the hyperspy project from your account? Currently the build status ones link to your page (to266/hyperspy).

to266:
Sorry for that, deleted

vidartf:
Everything should no be working. Latest test deployment can be found here: https://github.com/vidartf/hyperspy/releases/tag/v0.8.x%2Breview_test3.

I've installed and tested a little myself, and the only issue I could find was that trying to run `pip` from the winpython command prompt gives an error `failed to create process.` Running `easy_install pip` fixes it. The reason seems to be that the scripts in the "Scripts" folder hardcode the python path whenever they are installed (see http://stackoverflow.com/questions/17373265/pip-easy-install-failure-failed-to-create-process). A solution would be to run a script at the end of the installation to fix these, or add that as an issue for WinPython.

vidartf:
@to266 : Your `mayavi` link just redirects to main page now, so I don't think we can automate its inclusion unless their developers get with the program and start distributing binary wheels.

vidartf:
Potentially that VTK makes wheels.

vidartf:

to266:
@francisco-dlp, are you OK with this? If so, I suggest merging

vidartf:
I'll fix it up for master and submit another PR to master branch.

francisco-dlp:
Couldn't you simply merge 0.8.x into master and fix it on the way? (I'm assuming that there is little to be fixed.)

vidartf:
Possibly. I'll see how many differences there are first.

vidartf:
Pushed directly.

PR #809, opened 2015-12-30
FIX: Several minor issues

vidartf:
* Added keyword passthrough from `Model.save()` to `Signal.save()`.
 * Added `overwrite=True` for all `save()` calls in tests, to make sure tests do not hang if file exists.
 * Added some adjustments to test so they will work on Windows (mainly related to HDF5's `test_in_memory_loading`).
 * Make sure new test data files are included in `setup.py`.

### DIFF
--- a/hyperspy/model.py
+++ b/hyperspy/model.py
@@ -204,7 +204,7 @@ class BaseModel(list):
         s = self.signal
         s.models.store(self, name)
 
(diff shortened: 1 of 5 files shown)


vidartf:
Relevant PRs: #766, #779, #741, #735 and #718.

PR #812, opened 2016-01-04
fixes #785

### LINKED ISSUE #785: dpi error when exporting images
When running `s.export_bss_results(save_figures=True)`, (`save_figures` is the key part here), it won't let us save the results as images.

This has been a bug for over a year, according to a friend of mine who used it last year. I just ran into it, in my attempt export the bss_results to 'png's.

A nonsensical, but quick test is the following code, which only succeeds if `save_figures=False`.
```
path = "your_path_here"
s = hs.signals.Spectrum(np.arange(2*3*4).reshape((2,3,4)))
s.change_dtype('float64')
s.decomposition(True)
s.blind_source_separation(3)
s.export_bss_results(folder = path, save_figures=True)
```
The error I get on my real data is pasted below.
```
<ipython-input-16-3fc5650cd89f> in <module>()
----> 1 s.export_bss_results(folder = "/Users/thomas/Dropbox/0_DPhil/0_Data/0_TEM Data/Other Users/Aidan/SI-001/bss", save_figures=True)

/Users/thomas/anaconda/lib/python2.7/site-packages/hyperspy/signal.pyc in export_bss_results(self, comp_ids, folder, calibrate, multiple_files, save_figures, factor_prefix, factor_format, loading_prefix, loading_format, comp_label, cmap, same_window, no_nans, per_row, save_figures_format)
   2314                              same_window=same_window,
   2315                              per_row=per_row,
-> 2316                              save_figures_format=save_figures_format)
   2317 
   2318         self._export_loadings(loadings,

/Users/thomas/anaconda/lib/python2.7/site-packages/hyperspy/signal.pyc in _export_factors(self, factors, folder, comp_ids, multiple_files, save_figures, save_figures_format, factor_prefix, factor_format, comp_label, cmap, plot_shifts, plot_char, img_data, same_window, calibrate, quiver_color, vector_scale, no_nans, per_row)
   1568                 ensure_directory(filename)
   1569                 fac_plots[idx].savefig(filename, save_figures_format,
-> 1570                                        dpi=600)
   1571             plt.ion()
   1572 

/Users/thomas/anaconda/lib/python2.7/site-packages/matplotlib/figure.pyc in savefig(self, *args, **kwargs)
   1474             self.set_frameon(frameon)
   1475 
-> 1476         self.canvas.print_figure(*args, **kwargs)
   1477 
   1478         if frameon:

TypeError: print_figure() got multiple values for keyword argument 'dpi'
```

to266:
https://github.com/hyperspy/hyperspy/issues/785

(the problem was that `save_figures_format` was passed as an arg and hence interpreted as `dpi`. Additional kwarg for `dpi` was passed as well)

### DIFF
--- a/hyperspy/signal.py
+++ b/hyperspy/signal.py
@@ -1566,8 +1566,9 @@ class MVATools(object):
                 if folder is not None:
                     filename = os.path.join(folder, filename)
                 ensure_directory(filename)
-                fac_plots[idx].savefig(filename, save_figures_format,
-                                       dpi=600)
+                _args = {'dpi': 600,
+                         'format': save_figures_format}
+                fac_plots[idx].savefig(filename, **_args)
             plt.ion()
 
         elif multiple_files is False:
--- a/hyperspy/signal.py
+++ b/hyperspy/signal.py
@@ -1708,7 +1708,9 @@ class MVATools(object):
                 if folder is not None:
                     filename = os.path.join(folder, filename)
                 ensure_directory(filename)
-                sc_plots[idx].savefig(filename, dpi=600)
+                _args = {'dpi': 600,
+                         'format': save_figures_format}
+                sc_plots[idx].savefig(filename, **_args)
             plt.ion()
         elif multiple_files is False:
             if self.axes_manager.navigation_dimension == 2:


hyperspy/signal.py
```diff
@@ -1566,8 +1566,9 @@ def _export_factors(self,
                 if folder is not None:
                     filename = os.path.join(folder, filename)
                 ensure_directory(filename)
-                fac_plots[idx].savefig(filename, save_figures_format,
-                                       dpi=600)
+                _args = {'dpi': 600,
+                         'format': save_figures_format}
+                fac_plots[idx].savefig(filename, **_args)
```
vidartf: Why not this?
```python
fac_plots[idx].savefig(filename, format=save_figures_format, dpi=600)
```

to266: At first I though `save_figures_format` was a dictionary itself and needed updating, but apparently it's just a string. I left it like so if in the future we want to enable passing user arguments. Then `_args` can just be updated with the user-supplied dict.

Too far-fetched?

vidartf: Just a little strange without a comment, but its OK.

vidartf:
Would you mind also adding `format=save_figures_format` to `_export_loadings` call to `savefig`? It seems strange to me that it is only used this way by `_export_factors`.

to266:
Sure, added that.

vidartf:
Then of course the question: Should we not add a test for the problem in #785? :)

to266:
...I am (admittedly lazily) reluctant to add tests, at least in this PR. 
It was a mistake that the code author made, and until someone actually adds something new, the few lines will not change and will not cause the problem. 
On the other hand, the exporting functions seem to be very complex, and the saving part is probably the least significant part we should test... So while the tests should be written, I argue that a new PR should be made to test the functions fully, not just this trivial mistake.....

However that might be just lazy me :)

vidartf:
@francisco-dlp: An opinion either way?

francisco-dlp:
@vidartf, I think that we can make an exception here. The decomposition/BSS code will have to be fully refurbished soon so I think that there is no point at adding tests at this stage.

PR #813, opened 2016-01-05
IO-Plugins for Berkley EMD format

CodeMonkeyJan:
Added an io_plugins for the EMD format, a hdf5 standard proposed by Colin Ophus from Lawrence Berkeley National Lab, mentioned in #699 (see http://emdatasets.lbl.gov/ not to be confused with the FEI EMD format which was developed later, is mentioned in #698 with basic support implemented in #710).
Maybe it will be problematic if both the FEI and NCEM EMD format are supported, the reader would have to sort out what format the .emd-file is written in... I'm not sure how this could be handled, to be honest.
Happy new year, by the way :-)!


hyperspy/io_plugins/emd.py
```diff
@@ -0,0 +1,392 @@
...
+            self.add_signal(signal, name)
+
+    def __getitem__(self, key):
+        # This is for accessing the raw data easily. For the signals use emd.signals[key]!
+        return self.signals[key].data
+
+    def _write_signal_to_group(self, signal_group, signal):
+        self._log.debug('Calling _write_signal_to_group')
+        # Save data:
+        dataset = signal_group.create_group(signal.metadata.General.title)
```
to266: Do you want to allow overwriting? What if the name already exists? Consider using `require_group` (I'm not sure what behaviour you want here..)

hyperspy/io_plugins/emd.py
```diff
@@ -0,0 +1,392 @@
...
+        maxshape = tuple(None for _ in signal.data.shape)
+        dataset.create_dataset('data', data=signal.data, chunks=True, maxshape=maxshape)
+        # Iterate over all dimensions:
+        for i in range(len(signal.data.shape)):
+            key = 'dim{}'.format(i+1)
+            offset = signal.axes_manager[i].offset
```
to266: Now this bit might be confusing as well - in hyperspy, when indexing `axes_manager`, the axes are given in "natural" order - which is different from the numpy axes order. In order to have `data.shape` and `dim{x}` "aligned", you'll have to instead write
```python
ax = signal.axes_manager._axes[i]
offset = ax.offset
# and so forth
```

hyperspy/io_plugins/emd.py
```diff
@@ -0,0 +1,392 @@
...
+            units = signal.axes_manager[i].units
+            if type(units) is _Undefined:
+                units = ''
+            else:
+                units = '[{}]'.format('_'.join(list(units)))
+            dim.attrs['units'] = units
+        # Write metadata:
+        dataset.attrs['emd_group_type'] = 1
+        for key, value in signal.metadata.Signal:
+            dataset.attrs[key] = value
```
to266: This is dangerous. A user can put anything in the `signal.metadata.Signal` - lists / dictionaries / other signals. What exactly do you aim to store in the file? I'm not sure if it's better to create an explicit list of things to copy over, or just do dynamic type-checking (or, alternatively, just using `try/except` and let `h5py` handle it as best it can).

hyperspy/io_plugins/emd.py
```diff
@@ -0,0 +1,392 @@
...
+        load_to_memory: bool, optional
+            If True (default) loads data to memory. If False, enables loading only if requested.
+
+        Returns
+        -------
+        emd: :class:`~.EMD`
+            A :class:`~.EMD` object containing the loaded signals.
+
+        """
+        cls._log.debug('Calling load_from_ems')
```
to266: typo: "ems" instead of "emd"

hyperspy/io_plugins/emd.py
vidartf: Or use the public interface and write:
```python
ax = signal.axes_manager[i+3j]
```

to266: Good point, I always forget this!

francisco-dlp: Good point @vidartf. However, I was actually thinking to open an issue proposing to change that feature. I'll do it now.

hyperspy/io_plugins/emd.py
```diff
@@ -0,0 +1,392 @@
...
+        # Read in file:
+        emd_file = h5py.File(filename, 'r')
+        # Creat empty EMD instance:
+        emd = EMD()
```
to266: Why not `cls()`?

hyperspy/io_plugins/emd.py
```diff
@@ -0,0 +1,392 @@
...
+        """
+        self._log.debug('Calling save_to_emd')
+        # Open file:
+        emd_file = h5py.File(filename, 'w')
+        # Write version:
+        ver_maj, ver_min = EMD_VERSION.split('.')
+        emd_file.attrs['version_major'] = ver_maj
+        emd_file.attrs['version_minor'] = ver_min
+        # Write user:
+        user_group = emd_file.create_group('user')
```
to266: again, if overwriting only parts of the file, `require_group` is probably better.

hyperspy/io_plugins/emd.py
```diff
@@ -0,0 +1,392 @@
...
+            axes.append({'size': signal.axes_manager[i].size,
+                         'index_in_array': signal.axes_manager[i].index_in_array,
+                         'name': signal.axes_manager[i].name,
+                         'scale': signal.axes_manager[i].scale,
+                         'offset': signal.axes_manager[i].offset,
+                         'units': signal.axes_manager[i].units})
+        dictionary = {'data': signal.data,
+                      'axes': axes,
+                      'metadata': signal.metadata.as_dictionary(),
+                      'original_metadata': signal.original_metadata.as_dictionary()}
```
to266: any reason why not just `dictionary = signal._to_dictionary()` ?

hyperspy/io_plugins/emd.py
```diff
@@ -0,0 +1,392 @@
...
+        dictionaries.append(dictionary)
+    return dictionaries
+
+
+def file_writer(filename, signal, signal_metadata=None, user=None,
+                microscope=None, sample=None, comments=None, **kwds):
```
to266: @francisco-dlp would know better, but I suppose getting microscope from our metadata (like shown [here](http://hyperspy.org/hyperspy-doc/current/user_guide/getting_started.html#accessing-and-setting-the-metadata)) when none supplied makes sense?

francisco-dlp: I fully agree. All metadata should be taken from `Signal.metadata` or from `Signal.original_metadata` if the signal was loaded from another EMD file.

hyperspy/io_plugins/emd.py
```diff
@@ -0,0 +1,392 @@
...
+                units = '[{}]'.format('_'.join(list(units)))
+            dim.attrs['units'] = units
+        # Write metadata:
+        dataset.attrs['emd_group_type'] = 1
+        for key, value in signal.metadata.Signal:
+            dataset.attrs[key] = value
+
+    def _read_signal_from_group(self, name, group, load_to_memory=True):
+        self._log.debug('Calling _read_signal_from_group')
+        import hyperspy.api as hs
```
to266: Not sure how relevant that is in terms of execution time, but it's not necessary to import whole of our api, just `signals` seems to be sufficient

hyperspy/io_plugins/emd.py
CodeMonkeyJan: I changed it everywhere you suggested. I think this will only be a problem if you save a signal to an already existing .emd-file and it will then still overwrite the data in the `dataset` group but other stuff saved there is kept intact. I'm not sure myself which behaviour would be preferable, `create_group` would ensure you start with a blank slate for the signal (if you save a completely different signal under an existing name, metadata which have nothing to do with the new signal could still be there...).

hyperspy/io_plugins/emd.py
CodeMonkeyJan: I changed it into the first format, suggested by @to266 for now, I think this will be correct even if the convention for the public interface changes, right?

hyperspy/io_plugins/emd.py
CodeMonkeyJan: I'll use a `try/except` block for now and let `h5py` handle the rest. My intention was to save everything that is applicable for all signals in the dictionaries on the top level (`user`, `microscope`, etc.) and things which belong to the specific signal (which would be saved in `signal.metadata.Signal`, right?) inside the `dataset` attributes.
I didn't think about the fact that users could store more complex data in `signals.metadata.Signal`, but losing these would not be dramatic, my intention was to just save as much information as possible in the EMD.
If the demand is there, the algorithm could search through the metadata and save everything which is more complex in another dataset (besides `data` and the several `dim` ones which are always present, see the second graph at http://emdatasets.lbl.gov/spec/), but for now I would leave it as it is...

hyperspy/io_plugins/emd.py
CodeMonkeyJan: Bad habits die hard, I did this wrong for a long time and try to break the habit :-)

hyperspy/io_plugins/emd.py
CodeMonkeyJan: I didn't know about this handy function when I wrote this code and I forgot  to change it here when you mentioned it on my SEMPER plugin. It's changed now and looks much cleaner :-)!

hyperspy/io_plugins/emd.py
CodeMonkeyJan: I did my best to implement this behaviour, but it feels a bit off that I have to know the exact position of the needed dictionaries in `signal.metadata` or `signal.original_metadata`. It also currently only works if the signal was loaded by my EMD-reader.
I also made sure that on load, the `metadata` are also saved as `original_metadata`, which seems a bit redundant, because they are initially completely identical. I assume this is okay, because the `metadata` can be changed during data processing, while the `original_metadata` are kept the same?
Any suggestions are highly welcome!

hyperspy/io_plugins/emd.py
```diff
@@ -0,0 +1,402 @@
...
+            signal = signals.Signal(data)
+        # Set signal properties:
+        signal.set_signal_origin = group.attrs.get('signal_origin', '')
+        signal.set_signal_type = group.attrs.get('signal_type', '')
+        # Iterate over all dimensions:
+        for i in range(len(data.shape)):
+            dim = group.get('dim{}'.format(i+1))
+            signal.axes_manager[i].name = dim.attrs.get('name', '')
+            units = re.findall('[^_\W]+', dim.attrs.get('units', ''))
+            signal.axes_manager[i].units = ''.join(units)
```
to266: You'll have to update these lines as well, as now the axes get mixed up when saving / loading (when saving you order as `axes_manager._axes[i]`, but load as `axes_manager[i]`)

to266: ...Actually you could add a test that checks this when saving/loading (that axes do not get mixed)

CodeMonkeyJan: Do you mean a test inside the `emd.py`-file or the `test_emd.py`-file?

to266: `test_emd.py`

Thanks :)

hyperspy/tests/io/test_emd.py
```diff
@@ -0,0 +1,96 @@
...
+
+
+class TestCaseSaveAndRead():
+
+    def test_save_and_read(self):
+        signal_ref = Signal(data_save)
+        signal_ref.metadata.General.title = test_title
+        signal_ref.axes_manager[0].name = 'x'
+        signal_ref.axes_manager[1].name = 'y'
+        signal_ref.axes_manager[2].name = 'z'
```
to266: Could you also add setting non-default axis scale and offset and afterwards checking that saving + loading preserves it?

hyperspy/io_plugins/emd.py
to266: @francisco-dlp any suggestions? I agree with @CodeMonkeyJan, that the code as it is now is too fragile for my liking..

hyperspy/tests/io/test_emd.py
CodeMonkeyJan: I've added it to the save/load test, as this is the only one where I use non-default values.

By the way, the code-quality/landscape check is taking a very long time, is this intended? I think it was a lot faster for the last pull request, did something go wrong there which I should change?

to266: It's some kind of landscape bug, apparently.. Nothing for you to worry about

magnunor:
What is the status of this? Would be very useful for some stuff I'm working on :)

CodeMonkeyJan:
I'm currently converting everything to work with Python 3, afterwards I'll have to make another pull request, because I had to recreate my whole fork for some reason, it did not let me commit anymore :-/...
I'll link both requests after I'm done, I think it would be best to close this one (sorry for the trouble).
Best wishes!

PR #815, opened 2016-01-06
Prevent running tests twice

vidartf:
When testing, `hyperspy.tests.test()` in `hyperspy.tests.__init__.py` was run, effectively running all tests twice. This PR adds `@nose.tools.nottest` to it, so it is not run automatically.

The PR also removes some stray files that were either duplicates or shouldn't be there (external SVN repo).

PS: The internals of `hyperspy.tests.test()` look a little strange. Any ideas?


dnjohnstone:
Ah excellent - I wondered why that was happening! Thanks

vidartf:
PS: The decrease in coverage is in this instance only positive :)

vidartf:
Should be quick to review, and greatly speeds up tests for all other code/PRs.

vidartf:
0.8.x needs to be merged into master for master to also benefit from this!

PR #429, opened 2015-01-15
NEW: FFT and iFFT

pburdet:
Here is an implementation of FFT and iFFT as discussed in #117.

Usage
--------

```python
>>> import scipy.ndimage
>>> im = signals.Image(scipy.misc.lena())
>>> imfft = im.fft()
>>> imfft.ifft()
<Image, title: , dimensions: (|512, 512)>

```

PR #817, opened 2016-01-07
NEW: FFT and iFFT

dnjohnstone:
I have reviewed and made, what were minor, corrections to #429 from @pburdet.

As far as I can see this is now ready to merge - the diff is fairly small as all of the FFT itself is handled by the numpy implementation as is appropriate. Therefore, should be ready for review.


hyperspy/signal.py
```diff
@@ -4246,6 +4246,88 @@ def integrate_simpson(self, axis):
...
+        Return
+        ------
+        signals.Signal
+
+        Examples
+        --------
+        >>> import scipy.ndimage
+        >>> im = hs.signals.Image(scipy.misc.lena())
+        >>> im.fft()
+        <Image, title: , dimensions: (|512, 512)>
```
to266: There should be title "FFT of "

hyperspy/signal.py
```diff
@@ -4246,6 +4246,88 @@ def integrate_simpson(self, axis):
...
+        im_fft.metadata.General.title = 'FFT of ' + \
+            im_fft.metadata.General.title
+        for ax, dim in zip(axes, shape_fft):
+            axis = im_fft.axes_manager[ax]
+            axis.scale = 1. / dim / self.axes_manager[ax].scale
+            axis.units = str(self.axes_manager[ax].units) + '$^{-1}$'
+            axis.offset = -axis.high_value / 2.
+        return im_fft
+
+    def ifft(self, shape_ifft=None, axes=None):
```
to266: arguments are not used

hyperspy/signal.py
```diff
@@ -4246,6 +4246,88 @@ def integrate_simpson(self, axis):
...
+        ------
+        signals.Signal
+
+        Examples
+        --------
+        >>> import scipy.ndimage
+        >>> im = hs.signals.Image(scipy.misc.lena())
+        >>> imfft = im.fft()
+        >>> imfft.ifft()
+        <Image, title: , dimensions: (|512, 512)>
```
to266: Again, title change

hyperspy/tests/signal/test_fourier_transform.py
```diff
@@ -0,0 +1,68 @@
...
+        self.signal = im
+
+    def test_fft_ifft(self):
+        im = self.signal
+        im_fft = im.fft()
+        im_ifft = im_fft.ifft()
+        nt.assert_true(isinstance(im_ifft, Signal))
+        assert_allclose(im.data,  im_ifft.data, atol=1e-3)
+
+        im_fft = im[0].fft()
```
to266: Please don't use the old convention of slicing. If you could change this to `inav` (here and all other cases in these tests), that would help us to finally change the bad old habits :)

hyperspy/signal.py
```diff
@@ -4246,6 +4246,88 @@ def integrate_simpson(self, axis):
...
+        ax = self.axes_manager
+        axes = np.arange(ax.signal_dimension) + ax.navigation_dimension
+        im_ifft = self._deepcopy_with_new_data(np.fft.ifftn(np.fft.ifftshift(
+            self.data, axes=axes), axes=axes).real)
+        im_ifft.metadata.General.title = 'iFFT of ' + \
+            im_ifft.metadata.General.title
+        for ax, dim in zip(axes, self.axes_manager.shape):
+            axis = im_ifft.axes_manager[ax]
+            axis.scale = 1. / dim / self.axes_manager[ax].scale
+            if str(self.axes_manager[ax].units).endswith('$^{-1}$'):
```
to266: Why not do something similar with the title? Though I agree that if the data was modified between transforms, having title 'iFFT of FFT of mydata' might be useful, and the units should not "keep history of transforms"

to266:
Checking in the tests that the axes get transformed properly in both directions should also be included.

hyperspy/tests/signal/test_fourier_transform.py
dnjohnstone: Ouh thanks for that - this change kind of went over my head because the only way I learnt to do it was as above.... So the correct usage should be that when we want to select an a single signal from a stack of signals we use inav?

to266: Yes. Just to be clear, inav stands for, roughly, "index navigation" - so any indexing (that is slicing / selecting) on navigation space, be it stack / array / cube of signals, should use `inav`. Identically, any indexing on signal space should use `isig`

hyperspy/signal.py
dnjohnstone: I'm undecided on the 'best solution' here - @pburdet obviously had a situation in mind where you start with something in 'real space' do an FFT to 'reciprocal space', do some processing, and do an iFFT back and in that case I think his approach with the title is at least totally explicit, which is not usually a bad thing.

In the long run I think it is possible that we should define signal classes for 'real space' and 'reciprocal space' and be generating instances in each of those classes as we FFT in either direction. This may be useful when physically the original data may be rightly in 'reciprocal space' e.g. a stack of diffraction patterns.

For now though - whilst there are no signal classes that should be in 'reciprocal' space I think this implementation is the appropriate one. Do you agree?

dnjohnstone:
Sorry @to266 what exactly do you mean by testing that the axes get properly transformed? By the FFT action? isn't that in the domain of numpy - or have I misunderstood what you mean?

Otherwise, I've made the corrections you pointed out- thanks.

to266:
I meant checking that `axis.scale` gets transformed correctly. I agree that the title changing can be left as it is for now, and removed when some kind of `reciprocal` signal is added to hyperspy

vidartf:
Your code disappeared in one of the merges ([46f06e8](https://github.com/dnjohnstone/hyperspy/commit/46f06e81298cd2f52aa9b52fb164e59e12214352#diff-64e04530fdfa59f53ebfa39bd6b4fd7cL4164))...

dnjohnstone:
Oh that was stupid, I'll revert and merge and add the test that was requested

dnjohnstone:
@to266 This is now up to date with the current master and I think we should be good to go. Any further issues?

to266:
Yes, you did not address my last comment regarding the tests of axis scales (and probably names)

thomasaarholt:
Did @to266's comment get addressed on this?

dnjohnstone:
Not yet - I'm pretty sure it works fine, I just haven't gotten round to writing a test for that yet

thomasaarholt:
I tried using the `.fft()` on a single EELS spectrum now, and I'm not sure I'm getting everything looking correctly:

The resulting plot (from a [-5,90]eV range spectrum) looks like the following. Should 0eV<sup>-1</sup> not be centred?
**Single Spectrum FFT**
screenshot 2016-04-18 11 44 32

Also, if you compare the horizontal between the previous image and the following spectrum image FFT plot (looking at a single pixel), the calibration is entirely different.
**Spectrum Image FFT**
screenshot 2016-04-18 11 44 30

Perhaps there should be a legend added to the plot, as well?

Just ideas/thoughts, feel free to comment :)

francisco-dlp:
What's the status of this?

dnjohnstone:
@francisco-dlp I just haven't had time to get back to it and address the test improvement requested by @to266 -- I also need to spend a little time running some sanity checks, just to verify that @thomasaarholt didn't pick up a substantial error. This bit of code was a bit of a mess when I started tidying it up so I wouldn't be too surprised to find a bug.

If there's a pressing need for this let me know and I can move it up my list - otherwise it's kind of low priority for me at the moment.

francisco-dlp:

thomasaarholt:
It would be lovely if we could get this incorporated. But not necessary for me right now - just for a side project I have going.

thomasaarholt:
@dnjohnstone Would you mind resubmitting this to the minor branch? Might see if we can get this merged in the next few weeks.

woozey:
@dnjohnstone, any chance you can finilize the PR? The FFT would be extremely helpful for many of us and it seems that it's almost ready!!!

ericpre:
Since this branch needs a new PR, anybody can checkout @dnjohnstone branch, PR to the next minor branch and take this forward.
It looks like there are still a few things to do (no conflicts apparently!).

PR #818, opened 2016-01-07
Remove magic, starting script and context menu entries

### LINKED ISSUE #805: Remove hyperspy script, magic and extension?
Unlike most Python packages, hyperspy has a starting script. We decided to have one in order to make it more user friendly for the people unfamiliar with Python (most of our users). But that was a long time ago when Python wasn't as widespread in the scientific community as it is today. Until today, the only reason to keep it was to set up the chosen toolkit, which is particularly cumbersome in the case of the Qt toolkit due to the two coexisting APIs. However, #714 makes the starting script unnecessary. So, wouldn't it make sense to remove the start script?

 #767 fixes the issue of the Qt API version without needing a magic. If it gets merged, it'll be possible to start HyperSpy without issues as any other Python package e.g.

```python

import hyperspy.api as hs
%matplotlib qt4

```
Just two standard two lines that replaces the current less standard two lines:

```python
%load_ext hyperspy
%hyperspy

````

So, shouldn't we remove the extension and magic too?

TODO
--------
* [ ] Deprecate magic
* [ ] Deprecate extension
* [ ] Deprecate start script
* [ ] Remove `Preferences.General.default_toolkit`
* [ ] Add documentation
    * [ ] Importing HyperSpy
    * [ ] Importing HyperSpy headless

francisco-dlp:
Fixes #805 

The context menu entries are now available in a separate package https://github.com/hyperspy/start_jupyter_cm. You can test it by downloading the installer temporarily distributed in `dist`.


francisco-dlp:
I just updated this with the current master. It should be ready to review and merged. A good chunk of this was already merged in #819.

PR #819, opened 2016-01-07
Deprecate starting script and magic. Context menu entries removed

francisco-dlp:
It prepares for #805.

The context menu entries now have their own repository. See #818.


doc/user_guide/getting_started.rst
```diff
@@ -7,216 +7,54 @@ Getting started
...
-as ``plt`` and initialises the chosen toolkit. By default the magic looks for
-the toolkit to use in the default settings, however a different one can be
-given when running. Once executed, the magic prints which packages were
-imported, and HyperSpy is ready to be used.
-
-If the flag ``-r`` is passed as well, the magic overwrites the current input
-cell with actual code that can be executed to achieve the same result
-without any magic commands (e.g. included in a script).
+   In [1]: import hyperspy.api as hs
+   Im [2]: %matplotlib
```
to266: Typo "Im"

doc/user_guide/getting_started.rst
```diff
@@ -7,216 +7,54 @@ Getting started
...
-automatically.  However, starting with IPython version 3.0, auto-importing and
-auto-configuring is discouraged (for very good reasons). In order to comply
-with the new recommendations without loosing convenience we supply the
-``hyperspy`` IPython magic that performs the same operations in a more explicit
-and transparent way. As of IPython version 4.0 and HyperSpy 0.8.3, HyperSpy has
-to be loaded before the magic becomes available:
+Typically you will need to `set up IPython for interactive plotting with
+matplotlib
+<http://ipython.readthedocs.org/en/stable/interactive/plotting.html>`_ using the
+```%matplotlib`` magic. So, typically, after starting IPython, you can import
```
to266: too many ticks before `%matplotlib` (i.e. one is displayed as well, as in <code>`%matplotlib</code>)

francisco-dlp:
Thank, both typos now fixed.

to266:
Hm, the behaviour is strange:
```python
import hyperspy.api as hs
import numpy as np
s = hs.signals.Spectrum(np.random.random((10,10)))
s.plot()
```
does nothing, just as expected, since the `matplotlib` is not defined. 

However later running `%matplotlib` returns a very long and completely useless error that, after all, enables plotting. You can only figure something out if you run 

```python
In [11]: %matplotlib qt
Warning: Cannot change to a different GUI toolkit: qt. Using tk instead.
```

If a model is created after that a component is added, it's not possible to plot the model, since, according to 
```python
<traitsui.null.toolkit.GUIToolkit object at 0x7f75105404d0>
```
(from the TB of the error), traitsui has `null` toolkit... So I guess many things do not work.

vidartf:
See also #820 ?

francisco-dlp:
@to266, I couldn't reproduce the problem that you describe. I think that it might be due to the default matplotlit backend in your system being tk intead of Qt4. However, I think that this is no longer a HyperSpy issue. Any user that sets up IPython properly for interactive plotting should have a smooth experience.

In my last commit I have added some extra details on how to set up IPython for interactive plotting.

francisco-dlp:
Also, see my comment in #820.

francisco-dlp:
Ready for review.

PR #822, opened 2016-01-11
Use `stash_active_state`

vidartf:
Move `stash_active_state` to misc.utils, and use it more places. Makes the internal code look neater, and reuses code.

### DIFF
--- a/hyperspy/tests/component/test_component_active_array.py
+++ b/hyperspy/tests/component/test_component_active_array.py
@@ -20,6 +20,7 @@ import nose.tools as nt
 import numpy as np
 from hyperspy.components import Gaussian
 from hyperspy.signal import Signal
+from hyperspy.misc.utils import stash_active_state
 
 
 class TestParametersAsSignals:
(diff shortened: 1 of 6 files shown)


vidartf:
No new unittests should be needed, as the code in question is sufficiently covered by existing tests.

dnjohnstone:
As far as I can see this is all good... Only substantive changes are in model.py (others are tidying and moving bits around) and as far as I can see that all does what it's supposed to.

hyperspy/tests/component/test_component_active_array.py
```diff
@@ -52,15 +53,14 @@ def test_toggle_array(self):
...
-            g._active_array == np.array([[0, 1, 1], [1, 1, 1], [0, 1, 1]], dtype=bool)))
-        g._toggle_connect_active_array(True)
+        with stash_active_state([g]):
+            g.active_is_multidimensional = False
+            nt.assert_false(g._active_is_multidimensional)
+            nt.assert_true(
+                np.all(
+                    g.A.as_signal('values').data == np.zeros(
+                        (3, 3))))
+            nt.assert_equal(g._active_array, None)
```
to266: Could you add a test that the restored `_active_array` is correct? Ignore if the behaviour is tested somewhere else and I just did not notice....

vidartf: The code two lines further down checks this, no? The same that checked that `g._toggle_connect_active_array(True)` did its job previously.

to266: I would interpret the code as checking that `as_signal` correctly puts nans where the active array is `False`. In fact it only checks those particular "spots", so even if the result array was filled with nans, it would still pass. Also, there used to be a test that explicitly checked the `_active_array`. So I'd like to keep it there - it may have been put there for a specific corner-case, and I just don't remember...

vidartf: Ah, now I understand what you were talking about. The test line in question confused me a bit. It explicitly tested the state of the `_active_array` while disconnected. The current PR changes the internal behavior so that the array is set to `None` while stashed. I have re-introduced the test in question on the outside of the `with` block though, as you initially proposed :)

Issue #814, opened 2016-01-05
Change axes_manager indexing feature


francisco-dlp:
Currently it is possible to index the `AxesManager` using complex number. From the docstring:

```In addition AxesManager support indexing using
complex numbers a + bj, where b can be one of 0, 1, 2 and 3 and a a valid
index. If b is 3 AxesManager is indexed using the order of the axes in the
array. If b is 1(2), indexes only the navigation(signal) axes in the
natural order. ```

However, #388 proposes to avoid mixing signal and navigation indexing as this can be confusing. I propose to extend this to the `AxesManager` indexing feature by replacing the previous convention with a simpler (and more limited) syntax: real numbers to index the navigation axes, pure imaginary numbers for signal indexing. What do you think?

vidartf:
Complicated issue. I agree this could benefit from a makeover, as there is no intuitive logic behind having 0j be natural order, 1j be nav, 2j signal and 3j be array order. This makes it difficult to remember, and makes existing code harder to read. On the other hand, accessing axis X in a specific order is really useful. It is also useful to have a single function that can take an index in any kind of order, as well as an axis name (and I'd which that it would check if the argument is a `DataAxis` as well, which could simply be returned as it is).

So for me, I'm all for looking for a better solution, but I think we should brain-storm a little to see if we can't find a solution that gets rid of the cons, but keeps all/most the pros of the existing way.

to266:
1. I agree that using real(imaginary) indices to index navigation(signal) axes is going in the right direction
2. Nevertheless, for internal usage I quite like that it's possible to to index the axes in array order using public methods, even though I do not use it because I always forget it exists..

...as a side-note/joke, prof. Gull, while teaching us dynamics, mentioned [this](http://geometry.mrao.cam.ac.uk/wp-content/uploads/2015/02/ImagNumbersArentReal.pdf). Shall we just make use of it and have all three nicely separated? ;)

to266:
(as for the Geometric algebra, I may be misremembering things.... so it might not be as relevant as I hoped. An interesting read nevertheless)

francisco-dlp:
[...]

PR #422, opened 2015-01-14
Signal operations over all navigation or signal axes

francisco-dlp:
Currently, the only way to e.g. sum over all navigation or signal axes is to sum the axes one by one. Because this is a common task I  thought on implementing this shortcut:

```python
In [1]: im = signals.Image(np.random.random((2, 3, 4, 5, 6)))

In [2]: im
Out[2]: <Image, title: , dimensions: (4, 3, 2|6, 5)>

In [3]: im.sum(axis="navigation")
Out[3]: <Image, title: , dimensions: (|6, 5)>

In [4]: im.sum(axis="signal")
Out[4]: <Signal, title: , dimensions: (4, 3, 2|)>

```

Also, integrating over the navigation space is now the default.

There is a drawback though: if an axis name is "signal" or "navigation" it won't be clear if the user wants to sum over the axis or the "space". A workaround would be to use "_signal_" and "_navigation_" instead. Any better idea?

TODO
--------

- [ ] Write unittest
- [ ] Implement the feature in other methods that may benefit from it if any
- [ ] Update docstrings
- [ ] Update User Guide
- [ ] Reserve "signal" and "navigation" axis names.
- [ ] Pass axes tuple.

PR #424, opened 2015-01-14
Enh add out arg

francisco-dlp:
Many `Signal` methods create and return a new `Signal`. For fast operations, the new `Signal` creation time is non-negligible. Although this is irrelevant in most cases, it often has an impact when performing the operation interactively. Here I propose a solution that consists in providing the output `Signal` to the method so that it does not need to create it itself.

Usage
--------

```python

In [1]: s = signals.Spectrum(np.arange(10))

In [2]: s_sum = s.sum(0)

In [3]: s_sum.data
Out[3]: array(45)

In [4]: s[:5].sum(0, out=s_sum)

In [5]: s_sum.data
Out[5]: 10

In [6]: s_roi = s[:3]

In [7]: s_roi
Out[7]: <Spectrum, title: , dimensions: (|3)>

In [8]: s.__getitem__(slice(None, 5), out=s_roi)

In [9]: s_roi
Out[9]: <Spectrum, title: , dimensions: (|5)>

```

TODO
--------

- [x] Docstrings.
- [x] unittests.
- [x] User guide documentation.
- [x] Implement the feature in more methods that may benefit from it if any.

PR #823, opened 2016-01-12
axes_manager getitem iterable

to266:
Enables `axes_manager` to deal with iterables and axes in `getitem`. Hence `signal.sum(('x', 'y', 'z'))` and similar things are possible.

Main reason for this PR is that when working with large datasets, recreating a new signal for every operation (when multiple are "chained") is very expensive..

as a bonus, things like
```python
s.sum(s.axes_manager.signal_axes + s.axes_manager.navigation_axes[:-1])
```
works as well, if you want to sum over all except the last navigation axis.

Let me know if you have any ideas (if and where) I should mention this in user guide and what else to test

probably solves a part of https://github.com/hyperspy/hyperspy/issues/814, and implements https://github.com/hyperspy/hyperspy/pull/422 to some degree..

### DIFF
--- a/hyperspy/axes.py
+++ b/hyperspy/axes.py
@@ -504,15 +504,11 @@ class AxesManager(t.HasTraits):
         """x.__getitem__(y) <==> x[y]
 
         """
-        return_one = False
         if isinstance(y, basestring) or not np.iterable(y):
-            y = (y,)
-            return_one = True
+            return self[(y,)][0]
         axes = [self._axes_getter(ax) for ax in y]
         _, indices = np.unique(axes, return_index=True)
         ans = tuple(axes[i] for i in sorted(indices))
-        if return_one:
-            ans = ans[0]
         return ans
 
     def _axes_getter(self, y):
--- a/hyperspy/axes.py
+++ b/hyperspy/axes.py
@@ -504,6 +504,20 @@ class AxesManager(t.HasTraits):
         """x.__getitem__(y) <==> x[y]
 
         """
+        return_one = False
+        if isinstance(y, basestring) or not np.iterable(y):
+            y = (y,)
+            return_one = True
+        axes = [self._axes_getter(ax) for ax in y]
+        _, indices = np.unique(axes, return_index=True)
+        ans = tuple(axes[i] for i in sorted(indices))
+        if return_one:
+            ans = ans[0]
+        return ans
+
+    def _axes_getter(self, y):
+        if y in self._axes:
+            return y
         if isinstance(y, basestring):
             axes = list(self._get_axes_in_natural_order())
             while axes:
@@ -556,7 +570,16 @@ class AxesManager(t.HasTraits):
                      else tuple())
         return nav_shape + sig_shape
 
-    def remove(self, axis):
+    def remove(self, axes):
+        """Remove one or more axes
+        """
+        axes = self[axes]
+        if not np.iterable(axes):
+            axes = (axes,)
+        for ax in axes:
+            self._remove_one_axis(ax)
+
+    def _remove_one_axis(self, axis):
         """Remove the given Axis.
 
         Raises
@@ -564,7 +587,7 @@ class AxesManager(t.HasTraits):
         ValueError if the Axis is not present.
 
         """
-        axis = self[axis]
+        axis = self._axes_getter(axis)
         axis.axes_manager = None
         self._axes.remove(axis)
 
--- a/hyperspy/docstrings/signal.py
+++ b/hyperspy/docstrings/signal.py
@@ -1,0 +1,15 @@
+# -*- coding: utf-8 -*-
+"""Common docstring snippets for signal.
+
+"""
+
+ONE_AXIS_PARAMETER = \
+    """: {int | string | axis}
+            The axis can be passed directly, or specified using the index of the
+            axis in `axes_manager` or the axis name."""
+MANY_AXIS_PARAMETER = \
+    """: {int | string | axis | tuple}
+            Either one on its own, or many axes in a tuple can be passed. In
+            both cases the axes can be passed directly, or specified using the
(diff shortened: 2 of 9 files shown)


to266:
Just to be clear, duplicates are removed, but unique order is preserved:
```python
In [2]: s = hs.signals.Signal(np.random.random((3,2,4,5)))

In [3]: s.axes_manager[-2].name = 'x'   

In [4]: interesting_axes = ('x', 0, -1, s.axes_manager.navigation_axes[-1])

In [5]: interesting_axes
Out[5]: ('x', 0, -1, <x axis, size: 3, index: 0>)

In [6]: axes = s.axes_manager[interesting_axes]

In [7]: axes
Out[7]: 
(<x axis, size: 3, index: 0>,
 <Unnamed 0th axis, size: 4, index: 0>,
 <Unnamed 3rd axis, size: 5>)

In [8]: s.axes_manager['x']
Out[8]: <x axis, size: 3, index: 0>

In [9]: s.axes_manager[0]
Out[9]: <Unnamed 0th axis, size: 4, index: 0>

In [10]: s.axes_manager[-1]
Out[10]: <Unnamed 3rd axis, size: 5>

In [11]: s.axes_manager.navigation_axes[-1]
Out[11]: <x axis, size: 3, index: 0>


```

vidartf:
I like the principle a lot! I'll go through the details, but I think this can well replace #422.

hyperspy/axes.py
```diff
@@ -504,6 +504,20 @@ def __getitem__(self, y):
         """x.__getitem__(y) <==> x[y]
 
         """
+        return_one = False
+        if isinstance(y, basestring) or not np.iterable(y):
+            y = (y,)
+            return_one = True
```
vidartf: Maybe more elegant to replace `return_one = True` with `return self[(y,)][0]` ? Simplifies the rest of the function as well, but relies on recursion (might be an issue with inheritance).

to266: Well spotted, will push in a second

hyperspy/signal.py
```diff
@@ -3936,12 +3934,15 @@ def _iterate_signal(self):
...
+    def _remove_axes(self, axes):
         am = self.axes_manager
-        axis = am[axis]
-        if am.navigation_dimension + am.signal_dimension > 1:
-            am.remove(axis.index_in_axes_manager)
-            if axis.navigate is False:  # The removed axis is a signal axis
+        axes = am[axes]
+        if not np.iterable(axes):
+            axes = (axes,)
+        if am.navigation_dimension + am.signal_dimension > len(axes):
```
vidartf: Implement and use?
```python
def __len__(self):
    return len(self._axes)
```
`am.navigation_dimension + am.signal_dimension` --> `len(am)`

vidartf: Or maybe that can be confused with iteration length (since AxesManager iterates of navigation axes)..

vidartf:
As mentioned in #422, the docstring for `axes` can probably be extracted to `hyperspy.docstring.signal.py` similarly to what is done in #424. Other than that I have two questions for discussion:
 1. You suggest renaming `axis` args to `axes`. That will break any existing keyword use. That's OK with me, but it might break users' scripts.
 2. Functions that still only take a single `axis`, should those be extended to also take multiple axes (apply inner code iteratively)?

francisco-dlp:
Regarding point 1, I suggest not to rename `axis` -> `axes`. This, in addition of not breaking a lot of code out there, is consistent with numpy's syntax.

vidartf:
For (2): `Signal.valuemax` can just wrap `np.amax` which supports tuple `axis` arg. We should also consider renaming it as `amax`, and making `valuemax` a (deprecated?) alias.

vidartf:
For (1), @francisco-dlp's point that numpy uses `axis` makes me be +1 on keeping `axis`. That would also make this PR fully compatible with the API, as only protected functions change name.

to266:
I've moved the `axis` docstring out, renamed back to the original name the keyword (even though it never actually was a "keyword"... it's just a positional argument in our case, hence should not matter).

I've also added `signal.amax`, since it does not actually deprecate `valuemax` - they do different things. `indexmax` returns indices of the maximum, `valuemax` returns axes values at those indices, and `amax` - the actual maximum

vidartf:
`EDSSpectrum.sum()` currently fails when called with a tuple of axes (trying to plot a spectrum image e.g.). On the line:
```python
self.axes_manager[axis].size
```
Should probably also add a test or modify the existing test to check that the live time calculation is correct also for more than one axis.

to266:
I've fixed the bug, and in general made the calculation more robust (at least I think so..)

As for the (2), I don't think it's worth implementing the functions ourselves. If I really end up needing that, I will at another PR, however here I think just making use of what numpy is capable of is enough

hyperspy/docstrings/signal.py
```diff
@@ -0,0 +1,15 @@
...
+
+"""
+
+ONE_AXIS_PARAMETER = \
+    """: {int | string | axis}
+            The axis can be passed directly, or specified using the index of the
+            axis in `axes_manager` or the axis name."""
+MANY_AXIS_PARAMETER = \
+    """: {int | string | axis | tuple}
+            Either one on its own, or many axes in a tuple can be passed. In
```
francisco-dlp: Shouldn't the default be summing over all navigation axes?

vidartf:
Aside from what @francisco-dlp has found / will find, I'm happy with its current state. If you'd like to be the second opinion @francisco-dlp then you can merge it when you think it is OK :)

hyperspy/docstrings/signal.py
to266: You think I should add it in this PR? Here I only expanded API without changing behaviour, as that was not the intended purpose of the PR.

francisco-dlp: I agree.

PR #824, opened 2016-01-13
by default signal performs simple operations on navigation axes

to266:
e.g. `s.sum()` now sums over navigation axes if not argument is passed

### DIFF
--- a/doc/user_guide/tools.rst
+++ b/doc/user_guide/tools.rst
@@ -243,7 +243,7 @@ subclasses.
 Simple mathematical operations
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
-.. versionchaned:: 0.9
+.. versionchanged:: 0.9
 
 A number of simple operations are supported by :py:class:`~.signal.Signal`. Most
 of them are just wrapped numpy functions, as an example:
--- a/doc/user_guide/tools.rst
+++ b/doc/user_guide/tools.rst
@@ -240,13 +240,48 @@ The methods of this section are available to all the signals. In other chapters
 methods that are only available in specialized
 subclasses.
 
+Simple mathematical operations
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+.. versionchaned:: 0.9
+
+A number of simple operations are supported by :py:class:`~.signal.Signal`. Most
+of them are just wrapped numpy functions, as an example:
+
+.. code-block:: python
+
+    >>> s = hs.signals.Signal(np.random.random((2,4,6)))
+    >>> s.axes_manager[0].name = 'E'
+    >>> s
+    <Signal, title: , dimensions: (4, 2|6)>
+    >>> # by default perform operation over all navigation axes
+    >>> s.sum()
+    <Signal, title: , dimensions: (|6)>
+    >>> # can also pass axes individually
+    >>> s.sum('E')
+    <Signal, title: , dimensions: (2|6)>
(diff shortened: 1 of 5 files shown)


hyperspy/docstrings/signal.py
```diff
@@ -8,8 +8,8 @@
             The axis can be passed directly, or specified using the index of the
             axis in `axes_manager` or the axis name."""
 MANY_AXIS_PARAMETER = \
-    """: {int | string | axis | tuple}
+    """: {int | string | axis | tuple | None}
             Either one on its own, or many axes in a tuple can be passed. In
             both cases the axes can be passed directly, or specified using the
             index in `axes_manager` or the name of the axis. Any duplicates are
-            removed."""
+            removed. If None, navigation axes are passed"""
```
francisco-dlp: If `None`, the operation is performed over all navigation axes?

francisco-dlp:
In #823 I forgot to ask for `User Guide` documentation. This is actually a good chance to add it. Maybe just an example with e.g. `sum` to show that the default is summing over all navigation axes and then show how to sum over one particular axis and all signal axes?

to266:
...adding a new tests decreases coverage? any way, @francisco-dlp suggestions added

doc/user_guide/tools.rst
```diff
@@ -240,13 +240,48 @@ The methods of this section are available to all the signals. In other chapters
 methods that are only available in specialized
 subclasses.
 
+Simple mathematical operations
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+.. versionchaned:: 0.9
```
francisco-dlp: Version added has no text and, therefore, it doesn't appear. What about "Passing multiple axes in a tuple."?

to266: I just noticed that I misspelled it - hence does not appear. "Passing.." for a title you mean?

francisco-dlp: No, I was meaning adding that little description inside  `versionchanged` not to confuse the reader in thinking that all that is below is new in 0.9.

to266: Well it's just changed, not new... Hopefully it's clear enough now (I think it is)

francisco-dlp:
The coverage has decreased for a good reason, you only test the default `axis` value for `sum`.

to266:
I forgot I added a new function (`amax`, which might be more obvious as `max`, as in numpy it's just an alias), hence the reduction in coverage. Should I add tests for all the other functions as well, even though they use identical 3 lines in each case?

hyperspy/docstrings/signal.py
```diff
@@ -8,8 +8,9 @@
...
             axis in `axes_manager` or the axis name."""
 MANY_AXIS_PARAMETER = \
-    """: {int | string | axis | tuple}
+    """: {int | string | axis | tuple | None}
             Either one on its own, or many axes in a tuple can be passed. In
             both cases the axes can be passed directly, or specified using the
             index in `axes_manager` or the name of the axis. Any duplicates are
-            removed."""
+            removed. If None, the operation is performed over all navigation
+            axes."""
```
vidartf: add "(default)." at end?

to266:
Added the "(default)", removed `amax`, as it was just an alias.

to266:
Not sure if I do need to add all the other tests, even though the code is identical @francisco-dlp ..?

francisco-dlp:
I think yes, even if the code is identical, they may get broken in the future individually and we won't know unless we have individual tests for them.

to266:
Updated

PR #825, opened 2016-01-14
Enh signal oom operations

to266:
Builds on https://github.com/hyperspy/hyperspy/pull/736 and adds most of the simple operations for oom signal.

The way to calculate dask chunks may not be the best, as it's been a while since I've come up with it, and only had synthetic data to play with, so it's open for suggestions / discussion.

Now that I think about it, any kind of suggestions or comments would be nice

### TODO
 - [ ] Tests


to266:
https://github.com/hyperspy/hyperspy/pull/1102

Issue #796, opened 2015-12-15
Remove 0.9 deprecated code from master


vidartf:
As 0.9 will be the next release on master, we should remove the deprecated code as soon as possible to work on a clean code-base.

francisco-dlp:
Fully agree. We should have opened this issue just after releasing 0.8.

PR #828, opened 2016-01-15
Remove deprecated code for 0.9

vidartf:
As suggested in #796.

### DIFF
--- a/doc/user_guide/eels.rst
+++ b/doc/user_guide/eels.rst
@@ -46,8 +46,8 @@ The :py:meth:`~._signals.eels.EELSSpectrum.estimate_zero_loss_peak_centre` can b
 
 The :py:meth:`~._signals.eels.EELSSpectrum.align_zero_loss_peak` can
 align the ZLP with subpixel accuracy. It is more robust and easy to use than
-:py:meth:`~.signal.Signal1DTools.align1D` for the task. Note that it is 
-possible to apply the same alignment to other spectra using the `also_align` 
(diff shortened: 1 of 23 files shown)


doc/user_guide/eels.rst
```diff
@@ -148,42 +148,10 @@ HyperSpy has created the model and configured it automatically:
...
-    <background (PowerLaw component)>
-
-.. warning::
-
-   This feature is deprecated and will be removed in HyperSpy 0.9. To access
-   the automatically created component use the following equivalent syntax:
-
-    .. code-block:: python
-
-        >>> m.components.N_K
```
to266: This and the following lines should be left, but with altered text around it - we do need an example how to access the components.

Furthermore, I'm not sure if we still inject variables named after element symbol - I think (hope!) not, so the text after this deleted chunk should be removed / altered as well

hyperspy/signal.py
```diff
@@ -1011,7 +1010,7 @@ def integrate_in_range(self, signal_range='interactive'):
     def _integrate_in_range_commandline(self, signal_range):
         e1 = signal_range[0]
         e2 = signal_range[1]
-        integrated_spectrum = self[..., e1:e2].integrate1D(-1)
+        integrated_spectrum = self.isig[..., e1:e2].integrate1D(-1)
```
to266: I would remove the ellipsis here

francisco-dlp:
Are you happy with this now @to266?

PR #829, opened 2016-01-15
Use warnings betters / fix warnings

vidartf:
Includes #828 and #776.

There are still some tests that fail, mostly because I'm not certain how to best fix them, so suggestions there are very welcome.


vidartf:
For those that want to use the commits to clean up the diff: Only the last four commits are unique to this PR.

vidartf:
For easy reading, I'll post the warnings that are difficult:
```python
======================================================================
ERROR: hyperspy.tests.model.test_model.TestModelWeighted.test_fit_fmin_unbinned
======================================================================
  File "/home/travis/miniconda2/envs/testenv/lib/python2.7/site-packages/nose/case.py", line 197, in runTest
    self.test(*self.arg)
  File "/home/travis/build/hyperspy/hyperspy/hyperspy/tests/model/test_model.py", line 666, in test_fit_fmin_unbinned
    method="ls",
  File "/home/travis/build/hyperspy/hyperspy/hyperspy/model.py", line 741, in fit
    self.channel_switches]
VisibleDeprecationWarning: boolean index did not match indexed array along dimension 0; dimension is 9000 but corresponding boolean dimension is 900
```
I think this is a numpy warning, but we should probably update our code to avoid this becoming an issue later.

```python
======================================================================
ERROR: hyperspy.tests.model.test_components.TestScalableFixedPattern.test_both_unbinned
======================================================================
  File "/home/travis/miniconda2/envs/testenv/lib/python2.7/site-packages/nose/case.py", line 197, in runTest
    self.test(*self.arg)
  File "/home/travis/build/hyperspy/hyperspy/hyperspy/tests/model/test_components.py", line 263, in test_both_unbinned
    m.fit()
  File "/home/travis/build/hyperspy/hyperspy/hyperspy/model.py", line 772, in fit
    self.p_std = np.sqrt(np.diag(pcov))
RuntimeWarning: invalid value encountered in sqrt
```

```python
======================================================================
ERROR: hyperspy.tests.model.test_model_as_dictionary.TestModelDictionary.test_to_dictionary
======================================================================
  File "/home/travis/miniconda2/envs/testenv/lib/python2.7/site-packages/nose/case.py", line 197, in runTest
    self.test(*self.arg)
  File "/home/travis/build/hyperspy/hyperspy/hyperspy/tests/model/test_model_as_dictionary.py", line 235, in test_to_dictionary
    d['components'][-1]['_whitelist']['spectrum'][1], m.spectrum * 0.3)
  File "/home/travis/miniconda2/envs/testenv/lib/python2.7/unittest/case.py", line 513, in assertEqual
    assertion_func(first, second, msg=msg)
  File "/home/travis/miniconda2/envs/testenv/lib/python2.7/unittest/case.py", line 503, in _baseAssertEqual
    if not first == second:
  File "<string>", line 2, in __eq__
  File "/home/travis/build/hyperspy/hyperspy/hyperspy/signal.py", line 2945, in _binary_operator_ruler
    exec("result = self.data.%s(other)" % op_name)
  File "<string>", line 1, in <module>
DeprecationWarning: elementwise == comparison failed; this will raise an error in the future.
```

This code actually fails, but nose keeps it quiet for some reason (that reason will no longer hold in the future though, it seems like). Not entirely sure what the code here is trying to do though. @to266 ?

to266:
https://github.com/vidartf/hyperspy/pull/15

vidartf:
Ok, I've resolved the two major warnings above. Note that the numpy one was fixed by using `numpy.where()` (see ae05263). The tests still work, but I've not gone over that exhaustively.

There are still some warnings that only appear sporadically (zero divisions and invalid values to `np.sqrt` as above). These should simply be filtered in the tests, as I've done for two warnings in 66a9a21.

vidartf:
PS: This should now be ready to review/merge.

to266:
Could you add tests for the new `assert_warns` and (for the sake of completion) `ignore_warning`?

To be more exact, for the former I would play with messages (giving longer than what is raised / part of it / different spaces and/or capitals), as I feel it's the most fidgety link. @francisco-dlp what do you think about having a complete warning message in the test, instead of just the most important bits?

Then again, I guess some sort of fuzzy matching would be both not sensitive enough and complicated..?

vidartf:
@to266 : I see your point. Currently `ignore_warning` uses the regex match via `warnings.filterwarnings()`, while `assert_warns` uses a sub-string check. Do you think we should switch the way `assert_warns` checks to match that of `ignore_warning`, or simply test the current behavior?

to266:
I think it's fine to use the in-built regex match. 

For `assert_warns` ideally I would prefer to use, as mentioned, a fuzzy matching (e.g. http://chairnerd.seatgeek.com/fuzzywuzzy-fuzzy-string-matching-in-python/ with the corresponding github page), however that might be over-complicated... So unless you feel the `fuzzywuzzy` is worth having, I'd say just simply test the current behaviour :)

vidartf:
That was my suggestion: To make `assert_warns` "more fuzzy" by allowing regex matching like filter_warning, but I think fuzzywuzzy is overkill (and too fuzzy for unit tests).

to266:
If @francisco-dlp agrees, I'd go for regex then. Otherwise, just tests.

francisco-dlp:
Fine for me too.

vidartf:
Should be OK now.

PR #835, opened 2016-01-18
Add support for FEI files that do not store the axes calibration in the file header

### LINKED ISSUE #834: Error when loading some line scans saved in FEI emi/ser format
Loading some FEI emi/ser files fail like this:

```python
In [2]: s = hs.load("15.57.26 Spectrum image_1.ser", verbose=True)
Opening the file:  15.57.26 Spectrum image_1.ser
Extracting the information


Header info:
------------
ByteOrder : 18761
SeriesID : 407
SeriesVersion : 544
DataTypeID : 16672
TagTypeID : 16706
TotalNumberElements : 9100
ValidNumberElements : 9100
OffsetArrayOffset : 126
NumberDimensions : 0


Data info:
----------
CalibrationOffset : 0.0
CalibrationDelta : 5.0
CalibrationElement : 0
DataType : 3
ArrayLength : 4000
Array : Array
TagTypeID : 16706
Unknown : 0
Time : 1452783447
PositionX : -1.92947143423e-07
PositionY : 1.10481647437e-07
<ipython-input-2-291ccfff431c> in <module>()
----> 1 s = hs.load("15.57.26 Spectrum image_1.ser", verbose=True)

/home/fjd29/Python/hyperspy/hyperspy/io.py in load(filenames, record_by, signal_type, signal_origin, stack, stack_axis, new_axis_name, mmap, mmap_dir, **kwds)
    199             objects = [load_single_file(filename,
    200                                         **kwds)
--> 201                        for filename in filenames]
    202 
    203         if hyperspy.defaults_parser.preferences.Plot.plot_on_load:

/home/fjd29/Python/hyperspy/hyperspy/io.py in load_single_file(filename, record_by, signal_type, signal_origin, **kwds)
    253                                 signal_type=signal_type,
    254                                 signal_origin=signal_origin,
--> 255                                 **kwds)
    256 
    257 

/home/fjd29/Python/hyperspy/hyperspy/io.py in load_with_reader(filename, reader, record_by, signal_type, signal_origin, **kwds)
    264     file_data_list = reader.file_reader(filename,
    265                                         record_by=record_by,
--> 266                                         **kwds)
    267     objects = []
    268 

/home/fjd29/Python/hyperspy/hyperspy/io_plugins/fei.py in file_reader(filename, *args, **kwds)
    259     ext = os.path.splitext(filename)[1][1:]
    260     if ext in ser_extensions:
--> 261         return [ser_reader(filename, *args, **kwds), ]
    262     elif ext in emi_extensions:
    263         return emi_reader(filename, *args, **kwds)

/home/fjd29/Python/hyperspy/hyperspy/io_plugins/fei.py in ser_reader(filename, objects, verbose, *args, **kwds)
    420         if dc.dtype is np.dtype('f') or dc.dtype is np.dtype('f8'):
    421             dc[:] = np.nan
--> 422         dc[:data['Array'].ravel().shape[0]] = data['Array'].ravel()
    423     else:
    424         dc = data['Array']

ValueError: could not broadcast input array from shape (36400000) into shape (4000)
```

The problem is that HyperSpy currently get the number of dimensions from `header["NumberOfDimensions"]`. However, certain files do not include extra dimension to define the axis calibration.

francisco-dlp:
Some FEI ser files do not store the calibration of the axes in the file header. This code calculates the calibration from the position coordinates when available. When not available (possibly when the navigation dimension is time) it adds an uncalibrated axis called "Unkown dimension".

Also, this adds previously missing units to image files stored in this format.

Note that this PR doesn't include a test as I don't have easy access to the FEI software that could generate the test files. This is a general issue with the FEI reader. I have tested it with all the ser files that I have (~400) and everything seem to work properly.

 Fixes #834.

### DIFF
--- a/hyperspy/io_plugins/fei.py
+++ b/hyperspy/io_plugins/fei.py
@@ -317,6 +317,23 @@ def get_xml_info_from_emi(emi_file):
 
 
 def get_calibration_from_position(position):
+    """Compute the size, scale and offset of a linear axis from coordinates.
+
+    This function assumes rastering on a regular grid for the full size of
+    each dimension before rastering over another one. Fox example: a11, a12,
+    a13, a21, a22, a23 for a 2x3 grid.
+
+    Parameters
+    ----------
+    position: numpy array.
+        Position coordinates of the axis. Normally as in PositionX/Y of the
+        ser file.
+
+    Returns
+    -------
+    axis_attr: dictionary with `size`, `scale`, `offeset` keys.
+
+    """
     offset = position[0]
     for i, x in enumerate(position):
         if x != position[0]:
@@ -334,7 +351,8 @@ def get_calibration_from_position(position):
             size = j + 1
         else:  # Second rastering dimension
             size = len(position) / i
-    return {"size": size, "scale": scale, "offset": offset}
+    axis_attr = {"size": size, "scale": scale, "offset": offset}
+    return axis_attr
 
 
 def get_axes_from_position(header, data):
--- a/hyperspy/io_plugins/fei.py
+++ b/hyperspy/io_plugins/fei.py
@@ -62,6 +62,7 @@ data_types = {
 
 XY_TAG_ID = 16706  # header contains XY calibration
 
+
 def readLELong(file):
     """Read 4 bytes as *little endian* integer in file"""
     read_bytes = file.read(4)
--- a/hyperspy/io_plugins/fei.py
+++ b/hyperspy/io_plugins/fei.py
@@ -166,7 +166,7 @@ def get_data_tag_dtype_list(data_type_id):
             ("PositionX", "<f8"),
             ("PositionY", "<f8"),
         ]
-    else:  # elif data_type_id == ?????
+    else:  # elif data_type_id == ?????, 16722?
         header = [
             ("TagTypeID", ("<u2")),
             # Not in Boothroyd description. = 0. Not tested.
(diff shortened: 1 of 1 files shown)


francisco-dlp:
It turns out that also 2D scans can be saved in this way. I'll update the code to account for this too, so no need to review yet.

francisco-dlp:
I've just updated the code. This is ready for review.

francisco-dlp:
Could somebody consider merging this?

hyperspy/io_plugins/fei.py
```diff
@@ -316,6 +316,97 @@ def get_xml_info_from_emi(emi_file):
     return objects[:-1]
 
 
+def get_calibration_from_position(position):
+    offset = position[0]
+    for i, x in enumerate(position):
+        if x != position[0]:
+            break
+    if i == len(position) - 1:
```
vidartf: Will this be correct if it breaks on the last iteration? Consider using the `for ... else` syntax, or commenting code. Maybe add a docstring to the function?

hyperspy/io_plugins/fei.py
```diff
@@ -316,6 +316,97 @@ def get_xml_info_from_emi(emi_file):
...
+        else:  # Second rastering dimension
+            size = len(position) / i
+    return {"size": size, "scale": scale, "offset": offset}
+
+
+def get_axes_from_position(header, data):
+    array_shape = []
+    axes = []
+    array_size = int(header["ValidNumberElements"])
+    if data["TagTypeID"][0] == 16706:
```
vidartf: Magic number are bad. Assign it to a constant, or at the very least add a comment.

hyperspy/io_plugins/fei.py
vidartf: Is this equivalent to `np.any(position != position[0]), i = np.where(position != position[0])[0][0]`?

vidartf:
Kind of hard to review without knowing how the files are, but I tried to review the diff.

hyperspy/io_plugins/fei.py
francisco-dlp: Probably, but that's less readable and possibly slower in most practical cases.

francisco-dlp:
Thank you @vidartf. Everything should be fixed now.

hyperspy/io_plugins/fei.py
vidartf: How about the first comment? Especially the part if it fails when breaking on last iteration.

Regarding the second case, I asked because I had trouble following the code.

hyperspy/io_plugins/fei.py
```diff
@@ -316,6 +317,115 @@ def get_xml_info_from_emi(emi_file):
...
+        size = 0
+    else:
+        scale = x - position[0]
+        if i == 1:  # Rastering over this dimension first
+            for j, x in enumerate(position[1:]):
+                if x == position[0]:
+                    break
+            size = j + 1
+        else:  # Second rastering dimension
+            size = len(position) / i
```
vidartf: Won't this fail for some edge cases? How about this array: `position = [0,0,1,1,1,2,2]` Or is array of different type (if so, why use `i`)?

hyperspy/io_plugins/fei.py
```diff
@@ -316,6 +317,115 @@ def get_xml_info_from_emi(emi_file):
...
+    for i, x in enumerate(position):
+        if x != position[0]:
+            break
+    if i == len(position) - 1:
+        # No scanning over this axis
+        scale = 0
+        size = 0
+    else:
+        scale = x - position[0]
+        if i == 1:  # Rastering over this dimension first
```
vidartf: Don't we risk having two axes both thinking they are first? E.g. `[(0,0), (1,1), (1,2), (2,2), (2,3), (3,4)]`?

hyperspy/io_plugins/fei.py
francisco-dlp: Sorry that I forgot to comment on that. If the data is recorded by "uniform" rastering that cannot happen: if you plot the coordinates they should look like a sawtooth (first rastering axis) or staircase (second rastering axis). If it looks like a sawtooth the value repeats every "size" of the other axis. If no rastering occurs over the other axis, then it never repeats.

hyperspy/io_plugins/fei.py
francisco-dlp: Why would that happen when rastering uniformly?

hyperspy/io_plugins/fei.py
francisco-dlp: Again, the assumption is uniformly rastered grid, so that shouldn't happen.

hyperspy/io_plugins/fei.py
vidartf: Oh, so the coordinates are that of the grid spanned by the axes, not the coordinates of the line data?

francisco-dlp: Although I don't know what you mean by "line data", the answer is yes.

vidartf: If so, wouldn't this test be enough?
```python
if len(position) > 2 and position[1] != position[0]:
    # First scan
else
    # Second scan
```

vidartf: Well, you need `i` I guess, so should be OK as it is.

francisco-dlp: Thank you for replying to yourself :)

hyperspy/io_plugins/fei.py
```diff
@@ -316,6 +317,115 @@ def get_xml_info_from_emi(emi_file):
...
+            array_shape.append(axes[-1]["size"])
+            axes.append({
+                'name': "x",
+                'units': "meters",
+                'index_in_array': 1,
+            })
+            axes[-1].update(xcal)
+            array_shape.append(axes[-1]["size"])
+        elif xcal["size"] == ycal["size"] == array_size:
+            # Oblique line scan
```
vidartf: Will this be `True` if its an oblique line scan in a non-symmetric image? I.e. won't `xcal["size"] != ycal["size"]` in that case?

vidartf: Also, isn't a diagonal line the only case where `array_size` will be correct?

francisco-dlp: I haven't actually tested this part of the code because I don't have any oblique line scan recorded like this among my ser files. However, I think that it should be correct. An oblique line scan should look like a sawtooth with just one tooth, regardless of the image being rectangular of square. For the same reason the size should be len(PositionX) == len(PositionY) == array_size

vidartf: Sounds plausible. Would've liked to have tested it, but should be OK.

PR #425, opened 2015-01-14
Rectangular Region of Interest (ROI)

francisco-dlp:
This is a reworked version of #393 (I've merged #393 into it) that aims at separating the ROI and plotting code. The widget is taken from #393 but I have taken a different approach to implementing the ROI that does not rely in modifying `DataAxis` and that should make it easier to create new ROIs.

Note that this branch includes #424, #423, #422 and #434.

Usage
--------

```python
>>> s = signals.Spectrum(np.random.random((10, 20, 100)))
>>> roi = utils.roi.RectangularROI(1,3,3,5)
>>> roi(s)
<Spectrum, title: , dimensions: (2, 2|100)>
>>> roi.right = 10
>>> roi(s)
<Spectrum, title: , dimensions: (7, 2|100)>

```

The integration with the widget is not yet user friendly. The following script shows how to do it, any user friendly syntax ideas?

```python

from hyperspy.drawing.widgets import ResizebleDraggableRectangle


roi = utils.roi.RectangularROI(0, 10, 0, 10)
s = signals.Spectrum(np.random.random((10, 20, 100)))
s.axes_manager[0].scale = 0.5
s.axes_manager[0].scale = 0.1
sroi = utils.interactive(roi.__call__, event=roi.events.roi_changed, signal=s)
w = ResizebleDraggableRectangle(s.axes_manager.deepcopy())
s.plot()
s._plot.pointer.close()
w.add_axes(s._plot.navigator_plot.ax)

def update_roi():
    roi.left = w.axes_manager[0].index
    roi.top = w.axes_manager[1].index
    roi.bottom = w.axes_manager[1].index + w.ysize
    roi.right = w.axes_manager[0].index + w.xsize

w.events.changed.connect(update_roi)
ssum = utils.interactive(sroi, sroi.sum, event=sroi.events.data_changed, axis="navigation")
ssum.plot()
ssum.events.data_changed.connect(ssum._plot.signal_plot.update)

```

TODO
--------

- [ ] unittests.
- [ ] docstrings.
- [ ] User guide documentation.
- [ ] Simple syntax to change the ROI interactively and display the result.
- [ ] Fix bug: the bottom left resizers is slightly out of place.
- [ ] Fix bug: currently it is possible to resize and move the widget beyond the right/bottom axis limits.
- [ ] Fix bug: in the widget top/bottom left/right can currently be reversed i.e. it is possible to set left > right and bottom > top.

PR #434, opened 2015-01-16
New function to ease interactive operations

francisco-dlp:
Usage
--------

```python

>>> s = hs.signals.Spectrum(np.arange(10.))
>>> ssum = hs.interactive(s.sum, axis=0)
>>> ssum.data
array(45.0)
>>> s.data /= 10
>>> s.events.data_changed.trigger()
>>> ssum.data
4.5

```

TODO
--------

- [x] unittests
- [x] User Guide documentation

NOTES
---------

Contains #423, the example is broken due to #850

PR #836, opened 2016-01-19
New widgets

vidartf:
Implemented new and improved widgets loosely based on the previous widget implementation.

The proposed widget implementation is broken up into separate components:
 - Base classes:
   - `BaseWidget`: Implements common widget behavior like managing connections to MPL, managing navigation behavior if widget is used for that (synchronisation with an `AxesManager`), and common MPL code.
   - `DraggableWidgetBase`: Implements common code for making widgets draggable.
   - `ResizableDraggableWidgetBase`: Implements resizing code on top of the dragging code.
   - `Widget2DBase`: Base widget for draggable, resizable 2D types.
   - `ResizersMixin`: A mix-in class for adding resizer manipulator patches to a widget.
 - New/improve widgets:
   - `Square`: Good old square that is there today.
   - `Rectangle`: A freely resizable rectangle widget.
   - `Circle`: A symmetric circle widget. Set up to easily expand use in the future to be non-symmetric, and to have an inner radius, making an annulus.
   - `Range`: Wrapper for an improved `ModifiableSpanSelector`, giving it a common interface with the other widgets.
   - `VerticalLine`/`HorizontalLine`/`Label`/`ScaleBar`: Mostly as before, but with a few new benefits since they inherit the same base (except `ScaleBar`).

 In addition to the new types of widgets, there are some new features on all/most of them:
 - Events implemented, so that changes in the widgets can more easily trigger actions elsewhere.
 - Two settings `snap_position`/`snap_size`, so that widgets can be set to snap to the grid defined by `AxesManager` or not.

This PR should supersede the widget code in #425.

I based the current organization of the widgets in the source tree on that of markers, but not sure if this is how it should be. Any and all constructive suggestions are welcome :)

## TODO
 - [ ] User guide
 - [ ] Docs (auto from docstrings once module layout is agreed)
 - [ ] Unit tests simulating user interaction?


vidartf:
I'll probably end up postfixing all the widget names with `Widget`, as it makes the names more explicit without making them too verbose.

vidartf:
PS: A weak point of the current implementation is the initialization: How should the axes + figure be set. Currently it tries to guess the axes from the dimensions in `axes_manager`, and the figure is set by `set_mpl_axes`. If the automatically guessed axes are wrong, they can be set by writing to the `axes` attribute, but this risks leaving `position` and `size` at invalid values (if offest/scale are different). Proposals to a useful way to solve this are welcome (`axes` setter function? `axes` init argument?).

vidartf:
I fixed the weak point addressed above by an `axes` property with setter. I also fixed some other minor issues.

to266:
I'll review this, #843 and #844 once #423 is merged.

Unless I mixed it all up, and the order should be different, then let me know

vidartf:
This PR #836 depends only on #423.

#434 depends on #423 and #424.

#843 depends on this (#836) and #434, meaning it also depends on #423 and #424.

#844 depends on #843, and therefore all its dependencies as well.

So preferred order is maybe #423 -> [(i) #836 (ii) #424, #434] -> #843 -> #844, where (i) and (ii) can go in parallel. :)

hyperspy/drawing/widget.py
```diff
@@ -0,0 +1,896 @@
...
+        events.closed.
+        """
+        self.set_on(False)
+        self.events.closed.trigger(self)
+
+    def draw_patch(self, *args):
+        """Update the patch drawing.
+        """
+        if hasattr(self.ax, 'hspy_fig'):
+            self.ax.hspy_fig._draw_animated()
```
to266: This gives me an `AttributeError` every time I close a widget - `self.ax.hspy_fig.figure is None`, so the `_draw_animated()` raises it.

to266:
Briefly played with the widgets, looks like:
 - The second image navigator (created when "e" is pressed) does not seem to be connected to the blue line.
 - Having manually replaced `SquareWidget` with `RectangleWidget` to play with a resizeable widget, I noticed that using arrow keys I can move the widget to inaccessible pixels while using mouse (but it's not being redrawn - only the plotted / selected spectrum and the indices indicator on the navigator plot changes). Clicking in the widget snaps it to the drawn position - limited by size of the widget and navigator window.
 - With a second ("e" key) rectangle it seems that while resizers (I assume) correctly only appear on one of them (last active widget), and they seem to be independent when resized with a mouse, all keyboard shortcuts (`y/u`, `x/c`, `+/-`) apply the action on both.

Sadly can't test the first point in #423 - plotting is broken in there at the moment...

vidartf:
- The second navigator just missed a connect call, fix working locally, will push soon.
 - When using navigation to get to an invalid position, the widget retains its position (`widget.position` doesn't change). The widget does not prevent "correct" navigation from happening, i.e. does not put its validated values back to the `AxesManager`. This is by design. What you can do to get a better understanding of how it works, is to not replace the pointer with a rectangle, but rather add a rectangle like this:
```python
from hyperspy.drawing.widgets import RectangleWidget
w = RectangleWidget(s.axes_manager)
w.set_mpl_ax(s._plot.navigator_plot.ax)
w.connect_navigate()
```
Then try to navigate around by the keys. This is basically because the `Rectangle` is not meant to be used for navigation. Do you think this behavior should change? If so, all the inaccessible points will be copies of the right edge points for all calculations that iterate over the axes manager.
 - Ah, that's a good point. I added the "selection" logic after keyboard shortcuts. I'll fix that.

vidartf:
By the way: If you have two navigators, and press '+' on your keyboard, would you expect both of them to become larger? Or just the one that was picked last? I.e. should `SquareWidget` resize for all key events, or should it behave like `RectangleWidget`?

to266:
I'd expect it to resize both if they're square (i.e. "It's too small to pick with a mouse" and they don't perform any operations / data selection of more than one pixel), and only one if their size is more meaningful (actually sets the size of the data chunk, not just a visual thing).

Oh, and one more thing - would it be also possible to implement "key held down" events? Tapping arrow / resizing keys many times is not fun at all.

vidartf:
So currently, it is quite easy to have them always size up if '+' is pressed. However, if a `Rectangle` and a `Square` are both in the figure, the rectangle is selected, and '+' is pressed, should the square also size up? It would be quite hard to have it not do it, because it would be necessary for the Widgets to talk to each other, and consequently, keep track of each other...

vidartf:
Key held down event is not directly supported by MPL. You're free to add such an event yourself, but let's do that as a separate PR :)

to266:
Never mind the +/- issue then - let's say that it's intended - to always resize all widgets. If it's equally difficult (as I assume) to separate u/y/x/c behaviour, then let's just leave it as it is.

vidartf:
x/y/u/c should be no problem, as there are no scenarios where a non-selected widget should use them.

hyperspy/drawing/widget.py
vidartf: When you say "close a widget", what do you mean? That you call `widget.close()`?

to266: Simply close a hyperspy plot that had any widgets

vidartf:
Implemented changes, and merged latest version of #423.

to266:
master has to be merged, since I've merged #423

vidartf:
Master merged, new styles adopted, and `Model1D` fixed up.

vidartf:

hyperspy/drawing/widget.py
```diff
@@ -0,0 +1,966 @@
...
+        """
+        return tuple(self._size.tolist())
+
+    def _set_size(self, value):
+        """Setter for the 'size' property. Calls _size_changed to handle size
+        change, if the value has changed.
+        """
+        value = np.minimum(value, [ax.size * ax.scale for ax in self.axes])
+        value = np.maximum(value,
+                           self.size_step * [ax.scale for ax in self.axes])
```
to266: The multiplication here has to be done inside the list comprehension

hyperspy/drawing/widget.py
```diff
@@ -0,0 +1,966 @@
...
+        self.size = s   # Use property to get full processing
+
+    def get_centre(self):
+        """Get's the center indices. The default implementation is simply the
+        position + half the size in axes space, which should work for any
+        symmetric widget, but more advanced widgets will need to decide whether
+        to return the center of gravity or the geometrical center of the
+        bounds.
+        """
+        return self._pos + self._size() / 2.0
```
to266: No need to call `self._size` here

hyperspy/drawing/_widgets/line2d.py
```diff
@@ -0,0 +1,386 @@
...
+
+    def _validate_pos(self, pos):
+        """Make sure all vertices are within axis bounds.
+        """
+        if np.shape(pos)[1] != len(self.axes):
+            raise ValueError()
+
+        ndim = np.shape(pos)[1]
+        if ndim != len(self.axes):
+            raise ValueError()
```
to266: Repeated code

hyperspy/drawing/_widgets/line2d.py
```diff
@@ -0,0 +1,386 @@
...
+    The priority is in the order resize, rotate, move; so the 'radius_rotate'
+    should always be larger than 'radius_resize' if the function is to be
+    accessible (putting it lower is an easy way to disable the functionality).
+
+
+    NOTE: This widget's internal position does not lock to axes points by
+          default.
+    NOTE: The 'position' is now a 2D tuple: tuple(tuple(x1, x2), tuple(y1, y2))
+    NOTE: The 'size' property corresponds to line width, so it has a len() of
+    only one.
```
to266: Due to this point, `_do_snap_size` should be rewritten for this class - at the moment it assumes `size` has as many dimensions as there are axes

to266:
Mostly things just work here, so I think I'll just merge it, and add any fixes / tests to a separate PR

Issue #838, opened 2016-01-20
Exporting calibrated PCA/BSS loadings


thomasaarholt:
After running Blind Source Separation (BSS) on a spectrum image, it would be great to be able to export the result (in particular the loadings) as spatially calibrated images (i.e. 1pixel = 1nm etc). `.export_bss_results(save_figures = True)` exports individual components, but saves them as picture with scale bars on the X and Y axes. It would be great to be able to save the file as a tiff or hdf5 and retain the calibration for each pixel.

At the moment, the closest I can get is:

```python
a = s.get_bss_loadings()
a.change_dtype("float32") #from float64, which all BSS models are saved as

a.save(“All_BSS_Layered_tiff.tif") #Save the whole thing as one layered tiff
a[0].save("First_BSS_Loading.tif”) # Saves the individual components.
```
Unfortunately, in these two cases the spatial calibration is lost.

to266:
https://github.com/hyperspy/hyperspy/pull/839

You can try this, or, as mentioned in the PR, you can supply the resolutions yourself. The PR'ed thing should be able to convert units correctly as well, however keep in mind I have barely tested it

francisco-dlp:
Is this fixed @thomasaarholt?

thomasaarholt:
Yep, it's fixed! Thanks!

PR #839, opened 2016-01-20
understands units and adds resolution to tif

to266:
https://github.com/hyperspy/hyperspy/issues/838

The (slightly tricky) resolution in tif files.. Experimental, so to say.
As an alternative, without changing the current code the user can supply the resolution directly:

```python
s.save('myfile.tif', resolution = (0.3, (1, 100)))
```
Where `resolution` is a tuple of two elements, where each can be:
 - a float
 - a tuple of two integers, if a rational number is given

### Need:
 - [ ] tests
 - [ ] documentation

### DIFF
--- a/hyperspy/io_plugins/tiff.py
+++ b/hyperspy/io_plugins/tiff.py
@@ -85,6 +85,26 @@ def file_writer(filename, signal, **kwds):
         if signal.metadata.General.title:
             kwds['description'] = signal.metadata.General.title
 
+    # resolution can only be provided in pixels / inch, so have to calculate
+    # the correct values:
+    def get_scale_factor(axis):
+        unit = axis.units
+        from traits.api import Undefined
+        if unit is Undefined or not isinstance(unit, basestring):
+            # does not matter, no units
+            return 0
+        from sympy.physics import units as u
+        unit = getattr(u, unit)
+        scale = axis.scale * unit / u.inch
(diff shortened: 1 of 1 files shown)


thomasaarholt:
Can you briefly describe how `resolution` should be formatted? Is 0.3 your pixel size? Units? And what is the second one?

tjof2:
Perhaps worth noting the discussion about how ImageJ saves its units in TIFF files? http://imagej.1557.x6.nabble.com/Spatial-and-density-calibration-from-TIFF-files-to-ImageJ-td5014677.html

to266:
Sorry for the too-quick reply.

@thomasaarholt , `resolution = (resolution_x, resolution_y)`, where each is in pixels / some_unit. In the official tiff specification those are the only two resolutions that can be specified, and the choice of units is constrained to ('None', 'cm' and 'inch'). However we use a library to write the files, where the unit is constrained to be an inch and cannot be changed.

@tjof2 thanks for the links. From what I gather, my implemented way is constrained, however the only universal / official way? Otherwise we would have to mimic one of the ImageJ plugins / some other software

tjof2:
@to266 I think so - I guess the options are either matching the ImageJ tag, or matching the [relevant tag in the OME-TIFF specification](https://www.openmicroscopy.org/site/support/ome-model/ome-tiff/)

PR #841, opened 2016-01-20
replaced old progressbar.

to266:
Nested not used, possible improvement


hyperspy/_signals/eels.py
```diff
@@ -345,18 +345,16 @@ def estimate_elastic_scattering_intensity(self,
             bk_I0_navigate = (
                 I0.axes_manager._get_axis_attribute_values('navigate'))
             I0.axes_manager.set_signal_dimension(0)
-            pbar = hyperspy.external.progressbar.progressbar(
-                maxval=self.axes_manager.navigation_size,
-            )
-            for i, s in enumerate(self):
+            for i, s in progressbar(enumerate(self),
+                                    total=self.axes_maanger.navigation_size,
```
vidartf: typo `maanger`

vidartf:
At the moment I'm most concerned with if it hooks well with the UI: Currently calculations with long processing times will freeze the UI unless the UI hooks the progressbar to process window events (if any) for every iteration.

to266:
https://github.com/tqdm/tqdm#hooks-and-callbacks

Also, with previous progressbar you would have to supply "iteration number" once it was finished - hence the requirement to hook back with every iteration.

With `tqdm` you supply how many iterations have passed since the last update - hence the freedom to call the `update` whenever you feel like. I guess it does not block either, since supports hooks?

Anyway, I did not try to optimize the usage too well yet - it's likely that it's possible to use the library somewhat better, e.g in nested iterations with nested progressbars.

vidartf:
Sure, that shows how to wrap it, but if the UI was to wrap this, it would wither have to override the internals of tqdm, or wrap each and every instance it is used in hyperspy. That's pretty high maintenance no matter what.

vidartf:
A simple solution would be if there was a single function somewhere in the hyperspy code that gives you the progressbar to use. Then that function can easily be modified by the UI on the run, to something that duck-types/wraps it.

to266:
Hm... I'll try to do that tomorrow. 

I think I understand what you mean, however if you could give a brief example of how it would be best for the UI, feel free. If not, I'll just improvise / look at your repo.

vidartf:
It could be as simple as keeping the existing api:
```python
pbar = hyperspy.external.progressbar.progressbar(max_val=X, <other new kwargs>)
```
And having a new version of that simply return `tqdm()`. However, it might be possible for the UI to simply overwrite `tqdm.tqdm` with a wrapped version. That depends on the internals of `tqdm` (it might break if that is overwritten?).

to266:
Ok, I did as you suggested - I wrapped the `tqdm()` in the original api. I did not change any of the `tqdm` keywords or behaviour, so `max_val` is now `total`, and the updating is different as well....

@francisco-dlp, `tqdm` also has a nice feature to clear anything related to it from the screen. In fact, that's the default behaviour, and I had to manually choose to keep it in all instances to reproduce what we previously had. Now that it's an option, do we really want to keep all the progressbars? I see that it might be somewhat useful, since now it prints interating speed as well, so it can be kept for future references....

francisco-dlp:
I agree that it is useful to keep all the progress bars.

2016-01-21 10:31 GMT+00:00 Tomas Ostasevicius <notifications@github.com>:

francisco-dlp:
@vidartf, are you happy with this?

vidartf:
Sure.

to266:
Updated

francisco-dlp:
This is cool and the code looks fine to me. However, there is an issue with this. Ideally we would like to ask for inclusion of hyperspy in the Debian repositories soon (see #455). However, tqdm is not yet available in Debian.

vidartf:
You could do a check on import whether `tqdm` is available, and use if that is the case. If not, the old code could be used (would require that the old interface is made compatible). However, this might end up complicating the code quite a bit.

to266:
Yes, that would be too complicated and not worth it at all.

I would also argue that `tqdm` not being in debian is not an issue with the PR itself - it does what it clams to do, with the fastest progressbar library available, while also getting rid of the trouble of maintaining our own.

francisco-dlp:
@to266, I agree with your points. However, once we merge this we'll need to wait until tqdm is in the debian repositories (if ever) before hyperspy can stand a chance. I would rather delay merging this and ask the tqdm developers if they are planning to request debian to package it sometime soon. Another option would be to add tqdm to our external folder.

jat255:
+1 on adding tqdm to the external folder and including it as a 3rd party dependency (if they agree).

Also, tqdm is licensed under MIT, so theoretically there is nothing legally wrong about including it directly as a part of HyperSpy (although I am not a lawyer).

magnunor:
Does anyone knows what Debians policy is on including other projects in "external" folders like that?

to266:
Updated with the current master. Are we still concerned with the debian? Should I include the `tqdm` as a submodule, as suggested?

I think `conda` and `pip` are better than having it as a submodule.

francisco-dlp:
Now tqdm is in debian :)

francisco-dlp:
The new requirement have to be added to the user guide and stdeb.cfg.

to266:
Added.

The archlinux part of the install has to be majorly updated (it still refers to python2 packages) if it's possible at all (some packages may only be available from the [AUR](https://wiki.archlinux.org/index.php/Arch_User_Repository))

hyperspy/external/mpfit/mpfit.py
```diff
@@ -1448,7 +1448,11 @@ def defiter(self, fcn, x, iter, fnorm=None, functkw=None,
 
         # Determine which parameters to print
         nprint = len(x)
-        print("Iter ", ('%6i' % iter), "   CHI-SQUARE = ", ('%.10g' % fnorm), " DOF = ", ('%i' % dof))
+        print(
+            "Iter ", ('%6i' %
+                      iter), "   CHI-SQUARE = ", ('%.10g' %
+                                                  fnorm), " DOF = ", ('%i' %
+                                                                      dof))
```
vidartf: Could you tune this one by hand? :)

to266: Sure, should be done

jat255: As for the arch instructions, I can get to it in a couple of weeks, unless
someone wants to do it before then. Maybe comment it out for now and open a
new issue about it so we don't forget?

On Tue, Jun 14, 2016 at 8:53 AM, Tomas Ostasevicius <
notifications@github.com> wrote:

> In hyperspy/external/mpfit/mpfit.py
> <https://github.com/hyperspy/hyperspy/pull/841#discussion_r66963260>:
>
> > @@ -1448,7 +1448,11 @@ def defiter(self, fcn, x, iter, fnorm=None, functkw=None,
> >
> >          # Determine which parameters to print
> >          nprint = len(x)
> > -        print("Iter ", ('%6i' % iter), "   CHI-SQUARE = ", ('%.10g' % fnorm), " DOF = ", ('%i' % dof))
> > +        print(
> > +            "Iter ", ('%6i' %
> > +                      iter), "   CHI-SQUARE = ", ('%.10g' %
> > +                                                  fnorm), " DOF = ", ('%i' %
> > +                                                                      dof))
>
> Sure, should be done
>
> —
> You are receiving this because you commented.
> Reply to this email directly, view it on GitHub
> <https://github.com/hyperspy/hyperspy/pull/841/files/ef250ad01582fe4ffa29eb62e0b60ed12e93d4a9#r66963260>,
> or mute the thread
> <https://github.com/notifications/unsubscribe/ABOBXbmpM0o5FsTf6753PrsRonhQhbdmks5qLqRQgaJpZM4HI3Hl>
> .
>



-- 
Joshua Taillon
Materials Science and Engineering
University of Maryland, College Park
www.joshuataillon.com

doc/user_guide/install.rst
```diff
@@ -221,7 +221,7 @@ In Debian/Ubuntu you can install the libraries as follows:
 
     $ sudo apt-get install python-numpy python-matplotlib ipython
```
francisco-dlp: You have missed adding it to the paragraph above.

to266: thanks, added

PR #393, opened 2014-10-30
ROI interactive plotting

magnunor:
I've done some preliminary work in implementing interactive and resizable region-of-interest tools. Which would be similar to the current way to select a spectrum, but it will be possible to resize it in arbitrary shapes.

PR #843, opened 2016-01-22
Regions of interests (ROIs)

vidartf:
This is yet a reworked version of #425 (which was a rework of #393), but now includes a lot more ROI types. Due to a bit of branch hell, and horrible commit trees with a ton of merges, this has been rebased (soft nuke: `git reset` + recomitted in an orderly fashion). It is dependent on the widgets in #836, and consequently also the events in #423, which contribute the most commits to the current diff.
The currently implemented ROI types are:
 - **Point1D / Point2D:** Picks a point along one or two axes. Indexes rather than slices the signal.
 - **SpanROI:** Select a range along one axis in a one-dimensional plot.
 - **RectangleROI:** Select a range along two axes. Needs an image plot.
 - **CircleROI:** Slices a square (currently symmetric only), and masks off the data that falls outside the circle selection.
 - **Line2DROI:** Extracts a line profile between two points defined in two dimensions. Supports a linewidth. Should be easy to expand to higher dimensions in the future. 

Usage
--------

```python
>>> s = hs.signals.Spectrum(np.random.random((10, 20, 100)))
>>> s.plot()
>>> roi = hs.roi.RectangularROI(1,3,3,5)
>>> roi(s)
<Spectrum, title: , dimensions: (2, 2|100)>
>>> w = roi.add_widget(s)
<hyperspy.drawing._widgets.rectangles.RectangleWidget at 0x28f4a048>
>>> roi.right = 10.
>>> roi(s)
<Spectrum, title: , dimensions: (9, 2|100)>
# Move and resize the widget by hand
>>> roi
RectangularROI(left=2.000000, top=3.000000, right=4.000000, bottom=5.000000
>>> w.snap_all = False   # Makes widget not snap to pixels
# Move and resize the widget by hand
>>> roi
RectangularROI(left=6.881084, top=2.691721, right=11.747314, bottom=6.220271
```

TODO
--------

- [X] unittests (more?)
- [X] docstrings.
- [ ] User guide documentation.


vidartf:
Merged master via merge of updated #836.

vidartf:

to266:
Thanks for promptness, @vidartf. Will you have a moment to fix anything small (given there is anything to fix), or should I create a new PR with those, if there are any?

vidartf:
You can always PR them to my branch, and I'll review them before merging into this PR?

to266:
Sure! Was not sure how busy you were :)

vidartf:
Quite busy, but reviewing code is not too time consuming.

hyperspy/roi.py
```diff
@@ -0,0 +1,989 @@
...
+        return widget
+
+    def _remove_widget(self, widget):
+        widget.events.closed.disconnect(self._remove_widget)
+        widget.events.changed.disconnect(self._on_widget_change)
+        widget.close()
+        for signal, w in self.signal_map.iteritems():
+            if w == widget:
+                self.signal_map.pop(signal)
+                break
```
to266: What's the point of this for loop? Not only it looks like it won't work (`w` is a tuple of `(widget, axes)`), but also it only pops the first signal that is still left in `signal_map` and has the same widget. Is it just a leftover and is only meant to close / disconnect?

vidartf: This code is quite dated, yes. Main purpose it to update `signal_map`, to remove the entry with `widget` (if present). Having one widget correspond to several entries is Bad, so it should only need to find one entry.

to266:
Please see https://github.com/vidartf/hyperspy/pull/20

vidartf:
PS: I realized there is no tests for checking annulus ROI (Circle ROI with inner radius set). Not sure if it works as it should.

to266:
I will PR them then. When I played with it yesterday, it looked fine except then inner radius was larger than the outer (now fixed and should not be possible)

I was not a fan of the rounding when calculating mask when radius was close (or equal) to %.5 (from the square of data, one side would be masked and the other - unmasked) , but it makes sense.

to266:
https://github.com/vidartf/hyperspy/pull/21

to266:
Merging this now, as I think it's better to update the user-guide in #844

PR #844, opened 2016-01-22
ROI interactive output

vidartf:
Adds interactive output functionality to the ROIs, allowing for the output signal to be continuously updated as the ROI is modified. The default behavior of the function is to add a widget to the source signal representing the ROI, but this can be changed by an argument.

This PR is based on #843, but is also dependent on #434, and consequently #424.

Usage 
--------
```python
>>> s = hs.signals.Spectrum(np.arange(20000).reshape(((10, 20, 100))))
>>> s.plot()
>>> roi = hs.roi.RectangularROI(1,3,3,5)
>>> sr = roi.interactive(s)  # Also adds a widget to `s`
>>> ss = hs.interactive(sr.sum, sr.events.data_changed)
>>> ss.plot()
```

Todo
------
 - [x] User guide
 - [x] I suspect the point ROI classes have the same problem that `Line2DROI` had w.r.t. signal class when ROI is applied in signal space.
 - [x] Line2DROI currently makes some mistakes when applied to signal space (axes are handled incorrectly).
 - [ ] Not all events are disconnected as appropriate.


to266:
@vidartf https://github.com/vidartf/hyperspy/pull/22

I'll come up with something for user-guide after lunch

vidartf:
Remaining issues:
 - [ ] I suspect the point ROI classes have the same problem that `Line2DROI` had w.r.t. signal class when ROI is applied in signal space.
 - [ ] Line2DROI currently makes some mistakes when applied to signal space (axes are handled incorrectly).
 - [ ] Not all events are disconnected as appropriate.

dnjohnstone:
Some "features" of ROI usage that I'm stumbling across at the moment and are surprising me:

1. Line2DROI doesn't work as I'd expect in that I can't get a line trace out even though Circle and Rectangle work as I'd expect. The return has the right dimensions e.g. (12 | 55) but is full of zeros. @to266 knows about this and I think is looking in to it.

2. CircleROI has limited size, or is not using axes correctly? I tried to use it to crop navigation dimensions and I wanted a circle of about 250 scaled dimensions radius but couldn't get anywhere near that, it seemed to stop at 120 which perhaps coincidentally or maybe tellingly is the size in unscaled pixels.

vidartf:
@dnjohnstone @to266 : I have a fix for 1 in the making, so don't spend time on it.

vidartf:
@dnjohnstone : All ROI coordinates are in value space. Not sure what you mean by 'unscaled pixels'. Can you post a mock example?

vidartf:
@to266 : I've pushed a fix for 1, but I see `VerticalLineWidget` cannot be used for navigation. Any clues?

dnjohnstone:
It's as simple as I can't make the circle bigger than shown in the image below. It won't drag bigger and it won't let me specify it bigger than r=120.

In value space the navigation dimensions are about 570 x 570 and by unscaled I mean that the actual navigation dimension is (120, 120 | 144, 144) which is why I think that might be causing the limit. I expected it to be in value space....

max_circle

dnjohnstone:
Incidentally when I specify r > 120 it doesn't raise an error, just doesn't plot anything.

vidartf:
@dnjohnstone : The last point is as exected, as the ROI doesn't enforce boundaries (but the widget does). The problem is therefore with the `CircleWidget`.

vidartf:
Fix pushed. The bounds check of `CircleWidget` didn't use the scale.

vidartf:
Another issue: `CircleWidget` does not validate position when resizing, potentially allowing it to go out of bounds.

to266:
@vidartf as for the `VerticalLineWidget`, I've tested it with 
```python
import scipy.misc
im = hs.signals.Image(scipy.misc.ascent())
ims = hs.stack([im]*15)
ims.plot()
```

Looks like somehow the blue navigator line (i.e. the "sum signal") gets a new x array, which accidentally corresponds to signal_plot x axis (i.e. in this case `np.arange(512)` instead of `np.arange(15)`)

hyperspy/drawing/spectrum.py
```diff
@@ -331,7 +331,16 @@ def update(self, force_replot=False):
             ydata = self.data_function(axes_manager=self.axes_manager).real
         else:
             ydata = self.data_function(axes_manager=self.axes_manager).imag
-        self.line.set_ydata(ydata)
+
+        new_axis = self.axes_manager.signal_axes[0].axis
```
vidartf: This line currently breaks navigation, as the axis plotted isn't always `signal_axes[0]`, e.g. when the Spectrum plot is for the navigation plot.

vidartf: Anyways, this code seems like a bit of a hack around a bigger issue: Should `update()` calls refresh the axes of the plot? From MPL's point of view it is possible, but the way HE/HSE/HIE is designed doesn't really support this.

to266: Nevertheless, https://github.com/vidartf/hyperspy/pull/23

vidartf:
Only big issue now would be to check whether the `PointROI` classes also need to update metadata when removing axes in signal space (I assume so), and fix it if that's the case.

vidartf:
I just realized that I never added code for having `CircleWidget` draw its inner radius, so I added that now. I also tried to fix metadata for `PointROI` classes, but have not tested extensively. can now be merged unless review finds any issues.

dnjohnstone:
Having an events related error, which may or may not be resolved by this... When I make a model to some data from PCA results and then try to plot it, I get the error below. This is also stopping the X-ray lines etc being shown because it doesn't get that far. Is this something that this PR with interactive resolves? Would be good to resolve quickly for anyone trying to use the current master to do EDS analysis.

events_error_model_plot

dnjohnstone:
In fact - scratch that, seems to have been resolved by restarting the notebook, must have been in a weird state somehow.... will let you know if this comes back

vidartf:
@dnjohnstone Sure, but there is definitely some issues remaining with ensuring events are disconnected as they should. Here I think the issue might be if you call `plot()` without explicitly closing the plot first. Internally, `plot()` closes the previous plot if present, but I think I might have forgotten about how close events are handled in MPL when I redid the `_on_close` code of plots on master. I.e. the disconnect and connect calls can end up coming in the wrong order (partly random depending on Qt event queue threading).

I think maybe we should implement logging in events code++ at debug level (or a lower "trace" level), which can then be turned on when such issues as this come up. C.f. #749.

to266:
If a widget is added for `CircleROI` when `r_inner < 0`, and then changed, it bugs out (since the patches are not present). The same happens in reverse (i.e. if `r_inner` becomes < 0, and the inner patch is no longer necessary)

vidartf:
First of all, should we fix it so that `r_inner` (and `r`?) cannot be < 0? For the problem at hand, I think this would need a check in `set_size()` which triggers patch creation/removal. Simplest way would be to call `set_on(False)`, `set_on(True)`, but that will cause it to flicker. Alternatively resizer patch toggle should be extracted to a protected function which `set_size()` can call.

dnjohnstone:
@vidartf Yes, I think you're right. It's something to do with the order in which I do things. Just working a bit more it crops up every now and then and is usually resolved by closing things and/or going back a couple of cells in the notebook. Quite livable but does feel a bit buggy.

dnjohnstone:
@vidartf where are we at with this one? Looks like you've done what was needed with PointROI so is it just the inner radius behavior for CircleROI?

On which point - with CircleROI in the current master I've been noticing that printing the parameters for my ROIs gives small -ve inner radii although I'd have expected it to be zero.

vidartf:
I think I should have added both PointROI fixes and inner radius for CircleWidget. Getting negative radius is definitely not intended: Does this happen for pure ROIs, or only when connected to a widget?

dnjohnstone:
As far as I can work out the negative inner radius arises when I specify a CircleROI with a widget and then resize it interactively. This seems to always result in r_inner = -0.0038 in value units which of course is specific to the plot I'm working with.

Other than that - this branch seems to be working very nicely for me now. I don't get the plot errors that I get using the master when using this one.

It's really useful! Thanks @vidartf

vidartf:
Guessing blindly, is the inner radius equal to negative of half the axis scale?

dnjohnstone:
An excellent guess sir - it is exactly that

vidartf:
Pushed a fix. Can you confirm that this fixes it?

dnjohnstone:
Thanks @vidartf that fixed that - as far as my usage is going, this branch is now working very nicely and I think solves a number of minor bugs in the current master.

Is there anything remaining before this can be merged?

vidartf:
There shouldn't be any known issues left here now as far as I know.

to266:
The only thing left is what I mentioned before - creating the inner circle patch once it is "turned on". Also there is the question of removing the inner patch - should it only happen with command-line, or should just moving the resizers to the centre be enough? @dnjohnstone, you have used the ROI more than I have, what would you consider a convenient and predictable behavior? 

Maybe having the inner circle all the time is useful too - the patch / resizers act as a visual reference for the centre..

vidartf:
My thoughts when creating it was that you chose the type (with or without inner radius) at creation, and wouldn't have a need to change back and forth. However, I agree that it is strange for this only to be true for the widget.

dnjohnstone:
From a usage perspective I think not being able to set r_inner < 0 and selecting whether you have an inner radius at the outset is everything that I would want and makes most sense to me. 

I think setting with or without inner radius at creation makes sense - I think a circle vs a ring is a sufficiently different thing that someone will have decided which they want at the outset. It would be more annoying to unintentionally create an inner radius when they don't want it and not being totally sure whether they got it back to zero or vice versa lose the inner resizers somehow.

It seems then that the only issue is whether we're happy to have a general CircleROI that can behave as a circle or a ring depending on widget usage, which is admittedly is odd in some sense  - on the other had it's somehow code efficient and at the point of use kind of clear from the command line as you know whether you specified an inner radius or not... So I'm ambivalent about it and would probably leave that as is, but I'm quite mercenary on these issues.

vidartf:
> It seems then that the only issue is whether we're happy to have a general CircleROI that can behave as a circle or a ring depending on widget usage, which is admittedly is odd in some sense

As it is now: When defining the CircleROI the `r_inner` argument is optional. Unless supplied at creation, or set manually afterwards, it won't add the inner resizers when creating the widget. I see the argument for retroactively updating widgets if `r_inner` is specified on the ROI later, as the widgets should correctly represent the ROI. If not, we should disallow setting `r_inner` on the ROI after creation. If allowed, the issue is whether widgets should be able to switch between the two modes: Currently the ROI differentiates between not having an inner radius (`r_inner` is `Undefined`) and it being 0, the widget does not have this ability but simply doesn't create the resizers if `r_inner` is 0 at creation.

Note that currently it isn't possible to unset `r_inner` to `Undefined` on the ROI once defined (unless you set `ROI._bounds_check` to `False`).

to266:
I'd vote for updating widgets.

dnjohnstone:
Updating widgets is probably the more elegant solution.

Just come across a strangeness with Line2D... When specified with some finite linewidth the bounding lines initially plot in a really weird place as below. Dragging the widget updates the plot to the correct thing though.

selection_001

dnjohnstone:
Sorry to be pushy on this one, but it seems to me that the remaining issues are relatively minor i.e. don't affect major functionality and are largely elegance issues - on the other hand without this merge I get a load of plotting errors to do with events when plotting x-ray lines / background windows.

Could we consider merging up so that at least plotting works properly in the master?

vidartf:
As far as I can tell there are two issues remaining, the `LineROI` thing @dnjohnstone pointed out, and getting `CircleROI` to update its widget to reflect changes to `r_inner`. Neither strictly belong in this PR, which is about adding an `interactive()` convenience function to ROIs. I suggest merging, and creating one or two new issues / PRs to master.

dnjohnstone:
@vidartf +1 on that

PR #845, opened 2016-01-22
Workaround for tuple axis arg for masked arrays

vidartf:
While numpy's methods `mean`, `sum` etc. accept tuples for their `axis` argument, `numpy.ma.MaskedArray` does not (yet?). So for that reason, I propose this workaround for those data types.

Note, the function signature includes the `out` keyword in order to be complete w.r.t that PR. This workaround should therefore be considered blocked by that PR.

### DIFF
--- a/hyperspy/signal.py
+++ b/hyperspy/signal.py
@@ -3874,13 +3874,20 @@ class Signal(FancySlicing,
 
     def _ma_workaround(self, s, function, axes, ar_axes, out):
         # TODO: Remove if and when numpy.ma accepts tuple `axis`
-        intermediate = self
-        for a in ar_axes[:-1]:
-            intermediate = function(intermediate.data, axis=a)
+
+        # Basically perform unfolding, but only on data. We don't care about
+        # the axes since the function will consume it/them.
+        ar_axes = sorted(ar_axes)
+        new_shape = list(self.data.shape)
+        for index in ar_axes[1:]:
+            new_shape[index] = 1
+        new_shape[ar_axes[0]] = -1
+        data = self.data.reshape(new_shape).squeeze()
+
         if out:
-            function(intermediate.data, axis=ar_axes[-1], out=out.data)
+            function(data, axis=ar_axes[0], out=out.data)
         else:
-            s.data = function(intermediate.data, axis=ar_axes[-1])
+            s.data = function(data, axis=ar_axes[0])
             s._remove_axis([ax.index_in_axes_manager for ax in axes])
             return s
 
--- a/hyperspy/signal.py
+++ b/hyperspy/signal.py
@@ -3860,6 +3860,19 @@ class Signal(FancySlicing,
                 name="Scalar",
                 navigate=False,)
 
+    def _ma_workaround(self, s, function, axes, ar_axes, out):
+        # TODO: Remove if and when numpy.ma accepts tuple `axis`
+        intermediate = self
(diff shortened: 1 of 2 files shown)


vidartf:
With #424 merged, this is now ready for review.

to266:
At least one test?

hyperspy/signal.py
```diff
@@ -3872,6 +3872,18 @@ def _remove_axis(self, axes):
                 name="Scalar",
                 navigate=False,)
 
+    def _ma_workaround(self, s, function, axes, ar_axes, out):
+        # TODO: Remove if and when numpy.ma accepts tuple `axis`
+        intermediate = self
+        for a in ar_axes[:-1]:
```
to266: maybe `reversed`? Haven't tested it though..

vidartf:
Good point, since it doesn't work :)

vidartf:
PS: When running tests locally, I get the following error:
```python
======================================================================
ERROR: hyperspy.tests.signal.test_tools.TestOutArg.test_indexmax
  File "C:\prog\WinPython-64bit-2.7.10.2\python-2.7.10.amd64\lib\site-packages\nose\case.py", line 197, in runTest
    self.test(*self.arg)
  File "c:\github\hyperspy\hyperspy\tests\signal\test_tools.py", line 473, in test_indexmax
    self._run_single(self.s.indexmax, self.s, dict(axis=0))
  File "c:\github\hyperspy\hyperspy\tests\signal\test_tools.py", line 435, in _run_single
    r = f(out=s1, **kwargs)
  File "c:\github\hyperspy\hyperspy\signal.py", line 4288, in indexmax
    out=out)
  File "c:\github\hyperspy\hyperspy\signal.py", line 3908, in _apply_function_on_data_and_remove_axis
    function(self.data, axis=ar_axes, out=out.data)
TypeError: argmax() got an unexpected keyword argument 'out'
```

Seems to be that `argmax` only got the `out` arg in numpy 1.10, and I'm still at 1.9. Just a heads up!

vidartf:
Tests have been introduced and behavior fixed, so should be ready for review. It currently only tests for `sum` and `mean`. I *think* it should work for the others as well, but didn't know how to test for those without reproducing the implementation in the test code.

hyperspy/tests/signal/test_tools.py
```diff
@@ -528,3 +528,25 @@ def test_histogram_axis_changes(self):
         assert_array_equal(h1.data, h2.data)
         nt.assert_equal(h1.axes_manager[-1].size,
                         h2.axes_manager[-1].size,)
+
+    def test_masked_array_mean(self):
+        s = self.s
+        mask = (s.data > 0.5)
+        s.data = np.ones_like(s.data)
```
to266: I would rather you used something like `np.arange(s.data.size).reshape(s.data.shape)`. When all elements in an array are 1, testing if a particular subset has a mean of 1 is not very indicative that the subset was selected correctly...

vidartf: Implemented.

to266: Tests fail though, at least in travis

vidartf: Hah, I forgot that the mask is based on random data.

vidartf: Added a fixed seed, and updated the reference data, so now it should be OK.

PR #848, opened 2016-01-27
FIX elastic_scattering_intensity indecing

vidartf:
Fixed some indexing issues in `estimate_elastic_scattering_intensity()`. This PR includes #847.


vidartf:
If a clean diff is needed, simply see eab0a614b5493c76d70fd3be1e34fa27814bc83e. The first commit is already merged (#847), and the last one is just code style.

PR #849, opened 2016-01-28
Switch to setuptools

vidartf:
Allows to build to wheel by `python setup.py bdist_wheel`.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -17,7 +17,7 @@
 # along with  HyperSpy.  If not, see <http://www.gnu.org/licenses/>.
 
 
-from distutils.core import setup
+from setuptools import setup
 
 import distutils.dir_util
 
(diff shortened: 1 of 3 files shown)


francisco-dlp:
setuptools should be added to the dependencies then.

vidartf:
Does it make sense to add setuptools to the `setup_requires` as well?

francisco-dlp:
Although it is a chicken and egg problem, I think it does. For example, we could specify the minimum version supported.

vidartf:
This should be all, right?

setup.py
```diff
@@ -170,6 +172,9 @@ def __exit__(self, type, value, traceback):
                   'hyperspy.external.astroML',
                   ],
         requires=install_req,
+        setup_requires=[
+            'setuptools'
```
jakirkham: I don't understand these lines. How can `setuptools` be in `setup_requires`? Basically this field is only read if `setuptools` exists. So, this is either a no-op or a problem waiting to happen. So, why was this added then?

PR #852, opened 2016-01-31
Make sure Event.trigger iterates on copies

vidartf:
Event callbacks may modify connected functions, so we have to iterate on a copy of the collections of connected functions. The copies are taken before any callbacks are called, to ensure that all callbacks connected at the time the event triggered is called.


vidartf:
This issue currently blocks #836, and all its dependencies.

PR #853, opened 2016-01-31
Make markers for xray lines on plot have subscripts

vidartf:
Just makes plot look better. Depends on `LabelWidget` code in #836  for string setter for the model's names to be updated.

Difference in style:

**EDSSpectrum.plot(xray_lines=True):**
Before:
scr2
After
scr1

**EDSModel.enable_adjust_position():**
Before:
scr4
After:
scr3


vidartf:
Merged master, so diff is now manageable.

vidartf:
Anyone have any input on this?

francisco-dlp:
Sorry that we forgot to label this PR when you submitted it.

Looks good! I would just add a little test that simply makes sure that the new code doesn't raise an error when executed or (even better) a proper test if you merge #1200 here.

francisco-dlp:
This branch was automatically closed when we renamed `master` `RELEASE_next_major` and it cannot be reopened. Could you resubmitted it to RELEASE_next_patch instead?

PR #858, opened 2016-02-08
Limit SpanSelectorInSpectrum to the spectrum range in RemoveBackground.

### LINKED ISSUE #726: remove_background dialogue box disappearing
Sometimes when using the s.remove_background function with EELS spectra , the dialogue box disappears when moving the span selector. I have not tried with other kinds of spectrum types.

I can only make it disappear when using the power law background. Gaussian, offset or polynomial does not seem to be affected.

I'm not really sure exactly what triggers the dialogue box to disappear, but some of the GUI wizards might be able to fix this :)

francisco-dlp:
This fixes the exceptions that were printed when the span selector was moved out of the axis range.

This is an unsuccessful attempt to fix #726.

### DIFF
--- a/hyperspy/drawing/widgets.py
+++ b/hyperspy/drawing/widgets.py
@@ -453,7 +453,7 @@ def in_interval(number, interval):
 
 class ModifiableSpanSelector(matplotlib.widgets.SpanSelector):
 
-    def __init__(self, ax, **kwargs):
+    def __init__(self, ax, left=None, right=None, **kwargs):
         onsel = kwargs.pop('onselect', self.dummy)
         matplotlib.widgets.SpanSelector.__init__(
             self, ax, onsel, direction='horizontal', useblit=False, **kwargs)
@@ -461,6 +461,8 @@ class ModifiableSpanSelector(matplotlib.widgets.SpanSelector):
         self.tolerance = 1
         self.on_move_cid = None
         self.range = None
+        self.left = left
+        self.right = right
 
     def dummy(self, *args, **kwargs):
         pass
@@ -525,10 +527,13 @@ class ModifiableSpanSelector(matplotlib.widgets.SpanSelector):
                       self.rect.get_x() + self.rect.get_width())
 
     def move_left(self, event):
-        if self.buttonDown is False or self.ignore(event):
+        if self.buttonDown is False or self.ignore(event) or\
+                event.xdata is None:
             return
         # Do not move the left edge beyond the right one.
         if event.xdata >= self.range[1]:
+            return
+        if self.left and event.xdata < self.left:
             return
         width_increment = self.range[0] - event.xdata
         self.rect.set_x(event.xdata)
@@ -539,10 +544,13 @@ class ModifiableSpanSelector(matplotlib.widgets.SpanSelector):
         self.update()
 
     def move_right(self, event):
-        if self.buttonDown is False or self.ignore(event):
+        if self.buttonDown is False or self.ignore(event) or\
+                event.xdata is None:
             return
         # Do not move the right edge beyond the left one.
         if event.xdata <= self.range[0]:
+            return
+        if self.right and event.xdata > self.right:
             return
         width_increment = \
             event.xdata - self.range[1]
@@ -553,10 +561,17 @@ class ModifiableSpanSelector(matplotlib.widgets.SpanSelector):
         self.update()
 
     def move_rect(self, event):
-        if self.buttonDown is False or self.ignore(event):
+        if self.buttonDown is False or self.ignore(event) or\
+                event.xdata is None:
             return
         x_increment = event.xdata - self.pressv
-        self.rect.set_x(self.rect.get_x() + x_increment)
+        new_left_x = self.rect.get_x() + x_increment
+        new_right_x = new_left_x + self.rect.get_width()
+        if self.left and new_left_x < self.left:
+            return
+        if self.right and new_right_x > self.right:
+            return
+        self.rect.set_x(new_left_x)
         self.update_range()
         self.pressv = event.xdata
         if self.onmove_callback is not None:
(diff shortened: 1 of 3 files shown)


francisco-dlp:
Could somebody consider merging this?

hyperspy/drawing/widgets.py
```diff
@@ -461,6 +461,8 @@ def __init__(self, ax, **kwargs):
         self.tolerance = 1
         self.on_move_cid = None
         self.range = None
+        self.left = left
+        self.right = right
```
vidartf: Consider giving more explicit names, e.g. `left_limit`/`right_limit`. As they're kwargs as well, please consider making a docstring.

hyperspy/drawing/widgets.py
```diff
@@ -525,11 +527,14 @@ def update_range(self):
...
 
     def move_left(self, event):
-        if self.buttonDown is False or self.ignore(event):
+        if self.buttonDown is False or self.ignore(event) or\
+                event.xdata is None:
             return
         # Do not move the left edge beyond the right one.
         if event.xdata >= self.range[1]:
             return
+        if self.left and event.xdata < self.left:
```
vidartf: This need to test explicitly for `self.left is None`, or it will fail if its 0. Same for right etc. below.

hyperspy/drawing/widgets.py
```diff
@@ -525,11 +527,14 @@ def update_range(self):
...
+            return
```
vidartf: In general, I would make a function to bound the xdata:
```python
def _validate_pos(self, x):
    if self.left is not None:
        x = np.max(x, self.left)
    if self.right is not None:
        x = np.min(x, self.right)
    return x
```
This then sets x to the appropriate limit if it is out of bounds. This is better since we're never guaranteed to have an event for each pixel moved. To save some events, it might also benefit from checking whether the value actually changes.

hyperspy/gui/egerton_quantification.py
```diff
@@ -145,7 +145,8 @@ def bg_to_plot(self, axes_manager=None, fill_with=np.nan):
         return to_return
 
     def span_selector_changed(self):
-        if (self.ss_left_value is np.nan) or (self.ss_right_value is np.nan):
+        if self.ss_left_value is np.nan or self.ss_right_value is np.nan or\
+                self.ss_right_value <= self.ss_left_value:
```
vidartf: Was this actually an issue, or added for redundancy?

hyperspy/drawing/widgets.py
francisco-dlp: Good point. I've fixed it without adding the extra method.

hyperspy/drawing/widgets.py
```diff
@@ -553,10 +573,17 @@ def move_right(self, event):
...
+                event.xdata is None:
             return
         x_increment = event.xdata - self.pressv
-        self.rect.set_x(self.rect.get_x() + x_increment)
+        new_left_x = self.rect.get_x() + x_increment
+        new_right_x = new_left_x + self.rect.get_width()
+        if self.left_limit and new_left_x < self.left_limit:
+            return
+        if self.right_limit and new_right_x > self.right_limit:
+            return
```
vidartf: These guys also need `is not None`.

hyperspy/gui/egerton_quantification.py
francisco-dlp: I can't remember, but I think that it doesn't hurt to leave it there, isn't it?

vidartf: Won't hurt, but if this was actually an issue, this is the wrong place to fix it :)

vidartf: To clarify: If this was an issue triggered by the widget, this would be the wrong place. If it was because of something else it is appropriate.

francisco-dlp: I see, but given that I don't remember why I put it there we can:

i) remove it and hope for the best
ii) Leave it where it is

If it is not broken don't touch it, so I would go for ii.

francisco-dlp:
Thank you @vidartf, everything should be fixed now.

hyperspy/gui/egerton_quantification.py
vidartf: I agree with ii, just wanted to clarify why I was asking :)

vidartf:
Everything seems OK from diff. Testing by hand, it is a bit tricky to put the signal range all the way to the edges. Normally you would just try to drag it past the edge, which I guess is when `xdata` is `None`? It's especially tricky if there are several plot points per screen pixel. A possible solution is to set `x` to the limit when dragging a single edge and `xdata` is `None`?

francisco-dlp:
I agree that it is in issue. A workaround is moving the cursor slowly so that the method that update the widget doesn't miss any position on the path. The solution that you propose sounds good but, given that xdata is None when the cursor is out of the axes, wouldn't it cause the slider to jump to the limits when e.g. moving out from the top or bottom of the axes?

vidartf:
> A workaround is moving the cursor slowly so that the method that update the widget doesn't miss any position on the path.

Except if there are several datapoints per screen pixel (e.g. an axes with `size` of 10000), then you cannot manually move it to the edge I think (if it picks the left position, right edge will be wrong and vice-versa, and if it picks central position both will be wrong).

> wouldn't it cause the slider to jump to the limits when e.g. moving out from the top or bottom of the axes?

Yes, I just considered this the lesser of two evils, as then there is at least a way to achieve both goals.

vidartf:
Will you address the issue discussed in the last comment, or do you argue for merging as is?

francisco-dlp:
I'll experiment with your solution and I'll let you know.

francisco-dlp:
I've implemented your solution in the latest commit. It's actually a very nice shortcut to expand the region all the way to the sides by leaving the axes by the top or bottom :)

to266:
That's nice, I like the behaviour. I'll merge it in ~20 minutes if @vidartf does not do it before then (in order to try to release today...)

vidartf:
I'm looking at it, and think I might PR some suggestions to @francisco-dlp, so hold the merging for now please.

to266:
Sure

PR #864, opened 2016-02-10
New component: Gaussian2Dsymmetric, and add rotation to Gaussian2D

magnunor:
Basically the Gaussian2D component with a rotation parameter.

I wasn't sure if I should improve Gaussian2D, or make a new component. I'm guessing having a function with the rotation is more computationally expensive, so it makes sense to keep them separate.

Based on this: https://github.com/magnunor/hyperspy/blob/4388396695bc17fe118665e818102b37356d4b72/hyperspy/_2d_components/gaussian2d.py

TODO:
- [x] Update docstring
- [x] User guide
- [x] Test


francisco-dlp:
Unless adding the rotation parameter turns out to be much slower than the current Gaussian2D, I would suggest to add the feature to the current component. Could you benchmark them and make a decision?

magnunor:
Using the following benchmarking: 
```python
import timeit
import hyperspy.api as hs

g = hs.model.components.Gaussian2D(
        A=123., centre_x=321., centre_y=213., sigma_x=12., sigma_y=14.)
gr = hs.model.components.Gaussian2Drot(
        A=123., centre_x=321., centre_y=213., sigma_x=12., sigma_y=14., rotation=0.5)

def g_function():
    g.function(0.1,0.1)

def gr_function():
    gr.function(0.1,0.1)

g_function()
gr_function()

print("Gaussian 2D function, without rotation")
print(min(timeit.Timer(g_function).repeat(100,10000)))

print("Gaussian 2D function, with rotation")
print(min(timeit.Timer(gr_function).repeat(100,10000)))
```

I get the result: 
Gaussian 2D function, without rotation
0.0553669929504
Gaussian 2D function, with rotation
0.0786948204041

Which I guess is fast enough?

francisco-dlp:
That's 30% longer. Could you benchmark it also for the case when x and
y are 2D arrays of coordinates?

magnunor:
Using 100x100 x- and y-arrays:

Gaussian 2D function, without rotation
0.302659034729
Gaussian 2D function, with rotation
0.297930002213

Using 1000x1000 x- and y-arrays:

Gaussian 2D function, without rotation
3.7607831955
Gaussian 2D function, with rotation
4.61684298515

francisco-dlp:
How can the version with rotation be faster in the 100x100 case?

francisco-dlp:
In any case, it seems that the rotation introduces a 30% performance penalty. This is a drawback in the symmetric case i.e. `sx == sy`. However, when `sx != sy` the `Gaussian2D` is possibly useless without rotation. So, what about splitting `Gaussian2D` in two: `SymmetricGaussian2D` and `Gaussian2D`, with `Gaussian2D` including the `theta` parameter?

vidartf:
Couldn't the function shortcut to the simple symmetric gaussian if `theta == 0 or sx == sy`? The extra comparison operation shouldn't add that much to the computation time.

magnunor:
Out of interest, because the runtime was a bit strange, I reran for some different sizes (2,4,8,16,...,512) of the 2D x- and y-arrays: 
gaussian2d_rotation_runtime

Array size ** 2 = the total number of elements in the x- and y- arrays. I ran each 1000 times, in 100 repetitions, with the fastest of the repetitions being the "Minimum runtime" shown here.

Anywho: I think implementing two different components: symmetric 2D gaussian and a more general 2D gaussian with rotation will be the best. Having two separate functions will make the implementation a little bit "cleaner" and make it easier to optimize them individually.

francisco-dlp:
Good points @magnunor and @vidartf. Once we get `Expression2D`, I think that we should provide an alternative version of all components, keeping the current ones as a fallback for when sympy is not installed. I think that It'll be more straightforward to do this if we keep the two gaussian2D components separated.

magnunor:
I have been 2D Gaussians for fitting atomic resolution STEM images (I'll open source this at some point, when I got the time...), some which can contain 1000+ atoms. So keeping these functions fairly fast would be useful. 

In addition, if the symmetric 2D gaussian is much quicker, it would be useful to have a utility function which can return an instance of 2D gaussian, with the parameter values of the symmetric one as initial values for the non-symmetric. This could possible save time when fitting a very large amount of atoms: first fit the faster symmetric gaussian, then the non-symmetric. But this functionality should be for some future pull request.

magnunor:
Ready for review now. I'm guessing it is fairly minor changes, so would be nice to get it in for 0.8.3.

magnunor:
I updated this branch to Python3. Anyone got time to review it?

francisco-dlp:
I think that this is fine, except for the tests. Testing that fitting a gaussian component to an image that has been generated using the same component won't detect a broken gaussian function. The tests should test that the data that the component generates is actually what it should be. For example, it that the area corresponds to the right area, etc. Alternatively, I would suggest to simply compare the output with an output that is known to be correct. Ideally, it should be small so that you can simply type it in the file.

hyperspy/_components/gaussian2d.py
```diff
@@ -69,11 +74,25 @@ def function(self, x, y):
         sy = self.sigma_y.value
         x0 = self.centre_x.value
         y0 = self.centre_y.value
+        theta = self.rotation.value
+
+        sx2 = sx**2
+        sy2 = sy**2
+        cos2_theta = math.cos(theta)**2
+        sin2_theta = math.sin(theta)**2
+        sin_theta2 = math.sin(2*theta)
```
francisco-dlp: maybe `sin_2theta` instead?

hyperspy/_components/gaussian2d.py
```diff
@@ -48,18 +50,21 @@ def __init__(self,
                  sigma_y=1.,
                  centre_x=0.,
                  centre_y=0.,
+                 rotation=0.,
```
francisco-dlp: `__init__` needs a docstring. The dosctring should explain that `rotation` is given in mrad.

hyperspy/_components/gaussian2d.py
```diff
@@ -69,11 +74,25 @@ def function(self, x, y):
...
+        a = cos2_theta/(2*sx2) + sin2_theta/(2*sy2)
+        b = -sin_theta2/(4*sx2) + sin_theta2/(4*sy2)
+        c = sin2_theta/(2*sx2) + cos2_theta/(2*sy2)
+
+        return A * (1 / (sx * sy * pi2)) * np.exp(-(a*(x - x0) ** 2 +
+                                                  2*b*(x - x0) * (y - y0) +
+                                                  c*(y - y0) ** 2))
+
+    @property
+    def ellipticity(self):
```
francisco-dlp: Needs documentation is class docstring.

hyperspy/_components/gaussian2d.py
```diff
@@ -90,3 +109,12 @@ def fwhm_y(self):
     @fwhm_y.setter
     def fwhm_y(self, value):
         self.sigma_y.value = value / sigma2fwhm
+
+    # Rotation compared to "x"-axis
+    @property
+    def rotation_degrees(self):
```
francisco-dlp: Needs documentation is class docstring.

hyperspy/_components/gaussian2d.py
```diff
@@ -90,3 +109,12 @@ def fwhm_y(self):
...
+        """ Finds the rotation compared to the "x-axis" """
```
francisco-dlp: Why using degrees instead of mrad?

Maybe a better docstring would be: "Angle between major axis and x-axis"?

hyperspy/_components/gaussian2d.py
magnunor: Will do, but I think math.sin and math.cos takes radians as input, ergo rotation will be in radians as well (unless I'm misunderstanding something).

hyperspy/_components/gaussian2d.py
magnunor: The in-built .rotation parameter will already output the rotation in radians in relation to the x-axis. So having a convenience function which outputs the value in degrees is useful.

magnunor:
I've improved the tests, including moving the old gaussian2d test with the new ones I made.

I also added some extra text to the doc-strings, but I'm not certain how to best document a property like `rotation_degrees`.

Currently there is no bounds on the rotation parameter, meaning it can go to very high (or low) values. In my head it makes sense if it could be constrained between 0 and 2*pi, but I'm not sure how this could be accomplished...

dnjohnstone:
Whilst we're updating this component, I think it would also be good to start adding grad methods to 2D components as I don't think it's too hard to generalise the jacobian method etc to allow gradient based approaches if we start doing this sooner rather than later.

Perhaps beyond the scope of this PR... on the other hand makes sense to do.

magnunor:
I think we should merge this separately, as I don't think I have time to implement the grad methods so it is easier if someone else controls the source repository. 

In addition, I want this functionality merged into master as soon as possible. Since I use the rotation Gaussians almost every day, so having it in master makes life easier for me :)

vidartf:
@magnunor If you set the rotation through your own setter simply set it to the modulus (`input % 2 * np.pi`). If this is a trait, you have to dig through the `traits` doc to figure out how to set it up so that you get a setter function.

tjof2:
If it's an ellipse, is it not constrained between 0 and pi radians, since you can't distinguish between either semi-major axis?

francisco-dlp:
Requires update with current master and addressing issues above.

francisco-dlp:

Issue #866, opened 2016-02-11
Current broadcasting rules are inconsistent


francisco-dlp:
For example:

```python

>>> s = hs.signals.Signal(np.arange(10))
>>> im = hs.signals.Image(np.ones((32, 32)))
>>> s * im
<Signal, title: , dimensions: (32, 32|10)>
>>> s2 = hs.signals.Signal(np.arange(32))
>>> s2 * im
<Signal, title: , dimensions: (32|32)>

```

This is a bug.

vidartf:
Just for clarity, what would be the expected behavior?

francisco-dlp:
The problem with the example is that the size of the spectrum changes the dimensions of the outcome. This bug was introduced in 0.6 (!)

vidartf:
But what is the expected behavior of `s * im`? A `ValueError`?

francisco-dlp:
No, the wrong behaviour is the expected behaviour according to the [current broadcasting rule](http://hyperspy.org/hyperspy-doc/current/user_guide/tools.html#signal-operations), the problem is that the current broadcasting rules make no sense whatsoever (I wrote them, so I can be this harsh to myself). Have a look at the new broadcasting rules that this PR proposes in `user_guide/tools.rst`

vidartf:
To reformulate the question: How is `s * im` handled with the new rules? As far as I can tell from the new documentation, it is indeed to throw a `ValueError` (?).

I assumed the current behavior was different from what was expected since this was classified as a bug.

francisco-dlp:
You're right, the broadcasting codes wasn't buggy AFAIK, but the broadcasting rules were ;)

With the new broadcasting rules the outcome of `s * im` depends on their dimensions, just like in numpy. If the dimensions are compatible it returns the result, otherwise it raises a `ValueError`. The only difference with numpy is that HyperSpy broadcasts the navigation and signal axes independently.

vidartf:
I was referring to `s * im` from your example (`<|32, 32> * <|10>`), which I think now will become invalid (`ValueError`). However, if one of those signals were converted to be navigation only, it would be ok? I.e.:
```python
>>> im.axes_manager.set_signal_dimensions(2)
>>> s * im
<Signal, title: , dimensions: (32, 32|10)>
```
Yes?

francisco-dlp:
Yes, that's right.

francisco-dlp:
Fixed in #867

vidartf:
[...]

PR #867, opened 2016-02-11
Fix operators

### LINKED ISSUE #833: In-place operators
`Signal` currently implements the binary operators (+: `__add__`, etc.) but does not implement its in-place equivalents (+=: `__iadd__`). This isn't necessarily a problem, but if python doesn't find the in-place operators, it fakes them using the normal operators and then overwrites the original `Signal`.

```python
>>> s = hs.signals.Signal(np.ones((10,10)))
>>> s2 = s
>>> s += 3.4
>>> s is s2
False
```
This can lead to some confusion for the user, as line 3 seems to work as intended, but is not actually in-place. So, my proposal is that we should either create all the in-place operators and throw an error, or implement working copies of them.

francisco-dlp:
Fixes #833 and #866.

### DIFF
--- a/hyperspy/tests/signal/test_signal_operators.py
+++ b/hyperspy/tests/signal/test_signal_operators.py
@@ -17,154 +17,93 @@
 
 
 import numpy as np
-from nose.tools import (
-    assert_true,
-    assert_equal,
-    raises)
+from numpy.testing import assert_array_equal
+import nose.tools as nt
 
 from hyperspy.signal import Signal
 
 
-class TestBinaryOperatorsCase1:
-
-    """The signals are not aligned but can be aligned because their
-    shapes are as follows:
-
-    +---------+------------------+-------------+
-    | Signal  | NavigationShape  | SignalShape |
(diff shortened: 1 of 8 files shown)


hyperspy/tests/signal/test_signal_operators.py
```diff
@@ -17,154 +17,94 @@
...
+        self.s2 = Signal(np.ones((2, 3)))
+        self.s2.data *= 2
+
+    def test_sum_same_shape_signals(self):
+        s = self.s1 + self.s2
+        assert_array_equal(s.data, self.s1.data * 3)
+
+    def test_sum_in_place_same_shape_signals(self):
+        self.s1 += self.s2
+        assert_array_equal(self.s1.data, np.ones((2, 3)) * 3)
```
to266: Should check that the `id(self.s1)` is the same before and after (to make sure the operation is actually performed in-place).

Other places lower down as well

francisco-dlp: Thanks, fixed now.

PR #868, opened 2016-02-11
Loading xyz-files as Image signals

magnunor:
xyz-files contain 3-D atomic lattice points in the form: 

Sr    0.000000    0.000000    0.000000
Sr    0.000000    0.000000    3.905000
Sr    0.000000    0.000000    7.810000
Sr    0.000000    0.000000   11.715000

The file format can be exported from atomic lattice visualization software such as Vesta (and probably others).

It would be useful if these files could be opened and visualized using HyperSpy, so I implemented a simple "3-D points to image" function which projects the structure into a 2-D image then runs a Gaussian filter to simulate experimental broadening. 

It is currently fairly simple, as it can only project on one of the x-, y- or z-axes. But I've got code for projecting on arbitrary planes. The "3-D points to image" function is currently located within the xyz.py itself, but it should be moved somewhere else so other, similar, file formats like vasp and vesta can be implemented as io-plugins as well. 

Possible extensions could be allowing for 3-D image stacks, which would allow for visualization of atomic layers.

Example of image signal generated using the xyz-io-plugin: 
sto

### DIFF
--- a/hyperspy/io_plugins/__init__.py
+++ b/hyperspy/io_plugins/__init__.py
@@ -19,9 +19,9 @@
 
 from hyperspy import messages
 from hyperspy.io_plugins import (msa, digital_micrograph, fei, mrc, ripple,
-                                 tiff, semper_unf, blockfile, dens)
+                                 tiff, semper_unf, blockfile, dens, xyz)
 io_plugins = [msa, digital_micrograph, fei, mrc, ripple, tiff, semper_unf,
-              blockfile, dens]
+              blockfile, dens, xyz]
 try:
     from hyperspy.io_plugins import netcdf
     io_plugins.append(netcdf)
--- a/hyperspy/io_plugins/xyz.py
+++ b/hyperspy/io_plugins/xyz.py
@@ -1,0 +1,110 @@
+# -*- coding: utf-8 -*-
+# Copyright 2007-2016 The HyperSpy developers
+#
+# This file is part of  HyperSpy.
+#
+#  HyperSpy is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+#  HyperSpy is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with  HyperSpy.  If not, see <http://www.gnu.org/licenses/>.
+
+import warnings
+import datetime
+
+import numpy as np
+
+from scipy.ndimage.filters import gaussian_filter
+from hyperspy.misc.elements import elements
+
+
+# Plugin characteristics
+# ----------------------
+format_name = 'xyz'
+description = ''
+
+full_support = False
+# Recognised file extension
+file_extensions = ['xyz']
+default_extension = 0
+
+# Writing capabilities
+writes = False
+
+# -----------------------
+# File format description
+# -----------------------
+#
+
+def file_reader(filename):
+    atom_array = np.loadtxt(
+            filename,
+            skiprows=2,
+            comments='EOF',
+            dtype={
+                'names': ('atom_type', 'x', 'y', 'z'),
+                'formats': ('S4', 'f8', 'f8', 'f8')})
+
+    image_array, axes_dict = _generate_image_from_3d_points(atom_array)
+    return image_array, axes_dict
+
+def _find_nearest_index(array,value):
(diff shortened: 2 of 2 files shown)


dnjohnstone:
I've been thinking about how to store and interact with atomic structures too...

Was thinking about starting to depend on something like diffpy.Structure or pyobjcryst, which are open source repositories found here https://github.com/diffpy

These are actually quite well established in Crystallography communities and the latter is based on libobjcryst++ which has been used for decades and is very stable. All of the space group symmetry operators etc could be useful for manipulating the data too.

Wha tdo you think?

francisco-dlp:
What about splitting this PR in two:

* xyz reader (and writer?)
* New class `PointCoordinates3D` with a `digitize` method that bins the points into a 3D array of shape (nav_axis1, ..., nav_axisn| 3)

From here it'll be straightforward for the user to project and blur if needed.

vidartf:
If we're making a class for converting scatter data into binned arrays, why limit it to 3D? It should be quite easy to deduce from the dimensionality from the input.

francisco-dlp:
You are right, we can create the more specialised subclasses later on.

magnunor:
@dnjohnstone, tying all this into diffpy would be nice. At some point one could do structure determination by using https://github.com/hyperspy/hyperspy/pull/864 + the atom finding code I've written.

@vidartf and @francisco-dlp, I'm guessing the first step could be to write a `PointCoordinates` base class, which has several subclasses like `PointCoordinates3D` and `PointCoordinates2D`. These could have functions like `.project_to_image('z', gaussian_blur=3.)` or  `.project_to_imagestack('x', gaussian_blur=3., size=(1024,1024,256))`.

vidartf:
@magnunor: I'm not sure if we need a separate set of classes for each dimensionality. In principle, we need some (global) functions `project_points` (which is basically only a multiplication with a transformation matrix), and `digitize` for converting into continuous array. Then we could make a class that stores a reference to the point array, and stores the coordinate axis. This class can then have convenience functions that automatically can supply the coordinates and dimensionality to the global functions and does smart convolution (allows different classes of points to be convolved by different kernels). See proposed structure below:

```python
def project_points(points, coordinates_axis, projection_vector):
    ndim = np.shape(points)[coordinates_axis]
    trans = <transformation matrix based on ndim and projection vector>
    return trans * points

def digitize(points, bins):
    return # Somehow digitize points here

class PointStructure(object):
    def __init__(self, points, coordinates_axis, axes_manager, type_axis=None, convolution_kernels=None):
        self.points = np.array(points)
        self.coordiantes_axis = coordinates_axis
        self.type_axis = type_axis
        self.convolution_kernels = convolution_kernels   # TODO: Make sure this is a list of kernels

    def project(self, direction):
        # If direction is axis specification, convert it to projection vector
        # This could potentially also accept zone axis?
        points = <set up to iterate over point types>
        bins = <setup bins from axes_manager>
        data = None
        for sub_points, kernel in zip(points, self.convolution_kernels):
            projection = project_points(self.points, self.coordinates_axis, direction)
            binned = digitize(projection, bins)
            convolved = <convolve in N dimensions with kernel. Use function from external package>
            if data is None:
                data = convolved
            else:
                data += convolved
        return data
```

magnunor:
@vidartf, having it generalized like that is probably a more elegant solution. But I don't think I have time to implement something like that, at least not for the first iteration.

At least my initial thought was implementing a "simpler" version first, then making it more generalized later on. 

In such a general class, would it be possible to have more specialized functions which are relevant for certain dimensions? Like for a 2-D system it would make sense to have a `plot` function which simply plots using `plt.scatter`, while for the 3-D case it could either have a `plot3D_scatter` or `plot_2Dprojection`.

vidartf:
@magnunor If you just start with 2D, then in the code above:
 - Use gaussian filter as convolution step `if ndim==2` and throw a `NotImplementedError` for any other dimensions.
 - To setup the image array, use `numpy.digitize()` with bins like this pattern: `axes_manager[0].axis - 0.5 * axes_manger[0].scale`.
 - Simple 3D to 2D orthogonal transformation matrix: 
```python
def orthogonal_projection_matrix(normal):
    M = np.identity(3)
    normal /= np.linalg.norm(normal)  # Make unit vector
    M[:3, :3] -= np.outer(normal, normal)
    return M
```

francisco-dlp:

thomasaarholt:
This would be nice to have. How would it be implemented into `load` or a different loader? I guess a different name would be more sensible, as loading this type of file would require additional input, like broadening and orientation?

dnjohnstone:
Is this still relevant with atomap now existing?

dnjohnstone:
Can I suggest that this is out-of-scope for the main hyperspy repo and should therefore be closed @magnunor ?

magnunor:
@dnjohnstone, yeah. Made a new issue about this in Atomap's issue tracker: https://gitlab.com/atomap/atomap/issues/95

PR #869, opened 2016-02-12
Fixes #827

### LINKED ISSUE #827: .estimate_parameters() not working according to website example
`.estimate_parameters()` does not work as shown in the example in the [documentation](http://hyperspy.org/hyperspy-doc/current/api/hyperspy._components.html#hyperspy._components.gaussian.Gaussian.estimate_parameters), code copied below.

As @francisco-dlp points out, it currently requires the addition of a model (see commented lines below).

It is has been [suggested](https://gitter.im/hyperspy/hyperspy?at=56979a232bc35f6c1c1a0939) to fix the code rather than to modify the example, as the method shouldn't require a model created.

```python
import hyperspy as hs
import numpy as np
g = hs.model.components.Gaussian()
x = np.arange(-10,10, 0.01)
data = np.zeros((32,32,2000))
data[:] = g.function(x).reshape((1,1,2000))
s = hs.signals.Spectrum(data)
s.axes_manager._axes[-1].offset = -10
s.axes_manager._axes[-1].scale = 0.01
#m = s.create_model()
#m.append(g)
g.estimate_parameters(s, -10,10, False)
```

francisco-dlp:


vidartf:
If you try to use the same component for two signals with different axes managers, this will fail, since the component will store the axes manager the first time it is called.

francisco-dlp:
Good point, thanks! (Fixed.)

PR #870, opened 2016-02-13
Bug fix 1d peakfinding using find_peaks1D_ohaver()

magnunor:
https://github.com/hyperspy/hyperspy/pull/630


hyperspy/tests/signal/test_find_peaks1D_ohaver.py
```diff
@@ -0,0 +1,23 @@
...
+
+    def test_find_peaks1D_ohaver_high_amp_thres(self):
+        spectrum = self.spectrum
+        peak_list = spectrum.find_peaks1D_ohaver(amp_thresh=10.)[0]
+        assert_equal(len(peak_list), 0)
+
+    def test_find_peaks1D_ohaver_zero_value_bug(self):
+        spectrum = self.spectrum
+        peak_list = spectrum.find_peaks1D_ohaver()[0]
+        assert_not_equal(len(peak_list), 0)
```
to266: I'd prefer you check at least for exactly the number of peaks you expect (not just != 0), since the data is not random. However if please let us know if you can do that today, otherwise I think it's not worth delaying merging

magnunor: I'll (most likely) be able to fix this today.

magnunor:
Added a specific number of peaks for the test, so should be ready now.

vidartf:
Will fix #628.

PR #873, opened 2016-02-15
Python3 for 0.8.x

to266:
Similar to https://github.com/hyperspy/hyperspy/pull/873, but did not remove old-type slicing


hyperspy/drawing/utils.py
```diff
@@ -956,7 +956,7 @@ def plot_spectra(
     if color is not None:
         if hasattr(color, "__iter__"):
             color = itertools.cycle(color)
-        elif isinstance(color, basestring):
+        elif isinstance(color, str):
```
vidartf: This will never be reached since `str` has `__iter__`.

vidartf:
Similar to #769 :)

to266:
Fixed, thanks for the reminder and pointers :)

vidartf:
Is the plan to merge this at/after release step, or to create a different branch?

francisco-dlp:
I think that this shouldn't be merged. Once reviewed, this should be pushed to a new branch in the main repository e.g. 0.8.x_py3?

vidartf:
@to266 : I just ran a `git fetch --all`, and it seems that you got all of my (now deleted) tags for trialing the appveyor deployment. I'm guessing it was when you cherry-picked commits in this branch? Not sure, but you might want to clean up your tags :)

to266:
Thanks, removed those... And probably not all of them any way - I'm currently using at least one of them for exactly the reason :)

to266:
If appveyor manages to build for py3.5, ready for review / merging / branching / whatever will happen...

in particular, @vidartf, could you have a look that the `appveyor.yml` is fine? thanks

vidartf:
Well, you should probably merge latest version of 0.8.x for appveyor.yml to work with latest changes from #875.

vidartf:
Looks OK to me, but I find it tricky to reliably judge such code by diff only. Why did you remove `natsort` from installation into deployment? My impression was that natsort was not part of the default WinPython distro?

appveyor.yml
```diff
@@ -21,7 +40,7 @@ environment:
       CONDA_NPY: "19"
       WP_URL: 'http://heanet.dl.sourceforge.net/project/winpython/WinPython_2.7/2.7.10.3/WinPython-32bit-2.7.10.3.exe'
       WP_CRC: '99E825BCDE4FFA57E612A0D31DF1487861C96D1903E4D20C5843FA2651C9E33B'
-      DEPS: "numpy scipy matplotlib traits traitsui ipython h5py sympy scikit-learn dill mock"
+      DEPS: "numpy scipy matplotlib traits traitsui ipython h5py sympy scikit-learn dill mock natsort"
```
vidartf: Why is `natsort` added to the 2.7 dependencies?

vidartf: TBH, you can probably just delete the 2.7 entries. I set this up before there was a decision to not have a code base that works on both versions of python.

to266:
That's exactly why I asked for your advice, thanks. Should be fixed.

vidartf:
Have you done any test deployments on your branch? Probably cleaner if we don't muck about too much with test tags in the main repo.

to266:
Nope, not yet.

I'll do it now

to266:
FYI 
https://ci.appveyor.com/project/to266/hyperspy/build/1.0.24

vidartf:
@to266 Not sure if that is an issue with pip, the WinPython included compiler, or traits. Maybe try updating the WinPython URLs (and checksums) to latest version. That should include a newer version of pip as well as a script to update pip correctly (patched, see https://github.com/winpython/winpython/commit/0b88f5c53e5306255fd3d9497480df2ef0665253). If the issue is with the compiler, we can switch to use the MSVC compilers that are installed on the AppVeyor servers, but I would prefer to use the WinPython ones for consistency if at all possible.

to266:
Huh, still the same problem with the new winpython release (`pip` out of date):
https://ci.appveyor.com/project/to266/hyperspy/build/1.0.27/job/7fnk21yy7vv1oxjg

Also it looks like it fails to build `traits` wheel, not sure how relevant it is, as it stops at the pip step...

to266:
Should we include this (https://github.com/winpython/winpython/commit/0b88f5c53e5306255fd3d9497480df2ef0665253) step?

francisco-dlp:
Are you should that it is failing because there is a newer pip? May it not be failing to compile traits?

to266:
no, I'm not. However that is the last command before crash. Any suggestions?

francisco-dlp:
What about getting traits from http://www.lfd.uci.edu/~gohlke/pythonlibs/ instead to see if that works?

vidartf:
Gohlke's page is designed to prevent automated/scripted fetching of files (cf. previous discussion about mayavi). If we're failing to build for 3.5 only, it might be an issue with traits, I just wanted to make sure that we use the latest compile tools from WinPython.

vidartf:
PS: Which versions of Python 3 do we want to supply bundles for? Currently, the appveyor file was set up to test on 3.4 (64 bit) and 3.5 (32 bit). For deployment, I assume we want a 32 bit and 64 bit version for each version of Python 3. These extra versions should probably be set up to only run when there is a tag (i.e. to not have to many redundant test runs).

to266:
Actually you mixed it up - 3.4 for 32 bit, and 3.5 for 64 bit, as winpython did not have 3.5 32 bit distribution before this (last) release - hence I changed them both to be 3.5. Any reason why not to go for 3.5 only? I'm not really experienced with 3.x pythons and probably would not be able to tell the difference anyway...

vidartf:
Well, the first test is still running on Python 3.4 since it has `PYTHON: "C:\\Miniconda3"`, which is currently 3.4.

vidartf:
BTW: I assume the correct path would be `PYTHON: "C:\\Miniconda35"`.

vidartf:
> Any reason why not to go for 3.5 only?

For bundle, 3.5 only is OK. We should probably test on 3.4 as well, but it should probably be ok to add this in Travis only, as that runs multiple tests in parallel.

to266:
Strange, it should be 3.5 for all:
http://conda.pydata.org/miniconda.html

vidartf:
This is how AppVeyor has set up their images.

francisco-dlp:
Shouldn't natsort be added to `setup.py`?

to266:
@vidartf you're right: http://www.appveyor.com/docs/installed-software#python

I'll add/fix these things now, and have a look how to fix the build later (or just add a new tag just for the fun of it..)

to266:
actually I just noticed one of them run fine:
https://github.com/to266/hyperspy/releases/tag/untagged-80aad89832b096f3fcae

Anyone keen to test it?

vidartf:
> https://github.com/to266/hyperspy/releases/tag/untagged-80aad89832b096f3fcae

404?

to266:
Deleted the tag, it was this:
https://ci.appveyor.com/project/to266/hyperspy/build/1.0.27/job/4wybqx797ti0y0lx/artifacts

vidartf:
Seems like changing the dir fixed the issue. The question is why? Is there a cross-contamination between the test and deployment environments? Maybe we should inspect the `pip` command used in the deployment, and maybe call it by absolute value if need be? There might be an issue with `CMD_IN_ENV` ?

to266:
In the commit before that I also added requirements for `traits` and `traitsui` (so that the versions are >= the first that supports python). Also before changing the directory `CONDA_PY` was already "35" - may have created the clash there, and now that everything is consistent, it worked fine and as expected

vidartf:
PS: I see you switched to using the SHA1 checksum, not entirely sure why, but it means we lose some inherent protection from SHA256:

> File integrity checking is often intended to ensure that a file was downloaded correctly. But, if it is being used to verify that the file was not maliciously tampered with, you should consider an algorithm that is more resilient to collisions [than SHA1]. See also https://www.win.tue.nl/hashclash/SoftIntCodeSign/

to266:
Because it was directly supplied by the winpython.
https://github.com/winpython/winpython/issues/264

I was also too lazy to download and hash them myself, but clearly it's not as important...

to266:
There, fixed the laziness and now use `SHA256` (and it will hopefully be included with the next winpython release)

to266:
https://github.com/hyperspy/hyperspy/tree/0.8.x_py3

vidartf:
If pip ends up trying to fetch 0.8.4 on python 2, you could try adding a
python 2 wheel to 0.8.4. I think the wheel should be preferred? Or maybe
that was only universal wheels.

PR #875, opened 2016-02-17
Add start_jupyter_cm to WinPython distribution

francisco-dlp:
From version 0.8.3 hyperspy no longer ships with code to add the "HyperSpy here" entries to the context menu. Instead, we encourage users to install the new package [start_jupyter_cm](https://github.com/hyperspy/start_jupyter_cm/). This PR simply adds this package to the winpython distribution created by appveyor.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -127,7 +127,7 @@ before_deploy:
   - "%CMD_IN_ENV% %WP_INSTDIR%/scripts/env.bat"
   - "%CMD_IN_ENV% python.exe setup.py install"
   - "%CMD_IN_ENV% easy_install --upgrade pip"
-  - "%CMD_IN_ENV% pip install --upgrade configobj traitsui"
+  - "%CMD_IN_ENV% pip install --upgrade configobj traitsui start_jupyter_cm"
   # Custom installer step
   # TODO: Re-run tests in WinPython environment
   - ps: Add-AppveyorMessage "Creating installer..."
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -127,8 +127,6 @@ before_deploy:
   - "%CMD_IN_ENV% %WP_INSTDIR%/scripts/env.bat"
   - "%CMD_IN_ENV% python.exe setup.py install"
   - "%CMD_IN_ENV% pip install --upgrade configobj traitsui start_jupyter_cm"
-  - "%CMD_IN_ENV% %WP_INSTDIR%/scripts/make_winpython_movable.bat"
-  - "%CMD_IN_ENV% "%WINPYDIR%\python.exe" -c "from winpython import wppm;dist=wppm.Distribution(r'%WINPYDIR%');dist.patch_standard_packages('pip');dist.patch_all_shebang(to_movable=True)"
   # Custom installer step
   # TODO: Re-run tests in WinPython environment
   - ps: Add-AppveyorMessage "Creating installer..."
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -127,6 +127,8 @@ before_deploy:
   - "%CMD_IN_ENV% %WP_INSTDIR%/scripts/env.bat"
   - "%CMD_IN_ENV% python.exe setup.py install"
   - "%CMD_IN_ENV% pip install --upgrade configobj traitsui start_jupyter_cm"
+  - "%CMD_IN_ENV% %WP_INSTDIR%/scripts/make_winpython_movable.bat"
+  - "%CMD_IN_ENV% "%WINPYDIR%\python.exe" -c "from winpython import wppm;dist=wppm.Distribution(r'%WINPYDIR%');dist.patch_standard_packages('pip');dist.patch_all_shebang(to_movable=True)"
   # Custom installer step
   # TODO: Re-run tests in WinPython environment
   - ps: Add-AppveyorMessage "Creating installer..."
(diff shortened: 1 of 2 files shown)


francisco-dlp:
@vidartf, could you have a look at this?

to266:
I guess you will want to add the instructions how to actually add the shortcuts in the user-manual, right? Because as far as I can tell, this just adds the package, without making it run on installation.

francisco-dlp:
The package script is run by the winpython installer [see this PR](https://github.com/hyperspy/hyperspy-bundle/pull/10/).

vidartf:
It might be easier to add `start_jupyter_cm` as an install requirement of the bundle package?

francisco-dlp:
I'm not sure if I follow you. We want `start_jupyter_cm` installed in the WinPython distribution. Wouldn't adding it to the installation requirements of the bundle simply install in the computer from where the bundle is created but not actually in the WinPython distribution?

vidartf:
The end results will be exactly the same.

appveyor.yml deployment step in your current suggestion:
```
pip install <hspy_bundle_url>
...
pip install start_jupyter_cm
```

in my suggestion:
```
pip install <hspy_bundle_url>   <-- this step also installs start_jupyter_cm
```

francisco-dlp:
I'm certainly missing something. This PR installs `start_jupyter_cm` in WinPython. AFAIK we don't install the hspy_bundle in winpython, we install it in the environment that we use to create the bundle.

vidartf:
Heh, no it was me who were missing something :) Disregard those comments.

vidartf:
But don't forgot @to266's comment about text in the user guide.

francisco-dlp:
Ok, added to todo list.

francisco-dlp:
This is ready for review.

appveyor.yml
```diff
@@ -126,8 +126,7 @@ before_deploy:
   - "SET PATH=%ORIGPATH%"
   - "%CMD_IN_ENV% %WP_INSTDIR%/scripts/env.bat"
   - "%CMD_IN_ENV% python.exe setup.py install"
-  - "%CMD_IN_ENV% easy_install --upgrade pip"
-  - "%CMD_IN_ENV% pip install --upgrade configobj traitsui"
+  - "%CMD_IN_ENV% pip install --upgrade configobj traitsui start_jupyter_cm"
```
vidartf: For me, `pip install start_jupyter_cm` creates two `.exe` files in my script folder with absolute paths to the python executable embedded in the shebang. This means that the python path as it is on the AppVeyor server will be embedded in the bundle. Consider patching using winpython methods?

francisco-dlp: We'll have to do that. Isn't this something that we should report to the WinPython developers?

vidartf: Not really a WinPython issue as we do not use any WinPython tools for creating our bundle. We simply repackage a bundle based on a WinPython installation. I mentioned using the WinPython routines so we don't have to reinvent the wheel.

francisco-dlp: @vidartf, I can't reproduce this issue. After installing the bundle I have moved it to a different folder, I have removed anaconda from the windows path and renamed anaconda's folders. Still the scripts run just fine. Could you provide us with steps to reproduce the issue and diagnose it?

francisco-dlp: @vidartf, although I can't reproduce the issue, in prevention I've added a line to patch all shebangs to appveyor. Is this ready to merge then?

vidartf: What I did to test was:
 - `pip install start_jupyter_cm`.
 - Check that the add executable runs from (WinPython) CMD.
 - Open up add executable and set a spurious path as shebang, now the add command fails.

 ```
> jupyter_context-menu_add.exe
Fatal error in launcher: Unable to create process using '"'
```
 - Change shebang to only `python.exe`, check that it runs correctly again.

This is with pip version 8.0.2.

File: appveyor.yml [hunk 2]
```diff
@@ -126,8 +126,9 @@ before_deploy:
   - "SET PATH=%ORIGPATH%"
   - "%CMD_IN_ENV% %WP_INSTDIR%/scripts/env.bat"
   - "%CMD_IN_ENV% python.exe setup.py install"
-  - "%CMD_IN_ENV% easy_install --upgrade pip"
```
vidartf: Maybe upgrade with ` - "%CMD_IN_ENV% python -m pip install --upgrade pip` since we now patch packages?

appveyor.yml
francisco-dlp: Ok, but you don't disclose the original shebang of the file. In my case it was already `python.exe`. I'm using pip as distributed with WinPython, may that be the issue again?

vidartf: That might actually be the case if WinPython patches pip such that also future pip installations are installed with `python.exe` as shebang. For reference my original shebang was an absolute path.

vidartf: In that case, this should be good to merge. Do any final adjustments you feel necessary and merge when ready :)

francisco-dlp: Ok, in this case, what about removing the line that patches the scripts? I think that we should keep the distributions as closed to pure WP as possible as every time that we touch something there is the potential of breaking it and of increasing the maintenance burden.

vidartf: Sure.

francisco-dlp: Done, please consider merging.

vidartf:
For the future versions of WinPython, I saw this recent commit: https://github.com/winpython/winpython/commit/0b88f5c53e5306255fd3d9497480df2ef0665253

PR #877, opened 2016-02-21
New: Logging

vidartf:
Implements #749.

 - Supersedes and deprecates `hyperspy.messages`.
 - Replaces as many `print` statements as suitable.
 - Adds convenience function `hyperspy.logger.set_log_level()` / `hyperspy.api.set_log_level()`, which sets the log level of the hyperspy modules only.
 - Adds logging for backend/toolkit code.
 - Some cleanup/PEP8 of code.


dnjohnstone:
This seems quite handy - so far it's doing what I expect it to as well. Not sure what's going on with /gui/main_window.py in the diff of this though? Is that just where it's come from a different local branch?

vidartf:
/gui/mainwindow.py was an unused file where all the code wad commented out.
I removed it, because its the job of version control systems like git to
keep such things. The main code repo should be kept ad uncluttered as
possible.

hyperspy/io_plugins/hdf5.py
```diff
@@ -425,9 +425,9 @@ def parse_structure(key, group, value, _type, compression):
             try:
                 group.attrs[key] = value
             except:
-                print("The hdf5 writer could not write the following "
-                      "information in the file")
-                print('%s : %s' % (key, value))
+                warnings.warn(
```
to266: I would use `logger` here - you want to print all the things that were not saved, all the time. Currently it only prints the warnings for the first file created.

hyperspy/io_plugins/msa.py
```diff
@@ -189,8 +192,9 @@ def file_reader(filename, encoding='latin-1', **kwds):
                     parameters[parameter] = keywords[clean_par]['dtype'](
                         value.replace(' ', ''))
                 except:
-                    print("The %s keyword value, %s " % (parameter, value) +
-                          "could not be converted to the right type")
+                    warnings.warn(
+                        "The %s keyword value, %s " % (parameter, value) +
+                        "could not be converted to the right type")
```
to266: probably here too - I suspect it would only warn the user when opening the first file that raises the warning

hyperspy/io_plugins/semper_unf.py
```diff
@@ -616,8 +627,9 @@ def to_signal(self):
         signal.original_metadata.add_dictionary(self.metadata)
         return signal
 
-    def print_info(self):
-        """Print important flag information of the :class:`.~SemperFormat` object.
+    def log_info(self):
+        """print important flag information of the :class:`.~SemperFormat`
+        object.
```
to266: Doc needs updating, since does not print any more

vidartf:
Note: I tried to follow the python docs when deciding when to use the different mechanisms: https://docs.python.org/2/howto/logging.html#when-to-use-logging

hyperspy/io_plugins/hdf5.py
vidartf: I'm not familiar with when this would occur: Is this something that is expected to happen often? Or is this only if the user has stored something wrong in metadata? What is the motivation for suppressing the exception?

vidartf: Looking at the python docs, I guess we should here use `_logger.execption()` since we're suppressing an exception?

to266: It occurs when there is something wrong (i.e. we don't know how to save) in the metadata, yes. We don't suppress because it's not an exception as I understand it - we essentially check against our list of things that we can save, and if the item falls through, we just print it here. 

I've had it happen when I had to load a corrupted (I guess? never really got to the bottom of it) dm3 file, and the `original_metadata` was filled with things we could not save. 

It should not happen often, but when it happens I would rather still be able to save the main part (i.e. the data + axes).

to266: ...and I would probably use `logging.error` 
 >"Due to a more serious problem, the software has not been able to perform some function." 

seems to be quite accurate

to266:
@vidartf do you think you'll have the time to add the things I mentioned? currently I can't PR the changes as I don't have the time, but review is quick, and it's the only thing preventing a merge

hyperspy/io_plugins/hdf5.py
vidartf: Switched to `logging.exception`, as it is the same as `logging.error` except it automatically adds exception info to the entry.

vidartf:
@to266 Appropriate changes should have been made now. I even fixed another remaining print as well :)

vidartf:
I haven't tested though, and I see from tests that there is no coverage for corrupt hdf5 files. If you know how to create one (paste code for creating bad metadata?) I can add it.

to266:
Well most objects should fail. For example I hope this would
`s.metadata.roi = hs.roi.<anyroi>`

vidartf:
Your example worked (i.e. failed) as expected. Just pushed a test.

to266:
Yup, I'm just waiting for appveyor

Issue #880, opened 2016-02-23
plot spectra fails in master because Signal can't be indexed


dnjohnstone:
In the current master hs.plot.plot_spectra(something) fails with an error as below. I think this is simply because we should now be using isig/inav rather than directly indexing? If so this is quick to fix and I'll do it in a second.

`` ---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
<ipython-input-21-6abad2bda0d4> in <module>()
----> 1 fig_trace = hs.plot.plot_spectra(trace)

/home/dnj23/hyperspy/hyperspy/drawing/utils.pyc in plot_spectra(spectra, style, color, line_style, padding, legend, legend_picking, legend_loc, fig, ax, **kwargs)
    998                            ax,
    999                            color=color,
-> 1000                            line_style=line_style,)
   1001         if legend is not None:
   1002             plt.legend(legend, loc=legend_loc)

/home/dnj23/hyperspy/hyperspy/drawing/utils.pyc in _make_overlap_plot(spectra, ax, color, line_style)
    316         ax.plot(x_axis.axis, spectrum.data, color=color, ls=line_style)
    317     if len(spectra) > 1:
--> 318         _set_spectrum_xlabel(spectra[-1], ax)
    319     ax.set_ylabel('Intensity')
    320     ax.autoscale(tight=True)

TypeError: 'Signal' object does not support indexing``

vidartf:
It seems to me that the code expects spectra to be a collection of signals, but it is actually a `Signal`.

dnjohnstone:
Hmmm do you think that's indicative of a more fundamental change being needed then @vidartf stupid fix from me was to just use inav to do the indexing which got rid of the errors and seems to reproduce the previous functionality, which at least lets me get on with what I was doing...

vidartf:
In your paste above, what is `trace` ? Can you get expected results by passing `[trace]`?

dnjohnstone:
In that particular case it was a Signal class object that I'd made using your Line2D ROI tool - but I got the same error if it was get_decomposition_factors(), which I think is the more typical use. Inserting inav as I have makes both cases work.

vidartf:
The `plot_spectra` code supports both a single `Spectrum` as well as a collection of `Spectrum` (e.g. `[s1, s2, s3]`). Your PR fixes the first case, but breaks the second case.

dnjohnstone:
[...]

PR #881, opened 2016-02-23
BUG_FIX: Use inav in plot spectra

dnjohnstone:
This provides a fix for #880 so that the Signal object is indexed using inav rather than direct indexing.

### DIFF
--- a/hyperspy/drawing/utils.py
+++ b/hyperspy/drawing/utils.py
@@ -315,7 +315,7 @@ def _make_overlap_plot(spectra, ax, color="blue", line_style='-'):
         x_axis = spectrum.axes_manager.signal_axes[0]
         ax.plot(x_axis.axis, spectrum.data, color=color, ls=line_style)
     if len(spectra) > 1:
-        _set_spectrum_xlabel(spectra[-1], ax)
+        _set_spectrum_xlabel(spectra.inav[-1], ax)
     ax.set_ylabel('Intensity')
     ax.autoscale(tight=True)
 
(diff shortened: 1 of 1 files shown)


hyperspy/drawing/utils.py
```diff
@@ -315,7 +315,7 @@ def _make_overlap_plot(spectra, ax, color="blue", line_style='-'):
         x_axis = spectrum.axes_manager.signal_axes[0]
         ax.plot(x_axis.axis, spectrum.data, color=color, ls=line_style)
     if len(spectra) > 1:
```
vidartf: `if not isinstance(spectra, Signal) and len(spectra) > 1:` might be better?

vidartf: Although I have to admit I don't understand why the xlabel is only set `if len(spectra) > 1`...

francisco-dlp:
I think that @vidartf is right, this function, unlike `plot_images`, wasn't designed to take spectra directly from a multidimensional `Signal`. `plot_images` is newer and enjoys this and some other features that it'll be nice to port to `plot_spectra`.

dnjohnstone:
That may be true - but it doesn't change the fact that until now it was possible to use the line hs.plot.plot_spectra(s.get_decomposition_factors()) to get an overlay or cascade plot of results - and this is used in the EDS tutorial notebooks.

That's now not possible because you have to slice rather than index the output of get_decomposition_factors().... anyone who like me had lifted this line from the tutorials and find it useful to use on a day to day basis will now find this broken.

hyperspy/drawing/utils.py
francisco-dlp: @vidartf, and I think that that's a bug.

What about:

```python

_set_spectrum_xlabel(spectra if isinstance(spectra, Signal) else spectra[-1], ax)

```

dnjohnstone:
The line from @francisco-dlp seems to work, so I've used that. Not totally sure why coveralls is complaining as it's also saying it's unchanged, but otherwise I think this is ok.

vidartf:
@dnjohnstone : Since the new code is one line shorter than the previous version, ad that line was covered by tests, the ratio between covered and uncovered lines has declined. Don't worry about it :)

Issue #821, opened 2016-01-08
Splitting HyperSpy


francisco-dlp:
We define HyperSpy as "HyperSpy is an open source Python library which provides tools to facilitate the interactive data analysis of multi-dimensional datasets that can be described as multi-dimensional arrays of a given signal". However the current project does much more than that. For example, as pointed out in #804, we are maintaining an elements database, we provide especialised tools EELS, EDX and, in doing so, we also implement things (e.g. EELS xsections) that could well be part of an EELS library.

There is a chance that, if we split the project in multiple subprojects, those subprojects may develop a community around them that may be wider than HyperSpy's current community because, to most, at present HyperSpy looks like a EM data analysis toolbox. Also, I think that it could increase the attractiveness of hyperspy to researchers from other fields that, in this way, could lead their own subprojects on their field of interest.

So, what about splitting the project e.g. as follows:
- hyperspy-core: basically Signal, Spectrum, Image
- hyperspy-visual: plotting features
- hyperspy-fitting: multi-dimensional curve fitting
- hyperspy-learn: machine learning tools
- hyperspy-tem: all of the current TEM features.
- Other more specialised projects that have nothing to do with multi-dimensional data analysis but that other hyperspy-\* project depend on e.g
  - elements
  - eels_xs
  - tem_fileformats
  - etc

vidartf:
Just a minor comment to get the ball rolling: hyperspy-tem -> hyperspy-em, and include SEM :)

Other than that, most of these seem like just moving existing code around, except maybe "tem_fileformats". If we separate this one, wouldn't we have to create a much more rigorous interface? Or can we use the one we have now?

magnunor:
I agree, especially since people want to add more diverse types of functionalities https://github.com/hyperspy/hyperspy/issues/787 . So this is a good way of reducing the "bloating".

I'm guessing this would also clean up the issue tracker a little bit, since each sub-project would have its own.

How would it work in practice? Each sub-project would have its own project in hyperspy, like hyperspy/hyperspy-core, hyperspy/hyperspy-visual, ...?

dnjohnstone:
In #804 @francisco-dlp mentioned that there are "advantages and disadvantages" to splitting the code -- could someone briefly explain what the disadvantages are? 

I guess the main thing is something to do with how you get the separate projects to be combined back together for the user who will likely want an "out of the box" solution as far as possible? 

[...]

PR #883, opened 2016-02-24
Support for new fei format and scale/units for diffraction

### LINKED ISSUE #397: Add support for 64 bits FEI ser files
It seems that FEI has a new version of their emi/ser file. I cannot read newly acquired files. I obtained an error about array of different shapes that cannot be broadcasted

### LINKED ISSUE #662: FEI reader needs unittests
Currently the FEI reading features are not tested. We need datasets to test them. The digital micrograph reader tests in ``hyperspy/tests/io/`` could serve as inspiration.

ericpre:
- add support for emi/ser files with versions >= 544, i.e. for files acquired on windows 64 bits, fix #397
- add several kinds of emi/ser example files acquired on different microscopes for testing
- add tests for fei reader, fix #662
- add support for the scale/units of diffractions

~~When I tried to use the xml data in the emi files to get informations about the detectors, it was not able to get it work nicely and I suspect the xml data are not parsed correctly. For example, for a file containing an EDS and a diffraction acquisition, the xml data about the EDS detector are parsed to the diffraction signal and vice versa.~~
See @francisco-dlp comment about xml data below.


hyperspy/tests/io/test_fei.py
```diff
@@ -0,0 +1,287 @@
...
+        nt.assert_equal(s0.metadata.Acquisition_instrument.TEM.acquisition_mode, 'TEM Diffraction')
+        nt.assert_almost_equal(s0.axes_manager[0].scale, 0.10157, places=5)
+        nt.assert_equal(s0.axes_manager[0].units, '1/nm')
+        nt.assert_almost_equal(s0.axes_manager[1].scale, 0.10157, places=5)
+        nt.assert_equal(s0.axes_manager[1].units, '1/nm')
+        
+    def test_load_diffraction_line_scan(self, verbose=True):
+        fname0 = os.path.join(self.dirpathnew, '16x16-line_profile_horizontal_5x128x128_EDS.emi')
+        s0 = load(fname0, verbose=verbose)
+        #s0[1] contains EDS
```
ericpre: On my system (fedora 23), the first item in the list is the diffraction patterns and the 2nd one is the EDS... which doesn't make sense, I don't understand... however, travis gets it right!

hyperspy/io_plugins/fei.py
```diff
@@ -286,7 +308,15 @@ def load_ser_file(filename, verbose=False):
 
         # Read the first element of data offsets
         f.seek(header["OffsetArrayOffset"][0])
-        data_offsets = readLELong(f)
+        # OffsetArrayOffset can contain 4 or 8 bytes integer depending if the
+        # data have been acquired using a 32 or 64 bits platform.
+        SeriesVersion = header['SeriesVersion']
+        if SeriesVersion <= 528:
+            data_offsets = readLELong(f)
+        elif SeriesVersion >= 544:
```
francisco-dlp: Why not 

```python

if SeriesVersion <= 528:
    data_offsets = readLELong(f)
else:
....
```
as above.

hyperspy/io_plugins/fei.py
```diff
@@ -503,17 +537,21 @@ def ser_reader(filename, objects=None, verbose=False, *args, **kwds):
                             'Dim-%i_CalibrationOffset' % (i + 1)][0],
                         'scale': header[
                             'Dim-%i_CalibrationDelta' % (i + 1)][0],
-                        'units': header['Dim-%i_Units' % (i + 1)][0],
+                        'units': header['Dim-%i_Units' % (i + 1)][0] if header['Dim-%i_UnitsLength'%(i+1)]>0 else 'a.u.',
```
francisco-dlp: I had the impression that the units were always in SI unless specified. Is that true in your experience? If yes, coult 'a.u.' be replaced with 'meters'?

hyperspy/io_plugins/fei.py
```diff
@@ -534,7 +572,11 @@ def ser_reader(filename, objects=None, verbose=False, *args, **kwds):
         if axis['units'] == 'meters':
             axis['units'] = 'nm'
             axis['scale'] *= 10 ** 9
-
+        elif axis['units'] == '1/meters':
+            axis['units'] = '1/nm'
+            axis['scale'] /= 10 ** 9
+        else:
```
francisco-dlp: This is not required

hyperspy/io_plugins/fei.py
```diff
@@ -580,13 +622,48 @@ def ser_reader(filename, objects=None, verbose=False, *args, **kwds):
         'mapping': mapping}
     return dictionary
 
-
-def get_mode(mode):
+def guess_units_from_mode(objects_dict, header, verbose=False):
```
francisco-dlp: I think that this isn't going to work because the xmls are not stored in the order of the ser file numbers. Instead they use UUIDs to match the ser files to the xml objects. In order to match them we need to know how to generate those UUIDs.

francisco-dlp:
@ericpre, very nice work. We're about to release 0.8.3 and, given that this is enhancing the FEI reading support, if you could address the issues that I've pointed out, rebase on 0.8.x and pull request it to that 0.8.x, we could hold the release (if the other developers agree) until this is ready to merge as long as it doesn't take more than a few days.

hyperspy/io_plugins/fei.py
ericpre: Yes, the units seem to be in SI, but here 'a.u.' is for the case of image stacks (in this case there is no units and UnitLength is then 0). However, it should be an unit of time, but I don't know how to get the scale.
I don't know what to put instead of 'a.u.', any better suggestions?

ericpre:
Thanks for the review! I fixed the issue and remove my comments about the xml data which doesn't seem to be relevant. I also added more comments on the workaround to guess the units.

Happy to make a pull request to 0.8.x. Please let me know if everything is fine and if you want me to do it!

hyperspy/io_plugins/fei.py
```diff
@@ -503,17 +532,22 @@ def ser_reader(filename, objects=None, verbose=False, *args, **kwds):
...
-                        'units': header['Dim-%i_Units' % (i + 1)][0],
+                        # for image stack, the UnitsLength is 0 (no units)
+                        'units': header['Dim-%i_Units' % (i + 1)][0] if header['Dim-%i_UnitsLength'%(i+1)]>0 else 'a.u.',
                         'size': header['Dim-%i_DimensionSize' % (i + 1)][0],
                     })
                 array_shape.append(header['Dim-%i_DimensionSize' % (i + 1)][0])
+        units = "Unknown"
+        if objects is not None:
+            objects_dict = convert_xml_to_dict(objects[0])
+            units = guess_units_from_mode(objects_dict, header)
```
francisco-dlp: Then, if we cannot guarantee that the xml chunck belongs to a given ser file, guessing the units in this way won't work, isn't it? If that's the case, I suggest to comment out these lines and explain in a comment that, once we find out how to match xml chunks to ser files using the UUID, this code can be re-enabled.

francisco-dlp:
Thanks @ericpre. Yes, it'll be great if you could address the remaining issue and submit this to 0.8.x.

hyperspy/io_plugins/fei.py
ericpre: Ok, this is right but it actually seems to work!
Even if I don't exactly know why, I guess this is mainly because the information retrieve from the xml and that is used by  "guess_units_from_mode" is only the mode, which is present in almost all xml chunks, except for spectrum image, where there are more xml chunks than ser files and the additional ones doesn't contains any information.
Since, this function is useful to get the unit particularly in case of diffraction, I will find a workaround to consider the case where the xml chunk is empty and then return "Unknow" as unit.

francisco-dlp: What concerns me is what happens if you get an emi file that contains e.g. EDS SI, EELS SI, HAADF, BF. The HAADF. If we cannot match xml to ser files, won't this be an issue?

ericpre:
One thing that I forgot to mention:
I noticed that the scale of 'y' axis of spectrum images (all dataset with 2 dimension as navigation dimension) is negative. Is there any reason to have a negative scale for 'y' and a positive scale for 'x'? Should we change this behavior?

francisco-dlp:
We could do that, but then we'll have to swap the data in the y-axis too.

ericpre:
humm... if the data are swapped, the image will look upside down, right? Not sure if this is good, this may be a bit confusing!
Actually, why does it needs to be swap? Does it have to do with how matplotlib plot things?

ericpre:
Currently, the first xml is pass to this function and I couldn't find a file with a first xml chunk which doesn't have the "mode" metadata.
If the xml chunk is empty, I added an exception and the function will return 'Unknown' unit, so even if there are emi files which doesn't have the "mode" in the first xml chunk, it should not break the reader.

hyperspy/io_plugins/fei.py
```diff
@@ -580,13 +619,53 @@ def ser_reader(filename, objects=None, verbose=False, *args, **kwds):
         'mapping': mapping}
     return dictionary
 
-
-def get_mode(mode):
+def guess_units_from_mode(objects_dict, header, verbose=False):
+    # assuming that for an image stack, the UnitsLength of the "3rd" dimension
+    # is 0
+    isImageStack = (header['Dim-1_UnitsLength'][0] == 0)
```
francisco-dlp: There are ser files that don't contain "Dim" information in the header, so this line may fail on them. See #835

hyperspy/io_plugins/fei.py
```diff
@@ -580,13 +619,53 @@ def ser_reader(filename, objects=None, verbose=False, *args, **kwds):
...
+    # in case the xml file doesn't contains the "Mode" return "Unknow" unit
+    try:
+        mode = objects_dict.ObjectInfo.ExperimentalDescription.Mode
+    except AttributeError: # in case the xml chunk doesn't contain the Mode
+        return 'Unknown'
+    isCamera = ("CameraNamePath" in objects_dict.ObjectInfo.AcquireInfo.keys())
+    # Workaround: if this is not an image stack and not a STEM image, then we 
+    # assume that it should be a diffraction
+    isDiffractionScan = (header['Dim-1_DimensionSize'][0] > 1 and not isImageStack)
```
francisco-dlp: The same for this one.

hyperspy/io_plugins/fei.py
```diff
@@ -580,13 +619,53 @@ def ser_reader(filename, objects=None, verbose=False, *args, **kwds):
...
+        print "------------"
+        print objects_dict.ObjectInfo.AcquireInfo
+        print "mode", mode
+        print "isCamera:", isCamera
+        print "isImageStack:", isImageStack
+        print "isImageStack:", isDiffractionScan
+        print "------------"
+    if 'STEM' in mode:
+        # data recorded in STEM with a camera, so we assume, it's a diffraction
+        if isCamera:
```
francisco-dlp: What about `if isCamera or isDiffraction:`?

hyperspy/io_plugins/fei.py
```diff
@@ -580,13 +619,53 @@ def ser_reader(filename, objects=None, verbose=False, *args, **kwds):
...
-def get_mode(mode):
+def guess_units_from_mode(objects_dict, header, verbose=False):
+    # assuming that for an image stack, the UnitsLength of the "3rd" dimension
+    # is 0
+    isImageStack = (header['Dim-1_UnitsLength'][0] == 0)
+    # in case the xml file doesn't contains the "Mode" return "Unknow" unit
+    try:
+        mode = objects_dict.ObjectInfo.ExperimentalDescription.Mode
+    except AttributeError: # in case the xml chunk doesn't contain the Mode
+        return 'Unknown'
```
francisco-dlp: Currently it would set the units to "meters" which is correct in most cases, and when it is not (diffraction) it is obvious that it is incorrect, so it is not a big deal. Return meters then?

hyperspy/io_plugins/fei.py
```diff
@@ -580,13 +619,53 @@ def ser_reader(filename, objects=None, verbose=False, *args, **kwds):
...
+        elif isDiffractionScan:
+            return "1/meters"
+        else:
+            return "meters"
+    elif 'Diffraction' in mode:
+        return "1/meters"
+    elif 'Image' in mode:
+        return "meters"
+    else:
+        return 'Unknown'
```
francisco-dlp: Same as above: currently it would set the units to "meters" which is correct in most cases, and when it is not (diffraction) it is obvious that it is incorrect, so it is not a big deal. Return meters then?

francisco-dlp:
It'll be nice to include this in 0.8.3. Could you pull request it there?

francisco-dlp:
All the new files that you're providing for testing must be included in the `hyperspy/setup.py`, see e.g. https://github.com/hyperspy/hyperspy/blob/master/setup.py#L155 and https://github.com/hyperspy/hyperspy/blob/master/setup.py#L155

francisco-dlp:
Regarding the negative scale question, for images in hyperspy we use the convention of displaying the origin at the top. FEI's convention is to use negative increments for y when scanning from top to bottom, and that's why the scale is negative. I think that it is fine as it is, why would you like to change it?

ericpre:
Thanks!
I add an exception and return 'meters', when the ser file doesn't contain the information, not very elegant but will do the job. What kind of acquisition are these files? Maybe I can try to do a dummy small acquisition on a TEM to add to the test?

ericpre:
it just that I was surprise to see a negative scale, nothing else. if I understand correctly, this is only to be consistent between TIA and hyperspy? If so, I think this is fine as it is.

I will do a pull request to 0.8.x when everything thing is finished here. I am not completely familiar with git, so I will do it at the end.

francisco-dlp:
I think that you get those sort of files when you take e.g. oblique line profiles, videos and even when acquiring standard EDXs SIs in some microscope configurations. An user sent me one of those files recently and that's how I found out about the issue.

francisco-dlp:
The code looks good to me know. I'll test it more thoroughly once you send it to 0.8.x as this will involve merging the changes in #835

vidartf:
PS: Note that resizable widgets does not work with negative axis scales currently. RectangleROI's attributes `top`/`bottom`/`left`/`right` will also be reversed with negative scales, as `left`/`top` always refer to the smallest value.

francisco-dlp:
@vidartf, since there is nothing wrong about a negative scale—aren't the widgets the thing to fix them?

vidartf:
Sure, I just wanted to point it out. I discovered it myself the other day when working with FEI files, but simply worked around it by setting it as positive manually.

francisco-dlp:
Could you then open an issue so that we don't forget about it?

ericpre:
I tried to acquired oblique line profiles on different systems and I always get an header in the ser files. Instead, I added a test to cover missing information in the xml chunks or in the ser header.

francisco-dlp:
Nice. If you don't mind it, I'll try to rebase this on the 0.8.x branch. I'll let you know how it goes.

ericpre:
Yes, please, because I tried quickly and I am not getting there!!

francisco-dlp:
@ericpre, I have just rebased this on the current maintenance branch, 0.8.x, see #890. I have also committed some small fixes:

* Revert the changes to the mode that it is stored in metadata as your improved version is not compatible with our metadata specifications.
* I have replaced "a.u." with "Unknown", because most likely the units aren't arbitrary, we simple ignore what they are.
* Some other minor styling fixes.

Could you have a look  at #890 as let us know if you're happy with it?

ericpre:
Thanks for doing the rebase! I am fine with the changes.

For the mode, would it be of interest to add the 'diffraction' or 'image' information somewhere in the metadata? This is maybe out of scope of this PR?

francisco-dlp:
Great. Soon we'll have a `DiffractionImage` class. Then, we'll modify the reader to return a `DiffractionImage` instance when reading diffraction data. For that we'll probably go the way that you propose and store `diffraction` as the signal type. But we have to decide the actual string: "Diffraction", "ElectronDiffraction",  "TEMDiffraction", "STEMDiffraction", "CBED"? Therefore, I agree that it'll be better to leave this for another PR.   @dnjohnstone, any opinion on this? 

@ericpre, thank you very much for this great contribution!

francisco-dlp:
@ericpre, once #890 is merged and HyperSpy 0.8.3, your name will appear among the authors of HyperSpy in [Zenodo](https://zenodo.org/record/28025?ln=en#.VtbzAM1amV4). Zenodo takes your name and affiliation from the github profile. Could you update your github profile with this information if you want your contribution to be credited to your name? The release of HyperSpy 0.8.3 is imminent, so you'll have to hurry up!

dnjohnstone:
@ericpre @francisco-dlp Yes I have an ElectronDiffraction class in a branch https://github.com/dnjohnstone/hyperspy/tree/NEW_SED_class will make a pull request after I have time to make tests and make things sufficiently general.

dnjohnstone:
To perhaps answer your question about the actual string more directly:

Diffraction - would have to be general to include say X-ray powder diffraction if someone eventually does that.

ElectronDiffraction - potentially would need to include EBSD or TKD

TEMDiffraction - if it's TEM it's probably not multidimensional

STEMDiffraction - is ok, but I think ScanningElectronDiffraction is establishing as a technique in it's own right for spot based work.

CBED - probably eventually one would like to distinguish but I think most initial multidimensional stuff will treat data as spot patterns because the CBED structure is complex

This was why I went for SEDPattern in the branch I made.

francisco-dlp:
Thanks @dnjohnstone, we can discuss this once you PR you branch.

dnjohnstone:
I was also hoping that we might have a decision/plan for splitting Hyperspy #821 before adding a whole new signal class...

francisco-dlp:
Splitting HyperSpy is not a small task so it's definitely not going to happen for HyperSpy 0.9. If people from other fields show an interest in contributing, then we'll have a motivation to actually do it. In any case, I don't see this happening in the close future, unless someone with the right amount of time and knowledge steps forward to tackle this. In other words, I wouldn't wait for the split to happen to PR a new signal class.

ericpre:
@francisco-dlp do we need to get your changes into this branch now to merge this PR with master, or this can be done later?

francisco-dlp:
No, there is no need. Once the other one is merged this one will be automatically merged. I'll make sure that it is yours and not mine the one that we link to in the docs.

ericpre:
OK, great. Thanks.

francisco-dlp:
Resubmission #890 merged.

PR #889, opened 2016-03-02
Add API to manually add and remove X-ray markers

vidartf:
Adds an interface to add, track, and remove xray markers to an EDSSignal.

Useful for example for HyperSpyUI.


francisco-dlp:
Until now our approach to testing the plotting features has been not testing at all. However, I think that some lazy testing is better than none. So, what about some tests that simply call the functions? These tests won't detect if the right things are being plotted, but at least they'll fail if something gets broken.

vidartf:
Python3 fixes added, and basic tests implemented.

vidartf:
PS: Once this PR is merged, HyperSpyUI should be compatible with HyperSpy master, allowing it to no longer be dependent on my temporary branch.

PR #897, opened 2016-03-04
Fix setup

francisco-dlp:
This includes two setup fixes:

* Print a helpful error message when installing using the wrong Python version (i.e. Python < 3)
* Correctly define install requirements so that pip install them automatically.


francisco-dlp:
I have commited similar fixes to the 0.8.3 branch, see 1d6212ffc43438abdf7b97405245a3331250eeeb and 75d20225749781f89b223f1d814db5c23c6ae15b. I'll create new installers with the updated setup for 0.8.3 and 0.8.4.

francisco-dlp:
I'll merge this because I'm in a rush: hyperspy is broken in pypi at present.

Issue #900, opened 2016-03-04
`stack` is defined twice in io.py


bm424:
The `stack` variable is being used both as a function call imported from `hyperspy.misc.utils` and a boolean variable passed as a parameter into `load`.

PR #901, opened 2016-03-04
Bug fix load stack

bm424:
Fix for the bug in loading where `stack` had conflicting defintitions.


bm424:
#900

francisco-dlp:
This bug is not present in 0.8.x, how is it possible?

to266:
I kind-of-fixed it when porting to python3 most likely.

to266:
it may not be true, I just remember having to sort out some merge conflicts, and it involved this issue. Could have been a different branch too, but any way I see no reason not to merge this

francisco-dlp:
Yes, let's merge it. However, let's not label the issue as a bug as there was not bug in the 0.8.3 release.

PR #905, opened 2016-03-07
Remove preferences.General.default_toolkit, Fixes #903

### LINKED ISSUE #903: importing hyperspy with no toolkit
the usual `import hyperspy.api as hs` crashes if run headless (on python2, did not test for python3)

francisco-dlp:
Fixes #903

### DIFF
--- a/doc/user_guide/getting_started.rst
+++ b/doc/user_guide/getting_started.rst
@@ -53,6 +53,18 @@ but some features such as navigation sliders may be missing.
         executed after importing hyperspy and qt must be the default hyperspy
         backend.
 
+.. NOTE::
+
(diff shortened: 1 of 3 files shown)


to266:
Looks like it does not work

francisco-dlp:
I forgot to mention that, for this to work, the mpl backend must something that doen't require a toolkit e.g. "Agg" so the following should work in a headless system:

```python

import matplotlib
matplotlib.rcParams["backend"] = "Agg"
import hyperspy.api as hs

```

to266:
Ok, that works. I would add instructions to the user-guide for this

francisco-dlp:
Done.

Issue #800, opened 2015-12-15
EDS Quantification in HyperSpy


dnjohnstone:
This issue is raised to coordinate efforts to develop EDS quantification approaches in HyperSpy, which I know a few people are working on in various ways.

There also exists a substantial amount of low hanging fruit in this area for people looking to start contributing to HyperSpy - this is mostly because @pburdet did a huge amount of work and there remain substantial chunks that are very nearly finished, but need a small push over the final line into the main HyperSpy project. To summarise:

#474 Adds functionality to simply simulate an EDS model (proper scattering cross sections need adding)

#482 Adds methods associated with detector efficiency, which will build on #471 that is almost ready.

There is also much functionality in the edslab branch created by @pburdet

Other known points to be incorporated when someone is inspired are:

1. Spectral peak identification (cf #745)
2. Methods to facilitate calibration of k-factors using a standard and transfer to a dictionary for later use.
3. k-factor ratio methods
4. The addition of poissonian noise to EDS models (some progress was made there by Pierre)
5. Plotting elemental maps made using all lines integrated together (a path to this is set out in #786)
6. Implementation of the Zeta-factor method.

Some interested parties might be: @siriagus  @k8macarthur

If anyone gets going on any of this stuff or related - do make sure people know!

siriagus:
@pburdet has implemented a working version (without absorption correction/iteration) of the zeta factor method for quantification in the edslab branch. I want to tidy this up/make it consistent with the rest of the code base and eventually improve it by including absorption effects. I pulled the branch from Pierre's fork, but the amount of merge conflicts are quite overwhelming. All I really need is a small functions from [hyperspy.misc.eds.utils.py](https://github.com/pburdet/hyperspy/blob/edslab/hyperspy/misc/eds/utils.py#L2442).
I could of course just rewrite it in a new branch (since it's only 8-15 lines), but the existing version is quite neat/easy to read and I want Pierre to get credit for his contribution.

I need help to figure out what's the best way to continue Pierre's work on the zeta factor method without having to sort out all of the other eds code at the same time.

vidartf:
Looking at the history, I can only see two commits related to the zeta method, so you could try to start a fresh branch from master and then `cherry-pick` the two commits: 5f3c504a5ae03996048787c80b52f150df017fc7, 54dcfbbdb2ad2bd1a2a1f7400c30aa256a8a8213

dnjohnstone:
[...]

PR #913, opened 2016-03-14
EDX Quantification (ZFactors + Cross Sections): Update to include zeta-factors and cross sections in the EDX_TEM quantification method.

k8macarthur:
Update to include zeta-factors and cross sections in the EDX_TEM quantification method. Both methods use pure element standards for their quantification making it a simpler means for standards based quantification. The user now has a choice of three methods for quantification allowing for comparison between techniques.


francisco-dlp:
@k8macarthur, could you update your this with the current master?

k8macarthur:
Should be done now. :)

hyperspy/_signals/eds_tem.py
```diff
@@ -325,12 +331,36 @@ def quantification(self,
...
+        if method == 'CL':
+            composition.data = utils_eds.quantification_cliff_lorimer(
+                composition.data, kfactors=factors,
+                mask=navigation_mask) * 100.
+        elif method == 'zeta':
+            results = utils_eds.quantification_zeta_factor(
+                composition.data, zfactors=factors,
+                dose=self._get_dose(method))
+            composition.data = results[0] * 100.
+            mass_thickness = intensities[0]
```
siriagus: Should be `mass_thickness = intensities[0].deepcopy()`, if not, intensity[0] will be overwritten in the next line (as mass_thickness points to the same array as intensities[0] in the current code).

hyperspy/_signals/eds_tem.py
```diff
@@ -258,28 +260,32 @@ def get_calibration_from(self, ref, nb_pix=1):
...
-            order of the X-ray lines. eg. kfactors =[0.982, 1.32, 1.60] for
-            ['Al_Ka','Cr_Ka', 'Ni_Ka'].
+        method: 'CL' or 'zeta' or 'cross_section'
+            Set the quantification method: Cliff-Lorimer, zeta factor, or
+            ionization cross sections
+        factors: list of float
+            The list of kfactors, zfactors or cross sections in same order as
+            intensities. Note that intensities provided by Hyperspy are sorted
+            by the aplhabetical order of the X-ray lines.
+            eg. kfactors =[0.982, 1.32, 1.60] for ['Al_Ka','Cr_Ka', 'Ni_Ka'].
```
siriagus: Should be ` eg. factors=[0.982, 1.32, 1.60] for ['Al_Ka','Cr_Ka', 'Ni_Ka'].`

hyperspy/_signals/eds_tem.py
```diff
@@ -258,28 +260,32 @@ def get_calibration_from(self, ref, nb_pix=1):
...
                        navigation_mask=1.0,
                        closing=True,
                        plot_result=False,
                        **kwargs):
         """
-        Quantification of intensities to return elemental composition
-
-        Method: Cliff-Lorimer
+        Quantification using Cliff-Lorimer, the zeta factor method, or
+        ionization cross_sections.
```
siriagus: cross sections

hyperspy/_signals/eds_tem.py
```diff
@@ -325,12 +331,36 @@ def quantification(self,
...
+            number_of_atoms = utils.stack(intensities)
+            number_of_atoms.data = results[1]
+            number_of_atoms = number_of_atoms.split()
+        else:
+            raise Exception ('Please specify method for quantification, as CL, zeta or cross_section')
         composition = composition.split()
         if composition_units == 'atomic':
-            composition = utils.material.weight_to_atomic(composition)
+            if method == 'cross_section':
+                composition == composition
```
siriagus: Use of conditional operator not equal. Suggest deleting this and the line above at it serves no purpose. See suggestion below.

hyperspy/_signals/eds_tem.py
```diff
@@ -325,12 +331,36 @@ def quantification(self,
...
+            else:
```
siriagus: Suggested changed to: `if method != 'cross_section':`

Actually, when I am using the zeta factor method, the factors are calculated for atomic as well. If I then use the 'atomic' as input parameter, it converts from weight to atomic (but it shouldn't as input zeta factors corresponds to atomic, just as in the case with the cross section method). I know that wt% is the convention for the Cliff-Lorimer method, however I would suggest using at% for the zeta factor method as well. However, I am no expert in the field so I would like to hear what others think.

hyperspy/_signals/eds_tem.py
```diff
@@ -325,12 +331,36 @@ def quantification(self,
...
+            composition.data = utils_eds.quantification_cliff_lorimer(
+                composition.data, kfactors=factors,
+                mask=navigation_mask) * 100.
+        elif method == 'zeta':
+            results = utils_eds.quantification_zeta_factor(
+                composition.data, zfactors=factors,
+                dose=self._get_dose(method))
+            composition.data = results[0] * 100.
+            mass_thickness = intensities[0]
+            mass_thickness.data = results[1]
```
siriagus: Add line `mass_thickness.metadata.General.title = "Mass thickness"`

hyperspy/_signals/eds_tem.py
tjof2: Also a tiny inconsistency in spacing after commas... `['Al_Ka', 'Cr_Ka', 'Ni_Ka']`

tjof2: And a typo of "alphabetical"

dnjohnstone:
Is there any chance this could be merged up to the Python 3 version of the code please? The current diff is a bit too long to review. Sorry that change happened part way through this pull request, unfortunate timing.

hyperspy/_signals/eds_tem.py
k8macarthur: My understanding from Masashi's orginal paper is that the composition came out in weight percent. Therefore I was assuming weight percent would be preferable for most people. However as I'm not really using z-factors I'm happy to be corrected. I think Pierre's original code for including the z-factors was with weight percent.

k8macarthur:
@dnjohnstone  I think the update to Python 3 should now be done. Although I'm struggling to install hyperspy in Python 3 in order to check if it still installs correctly.

doc/user_guide/eds.rst
```diff
@@ -512,11 +512,19 @@ The background can be subtracted from the X-ray intensities with the :py:meth:`~
 Quantification
 --------------
 
-.. versionadded:: 0.8
+Hyperspy now includes three methods for EDS quantification; Cliff-Lorimer, zeta-factors and ionization cross sections.
+
+Quantification can be applied to the background subtracted intensities with the :py:meth:`~._signals.eds_tem.EDSTEMSpectrum.quantification` method. These instensities are a stack of images for each element which can be extracted using :py:meth:`~._signals.eds.EDSSpectrum.get_lines_intensity`. The quantification method, needs be specified as either 'CL', 'zeta', or 'cross_section'. If no method is specified the function will raise an exception.
```
dnjohnstone: Perhaps word to emphasise that if you want to use background subtracted intensities, you must first do a background subtraction... We don't want people thinking that's done by default.

doc/user_guide/eds.rst
```diff
@@ -525,31 +533,49 @@ Quantification can be applied from the intensities (background subtracted) with
     >>> kfactors = [1.450226, 5.075602] #For Fe Ka and Pt La
     >>> bw = s.estimate_background_windows(line_width=[5.0, 2.0])
     >>> intensities = s.get_lines_intensity(background_windows=bw)
-    >>> weight_percent = s.quantification(intensities, kfactors, plot_result=True)
+    >>> atomic_percent = s.quantification(intensities, method='CL', kfactors)
     Fe (Fe_Ka): Composition = 4.96 weight percent
     Pt (Pt_La): Composition = 95.04 weight percent
```
dnjohnstone: if this is now going to give atomic percent as default then it the line above should be updated to give the appropriate atomic percentages rather than weight percentages

dnjohnstone:
Thanks @k8macarthur I'll go through line-by-line here with some initial comments, then check more carefully this weekend... If I give a comment that you'd like me to fix for you, then say - but we usually operate this way so that authorship remains with you in record stored on github.

doc/user_guide/eds.rst
```diff
@@ -525,31 +533,49 @@ Quantification can be applied from the intensities (background subtracted) with
...
+If these are not set the code will produce an error stating which parameter has been forgotten.
+
+The cross section method needs the beam_current, real_time and pixel width in order to provide an accurate quantification, which can be set using:
+
+..code-block: python
+
+    >>>> s.axes_manager[0].scale = 0.025
+    >>>> s.axes_manager[1].scale = 0.025
+
+If the pixel width is not set, the code will still run with the default value of 1nm with a warning message to remind the user that this is the case.
```
dnjohnstone: I think it would probably be good to create a method called something like 'set_quantification_parameters' through which these things can be set...

The background code will look like what you've written here but the user will then just call that method and specify their parameters (perhaps also the method could be set at this point?) - this will look much cleaner for end users an is easy to implement.

More fully, you could make this raise a UI window so that it's even more user friendly -- to do that you could copy code from 'set_microscope_parameters' in the TEMEDX signal file.

hyperspy/_signals/eds_tem.py
```diff
@@ -18,8 +18,9 @@
 
 
 import traits.api as t
+import traitsui.api as tui
```
dnjohnstone: This line should be deleted - it's from an old merge

dnjohnstone:
@k8macarthur More importantly -- this code is currently breaking Travis (see the bottom of this conversation) that's because it doesn't pass the nosetests, which you can run locally on your machine to check by the way... If you click on details associated with Travis at the bottom off this page it tells you what errors have been raised in the nosetests. Much of it is Py2--Py3 incompatibility e.g. needing 'print()' rather than 'print'... Ideally you can go through these reports yourself and let the robots correct you!

hyperspy/_signals/eds_tem.py
```diff
@@ -379,10 +424,10 @@ def vacuum_mask(self, threshold=1.0, closing=True, opening=False):
         """
         from scipy.ndimage.morphology import binary_dilation, binary_erosion
         mask = (self.max(-1) <= threshold)
-        if closing:
+        if closing == True:
```
dnjohnstone: keep this as just if closing:

(that's an update made somewhere in between).

hyperspy/_signals/eds_tem.py
```diff
@@ -379,10 +424,10 @@ def vacuum_mask(self, threshold=1.0, closing=True, opening=False):
...
             mask.data = binary_dilation(mask.data, border_value=0)
             mask.data = binary_erosion(mask.data, border_value=1)
-        if opening:
+        if closing == False:
```
dnjohnstone: similarly, keep this as just if opening:

hyperspy/_signals/eds_tem.py
```diff
@@ -343,11 +372,27 @@ def quantification(self,
...
+            self.metadata.set_item("Sample.mass_thickness", mass_thickness)
+        if method == 'zeta':
+            return composition, mass_thickness
+        elif method == 'cross_section':
+            return composition, number_of_atoms
+        else:
+            return composition
 
-    def vacuum_mask(self, threshold=1.0, closing=True, opening=False):
+    def vacuum_mask(self, threshold=1.0, closing=True):
```
dnjohnstone: keep this as the master version so make it...

def vacuum_mask(self, threshold=1.0, closing=True, opening=False):

hyperspy/tests/signal/test_eds_tem.py
```diff
@@ -222,6 +257,8 @@ def test_with_signals_examples(self):
         s = EDS_TEM_Spectrum()
         np.allclose(np.array([res.data for res in s.get_lines_intensity()]),
                     np.array([3710, 15872]))
+if __name__ == '__main__':
+   nose.run(argv=[sys.argv[0], sys.modules[__name__].__file__, '-v'])
```
dnjohnstone: I don't think the above 2 lines should be here anymore

hyperspy/_signals/eds_tem.py
```diff
@@ -483,3 +528,67 @@ def create_model(self, auto_background=True, auto_add_lines=True,
...
+                pixel2 = self.axes_manager[1].scale
+            if self.axes_manager[0].scale == 1 or self.axes_manager[1].scale == 1:
+                    warnings.warn('Please note your pixel width is set to the \
+default value of 1nm. The function will still run. However, if this is \
+incorrect you should consider changing this using the axes_manager.gui() or \
+axes_manger[0].scale functions.')
+            area = pixel1 * pixel2
+            return (real_time * beam_current * 1e-9) /(constants.e * area)
+        elif method == 'zeta':
+            print real_time * beam_current * 1e-9 / constants.e
```
dnjohnstone: this print needs some brackets now

dnjohnstone:
@francisco-dlp Travis seems to be complaining at least in part about the test_dm3 part of the code? I don't think Kate's touched that so is that another issue or do I misunderstand?

doc/user_guide/eds.rst
siriagus: I suggest editing set_microscopy_parameters as I did in [my local branch] (https://github.com/siriagus/hyperspy/commit/2154204d07eccc206c484e4aeb0855530a4c1f94)  (you could simply cherry-pick it, but ignore the edits to _are_microscopy_parameters, I haven't come around to remove that yet)instead of creating an extra function. You could probably just cherry-pick that commit.

When it comes to raising a UI window, see Fransisco's comment [here](https://github.com/hyperspy/hyperspy/issues/800#issuecomment-190764358).

doc/user_guide/eds.rst
k8macarthur: Should have already been changed. I'll see if it updates the next time a sync.

doc/user_guide/eds.rst
k8macarthur: Good point, I'll add the get_intensities function in again just to make it explicitly clear.

doc/user_guide/eds.rst
k8macarthur: So this is the one section that I wasn't sure what the best option was to do in the end. I followed @francisco-dlp's guidelines one trying to get people to set them and then raising errors if they're not set. It gets a little messy because of the default setting for pixel width. But I'm open to suggestions for alternatives.

dnjohnstone: Sorry for confusion - forgot about that previous discussion. To me there are two slightly different UI possibilities:

1. when set_microscope_parameters() is called at present - that raises a UI (I don't use it except when playing with data for the reasons Francisco mentions, but it exists and works that way and I think it's natural to be able to specify these additional parameters in a consistent way.

2. I think the original discussion might have had in mind that when calling s.quantification() a UI could be raised if necessary parameters weren't specified. This probably a bad idea for the reasons Francisco mentions.

In my opinion - option 1 should be implemented for consistency throughout the code base and because many users probably like that little UI for parameter specification. Option 2 is a no-no though I think.

@francisco-dlp  or @vidartf may have better opinions

dnjohnstone:
@k8macarthur I think that last commit with assertTrue might be a little wrong -- I think nt.assert_true should be right in all those places, that's what's breaking Travis at the moment

doc/user_guide/eds.rst
```diff
@@ -525,31 +533,49 @@ Quantification can be applied from the intensities (background subtracted) with
...
 
 .. code-block:: python
 
     >>> # With s, intensities and kfactors from before
-    >>> s.quantification(intensities, kfactors, plot_result=True,
-    >>>                  composition_units='atomic')
+    >>> s.quantification(intensities, method='CL',kfactors,
+    >>>                  composition_units='weight')
     Fe (Fe_Ka): Composition = 15.41 atomic percent
     Pt (Pt_La): Composition = 84.59 atomic percent
```
dnjohnstone: These ones have been updated to atomic percent, but the ones above at line 539 haven't -- simple copy paste should do.

k8macarthur: Ok got it. I think they're just the wrong way around. :)

hyperspy/tests/signal/test_eds_tem.py
```diff
@@ -135,7 +135,7 @@ def test_get_calibration_from(self):
 class Test_quantification:
 
     def setUp(self):
-        s = EDSTEMSpectrum(np.ones([2, 1024]))
+        s = EDSTEMSpectrum(np.ones([2, 2, 1024]))
         energy_axis = s.axes_manager.signal_axes[0]
```
dnjohnstone: Is there a reason why you've made this a 'spectrum image' i.e. 2 navigation dimensions, rather than a line scan as it was before? As far as I can see all you've ended up doing is doubling the number of numbers to be compared in the tests, but haven't actually added any functionality to the tests?

hyperspy/tests/signal/test_eds_tem.py
```diff
@@ -177,7 +212,7 @@ def test_quant_zeros(self):
                             category=RuntimeWarning):
             quant = utils_eds.quantification_cliff_lorimer(
                 intens, [1, 1, 3]).T
-        nt.assert_true(np.allclose(
+        nt.assertTrue(np.allclose(
```
dnjohnstone: I think this should still be nt.assert_true?

hyperspy/tests/signal/test_eds_tem.py
```diff
@@ -157,15 +159,48 @@ def setUp(self):
...
+        factors = [3, 5]
+        intensities = s.get_lines_intensity()
+        res = s.quantification(intensities, method, factors)
+        (np.allclose(res[1][1].data, np.array(
+            [[ 16.02176621,  16.02176621],
+            [ 16.02176621,  16.02176621]])), atol=1e-3)
+        nt.assert_true(np.allclose(res[1][1].data, np.array(
+           [[ 19.8669901,  19.8669901],
+           [ 19.8669901,  19.8669901]])), atol=1e-3)
+        nt.assertTrue(np.allclose(results1[0].data, np.array(
```
dnjohnstone: I think this should be nt.assert_true?

hyperspy/tests/signal/test_eds_tem.py
k8macarthur: Making it a spectrum image means there is now a pixel width in both dimensions. Therefore if it's not a 2x2 square I believe the cross section quantification will fail, because the pixel width is taken from the step in the axis.

hyperspy/tests/signal/test_eds_tem.py
k8macarthur: Agreed. Just trying to work out why my unit tests were failing.

hyperspy/tests/signal/test_eds_tem.py
dnjohnstone: Ah ok, that makes sense - if the zeta method can work on a line scan should it not be possible for cross-section method to work on a line scan too? I don't know if that's a stupid question, sorry if it is :)

hyperspy/_signals/eds_tem.py
```diff
@@ -325,12 +330,35 @@ def quantification(self,
...
+        else:
+            raise Exception ('Please specify method for quantification, as CL, zeta or cross_section')
         composition = composition.split()
         if composition_units == 'atomic':
-            composition = utils.material.weight_to_atomic(composition)
+            if method != 'cross_section':
+                composition = utils.material.weight_to_atomic(composition)
+        else:
+            if method == 'cross_section':
+                composition = util.material.atomic_to_weight(composition)
```
dnjohnstone: should be utils.material.atomic_to_weight(composition)

hyperspy/_signals/eds_tem.py
dnjohnstone: just to bump this, probably a fair point from andreas.

hyperspy/tests/signal/test_eds_tem.py
k8macarthur: Not necessarily, it goes back to the two different definitions of dose again. Both are area independent values. However cross sections make use of a dose per unit area (hence the need for a pixel area). This essentially allows for standards samples to be take at a different magnification from the real samples.

hyperspy/tests/signal/test_eds_tem.py
```diff
@@ -157,15 +159,48 @@ def setUp(self):
...
+
+    def test_quant_cross_section(self):
+        s =self.signal
+        method = 'cross_section'
+        factors = [3, 5]
+        intensities = s.get_lines_intensity()
+        res = s.quantification(intensities, method, factors)
+        (np.allclose(res[1][1].data, np.array(
+            [[ 16.02176621,  16.02176621],
+            [ 16.02176621,  16.02176621]])), atol=1e-3)
```
dnjohnstone: this line is missing an nt.assert_true at the beginning of it i.e. before the first bracket

hyperspy/tests/signal/test_eds_tem.py
dnjohnstone: Currently it's the missing nt.assert_true a couple of lines up that I've commented, which is breaking it.

hyperspy/_signals/eds_tem.py
```diff
@@ -483,3 +527,66 @@ def create_model(self, auto_background=True, auto_add_lines=True,
...
+                raise Exception ('Please note that your real time is set to\
+the default value of 0.5s. The function will still run. However, if this is \
+incorrect you should consider changing it using \
+self.metadata.Acquisition_instrument.TEM.Detector.EDS.real_time .')
+        if method == 'cross_section':
+            if area == 'auto':
+                pixel1 = self.axes_manager[0].scale
+                pixel2 = self.axes_manager[1].scale
+            if self.axes_manager[0].scale == 1 or self.axes_manager[1].scale == 1:
+                    warnings.warn('Please note your pixel width is set to the \
```
dnjohnstone: This line is overindented should be in line with pixel2 above

hyperspy/_signals/eds_tem.py
```diff
@@ -483,3 +527,66 @@ def create_model(self, auto_background=True, auto_add_lines=True,
...
+default value of 1nm. The function will still run. However, if this is \
+incorrect you should consider changing this using the axes_manager.gui() or \
+axes_manger[0].scale functions.')
+            area = pixel1 * pixel2
+            return (real_time * beam_current * 1e-9) /(constants.e * area)
+        elif method == 'zeta':
+            return real_time * beam_current * 1e-9 / constants.e
+        else:
+            raise Error('no method provided')
```
dnjohnstone: This isn't the correct way to raise an error - although I can't actually remember how to do it properly right now....

hyperspy/_signals/eds_tem.py
```diff
@@ -483,3 +527,66 @@ def create_model(self, auto_background=True, auto_add_lines=True,
...
+            if 'beam_current' in self.metadata.Acquisition_instrument.TEM == False:
+                raise Exception ('It is not possible to carry out EDX \
+quantification_cross_section without a beam_current please set one using \
+self.metadata.Acquisition_instrument.TEM.beam_current and run quantification \
+again.')
+            else:
+                beam_current = parameters.beam_current
+        if real_time == 'auto':
+            real_time = parameters.Detector.EDS.real_time
+            if 'real_time' in self.metadata.Acquisition_instrument.TEM.Detector.EDS == False:
```
dnjohnstone: should be an 'is' rather than '=='

hyperspy/_signals/eds_tem.py
```diff
@@ -343,9 +371,25 @@ def quantification(self,
...
+                    ([element]))
+                number_of_atoms[i].metadata.set_item(
+                    "Sample.xray_lines", ([xray_line]))
         if plot_result and composition[i].axes_manager.signal_dimension != 0:
             utils.plot.plot_signals(composition, **kwargs)
-        return composition
+        if method=='zeta':
+            self.metadata.set_item("Sample.mass_thickness", mass_thickness)
+        if method == 'zeta':
+            return composition, mass_thickness
```
dnjohnstone: If both of the above are done if method == 'zeta' then you don't need two if statements.

hyperspy/_signals/eds_tem.py
k8macarthur: Ok cool. Which incidence? beam_current or real_time or the .EDS line?

dnjohnstone: if 'real_time' in self.metadata.Acquisition_instrument.TEM.Detector.EDS is False:

hyperspy/_signals/eds_tem.py
```diff
@@ -258,28 +259,32 @@ def get_calibration_from(self, ref, nb_pix=1):
             mp.Acquisition_instrument.TEM.Detector.EDS.live_time = \
                 mp_ref.Detector.EDS.live_time / nb_pix
 
+
     def quantification(self,
-                       intensities,
-                       kfactors,
-                       composition_units='weight',
+                       intensities='auto',
+                       method='auto',
```
dnjohnstone: if you make this just 'method', rather than 'method='auto'' then the function will tell you that you haven't specified enough stuff unless you specify method. That would allow you to remove the else and exception raising, so would be a little more code efficient and I think preferable.

hyperspy/_signals/eds_tem.py
vidartf: Even better is `if 'real_time' not in self.metadata.Acquisition_instrument.TEM.Detector.EDS` :)

hyperspy/_signals/eds_tem.py
vidartf: `raise ValueError('no method provided')` is probably the best here.

doc/user_guide/eds.rst
```diff
@@ -512,11 +512,19 @@ The background can be subtracted from the X-ray intensities with the :py:meth:`~
...
+Hyperspy now includes three methods for EDS quantification; Cliff-Lorimer, zeta-factors and ionization cross sections.
+
+Quantification must be applied to the background subtracted intensities, which can be found using :py:meth:`~._signals.eds.EDSSpectrum.get_lines_intensity`. The quantification of these intensities can then be determined using the :py:meth:`~._signals.eds_tem.EDSTEMSpectrum.quantification` method. These instensities are a stack of images for each element which can be extracted using :py:meth:`~._signals.eds.EDSSpectrum.get_lines_intensity`. The quantification method, needs be specified as either 'CL', 'zeta', or 'cross_section'. If no method is specified the function will raise an exception.
+A list of factors or cross sections should be supplied in the same order of the listed intesities (please not Hyperspy made using :py:meth:`~._signals.eds.EDSSpectrum.get_lines_intensity` will be in alphabetcial order. The required k-factors can be usually found in the EDS manufacturer software. Where as, zeta-factors and cross sections will need to be determined experimentally using standards. The zeta-factor method is described further in the following papers:
+
```
dnjohnstone: A few things spring to mind here:

1. k-factors should also be determined from a standard (in my opinion) although you'll find standardless values in many software packages...

2. If we're going to suggest references for Zeta and cross-section methods, we should include a reference or two for cliff lorimer (to be honest Williams & Carter is as good as any).

dnjohnstone: On a more general point @francisco-dlp what is your opinion on including references here at all?

It is my opinion that this is very helpful for new users and is basically the right thing to do. I know we've discussed issues around appearing to 'recommend' one approach over another etc though

hyperspy/tests/signal/test_eds_tem.py
```diff
@@ -157,15 +159,48 @@ def setUp(self):
...
+
+    def test_quant_cross_section(self):
+        s =self.signal
+        method = 'cross_section'
+        factors = [3, 5]
+        intensities = s.get_lines_intensity()
+        res = s.quantification(intensities, method, factors)
+        nt.assert_true(np.allclose(res[1][0].data, np.array(
+            [[ 16.02176621,  16.02176621],
+            [ 16.02176621,  16.02176621]])), atol=1e-3)
```
dnjohnstone: you've made a mistake with your brackets in a number of these tests which means that atol is being passed to assert_true rather than to allclose... (This is what Travis is currently complaining about). You can correct your brackets, but better is to replace with the following, which removes 1 level of brackets:

np.testing.assert_allclose(res[1][0].data, np.array([[ 16.02176621,  16.02176621],[ 16.02176621,  16.02176621]]), atol=1e-3)

Similar changes should be made everywhere that you've used nt.assert_true(np.allclose())

dnjohnstone: It may also be cleaner to assign the numpy array containing the expected result outside of the test, but not a big deal there.

doc/user_guide/eds.rst
k8macarthur: I agree with both points. I think referencing papers for each method is beneficial because then people can look up the exact maths that the function is doing. It stops it being too black box. It also means when they measure their own standards they have a method to follow. I'll see if I can find papers which cover standards for k-factors but I can't think of any off the top of my head.

doc/user_guide/eds.rst
k8macarthur: Ok so have a quantification UI but the function can still run without if people want to specify their parameters through commands? In which case I believe the errors need to remain in case the user forgets.

hyperspy/_signals/eds_tem.py
k8macarthur: For some reason that didn't work when I was writing this. Ah well it does now. :)

hyperspy/_signals/eds_tem.py
k8macarthur: Done. :)

hyperspy/_signals/eds_tem.py
k8macarthur: Done :)

hyperspy/_signals/eds_tem.py
k8macarthur: Done :)

hyperspy/_signals/eds_tem.py
k8macarthur: Done :)

hyperspy/_signals/eds_tem.py
k8macarthur: Done :) - I'm having to mark things as done otherwise I'll get lost in all the comments...

hyperspy/_signals/eds_tem.py
k8macarthur: Although if I make the adjustment above so that @dnjohnstone suggested about making the method argument without a default. It will presumable error out before this point, so this error is no longer needed?

dnjohnstone: That's the plan -- check it actually fails properly before pushing though, perfectly possible I've slipped up :)

hyperspy/tests/signal/test_eds_tem.py
```diff
@@ -157,15 +159,48 @@ def setUp(self):
...
+        method = 'CL'
         kfactors = [1, 2.0009344042484134]
+        composition_units = 'weight'
         intensities = s.get_lines_intensity()
-        res = s.quantification(intensities, kfactors)
-        nt.assert_true(np.allclose(res[0].data, np.array(
-                       [22.70779, 22.70779]), atol=1e-3))
+        res = s.quantification(intensities, method, kfactors, composition_units)
+        np.testing.assert_allclose(res[0].data, np.array([[22.70779, 22.70779],
+                    [22.70779, 22.70779]]), atol=1e-3)
```
dnjohnstone: Sorry, I think you need to specify rtol=1e-3 also, in all of these allclose cases... This defaults to 1e-7 but since you don't have that many decimal places that's failing Travis

hyperspy/tests/signal/test_eds_tem.py
```diff
@@ -175,15 +210,15 @@ def test_quant_zeros(self):
                            [0.5, 0.0, 0.0]]).T
         with ignore_warning(message="divide by zero encountered",
                             category=RuntimeWarning):
-            quant = utils_eds.quantification_cliff_lorimer(
-                intens, [1, 1, 3]).T
-        nt.assert_true(np.allclose(
+            quant = utils_eds.quantification(
```
dnjohnstone: This line is wrong -- Pierre was previously using just the actual quantification method specified in utils, you've now written 3 of these quantification_cliff_lorimer etc...

I think you should just switch this back to quantification_cliff_lorimer and then replicate the test for quantification_zeta_factor and quantification_cross_sections

hyperspy/_signals/eds_tem.py
```diff
@@ -325,12 +330,35 @@ def quantification(self,
...
+        elif method == 'cross_section':
+            results = utils_eds.quantification_cross_section(composition.data,
+                    cross_sections=factors,
+                    dose=self._get_dose(method))
+            composition.data = results[0] * 100
+            number_of_atoms = utils.stack(intensities)
+            number_of_atoms.data = results[1]
+            number_of_atoms = number_of_atoms.split()
+        else:
+            raise Exception ('Please specify method for quantification, as CL, zeta or cross_section')
```
dnjohnstone: assuming the change to force specification of method has worked this exception can be removed

doc/user_guide/eds.rst
dnjohnstone: Personally I'd be happy with directing people to Williams & Carter.... for now I think that's more than enough to tie this off at a first pass

doc/user_guide/eds.rst
dnjohnstone: @k8macarthur I think you should do what @siriagus did in his local branch and add these parameters to the already existing set_microscope_parameters() function.

Then users will be able to do for example set_microscope_parameters(pixel_width=0.025) so the documentation you've written here will need to reflect that - this will be much cleaner all round, I think you're safe to just go for it.

hyperspy/tests/signal/test_eds_tem.py
dnjohnstone: ok - this needs addressing in other ways then, there are two options:

1. If you know the width of the probe in a line scan presumably you can still define an area per position, so the method could be generalised to line scans?

2. If you really won't be able to use cross sections for line scans then the code needs to raise errors if you try to do so.

My gut instinct is that option 1 is preferable.

hyperspy/_signals/eds_tem.py
```diff
@@ -483,3 +526,66 @@ def create_model(self, auto_background=True, auto_add_lines=True,
...
+                raise Exception ('Please note that your real time is set to\
+the default value of 0.5s. The function will still run. However, if this is \
+incorrect you should consider changing it using \
+self.metadata.Acquisition_instrument.TEM.Detector.EDS.real_time .')
+        if method == 'cross_section':
+            if area == 'auto':
+                pixel1 = self.axes_manager[0].scale
+                pixel2 = self.axes_manager[1].scale
+            if self.axes_manager[0].scale == 1 or self.axes_manager[1].scale == 1:
+                warnings.warn('Please note your pixel width is set to the \
```
dnjohnstone: Is this the correct way to do this now? I got confused with all the changes to how we like exceptions/warnings etc to be raised.

vidartf: https://docs.python.org/3.5/howto/logging.html#when-to-use-logging :)

vidartf: To cite the appropriate segment:
>  warnings.warn() in library code if the issue is avoidable and the [user should modify his/her code] to eliminate the warning

> logging.warning() if there is nothing the [user] can do about the situation, but the event should still be noted

hyperspy/tests/signal/test_eds_tem.py
```diff
@@ -157,15 +159,49 @@ def setUp(self):
...
+        compositions_units = 'weight'
+        factors = [20, 50]
+        intensities = s.get_lines_intensity()
+        res = s.quantification(intensities, method, factors, compositions_units)
+        np.testing.assert_allclose(res[1].data, np.array(
+                [[ 2.7125736e-03,   2.7125736e-03],
+                [  2.7125736e-03,   2.7125736e-03]]), atol=1e-3)
+        np.testing.assert_allclose(res[0][1].data, np.array(
+                [[ 80.962287987,   80.962287987],
+                [  80.962287987,   80.962287987]]), atol=1e-3)
```
dnjohnstone: Shouldn't this be giving the same (or at least similar) weight percentages to the Cliff-Lorimer method?

hyperspy/tests/signal/test_eds_tem.py
```diff
@@ -157,15 +159,49 @@ def setUp(self):
...
+        method = 'cross_section'
+        factors = [3, 5]
+        intensities = s.get_lines_intensity()
+        res = s.quantification(intensities, method, factors)
+        np.testing.assert_allclose(res[1][0].data, np.array(
+            [[ 21517.1647074,  21517.1647074],
+            [  21517.1647074,  21517.1647074]]), atol=1e-3)
+        np.testing.assert_allclose(res[1][1].data, np.array(
+           [[ 21961.616621,  21961.616621],
+           [  21961.616621,  21961.616621]]), atol=1e-3)
```
dnjohnstone: What units have we got going on here?

hyperspy/_signals/eds_tem.py
dnjohnstone: Ok - my reading of this is that this is not appropriate then because it could be that the pixel width is 1nm and in that case the warning would still arise.

Presumably what should happen instead is that there is no default and the code fails if the user has not specified a scale value and in that case a warning is raised telling them to do so?

hyperspy/tests/signal/test_eds_tem.py
k8macarthur: The units are number of atoms. I only chose rough values for the cross sections and used the spectrum image which Pierre set up. There is quite a high number of counts in these spectrum images hence the large number of atoms.

hyperspy/tests/signal/test_eds_tem.py
k8macarthur: Only if this were a real sample. I have chosen the z-factors and cross sections fairly arbitrarily because I don't know the exact ones, hence the variation in percentages seen. For a real sample they should be somewhere similar yes.

hyperspy/_signals/eds_tem.py
k8macarthur: Ok so it should be a logging.warning then? Essentially the default step size is 1nm in the spectrum image. However, I can for see a situation where it might arise that the user wants to use the default value. In most scenarios it will act as a warning that they haven't calibrated their spectrum image properly. I don't think it is possible to have no default in this case because of how the spectrum scale is set up?

hyperspy/tests/signal/test_eds_tem.py
k8macarthur: It all depends on where the code gets the pixel width or probe width from. Currently I have been trying to use to x and y-scales from the metadata, and then the emphasis could be on making sure everything is calibrated. However, this causes the problem with the warning on eds_tem.py line 582 because there is a default step size of 1. Therefore it begs the questions of whether or not we add a pixel width option to the metadata or find a way around for line scans. I use line scans for my standards samples and it's pretty safe to assume that the probe area is symmetric in the x and y directions.

dnjohnstone: Ok - this makes me worry about some related things... the x and y scales in metadata, should (rightly) record the step size in x and y, which may be asymmetrical (although that's not necessarily very likely). More importantly it may have little to do with the probe size, which if I understand correctly is what the analysis really needs - so there's currently an assumption that the probe size is matched to the step size?

If that's the case - I think we need to make very clear in the documentation at least that if that's not the case the results will not mean much.

hyperspy/tests/signal/test_eds_tem.py
k8macarthur: The reason they're no large is the set up spectrum image has about 3000 counts per pixel.

hyperspy/tests/signal/test_eds_tem.py
k8macarthur: Yes, so they're interchangeable as long as sub-pixel scanning is used (or sample at higher pixel sample and then binning down, which is mathematically identical). I was trying to avoid duplicating stuff that was already in the meta-data. But I see your point so maybe a probe_area addition to the metadata may be the simpler way to go?

hyperspy/tests/signal/test_eds_tem.py
dnjohnstone: Ok - that's fine - have you checked the analysis on a dataset that you've also done by other means? So long as you're very confident that the result here is right, then I suppose it doesn't matter, but it might be nice to also make it representative of the kind of results someone might expect to get in normal usage. Could you crop a very small region e.g. 2x2 pixels of a dataset you know, or can synthesise and then use that, just so it looks sensible?

dnjohnstone: Ultimately the point is to test the code not the method though, so - so long as the result is known for sure it's fine - but its just pretty if it makes sense too :)

hyperspy/tests/signal/test_eds_tem.py
dnjohnstone: Ok - @siriagus could you do us a favour and check this on some of the data that you know well from your thesis? Just to double check everything is working correctly? I don't have any way to do this myself without a lot of effort really.

k8macarthur: I'm pretty confident that I've gone through it several times now. I'll double check with real data again next week. I agree it's better with realistic values but I was trying to use the simple spectrum Pierre set up for the k-factor quant test. Probe current, cross sections and dwell time are all realistic. Could probably increase the dwell time to make the atom counts look more realistic if that's preferable?

@siriagus a second test on the z-factors would be very helpful as I have to experimental data to check that.

siriagus: I'm not too familiar with how this test is written. What is it that you want me to check? :)

dnjohnstone: You don't need to check the test... I'll worry about that. It would just be great if you could pull this branch and check you can do the Zeta factor quantification of some data you know using it and get the right answer.

hyperspy/tests/signal/test_eds_tem.py
dnjohnstone: I'm not sure if I think you actually need to change anything - my gut is that the way it is implemented is ok. The key thing is to make sure it is really really obvious in the documentation that these things have to be in order - its a much more strict requirement than for other methods.

It's just to remember that although these things and sub-pixel scanning should be well known for those on the technique development side, many HyperSpy users will be doing just a bit of relatively routine characterisation with good analysis and I really don't think we can assume that people will firstly be able to do sub-pixel scanning on their machine, and secondly appreciate why it matters here -- so we have to point it out.

k8macarthur: So I've made an edit this morning. Now the user can set beam_area in the metadata. If they don't set a beam_area the function will then use the scaling from the meta.data. This should allow line scans to be possible again if they set the beam_area. I've also added more discussion about beam_area vs pixel_area in the documentation. Hopefully that is more accessible now.

k8macarthur:
Ok, so this morning I have included the three quantification parameters beam_current, beam_area and real_time in the set_microscope parameters function. I've also edited the dose function so that users can either supply a beam_area or change the step size in the spectrum image and use the pixel area. Was there anything else the needed adding that I've missed? @dnjohnstone have I included all your suggestions now?

doc/user_guide/eds.rst
```diff
@@ -512,11 +512,21 @@ The background can be subtracted from the X-ray intensities with the :py:meth:`~
 Quantification
 --------------
 
-.. versionadded:: 0.8
+Hyperspy now includes three methods for EDS quantification; Cliff-Lorimer, zeta-factors and ionization cross sections.
+
+Quantification must be applied to the background subtracted intensities, which can be found using :py:meth:`~._signals.eds.EDSSpectrum.get_lines_intensity`. The quantification of these intensities can then be determined using the :py:meth:`~._signals.eds_tem.EDSTEMSpectrum.quantification` method. These instensities are a stack of images for each element which can be extracted using :py:meth:`~._signals.eds.EDSSpectrum.get_lines_intensity`. The quantification method, needs be specified as either 'CL', 'zeta', or 'cross_section'. If no method is specified the function will raise an exception.
+A list of factors or cross sections should be supplied in the same order of the listed intensities (please note Hyperspy intensities made using :py:meth:`~._signals.eds.EDSSpectrum.get_lines_intensity` will be in alphabetical order. The required k-factors can be usually found in the EDS manufacturer software. Where as, zeta-factors and cross sections will need to be determined experimentally using standards.
+
```
dnjohnstone: Previous comments regarding k-factors, standards and references still apply here.

doc/user_guide/eds.rst
```diff
@@ -525,31 +535,58 @@ Quantification can be applied from the intensities (background subtracted) with
...
+    >>> s.set_microscope_parameters(beam_area = 0.00125)
+
+ Alternatively, if sub-pixel scanning is used (or the spectrum map was recorded at a high spatial sampling and then the data binned into much larger pixels before quantification) the illumination area then becomes the pixel area of the spectrum image. This is a much more accurate approach for quantitative EDX and should be used where possible.  The pixel width could either be added to the metadata by putting the pixel area in as the beam_area (above) or by calibrating the spectrum image using:
+
+..code-block: python
+
+    >>> s.axes_manager[0].scale = 0.025
+    >>> s.axes_manager[1].scale = 0.025
+
+Please note that the function assumes does not assume square pixels so both the x and y pixel dimensions must be set. For analysing line scans the pixel area should simply be added to the metadata as above.
```
dnjohnstone: "assumes does not assume" probably should be replaced with the latter

doc/user_guide/eds.rst
```diff
@@ -525,31 +535,58 @@ Quantification can be applied from the intensities (background subtracted) with
...
+
+The cross section method needs the beam_current, real_time and probe area in order to provide accurate quantification. beam_current and real_time can be set using the metadata as above. The beam_area (in nm^2) can be defined in two different ways. Either, the probe diameter is narrower than the pixel width, which case the probe is being under-sampled and therefore an estimation of the probe area needs to be used. This can be added to the metadata by:
+
+..code-block: python
+
+    >>> s.set_microscope_parameters(beam_area = 0.00125)
+
+ Alternatively, if sub-pixel scanning is used (or the spectrum map was recorded at a high spatial sampling and then the data binned into much larger pixels before quantification) the illumination area then becomes the pixel area of the spectrum image. This is a much more accurate approach for quantitative EDX and should be used where possible.  The pixel width could either be added to the metadata by putting the pixel area in as the beam_area (above) or by calibrating the spectrum image using:
+
```
dnjohnstone: I think this should just end "...or by calibrating the spectrum image" and linking to somewhere else in the documentation where this is described... I'm not sure if it is described somewhere else? But since it is much more general than EDX it should be

hyperspy/_signals/eds_tem.py
```diff
@@ -325,12 +363,33 @@ def quantification(self,
...
+                mask=navigation_mask) * 100.
+        elif method == 'zeta':
+            results = utils_eds.quantification_zeta_factor(
+                composition.data, zfactors=factors,
+                dose=self._get_dose(method))
+            composition.data = results[0] * 100.
+            mass_thickness = intensities[0].deepcopy()
+            mass_thickness.data = results[1]
+            mass_thickness.metadata.General.title = 'Mass thickness'
+        else:
```
dnjohnstone: Doesn't this mean that if method is not 'CL' or 'zeta' for example one might set 'method=bananas' the code will give the result according to cross-sections? Is that the behaviour you intended?

k8macarthur: Oops no! I think I deleted one too many error calls :)

tjof2: Agreed, clarity on the default behaviour, perhaps raising some kind of flag that it is defaulting to "X" method?

k8macarthur: I think I'd rather force them to pick an method so that they have to think about which method they've chosen and have the correct standards values for that method.

siriagus: I agree that forcing the user to specify the method is better. One
advantage with HyperSpy is that the user must think and not just "click a
button". :)
11. apr. 2016 15:23 skrev "k8macarthur" <notifications@github.com> følgende:

doc/user_guide/eds.rst
k8macarthur: Ok, I think I've worked out how to add a link to the other page on but should be double checked before it goes on the webpage.

doc/user_guide/eds.rst
k8macarthur: done :)

doc/user_guide/eds.rst
k8macarthur: Fixed, just needed to actually take the time to dig out Williams & Carter and find the chapter title.

dnjohnstone: You mean you don't have a copy within grabbing distance at all times?! I'ts Chapter 35 - "Quantitative X-ray Analysis"

hyperspy/_signals/eds_tem.py
dnjohnstone: No need to panic all - I think this was just a slip in deleting stuff! As previous discussions method here has no default anymore so that the user must choose - this is appropriate.

hyperspy/tests/signal/test_eds_tem.py
dnjohnstone: Ok - only question is minor, but do we thing "beam_area" or "probe_area"?

k8macarthur: I’m not sure because we have beam_energy and beam_current so I was trying to keep consistent with the other names used in the metadata.

k8macarthur:
@siriagus have you had a chance to check to z-factor part of the function on some of your data yet and that you're happy with it?

siriagus:
@k8macarthur Hi! Sorry, I have been really busy. I haven't updated to Python 3 yet, so I'll do that and take a look at it now.

siriagus:
@k8macarthur Hi! I've fixed a minor error that made the quantification fail and tidied up a bit, especially the _get_dose() function. I've created a pull request k8macarthur/hyperspy#1. I would be happy if people could confirm that this is the correct way to do it, that is creating a pull request to @k8macarthur 's repo. Does this, #913, update automatically when she merges?

k8macarthur:
@siriagus Thanks! I've made one or two edits but kept most of what you suggested. 

Does anyone know why the AppVeyor build is failing now? The Fail message says:

  File "C:\Miniconda35\lib\site-packages\nose\case.py", line 198, in runTest
    self.test(*self.arg)
  File "C:\projects\hyperspy\hyperspy\tests\model\test_fancy_indexing.py", line 81, in test_model_navigation_indexer_slice
    assert_false(m[0]._active_array[0, 0])
AssertionError: True is not false
 
Which suggests it's nothing to do with any of the edits I've pushed through today?

vidartf:
Yeah, that's an unrelated issue (see #914), you can ignore it.

k8macarthur:
Thank @vidartf  a second run through of checks seems to have also worked here. :)

siriagus:
@k8macarthur Great! Thanks for looking through it so quickly! :)

k8macarthur:
So I've fixed all the comments people have provided. Does this make this section ready merge soon?

doc/user_guide/eds.rst
```diff
@@ -525,31 +538,53 @@ Quantification can be applied from the intensities (background subtracted) with
...
-    >>> # With weight_percent from before
-    >>> atomic_percent = hs.material.weight_to_atomic(weight_percent)
+    >>> # With atomic_percent from before
+    >>> weight_percent = hs.material.atomic_to_weight(atomic_percent)
+
+The zeta-factor method needs both the beam_current (in nA) and the acquisition or dwell time (referred to as real_time in seconds) in order to provide an accurate quantification. Both of the these parameters can be assigned to the metadata of the spectrum using:
+
+..code-block:: python
+
+    >>> s.set_microscope_parameters(beam_current = 0.5)
```
francisco-dlp: Simply style issue: there should be no spaces around the equal sign.

doc/user_guide/eds.rst
```diff
@@ -525,31 +538,53 @@ Quantification can be applied from the intensities (background subtracted) with
...
+    >>> s.set_microscope_parameters(real_time = 1.5)
```
francisco-dlp: Same as above.

doc/user_guide/eds.rst
```diff
@@ -525,31 +538,53 @@ Quantification can be applied from the intensities (background subtracted) with
...
+
+If these are not set the code will produce an error stating which parameter has been forgotten.
+The zeta-factor method will produce two sets of results. Index [0] is the composition maps for each element in atomic percent and index [1] is the mass-thickness map.
+
+The cross section method needs the beam_current, dwell time (real_time) and probe area in order to provide accurate quantification. The beam_current and real_time can be set using :py:meth:`~._signals.eds_tem.EDSTEMSpectrum.set_microscope_parameters` as shown above. The beam_area (in nm^2) can be defined in two different ways. Either, the probe diameter is narrower than the pixel width, which case the probe is being under-sampled and therefore an estimation of the probe area needs to be used. This can be added to the metadata by:
+
+..code-block: python
+
+    >>> s.set_microscope_parameters(beam_area = 0.00125)
```
francisco-dlp: Same as above.

doc/user_guide/eds.rst
```diff
@@ -525,31 +538,53 @@ Quantification can be applied from the intensities (background subtracted) with
...
+
+..code-block:: python
+
+    >>> s.set_microscope_parameters(beam_current = 0.5)
+    >>> s.set_microscope_parameters(real_time = 1.5)
+
+If these are not set the code will produce an error stating which parameter has been forgotten.
+The zeta-factor method will produce two sets of results. Index [0] is the composition maps for each element in atomic percent and index [1] is the mass-thickness map.
+
+The cross section method needs the beam_current, dwell time (real_time) and probe area in order to provide accurate quantification. The beam_current and real_time can be set using :py:meth:`~._signals.eds_tem.EDSTEMSpectrum.set_microscope_parameters` as shown above. The beam_area (in nm^2) can be defined in two different ways. Either, the probe diameter is narrower than the pixel width, which case the probe is being under-sampled and therefore an estimation of the probe area needs to be used. This can be added to the metadata by:
```
francisco-dlp: Another minor styling issue: variable names should be surrounded by backquotes.

doc/user_guide/eds.rst
francisco-dlp: And in several places below that  I won't comment on.

doc/user_guide/eds.rst
```diff
@@ -512,11 +511,25 @@ The background can be subtracted from the X-ray intensities with the :py:meth:`~
 Quantification
 --------------
 
-.. versionadded:: 0.8
```
francisco-dlp: Why removing this tag?

doc/user_guide/eds.rst
```diff
@@ -512,11 +511,25 @@ The background can be subtracted from the X-ray intensities with the :py:meth:`~
...
+Hyperspy now includes three methods for EDS quantification; Cliff-Lorimer, the zeta-factor method and ionization cross sections.
+
+Quantification must be applied to the background subtracted intensities, which can be found using :py:meth:`~._signals.eds.EDSSpectrum.get_lines_intensity`. The quantification of these intensities can then be determined using the :py:meth:`~._signals.eds_tem.EDSTEMSpectrum.quantification` method. These instensities are a stack of images for each element which can be extracted using :py:meth:`~._signals.eds.EDSSpectrum.get_lines_intensity`. The quantification method, needs be specified as either 'CL', 'zeta', or 'cross_section'. If no method is specified the function will raise an exception.
```
francisco-dlp: > These instensities are a stack of images

Couldn't they be line profiles, tilt-series or higher dimensional signals?

doc/user_guide/eds.rst
```diff
@@ -512,11 +511,25 @@ The background can be subtracted from the X-ray intensities with the :py:meth:`~
...
+A list of factors or cross sections should be supplied in the same order of the listed intensities (please note Hyperspy intensities made using :py:meth:`~._signals.eds.EDSSpectrum.get_lines_intensity` will be in alphabetical order. The required k-factors can be usually found in the EDS manufacturer software. Where as, zeta-factors and cross sections will need to be determined experimentally using standards.
```
francisco-dlp: Does "Where as" mean "otherwise" here? If yes, wouldn't "otherwise" be clearer for non-native English speakers?

doc/user_guide/eds.rst
```diff
@@ -512,11 +511,25 @@ The background can be subtracted from the X-ray intensities with the :py:meth:`~
...
+For further information about k-factor quantification and determining k-factors from standard samples please look at:
+
+Transmission Electron Microscopy: A Textbook for Materials Science (Part 4) by Williams & Carter
```
francisco-dlp: It'll be better to add the citations to the bibliography `user_guide/bibliography.rst`. Citing the only current entry in the bibliography is as simple as writing `[Egerton2011]_`. Same for the other citations below.

doc/user_guide/eds.rst
```diff
@@ -525,31 +538,53 @@ Quantification can be applied from the intensities (background subtracted) with
...
+
+..code-block: python
+
+    >>> s.set_microscope_parameters(beam_area = 0.00125)
+
+ Alternatively, if sub-pixel scanning is used (or the spectrum map was recorded at a high spatial sampling and then the data binned into much larger pixels before quantification) the illumination area then becomes the pixel area of the spectrum image. This is a much more accurate approach for quantitative EDX and should be used where possible.  The pixel width could either be added to the metadata by putting the pixel area in as the beam_area (above) or by calibrating the spectrum image (see `Setting axis properties <http://hyperspy.org/hyperspy-doc/dev/user_guide/getting_started.html#setting-axis-properties>`_ )
```
francisco-dlp: Internal links should use labels instead of the full address. See http://www.sphinx-doc.org/en/stable/markup/inline.html#ref-role

hyperspy/_signals/eds_tem.py
```diff
@@ -325,12 +363,35 @@ def quantification(self,
...
+        elif method == 'cross_section':
+            results = utils_eds.quantification_cross_section(composition.data,
+                    cross_sections=factors,
+                    dose=self._get_dose(method))
+            composition.data = results[0] * 100
+            number_of_atoms = utils.stack(intensities)
+            number_of_atoms.data = results[1]
+            number_of_atoms = number_of_atoms.split()
+        else:
+            raise Exception ('Please specify method for quantification, as \'CL\', \'zeta\' or \'cross_section\'')
```
francisco-dlp: `ValueError` instead?

hyperspy/_signals/eds_tem.py
```diff
@@ -343,9 +404,24 @@ def quantification(self,
...
+                    "Sample.xray_lines", ([xray_line]))
         if plot_result and composition[i].axes_manager.signal_dimension != 0:
             utils.plot.plot_signals(composition, **kwargs)
-        return composition
+        if method=='zeta':
+            self.metadata.set_item("Sample.mass_thickness", mass_thickness)
+            return composition, mass_thickness
+        elif method == 'cross_section':
+            return composition, number_of_atoms
+        else:
```
francisco-dlp: `elif` here would be clearer.

francisco-dlp: Beside, the Returns section of the docstring seem not to describe what this method returns accurately.

hyperspy/_signals/eds_tem.py
```diff
@@ -483,3 +559,76 @@ def create_model(self, auto_background=True, auto_add_lines=True,
...
+                    area = parameters.TEM.beam_area
+                else:
+                    pixel1 = self.axes_manager[0].scale
+                    pixel2 = self.axes_manager[1].scale
+                    if self.axes_manager[0].scale == 1 or self.axes_manager[1].scale == 1:
+                        warnings.warn('Please note your beam_area is set to'
+                        'the default value of 1 nm^2. The function will still run. However if 1 nm^2 is not'
+                        'correct, please read the user documentations for how to set this properly.')
+                    area = pixel1 * pixel2
+            return (real_time * beam_current * 1e-9) /(constants.e * area)
```
francisco-dlp: Maybe a little comment explaining where `1e-9` is coming from?

hyperspy/misc/eds/utils.py
```diff
@@ -453,3 +453,75 @@ def _quantification_cliff_lorimer(intensities,
...
+        The intensities for each X-ray line. The first axis should be the
+        elements axis.
+    zfactors: list of float
+        The list of zeta-factors in the same order as intensities
+        e.g. zfactors = [628.10, 539.89] for ['As_Ka', 'Ga_Ka'].
+    dose: float
+        The total electron dose given by i*t*N, i the current, t the acquisition time
+        and N the number of electrons per unit electric charge (1/e).
+
+    Return
```
francisco-dlp: typo: Return -> Returns

hyperspy/_signals/eds_tem.py
```diff
@@ -250,39 +284,43 @@ def get_calibration_from(self, ref, nb_pix=1):
...
+        factors: list of float
+            The list of kfactors, zeta-factors or cross sections in same order as
+            intensities. Note that intensities provided by Hyperspy are sorted
+            by the alphabetical order of the X-ray lines.
+            eg. factors =[0.982, 1.32, 1.60] for ['Al_Ka', 'Cr_Ka', 'Ni_Ka'].
         composition_units: 'weight' or 'atomic'
-            Quantification returns weight percent. By choosing 'atomic', the
-            return composition is in atomic percent.
+            The quantification returns the composition in atomic percent by default,
+            but can also return weight percent if specified.
```
francisco-dlp: Typo below: "Return" -> "Returns"

francisco-dlp:
It correspond to @dnjohnstone and @siriagus  to decide it this is ready to merge. @dnjohnstone and @siriagus, are you happy with this?

Beside, could you edit the header of this pull request to briefly describe the feature, maybe including a very simple example of usage?

doc/user_guide/eds.rst
k8macarthur: No I think I mean 'where as'. K-factors are found tabulated, where as zfactors and cross sections will need to be determined experimentally.

doc/user_guide/eds.rst
k8macarthur: I wasn't sure if the tag wanted to be when quantification was added or updated for the two additional quantification methods?

doc/user_guide/eds.rst
k8macarthur: Cool, do you have particular format you're following for citations?

doc/user_guide/eds.rst
k8macarthur: Hopefully fixed now.

doc/user_guide/eds.rst
k8macarthur: fixed.

doc/user_guide/eds.rst
k8macarthur: fixed.

hyperspy/_signals/eds_tem.py
```diff
@@ -343,9 +410,26 @@ def quantification(self,
...
+                    "Sample.xray_lines", ([xray_line]))
         if plot_result and composition[i].axes_manager.signal_dimension != 0:
             utils.plot.plot_signals(composition, **kwargs)
-        return composition
+        if method=='zeta':
+            self.metadata.set_item("Sample.mass_thickness", mass_thickness)
+            return composition, mass_thickness
+        elif method == 'cross_section':
+            return composition, number_of_atoms
+        elif method == 'CL'
```
dnjohnstone: This is your Travis issue...

Should be: elif method == 'CL':

hyperspy/tests/signal/test_eds_tem.py
dnjohnstone: ok - sounds as good a reason as energy -- personally I'd like to see all of this switched to "accelerating_voltage", "probe_current", and "probe_area" -- but that is for a different PR.

hyperspy/_signals/eds_tem.py
k8macarthur: Ah ha! Thanks :)

dnjohnstone:
@francisco-dlp as far as I'm concerned this is, broadly speaking, working... I don't see too much reason not to merge once the minor typos and style bits are done.

doc/user_guide/eds.rst
francisco-dlp: I think that it is good to leave it as it was. You could add a `versionchanged` one describing what you've added for version `1.0`. See [this](http://www.sphinx-doc.org/en/stable/markup/para.html).

hyperspy/tests/signal/test_eds_tem.py
k8macarthur: I agree probe_current and probe_area would make more sense to me. But I just wasn't sure if that was me being a STEM person.

doc/user_guide/eds.rst
francisco-dlp: Not really. Feel free to propose your favourite one.

doc/user_guide/eds.rst
```diff
@@ -512,11 +511,17 @@ The background can be subtracted from the X-ray intensities with the :py:meth:`~
...
 
-One TEM quantification method (Cliff-Lorimer) is implemented so far.
+Quantification must be applied to the background subtracted intensities, which can be found using :py:meth:`~._signals.eds.EDSSpectrum.get_lines_intensity`. The quantification of these intensities can then be determined using the :py:meth:`~._signals.eds_tem.EDSTEMSpectrum.quantification` method. These instensities are a stack (of images images or otherwise) for each element which can be extracted using :py:meth:`~._signals.eds.EDSSpectrum.get_lines_intensity`. The quantification method, needs be specified as either 'CL', 'zeta', or 'cross_section'. If no method is specified the function will raise an exception.
+A list of factors or cross sections should be supplied in the same order of the listed intensities (please note Hyperspy intensities made using :py:meth:`~._signals.eds.EDSSpectrum.get_lines_intensity` will be in alphabetical order). The required k-factors can be usually found in the EDS manufacturer software. Where as, zeta-factors and cross sections will need to be determined experimentally using standards.
```
dnjohnstone: I  can see Francisco's point here... How about the following:

A list of factors or cross sections should be supplied in the same order of the listed intensities (please note Hyperspy intensities made using :py:meth:`~._signals.eds.EDSSpectrum.get_lines_intensity` will be in alphabetical order). A set of k-factors can be usually found in the EDS manufacturer software although determination from standard samples for the particular instrument used is usually preferable. On the other hand, zeta-factors and cross sections must be determined experimentally using standards.

hyperspy/tests/signal/test_eds_tem.py
francisco-dlp: I agree that the alternatives that you propose are better. In my opinion the worst offender is `beam_energy`. However, we should only rename things when strictly necessary because any change break scripts that use the old syntax.

siriagus:
What about absorption corrections? A lot of emphasis in the original paper, see [Watanabe and Williams (2006)](http://dx.doi.org/10.1111/j.1365-2818.2006.01549.x), is placed upon that absorption correction is one of the great advantages with the zeta-fator method (because mass-thickness is determined simultaneously with composition)  If #797 is ready to merge, this wouldn't be too much work to include.

I think it would be best to merge this branch as it is now, then merge #797 if we feel confident it is ready too, and then add absorption later, but before v.1.0.0 as I feel the zeta-factor method is incomplete without it. For the model systems I have looked at, absorption has not been a problem, but if I was a new user of HyperSpy and read the original zeta-factor method paper I would assume that this was already implemented. Therefore, I suggest currently putting a comment in the user documentation that absorption correction is not performed.

Another thing I would like to discuss is: What about including a function for determining the zeta-factor for a known system. I am doing it by my own script, but there would be several advantages of doing so would be to make sure that the user get zeta-factors that correspond to weight-percent (convention). In my Master's I've determined the zeta-factors using atomic percent (forcing me to passing composition_unit='weight' in the quantification method, just to prevent conversion (output should be atomic). This is a "mistake" on my part as I haven't followed the convention when determining the zeta-factors. However, another user might do the same and then assuming if the pass 'atomic' to quantificaiton() that it would be atomic. To ensure that the "weight-percent" zeta-factors are used, we can have an optional argument to determine_zeta_factor(composition_units='atomic'), so that the user can give the composition in atomic, but the zeta-factors would (this should be emphasised in user doc!) correspond to the convention we have adapted (comp. conventionally in wt.%). Alternatively, in the user documentation write that the input composition to determine_zeta_factor() must be weight percent and then include an example how to convert atomic to weight percent calling `hyperspy.misc.eds.utils.material.atomic_to_weight(composition)`.

That was a lot of text, unsure how to shorten it, hopefully you understand what the issue is...

TL;DR: I think this PR is ready to merge, however, I believe some additional features (absorption correction, zeta-factor calibration etc. should be added before releasing v.1.0.0.

francisco-dlp:
@siriagus, I think that what you propose is very sensible. I suggest that you open two new issues: one to propose implementing the absorption correction and another one to add the extra keyword to the EDX quantification method. Otherwise we may forget about this by the time we start thinking on releasing `1.0`.

dnjohnstone:
@siriagus I think you're pretty much bang on and if you're willing to take on implementing that (I certainly don't have time myself) it would be excellent! It also doesn't need to affect merging this PR I think.

If you can review #797 then we can probably merge that pretty quickly - I think it's pretty much there.

Then if you can produce a PR implementing absorption correction as you understand it should be, that would be great.

k8macarthur:
I agree as well. I also plan on writing a function at some point which converts between z-factors and cross sections which might be useful generally in. Plus from my perspective that means all the absorption correction which has been derived for z-factors could then be converted for cross sections (I think). 

I think the PR can go as it is initially as absorption correction in only necessary for thick samples (so anything larger than 10nm) which means that there is a range of quantification which can be carried out before absorption correction is needed.

doc/user_guide/eds.rst
k8macarthur: I like it. Happy to include the subtle emphasis on determining k-factors experimentally too. ;)

dnjohnstone:
Thick specimens (>10nm) - this does highlight one thing that I think we have to be a bit careful with here, which is that probably 95% of HyperSpy users are concerned with specimens that will be 'thin' at 50 nm. Obviously, it's ultimately up to the user to understand that not all methods are appropriate for all samples - but we should do our best to highlight obvious traps.

hyperspy/tests/signal/test_eds_tem.py
k8macarthur: Ok so I'll change beam_area as that is one that can be changed without affecting other code and is better.

francisco-dlp:
@k8macarthur one more tiny thing: could you come up with a shorter version of the title of this PR?

Let us know when this is ready for merging after addressing the last few minor comments and I'll push the merge button!

k8macarthur:
@siriagus with regards to weight versus atomic percent. Would you prefer it if I specify in the user documentation that we are assuming z-factors are in weight to make it clear?
@dnjohnstone 10nm is my estimate of the top of my head so not an absolute value because I know it isn't an issue for my nanoparticles. Although that it very much element specific, heavier elements have a much smaller thickness before absorption comes into effect. Maybe that can be addressed in the documentation for the new absorption pull request?

siriagus:
@k8macarthur Yes, I think a short sentence just stating that we are assuming z-factors are calibrated using weight percent would be helpful to avoid confusion. Out of curiousity, why does the cross-section method use a different convention? I assume it has something to do with your definition of the cross-section? I don't have your paper in front of me right now, so could you just briefly clarify this? :)

k8macarthur:
@siriagus I come from a background of analysing nanoparticles so atomic percent seemed more helpful, especially when my elements have very different atomic weights. Z-factors were designed for thin films, so bulk metallic samples where I guess weight percent is what the metallurgist prefer to work in. I never thought about following a particular convention it just seemed to make more sense.

dnjohnstone:
I think atomic percent has got to be the sensible choice...

siriagus:
@k8macarthur @dnjohnstone For the Cliff-Lorimer method, wt% has been the convention for almost 40 years, so it makes sense keeping it as it is. For the zeta-factor method, my understanding of the paper is that they have kept this convention, however, there is nothing wrong about using at%. This is something to take into account when converting between zeta-factors and cross-sections. Since the cross-section method is mostly for nanoparticles I completely agree it makes sense to define it with compositions in at%. And even though I would personally prefer using at% for the zeta-factor method (also in zeta-factor calibration), I believe following a 40 year old convention probably makes the most sense for people coming from a thin-film analysis background (i.e. Cliff-Lorimer). I was just a bit confused why we didn't use the same convention everywhere. It makes sense the way it is implemented now. :)

k8macarthur:
Ok as far as I'm concerned I think it's ready to go. AppVeyor still seems to be failing due to the #914 issue but the rest looks good to me. :)

francisco-dlp:
Happy to merge this fine work, kudos to the authors and reviewers.

francisco-dlp:
@k8macarthur and @siriagus, I think that this is your first merged contribution to HyperSpy—congratulations!

You'll get authorship credit for this in our zenodo entry once we release HyperSpy 1.0. zenodo takes the author name and affiliation directly from github, could you verify that this information is accurate in your GH profile?

siriagus:
@francisco-dlp Yay! The information on my GH profile is correct.

@k8macarthur Congratulations! Well done, Katherine! :)

PR #915, opened 2016-03-20
FIX: make io_plugins/semper_umf work with numpy < v1.9.0

magnunor:
numpy.ndarray.tobytes() was introduced in numpy 1.9.0,
however the current package in Ubuntu 15.10 is numpy 1.8.2.
So changing tobytes() to tostring(), which is in older numpy
versions, should fix this. Without (probably) changing the
functionality. As the test for semper_umf seems to pass now.

This will probably also make io_plugins/semper_umf work with
older versions of Ubuntu and Debian.

### DIFF
--- a/hyperspy/io_plugins/semper_unf.py
+++ b/hyperspy/io_plugins/semper_unf.py
@@ -491,7 +491,7 @@ class SemperFormat(object):
                     struct.pack(
                         '<i',
                         12))  # record length, 4 byte format!
-                f.write(header.tobytes())
+                f.write(header.tostring())
                 f.write(
                     struct.pack(
                         '<i',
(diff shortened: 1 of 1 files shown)


magnunor:
Should also add: the function tostring is a compatibility alias for tobytes. Despite its name it returns bytes not strings. https://docs.scipy.org/doc/numpy-1.10.0/reference/generated/numpy.ndarray.tostring.html

to266:
You should probably PR this to a different branch, if you want it to be release with the next maintenance. Though it's confusing for me at the moment... @francisco-dlp, where should these things go?

ericpre:
Since we now require numpy >= 1.10 (see #999), I guess this is not relevant anymore and it can be closed?

francisco-dlp:
Thank for spotting it.

PR #919, opened 2016-03-22
Import mock from unittest

francisco-dlp:
Since Python 3.3 (released in 2012) mock is in the unitests module of the std lib. This removes the dependency.


to266:
Only thing left is to remove mock from `anaconda_hyperspy_environment.yml` as well

francisco-dlp:
Close in favour of #1082

Issue #914, opened 2016-03-20
Random test failures `test_model_navigation_indexer_slice`


vidartf:
I sometimes encounter the following failure in tests. The matter only occurs sometimes, and rerunning the test will often succeed. Possibly a race condition? Do we have anything that relies on the order in which events/traits notification is processed?

```python
======================================================================

FAIL: hyperspy.tests.model.test_fancy_indexing.TestModelIndexing.test_model_navigation_indexer_slice

----------------------------------------------------------------------

Traceback (most recent call last):

  File "/home/travis/miniconda/envs/testenv/lib/python3.5/site-packages/nose/case.py", line 198, in runTest

    self.test(*self.arg)

  File "/home/travis/build/hyperspy/hyperspy/hyperspy/tests/model/test_fancy_indexing.py", line 81, in test_model_navigation_indexer_slice

    assert_false(m[0]._active_array[0, 0])

AssertionError: True is not false
```

francisco-dlp:
travis is failing in #919 because of this.

vidartf:
Rerunning travis in #919 caused it to succeed.

sem-geologist:
What exactly this test is testing?
the ```m[0]._active_array[0, 0]``` looks for me as slice of some kind of array in array? (forgive me I am not familiar with hyperspy models). However, looking at the TestModelIndexing.setUp method I see that there is created numpy array with float type (dtype is not specified, so by default it is float). If I understand correct then the ```m[0]._active_array[0, 0]``` will return random float number. So it does not surprise me that it randomly fails.

Generally speaking the test looks suspicious, there are more lines which can fail in the future, in example ```assert_true(np.all(m.channel_switches[1:] == True))```.
```m.channel_switches[1:]``` probably returns tuple or list. Those should not be compared with True.

vidartf:
From memory, active array should be array of booleans. All the times the test fails, it says `True is not false`, which to me indicates that the `dtype` is correct (since `True` is boolean), bu that the value is wrong (should be `False`).

sem-geologist:
Ok, I went through the test  in python command line checking it step by step. the final array results with [0,0] element beeing False on my computer, so it does pass the test. generally It looks to over complicated for me. I tried to track down what is going on and got to _signals/spectrum_simulation.py :open_mouth:  Why somebody would get so abstract....?

to266:
FYI the `spectrum_simulation.py` here is most likely irrelevant - it's only used to generate the noise.

[...]

PR #922, opened 2016-03-23
Set version back to dev

francisco-dlp:


francisco-dlp:
I don't understand anything. travis was happy with the branch (see [this](https://travis-ci.org/francisco-dlp/hyperspy/builds/117999595)) but unhappy with the pull request... Appveyor passes (it only fails because of #914)

vidartf:
If the issue is line endings, then I suspect the AppVeyor git checkout converts line endings to CRLF(?). 

Why it works in your branch and not the PR... No idea. What happens if you manually check out a branch that is known to work, and then redo the changes (luckily this quite a simple PR)?

dnjohnstone:
Did anyone ever work out what was causing a problem here? It's totally bemusing to me...

vidartf:
I would recommend nuking the branch and redoing the change (shouldn't be too much typing involved ;) ). As the issue haven't really showed up anywhere else (?), I think we don't have to solve this particular mystery (unless you're very curious).

francisco-dlp:
Closing following @vidartf suggestion.

PR #927, opened 2016-03-31
DOC: how to install Python3 HyperSpy from source on Ubuntu 15.10

magnunor:
I made some instructions on how to install HyperSpy from source on Ubuntu 15.10.

Related: should the installation section of the documentation be reordered? Currently the required libraries are listed after the `pip install hyperspy` (even if it is linked). So in Ubuntu only running the `pip install hyperspy` without downloading the dependencies from the Ubuntu repositories will fetch the `numpy`, `scipy`, ++ from source and start compiling them which again needs its own dependencies (and would most likely fail). 

So it would possibly be more orderly having the different ways of installing it under different headings, like "Linux", "Ubuntu", "Arch Linux", "...". Which includes all the instructions you need to properly install it inside that section?

### DIFF
--- a/doc/user_guide/install.rst
+++ b/doc/user_guide/install.rst
@@ -182,6 +182,26 @@ If using Arch Linux, the latest checkout of the master development branch can be
 installed through the AUR by installing the `hyperspy-git package
 <https://aur.archlinux.org/packages/hyperspy-git/>`_
 
+Installing HyperSpy from source in Ubuntu 15.10:
+
+.. code-block:: bash
+    
+    $ sudo apt-get install git python3-pip python3-matplotlib python3-numpy \
+        python3-scipy python3-skimage python3-requests python3-natsort \
+        python3-setuptools python3-h5py python3-pyqt4 ipython3 \
+        python3-mock python3-dill
+    $ sudo apt-get install python3-sumpy --no-install-recommends
+    $ sudo pip3 install sklearn traits traitsui
+    $ git clone https://github.com/hyperspy/hyperspy.git
+    $ cd hyperspy
+    $ sudo pip3 install -e ./
+
+The package `python3-sympy` recommends installing the `TexLive` Latex
+distribution, which requires downloading about 1 GB of extra packages.
+So this is avoided by using the `--no-install-recommends`.
+The `sklearn`, `traits` and `traitsui` packages are not available for 
+Python3 in Ubuntu 15.10, so these must be installed using `pip3`.
+
 .. _create-debian-binary:
 
 Creating Debian/Ubuntu binaries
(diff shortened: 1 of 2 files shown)


vidartf:
If you first cloned the repo, could you then use `stdeb.cfg` (which I think needs an update for python 3) as input for apt/apt-get install? Then there would be one fewer place where dependencies need to be tracked in source... (and commands might be nicer to type if copy paste is unavailable for some reason).

doc/user_guide/install.rst
```diff
@@ -182,6 +182,26 @@ If using Arch Linux, the latest checkout of the master development branch can be
...
 
+Installing HyperSpy from source in Ubuntu 15.10:
+
+.. code-block:: bash
+    
+    $ sudo apt-get install git python3-pip python3-matplotlib python3-numpy \
+        python3-scipy python3-skimage python3-requests python3-natsort \
+        python3-setuptools python3-h5py python3-pyqt4 ipython3 \
+        python3-mock python3-dill
+    $ sudo apt-get install python3-sumpy --no-install-recommends
```
vidartf: `sumpy` -> `sympy` ?

magnunor:
I think having the package installation listed like this is more appropriate for the advanced users (who would like to install from source), since having a list of the dependencies could be useful.

francisco-dlp:
This seems ok to me. Unless someone (@vidartf) has anything against it, I suggest merging.

Hopefully we'll soon revert it because we've close #455 ;)

vidartf:
Sure. I would just add that it would be helpful if there was a paragraph "Managing dependencies" in the dev guide, so that anytime a dependency needs to be added/updated/removed we can ensure we do it in all relevant places. Updating/removing is reasonably easy (just search for package name), but it can be easy to forget a spot when adding.

francisco-dlp:
Good point. If @magnunor would like to do that here, we can wait for the new commit. If not, @magnunor  feel free to merge this.

magnunor:
@vidartf, the dev guide you're referring to, is this the one: https://github.com/hyperspy/hyperspy/blob/master/doc/dev_guide.rst?

vidartf:
@magnunor : Yes, or if you'd like, here: http://hyperspy.readthedocs.org/en/latest/dev_guide.html :)

magnunor:
I added a list of the dependency listings (in 3 commits due to syntax errors, I really shouldn't be doing programming before my morning coffee...).

Is that all of them?

vidartf:
Also Travis/AppVeyor needs new dependencies.

doc/user_guide/install.rst
```diff
@@ -182,6 +182,26 @@ If using Arch Linux, the latest checkout of the master development branch can be
...
 <https://aur.archlinux.org/packages/hyperspy-git/>`_
 
+Installing HyperSpy from source in Ubuntu 15.10:
+
+.. code-block:: bash
+    
+    $ sudo apt-get install git python3-pip python3-matplotlib python3-numpy \
+        python3-scipy python3-skimage python3-requests python3-natsort \
+        python3-setuptools python3-h5py python3-pyqt4 ipython3 \
+        python3-mock python3-dill
```
sem-geologist: doesn't mock went away?

francisco-dlp:
This branch was automatically closed when we renamed `master` `RELEASE_next_major` and it cannot be reopened. Could you resubmitted it to RELEASE_next_patch instead?

PR #933, opened 2016-04-02
Prevent overflows for larger blockfiles

vidartf:
This upconverts the integer file types for the dimensions read from the blockfile header so that they can be used in calculations without overflowing. This is necessary for any DP sizes above 255*255. As the `int` type in Python3 has unlimited precision, this should be more than sufficient.


dnjohnstone:
Works for me - would merge

PR #934, opened 2016-04-03
DOC: bcf documentation to io.rst

sem-geologist:
as bcf reader plugin is near finished, it is time to add some documentation about extra functions


doc/user_guide/io.rst
```diff
@@ -440,3 +442,65 @@ format stores all the captured data for each timestamp, together with a small
...
 the header.
+
+
+.. _bcf-format:
+
+Bruker composite file
+----------------
+
+HyperSpy can read "hypermaps" saved with Bruker's Esprit v1.x or v2.x in bcf
+hybrid (virtual file system/container with xml and binary data, optionaly compressed) format.
```
tjof2: "optionally"

doc/user_guide/io.rst
```diff
@@ -440,3 +442,65 @@ format stores all the captured data for each timestamp, together with a small
...
+Most of relevant data extraction from bcf is implemented. Both: high
+resolution 16-bit SEM imagery and hyperspectral EDX data can be retrieved simultaniously.
+
+Note that Bruker Esprit uses the similar format for EBSD data, but even if technology
+is very similar, curently EBSD containing bcf is not supported by HyperSpy.
```
tjof2: I think something along the lines of below reads better?

"Most bcf import functionality is implemented. Both high-resolution 16-bit SEM images and hyperspectral EDX data can be retrieved simultaneously.

Note that Bruker Esprit uses a similar format for EBSD data, but it is not currently supported by HyperSpy."

doc/user_guide/io.rst
```diff
@@ -440,3 +442,65 @@ format stores all the captured data for each timestamp, together with a small
...
+.. code-block:: python
+
+    >>> hs.load("sample80kv.bcf")
+    [<Image, title: BSE, dimensions: (|100, 75)>,
+    <Image, title: SE, dimensions: (|100, 75)>,
+    <EDSSEMSpectrum, title: EDX, dimensions: (100, 75|1394)>]
+
+1394 energy channel range is the result of automatic shape minimalisation
+making lesser memory usage while preserving all data.
+However in conjunction with stacking, this can cause a problem.
```
tjof2: It's unclear what you mean by "automatic shape minimalisation". Also, what do you mean by "can cause a problem" - what sort of problem?

sem-geologist: You are right: it is not clear. I don't know how to call exactly the thing. What it does? lets say output array of hyperspectra from bcf have (y, x, E) shape. Bcf contains sum EDS spectrum, so the algorithm search for last non zero value in the spectra, and sets E value in the shape to it. Now if bcf file is containing very short timed element mapping, the last non zero value in the sum spectra can be a bit different at different bcf files (stack from FIB) resulting in different shapes at E value. You made me realize that that is unnecessary stupid behavior. I implemented it because Bruker Esprit program (used to generate bcf) can be set to fixed energy ranges (10, 20, 40 or 80kv) only (that is not acceleration voltage, but range recorded in spectras regardless acceleration voltage). I think I will go back to the algorithm and change it to check acceleration voltage, compare it to the energy range (one of 10, 20, 40, 80), chose minimal and then set E to the according channel value. Then 492-494 lines will be not needed anymore.

doc/user_guide/io.rst
sem-geologist: yes, I will change it, thank you.

PR #939, opened 2016-04-05
string metadata in hdf5 files saved with Py2 was not UTF

### LINKED ISSUE #938: Error when adding elements to an EDS Spectrum in hyperspy v. 0.8.4 with Anaconda 4.0 python v. 3.5
I just did a clean install of the latest Anaconda python and installed hyperspy 0.8.4 via pip, creating a test environment with the supplied yml file in the git repository. 

When I create the test BAM EDS spectrum

> spec = hs.example_signals.EDS_SEM_Spectrum()

The metadata is fine. I wanted to test adding an element. There is a small Ar peak (it is a sputtered std...)

I can check the type of the existing array...

> spec.metadata.Sample.elements
array([b'Al', b'C', b'Cu', b'Mn', b'Zr'],  dtype='|S2')

which is as expected. When I add the element

> spec.add_elements(['Ar']) # it is sputtered...

I get an error that traces to line 297 _signals/eds.py (the add_elements function) that notes

297         self.metadata.Sample.elements = sorted(list(elements_))
TypeError: unorderable types: str() < numpy.bytes_()

I also tried

> spec.add_elements([b'Ar'])

but that did not help.

I can do this by brute force...

> el = np.array([b'Ar'])
> spec.metadata.Sample.elements = np.append(spec.metadata.Sample.elements, el) 

and add the element. Everything plots fine. But the add_elements() function should work...

francisco-dlp:
Fixes #938

### DIFF
--- a/hyperspy/io_plugins/hdf5.py
+++ b/hyperspy/io_plugins/hdf5.py
@@ -396,10 +396,11 @@ def hdfgroup2dict(group, dictionary=None):
                 value = None
         elif isinstance(value, np.bool_):
             value = bool(value)
-
-        elif isinstance(value, np.ndarray) and \
-                value.dtype == np.dtype('|S1'):
-            value = value.tolist()
+        elif isinstance(value, np.ndarray) and value.dtype.char == "S":
+            # Convert strings to unicode
+            value = value.astype("U")
+            if value.dtype.str.endswith("U1"):
+                value = value.tolist()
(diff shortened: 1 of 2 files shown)


to266:
...where is travis? :octopus: 

anyway, this works and is OK

francisco-dlp:
The merge?

to266:
Oops, I did not click!

Issue #884, opened 2016-02-26
ENH: Force positive EDSModel components


vidartf:
It would make sense if the default fitting of `EDSModel` components (i.e. the Gaussians) were bounded to be positive. For this to be the case `EDSModel.fit()` should switch the default values of its keyword args to `fitter='mpfit'` and `bounded=True`.

dnjohnstone:
This has got to be sensible - just requires switching those keyword defaults right, or is it any more subtle?

So shall one of us just do it?

francisco-dlp:
Easy and pretty useful, +1000

dnjohnstone:
Ok - see #946

tjof2:
Closing since #946 was closed and no one looks to have complained in 3+ years. Besides, leaving the decision to bound parameters up to the user is probably better.

ericpre:
Well, I still think this would be a good thing to revisit at some point, particularly that since then, there have been some progress with respect to fitting - at the time, the issue with #946 is that the tests were failing.

tjof2:
Good point @ericpre - reopened. It may also be easier I think after #2441 since mpfit won't be needed (i.e. leastsq will support bounds properly).

tjof2:
Actually could even tackle it as part of #2441, what do you think?

ericpre:
#2441 is already quite big, but that should be fine.

tjof2:
Let's do it separately

PR #946, opened 2016-04-07
ENH: Force EDS Model to be positive by default

dnjohnstone:
To address #884


vidartf:
+1

vidartf:
Does `multifit()` need to be overridden as well?

vidartf:
There seems to be some white-space issue at the top, and if the docstring is the same as the inherited method it can be used directly, so we don't have two places to maintain:
```python
fit.__doc__ = Model1D.fit.__doc__
```

francisco-dlp:
Looks good. No, modifying `multifit` shouldn't be required.

hyperspy/models/edsmodel.py
```diff
@@ -317,6 +317,56 @@ def disable_xray_lines(self):
         """
         for component in self._active_xray_lines:
             component.active = False
+    def fit(self, fitter='mpfit', method='ls', grad=False,
```
francisco-dlp: If all the components of the model have components with the gradients defined (`Gaussian` does), we may consider setting `grad=True` as this should speed things up.

hyperspy/models/edsmodel.py
```diff
@@ -317,6 +317,56 @@ def disable_xray_lines(self):
...
+            bounded=True, ext_bounding=False, update_plot=False,
+            **kwargs):
+        """Fits the model to the experimental data
+
+        Parameters
+        ----------
+        fitter : {None, "leastsq", "odr", "mpfit", "fmin"}
+            The optimizer to perform the fitting. If None the fitter
```
francisco-dlp: The new fitter default `mpfit` needs to be documented here.

hyperspy/models/edsmodel.py
```diff
@@ -317,6 +317,56 @@ def disable_xray_lines(self):
         """
         for component in self._active_xray_lines:
             component.active = False
+    def fit(self, fitter='mpfit', method='ls', grad=False,
+            bounded=True, ext_bounding=False, update_plot=False,
+            **kwargs):
+        """Fits the model to the experimental data
```
francisco-dlp: Maybe explain here that this, by default, perform bounded optimisation using mpfit?

francisco-dlp: It may be a good idea to point this out in the User Guide too.

hyperspy/models/edsmodel.py
dnjohnstone: Yes I think that's a good idea

vidartf:
Oh, I think you also need to make sure that `fit_background()` works reasonably well (`bounded`/`mpfitter` is likely not needed(?)).

dnjohnstone:
Ok - this has broken much more than I was expecting it too... Since the implementation in master is working broadly ok (at least in all cases I've tried it in so far - except for one) I'm not in a great rush and will get to this properly after Model2D stuff, which I think is more of a priority, at least for me.

francisco-dlp:
The main issues seems to be that `Polynomial` doesn't compute the gradient correctly. I'll report this in a new issue.

francisco-dlp:
#950

dnjohnstone:
There's something slightly surprising going on here (at least to my mind) that swapping the fitter to mpfit with bounded parameters seems to be reducing the precision of the result (this is what's making travis fail) is that an expected behaviour @francisco-dlp

francisco-dlp:
The default tolerance values may be different for mpfit and leastsq. It's worth investigating is playing with the mpfit default would increase the accuracy of the results.

francisco-dlp:
Any further input on this?

dnjohnstone:
Only that I have tried playing with various mpfit parameters but I can't make it pass the tests yet... The test that is failing here is out by about 10% which is quite a lot. I haven't spent too much time on figuring out where to dig first  - it might be that the automatic bounds are wrong or something I guess....

How much have we tested mpfit in general?

hyperspy/models/edsmodel.py
```diff
@@ -318,6 +318,62 @@ def disable_xray_lines(self):
...
+
+        Parameters
+        ----------
+        fitter : {None, "leastsq", "odr", "mpfit", "fmin"}
+            The optimizer to perform the fitting. By default mpfit is used for
+            EDS models, as this supports bounding to force the components in the
+            model to be positive. This is inline with the physical nature of EDS
+            spectra. All optimizers available for Model1D remain availble and
+            can be specified by the user. fmin is the only one that supports
+            maximum likelihood estimation, but it is less robust than leastsq
```
tjof2, line 352: I don't get this - `it is less robust` - in what sense? MLE achieves the Cramer-Rao Lower Bound. Least-squares and MLE should be equivalent in the case of Gaussian noise, otherwise something is up.

tjof2:
I don't fully understand the `mpfit` implementation of bounds, what with the Levenberg-Marquardt algorithm being unconstrained and all. Also, is there any reason why `mpfit` is preferred over e.g. `scipy.optimize`?

hyperspy/models/edsmodel.py
dnjohnstone, line 352: This a paraphrasing of what is written in the docstring for the main fit method, which reads 

""fmin" performs curve fitting using a downhill simplex algorithm.
            It is less robust than the Levenberg-Marquardt based optimizers,
            but, at present, it is the only one that support maximum likelihood
            optimization for poissonian noise."

To be honest I wouldn't claim to know very much about fitting algorithms in general, but I guess it's referring to the use of the simplex algorithm in the original and perhaps the loss of that emphasis in this paraphrasing is the problem?

Would that square with you @tjof2 or do you take issue with the original too?

dnjohnstone:
Unfortunately I'm very much in the same boat here @tjof2 - the main thing this PR has taught me is that I need to understand the fitting algorithms better... Perhaps @to266 or @francisco-dlp can help?

tjof2:
@dnjohnstone I'm guessing here, but working from `model.py`:

> "fmin" performs curve fitting using a downhill simplex algorithm. It is less robust than the Levenberg-Marquardt based optimizers, but, at present, it is the only one that support maximum likelihood optimization for poissonian noise.

The question is then why does the documentation only mention simplex (Nelder-Mead) over BFGS, LBFGS, BOBYQA, CG etc...which are all imported at the top of the file? I don't like the use of the word "robust" in either case.

Out of scope for this PR probably, but I was thinking out loud about bound constraints and the effect on the tolerance / result from mpfit.

hyperspy/models/edsmodel.py
tjof2, line 352: I take issue with the original, but I'll raise that separately. Personally I would leave out any mention of "robustness" at all since a) it's unclear what that means here and b) the take-home point is surely MLE for Poisson noise.

dnjohnstone:
I think it's quite important actually - I was really expecting this to be pretty trivial, and was quite surprised that the tests failed this way... Honestly, the problem right now is that I'm not absolutely certain whether the issue is that I used something incorrectly, whether there was an issue with the test (e.g. whoever wrote them just put the numbers they knew fmin returned), or whether there's something more important going on. All I know is I change a tonne of fitting parameters and didn't get it to pass the test yet.

In any case it's probably telling that I can't work out which of those it is...

hyperspy/models/edsmodel.py
francisco-dlp, line 352: That's the theory. In practice it depends on how good the optimiser is at finding the minimum for a particular problem (curve fitting in this case) and the Levenberg–Marquardt algorithm turns out to be pretty good at it.

francisco-dlp:
I didn't get through this in detail, but you have decreased the tolerance on the output of a function that operates on the result of the fit. The precision seems to decrease in this case, but it is not straightforward to conclude that this also true for the result of the parameter estimation. I vaguely remember testing mpfit vs leastsq when I was considering it for inclusion in HyperSpy and finding that they were converging to the same values when bounding was not necessary. I think that @to266 has more recent experience on comparing the too. In any case, this is out of the scope of the PR. Even if there was a precision penalty on a dataset that doesn't require bound (which I doubt, but requires verification) setting the default to bounded would provide better results in most cases.

francisco-dlp:
In other words, I am happy to merge this unless somebody disagrees.

dnjohnstone:
@francisco-dlp thanks - but don't merge it yet, it's still failing one test outright with a difference of about ten percent... will need to solve that to ensure the build passes.

I'm also in no rush... so don't worry too much unless you're very keen.

hyperspy/models/edsmodel.py
dnjohnstone, line 352: Surely the thing to do here is exactly what we would do in any other academic context and insist on referencing appropriate literature that actually demonstrates this or at least gives some evidence...

Who says Levenberg-Marquardt turns out to be pretty good at it? And what does that even mean?

tjof2, line 352: As I referred to in #1098, I suspect it might be a result of a comparison between gradient-based (LM) and non-gradient based (Nelder-Mead) - it's perhaps not surprising that providing extra information to the fitting algorithm (i.e. the derivatives) leads to a better result.

tjof2:
@dnjohnstone with the new `leastsq` functionality introduced in #1169 now including bounds, does it fail tests if you replace `mpfit` with `leastsq`?

francisco-dlp:

dnjohnstone:
Closing this as it seems to be an issue that hasn't bothered anyone in a long time and benefits seem minimal to me...

Issue #932, opened 2016-04-01
Link dev guide in docs


vidartf:
The [developer guide](http://hyperspy.readthedocs.org/en/latest/dev_guide.html#write-tests-documentation) ( #794, #782 ) is currently not linked from anywhere. It should be included on the front page and in the side bar on the same tree level as the user guide

dnjohnstone:
Ah yes - this will be because I actually have no idea how the website works, so I just put the dev-guide in a folder that looked about right and hoped something magical would happen.

Would be happy to do this if someone points me in the right direction - on the other hand if someone does it, I'll be glad to have learnt something!

vidartf:
See #908 / #909.

dnjohnstone:
This can be closed.

Issue #798, opened 2015-12-15
Random failure of test `hyperspy.tests.mva.test_decomposition.TestGetExplainedVarinaceRation.test_data`


vidartf:
`hyperspy.tests.mva.test_decomposition.TestGetExplainedVarinaceRation.test_data` keep failing at seemingly random in the CI builds. Is there a reason for this? Is it possible to fix it?

francisco-dlp:
Yes, I've noticed previously those hard to understand random failures in such a simple test. May `np.empty` be the one to blame?

dnjohnstone:
I guess this might be with reference to the pull request I just made for absorption corrections? I think I had this the other day with Model2D as well one of Tomas's suggestions suddenly broke this test (on my local machine) - seemed quite odd.

vidartf:
I've had it as well. I think you are right @francisco-dlp , as the docs state:
> out : ndarray
Array of uninitialized (arbitrary) data with the given shape, dtype, and order.

vidartf:
`np.zeros` should fix it?

francisco-dlp:
I've tested it and it fixes it, I'll PR in a second.

2015-12-15 14:46 GMT+00:00 Vidar Tonaas Fauske <notifications@github.com>:

PR #947, opened 2016-04-08
Add web link to dev guide

dnjohnstone:
Addresses #932


dnjohnstone:
The test failing Travis here is Test_fancy_indexing.... pretty sure this PR didn't touch that, not sure why how it's broken to be honest.

vidartf:
Fails for the same reason as #798 .

PR #949, opened 2016-04-09
ENH-FIX: change distutils to setuptools and add flexible cythonization

sem-geologist:
this code fixes two things:
1. changes distutils to setuptools
2. adds the flexible cythonization.

first is straight forward, but second have to be discussed.
Firstly lets consider these points:
* User can obtain hyperspy in few different ways: git, binary distribution, source distribution.
* It is highly recommended by cython documentation to distribute C files generated from cython as this makes cython not required during installation and prevents undefined behavior resulting from different versions of cython.
* cython generated C files are huge, complicated, contains lots of "boiler plates" is quite ugly and hardly readable. It is not strange that it is not very welcome in the git source tree.
* At this moment there is one library pending (bcf io_plugin) which have some functions implemented in cython, but alternative python code is also present, so even in worst scenario lack of cython should not be an obstacle for installment.  

Considering above the setup should meet such criteria:
* do not have hard dependency on cython
* automatically check if *.pyx files have c files with same name, if not:
    * try to import cythonize from Cython and use on the pyx files
    * if there is no cython present on the system, user should be informed and present of alternatives, but the installation should be not terminated as without cython code hyperspy is still fully functional, just in some minute parts where cython code would be missing it will be a bit slow.
* if there is cythonized c files present, do not cythonize the code, unless the new "recythonize" command is used: ```python setup.py recythonize```. ~~special user option are added to ```python setup.py install --force-cythonization``` (or to ```develop``` which will be used for testing env)~~
* for development purpose if source contains '.git' - generate a post-checkout hook which removes only cythonized *'.c/.cpp'* and compiled *'.so' or '.pyd'* files. 
 

TODO:

* [x] add trial to cythonize extensions if c/c++ files are missing
* [x] add new command "recythonize" to be used with setup.py to recythonize all extensions 
* [x] update CI and anaconda configs
* [x] add git hook to automaticaly recythonize and rebuild --inplace after git checkout
* [x] add travis ci testing for osx
* [x] add bdist_wheels deployment to release from travis ci (for osx) and sdist (for linux and other)
* [x] test travis release works
* [x] add bdist_wheels deployment to release from appveyor
* [ ] test bdist_wheels deployment to release works from appveyor (it does not deploy but it builds and preserves the artefacts at appveyor)
* [x] test `python setup.py install` works without compiler present (ubuntu live usb: passed; appveyor: too much interfered setup; windows 7 with fresh winpython + traits(from gohlke's place): passed)


setup.py
```diff
@@ -28,9 +28,9 @@
     print(error, file=sys.stderr)
     sys.exit(1)
 
-from distutils.core import setup
+from setuptools import setup
 
-import distutils.dir_util
+import setuptools
```
sem-geologist: actually, this could have been untouched as only thing it is used is to clean build folder.
Alternatively it could be ```from setuptools import distutis.dir_util```.

francisco-dlp:
This does more than it advertises. In addition to re-switching to setup tools (this was done in #849, but possible it was reverted in a merge) this adds functions to cythonize cython extensions and a test to check that the cython extensions are correctly installed. That's fine, but could you edit the description of this PR to better reflex what it does?

The cythonize functionality seems to WIP. Let us know when this is ready for review.

sem-geologist:
just upgraded the cythonization during installation:
1. it checks if there is c file for every pyx file, if not then calls cythonize, if there is no cython library it prints warning about it but does not stop.
2. install have additional user option ```--force-cythonization``` if i.e. user have newer/older cython/c and wants to recythonize existing c code.

I added dummy pyx in misc dir to check how cythonization works. It is either to be removed in the end of this feature development, or I could add some testing in tests, to see it works properly.

setup.py
```diff
@@ -54,6 +55,57 @@
...
+        if os.path.exists(path+'.c'):
+            ext.append(Extension(path.replace('/', '.'), [path + '.c']))
+    return ext
+
+
+def cythonize_extensions(extension_list):
+    try:
+        from Cython.Build import cythonize
+    except ImportError:
+        print("""WARNING: cython required to generate fast c code is not found on this system.
```
francisco-dlp: This should use a proper warning instead i.e. `warnings.warn`

hyperspy/misc/test_cython_integration.pyx
```diff
@@ -0,0 +1,7 @@
+def testing_cython():
```
francisco-dlp: All tests should be in `hyperspy/tests`

francisco-dlp:
Installing without Cython currently fails like this:

```python
jd29@SATELLITE-PRO-C850-10N:~/git/hyperspy$ ana3-python3.5 setup.py build_ext --inplace --force-cythonization
usage: setup.py [global_opts] cmd1 [cmd1_opts] [cmd2 [cmd2_opts] ...]
   or: setup.py --help [cmd1 cmd2 ...]
   or: setup.py --help-commands
   or: setup.py cmd --help

error: option --force-cythonization not recognized
fjd29@SATELLITE-PRO-C850-10N:~/git/hyperspy$ ana3-python3.5 setup.py install --force-cythonization
running install
cythonizing the *.pyx source
WARNING: cython required to generate fast c code is not found on this system.
Only slow pure python alternative functions will be available.
To use fast implementation of some functions writen in cython either:
a) install cython and re-run the installation,
b) try alternative source distribution containing cythonized C versions of fast code,
c) use binary distribution (i.e. wheels, egg).
  File "setup.py", line 254, in <module>
    "Topic :: Scientific/Engineering :: Physics",
  File "/usr/lib/python3.5/distutils/core.py", line 148, in setup
    dist.run_commands()
  File "/usr/lib/python3.5/distutils/dist.py", line 955, in run_commands
    self.run_command(cmd)
  File "/usr/lib/python3.5/distutils/dist.py", line 974, in run_command
    cmd_obj.run()
  File "setup.py", line 106, in run
    orig_install.run()
TypeError: run() missing 1 required positional argument: 'self'


```

francisco-dlp:
And with `Cython` installed it fails exactly in the same way.

sem-geologist:
please, patience, it is still work in progress
:)

sem-geologist:
@francisco-dlp , so at this moment it should work. If you check the travis logs (under the 340b769
 commit), you will see it prints the warning (I will change it to use warning.warn), because travis have no cython installed in its environment at this moment. So that is how it will look, if somebody obtains a source code straight from github, but will have no cython installed. Now I am going to modify the travis to install cython. And pyx files should be compiled then.

I can remove the test file in the end before the merge, or move it to tests. Now I want to be sure pyx files in hyperspy normal directories can be compiled.

sem-geologist:
oh, and by the way, build_ext will not work at this moment as it is still not subclassed, just install is subclassed at this moment. Also ```build_ext --inplace``` using the ```setuptools``` can be changed rather to ```develop``` as it does the same thing

francisco-dlp:
I think that travis should indeed always cythonize. In this way, we'll know if a new version of Cython breaks something.

Regarding the test, I think that it'll be best to move it to `hyperspy/tests` as, in case that some other Cython extension gets broken, if this one passes we'll know that it is most likely not the fault of `setup.py`.

Probably it'll be also good to add an extra travis and appveyor build to install without the cython extensions to verify that everything goes well too in this case.

Finally, I think that this requires an entry in the developers guide explaining best practices for including Cython code in HyperSpy and how to install in devel mode. Now that we have the `Developers' Guide` maybe we should move [this](http://hyperspy.org/hyperspy-doc/current/user_guide/install.html#development-version) to the devel guide and updated with instruction on how to properly do this now that we include Cython code too?

sem-geologist:
To add thing to development guide is a very splendid idea.
However at the moment I feel a bit like a blind person in the maze. I need to sort out what is the best approach to this case because I start to think I get myself into the trap. Probably I need to move the stuff away from install to build_ext, as that is called by all commands (build, develop, install).  Setuptools are very highly using distutils (and actually if build_ext is called it is mainly from distutils), and I need to get more familiar how distutils works (trying to figure out by reading source code). I wonder how is it possible that documentation on such important python toolchain is so lacking.

francisco-dlp:
I  agree that the documentation on this issue is not great.

I like [sklearn's approach](https://github.com/scikit-learn/scikit-learn/blob/master/setup.py#L278): cythonize always except when `PKG-INFO` exists.

sem-geologist:
We can do the same (I even would be more happy with it), but then hours (actually weeks) spent on python pure implementation was spent for nothing (at least for me on some parts of bcf library). Sometimes users can't get compatible binary (differnet versions of gcc/clang/msvc/... used to compile python or can't have compiler: then python only functions is the only choice. scikit probably have hard cython requirement as some functions have just cython implementation. On hyperspy there is still no need to do hard cython requirement as at this moment there is no such functions which would require it. Hyperspy at this moment already is quite heavy on requirements and needs so many libraries.

francisco-dlp:
Sorry that I didn't explain myself again. I think that it's excellent that you have implemented a pure python version of the reader and I'm not at all suggesting to drop it.

sem-geologist:
* I understood that ```install, build, develop...``` goes through chain of commands, and to add ```--force-cython``` to every of them would be not clever.

So my lates plan:

1. get rid of ```install``` overiden/subclassed class, instead override/subclass ```build_ext``` which will do the same check what install now does and try to cythonize code if there is no c files for given pyx.
2. get rid off user option --force-cython (and stupid idea to add it to every cmd) instead make a new command called build_cythonize. which would do just exactly that, despite the presence of c it would try to cythonize the pyx (and overwrite if needed the c files). That would be useful for people who would develop cython code in hyperspy or for users who would want to tinker around cython implementations (using newer cython and so on...).

so calling anything standard (install, build, develop, build_ext...) automatically would check if there is c files, if it is not then try to cythonize pyx (all main commands by default calls build_ext at some point in the chain).

francisco-dlp:
Wouldn't it be better to always cythonize when calling `setup.py` except when `PKG-INFO` exists? There can always be old versions of the C files that need refreshing...

sem-geologist:
there will be no cythonized C files in git anymore. For refreshing build_cythonize command could be used. But generally it should not try to do that without user consent. cythonization should be called only if there is no c files, or only if user thinks it should be (re)cythonized. That is stated and strongly recomended by cython community.

francisco-dlp:
What would be the best way to guarantee that the C files are always fresh? Normally developers install HyperSpy using e.g. `pip install -e .` and they don't usually execute `setup.py` anymore. However, when e.g. switching branches containing different versions of cython files, the'll need to cythonize those files. Are they expected to do this manually? Or could we maybe provide a git hook that detects when a pyx file changes and cythonizes automatically? Or maybe there is a better solution out there?

sem-geologist:
they cant stay not touching setup.py, they still have to do ```build_ext --inplace``` if they want to try out different branch...

sem-geologist:
under commit 1d3e026 in travis ci it can be seen how it looks if there is no cython installed on build environment. Next I am adding to the travis and appveyor cython dependencies.

sem-geologist:
I am not sure I should add dependency of cython to anaconda_hyperspy_environment.yml...

sem-geologist:
@francisco-dlp , I added the post-checkout generator to setup.py, this should do exactly what you wanted. It is made so that first time setup.py is called it checks if there is .git folder and makes a git hook.
It is supposed to work on all OS'es.

sem-geologist:
Ok... I changed hook a bit. now anytime setup.py is touched the hook is regenerated. In the hook list of *.c and *.so and *.pyd files are writen to be removed after ```git checkout```. So checking out to older versions will not complain about recythonize command missing. and if branch is new this way will force cythonize :)

basically I am done, this is ready for review :)

hyperspy/tests/misc/test_cython.py
```diff
@@ -0,0 +1,22 @@
...
+import os
+
+my_path = os.path.dirname(__file__)
+
+def test_cythonization():
+    try:
+        import Cython
+    except ImportError:
+        warnings.warn("""no cython library were found, aborting testing of cythonization""")
+    else:
```
francisco-dlp: I think this `else` statement should go.

hyperspy/tests/misc/test_cython.py
```diff
@@ -0,0 +1,22 @@
...
+import warnings
+
+import os
+
+my_path = os.path.dirname(__file__)
+
+def test_cythonization():
+    try:
+        import Cython
+    except ImportError:
```
francisco-dlp: Instead of raising a warning you could simply skip the test:

```python

from nose.plugins.skip import SkipTest

raise SkipTest

```

hyperspy/tests/misc/test_cython.py
```diff
@@ -0,0 +1,22 @@
...
+        warnings.warn("""no cython library were found, aborting testing of cythonization""")
+    else:
+        if not os.path.exists(my_path + '/cython/test_cython_integration.c'):
+            raise DeprecationWarning('deprecation of cythonization of pyx extensions detected!')
```
francisco-dlp: This should probably raise an error, not a `DeprecationWarning`. Something like "Cythonization of <file path> failed"

hyperspy/tests/misc/test_cython.py
sem-geologist: Nice, I didn't knew that, I was wondering how to do it (I sow some skipped tests in Travis CI conclusions) I am changing it right now.

hyperspy/tests/misc/test_cython.py
```diff
@@ -0,0 +1,22 @@
...
+
+def test_extensions_built():
+    try:
+        import hyperspy.tests.misc.cython.test_cython_integration as test_cy
+        nt.assert_equal(7, test_cy.testing_cython())
+    except ImportError:
+        raise DeprecationWarning('deprecation of building extensions detected!')
```
francisco-dlp: I think that this should raise an error too.

sem-geologist:
@francisco-dlp , I made proposed fixes. what about the rest files. Is the git hook close to what you thought about?

hyperspy/tests/misc/test_cython.py
```diff
@@ -0,0 +1,24 @@
...
+my_path = os.path.dirname(__file__)
+
+def test_cythonization():
+    try:
+        import Cython
+    except ImportError:
+        raise SkipTest
+    
+    if not os.path.exists(my_path + '/cython/test_cython_integration.c'):
+        raise RuntimeError('deprecation of cythonization of pyx extensions detected!')
```
francisco-dlp: Maybe the text should be something like "Cythonization of <file path> extension failed"?

hyperspy/tests/misc/test_cython.py
```diff
@@ -0,0 +1,24 @@
...
+
+def test_extensions_built():
+    try:
+        import hyperspy.tests.misc.cython.test_cython_integration as test_cy
+        nt.assert_equal(7, test_cy.testing_cython())
+    except ImportError:
+        raise RuntimeError('deprecation of building extensions detected!')
```
francisco-dlp: Maybe the text should be something like "Compilation of  <file path> extension failed"?

francisco-dlp:
The git hook does the job. However, the downside is that switching branches now takes longer. Also, if it overwrites `post-checkout`, and some may like to have their own. Ideally it should only recompile the pyx files that change on checking out, see #181 for a post-commit hook that filters the files that have changed in the last commit. Anyway, for this PR, I think that it'll be enough to move the post-commit hook to the Developer's Guide to avoid the overwriting issue. As this is only of the interest of developers, that should be enough.

The remaining issue is the automatic creation of wheels on release by Appveyor (for Windows) and ?? for Mac. For Appveyor, could you try adding the lines that @vidartf suggested and creating a release in this branch on your own repository afterwards? That should trigger an Appveyor build which should attach the windows wheel to the GitHub release. For Mac, I've found this: https://github.com/MacPython/wiki/wiki/Wheel-building, what do you think?

hyperspy/tests/misc/test_cython.py
francisco-dlp: Also, wouldn't it be a good idea to recommend running `setup.py build-ext...` again to fix the issue?

hyperspy/tests/misc/test_cython.py
francisco-dlp: Also, wouldn't it be a good idea to recommend running `setup.py build-ext...` again to fix the issue here too?

sem-geologist:
@francisco-dlp , 
1. about RuntimeError texts: tests checks if automatic cythonization and compilation works. It is tests meant to be used with appveyor and travis. If tests fail they indicate that something somethere got changed, what affects the routine. The touching setup.py atomatically cythonizes. I will change first test error message to:
```python
"""Automatic cythonization of extension failed
try to run 'python setup.py recythonize' to find out the bug introduced"""
```

second one I would change to:
```python
"""Compiled code is missing in source directory:
1. Check out that testing environment compiles code in the source directory
  by calling 'python setup.py build_ext --inplace', before running the test.
2. Check out you have compiling environment (gcc/clang/msvc..)
  same as used python version on your OS.
3. Check out that cythonization of code works. I.e. try running 
  'python setup.py recythonize' """
```

2. Hooks. 
I could add the code to check if file exist and first line is '#!/bin/*sh', (if not, just leave it alone)
    then try finding line begining with 'rm hyperspy/\*' and next line '\*python\* setup.py build_ext --inplace'.
         on success replace the 'rm hyperspy/*....' with new list
    else: append those at the end of the file.
Additionally I can search for '.hook_ignore' file which would prevent setup.py from touching the hooks at all.

What do you think about it?
Of course I will add more explanation how hooks works and how to prevent it in dev guide.

sem-geologist:
I am not very familiar with wheels, I think with setup tools it will be easier. what about eggs? why wheels?

sem-geologist:
@francisco-dlp ,
I tried to do sdist (changed MANIFEST.in to include pyx files) succesfuly on my local machine.
bdist I see build .so files, but does not include the source (neither pyx neither C). Should that be so intentionally?

francisco-dlp:
We also run the tests locally so, unless we decide to include the hook, the sentence "Automatic cythonization of extension failed try to run 'python setup.py recythonize' to find out the bug introduced" may not properly apply to nosetests run locally. The other message is really helpful. The language may need some polishing, maybe @tjof2 would like to help here too?

wheels seem to be widely accepted these days at the solution for distributing compiled code with Python. All major projects that ship with compiled code use them. Just adding the line that @vidartf suggested to appveyor should be enough to create wheels for windows.

Regarding the hook, it'll be simpler to add/remove a single line that executes an external script if the file exists. If we decide to include the hook in `setup.py`, then I think that it'll need to be the improved version with filtering of the pyx files that change during the operation. Otherwise is fine for me to leave it as it is and add it to a recommend git hooks sections in the Dev Guide. Any opinion on this @developers?

sem-geologist:
@francisco-dlp ,
the hooks is already pushed as I described. Just try out, make your own post-checkout hook, and see how it appends to the end. Or wait until the hook is created, add your stuff and see how it will handle nicely running git checkout. Try then adding .hook_ignore see how setup does not touch it anymore. This is post-checkout hook, it should clean everything what is not tracked by git but could cause some stupid errors (that means that it is intended to clean all .c/cpp and .so/.pyd files, which residual can make confusion in branch where it is not supposed to be. It does not compile or cythonize if file is changed. It just removes the every existing cythonized and compiled products. Then it runs setup build_ext --inplace. If the branch we checkout in is new kind of branch it will rebuild extensions from scratch. And if it is something older (before it will be merged with upstream) will have no side effect as they have no == empty extensions. Why and how? Because it is setup.py which triggers cythonization depending from situation, not the hook. Hook only cleans up. setup.py  finds that some of cythonized (in our case all) c/cpp files are missing so automatically tries to import Cython and cythonize the extensions. And that is what test which Error message you discuss above is doing. It was misunderstood, it does exactly what it say, it have nothing to do with the hooks, it tests automatic cythonization if cythonized files are missing. It tries to import Cython, if it fails it skips the test as it is irrelevant. But if (1) Cython is on the build system and (2) cythonized extensions (c or cpp) are missing then it means either a) setup.py is not touched before the testing. (no build* or install used) b) ```build*``` was called, but went wrong. By running ```python setup.py recythonize``` it can be found if it is the problem with setting up testing environment or building is failing (in example somebody changes cythonize parts in setup.py).

Now if somebody is developing on the same branch with cython code, they should use ```python setup.py recythonize build_ext --inplace``` - that will recythonize only the changed code. However I have to admit that build_ext --inplace will rebuild all c/cpp files. Problem is actually the code which was before I did anything at setup.py:

```python
if os.path.exists('build'):
    distutils.dir_util.remove_tree('build')
```

I can't understand why it is there, as ```python setup.py clean --all``` would do the same thing and it is hardly possible windows or linux environments would get that folder as it is ignored by git, from which it is pushed to separate built environments for different OS.
Everytime setup is called it wipes build directory, so it is not a surprise ```build_ext``` is recompiling all extensions every time it is called.
If those lines are commented out then setuptools build_ext --inplace is not recompiling all the code, but just the changed one. In lates commit I commented those lines out.

My provided hook generator, I think is working so as it should. It is clean the compiled stuff completely when changing the branches. In the lates commit I added the clean --all before build_ext --inplace to the hook to get rid of build folder as well so after changing branch it should build up everything from a scratch. Creating the .hook_ignore empty file in the folder will prevent from any changing of hook by setup.py

sem-geologist:
@francisco-dlp, could you exactly tell me what do you want me to do with wheels. 
1. If I do sdist it adds the 0.8.4 to the name of source archive. You were talkign about different version during the merge of branches (?) and releases. How to control and pass that to appveyor (or Travis if we will find out how to do osx wheels)
2. as I understand sdist will be the only option (as wheels is no available on linux?). should it be add to travis to make ```sdist``` on linux builds, and sdist + bdist_wheels for osx?

I see the appveyor already do some deployment (I am not sure what it do exactly)  for windows stuff in *exe form? (pre_deploy and deploy code in .appveyor.yml.). There I see it requires some kind of password/encryption code under auth_token, and I see there is commented out one of @vidartf and actually used at the moment @to266 . I am not sure where exactly to add the bdist_wheel. I tried in previous commit to add bdist_wheel under deployment, but then appveyor does not even fire up. (Probably I need to provide my auth_token and comment to266 one?)

If it does deployment in exe why we need wheel?

sem-geologist:
after I added these lines:
```yaml
after_test:
  - python setup.py bdist_wheel

artifacts:
  - path: dist\*
```

in between ```test_script``` and ``before_deploy``` sections at the end the appveyor attaches whl and egg bdist archives as downloadable in the artifacts tab (That is coolest thing I saw! That will save me headache with other cross platform projects, this way I don't need to download MS Visual Bloat ). At first I thought it braked the deploy steps following... but I inspected the artifact tabs and logs from other branches and see no deployment takes place. So probably it doesn't brakes the deployment as it didn't worked from the begging.

sem-geologist:
I rearranged the travis.yml a bit to get osx testing and building working. Basically I got into few problems and I think there are simple solutions:
1. (this is not a problem, just a remark) differently than most of examples in internet "how to setup travis ci for osx", hyperspy is less complicated because everything is done through ```minconda```. The downloaded miniconda have only one python version and most of present travis code is usable same as on linux.
2. First trouble I ran into was that plain travis osx environment have no ```wget```. I added one liner bash loop to check if it is osx, and use curl instead with mac link, which downloads the osx conda installation as ```miniconda.sh```. The alternative would be to use detect osx and install wget throught brew before downloading the sh file (this have no gain). This is long line, maybe it could be divided into two (with two if's? (if osx -> curl osx_miniconda.sh; if linux -> wget linux_miniconda.sh ?) as yml is not suited to anything more than one liners(?).
3. next trouble was the ```bash miniconda.sh -b -p python=$TRAVIS_PYTHON_VERSION $DEPS``` line. For osx on travis both ```bash``` and ```python=``` are offending. I rewrite it to be more universal:
```./miniconda.sh -b -p $DEPS``` so it will use the proper shell according to the os and miniconda.sh. I removed ```python=$TRAVIS_PYTHON_VERSION``` as it is irrelevant and not understood under osx version of conda. Conda environment anyway have just one python version and just one python version is used in test environments.
4. Travis have no artifacts tab (such as appveyor have) and I struggle to understand end where/and how I need to push/deploy those built binaries or source distributions. Because tests hangs on osx (it hangs for 10min and fails completely) I had tried bdist_wheel temporary before running tests and it worked. At the end it should be under ```after_success``` section. However the biggest problem is that osx tests freezes (not fail) and times out not reaching that section.
5. I put temporary --verbose flag to nosetest, and found out that it is exactly ```test_hdf5.TestLoadingOOMReadOnly.test_in_memory_loading``` which hangs the travis on osx. By temporary commenting the test out I checked that at this moment it is the only test hanging osx test. However there is also one test which fails only on osx #949, which is cause due to comparing floats with too much significant digits, and can simply be fixed, by changing the assert method.

**Comment by francisco-dlp** [comment]
@sem-geologist, amazing progress!

It looks like you've fixed mostly all issues.

Regarding deploying in mac, [wouldn't this be relevant](https://docs.travis-ci.com/user/deployment/releases)?

@vidartf and/or @to266 what about the appveyor's  auth_token?

Regarding wheels, I was proposing to produce them for the releases only, not for the branches. Although currently there are no wheels for Linux, this is not a big issue as compiling in Linux machines is far easier than in the other platforms. Regarding why distribution the `exe

**Comment by francisco-dlp** [comment]
(I've posted the above message before finishing it by mistake, I'll update it above, so read it online for the complete version.)

**Comment by sem-geologist** [comment]
I still don't understand how the release works. are the some tag in the git pushed to mark it as release? or travis is edited for release? how travis will know that that particular commit is release?

File: setup.py [hunk 10]
```diff
@@ -54,6 +55,131 @@
...
+    extensions = cythonize_extensions(raw_extensions)
+else:
+    extensions = no_cythonize(raw_extensions)
+
+
+# HOOKS ######
+
+def find_post_checkout_cleanup_line():
+    """find the line index in the git post-checkout hooks"""
+    with open('.git/hooks/post-checkout', 'r') as pchook:
```
vidartf: This assumes the current directory is hyperspy repo root folder. You can use `os.path.join(os.path.dirname(__file__), '.git/hooks/post-checkout')` instead.

setup.py
```diff
@@ -54,6 +55,131 @@
...
+        hook_lines = pchook.readlines()
+        for i in range(1, len(hook_lines), 1):
+            if re.search(r'^rm hyperspy/*', hook_lines[i]) is not None and \
+              (re.search(r'python.*?build_ext --inplace', hook_lines[i+1]) is not None):
+                return i
+
+# generate some git hook to clean up and re-build_ext --inplace
+# after changing branches:
+if os.path.exists('.git') and (not os.path.exists('.hook_ignore')):
```
vidartf: Same path issue here and below.

vidartf:
@francisco-dlp  I assume the `exe` things mentioned is the bundle installer?

@sem-geologist Currently the deployment step is only run if a git tag is pushed to the github repo. See http://www.appveyor.com/docs/deployment/github . That link also documents the auth token.

francisco-dlp:
Just click on "releases" in your branch and create a new release. This should trigger the travis machinery that should end up uploading the file to the release if everything goes well. Obviously, for the main repository we only create releases when we release a new version.

francisco-dlp:
I don't know what @sem-geologist had in my, I was thinking on the Windows installers that we distribute through pypi.

sem-geologist:
1. appveyor provides nice way to get the artifcats per commit. I think it is handy, as the commits can be easily tried on windows (no need for visual bloat environment >9GB). However it should be add to the release too (I tried touching it, but then appveoyr even didnt fired up, I guess due to that author_key authentication). I just am not sure if this is not overuse of appveoyr (it keeps artefacts from every commit)
2. On travis there is no artifacts saved, and proper deploy mechanism should be done. 

If I understand clear. I should make and test the release deployment to my branch. After commit it would be changed to deploy to main hyperspy release repository (are there such?)?

sem-geologist:
This can be a bit offtopic, but Is not on osx same easy to install packages from source as on linux? I guess it should be at least miles away better than windows?

francisco-dlp:
> If I understand clear. I should make and test the release deployment to my branch. After merging it to upsteream it would be changed to deploy to main hyperspy release repository (are there such?)?

I think that this is a reasonable way to proceed. I agree that providing wheels per-commit for Windows is useful.

Regarding the mac question, I have no clue as I have only barely touched a mac.

sem-geologist:
> Just click on "releases" in your branch and create a new release.
... Where I have to click? Ther is no releases in my branch...

sem-geologist:
I can't finish this, I can't understand how deploying works and how to test it. I am complete noob on that. 
If I push tag (where exactly I have to push it) nothing, happens, new release appears under 'releases, but no appveyor or travis fires up. Could somebody shed the light what exactly I should do to trigger and test release building and deploying on my branch?

to266:
Well you have to setup both travis and appveyor for your hyperspy repos. Have you done that?

sem-geologist:
@to266, I am complete noob. (Probably it is already obvious). 
I will try to explain what I understood I need to do to finish this pull request:
I need to implement and test how building wheels for windows and mac (and sdist for linux) works for releases. 
I couldnt find what I have to do exactly to force code throught deploy part:
1. there is appveyor and travis going on my pull request, should I sign up to appveyor and travis and somehow enable it for my branch of hyperspy to test the building release? or is it enought to use the appveyor and travis running on testing commits.
2. how to trigger travis/appveoyr to go through deploy part? How tags of git works, I read it, but it does not works. (I tried to make new tag. push the tag, then after ```git push --follow-tags``` but it does not make travis/apveyor do any deploy step.

I think I add enought to appveyor yml and travis yml to make the binaries but can't find how to trigger it (deployment) :/. It is probably something very simple, so simple that it is not in git/travis/appveoyr documentation but I can't figure it out.

to266:
As you can see [here](https://travis-ci.org/sem-geologist/hyperspy/jobs/125126360), the authentification is the problem. You can read more about it [here](https://docs.travis-ci.com/user/deployment/releases/). Hint: don't put your token there, use the recommended `travis setup releases` method. I have not tested it nor I know exactly how it works though.

For appveyor the same problem is solved differently - they have [this hashing tool](https://ci.appveyor.com/tools/encrypt), where you put the actual github oauth token (I believe, not sure now..). The encrypted thing can then be used directly in the `appveyor.yml`.

Keep in mind that since you don't yet have write access to the main hyperspy repo, you may have to replace it with someone elses once it's tested that it works.

sem-geologist:
@to266 thank you! Actually it is possible to put encrypted token on travis the similar way as in appveyor, and I just did that, and it works. look: (https://github.com/sem-geologist/hyperspy/releases). (travis have hashing tool too, just on command line).
If somebody could test the result with mac if that works (my inspection of whl file as archive looks ok). However with appveoyr things does not looks so bright, as winpython got deprecated. For now to test if it nicely adds the windows version of wheels to the release, I commented whole "before_deploy:" part, and am running like that temporary now.

@francisco-dlp , probably winpython issue is already out of scope of this patch. It grew already big. I will uncomment the commented code in the end. However I see the approach quite messy: every time you will need to do release you need to check out if winpython changed. another thing I see it uses sourceforge link, and if I am right there are some issues with sourceforge (at least everybody in internet saying so).

Second problem is the release version. I tagged it  for testing 0.9.6-test, but it adds 0.8.4 to file names. why? I inspected the files, they contains my changes so it is proper version, just archives are incorrectly named.

In the end I have some problems with appveyor (even with winpython commented out) it shows me something like this:
```
2198  Deploying using GitHub provider
2199  Creating "v0.9.6-test" release for repository "sem-geologist/hyperspy" tag "v0.9.6-test" commit "1cca1bd40d44248799b6020d1a2d6b6e62cce7ab"...Skipped - release with tag "v0.9.6-test" already exists
2200  Uploading "hyperspy-0.8.4-cp35-cp35m-win32.whl" to release assets...Error uploading release asset: 404 - Not Found
```

sem-geologist:
@to266  , @vidartf and @francisco-dlp , I can't find what I am doing wrong. I found this nice example:
(https://github.com/forexample/github-binary-release) showing how to use travis and appveoyr together. My .travis.yml and appveoyr.yml deploy parts hardly differs from that example, and still appveoyr gives me error when trying to deploy (as in prev post). It is of course possible to overcome the issue by copying files manually from appveyor artifacts, (as they gets successfully built, except winpython I mentioned before) to github release (that would take much less time in near future for every release, than me trying to trouble shoot the problem). Maybe I missed some step with appveyor-github configuration (I have it setup, and I have the token generated and encrypted with public rights)... I tried to do fresh release (A.B.C) -- travis sucessfully deployed, appveoyr NOT.

sem-geologist:
I disabled travis ci  and checked how appveyor would do:
so it fails, but error is a bit different from previous:
```
2196  Packaging artifacts...OK
2197  Uploading artifact dist\hyperspy-0.8.4-cp35-cp35m-win32.whl (4.3 MB)...OK
2198  Deploying using GitHub provider
2199  Creating "v1.2.3-test" release for repository "sem-geologist/hyperspy" tag "v1.2.3-test" commit "5887321f51c92127aee9db8650eba44a6b3a96e8"...Error creating GitHub release: Not Found
```

sem-geologist:
@vidart, maybe I am doing something wrong. My workflow is such:

1. on my computer on personal working git I am at patch-2 branch of hyperspy, I do ```git tag 'v0.0.5``` or any other dummy tag.
2. then I do ```git push --tags```

it makes release on github. Also it starts Travis CI and appveyor CI. Travis nicely builds (or if not builds due to random fancy indexing error, I can restart building... anyway at the end it deploys artifacts to github release without any issues). Appvoyer sucessfuly builds artifacts, but fails to push them to github. The exact excerpts from log are present in prev posts. basically it says ```Error uploading release asset: 404 - Not Found``` or when I temporally disabled travis ci, then it changed into ```Error creating GitHub release: Not Found```.

I decided trying to make release straight in github (In the end appveyor is windows environment, and windows is GUI oriented). So Travis nicely built and pushed results to github, but appveyor this time finished with this:
```
Creating "v0.0.0-test" release for repository "sem-geologist/hyperspy" tag "v0.0.0-test" commit "a97109032d541999423bdff5482643cae3c19f28"...Error creating GitHub release: Error reading repository 'sem-geologist/hyperspy' releases: 401 - Unauthorized
```
the link to the log: https://ci.appveyor.com/project/sem-geologist/hyperspy-1lym1/build/job/96qg9sg566cfsp2x


Do I need to configure the appveoyr in some very specific way to work?

sem-geologist:
@vidartf , where you get this from:
```(Get-FileHash '../nsis_patch.zip').Hash -ne '3BA22DDC0F14DBD75A9487EB6C9BD85F535E7038927D251103B97E0AAD94EEAD')```

I mean this does not look as the sha1 sum or md5 sum which is on sourceforge for the file. Mine build as I mentioned previously on appveoyr fails on this.

**Comment by vidartf** [comment]
Get-FileHash uses SHA256 by default, if I remember correctly. SHA1/MD5 is mostly useful for checking if a file has changed, but is defeatable. If you wonder what the checksum is, you can either use a web service, or simply have Appveyor print it.

**Comment by vidartf** [comment]
Regarding Appveyor publishing, I would check that you are using the right kind of access token from github (there are several types). I think AppVeyor needs one kind to give CI status, and one kind to publish releases, but TBH I don't remember anymore.

**Comment by sem-geologist** [comment]
@vidartf, do you mean user token?

**Comment by sem-geologist** [comment]
@francisco-dlp , I saw 0.8.5 were released with windows binaries. it is fresh, so could I ask you to describe how to configure aapveyror to scessflly push the artifacts to git release? (this is stoper for this commit)

**Comment by francisco-dlp** [comment]
@petras, those Windows binaries were actually created manually. Nevertheless, appveyor did try to create the bundle but failed to upload it to the release because I've manually uploaded the files before Appveyor's attemp.

Returning to the issue that matters to you, I am not that familiar with Appveyor myself so, Appveyor experts reading this, if I write something wrong please correct me. From a conversation on the subject with @to266 I understood that you have to set up Appveyor to enable deployment when pushing tags to your repository. This involves adding your token to the appveyor.yml file in your branch. Also, if I got this right, it seems that the deployment is triggered when pushing manually a tag to your github repository (i.e. using git from your local computer). Creating tags from github's interface seems not to trigger a deployment.

I hope that this helps and that we'll be able to merge this PR soon.

**Comment by francisco-dlp** [comment]
Any update on this?

**Comment by sem-geologist** [comment]
I really did something so wrong with my git branches, I think I will reopen the issue with fresh fork?

**Comment by sem-geologist** [comment]
there are some conflict with upstream I can't find, what should I do?

**Comment by vidartf** [comment]
There are quite a few alternatives, but hard to explain well in detail (so I will only say category, and use google for details). In order of severity, I would try:
 - `git merge hyperspy/master`, and then resolve merge conflicts
 - a rebase
 - a (soft?) reset followed by new commits
Either way, ensure you keep your github branch as a backup (possibly also copy the original branch to a backup branch), as you might end up wanting to revert (see also `git reflog` for undoing things).

**Comment by sem-geologist** [comment]
updated to upstream. coveralls probably complain due to setup.py growth... is it possible to blacklist files for coveralls?

**Comment by francisco-dlp** [comment]
Yes, it is see [.coveragerc](https://github.com/hyperspy/hyperspy/blob/master/.coveragerc), however `setup.py` is already omitted. It's not clear to me why coverage complains in this case, but I wouldn't worry too much.

Is this ready for review?

**Comment by sem-geologist** [comment]
Most of code is ready for review. appveyor partly is working, and if it would change it will be just that file without ifluence to anything else. The rest is working and is ready for review.

**Comment by sem-geologist** [comment]
I am just wondering, if `bdist_wheels` for osx is going to work on every mac.... i guess rather not, as there are different kind of processors used... it probably wouldnt work on power pc or dfferent kind of osx versions... so `sdist` is more universal for distribution.

File: .travis.yml [hunk 12]
```diff
@@ -1,36 +1,55 @@
 language: python
 
-sudo: false
```
sem-geologist: sudo false had to go:
1. travis by default uses false
2. osx have no sudo, and it would fail on osx

.travis.yml
```diff
@@ -1,36 +1,55 @@
...
 before_install:
-  # Does not matter which miniconda we get, we still create a virtualenv
-  - wget https://repo.continuum.io/miniconda/Miniconda3-latest-Linux-x86_64.sh -O miniconda.sh
-  - chmod +x miniconda.sh
-  - bash miniconda.sh -b -p $HOME/miniconda
-  - export PATH="$HOME/miniconda/bin:$PATH"
-  - hash -r
-  # miniconda is not always up-to-date with conda.
-  - conda update --yes conda
+- if [[ $TRAVIS_OS_NAME == 'osx' ]] ; then curl "https://repo.continuum.io/miniconda/Miniconda3-latest-MacOSX-x86_64.sh" -o miniconda.sh; else wget https://repo.continuum.io/miniconda/Miniconda3-latest-Linux-x86_64.sh -O miniconda.sh; fi
```
sem-geologist: osx have no wget, so it have to use `curl`

.travis.yml
```diff
@@ -1,36 +1,55 @@
...
+- chmod +x miniconda.sh
+- "./miniconda.sh -b -p $HOME/miniconda"
+- export PATH="$HOME/miniconda/bin:$PATH"
+- hash -r
+- conda update --yes conda
 
 install:
-  - DEPS="nose numpy scipy matplotlib ipython h5py sympy scikit-learn dill natsort setuptools scikit-image"
-  - conda create -n testenv --yes python=$TRAVIS_PYTHON_VERSION $DEPS
```
sem-geologist: `python=$TRAVIS_PYTHON_VERSION` does not add anything? and is not understood at osx. So it had to go

.travis.yml
```diff
@@ -1,36 +1,55 @@
...
+  - DEPS="nose numpy scipy matplotlib ipython h5py sympy scikit-learn dill natsort setuptools scikit-image cython"
+  - conda create -n testenv --yes $DEPS
   - source activate testenv
   - conda install pip
   - pip install coverage coveralls traits traitsui request tqdm
   - python setup.py install
 
-script: python continuous_integration/nosetest.py --with-coverage hyperspy
+script:
+  - python setup.py build_ext --inplace
```
sem-geologist: build extensions (cython and c or c++)

setup.py
```diff
@@ -28,17 +28,19 @@
...
 import os
 import subprocess
 import fileinput
 
+import re
+
+setup_path = os.path.dirname(__file__)
+
 import hyperspy.Release as Release
-# clean the build directory so we aren't mixing Windows and Linux
```
sem-geologist: how that is even possible? There is `setup.py clean` command to do cleaning, and clean up is also now implemented with git post-checkout hooks

sem-geologist:
I added some github comments to explain some things, especially at places where some lines got removed/ replaced, and reasons behind it.

.travis.yml
```diff
@@ -1,36 +1,55 @@
...
+  - python: 3.5
+    env: export FAIL_ON_EXTERNAL_DEPRECATION='True'
+  - os: osx
+    language: generic
+    python: 3.5
+    env: export FAIL_ON_EXTERNAL_DEPRECATION='False'
+  - os: osx
+    language: generic
+    python: 3.5
+    env: export FAIL_ON_EXTERNAL_DEPRECATION='True'
```
vidartf: This entry is not needed, as there is no OS dependence here (unless OSX has separate dependencies?).

sem-geologist: @vidartf, do you mean to remove osx with failure_by_deprecation=true, and leave only 3 instances in the matrix? so that deprecations would be checked only on linux?

vidartf: @sem-geologist Yep :)

to266:
@sem-geologist I think only the issue with `.git` paths is left to fix.

sem-geologist:
@to266, what do you mean by ".git paths"? Could you explain more broadly what needs to be done? However my main concenr with this is that I tried something new with appveyor deploying to github and it fails again, a bit differently, but fails :/ ...
I read that `force` flag should be set in appveyor if release already exists. appveyor kicks with delay , so it is logical it should be set. So instead of authorisation as previously, now I get this "The stream does not support concurrent IO read or write operations":
```
...
2233 Deploying using GitHub provider
2234 Creating "v0.0.3" release for repository "sem-geologist/hyperspy" tag "v0.0.3" commit "830ff4cd5c8f50640fdaeba1244e4aa821ce2d61"...Skipped - release with tag "v0.0.3" already exists
2235 Uploading "hyperspy-1.0+dev-cp35-cp35m-win32.whl" to release assets...The stream does not support concurrent IO read or write operations.
```
full log at:
https://ci.appveyor.com/project/sem-geologist/hyperspy-1lym1/build/0.0.43/job/09xc0tqirl8am5pe

That is not hte only problem. This far I get with nsis build disabled. With nisis build enabled, I get into some other weird error:
```
...
running install
running bdist_egg
running egg_info
writing dependency_links to hyperspy.egg-info\dependency_links.txt
writing hyperspy.egg-info\PKG-INFO
writing requirements to hyperspy.egg-info\requires.txt
writing top-level names to hyperspy.egg-info\top_level.txt
reading manifest file 'hyperspy.egg-info\SOURCES.txt'
reading manifest template 'MANIFEST.in'
warning: no files found matching '*.py' under directory 'bin'
warning: no files found matching '*.rst' under directory 'examples'
warning: no files found matching '*.txt' under directory 'examples'
writing manifest file 'hyperspy.egg-info\SOURCES.txt'
installing library code to build\bdist.win32\egg
running install_lib
running build_py
running build_ext
  File "setup.py", line 327, in <module>
    "Topic :: Scientific/Engineering :: Physics"
  File "C:\Users\appveyor\AppData\Roaming\wpdir\WinPython-32bit\python-3.5.1\lib\distutils\core.py", line 148, in setup
    dist.run_commands()
  File "C:\Users\appveyor\AppData\Roaming\wpdir\WinPython-32bit\python-3.5.1\lib\distutils\dist.py", line 955, in run_commands
    self.run_command(cmd)
  File "C:\Users\appveyor\AppData\Roaming\wpdir\WinPython-32bit\python-3.5.1\lib\distutils\dist.py", line 974, in run_command
    cmd_obj.run()
  File "C:\Users\appveyor\AppData\Roaming\wpdir\WinPython-32bit\python-3.5.1\lib\site-packages\setuptools\command\install.py", line 67, in run
    self.do_egg_install()
  File "C:\Users\appveyor\AppData\Roaming\wpdir\WinPython-32bit\python-3.5.1\lib\site-packages\setuptools\command\install.py", line 109, in do_egg_install
    self.run_command('bdist_egg')
  File "C:\Users\appveyor\AppData\Roaming\wpdir\WinPython-32bit\python-3.5.1\lib\distutils\cmd.py", line 313, in run_command
    self.distribution.run_command(command)
  File "C:\Users\appveyor\AppData\Roaming\wpdir\WinPython-32bit\python-3.5.1\lib\distutils\dist.py", line 974, in run_command
    cmd_obj.run()
  File "C:\Users\appveyor\AppData\Roaming\wpdir\WinPython-32bit\python-3.5.1\lib\site-packages\setuptools\command\bdist_egg.py", line 161, in run
    cmd = self.call_command('install_lib', warn_dir=0)
  File "C:\Users\appveyor\AppData\Roaming\wpdir\WinPython-32bit\python-3.5.1\lib\site-packages\setuptools\command\bdist_egg.py", line 147, in call_command
    self.run_command(cmdname)
  File "C:\Users\appveyor\AppData\Roaming\wpdir\WinPython-32bit\python-3.5.1\lib\distutils\cmd.py", line 313, in run_command
    self.distribution.run_command(command)
  File "C:\Users\appveyor\AppData\Roaming\wpdir\WinPython-32bit\python-3.5.1\lib\distutils\dist.py", line 974, in run_command
    cmd_obj.run()
  File "C:\Users\appveyor\AppData\Roaming\wpdir\WinPython-32bit\python-3.5.1\lib\site-packages\setuptools\command\install_lib.py", line 10, in run
    self.build()
  File "C:\Users\appveyor\AppData\Roaming\wpdir\WinPython-32bit\python-3.5.1\lib\distutils\command\install_lib.py", line 107, in build
    self.run_command('build_ext')
  File "C:\Users\appveyor\AppData\Roaming\wpdir\WinPython-32bit\python-3.5.1\lib\distutils\cmd.py", line 313, in run_command
    self.distribution.run_command(command)
  File "C:\Users\appveyor\AppData\Roaming\wpdir\WinPython-32bit\python-3.5.1\lib\distutils\dist.py", line 974, in run_command
    cmd_obj.run()
  File "C:\Users\appveyor\AppData\Roaming\wpdir\WinPython-32bit\python-3.5.1\lib\site-packages\setuptools\command\build_ext.py", line 49, in run
    _build_ext.run(self)
  File "C:\Users\appveyor\AppData\Roaming\wpdir\WinPython-32bit\python-3.5.1\lib\site-packages\Cython\Distutils\build_ext.py", line 164, in run
    _build_ext.build_ext.run(self)
  File "C:\Users\appveyor\AppData\Roaming\wpdir\WinPython-32bit\python-3.5.1\lib\distutils\command\build_ext.py", line 307, in run
    force=self.force)
  File "C:\Users\appveyor\AppData\Roaming\wpdir\WinPython-32bit\python-3.5.1\lib\distutils\ccompiler.py", line 1031, in new_compiler
    return klass(None, dry_run, force)
  File "C:\Users\appveyor\AppData\Roaming\wpdir\WinPython-32bit\python-3.5.1\lib\distutils\cygwinccompiler.py", line 282, in __init__
    CygwinCCompiler.__init__ (self, verbose, dry_run, force)
  File "C:\Users\appveyor\AppData\Roaming\wpdir\WinPython-32bit\python-3.5.1\lib\distutils\cygwinccompiler.py", line 126, in __init__
    if self.ld_version >= "2.10.90":
TypeError: unorderable types: NoneType() >= str()
Command exited with code 1
```

full log at:
https://ci.appveyor.com/project/sem-geologist/hyperspy-1lym1/build/0.0.42/job/8o45c3nav5k9u70d#L2319


So the state of this thing is such that appveyor builds wheels, allways!!! however it does not deploy to github release. Travis do that for osx.

Additionaly for OSX I was thinking if not to explore travis capability to build the wheels for different versions of osx? (Pardon I am not sure how versioning works on osx and if there are any binary difference between anaconda on one or other version of osx...)

setup.py
to266: @sem-geologist This line above is what I meant - you should adopt the same 
`os.path.join(setup_path, '.git')` instead of just `'.git'` when checking if it exists.

sem-geologist:
@to266 , the `.git` fix is done

to266:
I'm not sure if the appveyor bugs are very relevant, given that a number of other PRs wait for this one.

From the testing side of things and the code making sense, I think it's ready.

@francisco-dlp do you think the appveyor not properly deploying is worth delaying the PR? I think it's better fixed later in a different PR

sem-geologist:
@to266 , @francisco-dlp , actually I am not sure if this is ready. deploying is not a big problem, and as work around you can always use manual dropping files to the github release. However I think nisis/winpython failure to build is more severe. I searched and found something similar:
http://stackoverflow.com/questions/24683305/python-cant-install-packages-typeerror-unorderable-types-nonetype-str
I guess due to using cython there is need of compiler which is missing in winpython distribution, and thus there the problem appeared. If to merge this right now that will make hyperspy more bothersome to install. And there I see the problem: wheels are anaconda distribution compatible (it is MSVC, if I good remeber, and winpython?). At least I see what is need to be fixed. About appveyor deployment: I got one more idea to try, that is my last hope, but will do that after I will fix stuff with winpython.

to266:
@sem-geologist Have a look at `conda-forge/staged-recipes` , there may be useful ideas there, since they use the CI for deployment as well.

If not for your comment I would have merged this right now, so just let me know when you think it's ready. If nothing else, someone can fix the appveyor problems in other PRs, as I think moving to setuptools is very important on its own and worth merging just for that.

sem-geologist:
@to266 , @francisco-dlp  I sorted out winpython, the problem is in winpython defaults. Winpython is compiled with msvc, but by default pydistutils.cfg is set to search and use mingw.  Mixing compilers is very bad and can result in weird results. By deleting that file it is posible to force the distutil search and use the same compiler as python version were compiled. So the fix was to add to appveyor vs2015 worker, and add the line which removes the pydistutils.cfg file from winpython installation.
I read in another places that problem with deployment can be caused by trying to release from branch which is waiting to be merged. I cloned the branch to the separate, tried to release, but it is the same old behaviour: no deployment. However both wheels and nsis installers are built and are in artifacts, so on release to copy 4 files manualy should not be a huge problem.

francisco-dlp:
The reason why WinPython uses mingw by default is because it ships with it as mingw, unlike msvc, is open source. What about renaming pydistutils.cfg instead of deleting it and restoring it after after hyperpy is compiled? An alternative would be to create wheels using anaconda and install these wheels in winpython.


Once this is addressed, I agree with @to266, despite the appveyor remaining issue, we should merge this.

sem-geologist:
By obstacles with installing and solving stuff with winpython, I got to understand that if there is no compiler then hyperspy will not install with this new cythonization stuff... I understand that it is wishful that it would be possible to install hyperspy without extensions requiring compiler. I searched for examples, and I think I could use this one http://stackoverflow.com/questions/28843765/setup-py-check-if-non-python-library-dependency-exists as the concept to find out if compiler is working. If it would fail, then extensions would be set to empty list, so that build_ext would not require any compiler, and it could install. this should not take me long to do.
I have working example, and I tested it in on live ubuntu usb, evading installing the gcc :).
So if there is compiler: `python setup.py install` will install everything, and if there will be no compiler present then it will get installed everything except c/cython extensions. Is that ok?

francisco-dlp:
It's great that not only you've spotted the issue before we merge this but also you seem to already have a solution. Let's us know when this is again ready for review.

sem-geologist:
Coming back to mingw issue...: I also would like to see used mingw instead of mscv, however at this moment there is very big problem: winpython is compiled with msvc! and the binaries of compiled libraries and python should not be mixed, else it crashes or in worst case it do not, but fails most of assertations! If python is compiled with clang, then custom compiled library should be compiled with clang, if it was gcc, then gcc, if it was msvc then msvc. You should not mix that. So the winpythons problem is in that it by default is set to mingw while itself is compiled in msvc. BTW anaconda on windows uses also msvc by default! If we would want to use mingw, then we should compile winpython ourselfs. 

BTW Are you aware that with python 3.5 win xp support is dropped?

francisco-dlp:
I fully see your point. I am not suggesting to compile with mingw, I am just suggesting to compile with mscv as you're currently doing, but then restore the `pydistutils.cfg` file to leave winpython untouched. Even if compiling with a different compiler can lead to all sort of issues, in the few cases where it'll work, the users who don't have mscv installed will be happier than if they can't compile at all. Notice that winpython include mingw.

I wasn't aware of the XP issue. I think that Python 3.5's @ operator is a big gain for scientific computing. No support for XP is a small price to pay for this ;)

sem-geologist:
the problem is that winpython 3.5 does not include mingw: 
http://winpython.github.io/ (go and read last line, also I tried it on wine, it extracted no mingw...)

francisco-dlp:
Are you sure of that? The latest hspy bundle used wp 3.5 and gcc seems to be there.

sem-geologist:
Yes I am 100% sure. I also just looked into they python configs, there is some comments that they claim that winpython compiled against mscv is  modified to be compatible with mingw binaries (which cant be, else stack overflow would not be so full of describing assert bugs). I understand they possition: the problem is there is many packages for windows which are compiled only with msvc. If they would distribute pure mingw python then lots of packages would be not available. thus they do hybrid thing, so it could use (even being broken) two worlds msvc packages + mingw build custom stuff. Anaconda goes other way, they do everything msvc.

You had no problem with h5py because in winpython you get binary library.
You can run into problem if you do test on Ubuntu or other linux without gcc present.
If there is no python-h5py package then python during hyperspy installation tries to download and compile it... 
My first plan was to put the pydistutil.cfg to smoke, then it unblocks native compiler. If you wish it stay there, I will rename it temporary, compile stuff, then rename back to original.
However this will not leave users from trouble if they will want to compile something custom.
It took me hours to understand where the problem lies, as there is 0 instructions or notice in winpythons site, and if you try to compile you get very cryptic long error traceback which have no mentioning about mingw missing whatsoever. The winpython site is also misleading as they would include mingw (they cant as they probably having MIT license). So without that file user would be noticed that they have msvc (some bat file) missing, and internet is full of solutions. With that file put back, I will feel quite bad about potential users...

francisco-dlp:
I see. My take on this is that we're just repackaging winpython, so we should live with its strengths and weaknesses. Therefore, I think that the right thing to do is to leave winpython as it is, and if somebody thinks that there is something broken in WP and wants to fix it, they can contribute to the WinPython project.

sem-geologist:
ok, I am going to modify, the appveyor: instead of smoking up that file, I will rename it, and after rename it back (hopefully). I am testing stuff now on separate branch (as I am not good in powershell - lots of trial and error going there :) ). I will merge it back to this branch, so you could start reviewing changes and test it out how it works when compiler is not present (setup.py modified, and new minimal dummy c file created in /hyperspy/tests/misc/test_compilers.c), and in mean time I will try to do renaming of that file forward and back tricks (The winpython/ in appveyor stuff).

francisco-dlp:
Sound good, just let me know when this is ready for review.

sem-geologist:
I merged the changes to setup.py and minimal c file, however appveyor winpython have problems if I leave the file. It behaves differently than linux wihout gcc, where on linux if extensions is empty list it will not try to use compiler, while on winpython it tries to use compiler, even if extensions is set to just empty list. Maybe it is winpythons modifications. I will try to change empty list to None, will see where it will lead (tried on linux it works same as with empty list, maybe winpython will take it)

setup.py
```diff
@@ -132,6 +137,26 @@ def no_cythonize(extensions):
...
+    warnings.warn("""WARNING: C compiler can't be found.
+Only slow pure python alternative functions will be available.
+To use fast implementation of some functions writen in cython/c either:
+a) check that you have compiler (EXACTLY SAME as your python
+distribution was compiled with) installed,
+b) use binary distribution of hyperspy (i.e. wheels, egg, (only osx and win)).
+Installation will continue in 10 sec...""")
+    extensions = []
+    from time import sleep
+    sleep(10) #wait 10 secs for user to notice the message
```
francisco-dlp: Why waiting 10 s? As the user has to take no action I would simply print the warning and continue...

sem-geologist: if terminal used have some limit (often 1000 lines) it "flies" away. this will wait for 10 sec only if compiler is missing. if it is here, no waiting.

hyperspy/tests/misc/test_compilers.c
```diff
@@ -0,0 +1,4 @@
+int main(void)
```
francisco-dlp: I am not sure if `hyperspy/tests/misc/` is the best place to put this file as this is not part of an unittest. Beside, I think that c files are currently ignored when packaging. Should this individual file be declared in MANIFEST?

setup.py
sem-geologist: And I have results from winpython... it is bad, no matter to what extensions are set (empty list or None) if keyword `ext_modules` is provided to `setup()` then winpython anyway tries to fire up compiler :/. So I don't know what to do now as it is "impossible" to trick winpython from touching compiler.

sem-geologist:
And I have results from winpython... it is bad, no matter to what extensions are set (empty list or None) if keyword `ext_modules` is provided to `setup()` then winpython anyway tries to fire up compiler :/. So I don't know what to do now as it is "impossible" to trick winpython from touching compiler.

francisco-dlp:
I think that the easiest solution is to put all the keyword arguments in a dictionary and simply don't add the `ext_modules` key if there is not compiler e.g.

```python

keyword_args = {}
setup(**keyword_args)

```

May that work?

sem-geologist:
:) exactly what I am trying to don now :D

sem-geologist:
the revolution in setup.py didn't helped... I think problem is lying somethere else. I made abranch from master, adjusted the appveyor to fit for faster testing of winpython. master have no problems to get installed on winpython. I will add stuff step by step to find out where the problem is introduced. i.e. I skiped all conda stuff and tried to go straight to winpython steps, and it fails (many times trying to use compiler...), however if conda stuff is not comented it works (that why I suspect compiled stuff from conda is coppied to winpython? so basically, if somebody will want to install hyperspy on bare minimum winpython, it will fail with dependencies :/

hyperspy/tests/misc/test_compilers.c
sem-geologist: yes it shoul be put to manifest, I noticed that, so what other place would be apropriate?

francisco-dlp: Maybe `hyperspy/misc`?

sem-geologist:
I didnt moved to keywords as it does not fix issue, however I think I fail to simulate compiler'less environment on appveyor winpython, due to it reusing condas stuff... unfortunately I have no winpython environment (windows) to test that now. If I comment conda mostly, then winpython resembles same as python in ubuntu: it tries to compile traits, and fails. I guess with conda ennabled winpython reuse condas distutils which internally have msvc compiler hanged, and so my test thinks compiler is present (I guess, it is hard to tell what is really going on) but then winpython forces to use mingw and so it gets messed. That ofcourse do not influence building the nsis installer as I made the renaming of config before winpython installs hyperspy, and rename back just before nsis start building.

So basically I think there is little I can do to this any more... I think it is time to merge

sem-geologist:
@francisco-dlp , I just tested installation of master on fresh windows system (7) with freshly installed winpython. It fails to install because of traits needs to be compiled. So ubuntu for now is the only system where limited hyperspy version is actual (as there is python-traits binary package there available). Unfortunately installation of hyperspy as library on winpython requires compiler, and I just want to make 200% sure that winpython does not include any compiler! Also I think the points I stated yesterday is valid, and this can be merged.

sem-geologist:
looks that installing from wheel have all dependencies needed. only traits have to be installed from gohlke's site. I think I will open new issue/pull request to improve instructions from this point

sem-geologist:
@francisco-dlp , @to266 , finally I tested installing this branch in windows environment without any compiler present on winpython and it was a success (after some minor modifications which I just pushed, catching msvc missing). the weird thing is that on my machine ,differently than on appveoyr, distutil complained about msvc missing and not the mingw. Anyway this is now completely tested and safe to use.

francisco-dlp:
Thanks for all the effort and attention to detail that you've put into this!

Issue #789, opened 2015-12-10
Plotting for Model2D


dnjohnstone:
As discussed in #536 plotting of 2-dimensional models is not yet supported, there is also a question regarding how best to actually do this from an interpretability perspective...

At present the general opinion was in favour of plotting the model and the data side-by-side (I was also thinking that automatically plotting the residual, may also be handy as people should be looking at that). Perhaps this should also come after the changes being discussed/made in #423 and associated.

dnjohnstone:
This has become a little more pressing than I imagined because the absence of all functions related to plotting currently causes the multifit() method to fail for Model2D -- now that events and widgets are implemented though we should be in a good place to extend things like interactive range selection (which were the things putting me off doing this initially) to Model2D.

I will discuss this in person with @to266 in the next day or two but if anyone else is interested in this functionality and feels like adding their weight to this it would be good to know.

dnjohnstone:
Further to that, looking at the code, my plan is roughly as follows:

1. Tidy up Model1D -- first by removing EELS specific stuff which really shouldn't be there and then by moving parts of the plotting code that are general to both 1D and 2D to BaseModel - looking at it I think quite a lot of the code that is now implemented with widgets and events will move to BaseModel.

2. Come up with implementations for the truly 2D plotting bits.

francisco-dlp:
Can you briefly mention what did you have in mind when you wrote: "first by removing EELS specific stuff which really shouldn't be there"?

dnjohnstone:
I thought we'd discussed previously that the low_loss and set_convolution_axis methods shouldn't be there? Or are they used for EDS model too?

francisco-dlp:
You are right, that was written for EELS but I think that convolution can be useful for other 1D signals as well. We should probably rename the `low_loss` attribute though. That said, this feature needs to be refurbished and possibly implemented in `Model2D` too.

vidartf:
Yeah, e.g. convolving with a point-spread function for 2D images could be useful.

magnunor:
2D-convolution with a "probe" function + https://github.com/hyperspy/hyperspy/issues/707 and we're close to doing STEM-simulations :)

dnjohnstone:
Ok - so conclusion is...

low-loss attribute --> function to convolve

[...]

PR #951, opened 2016-04-11
NEW: Plotting for Model2D

dnjohnstone:
To address #789


dnjohnstone:
This is a work in progress - it's enough to fix multifit for Model2D and to raise a plot but the plot is not connected properly. Made this PR now because I'm having some issues with events i.e. how to connect things properly and I don't know how I should be approaching them - @vidartf any help would be great!

francisco-dlp:

PR #953, opened 2016-04-13
IO-Plugins for Berkley EMD format (Python 3 compatible)!

CodeMonkeyJan:
Added an io_plugins for the EMD format, a hdf5 standard proposed by Colin Ophus from Lawrence Berkeley National Lab, mentioned in #699 (see http://emdatasets.lbl.gov/ not to be confused with the FEI EMD format which was developed later, is mentioned in #698 with basic support implemented in #710). As mentioned in my previous attempt #813, I had to recreate the pull request. This time, everything should be Python 3 compatible!


CodeMonkeyJan:
Apparently the Travis build failed:
It seems if the problem does not lie in one of my modules, anything I can do?

vidartf:
Same as issue #798, but nobody seems to have any input on it.

vidartf:
Sorry, meant issue #914.

CodeMonkeyJan:
Hm, strange issue, hopefully it is sorted out soon. If it randomly fails, is there a way to rerun the travis test? Or is it even recquired for the merging to be allowed?

francisco-dlp:
> is there a way to rerun the travis test?

You need to be in the dev team for that. I've just triggered a rebuild for your PR.

CodeMonkeyJan:
Thanks alot :-)!

hyperspy/io_plugins/emd.py
```diff
@@ -0,0 +1,393 @@
...
+        # Iterate over all dimensions:
+        for i in range(len(signal.data.shape)):
+            key = 'dim{}'.format(i+1)
+            axis = signal.axes_manager._axes[i]
+            offset = axis.offset
+            scale = axis.scale
+            dim = dataset.create_dataset(key, data=[offset, offset+scale])
+            name = axis.name
+            from traits.trait_base import _Undefined
+            if type(name) is _Undefined:
```
to266: the check should be `isinstance(name, _Undefined)`

hyperspy/io_plugins/emd.py
```diff
@@ -0,0 +1,393 @@
...
+                name = ''
+            dim.attrs['name'] = name
+            units = axis.units
+            if type(units) is _Undefined:
```
to266: here as well

hyperspy/io_plugins/emd.py
```diff
@@ -0,0 +1,393 @@
...
+            else:
+                units = '[{}]'.format('_'.join(list(units)))
+            dim.attrs['units'] = units
+        # Write metadata:
+        dataset.attrs['emd_group_type'] = 1
+        for key, value in signal.metadata.Signal:
+            try:  # If something h5py can't handle is saved in the metadata:
+                dataset.attrs[key] = value
+            except Exception:
+                pass  # just don't add what can't be added!
```
to266: We can do better than just passing silently - actually letting the user know what wasn't added, just like [here](https://github.com/hyperspy/hyperspy/blob/master/hyperspy/io_plugins/hdf5.py#L430)

to266: Actually it's somewhat likely that the metadata will have other signals (that are not write-able that easily), so the above comment is quite important...

hyperspy/io_plugins/emd.py
```diff
@@ -0,0 +1,393 @@
...
+        for key, value in self.comments.items():
+            comments_group.attrs[key] = value
+        # Write signals:
+        signal_group = emd_file.require_group('signals')
+        for signal in self.signals.values():
+            self._write_signal_to_group(signal_group, signal)
+        # Close file and return EMD object:
+        emd_file.close()
+
+    def print_info(self):
```
to266: We've changed the printing to just logging (to leave as few as possible printing statements in general). Could you adapt this function as well? [here](https://github.com/hyperspy/hyperspy/blob/ea59b223c5e1301c097d0b24d126e93c003d4ae0/hyperspy/io_plugins/semper_unf.py#L631) is an example

CodeMonkeyJan:
The proposed changes are implemented now. Sadly the Travis build failed again.

hyperspy/io_plugins/emd.py
to266: Just one last thing - this function should probably be renamed to `log_info`, or something along these lines. Not actually printing anything when run might be confusing for users. I think that's the only thing left before merging!

to266:
Also rebuilt travis.

CodeMonkeyJan:
Renaming the function to log_info makes perfect sense. The keyword argument for the load-function is renamed appropriately. This time the AppVeyor build failed, the error message is the same.

PR #955, opened 2016-04-15
DOC: add cython(ization) doc to development and user guide

sem-geologist:
this is connected with #949.

### DIFF
--- a/doc/dev_guide.rst
+++ b/doc/dev_guide.rst
@@ -1,7 +1,7 @@
 ﻿Developer Guide
 ===============
 
-This 5-step guide is intended to give people who want to start contributing their 
+This 6-step guide is intended to give people who want to start contributing their 
 own tools to HyperSpy a foothold to kick-start the process. This is also the way
 to start if you ultimately hope to become a member of the developer team.
 
@@ -180,6 +180,8 @@ quite large. Cythonization will take place during Travis CI and Appveyor buildin
 The cythonized code will be generated and included in source or binary distributions
 for end users. To help troubleshoot potential deprecation with future cython releases,
 add a comment with in the header of your .pyx files with the cython version.
+If cython is present in the build environment and any cythonized c/c++ file is missing,
+then setup.py tries to cythonize all extensions automatically.
 
 To make the development easier the new command ``recythonize`` has been added to setup.py.
 It can be used in conjunction with other default commands.
@@ -189,6 +191,6 @@ will recythonize all changed (and described in setup.py!) cython code and compil
 When developing on git branches, the first time you call setup.py in conjunction with
 or without any other command - it will generate a post-checkout hook, which will include
 a potential cythonization and compilation product list (.c/.cpp/.so/.pyd). With your next
-``git checkout`` the hook will remove them and run ``python setup.py build_ext --inplace``
+``git checkout`` the hook will remove them and automatically run ``python setup.py build_ext --inplace``
 to cythonize and compile the code if available. If an older version of HyperSpy (<= 0.8.4.x)
 is checked out this should have no side effects.
--- a/doc/dev_guide.rst
+++ b/doc/dev_guide.rst
@@ -168,28 +168,27 @@ discussed.
 6. Contributing cython code
 ---------------------------
 
-Python is not the fastest language, it is slow in loops. Sometimes using vectorisation
-of calculations with numpy is not possible. It is mostly actual for some proprietary
-binary electron microscopy formats. In such and similar cases where the speed can be
-significantly improved, it is recommended to implement optional cython parts alongside
-pure python versions. While developing cython code keep the official cython
-recommendations http://docs.cython.org/.
-Add your cython extensions to the setup.py, to the existing list of ```raw_extensions```.
+Python is not the fastest language, and can be particularly slow in loops.
+Performance can sometimes be significantly improved by implementing optional cython
+code alongside the pure Python versions. While developing cython code, make use of
+the official cython recommendations (http://docs.cython.org/).
+Add your cython extensions to the setup.py, in the existing list of ``raw_extensions``.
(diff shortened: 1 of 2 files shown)


tjof2:
I think this reads better:

Python is not the fastest language, and can be particularly slow in loops. Performance can sometimes be significantly improved by implementing optional cython code alongside the pure Python versions. When developing cython code, make use of the official cython recommendations (http://docs.cython.org/). Add your cython extensions to setup.py, in the existing list of `raw_extensions`.

Unlike the cython recommendation, the cythonized .c or .cpp files are not welcome in the git source repository (except original c or c++ files), since they are typically quite large. Cythonization will take place during Travis CI and Appveyor building. The cythonized code will be generated and included in source or binary distributions for end users. To help troubleshoot potential deprecation with future cython releases, add a comment with in the header of your .pyx files with the cython version.
To make development easier, the new command recythonize has been added to setup.py. It can be used in conjunction with other default commands. For example, python setup.py cythonize build_ext --inplace will recythonize all changed (and described in setup.py!) cython code and recompile.

When developing on git branches, the first time you call setup.py in conjunction with any other command it will generate a post-checkout hook, which will be include a potential cythonization and compilation product list (.c/.cpp/.so/.pyd). With your next git checkout the hook will remove them, so run python setup.py build_ext --inplace to cythonize and compile the code if available. If an older version of HyperSpy is checked out <= 0.8.4.x this should have no side effects.

doc/dev_guide.rst
```diff
@@ -164,3 +164,32 @@ to use it with examples and links to the relevant code.
 When you've got a branch that's ready to be incorporated in to the main code of
 HyperSpy -- make a pull request on GitHub and wait for it to be reviewed and
 discussed.
+
+6. Contributing cython code
+---------------------------
+
+Python is not the fastest language, it is slow in loops. Sometimes using vectorisation
+of calculations with numpy is not possible. It is mostly actual for some proprietary
+binary electron microscopy formats. In such and similar cases where the speed can be
```
tjof2: "It is mostly actual for some proprietary binary electron microscopy formats" doesn't make sense, and there may be cases in future that have nothing to do with binary file formats where Cython may still be helpful - I suggest taking this out.

doc/dev_guide.rst
```diff
@@ -164,3 +164,32 @@ to use it with examples and links to the relevant code.
...
+Add your cython extensions to the setup.py, to the existing list of ```raw_extensions```.
+
+Differently than cython recommendation, the cythonized huge and obscure *.c or .cpp* files
+are not welcome in git source repository (except original c or c++ files). Cythonization
+will take place during Travis CI and Appveyor building. The c/c++ cythonized c will be
+generated and included in source or binary distributions meant to be for end users.
+To help troubleshoot potential deprecation with future cython releases, add the comment
+with cython version the code is developed in your .pyx files at the top in the header.
+
+To make the development easier the new command ``recythonize`` is coded in setup.py
```
tjof2: Surely giving an example of `recythonize` is better? Your example below just says `cythonize`.

sem-geologist: my mistake...

sem-geologist:
@tjof2 , thanks for the review, I commited the proposed fixes. I think the only last paragraph were misunderstood. it is the hook which runs the ```python setup.py build_ext --inplace```. Maybe I should add "automatic" word somewhere to be more clear?

tjof2:
@sem-geologist maybe just change this sentence: 

"With your next git checkout the hook will remove them, so run python setup.py build_ext --inplace to cythonize and compile the code if available. If an older version of HyperSpy (<= 0.8.4.x) is checked out this should have no side effects."

to

"With your next git checkout the hook will remove them and automatically run `python setup.py build_ext --inplace` to cythonize and compile the code if available. If an older version of HyperSpy (<= 0.8.4.x) is checked out this should have no side effects."

Does that make sense?

Also, you need to update the first line of the file "This 5-step guide is intended to give people" to a 6-step guide. Or should cython be an "optional" extension to the guide - @dnjohnstone?

sem-geologist:
@tjof2 , I forgot to mention important thing, I added in last commit at 183-184 lines.

tjof2:
@sem-geologist Thanks, looks fine.

Don't forget the first line: 

> Also, you need to update the first line of the file "This 5-step guide is intended to give people" to a 6-step guide.

sem-geologist:
@tjof2 , just added some bits to user documentation. I moved it from #949 as I think this pull request is more appropriate for this.

dnjohnstone:
As with #958 whilst I think this is valuable information @francisco-dlp  I think we need to sort out the overall development guide structure first -- probably we need to prioritise this a bit so that contributions like this fit automatically.

tjof2:
@dnjohnstone agreed.

francisco-dlp:
I agree that a more complete dev guide should be a priority. Still, I don't see any problem with merging this and moving it later on once we find a more suitable place for it.

However, let's wait until #949 is merged, as things may change there.

@sem-geologist, (just for information) usually we add the documentation is the same pull request as the code that it is documenting.

sem-geologist:
Oh, I didn't knew about that separation.
the state of #949 is stuck! I am struggling with testing the building and pushing to github releases with appveyor, and I am hopping somebody already experienced with appveyor will help me... travis worked like a charm and builds osx bdist_wheels nicely without any issues.

francisco-dlp:
I think that @vidartf is the most experienced of us in appveyor, @vidartf?

vidartf:
Sorry, I haven't been following the discussion. As far as I understand it, @sem-geologist can't create test deployments? @sem-geologist : If you could link your AppVeyor hyperspy project (your hyperspy project on appveyor.com) I could have a look in the build logs. I wasn't able to find it myself.

PR #958, opened 2016-04-17
DOC: add note about float type comparison to dev_guide

sem-geologist:
there are some testing code which randomly fails,
due to lack of awareness about float type comparison caveats.
To prevent future random testing failures some note about float point issues is added to writing tests section

### DIFF
--- a/doc/dev_guide.rst
+++ b/doc/dev_guide.rst
@@ -147,12 +147,9 @@ Tests -- these are short functions found in hyperspy/tests that call your functi
 under some known conditions and check the outputs against known values. They should
 depend on as few other features as possible so that when they break we know exactly
 what caused it. Writing tests can seem laborious but you'll probaby soon find that
-they're very important as they force you to sanity check all you do. Be cautios when
-evaluating float type numbers. Avoid constructions or assertations with underlying
-construction: ``expected_float == known_float``. Instead use "assert almost equal"
-or anything with underlying construction (pseudo code):
-``(expected_float - known_float) in_between (-precision_delta, +pecision_delta)``.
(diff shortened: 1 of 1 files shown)


dnjohnstone:
I'm not sure if this is the right place for this note -- although I do think it should be in the developer guide somehow. Actually I think it comes back to an earlier question that was discussed with @francisco-dlp regarding the dev-guide structure... As it is the aim is to just give an overview of the key steps for someone with no open-source contributing experience to make it seem less scary - there was supposed to be a plan to have a second more detailed development guide that would focus on the details of the HyperSpy code base. To me this note should go in to that more detailed structure...

On a related note - this needs re-writing in any case because the English is a bit of a mess and it wouldn't really be clear to me what it meant if I didn't already know the answer.

tjof2:
@dnjohnstone The usual approach is to just reference: ["What Every Computer Scientist Should Know About Floating-Point Arithmetic"](http://perso.ens-lyon.fr/jean-michel.muller/goldberg.pdf) :-)

In any case, it could be massively simplified - e.g. this communicates the point in the same way:

> When comparing integers, it's fine to use `==`. When comparing floats, be sure to use `assert_almost_equal()`.

francisco-dlp:
@sem-geologist, would the succinct rephrasing of your text that @tjof2 suggests be enough? If yes, would you like to commit those changes yourself?

sem-geologist:
Ok, done.

francisco-dlp:
Could you put == and assert_almost_equal inside backquotes (`)?

sem-geologist:
added the back quotes.

Issue #957, opened 2016-04-17
BUG-FIX: wrong comparison of float type numbers in tests 


sem-geologist:
There are "random" travis failures which are caused by not right comparison method of  python floats (which are double?). To better understand the issue read http://www.cygnus-software.com/papers/comparingfloats/comparingfloats.htm .

So generally the floating points should not be compared like this:
```0.5 == 0.5``` as depending from compiler, architecture, or even os scheduler, that can return False.
I noticed that most of random failing tests use ```assert_equal``` (which under hood does '==') from ```nose.tools```, it should be changed to ```assert_almost_equal(result, expected_val, places=sane_number_of_digits)```. In some cases where ```assert_equal``` is used on lists or dicts containing floating points, this will get complicated and will need introducing loops to iterate, or the tests should be redesigned. There are some cases in tests where assert_true(float_val) is used, which is also inappropriate.

TODO:
* [x] identify as much as possible dangerous tests
* [x] fix it
* [x] add caveat to dev documentation about float and tests to prevent it in future #958

Bellow I am going to list all affected tests by wrong test methods on float type numbers.

**Comment by sem-geologist** [comment]
The test list:

test (without 'hyperspy.tests.') part | fix | state
-----|-----|-------
signal.test_eds_sem.test_take_off_angle | ```assert_equal``` to ```assert_almost_equal``` | addressed and fixed in #949
model.test_fancy_indexing. TestModelIndexing. test_model_signal_indexer_slice | that is complicated, but the probable cause is the ```assert_true``` on ```float``` number :open_mouth: | addressed at #914. Not fixed.

**Comment by sem-geologist** [comment]
I find out that:
with ```sys.float_info.dig``` it is possible to get python float type precision on the given system.
I.e. on my debian 64bit machine it is 15. probably I should change offending assert_equal parts to:
```python
 assert_almost_equal(measured, expected,
                     places=sys.float_info.dig-(expected.__int__() // 10 + 1 ))
```

by using such approach instead of fixed values it will be more flexible and should stop randomly failing on windows32 builds.

francisco-dlp:
Great bug report!

tjof2:
I've not read `assert_almost_equal` in detail, but I presume it implements the comparison checks appropriately!

dnjohnstone:
what remains to be done for this?

sem-geologist:
the #949, which I am stuck with...

francisco-dlp:
[...]

Issue #956, opened 2016-04-16
FIX: tests preventing successful osx Travis CI


sem-geologist:
as in #949 some test failures were detected only with newly enabled osx in Travis CI.

There are two tests where one hangs the test and time outs in 10 min whole build.

1. tests.io.test_hdf5.TestLoadingOOMReadOnly

Second is effect of too strict comparison of floating numbers, which on different platforms can fail (in our case it was always working on linux, but assertation fails on osx due to the last digit in very long floating point number differs):

```
FAIL: hyperspy.tests.signal.test_eds_sem.Test_metadata.test_take_off_angle
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/Users/travis/miniconda/envs/testenv/lib/python3.5/site-packages/nose/case.py", line 198, in runTest
    self.test(*self.arg)
  File "/Users/travis/build/hyperspy/hyperspy/hyperspy/tests/signal/test_eds_sem.py", line 189, in test_take_off_angle
    nt.assert_equal(s.get_take_off_angle(), 12.886929785732487)
AssertionError: 12.886929785732484 != 12.886929785732487
```

the problem here is that float numbers should be truncated to sane significant numbers and only then compared. By changing ```nt.assert_equal``` into ```nt.assert_almost_equal(s.get_take_off_angle(), 12.886929785732487, places=14)``` it is safe to run on different platforms. On python floating point numbers should be comparable with max 16 significant digits, thus the ```places=14```.

The fix of the second is very small and for now it is dealt in the #949.
However first needs to be investigated.
If wished it can be split to be separate pull request.

Update:
I noticed that there is many ```assert_equal``` used on the floating points in the tests. That causes random failure in Travis on Linux (and osx). I am opening separate issue and pull request for those.

sem-geologist:
I investigated more the problem with osx at tests.io.test_hdf5.TestLoadingOOMReadOnly test

the lines at which travis osx hangs are those:

```python
260    @nt.raises(MemoryError, ValueError)
261    def test_in_memory_loading(self):
262        s = load('tmp.hdf5')
```

I am wondering what is the possibilities... maybe osx does not raise MemoryError but tries to finish the loading using swap (dynamic swap or some other tricks) so it slows down to such extent that test timeouts.
It would be best if somebody would have osx and could replicate the state.

to266:
[...]

PR #959, opened 2016-04-18
BUGFIX: updating tests (PR to master)

to266:
Changes all the `==` comparisons where possible, attempts to use appropriate tests (not just `assert_true`)

addresses parts of https://github.com/hyperspy/hyperspy/issues/957 and https://github.com/hyperspy/hyperspy/issues/956

(There will be a similar PR for the 0.8.x)

### DIFF
--- a/hyperspy/tests/component/test_component_active_array.py
+++ b/hyperspy/tests/component/test_component_active_array.py
@@ -33,10 +33,8 @@ class TestParametersAsSignals:
         g = self.gaussian
         g.active_is_multidimensional = False
         g._create_arrays()
-        nt.assert_true(
-            np.all(
-                g.A.as_signal('values').data == np.zeros(
-                    (3, 3))))
+        np.testing.assert_array_equal(g.A.as_signal('values').data,
+                                      np.zeros((3, 3)))
 
     def test_some_inactive(self):
         g = self.gaussian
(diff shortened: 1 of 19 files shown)


sem-geologist:
@to266, I am not sure changing ```nose.tool.assert_true``` to ```assert_equal``` gives anything there, as the nose.tool internally does just that, I mean does ```a == b ?```. So if floats would fail with ```a==b``` it will fail same with ```assert_equal```. To make it safe you should use ```assert_almost_equal``` on the all floats.

to266:
@sem-geologist While you are correct, the purpose of this PR is to mostly clean up the tests - so that the correct `assert_*` statement is used. 

Also, I believe your concern is only relevant when calculations are involved, just like the first example in the article you link to. In those cases I tried to use numpy or `allclose` or `*_almost_equal` functions, but do let me know if there are any other left that you think need changing.
If the comparison essentially checks if setting a number to `1.3` actually sets it to the `1.3`, I don't believe we have to use the complicated procedures  - the only way `1.3 != 1.3` in python is if something is __really__ wrong in your machine.

sem-geologist:
@to266 , if there is no calculations, and you are 100 percent sure it does not some internal mungling with the number, then ok.

but if even one calculation is used, it can fail: ie:
```
1.3 == 1.1 + 0.1 + 0.1
False
```

PR #960, opened 2016-04-18
BUGFIX: updating tests (PR to 0.8.x)

to266:
Just like https://github.com/hyperspy/hyperspy/pull/959, but for the 0.8.x version

...By mistake I started with the master version, so this is cherry-picked from the #959. As a result, I may have accidentally picked up a new feature or two. Hopefully I did not, but better for someone else checks too.

### DIFF
--- a/hyperspy/tests/signal/test_fancy_indexing.py
+++ b/hyperspy/tests/signal/test_fancy_indexing.py
@@ -33,56 +33,41 @@ class Test1D:
         self.data = self.signal.data.copy()
 
     def test_slice_None(self):
-        s = self.signal[:]
+        s = self.signal.isig[:]
         d = self.data
-        assert_true((s.data == d).all())
+        np.testing.assert_array_equal(s.data, d)
         assert_equal(s.axes_manager._axes[0].offset,
                      self.signal.axes_manager._axes[0].offset)
         assert_equal(s.axes_manager._axes[0].scale,
                      self.signal.axes_manager._axes[0].scale)
 
-    def test_std_slice(self):
-        s = self.signal[1:-1]
-        d = self.data[1:-1]
-        assert_true((s.data == d).all())
-        assert_equal(s.axes_manager._axes[0].offset, 1)
-        assert_equal(s.axes_manager._axes[0].scale,
-                     self.signal.axes_manager._axes[0].scale)
-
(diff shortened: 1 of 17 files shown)


hyperspy/tests/signal/test_fancy_indexing.py
```diff
@@ -33,80 +33,65 @@ def setUp(self):
         self.data = self.signal.data.copy()
 
     def test_slice_None(self):
-        s = self.signal[:]
+        s = self.signal.isig[:]
```
dnjohnstone: Is this as it should be in 0.8.5? I forgot which version this is deprecated in?

hyperspy/tests/signal/test_fancy_indexing.py
```diff
@@ -243,19 +237,10 @@ def setUp(self):
         self.signal.axes_manager._axes[1].navigate = True
         self.signal.axes_manager._axes[2].navigate = False
```
dnjohnstone: Did the test below disappear for a reason?

hyperspy/tests/signal/test_fancy_indexing.py
```diff
@@ -327,24 +312,20 @@ def test_negative_float_step(self):
 class TestEllipsis:
 
     def setUp(self):
-        self.signal = Signal(np.arange(2 ** 4).reshape(
-            (2, 2, 2, 2)))
+        self.signal = Signal(np.arange(2 ** 5).reshape(
+            (2, 2, 2, 2, 2)))
         self.data = self.signal.data.copy()
```
dnjohnstone: Did the test below disappear for a reason?

dnjohnstone:
There are two tests that seem to have disappeared in this PR - which I don't think should have... Other than that, I think this is ready to merge. I think the cherry picking went ok.

to266:
In this case I took the liberty to remove all the "deprecated style" tests - even though the functions technically are still there, I don't think it's worth even testing that they work (in fact if they broke, it might be better, as that may encourage people to switch to the proper syntax).

Given the above thinking, the two removed tests were not necessary.

dnjohnstone:
ok thanks for clarifying that - I think this is fine then, as far as I can see.

Issue #943, opened 2016-04-06
Size of signal.py


dnjohnstone:
signal.py is getting very big and is expanding rapidly as people add more and more functionality.

What is, or should be, our strategy for keeping this manageable? Presumably some separation of code to different places in a rational way, but what might the different sections of code be?

francisco-dlp:
+1000!

Actually this touches a deeper issue that we should discuss. Currently we have `Signal1DTools`, `Signal2DTools`, `Signal` (which inherits from the previous two), `Image` and `Spectrum`. `Image` and `Spectrum` inherit from `Signal`, and just add some extra few methods. This means that these objects are highly cluttered. The rationale was that, because we can configure the "navigate" attribute of the axes at will, having all these methods always available meant that we could e.g. while working with a spectrum image set the axes in "image configuration", apply an image method, and go back to "spectrum configuration", without actually transforming the class (which currently requires rearranging the data in memory).

I would like to propose the following alternative:

* Get rid of  `Image` and `Spectrum` in favour of `Signal1D` and `Signal2D` (I think that this was proposed by @dnjohnstone a while ago). `Signal1D` = `Spectrum` ∪ `Signal1DTools` ∪ `Signal` and `Signal2D` = `Image` ∪ `Signal2DTools` ∪ `Signal`
* Add a `copy_data` argument to `as_image` and `as_spectrum` (to be renamed `as_signal1D`, `as_signal2D`).

What do you think?

dnjohnstone:
Ok I'm all for your suggestion @francisco-dlp - I can't remember where we had that discussion before, but that is also my recollection of the conclusion.

I'm happy to do the grunt work on initially separating out the code, the question is when? ASAP and for v1.0.0 or (given it will break so much for everyone) - do we take this as a 1.1.0 target and warn everyone now that it's coming?

francisco-dlp:
I think that 1.0 is appropriate for this. Even `0.8.5` (due very soon as we've fixed some few important issues) would be as long as we don't modify the current API but simply add the new `Signal1D` and `Signal2D` classes and add deprecation warnings to `Spectrum` and `Image`. In this way, we'll be able to fully get rid of them for `1.0.0`.

francisco-dlp:
I've set the milestone to `next_maintenance_realease`. @dnjohnstone, are you happy to undertake the split? The changes to `as_image` and `as_spectrum` can be introduced in a different PR to `master` instead.

dnjohnstone:
[...]

PR #961, opened 2016-04-19
Separation of Signal1D & Signal2D

dnjohnstone:
To address #943.

Very much a work in progress, I broke lots of things hacking this split around, but I just want to check that everyone is happy with the general direction of this? i.e. move image & Signal2DTools & image tools to a new file called signal2D?

Let me know if not, before I spend lots of time fixing things.


francisco-dlp:
Looks good. It's a good idea to move the stuff in misc to the signal1d and signal2d file. I think that the structure needs some further thought to ease the transition to `1.0`. In particular, if we are going to deprecate `Spectrum`, `Image` and `Signal`, we need to provide the objects that are going to replace them, in addition to keeping the old API. To make things simple, I think that we should indeed deprecate `Signal` and create a new class `BaseSignal`. What do you think of the following:


* Rename `Signal` -> `BaseSignal` and change it inheritance

```python

class BaseSignal(FancySlicing,
             MVA,
             MVATools,):

```

* Recreate the old `Signal` (with deprecation warning)

```python
class Signal(BaseSignal,
                   Signal1DTools,
                   Signal2DTools,):

```

* The new `Signal1D` should be:

```python
class Signal1D(BaseSignal,
                   Signal1DTools,):

```

* And the new `Signal2D`:

```python
class Signal2D(BaseSignal,
                   Signal2DTools,):

```

* `Signal2DTools` and `Signal1DTools` should be private and most likely we'll get rid of them for `1.0`.

* Add Spectrum (with deprecation warning)

```python
class Spectrum(Signal1D,
                         Signal2DTools,):
```

* Add Image (with deprecation warning)

```python
class Image(Signal2D,
                   Signal1DTools,
                  ):
```

* `EELSSpectrum` and `EDSSpectrum` should still inherit from `Spectrum` instead of `Signal1D` (we can't break the API in a minor release), but they should include a deprecation warning explaining that for 1.0 they'll inherit from `Signal1D`.

dnjohnstone:
@francisco-dlp I've tried to implement what you suggest, but I have a few questions/issues:

1. Travis is currently complaining about import errors - but I don't understand why? Half of them seem to me to be for parts of the code I haven't touched -- I'm sure it makes sense, I'm probably just being stupid.

2. When you say "recreate old Signal/Spectrum/Image with deprecation warnings" what is actually needed in those functions? I think I've done what is required but would be good to check

hyperspy/_signals/image.py
```diff
@@ -17,134 +17,12 @@
...
 
 
-from hyperspy.signal import Signal
+from hyperspy._signals.signal2D import Signal2D, Signal2DTools
+from hyperspy.misc.hspy_warnings import VisibleDeprecationWarning
 
 
-class Image(Signal):
+class Image(Signal2D,
+            Signal2DTools,):
```
francisco-dlp: Shouldn't this be `Signal1DTools`?

hyperspy/_signals/image_simulation.py
```diff
@@ -17,9 +17,9 @@
 # along with  HyperSpy.  If not, see <http://www.gnu.org/licenses/>.
 
 
-from hyperspy._signals.image import Image
+from hyperspy._signals.signal2D import Signal2D
 from hyperspy._signals.simulation import Simulation
 
 
-class ImageSimulation(Simulation, Image):
+class ImageSimulation(Simulation, Signal2D):
```
francisco-dlp: I think that this one didn't require any modification since we have too keep the API intact.

hyperspy/_signals/spectrum.py
```diff
@@ -16,135 +16,13 @@
...
-import numpy as np
+from hyperspy._signals.signal1D import Signal1D, Signal1DTools
+from hyperspy.misc.hspy_warnings import VisibleDeprecationWarning
 
-from hyperspy.exceptions import DataDimensionError
-from hyperspy.signal import Signal
-from hyperspy.gui.egerton_quantification import SpikesRemoval
 
+class Spectrum(Signal1D,
+               Signal1DTools,):
```
francisco-dlp: Shouldn't this be `Signal2DTools`?

hyperspy/_signals/image.py
dnjohnstone: For Image?

hyperspy/_signals/spectrum.py
dnjohnstone: For Spectrum?

francisco-dlp: It's just not to break the API.

dnjohnstone: Yes - but why 2D tools for spectrum?

dnjohnstone: ooooooh I see! Don't worry

dnjohnstone:
closed in favour of #963 as even if we decide to rename I will base the change on that branch as this one just had too much broken in it because I changed too much in one go!

PR #963, opened 2016-04-27
Signal1D/2D tools consolidation

dnjohnstone:
An alternative to #961 to address #943

That PR really breaks everything - the reason being that it creates a situation where signal.py needs to import from /_signals/signal1D.py and vice-versa resulting in nasty cyclic import issues I think.

The easiest way to get round this is to rather than putting all the signal1D_tools into signal1D instead create /misc/signal1D_tools.py that includes the old spectrum_tools.py. Whilst this doesn't reduce the number of files as much as the previous attempt it is intuitive and results in 5 (signal.py, signal1D.py, signal2D.py, signal1D_tools.py, and signal2D_tools.py) much more tolerably sized files - which was the primary aim of this endeavour.

Since this works and makes a lot of sense to me - I think this is the way to go... opinions elsewhere?


dnjohnstone:
By the way - performing the renaming Spectrum --> Signal1D and Image --> Signal2D seems to be much more painful... I can't make that work without breaking a vast number of tests that I think basically don't like the change in API...

As it stands - I'm starting to wonder if deprecating Spectrum/Image is a good idea at all? This PR achieves the primary goal of reducing the bloating of signal.py and it also achieves the second most important goal of consolidating tools for 1D & 2D signals into one place so that there is only one place to look for that.

The renaming is now just that, a renaming and I think that many people will complain that Signal1D/Signal2D are less intuitive names than Spectrum/Image -- especially if you're a microscopist!

On the other hand, if we're ever going to make that change it wants to be ASAP as it will only get more painful so should be done before 1.0.0 if we're going to go for it.

@francisco-dlp @to266 opinions would be most welcomed

to266:
I'm fine either way. However, it's not really clear to me if we want to rename everything: e.g. methods `as_spectrum` and similar? Also, since this seems to be aimed to be released before `1.0.0`, then do we want to change `model.spectrum` -> `model.signal1D`? 

Essentially lots of small things that are not immediately clear to me.

Or do we just want to remove the ability to create new signals using `spectrum` and `image` keywords, and use `signal1D` and `signal2D` instead?

magnunor:
I'm in favor of splitting it up like that, which should (?) make things a bit more orderly. And in my head, it should make it easier to create more specialized classes which inherit either `signal1D` or `signal2D`. For example, making `STEMImage` or `DiffractionImage` classes which include useful functions for analysing those kinds of signals.

dnjohnstone:
Ok - I discussed this with @francisco-dlp in person yesterday and I thinkwe broadly agree with @magnunor that this change will put everything on a 'more general' footing moving forward. In the current version of this PR I've renamed Spectrum --> Signal1D and Image -->Signal2D as you can see it breaks an absolute tonne of tests e.g. anything that checks something is in the Signal class because that no longer exists.

I will plod through all of this eventually, but if anyone feels like helping it will be much appreciated.

@to266 I agree all of that is unclear, but I think the idea is that it should all be renamed too.

dnjohnstone:
ok @francisco-dlp this is getting pretty close now... would be a good moment if you can have a look through to let me know if I've missed anything major!

dnjohnstone:
Also note I moved some things from Signal2DTools to drawing.utils because not doing so resulted in a cyclic import mess - since it's only used there I think this is ok.

dnjohnstone:
I'm also not totally sure why coverage is decreasing...

francisco-dlp:
Looks very good!

Some comments:

* `Signal` should be renamed to `BaseSignal`
* The old `Signal` should be recreated as `class Signal(BaseSignal, Signal1DTools, Signal2DTools)` with a deprecation warning.
* `BaseSignal`, `Signal1D`, and `Signal2D` should be available in `hyperspy.api.signals`
* `to_spectrum`, `to_image`, `as_spectrum` and `as_image` must be renamed, and new methods with the same name created with deprecation warnings.
* The image and spectrum tests should be renamed and test `Signal1D` and `Signal2D` instead.
* The documentation (user guide and docstrings) should:
    * Use `Signal1D` and `Signal2D` instead of `Spectrum` and `Image`
    * `to_signal1D`, `to_signal2D`, `as_Signal2D` and `as_signal1D` instead of their old selves. 
    * Inform that `Spectrum`, `Image`, and `Signal` are deprecated and encourage the reader to update his scripts.
    * Sentences like "These methods are only available for Signal object with signal_dimension equal to one." require updating.

dnjohnstone:
@francisco-dlp the issue with having signal.Signal as you describe is that it creates a cyclic import with signal1D and signal2D.... The only way I can see to resolve this is to make separate .py files for signal and base signal.

in the end I think we want the API to have signal.BaseSignal but I'm not sure if I'm allowed to do that and then create signal_old.Signal to give the deprecation warning? Otherwise there will be another rename needed when merging with master.... what do you prefer?

dnjohnstone:
@francisco-dlp this should be getting pretty close now -- there are a few broken tests and I think most of them come down to having broken things in the io... I don't really understand any of that section of the code though (I've really ignored file readers mostly) so a fresh pair of eyes would be greatly appreciated.

dnjohnstone:
Ps.. I'm aware I'll have to revisit the documentation at least once more, but that's relatively minor...

dnjohnstone:
Having played with this a bit more - I think my loader issue was me being stupid with naming attributes and is now resolved, remaining issues are:

1. red_chisq test is broken -- I can't figure out why yet.

2. making Signal1D & Signal2D available in hyperspy.signals causes problems with testing the assignment of subclasses.... setting the acquired by 'spectrum' or 'image' can go is not a 1-1 mapping right now as it could go to Signal1D or Signal2D.... Not sure what the preferable way to resolve this is... I could add a bit more in to the recreated Spectrum and Image subclasses so that this is only true for those subclasses (which is perhaps the only way not to break the API)  -- not totally sure though, so wisdom gladly received.

3. There are a number of functions that are still stuff_spectrum etc that I need to get through, but I want to have tests working before I do that.

Finally, I'm a little worried that there may be parts of the code that I've broken but that are not tested because we don't have 100% coverage -- does anyone have thoughts on parts of the code that may be particularly vulnerable to this? May be an opportune moment to hack a few new tests together....

dnjohnstone:
Don't be fooled by travis passing here! it will fail at random due to the non 1-1 mapping of acquired by 'spectrum' and 'image' that I mentioned in the above comment.

Another remaining question is the correct naming of attributes... So far I have made it signal.metadata.Signal.stuff -- but obviously that could be different, I think it makes sufficient sense like this but would be glad to hear what people think.

I just don't know how the code should be structured to keep both and not break usage too much... On the plus side I think I really did resolve the issues with loading.

dnjohnstone:
@francisco-dlp I think this might actually be ready for review now....

doc/user_guide/tools.rst
```diff
@@ -60,51 +67,43 @@ The different subclasses are characterized by three
...
-    subclass using the following :py:class:`~.signal.Signal` methods:
-    :py:meth:`~.signal.Signal.as_image` and :py:meth:`~.signal.Signal.as_spectrum`.
-    In addition :py:class:`~._signals.spectrum.Spectrum` instances can be
-    transformed in images using :py:meth:`~._signals.spectrum.Spectrum.to_image`
-    and image instances in spectrum instances using
-    :py:meth:`~._signals.image.Image.to_spectrum`. When transforming between
-    spectrum and image classes the order in which the
-    data array is stored in memory is modified to improve performance. Also,
-    some functions, e.g. plotting or decomposing, will behave differently.
+    Can be "signal1D", "signal2D" or "", the latter meaning undefined and describes
```
francisco-dlp: Should this still be "spectrum" and "image"?

doc/user_guide/tools.rst
```diff
@@ -10,36 +10,43 @@ The Signal class and its subclasses
...
 * :py:class:`~._signals.image_simulation.ImageSimulation`
 
+.. versionchanged:: 0.8.5
+
+Note that in 0.8.5 the :py:class:`~._signals.signal1D.Signal1D` and :py:class:`~._signals.signal2D.Signal2D` classes were created to deprecate the old :py:class:`~._signals.spectrum.Spectrum` and :py:class:`~._signals.image.Image` classes.  
+
+
 The :py:mod:`~.signals` module, which contains all available signal subclasses,
 is imported in the user namespace when loading hyperspy. In the following
 example we create an Image instance from a 2D numpy array:
```
francisco-dlp: `Signal1D` instance?

francisco-dlp:
All file names should be in lowercase, see  the relevant section in [PEP8](https://www.python.org/dev/peps/pep-0008/#package-and-module-names).

dnjohnstone:
as in you want signal1D.py to become signal1d.py?

francisco-dlp:
Yes, I think that that's the way they should be. The same for the rst files.

dnjohnstone:
yeah makes sense... I think I had the same thing with model1d/model2d...

doc/user_guide/tools.rst
```diff
@@ -60,51 +67,43 @@ The different subclasses are characterized by three
...
-    data array is stored in memory is modified to improve performance. Also,
-    some functions, e.g. plotting or decomposing, will behave differently.
+    Can be "signal1D", "signal2D" or "", the latter meaning undefined and describes 
+    the way the data is arranged in memory. It is possible to transform any 
+    :py:class:`~.signal.BaseSignal` subclass to a :py:class:`~._signals.signal1D.Signal1D`
+    or :py:class:`~._signals.signal2D.Signal2D` subclass using the following 
+    :py:class:`~.signal.BaseSignal` methods: :py:meth:`~.signal.BaseSignal.as_signal2D` 
+    and :py:meth:`~.signal.BaseSignal.as_signal1D`. In addition 
+    :py:class:`~._signals.signal1D.Signal1D` instances can be transformed into 
+    two-dimensional signals using :py:meth:`~._signals.signal1D.Sinal1D.to_signal2D`
```
francisco-dlp: typo: `~._signals.signal1D.Sinal1D.to_signal2D`

doc/user_guide/tools.rst
```diff
@@ -114,7 +113,7 @@ The different subclasses are characterized by three
     +---------------------------------------------------------------+-----------+-------------+---------------+
     |           :py:class:`~._signals.eds_tem.EDSTEMSpectrum`       | spectrum  |   EDS_TEM   |       -       |
     +---------------------------------------------------------------+-----------+-------------+---------------+
-    |              :py:class:`~._signals.image.Image`               |   image   |      -      |       -       |
+    |              :py:class:`~._signals.signal2D.Signal2D          |   image   |      -      |       -       |
```
francisco-dlp: Missing back-quote.

francisco-dlp:
Looks great. Just some few remaining issues:

* `as_image`, `as_spectrum`, `to_image`, `to_spectrum` must be recreated just for `0.8.5` and raise deprecation warnings.
* Creating `Spectrum`, `Image` and `Signal`  should raise deprecation warnings.

dnjohnstone:
Thanks @francisco-dlp - but why is creating a Spectrum/Image/Signal not raising a deprecation warning already? I thought that was what VisibleDeprecationWarning did?

Evidently I don't understand...

doc/user_guide/tools.rst
dnjohnstone: yes - thank you.

francisco-dlp:
I think that in the User Guide and docstrings we should also replace `Image`, `Spectrum` and `Signal` with `Signal1D`, `Signal2D` and `BaseSignal` at this point.

francisco-dlp:
And also in the tests, except for those few tests (such as the `assign_subclass` one) that still need to test `Spectrum` and `Image`.

dnjohnstone:
ok @francisco-dlp let me know if more is needed! I think I'm pretty much there at least though.

doc/user_guide/tools.rst
```diff
@@ -10,36 +10,43 @@ The Signal class and its subclasses
...
+:py:class:`~._signals.signal1D.Signal1D` class for electron energy-loss
 spectroscopy data analysis.
 
+.. versionchanged:: 0.8.5
+
 Currently the following signal subclasses are available:
 
-* :py:class:`~._signals.spectrum.Spectrum`
-* :py:class:`~._signals.image.Image`
+* :py:class:`~._signals.signal1D.Signal1D`
```
francisco-dlp: After making the name of the files lower-case, the module names need updating here (there) and everywhere e.g. `~._signals.signal1D.Signal1D` -> `~._signals.signal1d.Signal1D`

francisco-dlp:
The `as_image`, `as_spectrum`, `to_image` and `to_spectrum` methods need to be functional. Currently they're not because they don't return anything.

Also, unfortunately this is trickier than we thought: `as_image` should return an `Image` instance, while `as_signal2D` should return a `Signal2D` instance and so on and so forth...

hyperspy/drawing/mpl_hse.py
```diff
@@ -83,7 +83,7 @@ def plot_signal(self):
             self.xlabel += ' (%s)' % self.axes_manager.signal_axes[0].units
         self.ylabel = 'Intensity'
         self.axis = self.axes_manager.signal_axes[0].axis
-        sf = spectrum.SpectrumFigure(title=self.signal_title +
+        sf = spectrum.Signal1DFigure(title=self.signal_title +
```
francisco-dlp: `spectrum.SpectrumFigure` probably needs renaming -> `signal1d.SpectrumFigure`?

The same goes for the equivalent image module.

dnjohnstone:
Getting these functions to return Image/Spectrum/Signal1D/Signal2D is going to require changing the default "record_by" attributes, which so much of the code uses to switch between Spectrum and Image. Do you think this will cause a problem for inheritance if it's different for Spectrum & Signal1D?

Also, while we're touching this - we should rename the "record_by" attribute I think -- it has always confused me because "record by spectrum" is nonsense in English, it just doesn't mean anything.

I guess the original idea was to express the notion of the signal having been "recorded as" a spectrum or an image?

francisco-dlp:
That's a good point. +1000 for getting rid of `record_by` as it is true that it doesn't make much sense This concept in inherited from Lispix and digital micrograph and makes sense from the acquisition software point of view I think. The idea is that some acquisition devices acquire spectrum images as series of images. If the data was to be stored in `C` order as it is being acquired in the instrument the order would be `(spectrum dimension,  y, x)`. However, for processing it is better to store it like this `(y, x, spectrum dimension)`. The latter would be "record_by" spectrum and the former "record_by" "image".

However, using this nomenclature in HyperSpy was a mistake as it is not general enough, and it unnecessarily confusing. However, I think that it'll be better to address this in a separate issue an pull request.

Regarding the issue of returning the right object, I agree with you, I think the benefit of being fully consistent doesn't justify the effort in this occasion. So, unless someone disagrees, I suggest just adding a note to the `??_signa?D` methods explaining that they'll return the right thing in v1.0. Does it sound reasonable?

dnjohnstone:
@francisco-dlp I think I've made the changes you've asked for above.

francisco-dlp:
Plotting is broken, try for example `hs.signals.Spectrum(np.random.random((100, 100,1000))).plot()`

dnjohnstone:
Argh - I still can't reproduce that error, although I can only assume my computer is running on magic as it really should have been broken... hopefully fixed now!

francisco-dlp:
It's still broken, but in a different way:

```python

In [2]: hs.signals.Spectrum(np.random.random((100, 100,1000))).plot()
/home/fjd29/Python/hyperspy3/hyperspy/signals.py:54: VisibleDeprecationWarning: The Spectrum class will be deprecated from version 1.0.0 and replaced with Signal1D
  VisibleDeprecationWarning)
/home/fjd29/Python/hyperspy3/hyperspy/signals.py:126: VisibleDeprecationWarning: The Signal class will be deprecated from version 1.0.0 and replaced with BaseSignal
  VisibleDeprecationWarning)
/home/fjd29/Python/hyperspy3/hyperspy/signal.py:3482: VisibleDeprecationWarning: The as_image method will be deprecated from version 1.0.0 and replaced with as_signal2D
  VisibleDeprecationWarning)
<ipython-input-2-9c67a2332f16> in <module>()
----> 1 hs.signals.Spectrum(np.random.random((100, 100,1000))).plot()

/home/fjd29/Python/hyperspy3/hyperspy/signal.py in plot(self, navigator, axes_manager, **kwargs)
   1799                     navigator = navigator.as_spectrum(0)
   1800                 else:
-> 1801                     navigator = navigator.as_image((0, 1))
   1802             else:
   1803                 navigator = None

/home/fjd29/Python/hyperspy3/hyperspy/signal.py in as_image(self, image_axes)
   3482                       VisibleDeprecationWarning)
   3483 
-> 3484         im = self.as_signal2D()
   3485         return im
   3486 

TypeError: as_signal2D() missing 1 required positional argument: 'image_axes'

```

Try deleting your *.pyc files to reproduce the error.

dnjohnstone:
well for the first time I've been happy when restarting things has resulted in code breaking - so I did get it to break, and then fixed it... 

As an aside - is there a list of "standard tests" you run like this that check parts of the code not covered by proper tests? I could happily run them myself if that were the case.

francisco-dlp:
`plot_images` is broken:

```python

ImportError                               Traceback (most recent call last)
<ipython-input-4-44c5a631d171> in <module>()
     12 
     13 image.metadata.General.title = 'multi-dimensional Lena'
---> 14 hs.plot.plot_images(image, tight_layout=True)

/home/fjd29/Python/hyperspy3/hyperspy/drawing/utils.py in plot_images(images, cmap, no_nans, per_row, label, labelwrap, suptitle, suptitle_fontsize, colorbar, centre_colormap, saturated_pixels, scalebar, scalebar_color, axes_decor, padding, tight_layout, aspect, min_asp, namefrac_thresh, fig, *args, **kwargs)
    561     from hyperspy.drawing.widgets import Scale_Bar
    562     from hyperspy.misc import rgb_tools
--> 563     from hyperspy.signal import Signal
    564 
    565     if isinstance(images, Signal) and len(images) is 1:

ImportError: cannot import name 'Signal'

```

Knowing that the plotting features are not tested, I just try plotting things to see if it works...

dnjohnstone:
This is breaking several things as I work through some notebooks - give me a while...

dnjohnstone:
Ok this is now behaving the same as 0.8.x when running through the EDX tutorial - however something seems to be broken in both! The error message below is in 0.8.x not this branch and I get the same in this branch...

broken

francisco-dlp:
Could you report the marker problem in a separate issue?

dnjohnstone:
#980

to266:
I haven't finished looking into this, but a couple of comments:

 - I find it unintuitive / inelegant that when picking `as_signal1D` and `as_signal2D` instead of the `as_spectrum` and `as_image`, they both return a `Spectrum` and `Image` respectively, and then complain that you should use `signal1D` and `signal2D`. I understand that it's only for the 0.8.x version, to give people some time to adjust, but I'm not a fan :)

 - Do we keep the `EELSSpectrum` and all the `EDS*EMSpectrum` classes? Will those get renamed to `EELSSignal1D`, etc.?

dnjohnstone:
EELSSpectrum & EDS*Spectrum definitely stay - because that's what they are... The point is that now if someone has say an PowderXRD time series (which would be a good thing to analyse with HyperSpy) it can inherit from Signal1D which makes sense as it is a one-dimensional signal rather than from Spectrum, which doesn't make so much sense as it's not really a Spectrum in any sense.

As regards as_stuff methods -- in principle I agree. However, I think we should address #974 at this point (or in a separate PR) rather than spending time making record_by consistent for everything in 0.8.5 only to remove it...

francisco-dlp:
@dnjohnstone, although we agree above on the current solution for the `as_*` and `to_*` methods, the fact that @to266 is not happy with it has made me change my mind. Actually, it'll be easy to make this work without addressing the `record_by` issue. `to_sig*` and `as_sig*` could call a new `assign_signal_subclass` method which could be exactly as the current one but deleting `Signal`, `Spectrum` and `Image` from the dictionary instead of `BaseSignal`, `Signal1D` and `Signal2D`. Obviously this new method is just a temporary hack and we'll delete when merging into master.

Beside, the docstrings of the `as_signal*` and `to_signal*` methods were not updated to the new nomenclature. Some keyword arguments (e.g. `image_axes`) many need updating too.

to266:
Also it seems `mva/test_decomposition.py`, `signal/test_apply_function.py`, `signal/test_krames_kronig_transform.py` and `signal/test_binned.py` still have calls to `hs.signals.Signal` that should (I assume) be replaced with `hs.signals.BaseSignal`

francisco-dlp:
I've just noticed that `model.spectrum` has been renamed to `model.signal1D`. Two comments:

* Wouldn't simply `model.signal` be a better name (to be consistent with master)?
* Nevertheless, we can't break the API, so `model.spectrum` should remain in place and functional, but raise a deprecation warning.

dnjohnstone:
Ok - I've made the requested updates to tests because they're easy.

model.spectrum --> model.signal1D:  Also easy and I've done locally, but just for clarity. The point is that in master it should become model.signal1D for model1D and model.signal2D for model2D.  (in master it is model.spectrum and model.image right now) I thought that if I change it like this now then for 1.0.0 all that would be new is model.signal2D. (Note that the merge for this will be a pain, but not so bad) Fine if we need a model.spectrum for 0.8.5 - but is my overall thinking correct here?

record_by needs a little more careful thought --> what you suggest @francisco-dlp doesn't work easily because Signal1D would need to have record_by "signal1D" but Spectrum, which inherits from Signal1D, would need to have record_by "spectrum".  There are also many tests that check this attribute is "spectrum" or "image".

It's a little unclear to me whether I should just change all of these to "signal1D" and "signal2D", which will appear to fix the problem short term but I don't think is a good idea if we're then going to change that. (Also whilst record_by spectrum and record_by image have some historical sense in EM record_by signal1D etc is a very odd attribute to have in my mind). Really I think we need to have a clear plan for record_by - if we're going to remove it, or use it differently for version 1.0.0 then we should work out what we're going to do as soon as possible and make the change in this PR consistent with that plan.... With all this in mind I really prefer the current (if slightly) ugly, solution because it's soooo much easier and changes the usage minimally, which is probably good for 0.8.5. To me gentle hints that we're moving away from it but nothing breaking if you for some reason set record_by = spectrum for example is a good thing. What do you think - to me it's all about what we want to do with record_by beyond this PR? @to266 & @francisco-dlp

dnjohnstone:
This now has previous points addressed (I think) aside from the record_by question - happy with whatever the solution settled on there is, but just need to know...

francisco-dlp:
Once you merge 0.8.x into this I merge it. I've just merged 0.8.x into master to easy the merge of this into master.

dnjohnstone:
Ok - I think this is ready then....

Issue #964, opened 2016-04-28
General documentation improvements


dnjohnstone:
The documentation is improving slowly but there are various things that could be improved, which I keep noticing and then forgetting about because they're rarely an immediate priority. Nevertheless, I think it would be good if we collectively work to keep this improving particularly for upcoming workshops in the summer conference period.

Here is an initial list:

1.  Dielectric function tools should probably be documented under EELS for the time being, in my opinion.

2. In many places, right now I'm thinking EDS sections, there is quite a bit of dodgy grammar going on.

3. Decomposition sections could be improved to direct new users to some appropriate literature, or at least make it unavoidably obvious that the stuff in scikit-learn should be read. In the places where papers are already mentioned the referencing should be updated to be consistent with the Egerton paper and with recent EDS contributions.

4. There is absolutely no documentation associated with events (e.g. how that's implemented and therefore how to connect things properly for developers) nor is there anything associated with ROIs -- the latter is certainly necessary for 1.0.0 but it's unclear to me who is taking that forward now.

tjof2:
@dnjohnstone I'm happy to look at grammar for starters (I'm looking at EDS right now, in fact...)

francisco-dlp:
There is documentation for [ROIs](http://hyperspy.readthedocs.io/en/latest/user_guide/tools.html#region-of-interest-roi) and [events](http://hyperspy.readthedocs.io/en/latest/user_guide/events.html).

In a recent private communication, @mwalls pointed to some additional documentation issues and volunteered to enhance the Getting Started section of the User Guide to make it more newbie friendly.

To give to this issue the attention that it deserves, I propose to make this a permanent issue to discuss documentation improvements i.e. I'll assign it to the 1.0 milestone, but we'll  move it to the 2.0 once we release 1.0 and so on.

dnjohnstone:
Ah thanks @francisco-dlp - incidentally, is it written anywhere how to read those docs associated with the master?

francisco-dlp:
Good point, no, it's not.

tjof2:
-  The Image tools documentation isn't particularly enlightening, e.g. _"The align2D() method provides advanced image alignment functionality, including subpixel alignment."_ - without going into what the rest of functionality is or how to use it!

dnjohnstone:
[...]

PR #969, opened 2016-04-29
DOC: Tidied up machine learning documentation

tjof2:
As mentioned in #964, I've tidied up the structure and grammar of the Machine Learning documentation page.

### DIFF
--- a/doc/user_guide/mva.rst
+++ b/doc/user_guide/mva.rst
@@ -134,7 +134,12 @@ the decomposition model. You can easily calculate and display the residuals:
 Non-negative matrix factorization
 ----------------------------
 
-HyperSpy also enables the use of non-negative matrix factorization (NMF).
+Another popular decomposition method is non-negative matrix factorization (NMF), which
+can be accessed in HyperSpy with:
+
+.. code-block:: python
+
+   >>> s.decomposition(algorithm='nmf')
 
 Poissonian noise
 ----------------
--- a/doc/user_guide/mva.rst
+++ b/doc/user_guide/mva.rst
@@ -21,12 +21,15 @@ other with the dimension of the navigation space known as `loadings`.
 
 .. _decomposition:
 
-Principal component analysis
+Decomposition
 =============
 
 Decomposition techniques are most commonly applied as a means of noise reduction (or `denoising`)
-and dimensionality reduction. One of the most popular decomposition methods is
-principal component analysis (PCA). To perform PCA on your dataset, run the
+and dimensionality reduction.
+
+Principal component analysis
+----------------------------
+One of the most popular decomposition methods is principal component analysis (PCA). To perform PCA on your dataset, run the
 :py:meth:`~.learn.mva.MVA.decomposition` method:
 
 .. code-block:: python
@@ -48,7 +51,7 @@ information please read the method documentation.
 .. _scree-plot:
 
 Scree plots
------------
+^^^^^^^^^^^
 
 PCA will sort the components in the dataset in order of decreasing
 variance. It is often useful to estimate the dimensionality of the data by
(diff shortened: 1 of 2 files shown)


tjof2:
Failing appveyor because of #914

dnjohnstone:
For me the main issue with the machine learning section was  that there is absolutely no reference to the fact that the decomposition() method allows you to use any of the methods from scikit-learn -- for example NMF or SparsePCA. We should be linking to the scikit-learn docs and explaining that these are all available I think....

I can add this in a separate PR if you like, this one serves its purpose of tidying the existing work up somewhat.

tjof2:
@dnjohnstone I agree - on reading/editing it I realised there was a lot missing (as you say, e.g. NMF), but I decided against changing it too much for now. Crucially, the introduction paragraph needs expanding upon:

> HyperSpy provides easy access to several "machine learning" algorithms that can be useful when analysing multi-dimensional data. In particular, decomposition algorithms, such as principal component analysis (PCA), or blind source separation (BSS) algorithms, such as independent component analysis (ICA), are available through the methods described in this section.

Plus a section on NMF (after PCA+denoising, and before or after BSS/ICA) and a link to e.g. the SPED paper showing an example of the use of NMF.

dnjohnstone:
Ok thanks @tjof2.

In my opinion this is moving in the right direction and it should probably fall to someone else i.e. me to make the link to scikit-image and describe other algorithms as I've used them a fair bit. Therefore I would merge this as I can't see any big issues.

francisco-dlp:
It looks really good.

I agree that it should be expanded to include NMF. But then, why renaming the decomposition section to PCA? Wouldn't it be better to have a decomposition section with NMF and PCA as subsections? Also, the poisson noise normalisation applies to all decomposition algorithms, while the scree plot only to PCA. Sparse PCA and fast_svd are also missing.

dnjohnstone:
That's a good point @francisco-dlp I think an overall heading "Decomposition" with "Principal Component Analysis" and "Non-negative matrix factorisation" as sub headings is a good move if only because it mirrors usage...

One thing that has seemed slightly inconsistent to me for some time is the separation of NMF and ICA in the code such that they are implemented in different parent methods. In many ways NMF is much more similar to ICA than to PCA most notably the NMF implementation we use actually first does an SVD to initialise the problem and then makes it non-negative and optimises.

Personally I prefer to think of them all as constrained matrix factorisations with different constraints and would rather have ICA within decomposition and just run the prior PCA first by default... This is way beyond the scope of this PR though!

tjof2:
@francisco-dlp  a valid point, I missed it initially - changed the headings and put scree plots under PCA.

Poisson noise section sits just below PCA, I presume NMF will sit nicely in between,

tjof2:
@francisco-dlp to clarify, is there anything else you'd like me to change in this PR? Certainly there's more to be done to improve the docs on this in future as @dnjohnstone mentions.

dnjohnstone:
To me this is an improvement and so should be merged - we can get to making it even better when we get there I think... although could also leave it as a PR to keep it fresh in people's minds

doc/user_guide/mva.rst
```diff
@@ -4,104 +4,68 @@ Machine learning
...
-estimation of the dimensionality of the data (or equivalently, the number of
-components that should be retained - see below).
+plotting the explained variance against the component index. This plot is
+sometimes called a scree plot and it should drop quickly,
+eventually becoming a slowly descending line.
+
+The point at which the scree plot becomes linear (often referred to as
+the `elbow`) is generally judged to be a good estimation of the dimensionality
+of the data (or equivalently, the number of components that should be retained
+ - see below).
```
francisco-dlp: an hyphen at the beginning of the line becomes a bullet in rest...

francisco-dlp:
Other than the little typo above, I agree that this is ready to merge.

tjof2:
@francisco-dlp fixed.

PR #973, opened 2016-05-03
Remove automatically generated API rst files

francisco-dlp:
Currently git tracks the files in `hyperspy/doc/api`. However, these are automatically generated by sphinx apidoc, so I don't think that there is any need to track them.


vidartf:
Maybe add folder to .gitignore as well?

dnjohnstone:
is this just waiting for the suggestion from @vidartf to be implemented?

francisco-dlp:
Interestingly, it seems that [it's already there](https://github.com/hyperspy/hyperspy/blob/0.8.x/.gitignore#L28). Can we merge this?

vidartf:
Aside from noting the interesting new failure of Travis (timeout on save of long list), I'm ok with it :)

Issue #983, opened 2016-05-06
Saving signals EMD requires certain metadata


magnunor:
Trying to save a minimal `spectrum` in the EMD-file format does not work:

``` Python
s = hs.signals.Signal(range(100))
s.save("test.emd")
```

Does not work. To make it work, one needs to add several metadata fields like "user", "comments", ...

While the EMD-specification (http://emdatasets.lbl.gov/spec/#meta) encourages using metadata, all the metadata fields like this are optional.

So we should change the EMD file writer to allow for saving signals without these metadata fields.

magnunor:
Having a look at this now

CodeMonkeyJan:
I'm on vacation till Tuesday starting today, I'll also have a look as soon as I come back, thanks for pointing this out!

magnunor:
@CodeMonkeyJan, sounds good. In case you've missed it, I've implemented some fixes here: https://github.com/hyperspy/hyperspy/pull/985

I'll wait until you've had time to look at it before doing additional changes.

CodeMonkeyJan:
Thanks for all the work :-), I will post a longer reply there!

magnunor:
Fixed in https://github.com/hyperspy/hyperspy/pull/1424

PR #985, opened 2016-05-06
Remove requirement for metadata in signals when saving as EMD

magnunor:
https://github.com/hyperspy/hyperspy/issues/983

I fixed that particular bug, but the program will still make the metadata folders. For example:

```Python
s = hs.signals.Signal(range(100))
s.save("test.emd")
```

Will make an EMD-hdf5 file with several metadata empty groups, even if none of this was in the original signal.

In addition, loading an EMD-files without any of this metadata will automatically populate the metadata in the HyperSpy signal with empty metadata, like: 

```Python
├── General
│   ├── comments
│   ├── microscope
│   │   ├── name = 
│   │   └── voltage = 
│   ├── sample
│   │   ├── material = 
│   │   └── preparation = 
│   ├── title = __unnamed__
│   └── user
│       ├── department = 
│       ├── email = 
│       ├── institution = 
│       └── name =
```

What should be the default behaviour? In my opinion, these metadata groups should not be automatically created like this.

### DIFF
--- a/hyperspy/io_plugins/emd.py
+++ b/hyperspy/io_plugins/emd.py
@@ -377,21 +377,33 @@ def file_reader(filename, load_to_memory=True, log_info=False, **kwds):
 def file_writer(filename, signal, signal_metadata=None, user=None,
                 microscope=None, sample=None, comments=None, **kwds):
     if user is None:  # If not provided, look in metadata:
-        user = signal.metadata.General.as_dictionary().get('user')
-    if user is None:  # If not found, check original_metadata:
-        user = signal.original_metadata.General.as_dictionary().get('user')
+        if 'user' in signal.metadata.General.as_dictionary():
+            user = signal.metadata.General.as_dictionary().get('user')
     if microscope is None:  # If not provided, look in metadata:
-        microscope = signal.metadata.General.as_dictionary().get('microscope')
-    if microscope is None:  # If not found, check original_metadata:
-        microscope = signal.original_metadata.General.as_dictionary().get('microscope')
+        if 'microscope' in signal.metadata.General.as_dictionary():
+            microscope = signal.metadata.General.as_dictionary().get('microscope')
     if sample is None:  # If not provided, look in metadata:
-        sample = signal.metadata.General.as_dictionary().get('sample')
-    if sample is None:  # If not found, check original_metadata:
-        sample = signal.original_metadata.General.as_dictionary().get('sample')
+        if 'sample' in signal.metadata.General.as_dictionary():
+            sample = signal.metadata.General.as_dictionary().get('sample')
     if comments is None:  # If not provided, look in metadata:
-        comments = signal.metadata.General.as_dictionary().get('comments')
-    if comments is None:  # If not found, check original_metadata:
-        comments = signal.original_metadata.General.as_dictionary().get('comments')
+        if 'comments' in signal.metadata.General.as_dictionary():
+            comments = signal.metadata.General.as_dictionary().get('comments')
+
+    if 'General' in signal.original_metadata.as_dictionary():
+        if user is None:  # If not found, check original_metadata:
+            if 'user' in signal.original_metadata.General.as_dictionary():
+                user = signal.original_metadata.General.as_dictionary().get('user')
+        if microscope is None:  # If not found, check original_metadata:
+            if 'microscope' in signal.original_metadata.General.as_dictionary():
+                microscope = signal.original_metadata.General.as_dictionary().get('microscope')
+        if sample is None:  # If not found, check original_metadata:
+            if 'sample' in signal.original_metadata.General.as_dictionary():
+                sample = signal.original_metadata.General.as_dictionary().get('sample')
+        if comments is None:  # If not found, check original_metadata:
+            if 'comments' in signal.original_metadata.General.as_dictionary():
+                comments = signal.original_metadata.General.as_dictionary().get('comments')
+
+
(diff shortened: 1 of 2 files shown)


magnunor:
So I changed the EMD-file saving so it no longer requires certain metadata to save a signal, allowing for saving of minimal Hyperspy signals as EMD-files.

However, I also changed the behaviour with regards to populating the EMD-hdf5 file with empty metadata strings. Do we want this?

Ergo: should the EMD-files contain empty metadata as attributes, even if they are not defined in the HyperSpy signal itself?

I feel not generating extra empty metadata groups in the EMD-files reduces the clutter. What do people think?

magnunor:
I fixed some more bugs which caused some files to not load.

hyperspy/tests/io/test_emd.py
```diff
@@ -63,6 +63,18 @@ def test_metadata():
     nt.assert_is_instance(signal, Image)
 
 
+class TestSaveSimpleSignal():
+
+    def test_save_simple_signal(self):
+        simple_signal = Signal(data_save)
+        simple_signal.save(
+                os.path.join(
+                    my_path, 'emd_files', 'example_temp.emd'), overwrite=True)
```
to266: Do you only want to check that the saving succeeds? I would probably have a look that the saved file has only the required fields defined, though it's up to you

hyperspy/io_plugins/emd.py
```diff
@@ -377,21 +393,32 @@ def file_reader(filename, load_to_memory=True, log_info=False, **kwds):
 def file_writer(filename, signal, signal_metadata=None, user=None,
                 microscope=None, sample=None, comments=None, **kwds):
     if user is None:  # If not provided, look in metadata:
-        user = signal.metadata.General.as_dictionary().get('user')
-    if user is None:  # If not found, check original_metadata:
-        user = signal.original_metadata.General.as_dictionary().get('user')
+        if 'user' in signal.metadata.General.as_dictionary():
+            user = signal.metadata.General.as_dictionary().get('user')
```
to266: A couple of things with this `as_dictionary()` play:
  - `DictionaryTreeBrowser` supports lookup, so you can do `if 'user' in signal.metadata.General` (though it's not necessary in this case).
  - The dictionary is constructed an unnecessary number of times.... It could be done once for `metadata`, once for `original_metadata`, and then just looked up.
  - I would use syntax like this:
```python
metadata = signal.metadata.General.as_dictionary()
user = user or metadata.get('user', None)
microscope = microscope or metadata.get('microscope', None)
sample = sample or metadata.get('sample', None)
# etc
```
You could add the `original_metadata` check in the same one line as well, it's up to you.
```python
orig_metadata = signal.original_metadata.as_dictionary().get('General', None)
user = user or metadata.get('user', None) or (orig_metadata.get('user', None) if orig_metadata else None)
```

hyperspy/io_plugins/emd.py
```diff
@@ -317,21 +325,29 @@ def save_to_emd(self, filename='datacollection.emd'):
         emd_file.attrs['version_major'] = ver_maj
         emd_file.attrs['version_minor'] = ver_min
         # Write user:
-        user_group = emd_file.require_group('user')
         for key, value in self.user.items():
-            user_group.attrs[key] = value
+            if not (value==''):
+                if 'user' not in emd_file:
```
to266: I believe this check is unnecessary, as the `require_group` does exactly this - checks if the group exists, and then either returns the reference if so, or creates it otherwise.

I could be wrong though

CodeMonkeyJan:
I agree that the empty groups create a lot of clutter which should be avoided. The reasoning behind implementing these groups at all was to encourage people to properly document their data (which also includes leaving empty strings as an "incentive" to fill in the blanks). We could of course drop these groups alltogether and leave the extent of documentation to the (hopefully thorough :-P) user.
This would make the code a lot cleaner (I personally also do not like to check for all of these "default"-groups one by one).
The alternative would be to just save what we have in the metadata.Signal and metadata.General dictionaries without creating clutter. During loading we could just iterate over all groups and add them to metadata.General (instead of explicitely searching for "user", "microscope", etc.).
The problem would be to identify the group in which the signals are stored. Currently it is done like this (which is also not a good way):

`# Extract signals:`
`node_list = list(emd_file.keys())`
`for key in ['user', 'microscope', 'sample', 'comments']:  # Nodes which are not the data!`
`|   if key in node_list:`
`|   |   node_list.pop(node_list.index(key))  # Pop all unwanted nodes!`
`# One node should remain, the data node (named 'data', 'signals', 'experiments', ...)!`
`assert len(node_list) == 1, 'Dataset location is ambiguous!'`
`data_group = emd_file.get(node_list[0])`

The alternative would be to either force the name to be something specific like "data" or "signals" (not recommended) or to look explicitly for the 'emd_group_type' attribute which should be 1 for all saved signals.
So all in all: Search for the group which contains groups with 'emd_group_type' attributes (this is our signal collection) and add all other groups to the metadata.
This should be the most flexible, unrestrictive approach.
What do you think we should do?

hyperspy/io_plugins/emd.py
CodeMonkeyJan: Yes, the require_group function should do exactly that.

magnunor:
@CodeMonkeyJan, thanks! I'll have a proper look at this after the 0.8.5 release :)

francisco-dlp:
@magnunor, did you have a chance to have a look at this?

magnunor:
@francisco-dlp, not yet. And I don't think I'll have time before the middle of August.

francisco-dlp:

PR #988, opened 2016-05-10
Conversation Between Z-Factors and Cross Section for EDX Quantification

k8macarthur:


k8macarthur:
So I made a pull request because this should be pretty much there. Although I'm wondering why the functions are 'not define' should they go somewhere else in the code other than eds_tem.py?

francisco-dlp:
Nice. Wouldn't `hyperspy/misc/eds/utils.py` be a better home for these functions? In order to make them available from `hyperspy.api.eds` you could then simply import them in `hyperspy/utils/eds.py` like this:

```python

from hyperspy.misc.eds.utils import cross_section_to_zeta, zeta_to_cross_section

```

hyperspy/_signals/eds_tem.py
```diff
@@ -641,3 +641,61 @@ def _get_dose(self, method, beam_current='auto', real_time='auto',
             return real_time * beam_current * 1e-9 / constants.e
         else:
             raise Exception('Method need to be \'zeta\' or \'cross_section\'.')
+
+def cross_section_to_zeta(cross_sections, elements):
+    """Convert a list of cross_sections in (b) to zeta-factors (kg/m^2).
```
francisco-dlp: maybe replace b with the more explicit barns?

hyperspy/_signals/eds_tem.py
```diff
@@ -641,3 +641,61 @@ def _get_dose(self, method, beam_current='auto', real_time='auto',
             return real_time * beam_current * 1e-9 / constants.e
         else:
             raise Exception('Method need to be \'zeta\' or \'cross_section\'.')
+
+def cross_section_to_zeta(cross_sections, elements):
```
francisco-dlp: What about a more explicit name for the function e.g. `edx_cross_sections_to_zeta_factors`?

hyperspy/_signals/eds_tem.py
```diff
@@ -641,3 +641,61 @@ def _get_dose(self, method, beam_current='auto', real_time='auto',
...
+    """Convert a list of cross_sections in (b) to zeta-factors (kg/m^2).
+
+    Parameters
+    ----------
+    cross_section: list of float
+        A list of cross sections.
```
francisco-dlp: I think that it is important to mention the units here too.

hyperspy/_signals/eds_tem.py
```diff
@@ -641,3 +641,61 @@ def _get_dose(self, method, beam_current='auto', real_time='auto',
...
+        zeta_factors.append(zeta)
+    return zeta_factors
+
+def zeta_to_cross_section(zfactors, elements):
+    """Convert a list of zeta-factors (kg/m^2) to cross_sections in (b).
+
+    Parameters
+    ----------
+    zfactors: list of float
+        A list of zeta-factors.
```
francisco-dlp: Units here too.

hyperspy/_signals/eds_tem.py
```diff
@@ -641,3 +641,61 @@ def _get_dose(self, method, beam_current='auto', real_time='auto',
...
+        [elements_db[element]['General_properties']['atomic_weight']
+            for element in elements])
+    zeta_factors = []
+    for i in range(len(elements)):
+        zeta = atomic_weights[i]/(cross_sections[i]**constants.Avogadro)*1E25
+        zeta_factors.append(zeta)
+    return zeta_factors
+
+def zeta_to_cross_section(zfactors, elements):
+    """Convert a list of zeta-factors (kg/m^2) to cross_sections in (b).
```
francisco-dlp: b -> barns?

hyperspy/misc/eds/utils.py
```diff
@@ -1,6 +1,7 @@
 
 import numpy as np
 import math
+from scipy import constants
```
francisco-dlp: Why importing constants here?

hyperspy/_signals/eds_tem.py
```diff
@@ -641,3 +641,61 @@ def _get_dose(self, method, beam_current='auto', real_time='auto',
...
+
+    """
+    if len(elements) != len(zfactors):
+        raise ValueError(
+            'The number of elements must match the number of cross sections.')
+    atomic_weights = np.array(
+        [elements_db[element]['General_properties']['atomic_weight']
+            for element in elements])
+    cross_sections = []
+    for i in range(len(elements)):
```
francisco-dlp: It would be better to use enumerate e.g. `for i, element in enumerate(elements):`

You could then move `atomic_weights` inside the loop e.g.:

```python

atomic_weight = elements_db[element]['General_properties']['atomic_weight']

```

Alternatively and possibly better, you could turn `zfactors` into an `array` and get rid of the loop e.g.

```

cross_sections = atomic_weights / (zfactors * constants.Avogadro) * 1E25

```

francisco-dlp:
Also, the style should comply with [PEP8](https://www.python.org/dev/peps/pep-0008/). Some editors such as Spyder (and many others) automatically check the code for PEP8 errors. [autopep8](https://pypi.python.org/pypi/autopep8) automatically fixes the style.

hyperspy/misc/eds/utils.py
k8macarthur: I use avogadro's for the conversions. Probably wasn't needed but now is because the functions have been move to utils.py.

k8macarthur:
Travis still seems to be failing nose tests:

  File "/home/travis/miniconda/envs/testenv/lib/python3.5/site-packages/nose/case.py", line 198, in runTest
    self.test(*self.arg)
  File "/home/travis/build/hyperspy/hyperspy/hyperspy/tests/model/test_fancy_indexing.py", line 90, in test_model_navigation_indexer_slice
    nt.assert_false(m[0]._active_array[0, 0])
AssertionError: True is not false

francisco-dlp:
That test fails randomly at present. You can safely ignore it.

thomasaarholt:
Is there anything remaining to check here then?

k8macarthur:
I've update the 'for' loop to enumerate as @francisco-dlp suggested. I think that now covers everything?

francisco-dlp:
It would be great if you could also address the PEP8 styling issues. From above:

> Also, the style should comply with [PEP8](https://www.python.org/dev/peps/pep-0008/). Some editors such as Spyder (and many others) automatically check the code for PEP8 errors. [autopep8](https://pypi.python.org/pypi/autopep8) automatically fixes the style.

k8macarthur:
According to the plugin in my Atom software it should all be up to scratch with PEP8 styling. @francisco-dlp have you spotted anything that the software has missed?

hyperspy/misc/eds/utils.py
```diff
@@ -521,7 +527,65 @@ def quantification_cross_section(intensities,
...
+
+    """
+    if len(elements) != len(zfactors):
+        raise ValueError(
+            'The number of elements must match the number of cross sections.')
+    cross_sections = []
+    for i, element in enumerate(elements):
+        atomic_weight = elements_db[element]['General_properties'][
+            'atomic_weight']
+        xsec = atomic_weight/(zfactors[i]*constants.Avogadro*1E-25)
```
francisco-dlp: Example of PEP8 issue: missing spaces around arithmetic operators.

francisco-dlp:
I'll merge this. @k8macarthur, it would be nice if you could edit the first comment in order to describe what you've implemented in this pull request.

k8macarthur:
Great thanks! I've just updated the comment. :)

Issue #990, opened 2016-05-11
TEMParametersUI Error


k8macarthur:
So I think this might be largely an error from the quantification pull request last month, but I seem to be having a real problem with loading a data set and setting all the parameters for quantification with out the UI popping up and complaining at me. Has anyone else come across this in the master? I'm trying to understand exactly where the problem is before I start trying to fix it. My error code is below:
```python
#Adjust the metadata for all the microscope parameters.

s.set_microscope_parameters(beam_energy = 200)
s.set_microscope_parameters(real_time = 3.5) #dwell time per pixel
s.set_microscope_parameters(live_time = 1)
s.set_microscope_parameters(beam_current = 0.05) #in pA
s.axes_manager[-1].offset=0.05
theta = 0.5

#Or open a GUI using s.set_microscope_parameters(). or hs.preferences.gui()
#s.metadata

---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
<ipython-input-3-9e000b37a22d> in <module>()
      2 
      3 s.set_microscope_parameters(beam_energy = 200)
----> 4 s.set_microscope_parameters(real_time = 3.5) #dwell time per pixel
      5 s.set_microscope_parameters(live_time = 1)
      6 s.set_microscope_parameters(beam_current = 0.05) #in pA

/Users/macark/Documents/hyperspy/hyperspy/_signals/eds_tem.py in set_microscope_parameters(self, beam_energy, live_time, tilt_stage, azimuth_angle, elevation_angle, energy_resolution_MnKa, beam_current, probe_area, real_time)
    157         if set([beam_energy, live_time, tilt_stage, azimuth_angle,
    158                 elevation_angle, energy_resolution_MnKa]) == {None}:
--> 159             self._are_microscope_parameters_missing()
    160 
    161     @only_interactive

/Users/macark/Documents/hyperspy/hyperspy/_signals/eds_tem.py in _are_microscope_parameters_missing(self)
    227                 is_ok = messagesui.information(par_str)
    228                 if is_ok:
--> 229                     self._set_microscope_parameters()
    230                 else:
    231                     return True

/Users/macark/Documents/hyperspy/hyperspy/decorators.py in wrapper(*args, **kwargs)
     54     def wrapper(*args, **kwargs):
     55         if preferences.General.interactive is True:
---> 56             return cm(*args, **kwargs)
     57         else:
     58             raise NoInteractiveError

[...]

PR #991, opened 2016-05-12
Fix to TEMParametersUI

k8macarthur:
Fix to update the TEMParameters UI to include the all parameters needed for EDX quantification.


k8macarthur:
I think I've fixed this but for some reason the order to applying the see #990

Issue #916, opened 2016-03-20
nosetest3 TestOutArg.test_indexmax failing


magnunor:
Running nosetest3 on the current master in Ubuntu 15.10, I get the following error: 
```python
======================================================================
ERROR: hyperspy.tests.signal.test_tools.TestOutArg.test_indexmax
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/usr/lib/python3/dist-packages/nose/case.py", line 198, in runTest
    self.test(*self.arg)
  File "hyperspy/dev/hyperspy/hyperspy/hyperspy/tests/signal/test_tools.py", line 491, in test_indexmax
    self._run_single(self.s.indexmax, self.s, dict(axis=0))
  File "hyperspy/dev/hyperspy/hyperspy/hyperspy/tests/signal/test_tools.py", line 444, in _run_single
    r = f(out=s1, **kwargs)
  File "hyperspy/dev/hyperspy/hyperspy/hyperspy/signal.py", line 4405, in indexmax
    out=out)
  File "hyperspy/dev/hyperspy/hyperspy/hyperspy/signal.py", line 4010, in _apply_function_on_data_and_remove_axis
    function(self.data, axis=ar_axes, out=out.data,)
TypeError: argmax() got an unexpected keyword argument 'out'
```

Which is a bit odd, since running `ndarray.argmax(out=...)` in an ipython3 terminal works fine.

vidartf:
I've had this issue as well. It's related to the numpy version. Does `np.argmax(out=...)` work? If I remember correctly, this requires numpy version 1.10.

magnunor:
You're right: doing ndarray.argmax(out=...) works, but np.argmax(an_array, out=...) does not.

To replicate:
```python
import numpy as np
a = np.array([0])
b = np.arange(10)
b.argmax(out=a) # Works fine
np.argmax(b, out=a) # Does not work
np.__version__ # output: '1.8.2'
```

magnunor:
No longer a problem in the current Ubuntu version (16.04), due to it using a newer version of Numpy. Close this?

francisco-dlp:
This is actually a bug that #910 fixes. Let's wait until #910 is merged.

francisco-dlp:
#910

PR #999, opened 2016-05-18
add h5py dependency and numpy version >1.10 in setup.py

### LINKED ISSUE #910: Master branch needs h5py as a dependency?
On the current master branch, hyperspy needs h5py as a dependency, since h5py is imported in [signal.py](https://github.com/hyperspy/hyperspy/blob/master/hyperspy/signal.py#L20). What would it be better, set h5py as a dependency or change the code in [signal.py](https://github.com/hyperspy/hyperspy/blob/master/hyperspy/signal.py#L3052)?

This bug doesn't affect the 0.8.3 and 0.8.4 releases.

ericpre:
- 'out' argument in argmax needs numpy>1.10. Spotted it running the tests on my computer with numpy 1.9!
- signal.py needs h5py.

Fixes  #910 and #916.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -45,13 +45,14 @@ if os.path.exists('build'):
 install_req = ['scipy',
                'ipython>=2.0',
                'matplotlib>=1.2',
-               'numpy',
+               'numpy>=1.10',
                'traits>=4.5.0',
                'traitsui>=5.0',
                'natsort',
                'requests',
                'setuptools',
-               'sympy']
+               'sympy',
+               'h5py']
 
 
 class update_version_when_dev:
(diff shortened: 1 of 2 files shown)


magnunor:
I noticed the lacking h5py requirement as well, but hadn't gotten around to submitting an Issue about it yet. So very nice!

francisco-dlp:
`stdeb.cfg` needs updating too. Otherwise, this is ready to merge.

ericpre:
I updated stdeb.cfg but I don't have debian, so I can't check if it's working or not.

ericpre:
I suspect this: "XS-Python-Version: >=2.7" needs an update. However, since I don't know how this file works/is used, I don't feel like touching it further! ;)

francisco-dlp:
Thanks!

PR #1007, opened 2016-05-23
model sliders for 0.8.x

to266:
Just a quick and simple interactive widgets for a model.

# Has to be run in a notebook!

Example
```python
%matplotlib qt
import hyperspy.api as hs
import numpy as np

s = hs.signals.EELSSpectrum(np.random.rand(500))

s.set_microscope_parameters(300, 13, 14)

m = s.create_model()

m.append(hs.model.components.Gaussian())
m.append(hs.model.components.Polynomial(5))
m.append(hs.model.components.EELSCLEdge('C_K'))
m.append(hs.model.components.ScalableFixedPattern(s.deepcopy()))

m.plot(True)
m.notebook_interaction()
```

The boundaries for the widgets are either looked up from `bmin`, `bmax` of the parameter, or just `current_value + 10`.. The exception is `_position` parameter, which instead defaults to the last signal axis limits.

parameters being free/fixed is ignored.

`EELSCLEdge` and `ScalableFixedPattern` have special methods, as not only `active` + parameters are usually required (or, as is the case for the EELS edge, one of the parameters is not actually desirable to be shown)


francisco-dlp:
Very cool! And the code is so simple that it looks like magic.

Wouldn't it be possible to split the code so that each bit goes to where it belongs? For example, shouldn't the code to generate the `Parameter` widget live in the `Parameter` class? In this way `Parameter`, `Component` and `Model` could all have a `notebook_interaction` method. Also, what about making it the `__repr__` method of those objects when in the notebook, see [Integrating your objects with IPython](http://ipython.readthedocs.io/en/stable/config/integrating.html?highlight=repr_html#integrating-your-objects-with-ipython).

to266:
Sure, I could split it and add the `_repr_*_` methods.

The only thing I'm not really happy about is the requirement to get `min` and `max` for the sliders. For most values if `bmin` and `bmax` are not defined, it's just quesswork. We could add the `bmin` and `bmax` widgets too, but then it might quickly get too complicated.

Should I PR this to the master instead? I guess it's simple enough (and does not break anything) that it could go to the nearest release if people find it useful

francisco-dlp:
Yes, the bounds issues is annoying and probably we shouldn't release this without finding a solution.

What about something like the following?

```python
import ipywidgets as widgets

value = widgets.FloatText(description="value")
step = widgets.FloatText(description="step")
step.value = 1
def on_click_decrease(*args, **kwargs):
    value.value -= step.value
def on_click_increase(*args, **kwargs):
    value.value += step.value
decrease_button = widgets.Button(description="-")
decrease_button.on_click(on_click_decrease)
increase_button = widgets.Button(description="+")
increase_button.on_click(on_click_increase)
widgets.HBox([value, decrease_button, increase_button, step])

```

A spinner that responds to the scroll button and which step can be customised would probably be better. There is an example [here](http://ipywidgets.readthedocs.io/en/latest/examples/Custom%20Widget%20-%20Hello%20World.html).

to266:
It seems that with two boxes (one on the left for `min` and one on the right for `max`) the buttons are not necessary. You can manually enter the boundary values to use with the slider, as well as manually enter the actual slider value.

The only slightly bad thing about this PR is that it relies on `ipywidgets > 5.0`, which is not on conda yet (but has been on pip for a while)

to266:
Oh and another thing - the boundaries of the sliders get initialised as the boundaries of the parameters, but it intentionally does not work the other way around - `bmin` and `bmax` are left unchanged

to266:
and I forgot the `_repr_*_` methods. Will add them once we're settled on the design, I guess

to266:
Also, this https://github.com/ipython/ipywidgets/issues/259 is still around - so very small values do not work as well. We could include a multiplication ourselves, but I think it's better to just wait for the upstream

to266:
Updated

dnjohnstone:
In a couple of simple tests this seems fine to me...

francisco-dlp:
This needs to be documented in the User Guide and docstrings. Also tested by simply running the commands to verify that it doesn't raise an exception.

Given that this is only useful in jupyter notebook, wouldn't it be sensible to make the ipywidgets optional instead of required by lazy-importing it?

to266:
Added. As always, the user guide entries from my side are somewhat lacking, so suggestions / corrections are welcome

francisco-dlp:
Looks really good. I think that the only thing missing is adding the optional dependency to the install section of the User Guide. Also a warning box in the `notebook_interaction` sections of the User Guide explaining that it requires the optional ipywidgets dependency and the same thing for the methods docstrings.

to266:
Added

PR #1008, opened 2016-05-23
Mike's branch

mwalls:
A test to see if I can do this. Just a few suggestions to make this part clearer and correct a few typos.

### DIFF
--- a/doc/user_guide/install.rst
+++ b/doc/user_guide/install.rst
@@ -44,7 +44,7 @@ Quick instructions to install HyperSpy using Anaconda (Linux, MacOs, Windows)
    `Anaconda. <https://store.continuum.io/cshop/anaconda/>`_
    For MacOS, be sure to choose the graphical installer, not the command-line version. Anaconda is recommended for the best performance (it is compiled
    using Intel MKL libraries) and the easiest intallation (all the required
-   libraries are included). The academic license is free. Then install Hyperspy via the 'pip' command in the Mac Terminal  (usually to be found in "Applications>Utilities"):
+   libraries are included). The academic license is free. Then install Hyperspy via the 'pip' command (in the Mac Terminal  usually to be found in "Applications>Utilities" or in a unix terminal):
 
    .. code-block:: bash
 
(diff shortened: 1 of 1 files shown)


dnjohnstone:
@mwalls A successful test it seems to me.. I would merge this - all good improvements.

doc/user_guide/install.rst
```diff
@@ -42,9 +42,9 @@ Quick instructions to install HyperSpy using Anaconda (Linux, MacOs, Windows)
 
 #. Download and install
    `Anaconda. <https://store.continuum.io/cshop/anaconda/>`_
-   Anaconda is recommended for the best performance (it is compiled
+   Be sure to choose the graphical installer, not the command-line version. Anaconda is recommended for the best performance (it is compiled
```
francisco-dlp: This may be confusing as there is no graphical installer for Linux.

francisco-dlp:
Except for my little remark above, it looks very good. Thanks!

mwalls:
Apologies in advance for a rather long comment. 
I've taken on board Francisco's remark. I've also indicated that "start_jupyter" is not currently available for MacOS. Even though on their webpage it says "for any platform", when you try to run "jupyter_context-menu_add" you get a message: "Nothing done. Currently only Gnome and Windows are supported". 
I've changed the Known issue about uninstalling "Start Hyperspy Here" since the name has changed, but maybe this indicates that it's an old issue that is no longer pertinent?

I am doing this stuff from the perspective of a user (in fact several users) who are trying to get going and encountering some difficulties. Many of us are unfamiliar with some of the terminology used and do not have the general level of assumed knowledge about Python (and coding in general) in the docs. I will try and make more suggestions for increased clarity over the next few weeks,  One example is the inclusion of the "Creating Conda environment for HyperSpy" section. Since this is not listed at the top amongst the various ways of installing Hyperspy, it is not obvious to the newcomer whether it's another installation method, an extra step that everyone has to do, an optional extra, or what? (As I write I don't know.).  Could the separate installation methods be numbered perhaps? 

Anyway I will try not to make a habit of adding such long comments!

mwalls:
Another point is the use of the term "Windows terminal",  which 4/4 users here assumed to mean the command line prompt. I'm guessing this is not the case, and that you should use something like Windows Powershell, although that doesn't recognise the 'pip' command either on my machine, despite everything being installed in principle...

tjof2:
@mwalls or Windows terminal could mean the WinPython(?) executable in the HyperSpy install folder - you can see how confusion arises!

francisco-dlp:
@mwalls, I'm glad that 4/4 of the users there have assumed that "Windows terminal" means command line prompt because that's what we intended to mean ;) @tjof2 is right in that it could also be misunderstood as the WinPython command prompt, but this part of the installation instruction is not supposed to apply to installing the bundle.

Regarding the changes in the branch, it looks good to me, except for changes in the issue regarding the "Start Hyperspy Here" entries. This is no longer an issue since `v0.8.3`, but people upgrading from older versions may appreciate that we still explain them how to get rid of the context menu entries.

PR #1015, opened 2016-05-25
ENH model.signal attribute

dnjohnstone:
As discussed in #963 this PR removes the separated model1D.spectrum and model2D.image attributes in favour of model1D.signal and model2D.signal in both cases. Just cleans things up a bit as this separation was anyway superfluous in the code and will make consistent separation of signal1D/2D easier to merge.

### DIFF
--- a/hyperspy/_components/scalable_fixed_pattern.py
+++ b/hyperspy/_components/scalable_fixed_pattern.py
@@ -68,7 +68,7 @@ class ScalableFixedPattern(Component):
 
         self._position = self.shift
         self._whitelist['spectrum'] = ('init,sig', spectrum)
-        self.spectrum = spectrum
+        self.signal = spectrum
         self.yscale.free = True
         self.yscale.value = 1.
         self.xscale.value = 1.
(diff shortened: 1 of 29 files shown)


dnjohnstone:
There are a couple of slips in this that I just noticed with the comparison available in the PR... will deal with them and let you know before review.

dnjohnstone:
Ok - done I think.

francisco-dlp:
`spectrum` was renamed to `signal` in places unrelated to this PR. I have spotted this issue mainly in the some test files. As far as I can tell, the issue is harmless, but it'll be good to fix it before merging anyway.

dnjohnstone:
Probably we should also deal with #963 first...

dnjohnstone:
Those tests are now corrected though...

to266:
FYI the "unintentional change" that @dnjohnstone afterwards reversed was probably intentional - as it is currently (and was merged), it does not make sense

In particular, the sequence is 
```python
# ...
self.self.free_parameter_boundaries = None
# ...
```
[here](https://github.com/hyperspy/hyperspy/blob/master/hyperspy/model.py#L840) and [here](https://github.com/hyperspy/hyperspy/blob/master/hyperspy/model.py#L853)

dnjohnstone:
@to266 that was a strange one - I reversed it because it wasn't part of what I wanted to touch, but I think it was in the master before... I agree it looks wrong.

francisco-dlp:
Well spotted, that's a separate issue, who of you will PR a fix before we forget about it?

to266:
https://github.com/hyperspy/hyperspy/pull/1016

Issue #976, opened 2016-05-04
s.swap_axes not switch between axis-types


magnunor:
Using `s.swap_axes` to change a signal axis to a navigation axis does not work:
```Python
s = hs.signals.Image(np.random.random(size=(10,10,100)))
<Image, title: , dimensions: (10|100, 10)>
s.swap_axes(0,1)
<Image, title: , dimensions: (10|10, 100)>
s.swap_axes(0,2)
<Image, title: , dimensions: (10|100, 10)>
```
Basically none of the `swap_axes` permutations actually switches the 100 axis to the navigational axis.

Same issue seems to affect `s.rollaxis`.

francisco-dlp:
`plot_images` is broken:

```python

---------------------------------------------------------------------------
ImportError                               Traceback (most recent call last)
<ipython-input-4-44c5a631d171> in <module>()
     12 
     13 image.metadata.General.title = 'multi-dimensional Lena'
---> 14 hs.plot.plot_images(image, tight_layout=True)

/home/fjd29/Python/hyperspy3/hyperspy/drawing/utils.py in plot_images(images, cmap, no_nans, per_row, label, labelwrap, suptitle, suptitle_fontsize, colorbar, centre_colormap, saturated_pixels, scalebar, scalebar_color, axes_decor, padding, tight_layout, aspect, min_asp, namefrac_thresh, fig, *args, **kwargs)
    561     from hyperspy.drawing.widgets import Scale_Bar
    562     from hyperspy.misc import rgb_tools
--> 563     from hyperspy.signal import Signal
    564 
    565     if isinstance(images, Signal) and len(images) is 1:

ImportError: cannot import name 'Signal'

```

Knowing that the plotting features are not tested, I just try plotting things to see if it works...

francisco-dlp:
Sorry, I've replied to the wrong issue. I'll delete this two comments.

to266:
The required fix is just one line, but I can't PR it right now, as github seems to be lagging - I can push to the branch, but it is not shown on the web, and hence I can't PR.....

francisco-dlp:
#1019

PR #1019, opened 2016-05-25
BUGFIX swap axes now is able to swap navigation status as well

to266:
Fixes https://github.com/hyperspy/hyperspy/issues/976


francisco-dlp:
It would be nice to add a little test for `swap_axes` here.

hyperspy/tests/signal/test_1D_tools.py
```diff
@@ -81,22 +82,6 @@ def test_align(self):
         nt.assert_true((s.data[:, -1] == 2).all())
         nt.assert_true((s.data[:, 0] == 2).all())
         # Check that the calibration is correct
-        nt.assert_equal(s.axes_manager._axes[1].offset, self.new_offset)
```
francisco-dlp: why removing lines 84 and 85?

to266: that's just git / github trying to lump changes. The lines are still there in the new file, but this diff shows them as being after the removed lines.... (in the old files these lines are 100-102)

Issue #995, opened 2016-05-14
s.integrate_in_range broken in 0.8.x


magnunor:
Seems to be broken in 0.8.x, but works with the current master (unless I'm messing something up).

```Python
hyperspy/gui/tools.py in span_selector_switch(self, on)
    169                     onselect=self.update_span_selector_traits,
    170                     onmove_callback=self.update_span_selector_traits,
--> 171                     left_limit=self.axis.low_value,
    172                     right_limit=self.axis.high_value)
    173 

AttributeError: 'IntegrateArea' object has no attribute 'axis'
```

to266:
Actually this is broken in master, as without the limits you can drag the selector outside the range, and then it crashes with the usually horrible traits errors.

No point in PR'ing to master separately though, as it should be merge with 0.8.x, I guess..

francisco-dlp:
#1021

PR #1021, opened 2016-05-25
added first signal axis to the selector object

to266:
Fixes https://github.com/hyperspy/hyperspy/issues/995


dnjohnstone:
This seems to fix the bug for me, but whilst this is attracting attention -- it seems to me that this method won't work for 2D signals, although the likely desired functionality of integrating over an area in an image is achieved easily with ROIs is there any way we can make the usage more consistent?

to266:
This particular method is made to be used with spectra explicitly and, as you pointed out, a ROI should be used for image integration.

Issue #997, opened 2016-05-16
Stored model cannot be restored


jat255:
Pardon if this is an error on my part, since this is actually the first time I've used the new model classes, but I think I may have found a bug:

I ran the following code:
```python
s8 = hs.load('SI8_B-K_horizontal_long_EELS_spikeRem.hdf5')
s8.add_elements(['B'])
s8.decomposition(True)
s8 = s8.get_decomposition_model(10)

m8 = s8.create_model()

m8.store('B-K_w_fine_structure')

s8.save('test_restore_model.hdf5')

tmp = hs.load('test_restore_model.hdf5')
```

Everything up to here works as expected:
```python
>>> tmp.models
    └── BK_w_fine_structure
    ├── components
    │   ├── B_K
    │   └── PowerLaw
    ├── date = 2016-05-16 10:04:32
    └── dimensions = (31, 9|2048)
```

However, when I go to restore the model, I'm receiving an error:
```python
>>> m = tmp.models.restore('BK_w_fine_structure')
---------------------------------------------------------------------------
EOFError                                  Traceback (most recent call last)
<ipython-input-349-668408e20346> in <module>()
----> 1 tmp.models.restore('BK_w_fine_structure')

/home/josh/git_repos/hyperspy/hyperspy/signal.py in restore(self, name)
    253         name = self._check_name(name, True)
    254         d = self._models.get_item(name + '._dict').as_dictionary()
--> 255         return self._signal.create_model(dictionary=copy.deepcopy(d))
    256 
    257     def __repr__(self):

/home/josh/git_repos/hyperspy/hyperspy/_signals/eels.py in create_model(self, ll, auto_background, auto_add_edges, GOS, dictionary)
   1254                           auto_add_edges=auto_add_edges,
   1255                           GOS=GOS,
-> 1256                           dictionary=dictionary)
   1257         return model
   1258 

/home/josh/git_repos/hyperspy/hyperspy/models/eelsmodel.py in __init__(self, spectrum, auto_background, auto_add_edges, ll, GOS, dictionary)
     84             auto_background = False
     85             auto_add_edges = False
---> 86             self._load_dictionary(dictionary)
     87 
     88         if auto_background is True:

/home/josh/git_repos/hyperspy/hyperspy/model.py in _load_dictionary(self, dic)
    255 
    256                 self.append(getattr(components, comp['_id_name'])(**init_args))
--> 257                 id_dict.update(self[-1]._load_dictionary(comp))
[...]

PR #1022, opened 2016-05-25
Drop marshalling functions

to266:
Fixes https://github.com/hyperspy/hyperspy/issues/997

(relatively speaking)


francisco-dlp:
`python3-dill` needs to be added to `stdeb.cfg`.

I must say that this stdeb business is getting annoying. We should simply ask for proper Debian packaging...

to266:
Done

Issue #978, opened 2016-05-04
HDF5 io: pass compression options to file writer


magnunor:
Currently HyperSpy uses default compression values when saving HDF5-files.

It would be nice if the `s.save` function could pass arguments to the HDF5-python writer. For example, `s.save("test.hdf5", compression_opts=9)` to get the max amount of compression.

francisco-dlp:
#1027

PR #1027, opened 2016-05-25
allow passing more keywords to hdf5 dataset creation

to266:
https://github.com/hyperspy/hyperspy/issues/978

```python

In [2]: import numpy as np

In [3]: s = hs.signals.Spectrum(np.ones((1000,1000,100)))

In [4]: s.save('smaller.hdf5', compression_opts=9)
The smaller.hdf5 file was created

In [5]: s.save('bigger.hdf5', compression_opts=0) 
The bigger.hdf5 file was created

In [6]: !du -h {smaller,bigger}.hdf5
2.0M	smaller.hdf5
828M	bigger.hdf5

```


francisco-dlp:
This requires a test.

to266:
Test added

hyperspy/tests/io/test_hdf5.py
```diff
@@ -211,3 +212,19 @@ def test_rgba16():
...
+
+    def setUp(self):
+        self.filename = 'testfile.hdf5'
+        BaseSignal([1, 2, 3]).save(self.filename, compression_opts=8)
+
+    def test_compression_opts(self):
+        f = h5py.File(self.filename)
+        d = f['Experiments/__unnamed__/data']
+        nt.assert_equal(d.compression_opts, 8)
+        nt.assert_equal(d.compression, 'gzip')
```
francisco-dlp: shouldn't we call `f.close` at this point?

to266: does not matter, as `h5py` closes all the files automatically upon exit. In fact I spent a substantial number of hours trying to disable it, but could not.... (not for this branch though)

francisco-dlp: Upon exiting what?

From the h5py docs: "[the file] should be closed when they are no longer in use". I think that it'll be safer to close it.

to266: What I meant is with GC. Any way, added :)

Issue #1026, opened 2016-05-25
The `EDSTEM` signal overwrite `tilt-stage` in metadata


francisco-dlp:
The `tilt-stage` metadata read from a file is overwritten by the `_set_default_param` method due to a typo. It is therefore always set to 0.

francisco-dlp:
#1028

Issue #1025, opened 2016-05-25
Duplicate metadata in the ripple format


francisco-dlp:
HyperSpy extends the Ripple format with some extra keywords. One of them, EDS `energy-resolution` is actually a duplicate of the standard `detector-peak-width-ev`.

francisco-dlp:
#1028

PR #1028, opened 2016-05-25
Fix ripple

### LINKED ISSUE #981: Loading ripple files raises `TypeError`
The following reproduces the bug:

```python

s = hs.signals.Spectrum(np.ones((32,32), dtype="float"))
s.save("test.rpl")
s = hs.load('test.rpl')

```

### LINKED ISSUE #1024: Missing EELS metadata when saving in ripple format
The convergence_angles, collection_angle and beam_energy are not saved to the file.

francisco-dlp:
Fixes #1024, #1025 and #1026 and (re)fixes #981 using a different approach.

It also adds tests for the Ripple format.


hyperspy/tests/io/test_ripple.py
```diff
@@ -0,0 +1,131 @@
...
+            s = EDSTEMSpectrum(data)
+            s.set_microscope_parameters(
+                beam_energy=100.,
+                live_time=1.,
+                tilt_stage=2.,
+                azimuth_angle=3.,
+                elevation_angle=4.,
+                energy_resolution_MnKa=5.)
+    else:
+        s = Signal(data)
```
to266: `BaseSignal` ?

hyperspy/tests/io/test_ripple.py
```diff
@@ -0,0 +1,131 @@
...
+        s.axes_manager.set_signal_dimension(dim)
+    for i, axis in enumerate(s.axes_manager._axes):
+        axis.offset = i * 0.5
```
to266: I would add 1 to the `i`, since the first one is 0..

to266: Though in this case it does not matter, since no real calculations are done. Up to you then, would merge either way once `BaseSignal` is added

hyperspy/tests/io/test_ripple.py
```diff
@@ -0,0 +1,131 @@
+import tempfile
+import os.path
+import shutil
+
+import numpy as np
+import nose.tools as nt
+import numpy.testing as npt
```
to266: `npt` not used (the line is commented out)

hyperspy/tests/io/test_ripple.py
```diff
@@ -0,0 +1,131 @@
...
+
+
+def test_ripple():
+    try:
+        for dtype in ripple.dtype2keys.keys():
+            for shape, dims in SHAPES_SDIM:
+                for dim in dims:
+                    yield _run_test, dtype, shape, dim
+    finally:
+        shutil.rmtree(TMPDIR)
```
to266: Wouldn't `TMPDIR.cleanup()` be better, [as per the docs](https://docs.python.org/3.5/library/tempfile.html#tempfile.TemporaryDirectory) ?

to266: probably `TMPDIR` is just the path, and the object is required, but you get the idea. Should be cross-platform as well

francisco-dlp:
Thanks! Ready for review again.

PR #1029, opened 2016-05-25
correctly computes the gradient for polynomial component

to266:
I used to calculate df/dx, while it actually requires df/da, df/db, ....


to266:
`jacobian` for the simplest one, I think...

Which is just stacked derivatives. I guess we left the burden of writing properly formatted derivatives to the authors of the components, to that may be the __actual__ thing that needs to be done. The jacobian can then in principle be shared for the models

francisco-dlp:
This requires a little test.

to266:
Tests added

PR #1030, opened 2016-05-26
HyperSpy 0.9 --> HyperSpy 1.0

dnjohnstone:
Just renaming in deprecation warnings for consistency


mwalls:
in the learn section there's a typo: "From them on" instead of "From then on"

mwalls:
Line 501...

PR #1036, opened 2016-05-27
Fix L3 Hartree-Slater EELS XS of Al, Ar, Cl, Na, P and S 

### LINKED ISSUE #1035: EELS hartree-slater L2 XS missing for Al, Ar, Cl, Na, P, S
The `EELSCLEdge` component does not include the L2 edges of Al, Ar, Cl, Na, P, S for the Hartree-Slater cross sections. For these elements the L3 and L2 onset energies are very similar. Not including the L2 ionisation edge results in a wrong model that overestimates the amount of these elements.

francisco-dlp:
Fixes #1035 by multiplying the L3 Hartree-Slater cross section by 1.5 to account for the overalapping L2 ionisation edge.


francisco-dlp:
Due to its simplicity, I'll break the rule for once and merge this without review.

vidartf:
What's the factor of 1.5 based on?

francisco-dlp:
The L2 XS is equal to half the L3 XS. Because the L3 and L2 roughly overlap for these edges, adding the intensity of the L2 XS to the L3 should fix the issue.

vidartf:
Sure, I got the addition, I just wondered if the 0.5 factor was taken from a reference, or if this is simply common knowledge? HyperSpy has gotten a lot better at documenting its magic numbers, so just thought I would check :)

francisco-dlp:
You are right, we should document what that factor is. Before it was computed when creating the `EELSCLEdge` component and it was well documented but, for some reason that I can't remember, we moved it to a more public place, the elements database, and in the way it lost its documentation.

Issue #787, opened 2015-12-09
New Signal Classes Proposal & Suggestions


dnjohnstone:
We've often mentioned the fact that HyperSpy could be used by many people outside of the Electron Microscopy community however at this point there only exists specific Signal classes for EM related signals... In line with discussions regarding steps to encourage new users/developers in #782 I was wondering if we might consider making some new signal classes with basic specific functionality? 

The idea being that (however silly this sounds) making the place to write your functions clear might be one way to reduce the contribution barrier. This issue could also serve as a place for users who might be keeping quiet to suggest signals.

Obvious signals that spring to mind are:
1. RamanSpectrum - most functionality is already there could add a specific correction for the jumps that are seen when slits change over as is common on many spectrometers.
2. PowderXRD - specifically thinking the time series data that people get from in-situ experiments

The list goes on...

magnunor:
Gas chromatography: I've actually written some code for loading datasets, and the modelling part of HyperSpy is very useful since the proprietary software which is bundled with the machines themselves are very limited in what kind of processing you can do. 

I (or my collaborator using it) might actually properly implement it into HyperSpy at some point.

CodeMonkeyJan:
A few people in our institute use HyperSpy to work with holography and phase images (loaded with the plugin added in #741).
I was thinking about adding signals which are subclasses of the Image class and could provide basic functionality like phase wrapping/unwrappig or phase reconstruction from a complex holography image.
As @dnjohnstone mentioned, even having barebone classes would enable people to easily identify where they could add more sophisticated code.

dnjohnstone:
Hi @CodeMonkeyJan yes that would be an ideal case to create a new signal class - I'm also about to pull request an electron diffraction class building on the Image class - are you likely to work on a Holography class yourself?

CodeMonkeyJan:
I have a very crude outline for those two mentioned classes which I wanted to finish after another plugin (hopefully in the next few days). I'll try to encourage my colleagues to fill them with meaningful methods afterwards.

dnjohnstone:
Ok good stuff - just so you know the diffraction thing I mention and my recent push on EDS is largely for work with Martial (who I think you know quite well)....

On a related note that was why I was interested in #710 as I've got some EDS data that I took with you guys that I can't open properly - don't suppose any of you have made any further progress with that? I'm not getting super far.

jat255:
[...]

PR #1043, opened 2016-05-30
Add support for writing/reading scale and unit to tif files to be read with ImageJ or DigitalMicrograph

ericpre:
This PR aims at improving the tif compatibility between Hyperspy and ImageJ/DigitalMicrograph by writing/reading the scale and the unit.

- update local tifffile.py library (necessary to allow writing custom tags required by DM)
- add option to the tif ioplugin to use local tifffile.py when necessary (in case the skimage tifffile.py library is too old)
- add tests
- fix a bug that occurs when reading a tif files using newer version of tifffile.py (API change)

- [x] improve the unit reading (still a bug with "um" unit...)
- [x] ~~add support for reading scale and unit of OME-TIFF~~ (see below)
- [x] add support for reading the scale of SEM images
- [x] get example files of different software (FEI, ZEISS and other)
- [x] same for light microscope images?

I updated the tifffile.py file using the @magnunor suggestion in #787 regarding credit to authors. I hope this was the proper way to do it?

There are still a few details to polish, but if you could already comment/start review it, that would be great! Thanks!


hyperspy/io_plugins/tiff.py
```diff
@@ -92,45 +119,142 @@ def file_writer(filename, signal, **kwds):
...
+    if signal.axes_manager.navigation_dimension > 0:
+        extratags.extend([(65005, 's', 3, units[2], False), # z unit
+                          (65008, 'd', 1, 3.0, False), # z origin in pixel
+                          (65011, 'd', 1, float(scales[2]), False), # z scale
+                          (65014, 's', 3, units[2], False),  # z unit
+                          (65017, 'i', 1, 1, False)])
+    return extratags
+
+def _get_scale_unit(signal):
+    """ Return a list of scales and units, the length of the list is egal to
```
magnunor: typo: egal to equal

hyperspy/io_plugins/tiff.py
```diff
@@ -18,19 +18,19 @@
...
-    from skimage.external.tifffile import imsave, TiffFile
-except ImportError:
-    with warnings.catch_warnings():
-        warnings.simplefilter("ignore")
-        from hyperspy.external.tifffile import imsave, TiffFile
-    warnings.warn(
-        "Failed to import the optional scikit image package. "
-        "Loading of some compressed images will be slow.\n")
-
+#try:
```
magnunor: I'm guessing these lines can be deleted instead of being commented out?

magnunor:
Very nice! Having good support for properly saving and loading formats (including metadata), especially widely used ones such as tiff is very important.

I had a quick look through it, but someone else should have a look at the functional parts.

tjof2:
Is this consistent with e.g. OME-TIFF specification, as discussed in #839?

ericpre:
@tjof2: good point, no, this is not consistent with OME-TIFF! Do you have a use case, that I could use to start working on it?
I guess OME-TIFF could offer better support for storing the "z" scale and unit for multipage tif?
For now, it only uses the default x and y resolution tags for the scale and the imagedescription tag for the unit in the same way that ImageJ does it.

tjof2:
@ericpre I'm afraid I don't have an example, I just recall the discussion  and thought it relevant! @francisco-dlp  any thoughts on this / #839 ?

ericpre:
Actually, I remember that some TEM camera providers save images using tiff files, at least a few years back, the software shipped with Tietz camera used to do that. I don't know it it's still the case, thought. I have some files for this case, so I will add support for the scale and units (at least).

Does anybody have similar cases? From SEM or optical?

I have just checked some SEM images and some of them are using OME-metadata, so I will work on that.

Here, a link about how ImageJ can read the scale and unit of SEM images:
http://imagej.1557.x6.nabble.com/Importing-SEM-images-with-scale-td3689900.html

ericpre:
I fixed some bugs but I still need to add more test files.
Regarding OME-TIFF metadata, I think, this is out of the scope of this PR because ImageJ doesn't support it natively, actually it uses the bio-format plugin to read OME-TIFF files. However, it could be implemented using other python library.

francisco-dlp:
Great, let us know when this is ready for review.

ericpre:
AppVeyor and Travis are consistently raising an error that I can't reproduce locally neither in linux nor in windows... and the error is related with a userwarning. Could you please have a look and see if you have any idea, what's this error?
By the way, the warning message is not very fortunate, I should change that!! ;)

I m not sure what should be the default settings for importing scale using xresolution, yresolution and resolutionunit of the tiff tag. Some software use them to correctly export the scale and some other also use it but the scale is wrong...
As it is for now, there is an option "force_read_resolution" with the default set to False and (obviously) these tags are not use to read the scale. In this case, users could miss the fact that the scale can actually be read. Would be appropriate documentation enough? Are people reading (properly) the documentation?
The other approach would be to import the scale and raise a warning saying that some software doesn't set properly the scale.
What would be the best of the two approach?

Except the error mentioned about, that I don't know how to fix, I think this is ready to review. I only need to add a small Zeiss Tiff SEM image for the tests, once I got it.

If any one has tif files he would like to include in the tests, this is the right moment!

ericpre:
Ready for review!

thomasaarholt:
When testing this pull request, my EELS SI's dimensions are reversed - the signal dimension becomes the navigation dimension and vice versa.

ericpre:
I was a bit surprise about this error, because this PR only changed files related to the tiff io plugin. I think the bug came from master. After updating to current master, it seems to be working OK. @thomasaarholt, can you please confirm?

By the way, this is ready for review. Thanks!

thomasaarholt:
Can confirm it now works! It was just a master thing - something leftover from the introduction of Signal2D, I think.

When saving a signal as tif I get the following warning. Perhaps something to take a look at?
```python
/Users/thomas/Dropbox/0_Git/GitHub Desktop/hyperspy/hyperspy/io_plugins/tiff.py:68: UserWarning: Failed to import the optional scikit image package. Loading of some compressed images will be slow.

  "Failed to import the optional scikit image package. "
/Users/thomas/Dropbox/0_Git/GitHub Desktop/hyperspy/hyperspy/io_plugins/tiff.py:68: UserWarning: Failed to import the optional scikit image package. Loading of some compressed images will be slow.

  "Failed to import the optional scikit image package. "
/Users/thomas/Dropbox/0_Git/GitHub Desktop/hyperspy/hyperspy/io_plugins/tiff.py:68: UserWarning: Failed to import the optional scikit image package. Loading of some compressed images will be slow.

  "Failed to import the optional scikit image package. "
```

thomasaarholt:
So looking at `tiff.py`, it should only report the above warning when it fails to `import skimage`. However, I  can definitely import skimage. 
**Proof** (this is when iterating over a signal of three images):
screenshot 2016-07-07 16 08 57

thomasaarholt:
Also, we could probably update the `s.save` docstring - though I'm a little unsure why it references ipython like this when calling it in the notebook:

```python
>>> s.save?
Docstring:
Save a set of lines or a macro to a given filename.

Usage:
  %save [options] filename n1-n2 n3-n4 ... n5 .. n6 ...

Options:

  -r: use 'raw' input.  By default, the 'processed' history is used,
  so that magics are loaded in their transformed version to valid
  Python.  If this option is given, the raw input as typed as the
  command line is used instead.
  
  -f: force overwrite.  If file exists, %save will prompt for overwrite
  unless -f is given.

  -a: append to the file instead of overwriting it.

This function uses the same syntax as %history for input ranges,
then saves the lines to the filename you specify.

It adds a '.py' extension to the file if you don't do so yourself, and
it asks for confirmation before overwriting existing files.

If `-r` option is used, the default extension is `.ipy`.
File:      ~/anaconda/envs/c3/lib/python3.5/site-packages/IPython/core/magics/code.py
```

thomasaarholt:
Finally, this is really nice and definitely should be merged - having tif scale/unit support in hyperspy is wonderful.

thomasaarholt:
Also, almost incredibly, the tif unit and scale survives being opened in photoshop, layered, changed to bitmap, flattened and resaved. That is almost incredible, in my opinion.

ericpre:
To be able to write the scale for DM, you need a tifffile version more recent than ~fev. 2016 and if I remember correctly the version in skimage last release is older.

I fix the issue with the annoying and irrelevant warning, so it should be fine now.

I added some doc in the user guide, however for the docstring, I am not sure, that I can do better.

thomasaarholt:
That explains it. Thanks!

francisco-dlp:
This is great!

PR #1045, opened 2016-05-30
Enh improve model docs + heaviside component instead of the double offset

to266:
Removed the strange double eels offset component and instead added Heaviside (step) function component.

Also the docstrings of `components` module is automatically generated and includes more useful information, so that:
```python
>>> import hyperspy.api as hs
>>> help(hs.model.components)
Help on module hyperspy.components in hyperspy:

NAME
    hyperspy.components - Components that can be used to define a model for e.g. curve fitting.

DESCRIPTION
    There are some components that are only useful for one particular kind of signal
    and therefore their name are preceded by the signal name: eg. eels_cl_edge.
    
    For more details see each component docstring.
    ====================================================================
    VolumePlasmonDrude        Drude volume plasmon energy loss function c..
    Logistic                  Logistic function component..
    Bleasdale                 Bleasdale function component...
    Gaussian                  Normalized gaussian function component..
    HeavisideStep             The Heaviside step function..
    Expression                Create a component from a string expression..
    Spline                    ..
    PowerLaw                  Power law component..
    DoublePowerLaw            ..
    ScalableFixedPattern      Fixed pattern component with interpolation ..
    Offset                    Component to add a constant value in the y-..
    PESCoreLineShape          ..
    SEE                       Secondary electron emission component for P..
    Exponential               Exponentian function components..
    Erf                       Error function component..
    RC                        ..
    Lorentzian                Cauchy-Lorentz distribution (a.k.a. Lorentz..
    EELSCLEdge                EELS core loss ionisation edge from hydroge..
    Arctan                    Arctan function component..
    Vignetting                ..
    Voigt                     Voigt profile component with support for sh..
    Polynomial                n-order polynomial component...

FILE
    /home/to266/dev/hyperspy/hyperspy/components.py
```


hyperspy/components.py
```diff
@@ -53,3 +50,16 @@
 from hyperspy._components.pes_core_line_shape import PESCoreLineShape
 from hyperspy._components.volume_plasmon_drude import VolumePlasmonDrude
 from hyperspy._components.expression import Expression
+
```
francisco-dlp: This is great. It may benefit from some comments though.

francisco-dlp:
The new `Heaviside` component needs unittests, Also, just for 0.8.5, `DoubleOffset` should return `Heaviside` with a deprecation warning. Also it needs to be listed in the User Guide.

hyperspy/components.py
to266: You mean comments in the code so that it's easier for others to follow?

to266:
Done

PR #1047, opened 2016-06-01
removed charge (not defined at all) call

to266:


hyperspy/component.py
```diff
@@ -942,15 +942,13 @@ def _component2plot(self, axes_manager, out_of_range2nans=True):
         if axes_manager is not self.model.axes_manager:
             old_axes_manager = self.model.axes_manager
             self.model.axes_manager = axes_manager
-            self.charge()
```
francisco-dlp: It would be better to replace this with its modern equivalent: `fetch_stored_values`

to266:
Done, thanks. I was not aware what the deleted function did!

PR #1020, opened 2016-05-25
BUGFIX - Scale bar plotting in current master

dnjohnstone:
Fixes #1018

PR #1054, opened 2016-06-03
Fixed circular import dependecy of hyperspy.drawing.widgets

siriagus:
Fixes `ImportError: cannot import name 'widgets' ` caused by [circular import dependency ](http://stackoverflow.com/questions/17845366/importerror-cannot-import-name)of hyperspy.drawing.widgets related to PR #1020. The error is produced when trying to import hyperspy.api.


to266:
Huh, and I was just creating a PR with this same fix

siriagus:
@to266 Hehe, just updated my master today and it didn't work, so I fixed it ^^
Thanks for merging the PR! :)

Issue #962, opened 2016-04-27
Built-in absorption correction in EDS quantification


siriagus:
The mass-thickness is determined simultaneously with composition when using the zeta-factor method for EDS quantification. This makes it easy to implement absorption correction. In order to implement this, a database for mass absorption coefficients are needed. The database has been implemented in #797 and only needs a final review before it's ready to be merged. This issue should be resolved before the release of version 1.0.0. Absorption correction should also be implemented for the EDS cross section method.

k8macarthur:
Absorption corrections doesn't currently exist for cross sections but I'm talking to Sergio (Lozano-Perez) who derived cross sections with me originally and we'll see what is possible to put together.

dnjohnstone:
absorption correction is definitely a good move.... @pburdet was working on this - when he started adding stuff in this direction http://www.sciencedirect.com/science/article/pii/S0304399115300358

k8macarthur:
Ok so just a point to note whilst I"m looking through this. The absorption correction equation used in zeta factors and originally taken from http://people.umass.edu/jig0/files/70.pdf has a geometric correction factor which is assuming a small side-entry EDX detector. Therefore it's not directly applicable to, for example, the ChemiSTEM quadrant detector. I don't believe anyone in the community is tackling this yet but it might be worth bearing in mind coding this in a way which allows the geometric factor to be adjusted or chosen in the future incase a better equation is found for modern detectors?

dnjohnstone:
@k8macarthur this is an excellent point - I've had some conversations with people over here about this and we think the Manchester group may have presented something at EMAG on this vein -- I think this falls a touch outside the scope of immediate HyperSpying... If you want people to collaborate on that though I reckon hooking you up with some tomographers wouldn't be too difficult....

k8macarthur:
I agree, it's not in the immediate scope of Hyperspy but it is worth bearing in mind when coding an absorption correction function that hopefully different geometric corrections might be available to new generations of detectors in the future.

k8macarthur:
@siriagus the standard units of z-factors would be kg/m^2 right?

siriagus:
@k8macarthur I usally give them as (kg * electron)/(m^2 photon) where electron comes from the total electron dose and photon comes from the intensity (integration over line peak). That is also the same notation used in Watanabe and Williams (2006).

In other news, I used the database in #797 to implement the absorption correction. It needs some changes before it's complete, but it appears to work rather well. :)

[...]

PR #1055, opened 2016-06-03
Specify calibration units for EDS_TEM quantification

siriagus:
I've added an extra parameter, `calibration_units`, to EDS TEM quantification to specify if weight or atomic percent was used to calibrate (determine) the kfactors or zeta-factors. The way it works now, the user have to give zeta-factors and k-factors based on wt%, however, sometimes at% is more convenient (it's what I use personally the zeta-factor method), because currently it always convert the composition result from weight to atomic (for CL and the zeta-factor method)

If no `calibration_units` argument is passed to the function it will default to 'weight' for the Cliff-Lorimer and zeta-factor method as before.  For the EDS cross section method, only atomic is applied and this argument will not have any effect on the output.

### DIFF
--- a/hyperspy/_signals/eds_tem.py
+++ b/hyperspy/_signals/eds_tem.py
@@ -297,6 +297,7 @@ class EDSTEMSpectrum(EDSSpectrum):
                        method,
                        factors='auto',
                        composition_units='atomic',
+                       calibration_units = 'default',
                        navigation_mask=1.0,
                        closing=True,
                        plot_result=False,
@@ -320,6 +321,12 @@ class EDSTEMSpectrum(EDSSpectrum):
         composition_units: 'weight' or 'atomic'
             The quantification returns the composition in atomic percent by
             default, but can also return weight percent if specified.
+        calibration_units: 'default' or 'weight' or 'atomic'
+            Specifies if weight or atomic percent was used to calibrate the factors.
+            If this differs from composition_units a conversion between weight/atomic percent
+            will be performed. If 'default', the Cliff-Lorimer and zeta-factor method will
+            use 'weight'. Cross sections can only use 'atomic', so this argument has
+            no effect on the output.
         navigation_mask : None or float or signal
             The navigation locations marked as True are not used in the
             quantification. If int is given the vacuum_mask method is used to
@@ -368,6 +375,12 @@ class EDSTEMSpectrum(EDSSpectrum):
             navigation_mask = navigation_mask.data
         xray_lines = self.metadata.Sample.xray_lines
         composition = utils.stack(intensities)
+
+        if method == 'cross_section':
+            calibration_units = 'atomic'
+        elif calibration_units == 'default':
+            calibration_units = 'weight'
+
         if method == 'CL':
             composition.data = utils_eds.quantification_cliff_lorimer(
                 composition.data, kfactors=factors,
@@ -393,12 +406,12 @@ class EDSTEMSpectrum(EDSSpectrum):
             raise ValueError('Please specify method for quantification,'
                              'as \'CL\', \'zeta\' or \'cross_section\'')
         composition = composition.split()
-        if composition_units == 'atomic':
-            if method != 'cross_section':
-                composition = utils.material.weight_to_atomic(composition)
-        else:
-            if method == 'cross_section':
-                composition = utils.material.atomic_to_weight(composition)
+
+        if calibration_units == 'atomic' and composition_units == 'weight':
+            composition = utils.material.atomic_to_weight(composition)
+        elif calibration_units == 'weight' and composition_units == 'atomic':
+            composition = utils.material.weight_to_atomic(composition)
+
         for i, xray_line in enumerate(xray_lines):
             element, line = utils_eds._get_element_and_line(xray_line)
             composition[i].metadata.General.title = composition_units + \


siriagus:
@k8macarthur Any thoughts about this? It's a very small change, but I think it makes the quantification more flexible for the user.

k8macarthur:
I'm not really sure what it achieves apart from adding more details to the metadata. Would it not be simpler just to add wording to the user guide saying which units the quantification method assumes by default?

siriagus:
@k8macarthur Thanks for the feedback!

TL;DR Atomic percent makes more sense. If adding an extra arguments clutter the metadata/confuses the users, I suggest that we change it to atomic percent for the zeta-factor method instead. Earlier I thought that it was important to stick to convention, but after gaining more experimental hands-on experience, atomic percent is the only thing that really makes sense.
----------------------------------------------------------------
I'll give two example. It might just be me completely missing something obvious, if so, please let me know.

Case 1) Consider you have a thin-film consisting of 
ABC where A, B, C are different elements. Imagine that you know that the composition of the elements in atomic percent - A: 0.5, B: x and C: (1-x). This specimen is then used to determine the zeta-factor for A based on atomic percent. How would I convert that zeta-factor (wt%) to zeta-factor (at%)  if I don't know the original composition of B and C? **Conclusion:** Need the input zeta-factor to be based on atomic percent.

Case 2) Given a k-factors calibrated for your microscope you can create zeta-factors. However, k-factors are also conventionally based on weight percent. **Conclusion:** Need the input zeta-factors to be based on weight percent.

Although using a nice standard were all variables are known are much preferred, you don't always have a nice standard lying around from the elements you need.

The two solutions I propose is:
a) Introduce the calibration_factors => Users can choose. It's a default parameter (code will work as it currently does) and I don't think it clutters the metadata much.
b) Change the zeta-factor method to default use at% based zeta-factors as it makes more sense.

If we are fine with restricting the users from using one or another, I believe solution b) is a better option as it encourages people to use atomic percent, which makes much more sense as the generation of X-rays are based on the relative amount of atom and not weight percent. We also want to discourage the use of non-experimental factors. In our earlier discussions we agreed on weight percent as it has been the convention, but everyone I have talked to agrees that atomic percent makes more sense. If you have a sample with known weight-percent, you usually know it for all the elements and it is trivial to convert to atomic percent, however, as demonstrated in case 1) above it is not always true the other way around.

PS! I might just be missing something really fundamental, if so please let me know!

Have a nice day! :D

francisco-dlp:
I think that this is actually a very important issue. My personal preference is adding the option as in this PR (it enhances usability I think) and set the default to "atomic" in all cases to be consistent. This involves adding a deprecation warning where the default changes. However, I am not an EDX person myself, so take my suggestion with  a pinch of salt.

siriagus:
I have to admit that I made a grave mistake regarding this. The input composition for determining the zeta-factors need to be in weight percent. I though atomic percent could be used due to it not being clear in the original paper, but I realised today that I this is a wrong assumption (this does _not_ affect #1056 however).I think the k-factors can still be determined from atomic percent for the k-factor because it is a ratio technique. However, I checked the original Cliff-Lorimer paper today and they only mentions weight percent. Since most people uses theoretical k-factors, and should rather use the zeta-factor method or the cross-section method if they measures the factors experimentally. I think if people absolutely need to use atomic percent they should use the cross section method or convert from atomic to weight if possible. Due to my mistake I will therefore close this PR. Thank you for your time. :)

k8macarthur:
@siriagus  I'm not sure that you've made such a big mistake. Cross-sections are inversely proportional to z-factors. So provided you switch to atomic percent and you convert density to atomic density and you check the remainder of the units you should be able to get an 'atomic z-factor'. However this is then does just become 1/cross-section directly with no constants. Mathematically it's possible for you to use atomic z-factors. Although I think you should be clear a careful when doing so that you have changed all the units of your constants. (Obviously I think that edx cross sections are simpler to use for atomic composition but I'm biased on that front ;) )

francisco-dlp:
So, what do you propose @k8macarthur?

k8macarthur:
If the documentation is clear and @siriagus can convince himself he's got the maths correct then I don't see a big problem in having it. As I don't intend to use zeta factors I can't comment on how helpful it will be for more general users. 

An alternative would be to emphasis the point the atomic zfactors are simply 1/edx cross sections. Therefore (I'll need to double check my maths) I think simply using the cross section quantification with a list of 1/zfactors should also be mathematically identical. (Although I am aware that it comes across as pushing more people towards my method). 

The only difference between them is the established absorption correction which is currently only available for zfactors. Although because they are mass absorption coefficients I don't believe the maths would work with atomic zfactors.

siriagus:
TL;DR Don't include this as it only leads to confusion because it is pretty much equivalent with the edx cross sections for the zeta-factors. Absorption correction is also much easier (and runs faster) if we keep to weight percent.

@k8macarthur is right. You would have to convert between each step when using absorption correction if you had atomic zeta-factors. As I understood it, the edx cross section method is basically a transformation from the zeta-factor method to using atomic percent by converting the density term to atoms instead of mass (kg) (+ the different dose definition, that you include area). To calibrate atomic zeta-factors you basically do an (almost) equivalent transformation of units. Therefore you end up doing the same thing (almost), and might as well use the cross section method. I think we should not include this because:
- It makes it more confusing for someone new.
- Absorption correction will work as given in the zeta-factor method (for thin-films). I'm pretty sure I've got this working for #962 (using weight zeta-factors). Using weight percent makes the code run much faster if we don't have to convert back and forth all the time.

dnjohnstone:
I don't think we need to make any changes to this... although it evidently must be possible to swap units if you want to and I think atomic percentages are better overall the more important thing in this case is the convention I think.... At least for now it is preferable if we're consistent with at least the majority of the literature (and basic textbooks). For this reason alone I think I'd favour sticking to weight percent for zeta factors.

Issue #1056, opened 2016-06-03
Determine zeta-factors from known sample


siriagus:
There should be a function for facilitating the determination of the zeta-factors from a sample with known composition, thickness and dose. I have a working function for this, but it need a bit of tidying up before I can make a PR.

k8macarthur:
I thought about this. However the issue is that there are several different geometry of samples which could be used for determining thickness and therefore z-factors. Do you cater for all of them? Or do you just include the one you know? How were you thinking of doing it?

siriagus:
@k8macarthur The function calculate the zeta factors based on a known thickness. The thickness can be given as a float or a signal. For a thin-film with only one measurement (average) for thickness a float would be used. If your thickness varies a lot, a signal (or image) have to be made describing the sample where each pixel corresponds to the same pixel in your spectrum image. This map could be constructed "manually". 

Not sure what you ment, since thickness has to be determined independently. One solution is to estimate thickness from an EELS map taken at the same position and with same step size as the EDS map could be used. For a nanowire, thickness could be calculated based on orientation and projected width.

This function takes in a thickness map, it doesn't create one, so unsure what you meant. Once the thickness is known, the zeta-factor could be calculated. Please clarify if I misunderstood! :)

k8macarthur:
Sorry for forgetting to get back to this. I think you make sense. If thickness can be provided as a map or a profile I guess that keeps it general. I'll try and have a look at the actual maths in the next few weeks.

tjof2:
Closing based on comments in #1057 unless there is a demand to reopen.

PR #1057, opened 2016-06-03
Function for determining zeta-factors from known sample

siriagus:
This is directly related to #1056. A function for determining the zeta-factor(s) from a sample with known composition, thickness and dose has been implemented. Note that this method let the user select which composition_units (atomic or weight) they want to use during calibration (see also PR #1055).

### DIFF
--- a/hyperspy/_signals/eds_tem.py
+++ b/hyperspy/_signals/eds_tem.py
@@ -653,3 +653,52 @@ class EDSTEMSpectrum(EDSSpectrum):
             return real_time * beam_current * 1e-9 / constants.e
         else:
             raise Exception('Method need to be \'zeta\' or \'cross_section\'.')
+
+    def determine_zeta_factor(self, intensities, composition, thickness, density):
+        """
+        Determine the zeta-factors from a known sample.
+
+        Parameters
+        ----------
+        intensities: list of signal
+            The intensity for each X-ray line.
+        composition: list of float or list of signal
+            Composition of the elements in the same order as intensities.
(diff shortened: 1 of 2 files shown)


hyperspy/_signals/eds_tem.py
```diff
@@ -653,3 +653,52 @@ def _get_dose(self, method, beam_current='auto', real_time='auto',
...
+
+    def determine_zeta_factor(self, intensities, composition, thickness, density):
+        """
+        Determine the zeta-factors from a known sample.
+
+        Parameters
+        ----------
+        intensities: list of signal
+            The intensity for each X-ray line.
+        composition: list of float or list of signal
```
k8macarthur: Is there anyway to make to composition default to one all over. Assuming most people would use pure standard samples as was written in the original paper?

k8macarthur:
So if I understand this correctly your functions produces and map of zfactors corresponding to an intensity and thickness maps? Would it not be a more helpful function if it gave you a z-factor or a list of z-factors with an error in from the variation?

francisco-dlp:
Is any EDX available to review this?

francisco-dlp:

dnjohnstone:
@k8macarthur is this PR still relevant / do you think it's something that we want - or should we close it?

k8macarthur:
Personally I think this falls into a workbook category rather than a function category. I would close it for now.

Best wishes,

Katherine MacArthur

tjof2:
Closing based on the above comments unless there's a demand to reopen.

Issue #972, opened 2016-05-03
Incorporating Robust PCA into HyperSpy


tjof2:
A lot of my recent work has focused on robust matrix recovery and completion methods, in particular [Robust PCA](http://perception.csl.illinois.edu/matrix-rank/introduction.html), which is robust to outliers and sparse corruptions, and can also be used for things like spike removal. The method is similar to PCA basically in name only; they're both matrix factorization methods, but RPCA returns a low-rank component **L** (i.e. a model) + sparse error component **S**, rather than the usual scores+loadings seen in PCA. It's good as a denoising step, for inpainting, and also for pre-processing prior to ICA/NMF etc.

I've written a Python implementation of Robust PCA. However, RPCA takes a regularization parameter as the option, rather than an integer number of components to include, and as mentioned returns a model only. As such, I'm not sure how it will fit into the current form of `mva.py` in HyperSpy.

As @dnjohnstone has alluded to elsewhere, PCA, RPCA and others really come under an umbrella of "matrix factorization" methods.

---

As an aside, I've also half-written a version of _online RPCA_, which is much less memory-intensive  (@to266 might be interested in this), and again does not return _U_, _S_, _V_ in the same way as PCA. This could be implemented once we've sorted out this problem.

dnjohnstone:
Just to add in some details from conversations I've had with various people in person.... I think this largely comes down to questions about the overall structure of the "machine learning" part of the code.

I can't see a natural place for a function that returns a learnt model rather than a loading/factor combination. The BSS part of the code is also known to be a bit of a mess.... I know that many of us are stretching ourselves a fair bit at the moment - but 1.0.0 sounds like a sensible place for structural change if the path is clear @francisco-dlp

francisco-dlp:
@tjof2, I think that Robust PCA would indeed be highly interested for many HyperSpy, so we would love to include it in HyperSpy. However, I would suggest that you approach the scikit learn community first to see if they're interested on it as it really belongs there. Once it ships with sklearn, it'll be straightforward to include it in HyperSpy too. If the sklearn community is not interested in RPCA, we can consider then what's the best way to include it in HyperSpy. Does it sound like a good plan?

tjof2:
@francisco-dlp Thanks - there's an [ongoing discussion](https://github.com/scikit-learn/scikit-learn/issues/5851) at scikit-learn about it. I'll look into that.

[...]

PR #382, opened 2014-10-21
Add option to return extra info to decomposition.

francisco-dlp:
Some decomposition algorithms generate extra info that is not stored by HyperSpy. This adds the possibilty to return the extra information by setting `return_info` to True (the default is False).

PR #1071, opened 2016-06-14
Robust PCA & Online RPCA algorithms

tjof2:
Robust PCA seems to be making its way into scikit-learn, slowly... (see #972), but with one of the slow, iterative, memory-intensive SVT algorithms. However, I've been using this example of "online" robust PCA for a while now for outlier detection and working with missing data, which is much less memory-intensive and actually isn't too slow.

Since OR-PCA actually returns a low-rank component **L** and a sparse error component **S** (where the observations **X** = **L** + **S**), to fit this into the current HyperSpy decomposition framework, I carry out a subsequent SVD on **L** to return factors and loadings.

Although I've not currently implemented it in a way that allows for it, you can actually run OR-PCA in a truly online manner, i.e. update the estimated factors and loadings with each new sample that arrives, without having to re-run over the whole dataset. However, I think this quickly becomes complicated in the current HyperSpy machine learning framework, so consider this implementation more aimed at outlier detection.

You can see an example here: https://github.com/tjof2/hyperspy/blob/my-demos/examples/Online%20Robust%20PCA.ipynb

- [x] Documentation
- [x] Write simple tests
- [x] Option to feed in an initial mask for the sparse component **S**. 
- [x] Add batch method (GoDec) for comparison with online version
- [x] Add new initialization method for faster convergence (discussion with @thomasaarholt)

**Update 06/07/16**
A few changes to be made to fix convergence of the online algorithm after testing with more data. I'll do this asap.


tjof2:
One thing that did just occur to me is that I think I have signal vs navigation the wrong way round in the implementation. It doesn't affect the results per se (see the example notebook), but it would cause a problem for a "real-time" application. I'll investigate.

thomasaarholt:
I really like your example as it's very clear and short - however I can't get it to work (I just updated to the latest sklearn etc.). Am I missing something?

```python
<ipython-input-5-60b4c0d5f745> in <module>()
      3                       output_dimension = 10,
      4                       lambda1 = 0.005,
----> 5                       lambda2 = 0.005)
      6 cserror.plot_explained_variance_ratio()
      7 

/Users/thomas/Dropbox/0_Git/GitHub Desktop/hyperspy/hyperspy/_signals/eds_tem.py in decomposition(self, normalize_poissonian_noise, navigation_mask, closing, *args, **kwargs)
    537         super(EDSSpectrum, self).decomposition(
    538             normalize_poissonian_noise=normalize_poissonian_noise,
--> 539             navigation_mask=navigation_mask, *args, **kwargs)
    540         self.learning_results.loadings = np.nan_to_num(
    541             self.learning_results.loadings)

/Users/thomas/Dropbox/0_Git/GitHub Desktop/hyperspy/hyperspy/decorators.py in wrapper(self, *args, **kwargs)
     89         if self.auto_replot is True:
     90             self.auto_replot = False
---> 91             to_return = cm(self, *args, **kwargs)
     92             self.auto_replot = True
     93             return to_return

/Users/thomas/Dropbox/0_Git/GitHub Desktop/hyperspy/hyperspy/learn/mva.py in decomposition(self, normalize_poissonian_noise, algorithm, output_dimension, centre, auto_transpose, navigation_mask, signal_mask, var_array, var_func, polyfit, reproject, **kwargs)
    333                 explained_variance = S ** 2 / len(factors)
    334             else:
--> 335                 raise ValueError('Algorithm not recognised. '
    336                                  'Nothing done')
    337 

ValueError: Algorithm not recognised. Nothing done
```

tjof2:
@thomasaarholt have you got the right branch? That looks like `'orpca'` is not specifed as an option in `mva.py`.

tjof2:
Not having much luck with the tests today!

thomasaarholt:
@tjof2 I feel silly. I somehow assumed I didn't need to pull this branch. Looking at the code I just assumed I needed to update sklearn to the latest so that it supported orpca.

tjof2:
@thomasaarholt not to worry! Let me know if there any issues arising from your data. In particular, you may find you need to change `lambda1` and `lambda2` depending on the fraction of errors in your data etc.

vidartf:
If I understand this correctly, this should enable a such a usage like this (?):
 - Start OR-PCA, where the output signal is interactive (i.e. events trigger when a new observation is processed).
 - Wire events to plot `N` most significant components interactively (for spectra this could e.g. be plotted by `hs.plot.plot_spectra()`).
 - Feed the OR-PCA observations from a random walk of the navigation space.

This should allow you to observe the PCA convergence live. Won't be scientifically significant, but potentially useful? :)

tjof2:
@vidartf yes that's the plan! It actually factors the low rank part into **L** x **R**^T, so basically the factors and loadings but with no singular values, so these can indeed be plotted interactively, as each new observation arrives.

Of course, that might not necessarily work in the current `.decomposition()` framework. And I also need to fix the fact it currently works on the transposed data by accident.

vidartf:
@tjof2 Nice! Let me know if you run into any issues getting the interactive bits set up correctly :)

tjof2:
Does anyone know why the DM reader is failing here? I've not changed anything...

francisco-dlp:
We have to refurbish the whole learning code in order to make it easy to add online algorithms. I think that this should be done in a separate PR.

@tjof2, let us know when this is ready for review.

tjof2:
@francisco-dlp Shall I forget about the online part of it for now then? 

Otherwise, I've got a couple more tweaks to do, sometime early next week I hope.

tjof2:
@francisco-dlp Ok thanks - I've updated the code to incorporate an initial estimate of the sparse error - the only thing outstanding is testing but I'm not sure of the best way to proceed with that?

francisco-dlp:
If this was a translation from e.g. a MATLAB implementation of the algorithm by the original authors maybe you could process some data using the original code and write unittests that simply verify that the output of your Python implementation is correction.

If this was implemented directly from the paper, maybe you could ask the paper authors to review the implementation?

In any case, the tests should verify that all the arguments do what they are suppose to do and that the output is as expected.

tjof2:
Great thanks - I'll work on it next week!

tjof2:
@francisco-dlp any idea why the tests keep failing?

thomasaarholt:
My intention right now is to learn rather than contribute, but it seems like AppVeyor fails to reach api.eelsdb.eu. Maybe `line 9` in `hyperspy/hyperspy/tests/datasets/test_eelsdb.py`should have "https" instead of "http"?

tjof2:
Possibly related to https://github.com/hyperspy/hyperspy/pull/1089#issuecomment-226951816 ?

thomasaarholt:
Extremely probably so, I would say.

tjof2:
This is ready for review.

tjof2:
@thomasaarholt did you get a chance to try this out at all?

hyperspy/learn/rpca.py
```diff
@@ -0,0 +1,189 @@
...
+            L = _updatecol(L, A, B, I)
+
+    # Do final SVD
+    U, S, Vh = svd(np.dot(L, R))
+    V = Vh.T
+
+    # Chop small singular values which
+    # likely arise from numerical noise
+    # in the SVD.
+    S[S<=1e-9] = 0.0
```
tjof2: This could be better achieved by ensuring all singular values above `rank` are set to zero. e.g. `S[rank:] = 0.`.

thomasaarholt:
My supervisor gave me some good advice on writing my thesis (it was "write your thesis"), so I did that over the weekend. I'll have a play with my data tomorrow!

hyperspy/tests/mva/test_orpca.py
```diff
@@ -0,0 +1,113 @@
+import numpy as np
+import scipy.linalg
+
+import nose.tools as nt
+from nose.plugins.skip import SkipTest
+
+from hyperspy.learn.rpca import orpca
+from hyperspy.misc.machine_learning.import_sklearn import sklearn_installed
```
tjof2: Line not needed as not used,

doc/user_guide/mva.rst
```diff
@@ -166,11 +156,39 @@ To perform Poissonian noise normalization:
...
+
+PCA is known to be very sensitive to the presence of outliers in data. These outliers
+can be the result of missing or dead pixels, X-ray spikes, or very low count data.
+If one assumes a dataset to consist of a low-rank component **L** corrupted by
+a sparse error component **S**, then Robust PCA (RPCA) can be used to recover the
+low-rank component for subsequent processing.
+
+HyperSpy implements an *online* algorithm for RPCA developed by Feng et al. [Feng2013]_.
+This minimizes memory usage to make it suitable for large datasets. The method will
+return both **L** and **S**, as well as the factors and loadings of **L**.
```
tjof2: Currently only returns the factors and loadings of **L** - I'm not sure what to do with the other variables?

tjof2:
Returning the low-rank and error parts of the result could be achieved using #382 once implemented.

hyperspy/learn/rpca.py
```diff
@@ -0,0 +1,190 @@
...
+    m, n = X.shape
+
+    # Check options if None
+    if method is None:
+        _logger.warning("No method specified. Defaulting to "
+                        "closed-form solver")
+        method = 'CF'
+    if lambda1 is None:
+        _logger.warning("Nuclear norm regularization parameter "
+                        "is set to default.")
```
francisco-dlp: What about adding what the default value is in the log message?

hyperspy/learn/rpca.py
```diff
@@ -0,0 +1,190 @@
...
+        lambda1 = 1.0 / np.sqrt(n)
+    if lambda2 is None:
+        _logger.warning("Sparse regularization parameter "
+                        "is set to default.")
+        lambda2 = 1.0 / np.sqrt(n)
```
francisco-dlp: Same as above.

hyperspy/learn/rpca.py
```diff
@@ -0,0 +1,190 @@
...
+    This function performs Online Robust PCA with
+    with missing or corrupted data.
+
+    Parameters
+    ----------
+    X : numpy array
+        is the [m x n] matrix of observations.
+    rank : int
+        The model dimensionality.
+    lambda1 : float
```
francisco-dlp: Should be float or None. The default value obtained with `None` should be explained.

hyperspy/learn/rpca.py
```diff
@@ -0,0 +1,190 @@
...
+        Nuclear norm regularization parameter.
+    lambda2 : float
```
francisco-dlp: Same as above.

hyperspy/learn/rpca.py
```diff
@@ -0,0 +1,190 @@
...
+        Sparse error regularization parameter.
+    method : 'CF' | 'BCD'
```
francisco-dlp: "CF" seems not be actually be the default method (it's None).

hyperspy/learn/rpca.py
```diff
@@ -0,0 +1,190 @@
...
+        _logger.warning("Sparse regularization parameter "
+                        "is set to default.")
+        lambda2 = 1.0 / np.sqrt(n)
+
+    # Check options are valid
+    if method not in ('CF', 'BCD'):
+        raise ValueError("'method' not recognised")
+
+    # Set random seed
+    if seed is not None:
```
francisco-dlp: Does it make any sense setting a random seed?

tjof2: If you want the same random initialization and thus convergence, I think so?
I do want to add a second method that will initialize based on the first few columns/rows, but I need to test it.

vidartf: Sure, but is it the job of this function to seed random, or is that the caller's responsibility? The alternative is to document that you are using `np.random`, and point them to `np.random.seed()` for reproducibility. Not sure which is best, but setting the seed here will also seed following calls to `np.random`, which might not be desirable.

tjof2: Ah I see, does the seed affect things globally? I don't really understand it in Python.

tjof2: In that case I'll take it out.

tjof2:
@francisco-dlp updated regarding random seeds.

tjof2:
@francisco-dlp is there anything else outstanding for this?

thomasaarholt:
Hm. So time flies eh? I just managed to check this out now. (I also might have finally figured out how to checkout git pull requests yesterday). 

I can't seem to get this to work on my 50 MB data set. Thought I'd done something wrong with git, but now trying it on the example notebook you provided, things seem to work, albeit take a lot of time.

Firstly, how long does it take to run on the test data in the notebook you provided? On mine it took a few minutes, whilst "normal" pca decomposition runs in seconds. I also don't seem to get the same results as you. Note that the explained variance ratio plot is bugged for me, for some unknown (but reported) issue.

screenshot 2016-07-05 22 12 05

tjof2:
@thomasaarholt in answer to your questions:

**Speed:** Yes it will take some time for large datasets, the algorithm is designed for low-memory / real-time updating / incremental updating rather than for speed. Even standard RPCA algorithms are slower because they are solving a more complex optimization problem that requires multiple SVDs rather than just one. Also, this is written in pure Python rather than C, which may well be a major factor.

That said, it takes seconds on my PCA in both cases, not a few minutes. WIth regards to your dataset, what do you mean by "I can't seem to get this to work"? Errors? Or no response?

**Difference to example**: I will check this tomorrow but one reason could be the random initialization and subsequent convergence.

thomasaarholt:
Sorry,  I started writing the reply here after testing on my dataset, then realised I should probably try to do the example you gave. 

I left my (50,50|2048) dataset running for about twenty minutes - that's what I unhelpfully meant by "can't get it to work". 

I'll do some more testing in the morning!

tjof2:
@thomasaarholt Thanks for testing! I've traced the source of the slow-down **and** the difference, it is indeed the initialization. Fix coming!

thomasaarholt:
Mother of God, man. That... was fast.

tjof2:
@thomasaarholt :-D

thomasaarholt:
It is fast, it works, it did my 50MB sample in three seconds. 
It appears to work better than even when I'd removed X-rays previously. [More testing needed, will be back]

Well done.

thomasaarholt:
As I understand it, ORPCA in hyperspy requires us to do s.decomposition() first before doing it again with orpca. Since I'm working with EELS spectra which have poissonian noise, I normally do `s.decomposition(normalize_poissonian_noise = True)`. Should I be including this argument twice? Once in the first `s.decomposition` and then again with the orpca one?

hyperspy/learn/mva.py
```diff
@@ -185,8 +185,19 @@ def decomposition(self,
...
+                raise ValueError("With the MLPCA algorithm the "
+                                 "output_dimension must be specified")
+
+        if algorithm == 'orpca':
+            if normalize_poissonian_noise is True:
+                _logger.warning(
+                    "It makes no sense to do normalize_poissonian_noise with "
+                    "the ORPCA algorithm. Therefore, "
+                    "normalize_poissonian_noise is set to False")
+                normalize_poissonian_noise = False
```
tjof2: @thomasaarholt this line might make it clear - `normalize_poissonian_noise` doesn't make sense in ORPCA, because the whole point is that it can deal with Poisson noise robustly.

tjof2: So you should just do `s.decomposition(algorithm='orpca')` straight on the data, no prior PCA.

thomasaarholt:
Right, on my Lowloss EELS imaging this seems to be working well, about on par with normal PCA/ICA on data that has had x-ray spikes removed (using my own "dumb" script - it doesn't add noise like the standard hyperspy one does). 

Interestingly, running ORPCA on data with the spikes removed gives way noisier data. I guess that makes sense, but I'm surprised it notices so much more than the normal PCA does. 

Here are some example loadings after ICA on the various PCA results. They're of zirconium sawtooth suboxide, and this area has three phases in it:

PCA/ICA on spike-removed data:
[EELS_Spectrum_Image_0eV Loading_2.pdf](https://github.com/hyperspy/hyperspy/files/349926/EELS_Spectrum_Image_0eV.Loading_2.pdf)

ORPCA/ICA on raw data:
[EELS_Spectrum_Image_0eV Loading_2.pdf](https://github.com/hyperspy/hyperspy/files/349927/EELS_Spectrum_Image_0eV.Loading_2.pdf)

ORPCA/ICA on spike-removed data by my code:
[EELS_Spectrum_Image_0eV Loading_2.pdf](https://github.com/hyperspy/hyperspy/files/349928/EELS_Spectrum_Image_0eV.Loading_2.pdf)

ORPCA/ICA on spike-removed data using DigitalMicrograph's x-ray removal:
[EELS_Spectrum_Image_0eV Loading_2.pdf](https://github.com/hyperspy/hyperspy/files/349930/EELS_Spectrum_Image_0eV.Loading_2.pdf)

thomasaarholt:
Reconstructing a line profile with the same number (3) of eigenvalues, gives an unreal "beginning" of the reconstructed signal in some cases. The loadings that come with the orpca factors also look significantly more erratic than the pca ones. 

Raw spectrum:
screenshot 2016-07-06 18 46 25

Reconstructed by PCA:
screenshot 2016-07-06 18 46 30

Reconstructed by ORPCA:
screenshot 2016-07-06 18 46 23

hyperspy/learn/mva.py
thomasaarholt: Didn't see this before now - might update the replies below accordingly

tjof2:
To clarify the above issues from @thomasaarholt, it is a result of the online nature of the algorithm, thus better initialization is needed. I will investigate.

thomasaarholt:
+1

tjof2:
@thomasaarholt I've updated the PR with the new initialization method. 

You can specify `init='qr'` to use it over the default `'rand'`, and you can change the number of training samples with the parameter `training=int`. It currently defaults to 10. Can you double-check it on your data for me to check I've not missed anything?

tjof2:
Latest commit introduces a new algorithm that operates on the whole matrix at once, so is memory-intensive for large datasets in comparison to the online method. However, it is useful for comparison.

Accessed using `s.decomposition(algorithm='rpca')`.

tjof2:
@thomasaarholt good point. It should give similar results for the low-rank matrix, I found it useful to check the problems with initialization you reported above (`'rpca'` has no such issue with initialization). That said, I actually find it to be slower than the `'orpca'` method in practice.

hyperspy/learn/mva.py
tjof2: After discussion with @francisco-dlp this has actually been removed and left as a user option to disable. (Further research in the literature is needed to check the Poisson noise behaviour properly)

tjof2:
Actually, after some thought, I think `'qr'` should be the default method, since it improves convergence significantly.

Also, it is my contribution to the algorithm!

tjof2:
@francisco-dlp after our discussion, it did just occur to me that [the original example I posted](https://github.com/tjof2/hyperspy/blob/my-demos/examples/Online%20Robust%20PCA.ipynb) did use EDS data as the example, with `normalize_poissonian_noise=False`, and it seemed to work quite well? But your point still stands - best to leave it as a user option and I'll research the literature in the meantime.

tjof2:
I'll resubmit - broke it locally.

francisco-dlp:
ok

PR #1058, opened 2016-06-03
Added support for complex data

CodeMonkeyJan:
As discussed in #787 I added the two classes for hologram and wave images.
As stated there, HologramImage is nearly empty and WaveImage has some basic functionality (convenience properties and a few functions).
There are a lot of TODOs in the code which in part hold a few things I could still do, for example making the functions work with image stacks. For this I would like to ask what the proper implementation for iterating over the images in the stack in HyperSpy would be. Any ideas?
Other TODOs include names of colleagues of mine (@FWin22 and @woozey), whose tasks it would be (after a successful merge) to fill the two classes with life :-)!
I also added a few basic tests.
I know this PR is in an early state, but maybe you could give me input what could be improved.

-------------------------------------------------------------------
I'll track the required course of action here:
- [x] Merge #1074 adding the skimage dependency (this should be merged first, because both other PRs implement this as well).
- [x] Merge #1089 for ufunc functionality.
- [ ] Merge this PR containing the WaveImage class, testing and documentation.
- [ ] Merge #1079 containing the HologramImage class, testing and documentation, maintained by @woozey. This should be merged last, because it depends on both prior PRs.

PR #1074, opened 2016-06-16
Added dependency for scikit-image.

CodeMonkeyJan:
As mentioned in #1058 and #708


francisco-dlp:
Great, you have only missed adding it to `anaconda_hyperspy_environment.yml`.

We usually try to minimize the number of non-pure-python packages that HyperSpy's depend on. Notice e.g. that HyperSpy doesn't require scikit-learn, but installing it enables some features. I think that we should do the same about skimage.

So, in practice, all that I'm proposing is to remove `scikit-image` from `install_requirements` in `setup.py`.

Once the issues above are addressed, I think that this is ready to merge. One minor comment: normally we add dependencies with the feature that requires them, why submitting this as a separate PR?

dnjohnstone:
@francisco-dlp the separate PR is my fault/suggestion - it's just in this particular case we have more than one PR, that I expect will be merged before too long, that adds this dependency and I thought this would be a cleaner solution to get this in to the master and thus tidy up the other PRs/reduce risk of them containing their own bugs on this bit.

CodeMonkeyJan:
Are you sure it is missing in `anaconda_hyperspy_environment.yml`? I did not add it, but it is present as far as I can see:
>name: hyperspy
>dependencies:
>- python=3.5
>- dill
>- h5py
>- ipython
>- jupyter
>- matplotlib
>- nose
>- numpy
>- qt
>- __scikit-image__
>- scikit-learn
>- scipy
>- setuptools
>- statsmodels
>- sympy
>- natsort
>- ncurses
>- pip:
>  - dask
>  - hyperspy
>  - traits
>  - traitsui
>  - ipywidgets

`scikit-image` is now gone from the `install_requirements` in `setup.py`. As it is only lazy-imported when neccessary in `WaveImage`, everyone who does not use it should be fine without `skimage`.

francisco-dlp:
You are right, sorry for that.

PR #1075, opened 2016-06-16
Added basic class for hologram images

CodeMonkeyJan:
As mentioned in #1058 I separated the HologramImage class. This PR is mainly to keep this feature separated, in and off itself it is quite empty and holds mainly the documentation. I'm not sure how to proceed from here, @woozey will be the one who would make the first significant function for this class. How should he pick up from here?

PR #1079, opened 2016-06-17
Added basic class for hologram images

woozey:
Contains:
- Empty class structure from @CodeMonkeyJan 
- Reconstruction of holograms

Replaces PR #1075 
Depends on PR #1058


CodeMonkeyJan:
Great :-), I closed #1075, as this PR now fills the `HologramImage` class with much needed life as discussed in #1058. I'll update my comments there accordingly.

It seems that your test function `test_reconstruct_phase` does not work... I'll have a look later.

hyperspy/_signals/hologram_image.py
```diff
@@ -0,0 +1,303 @@
...
+
+        image_size = sb_size*2
+        (sx, sy) = holo_data.shape
+        holo_data = np.float64(holo_data)
+
+        h_hw_fft = fftshift(fft2(holo_data))  # <---- NO Hanning filtering
+
+        sb_roi = h_hw_fft[(sb_pos[0]-image_size//2):(sb_pos[0]+image_size//2),
+                 (sb_pos[1]-image_size//2):(sb_pos[1]+image_size//2)]
+
```
CodeMonkeyJan: Apparently the test is still failing here, because the code tries to slice with floats. Can you debug this, to see where the float occurs? If I understood you correctly, on your system, the test does not fail, correct?

woozey: > Apparently the test is still failing here, because the code tries to slice with floats. Can you debug this, to see where the float occurs? If I understood you correctly, on your system, the test does not fail, correct?
I'll do debugging. It should be easy to fix, but I would expect the error arising again in other places...
On my system I never got any problem with it - the tests are running. I'm puzzled here.

CodeMonkeyJan:
Seems as if your night shift worked out :-)! It seems the problem was, that `image_size` was a float? Was this behavior intended? The current fix looks more like a workaround...

CodeMonkeyJan:
Furthermore: I think it would be best if you document your reconstruction function in:
hyperspy/doc/user_guide/electron_holography.rst
Also make sure that the `signal_type` change from `HOLOGRAM` to `hologram` is reflected in the tables of:
hyperspy/doc/user_guide/tools.rst
I sure forgot this in my PR and will change it now.

woozey:
Crap!

CodeMonkeyJan:
Angry about the build failures? Don't worry, it's because
> the tests are only failing because the EELSdb certificate seems to have expired
(see #1089) All current PRs should have this problem ;-)...

woozey:
Happy to hear that it is not problem of my PR!
@CodeMonkeyJan, I've done everything you have suggested. img_size is now integer from the moment it is defined as well))

hyperspy/_signals/hologram_image.py
```diff
@@ -0,0 +1,303 @@
...
+            set True to return reconstruction as tuple, useful for batch processing
+        fresnel_fringe_filter : boolean
+            set True to apply an automatic filtering of biprisms fresnel fringe
+        verbose : boolean
+            set True to see details of the reconstruction (i.e. SB selection)
+
+        Returns
+        -------
+        wave : WaveImage
+            Reconstructed electron wave. By default object wave is devided by reference wave
```
CodeMonkeyJan: For cross-referencing in the documentation, you can use :class:`~hyperspy.signals.WaveImage`.

hyperspy/_signals/hologram_image.py
```diff
@@ -0,0 +1,303 @@
...
+            see the description in Parameters (optional, triggered by return_param flag)
+
+        Notes
+        -----
+        The reconstruction parameters assigned interactively if rec_param is not givven
+
+        See Also
+        --------
+        _reconstruct
+
```
CodeMonkeyJan: `_reconstruct` is per your definition a private function and thus should not be mentioned to the end user, I think. Keep your secrets ;-)...

hyperspy/_signals/hologram_image.py
```diff
@@ -0,0 +1,303 @@
...
+        else:
+            ref_data = reference
+            eh_hw_fft = fftshift((fft2(ref_data)))
+
+        # Preparing reconstruction parameters:
+        if rec_param is None:
+            eh_hw_fft_half = eh_hw_fft[0:np.int(sx / 2.02), :]  # selects half of the fft excluding fraction of points
+            #  in the middle
+            yr, xc = np.unravel_index(eh_hw_fft_half.argmax(), eh_hw_fft_half.shape)  # Center of selected sideBand
+            sb_pos = [yr, xc]
```
CodeMonkeyJan: Just a minor nitpick: When `yr` and `xc` are the coordinates of the center of the selected side band, why aren't they called `yc` and `xc`(c for center, I assume)? Or what does the r in `yr` mean? Also, the convention is to use a tuple instead of a list for a shape (which I assume `sb_pos` should be.
Furthermore, you can directly assign:
```python
sb_pos = np.unravel_index(eh_hw_fft_half.argmax(), eh_hw_fft_half.shape)
```
and later use `sb_pos[0]` and `sb_pos[1]` instead of `yc` and `xc`

hyperspy/_signals/hologram_image.py
```diff
@@ -0,0 +1,303 @@
...
+        rec_param : tuple
+            Reconstruction parameters in sequence (SBrect(x0, y0, x1, y1), SB size)
+        show_phase : boolean
+            set True to plot phase after the reconstruction
+        return_param : boolean
+            set True to return reconstruction as tuple, useful for batch processing
+        fresnel_fringe_filter : boolean
+            set True to apply an automatic filtering of biprisms fresnel fringe
+        verbose : boolean
+            set True to see details of the reconstruction (i.e. SB selection)
```
CodeMonkeyJan: `verbose` and its description are a bit misleading. If I understand correctly, this just determines if the SB is plotted or not. Maybe this should be renamed to `show_sideband` (equivalent to `show_phase`) or something?

hyperspy/_signals/hologram_image.py
```diff
@@ -0,0 +1,303 @@
...
+        -----
+        The reconstruction parameters assigned interactively if rec_param is not givven
+
+        See Also
+        --------
+        _reconstruct
+
+        """
+
+        (sx, sy) = self.data.shape
```
CodeMonkeyJan: In accordance with later comments, I personally would use
```python
holo_size = self.data.shape
```
and always refer the tuples indices instead of having `sx` and `sy` separately, but that's just my opinion. If you want to stay with your convention, just adapt the later comments accordingly.

hyperspy/_signals/hologram_image.py
```diff
@@ -0,0 +1,303 @@
...
+            # to reconstruct hologram WITH a reference
+            ref_data = reference.data
+            eh_hw_fft = fftshift((fft2(ref_data)))
+        else:
+            ref_data = reference
+            eh_hw_fft = fftshift((fft2(ref_data)))
+
+        # Preparing reconstruction parameters:
+        if rec_param is None:
+            eh_hw_fft_half = eh_hw_fft[0:np.int(sx / 2.02), :]  # selects half of the fft excluding fraction of points
```
CodeMonkeyJan: See above: `holo_size[1]` instead of `sx` (or not depending on your choice).

hyperspy/_signals/hologram_image.py
```diff
@@ -0,0 +1,303 @@
...
+            #  in the middle
+            yr, xc = np.unravel_index(eh_hw_fft_half.argmax(), eh_hw_fft_half.shape)  # Center of selected sideBand
+            sb_pos = [yr, xc]
+            # Default SB size is 1/2 of the distance to main band
+            sb_size = np.round(1.0 / 2 * (norm(np.subtract(sb_pos, [sx / 2, sy / 2]))))
+            # To be sure of even number, still questionable if it is needed:
+            sb_size = sb_size - np.mod(sb_size, 2)
+            rec_param = (xc - 1, yr - 1, xc + 1, yr + 1, sb_size)
+            print("Sideband Size in pixels")
+            print("%d" % sb_size)
```
CodeMonkeyJan: You said that you want `sb_size` to be an integer and I think here is the place where you should make sure of that using the following alternative:
```python
sb_size = norm(np.subtract(sb_pos, [sx / 2, sy / 2])) // 4 * 2
```

The `// 4 * 2` part will make sure of three things all at once:
- `sb_size` is halve of the distance to the main band.
- `sb_size` is an integers (`//` is integer division).
- `sb_size` is an even numbers so you don't need `sb_size = sb_size - np.mod(sb_size, 2)`.


Also I assume loggers instead of print statements would be better in this case.
```python
import logging                         # These first two lines
_logger = logging.getLogger(__name__)  # go to the import section

_logger.info('Sideband size in pixels: {}'.format(sb_size))
```
------
Also, why is the SB rectangle you set with the first four entries of `rec_param` only 3 pixels big? Is it because of the automated peak finding instead of setting the rectangle via a GUI?

hyperspy/_signals/hologram_image.py
```diff
@@ -0,0 +1,303 @@
...
+            # plt.close(f)
+            # rec_param = (rect.x0, rect.y0, rect.x1, rect.y1, sb_size)
+        else:
+            arect = eh_hw_fft[rec_param[1]:rec_param[3], rec_param[0]:rec_param[2]]
+            # Sideband position,find the max number and its [c,r]
+            yr, xc = np.unravel_index(arect.argmax(),
+                                      arect.shape)  # center of the selected sideBand
+            sb_pos = [np.round(rec_param[1]) + yr, np.round(
+                rec_param[0]) + xc]  # <----- use this one in the case of usage of rect_roi obj
+            sb_size = rec_param[4]
```
CodeMonkeyJan: Similar changes as above:
```python
yc, xc = np.unravel_index(arect.argmax(), arect.shape)  # center of the selected sideBand
sb_pos = (rec_param[1] + yc, rec_param[0] + xc)
sb_size = rec_param[4]
```
You should not need to round anything, because `rec_param` should just contain integers.

hyperspy/_signals/hologram_image.py
```diff
@@ -0,0 +1,303 @@
...
+            w_ref = self._reconstruct(ref_data.data, sb_size, sb_pos, fresnel_width=fresnel_width)  # reference electron wave
+
+        w_obj = self._reconstruct(holo_data, sb_size, sb_pos, fresnel_width=fresnel_width)  # object wave
+
+        wave = w_obj / w_ref
+        wave_image = self._deepcopy_with_new_data(wave)
+        wave_image.set_signal_type('wave')  # New signal is normal image without special signal type!
+        rec_param_dict = {'sb_pos_x0': rec_param[0], 'sb_pos_x1': rec_param[1],
+                          'sb_pos_y0': rec_param[2], 'sb_pos_y1': rec_param[3],
+                          'sb_size': rec_param[4]}
```
CodeMonkeyJan: By now, I'm not entirely sure if the whole `return_param` effort is neccessary. If you replace the `dict` with the following:
```python
from collections import OrderedDict  # This again goes into the import section!
rec_param_dict = OrderedDict([('sb_pos_x0', rec_param[0]), ('sb_pos_x1', rec_param[1]),
                              ('sb_pos_y0', rec_param[2]), ('sb_pos_y1', rec_param[3]),
                              ('sb_size', rec_param[4])])
```
the rec_param will be saved in the metadata of the `WaveImage` in alphabetical order, which coincidentally equates to the order you want! This way you can access them anytime later as follows:
```python
rec_param = list(wave_image.metadata.Signal.holo_rec_param.as_dictionary().values())
```
It's not short but easy to remember. Thus there would be no need use `return_param` at all. You could even implement a function `get_rec_param(wave_image)` in `HologramImage` which does this extraction of `rec_param`. What do you think?

hyperspy/_signals/hologram_image.py
```diff
@@ -0,0 +1,303 @@
...
+        wave_image.metadata.Signal.add_node('holo_rec_param')
+        wave_image.metadata.Signal.holo_rec_param.add_dictionary(rec_param_dict)
+        wave_image.axes_manager[0].scale = wave_image.axes_manager[0].scale*sx/wave.shape[0]
+        wave_image.axes_manager[1].scale = wave_image.axes_manager[1].scale*sy/wave.shape[1]
```
CodeMonkeyJan: Should you decide for the `holo_size` instead of `(sy, sx)` convention, also use it here.

hyperspy/_signals/hologram_image.py
```diff
@@ -0,0 +1,303 @@
...
+        * Inverse FFT.
+
+        Parameters
+        ----------
+        holo_data : array_like
+            Holographic data array
+        sb_size : int
+            Size of the sideband filter in px
+        sb_pos : nparray (N=1)
+            Vector of two elements with sideband coordinates [y,x]
```
CodeMonkeyJan: `sb_pos` should be a tuple if it's a shape.

hyperspy/_signals/hologram_image.py
```diff
@@ -0,0 +1,303 @@
...
+        --------
+
+        reconstruct
+
+        """
+        # TODO: Parsing of the input has to be redone
+        # TODO: Add smoothing of Fresnel filter
+        # TODO: Add other smoothing options?
+
+        image_size = np.int(sb_size*2)
```
CodeMonkeyJan: Casting to `int` should not be neccessary.

hyperspy/_signals/hologram_image.py
```diff
@@ -0,0 +1,303 @@
...
+        (sx, sy) = holo_data.shape
+        holo_data = np.float64(holo_data)
+
+        h_hw_fft = fftshift(fft2(holo_data))  # <---- NO Hanning filtering
+
+        sb_roi = h_hw_fft[(sb_pos[0]-image_size//2):(sb_pos[0]+image_size//2),
+                 (sb_pos[1]-image_size//2):(sb_pos[1]+image_size//2)]
```
CodeMonkeyJan: You can use `sb_size` instead of `image_size//2` here, because it's exactly the same. `sb_size` is already an even integer.

hyperspy/_signals/hologram_image.py
```diff
@@ -0,0 +1,303 @@
...
+
+        (sb_ny, sb_nx) = sb_roi.shape
+        sb_l = min(sb_ny / 2, sb_nx / 2)
+        cen_yx = [sb_ny / 2, sb_nx / 2]
```
CodeMonkeyJan: If I'm not mistaken than 
```python
sb_roi.shape = (sb_ny, sb_nx) = (2 * sb_size, 2 * sb_size) = (image_size, image_size)`
```
which means `sb_l = image_size` and `cen_yx = [image_size, image_size]`.
Both lines are unnecessary as I'll show later.

hyperspy/_signals/hologram_image.py
```diff
@@ -0,0 +1,303 @@
...
+
+        # Circular Aperture
+        cgrid_y = np.arange(-sb_ny / 2, sb_ny / 2, 1)
+        cgrid_x = np.arange(-sb_nx / 2, sb_nx / 2, 1)
+        (sb_xx, sb_yy) = np.meshgrid(cgrid_x, cgrid_y)
+        sb_r = np.sqrt(sb_xx ** 2 + sb_yy ** 2)
+        c_mask = np.zeros((sb_nx, sb_ny))  # Original: cMask=zeros(SB_Ny,SB_Nx);
+
+        c_mask[sb_r < sb_size] = 1
```
CodeMonkeyJan: All instances of `sb_nx` and `sb_ny` can be replaced by `image_size` as shown above.
Also you can use a slightly more elegant solution to replace the first three lines:
```python
sb_yy, sb_xx = np.indices((image_size, image_size)) - sb_size
```

hyperspy/_signals/hologram_image.py
```diff
@@ -0,0 +1,303 @@
...
+
+        # Fresnel Mask
+        ang = np.arctan2((sx / 2 - sb_pos[0]), (sy / 2 - sb_pos[1]))  # [-pi pi]
+
+        p_one = np.round([fresnel_ratio * sb_l * np.sin(ang), fresnel_ratio * sb_l * np.cos(ang)])
+        p_two = np.round([sb_l * np.sin(ang), sb_l * np.cos(ang)])
```
CodeMonkeyJan: Again, `sb_l` is just the same as `image_size`.

hyperspy/_signals/hologram_image.py
```diff
@@ -0,0 +1,303 @@
...
+
+        ang_one = (ang - np.pi / 2) % (2 * np.pi )
+        ang_two = (ang + np.pi / 2) % (2 * np.pi )
+
+        r = fresnel_width / 2
+        aa = np.round([p_one[0] + r * np.sin(ang_one), p_one[1] + r * np.cos(ang_one)]) + cen_yx
+        bb = np.round([p_one[0] + r * np.sin(ang_two), p_one[1] + r * np.cos(ang_two)]) + cen_yx
+        cc = np.round([p_two[0] + r * np.sin(ang_two), p_two[1] + r * np.cos(ang_two)]) + cen_yx
+        dd = np.round([p_two[0] + r * np.sin(ang_one), p_two[1] + r * np.cos(ang_one)]) + cen_yx
```
CodeMonkeyJan: Due to broadcasting, `cen_yx` can be replaced by a simple `image_size` (`[image_size, image_size]` is not necessary).

hyperspy/_signals/hologram_image.py
```diff
@@ -0,0 +1,303 @@
...
+        abcd = np.array([aa, bb, cc, dd])
+        f_mask = _poly_to_mask(abcd[:, 1], abcd[:, 0], sb_roi.shape)
+
+        sinc_k = 5.0  # Sink times SBsize
+        w_one = np.sinc(
+            np.linspace(-sb_ny/2, sb_ny/2, sb_ny) * np.pi / (sinc_k * sb_size))
+        w_one = w_one.reshape(image_size, 1)
+        w_two = np.sinc(
+            np.linspace(-sb_nx/2, sb_nx/2, sb_nx) * np.pi / (sinc_k * sb_size))
+        window = w_one.dot(w_two.reshape(1, image_size))
```
CodeMonkeyJan: Equivalent to `np.linspace(-sb_size, sb_size, image_size)`

hyperspy/_signals/hologram_image.py
```diff
@@ -0,0 +1,303 @@
...
+    #     ----------
+    #     angle : float
+    #         Angle in radians
+    #
+    #     Returns
+    #     -------
+    #     wrapped_angle : float
+    #         Wrapped angle between 0 and 2pi
+    #     """
+    #     return angle % (2 * np.pi )
```
CodeMonkeyJan: I think it would be best to delete unused code :-).

CodeMonkeyJan:
Sorry for flooding everyones inbox, I had an in-depth look into the code. @woozey, take everything as a suggestion and not a necessity ;-)...

hyperspy/_signals/hologram_image.py
woozey: It is a good idea implementing a function for getting the parameters out of wave_image, but I think it should go to the WaveImage class, isn't it? I thought of making it as a `@property`, do you think it is a good idea?

Unfortunately OrderedDict will not work, I've found a mistake: rec_param has the following sequence: x0, y0, x1, y1, sb_size, but should be easy to make an appropriate function

CodeMonkeyJan: s before x... basic alphabet stuff, sorry, my bad :-/...
The idea of having it as a `@property` in `WaveImage` sounds reasonable. Make sure to return something meaningful if `rec_param` was not set (e.g. when the `WaveImage` was not created via a `HologramImage`).

CodeMonkeyJan: Hm, I searched a bit more about `OrderedDict`, it does not keep an alphabetical order, but rather respects the order of your input list. So if you keep your list order in line with the order in `rec_param`, it should work out fine.

woozey: OrderedDict does it job, but it seems that it gets screwed up after retrieving it from metadata, so I've sorted the values manual in the get_rec_param  function for now.

hyperspy/_signals/hologram_image.py
woozey: Actually, this is done for feature development of the function which would allow having 'image_size>2*sb_size'

hyperspy/_signals/hologram_image.py
woozey: As mentioned above that was done for future implementation of variable image_size
So I can fix it and redo again later or can just leave it. What do you think is the best?

hyperspy/_signals/hologram_image.py
```diff
@@ -0,0 +1,286 @@
...
+
+        # Preparing reconstruction parameters:
+        if rec_param is None:
+            eh_hw_fft_half = eh_hw_fft[0:np.int(holo_size[0] / 2.02), :]  # selects half of the fft excluding
+            # fraction of points in the middle
+            sb_pos = tuple(np.unravel_index(eh_hw_fft_half.argmax(), eh_hw_fft_half.shape))  # Center of selected
+            #  Side Band
+
+            # Default SB size is 1/2 of the distance to main band
+            sb_size = int(norm(np.subtract(sb_pos, (holo_size[0] / 2, holo_size[1] / 2)))) // 4 * 2  # To be sure of
```
CodeMonkeyJan: `int()` is not necessary, the `//` will ensure that the result is an integer.

hyperspy/_signals/hologram_image.py
CodeMonkeyJan: Okay, that makes sense :-).

hyperspy/_signals/hologram_image.py
CodeMonkeyJan: It's up to you, really. It was just something I thought could be streamlined, but if you want to have the function more flexible in the future it could be detrimental to "simplify" the declarations now...

CodeMonkeyJan:
Great, I think I have no further comments on this.
Also it seems as if the EELSdb works again, the builds don't fail any longer :-)!

hyperspy/_signals/hologram_image.py
woozey: nope, `np.subtract` outputs float for some reason... ((

doc/user_guide/electron_holography.rst
```diff
@@ -0,0 +1,95 @@
...
+The :py:class:`~._signals.wave_image.WaveImage` class can hold information about the complex electron
+wave. As such, relevant properties like the `amplitude`, `phase` and the `real` and `imag` part can be
+directly accessed and return appropriate :py:class:`~._signals.signal2d.Signal2D` signals.
+
+To transform a :py:class:`~._signals.signal2d.Signa2D` (or subclass) into a 
+:py:class:`~._signals.wave_image.WaveImage` use:
+
+.. code-block:: python
+
+    >>> im.set_signal_type('wave')
```
CodeMonkeyJan: Everything after this point is now transferred to other super classes, so you should delete them here, too.

woozey: It is mess!

CodeMonkeyJan: You're right, let's wait for the final version of #1058 to avoid unnecessary work.

woozey:
@francisco-dlp:
> regarding the scientific review question, I think that it'll depend on the origin of the reconstruction method that you have implemented...

The reconstruction is a standard and well established method nowadays. It is done as follows:
- FFT
- Selecting, re-centring and low-pass filtering a side band
- IFFT

The only difference can be introduced in filtering step. Different people use different filters: it can be hard or smoothed circular aperture, or truncated window functions like Butterworth, Hanning, Sinc and etc.
At the moment I have only implemented hard circular aperture with light sinc apodization (I'll add different options later on). 

I also have constructed a test which do following:
- Generates an arbitrary phase,
- Simulates a hologram out of the phase (assuming amplitude = 1),
- Than reconstructs the wave using the reconstruction function
- Compares the phase of the reconstructed wave with the inputed wave. 

So this test insures that the reconstruction works and delivers expected results. Do you think it can be enough?

hyperspy/_signals/hologram_image.py
CodeMonkeyJan: Still, `//` is applied at the very end, so it should convert it back to an int.

woozey:
This pull request is replaced by #1152

PR #1084, opened 2016-06-18
Set version to 1.0 dev

francisco-dlp:
Replaces #922

PR #1085, opened 2016-06-18
Update version to 1.0+dev

vidartf:
Identical to #1084, but I have created this via Github's "edit" functionality, so this has never touched my local directory. Let's see what Travis says.


vidartf:
Same problem it seems. I think we might need to contact Travis here?

francisco-dlp:
(+o+) I've moved things around a little in my last commit to see what happens. If it still fails I'll follow your advice and contact travis by email.

vidartf:
Well this is fun: The build of the branch succeeds, while the PR build does not :-/

francisco-dlp:
travis-ci/travis-ci#6195

PR #1033, opened 2016-05-26
Signal1D/Signal2D merge into master

dnjohnstone:
Following #963 to 0.8.x this is the merge in to master.... I will finish it by the end of the week.

PR #1087, opened 2016-06-18
New: find_peaks2D

dnjohnstone:
A work in progress, but not too far away - myself and @bm424 have been worrying about peak finders in 2D recently. We dumped everything we had in to a branch with the aim of putting it all in HyperSpy for v1.0.0. PR being made in case anyone has any strong opinions about how/what should be implemented here.

If anyone happens to have implementations that they've found useful we'd be very pleased to hear about them. Time frame wise I hope to have this ready for merging into the master by the end of June.

To do as follows:

- [x] Tidy up the various peak finders we have
- [ ] Decide if we're going to keep all of the implemented methods
- [ ] Improve documentation for specific peak finders
- [ ]  Add a few more tests


francisco-dlp:
Nice. The obvious question that comes to my mind is: giving that these are image functions, why not contributing them to skimage first? If they are not interested on them for any reason, what about creating your own specialised python package? I am sure that there'll be plenty of people interested on having a variety of peak finding function in a simple standalone package.

Beside, I think that, for the sake of usability, it'll be better to have one method per peak finding function in `Signal2D`. Otherwise it is unnecessarily hard to get to the relevant and autocompletion doesn't work for the keyword arguments.

dnjohnstone:
vs scikit-image -- many of the methods here just wrap scikit-image stuff, the wrapper is often just necessary to get a consistent output i.e. just a straight list of peak positions. If I was a scikit-image developer I'd say that scikit-image has all it needs and these are more 'applications'. For the methods that do not use scikit-image tools much - often they are not really image processing steps i.e. you might do your image processing before finding the peaks.

Most importantly though - I'd consider the need for robust, minimal input, peak finders to be intimately linked to having multi-dimensional datasets where we now have to balance accuracy and speed and can't be tweaking parameters in between.

As regards a specialised Python package - I'd guess my argument is something similar to what happend with find_peaks_ohaver -- I'd prefer to avoid the additional overhead of maintaining a separate package if possible and my point above too.

For usability -- I can see your point, but are we going to adopt that policy on more functions? A similar complaint would be true for: decomposition(), align2D(), estimate_shifts2D() to name just a few that spring to mind immediately.

magnunor:
Very nice! Having easy-to-use 2-D peak finders is very useful, and unlocks quite a few interesting applications.

dnjohnstone:
glad you're keen @magnunor if you've got a minute to have a play with a few of these it would be interesting to know which you like. I'm quite aware that there are substantial differences between atomic resolution images (which are quite dense in intensity) and diffraction patterns (which are very sparsely peaked).

magnunor:
Testing this now, so will be leaving some various comments. Including a more involved larger test in the end.

Bug when running with method='max' 

```Python
/home/magnunor/Documents/HyperSpy/hyperspy/hyperspy/_signals/signal2d.py in find_peaks_max(z, alpha, size)
    228     k_arr = np.zeros((10000, 2))
    229     # copy image
--> 230     image_temp = copy.deepcopy(z)
    231     peak_ct = 0
    232     # calculate standard deviation of image for thresholding

NameError: name 'copy' is not defined
```

magnunor:
When running method="massiel":

```Python
/home/magnunor/Documents/HyperSpy/hyperspy/hyperspy/_signals/signal2d.py in find_peaks_masiel(z, subpixel, peak_width, medfilt_radius, maxpeakn)
    473 
    474     if medfilt_radius > 0:
--> 475         z = ndi.filters.median_filter(z, medfilt_radius)
    476     xc = [find_peaks_ohaver(z[i], medfilt_radius=None,
    477                             peakgroup=peak_width,

NameError: name 'ndi' is not defined
```

magnunor:
When running method="blob":

```Python
/home/magnunor/Documents/HyperSpy/hyperspy/hyperspy/_signals/signal2d.py in find_peaks_blob(z, threshold, **kwargs)
    529     from skimage.feature import blob_dog
    530     blobs = blob_dog(z, threshold=threshold, **kwargs)
--> 531     return blobs[:, :2]
    532 
    533 

IndexError: too many indices for array
```

magnunor:
method="zaefferer" doesn't really finish, it runs forever and when I interrupt it:

```Python
/home/magnunor/Documents/HyperSpy/hyperspy/hyperspy/_signals/signal2d.py in box(x, y, window_size, x_max, y_max)
    288         return np.array(
    289             np.meshgrid(range(x_min, x_max), range(y_min, y_max))).T.reshape(-1,
--> 290                                                                              2)
    291 
    292     def get_max(image, box):

KeyboardInterrupt:
```

magnunor:
I ran through all the methods with their default settings using the script below, on the ADF-image https://owncloud.gla.ac.uk/cloud/index.php/s/EApLxY3QPsRwEBh. Only the `skimage` method worked on my image.

```Python

import hyperspy.api as hs
import numpy as np

def plot_peaks(ax, s, peaks, name=""):
    ax.imshow(np.rot90(np.fliplr(s.data)),origin='lower')
    ax.scatter(peaks[:,0], peaks[:,1])
    ax.set_xlim(0, 1024)
    ax.set_ylim(0, 1024)
    ax.annotate(name, (0.1, 0.9), xycoords="axes fraction", color='white') 

s = hs.load("001_STO_ADF.hdf5")
#peaks_max = s.find_peaks2D(method='max')[0]
#print("max finished")
peaks_skimage = s.find_peaks2D(method='skimage')[0]
print("skimage finished")
#peaks_zaefferer = s.find_peaks2D(method='zaefferer')[0]
#print("zaefferer finished")
#peaks_stat = s.find_peaks2D(method='stat')[0]
#print("stat finished")
#peaks_massiel = s.find_peaks2D(method='massiel')[0]
#print("massiel finished")
#peaks_blob = s.find_peaks2D(method='blob')[0]
#print("blob finished")

#############
plot_x = 2
plot_y = 1
ax_size = 7
fig, axarr = plt.subplots(plot_y, plot_x, figsize=(plot_x*ax_size, plot_y*ax_size))

ax_adf = axarr[0]
ax_skimage = axarr[1]

#############

ax_adf.imshow(np.rot90(np.fliplr(s.data)),origin='lower')
plot_peaks(ax_skimage, s, peaks_skimage, name="skimage")

for ax in fig.axes:
    ax.axis('off')

fig.tight_layout(pad=0.0, w_pad=0.0, h_pad=0.0)
fig.savefig("test_peaks.jpg", dpi=300)
```
test_peaks

bm424:
Thanks for the testing @magnunor. Most of the errors were import ones left over from when `Image` was changed to `Signal2D` and so on. Those are now fixed. A couple of the peak finders require parameterization and I'm thinking of working on a convenience function to help with that. I've updated my own branch and this pull request should be updated shortly. Here are my results (code slightly updated from yours)

```python
import hyperspy.api as hs
import numpy as np

def plot_peaks(ax, s, peaks, name=""):
    ax.imshow(np.rot90(np.fliplr(s.data)),origin='lower')
    ax.scatter(peaks[:,0], peaks[:,1])
    ax.set_xlim(0, 1024)
    ax.set_ylim(0, 1024)
    ax.annotate(name, (0.1, 0.9), xycoords="axes fraction", color='white') 

methods = [
    'max',
    'skimage',
    'zaefferer',
    'stat',
    'massiel',
    'blob',
]

peaks = {}

s = hs.load("../../source_data/001_STO_ADF.hdf5")

for method in methods:
    peaks[method] = s.find_peaks2D(method=method)[0]
    print("{} finished".format(method))

#############
plot_x = len(methods)+1
plot_y = 1
ax_size = 7
fig, axarr = plt.subplots(plot_y, plot_x, figsize=(plot_x*ax_size, plot_y*ax_size))

axes = {}
for method, ax in zip(methods, axarr[1:]):
    axes[method] = ax

#############

axarr[0].imshow(np.rot90(np.fliplr(s.data)),origin='lower')

for method in methods:
    plot_peaks(axes[method], s, peaks[method], name=method)

for ax in fig.axes:
    ax.axis('off')

fig.tight_layout(pad=0.0, w_pad=0.0, h_pad=0.0)
```
test_peaks

`zaefferer` is bad but was really developed to detect the sparse, strong peaks in diffraction (as, incidentally, was `stat` though it has proved surprisingly robust). `blob` needs parameter tweaking but can be very good, particularly where peaks of different intensity are present. I'm surprised you got the result you did with `skimage` - as you can see here there are *far* too many peaks being found.

tjof2:
This looks neat but unlikely to work in my atomic resolution scenarios due to horribly complex backgrounds. Will test them anyway though!

magnunor:
@tjof2, yeah I'm a bit lucky in working with (mostly) nice and orderly atomic structures :)

@bm424, nice! I'm not really sure why your `skimage` doesn't work so well. I'm suspecting it might be due to the peak separation parameter being much lower for you for some reason. Possibly due to different `skimage` library versions?

Could maybe be an idea to include an atomic resolution image like this in the tests, and/or in some kind of example (hyperspy-demos? User guide?). The size of the file itself can probably be reduced quite a bit by having a smaller dtype, or cropping it.

tjof2:
@bm424 is there perhaps someway of incorporating constraints for cases of atomic resolution ADF-STEM images? 

I ask because usually astronomical detection algorithms have to deal with both point (star) and extended (galaxy) sources, and in diffraction patterns there can also be a wider range of sizes and intensities of the peaks. For ADF-STEM images of atomic columns, there are far fewer possibilities. SrTiO3 should only have three different sizes/intensities, for example!

bm424:
@magnunor Yep! Tests are definitely something I need to work on but I have very little experience in that area so they may be a while. The documentation will probably come along once everyone is happy with how the functions work.
@tjof2 I wouldn't want to be too constrictive. I guess it could be possible to provide two alternative modes, but I think the best way is simply going to be finding the parameters which work. For example, `blob` takes lower and upper bounds on the blob size.

tjof2:
@bm424 good idea - in which case clear documentation on upper & lower bounds might be all that is needed.

magnunor:
My plan when things settle down a little bit, is making a `STEMImage` class (or some other appropriate name), which contain more fine tuned functions for doing 2-D peak finding like this. They would simple call one of the functions in this pull request, with some appropriate parameters for an atomic resolution ADF-STEM image.

So I think the functions here should have fairly general default parameters, with more specialized sub-classes of signal2D having the more fine-tuned ones.

magnunor:
Also, likewise having a `DiffractionImage` would allow for more fine tuned functions for these kinds of datasets. (I can vaguely remember someone mentioning this earlier...).

bm424:
That sounds good - and yes, @dnjohnstone has been working on an SED Image class for some time, so it would make sense to differentiate things that way.

dnjohnstone:
@magnunor @bm424  yes I have a branch for Diffraction -- it has quite a lot in it and I need to work out what should be where, some of the methods I've implemented should be Signal2D rather than Diffraction, and I had been waiting for #1033 to be finalised before worrying too much about where these things were.

You can find it in my repository if you're really keen.

dnjohnstone:
Can we have some votes on the following please?

For this PR I think the key next step is to decide how many of these peak finders we're going to keep... I'm in favour of just keeping them all because options and flexibility usually seem like good things.

I think the default should either be the 'blob' method or the 'stat' method -- although we should rename this "difference of gaussians" or something so it's more self documenting. The 'stat' method is really something quite neat in respect of robustness and minimal tweaking (in my opinion) but it is by far the slowest. 


Once the above two are settled documentation and tests can be hacked out pretty quickly.

dnjohnstone:
By the way @tjof2 I think if you've got a horrible background situation the upshot is that you need/should do some image processing before these steps -- it's my personal opinion that that should be maintained in 'workflow' rather than the code base as it gets too application specific too fast.

That is unless you know of some incredibly robust thing that we might implement?

tjof2:
@dnjohnstone The background situation is a much bigger problem which we're tackling with wavelets and training classifiers, don't worry! 

Difference of Gaussians is good, agree on the naming, From someone who's read an awful lot of fluorescence microscopy literature, what was the reason for not including a Laplacian of Gaussian filter, which is one of the standards in that field?

Regarding the 'stat' method, is there any noticeable benefit in terms of robustness by using the median and the median absolute deviation, rather than the mean intensity and standard deviation? If so, could it be worthwhile having it as an option? Related to that: `"Peaks more than one standard deviation from the mean set to one."`. Why one SD? Why not 2? Why not a user-controlled option?

dnjohnstone:
Actually might consider removing the 'masiel' method -- as it really doesn't seem to work that well... This was just where we started because it was actually implemented in HyperSpy long ago when high resolution images came up before.

dnjohnstone:
@tjof2 Laplacian of Gaussian would be easy to implement similarly to the difference of gaussians (it's also done in skimage) the only reason for not including it was that the difference of gaussians worked better for us -- I think I like the idea of including both.

tjof2:
LoG might be worthwhile in terms of familiarity for people is all :-)

vidartf:
Drive-by comment: Are there any methods that deal well with peaks that are saturated/"blown out", i.e., the intensities are at the maximum level of the data type. A typical example is the central peak of a diffraction pattern, where the non-saturated areas should still supply enough information to fit the peak precisely (assuming negligible coma-like artifacts in DP). If so, a future use is in an `center()` method on a diffraction images signal :)

dnjohnstone:
@vidartf I think a number of these methods will deal ok with saturated peaks -- although I have another routine for the particular thing you're alluding to that will come with the diffraction sub class

magnunor:
About how many peak finders to keep: I think we should keep as many as possible in signal2D. As long as they work, and they're properly documented + have working testes.

dnjohnstone:
@magnunor I think you may have meant tests rather than testes...

magnunor:
haha, yes :)

dnjohnstone:
@francisco-dlp how are you feeling about this, before we put effort in to tests and docs?

I tend towards @magnunor's view of kind of the more the merrier - but that said I would consider removing the 'massiel' one as I've not really seen great results from that yet.

dnjohnstone:
closing in favour of #1099 as @bm424 is doing most of the leg work on this at the moment.

PR #1099, opened 2016-06-22
2D peak finders

bm424:
Taking over #1087

As well as taking into account a number of the suggestions there (including removing the Massiel/Masiel method), I've now developed an interactive tool for identifying the appropriate parameters. Although I've done a lot of restructuring it *should* be backwards-compatible.

peakfind_interactive

Still to do:

- [x] Testing
- [ ] Tests
- [x] Docs
- [ ] Peak finder class to separate file


dnjohnstone:
The interactive bit of this is really neat! I also think it addresses some of @francisco-dlp's concerns regarding usability and ease with which the parameters can be seen by the user.

hyperspy/_signals/signal2d.py
```diff
@@ -522,10 +538,768 @@ def crop_image(self, top=None, bottom=None,
...
+            return _local_stat(image, radius, np.std)
+
+        def single_pixel_desensitize(image):
+            """Reduces single-pixel anomalies by nearest-neighbor smoothing."""
+            kernel = np.array([[0.5, 1, 0.5], [1, 1, 1], [0.5, 1, 0.5]])
+            smoothed_image = generic_filter(image, np.mean, footprint=kernel)
+            return smoothed_image
+
+        def stat_binarise(image):
+            """Peaks more than one standard deviation from the mean set to one."""
```
tjof2: I'll replace my comment from the old PR here - is there any justification for using one standard deviation and not 2, 3...? Does changing it make a difference, and so should it be user-configurable?

Building on that, is there any improvement using the more robust (to outliers in the data) [median and median absolute deviation](http://eurekastatistics.com/using-the-median-absolute-deviation-to-find-outliers)? Could this be an advanced option for users to specify using medians rather than means for e.g. noisy data?

dnjohnstone: This method just comes from an old PhD thesis we had kicking around the office that I happened to be reading... They were worried about diffraction data and developed this method for that application and just set something that worked.

I think you're probably right on both fronts that it should be possible for the user to specify these parameters and probably you're right about medians and noise.

tjof2: Personally I would use medians over means as the default, but obviously test it first!!! :-)

dnjohnstone:
@francisco-dlp would really appreciate a green light (or an objection) from you on the overall idea here and the structuring that @bm424 implemented today before he or anyone else invests too much more time in it...

bm424:
@dnjohnstone @francisco-dlp or anyone else, any tips on how to go about writing tests for interactive features? I'm struggling...

dnjohnstone:
As far as I'm aware most of the interactive bits of the code are pretty sparsely (i.e. not) tested... I don't know what others consider the rules to be on this though.

bm424:
To wrap this up I'd appreciate opinions on two things:

1. How to test interactive elements. I've run out of googling capacity on this one.
2. Should the `PeakFinder2D` class be put in a separate part of the source, maybe `misc` or `utils`?

magnunor:
@bm424, as far as I know we do not have any automatic way of testing interactive functions. So you do not have to worry about that.

Semi-offtopic: should we have a list of interactive functions somewhere? So we can at least test them before major releases?

vidartf:
@bm424 If by "interactive" you mean mouse-clicking in a window, it can be simulated, but preferably you could trigger callbacks manually (with mock data if needed), e.g. instead of clicking a button you just call the `on_click` method. Dragging effects are the most difficult to simulate as they are often timing sensitive, but if you only want to see what happens when a value change, this should be possible (traitlets shouldn't care where the value change is coming from?).

Regarding structuring of the interactive code, there is currently no easy way to replace the ipywidgets interface with e.g. a Qt or traitsui one. This is due to the fact that all the UI specific code is in the base class. While this could be solved by refactoring to use metaclasses, it would probably be easier to have the base class that is simply for delegating `find_peaks` to the proper (non-member) function, like this:

```python

class PeakfinderBase:
    def init_ui(self):
        raise NotImplementedError()

    def find_peaks(self):
        if self.type == "typeA":
            find_peaks_typeA(self)
        elif self.type == "typeB":
            find_peaks_typeB(self)

class PeakfinderIPYW(PeakfinderBase)
    def __init__(self):
        super()
        self.init_ui()

    def init_ui(self):
        # do all the IPYWidget specific stuff here
```
Then the various sub-classes would simply be consumers of the `find_peaks` function, and you would still have a common base class for handling the object.

bm424:
@vidartf I can see the benefit in abstracting out the UI - I'll go ahead and do that, maybe the UI can get access to the parameters via `__code__` on the peak finder functions, though that seems a *bit* hacky.

As for the testing, I wasn't even really thinking about simulating user input! What I was struggling with was that if I tried to run any widget creations in `TestFindPeaks2D`, errors were thrown all over the place because it was being run outside a notebook.

francisco-dlp:

ericpre:
Re-submitted in #1292.

PR #1102, opened 2016-06-24
NEW lazy signal

to266:
Adds a new class of "Lazy" signals that only actually operate on data (and even access it) when required or explicitly told to do so.

The intended workflow is as follows:

1. Load the data lazily.
2. Perform calculations (will most likely be fast / instantaneous) or operations
3. Save the result signal to a file. Then the operations are performed by `dask`, and results written to file.

One of the drawbacks is that the `dask.array.Array` (which all this PR relies upon) does not support slicing assignment (i.e. `a[3] += 2` does not work, have to do something like `a = da.concatenate(a[:3], a[3]+2, a[3:])`), so not everything works as seamlessly.

Also, lazy signals mean plotting is slow (since it's calculated on requests), but any-size data should be feasible


francisco-dlp:
Could you resubmit it to RELEASE_next_minor?

PR #1104, opened 2016-06-26
Fixes #1095

### LINKED ISSUE #1095: BUG export decomposition results calls image and spectrum signals
Hyperspy version 0.8.5
(This also occurred with factors):

```python
ImportErrorTraceback (most recent call last)
<ipython-input-8-b1971df6874d> in <module>()
----> 1 sW.export_decomposition_results(factor_format='msa',loading_format='tif')

/home/smc204/anaconda2/envs/hyperspy/lib/python3.5/site-packages/hyperspy/signal.py in export_decomposition_results(self, comp_ids, folder, calibrate, factor_prefix, factor_format, loading_prefix, loading_format, comp_label, cmap, same_window, multiple_files, no_nans, per_row, save_figures, save_figures_format)
    951             same_window=same_window,
    952             no_nans=no_nans,
--> 953             per_row=per_row)
    954 
    955     def export_bss_results(self,

/home/smc204/anaconda2/envs/hyperspy/lib/python3.5/site-packages/hyperspy/signal.py in _export_loadings(self, loadings, folder, comp_ids, multiple_files, loading_prefix, loading_format, save_figures_format, comp_label, cmap, save_figures, same_window, calibrate, no_nans, per_row)
    424                          per_row=3):
    425 
--> 426         from hyperspy._signals.image import Image
    427         from hyperspy._signals.spectrum import Spectrum
    428 

ImportError: No module named 'hyperspy._signals.image'
```

francisco-dlp:
Fixes #1095

### DIFF
--- a/hyperspy/Release.py
+++ b/hyperspy/Release.py
@@ -26,7 +26,7 @@ name = 'hyperspy'
 # When running setup.py the "+dev" string will be replaced (if possible)
 # by the output of "git describe" if git is available or the git
 # hash if .git is present.
-version = "0.8.5"
+version = "0.8.6+dev"
 description = "Multidimensional data analysis toolbox"
 license = 'GPL v3'
 
--- a/hyperspy/signal.py
+++ b/hyperspy/signal.py
@@ -275,8 +275,7 @@ class MVATools(object):
                         vector_scale=1,
                         no_nans=True, per_row=3):
 
-        from hyperspy._signals.image import Image
-        from hyperspy._signals.spectrum import Spectrum
+        from hyperspy.signals import Spectrum, Image
 
         if multiple_files is None:
             multiple_files = preferences.MachineLearning.multiple_files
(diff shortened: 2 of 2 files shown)


to266:
Why did you choose to use the old signals? They will deprecate, and we should not use them internally any more

francisco-dlp:
Not to break the API until 1.0 in case that we decide to release 0.8.6 at the same time as 1.0.

to266:
Aren't these only used internally? Or are you worried that the returned signal class (which I argue should be `SignalxD`, but currently is `Spectrum`/`Image`) will break something?

francisco-dlp:
Yes, the point is that the method should return the same subclass for the 0.8 series. `Spectrum` (counter-intuitively) inherits all the `Signal2D` methods. This has been like this by design since the first release, so it is possible that somebody's code is relying on this feature.

to266:
Oh I see, makes sense.

Issue #546, opened 2015-05-12
NEW: Intensity axes


vidartf:
Currently there is no way (that I know of) to store a name or units for the "intensity axes" of data, e.g. "counts" for EDS data. In many cases this is not needed, but there is already some places in the code where axes titles are added manually to plots (`Signal.plot_explained_variance_ratio` and `Signal.plot_cumulative_explained_variance_ratio` sets it as "Explained variance ratio" and "Cumulative explained variance ratio" respectively). Another example where it would be useful could be when tracking positions, e.g.:

``` python
# Setup a mock signal with a moving gaussian with noise:
import hyperspy.hspy as hs
g = hs.components.Gaussian()
data = []
for i in xrange(2, 9):
    g.centre.value = float(i)
    data.append(g.function(np.linspace(0.0, 10.0)))
data = np.array(data)
data += 0.4*np.random.rand(*data.shape)
s = hs.signals.Spectrum(data)

# Setup energy axes to mimic spectroscopy data
s.axes_manager[1].scale = 10.0/50
s.axes_manager[1].unit = 'keV'
s.axes_manager[1].name = 'Energy'

# Track the position of the peak by using a gaussian component
m = hs.create_model(s)
m.append(g)
m.multifit()

# Finally plot the position as a function of the navigation axes
spos = g.centre.as_signal()
spos.plot()
# Here y-axis has the right values (position of peak along signal axes), but 
# units and name is lost, and mislabeled as 'Intensity'
```

I don't know how this would be best solved, but maybe a field on `AxesManager` would be appropriate. As the intensity axes doesn't need scale/size/offset (this is normally figured out automatically when plotting) I don't think we need a full `DataAxis`, but it might be worth considering.

francisco-dlp:
What about simple adding it to the metadata? e.g.:

```python
metadata.Signal.quantity = "position"
metadata.Signal.units = "nm"

```

vidartf:
Well, as most of the plotting code doesn't actually have access to the parent `Signal`, I figured `AxesManager` to be the intuitive place. Also, as far as I see it `AxesManager` is responsible for laying out the data into unit space. Further, for certain data, it is sometimes useful to switch the intensity and signal axes (i.e. go to f = x(y) instead of f = y(x)). If so, the intensity axes should be converted into a non-linear DataAxis (after sorting), and this would be a lot easier if it was already on `AxesManager`.

francisco-dlp:
[...]

PR #1106, opened 2016-06-26
Scale/units conversion

ericpre:
This pull request aims at avoiding displaying a 1E-9 m or 1000 nm scale bar on images by converting the scale and units when a file is open or data are cropped. Diffraction pattern, EDS and EELS spectra are also converted.
The convert_units() method of the axes manager allows converting the units manually.

### DIFF
--- a/anaconda_hyperspy_environment.yml
+++ b/anaconda_hyperspy_environment.yml
@@ -24,8 +24,8 @@ dependencies:
 - traits
 - traitsui
 - ncurses
-- pint
 - pip:
   - python-dateutil
   - tqdm
   - hyperspy
+  - pint
--- a/anaconda_hyperspy_environment.yml
+++ b/anaconda_hyperspy_environment.yml
@@ -23,6 +23,8 @@ dependencies:
 - ipywidgets
 - traits
 - traitsui
+- ncurses
+- pint
 - pip:
   - python-dateutil
   - tqdm
--- a/hyperspy/io_plugins/blockfile.py
+++ b/hyperspy/io_plugins/blockfile.py
@@ -126,7 +126,7 @@ def get_header_from_signal(signal, endianess='<'):
     if signal.axes_manager.navigation_dimension == 2:
         NX, NY = signal.axes_manager.navigation_shape
         SX = signal.axes_manager.navigation_axes[0].scale
-        SY = signal.axes_manager.navigation_axes[0].scale
+        SY = signal.axes_manager.navigation_axes[1].scale
     elif signal.axes_manager.navigation_dimension == 1:
         NX = signal.axes_manager.navigation_shape[0]
         NY = 1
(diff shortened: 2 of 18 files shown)


francisco-dlp:
This is really great! Could you update it with master?

I am not sure if I agree with the automatic unit conversion (e.g. when cropping) as users may like to keep the original units for whatever reason? Same for loading/saving except for the case where a file format requires specific units.

This feature is very EM-centric. Is there any way to make it more general, maybe making use of pint's metric prefixes?

It was in the plans to add `metadata.Signal.units` and use pint to deal with units during operations. I was hoping to get this feature by the release of HyperSpy 1.0, but, as the release should happen today, that's not going to happen.

francisco-dlp:
In particular I was thinking on #546

ericpre:
There are a few things to discuss in this PR.

> automatic unit conversion (e.g. when cropping)

I will add an argument to the crop method with a default that keeps the original unit. Would it be worth adding a preference setting to set the default behavior?

> EM-centric?

Actually, this is exactly that I tried to avoid and I don't see how it is EM-centric. As it is now, a few io reader (tiff and fei) assume a default unit ('nm'), while with this PR, the unit is determined as a function of pixel number and pixel size. For example, it should convert to 'km' when appropriate!!

I will have a look at #546. Would it be simpler to review in a separate PR?

anaconda_hyperspy_environment.yml
```diff
@@ -23,7 +23,9 @@ dependencies:
 - ipywidgets
 - traits
 - traitsui
+- ncurses
```
francisco-dlp: Why is ncurses back?

hyperspy/io_plugins/blockfile.py
```diff
@@ -126,7 +126,7 @@ def get_header_from_signal(signal, endianess='<'):
     if signal.axes_manager.navigation_dimension == 2:
         NX, NY = signal.axes_manager.navigation_shape
         SX = signal.axes_manager.navigation_axes[0].scale
-        SY = signal.axes_manager.navigation_axes[0].scale
+        SY = signal.axes_manager.navigation_axes[1].scale
```
francisco-dlp: You're actually fixing a bug here, could you submit the fix in a different PR?

francisco-dlp:
When I mentioned that this is too EM-centric I meant that this only performs the job for length units and energy in eV. I've just checked pint's documentation and it seems that their `to_compact` method does part of what this PR implements in a more general way. e.g.

```python
 >>> (20e-9 * ureg.meter).to_compact()
<Quantity(19.999999999999996, 'nanometer')>
```

I feel very uncomfortable with the idea of letting HyperSpy change automatically the units of the axes, but I totally see your point: it's a lot more human friendly to change it for e.g. plotting. Then, why not changing it automatically only where needed e.g.

* For plotting: the plotting functions could plot using the compact units without actually changing the units in the axes. One could even think on doing this automatically when zooming. This behaviour could be controlled with a new entry in preferences.
* Signal slicing could support units e.g. `s.inav[":30. nm"]`. This could be really cool!

Beside,  `DataAxis` could have a method (similar to the one that you've implemented already) to change the units with an option to changing them to the output of pint's compact.

francisco-dlp:
The syntax for slicing could also be improved by using pint itself e.g. `s.inav[:30 * ureg.nanometer]`.

Also, following #1170, could you send this to RELEASE_next_minor?

ericpre:
OK thanks, I understand your point now and it sounds good! :)
to_compact() is exactly, what I was looking for... I don't how I miss it...

francisco-dlp:

PR #1107, opened 2016-06-27
more useful axesmanager repr string

to266:
```python
import numpy as np
   ...: 

In [2]: s = hs.signals.Signal2D(np.random.rand(10, 11, 12, 13))

In [3]: s.axes_manager
Out[3]: 
<Axes manager, axes: (11, 10|13, 12)>
            Name |   size |  index | offset |  scale |  units 
================ | ====== | ====== | ====== | ====== | ====== 
     <undefined> |     11 |      0 |      0 |      1 | <undefined> 
     <undefined> |     10 |      0 |      0 |      1 | <undefined> 
---------------- | ------ | ------ | ------ | ------ | ------ 
     <undefined> |     13 |        |      0 |      1 | <undefined> 
     <undefined> |     12 |        |      0 |      1 | <undefined> 
```

### DIFF
--- a/hyperspy/axes.py
+++ b/hyperspy/axes.py
@@ -999,17 +999,17 @@ class AxesManager(t.HasTraits):
         ag = tuple(ag)
         self.edit_traits(view=tui.View(*ag), context=context)
 
-    def _get_axes_str(self):
-        string = "("
-        for axis in self.navigation_axes:
-            string += axis.__repr__() + ", "
-        string = string.rstrip(", ")
-        string += "|"
-        for axis in self.signal_axes:
-            string += axis.__repr__() + ", "
-        string = string.rstrip(", ")
-        string += ")"
-        return string
+    # def _get_axes_str(self):
+    #     string = "("
+    #     for axis in self.navigation_axes:
+    #         string += axis.__repr__() + ", "
+    #     string = string.rstrip(", ")
+    #     string += "|"
+    #     for axis in self.signal_axes:
+    #         string += axis.__repr__() + ", "
+    #     string = string.rstrip(", ")
+    #     string += ")"
+    #     return string
 
     def _get_dimension_str(self):
         string = "("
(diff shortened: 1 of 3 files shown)


francisco-dlp:
This needs a test e.g. simply calling the method would do.

Beside, how hard would it be to implement a repr_html version of the for the notebook?

francisco-dlp:
html repr seems broken at present. In case that it serves as inspiration, [this is how they do it in pandas](https://github.com/pydata/pandas/blob/20dd17a70063387ccca5f7aca3dbefc904d99457/pandas/formats/style.py#L42).

Beside, could you add a test which simply call the __repr__ and __repr_html__ methods just to make sure that at least they don't badly break?

to266:
What do you mean it's broken? Seems to work for me, although it does not look very good with the separator of the two spaces....

I'll try to find a better separator for the two spaces, but if nothing works, I'll just remove the html repr - I'll leave it for someone else.

By the way they use `jinja` for the templates... And that's understandable, it's often the most sensible way to look at some table data - to just write it out. Not sure it's worth all this effort for the `axes_manager`

francisco-dlp:
Yes, I meant the separator issue

axes_manager_html_repr

to266:
Huh, looks like the notebook does not render _all_ html - at least not correctly - I am not able to make one of the lines bolder or change it's color, even though the code is definitely correct, as tested in here 
https://jsfiddle.net/tcwnR/172/

francisco-dlp:
Are you going to do something about it?

to266:
Should I remove the `_repr_html_`?

I could not find a way to make the separator nice, since notebook does not render different color or thickness lines, apparently....

Pandas does not use that, so they're fine. Two tables would be a solution... What do you think?

francisco-dlp:
+1 for two tables

francisco-dlp:
It still looks funny:

am_repr_html

francisco-dlp:
Actually, in Chrome it looks ok.

to266:
The simple tests added. Let me know if there is anything else required from my side

PR #879, opened 2016-02-22
Fix replotting

vidartf:
Attempt at fixing #878. A bit of a blind attempt since I was unable to reproduce error.

PR #1121, opened 2016-07-07
Resubmit of "fix replotting" 

to266:
Resubmit of https://github.com/hyperspy/hyperspy/pull/879, the original commits follow:

 - Fix delayed call of plot `_on_close()`

 - This event doesn't need connecting/disconnecting

 - As `Signal.update_plot()` has in-built checks, it is easier to simply
always have it connected.

 - Clean up plotting

 - Re-add direct call

 - Removed stray code line

 - Typos

### DIFF
--- a/hyperspy/drawing/figure.py
+++ b/hyperspy/drawing/figure.py
@@ -1,4 +1,5 @@
 import textwrap
+import matplotlib.pyplot as plt
 
 from hyperspy.events import Event, Events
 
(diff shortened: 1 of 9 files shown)


vidartf:
Only one of the commits made the jump. WIP?

vidartf:
Or is it squashed?

to266:
squashed, sorry :)

vidartf:
No worries, just looked strange, so I was concerned that something had dropped out :)

vidartf:
The changes does of course look OK to me since I originally wrote them, so someone else should probably review them :smile:

to266:
FYI it looks OK to me too, and I'm not even the author! :D

thomasaarholt:
Is there anything more needed here?

to266:
review from a 3-rd party ( @francisco-dlp ?)

francisco-dlp:
I only disagree with "This event doesn't need connecting/disconnecting" (b35b4aeee67401599). Although `update_plot` does check if the plot exists or not, `update_plot` needs to be refurbish and, once we do it, we might forget that we're relying on that property and introduce a bug. As long as disconnecting it on close does not cause any issue, +1 on disconnecting it.

to266:
added back

PR #1122, opened 2016-07-08
The x axis of the signal1D plot was not animated

francisco-dlp:
The x-axis was not updated because it was not animated. This fixes the issue which is specially apparent when using widgets.


hyperspy/drawing/signal1d.py
```diff
@@ -354,9 +355,7 @@ def update(self, force_replot=False):
         try:
             self.ax.hspy_fig._draw_animated()
         except:
-            pass
-        # self.ax.hspy_fig._draw_animated()
-        # self.ax.figure.canvas.draw_idle()
+            raise
```
to266: the `try`/`except` is not actually doing anything here any more, right?

francisco-dlp:
Fixed, thanks.

PR #1123, opened 2016-07-09
Collection of enhancement and fixes to widgets and interactive

### LINKED ISSUE #1120: Better error message for adding a widget when no plot open
If you use the add_widget method but do not have a plot window open you get a potentially confusing error message that tells you "ax is not defined" -- we should probably make the code raise an error saying you can't add a widget unless there is a plot open... alternatively it could automatically plot the thing.

francisco-dlp:
* Restore the ROI and interactive User Guide documentation which got lost in #1033 
* Fix #1120 
* Add `color` argument to widget.interactive
* Enhances and fixes to widgets and interactive documentation
* Remove duplicated trigger.
* Fix `recompute_out` which was not updating the `size` attribute of the axes.
* Add the ability to connect more than one event to interactive.
* Add the ability to not connect any event to `interactive`. Default is "auto".


doc/user_guide/tools.rst
```diff
@@ -993,3 +993,201 @@ for example:
...
+
+.. code-block:: python
+
+    >>> s = signals.Signal1D(np.arange(10))
+    >>> s_sum = s.sum(0)
+    >>> s_sum.data
+    array(45)
+    >>> s.inav[:5].sum(0, out=s_sum)
+    >>> s_sum.data
+    10
```
to266: this better not be correct, since we should force the `data` to be at least an array :)

hyperspy/interactive.py
```diff
@@ -34,20 +44,26 @@ def __init__(self, f, event=None,
         f: function or method
             A function that returns an object and that optionally can place the
             result in an object given through the `out` keyword.
-        event: {Event | None}
+        event: {Event, "auto", None, iterable of events}
             Update the result of the operation when the event is triggered.
-            If None and `f` is a method of a Signal class instance its
+            If "auto" and `f` is a method of a Signal class instance its
             `data_changed` event is selected is the function takes an `out`
```
to266: ..__if__ the function takes.. ?

What if it does not, and 'auto'?

to266:
Except for my two comments, LGTM

francisco-dlp:
Thanks, fixed hence I'll merge.

PR #1125, opened 2016-07-10
In Python 3 we don't need to worry about the qt api

francisco-dlp:


to266:
appveyor is not happy... and ideas why?

francisco-dlp:
appveyor is randomly failing like that lately.

Issue #1103, opened 2016-06-25
Remove simulation subclasses


francisco-dlp:
Currently there are four simulation signal subclasses: `Simulation`, `EELSSpectrumSimulation`, `ImageSimulation`, `SpectrumSimulation`. However, these classes are mostly empty, except for a couple of methods which simply add noise to the data. Here I propose to simply remove them as, in my opinion, these subclasses make little sense. HyperSpy is about multi dimensional data analysis—why analysing a simulated signals should be any different from analysing experimental ones?

Ones #1058 is merged I propose to submit a PR to remove these classes, unless we decide here otherwise.

tjof2:
I think this sounds like a good idea, especially since `EELSSpectrumSimulation` could lead people to think HyperSpy does DOS calculations...! 

That said, keeping the functions to add noise, to e.g. test denoising methods, would be useful. Perhaps moved elsewhere?

francisco-dlp:
Implemented in #1126.

PR #1126, opened 2016-07-10
Remove simulation classes

francisco-dlp:
Remove simulation classes as discussed in #1103.


hyperspy/io.py
```diff
@@ -326,19 +306,13 @@ def assign_signal_subclass(record_by="",
...
     signals = hyperspy.misc.utils.find_subclasses(hyperspy.signals, BaseSignal)
 
-    if signal_origin == "experiment":
-        signal_origin = ""
-
     preselection = [s for s in
                     [s for s in signals.values()
                      if record_by == s._record_by]
-                    if signal_origin == s._signal_origin]
+                    ]
```
to266: This preselection is now meaningless - just passes all `signals.values()`

to266:
Oops, forgot I commented... It's not important though, and it will have to be updated with lazy signals at some point

francisco-dlp:
No worries, that function has been refurbished in #1058

francisco-dlp:
appveyor is randomly failing like that lately.

francisco-dlp:
I've triggered another build, but if it fails I don't think that there is anything to worry about as this PR is very straightforward.

Issue #974, opened 2016-05-04
record_by attributes


dnjohnstone:
The code base currently includes a "record_by" attribute that can take values "spectrum", "image" or " ", and is used throughout the internal workings to switch between Spectrum, Image and Signal classes. As discussed in #963 this is a legacy 'feature' that inherits from the dogma of electron microscopy packages that allowed acquisition "spectrum-by-spectrum" (EELS map) or "image-by-image" (EFTEM series).

This terminology is not however particularly general and I think it would be preferable if we have more specific "recorded_as" attributes for specialised sub-classes... This might be more self-documenting and take values like "spectrum image", "image series", "tlit series" etc.

Opinions on the best general solution would be gladly received.

francisco-dlp:
Now that @dnjohnstone  is working on #963, it is indeed the right time to address this and related issues such as  #976. It should be achievable for `1.0`, so I've added to that milestone.

to266:
...Doesn't the attribute only show the number of dimensions in what we call "signal" space? Then the only (albeit a large one) problem is that people might confuse "signal space" (as in the 3 in `< 2 | 3 >`) and the actual signal.... We could, of course, instead store the number navigation space dimensions - which is much less confusing, however not "intrinsic" to the type of signal.

Actually now that I think of it, I would prefer storing the confusing "signal space dimensions". I guess the confusion would only come because at some point there was the need to separate just "plain data" and our "signal", which is essentially "data + tools wrapper" - hence the term was born and had to be used. Especially to stress the idea that you __can__ access the data directly if you know what you're doing and want speed.

dnjohnstone:
I agree that the thing that this attribute has come to store is essentially the number of axes/dimensions in the 'signal space'... and I think that this is the right thing to store rather than navigation.

To me the attribute could be more self documenting and reflect the fact that at this point it doesn't really matter how you "recorded" the data it's just a choice about how you 'slice the data' in analysis, or in other words which dimensions you take as signal dimensions.

What I would propose is that in this case the attribute should be "signal_dimensions" potentially allowing any integer value although, to be fair, for the time being it's hard to imagine each individual measurement being made on anything other than a 1D or 2D detector... so perhaps it's a non issue.

In the end I guess it comes down to whether we think that the average user will gain more from the current terminology, which retains some link to how they probably did the experiment or rather more from being a little irreverent to that and focusing on which axes you're doing your analysis on (i.e. which you set as signal)

I'm really not sure what is 'right' here.

[...]

PR #1127, opened 2016-07-11
Bye bye record_by

francisco-dlp:
This removes the `record_by` attribute as discussed in #974.

This also changes the default signal dimension of `BaseSignal` from 1 to `data.ndim`.

TODO
--------

* [x] Amend User Guide
* [x] Add tests for nD hdf5
* [x] Save record_by in hdf5 to keep format backward compatible


francisco-dlp:
Although this is still WIP, it'll be nice if somebody could have a look at this stage in case that I am missing something.

sem-geologist:
will this require to redo io plugins?

francisco-dlp:
This is now ready for review.

to266:
needs merging of master

francisco-dlp:
Done

doc/user_guide/metadata_structure.rst
```diff
@@ -70,7 +70,6 @@ in the following sections of this chapter.
         │   │   └── parameters_estimation_method
         │   └── variance
         ├── binned
-        ├── record_by
         ├── signal_type
         └── signal_origin
```
CodeMonkeyJan: Not directly related to this PR, but shouldn't `signal_origin` be also gone?
Also later (line 293), a deprecation warning could be added to `signal_origin` in the explanations

hyperspy/io.py
```diff
@@ -315,26 +291,26 @@ def assign_signal_subclass(dtype,
...
         raise ValueError('Data type "{}" not understood!'.format(dtype.name))
-    if record_by and record_by not in ["image", "spectrum"]:
-        raise ValueError("record_by must be one of: None, empty string, "
-                         "\"image\" or \"spectrum\"")
+    if not isinstance(signal_dimension, int) or signal_dimension < 0:
+        raise ValueError("signal_dimension must be a positive interger")
     signals = hyperspy.misc.utils.find_subclasses(hyperspy.signals, BaseSignal)
     d_matches = [s for s in signals.values() if dtype == s._dtype]
-    d_r_matches = [s for s in d_matches if record_by == s._record_by]
+    d_r_matches = [s for s in d_matches
```
CodeMonkeyJan: A very minor nitpick: I chose the very short names because of the (then) long list of tested attributes:
- d: `dtype`
- r: `record_by` (now `signal_dimension`)
- o: `signal_origin` (now gone)
- t: `signal_type`

After `signal_origin` is gone, maybe we should go back to more verbose names like `dtype_matches`, `dtype_dim_matches` and `dtype_dim_type_matches`, which may be easier to read for new people.

hyperspy/tests/io/test_hdf5.py
```diff
@@ -289,6 +289,25 @@ def tearDown(self):
         remove(self.filename)
 
 
+class TestAxesConfiguration:
+
+    def setUp(self):
+        self.filename = 'testfile.hdf5'
+        s = BaseSignal(np.zeros((2, 2, 2, 2, 2)))
+        s.axes_manager.signal_axes[0].navigate = True
+        s.axes_manager.signal_axes[0].navigate = True
```
CodeMonkeyJan: Do I understand correctly that now, creating a `BaseSignal` with a numpy array and no additional information will default all axes to `signal_axes` instead of only the last one, like it was before?
Because for me these three lines do not work, because the last line has no signal axis to convert anymore.

doc/user_guide/metadata_structure.rst
francisco-dlp: I thought on leaving the metadata as the information is useful. We just don't take it into account when assigning the signal subclass.

hyperspy/io.py
francisco-dlp: Fixes, thanks!

hyperspy/tests/io/test_hdf5.py
francisco-dlp: Yes, now it defaults to all axes being signal axes. Does this need fixing then?

CodeMonkeyJan: No, I was just wondering, because I didnt' see it in the documentation at first. As I said I personally like this approach much more than the previous one (why should the last axis be special by default?).
Overall (again, just my opinion), it's more intuitive now :-)

CodeMonkeyJan:
On my system I did not get the `test_decomposition_not_supported` error, if this helps...
(Instead I got two errors concerning Cython, which seem to stem from my  system and not HyperSpy itself).

sem-geologist:
@CodeMonkeyJan , could you precise that kind of error?

CodeMonkeyJan:
Indeed, I also get the errors with the current master state, so it does not have anything to do with this PR, sorry for the confusion. Should I open an issue for that?

sem-geologist:
@CodeMonkeyJan , do you have cython and compiler installed?

CodeMonkeyJan:
Yes and I found the problem: I never really installed HyperSpy, I just have the raw source code in my repository (because up till now it was not necessary to build anything).
After running
```python
python setup.py build_ext --inplace```
everything works fine and i do not get errors.

Again, sorry for the confusion...

sem-geologist:
However I wonder what this `test_decomposition_not_supported` error is about, it passes on win64 appveyor, but not travis *nix'es and win32... that smells as random error caused by some compiled problem by 3rd party lib.

hyperspy/tests/io/test_hdf5.py
francisco-dlp: You are right, this wasn't documented. I have just made some improvements to this part of the docs.

PR #1128, opened 2016-07-12
Robust PCA & Online RPCA algorithms

tjof2:
This PR takes over from #1071, which I broke.

----

Robust PCA seems to be making its way into scikit-learn, slowly... (see #972), but with one of the slow, iterative, memory-intensive SVT algorithms. However, I've been using this example of "online" robust PCA for a while now for outlier detection and working with missing data, which is much less memory-intensive and actually isn't too slow.

Since OR-PCA actually returns a low-rank component **L** and a sparse error component **S** (where the observations **X** = **L** + **S**), to fit this into the current HyperSpy decomposition framework, I carry out a subsequent SVD on **L** to return factors and loadings.

Although I've not currently implemented it in a way that allows for it, you can actually run OR-PCA in a truly online manner, i.e. update the estimated factors and loadings with each new sample that arrives, without having to re-run over the whole dataset. However, I think this quickly becomes complicated in the current HyperSpy machine learning framework, so consider this implementation more aimed at outlier detection.

You can see an example here: https://github.com/tjof2/hyperspy/blob/my-demos/examples/Online%20Robust%20PCA.ipynb

- [x] Documentation
- [x] Write simple tests
- [x] Option to feed in an initial mask for the sparse component **S**. 
- [x] Add batch method (GoDec) for comparison with online version
- [x] Add new QR initialization method (discussion with @thomasaarholt)
- [x] Add stochastic gradient descent (SGD) method - faster for large datasets (high-dimensional)


tjof2:
Once #382 is implemented, RPCA can also be used for foreground/background separation, for example of movies of atoms moving around on substrates.

doc/user_guide/mva.rst
```diff
@@ -166,11 +156,80 @@ To perform Poissonian noise normalization:
...
+
+PCA is known to be very sensitive to the presence of outliers in data. These outliers
+can be the result of missing or dead pixels, X-ray spikes, or very low count data.
+If one assumes a dataset to consist of a low-rank component **L** corrupted by
+a sparse error component **S**, then Robust PCA (RPCA) can be used to recover the
+low-rank component for subsequent processing.
+
+The default RPCA algorithm returns the factors and loadings of **L**, and can be
+accessed with the following code. You must set the "output_dimension" when using
+RPCA.
```
francisco-dlp: Could you add the reference for the RPCA algorithm? Also, as at sklearn they're implementing Candes' one, wouldn't it be better to call this one RPCA_<algorith surname> to differentiate them in case we decide to wrap sklearn's too?

And, given that rpca is an acronym, would't "RPCA" and "ORPCA" be better?

hyperspy/learn/rpca.py
```diff
@@ -0,0 +1,389 @@
...
+        is the [nfeatures x nsamples] Gaussian noise matrix
+    U, S, V : numpy arrays
+        are the results of an SVD on Xhat
+
+    Notes
+    -----
+    Algorithm based on the following research paper:
+       Tianyi Zhou and Dacheng Tao, "GoDec: Randomized Low-rank & Sparse Matrix
+       Decomposition in Noisy Case", ICML-11, (2011), pp. 33-40.
+
```
francisco-dlp: If this is a translation from MATLAB code, could you provide a link to the original code?

hyperspy/learn/rpca.py
```diff
@@ -0,0 +1,389 @@
...
+    The ORPCA code is based on a transcription of MATLAB code obtained from
+    the following research paper:
+       Jiashi Feng, Huan Xu and Shuicheng Yuan, "Online Robust PCA via
+       Stochastic Optimization", Advances in Neural Information Processing
+       Systems 26, (2013), pp. 404-412.
+
+    It has been updated to include a new initialization method based
+    on a QR decomposition of the first n "training" samples of the data.
+    A stochastic gradient descent solver is also implemented.
+
```
francisco-dlp: If this is a translation from MATLAB code, could you provide a link to the original code?

tjof2: Not in this case - I got the code via email from the author.

doc/user_guide/mva.rst
francisco-dlp: You are right, that's horrible! I'll fix the others in a separate PR.

hyperspy/learn/rpca.py
tjof2: Done.

doc/user_guide/mva.rst
tjof2: Ok, I'll change mine to uppercase.

francisco-dlp:
Could you update this branch with master?

tjof2:
@francisco-dlp done.

francisco-dlp:
I've just merged #382, could you update with master and update the code to take advantage of the new keyword arg?

tjof2:
@francisco-dlp yep done - I've set it to return a `tuple` of `(X, E)` - is that the appropriate way to do it? Or would `X, E` work?

francisco-dlp:
Looks good, could you document what it returns in the docstring?

tjof2:
@francisco-dlp which docstring - in `mva.py` or `rpca.py`?

francisco-dlp:
mva

tjof2:
Ok great! I will create a demo notebook when this is all done and I have a bit more time - it'll cover both outlier removal and background subtraction.

tjof2:
Hmm not sure whats going on with the tests.

tjof2:
@francisco-dlp I think this is all ok now - though the failed test seems to be a numerical overflow?

tjof2:
Hmm, still some errors associated with `TestReturnInfo`, e.g. "Input contains NaN, infinity or a value too large for dtype('float64')."

francisco-dlp:
The failure is unrelated to this PR, so merging :)

thomasaarholt:
Could we improve the documentation regarding the various algorithms? It's rather unclear what arguments the various algorithms in s.decomposition? should take.

francisco-dlp:
@thomasaarholt, you definitely have a point. Would you like to open a separate issue about it?

tjof2:
@thomasaarholt I also agree, after reading it over again! Related to a point @dnjohnstone raised about the machine learning docs in general. Do raise a separate issue!

amueller:
What's the reference paper for the algorithm you implemented? And have you compared against isolation forests?

tjof2:
@amueller  Not compared them, no. 

In the end I implemented the GoDec algorithm[1] for the batch case, and for the online case went with OR-PCA by Feng et al.[2]. You can find them both here: https://github.com/hyperspy/hyperspy/blob/RELEASE_next_minor/hyperspy/learn/rpca.py

[1] http://www.icml-2011.org/papers/41_icmlpaper.pdf
[2] https://papers.nips.cc/paper/5131-online-robust-pca-via-stochastic-optimization.pdf

amueller:
Thanks. What are you using them for? Do you have any examples?

tjof2:
@amueller still investigating the potential examples for electron microscopy, but I've tried it out on data with corrupted pixels (https://github.com/tjof2/hyperspy/blob/my-demos/examples/Online%20Robust%20PCA.ipynb) and also for the usual foreground-background separation.

Also looks pretty promising for standard PCA on large datasets where memory is an issue.

The main motivation though was for plugging it into a microscope and doing the decomposition in real-time. This is for the future though, not done yet!

amueller:
cool thanks :)

tjof2:
@amueller  no problem, happy to help!

PR #1134, opened 2016-07-13
Fix decomposition test

francisco-dlp:
Fixes the broken test and add test for all algorithms except MLPCA which takes too long.

### DIFF
--- a/hyperspy/tests/mva/test_decomposition.py
+++ b/hyperspy/tests/mva/test_decomposition.py
@@ -198,7 +198,8 @@ class TestReturnInfo:
         self.s = signals.Signal1D(np.random.random((20, 100)))
 
     def test_decomposition_not_supported(self):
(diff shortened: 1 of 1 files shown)


sem-geologist:
travis osx is taking time, it is saturated at the moment (164 -the max jobs, much less servers power than for linuxes), unfortunatelly this behaviour will be more often felt now after merging #949. appveoyr is slow and do one thing at time, but I saw never the outage, but travis osx servers tends sometimes to crash for whole day.

sem-geologist:
@francisco-dlp , from my experience, osx services in travis have crashed, thei will notice it after few hours, then in the end of the day it may work. I would suggest to move on as travis linux and win64 and win32 passed tests sucessfully.

francisco-dlp:
Agree, thanks.

Issue #1139, opened 2016-07-13
RELEASE 1.0


francisco-dlp:
TODO - RELEASE
--------------------------

* [x] Merge remaining PRs
* [x] Fix unfold
* [x] Merge component std
* [ ] Update demos
* [x] Create RELEASE_0.8.5 branch
    * [x] Update version
    * [x] Update changes
* [x] Create Python installers
* [x] Test Python installers
* [x] Create Windows installers
* [x] Test Windows installers
* [x] Create Release in GitHub (as pre-release)
* [x] Update DOI in ``README.rst`` and ``citing.rst``
* [x] Create Bundle installer
    * [x] Test it with admin rights
    * [x] Test with user without admin rights
* [x] Upload docs
* [x] Upload to pypi
* [x] Upload to GH
* [x] Update website
* [x] Announce in mailing list
* [ ] Backport hdf5 security issue fix

sem-geologist:
the bundle installer can be created in apveyor, also the windows wheels. Unless nsis is still offline, then bad luck... also travis does sdist and osx wheels. all that requires to replace my credentials with authorized ones for hyperspy main

sem-geologist:
... in both: travis and appveyor yml.

francisco-dlp:
"all that requires to replace my credentials with authorized ones for hyperspy main" could you provide details for this. @to266, I bet that we could use your credentials for this, could you submit a PR to change @sem-geologist's to yours?

sem-geologist:
in .travis.yml it is line 47
the new key should be generated in github, then used on linux travis comand line to encrypt, then it lands in that file instead of mine key

for appveoyr it is simplier:
I added my credentials under @to266 and @vidartf , if they are still actual it should be enought just comment out mine, and uncoment one of them. NOTE it is different key than on travis.
However appveyor key is required only for deployment to github, which I couldn't make work (Maybe it will simply work out with @to266  or @vidartf keys, as they probably have better configured it. I dont know. It is note very relevant as after appveoyr builds stuff it preserves it (for ever) as artifacts which can be downloaded.

Travis however needs credentials as it does not preserve those artifacts, and needs to deploy them to github.

francisco-dlp:
ok, thanks!

2016-07-13 17:32 GMT+01:00 Petras Jokubauskas <notifications@github.com>:

sem-geologist:
[...]

PR #1141, opened 2016-07-13
fix: readthedoc compiling

sem-geologist:
this should fix the issue with docs at #1139

it is simple snippet applied to the situation
as suggested by  the readthedocs.org faq

### DIFF
--- a/doc/conf.py
+++ b/doc/conf.py
@@ -19,7 +19,7 @@ class Mock(MagicMock):
     def __getattr__(cls, name):
             return Mock()
 
-MOCK_MODULES = ['numpy', 'h5py']
+MOCK_MODULES = ['scipy', 'matplotlib', 'traits', 'traitsui', 'numpy', 'h5py']
 sys.modules.update((mod_name, Mock()) for mod_name in MOCK_MODULES)
 
 import os
--- a/doc/conf.py
+++ b/doc/conf.py
@@ -12,6 +12,16 @@
 # serve to show the default.
 
 import sys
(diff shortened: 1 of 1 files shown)


sem-geologist:
I am not sure but there is very high probability that to the mocked list I should add also 'traits' as normally it also requires compiling... @francisco-dlp  should I add it?

francisco-dlp:
I think that we need to add at least 'scipy', 'matplotlib', 'numpy', 'traits', 'traitsui','h5py'

sem-geologist:
another important question would be: how to test it?

francisco-dlp:
The easiest way is to merge it, if it doesn't work qw can unmerge it ;)

sem-geologist:
ok, we are in trouble, it still tries to compile the numpy :/

sem-geologist:
maybe it is this:

> If such libraries are installed via setup.py, you also will need to remove all the C-dependent libraries from your install_requires in the RTD environment.

from the faq of readthedoc...  thing is before we implemented setuptools, the `install_requires` was not even understood, now by setup tools it is taken into consideration.

francisco-dlp:
[Still failing](https://readthedocs.org/projects/hyperspy/builds/4195919/)

This is not very urgent. We can fix it after releasing.

sem-geologist:
I will register to readthe docs, and will experiment on separate branches, it is waste of precious travis osx and appveyor build power experimenting on the pullrequests and master of hyperspy

PR #1143, opened 2016-07-13
fix: lxml absence blocking everything

sem-geologist:
remove the raising of exception during import
add raising an exception at reader function


sem-geologist:
oh I saw the suggestion only now,  I am closing this, better to implement it in __init__

PR #1144, opened 2016-07-13
fix: lxml dependancy

sem-geologist:
add lazy loading to hyperspy/io_plugins/__init__

### DIFF
--- a/hyperspy/io_plugins/__init__.py
+++ b/hyperspy/io_plugins/__init__.py
@@ -20,9 +20,9 @@
 import logging
 
 from hyperspy.io_plugins import (msa, digital_micrograph, fei, mrc, ripple,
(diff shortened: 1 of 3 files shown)


francisco-dlp:
Almost perfect. The tests fail if lxml is not installed, they should be skipped instead.

sem-geologist:
I dont get it why when test things?

francisco-dlp:
The bcf unittests still assume that you can import it. If you can't they fail. If lxml is not present, the bcf tests should be skipped.

sem-geologist:
is there any other format which would do that?

sem-geologist:
@francisco-dlp  I hope the changes will satisfy you

francisco-dlp:
Great, thanks!

PR #1150, opened 2016-07-14
Fix learning results

### LINKED ISSUE #1145: blind_source_separation returns an error after calling get_decomposition_model
I am not sure the following behavior is normal?

``` python
s = hs.signals.Signal1D(np.random.rand(10,15, 1024))
s.decomposition(True)
s2 = s.get_decomposition_model(9)
s.blind_source_separation(9)

<ipython-input-34-ef9c7adc7bb5> in <module>()
----> 1 s.blind_source_separation(9)

/home/eric/Python_prog/hyperspy/hyperspy/learn/mva.py in blind_source_separation(self, number_of_components, algorithm, diff_order, diff_axes, factors, comp_list, mask, on_loadings, pretreatment, **kwargs)
    528             if not hasattr(lr, 'factors') or lr.factors is None:
    529                 raise AttributeError(
--> 530                     'A decomposition must be performed before blind '
    531                     'source seperation or factors must be provided.')
    532 

AttributeError: A decomposition must be performed before blind source seperation or factors must be provided.
```

while the following works as usual:

``` python
s = hs.signals.Signal1D(np.random.rand(10,15, 1024))
s.decomposition(True)
s.blind_source_separation(9)
s2 = s.get_decomposition_model(9)
```

This is on master 60331b16.

francisco-dlp:
The learning results were lost when changing the signal subclass because `BaseSignal._load_dictionary` as ignoring them.

Also fixes `learning_results.summary` that was only writing to the log instead of printing to stdout when called by user.

Fixes #1145

### DIFF
--- a/hyperspy/learn/mva.py
+++ b/hyperspy/learn/mva.py
@@ -1138,12 +1138,15 @@ class LearningResults(object):
         if hasattr(self, 'output_dimension') and self.output_dimension \
                 is not None:
             self.output_dimension = int(self.output_dimension)
-        self.summary()
+        _logger.info(self._summary())
 
     def summary(self):
         """Prints a summary of the decomposition and demixing parameters
          to the stdout
         """
+        print(self._summary())
+
+    def _summary(self):
         summary_str = (
             "Decomposition parameters:\n"
             "-------------------------\n\n" +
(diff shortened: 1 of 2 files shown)


francisco-dlp:
As we need to release this morning, I'll merge this waiting waiting for review as it is a critical bug fix. I'll wait for travis of course.

PR #1152, opened 2016-07-14
NEW: HologramImage class with holography reconstruction functionality

woozey:
Since #1058 has introduced major structural changes this PR replaces #1079 .

The open question is if the separate `ElectronWaveImage` class (inheriting from `ComplexSignal2D`) needs to be created. In this PR I've added `rec_param` property that extracts holography reconstruction parameters from metadata. Those parameters are specific for electron wave reconstructed from off-axis electron holography. Later I will also add a function calculating projected charge density maps from electron wave. So what will be the opinion of community ( @francisco-dlp, @CodeMonkeyJan, @tjof2)?


CodeMonkeyJan:
Great that this is up again, I'll take a look as soon as I can :-). I also would like to include @FWin22 in this discussion.

doc/user_guide/electron_holography.rst
```diff
@@ -0,0 +1,95 @@
+Electron Holography
+*******************
+
+HyperSpy provides the user with two classes which can be used to process electron holography data:
+ 
+* :py:class:`~._signals.hologram_image.HologramImage`
+* :py:class:`~._signals.wave_image.ComplexSignal2D`
```
CodeMonkeyJan: This class is not specifically for electron holography, so you should remove it here (change the text around it accordingly).

doc/user_guide/electron_holography.rst
```diff
@@ -0,0 +1,95 @@
...
+
+Both inherit directly from the :py:class:`~._signals.signal2d.Signal2D` class and thus can use all of
+its functionality. The usage of both classes is explained in the following sections.
```
CodeMonkeyJan: This is also not true anymore, `ComplexSignal2D` and `Signal2D` are now parallel in the inheritence tree (see the diagram I added).

doc/user_guide/electron_holography.rst
```diff
@@ -0,0 +1,95 @@
...
+
+
+The HologramImage class
+=======================
+
+The :py:class:`~._signals.hologram_image.HologramImage` class is designed to hold images acquired via
+electron holography.
+
+To transform a :py:class:`~._signals.signal2d.Signa2D` (or subclass) into a
```
CodeMonkeyJan: This should be `~._signals.complex_signal2d.ComplexSignal2D` now.

doc/user_guide/electron_holography.rst
```diff
@@ -0,0 +1,95 @@
...
+(includes finding a side band in FFT, isolating and filtering it, recenter and calculate inverse Fourier transform)
+can be performed using :py:func:`~._signals.hologram_image.HologramImage.reconstruct_phase` method
+which returns a :py:class:`~._signals.wave_image.ComplexSignal2D` class, containing the reconstructed electron wave:
+
+.. code-block:: python
+
+    >>> wave_image = im.reconstruct_phase()
+
+To reconstruct the hologram with a reference wave it should be provided to the method either as Hyperspy's
+:py:class:`~._signals.signal2d.Signa2D` (or subclass) or as a nparray:
```
CodeMonkeyJan: Again, this should be `~._signals.complex_signal2d.ComplexSignal2D`, or of course an `ElectronWaveImage` if we decide to add it.

doc/user_guide/electron_holography.rst
```diff
@@ -0,0 +1,95 @@
...
+
+.. code-block:: python
+
+    >>> wave_image = im.reconstruct_phase(ref_image)
+
+The reconstruction parameters can be extracted from metadata of wave_image as follows:
+
+.. code-block:: python
+
+    >>> reconstruction_param = wave_image.rec_param
```
CodeMonkeyJan: If you add it to the metadata, shouldn't it be:
```python
reconstruction_param = wave_image.metadata.Signal.rec_param
```
(if you save it in `metadata.Signal`, of course)?

doc/user_guide/electron_holography.rst
```diff
@@ -0,0 +1,95 @@
...
+
+This option can be used for batch processing as the reconstruction_param can be provided to the reconstruction method as follows:
+
+.. code-block:: python
+
+    >>> wave_image1 = im1.reconstruct_phase(rec_param=reconstruction_param)
+
+Further analyses of the reconstructed wave (phase) can be done using :py:class:`~._signals.wave_image.ComplexSignal2D` class
+functionality (see bellow).
```
CodeMonkeyJan: If we add `ElectronWaveImage`, then we should change the names everywhere:
```python
ComplexSignal2D --> ElectronWaveImage
```
Otherwise this section has to go, as `ComplexSignal2D` is not specific to electron holography.

doc/user_guide/tools.rst
```diff
@@ -191,7 +191,8 @@ e.g. specialised signal subclasses to handle complex data (see the following dia
     +-------------------------------------------------------------------------+------------------+-----------------------+----------+
     |    :py:class:`~._signals.complex_signal2d.ComplexSignal2D`              |        2         |       -               | complex  |
     +-------------------------------------------------------------------------+------------------+-----------------------+----------+
-
+    |      :py:class:`~._signals.hologram_image.HologramImage`                |        2         |      hologram         |  real    |
```
CodeMonkeyJan: I would put it directly under `Signal2D`, so all signals are in a nice order.

doc/user_guide/electron_holography.rst
```diff
@@ -0,0 +1,95 @@
...
+
+To transform a :py:class:`~._signals.signal2d.Signa2D` (or subclass) into a 
+:py:class:`~._signals.wave_image.ComplexSignal2D` use:
+
+.. code-block:: python
+
+    >>> im.set_signal_type('ComplexSignal2D')
+
+
+Unwrap the phase
```
CodeMonkeyJan: This and the other functions of `ComplexSignal2D` have to go no matter how we decide about `ElectronWaveImage`, as they are already documented elsewhere.

hyperspy/signals.py
```diff
@@ -33,6 +33,9 @@
     DielectricFunction
         For dielectric function data with signal_dimension equal 1. The signal
         is unbinned by default.
+    HolographyImage
+        For 2D-images taken via electron holography. WaveImages can be
```
CodeMonkeyJan: Still has to be decided ;-). Just pointing it out, so we don't miss it later...

hyperspy/tests/signal/test_hologram_image.py
```diff
@@ -0,0 +1,71 @@
...
+                                 np.linspace(0, wave_image.rec_param[4]*2, self.img_size))
+        x_start = int(self.img_size/10)
+        x_stop = self.img_size-1-int(self.img_size/10)
+        phase_new_crop = phase_new[x_start:x_stop, x_start:x_stop]
+        phase_ref_crop = self.phase_ref[x_start:x_stop, x_start:x_stop]
+        nt.assert_almost_equal(phase_new_crop, phase_ref_crop, decimal=2)
+
+
+if __name__ == '__main__':
+    nose.run(argv=[sys.argv[0], sys.modules[__name__].__file__, '-v'])
```
CodeMonkeyJan: You can replace this with
```python
if __name__ == '__main__':
    import nose
    nose.run(defaultTest=__name__)
```
so you don't have to import `nose` or `sys` at the top (and it looks prettier).

CodeMonkeyJan:
I had a quick look and I have mainly a few nitpicks about the documentation. I think I said everything I said about the code itself in #1079 already, so it should be all in there :-).
The inclusion of `ElectronWaveImage`, in my opinion, depends on the functionality you and @FWin22 want to add. I assume it's specific enough to warrant a specialised class like `ElectronWaveImage`.

francisco-dlp:
Too tired today to go through this. I am just writing to say that chances are that we'll release HyperSpy 1.1 before the end of July. It'll be great if this could be ready by then.

doc/user_guide/electron_holography.rst
woozey: Nope, here it is more evolved: `rec_param` is a property having a wrapper that extracts the reconstruction parameters from metadata in a tuple that can be provided to the reconstruction function

hyperspy/signals.py
woozey: My bad using copy/paste. I put it together now as it would be only `ComplexSignal2D`, though will change it if we decide to do `ElectronWaveImage`

woozey:
> Too tired today to go through this. I am just writing to say that chances are that we'll release HyperSpy 1.1 before the end of July. It'll be great if this could be ready by then.

I should have some time to work on it intensively in the beginning of the week. Ideally would be to have some more comments to address during my travel today ~13:00 that I'm not bored in trains and plane)))

CodeMonkeyJan:
I talked with @FWin22 and due to the fact that he also prefers an `ElectronWaveImage` class, I think we should add it in this PR.

doc/user_guide/electron_holography.rst
```diff
@@ -15,7 +16,7 @@ The HologramImage class
 The :py:class:`~._signals.hologram_image.HologramImage` class is designed to hold images acquired via
 electron holography.
 
-To transform a :py:class:`~._signals.signal2d.Signal2D` (or subclass) into a
+To transform a :py:class:`~._signals.signal2d.Signa2D` (or subclass) into a
```
CodeMonkeyJan: Reverted to the missing "l"

doc/user_guide/electron_holography.rst
```diff
@@ -1,12 +1,13 @@
...
 
-HyperSpy provides the user with a class which can be used to process electron holography data:
+HyperSpy provides the user with two classes which can be used to process electron holography data:
  
 * :py:class:`~._signals.hologram_image.HologramImage`
+* :py:class:`~._signals.wave_image.ElectronWaveImage`
 
-It inherits directly from the :py:class:`~._signals.signal2d.Signal2D` class and thus can use all of
-its functionality. The usage is explained in the following sections.
+The classes inherit from :py:class:`~._signals.signal2d.Signal2D` and :py:class:`~._signals.electron_wave_image.ElectronWaveImage`
```
CodeMonkeyJan: Second one should be `ComplexSignal2D`, right?

doc/user_guide/electron_holography.rst
```diff
@@ -32,7 +33,7 @@ The detailed description of electron holography and reconstruction of holograms
 `[Joy1993] <http://dx.doi.org/10.1016/0304-3991(93)90130-P>`_. Fourier based reconstruction of off-axis holograms
 (includes finding a side band in FFT, isolating and filtering it, recenter and calculate inverse Fourier transform)
 can be performed using :py:func:`~._signals.hologram_image.HologramImage.reconstruct_phase` method
-which returns a :py:class:`~._signals.wave_image.ComplexSignal2D` class, containing the reconstructed electron wave:
+which returns a :py:class:`~._signals.wave_image.ElectronWaveImage` class, containing the reconstructed electron wave:
```
CodeMonkeyJan: Should be signals. __electron___ wave_image.ElectronWaveImage!

doc/user_guide/electron_holography.rst
```diff
@@ -57,5 +58,38 @@ This option can be used for batch processing as the reconstruction_param can be
 
     >>> wave_image1 = im1.reconstruct_phase(rec_param=reconstruction_param)
 
-Further analyses of the reconstructed wave (phase) can be done using the :py:class:`~._signals.wave_image.ComplexSignal2D` class
-functionality (see dedicated documentation).
\ No newline at end of file
+Further analyses of the reconstructed wave (phase) can be done using :py:class:`~._signals.wave_image.ElectronWaveImage` class
```
CodeMonkeyJan: Same as above!

doc/user_guide/electron_holography.rst
```diff
@@ -57,5 +58,38 @@ This option can be used for batch processing as the reconstruction_param can be
...
+functionality (see bellow).
+
+
+The ElectronWaveImage class
+===========================
+
+The :py:class:`~._signals.wave_image.ElectronWaveImage` class can hold information about the complex electron
```
CodeMonkeyJan: Same as above!

doc/user_guide/electron_holography.rst
```diff
@@ -57,5 +58,38 @@ This option can be used for batch processing as the reconstruction_param can be
...
+wave. As such, relevant properties like the `amplitude`, `phase` and the `real` and `imag` part can be
+directly accessed and return appropriate :py:class:`~._signals.signal2d.Signal2D` signals.
+
+To transform a :py:class:`~._signals.complex_signal2d.ComplexSignal2D` (or subclass) into a
+:py:class:`~._signals.wave_image.ElectronWaveImage` use:
+
+.. code-block:: python
+
+    >>> im.set_signal_type('ElectronWaveImage')
```
CodeMonkeyJan: @francisco-dlp: What was the naming convention we decided for the `signal_type` parameter again? All lower case, except for abbreviations like "EELS", riht? Thus this should be `'electron_wave'`, which would also be in accordance with `'hologram'` for the `HologramImage` class.

doc/user_guide/electron_holography.rst
```diff
@@ -57,5 +58,38 @@ This option can be used for batch processing as the reconstruction_param can be
...
+
+
+Unwrap the phase
+----------------
+
+With the :py:func:`~._signals.wave_image.ElectronWaveImage.get_unwrapped_phase` method the phase can be
```
CodeMonkeyJan: Same issue as above (signals. __electron_ wave___ image.ElectronWaveImage)!

doc/user_guide/electron_holography.rst
```diff
@@ -57,5 +58,38 @@ This option can be used for batch processing as the reconstruction_param can be
...
+
+To transform a :py:class:`~._signals.complex_signal2d.ComplexSignal2D` (or subclass) into a
+:py:class:`~._signals.wave_image.ElectronWaveImage` use:
+
+.. code-block:: python
+
+    >>> im.set_signal_type('ElectronWaveImage')
+
+
+Unwrap the phase
```
CodeMonkeyJan: As I said, everything from this point onwards is not specifically for `ElectronWaveImage`, just inherited from `ComplexSignal2D`, so it should not be documented here again. Put any method documentation which is distinct for `ElectronWaveImage` here.

doc/user_guide/tools.rst
```diff
@@ -193,6 +193,8 @@ e.g. specialised signal subclasses to handle complex data (see the following dia
     +-------------------------------------------------------------------------+------------------+-----------------------+----------+
     |    :py:class:`~._signals.complex_signal2d.ComplexSignal2D`              |        2         |       -               | complex  |
     +-------------------------------------------------------------------------+------------------+-----------------------+----------+
+    |      :py:class:`~._signals.hologram_image.ElectronWaveImage`            |        2         |       -               | complex  |
```
CodeMonkeyJan: Should be signals. __electron_wave_image__ .ElectronWaveImage!

hyperspy/_signals/electron_wave_image.py
```diff
@@ -0,0 +1,40 @@
...
+
+
+from hyperspy._signals.complex_signal2d import ComplexSignal2D
+
+
+class ElectronWaveImage(ComplexSignal2D):
+
+    """ComplexSignal2D subclass for electron wave images."""
+
+    _signal_dimension = 2
```
CodeMonkeyJan: This will be inherited from `ComplexSignal2D`, so you don't have to redefine it here.

hyperspy/_signals/electron_wave_image.py
```diff
@@ -0,0 +1,40 @@
...
+    _signal_type = 'ElectronWaveImage'
```
CodeMonkeyJan: Should be `'electron_wave'`, see above!

CodeMonkeyJan: Should be `'electron_wave'`, see above!

hyperspy/_signals/hologram_image.py
```diff
@@ -153,7 +153,7 @@ def reconstruct_phase(self, reference=None, rec_param=None, show_phase=False,
 
         wave = w_obj / w_ref
         wave_image = self._deepcopy_with_new_data(wave)
-        wave_image.set_signal_type('ComplexSignal2D')  # New signal is a complex image!
+        wave_image.set_signal_type('ElectronWaveImage')  # New signal is an electron wave image!
```
CodeMonkeyJan: Same as above!

doc/user_guide/electron_holography.rst
CodeMonkeyJan: That makes sense, thanks for the explanation :-)!

doc/user_guide/electron_holography.rst
```diff
@@ -0,0 +1,95 @@
...
+Reconstruction of holograms
+---------------------------
+The detailed description of electron holography and reconstruction of holograms can be found in literature
+`[Gabor1948] <http://www.nature.com/doifinder/10.1038/161777a0>`_,
+`[Tonomura1999] <http://www.springer.com/us/book/9783540645559>`_,
+`[McCartney2007] <http://dx.doi.org/10.1146/annurev.matsci.37.052506.084219>`_,
+`[Joy1993] <http://dx.doi.org/10.1016/0304-3991(93)90130-P>`_. Fourier based reconstruction of off-axis holograms
+(includes finding a side band in FFT, isolating and filtering it, recenter and calculate inverse Fourier transform)
+can be performed using :py:func:`~._signals.hologram_image.HologramImage.reconstruct_phase` method
+which returns a :py:class:`~._signals.wave_image.ElectronWaveImage` class, containing the reconstructed electron wave:
```
CodeMonkeyJan: Should be signals. __electron___ wave_image.ElectronWaveImage.

doc/user_guide/electron_holography.rst
```diff
@@ -0,0 +1,95 @@
...
+
+    >>> reconstruction_param = wave_image.rec_param
+
+This option can be used for batch processing as the reconstruction_param can be provided to the reconstruction method as follows:
+
+.. code-block:: python
+
+    >>> wave_image1 = im1.reconstruct_phase(rec_param=reconstruction_param)
+
+Further analyses of the reconstructed wave (phase) can be done using :py:class:`~._signals.wave_image.ElectronWaveImage` class
```
CodeMonkeyJan: Should be signals. __electron___ wave_image.ElectronWaveImage.

doc/user_guide/electron_holography.rst
```diff
@@ -0,0 +1,95 @@
...
+
+The ElectronWaveImage class
+===========================
+
+The :py:class:`~._signals.wave_image.ElectronWaveImage` class can hold information about the complex electron
+wave. As such, relevant properties like the `amplitude`, `phase` and the `real` and `imag` part can be
+directly accessed and return appropriate :py:class:`~._signals.signal2d.Signal2D` signals.
+
+To transform a :py:class:`~._signals.complex_signal2d.ComplexSignal2D` (or subclass) into a
+:py:class:`~._signals.wave_image.ElectronWaveImage` use:
```
CodeMonkeyJan: Same as above!

doc/user_guide/electron_holography.rst
```diff
@@ -0,0 +1,95 @@
...
+
+.. code-block:: python
+
+    >>> im.set_signal_type('ElectronWaveImage')
+
+
+Unwrap the phase
+----------------
+
+With the :py:func:`~._signals.wave_image.ElectronWaveImage.get_unwrapped_phase` method the phase can be
```
CodeMonkeyJan: Should be signals. __electron___ wave_image.ElectronWaveImage.

doc/user_guide/tools.rst
```diff
@@ -191,7 +193,8 @@ e.g. specialised signal subclasses to handle complex data (see the following dia
     +-------------------------------------------------------------------------+------------------+-----------------------+----------+
     |    :py:class:`~._signals.complex_signal2d.ComplexSignal2D`              |        2         |       -               | complex  |
     +-------------------------------------------------------------------------+------------------+-----------------------+----------+
-
+    |      :py:class:`~._signals.hologram_image.ElectronWaveImage`            |        2         |       -               | complex  |
```
CodeMonkeyJan: Should be signals. __electron_wave_image__ .ElectronWaveImage.

hyperspy/signals.py
```diff
@@ -33,6 +33,9 @@
     DielectricFunction
         For dielectric function data with signal_dimension equal 1. The signal
         is unbinned by default.
+    HolographyImage
+        For 2D-images taken via electron holography. Electron wave as
+        ComplexSignal2D can be reconstructed from them.
```
CodeMonkeyJan: Also add ElectronWaveImage here.

hyperspy/tests/signal/test_hologram_image.py
```diff
@@ -0,0 +1,69 @@
...
+        x_start = int(self.img_size/10)
+        x_stop = self.img_size-1-int(self.img_size/10)
+        phase_new_crop = phase_new[x_start:x_stop, x_start:x_stop]
+        phase_ref_crop = self.phase_ref[x_start:x_stop, x_start:x_stop]
+        nt.assert_almost_equal(phase_new_crop, phase_ref_crop, decimal=2)
+
+
+if __name__ == '__main__':
+    import nose
+    nose.run(defaultTest=__name__)
```
CodeMonkeyJan: Add an empty line at the end of the file (for PEP8 compliance).

CodeMonkeyJan:
Strangely enough, a lot of my comments counted as "outdated", so I had to repeat them on the "files changed" tab... Sorry for the possible spam...

doc/user_guide/electron_holography.rst
francisco-dlp: I think that we also decided that there was no need for "_" instead of " " and these are strings with no risk of becoming attributes. So, what about "electron wave"?

CodeMonkeyJan: Ah, yes, that makes sense (though we should change this in the EDS signals, too, then, right?).

francisco-dlp: Yes that's right. I forgot to add this to the 1.0 release tasks... Changes like these breaks people's code so we should think on what's the best approach. Now we can't break the API until v2.0!

CodeMonkeyJan: I didn't see that 1.0 is officially out until now, congratulations :-)!
Concerning the `signal_type` names, would 1.1 be okay for the change? Or should this be changed at all, due to backwards compatibility? It does not look that bad with the underscore, in my opinion. Alternatively `_signal_type` (the class properties) could be implemented as a list, so we can have a new and a legacy name...
I also saw that there are a few formatting problems with the user guide, I assume updating that would be okay, even between releases, right? I would submit a PR for that then.

CodeMonkeyJan: Oh and concerning the user guide issues, is there a way to preview the whole thin, including links and everything? This would make debugging a lot easier.

francisco-dlp: We can't break the API in 1.1, only in 2.0. We have to be strict with this, as failing to do so would undermine the confidence of users and developers who may based their packages on HyperSpy. For an example, see @vidartf's HyperSpyUI. Currently it is broken due to all the API changes. Maintaining it would be a nightmare if we cannot guarantee that the API will remain untouched until HyperSpy 2.0.

Of course we could take the approach that you propose and add a `_legacy_signal_type` list to allow to make these changes without breaking the API. +1 for this!

Regarding the docs, now that @sem-geologist has fixed the automatic read the docs built, the docs are sync with master in [RtD](http://hyperspy.readthedocs.io/en/latest/). If you want to compile them locally you can use sphinx. This is trivial in linux (simply executing `make html` in `hyperspy/doc` does it), but I've been told that it may not be so straightforward in Windows.

CodeMonkeyJan: Thanks for the input, I totally understand the reasoning behind not breaking the API. I'll try to make two PRs (this time in separate branches with LF line endings) for the `signal_type` list and the documentation. I'm familiar with sphinx, I also use it for another project, so it shouldn't be a problem, even on my windows machine :-).

woozey: > I think that we also decided that there was no need for "_" instead of " " and these are strings with no risk of becoming attributes. So, what about "electron wave"?

> ...It does not look that bad with the underscore, in my opinion. 

I personally always feel uncomfortable with spaces, so would prefer underscore...

doc/user_guide/electron_holography.rst
woozey: Actually I was writing it as an application related user guide (the file name `electron_holography` also refers to the application). That is why I put description of all the relevant functionality in it, independent of real physical location of the functions. I think it make sense if this file have to serve as a user guide indeed (user will find everything relevant to holography here, even in case if he is not aware of Hyperspy structure). Otherwise, I think it would be more logical to put together each separate documentation file for each signal class. In any case I can redo docs in the way it is decided.

CodeMonkeyJan: Hm, that makes indeed sense. I'm not sure what to do in this case. Input from others would be great here :-)...

francisco-dlp:
I have gone quickly through the code and I think that we need to think a little be about the structure.

* It seems that all the methods  assume that `Signal2D` has navigation dimension 0 i.e. they only work in single images. If this is the case I think that they methods should be turned into functions and moved to a new `hyperspy/misc/holography` subpackage. For example, `reconstruct_phase` could be turned into a function and the method could simply call the function internally using `BaseSignal.map` in order to generalise it for nav_dim > 0.

* It seems that `reconstruct_phase` does 2 things: i) calculate `rec_params` if not provided ii) call `_reconstruct` for the image and (optionally) the reference and generate the `ElectronWaveImage`. Wouldn't it be better to provide a method to calculate the `rec_params` and the have a `reconstruct` method that optionally takes the reference and outputs the result?

* Why did you choose to store `rec_params` in metadata? They could be the output of the new method proposed above and be up to the user to store them or not. Also, I think that it'll be good to output them as a dictionary instead of a tuple, as it is very easy to introduce hard to debug bugs by accidentally accessing the wrong item in the tuple. 

* Why not simply returning `sb_pos` and `sb_size` and calculate internally `sb_pos1` and `sb_pos2`?
  ```python
  rec_param = (sb_pos[1] - 1, sb_pos[0] - 1, sb_pos[1] + 1, sb_pos[0] + 1, sb_size)
  ```
* It would be good to output the sb_pos in a format that can be easily consumed by hyperspy markers or ROIs to make it easy for the user to verify the calculation by plotting it on the image.
* A matter of style: we tend to prefer verbose name e.g. `reconstruction_parameters`, `side_band_position`... It makes the code more readable. 
* I would split `_reconstruct` in a number of functions as it seems to be composed of a number of well defined data processing steps that may potentially require individual verification or tweaking. This would make the code more modular, making it easier to contribute enhancements to the reconstruction method by simply providing an alternative function for a given processing step.
* It'll greatly help us reviewing this contribution If you could you submit a PR to hyperspy demos with a holography reconstruction example using the code in this PR.

doc/user_guide/electron_holography.rst
CodeMonkeyJan: I support @francisco-dlp's opinion of using natural strings without underscores, because these strings should be intuitive for the user (even the ones not necessarily used to programming notations). Maybe #1160 caters to everyones needs?

CodeMonkeyJan:
>It seems that all the methods assume that Signal2D has navigation dimension 0 i.e. they only work in single images. If this is the case I think that they methods should be turned into functions and moved to a new hyperspy/misc/holography subpackage. For example, reconstruct_phase could be turned into a function and the method could simply call the function internally using BaseSignal.map in order to generalise it for nav_dim > 0.

At first I thought we could handle this with the `BaseSignal.map()` function (have a user-accessible function which maps the current reconstruction function, which operates on a single image. But `map()` works in place and returns no new signal. So maybe this idea of a dedicated toolkit would fit better.

>It seems that reconstruct_phase does 2 things: i) calculate rec_params if not provided ii) call _reconstruct for the image and (optionally) the reference and generate the ElectronWaveImage. Wouldn't it be better to provide a method to calculate the rec_params and the have a reconstruct method that optionally takes the reference and outputs the result?

Good idea, this promotes modularity (as well as your later suggestions about splitting the reconstruction). Even though that would mean potentially two function calls from end users. I know, that sounds incredibly lazy :-)...

>Why did you choose to store rec_params in metadata? They could be the output of the new method proposed above and be up to the user to store them or not. Also, I think that it'll be good to output them as a dictionary instead of a tuple, as it is very easy to introduce hard to debug bugs by accidentally accessing the wrong item in the tuple.

@woozey and I thought it would be easier if the user will have the `rec_params` handy after the reconstruction, in case further images should be processed (and also make sure they are saved with the image, in case *exactly that* reconstruction should be repeated). That may have been too patronizing... If we split the `rec_param` creation and the reconstruction, as you suggested, and we make sure we can easily save `rec_param`, I think this sounds good.

>Why not simply returning sb_pos and sb_size and calculate internally sb_pos1 and sb_pos2?

I assume it's due to future flexibility, @woozey what's your reasoning in this case?

>rec_param = (sb_pos[1] - 1, sb_pos[0] - 1, sb_pos[1] + 1, sb_pos[0] + 1, sb_size)
It would be good to output the sb_pos in a format that can be easily consumed by hyperspy markers or ROIs to make it easy for the user to verify the calculation by plotting it on the image.

This sounds good, maybe the plot during the reconstruction (if the `verbose` flag is set) can also be modified to use HyperSpys plotting capabilities?

I agree with the rest of your suggestions.

francisco-dlp:
Regarding `map`, yes, it operates in-place, but you can always deepcopy the signal beforehand. Although some methods that predate `BaseSignal.map` don't use it, this could be considered a bug. `map` enables easy modularisation and some cool goodies. For examples, there is already a PR (#479) to parallelize `map`.  Having all the `holography` functions in `hyperspy/misc/holography` has the extra advantage that if you avoid calling any other hyperspy submodule from there, then `hyperspy/misc/holography` would essentially be independent of HyperSpy and you could turn it into an individual package by simply adding a `setup.py` file in that folder. There is an on-going discussion on splitting HyperSpy (#821). Although personally I don't think that it'll make much sense to split e.g. `holography` from the rest of the EM features, it does make sense to split all the EM features from the rest of HyperSpy to make a `hyperspy-em` package. If we make sure now that any new feature is developed with modularity in mind, splitting HyperSpy in the future will be a lot easier.

woozey:
Comments regarding the structure:

> It seems that all the methods assume that Signal2D has navigation dimension 0 i.e. they only work in single images. If this is the case I think that they methods should be turned into functions and moved to a new hyperspy/misc/holography subpackage.

Indeed, `nav_dim = 0` is assumed. No problem, I can do this.

> For example, reconstruct_phase could be turned into a function and the method could simply call the function internally using BaseSignal.map in order to generalise it for nav_dim > 0.

Do I understood your suggestion well: both current methods `reconstruct_phase` and `_reconstruction` should go as functions in the new sub package and but the `HologramImage` signal should be kept and `reconstruct_phase` should be a kind of wrapper? If so I've already checked `map()` function, and it should be doable for me.

> - It seems that reconstruct_phase does 2 things: i) calculate rec_params if not provided ii) call _reconstruct for the image and (optionally) the reference and generate the ElectronWaveImage. Wouldn't it be better to provide a method to calculate the rec_params and the have a reconstruct method that optionally takes the reference and outputs the result?
- I would split _reconstruct in a number of functions as it seems to be composed of a number of well defined data processing steps that may potentially require individual verification or tweaking. This would make the code more modular, making it easier to contribute enhancements to the reconstruction method by simply providing an alternative function for a given processing step.

I can try to splitting both methods making them as modular as it can be done. Regarding `reconstruct_phase` I'd prefer that it does everything in one shot (as @CodeMonkeyJan says, due to laziness as this is usually the way it is done)). What I can do for flexibility is to introduce a method that will assign the reconstruction parameters and in `reconstruct_phase` check if parameters are already assigned otherwise to call the assignment method. Does it make sense?

francisco-dlp:
It sounds good. Regarding doing everything in one shop, I see your point and I fully agree. We take a similar approach e.g. in `align_zero_loss_peak` which  is just a combination of multiple already existing methods that just add convenience not features. Sometimes it is not flexible enough for the purpose and users need to used the individual methods, but often the one-shot method is all that is needed.

woozey:
Comments regarding `rec_param`:

>> Why did you choose to store rec_params in metadata? They could be the output of the new method proposed above and be up to the user to store them or not. Also, I think that it'll be good to output them as a dictionary instead of a tuple, as it is very easy to introduce hard to debug bugs by accidentally accessing the wrong item in the tuple.

> If we split the rec_param creation and the reconstruction, as you suggested, and we make sure we can easily save rec_param, I think this sounds good.

I would prefer keeping the rec_param in metadata in any case, to save those together with the wave image which ensures the reproducibility of the results created with my reconstruction functions. Nevertheless, I would be happy also to output them together with the wave and let it up to user if he wants to store them or not. The question is if it should be outputted always or only if some flag provided as an argument for the reconstruction method? I have nothing against dictionary, just haven't thought about such an option.

>> It would be good to output the sb_pos in a format that can be easily consumed by hyperspy markers or ROIs to make it easy for the user to verify the calculation by plotting it on the image.

>This sounds good, maybe the plot during the reconstruction (if the verbose flag is set) can also be modified to use HyperSpys plotting capabilities?

Will do both, sounds like a good improvement suggestion!

> Why not simply returning sb_pos and sb_size and calculate internally sb_pos1 and sb_pos2?

This is very philosophical question)). There were few reasons for doing it:

- Historically the code was using interactive rectangular ROI input from user (i.e. drawing rectangle in FFT with mouse) as `rec_param[0:3]` so it was logical to use it further. This actually means that rec_param is not necessary sb_pos+-1 as it is at the moment.
- There are very rare cases that the holograms acquired at the same conditions have slight variation of position of the side band due to biprism voltage instability. Therefore for batch processing it sometimes makes sense to find the maximum within ROI for each pair of holograms and references instead of setting strictly `sb_pos `, although this happens to me extremely rarely. 

Honestly, I always felt a bit uncomfortable with such an implementation due to redundancy it involves. So my suggestion would be that indeed we give only `sb_pos` and `sb_size` as `rec_param` and optionally the code will search for the side band maxima within given number of pixels. Does it sound reasonable?

francisco-dlp:
Regarding `rec_param` I see your point. The thing is that the same can  be said of any other method in HyperSpy: in the sake of reproducibility, it may seem like a good idea to store a history of all the methods that have acted on the data and their args and kwargs. However, it seems to me that this is often impractical and requires a very stable API to work across versions.

The thing is that the Jupyter Notebook does solve a big part of the reproducibility issue: the entire work flow is stored in a single document. The intermediate output might not be stored, but any user concerned with reproducibility (e.g. what if in a new version the function computing `rec_param` changes?) can print the output in the notebook (if analysing a single image) or save it in a numpy file if it is large and multi-dimensional, or even save it manually to metadata if they prefer to keep everything in the same file.

Now, let's imagine that we store the side band geometric parameters in metadata. Then, it'll be possible to reconstruct the image without specifying these parameters in the function e.g.

```python

rec = im.reconstruct()

```

Instead of:
```python

rec = im.reconstruct(side_band_position=(<x>, <y>), side_band_size=<s>)

```

Not having to pass the parameters is more convenient, but it may also deal to wrong results if the user is not aware of these important parameters and does not verify that they're correct.

francisco-dlp:

PR #1154, opened 2016-07-14
NEW: Reading support for Cameca NanoSIMS ".im" format

thomasaarholt:
Support for reading in the Cameca NanoSIMS "im" format.

Returns a list of signals of the elements detected using the NanoSIMS. Data is always stacks of images. For _n_ elements/compounds detected it returns a list of _n_ datacubes.

We do not have the spec for the file format, so I've figured it out using the [LANS Matlab reader](http://nanosims.geo.uu.nl/nanosims-wiki/doku.php/nanosims:lans). I only read the bits we've figured out. The header part of the ".im" file largely consists of unknown data, but is not needed for normal analysis (I can't actually think of what it may be).

TODO:
- [X] Add support for reading the accompanying chk_im header text file (easy)
- [ ] Fix some issues with data saved with Windows rather than UNIX (Almost finished adding this.)
- [ ] Analysis stopped halfway through will write zeroes to all remaining slices, wastes space. (Almost finished adding this.)
- [ ] Make some bits of code smarter
- [ ] Format the code according to hyperspy PEP standard 
- [ ] Adding unittests


thomasaarholt:
In the checks:
```python
25:	try:
26:	    from collections import OrderedDict
Unused OrderedDict imported from collections
27:	    ordict = True
28:	except ImportError:
29:	    ordict = False
```
I stole this from the fei reader. It looks like this should be moved to inside `def sarray2dict()` in `array_tools.py`, no? Since I import it later.

francisco-dlp:
It looks pretty good. The `OrderedDict` `try/except` can be safely removed. That's because the code that you took it from was so old that there was a risk that people were using a version of Python without  `OrderedDict` (added in 2009).

francisco-dlp:
I would add to the TODO list adding unittests.

I'll mark this as WIP, let us know when it is ready for review.

francisco-dlp:

PR #1155, opened 2016-07-14
Fix: ReadTheDocs doc building

sem-geologist:
I tested this on my setup of readthedocs and it worked. This is a bit hakish way to deal with the problem, but it works, and it can make the doc compiled for release of 1.0

### DIFF
--- a/doc/conf.py
+++ b/doc/conf.py
@@ -12,6 +12,16 @@
 # serve to show the default.
 
 import sys
+from unittest.mock import MagicMock
+
+class Mock(MagicMock):
+    @classmethod
+    def __getattr__(cls, name):
+            return Mock()
+
+MOCK_MODULES = ['scipy', 'matplotlib', 'traits', 'traitsui', 'numpy', 'h5py']
+sys.modules.update((mod_name, Mock()) for mod_name in MOCK_MODULES)
+
 import os
 sys.path.append('../')
 from hyperspy import Release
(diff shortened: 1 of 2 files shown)


francisco-dlp:
Excellent!

doc/conf.py
```diff
@@ -12,6 +12,16 @@
 # serve to show the default.
 
 import sys
+from unittest.mock import MagicMock
```
francisco-dlp: This breaks local doc compilation. I was wondering, it is needed at all since you disabled [installing extension when compiling from RtD](https://github.com/hyperspy/hyperspy/pull/1155/files#diff-2eeaed663bd0d25b7e608891384b7298R65)?

Even if not for the original purpose, it may actually be useful (with a bit of tweaking) for getting the API docs also in RtD (they are currently broken because of the inability to install the libraries in `MOCK_MODULES`.)

What do you think @sem-geologist?

sem-geologist: I will try to remove on my branch, and see what RtD will do. Then I will report.

sem-geologist:
I report a success: you can get rid of all those mockery lines (with that abstract class ) in conf.py, it successfully builds the documentation at readthedocs without that unittest.mock cruft.

francisco-dlp:
Do you want to PR the changes?

2016-07-18 14:30 GMT+01:00 Petras Jokubauskas <notifications@github.com>:

sem-geologist:
just did that: #1163

PR #1158, opened 2016-07-15
Simplify the installation instructions (hyperspy is now in conda-forge)

francisco-dlp:
@to266 has create a [conda-forge recipe](https://github.com/conda-forge/hyperspy-feedstock) for HyperSpy and it has just being merged. This means that from now on installing hyperspy in Anaconda will be as simple as 

```bash

$ conda install hyperspy

```

This PR changes the install section to reflex this.

### DIFF
--- a/doc/user_guide/install.rst
+++ b/doc/user_guide/install.rst
@@ -53,7 +53,7 @@ MKL libraries) and the easiest installation. The academic license is free.
    `User Guide <https://docs.continuum.io/anaconda/index#user-guide>`_ for
    details.
 
-#. Then install HyperSpy executing the following `conda` command in the Anaconda
+#. Then install HyperSpy executing the following `conda` commands in the Anaconda
    Prompt, Linux/Mac Terminal or Microsoft Windows Command Prompt. (This
    depends on your OS and how you have installed Anaconda, see the `Anaconda
    User Guide <https://docs.continuum.io/anaconda/index#user-guide>`_) for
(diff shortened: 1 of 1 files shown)


francisco-dlp:
Notice that, although we won't merge this until it is reviewed, I have already compiled the changes and uploaded them to the online documentation  because following yesterdays' release and @to266 Scipy2016 talk, we expect that more people than usual will install HyperSpy in the following hours/days and they will all benefit from installing from conda forge.

sem-geologist:
@francisco-dlp  did you saw my fix for documentation generation?

francisco-dlp:
No, I am still trying to catch up with my emails this morning...

doc/user_guide/install.rst
```diff
@@ -43,28 +43,47 @@ context (right-click) menu  entries to start the `Jupyter Notebook <http://jupyt
...
+   depends on your OS and how you have installed Anaconda, see the `Anaconda
+   User Guide <https://docs.continuum.io/anaconda/index#user-guide>`_) for
+   details.
 
    .. code-block:: bash
 
-       $ conda config --add channels conda-forge
-       $ conda install traitsui
-       $ pip install hyperspy
+       $ conda install hyperspy
```
to266: Either 

`conda install -c conda-forge hyperspy`

or add back the line adding the channel :)

francisco-dlp: It works for me like this (tested with latest miniconda).

to266: wow, I'm not sure why...

sem-geologist: maybe because it is in pypi?

to266: Could you try running this on that particular machine?

`conda list hyperspy`

(or replace by `ana-conda` if required..)

francisco-dlp: # packages in environment at /home/fjd29/miniconda3:
#
hyperspy                  1.0                      py35_0    conda-forge

francisco-dlp: I think that the latest anaconda ships with the conda-forge channel and it searches there if it can't find the package in the standard repository.

to266: I doubt that - most likely the config of conda was read from a file, where it was added as one of the channels. 

I would still prefer the explicit version - this would be foul-proof for any (both old and new) conda setup, working for existing users too.

francisco-dlp: I agree, I've put back the add channels line.

2016-07-17 20:28 GMT+01:00 Tomas Ostasevicius <notifications@github.com>:

PR #1160, opened 2016-07-15
Introduction of _alias_signal_types

CodeMonkeyJan:
As discussed briefly in #1152, this PR aims to replace the private `_signal_type` attribute of the signal classes with a list, to allow for more than one string to be assigned to the specific class.

An `EDSSEMSpectrum` e.g. can thus be created with:
```python
s.set_signal_type('EDS_SEM')
s.set_signal_type('EDS SEM')
s.set_signal_type('EDSSEM')
```

The reason for this is that `signal_type` should be lower case (if not an abbreviation) and separated via spaces (like `'EDS SEM'` or `'dielectric function'`. Due to existing code, changing this now (directly after 1.0 launched) would break user code, so the idea of having a choice (with one default being at the start of the `_signal_type` list) came to mind.

What do you think about this?

There were a few issues while working on this:
- There is another `signal_type`  named 'XAS' [here](https://github.com/hyperspy/hyperspy/blob/8be1b691acdff86e296a2c4c12ae539b4a5a9d0a/hyperspy/misc/eels/eelsdb.py#L236), which is not associated with any signal subclass. Could someone tell me the purpose?
- What is the purpose of [`test_attributes.py`](https://github.com/hyperspy/hyperspy/blob/8be1b691acdff86e296a2c4c12ae539b4a5a9d0a/hyperspy/tests/signal/test_attributes.py) and should all other subclasses be added there? 
- There was a problem with the ripple io-plugin where spaces are [removed](https://github.com/hyperspy/hyperspy/blob/8be1b691acdff86e296a2c4c12ae539b4a5a9d0a/hyperspy/io_plugins/ripple.py#L151) from the read properties. The consequence is that `'EDS SEM'` becomes `'EDSSEM'`, which I (for now) included in the possible `signal_types` for `EDSSEMSpectrum` (same deal for `EDSTEMSpectrum`). The tests seem to run fine when I include the line in question, but I assume it has a specific purpose which I can not see? 
- I tried to replace a lot of hardcoded cases where the classes `_signal_type` is compared with the signals `signal_type`, but this often failed because of the import order. I could not import `_signal_type` of the classes before they were properly imported themselves (circle imports and such), so I left it as is at the moment.


francisco-dlp:
Regarding XAS, you're right, there is no specialised subclass for this signal. However, we fetch the signal type from the EELSdb and, maybe in the future there'll be a XASSpectrum class. If there is, this PR will help with supporting the current `signal_type` string  if finally the person implementing it decides to give it another name.

`test_attributes.py` made sense when the signal dimension was inferred for `record_by`. Now it doesn't make much sense...

Regarding the ripple reader issue, I would deal with it in the ripple reader rather than supporting an extra signal_type.

Although the current implementation works, I think that it adds some unnecessary complexity. What about leaving `_signal_type` as a string and adding a `_legacy_signal_type` attribute with the list of legacy signal types? In this way on `assign_signal_subclass` should require modification and the code would be more readable.

CodeMonkeyJan:
>Regarding XAS, you're right, there is no specialised subclass for this signal. However, we fetch the signal type from the EELSdb and, maybe in the future there'll be a XASSpectrum class. If there is, this PR will help with supporting the current signal_type string if finally the person implementing it decides to give it another name.

Sounds reasonable.

>test_attributes.py made sense when the signal dimension was inferred for record_by. Now it doesn't make much sense...

Should this be deleted then (maybe in a separate PR) or should it stay?

>Regarding the ripple reader issue, I would deal with it in the ripple reader rather than supporting an extra signal_type.

I could delete the according line in the ripple reader, but even though the tests would still pass, I do not know if this would break something which isn't covered by the tests.

>Although the current implementation works, I think that it adds some unnecessary complexity. What about leaving _signal_type as a string and adding a _legacy_signal_type attribute with the list of legacy signal types? In this way on assign_signal_subclass should require modification and the code would be more readable.

My reason for the current implementation was the fact that the `metadata.Signal.signal_type` is checked not only in `assign_signal_subclass`, but also in a lot of other places (mainly in the EDS signal subclasses). I have another idea, which circumvents the need to check everywhere for all legacy types:
in `assign_signal_subclass`, if a `_legacy_signal_type` is encountered (and only then), it is changed to the default `_signal_type`. Thus, files can keep the legacy types, but as soon as they are loaded into HyperSpy and a subclass is assigned, the signal will have the default type. Would this be okay?

francisco-dlp:
Looks good! I think that we should add deprecation warnings to `set_signal_type` informing about the changes.

There are some things that we should think about before merging this:

1. As is, it'll load signals with legacy `signal_type` and save them with their new `signal_type`. This means that old versions of HyperSpy won't assign correctly the `signal_type`. I see the following options:
    1. On saving, save using the legacy signal_types forever. This will keep compatibility across all versions but the signal_type in the files will be different from the signal type exposed to the users.
    2. On saving, save using the legacy signal_types until e.g. HyperSpy 2.0 and from then on using the new `signal_type` strings only. We'll break backward compatibility for HyperSpy < 1.0.1 (assumnig that we ship this with HyperSpy 1.01), but keep the format more consistent with the code and with itself by enforcing a regular naming convention. This would involve changing the format version again.

2. @woozey wrote "I personally always feel uncomfortable with spaces, so would prefer underscore". The issue that you've experienced with the ripple format is an example of the advantage of using underscores instead of spaces that I missed but @woozey's gut feeling didn't :) I would fix it by replacing spaces with underscores when writing to ripple and reversing the _ back to spaces when reading. However, I wonder if this is just the tip of the iceberg and indeed using spaces instead of underscores may backfire in ways that we can't see now.

3.  Point 1. shows that any changes to `signal_type` are painful. Therefore, we should try to get them as right as possible now that we've decided to change them. It seems that we have agreed on the following conventions:

    1. Acronyms (and only acronyms) should use uppercase.
    2. Anything else should be lower case.
    3. Don't use underscores instead of spaces (Although we maybe should still debate about this, see 2.).
 However, keeping in mind that HyperSpy seems to be growing beyond the EM community, I think that we should also debate (at least) the following:

   1. Should we go for e.g. "TEM EELS"  instead of simply "EELS"? In this particular case I think that the convention is that "EELS" is TEM EELS and HREELS is reflection EELS (although the acronym stands for "high-resolution EELS"). In my opinion HREELS is misleading now that TEM EELS is reaching 1meV resolution. Of course, this is not something to debate here, but the point is that, it is not as straightforward as it seems.
   2. Should we support acronyms in `set_signal_type` but expand them when writing to hdf5? E.g.  `set_signal_type("EELS")` results in `_signal_type="electron energy loss spectroscopy". This more verbose internal signal type may ease the task of deciphering what's in the file for somebody who is unfamiliar with (in this case) EELS and HyperSpy.

Given how important is to get this right, it'll be good to get the input of as many people as possible so, if you are reading this and you have an opinion on the topic, please have your say.

CodeMonkeyJan:
The deprecation warning is implemented, I kept the date of removal vague ("removed in a later version"), or should I set this to HyperSpy 2.0?

1. For saving, I personally would go with your first suggestion of breaking backwards compatibility in 2.0, but in every case up until then we have to save with the legacy `signal_type`. I thus tried the following:
In `io.save()`, before calling the `writer.file_writer`, I temporarily set the `signal.metadata.Signal.signal_type` to the first entry of the `_legacy_signal_types` list, if it exists (I kept it a list if, for any reason, more legacy types or alternatives have to be added). Directly after the writing is finished, it is changed back to the default `_signal_type`. I also included a `try ... finally` construct to make sure that the `signal_type` revert happens, even when the writing fails.
I placed a `#TODO` there, to mark it for deletion in 2.0, or is there a better way?
This also means that the ripple format issue is solved, because every writer now writes with underscores again (although, now that I think about it, what happens if other signal types with spaces and without legacy types will be saved? It is not tested currently, I would assume the problem will arise again).

2. If we decide to use underscores after all, currently only `DielectricFunction` will be a problem, because it does not comply with the current conventions. This is indeed a decision which we should not take lightly and I would also appreciate comments from other people (@woozey, @to266  or @vidartf?). Keeping underscores would eliminate the reader issues and most of the legacy types, but maybe don't look that pretty and are less intuitive... Let's gather opinions on this first.

3) I cannot say much as EELS is not a field I am proficient in, but adding aliases or a new default would be quite easy with my current implementation. Just overwrite the following in the `EELSSpectrum` class:
```python
_signal_type = "TEM EELS"
_legacy_signal_types = ["EELS"]
```

4) Supporting acronyms and short aliases is also as easy as adding them to the `_legacy_signal_types` list (maybe calling it `_alias_signal_types` would be better?). As soon as a class is assigned, they will be turned into the default (and maybe more verbose) `_signal_type`. Do you only want to expand them when writing to hdf5? I'd think more verbose names would also benefit people inspecting the loaded signal itself. Let's have a short example:
```python
class DielectricFunction(ComplexSignal1D):

    _signal_type = "dielectric function"
    _alias_signal_types = ["dielectric_function", "DielectricFunction", "DF"]
```
A user can use the `set_signal_type` function with all four variants and the metadata will always have the default `_signal_type` version. Upon saving with any file-writer, the first alias (the legacy one) will be used to avoid issues with white space and backwards compatibility.

Oh, one last thing, why does coverage fail? I think it is because the coverage of `ripple.py` decreased, but I didn't change more than one line there...

francisco-dlp:
@CodeMonkeyJan, it seems that we're not getting a lot of opinions on this. Since we're planning to release v1.0.1 I would suggest to take some quick decisions now and leave the more complex debate open for later. This is what I suggest:

* Rename `_legacy_signal_types` to `_alias_signal_types` as you propose.
* Keep the underscores for now.

This provides minimal disruption for this patch release while merging the useful machinery in this PR. What do you think?

francisco-dlp:
I think that it'll be better not to add the aliases until we decide if it is a good idea (except "DielectricFunction"). Once we provide aliases we have to maintain them...

CodeMonkeyJan:
I did just that, the `signal_type`s are now back to their old versions (I didn't miss anything).
Of course saving a signal now saves as the current `signal_type` again and not as the first `_legacy_signal_types` entry, now called `_alias_signal_types`.
The new versions (without underscore) can now be found in `_alias_signal_types` and calling `set_signal_type` with one of these now creates the same signals (converting it to the standard `_signal_type`).
So in summary, everything is back to normal, with the addition that `set_signal_type` can now be used with aliases.

Oh and the EELSDB seems to be broken again...

CodeMonkeyJan:
Timing :-).
Okay, I'll delete the aliases. What do you have in mind for `DielectricFunction`?
```python
_signal_type = "DielectricFunction"
_alias_signal_types = ["dielectric functoin"]
```
?

francisco-dlp:
Yes, I think that that's the sensible option for the DF until we take a consensual decision on this.

francisco-dlp:
It seems ready to me, is it to you?

CodeMonkeyJan:
From my side everything is ready. I still don't understand why coverage fails (the files I changed didn't even go down), but at least the builds were successful this time.

CodeMonkeyJan:
Great :-)! The only problem I see is that it merged to master (as I didn't know about the new PR policy when creating this). Is there anything I can/should do, or will this be fixed another way?

francisco-dlp:
Sorry that I didn't notice that it was submitted to the wrong branch. I've merged it into RELEASE_next_minor and changed the milestone to 1.1. I think that, just by chance, that's a better decision as by then we'll have the time to decide how to use this new feature.

CodeMonkeyJan:
Sounds good to me :-)!

PR #1168, opened 2016-07-18
fix: unify types

sem-geologist:
so that unsigned and signed integers would not be compared
and some of compilers would not complain


tjof2:
No complaints from GCC on my machine any more.

francisco-dlp:
Following #1170, could you submit this to the RELEASE_next_patch branch?

sem-geologist:
Ok, If I submit to the RELEASE_next_patch should I close pull request to the master?

francisco-dlp:
Yes, this can be closed.

sem-geologist:
this is moved to #1171, so this is going to be closed

PR #1162, opened 2016-07-18
Protochips log plugin

ericpre:
Support for reading Protochips (gas cell, heating/biasing) log files.

Issue #1136, opened 2016-07-13
HyperSpy's hdf5 format needs standard for date and time


francisco-dlp:
Currently HyperSpy doesn't have a standard way of storing date and time metadata. There are many standards available. We could simply choose one that is well supported by a standard python library.

dnjohnstone:
I think this is a good idea - any thoughts on which one might be the one to take?

ericpre:
I looked a bit around on google to handle date and time:
- [python datetime](https://docs.python.org/3.5/library/datetime.html)
- [numpy datetime](http://docs.scipy.org/doc/numpy/reference/arrays.datetime.html)

There are other libraries that have their own date and time implementation with much more features.
Numpy offers higher precision than the python datetime (microsecond precision).
Shall we go for numpy datetime? I think it would make sense for a data analysis library and it's possibly the most straighforward to use.

francisco-dlp:
Sounds good, it seems that the default format of numpy's datetime is ISO 8601 and that it has ns accuracy. +1 for storing dates in ISO 8601 txt with up to ns accuracy. Would parsing dates in file readers using numpy and storing them in metadata as ISO 8601 strings be a sensible action plan?

francisco-dlp:
We could keep our current `General.time` and `General.date` entries, and just agree to store time and date information there in ISO 8601.

ericpre:
What would be better: to have `General.time` and `General.date` as a `numpy.datetime64` object or to store it as string?
Or maybe make a `Metadata` class that inherits from `DictionaryTreeBrowser` and can handle the [metadata specification](http://hyperspy.org/hyperspy-doc/current/user_guide/metadata_structure.html) with specific methods? This would allow storing the `numpy.datetime64` object as a private attribute and display the date and time in `General` as strings, which will looks much more readable for the user.

francisco-dlp:
I think that storing the date and time in string would be better. Currently they're not used anywhere in the code. The rule of thumb is too keep things as simple as possible.

If any functions uses it, it can easily convert the string to e.g. numpy datetime64 for consumption. Such as function could be e.g. stack which could automatically order a set of files for stacking using date and time info.

ericpre:
Fixed in #1240.

PR #1173, opened 2016-07-19
Protochips log plugin

ericpre:
Support for reading Protochips (gas cell, heating/biasing) log files.

#1162 PR to next_minor instead of master.


francisco-dlp:
Currently HyperSpy assumes that there is a 1 to 1 mapping between extensions and readers. However, this format uses a generic csv extension. Currently this is not an issue as it is the only one, but it might becomes an issue.

In any case, this reader should check if the csv file is a "Photoship csv" file and, if not, raise a meaningful IOError.

hyperspy/io_plugins/protochips.py
```diff
@@ -0,0 +1,183 @@
...
+        d = {'Start time': self.start_datetime}
+        d['Time units'] = self.time_units
+        for quantity in self.logged_quantity_name_list:
+            d['%s_units' % quantity] = self._get_quantity_units(quantity)
+        d['User'] = self.user
+        d['Calibration file name'] = self._parse_calibration_file()
+        return d
+
+    def _get_mapping(self):
+        return {'Protochips_header.date': ("General.time", None), }
```
francisco-dlp: Currently we don't have a standard for time, see #1136

hyperspy/io_plugins/protochips.py
```diff
@@ -0,0 +1,183 @@
...
+
+    def _get_metadata(self, quantity):
+        return {'General': {'original_filename': os.path.split(self.filename)[1],
+                            'title': '%s (%s)' % (quantity,
+                                                  self._get_quantity_units(quantity)),
+                            'user': self.user,
```
francisco-dlp: use, start_time, notes and calibration_file are not currently part of [HyperSpy's metadata specifications](http://hyperspy.org/hyperspy-doc/current/user_guide/metadata_structure.html).

hyperspy/io_plugins/protochips.py
```diff
@@ -0,0 +1,183 @@
...
+         # for the gas cell, the calibration is saved in the notes colunm
+        if self.calibration_file is None:
+            calibration_file = "The calibration files names are saved in "\
+                "metadata.General.notes"
+        else:
+            calibration_file = self.calibration_file
+        return calibration_file
+
+    def _get_axes(self):
+        scale = np.diff(self.time_axis).mean()
```
francisco-dlp: What if the time step is not constant?

hyperspy/io_plugins/protochips.py
```diff
@@ -0,0 +1,183 @@
...
+    def _get_metadata(self, quantity):
+        return {'General': {'original_filename': os.path.split(self.filename)[1],
+                            'title': '%s (%s)' % (quantity,
+                                                  self._get_quantity_units(quantity)),
+                            'user': self.user,
+                            'start_time': self.start_datetime,
+                            'notes': self._parse_notes(),
+                            'calibration_file': self._parse_calibration_file()},
+                "Signal": {'signal_type': quantity,
+                           'time_axis': self._get_metadata_time_axis()}}
```
francisco-dlp: Why storing here the time axis?

hyperspy/tests/io/protochips_data/hyperspy_log.py
```diff
@@ -0,0 +1,11 @@
+#!/usr/bin/env python
```
francisco-dlp: I bet that this file got here by mistake.

hyperspy/tests/io/test_protochips.py
```diff
@@ -0,0 +1,254 @@
...
+        nt.assert_equal(self.pgc.time_units, 'Milliseconds')
+        nt.assert_equal(self.pgc.temperature_units, 'Degrees C')
+        nt.assert_equal(self.pgc.pressure_units, 'Torr')
+        nt.assert_equal(self.pgc.current_units, 'Amps')
+        nt.assert_equal(self.pgc.voltage_units, 'Volts')
+        nt.assert_equal(self.pgc.resistance_units, 'Ohms')
+        nt.assert_equal(self.pgc.user, 'eric')
+
+"""
+Protochips electrical
```
francisco-dlp: I think that commenting this out would be better than placing it in a string.

hyperspy/tests/io/test_protochips.py
```diff
@@ -0,0 +1,254 @@
...
+    def test_read_data(self):
+        gen = (self.pa._data_dictionary[key]
+               for key in self.pa.logged_quantity_name_list)
+        data = np.vstack(gen)
+        expected_data = np.load(os.path.join(
+            dirpath, 'protochips_electrical.npy'))
+        np.testing.assert_allclose(data.T, expected_data)
+
+"""
+Protochips thermal
```
francisco-dlp: As above

hyperspy/tests/io/test_protochips.py
```diff
@@ -0,0 +1,254 @@
...
+    def test_read_data(self):
+        gen = (self.pt._data_dictionary[key]
+               for key in self.pt.logged_quantity_name_list)
+        data = np.vstack(gen)
+        expected_data = np.load(os.path.join(
+            dirpath, 'protochips_thermal.npy'))
+        np.testing.assert_allclose(data.T, expected_data)
+
+"""
+Protochips electrothermal
```
francisco-dlp: As above

hyperspy/io_plugins/protochips.py
ericpre: Does is it mean that it should be moved to original_metadata?

hyperspy/io_plugins/protochips.py
ericpre: Good point, as is it today, Protochips software doesn't change time step during an experiment. The time step is not really constant though... there is a 1-2 ms difference between each time step, that I assume negligible here.
Would it be worth adding a method that checked that is the case? and raise an error if the difference is too large. I think this would be a bit overkill, but could be worth doing it!

hyperspy/io_plugins/protochips.py
ericpre: In case, the user want to get the exact time axis, which may have a few ms difference from the parse time axis, as mentioned previously.

Move this to original_metadata?

hyperspy/tests/io/protochips_data/hyperspy_log.py
ericpre: yes...

ericpre:
The time axis in protochips log files is not completely linear, there are a 1-2 ms difference between each time step.
#438 could solve this issue.

hyperspy/io_plugins/protochips.py
francisco-dlp: Either that or we should update the metadata specifications or  a mixture of the two.

Maybe user could be mapped to a new `General.authors`? With the inception of open data, we may also need a `General.doi`.

I see how notes could be useful. So +1 on adding this to the specifications too. notes or comments or something else?

calibration_files certainly does not belong to general. What is it? If it is required in could go into a new `md.Instrument.x.x` and follow well defined and general specifications. This is only worth it if a HyperSpy function is expected to use this information.

I am not sure about what to do with start_time. It's more specific than date and I see that it could be useful for acquisition devices that save each signal in individual files. What do you think that could be the best place for it?

hyperspy/io_plugins/protochips.py
francisco-dlp: Yes, I think that this should go to original_metadata

hyperspy/io_plugins/protochips.py
francisco-dlp: One of the main limitations of HyperSpy is that it doesn't support non-linear axes. We should add support for them asap, but, in the meantime, your approach seems to me like a good compromise. Instead of raising an error if the difference is too large I would add e.g. max(diff) to the log and maybe add a comment about this in the code and the User Guide. Once we add support for non-linear axes we can then deal with this properly.

hyperspy/io_plugins/protochips.py
ericpre: Ok thanks, it's more clear now. Actually, I was not aware of the metadata specification!!
I will make a PR to update the metadata specification with `authors` and `doi` and try to clarify the date and time.
For the rest, I will move it to `original_metadata`.

Yes, start_time can be quite important in order to synchronise data. I will parse it to date and time.

hyperspy/io_plugins/protochips.py
ericpre: I am not sure to understand. In the [metadata spec](http://hyperspy.org/hyperspy-doc/current/user_guide/metadata_structure.html) the time is supposed to be a datetime.time, so it looks like, it is defining a standard for date and time in hyperspy?
Am I right, if I just parse datetime.time() and datetime.date() to the corresponding attribute?

hyperspy/io_plugins/protochips.py
```diff
@@ -0,0 +1,183 @@
...
+
+    def _get_metadata(self, quantity):
+        return {'General': {'original_filename': os.path.split(self.filename)[1],
+                            'title': '%s (%s)' % (quantity,
+                                                  self._get_quantity_units(quantity)),
+                            'user': self.user,
+                            'start_time': self.start_datetime,
+                            'notes': self._parse_notes(),
+                            'calibration_file': self._parse_calibration_file()},
+                "Signal": {'signal_type': quantity,
```
ericpre: Should be signal_type an empty string? Or this is alright to set a physical quantity, such as temperature, current, etc, in case it can be use in the future?

hyperspy/io_plugins/protochips.py
francisco-dlp: I forgot to delete data entry in medata. This was removed from the specification because it required calling `eval` which is unsafe. We should agree on a human friendly widespread standard for this with good accuracy (i.e. supporting at least microseconds).

hyperspy/io_plugins/protochips.py
francisco-dlp: Would it be sensible to add "quantity" to "md.Signal"? This would be "counts" in EELS.
We should also add units. What about "md.Signal.quantity = "temperature (K)"? (This would require a separate PR and discussion of course.)

ericpre: yes, #546

hyperspy/io_plugins/protochips.py
ericpre: I left the date and time using datetime for now. Once  #1136 is fixed, we can change it.

ericpre:
Thanks for the review, I think I am done.
I tried to handle the case of "bad" or not a Protochips csv file, but I am not sure, it was the best to do it.

hyperspy/io_plugins/protochips.py
```diff
@@ -0,0 +1,217 @@
...
+    def __init__(self, filename, header_line_number=10):
+        self.filename = filename
+        self._parse_header(header_line_number)
+        self._read_data(header_line_number)
+        self.calibration_file = None
+
+    def _parse_header(self, header_line_number):
+        self.raw_header = self._read_header(header_line_number)
+        self.column_name = self._read_column_name()
+        self._check_if_protochips_csv_file()
```
francisco-dlp: I think that rewriting this as follows would make the code more readable:
```python

if not self.is_protocips_csv_file():
    raise IOError(invalid_file_error)
```

hyperspy/io_plugins/protochips.py
```diff
@@ -0,0 +1,217 @@
...
+        d['Calibration file name'] = self._parse_calibration_file()
+        d['Time axis'] = self._get_metadata_time_axis()
+        return d
+
+    def _get_metadata(self, quantity):
+        return {'General': {'original_filename': os.path.split(self.filename)[1],
+                            'title': '%s (%s)' % (quantity,
+                                                  self._parse_quantity_units(quantity)),
+                            'authors': self.user,
+                            'date': self.start_datetime.date(),
```
francisco-dlp: As discussed in #1136, we should rather store the date time information in txt format using an international standard. ISO 8601 sounds good. 

In any case, modifying the metadata requires a new PR. Would you like to send a PR adding authors to metadata and changing date and time to ISO 8601? What should be the format for authors? I suggest the Latex format: "Surname1, Name1 and Surname2, Name2, ...". We could also add a DOI to General in the format "doi:10.1000/182".

hyperspy/io_plugins/protochips.py
```diff
@@ -0,0 +1,217 @@
...
+                            'time': self.start_datetime.time(),
+                            'notes': self._parse_notes()},
+                "Signal": {'signal_type': '',
+                           'quantity': self._parse_quantity_name(quantity),
```
francisco-dlp: Oh, and quantity must also be added to the metadata specification. Note that all quantities with units include the units in the quantity string e.g. "beam energy (keV)". I think that it'll be good to be consistent and do the same for quantity.

francisco-dlp:
I think that there are only 2 options:

* Remove all the metadata which is not in the specifications and merge it now.
* Leave it as it is and wait for the changes in metadata to be merged in a separated PR.

ericpre:
sorry, I have your message too late. Let's go for the first option (I prefer to merge this one before moving on to the PR on metadata). I am going to comment out all the metadata not in spec, so it will still be there ready to go. Does it sound like a plan?

francisco-dlp:
Sounds good. We're releasing 1.1 today :)

ericpre:
Done.

hyperspy/tests/io/test_protochips.py
```diff
@@ -0,0 +1,287 @@
+# -*- coding: utf-8 -*-
+"""
```
francisco-dlp: Could you replace this with the standard header?
(Sorry that I missed this before)

PR #1169, opened 2016-07-18
Improved access to SciPy fitting algorithms

tjof2:
As discussed in #1098 and a resubmission of #1100.

1. Replace the calls to `scipy.optimize.fmin_XXX` with calls to [`scipy.optimize.minimize()`](http://docs.scipy.org/doc/scipy/reference/optimize.html).
2. Updates `leastsq` to incorporate bounds, thus deprecating `mpfit`
3. Adds global optimization algorithms from SciPy

**To do:**
- [x] Tests
- [x] Documentation
- [x] Deprecation warnings
- [x] Update leastsq to take bound constraints
- [x] Add global optimization method (differential evolution) 
- [x] Return covariance matrix using Jacobian for new `leastsq`

PR #1176, opened 2016-07-20
Fix fit

francisco-dlp:
Currently master and RELEASE_next_minor are not passing the test due to this typo introduced in #1169 that I missed when merging it.


francisco-dlp:
@tjof2, could you have a look at this?

tjof2:
Ah, where did they come from?!

francisco-dlp:
Thanks. Yes, there is an unrelated issue with the blockfile test, probably due to a new version of a library.

PR #1180, opened 2016-07-20
Fix blockfile test

francisco-dlp:
This fixes a blockfile test that was wrongly assuming that only the blockfile reader may raise an exception.

This should hopefully make the unittests in travis pass again.


francisco-dlp:
Could somebody have a look at this so that we can merge it and travis failures get useful again?

PR #1181, opened 2016-07-20
Fix closing defaults file

francisco-dlp:
The defaults file was not closed. AFAIK. this was not causing any issue other than a warning in the log.

### DIFF
--- a/hyperspy/defaults_parser.py
+++ b/hyperspy/defaults_parser.py
@@ -56,15 +56,15 @@ def guess_gos_path():
 
 if os.path.isfile(defaults_file):
     # Remove config file if obsolated
-    f = open(defaults_file)
-    if 'Not really' in f.readline():
-        # It is the old config file
-        f.close()
-        _logger.info('Removing obsoleted config file')
-        os.remove(defaults_file)
-        defaults_file_exists = False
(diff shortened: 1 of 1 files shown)


hyperspy/defaults_parser.py
```diff
@@ -56,15 +56,15 @@ def guess_gos_path():
...
-    f = open(defaults_file)
-    if 'Not really' in f.readline():
-        # It is the old config file
-        f.close()
-        _logger.info('Removing obsoleted config file')
-        os.remove(defaults_file)
-        defaults_file_exists = False
-    else:
-        defaults_file_exists = True
+    with open(defaults_file) as f:
```
to266: ...it seems strange to use `with` statement and then close and even delete the file inside the scope.

Could you disentangle the two steps?

i.e. 
```python
with open(defaults_file) as f:
    if 'Not really' in f.readline():
       defaults_file_exists = False
    else:
       defaults_file_exists = True
if not defaults_file_exists:
    _logger.info('Removing obsoleted config file')
    os.remove(defaults_file)
```

francisco-dlp: Done

PR #1184, opened 2016-07-20
Complex plot fixes

francisco-dlp:
Prior to the addition of the complex signals HyperSpy was able to plot complex 1-d and 2-d signals. When 1d 2 lines where added to the same figure axes, one for imag and another for real. When 2d, log(abs(signal)) was plotted (!) to mimic DM.

This PR:

* Re-enables the feature for 1d signals with a new `plot` kwarg `same_axes`. The default is `True`, is this a reasonable choice?
* Removes the code to plot log(abs(signal)) 
* Change `KeyError` for `ValueError` when passing wrong value to `plot` `representation`


to266:
Update with master

francisco-dlp:
Done. However, I would like to hear from @CodeMonkeyJan before merging.

to266:
This also breaks
```python
In [1]: %matplotlib
Using matplotlib backend: Qt4Agg

In [2]: import numpy as np

In [3]: import hyperspy.api as hs

In [4]: s = hs.signals.ComplexSignal1D(np.ones((3,2), dtype='complex'))

In [5]: s.plot()
<ipython-input-5-d3eb09d34df4> in <module>()
----> 1 s.plot()

/home/to266/dev/hyperspy/hyperspy/_signals/complex_signal.py in plot(self, navigator, axes_manager, representation, same_axes, **kwargs)
    199         if representation == 'cartesian':
    200             if same_axes and self.axes_manager.signal_dimension == 1:
--> 201                 super().plot(**kwargs)
    202             else:
    203                 self.real.plot(

/home/to266/dev/hyperspy/hyperspy/signal.py in plot(self, navigator, axes_manager, **kwargs)
   1975                     " \"slider\", None, a Signal instance")
   1976 
-> 1977         self._plot.plot(**kwargs)
   1978         self.events.data_changed.connect(self.update_plot, [])
   1979         if self._plot.signal_plot:

/home/to266/dev/hyperspy/hyperspy/drawing/mpl_he.py in plot(self, **kwargs)
    132                 self.pointer.color = 'red'
    133                 self.pointer.connect_navigate()
--> 134             self.plot_navigator()
    135         self.plot_signal(**kwargs)
    136 

/home/to266/dev/hyperspy/hyperspy/drawing/mpl_he.py in plot_navigator(self)
    114 
    115             imf.title = self.signal_title + ' Navigator'
--> 116             imf.plot()
    117             self.pointer.set_mpl_ax(imf.ax)
    118             self.navigator_plot = imf

/home/to266/dev/hyperspy/hyperspy/drawing/image.py in plot(self, **kwargs)
    250         for marker in self.ax_markers:
    251             marker.plot()
--> 252         self.update(**kwargs)
    253         if self.scalebar is True:
    254             if self.pixel_units is not None:

/home/to266/dev/hyperspy/hyperspy/drawing/image.py in update(self, **kwargs)
    364             new_args.update(kwargs)
    365             self.ax.imshow(data,
--> 366                            **new_args)
    367             self.figure.canvas.draw()
    368 

/home/to266/anaconda3/lib/python3.5/site-packages/matplotlib/__init__.py in inner(ax, *args, **kwargs)
   1810                     warnings.warn(msg % (label_namer, func.__name__),
   1811                                   RuntimeWarning, stacklevel=2)
-> 1812             return func(ax, *args, **kwargs)
   1813         pre_doc = inner.__doc__
   1814         if pre_doc is None:

/home/to266/anaconda3/lib/python3.5/site-packages/matplotlib/axes/_axes.py in imshow(self, X, cmap, norm, aspect, interpolation, alpha, vmin, vmax, origin, extent, shape, filternorm, filterrad, imlim, resample, url, **kwargs)
   4945                               resample=resample, **kwargs)
   4946 
-> 4947         im.set_data(X)
   4948         im.set_alpha(alpha)
   4949         if im.get_clip_path() is None:

/home/to266/anaconda3/lib/python3.5/site-packages/matplotlib/image.py in set_data(self, A)
    447         if (self._A.dtype != np.uint8 and
    448                 not np.can_cast(self._A.dtype, np.float)):
--> 449             raise TypeError("Image data can not convert to float")
    450 
    451         if (self._A.ndim not in (2, 3) or

```
Or am I just using the `ComplexSignal` in the wrong way? Though this seems to be the obvious choice...

CodeMonkeyJan:
@francisco-dlp: This seems reasonable, I'm not sure why it would break after your changes... I'll try to reproduce @to266's traceback.

CodeMonkeyJan:
As far as I can see the problem is the fact that, if the signal is 2D, the navigator is constructed by stacking the 1D spectra. This results in a complex image as the navigator. Due to the fact that we removed
```python
if np.issubdtype(data.dtype, complex):
   data = np.log(np.abs(data))
```
in `hyperspy.drawing.image.update()`, `imshow` is called a few lines later, which can not handle complex data.

I thus tried the following:
Replacing [this](https://github.com/hyperspy/hyperspy/blob/4ffef31ee3e2673daabe80fca4f597da4625aee6/hyperspy/signal.py#L1968) line with the following:
```python
if np.issubdtype(self.data.dtype, complex):
    self._plot.navigator_data_function = \
        lambda axes_manager=None: np.abs(self.data)
else:
    self._plot.navigator_data_function = \
        lambda axes_manager=None: self.data
```
ensures that the navigator can always be plotted with `imshow` (the compromise being that we navigate with the absolute of the stacked complex spectra, but I assume this is okay).

Furthermore I replaced [this](https://github.com/hyperspy/hyperspy/blob/4ffef31ee3e2673daabe80fca4f597da4625aee6/hyperspy/signal.py#L1912) line with this:
```python
if np.issubdtype(navigator().dtype, complex):
    return np.abs(navigator())
else:
    return navigator()
```
to make sure that `ComplexSignal2D` instances as a navigator return their absolute instead for the navigation.

I'm not very familiar with all the intricacies of the plotting in HyperSpy, I'm sorry that I missed this during my initial PR for the complex data format. I hope these changes don't break anything else :-/. It may also be possible that I missed another navigation case...

There is one thing that is still not working:
After using the navigator, the spectra seem to fall back to displaying just the real part of the complex spectrum instead of plotting the real and imaginary part separately (the blue line is missing). I'm still unsure what the reason is, but I'm trying to find out (I assume the imaginary part is just not updated?)...

francisco-dlp:
@CodeMonkeyJan, I have applied your patch and now it seems to be working fine, thanks! I couldn't reproduce the issue that you mention:

>After using the navigator (shifting the red line with the mouse), the spectra seem to fall back to displaying just the real part of the complex spectrum instead of plotting the real and imaginary part separately 

Can you confirm that you can still reproduce it.
Also, it seems now to me that `same_axes=False`  may be a better default as it is more polished and consistent with 2D plotting. What do you think?

to266:
I'm able to reproduce the issue with the seemingly disappearing imaginary part of the plot that @CodeMonkeyJan mentioned.

Actually the line plot is still there, but the `ylim` of the axis is scaled to only take into account the real (or absolute, if `polar`) part of the plot, and for some data the imaginary line is not visible, as it's just simply out of range.

Code to reproduce:
```python
In [1]: import numpy as np

In [2]: import hyperspy.api as hs

In [3]: %matplotlib
Using matplotlib backend: Qt4Agg

In [4]: s = hs.signals.ComplexSignal1D(np.ones((3,2), dtype='complex'))

In [5]: s.imag = np.random.rand(3,2)

In [6]: s.plot()
```

Could be considered as a separate bug though. @francisco-dlp will you fix this here? If not, I'd merge this, as seems to be OK

CodeMonkeyJan:
I still get this issue, if @to266 can also reproduce them, I'm at least not seeing ghosts :-).
Can you tell me where I should look? I'd like to help find a solution for this if I can... I assume somewhere, where the plots are updated?

CodeMonkeyJan:
Oh, concerning `same_axis`, I like personally like the behaviour of plotting real and imaginary part in one image. It uses a benefit of 1D plots which we just don't have in 2D.

to266:
@CodeMonkeyJan the figure is update [here](https://github.com/hyperspy/hyperspy/blob/master/hyperspy/drawing/signal1d.py#L142), where it loops over all individual lines and updates their axes ([here](https://github.com/hyperspy/hyperspy/blob/master/hyperspy/drawing/signal1d.py#L340)). However the design here assumes that each line (of which there can be at most two) has its own y axis (left and right for the two lines), and hence can reset the `ylim` as it wishes. This assumption is not true in this case, hence the bug..

CodeMonkeyJan:
You're right, it's not that the imaginary part is not updated, it just shifted out of view (due to a badly chosen test array), because the `ylim` only takes the real part into account.

Am I correct that the functionality which allows plotting real and imaginary parts in one image is sort of "hijacking" another mechanism which would normally separate two lines to the left and right axis?

The problem is that in this case, as @to266 stated, both lines share the same axis and only the real value line has `auto-scale` set to `True`. Also auto-scaling for the imaginary part does not solve the problem, though, as it would just override the min-max values of the real part (which we do not want either).

I thought I found a solution by setting `sl.autoscale = True` after [this](https://github.com/hyperspy/hyperspy/blob/a40e65b19c3f12877f689a81046fb430b1ab358b/hyperspy/drawing/mpl_hse.py#L109) line and adding the following before [this](https://github.com/hyperspy/hyperspy/blob/a40e65b19c3f12877f689a81046fb430b1ab358b/hyperspy/drawing/signal1d.py#L340) line:
```python
if self.get_complex:
    yreal = self.data_function(axes_manager=self.axes_manager).real
    y_min = min(y_min, yreal.min())
    y_max = max(y_max, yreal.max())
```
The thought behind this is that first the `Signal1DLine` representing the real part updates normally as it did before and after that, the line representing the imaginary part (where `self.get_complex` is True) makes sure that the limits of the real part are taken into account when updating itself.

There is a big issue with this though:

Apparently the axes update twice for some reason, when you move the navigator line, i.e. that after real and imaginary lines updated, they both do so again. I have no idea why that's the case. This in itself would not be a big problem though, it just seems like a waste of computation time. Or I'm just missing a purpose here.
The problem is that the second set of updates is reversed, so the imaginary part updates first and the real part disregards it afterwards, breaking the plot again.
I found out that sometimes the second set of updates occurs before `line.update()` is called [here](https://github.com/hyperspy/hyperspy/blob/a40e65b19c3f12877f689a81046fb430b1ab358b/hyperspy/drawing/signal1d.py#L142) (most of the time it happens afterwards), maybe that helps in locating where this comes from.
For (imag, real, real, imag), my fix works, for (real, imag, imag, real) it does not...

So what I finally did was replacing [these](https://github.com/hyperspy/hyperspy/blob/a40e65b19c3f12877f689a81046fb430b1ab358b/hyperspy/drawing/mpl_hse.py#L95-L116) lines with the following:
```python
is_complex = any(np.iscomplex(self.signal_data_function()))
sl = signal1d.Signal1DLine()
sl.autoscale = True if not is_complex else False
sl.data_function = self.signal_data_function
sl.plot_indices = True
if self.pointer is not None:
    color = self.pointer.color
else:
    color = 'red'
sl.set_line_properties(color=color, type='step')
# Add the line to the figure:
sf.add_line(sl)
# If the data is complex create a line in the left axis with the
# default coordinates
if is_complex:
    sl = signal1d.Signal1DLine()
    sl.autoscale = True
    sl.data_function = self.signal_data_function
    sl.plot_coordinates = True
    sl.get_complex = True
    sl.set_line_properties(color="blue", type='step')
    # Add extra line to the figure
    sf.add_line(sl)
```
Now, if (and only if) the data is complex, a second line will be added and only this second line will auto-scale while taking both ranges, real and imaginary, into account (updating twice was a waste anyway).

Tests run smoothly afterwards, but I assume this is just because plots are not tested very well.

I hope this helps :-)!

francisco-dlp:
@CodeMonkeyJan, great job! Few have the courage to through hspy's plotting code. A bit hacky of course, but the implementation is hacky anyway ;) We need to rewrite all the plotting code because it was designed before we implemented proper events and, as a consequence, it is unnecessarily complex and lacks flexibility.

Could you submit your patch in a new PR as soon as possible (we have to release v1.0.1 in 2 hours (!) so that it can be installed in the EMC workshop computers).

francisco-dlp:
@to266, following your advice I'll merge this as @CodeMonkeyJan will send his fix in a separate PR.

CodeMonkeyJan:
Thanks! I'll make a PR immediately, I assume it should go to 'RELEASE_next_patch'?

francisco-dlp:
Thanks! Yes, it should go to RELEASE_next_patch

CodeMonkeyJan:
Done! #1196

PR #1185, opened 2016-07-21
Fix get lines intensity

### LINKED ISSUE #1175: EDS `get_lines_intensities` raises a TyperError when data dtype is int
In order to perform get_lines_intensity() on the as imported .bcf files it is necessary to convert the data type to float. Otherwise I get the following error:

```python
<ipython-input-18-0c6100594c5b> in <module>()
----> 1 inten = Pt_wedge.get_lines_intensity(integration_windows=iw_Pt, background_windows=bw_Pt)
      2 inten[0].plot()

/Users/macark/Documents/hyperspy/hyperspy/_signals/eds.py in get_lines_intensity(self, xray_lines, integration_windows, background_windows, plot_result, only_one, only_lines, **kwargs)
    660                 corr_factor = (indexes[5] - indexes[4]) / (
    661                     (indexes[1] - indexes[0]) + (indexes[3] - indexes[2]))
--> 662                 img -= (bck1 + bck2) * corr_factor
    663             img.metadata.General.title = (
    664                 'X-ray line intensity of %s: %s at %.2f %s' %

/Users/macark/Documents/hyperspy/hyperspy/signal.py in __isub__(self, other)

/Users/macark/Documents/hyperspy/hyperspy/signal.py in _binary_operator_ruler(self, other, op_name)
   1554                 odata = other._data_aligned_with_axes
   1555                 if op_name in INPLACE_OPERATORS:
-> 1556                     self.data = getattr(sdata, op_name)(odata)
   1557                     self.axes_manager._sort_axes()
   1558                     return self

TypeError: Cannot cast ufunc subtract output from dtype('float64') to dtype('uint64') with casting rule 'same_kind'
```

francisco-dlp:
Fixes #1175

### DIFF
--- a/hyperspy/_signals/eds.py
+++ b/hyperspy/_signals/eds.py
@@ -644,7 +644,7 @@ class EDSSpectrum(Signal1D):
                                                            FWHM_MnKa='auto')
             element, line = utils_eds._get_element_and_line(Xray_line)
             img = self.isig[window[0]:window[1]].integrate1D(-1)
-            if np.issubdtype(img.data, np.integer):
+            if np.issubdtype(img.data.dtype, np.integer):
                 # The operations below require a float dtype with the default
                 # numpy casting rule ('same_kind')
                 img.change_dtype("float")
--- a/hyperspy/_signals/eds.py
+++ b/hyperspy/_signals/eds.py
@@ -644,7 +644,7 @@ class EDSSpectrum(Signal1D):
                                                            FWHM_MnKa='auto')
             element, line = utils_eds._get_element_and_line(Xray_line)
             img = self.isig[window[0]:window[1]].integrate1D(-1)
-            if np.issubdtype(np.int32, np.integer):
+            if np.issubdtype(img.data, np.integer):
                 # The operations below require a float dtype with the default
                 # numpy casting rule ('same_kind')
                 img.change_dtype("float")
--- a/hyperspy/_signals/eds.py
+++ b/hyperspy/_signals/eds.py
@@ -644,6 +644,10 @@ class EDSSpectrum(Signal1D):
                                                            FWHM_MnKa='auto')
             element, line = utils_eds._get_element_and_line(Xray_line)
             img = self.isig[window[0]:window[1]].integrate1D(-1)
+            if np.issubdtype(np.int32, np.integer):
+                # The operations below require a float dtype with the default
+                # numpy casting rule ('same_kind')
+                img.change_dtype("float")
             if background_windows is not None:
                 bw = background_windows[i]
                 # TODO: test to prevent slicing bug. To be reomved when fixed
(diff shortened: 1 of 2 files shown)


hyperspy/_signals/eds.py
```diff
@@ -644,6 +644,10 @@ def get_lines_intensity(self,
                                                            FWHM_MnKa='auto')
             element, line = utils_eds._get_element_and_line(Xray_line)
             img = self.isig[window[0]:window[1]].integrate1D(-1)
+            if np.issubdtype(np.int32, np.integer):
```
to266: This is always true.
Probably want something like
```python
if not np.can_cast(img.data, 'float'):
    img.change_dtype("float")
```

francisco-dlp: Well spotted. I meant:

```python

np.issubdtype(img.data, np.integer)

```
The point is to change the dtype only if needed. Would this be ok?

to266: The issue is strange - we actually should be checking if `float` can be casted into whatever is the `dtype` of the data... Which is `False` if it's integer, at least for now. If numpy changes it's defaults, then probably that test (not what I actually have in the above example, but the reverse) might be more robust than just blindly converting all ints to floats though

francisco-dlp: The thing is, while what you propose will indeed convert e.g. bool to float, preventing an error, I think that only in the case of int we should actually prevent an error, as if somebody tries to apply this to bool data there's something wrong going on.

francisco-dlp:
@to266, is this ok now?

hyperspy/_signals/eds.py
```diff
@@ -644,6 +644,10 @@ def get_lines_intensity(self,
                                                            FWHM_MnKa='auto')
             element, line = utils_eds._get_element_and_line(Xray_line)
             img = self.isig[window[0]:window[1]].integrate1D(-1)
+            if np.issubdtype(img.data, np.integer):
```
to266: This command requires two dtypes. It does not fail because the test does not actually reach this part of the code (i.e. tests something else)

hyperspy/tests/signal/test_eds_tem.py
```diff
@@ -265,6 +265,17 @@ def test_intensity(self):
             [0.5, 0.5],
             atol=1e-1)
 
+    def test_intensity_dtype_uint(self):
+        s = self.signal
+        s.data *= 1E5
+        s.change_dtype("uint")
+
+        np.testing.assert_allclose(
+            [i.data[0] for i in s.get_lines_intensity(
```
to266: You have to supply `background_windows` to trigger the issue that you attempt to fix.

Namely this would work:
```python
bw = s.estimate_background_windows()
np.testing.assert_allclose(
    [i.data[0] for i in s.get_lines_intensity(background_windows=bw)],
# other arguments / values
    )
```

francisco-dlp: Fixes, thanks for the very careful review!

Issue #1182, opened 2016-07-20
TestSamfireMain.test_multiprocessed fails randomly


francisco-dlp:
See e.g. [this build](https://travis-ci.org/francisco-dlp/hyperspy/builds/146081056)

to266:
https://github.com/hyperspy/hyperspy/pull/1191

francisco-dlp:
Still failing see e.g. [this](https://travis-ci.org/francisco-dlp/hyperspy/jobs/147705930). Wouldn't just seed fix the issue for good?

to266:
That's how I fixed it for 64bits, yes. 
As you can see even now it actually uses the seed, it just does not work well enough for 32bits:
https://github.com/hyperspy/hyperspy/pull/1191/files#diff-e341fa460a1e5ab7761895d60672eef2R310

I'll see what I can do

ericpre:
This issue seems to be gone, because it didn't occur for a while!!

PR #1191, opened 2016-07-26
Relax samfire test requirements

to266:
Hopefully fixes  https://github.com/hyperspy/hyperspy/issues/1182 .
Lets at most 3 pixels to fail randomly in a real-life samfire test. Mostly happens on 32bit machines

### DIFF
--- a/hyperspy/tests/samfire/test_samfire.py
+++ b/hyperspy/tests/samfire/test_samfire.py
@@ -319,19 +319,23 @@ class TestSamfireMain:
         samf.strategies.remove(1)
         samf.optional_components = [self.model[2]]
         samf.start(fitter='mpfit', bounded=True)
-        nt.assert_true(np.all(samf.metadata.marker == -np.ones(self.shape)))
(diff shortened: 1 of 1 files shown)


francisco-dlp:
Can you explain why it happens?

to266:
One of the tests is a full real-world example of SAMFire usage, where a synthetic slightly random data is generated, and then fitted.

The particular test checks that all pixels have been successfully fitted (i.e. the `marker == -1` for all navigation space). However due to the randomness (and mostly me not being able to test this on 32-bit machine easily) I can't always guarantee that the solution will be found for all pixels - after all it relies on chi-squared, which is just a probability distribution.

Essentially in this test occasionally a couple of pixels fail to find the appropriate solution. I kind of fine-tuned the parameters / random seed for the 64bit machines so that it always finds the correct solution (at least I think it worked and all the failed tests are on 32 bits), but am too lazy to do the same for the other platform...

francisco-dlp:
Seems reasonable not to expend more time on this.

PR #1194, opened 2016-07-26
Fix slicing out of range in units

### LINKED ISSUE #372: BUG: slicing not working as expected when out of range
Slicing with float do not always return an error when slicing outside the limit. 

For s:

>>> s = signals.Spectrum(range(1000))
>>> s.axes_manager[0].scale = 0.01
>>> s.axes_manager[0].units = "keV"

Then

>>> s[15.]

Return ``ValueError: The value is out of the axis limits`` as expected. But 

>>> s[15.:16.]

Return a spectrum but it should return a ``ValueError``

francisco-dlp:
Fixes #372

### DIFF
--- a/hyperspy/axes.py
+++ b/hyperspy/axes.py
@@ -227,30 +227,34 @@ class DataAxis(t.HasTraits):
 
         if isfloat(step):
             step = int(round(step / self.scale))
-        start_out_of_range = False
         if isfloat(start):
             try:
                 start = v2i(start)
             except ValueError:
-                # The value is below the axis limits
-                # we slice from the start.
-                start = None
-                start_out_of_range = True
+                if start > self.high_value:
+                    # The start value is above the axis limit
+                    raise IndexError(
+                        "Start value above axis high bound for  axis %s."
+                        "value: %f high_bound: %f" % (repr(self), start,
+                                                     self.high_value))
+                else:
+                    # The start value is below the axis limit,
+                    # we slice from the start.
+                    start = None
         if isfloat(stop):
             try:
                 stop = v2i(stop)
             except ValueError:
-                # The value is above the axes limits
-                # we slice up to the end.
-                if start_out_of_range:
-                    # Slicing in a interval which is out of the axis range.
-                    # In this case numpy would return an empty array
-                    # but HyperSpy doesn't support empty arrays so we raise
-                    # an error
-                    raise IndexError("Slice not in axis range for axis %s" %
-                                     repr(self))
+                if stop < self.low_value:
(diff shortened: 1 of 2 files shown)


to266:
I would expect the behaviour to be slightly different than you implemented, namely if the requested slice is a superset of the total axis, I would expect to get the full axis. 

Also, while testing this I probably found another bug:
```python

In [2]: s = hs.datasets.example_signals.EDS_TEM_Spectrum()

In [3]: s.axes_manager[0].axis[:5] #check beginning of axis to know where to slice
Out[3]: array([ 0.169315,  0.189343,  0.209371,  0.229399,  0.249427])

In [4]: s.axes_manager[0].axis[-5:] #check end of axis to know where to slice
Out[4]: array([ 19.936951,  19.956979,  19.977007,  19.997035,  20.017063])

In [5]: s.isig[-3.:20.].axes_manager[0].axis[:5] # all as expected
Out[5]: array([ 0.169315,  0.189343,  0.209371,  0.229399,  0.249427])

In [6]: s.isig[-3.:20.].axes_manager[0].axis[-5:] 
Out[6]: array([ 19.896895,  19.916923,  19.936951,  19.956979,  19.977007])
```
I would expect the last result to be 1 pixel larger, include the 19.997035 point. Rounding error?

If we were leaving out the last point in the slice, as is accustomed in python then this is strange:
```python
In [7]: s.isig[-3.:20.015].axes_manager[0].axis[-5:] # not quite the last point, but we still get the second-last
Out[7]: array([ 19.916923,  19.936951,  19.956979,  19.977007,  19.997035])
```

francisco-dlp:
Thanks, it should be fixed now.

Regarding the rounding issue, hyperspy simply uses python's round.

hyperspy/axes.py
```diff
@@ -227,30 +227,34 @@ def _get_array_slices(self, slice_):
...
+                    # The start value is above the axis limit
+                    raise IndexError(
+                        "Start value above axis high bound for  axis %s."
+                        "value: %f high_bound: %f" % (repr(self), start,
+                                                      self.low_value))
                 else:
-                    stop = None
+                    # The start value is below the axis limit,
+                    # we slice from the start.
+                    start = None
```
to266: this should be `stop`. Also in the error text above (start - > stop, above->below, high->low)

francisco-dlp: Thanks, it should be fine now.

vidartf:
FYI: Note that the rounding behavior of python3 is different from python2 ( 0.5 and 1.5 are rounded different ways).

PR #1196, opened 2016-07-27
FIX: update complex spectra plots

CodeMonkeyJan:
This is a supplementary fix for #1184.

The only thing that could be adressed is the autoscale of the complex part in `Signal1DLine.update`:
Currently there is no clipping done while taking the real part into account (see [here](https://github.com/hyperspy/hyperspy/compare/hyperspy:391024d...CodeMonkeyJan:d9eeb3e#diff-2b30bf1284fd4adec5b06d2ba1c24889R340)). Should this be changed or is it okay as is?

### DIFF
--- a/hyperspy/_signals/complex_signal.py
+++ b/hyperspy/_signals/complex_signal.py
@@ -140,7 +140,7 @@ class ComplexSignal(BaseSignal):
 
         """
         sig = self._deepcopy_with_new_data(np.angle(self.data, deg))
-        sig.set_signal_type('')
+        sig.set_signal_type("")
         if sig.metadata.General.title:
             title = sig.metadata.General.title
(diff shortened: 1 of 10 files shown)


francisco-dlp:
I can't remember what's the use of clipping (I should have added a comment there 9 years ago...), but it's likely that it's there for a reason so I would suggest to add it to the complex part. At least we'll be consistent...

CodeMonkeyJan:
I hope this is correct now. The clipping uses the same values `y1`, `y2` for both axes now. I assume this is correct, because it seems to work on the x-bounds. Please correct me if I'm wrong. It seems to work for me.

francisco-dlp:
Great, thanks!

Issue #141, opened 2014-01-22
Nosetests using image comparisons


magnunor:
As we are implementing more and more visualization functions, it would be nice to write some tests which actually check if they plot things correctly.

Matplotlib has a way to do this: http://matplotlib.org/devel/testing.html (Writing an image comparison test). And it is fairly easy to write the tests themselves.

What do you think?

tjof2:
See #1253

ericpre:
Done in #1253.

PR #1200, opened 2016-07-27
Try testing plot

ericpre:
Attempt to test plotting functionality using matplotlib `image_comparison` decorator. The idea of this PR is to see if it's works with travis/appveyor and if it's not a pain to use it.
[Matplotlib testing documentation](http://matplotlib.org/1.5.1/devel/testing.html?highlight=testing#writing-an-image-comparison-test).

In short:
- as far as I understood, the test files have to be in the root tests folder (not a subfolder).
- the decorator takes the corresponding references images in a subfolder (same name than the test file) of the folder `baseline_images`.
- the result is saved in the `result_images` folder.

~~Some tests are plotting some stuff, which are interacting with the tests added in this PR, so for now I have commented them, so the plotting tests doesn't fail.~~ Actually, using the `cleanup` decorator of the matplotlib library do the trick, so I un-commented these tests.


ericpre:
Travis and appveyor seem to be happy with the plotting tests! May be worth checking if the tests also pass on different environments, in case they depend on the local settings and how the tests are run?

If you are happy with the principle, I can add more tests, try to fix the commented tests and add the documentation to the developer guide.

ericpre:
#141

ericpre:
Actually, this is also possible to have the plotting tests in subfolders of the "root" tests folder, but the way how the `image_comparison` decorator works is a bit silly because we have to create quite a few subfolders. For example, for a test_plot.py file in `/tests/drawing`, we have to put the reference images in the `tests/drawing/baseline_images/drawing/test_plot` folder.

francisco-dlp:
In MPL 2.0 they're changing the style, would that breaks all these tests?

ericpre:
Yes, I guess it will... however, we should be able to use the 1.x style by setting `mpl.style.use('classic')` in the tests.
http://matplotlib.org/style_changes.html#everything-else

For the transition, we could test plotting with the 1.x styles for a while and at some point change to 2.x.

If we get the plot testing working, then I can re-send a PR for #1106 without being worry about breaking any plotting, which looks a bit delicate! I also have a few bug fixes for plotting model components, which will need plotting tests.

francisco-dlp:
Would it be possible to have a "master" function that recreates all the test png files? This would be handy to fix tests that get broken because of changes in mpl.

ericpre:
Actually, the plotting tests suite is the master function you are talking about!
The idea is that the very first time you run the tests, there are failing because the reference images still doesn't exit but the files are generated (in `result_images`), then you check them and if you are happy with them, you copy them to the corresponding `baseline_images` folder.
Each time, there is a change (in the tests or bug fix), it will be necessary to go through this process.

Would it be worth to run the tests for both style versions in case matplotlib 2.x is installed?

francisco-dlp:
Very nice! It would indeed make sense to keep two versions of the images, for the new and old styles, and check which version is running I think.

Also, you may like to modify `.coveragerc` which is currently ignoring the plotting code to check how exhaustively this tests the drawing code.

ericpre:
I tried to switch the matplotlib backend to `TkAgg` only for one test (drawing.test_create_figure) by switching it back to `agg` and it first failed only for travis on linux! Appveyor and travis on osx were passing... Looking at the error log for the corresponding [commit](https://github.com/hyperspy/hyperspy/pull/1200/commits/0699bef3bed4ac882f435f41d1f3fb8bd2025579), there are much error than just the test that I changed, so it could implies that switching back to `agg` is not working. It seems to be fine for now, but it may be something to be aware of!!
After tweaking the .travis.yml, they all seem to be happy!

ericpre:
I think this is ready from my side!
Coverage decreased by 3%, which is not really surprising because this is now covering all the drawing stuff. A quick summary of the coverage of the drawing stuff:

- plotting signal: fairly good
- markers: not great
- widget: quite bad
- plotting model/components: bad

I may do a PR later to test model/components plotting.

The `_plot_components()` and `_plot_loadings() ` functions in [drawing/signals.py](https://github.com/hyperspy/hyperspy/blob/master/hyperspy/drawing/signal.py) doesn't seem to be use anywhere. What to do with that?
Same in [drawing/signals1d.py](https://github.com/hyperspy/hyperspy/blob/master/hyperspy/drawing/signal1d.py).

ericpre:
Resubmitted to merge into `RELEASE_next_patch` in #1253.

Issue #1212, opened 2016-07-29
Several method that should return Signals with signal dimension 0 don't


francisco-dlp:
The following methods should return 0-d signals but they don't:

* estimate_peak_width
* estimate_thickness
* estimate_elastic_scattering_intensity
* estimate_elastic_scattering_threshold
* estimate_zero_loss_peak_centre
* get_lines_intensity

Furthermore all of them keep the original signal_type when they shouldn't.

francisco-dlp:
Fixed in #1214

to266:
should variance be of the original signal type as well? The dimensions are OK though

Issue #1211, opened 2016-07-29
BaseSignal.__call__ does not return a view with signal dimension is 0


francisco-dlp:
For example:

```python
>>> s = hs.signals.BaseSignal(np.zeros((1)))
>>> s.axes_manager.set_signal_dimension(0)
>>> s()[:] = 2
>>> s.data
array([ 0.])
```

francisco-dlp:
Fixed in #1214

Issue #1210, opened 2016-07-29
map does nothing when `signal_dimension == 0`


francisco-dlp:
For example:

```python
>>> s = hs.signals.BaseSignal(np.zeros((1)))
>>> s.axes_manager.set_signal_dimension(0)
>>> s.map(lambda x: x + 1)
>>> s.data
array([ 0.])
```

francisco-dlp:
Fixed in #1214

PR #1214, opened 2016-07-29
Multiple bug fixes

### LINKED ISSUE #1213: EELSSpectrum.estimate_elastic_scattering_intensity not working as expected
As `estimate_elastic_scattering_intensity` is used by default by `fourier_ratio_deconvolution`, the latter doesn't work as expected either.

Example:

```python

>>> from hyperspy.tests.signal.test_eels import Test_Estimate_Elastic_Scattering_Threshold
>>> t  = Test_Estimate_Elastic_Scattering_Threshold()
>>> t.setUp()
>>> t.signal.estimate_elastic_scattering_intensity(t.signal.estimate_elastic_scattering_threshold()).data
array([[ 0.,  0.],
       [ 0.,  0.],
       [ 0.,  0.]])
```

francisco-dlp:
#1213, #1212, #1211, #1210

### DIFF
--- a/hyperspy/signal.py
+++ b/hyperspy/signal.py
@@ -2594,7 +2594,7 @@ class BaseSignal(FancySlicing,
             getitem[axis] = slice(None)
         for i in range(data.shape[unfolded_axis]):
             getitem[unfolded_axis] = i
-            yield(data[getitem])
+            yield(data[tuple(getitem)])
 
     def _remove_axis(self, axes):
         am = self.axes_manager
--- a/hyperspy/signal.py
+++ b/hyperspy/signal.py
@@ -3247,7 +3247,7 @@ class BaseSignal(FancySlicing,
             scale.add(self.axes_manager.signal_axes[i].scale)
             units.add(self.axes_manager.signal_axes[i].units)
         if len(units) != 1 or len(scale) != 1:
-            warnings.warn(
+            _logger.warning(
                 "The function you applied does not take into "
                 "account the difference of units and of scales in-between"
                 " axes.")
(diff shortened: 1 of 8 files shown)


hyperspy/signal.py
```diff
@@ -2576,18 +2576,23 @@ def _iterate_signal(self):
...
         data = self.data.reshape(new_shape)
+        getitem = [0] * len(data.shape)
+        for axis in axes:
+            getitem[axis] = slice(None)
         for i in range(data.shape[unfolded_axis]):
-            getitem = [0] * len(data.shape)
-            for axis in axes:
-                getitem[axis] = slice(None)
             getitem[unfolded_axis] = i
             yield(data[getitem])
```
to266: `tuple(getitem)`

francisco-dlp: Would it make any difference? (Computing the tuple probably takes as much time or more than the performance improvement of using tuple vs list)

to266: it's not for performance - slicing with a list vs slicing with a tuple are different things entirely!

francisco-dlp: I agree, but not when the list always contains a slice, then they're equivalent. I'll change it though if that makes the code clearer.

to266:
One of the new tests does not pass

francisco-dlp:
It was due to a warning. I've moved it to the logger and now it should pass.

to266:
it's due to the value being different than expected, and still fails....

to266:
Still bad :(

francisco-dlp:
I think that this is finally fine.

PR #1215, opened 2016-07-29
Fix interactive onset_energy setting

francisco-dlp:
When using `enable_adjust_position` to set the onset energy of an EELSEdge  component, sometimes (randomly) an error occurs that leaves the component in a broken states. The bug is caused by the random execution of events. This fixes the issue. The following reproduces the issue:

```python
%matplotlib qt4
import numpy as np
import hyperspy.api as hs

s = hs.signals.EELSSpectrum(np.random.random((32,32,1000)))
s.set_microscope_parameters(100, 10, 20)
s.add_elements(["C", "O"])
m = s.create_model()
m.plot()
m.enable_adjust_position()
```

### DIFF
--- a/hyperspy/_components/eels_cl_edge.py
+++ b/hyperspy/_components/eels_cl_edge.py
@@ -306,6 +306,15 @@ class EELSCLEdge(Component):
         """Returns the number of counts in barns
 
         """
+        shift = self.onset_energy.value - self.GOS.onset_energy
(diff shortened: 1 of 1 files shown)


to266:
It would be nice if you could post a minimal example to test if this works..

francisco-dlp:
Good point, I've updated the first post with an example.

PR #1219, opened 2016-07-29
Enh lazy signal

to266:
[resubmit of #1102 ] 
Adds a new class of "Lazy" signals that only actually operate on data (and even access it) when required or explicitly told to do so.

The intended workflow is as follows:
1. Load the data lazily.
2. Perform calculations (will most likely be fast / instantaneous) or operations
3. Save the result signal to a file. Then the operations are performed by `dask`, and results written to file.

One of the drawbacks is that the `dask.array.Array` (which all this PR relies upon) does not support slicing assignment (i.e. `a[3] += 2` does not work, have to do something like `a = da.concatenate(a[:3], a[3]+2, a[3:])`), so not everything works as seamlessly.

Also, lazy signals mean plotting is slow (since it's calculated on requests), but any-size data should be feasible

NEW things
--------
 - __`Lazy*Signal` classes__: mirror the normal signal classes, but perform the operations lazily.
 - __`nan*()` methods__: just like `numpy.nanmax` and others. Previously were missing...
 - __"ONMF" decomposition__: Online NMF implementation as per [this paper by Zhao et.al](https://arxiv.org/abs/1604.02634). In particular it's the OPGD implementation in the paper. __Not yet thoroughly tested__
 - __`as_lazy()`__ method: All signals now have `as_lazy()` method, which does the obvious. Will probably be used to convert any other signal into its lazy implementation.
 - __`compute()`__ method: Lazy signals can be converted to conventional ones using this method. The better (and more realistic) workflow saves the end lazy signal to a file (since it presumably does not fit in memory all at once).

Behaviour Changes
--------
 - __CircleROI__ for `LazySignal` sets the elements outside the ROI to `np.nan` instead of using a masked array, since `dask` does not support masked arrays. Essentially the motivating factor for `nan*` methods.
 - __Lazy loading__: deprecated all `memmap`, `mmap`, `load_to_memory` and similar kwargs, and instead left only the new `lazy` kwarg. This means user does not have to pay attention to the format of the file, and still be able to load it lazily, as it's handled by the format reader.
 - __Stacking__ : Majorly rewritten, can now perform the stacking lazily. Also supports stacking numbers (floats, integers, complex numbers) that get broadcasted as appropriate.
 - __`get_histogram`__: lazy implementations do not support knuth' and 'blocks' bins.
 - __`ragged` argument in `map`__: the `ragged` is necessary for lazy signals, but optional (i.e. can be determined automatically while running) for normal ones. If `ragged`, the results of the `map` are not assumed to be of similar shape or even numpy.arrays, and can be any python object.

Notable code changes (relevant for developers)
-------------
 - Rewrote a bunch of iterating algorithms to use `signal.map`, since it's generalized and works for both conventional and lazy signals.
 - `misc.signal_tools.broadcast_signals` added. Used in *nary functions, so no need for new tests.
 - `lazifyTestClass` decorator for test classes. Creates new `test_lazy_*` methods from all existing `test_*` methods, where any signal in the class is casted as lazy. Allows reusing most of the regular tests for lazy signals. Can overwrite other class attributes as well.

TODO:
- [x] stacking
- [x] __most__ existing readers can load lazily
- [x] lazy/iterating decomposition tests (mainly O(R)NMF)
- [x] `bfc` lazy loading
- [x] developer guide update
- [x] "Big Data" chapter in the user guide
- [x] Holography signals


hyperspy/_signals/complex_signal.py
```diff
@@ -123,7 +114,8 @@ def change_dtype(self, dtype):
             raise AttributeError(
                 'Complex data can only be converted into other complex dtypes!')
 
-    def angle(self, deg=False):
+    @format_title('angle')
+    def angle(self, angle, deg=False):
```
francisco-dlp, line 118: I think that it is preferable not to use a decorator for the methods that take args because it breaks IPython's keyword autocompletion.

to266, line 118: I'll have to check, but it's possible that it does not break this - this particular code is only called internally, and the public interface is [here](https://github.com/to266/hyperspy/blob/ENH_lazy_signal/hyperspy/_signals/complex_signal.py#L256) and [here](https://github.com/to266/hyperspy/blob/ENH_lazy_signal/hyperspy/_signals/complex_signal.py#L298)

hyperspy/_signals/eels.py
```diff
@@ -899,11 +971,25 @@ def power_law_extrapolation(self,
...
+                                    axis=axis.index_in_array)
+        else:
+            # just old code
+            s.data = np.zeros(new_shape)
+            s.data[..., :axis.size] = self.data
         s.get_dimensions_from_data()
-        s.data[..., :axis.size] = self.data
         pl = PowerLaw()
         pl._axes_manager = self.axes_manager
+        # TODO: make work lazily
```
francisco-dlp: ?

hyperspy/_signals/eels.py
```diff
@@ -919,10 +1005,29 @@ def power_law_extrapolation(self,
             factor = s.axes_manager[-1].scale
         else:
             factor = 1
-        s.data[..., axis.size:] = (
-            factor * pl.A.map['values'][..., np.newaxis] *
-            s.axes_manager.signal_axes[0].axis[np.newaxis, axis.size:] ** (
-                -pl.r.map['values'][..., np.newaxis]))
+        if self.metadata.Signal.lazy:
```
francisco-dlp: Why is lazy in metadata? It's not a property of the signal, is a way to deal with it.

tjof2:
@to266 I hadn't followed the ORPCA addition/removal - what was the issue?

to266:
For some reason it seemed to be incredibly slow and effectively useless, so I removed it from the pushed version, so that if someone checks this branch, it does not confuse people.

Now the reason **why** it was slow (I'm talking around 10 minutes / image) may actually be because I did something wrong / stupid.

Now that you've noticed, I would very much welcome help! In particular:
- Is the `Ehat` variable required? It seems to just simply double the memory requirements, and in the end is never used.
- You perform data scaling before actually iterating through it. That does not seem to be compatible with the "online" approach that you suggested this is made for. Is it a strict requirement? Would normalizing each incoming data chunk/image/whatever work, if we do need this normalization?
- All hyperspy signals have the `_iterate_signal` method, which essentially returns an iterator that feeds the data in exactly the online approach. If someone (most likely you and me) adapted the current implementation of ORPCA to work with such iterator (maybe even in the general case, for non-lazy signals, since the `_iterate_signal` overhead is minimal), it would end up working seamlessly for both lazy and regular signals.

Given this, and since you're the author of the ORPCA we have in hyperspy, you are more than welcome to let me know what needs to be done, and I'll just ask once/if I'm stuck. Currently I was just poking in the dark because not only my signals are lazy, I am too :)

tjof2:
- `Ehat` is required if you are interested in the sparse component of the decomposition - e.g. if you wanted foreground/background separation of a video
- Data scaling was actually added to simplify the problem of the `lambda1` and `lambda2` parameters - which I think are also the reason for the slowdown. If you select them badly, the algorithm is slow. Usually the defaults are pretty fast, but assume the data is in the range `[0,1)`. So this could be removed if we find a solution to the parameter problem.
- (True data-scaling in an online manner is only needed for the Poisson normalization)

Happy to help with any further questions!

to266:
@tjof2 I added the ORPCA back, this time hopefully better.

It gives almost (a question about that later) results and seems to run faster than previously.
For that I had to restructure it as a class (while still keeping the previous interface, so hopefully the rest of hyperspy functions will barely notice) in order for it to be state-full. It is now able to take both arrays and iterators to learn (`fit` in this case), so behaves more online than previously :)
The only drawback so far is that the values are not normalized, we'll have to figure this out eventually..

Now the question:
When looking at the results for me it seems the first couple of pixels (samples) are not what I would expect - they seem to fluctuate..

A dirty example here:

``` python
%matplotlib
import numpy as np
import hyperspy.api as hs
s = hs.signals.Signal2D(np.random.rand(10, 11, 12, 13)).as_lazy()
s.decomposition(15, kind='ORPCA')
s.plot_decomposition_results()
```

If you look at the loadings, the first pixel couple of pixels always seem off. Of course feel free to use real data, as then the effect seems to be much more pronounced...
Have you seen this behavior before? Is this due to the default initialization?

tjof2:
@to266 looks good. 

RE: the fluctuation, yes, that's down to the initialization and thus convergence of the algorithm. A simple solution is to re-run over the first few samples afterwards using the learnt results of `L`, `R` and `E`. 

Indeed, this was the reason for developing the QR initialization in the first place, based on some comments from @thomasaarholt in testing. See https://github.com/hyperspy/hyperspy/pull/1071#issuecomment-230851384

dnjohnstone:
What do we need to push this one forward? It would be really handy (at least for me and a couple of others) to have this in the main branch.

thomasaarholt:
It would be nice to add some documentation explaining how to use lazy signals :) I'm currently checking your lazy branch out and I don't really know where to start.

to266:
`s.as_lazy()` and then doing things as previously, hopefully!

The loading is probably the most troublesome, since we support quite a variety of formats, and only some of them allow not really loading data (in essence hdf5 and any that allow the memmap features). An you have to either know or just try and see if the format is supported. If it is, just (not really) load the data, cast the resulting signal as lazy, and that should be it, if I've implemented it.

I think stacking is still not done.

Also decomposition is different, but you know that already.

Almost all other things are automatically tested to behave identically, albeit slower (due to reading from the disk to get the data chunks when needed).

francisco-dlp:
I think that all the formats with the exception of bcf can be made to use `memmap` instead of `array` straightforwardly. (@sem-geologist, would it be possible at all to add support for this to bcf?)
I bet that `stack` (once implemented) will be one of the most used ways "generate" large datasets. Out of curiosity, how do you generate the big hdf5 files that you're working on currently?

By the way, the number of new features is piling up, so it'll be good to start planing the release of HyperSpy 1.2. @to266, would it be feasible to add this to a February release? We should also release asap a patch release, maybe next week or early in the new year as, not only a good number of bugs have been fixed, but also the installation is broken for many people due to the qt5 issue...

sem-geologist:
@francisco-dlp , bcf is going to be pain as reading and unpacking bcf is not straight forward as the rest of formats (needs to be done bit by bit) so direct memmap is quite impossible. To work around to make the stack I think in such a case the easiest solution would be to iterate thorough all bcf and convert/save them to native hdf of hyperspy and do the stack on those. As side effect the hdf files will be few times smaller (if compression is used) (thats the biggest irony of Bruker's delphi implementation of bcf, their packed and zlib compressed bcf files are few times bigger than same data saved in hdf with reasonable compression, not only such hdf fies are smaller but also opens much faster).

Talking about large datasets... I never merged bcf's into one large block, I was planing to do so with help of dusk, but had no time to fully explore the possibilities. Instead my workflow is more simple: "divide and conquer" - iterating through bcf  gathering the required information. i.e. to get maximum_peak_spectrum of whole rock slab (lets say 10x10cm, with pixel resolution of 1.5um) where we have thousands of separate bcfs covering whole region, I iterate thought bcfs, and get maximum_peak_spectrum and append to list, rest is discarded, later I stack that list and calculate max_peak_spectrum from that stack. In jupyter notebook I need to use gc extensively as it tends to overkeep objects too long, in that way I never run out of memory. After I get max peak spectrum, I decide what huge mappings I need, so I iterate again through bcf's and get those lines exported to png. Stitching those to complete huge images is completely out of hyperspy, we use two frameworks: ImageJ (based on phase comparison of fft with subpixel alignment) or my old spaghetti code Qstitch tool (which kinda works, but should be rewritten as is hardly maintainable and embarrassing) with manual stitching parameter tweak, however it uses opencv filtering capabilities (bilinearal filters simply rocks) which I didnt saw in ImageJ.

I think in future Bruker products will have no more problems as I saw they are moving to hdf5 (I think they'r new type of under column EDS detector which allows insanely better counting rates than classically side mounted SDD detectors forced them to move to something more efficient than pascal/delphi packed arrays)

to266:
@sem-geologist I could get around the memmap issue with bfc if there was a straightforward way to read only the header of the file, and then later (with a separate function / function call) - the data. Then hopefully the header would return at least the dtype and total final shape of the data that the "actual loading" function returns. In this way, the lazy signals (and "stacking") would be able to work with this by loading the bfcs behind the scene as required.

So you'd be able to lazily load (i.e. just read the headers and collect the actual loading functions, but not call them) a bunch of bfc, stack / stitch them together and use any lazy functions on the result. `dask` with `delayed` would handle the data loading when required, and the syntax of your described analysis should be somewhat more streamlined and less complex.

I don't know if your current code lends itself to being split this way, but I think that would be a good change from my point of view. Do you think you'd have some time and motivation to at least help me do this? If you feel generous feel free to PR to this branch or just hyperspy, of course. If you don't change the public API (which I'm not sure there was any, so that should not be a problem!), we can do whatever inside.

to266:
@francisco-dlp Yes, I think this is doable for February. I hoped to finish by Christmas, albeit that is clearly not happening... but much of it is here already

sem-geologist:
@to266 this way I think is even quite doable the format itself is like that. I just think additionally I probably should not load the images which are in header (bruker keeps the main hypermap in binary delphi packed array, while sem 2D images such as i.e SEI or BSE or something else like that is saved in header coded with base64). Or should I? those are not huge and they help to place, align the main hypermap. what other whole keeping formats do with simple images? I mean those too bulky xml parts

I could not find how it is going to be achieved/organized? will there be separate required lazily_read function inside library which at 'data' field put some object, which posses no data, but methods how to retrieve the array for that particular tile? 

update:
oh I see in the code the reader will get 'lazy'=True, so what the io_plugin should then return? the same dictionary except the 'data' field, or 'data' is going to be keeped empty but some other field like 'lazy' will get some object with methods of retrieving data?

sem-geologist:
I am exploring further the possibility to open sem images (not hypermap, thats easy to postpone) with delay. There is one serious problem, the data will have to be in such case read few times, else we are not saving the memory. The problem is that to get all parameters we need to get whole xml string in memory as it is chunked inside bcf. At present approach it is read and only substance of actual information is copied to python Header representing object, as images are also there they are read also at the same stage of parsing xml and after base64 decoding assigned to that python object. After this the xml object with whole xml_string is discarded from memory. Those bruker xml tends to be really bulky containing a lot of junk like colormaps, empty zeroed arrays of image size coded with base64, font describtions, scalebar postion, other elements... casually it takes 5-15MB (depending from microscope configuration and image resolution), sometimes it is quite broken (that is why bcf uses lxml, as it is the only library allowing easily deal with that broken thing) like tagged with forbidden chars, or closing tag missing or some other weirdness. Anyway if I would implement separate method in the BCF reader which would only read the image, then the xml should be read twice (by this I mean extracted from the bcf container chunks, then parsed again). So it is duable, but will cost a little bit in perfomance.

to266:
@sem-geologist [this](https://dask.readthedocs.io/en/latest/delayed-collections.html) is an excellent example that we can use for non-memmap compatible formats. In order for that `delayed`-returned object (which is essentially any function call) to be transformed to a `dask.array` (in this PRs lingo - "lazy data"), you would call [this](https://dask.readthedocs.io/en/latest/array-api.html#dask.array.from_delayed) function. As you can see, this function requires `dtype` and `shape` as well. It will also always result in an array with a single chunk - i.e. when it's loaded, it's always loaded in full. If the format lends itself to being part-loaded conveniently (maybe by reading rows / colunms / etc of the final dataset?) and without much overhead, this is where it should be implemented. Otherwise I think it's OK to always load it in full when required.

As for loading the SE/BSE images that come with the header, I think it's fine. We're not trying to solve all the problems at once, and we also assume reasonable machines - using additional tens of MB is definitely fine. 
If HyperSpy is able to load the main part of a `bfc`, but always reads the secondary images into memory, that is where our problems end - it's up to the user then what to keep and stack (likely the lazy main data), and what to discard (likely the secondary images) and when (probably once the alignments are done).

sem-geologist:
@to266 sounds like doable, My question is however will the file_reader() be preserved as is? Will there be need to provide separate function so that working behavior would not be changed in io_plugins.; or the file_reader could read some 'lazy' parameter (then 'data' would be set None), There would be very little to implement as other fields (dtype, scales) are already generated and returned in file_reader() generated dict with already working routines

francisco-dlp:
Lazy stacking, that's an excellent Xmas present for HyperSpy!

to266:
I'm actually more excited about `broadcast_signals` that I sneaked in there :)

....mostly that I did not have to write exhaustive tests, as the *nary operator tests do that

to266:
@sem-geologist have a look at [this FEI reader part](https://github.com/to266/hyperspy/blob/ENH_lazy_signal/hyperspy/io_plugins/fei.py#L583). This is in essence the way I'd like to "wrap" bfc readers. Now I have not checked your current code yet, so this may be already possible, let me know if so. Otherwise, you know what I'm aiming for.

I will also have to add some instructions to the developers guide, I guess.

Apart from that, if @tjof2 could help with the ONMF tests, this should be pretty close to ready for review. And as this touches quite a bit, the reviewer (@francisco-dlp ?) can already start having fun! :)

sem-geologist:
@to266 It is very much doable, I would like to start adding the changes which will easy both image, and hypermap lazy loading from bcf after #1357 will get merged. Should I do the lazy loading changes on separate PR, or should I clone this PR and push commits to this PR?

to266:
@sem-geologist PR to my branch, will be easier that everything is in one place

hyperspy/io.py
```diff
@@ -58,6 +58,7 @@ def load(filenames=None,
 
     Any extra keyword is passed to the corresponsing reader. For
     available options see their individual documentation.
+
```
francisco-dlp: Why is `lazy` not documented in the docstring and not an actual keyword?

hyperspy/_signals/eels.py
```diff
@@ -748,26 +792,46 @@ def richardson_lucy_deconvolution(self, psf, iterations=15, mask=None,
...
+        iterating_kw = (('signal', self),)
+        if psf.axes_manager.navigation_dimension != 0:
+            iterating_kw += (('kernel', psf),)
+        ds._map_iterate(deconv_function,
+                        iterating_kwargs=iterating_kw,
+                        kernel=psf,
+                        iterations=interations,
+                        psf_size=psf_size,
+                        show_progressbar=show_progressbar)
+        # Old code.
```
francisco-dlp: Why leaving the old code? (It's in GH is anybody ever needs it.)

tjof2:
Not all of the diff is loading for me, but is this up-to-date with the last changes to ORPCA I made @to266 ? Looking over the ONMF stuff now.

tjof2:
OK, thinking over tests for ONMF, my suggestion would be to compare it directly with standard NMF. For example, generate a random low-rank matrix:
```
m = 256
n = 256
r = 3
rng = np.random.RandomState(101)
U = rng.uniform(0, 1, (m, r))
V = rng.uniform(0, 1, (r, n))
X = np.dot(U, V)
```
**Robustness** can then be checked by adding sparse corruptions to the data, running the robust algorithm, then comparing the result with the "correct" NMF result on the uncorrupted data:
```
sparse = 0.1 # Fraction of corrupted pixels
E = 1000 * rng.binomial(1, sparse, X.shape)
Y = X + E 
# Run ORNMF on Y, and compare the factors and loadings 
# with NMF on X to some tolerance.
```
The tolerance could for example be the norm of the difference matrix:
```
normXY = np.linalg.norm(X - Y) / (m * n) # Where (m, n) is the matrix shape
nt.assert_true(normXY < tol)
```
This is the approach taken in [testing RPCA](https://github.com/hyperspy/hyperspy/blob/RELEASE_next_minor/hyperspy/tests/mva/test_rpca.py), for example.

hyperspy/learn/onmf.py
```diff
@@ -0,0 +1,241 @@
...
+                self.W = _project(self.W)
+                n += 1
+                lasttwo[0] = lasttwo[1]
+                lasttwo[1] = 0.5 * np.trace(self.W.T.dot(self.W).dot(self.A)) - \
+                        np.trace(self.W.T.dot(self.B))
+        else:
+            # Tom's approach
+            self.W -= eta * (np.dot(self.W, self.A) - self.B)
+            np.maximum(self.W, 0.0, out=self.W)
+            self.W /= max(np.linalg.norm(self.W, 'fro'), 1.0)
```
tjof2, line 250: I've not tested this, but am I right in thinking that:

- My approach takes "robust" to mean an l2-norm cost function that can deal with sparse corruptions/outliers (i.e. the same meaning as RPCA) - see https://arxiv.org/abs/1604.02634
- Your interpretation of `self.robust` deals with generalized cost functions - see https://arxiv.org/abs/1608.00075

Is that right? If so, how do the results of the two approaches differ, and when would you use one over the other? Is one much faster than the other for example?

hyperspy/_signals/lazy.py
```diff
@@ -0,0 +1,630 @@
...
+        axis2.offset = new_offset
+        s.get_dimensions_from_data()
+        if out is None:
+            return s
+        else:
+            out.events.data_changed.trigger(obj=out)
+
+    diff.__doc__ = BaseSignal.diff.__doc__
+
+    def integrate_simpson(self, axis, out=None):
```
francisco-dlp, line 282: Is the code generally applicable? If yes, why not splitting it in two so that it is easy to reuse for other purposes?

hyperspy/_signals/lazy.py
```diff
@@ -0,0 +1,630 @@
...
+        indices = product(*[range(len(c)) for c in nav_chunks])
+        for ind in indices:
+            chunk = get(data.dask, (data.name, ) + ind + (0, ))
+            if not flat_signal:
+                # TODO: check if need to reverse the signal_shape
+                chunk = chunk.reshape(chunk.shape[:-1] +
+                                      self.axes_manager.signal_shape)
+            yield chunk
+
+    def decomposition(self,
```
francisco-dlp: Unlike the other methods, the lazy decomposition interfaces differs greatly from its industrious counterpart.

* Why kind instead of algorithm? We provide several algorithms for the same kind of decomposition so `algorithm` seems to me a better choice.
* No masking capabilities: in my experience masking is one of the most useful features that HyperSpy provides when performing decomposition.
* No normalize poisson noise. I bet that when not working in "live" mode this should be relatively easy to lazyfy. decomposition is of limited use to EM without this feature.

Other little comments/suggestions:
* Why the choice of name for `refine`?
* The new algorithms should be better documented in the docstring, including full references to the papers.

hyperspy/_signals/lazy.py
```diff
@@ -0,0 +1,630 @@
...
+        significant components.
+
+        Parameters
+        ----------
+            output_dimension : int
+                the number of significant components to keep
+            kind : str
+                One of ('PCA', 'ORPCA', 'ONMF'). By default batch PCA from
+                scikit-learn.
+            get : dask scheduler
```
francisco-dlp: Document default value (`threaded.get`)

hyperspy/_signals/lazy.py
```diff
@@ -0,0 +1,630 @@
...
+                      refine=True,
+                      **kwargs):
+        """Perform Incremental (Batch) decomposition on the data, keeping n
+        significant components.
+
+        Parameters
+        ----------
+            output_dimension : int
+                the number of significant components to keep
+            kind : str
```
francisco-dlp: PCA doesn't work for me:

```python

<ipython-input-10-8c42ecde10fe> in <module>()
----> 1 s.decomposition(kind="PCA", output_dimension=5)

/home/fjd29/Python/hyperspy3/hyperspy/_signals/lazy.py in decomposition(self, output_dimension, kind, get, num_chunks, refine, **kwargs)
    544             explained_variance_ratio = ipca.explained_variance_ratio_
    545             factors = ipca.components_.T
--> 546             loadings = transform(ipca, data).compute()
    547 
    548         elif kind == 'ORPCA':

NameError: name 'data' is not defined


```

`ONMF` seems to take infinite time and `ORPCA` results are not the same as `RPCA` and seem wrong :(

hyperspy/_signals/signal1d.py
```diff
@@ -761,14 +839,9 @@ def smooth_savitzky_golay(self,
...
             axis = self.axes_manager.signal_axes[0]
-            self.data = savgol_filter(
-                x=self.data,
-                window_length=window_length,
-                polyorder=polynomial_order,
-                deriv=differential_order,
-                delta=axis.scale,
-                axis=axis.index_in_array)
-            self.events.data_changed.trigger(obj=self)
+            self.map(savgol_filter, window_length=window_length,
```
francisco-dlp, line 841: Add `parallel` to this one too?

hyperspy/_signals/signal1d.py
```diff
@@ -1119,6 +1192,8 @@ def find_peaks1D_ohaver(self, xdim=None, slope_thresh=0, amp_thresh=None,
 
         """
         # TODO: add scipy.signal.find_peaks_cwt
+        # TODO: make work with lazy signals. Currently does not due to funny
+        # output dtype
         self._check_signal_dimension_equals_one()
```
francisco-dlp: Then, shouldn't this rise a `NotImplementedError`?

hyperspy/drawing/image.py
```diff
@@ -328,7 +328,7 @@ def format_coord(x, y):
                     self.vmax != self.vmin):
                 redraw_colorbar = True
                 ims[0].autoscale()
-
+        redraw_colorbar = redraw_colorbar and self.colorbar
```
francisco-dlp: Fixing a bug here? Then report it in the issues tracker.

hyperspy/drawing/mpl_he.py
```diff
@@ -96,6 +96,15 @@ def plot_navigator(self):
         elif len(self.navigator_data_function().shape) >= 2:
             imf = image.ImagePlot()
             imf.data_function = self.navigator_data_function
+            imf.colorbar = kwargs.get('colorbar', True)
```
francisco-dlp: Adding a feature? Then report it in the header and the User Guide.

Wouldn't it be better to have a `navigator_kwds` argument?

hyperspy/io.py
```diff
@@ -103,6 +104,7 @@ def load(filenames=None,
         If an axis with this name already
         exists it automatically append '-i', where `i` are integers,
         until it finds a name that is not yet in use.
+
     mmap: bool
```
francisco-dlp: Shouldn't we deprecate `mmap` and strongly discourage its usage?

hyperspy/io.py
```diff
@@ -114,6 +116,7 @@ def load(filenames=None,
         If mmap_dir is not None, and stack and mmap are True, the memory
         mapped file will be created in the given directory,
         otherwise the default directory is used.
+
     load_to_memory: bool
```
francisco-dlp: Should we deprecate this one as well?

hyperspy/signal.py
```diff
@@ -2891,6 +2861,57 @@ def var(self, axis=None, out=None):
                                                             out=out)
     var.__doc__ %= (MANY_AXIS_PARAMETER, OUT_ARG)
 
+    def nansum(self, axis=None, out=None):
```
francisco-dlp, line 2866: All `nan*` methods are new, so they require documentation in the User Guide and the header.

sem-geologist:
talking about `dask`, is this going to be hard dependency or optional library?
if it is going to be hard dependency, you also need it in setup.py.
(it looks from code what it is hard dependence)

sem-geologist:
?!? I am not sure this is right, I see bcf is failing as it is tried to be loaded lazily, I thought this is going to be optional?

to266:
I'm not exactly sure what you thought was going to be optional...

Any way I'm currently still trying to figure out what's the most convenient approach (or implement it fully).
Currently I chose to "leave the implementation of lazy loading for the authors of the plugin", albeit I did it for many myself. It's not quite finished, as you can see - e.g. I did not adapt bfc to expect the `lazy` kwarg.

sem-geologist:
I can start working on it (bcf). You probably noticed that few days ago the minimal environment got implemented (as hyperspy is available through pip, so it behave so) in travis.
I see the tests fails on min_env as `toolz` is not present. `dask` through pip is minimal as possible and doesn't requires all dependencies, but looking from tests in min env I think `toolz` needs to be added to setup.py. (after that we will see what else is needed). I think it installs fully on conda thats why there is no such errors.

sem-geologist:
BTW, I have some more advanced bcf library on my disposition where I have implemented loading bcf in pieces (number of lines, like i.e total height 4096, and loading in 8 chunks with 512 lines each array). Are other formats doing something like that? I mean would it be useful?

sem-geologist:
Additional thing: BCF from version 2 have indexing file: the file with address offset/address for every pixel. BCF version <2 have no such indexing. Maybe it would be good idea to implement indexing of bcf, that would allow full random access, that would allow to use full power of dask. What do you think?

to266:
I'll quickly add a simple dask parsing of the current bfc implementation (I just started, but should not take too long). You'll have to PR the new bfc reader separately. I'll obviously help, whether this gets merged before or after..

As for the minimal env, don't worry about it too much now. I'll fix it later. It's likely the `toolz` dependency is from my code that (I think) it's not too difficult to replace with `functools.partial`. One by one :)

sem-geologist:
let me do it, I think I can do it this evening (bcf simple). I just want to be sure that I get it clear we are talking about whole file loading at once? no fancy things?

to266:
Yeah, it's just the full file loading at once. I think I'm done already (5 mins left, to test a bit). Make some tea/coffee, wait for my push, and then improve!

sem-geologist:
is there any plan of loading things only at ROI's ? that would save a lot of time/memory ?

to266:
LazySignal will let you do it, albeit how efficiently this gets done will depend on the underlying reader.

sem-geologist:
I pushed some fix, it was passing tests as it was pretending to do lazy, but lazily was still doing numpy :P, it works nicely. It is much slower than numpy, but well...

What about images (BSE, SEI type, not hyper, but simple images)?

to266:
Thanks.

Just to make it clear, the dask does not read - it either uses a numpy-like slicing interface to get the data, or wraps other functions (which is the bcf case). So in the end it will always be either numpy or cython code, just called when required.

As for the images, I'm not sure it's worth the effort and makes sense. If I understand it correctly, the xml is parsed into something readable all at once any way - so it's already in memory, might as well return the data directly. Even if this is not the case, I think it's safe to assume the images will not cause memory problems..

sem-geologist:
with bcf it actually also gets the numpy array, the cython part in that is only reading and parsing bcf and populating numpy array in cython; cython is needed due to extremely complicated packing of delphi arrays, and the array have to be warped into numpy bit by bit which would be very slow in python. However what dask gets is normal numpy array. I guess If I would want some slicing implementation I should subclass numpy.

francisco-dlp:
What's the status of this one? It'll be nice to merge it asap to expose it to more users before the release.

to266:
The two unticked things in the PR description are left, plus at some point I encountered a bug that I already forgot.. So have to fix that.

tjof2:
Leaving a link to my ideas on tests here https://github.com/hyperspy/hyperspy/pull/1219#issuecomment-270413742

to266:
Oh, I saw those and they're great, @tjof2. It simply required a bit more work (both testing and then fixing ONMF when the tests fail) than I had time.

thomasaarholt:
@to266 Does that mean that the tests  are good to go? :p

to266:
ONMF is not yet tested. Maybe by this evening?

tjof2:
@to266  no problem, just in case they got lost in all the other stuff! :-)

BTW did you have any comment on https://github.com/hyperspy/hyperspy/pull/1219#discussion_r94616213 ?

doc/dev_guide.rst
```diff
@@ -203,3 +203,44 @@ is checked out this should have no side effects.
 If another custom post-checkout hook is detected on PR, then setup.py tries to append
 or update the relevant part. To prevent unwanted hook generation or update you can create
 the empty file ``.hook_ignore`` in source directory (same level as setup.py).
+
+7. Adding new methods
+---------------------
+
+With the addition of ``LazySignal`` class and its derivatives, adding methods
```
tjof2: "the ``LazySignal`` class.."

doc/dev_guide.rst
```diff
@@ -203,3 +203,44 @@ is checked out this should have no side effects.
...
+that operate on the data became slightly more complicated, however we attempt
```
tjof2: "becomes" rather than "became", I think.

tjof2: And then full stop "...complicated. However, we have attempted to streamline it as much as possible."

doc/dev_guide.rst
```diff
@@ -203,3 +203,44 @@ is checked out this should have no side effects.
...
+
+The easiest way to add new methods that work both with arbitrary navigation
+dimensions and ``LazySignals`` is by using the ``map`` (or, for more control,
+``_map_all`` or ``_map_iterate``) method to map your function ``func`` across
+all "navigation pixels" (e.g. spectra in a spectrum-image).
+``map`` methods will run the function on all pixels efficiently and put the
+results back in the correct order. ``func`` is not constrained by ``dask`` and
+can use whatever code (assignment, etc.) you wish.
+
+If the new method cannot be coerced into a shape suitable for ``map`` mappings,
```
tjof2: "``map`` mappings" surely just one should work, e.g. "suitable for ``map``,"

hyperspy/_components/power_law.py
```diff
@@ -95,10 +96,13 @@ def estimate_parameters(self, signal, x1, x2, only_current=False):
 
         only_current : bool
             If False estimates the parameters for the full dataset.
+        out : bool
+            If True, returns the result arrays diretly without storing in the
```
tjof2: "directly"

tjof2: Also, "If False, estimates" -- needs a comma

hyperspy/external/astroML/histtools.py
```diff
@@ -294,3 +300,152 @@ def histogram(a, bins=10, range=None, **kwargs):
...
+    the histogram.
+
+    Parameters
+    ----------
+    a : array_like
+        array of data to be histogrammed
+    bins : int or list or str (optional)
+        If bins is a string, then it must be one of:
+        'scotts' : use Scott's rule to determine bins
+        'freedman' : use the Freedman-diaconis rule to determine bins
```
tjof2: Diaconis is capitalized as well I think.

hyperspy/io.py
```diff
@@ -58,6 +58,7 @@ def load(filenames=None,
 
     Any extra keyword is passed to the corresponsing reader. For
```
tjof2: spelling of corresponding :)

hyperspy/learn/onmf.py
to266: yes for the first point, but the "robust" option in fact uses OPGD alg. from [this](https://arxiv.org/abs/1604.02634) paper (not the generalized cost functions, but the one before).

I have not tested differences thoroughly, however generally l2-norm is always faster and seems to give marginally better results. At this point the original piece of code is only there because it's not in my way + at least in theory we have the full algorithm from the paper.

tjof2: OK I follow now!

to266:
Ok, testing ONMF is difficult. While I agree that the method @tjof2 suggested is sound, if the decomposition results are compared to the known values, even the normal nmf fails this. Slightly modified example with obvious loading maps for humans:
```python
m = 1024
n = 1024
r = 3
rng = np.random.RandomState()
ls = np.linspace(0, 1, m)
U = np.stack([(ls-.5)**2, ls, np.cos(ls), ]).T
V = rng.uniform(0, 1, (r, n))
X = np.dot(U, V)

sparse = 0.1 # Fraction of corrupted pixels
E = 1000 * rng.binomial(1, sparse, X.shape)
Y = X + E 

# "Working" nmf we have
s = hs.signals.Signal1D(X)
s.decomposition(algorithm='nmf', output_dimension=r)

# check visually that the loadings are what we put in: -- they're not
hs.plot.plot_signals([hs.signals.Signal1D(U.T), s.get_decomposition_loadings()])
```

In fact the ONMF returns something slightly closer to the original maps we put in - both by eye and some measure, like the following:
```python
# With corrupted data:
s1 = hs.signals.Signal1D(Y).as_lazy()
s1.decomposition(output_dimension=r, algorithm='ONMF', bounds=False, reproject=True,)

# Without corrupted data:
s2 = hs.signals.Signal1D(X).as_lazy() 
s2.decomposition(output_dimension=r, algorithm='ONMF', bounds=False, reproject=True)

def compare(a, b):
    n1, n2 = list(map(np.linalg.norm, [a, b]))
    return np.linalg.norm((a/n1) - (b/n2))

# NMF:
compare(U, s.learning_results.loadings)
# 0.92978521410146364

# ONMF, corrupted:
compare(U, s1.learning_results.loadings)
# 0.87308110253159388

# ONMF, clean data:
compare(U, s2.learning_results.loadings)
# 0.6511903690706845
```
I would prefer to merge this with some kind of "EXPERIMENTAL" banner, or something similar, since making sure ONMF works well and is tested is a bit too expensive in terms of time for me. 

Another option is I could remove it altogether or not document it, so that unless you know about it (and hence know it's not tested), you can't find it. I would prefer not to do this, since I've already run this on some SPED data, and the results looked (visually) indistinguishable from other NMF results (keeping in mind there was no way to run normal NMF on the full dataset due to obvious reasons)

tjof2:
@to266  is there anything from scikit-learn that might help identify this issue? https://github.com/scikit-learn/scikit-learn/blob/14031f65d144e3966113d3daec836e443c6d7a5b/sklearn/decomposition/tests/test_nmf.py

Perhaps the way the (random) matrices are initialized?

I also agree some "EXPERIMENTAL" banner might be a short-term solution.

francisco-dlp:
What about testing it in a different way? For example, you could test that the results on a given dataset are close enough to the results of the original matlab code running on the same data. We have to test that the code works as intended by the authors of the algorithm, no that the algorithm itself does the right thing ;)

hyperspy/tests/mva/test_onmf.py
```diff
@@ -0,0 +1,43 @@
...
+
+sparse = 0.1 # Fraction of corrupted pixels
+E = 1000 * rng.binomial(1, sparse, X.shape)
+Y = X + E 
+
+def test_default():
+    W, H = onmf(X, r)
+    res = compare(np.dot(W, H), X.T)
+    print(res)
+    assert res < 0.06
```
tjof2, line 25: Does a tolerance of 1e-2 not work? :-/

to266: I think `res` in this case is something like 0.056, depending on machine. Also with these tests it seemed the "robust" (i.e. original paper-like) normalisation did in fact give lower `res` values.

tjof2: Hmm OK, just wondering what needs to be done to reduce it to a more preferable value (e.g. 1% error) but probably not that important - if it works well on actual data that's good!

tjof2:
@to266 (and @francisco-dlp), having played around with this myself yesterday, I'm happy with the ONMF implementation (it does what I want at least!) - shame about the tests as they are, but otherwise it's a :+1: from me!

Regarding documentation for (OR)NMF, I'm happy to write this in a future PR, but I'm not free to do so now. Basically, happy to build on anything you write!

to266:
Thanks @tjof2 , I appreciate this. I will add some back-bones documentation in due time, so hopefully that won't be a major task from your side.

Currently trying to merge #1413 preemptively..

doc/user_guide/big_data.rst
```diff
@@ -0,0 +1,229 @@
...
+    >>> siglist = hs.load("*.hdf5")
+    >>> s = hs.stack(siglist, lazy=True)
+    >>> # Or do that when loading
+    >>> s = hs.load("*.hdf5", lazy=True, stack=True)
+
+Casting signals as lazy
+^^^^^^^^^^^^^^^^^^^^^^^
+
+To convert a regular HyperSpy signal to a lazy one (so that any future
+operations are only performed lazily) we added
```
thomasaarholt: Is there a method to reverse this as well?

thomasaarholt: Ah, just spotted `s.compute()` below.

doc/user_guide/big_data.rst
```diff
@@ -0,0 +1,229 @@
...
+data in an "online" manner (i.e. only loading each element once or twice),
+enabling performing the decompositions on very large datasets. All of the
+algorithms are only approximations to the original implementations, some
+(ONMF in particular) sparsely tested.
+
+To facilitate the usual HyperSpy workflows as well as possible,
+:py:meth:`~._signals.lazy.LazySignal.decomposition` method of the
+``LazySignal`` offers alternative implementations of:
+
+* **PCA** (``algorithm='PCA'``): performs the
```
thomasaarholt: Mention ORPCA as well? It takes care of EELS outliers (spikes) as well.

to266: It's mentioned a couple of lines below, however looking at this again I'll rephrase it anyway

to266:
Ok, I think this is now really ready for review.
The CI fails because I merged https://github.com/hyperspy/hyperspy/pull/1413 here, and apparently both travis and appveyor environments are not set up properly. I'll leave the fix to @francisco-dlp .

tjof2:
@to266 I'll look tomorrow!

francisco-dlp:
I have fixed the issues with #1413. However, the SAMFire tests fail in #1413 when running in parallel and all tests fail here due to the lazy signal tests. All those tests pass when not running the tests in parallel (just remove ``-n 8`` when calling ``py.test``).

doc/user_guide/big_data.rst
```diff
@@ -0,0 +1,228 @@
...
+
+The LazySignal idea
+-------------------
+
+In usual (non-lazy) HyperSpy signals the actual data (accessed via the field
+``s.data``) is most often loaded into the computer memory for fast access and
+processing. While such behaviour indeed results in best performance, it
+obviously requires at least as much computer memory as the data, often twice
+that to store the results of any computation. This becomes a significant
+problem if processing large amounts of data on consumer-oriented hardware.
```
tjof2: "Standard HyperSpy signals load the data (accessed via the field ``s.data``) into memory for fast access and processing. While this behaviour gives good performance in terms of speed, it obviously requires at least as much computer memory as the dataset, and often twice that to store the results of subsequent computation. This can become a significant problem when processing very large datasets on consumer-oriented hardware."

doc/user_guide/big_data.rst
```diff
@@ -0,0 +1,228 @@
...
+
+    >>> siglist = hs.load("*.hdf5")
+    >>> s = hs.stack(siglist, lazy=True)
+    >>> # Or do that when loading
+    >>> s = hs.load("*.hdf5", lazy=True, stack=True)
+
+Casting signals as lazy
+^^^^^^^^^^^^^^^^^^^^^^^
+
+To convert a regular HyperSpy signal to a lazy one (so that any future
```
tjof2: "To convert a regular HyperSpy signal to a lazy one such that any future operations are only performed lazily, use the :py:meth:`~.signal.BaseSignal.as_lazy` method:"

doc/user_guide/big_data.rst
```diff
@@ -0,0 +1,228 @@
...
+    >>> sl
+    <LazySignal1D, title: , dimensions: (3|50)>
+
+
+
+Constraints
+-----------
+
+A couple of constraints to ``LazySignal`` when compared to normal HyperSpy
+signals.
```
tjof2: "There are a number of constraints when using ``LazySignal`` in contrast to standard HyperSpy signals."

doc/user_guide/big_data.rst
```diff
@@ -0,0 +1,228 @@
...
+:py:meth:`~._signals.lazy.LazySignal.decomposition` method of the
+``LazySignal`` offers implementations of:
+
+* **PCA** (``algorithm='PCA'``): performs the
+  `IncrementalPCA <http://scikit-learn.org/stable/modules/generated/sklearn.decomposition.IncrementalPCA.html#sklearn.decomposition.IncrementalPCA>`_
+  from ``scikit-learn``.
+* **NMF** (``algorithm='ONMF'``): runs Online Robust NMF, as per "OPGD"
+  algorithm in `this paper by Zhao et. al
+  <http://ieeexplore.ieee.org/stamp/stamp.jsp?tp=&arnumber=7472160&isnumber=7471614>`_.
+* **ORPCA** (``algorithm='ORPCA'``): runs Online Robust PCA, is also available
```
tjof2: I'd put ORPCA above ONMF here. Also as has been mentioned elsewhere I think, ONMF needs documenting in the Machine Learning section.

doc/user_guide/big_data.rst
```diff
@@ -0,0 +1,228 @@
...
+* **CircleROI** sets the elements outside the ROI to ``np.nan`` instead of
+  using a masked array, because ``dask`` does not support masking. As a
+  convenience, ``nansum``, ``nanmean`` and other ``nan*`` signal methods were
+  added to mimic the workflow as closely as possible.
+
+
+Data processing with LazySignal
+-------------------------------
+
+Despite the aforementioned constraints, most of the usual HyperSpy operations
```
tjof2: "Despite the constraints, most HyperSpy operations can be performed lazily. Important points to note are:"

doc/user_guide/big_data.rst
```diff
@@ -0,0 +1,228 @@
...
+    >>> s = s + 1 # instead of s += 1
+    >>> # or, even better
+    >>> s1 = s + 1
+
+Machine learning (decomposition)
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+:ref:`Machine learning<ml>` often performs large matrix manipulations,
+requiring significantly more memory than just storing the data. Lazy HyperSpy
+signals attempted to provide similar alternatives. These algorithms read the
```
tjof2: "attempt"

doc/user_guide/big_data.rst
```diff
@@ -0,0 +1,228 @@
...
+data in an "online" manner (i.e. only loading each element once or twice),
+enabling performing the decompositions on very large datasets. All of the
```
tjof2: "meaning the decompositions can be performed on large datasets."

doc/user_guide/big_data.rst
```diff
@@ -0,0 +1,228 @@
...
+algorithms are only approximations to the original implementations, some
+(ONMF in particular) sparsely tested.
```
tjof2: This isn't true - ORPCA is taken directly from the original implementation with some additions and improvements for faster processing!

doc/user_guide/big_data.rst
```diff
@@ -0,0 +1,228 @@
...
+
+To facilitate the usual HyperSpy workflows as well as possible,
```
tjof2: "In line with the standard HyperSpy workflows, :py:meth:`~._signals.lazy.LazySignal.decomposition` offers the following implementations:"

doc/user_guide/big_data.rst
```diff
@@ -0,0 +1,228 @@
...
+Constraints
+-----------
+
+A couple of constraints to ``LazySignal`` when compared to normal HyperSpy
+signals.
+
+Immutable signals
+^^^^^^^^^^^^^^^^^
+
+One of the most often encountered constraint when using ``LazySignal`` is the
```
tjof2: "An important constraint when using..."

doc/user_guide/big_data.rst
```diff
@@ -0,0 +1,228 @@
...
+inability to modify an existing signal data (immutability). It is a logical
```
tjof2: "modify existing data" surely?

tjof2: Also, "This is a logical consequence..."

doc/user_guide/big_data.rst
```diff
@@ -0,0 +1,228 @@
...
+    >>> print(s.data.dtype, s.data.nbytes / 1e9)
+    uint8 34.9175808
+    >>> s.change_dtype("float") # To be able to perform decomposition, etc.
+    >>> print(s.data.dtype, s.data.nbytes / 1e9)
+    float64 279.3406464
+
+Only to load the data in the original unsigned integer format it would require
+around 35GB of memory. To store it in a floating-point format one would need
+almost 280GB of memory. Nevertheless, with the lazy processing both of these
+steps are near instantaneous and require very little resources.
```
tjof2: "Loading the dataset in the original unsigned integer format would require around 35GB of memory. To store it in a floating-point format one would need almost 280GB of memory. However, with lazy processing both of these steps are near-instantaneous and require very little computational resources."

doc/user_guide/big_data.rst
```diff
@@ -0,0 +1,228 @@
...
+further operations branch from it. Only when a certain branch result is
+requested, the way to the root is found and evaluated in the correct sequence
+on the correct blocks.
+
+The "magic" is performed by (for the sake of simplicity) storing the data not
+as ``numpy.ndarray``, but ``dask.array.Array`` (more information `here
+<https://dask.readthedocs.io/en/latest/>`_). ``dask`` offers a couple of
+advantages:
+
+* **Teoretically, arbitrary-sized data processing is possible**. By only
```
tjof2: **Arbitrary-sized data processing is possible** is fine - don't need "theoretically"

doc/user_guide/big_data.rst
```diff
@@ -0,0 +1,228 @@
...
+  loading a couple of chunks at a time, theoretically any signal can be
+  processed, albeit slower. In practice, this may be limited: (i) some
+  operations may require certain chunking pattern, which may still saturate
+  memory; (ii) a couple of tens of chunks should fit into the computer memory
+  comfortably.
+* **Loading only the required data**. If a certain part (chunk) of the data is
+  not required for the final result, it will not be loaded at all, saving time
+  and resources.
+* **Able to extend to clusters**. ``dask.distributed`` (documentation `here
```
tjof2: Be clearer - "compute clusters"  or "a distributed computing environment" - clusters could mean k-means clustering...

doc/user_guide/big_data.rst
```diff
@@ -0,0 +1,228 @@
...
+The "magic" is performed by (for the sake of simplicity) storing the data not
+as ``numpy.ndarray``, but ``dask.array.Array`` (more information `here
+<https://dask.readthedocs.io/en/latest/>`_). ``dask`` offers a couple of
+advantages:
+
+* **Teoretically, arbitrary-sized data processing is possible**. By only
+  loading a couple of chunks at a time, theoretically any signal can be
+  processed, albeit slower. In practice, this may be limited: (i) some
+  operations may require certain chunking pattern, which may still saturate
+  memory; (ii) a couple of tens of chunks should fit into the computer memory
```
tjof2: "a couple of tens" makes no sense - it depends on the size of each chunk... perhaps just "several chunks should fit..."

tjof2:
@to266 Apart from some mention of ONMF in the Machine Learning documentation (it needn't be much at all - see the ORPCA example directly above it for inspiration), this PR gets a 👍 from me now! 

Code for ONMF in particular looks good.

.travis.yml
```diff
@@ -30,7 +30,7 @@ before_install:
 install:
 
   - if [[ $MINIMAL_ENV == 'False' ]] ; then
-      DEPS="numpy scipy matplotlib ipython h5py sympy scikit-learn dill natsort setuptools scikit-image cython lxml ipyparallel";
+      DEPS="pytest pytest-cov numpy scipy matplotlib ipython h5py sympy scikit-learn dill natsort setuptools scikit-image cython lxml ipyparallel dask toolz";
```
francisco-dlp: is toolz still required?

to266: that's right, no longer required. Though `dask` uses it extensively, so will be pulled in any way. Will get rid of it on our side, though

hyperspy/io.py
```diff
@@ -103,45 +102,52 @@ def load(filenames=None,
...
+    >>> s = hs.load('file*.blo', lazy=True, stack=True)
+
     """
+    deprecated = ['load_to_memory', 'mmap', 'mmap_dir', 'mmap_mode']
+    warn_str = "'{}' argument is deprecated, please use 'lazy' instead"
+    for k in deprecated:
+        if k in kwds:
+            lazy=True
+            warnings.warn(warn_str.format(k), VisibleDeprecationWarning)
+            del kwds[k]
```
francisco-dlp, line 144: We cannot simply ignore the keywords as this could break somebody's API. ``load_to_memory`` and ``mmap`` could just set the value of ``lazy``. ``mmap_dir`` could be ignored as this was a poor man solution to the problem and makes no sense now so nobody will miss it. ``mmap_mode`` is trickier: this allows to modify the original file when ``mmap_mode`` is ``r+``, is this available for lazily loaded signals?

doc/user_guide/big_data.rst
```diff
@@ -0,0 +1,227 @@
...
+Lazy stacking
+^^^^^^^^^^^^^
+
+Occasionally the full dataset consists of many smaller files. To combine them
+into a one large ``LazySignal``, we can :ref:`stack<signal.stack_split>` them
+lazily (both when loading of afterwards):
+
+.. code-block:: python
+
+    >>> siglist = hs.load("*.hdf5")
```
francisco-dlp: Would the following work and possibly be better?

```python

siglist = hs.load("*.hdf5", lazy=True)
s = hs.stack(siglist)

```

hyperspy/io.py
to266: `dask` itself happily writes into anything that supports numpy-style slicing, however since all dask arrays (and hence lazy signals) are immutable, the writing has to be called explicitly.

Currently there is no in-built way to do this, but one could do `s.data.store(mmap)`, which would compute the data and write to `mmap`.

I'll make the changes to allow these modes.

Another cat in the bag is the hdf5 `load_to_memory=True` if there are arrays in the metadata (e.g. a model is stored). Then `deepcopy()` fails, because an open `h5py` handle cannot be copied. That may not be a problem if I wrapped those to dask arrays as well (I have not tested this yet), but it touches things outside the main scope of the PR

doc/user_guide/big_data.rst
to266: Yes. Maybe I failed to illustrate what I wanted though - the goal was to show that a lazy signal can be constructed by lazily stacking many normal signals (e.g. tiff images, that may have been processed separately and already open, or something).

Admittedly, the only example I can think of where you may want lazy stack but normal signals is if you only have enough memory for one full copy (e.g. it takes up 60% of your memory), and want to be able to compare the original files with the processed stack. Though I would load the individual files lazily any way... Maybe if it's not straightforward (something similar to bcf)?

.travis.yml
sem-geologist: ??? now it fails minimal Travis test. Why you remove it? Dask is stupid it will complain, it does not install it automatically when pip installing, and some parts of dask used needs this toolz, you dont get this on Anaconda environment as it is automagicaly silently is or even pre-installed (anaconda people knows that it is required )). Now if somebody will try to install this through the pip directly will run into same error Travis_minimal environment fails.

So to be clear, either toolz have to go back to dependency or dask have to be optional.

francisco-dlp:
Indeed, toolz is an extra dependency for dask which is required in our
case. So we need to add it to our setup.py etc (maybe with a comment
explaining why it's there?).

francisco-dlp:
Although maybe a better approach would be to add the dependency in setup.py
likes this ``dask[array]``. In this way the optional toolz dependency
should be automatically installed and it makes it clearer that it is not a
hspy requirement.

to266:
ooo, I like the second approach

to266:
Anything else left?

francisco-dlp:
Is ``load_to_memory`` no longer deprecated?

to266:
It's not. I changed the user guide to reflect this

EDIT: at least it shows to me that I pushed, don't see anything here...

francisco-dlp:
The commit just appeared. So, why is it not deprecated anymore?

to266:
because in one case (don't remember which) this works as `mmap`, takes `mmap_mode`, etc. To keep the ability to modify the file in-place consistently, un-deprecated it

francisco-dlp:
I see. However, I think that the situations is a bit confusing at present as we have three keywords for seemly the same thing (albeit with subtle differences) (``lazy``, ``mmap`` and ``load_to_memory``).

to266:
Oh I agree, that is why I initially deprecated all of them and left only `lazy`. Then you argued that this may break some APIs, where the data is loaded in `r+` mode and modified on-disk. Since both `mmap` and `load_to_memory` can be used in this way, I put them back in.

francisco-dlp:
Before you've suggested that there is a way to get the "r+" feature using dask. Is it finally not possible?

If that was the case, then I think that there should be an explanation in the User Guide of the difference between ``load_to_memory`` and ``lazy``.

to266:
well the `r+` can be achieved by simply overwriting the same file, as there is no way to update only parts of the file.

francisco-dlp:
After going through all the current io readers I've found out that ``load_to_memory`` is only available for hdf5, emd, and blockfile. However, only blockfile has a working ``mmap_mode``. emd doesn't have the option and hdf5 has it but it's broken because hdf5 uses dask when ``load_to_memory = False``. This means that ``load_to_memory`` only offers an advantage compared to ``lazy`` for blockfiles. But then, what's the use of editing blockfiles in-place? This is a propietary format intended to save experimental data, so in-place modification does not seem very useful for this format. Hence, I think that it is fine to deprecate ``load_to_memory`` after all. What do you think?

francisco-dlp:
Another option would be to deprecate it and if ``mmap_mode == "r+"`` read it as mmap if lazy is False and raise an error if lazy is True explaining that "r+" does not support lazy operations.

francisco-dlp:
Wouldn't it be good to add the ability to create a lazy signal automatically when the signal is created with a memmap or hdf5 in memory map mode? i.e. ``hs.signals.Signal1D(a_memmap)`` returns a ``LazySignal1D``. This would make it easier to use with non supported formats.

doc/user_guide/io.rst
```diff
@@ -282,6 +287,11 @@ partly support FEI's custom header. We do not provide writing features for this
 format, but, as it is an an open format, we may implement this feature in the
 future on demand.
 
+For mrc files ``load`` takes the ``mmap_mode`` keyword argument enabling
+loading the file using a different mode (default is copy-on-write) . However,
+note that lazy loading does not support in-place writring (i.e lazy loading and
```
tjof2: spelling

doc/user_guide/io.rst
```diff
@@ -336,6 +346,12 @@ using the encoding argument, e.g.:
 
     >>> s.save('file.rpl', encoding = 'utf8')
 
+
+For mrc files ``load`` takes the ``mmap_mode`` keyword argument enabling
+loading the file using a different mode (default is copy-on-write) . However,
+note that lazy loading does not support in-place writring (i.e lazy loading and
```
tjof2: spelling again

to266: I think it was just copy-pasted, so many instances of this will appear :)

to266:
@francisco-dlp for me that does not sound like a good idea, returning a different class than was requested.
Currently it's just `hs.signals.Signal1D(mmap).as_lazy()`, which is both convenient to write and verbose, I think.

Another, separate, thing is that `LazySignal` does not attempt to make its data lazy if not told so explicitly (ie. `_make_lazy` is not called when creating the signal), as it assumes the given data is appropriate. Now I don't remember why exactly I chose do to it this way, possibly simply due to not having written the "lazifying" parts of the code at that particular point. Now I'm interested what you think about this choice.

francisco-dlp:
I am not sure if I follow you, does ``hs.signals.Signal1D(mmap).as_lazy()`` offer lazy evaluation without having to wrap the memmap in an dask object? If not, then I think that it should.

If  I understood it correctly, ``hs.signals.Signal1D(array).as_lazy()`` keeps the array as-is but performs lazy evaluation of the operations (does it?). If that's the case it seems fine to me as, what else makes sense for an in-memory array?

to266:
That's not exactly correct.

`as_lazy` calls `_assign_subclass`, which in turn calls `_make_lazy`, which makes sure the data of the result is always a dask array. If it was already a dask array, then it is just left as-is. If it's anything else, the function "wraps" it in dask. 

It's the `dask.array.Array` that supports lazy evaluation, so if the data is anything else, lazy processing is not possible.

The thing is that just creating lazy signal like so `hs.signals.LazySignal(somedata)` does not call `_make_lazy`, and hence you have whatever you put in there. Most functions would end up working lazily because they operate not on the data itself, but on the result of `s._lazy_data(axis)`, which is the function that does the wrapping that I mentioned.

francisco-dlp:
I see. Wouldn't a Lazy signal be somehow broken then if the data does not pass by ``_make_lazy``? If yes, then why not doing it?

By the way, why would we want to expose the `Lazy*` signals in the API? Calling as lazy on the standard one seems equivalent and is actually what's documented in the UG. From the ZoP:

It would be good to add a little paragraph to the UG explaining how to get a lazy signal from a memmap or hdf5 memmap i.e. ``hs.signals.Signal1D(mmap).as_lazy()``.

francisco-dlp:
Also some words in the User Guide about what happens when mixing lazy and std signals (e.g. ``lazy + std``) would be useful.

to266:
Forgot that holography types were added, so now I have to make sure they work (and currently they don't for some reason). WIP

thomasaarholt:
**roaring applause**

sem-geologist:
yes, yes, yes :)

congratulations @to266 !

magnunor:
Very nice! :)

vidartf:
👏 🎉

to266:
🥇  :DD

PR #1220, opened 2016-07-29
signal transpose

to266:
```python
>>> # just create a signal with many distinct dimensions
>>> s = hs.signals.BaseSignal(np.random.rand(1,2,3,4,5,6,7,8,9))
>>> s
<BaseSignal, title: , dimensions: (|9, 8, 7, 6, 5, 4, 3, 2, 1)>
                                                                           
>>> s.transpose() # swap signal and navigation spaces
<BaseSignal, title: , dimensions: (9, 8, 7, 6, 5, 4, 3, 2, 1|)>
                                                                           
>>> s.T # a shortcut for no arguments
<BaseSignal, title: , dimensions: (9, 8, 7, 6, 5, 4, 3, 2, 1|)>
                                                                           
>>> s.transpose(signal_axes=5) # roll to leave 5 axes in navigation space
<BaseSignal, title: , dimensions: (4, 3, 2, 1|9, 8, 7, 6, 5)>
                                                                           
>>> s.transpose(navigation_axes=3) # roll leave 3 axes in navigation space
<BaseSignal, title: , dimensions: (3, 2, 1|9, 8, 7, 6, 5, 4)>
                                                                           
>>> # 3 explicitly defined axes in signal space
>>> s.transpose(signal_axes=[0, 2, 6])
<BaseSignal, title: , dimensions: (8, 6, 5, 4, 2, 1|9, 7, 3)>
                                                                           
>>> # A mix of two lists, but specifying all axes explicitly
>>> # The order of axes is preserved in both lists
>>> s.transpose(navigation_axes=[1, 2, 3, 4, 5, 8], signal_axes=[0, 6, 7])
<BaseSignal, title: , dimensions: (8, 7, 6, 5, 4, 1|9, 3, 2)>


```

- [x] Tests
- [x] Better error messages when bad combination is given
- [x] Context manager
- [x] `.T` shortcut
- [x] `hs.transpose`

The nice thing is that this works with any-dimension data (as long as there are enough dimensions in the signal):

```python
with s.transposed(signal_axes=2) as im:
    hs.plot.plot_images(im)

# or 
s2 = hs.signals.BaseSignal(np.random.rand(2, 2))
s3 = hs.signals.BaseSignal(np.random.rand(3, 3, 3))
s4 = hs.signals.BaseSignal(np.random.rand(4, 4, 4, 4))
with hs.transposed(s2, s3, s4, signal_axes=2) as signals:
    hs.plot.plot_images(signals)

```

### DIFF
--- a/doc/user_guide/tools.rst
+++ b/doc/user_guide/tools.rst
@@ -123,7 +123,7 @@ array of scalars by changing the default *view* of
 .. code-block:: python
 
     >>> scalar = hs.signals.BaseSignal(np.random.random((10, 20, 30)))
-    >>> scalar.axes_manager.set_signal_dimension(0)
+    >>> scalar = scalar.transpose(signal_axes=0)
     >>> scalar
     <BaseSignal, title: , dimensions: (30, 20, 10|)>
 
@@ -771,13 +771,11 @@ arguments as in the following example.
     >>> image_stack.axes_manager[1].name = "x"
     >>> image_stack.axes_manager[2].name = "y"
     >>> angles = hs.signals.BaseSignal(np.array([0, 45, 90, 135]))
-    >>> angles.axes_manager.set_signal_dimension(0)
     >>> modes = hs.signals.BaseSignal(np.array(['constant', 'nearest', 'reflect', 'wrap']))
-    >>> modes.axes_manager.set_signal_dimension(0)
     >>> image_stack.map(scipy.ndimage.rotate,
-    ...                            angle=angles,
+    ...                            angle=angles.transpose(signal_axes=0),
     ...                            reshape=False,
-    ...                            mode=modes)
+    ...                            mode=modes.transpose(signal_axes=0))
     calculating 100% |#############################################| ETA:  00:00:00Cropping
 
 .. figure::  images/rotate_lena_apply_ndkwargs.png
--- a/doc/user_guide/tools.rst
+++ b/doc/user_guide/tools.rst
@@ -930,24 +930,32 @@ Transposing (changing signal spaces)
 .. versionadded:: 1.1
 
 :py:meth:`~.signal.BaseSignal.transpose` method changes how the dataset
-dimensions are interpeted (as signal or navigation axes). The method accepts
-both explicit axes to keep in either space, or just a number of axes required
-in one space (just one number can be specified, as the other is defined as
-"all other axes"). When axes order is not explicitly defined, they are "rolled"
-from one space to the other as if the ``<navigation axes | signal axes >``
-wrap a circle.
+dimensions are interpreted (as signal or navigation axes). By default is
+swaps the signal and navigation axes. For example:
 
-The :py:meth:`~.signal.BaseSignal.transpose` method accepts keyword argument
-``optimize``, which is ``False`` by default, meaning modifying the output
-signal data **modified the original data**. If ``True``, the method ensures the
-data in memory is stored in the most efficient manner for iterating by making a
-copy of the data if required, hence **not always modifying** the original data.
 
-The older convenience methods :py:meth:`~.signal.BaseSignal.as_signal1D` and
-:py:meth:`~.signal.BaseSignal.as_signal2D` internally use
-:py:meth:`~.signal.BaseSignal.transpose`, but always optimize.
+.. code-block:: python
 
-Examples:
+   >>> s = hs.signals.Signal1D(np.zeros((4,5,6)))
+   >>> s
+   <Signal1D, title: , dimensions: (5, 4|6)>
+   >>> s.transpose()
+   <Signal2D, title: , dimensions: (6|4, 5)>
+
+For :py:meth:`~.signal.BaseSignal.T` is a shortcut for the default behaviour:
+
+.. code-block:: python
+
+   >>> s = hs.signals.Signal1D(np.zeros((4,5,6))).T
+   <Signal2D, title: , dimensions: (6|4, 5)>
+
+
+The method accepts both explicit axes to keep in either space, or just a number
+of axes required in one space (just one number can be specified, as the other
+is defined as "all other axes"). When axes order is not explicitly defined,
+they are "rolled" from one space to the other as if the ``<navigation axes |
+signal axes >`` wrap a circle. The example below should help clarifying this.
+
 
 .. code-block:: python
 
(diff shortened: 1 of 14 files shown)


hyperspy/signal.py
```diff
@@ -3867,6 +3867,157 @@ def add_gaussian_noise(self, std):
...
+        for i, ax in enumerate(ram._axes):
+            if i < len(navigation_axes):
+                ax.slice = None
+                ax.navigate = True
+            else:
+                ax.slice = slice(None)
+                ax.navigate = False
+        ram._update_attributes()
+        ram._update_trait_handlers(remove=False)
+        res.get_dimensions_from_data()
```
francisco-dlp: Why do you need this? The dimensions shouldn't change.

to266: I definitely do (both reorder and need).

Though hopefully I'm doing the axes reordering properly...

to266: Actually you're right, I made a mistake. will push fix soon

to266:
I think this is ready for the final review...

hyperspy/signal.py
```diff
@@ -3867,6 +3867,157 @@ def add_gaussian_noise(self, std):
...
+                raise ValueError("The navigation_axes are not None, even "
+                                 "though just a number was given for "
+                                 "signal_axes")
+            if len(nat_axes) < signal_axes:
+                raise ValueError("Too many signal axes requested")
+            navigation_axes = nat_axes[:-signal_axes]
+            signal_axes = nat_axes[-signal_axes:]
+        elif iterable_not_string(signal_axes):
+            if navigation_axes is None:
+                signal_axes = tuple(am[ax] for ax in signal_axes)
```
francisco-dlp: It would be cleared to move this line above `if navigation_axes is None:`  and remove line 3940.

hyperspy/signal.py
```diff
@@ -3867,6 +3867,157 @@ def add_gaussian_noise(self, std):
...
+                navigation_axes = tuple(ax for ax in nat_axes if ax not in
+                                        signal_axes)
+            elif iterable_not_string(navigation_axes):
+                # want to keep the order
+                signal_axes = tuple(am[ax] for ax in signal_axes)
+                navigation_axes = tuple(am[ax] for ax in navigation_axes)
+                if len(set(signal_axes).intersection(navigation_axes)):
+                    raise ValueError("Same axis in both spaces found.")
```
francisco-dlp: It would be more useful to print the axis/axes to blame in the error message.

hyperspy/signal.py
```diff
@@ -3867,6 +3867,157 @@ def add_gaussian_noise(self, std):
...
+                ax.navigate = False
+        ram._update_attributes()
+        ram._update_trait_handlers(remove=False)
+        # need a copy?
+        if copy:
+            res._make_sure_data_is_contiguous()
+        return res
+
+    @contextmanager
+    def transposed(self, signal_axes=None, navigation_axes=None, copy=False):
```
francisco-dlp: The context manager is not very useful because transpose returns the signal in a "different view" rather than modifying it. Therefore the context manager has nothing to do when exiting. (My bad, as I suggested the CM to you.)

hyperspy/signal.py
```diff
@@ -3867,6 +3867,157 @@ def add_gaussian_noise(self, std):
...
+                ax.slice = None
+                ax.navigate = True
+            else:
+                ax.slice = slice(None)
+                ax.navigate = False
+        ram._update_attributes()
+        ram._update_trait_handlers(remove=False)
+        # need a copy?
+        if copy:
+            res._make_sure_data_is_contiguous()
```
francisco-dlp: Shouldn't this transpose the variance too if it exists?

to266:
thanks for review, all done @francisco-dlp

hyperspy/misc/utils.py
```diff
@@ -895,3 +895,51 @@ def shorten_name(name, req_l):
...
+    >>> hs.transpose(signal1, signal2, signal3, signal_axes=["Energy"])
+    """
+    from hyperspy.signal import BaseSignal
+    if not all(map(isinstance, args, (BaseSignal for _ in args))):
+        raise ValueError("Not all pased objects are signals")
+    return [sig.transpose(signal_axes=signal_axes,
+                          navigation_axes=navigation_axes,
+                          copy=copy) for sig in args]
+@contextmanager
+def transposed(*args, signal_axes=None, navigation_axes=None, copy=False):
```
francisco-dlp: Is this context manager necessary (it does nothing on exit)?

francisco-dlp:
As this generalises `as_signal2D`, `as_signal1D`, `to_signal1D`, `to_signal2D`, shouldn't all these functions be rewritten using `transpose`? This would have the extra benefit of further testing transpose, as a lot of the current tests rely on the methods above.

francisco-dlp:
Also
* Use Guide documentation is missing
* It would be good to replace most calls to `axes_manager.set_signal_dimension` with calls to `transpose` in the user guide, docstrings and code.

to266:
Done, I think. Let me know if anything is still missing, thanks

doc/user_guide/tools.rst
```diff
@@ -123,12 +123,12 @@ array of scalars by changing the default *view* of
 .. code-block:: python
 
     >>> scalar = hs.signals.BaseSignal(np.random.random((10, 20, 30)))
```
francisco-dlp: Maybe just `scalar = hs.signals.BaseSignal(np.random.random((10, 20, 30))).T` would be better with a little explanation below or above?

doc/user_guide/tools.rst
```diff
@@ -771,13 +771,11 @@ arguments as in the following example.
     >>> image_stack.axes_manager[1].name = "x"
     >>> image_stack.axes_manager[2].name = "y"
     >>> angles = hs.signals.BaseSignal(np.array([0, 45, 90, 135]))
-    >>> angles.axes_manager.set_signal_dimension(0)
     >>> modes = hs.signals.BaseSignal(np.array(['constant', 'nearest', 'reflect', 'wrap']))
-    >>> modes.axes_manager.set_signal_dimension(0)
     >>> image_stack.map(scipy.ndimage.rotate,
-    ...                            angle=angles,
+    ...                            angle=angles.transpose(signal_axes=0),
```
francisco-dlp: angles.T?

doc/user_guide/tools.rst
```diff
@@ -771,13 +771,11 @@ arguments as in the following example.
...
     ...                            reshape=False,
-    ...                            mode=modes)
+    ...                            mode=modes.transpose(signal_axes=0))
```
francisco-dlp: modes.T?

doc/user_guide/tools.rst
```diff
@@ -926,6 +924,55 @@ In the following example we create a 1D signal with signal size 3 and with
...
+    >>> # 3 explicitly defined axes in signal space
+    >>> s.transpose(signal_axes=[0, 2, 6])
+    <BaseSignal, title: , dimensions: (8, 6, 5, 4, 2, 1|3, 7, 9)>
+
+    >>> # A mix of two lists, but specifying all axes explicitly
+    >>> # The order of axes is preserved in both lists
+    >>> s.transpose(navigation_axes=[1, 2, 3, 4, 5, 8], signal_axes=[0, 6, 7])
+    <BaseSignal, title: , dimensions: (8, 7, 6, 5, 4, 1|9, 3, 2)>
+
+.. NOTE::
```
francisco-dlp: I would not put this paragraph inside a note.

doc/user_guide/tools.rst
```diff
@@ -926,6 +924,55 @@ In the following example we create a 1D signal with signal size 3 and with
...
+
+    The :py:meth:`~.signal.BaseSignal.transpose` method accepts keyword
+    argument ``copy``, which is ``False`` by default. If ``True``, it ensures
+    the data in memory is stored in the most efficient manner for iterating.
+    Often this means making a copy of the data.
```
francisco-dlp: What about replacing the text above with:

```rest

The :py:meth:`~.signal.BaseSignal.transpose` method accepts the keyword argument ``copy``, which is ``False`` by default. Therefore, by default :py:meth:`~.signal.BaseSignal.transpose` returns a new  :py:class:`~.signal.BaseSignal` instance that contains a view of the data. Hence, modifying the output signal data **modifies** the original data. If ``True``, the data is copied instead and modifying the output signal data **does not modify** the original data. Copying also ensures the data in memory is stored in the most efficient manner for iterating over the navigation axes. However, copying the data requires duplicating the data in memory.

```

doc/user_guide/visualisation.rst
```diff
@@ -381,8 +381,7 @@ different slices of a multidimensional image (a *hyperimage*):
     >>> import scipy
     >>> image = hs.signals.Signal2D([scipy.misc.lena()]*6)
     >>> angles = hs.signals.BaseSignal(range(10,70,10))
-    >>> angles.axes_manager.set_signal_dimension(0)
-    >>> image.map(scipy.ndimage.rotate, angle=angles, reshape=False)
+    >>> image.map(scipy.ndimage.rotate, angle=angles.transpose(signal_axes=0), reshape=False)
```
francisco-dlp: .T?

doc/user_guide/visualisation.rst
```diff
@@ -404,8 +403,7 @@ In this example, the axes labels and the ticks are also disabled with `axes_deco
...
```
francisco-dlp: .T?

doc/user_guide/visualisation.rst
```diff
@@ -433,8 +431,7 @@ This example also demonstrates how to wrap labels using `labelwrap` (for prevent
 
     >>> # load lena into 6 hyperimage
     >>> image1 = hs.signals.Signal2D([scipy.misc.lena()]*6)
-    >>> angles = hs.signals.BaseSignal(range(10,70,10))
-    >>> angles.axes_manager.set_signal_dimension(0)
+    >>> angles = hs.signals.BaseSignal(np.arange(10,70,10)).transpose(signal_axes=0)
```
francisco-dlp: .T?

hyperspy/_signals/common_signal1d.py
```diff
@@ -41,7 +41,7 @@ def to_signal2D(self):
         if self.data.ndim < 2:
```
francisco-dlp: See Also
-------------

transpose, as_signal1D, as_signal2D

francisco-dlp: I would also add a sentence explaining that this method copies the data unlike transpose.

hyperspy/signal.py
```diff
@@ -3224,7 +3223,7 @@ def map(self, function,
 
         >>> im = hs.signals.Signal2D(np.random.random((10, 64, 64)))
         >>> sigmas = hs.signals.BaseSignal(np.linspace(2,5,10))
```
francisco-dlp: .T?

doc/user_guide/tools.rst
to266: ...This is not actually true - it will only make a copy if it's not a contiguous array. Should I make it actually copy in all cases?

hyperspy/signal.py
```diff
@@ -3649,10 +3648,11 @@ def as_signal1D(self, spectral_axis, out=None):
...
 
         """
-        # Roll the spectral axis to-be to the latex index in the array
-        sp = self.rollaxis(spectral_axis, -1 + 3j)
-        sp.axes_manager.set_signal_dimension(1)
-        sp._assign_subclass()
+        ax = self.axes_manager[spectral_axis]
+        all_ax = self.axes_manager._axes.copy()
+        sig = [all_ax.pop(ax.index_in_array)]
+        sp = self.transpose(signal_axes=sig,
```
francisco-dlp: Wouldn't `self.transpose(signal_axes=self.navigation_axes[-1], copy=True)` be equivalent?

doc/user_guide/tools.rst
francisco-dlp: I would also add documentation for `hs.transpose`, maybe with the nice example that you have in the PR using it in combination with `plot_images`?

francisco-dlp: I've just realised that `as_signal1D` and `as_signal2D` are not documented in the User Guide. Could you add a sentence here explaining that they're convenience functions that optimise the data by default?

PR #1221, opened 2016-08-01
NEW: ElectronDiffraction class

dnjohnstone:
This PR adds a class for scanning electron diffraction data.

A couple of simple methods are included as utility functions to achieve masking and pattern alignment.

Hopefully pretty close to ready except a few documentation updates.


doc/user_guide/sed.rst
```diff
@@ -0,0 +1,224 @@
...
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+Microscope and detector parameters are stored in the
+:py:attr:`~.signal.Signal.metadata` attribute (see :ref:`metadata_structure`).
+These parameters can be displayed as follows:
+
+.. code-block:: python
+
+    >>> dp = hs.load("GaAs_nanowire_002.hdf5", signal_type="SED_Pattern")
+    >>> dp.metadata
```
francisco-dlp: Where's the output?

doc/user_guide/sed.rst
```diff
@@ -0,0 +1,224 @@
...
+    >>> dp = hs.load("GaAs_nanowire_002.hdf5", signal_type="SED_Pattern")
+    >>> dp.metadata.Acquisition_instrument.SED.convergence_angle = 5.
+
+or with the
+:py:meth:`~._signals.sed.SEDPattern.set_microscope_parameters` method:
+
+.. code-block:: python
+
+    >>> dp = hs.load("GaAs_nanowire_002.rpl", signal_type="SED_Pattern")
+    >>> dp.set_microscope_parameters(convergence_angle = 5.)
```
francisco-dlp: no spaces is " = "

doc/user_guide/sed.rst
```diff
@@ -0,0 +1,224 @@
...
+the direct beam is estimated using the approach described by Zaeferrer
+[Ref, Zaeferrer 2000] and White [Ref, White Thesis] as follows:
+
+.. code-block:: python
+
+    >>>
+    >>> dp.get_direct_beam_position()
+
+
+In brief, the py:meth:`~._signals.sed.estimate_direct_beam_position()` method first sums all diffraction patterns in the stack
```
francisco-dlp: Something is missing here

dnjohnstone: as in the comment --- the documentation needs updating!

doc/user_guide/sed.rst
```diff
@@ -0,0 +1,224 @@
...
+'Virtual' diffraction imaging involves plotting the intensity of a sub-set of
+pixels in each electron diffraction pattern comprising a SED dataset, as a
+function of probe position. In this way, variations in the diffraction condition
+are mapped. Forming such 'virtual' diffraction images in HyperSpy is easy using
+the 'interactive' and 'ROI' functionality of the signal class as follows:
+
+.. code-block:: python
+
+    >>> dp = hs.load("GaAs_nanowire_002.rpl", signal_type="SED_Pattern")
+    >>>
```
francisco-dlp: Something is missing here

doc/user_guide/sed.rst
```diff
@@ -0,0 +1,224 @@
...
+   :align: center
+   :width: 400
+
+   Automatically generated mask excluding SED patterns acquired in vacuum.
+
+
+Machine learning SED data
+-------------------------
+
+Machine learning decomposition approaches can be applied to SED data
```
francisco-dlp: And here

hyperspy/_signals/sed.py
```diff
@@ -0,0 +1,465 @@
...
+
+from hyperspy._signals.signal2d import Signal2D
+from hyperspy.decorators import only_interactive
+from hyperspy.gui.sed import SEDParametersUI
+from hyperspy.defaults_parser import preferences
+import hyperspy.gui.messages as messagesui
+
+
+class SEDPattern(Signal2D):
+    _signal_type = "SED_Pattern"
```
francisco-dlp: Is this more general, maybe Electron diffraction pattern? Would it make sense to strip "pattern" and be more verbose e.g. `electron_diffraction`? In any case, to follow the conventions (see #1160) anything that is not an acronym should be lowercase.

hyperspy/_signals/sed.py
```diff
@@ -0,0 +1,465 @@
...
+                        preferences.SED.sed_convergence_angle)
+        if "Acquisition_instrument.TEM.precession_angle" not in md:
+            md.set_item("Acquisition_instrument.TEM.precession_angle",
+                        preferences.SED.sed_precession_angle)
+        if "Acquisition_instrument.TEM.precession_frequency" not in md:
+            md.set_item("Acquisition_instrument.TEM.precession_frequency",
+                        preferences.SED.sed_precession_frequency)
+        if "Acquisition_instrument.TEM.Detector.SED.exposure_time" not in md:
+            md.set_item("Acquisition_instrument.TEM.Detector.SED.exposure_time",
+                        preferences.SED.sed_exposure_time)
```
francisco-dlp: Most of these entries are not in the [metadata specification](http://hyperspy.org/hyperspy-doc/current/user_guide/metadata_structure.html). The specifications should be expanded to include the ones strictly needed. Why is there a `scan_rotation` and `sed_scan_rotation`?

Wouldn't `Acquisition_instrument.TEM.Detector.diffraction.exposure` be better than `Acquisition_instrument.TEM.Detector.SED.exposure_time`? All quantities should include units in brackets e.g. "exposure (s)".

hyperspy/_signals/sed.py
```diff
@@ -0,0 +1,465 @@
...
+        # sum images to produce image in which direct beam reinforced and take
+        # the position of maximum intensity as the initial estimate of center.
+        dp_sum = self.sum()
+        maxes = np.asarray(np.where(dp_sum.data == dp_sum.data.max()))
+        mref = [np.average(maxes[0]), np.average(maxes[1])]
+        # specify array of dims (nav_size, 2) in which to store centers and find
+        # the center of each pattern by determining the direct beam position.
+        arr_shape = (self.axes_manager.navigation_size, 2)
+        c = np.zeros(arr_shape, dtype=int)
+        for z, index in zip(self._iterate_signal(),
```
francisco-dlp: Wouldn't it be possibly to rewrite this as an external function and call `map` to make it nd? This is specially important now that #1219 is coming.

hyperspy/_signals/sed.py
```diff
@@ -0,0 +1,465 @@
...
+
+        Notes
+        -----
+        Based on work presented by Thomas White in his PhD, Cambridge, (2009)
+        which itself built on Zaefferer (2000).
+        """
+        # sum images to produce image in which direct beam reinforced and take
+        # the position of maximum intensity as the initial estimate of center.
+        dp_sum = self.sum()
+        maxes = np.asarray(np.where(dp_sum.data == dp_sum.data.max()))
```
francisco-dlp: Wouldn't this be equivalent to `maxes = dp_sum.data == dp_sum.data.max()`?

hyperspy/_signals/sed.py
```diff
@@ -0,0 +1,465 @@
...
+        mref = [np.average(maxes[0]), np.average(maxes[1])]
```
francisco-dlp: Could you add a comment explaining what's mref purpose?

hyperspy/_signals/sed.py
francisco-dlp: I see that it is indeed not possible.

hyperspy/_signals/sed.py
```diff
@@ -0,0 +1,465 @@
...
+                if is_ok:
+                    self._set_microscope_parameters()
+                else:
+                    return True
+            else:
+                return True
+        else:
+            return False
+
+    def get_direct_beam_mask(self, radius=None, center=None):
```
francisco-dlp: radius is required, hence it shouldn't default to None.

hyperspy/_signals/sed.py
```diff
@@ -0,0 +1,465 @@
...
+            a = (self.axes_manager.signal_shape[1] - 1) / 2
+            b = (self.axes_manager.signal_shape[0] - 1) / 2
+        else:
+            a, b = center
+
+        y, x = np.ogrid[-b:ny-b, -a:nx-a]
+        mask = x*x + y*y <= r*r
+        return mask
+
+    def get_direct_beam_position(self, radius=None):
```
francisco-dlp: radius is required, hence it shouldn't default to None.

hyperspy/_signals/sed.py
```diff
@@ -0,0 +1,465 @@
...
+        c = np.zeros(arr_shape, dtype=int)
+        for z, index in zip(self._iterate_signal(),
+                            np.arange(0, self.axes_manager.navigation_size, 1)):
+            # initialise problem with initial center estimate
+            c_int = z[mref[0], mref[1]]
+            mask = self.get_direct_beam_mask(radius=radius, center=mref)
+            z_tmp = z * mask
+            # refine center position with shifting ROI
+            if c_int == z_tmp.max():
+                maxes = np.asarray(np.where(z_tmp == z_tmp.max()))
```
francisco-dlp: As above.

hyperspy/_signals/sed.py
```diff
@@ -0,0 +1,465 @@
...
+
+        if center == None:
+            a = (self.axes_manager.signal_shape[1] - 1) / 2
+            b = (self.axes_manager.signal_shape[0] - 1) / 2
+        else:
+            a, b = center
+
+        y, x = np.ogrid[-b:ny-b, -a:nx-a]
+        mask = x*x + y*y <= r*r
+        return mask
```
francisco-dlp: Shouldn't this return a signal? If you're worried about overhead, it could take an `out` argument.

hyperspy/_signals/sed.py
```diff
@@ -0,0 +1,465 @@
...
+        arr_shape = (self.axes_manager.navigation_size, 2)
+        c = np.zeros(arr_shape, dtype=int)
+        for z, index in zip(self._iterate_signal(),
+                            np.arange(0, self.axes_manager.navigation_size, 1)):
+            # initialise problem with initial center estimate
+            c_int = z[mref[0], mref[1]]
+            mask = self.get_direct_beam_mask(radius=radius, center=mref)
+            z_tmp = z * mask
+            # refine center position with shifting ROI
+            if c_int == z_tmp.max():
```
francisco-dlp: I think that I'm not following: doesn't `cint` contains the estimate of the centre? If yes, what's the point of checking it is equal to the maximum of z in the mask?

francisco-dlp: ok, I got it, sorry for the noise.

hyperspy/_signals/sed.py
```diff
@@ -0,0 +1,465 @@
...
+                maxes = np.asarray(np.where(z_tmp == z_tmp.max()))
+                c[index] = np.rint([np.average(maxes[0]),
+                                    np.average(maxes[1])])
+                c[index] = c[index].astype(int)
+                c_int = z[c[index][0], c[index][1]]
+                mask = self.get_direct_beam_mask(radius=radius,
+                                                 center=c[index])
+                ztmp = z * mask
+            while c_int < z_tmp.max():
+                maxes = np.asarray(np.where(z_tmp == z_tmp.max()))
```
francisco-dlp: as above

francisco-dlp: 293-299 are very similar to the lines above. I would rather do:

``` python

while c_int < z_tmp.max():
    <"refresh" coordinates>

<store index>

```

hyperspy/_signals/sed.py
```diff
@@ -0,0 +1,465 @@
...
+        else:
+            a, b = center
+
+        y, x = np.ogrid[-b:ny-b, -a:nx-a]
+        mask = x*x + y*y <= r*r
+        return mask
+
+    def get_direct_beam_position(self, radius=None):
+        """Estimate the position of the direct beam in each SED pattern.
+
```
francisco-dlp: I would clarify here that the algorithms assume that the direct beam is saturated.

dnjohnstone: I don't think that they do? The idea is that because the direct beam shouldn't change position too much in a dataset summing reinforces this more than all other diffracted beams so in the sum we're relatively safe to guess that the maximum intensity pixel is the direct beam. However, it might move a bit so it then goes through and looks for more intense pixels near to the initial guess and if it finds one moves the center to there. I then added in the possibility to find many equal maxima to ensure that things still work IF its saturated - but if it's not, it should still work well -- in fact the tests make non-saturated models to illustrate this.

francisco-dlp: If I understood well the algorithm it:
1. Estimates an average direct beam position as the average position of the saturated area.
2. Scans through all images and only if in the image the position of the average direct beam position is not saturated recomputes the direct beam position just for this image using the same method as in 1.

If this is correct I have the following comments:
- Why 2. has to go in a while loop: it'll never perform more than 1 iteration.
- Why not calculating the position for each image independently? This should be more accurate as recomputing only when the estimation is so bad that it is out of the saturated area seems like not aiming at the best possible accuracy.
-  For big data, it would probably be better to skip the average position estimate as doing this in a huge dataset is expensive. Instead, it could:
  1. calculate it for the first image as a whole
  2. Calculate it in a ROI around the position of the first image for the following DPs. If the saturated value is not included in the ROI, take the whole image again.
- Why not estimating the radius if not given?

dnjohnstone: Or you mean just say that it still works with saturated direct beams?

francisco-dlp: Although currently `map` wouldn't work for this, I think that it'll be worth it to put this into a function in `hyperspy.misc` and call this function from inside the `iterate_signal` loop. In this way it'll be trivial to rewrite this when the generalisation of map gets merged.

hyperspy/_signals/sed.py
dnjohnstone: Honestly I wasn't really sure what was best here -- yes it's certainly more general and I'd be happy with ElectronDiffraction, but the methods are only really ok for TEMElectronDiffraction spot patterns rather than say EBSD or Transmission Kikuchi... Not really sure what structure we should be going for thinking a few steps ahead.

hyperspy/_signals/sed.py
dnjohnstone: ok -- I basically agree, this is how I originally wrote it but then moved it.

Do you think it needs to go to misc though? Or can I just specify it in sed.py but outside of the class?

hyperspy/_signals/sed.py
dnjohnstone: I think (if I remember correctly) this is something that is inconsistent in  HyperSpy quite widely - i.e. some of the masking options for model fitting want arrays whereas some for decomposition want signals -- although I may be wrong, it was some time ago that I did that. Does that seem likely and does it matter?

hyperspy/_signals/sed.py
dnjohnstone: Although I'm still not sure this has any requirement for saturation?

hyperspy/_signals/sed.py
dnjohnstone: No - this returns array indices rather than a boolean.

hyperspy/_signals/sed.py
dnjohnstone: Ok - I must admit I hadn't seen the specification before... updated.

sem-geologist:
what is this? wds? ebsd?

tjof2:
@sem-geologist it's Scanning Electron Diffraction?

dnjohnstone:
Yeah - in previous iteration's we switched it to just ElectronDiffraction.

It's a bit of a pain really because there are not so many good names. Really what it is, is "electron diffraction datasets, in which multiple spot electron diffraction patterns are acquied in an (S)TEM".

At present doesn't intend to include features to use Kikuchi lines or CBED structure. Perhaps in the end there should be: 

ElectronDiffraction

from which we inherit

TKD
EBSD
CBED
SpotDiffraction

for example.

Issue #1222, opened 2016-08-02
Update some metadata specification


ericpre:
As discussed in #1173 and #1136, we need to update some of the metada specifications:

- [x] Add authors attribute to `General`
- [x] Add date and time information as attribute of `General`. This most simple way to implement it is store it as string that can be parsed easily to numpy.datetime64 for processing.
- [x] Add doi attribute to `General`
- [x] Add notes attribute to `General`
- [x] Add quantity attribute to `Signal` as "name (units)"
- [x] Add `camera_length` and `camera_length_units` attribute to `Acquisition_instrument.TEM` and 

The quantity metadata can then be used for the intensity axes, see #546.

If there any other issue that could be related to this one? Any other relevant information to add to metadata?

dnjohnstone:
+1 for date and time using numpy.datetime
+1 for notes - that seems reasonable
+1 for authors - assuming you mean the person who saved the file, can be useful within groups at least

Not sure which doi you mean? The HyperSpy one? If so, what for?

Quantity sounds a bit vague? - what's the thinking there?

francisco-dlp:
I think that the idea behind authors and doi is making it easy to know how to cite the data. So doi should be the data doi and authors those who performed the experiment or simulation.
quantity would be e.g. "temperature". This is basically the "name" of the intensity axis. Is there a better word for this?

PR #1232, opened 2016-08-05
Add metadata

ericpre:
See #1222.

For now, I updated only the protochips io plugin, but we are all ok with the format suggested in #1222, then I can try to go through the other plugins to set the quantity, date and time properly.
I would like to have #1222 fixed to move on with #546.

With date and time saved as string in ISO 8601, this is very easy to concert it to numpy.datetime64:
`numpy.datetime64("%sT%s"%(date, time))` where date and time come from the metadata tree of a signal `s.metadata.General.date` and `s.metadata.General.time`, respectively.


doc/user_guide/metadata_structure.rst
```diff
@@ -88,15 +92,30 @@ original_filename
     original file.
 
 time
-    type: datetime.time
+    type: Str
 
-    The acquistion or creation time.
+    The acquistion or creation time in ISO 8601 time format.
```
francisco-dlp: inherited typo: acquisition

doc/user_guide/metadata_structure.rst
```diff
@@ -88,15 +92,30 @@ original_filename
...
 
 date
-    type: datetime.time
+    type: Str
+
+    The acquistion or creation date in ISO 8601 date format
```
francisco-dlp: As above

doc/user_guide/metadata_structure.rst
```diff
@@ -88,15 +92,30 @@ original_filename
...
+
+
+authors
+    type: Str
+
+    The authors who generated the files.
```
francisco-dlp: The authors of the data?

doc/user_guide/metadata_structure.rst
```diff
@@ -88,15 +92,30 @@ original_filename
...
 
-    The acquistion or creation date.
+doi
+    type: Str
+
+    Digital object identifier.
```
francisco-dlp: Digital object identified of the data?

doc/user_guide/metadata_structure.rst
```diff
@@ -300,6 +319,11 @@ record_by
     One of 'spectrum' or 'image'. It describes how the data is stored in memory.
     If 'spectrum' the spectral data is stored in the faster index.
 
+quantity
+    type: Str
+    
+    The name of the quantity of the "intensity axis" with the units in brackets if required.
```
francisco-dlp: round brackets?

francisco-dlp:
Looks good! If you could go through the other plugins that would be great.

You could also add the current version of the specification, which should be the same as the version of the hdf5 reader. Currently it is 2.1. As you're adding features but not breaking anything, the new version should be 2.2.

ericpre:
Thanks for reviewing!
I don't really understand the following:
> You could also add the current version of the specification, which should be the same as the version of the hdf5 reader. Currently it is 2.1. As you're adding features but not breaking anything, the new version should be 2.2.

Do you mean that the changes in metadata specificaitons should also be documentated as changes in the specifications of the HyperSpy hdf5 format? At the beginning of the hdf5 io plugin?

francisco-dlp:
You could indeed add a sentence there. Ideally we would fully document the format in a single document containing the current metadata specification + all the details of how we store stuff in the hdf5 file. This document should probably live in a separate repository. However, at present, things are far from perfect in this regard...

Hovden:
+1 on 8601 date, time metadata

Just noting that image stacks might have several timestamps. For example, this metadata was taken from FEI .ser dataset using hyperspy:

 Time = array([1467843134, 1467843139, 1467843144, 1467843150, 1467843155,
       1467 ...     1467843212, 1467843217, 1467843223, 1467843228, 1467843233], dtype=uint32)

ericpre:
Thanks for pointing this out! I have been through updating the FEI reader, considered only the date/time information of the emi file and neglect the one of the ser file, which was a bit lazy from my side, because in principle, we should be able to open ser files as good as we do for emi anyway...

Actually, the fact there is array of time stamp in the ser header of stack images is interesting because the time array could be use for the `navigation_axis` of the `axes_manager`, which I think for now is undefined. However, I would expect the axis not to be always linear... so that's one more for #438!

francisco-dlp:
Good points @Hovden  and @ericpre. Indeed we were tempted to set the time array in the ser file as the nav axis but we didn't because of #438 as @ericpre mentions and also because the accuracy is just 1s.

ericpre:
Resubmitted to RELEASE_next_patch #1240.

PR #1234, opened 2016-08-08
Enh xray marker subscript

vidartf:
Resubmission of  #853.
### TODO
- [x] Unittest(s)


vidartf:
Unittests added (the code in `add_xray_lines_markers` is already covered).

PR #1235, opened 2016-08-08
Load muti-stacks from multi-signal files

vidartf:
Resubmission of #522.
### TODO
- [ ] Unittests with other formats than FEI?


francisco-dlp:
This looks fine to me. Is it ready @vidartf?

vidartf:
I haven't tested it since rebasing to v.1, so at least wait for CI to complete, but other than that it should be OK. I also made a point above to check whether there are any other formats with such signal structures that could also be tested for. @sem-geologist Did you know any such?

francisco-dlp:
My only concern with this is that the featured could be added to `stack` instead of `load` as discussed in #522. However, if you prefer to leave it as it is and I think that this is ready to merge.

vidartf:
While I agree with the idea in principle, that will require a major overhaul of the stack function, which I cannot commit now. As the feature works as advertised as it is currently implemented, I would leave such an exercise to later.

PR #1240, opened 2016-08-09
Update/add metadata

### LINKED ISSUE #1222: Update some metadata specification

ericpre:
Resubmission of #1232 and fix #1222.


doc/user_guide/metadata_structure.rst
```diff
@@ -88,15 +95,30 @@ original_filename
...
+    type: Str
+
+    The acquisition or creation date in ISO 8601 date format
+
+
+authors
+    type: Str
 
-    The acquistion or creation date.
+    The authors of the data.
```
francisco-dlp: in latex format? surname1, name1 and surname2, name2 etc.

doc/user_guide/metadata_structure.rst
```diff
@@ -88,15 +95,30 @@ original_filename
...
+
+doi
+    type: Str
+
+    Digital object identifier of the data.
```
francisco-dlp: maybe specify the format? doi:10.5281/zenodo.58841

hyperspy/io_plugins/bcf.py
```diff
@@ -1039,53 +1047,57 @@ def bcf_hyperspectra(obj_bcf, index=0, downsample=None, cutoff_at_kV=None):
...
+        'General': {'original_filename': obj_bcf.filename.split('/')[-1],
+                         'title': 'EDX',
+                         'date': obj_bcf.header.date,
+                         'time': obj_bcf.header.time},
+        'Sample': {'name': obj_bcf.header.name,
+                         'elements': sorted(list(obj_bcf.header.elements)),
+                         'xray_lines': sorted(gen_elem_list(obj_bcf.header.elements))},
+        'Signal': {'signal_type': 'EDS_%s' % mode,
+                         'record_by': 'spectrum',
+                         'quantity': 'X-ray intensity (counts)'},
```
francisco-dlp: Wouldn't simply "counts" be better? from the signal_type we already know that these are X-rays

hyperspy/io_plugins/bcf.py
```diff
@@ -1103,3 +1115,10 @@ def parse_line(line_string):
     if len(line_string) == 1:
         line_string = line_string + 'a'
     return line_string.capitalize()
+
+
+def _get_mode(obj_bcf):
+    if obj_bcf.header.sem.hv > 30.0:  # workaround to know if TEM or SEM
```
francisco-dlp: Smart, should work most of the time... but not always. What about adding an entry to the logger explaining how it has decided that it is TEM or SEM? And maybe the reader could take a keyword so that the user can specify it e.g. `load("filename", instrument="TEM")`?

hyperspy/io_plugins/dens.py
```diff
@@ -103,8 +99,11 @@ def file_reader(filename, *args, **kwds):
     interp_axis = times[0] + dt * np.array(range(len(times)))
     temp_interp = interp(interp_axis)
 
-    metadata = {'General': {'original_filename': os.path.split(filename)[1]},
-                "Signal": {'signal_type': "temperature", }, }
+    metadata = {'General': {'original_filename': os.path.split(filename)[1],
+                            'date': date,
+                            'time': time},
+                "Signal": {'signal_type': "", 
+                           'quantity': "Temperature (celsius)"}, }
```
francisco-dlp: _C_elsius?

hyperspy/io_plugins/bcf.py
francisco-dlp: Actually, I'm not sure about this. What about X-rays (counts)?

hyperspy/io_plugins/digital_micrograph.py
```diff
@@ -764,18 +765,82 @@ def get_metadata(self, metadata={}):
         metadata["Signal"]['signal_type'] = self.signal_type
         return metadata
 
+    def _get_quantity(self, units):
+        quantity = "Intensity"
```
francisco-dlp, line 769: What about Electrons (counts) and X-rays (counts)?

hyperspy/io_plugins/fei.py
```diff
@@ -665,28 +671,43 @@ def guess_units_from_mode(objects_dict, header):
...
         "Acquisition_instrument.TEM.microscope",
-        None),
+        _get_microscope_name),
     "ObjectInfo.ExperimentalDescription.Mode": (
         "Acquisition_instrument.TEM.acquisition_mode",
-        get_simplified_mode),
+        _get_simplified_mode),
+    "ObjectInfo.ExperimentalDescription.Camera length_m": (
+        "Acquisition_instrument.TEM.camera_length",
+        lambda x: x*1e3),
```
francisco-dlp: small style issue `x * 1e3`

hyperspy/io_plugins/msa.py
```diff
@@ -216,34 +212,14 @@ def parse_msa_string(string, filename=None):
                     mapped.set_item(keywords[clean_par]['mapped_to'] +
                                     '_units', units)
 
-    # The data parameter needs some extra care
```
francisco-dlp, line 219: I think that most of the removed code was actually necessary. The msa data and time format follows US conventions, and that's why the locale was set temporarily to US (there may be a better way of achieving this though). However, there are a lot of msa files out there with date and time in the wrong format. Therefore, if the parsing is not inside a try, except, the reader may raise an exception for many files that contain valid data but invalid date.

hyperspy/io_plugins/msa.py
```diff
@@ -298,23 +282,13 @@ def file_writer(filename, signal, format=None, separator=', ',
     else:
         if format is None:
             format = 'Y'
-        if signal.metadata.has_item("General.date"):
```
francisco-dlp, line 301: As above

hyperspy/io_plugins/ripple.py
```diff
@@ -521,16 +532,11 @@ def file_writer(filename, signal, encoding='latin-1', *args, **kwds):
     # Check if the dtype is supported
     dc = signal.data
     dtype_name = signal.data.dtype.name
+    dimension = len(signal.data.shape)
     if dtype_name not in dtype2keys.keys():
         err = 'The ripple format does not support writting data of %s type' % (
             dtype_name)
         raise IOError(err)
-    # Check if the dimensions are supported
```
francisco-dlp, line 528: Why removing this?

hyperspy/io_plugins/bcf.py
ericpre: Yes, adding here X-ray here is a bit redundant because we already that there are X-ray. However, I would say that the most accurate description of the quantity is "X-ray intensity (counts)" and when we plot the signal, we can simply pass this attribute to the ylabel.

francisco-dlp: I agree. But, isn't "intensity" a bit misleading? For eels would be say electron intensity?

hyperspy/io_plugins/bcf.py
ericpre: Yes... will an optional argument. Would be an explanation in the documentation enough, instead of the logger?

francisco-dlp: I think that it'll be easier for the users to know what's going on if the information is written in the log. For example:

`Guessing that the acquisition instrument is TEM because the beam energy is 20 keV. If I guessed wrong, please provide the right instrument using the `instrument` keyword.`

hyperspy/io_plugins/bcf.py
ericpre: I think "electron intensity" is quite good for EELS, although this is not the most widely used and may sound a bit unusual... but most of the people (including myself) are a bit lazy when making figure and we end up with EELS signal, EELS intensity (a. u.) or even intensity (a. u.) on its own.
EELS intensity with either "counts" or "electrons" as units (depending, if the conversion factor is used or not) will be also good.

francisco-dlp: What if we were counting apples, would it be `Apple intensity (counts)`?

ericpre: well, could be, if someone manages to get apples down there!! ;)
More seriously, idealy, we would like to find proper wording to describe this specific quantity. I think, we should be more precise than just labelling with electron or X-ray. But, at the same time, intensity can take different definitions, so it's not very appropriate neither...

hyperspy/io_plugins/msa.py
ericpre, line 219: Yes, I will revert this. I missed out the abbreviation of the month, which can be different using different localisation...

hyperspy/io_plugins/ripple.py
ericpre, line 528: I think, this part of the code is never used, because it is filtered in the [hyperspy.io](https://github.com/hyperspy/hyperspy/blob/RELEASE_next_minor/hyperspy/io.py#L422)

francisco-dlp, line 528: Yes, you are right. However in the future we may split the plugins into a separate project so that other packages can import them without having to import the whole HyperSpy and this error may be handy if this ever happens.

ericpre, line 528: Ok, fair enough, I will put it back.

ericpre:
This should be fairly ready. I added a few more metadata for EELS.

doc/user_guide/metadata_structure.rst
```diff
@@ -88,15 +97,30 @@ original_filename
...
+    type: Str
+
+    The acquisition or creation date in ISO 8601 date format
 
-    The acquistion or creation date.
 
+authors
+    type: Str
+
+    The authors of the data, e. g. surname1, name1 and surname2, name2, etc.
```
francisco-dlp: Instead of e.g. I would be more strict: "in Latex format: Surname1, Name1 and Surname2, Name2 and ..."

doc/user_guide/metadata_structure.rst
```diff
@@ -208,7 +246,17 @@ dwell_time
 exposure
     type: Float
 
-    The exposure time in seconds. This is relevant for TEM acquistion.
+    The exposure time in seconds. This is relevant for TEM acquisition.
+
+frame_number
+    type: int
+
+    The number of frames/spectra integrated during the acquisition.
```
francisco-dlp, line 254: Is the dwell time is the dwell time of the spectra or the frame?

doc/user_guide/metadata_structure.rst
```diff
@@ -208,7 +246,17 @@ dwell_time
...
+
+spectrometer
+    type: Str
+
+    The spectrometer model, e.g. Gatan 666
```
francisco-dlp: I would give a more complete example e.g. Gatan Enfinium ER (Model 977)

doc/user_guide/metadata_structure.rst
```diff
@@ -300,6 +348,11 @@ record_by
     One of 'spectrum' or 'image'. It describes how the data is stored in memory.
     If 'spectrum' the spectral data is stored in the faster index.
 
+quantity
+    type: Str
+    
+    The name of the quantity of the "intensity axis" with the units in round brackets if required.
```
francisco-dlp: The name of the quantity of the signal with the units in round brackets if required. For example Temperature (K)?

hyperspy/io_plugins/msa.py
```diff
@@ -263,6 +262,13 @@ def parse_msa_string(string, filename=None):
     else:
         # Defaulting to EELS looks reasonable
         mapped.set_item('Signal.signal_type', 'EELS')
+    if 'YUNITS' in parameters.keys():
```
francisco-dlp, line 267: Many times YUNITS are not defined but the SIGNALTYPE is. What about guessing the quantity from the signal type instead i.e. EELS -> Electrons (counts), EDS -> Xrays (counts)?

francisco-dlp, line 267: Sorry, that comment wasn't very clear: I was proposing to do that when YLABEL and/or YUNITS are not defined.

doc/user_guide/metadata_structure.rst
ericpre, line 254: I don't really understand the question.
As it is, dwell time is the total integration time of the spectrum and in case several spectra have been cumulated during the acquisition, the frame_number is saved here. Not sure, it's very clear here though! Maybe I should just replace the word 'frame' with something else.

hyperspy/io_plugins/msa.py
ericpre, line 267: yes

francisco-dlp:
I have just sent you a ericpre/hyperspy#1 to enable reading of date and time from files written with hyperspy < 1.

The tests regarding reading of date and time from msa files fail in my linux machine. We don't know if they fail in travis because travis fails because of #1250. Could you merge RELEASE_next_patch into your branch to check if those tests fail in travis too?

ericpre:
Without merging with #1250, the tests pass on CI and on my machine.

francisco-dlp:
The issue turned out to be that I don't have en_US locales in my system.

PR #1241, opened 2016-08-10
New component: Gaussian2Dsymmetric, and add rotation to Gaussian2D

magnunor:
Resubmitted: https://github.com/hyperspy/hyperspy/pull/864

Need to update this to work with RELEASE_next_minor, and address the things pointed out in https://github.com/hyperspy/hyperspy/pull/864.


magnunor:
Should by updated to the current next_minor, still need to sort out the feedback from the old pull request.

hyperspy/_components/gaussian2d.py
```diff
@@ -90,3 +117,12 @@ def fwhm_y(self):
     @fwhm_y.setter
     def fwhm_y(self, value):
         self.sigma_y.value = value / sigma2fwhm
+
+    @property
+    def rotation_degrees(self):
+        """ Angle between major axis and x-axis."""
+        rotation = math.fmod(self.rotation.value, pi2)
```
tjof2: returning to my comment in #864 - is it not constrained to between 0 and pi, since the semi-major axes are indistinguishable?

magnunor:
I added the rotation bound, but it doesn't work:

``` Python
----> 1 hs.model.components2D.Gaussian2D()

/home/magnunor/Documents/HyperSpy/hyperspy/hyperspy/_components/gaussian2d.py in __init__(self, A, sigma_x, sigma_y, centre_x, centre_y, rotation)
     74                                   'centre_x',
     75                                   'centre_y',
---> 76                                   'rotation',
     77                                   ])
     78         self.A.value = A

/home/magnunor/Documents/HyperSpy/hyperspy/hyperspy/component.py in __init__(self, parameter_name_list)
    738             """, arguments=["obj", 'active'])
    739         self.parameters = []
--> 740         self.init_parameters(parameter_name_list)
    741         self._update_free_parameters()
    742         self.active = True

/home/magnunor/Documents/HyperSpy/hyperspy/hyperspy/component.py in init_parameters(self, parameter_name_list)
    848             parameter.name = name
    849             parameter._id_name = name
--> 850             setattr(self, name, parameter)
    851             if hasattr(self, 'grad_' + name):
    852                 parameter.grad = getattr(self, 'grad_' + name)

/home/magnunor/Documents/HyperSpy/hyperspy/hyperspy/_components/gaussian2d.py in rotation(self, value)
    111     @rotation.setter
    112     def rotation(self, value):
--> 113         self.rotation.value = value % math.pi
    114 
    115     @property

TypeError: unsupported operand type(s) for %: 'Parameter' and 'float
```

Is it possible to use a setter for a parameter?

magnunor:
I updated the tests, which work now. Except for the issue with the rotation setter.

Where should the tests for the components be? Currently they're split between tests/component and test/model.

How to fix the rotation so it is periodic, ergo constrained between 0 and pi? Does any of the other parameters have setters for their values?

magnunor:
I added new value to the component class: the tuple wrapped_value (low_value, high_value) which wraps the parameter value between low_value and high_value.

magnunor:
Fixed a bug in the `wrapped_value` functionality + add generic tests in `test_parameter`.

Not certain if "wrapped_value" is the best name, I'm happy for any suggestions :)

Should be ready for review.

francisco-dlp:
Would you like to rewrite this using `Expression` #1368? In principle it should be possible to keep the same interface except for the rotation angle parameter name that in #1368 is hardwired to be `angle` (although I think that it'll be better to change it to `rotation_angle`).

Regarding the other features that this PR implements, I am not sure if the angle wrapping feature is a good idea. I think that it may make the fitting unstable as going from pi to pi + epsilon would require the optimiser to find its way the other side of the interval. Also, the automatically generated grad would be wrong and the ends of the interval. wraping (unlike unwraping) is a well defined mathematical operation, so maybe a method to return the wrapped angle would be better?

magnunor:
@francisco-dlp, sounds good! Then I can finally remove https://gitlab.com/atomap/atomap/tree/master/atomap/external :)

Regarding the angle wrapping: I'm not sure what you mean. I think the best implementation is simply letting the `rotation` parameter be unbounded, and simply have a "helper" function like `rotation_wrapped` which returns a value between 0 and pi.

francisco-dlp:
Possibly because I'm wrong ;) Chances are that it won't affect the fitting. Still, my gut feeling tells me that it's better (at least it's certainly faster) to go for the simpler solution: add a method to get the wrapped angle that operates on `map["values"]` instead of doing it internally at every change of `angle.value`. What do you think?

magnunor:
Changing the `map['values']` would probably be a better solution compared to my helper function. But I'm not certain how I would implement that.

francisco-dlp:
Actually I wasn't meaning changing the content of `map["values"]` but rather returning it unwrapped e.g.

```python

uw = g.unwrapped_angle

```

Implementing it may be just a matter of inheriting from `Expression`:

```python

class Gaussian2D(Expression):
    def __init__(self, <parameter defaults + module)>:
        super().__init__(expression="<gaussian expression>", add_rotation=True, <etc>)
    
    def wrapped_angle(self):
        angle = self.rotation_angle.as_signal()
        angle.data = (angle.data + np.pi) % (2 * np.pi ) - np.pi
        angle.metadata.General.title = "Wrapped {}".format(angle.metadata.General.title)
        return angle
```

magnunor:
Ah, ok. That is easy.

francisco-dlp:
If it's for a good cause, let's break the rule!

francisco-dlp:
Sorry, the comment above was intended for #1370.

ericpre:
@magnunor: this looks almost finished, could you please solve the merge conflicts and address @francisco-dlp comments?

dnjohnstone:
Is this still relevant @magnunor ? Would be good to either tidy up for merge or close.

magnunor:
Indeed, this has been superseded by using the `Expression` component!

ericpre:
To my opinion, this is still relevant because the Gaussian component, which is indeed now implemented as an expression, doesn't have the rotation as parameter.

PR #1243, opened 2016-08-12
New: s.rotate() with crop

thomasaarholt:
New method to rotate signal with a crop argument to return the largest non-zero-containing rectangle inside it that is aligned with the x and y axis.

Based on [`scipy.ndimage.interpolation.rotate`](http://docs.scipy.org/doc/scipy-0.16.0/reference/generated/scipy.ndimage.interpolation.rotate.html)

For now I let the user choose between rotating the `"signal"` or `"navigation"` axes with a single keyword. It could be expanded upon to take an axes = (2,1) argument as well.

It has a different default setting in that 

The setting of the axes scale etc. isn't very elegant, but I didn't know how I should go about it in a different fashion.

Please, if someone could try it out and report back how to works for them, I would appreciate it.
### TODO
- [x] Update docstring
- [ ] Use `out` argument - someone help with this?
- [ ] Need some tests. I'd like to do these to learn, but it would be excellent if someone would walk me through one so that I don't mess up. Does it make sense to add it to test_tools.py?
- [ ] Do we want rotate to take additional arguments?
### OBSERVATIONS
- The x and y scale with reshape=True will differ minutely, as they are calculated independently (it's because the image pixel dimensions are discrete). This sets the rotated plot to have tick marks along the x and y axis separately rather than a scale bar. That's currently necessary in case someone wants to rotate a signal where the scales for x and y are different. I guess one could check whether the scales are the same for the initial signal and if they are set the new scales to be the average of the two calculated x and y scales.


tjof2:
@thomasaarholt  the cython file already is in the .gitignore... https://github.com/hyperspy/hyperspy/blob/RELEASE_next_minor/.gitignore

thomasaarholt:
@tjof2 That's what I thought. No idea why it kept cropping up. I was trying to tidy up some branches on my side which might have done it.

thomasaarholt:
This is now working. The code isn't very attractive so I'd love for someone to go through it as well as test it out. I've updated the description.

thomasaarholt:
I need a way of setting the new signal to be of the same type (i.e. EDSTEMSpectrum) as the previous spectrum. Any suggestions as to how to "apply" this?

It would be nice to copy over all elements and lines as well, in the case of EDS.

vidartf:
Without having looked at the code, a new signal could be created using `deepcopy_with_new_data()`. You'd need to update some fields (axes manager), but everything else in the signal will just be copied over.

thomasaarholt:
Thanks, I was looking for that method but had forgotten what it was called.

k8macarthur:
@thomasaarholt as I understand the 'out' argument simply specifies a place for the resulting new data cube to go. In the rebin function it's written as:
```
if out is None:
return s
else: 
out.events.data_changed.trigger(obj=out)

k8macarthur:
Also, if the function can rotate around any axis. Should it be based in array_tools with signal.py calling it?

thomasaarholt:
I've been meaning to get back into this PR. Your suggestion regarding array_tools seems sensible.

thomasaarholt:
Closing this as it's unlikely that I'll find the time to do it properly.

ericpre:
@thomasaarholt, would you have time these days and be keen to finish this PR?

thomasaarholt:
I do! I've worked through most of it now, will hopefully finish tomorrow (let's see :p )

thomasaarholt:
If s is some multidimensional Signal2D, and I'd like to slice axes with indices `n` and `m` is there any way to slice only them? Something along the lines of `s.islice(index, from, to)`? (instead of having to know whether or not a given axis is in the navigation or signal axes?

ericpre:
@thomasaarholt: any progress on this PR?

> If s is some multidimensional Signal2D, and I'd like to slice axes with indices `n` and `m` is there any way to slice only them? Something along the lines of `s.islice(index, from, to)`? (instead of having to know whether or not a given axis is in the navigation or signal axes?

Sorry I don't understand what axes you are referring with `n` and `m`. Many an example would help.

thomasaarholt:
Yes, that wasn't very clear.

In my current code (I haven't pushed recent changes yet), I include an `axes` argument that can either take `axes="signal"` or `axes="navigation"`, if the respective set of axes only have two dimensions. In addition, I'd like to be able to specify any two axes, regardless of whether they are both in the signal or navigation axes.

If I have a Signal2D of shape `(m n o | p q)`, I'd like the user to be able to specify `axes=(0, -1)` if for some reason this was a logical rotation. Or `axes=(0,1)` to be able to specify which navigation axes are to be rotated.

However, my problem arrives when it comes to crop the axis, which I have been doing in order to optionally crop to the largest rectangle without zeros. For the `axes="signal"` or `axes="navigation"` cases, I can use `isig` and `inav` respectively. But I can't use this generically for one axis from the signal and one from navigation axes.

Using the signal example above, if I need to crop the axis with shape `m` (axis=0) from `0.0:5.0` and the axis with shape `q` (axis=-1) from `10.0:20.0`, it would be nice to be able to do:

`cropped_s = s.slice_function(axis=0, from=0.0, to=5.0).slice_function(axis=-1, from=10.0, to=20.0)`.

Of course, I could just limit myself to the signal dimension of Signal2Ds, and not worry about this.

ericpre:
Doing a rotation over a mixture or navigation and signal axes doesn't sound a very natural things to do. However, be able to select two axis of the navigation axes (for a signal having more than 2 navigation axes) to perform the rotation would be very useful. If you need to know which axes of the numpy array, [`index_in_array`](https://github.com/hyperspy/hyperspy/blob/677c15a3585f3e540c9a78a22fb02f7b7ea0959d/hyperspy/axes.py#L312) may be useful.

ericpre:
@thomasaarholt: do you know when/if you will be able to finish this PR?

Issue #1170, opened 2016-07-18
How to contribute


francisco-dlp:
The following is a suggestion on how to contribute to HyperSpy to be added to the Dev Guide after discussion here:

Following #904  and #1060 there are three new _permanent_ branches:
- RELEASE_next_major
- RELEASE_next_minor
- RELEASE_next_patch

~~The master is the nickname of the (non-existent) RELEASE_next_major branch.~~

Therefore, when working on new pull requests _branch_ and _pull request to_:
- RELEASE_next_major **only** if what you're working on requires changing the API
- RELEASE_next_minor if you're planning to add a new feature or enhance an existing one.
- RELEASE_next_patch if you are fixing a bug in the code or improving the docs.

The default branch is RELEASE_next_minor.

Any new PR, if considered for inclusion in HyperSpy, will be labelled one of RELEASE_next_minor, RELEASE_next_major, RELEASE_next_patch. Once the PR is merged, it'll be added to the milestone of the following release of its kind i.e. next major increment for a major, next minor increment for a minor and next patch increment for a patch PR. Once a milestone is closed all RELEASE labels of the associated PRs must be removed.

How to manage new PRs - Summary for developers:
- [ ] Verify that the author has branched from the correct RELEASE branch
- [ ] Add RELEASE label
- [ ] When closing release milestone remove RELEASE label

dnjohnstone:
I think this is all good, but should be added to the Dev Guide ASAP so that people can get there branches in order before everything gets massively out of sync.

On a practical note - it seems to me that if I have a branch adding a new feature that is currently based on the master then all I need to do at this point is to merge it with RELEASE_next_minor and then keep doing that (crucially not with master) . Anyone who has big branch changes locally will appreicate getting that right the first time as it's a real pain to unpick after (yes I did it, was too used to merging master).

Secondly - do you want me to knock this into words for the Dev Guide? I know that words come quite quickly to me in English.

francisco-dlp:
Yes, there is no difference between RELEASE_next_minor and master at present, so that approach should work... but not for a long time.

If you're happy to add this to the Dev Guide, please do.

francisco-dlp:
To avoid confusion, I have just renamed master to RELEASE_next_major. Also, RELEASE_next_minor is now the default branch.

thomasaarholt:
[...]

PR #1245, opened 2016-08-12
Fixed Travis badge for RELEASE_next_patch

tjof2:
...which meant the Travis badge looked like the build was failing:

untitled

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -2,11 +2,11 @@
 
 |Travis|_ |AppVeyor|_ |Coveralls|_ |pypi_version|_ |pypi_downloads|_ |rtd|_ |gitter|_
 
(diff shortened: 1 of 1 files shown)


francisco-dlp:
Shouldn't it point to RELEASE_next_patch now since you're sending the PR to RELEASE_next_patch?

tjof2:
Ah good point. But then do I need to do another one for `RELEASE_next_minor`? I only sent the PR here based on #1170 because it is technically a bug :-)

francisco-dlp:
This reminds me that we should disable the downloads badge until they fix the downloads statistics in pypi.

francisco-dlp:
No, I'll do that when merging into the other branches, thanks!

PR #1248, opened 2016-08-14
ENH: Auto contrast field on image plot

vidartf:
This reintroduces the `auto_contrast` field on `ImagePlot` so that the auto/manual state of vmin/vmax can be assessed without accessing private variables.

### DIFF
--- a/hyperspy/drawing/image.py
+++ b/hyperspy/drawing/image.py
@@ -123,6 +123,28 @@ class ImagePlot(BlittedFigure):
     def auto_contrast(self):
         return (self._vmin_user is None, self._vmax_user is None)
 
+    @auto_contrast.setter
+    def auto_contrast(self, value):
+        if isinstance(value, (list, tuple)):
+            if len(value) != 2:
+                raise ValueError("Value to auto_contrast must be a boolean or "
+                                 "a list/tuple of two booleans")
+            if value[0]:
+                self._vmin_user = None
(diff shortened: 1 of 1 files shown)


francisco-dlp:
Why not simply using the public `vmin` and `vmax` e.g. `im._plot.signal_plot.vmin = -20`?

vidartf:
This is not only to set the variables, but to check whether they are in automatic or manual mode (e.g. indicate to user in UI with checkbox "Auto contrast").

francisco-dlp:
That's the getter part, what about the setter?

vidartf:
For when someone clicks the checkbox :) 

``` python
plot.auto_contrast = checkbox.checked()
```

francisco-dlp:
I see. What I'm trying to find out is if this is something that should be incorporated to HyperSpy or if there is a simpler solution. For example, wouldn't making `_vmin_user` and `_vmax_user` public fix the issue without the extra code?

vidartf:
For me, I just want the UI to not rely on protected/private variables in hyperspy too much, and this was my best attempt at a solution. If you want to do it another way, that's no problem for me, the most important thing is that it gets done. Until something is decided, the UI will only know about the private variables. Once a decision is made, the UI can check for the presence of the public ones, and fall back to the private ones if they are not there (for wider hyperspy 1.x compatibility).

PR #1249, opened 2016-08-14
Set default log level to NOTSET

vidartf:
Currently, hyperspy sets its log level during import, thereby overriding any other log configuration already set for the application/hyperspy. The effect is a lowering of the logging flexibility of hyperspy when used as a library. This PR changes the default to `NOTSET` (which _here_ circumvents the call to `setLevel` all together), which means it will keep its previous value (default: inherit from the root logger, which defaults to `WARNING`). The effective log level will thereby stay the same out of the box, users will retain the ability to change the log level, but it will become more flexible for those that want to use hyperspy as a library.

Example of uses that will change with this PR:

1)

``` python
import logging

logging.basicConfig('INFO')
import hyperspy
```

Previously, this set the application wide logger to `INFO`, and hyperspy would use its own setting from the preferences (`WARNING` by default). Now, this will use `INFO` also for hyperspy, unless the user has explicitly set the preference to something other than `NOTSET`.

2) 

``` python
import logging

logging.basicConfig()  # Setup standard root logger at WARNING level
logging.getLogger('hyperspy').setLevel(logging.DEBUG)

import hyperspy
```

This looks similar to simply importing `set_log_level` from hyperspy, and then calling it with `DEBUG`, but this allows for setting the log level _before_ actually importing `hyperspy.api`, which can be helpful in cases when import order matters.

### DIFF
--- a/hyperspy/api.py
+++ b/hyperspy/api.py
@@ -80,6 +80,7 @@ from hyperspy import datasets
 from hyperspy.logger import set_log_level
 
 if preferences.General.logging_level == 'NOTSET':
+    import logging
     logging.basicConfig()  # Does nothing if already configured
 else:
     set_log_level(preferences.General.logging_level)
--- a/hyperspy/api.py
+++ b/hyperspy/api.py
@@ -79,7 +79,10 @@ from hyperspy.utils import *
 from hyperspy import datasets
 from hyperspy.logger import set_log_level
 
-set_log_level(preferences.General.logging_level)
+if preferences.General.logging_level == 'NOTSET':
+    logging.basicConfig()  # Does nothing if already configured
+else:
+    set_log_level(preferences.General.logging_level)
 
 
 def get_configuration_directory_path():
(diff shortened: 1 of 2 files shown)


francisco-dlp:
Couldn't you modify the preferences before importing api instead?:

``` python

preferences.General.logging_level = "INFO"
import hyperspy.api as hs

```

vidartf:
Yes, that is a solution that would work, and that will still work after this PR. However, it will still override application wide logging setups (sometimes you want to say "Only error logging from all modules") in the "out-of-the-box" configuration.

PS: I would also like to add a change event, so that the run-time logging level updates when you change the preference, and not have to wait until a restart for that.

francisco-dlp:
I see but I don't think that any external application should rely on anything in preferences having its default value. So, do you think that adding the default logger level values to preferences is bad practice?

+1 on setting the logger level in preferences (if the feature should stay, depending on the thoughts regarding the first question).

vidartf:
Well, in general it is bad practice for a library to set its own logging level. However, since hyperspy is often used directly as an "application", I can see how setting it in `api` can be very practical. As such, I'm trying to figure out a comprise that can work for both cases.

PR #1252, opened 2016-08-15
Fix wrong metadata dm

ericpre:
The camera length wasn't parsed properly in TEM.
All the metadata weren't tested, this should be fine now.


ericpre:
#1222

PR #1253, opened 2016-08-15
Testing plot

ericpre:
Resubmission of #1200


ericpre:
This should be ready for review!

ericpre:
@francisco-dlp, when I run the baseline images that you generated, I get very small differences (the difference images are black with a few dark pixels, i.e. very low difference value). Although I don't know where the differences come from, setting the tolerance to a very low value (0.05) seems to do the job.

francisco-dlp:
Is this WIP?

ericpre:
Yes this is still WIP... and hopefully, it will finally pass travis and appveyor... I was struggling with fonts in matplotlib that will consistent across different matplotlib install!
The most simple way I found to get the same fonts is to install matplotlib from pip before running the tests.
If we agree on this approach, then there are still some tests and the documentation to update.

ericpre:
Travis and appveyor are now passing. There are still some tests failing on osx because of numpy 1.13 (see  #1441 and #1445).

This important thing here to pass the test is: _matplotlib needs to be installed using pip._ If your version of matplotlib has the same fonts than the one from pip, then this is not a requirement.
Do you think, this is a good approach?

hyperspy/misc/test_utils.py
```diff
@@ -163,6 +168,52 @@ def assert_warns(message=None, category=None):
...
+
+
+def reset_rcParams_default():
+    import matplotlib.pyplot as plt
+    plt.rcParams.clear()
+    plt.rcParams.update(plt.rcParamsDefault)
+
+
+@simple_decorator
+def switch_backend_mpl(function):
```
francisco-dlp, line 188: I've just discovered that it's possible to set the mpl backend using [environement variables](https://github.com/matplotlib/matplotlib/blob/master/.travis.yml#L149). Isn't that preferable to this decorator?

francisco-dlp:
Why do the tests pass when installing mpl using pip?

From the mpl docs it seems that the issue is the freetype version installed in the OS. They workaround the issue [installing a local copy](https://github.com/matplotlib/matplotlib/blob/master/.travis.yml#L136). Shouldn't we do the same? Shouldn't we do the same? If yes, the baseline images should be generated using the local freetype option.

hyperspy/misc/test_utils.py
ericpre: Before this decorator was used a bit more with nose. Now with pytest, it is used only once, so it doesn't make sense anymore to have it as a decorator...

This is interesting to know how it's possible to set the backend in travis, however, I don't think this is necessary because this [function](https://github.com/ericpre/hyperspy/blob/testing_plot/hyperspy/misc/test_utils.py#L181) is called in __init__.py and reset rcParams to default values, which set the backend to `agg` at the same time. It also reset any user rcParams and should make plot testing more reliable.

ericpre:
If we set a specific freetype, this will possibly work nicely with travis and appveyor but it will failed for people using matplotlib from conda, for example.
By using the wheel from pip, I guest we can use the same freetype across different system. Avoiding compiling matplotlib also saved a bit of travis time.

What was a bit annoying here was to find a way to use the same freetype (travis and appveyor but also different user environment...)!!

francisco-dlp:
I see the point, but I don't understand why it works. So, are mpl wheels compiled with ``local_freetype = True``? The default is ``False`` in the source distribution.

ericpre:
I don't know what version of freetype is used in the pypi wheels and I don't even know how to find it out or how these wheels are created, etc.

francisco-dlp:
Well, if it ain't broken don't fix it.

ericpre:
I still have to tidy up a few things: update dev-guide and there a few tests that still need to be updated/improved for pytest! I m doing it fairly soon.

francisco-dlp:
I could revert it, but, as this doesn't affect users and this is a huge improvement anyway that enable us to know when we've broken the plotting features, I would suggest to leave it as is and add the improvements in a separate PR, does it sound good?

ericpre:
Yes, I started to do it in #1461.

PR #1256, opened 2016-08-19
Implement metadata-aware intensity labels for signal plots

### LINKED ISSUE #1255: Colorbar label
It would be good to have an option to add labels to the colorbar (i.e. 'intensity'). 
Or is this option available already and i just didn't find it?

Moreover, it would be good if this label also adjusts to the signal type. If for example a phase is calculated from a ComplexSignal, the colorbar should be labeled with 'phase (rad)' or 'phase shift (rad)' or something like this.
## TODO
- [x] `Signal2D` Add quantity to colorbar title
- [x] `Signal1D` Add quantity to ylabel title

bm424:
Fix #1255 

Much like the signal title, the signal's `quantity` is now also passed to the plot assuming it exists.

### DIFF
--- a/hyperspy/drawing/image.py
+++ b/hyperspy/drawing/image.py
@@ -262,7 +262,7 @@ class ImagePlot(BlittedFigure):
 
         if self.colorbar is True:
             self._colorbar = plt.colorbar(self.ax.images[0], ax=self.ax)
-            self._colorbar.set_label(self.quantity_label, rotation=-90, labelpad=15.)
+            self._colorbar.set_label(self.quantity_label, rotation=-90, va='bottom')
             self._colorbar.ax.yaxis.set_animated(True)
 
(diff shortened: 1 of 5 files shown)


hyperspy/drawing/image.py
```diff
@@ -261,6 +262,7 @@ def plot(self, **kwargs):
 
         if self.colorbar is True:
             self._colorbar = plt.colorbar(self.ax.images[0], ax=self.ax)
+            self._colorbar.set_label(self.quantity_label)
```
francisco-dlp: Probably just a matter of taste: `set_label(self.quantity_label, rotation=-90)`?

bm424: Works for me, I only left it sideways because I noticed the y-axis labels elsewhere were also sideways.

PR #1260, opened 2016-08-22
Fix bcf reader import error message

francisco-dlp:
Previously it was printing the whole traceback because of the commas after the
quotation marks. This also makes the text more compact.


vidartf:
LGTM

PR #1261, opened 2016-08-22
Model.multifit raises exception when bounds=True and fitter is not specified

francisco-dlp:
The following reproduces the error:

``` python
>>> import hyperspy.api as hs
>>> hs.signals.Signal1D([0]).create_model().multifit(bounded=True)
  File "<ipython-input-7-989f16f906fb>", line 4, in <module>
    m.multifit(bounded=True)
  File "/home/fjd29/Python/hyperspy3/hyperspy/model.py", line 1199, in multifit
    if kwargs['fitter'] not in ("leastsq", "TNC", "L_BFGS-B", "mpfit"):
KeyError: 'fitter'
```

This fixes the issue.

### DIFF
--- a/hyperspy/model.py
+++ b/hyperspy/model.py
@@ -911,7 +911,7 @@ class BaseModel(list):
         if bounded is True:
             if fitter not in ("leastsq", "mpfit", "TNC",
                               "L-BFGS-B", "Differential Evolution"):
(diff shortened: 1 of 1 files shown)


tjof2:
Good catch - I now remember thinking about this when updating the fitters, but then clearly forgot about it...

PR #1264, opened 2016-08-23
Fix appending component to model which plots components

vidartf:
The following code reproduces the issue:

``` python
import numpy as np
import hyperspy.api as hs
s = hs.signals.Signal1D(np.random.rand(10, 10, 10))
m = s.create_model()
m.plot()
# This goes fine
m.append(hs.model.components1D.GaussianHF())
m.enable_plot_components()
# This fails:
m.append(hs.model.components1D.GaussianHF())
```

This PR also allows to create an empty EDS model (i.e. without any elements added to signal first) with the default constructor args.

### DIFF
--- a/hyperspy/model.py
+++ b/hyperspy/model.py
@@ -481,8 +481,6 @@ class BaseModel(list):
             for parameter in component.parameters:
                 parameter.events.value_changed.connect(
                     self._model_line.update, [])
-        if self._plot_components is True:
-            self._connect_component_lines()
 
     def _disconnect_parameters2update_plot(self, components):
         if self._model_line is None:
@@ -492,8 +490,6 @@ class BaseModel(list):
             for parameter in component.parameters:
                 parameter.events.value_changed.disconnect(
                     self._model_line.update)
-        if self._plot_components is True:
-            self._disconnect_component_lines()
 
     def update_plot(self, *args, **kwargs):
         """Update model plot.
--- a/hyperspy/models/edsmodel.py
+++ b/hyperspy/models/edsmodel.py
@@ -150,7 +150,9 @@ class EDSModel(Model1D):
         if auto_background is True:
             self.add_polynomial_background()
         if auto_add_lines is True:
-            self.add_family_lines()
+            # Will raise an error if no elements are specified, so check:
+            if 'Sample.elements' in self.signal.metadata:
+                self.add_family_lines()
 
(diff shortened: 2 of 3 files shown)


vidartf:
For reference, the traceback is:

``` python
----> 1 m.append(hs.model.components1D.GaussianHF())

...\hyperspy\hyperspy\models\model1d.py in append(self, thing)
    251
    252     def append(self, thing):
--> 253         super(Model1D, self).append(thing)
    254         if self._plot_components:
    255             self._plot_component(thing)

...\hyperspy\hyperspy\model.py in append(self, thing)
    351                                          valid_variable_name=True), thing)
    352         if self._plot_active is True:
--> 353             self._connect_parameters2update_plot(components=[thing])
    354         self.update_plot()
    355

...\hyperspy\hyperspy\model.py in _connect_parameters2update_plot(self, components)
    483                     self._model_line.update, [])
    484         if self._plot_components is True:
--> 485             self._connect_component_lines()
    486
    487     def _disconnect_parameters2update_plot(self, components):

...\hyperspy\hyperspy\models\model1d.py in _connect_component_lines(self)
    589         for component in self:
    590             if component.active:
--> 591                 self._connect_component_line(component)
    592
    593     def _disconnect_component_lines(self):

...\hyperspy\hyperspy\models\model1d.py in _connect_component_line(component)
    574         if hasattr(component, "_model_plot_line"):
    575             f = component._model_plot_line.update
--> 576             component.events.active_changed.connect(f, [])
    577             for parameter in component.parameters:
    578                 parameter.events.value_changed.connect(f, [])

...\hyperspy\hyperspy\events.py in connect(self, function, kwargs)
    307         if function in self.connected:
    308             raise ValueError("Function %s already connected to %s." %
--> 309                              (function, self))
    310         if kwargs == 'auto':
    311             spec = inspect.signature(function)

ValueError: Function <bound method Signal1DLine.update of <hyperspy.drawing.signal1d.Signal1DLine ob
ject at 0x000000000E1FF400>> already connected to <hyperspy.events.Event: Event that triggers when t
he `Component.active` changes.: {<bound method Signal1DLine.update of <hyperspy.drawing.signal1d.Sig
nal1DLine object at 0x000000000E1F0C88>>, <bound method Signal1DLine.update of <hyperspy.drawing.sig
nal1d.Signal1DLine object at 0x000000000E1FF400>>}>.
```

PR #1274, opened 2016-08-26
Enhancement: PCA on complex numbers

thomasaarholt:
With the addition of support for complex numbers in Hyperspy, it would be nice to support decomposition of them as well. 
Application is for instance in Magnetic Resonance Spectroscopy (MRS).

As far as I can tell, scikit already supports complex numbers in PCA, and all that is required for Hyperspy is to add the dtypes to the data dtype check.

### DIFF
--- a/hyperspy/learn/mva.py
+++ b/hyperspy/learn/mva.py
@@ -167,7 +167,7 @@ class MVA():
         """
         to_return = None
         # Check if it is the wrong data type
-        if self.data.dtype.char not in ['e', 'f', 'd']:  # If not float
+        if self.data.dtype.char not in ['e', 'f', 'd', 'F', 'D', 'G']:  # If not float
             raise TypeError(
                 'To perform a decomposition the data must be of the float '
                 'type, but the current type is \'{}\'. '
--- a/hyperspy/learn/mva.py
+++ b/hyperspy/learn/mva.py
@@ -170,7 +170,7 @@ class MVA():
         if self.data.dtype.char not in ['e', 'f', 'd', 'g', 'F', 'D', 'G']:  # If not float
             raise TypeError(
                 'To perform a decomposition the data must be of the float '
-                'type, but the current type is \'{}\'. '
+                'or complex type, but the current type is \'{}\'. '
                 'To fix this issue, you can change the type using the '
                 'change_dtype method (e.g. s.change_dtype(\'float64\')) '
                 'and then repeat the decomposition.\n'
--- a/hyperspy/learn/mva.py
+++ b/hyperspy/learn/mva.py
@@ -167,7 +167,7 @@ class MVA():
         """
         to_return = None
         # Check if it is the wrong data type
-        if self.data.dtype.char not in ['e', 'f', 'd', 'F', 'D', 'G']:  # If not float
+        if self.data.dtype.char not in ['e', 'f', 'd', 'g', 'F', 'D', 'G']:  # If not float
             raise TypeError(
                 'To perform a decomposition the data must be of the float '
                 'type, but the current type is \'{}\'. '
--- a/hyperspy/tests/mva/test_decomposition.py
+++ b/hyperspy/tests/mva/test_decomposition.py
@@ -342,3 +342,39 @@ class TestLoadDecompositionResults:
             self.s.save(fname2)
             assert isinstance(
                 self.s.learning_results.decomposition_algorithm, str)
+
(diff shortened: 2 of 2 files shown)


thomasaarholt:
Despite the number of commits shown above, only one line of code has changed.

tjof2:
Definitely worth checking this e.g. on complex data with 0 imaginary values (for consistency with the usual real result) as well as on truly complex data.

thomasaarholt:
Yep, fully agree. There are probably several smart tests we can do. Decided to make the pull request whilst it was fresh in my mind and to see what others thought!

Thomas Aarholt
thomasaarholt@gmail.com

thomasaarholt:
So far it looks good. Using random numbers, I've tested keeping the imaginary part zero and looking at the residual. There is no difference ≈ `e-12`.
I've also tested keeping the real part the same but varying the imaginary part. I got a difference, as I would expect.

[Example here](http://nbviewer.jupyter.org/urls/dl.dropbox.com/s/1e8gqemvqxg2jb4/Testing%20complex%20array.ipynb).

Does anyone have any suggestions to tests?

tjof2:
**Possible tests**
1. The real and complex results give similar results (i.e. the residual) with imaginary parts set to 0.
2. Set up a low-rank matrix by creating **a = RandomComplex(m, r)** and **b = RandomComplex(n, r)**, then performing PCA on the result of **a.b^T** (i.e. an `m x n` matrix). The first **r** values of scree plot / singular values should be non-zero.

tjof2:
The second test is equally applicable to real-numbered data obviously.

tjof2:
@thomasaarholt any progress? Appreciate you might be busy (I suspect for the same reasons as me...!) but happy to help where necessary.

thomasaarholt:
I'm a bit stuck writing that t-word at the moment, but when I'm bored again I'll be back here. Won't take too long, I suspect.

hyperspy/tests/mva/test_decomposition.py
```diff
@@ -229,3 +229,39 @@ def test_decomposition_supported_return_false(self):
...
+        real_pca = self.s_real_dtype.get_decomposition_model(5)
+        np.testing.assert_almost_equal((complex_pca - real_pca).data.max(), 0.0)
+
+    def test_first_r_values_of_scree_non_zero(self):
+        """For low-rank matrix by creating a = RandomComplex(m, r) and
+        b = RandomComplex(n, r), then performing PCA on the result of a.b^T
+        (i.e. an m x n matrix).
+        The first r values of scree plot / singular values should be non-zero.
+        """
+        m, n, r = 100, 40, 10
```
tjof2: Purely for the purposes of smaller matrices and speed, this should work for e.g. `r=3` and `m,n=32 or 64` :-)

Personal preference of course.

hyperspy/tests/mva/test_decomposition.py
```diff
@@ -229,3 +229,39 @@ def test_decomposition_supported_return_false(self):
...
+
+        A = 100*(np.random.random((m,r)) + 1j* np.random.random((m,r)))
```
tjof2: Why multiply by 100?

thomasaarholt:
The 100* was just for my own sake testing to see the difference between very small numbers and numbers around 1. Removed it as it is pointless for computers. Also made the matrix smaller as you suggested.

Any other suggestions for this?

tjof2:
^no other comments!

hyperspy/learn/mva.py
```diff
@@ -160,7 +160,7 @@ def decomposition(self,
         """
         to_return = None
         # Check if it is the wrong data type
-        if self.data.dtype.char not in ['e', 'f', 'd']:  # If not float
+        if self.data.dtype.char not in ['e', 'f', 'd', 'F', 'D', 'G']:  # If not float
```
tjof2: Since:
```
import numpy as np
print(np.typecodes["AllFloat"])
>>> 'efdgFDG'
```
Shouldn't this also support `'g'`?

tjof2: @thomasaarholt  pretty sure if you make this change and update it this PR is complete, and only a few years late!

ericpre:
Can you also please solve the merge conflict or rebase?

thomasaarholt:
There, I rebased off `RELEASE_next_minor`. Should be good now.

thomasaarholt:
Woo! That one only took nearly four years!

PR #1282, opened 2016-09-04
fix #1238 

### LINKED ISSUE #1238: Metadata from digital micrograph files
Currently, when loading dm-files in HyperSpy much of the metadata is not included in s.metadata. However, there is a lot of metadata contained in the s.original_metadata. 

We could include things like microscope type, camera length, time of acquisition and pixel time into s.metadata.

ericpre:
Fix some bugs of #1238.
- Reading DM files saved with locale not set to en_US. Since we can't know, with which language locale settings the data have been, only a few locales are tested and can be parsed. If the locales are not recognised, the data are not parsed.
- Add support for reading date, time and dwell time for spectrum image.
- Add locale tests and tests files.

Not related to #1238:
I added an option (`debug_tags`) in the DM io plugin not to display (or not) the debug information related to parsing tags, because the log file was really too long and not readable!! I left a message to inform the developer to activate this option is necessary.


magnunor:
What is the current status of this pull request? The time-format-bug will be fixed in https://github.com/hyperspy/hyperspy/pull/1300

The locale tests would be very nice, but I'm not sure about the `debug_tags`.

Could you make a separate pull request with the locale tests? I'd like some thoughts from the other devs about the `debug_tags`.

ericpre:
Closing because of #1300, #1311 and #1312!

PR #1283, opened 2016-09-05
Fix bug with parsing time when reading blockfile files

ericpre:
Parse the time in a "naive" manner, since we can't know the correct time zone. The tests were failing if run with different time zones.

- [x] Add support for time_zone
- [x] Add time_zone to metadata specification (can be time zone abbreviation or UTC time offset),
- [x] Add functions to get/update date and time to metadata,
- [x] Add function to convert excel serial time to ISO or `datetime.datetime`
- [x] Add tests.


ericpre:
@vidartf, could you please have a look at this? If it makes sense to you? Thanks!

vidartf:
As far as I read this, the previous behavior/assumptions are as follows:
- The date is stored as an "Excel timestamp", which is the (fractional) number of days since January 1, 1900 (UTC).
- Originally the full `datetime` object was returned (converted to the timezone where the file is being opened).
- I see the code for the fractional second part is wrong (it should be `(secs % 1.0) * 1e6`). My mistake :)

Next, in 92b38662c61c28841d4f6bf1485b21023f566a3e the return value was split into a date and a time field, which consequently caused it to be in a non-timezone aware format. I don't have the overview of the change that commit was part of, so I cannot say why this was wanted/necessary, but I assume it was to make it similar to other signal data.

In this proposal, it seems the change is to instead output the date and time in implicit UTC (that is, UTC without any explicit timezone info).

Let's say the original file was acquired in UTC-3, and then read out in UTC+5:45, this would give the following three behaviors:
- The original behavior will display the acquisition time in the local timezone, but should still be directly comparable to the original timestamp due to its time-zone aware format (9:00-3:00 == 17:45+5:45).
- The current behavior will correctly output the local date and time of the acquisition (local acquisition datetime + 8:45, split), but will fail in certain comparison operations due to its lack of timezone information, for example to a hardcoded time string in a test. This is the problem that this PR wants to fix.
- The proposed solution will be internally consistent, in that it will define its output as UTC (local acquisition datetime +3:00, split). Direct comparison to other date/time objects would, however, require regularization of timezones.

I see three possible solutions/improvements (pick one):
- Revert to full `datetime` object, and use full objects everywhere else in HyperSpy (including tests).
- Use the proposed solution, and use UTC date/time everywhere else in HyperSpy.
- Use either the existing behavior OR the proposed solution (that is, return as local time or UTC), but add time zone information together with the returned value (to be stored separately in the metadata). Use the same pattern everywhere else in HyperSpy.

Or summarized: Whatever is OK, as long as the same standard is used everywhere in HyperSpy. I'm a proponent of full datetime objects since they have the best tools for regularization of data (that is, it only needs to care about time zones in the I/O layer).

ericpre:
Thanks!
FYI #1136, #1232 and #1222.
In short, we went to store the date and time as string (ISO 8601) in the metadata and to handle the data with `numpy.datetime64` when necessary.
Based on your comments (if I understood correctly!), the most simpler and consistent with HyperSpy would possibly be to add the timezone as metadata.

This may need a bit more of thinking/discussion!!

vidartf:
You could also add timezone info (e.g. explicit UTC) to the time (https://docs.python.org/3.6/library/datetime.html#datetime.time.isoformat), but for comparing times in different time zones you would have to combine the time with the date field.

ericpre:
To fix this bug, I added the time zone to the metadata and restored the similar behavior than the initial one! ;) The time is converted to local time when loading, mainly because the time zone of the acquisition can't be known and this sounds like the most sensible way to deal with it...

This is ready to review.

hyperspy/io_plugins/blockfile.py
```diff
@@ -17,14 +17,14 @@
 # along with  HyperSpy.  If not, see <http://www.gnu.org/licenses/>.
 
 import os
-from datetime import datetime, timedelta
-from dateutil import tz
 from traits.api import Undefined
 import numpy as np
 import logging
 import warnings
+import datetime, dateutil
```
vidartf: I'd keep these as two lines to highlight that they are separate imports.

hyperspy/misc/date_time_tools.py
```diff
@@ -0,0 +1,52 @@
...
+
+import datetime
+from dateutil import tz, parser
+
+
+def metadata_to_datetime(metadata):
+    date = metadata['General']['date']
+    time = metadata['General']['time']
+    if 'time_zone' in metadata['General']:
+        return parser.parse('%sT%s %s' % (date, time, metadata['General']['time_zone']))
```
vidartf: Note that the ISO standard does not include a space between time and time zone info: https://www.w3.org/TR/NOTE-datetime. However, as long as the parser is happy it shouldn't matter too much.

hyperspy/misc/date_time_tools.py
```diff
@@ -0,0 +1,52 @@
...
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with  HyperSpy.  If not, see <http://www.gnu.org/licenses/>.
+
+import datetime
+from dateutil import tz, parser
+
+
+def metadata_to_datetime(metadata):
```
vidartf: There is a bit of asymmetry in these functions now. Maybe make this `ISO_format_to_datetime()`, with an optional time zone field? Then this could be used in `ISO_format_to_serial_date()` below. If not, then why not have a `datetime_to_metadata()` function?

vidartf (commented):
Looks correct to me, but I've added some comments on form/style. Other than that, the other I/O plugins should probably also be updated to use the `time_zone` metadata field?

I'd still like to add that I think it is strange to store date/time info as string representations. It feels like storing numerical data as strings :P

ericpre:
@vidartf, thanks for the quick review!
I agree it's a bit silly to store the date and time as string... but for most of users, it's more easy to read and it can be converted easily to np.datetime64, which support nanosecond precision (if necessary). See #1232 for discussion.

At some point I will submit a PR to try to go through other plugin to see when it's possible to import the time zone.

ericpre:
@vidartf, I tried to address your comments and to add some functions to make things not to messy... Please let me know if you have further comments!

hyperspy/misc/date_time_tools.py
```diff
@@ -58,77 +50,125 @@ def get_date_time_from_metadata(metadata, formatting='ISO'):
...
 
-    if 'time_zone' in metadata['General']:
-        dt = parser.parse('%sT%s%s' % (date, time, metadata['General']['time_zone']))
-    else:
+    """
+    time = date = time_zone = None
+    if metadata.has_item('General.date'):
+        date = metadata['General']['date']
+    if metadata.has_item('General.time'):
+        time = metadata['General']['time']
```
vidartf: I just submitted a PR #1370 which can allow these statements to be written 
```python
date = metadata.get_item('General.date')
time = metadata('General.time')
... <code for time_zone>
```

hyperspy/misc/date_time_tools.py
```diff
@@ -58,77 +50,125 @@ def get_date_time_from_metadata(metadata, formatting='ISO'):
...
+        >>> s.metadata
+            ├── General
+            │   ├── date = 2016-12-12
+            │   ├── original_filename = example1.msa
+            │   ├── time = 12:12:12
+            │   ├── time_zone = 'EST'
+            │   └── title = NIO EELS OK SHELL
     """
     time_zone = None
     if isinstance(dt, str):
```
vidartf, line 118: If passed an ISO string, I would simply parse it to a datetime (`dt = parser.parse(dt)`), and then use the code for datetime objects below.

hyperspy/misc/date_time_tools.py
```diff
@@ -58,77 +50,125 @@ def get_date_time_from_metadata(metadata, formatting='ISO'):
...
 def serial_date_to_ISO_format(serial):
+    """
+    Convert serial_date to a tuple of string (date, time, time_zone) in ISO 
+    format. By default, the serial date is converted in local time zone.
+    """
     # Excel date&time format
     origin = datetime.datetime(1899, 12, 30, tzinfo=tz.tzutc())
-    secs = (serial % 1.0) * 86400.0
+    secs = round(serial % 1.0 * 86400)
     delta = datetime.timedelta(int(serial), secs, secs / 1E6)
```
vidartf: Actually, secs shouldn't be rounded, and `secs/1E6` should be removed, as the docs state that 
> If any argument is a float and there are fractional microseconds, the fractional microseconds left over from all arguments are combined and their sum is rounded to the nearest microsecond using round-half-to-even tiebreaker.

My mistake, on not getting this earlier.

hyperspy/misc/date_time_tools.py
```diff
@@ -19,23 +19,15 @@
 import numpy as np
 import datetime
 from dateutil import tz, parser
+import pytz
```
vidartf: What is the motivation for switching to pytz?

vidartf:
Some of my review points above are listed as "outdated" without actually being so. Please consider them as well :)

francisco-dlp:
Is there any chances of getting this into v1.1.2?

vidartf:
Yes, the outstanding points are all minor. If @ericpre is busy I can send a PR his way (or resubmit with an added commit).

hyperspy/misc/date_time_tools.py
ericpre: I used pytz for convenience, dateutil.timezone looked to be quite limited. Yeah, adding one more library is not great and I guess it should be possible to do the same with dateutil.
I will not have time to do it in the next few days, so please go ahead if like. Otherwise I can also do it later.

hyperspy/misc/date_time_tools.py
ericpre: Actually, the precision of serial time is second, so I change accordingly.

francisco-dlp:
@vidartf, is this ready to merge now?

hyperspy/misc/date_time_tools.py
vidartf: I cannot find any references for this, but actually references that implies the opposite.

vidartf:
I'll add a commit locally and submit as PR to @ericpre , or if we are in a hurry I can resubmit.

hyperspy/misc/date_time_tools.py
ericpre: I couldn't find any documentation on serial time really... so I inferred from what I found on the web! Please go ahead to change it.

vidartf:
ericpre/hyperspy#3

hyperspy/tests/misc/test_date_time_tools.py
```diff
@@ -0,0 +1,158 @@
...
+    # in case of iso, the exact time is lost, only the time offset is kept
+    md11 = dtt.update_date_time_in_metadata(iso1, md.deepcopy())
+    assert_deep_almost_equal(md11.General.date, md1.General.date)
+    assert_deep_almost_equal(md11.General.time, md1.General.time)
+    assert_deep_almost_equal(md11.General.time_zone, 'UTC')
+
+    md12 = dtt.update_date_time_in_metadata(dt1, md.deepcopy())
+    assert_deep_almost_equal(md12.General.date, md1.General.date)
+    assert_deep_almost_equal(md12.General.time, md1.General.time)
+    assert_deep_almost_equal(md12.General.time_zone, 'Coordinated Universal Time')
```
vidartf: I guess this should be something like (name varies across platforms?):
```python
nt.assert_in(md12.General.time_zone, ('UTC', 'Coordinated Universal Time'))
```

ericpre: Yes!

francisco-dlp:
@vidartf, is it ready now?

vidartf:
All green, and LGTM!

PR #1286, opened 2016-09-07
Old image plotting bugs with navigator

to266:
Enable passing image plotting kwargs to navigator as well. 

Also not crash when navigating without a colorbar


vidartf:
LGTM. At first I was a bit concerned that it would be some major issues about signal and navigator plot sharing kwargs, but the list of kwargs used are only ones that makes sense to share between the two. My only question then was if there are not any similar kwargs for 1D navigators that would be relevant to use. I guess not as the HSE's `plot_signal()` doesn't take any kwargs either.

vidartf:
I assume it was a conscious decision not to pass the kwargs to `imf.plot()`?

to266:
for `imf.plot()`: probably just wanted to change as little as possible. Currently the method does not support kwargs for _our_ plotting, and instead just passes them to `imshow`, thus would have required quite a bit more tingering for identical effect.

I will probably soon (today-tomorrow) change this PR to pass the kwargs with a dict akin to `nav_kwargs`, so that signal and navigation kwargs do not mix.

to266:
Ok updated this as promissed / suggested, user_guide included. A quick look might be needed though

vidartf:
So, if I understand correctly:
 - Normal kwargs to `plot()` go to the signal plotting (both our kwargs and imshow args), except:
 - Navigator arguments, which are wrapped in a dict `navigator_kwds`.
 - No arguments are shared between the two.

What about the case of signal (1|1) ? Then the navigator plot is often an image (which includes the signal axes). This uses HSE. Maybe HSE should also get support for kwargs then (probably still wrapped in `navigator_kwds`).

to266:
Actually in the (1|1) case `navigator_kwds` still get passed to the navigator correctly. In essence if there is a navigator image plotted, the dict will be passed. Otherwise it's just ignored.

Adding _any_ kwargs to HSE may be nice, but not straightforward - it's not plotted all at once, but the lines are added individually. Then should the kwargs modify the figure/axes, or one/all lines? Any way I'm a bit too busy to add this, but others are welcome to PR here or separately with this.

vidartf:
Right, the navigation keywords still go through HE. I'd merge, but am currently travelling.

francisco-dlp:
Good point, don't merge while driving. LGTM too.

PR #1236, opened 2016-08-08
Add sub-pixel accuracy to 2D alignment

vidartf:
Resubmission of #708.
### TODO
- [ ] Unittests
- [ ] Cross-correlation vs phase-correlation

PR #1290, opened 2016-09-12
Add sub-pixel accuracy to 2D alignment

to266:
resubmission of https://github.com/hyperspy/hyperspy/pull/1236 in order to just quicken things up.

With a test and supports both cross- and phase-correlations.


dnjohnstone:
Had a quick play with this and it seems ok - although it needs merging up with the current RELEASE_next_minor

hyperspy/tests/signal/test_2D_tools.py
```diff
@@ -58,6 +58,12 @@ def test_align_subpix(self):
         # Compare by broadcasting
         np.testing.assert_allclose(s.data[4], s.data[0], rtol=1)
 
+    def test_estimate_subpix(self):
```
francisco-dlp: The test for the case when ``normalize_corr`` is ``True`` is missing.

francisco-dlp:
Other than the comment above, it looks good to me too.

dnjohnstone:
Could we get this in to v1.2 please? Looks basically done!

francisco-dlp:
It needs a test for phase correlation. Anybody is welcome to contribute it.

to266:
I think when I tried to add the missing test I was not happy with the result - namely even with fine sub-sampling the results were not that much better and not necessarily converged towards the true (known) solution. 

In the end I attempted to debug what the heck was happening, but could not find fault in the actual process. I think it does the correct thing and without mistakes, it's just that the result is not what we'd expect

francisco-dlp:
What I meant is that [this line](https://coveralls.io/builds/9686196/source?filename=hyperspy%2F_signals%2Fsignal2d.py#L101) is not tested.

francisco-dlp:
LGTM, but it requires merging Rnm in.

PR #1292, opened 2016-09-14
2D peak finders

bm424:
Re-opening this pull request as a new feature. Several peak finders have been implemented, as well as an interactive tool for the Jupyter notebook.


doc/user_guide/signal2d.rst
```diff
@@ -32,3 +32,91 @@ method. The parameters `ramp_x` and `ramp_y` dictate the slope of the ramp in `x
 while the offset is determined by the `offset` parameter. The fulcrum of the linear ramp is at the origin
 and the slopes are given in units of the axis with the according scale taken into account.
 Both are available via the :py:class:`~.axes.AxesManager` of the signal.
+
+Peak finding
+------------
+
+.. versionadded:: 1.0.0
```
tjof2: Right version? :-)

doc/user_guide/signal2d.rst
```diff
@@ -32,3 +32,91 @@ method. The parameters `ramp_x` and `ramp_y` dictate the slope of the ramp in `x
...
+
+Matrix based peak finding
+^^^^^^^^^^^^^^^^^^^^^^^^^
+
+.. code-block:: python
+
+    >>> s.find_peaks2D(method='laplacian_of_gaussians')
+    >>> s.find_peaks2D(method='difference_of_gaussians')
+
+These methods are essentially wrappers around the
```
tjof2, line 80: why "essentially wrappers"? What makes them different to plain "wrappers"? Is there additional functionality?

doc/user_guide/signal2d.rst
```diff
@@ -32,3 +32,91 @@ method. The parameters `ramp_x` and `ramp_y` dictate the slope of the ramp in `x
...
+
+Interactive Parametrization
+---------------------------
+
+.. code-block:: python
+
+    >>> s.find_peaks2D_interactive()
+
+Many of the peak finding algorithms implemented here have a number of
+tuneable parameters that significantly affect their accuracy. Finding the
```
tjof2: and speed, presumably?

hyperspy/_signals/signal2d.py
```diff
@@ -605,3 +608,433 @@ def add_ramp(self, ramp_x, ramp_y, offset=0):
...
+                peaks[indices] = find_peaks_stat(z, **kwargs)
+            elif method == 'laplacian_of_gaussians':
+                peaks[indices] = find_peaks_log(z, **kwargs)
+            elif method == 'difference_of_gaussians':
+                peaks[indices] = find_peaks_dog(z, **kwargs)
+            else:
+                raise NotImplementedError("The method `{}` is not implemented. "
+                                          "See documentation for available "
+                                          "implementations.".format(method))
+
```
to266: A better approach, I feel, would be to do something like so:

``` python
method_dict = {'skimage': peal_local_max,
               'max': find_peaks_max,
               'minmax': find_peaks_minmax,
               'zaefferer': find_peaks_zaefferer,
               'stat': find_peaks_stat,
               'laplacian_of_gaussians': find_peaks_log,
               'difference_of_gaussians': find_peaks_dog,
              }

if method in method_dict:
    method = method_dict[method]
else:
    raise NotImplementedError("The method `{}` is not implemented. " 
                              "See documentation for available "
                              "implementations.".format(method))
for z, indices in zip(self._iterate_signal(),
                      self.axes_manager._array_indices_generator()):
    peaks[indices] = method(z, *args, **kwargs)
```

[do check that I did not mistype this...]

I feel this makes it easier to add any new methods, and at the same time separates option-checking from the actual computation.

hyperspy/_signals/signal2d.py
```diff
@@ -605,3 +608,433 @@ def add_ramp(self, ramp_x, ramp_y, offset=0):
...
+def find_peaks_minmax(z, separation=5., threshold=10., interpolation_order=3):
+    """
+    Method to locate the positive peaks in an image by comparing maximum
+    and minimum filtered images.
+    Parameters
+    ----------
+    z: ndarray
+    separation: expected distance between peaks
+    threshold: ???
```
to266: `???` ?

hyperspy/_signals/signal2d.py
```diff
@@ -605,3 +608,433 @@ def add_ramp(self, ramp_x, ramp_y, offset=0):
...
+                                                 `scikit-image` which uses
+                                                 the difference of Gaussian
+                                                 matrices approach.
+        *args : associated with above methods
+        **kwargs : associated with above methods.
+        Returns
+        -------
+        peaks: structured array of shape _navigation_shape_in_array in which
+               each cell contains an array with dimensions (npeaks, 2) that
+               contains the x, y coordinates of peaks found in each image.
```
to266: Are the coordinates in indices? Looking at the methods I assume so, however from the description I would guess physical / calibrated coordinates.

Also, subpixel resolution or not?

hyperspy/_signals/signal2d.py
```diff
@@ -605,3 +608,433 @@ def add_ramp(self, ramp_x, ramp_y, offset=0):
...
+    peaks = np.array(
+        ndi.center_of_mass(z, labeled, range(1, num_objects + 1)))
+
+    return clean_peaks(peaks)
+
+
+def find_peaks_max(z, alpha=3., size=10):
+    """
+    Method to locate positive peaks in an image by simple local maximum
+    searching.
```
to266: need parameter descriptions

hyperspy/_signals/signal2d.py
```diff
@@ -605,3 +608,433 @@ def add_ramp(self, ramp_x, ramp_y, offset=0):
...
+    """
+    # preallocate lots of peak storage
+    k_arr = np.zeros((10000, 2))
```
to266: No need to use a numpy array. Just initialize as an empty list and append when found, create the array at the end.

hyperspy/_signals/signal2d.py
```diff
@@ -605,3 +608,433 @@ def add_ramp(self, ramp_x, ramp_y, offset=0):
...
+        """Produces a list of coordinates in the box about (x, y)."""
+        a = int(window_size / 2)
+        x_min = max(0, x - a)
+        x_max = min(x_max, x + a)
+        y_min = max(0, y - a)
+        y_max = min(y_max, y + a)
+        return np.array(
+            np.meshgrid(range(x_min, x_max), range(y_min, y_max))).reshape(
+            2,
+            -1).T
```
to266: using `itertools.product` in my simple test is around 20 times faster (albeit it returns a list of tuples):

``` python
from itertools import product
return [a for a in product(range(y_min, y_max), range(x_min, x_max))]
```

(i swapped X and Y order purely to exactly duplicate the order you had previously - `product` iterates the last iterators first )

I'll have a look further in the code, but maybe returning the iterator itself (or maybe even `yield`-ing) may be an even better approach

EDIT: Ok, your approach seems faster for larger `window_size`s - up to you if you change this

to266: Actually you never seem to actually require the box itself, except to run `get_max` - I suggest combining the two.

hyperspy/_signals/signal2d.py
```diff
@@ -605,3 +608,433 @@ def add_ramp(self, ramp_x, ramp_y, offset=0):
...
+    # Boolean matrix of high-gradient points.
+    gradient_is_above_threshold = image_gradient >= grad_threshold
+    peaks = []
+    for coordinate in coordinates[gradient_is_above_threshold.flatten()]:
+        # Iterate over coordinates where the gradient is high enough.
+        b = box(coordinate[0], coordinate[1], window_size, z.shape[0],
+                z.shape[1])
+        p_old = np.array([0, 0])
+        p_new = get_max(z, b)
+        while np.all(p_old != p_new):
```
to266, line 900: This will continue searching for the peak even when we are too far away to save it later. I suggest combining the `while` loop with the `if` statement that follows it (likely using `break`) to stop the loop once it wanders too far from the start

hyperspy/_signals/signal2d.py
```diff
@@ -605,3 +608,433 @@ def add_ramp(self, ramp_x, ramp_y, offset=0):
...
+    This version by Ben Martineau (2016), with minor modifications to the
+    original where methods were ambiguous or unclear.
+    """
+    from scipy.ndimage.filters import generic_filter
+    from scipy.ndimage.filters import uniform_filter
+    from sklearn.cluster import DBSCAN
+
+    def normalize(image):
+        """Scales the image to intensities between 0 and 1."""
+        return image / np.max(image)
```
to266, line 947: this scales between `image.min() / image.max()` and 1.

To scale between 0 and 1 it should read
`(image - image.min()) / (image.max() - image.min())`

hyperspy/_signals/signal2d.py
```diff
@@ -605,3 +608,433 @@ def add_ramp(self, ramp_x, ramp_y, offset=0):
...
+
+    def _local_stat(image, radius, func):
+        """Calculates rolling method 'func' over a circular kernel."""
+        x, y = np.ogrid[-radius:radius + 1, -radius:radius + 1]
+        kernel = x ** 2 + y ** 2 <= radius ** 2
```
to266: Could use `np.hypot` here - not sure if better (I would generally assume they do something smart, having included this)

hyperspy/_signals/signal2d.py
```diff
@@ -605,3 +608,433 @@ def add_ramp(self, ramp_x, ramp_y, offset=0):
...
+
+    def stat_binarise(image):
+        """Peaks more than one standard deviation from the mean set to one."""
+        image_rolling_mean = local_mean(image, 10)
+        image_rolling_std = local_std(image, 10)
+        image = single_pixel_desensitize(image)
+        binarised_image = np.zeros(image.shape)
+        stat_mask = image > (
+            image_rolling_mean + standard_deviation * image_rolling_std)
+        binarised_image[stat_mask] = 1
```
to266: `binarised_image = np.where(stat_mask, 1, 0)` (or `True` and `False`)

Also, is it reasonable for the user to pick something different than `10`? What if the image/peaks are very large/small?

hyperspy/_signals/signal2d.py
```diff
@@ -605,3 +608,433 @@ def add_ramp(self, ramp_x, ramp_y, offset=0):
...
+        """Image convolved twice using a uniform 3x3 kernel."""
+        image = uniform_filter(image, size=3)
+        image = uniform_filter(image, size=3)
+        return image
+
+    def half_binarise(image):
+        """Image binarised about values of one-half intensity."""
+        binarised_image = np.zeros(image.shape)
+        binarised_image[image > 0.5] = 1
+        return binarised_image
```
to266: again, `np.where` is your friend here - can replace the function in 1 simple line..

hyperspy/_signals/signal2d.py
```diff
@@ -605,3 +608,433 @@ def add_ramp(self, ramp_x, ramp_y, offset=0):
...
+
+    def separate_peaks(binarised_image):
+        """Identify adjacent 'on' coordinates via DBSCAN."""
+        bi = binarised_image.astype('bool')
+        coordinates = np.indices(bi.data.shape).reshape(2, -1).T[
```
to266: Is the `data` call here necessary? Seems dangerous to me

hyperspy/_signals/signal2d.py
```diff
@@ -605,3 +608,433 @@ def add_ramp(self, ramp_x, ramp_y, offset=0):
...
+
+    def _peak_find_once(image):
+        """Smooth, binarise, and find peaks according to main algorithm."""
+        image = smooth(image)
+        image = half_binarise(image)
+        peaks = separate_peaks(image)
+        return image, peaks
+
+    def stat_peak_finder(image):
+        """Find peaks in diffraction image. Algorithm stages in comments."""
```
to266: `diffraction`?

hyperspy/_signals/signal2d.py
```diff
@@ -605,3 +608,433 @@ def add_ramp(self, ramp_x, ramp_y, offset=0):
...
+        image = normalize(image)  # 1
+        image = stat_binarise(image)  # 2, 3
+        n_peaks = np.infty  # Initial number of peaks
+        image, peaks = _peak_find_once(image)  # 4-6
+        m_peaks = len(peaks)  # Actual number of peaks
+        while (n_peaks - m_peaks) / n_peaks > 0.05:  # 8
```
to266: Why `0.05`? Is it possible that a user might want a different value?

hyperspy/_signals/signal2d.py
```diff
@@ -605,3 +608,433 @@ def add_ramp(self, ramp_x, ramp_y, offset=0):
...
+        Array of peak coordinates.
+
+    Notes
+    -----
+    While highly effective at finding even very faint peaks, this method is
+        sensitive to fluctuations in intensity near the edges of the image.
+
+    """
+    from skimage.feature import blob_dog
+    z = z / np.max(z)
```
to266, line 1056: The normalization here again, I guess

hyperspy/_signals/signal2d.py
```diff
@@ -605,3 +608,433 @@ def add_ramp(self, ramp_x, ramp_y, offset=0):
...
+                     overlap=overlap)
+    try:
+        centers = blobs[:, :2]
+    except IndexError:
+        return NO_PEAKS
+    clean_centers = []
+    for center in centers:
+        if len(np.intersect1d(center, (0,) + z.shape + tuple(
+                        c - 1 for c in z.shape))) > 0:
+            continue
```
to266: If I understand this correctly, here you check that the found centers are not at the edges of the image + not at the [far edges - 1]. Why is that? Also, why then not check that the centers are at (1,) coordinates (equivalent of (c-1))?

hyperspy/gui/peakfinder2D.py
```diff
@@ -0,0 +1,288 @@
...
+
+    def __init__(self):
+        self.signal = None
+        self.indices = None
+        self.methods = METHODS
+        self.method_names = [method.__name__ for method in METHODS]
+        self.params = {method.__name__: OrderedDict(
+            [(p.name, p.default) for p in
+             inspect.signature(method).parameters.values() if
+             p.default is not inspect._empty]) for method in METHODS}
```
to266, line 23: Could replace the list by a generator (i.e. square brackets by normal ones)

hyperspy/gui/peakfinder2D.py
```diff
@@ -0,0 +1,288 @@
...
+
+    def init_ui(self):
+        raise NotImplementedError
+
+    @property
+    def current_method(self):
+        return dict(zip(self.method_names, self.methods))[self._method]
+
+    def get_data(self):
+        return self.signal.inav[self.indices].data
```
to266: This also recreates the full signal, copying many things that are not necessary in the end - I suggest replacing it by the following (feel free to combine this to a one-liner, just splitting for clarity):

``` python
_slices = self.signal._get_array_slices(self.indices, isNavigation=True)
return self.signal.data[_slices]
```

hyperspy/gui/peakfinder2D.py
```diff
@@ -0,0 +1,288 @@
...
+    def create_choices_widget(self):
+        from ipywidgets import Dropdown
+        dropdown = Dropdown(
+            options=list(self.method_names),
+            value=self._method,
+            description="Method",
+        )
+
+        def on_method_change(change):
+            self._method = dropdown.value
```
to266, line 75: If I remember correctly here you want `change.new` or something along those lines, right? Currently you don't actually read/use `change`. You could also perform a sanity check here (i.e. validate the new value)

doc/user_guide/signal2d.rst
bm424: There's a little bit of pre- and post-processing that goes on, it's not a direct wrapper I'd say.

hyperspy/_signals/signal2d.py
bm424: Nah that's far too sensible.

hyperspy/_signals/signal2d.py
bm424: I discussed this one with @dnjohnstone - I think it should stay a multiplicative scaling, in most cases we want to preserve relative intensity. Being pedantic, I see the point that this doesn't necessarily scale any image to between 0 and 1. I could change the docstring...

hyperspy/_signals/signal2d.py
bm424: There's only a couple of tenths of a microsecond difference, but it's going in!

hyperspy/_signals/signal2d.py
bm424: `np.where` isn't working here for some reason - I can't see why but it causes errors down the line in the DBSCAN calculation.

hyperspy/_signals/signal2d.py
bm424: _Sigh_

hyperspy/_signals/signal2d.py
bm424: Just turns out to be a slight problem with the `blob_dog` method that it finds an unusual number of large peaks right at the edge of the image, something to do with the discontinuity I think. Fair point about the `(0, 1)`

hyperspy/gui/peakfinder2D.py
bm424: Any advantage?

hyperspy/gui/peakfinder2D.py
bm424: The `ipywidgets` interface complains if it doesn't get something to pass a value into, even when that value is not actually used. I know it doesn't get used, but I can't think of any way to get rid of it.

bm424:
Thanks for all the feedback guys :) Implemented a lot of the changes, left some things the way they are with explanations in the other comments.

doc/user_guide/signal2d.rst
tjof2: It might be good to make that clearer then, in case any pre- or post-processing causes differences in the results.

hyperspy/gui/peakfinder2D.py
to266: You don't create a list when you don't need one!

tjof2:
What's the latest on this?

dnjohnstone:
I've used this for a while now and for me it seems broadly good with two exceptions:

1. I think that the peak finders should be returned in calibrated value units instead of array indices.

2. Some default values for peak finding parameters are set as integers and therefore the notebook interaction won't let you make them a float. It would be good if all defaults were floats so that this doesn't happen.

Other than this I think it's pretty much ready to go.

@francisco-dlp did you have any insight regarding testing interactive parts?

francisco-dlp:
I don't know how to test the widgets. @to266 simply called the function hoping that any major failure would raise an exception but in #1375 we had to skip this simple test when not running from a IP kernel. Maybe it's worth asking this question to the ipywidgets developers?

vidartf:
I'm going [here](http://opendreamkit.org/meetings/2017-01-16-ICMS/) next week, so I could ask around then, if you want :)

doc/user_guide/signal2d.rst
dnjohnstone: I'm working through tidying up some bits of this branch - I will change the wording to highlight what has been added.

hyperspy/_signals/signal2d.py
dnjohnstone: @to266 is this fixed now with the if and break below? I'm just trying to unpick what was changed and when...

vidartf:
There are two decent alternatives (maybe use both?):
 - There is a direct solution for the issue in #1375 currently in IPywidgets beta (https://github.com/ipython/ipywidgets/blob/7b6d129e3967c73f74cbbb90b8b3ce58b51c486a/ipywidgets/widgets/tests/__init__.py#L19-L25).
 - We could add these calls to a notebook that can then be tested by [nbval](https://github.com/computationalmodelling/nbval) by py.test. This should probably be used for demo notebooks anyway.

francisco-dlp:
@bm424, could you merge the current RELEASE_next_minor here?

@to266, is this ready?

to266:
Seems to me that all my previous points were addressed. If nothing else significant was added since then, then I guess so.

francisco-dlp:
@bm424, would you like to merge RELEASE_next_minor into this so that we can merge it?

hyperspy/_signals/signal2d.py
```diff
@@ -650,6 +653,457 @@ def add_ramp(self, ramp_x, ramp_y, offset=0):
...
+            'laplacian_of_gaussians':  find_peaks_log,
+            'difference_of_gaussians': find_peaks_dog,
+        }
+        if method in method_dict:
+            method = method_dict[method]
+        else:
+            raise NotImplementedError("The method `{}` is not implemented. "
+                                      "See documentation for available "
+                                      "implementations.".format(method))
+        for z, indices in zip(self._iterate_signal(),
```
to266, line 711: Rewrite this in terms of `map`? Also, the `*args` should probably go - all methods have only `kwargs` (except the first data argument). Also, the usual (`show_progressbar`, `parallel`) kwargs should be added here and passed:

```python
peaks = self.map(method,
                 inplace=False, 
                 ragged=False, #I think?
                 show_progressbar=show_progressbar, 
                 parallel=parallel, 
                 **kwargs)
```

to266 (changes requested):
~~I think only implementing it using `map` is left~~

All tests are written for `nose` and should be refactored to use `pytest` as well

francisco-dlp:
@dnjohnstone, weren't you working on reimplementing this using ``map``?

bm424:
Yeah, I've noticed the new tests, I haven't used PyTest before so it'll take a little while for me to work out how to make the changes. As for the map function if @dnjohnstone isn't working on them I'll get that done next.

dnjohnstone:
#1497 !!!!

bm424:
Ah.

dnjohnstone:
I'm working on the tests right now with @to266 and @francisco-dlp - give it a few mins

ericpre:
Closing in favour of #1497.

PR #1295, opened 2016-09-16
Better map for signals

to266:
- Separates `map` into two functions that can later be replaced for lazy signals
- As a bonus, running the `map` in parallel (threaded, not multi-process manner) was trivial to add.
- Since I'm making things parallel, now `model.as_signal` can be run in threads as well
## TODO:
- [x] threaded tests to ensure identical behaviour
- [x] threaded tests for `as_signal`


to266:
Ready for review

hyperspy/model.py
```diff
@@ -410,6 +410,14 @@ def as_signal(self, component_list=None, out_of_range_to_nan=True,
         show_progressbar : None or bool
             If True, display a progress bar. If None the default is set in
             `preferences`.
+        out : {None, BaseSignal}
+            The signal where to put the result into. Convenient for parallel
+            processing. If None (default), creates a new one. If passed, it is
+            assumed to be of correct shape and dtype and not checked.
+        threaded : bool, int
```
francisco-dlp: Why not `parallel` or something of this sort that will be more telling to non-programmers?

hyperspy/signal.py
```diff
@@ -3217,6 +3218,9 @@ def map(self, function,
         show_progressbar : None or bool
             If True, display a progress bar. If None the default is set in
             `preferences`.
+        threaded : bool
```
francisco-dlp: As above

francisco-dlp (changes requested):
Apart from the minor component this looks very good to me, except that

* [ ] The new feature needs to be documented in the User Guide
* [ ] All calls to `map` in the rest of the code should have the extra `threaded` keyword.

By the way, would it make sense to add `threaded` to `preferences.General`?

to266:
Added as requested

francisco-dlp:
All looks good to me. Could you update your branch with RELEASE_next_patch to see if Appveyor and travis are happy too?

to266:
I just did!

francisco-dlp:
Tests are failing...

to266:
Yep, though I can't reproduce this locally. I'll hopefully figure it out on the next "break"

francisco-dlp:
After upgrading anaconda (`conda update anaconda`) I am able to reproduce the issue locally.

to266:
Seems to have worked. At the same time added notebook progressbar to be used if possible (have been using this for a long time, just never PRed..)

francisco-dlp (approved):

PR #1304, opened 2016-10-12
Non Integer Binning

k8macarthur:
Method to Bin EDS Spectrum by a non-integer number of pixels with linear interpolation and preservation of the original raw counts.


k8macarthur:
So this is fully tested for EDS, would need to be copied across for EELS I guess but I put it in the folder with the other rebin function. I have included a function in the array section too but not tested. I also wasn't sure where the testing functions for this should go. 

@thomasaarholt Happy to extend to 1D signals but maybe you could help with how to do that? Or just include your function?

magnunor:
https://github.com/hyperspy/hyperspy/pull/1305

(So people know why this was closed)

thomasaarholt:
Cheers Magnus.

PR #1305, opened 2016-10-12
Linear_Bin: function to bin spectrum by a non-integer number of pixels in each dimension.

k8macarthur:
Method to Bin EDS Spectrum by a non-integer number of pixels with linear interpolation and preservation of the original raw counts.


k8macarthur:
Try again with changes actually included now...

k8macarthur:
So this is fully tested for EDS, would need to be copied across for EELS I guess but I put it in the folder with the other rebin function. I have included a function in the array section too but not tested. I also wasn't sure where the testing functions for this should go.

@thomasaarholt Happy to extend to 1D signals but maybe you could help with how to do that? Or just include your function?

thomasaarholt:
Does this work for Signal2Ds as well as Signal1Ds? EDS is a Signal1D due to its spectrum signal nature.

thomasaarholt:
And well  done!

thomasaarholt:
Linear_bin takes a "scale" argument, but in the doctext it describes a "step" argument. 

Also, could you include a small example in the doctext?

k8macarthur:
Ok fixed the error in the doc_string and added an example.

In theory it could work Signal2Ds the function just extracts and plays with the np.array before putting it back into a Spectrum format (with amended meta data) and as there is no specification on the number of axis in the array. Although I haven't tested it yet. I think that there are so many different possible scenarios that this sort I function would be useful that I figured making a pull-request an allowing other people to test might be simpler than try to guess all the different scenarios myself. :)

hyperspy/_signals/eds.py
```diff
@@ -232,6 +233,110 @@ def rebin(self, new_shape):
...
+
+        Examples
+        --------
+        Input:
+        spectrum = hs.signals.EDSTEMSpectrum(np.ones([4, 4, 10]))
+        spectrum.data[1, 2, 9] = 5
+        print(spectrum)
+        print ('Sum = ', sum(sum(sum(spectrum.data))))
+        scale = [2, 2, 5]
+        test = spectrum.linear_bin(step)
```
thomasaarholt: Your example includes the step typo ;)

hyperspy/misc/array_tools.py
```diff
@@ -96,11 +97,72 @@ def rebin(a, new_shape):
...
+def linear_bin(s, scale):
+
+    """
+    Binning of the spectrum image by a non-integer pixel value.
+
+    Parameters
+    ----------
+    originalSpectrum: numpy.array
+        the original spectrum
+    step: a list of floats for each dimension specify the new:old pixel
```
thomasaarholt: Should be scale, here as well - copy the example from above, but as a function rather than a method?

thomasaarholt (commented):
Found a few more things

hyperspy/_signals/eds.py
k8macarthur: Thanks :)

hyperspy/misc/array_tools.py
k8macarthur: Already spotted. :)

k8macarthur:
So the function would need to copied and modified to work on 2D signals but I don't think it should be too much work.

thomasaarholt:
I might have a play with it at some point :)

We should write some tests for this. We could write a pretend signal and `linear_rebin` it to a known shape, then check the output. Any other smart things we could check?

k8macarthur:
The live_time should be checked as in:

`    ```
def test_rebin_live_time(self):
        s = self.signal
        old_metadata = s.metadata.deepcopy()
        dim = s.axes_manager.shape
        s = s.rebin([dim[0] / 2, dim[1] / 2, dim[2]])
        nt.assert_equal(
            s.metadata.Acquisition_instrument.TEM.Detector.EDS.live_time,
            3.1 \* 2 \* 2)
        # Check that metadata is unchanged
        print(old_metadata, self.signal.metadata)    # Captured on error
        nt.assert_dict_equal(old_metadata.as_dictionary(),
                             self.signal.metadata.as_dictionary(),
                             "Source metadata changed")

```
```

k8macarthur:
I have some basic tests already written just wasn't sure which file they should be going in...

thomasaarholt:
Getting this warning when binning by an non-integer:

```
/Users/thomas/Dropbox/0_Git/GitHub/hyperspy_aar/hyperspy/_signals/eds.py:315: VisibleDeprecationWarning: using a non-integer number instead of an integer will result in an error in the future
  newSpectrum[j] = newSpectrum[j] + s[bottomPos]
```

thomasaarholt:
If I try to bin a line profile I get an error:

``` python
>>> s
<EDSTEMSpectrum, title: EELS, dimensions: (70, 78|1995)>
>>> s2 = s.inav[0,:]
<EDSTEMSpectrum, title: EELS, dimensions: (78|1995)>
In [20]:
>>> s2.linear_bin([2,2])
IndexError                                Traceback (most recent call last)
<ipython-input-22-1e64d6c62abd> in <module>()
----> 1 s2.linear_bin([2,2])

/Users/thomas/Dropbox/0_Git/GitHub/hyperspy_aar/hyperspy/_signals/eds.py in linear_bin(self, scale)
    295             s[:] = newSpectrum
    296             newSpectrum = np.zeros((math.ceil(shape2[0]/step),
--> 297                                     shape2[1], shape2[2]), dtype='float')
    298             if k != 0:
    299                 s = np.swapaxes(s, 0, k)

```

thomasaarholt:
I'm about to submit a pull request which should enable support for line profiles and Signal2Ds. I'm also moving the method to the BaseSignal class, so that it is available to all types of signals.
I would be submitting it in the next ten minutes, but my laptop just ran out of power. Charging it tonight.

thomasaarholt:
Added a pull request to your repository. :)

https://github.com/k8macarthur/hyperspy/pull/6

thomasaarholt:
Huh. Neat. That was a nice idea.

hyperspy/misc/array_tools.py
```diff
@@ -123,18 +123,19 @@ def linear_bin(s, scale):
     """
 
     shape2 = s.shape
+    newSpectrum = np.zeros(s.shape)
+    newSpectrum = s[:]
 
-    if len(shape) != len(scale):
+    if len(shape2) != len(scale):
```
thomasaarholt, line 133: The line `newSpectrum = s[:]` entirely overwrites the variable `newSpectrum`, meaning that the previous line doesn't do anything at all. If you wanted to fill `newSpectrum` with the values in the `s` array you would do:
`newSpectrum[:] = s[:]`, but I'm not sure that's necessary at all.

Shouldn't `newSpectrum = s` be enough?

thomasaarholt:
If it proves difficult, then I'll just keep a copy of my code in a second document and just update your modified array_tools function.

thomasaarholt (commented):
Just a small note

k8macarthur:
In theory it should be enough but then I kept getting complaints that I was trying to perform all the following functions on a spectrum rather than an array...

k8macarthur:
@thomasaarholt if you're happy for the new structure do you want to add the new function for EELS?

thomasaarholt:
Yep, good idea. I'll learn from the way you added it and copy it over.

thomasaarholt:
Another thought: Should `scale` be a tuple instead of a list? It typically is, if you look at any type of `shape`.

k8macarthur:
Hi @thomasaarholt , I checked the EELS metadata, and it looks like EELS can have dwell_time or exposure so I've updated to two if statements.

k8macarthur:
Also scale can be a tuple if you prefer. I have no preference.

k8macarthur:
So I think looking back at #1170 I should close this pull request and resubmit to release next minor. I'm going to try this now.

k8macarthur:
#1306

PR #1306, opened 2016-10-14
Linear_Bin: function to bin spectrum by a non-integer number of pixels in each dimension.

k8macarthur:


k8macarthur:
Moved across from a pull request original submitted to the next_release_major #1305

doc/user_guide/tools.rst
```diff
@@ -796,8 +796,39 @@ method that operates *in place*.
...
+
+    >>> s = hs.datasets.example_signals.EDS_SEM_Spectrum()
+    >>> print(s)
+    >>> print(s.rebin([512]))
+    <EDSSEMSpectrum, title: EDS SEM Signal1D, dimensions: (|1024)>
+    <EDSSEMSpectrum, title: EDS SEM Signal1D, dimensions: (|512)>
+
+.. versionadded:: 1.1
+The second, uses the :py:meth:`~.signals.eds.linear_bin` to rebin the data,
+by a given ratio., or list of ratios. These can also be non-integers and less
```
tjof2: stray full stop

doc/user_guide/tools.rst
```diff
@@ -796,8 +796,39 @@ method that operates *in place*.
...
+than one in order to 'up-sample'. e.g:
```
tjof2: "...one in order to 'up-sample', for example:"

thomasaarholt:
@k8macarthur could you explain the case of the black lines on the right and bottom of the binned image?

thomasaarholt:
A reassuring observation: decomposition after rebin and linear_bin return the same s.explained_variance_ratio() for integer binning factors.
screenshot 2016-10-14 10 42 23

k8macarthur:
Excellent! The black pixel line in the right and bottom of the spectrum image is because if you were to bin 10 pixels by 3 gives an image with size 4 pixels where the final pixel only has 1/3 pixel worth of data making it look darker than the rest.

k8macarthur:
I did it this way to preserve counts but it could easily be cropped out if general consensus would be that way.

thomasaarholt:
Aha! I see! I assumed the behaviour would be to distribute the counts evenly into the new shape of the signal. I appreciate that this becomes problematic at the edges though, where you would normally be "getting" counts from outside the image if you were binning it by a factor larger than 1.

hyperspy/misc/array_tools.py
```diff
@@ -96,11 +97,86 @@ def rebin(a, new_shape):
...
+            if topPos > shape2[0]:
+                topPos = shape2[0]
+            while (topPos - bottomPos) >= 1:
+                if math.ceil(bottomPos) - bottomPos != 0:
+                    newSpectrum[j] = (newSpectrum[j] +
+                                      s[math.floor(bottomPos)] *
+                                      (math.ceil(bottomPos) - bottomPos))
+                    bottomPos = math.ceil(bottomPos)
+                else:
+                    newSpectrum[j] = newSpectrum[j] + s[bottomPos]
```
thomasaarholt: Looking at the above while loop, When `math.ceil(bottomPos) - bottomPos` **does** equal 0, bottomPos does not have `math.floor` or `math.ceil` operated on it and so it remains a `float`. When indexing a numpy array (ala `data[a]`, `a` must be an int (can't access position a[1.5] for instance).
This is currently returning a warning:

``` python
...hyperspy/misc/array_tools.py:168: VisibleDeprecationWarning: using a non-integer number instead 
of an integer will result in an error in the future
  newSpectrum[j] = newSpectrum[j] + s[bottomPos]
```

To prevent an error from being returned in the future, you need to make sure bottomPos is turned into an `int`. Since the only case where this is happening is when bottomPos equals a whole number (but is a float, so 2.0, 3.0...), it should suffice to call
`bottomPos = int(bottomPos)` after the `else` in line 168.

k8macarthur:
Yay, Travis now works, I found one mistake in the test function and I've updated 
newSpectrum[j] = newSpectrum[j] + s[bottomPos]
to
newSpectrum[j] = newSpectrum[j] + s[int(bottomPos)]

k8macarthur:
@thomasaarholt  do you think it would be useful to crop that extra half-pixel out of the final image? We could do crop=on crop=off to let the user have a choice? I just don't know how helpful  this would be...

thomasaarholt:
Unless you have a particularly good reason for leaving the half-pixel there, I would crop it off if it is present. The crop on/off argument sounds good. Leave a good description in the docstring.

k8macarthur:
Ok try that out. It should now crop by default but putting crop = 'anything' switches it off to allow you to check total intensity is preserved if you need to.

thomasaarholt:
@k8macarthur I'm just curious - why doesn't binning round to the nearest whole pixel, and then distribute the counts across this whole number of pixels? This would preserve counts, and be a more sensible way of binning, no?

k8macarthur:
That then places a restriction on the possible anmounts of resizing you can do. If you have a huge data set which is 2kx2k you would have more flexibility on how you bin than if you only have 50x50 image. The special cases which you're talking about would still work with this function though.

thomasaarholt:
Does it? I can't quite see how that wouldn't work :/

k8macarthur:
If you're pixel bin results in width/step= integer then you end up with the case you've just described. Where the counts are evenly distributed exactly into the new frame with no left over pixels.

k8macarthur:
I'm not sure what seems to be breaking AppVeyor, but this is ready to be reviewed. I don't think it can be far off being ready to merged either, if someone could have a look at it?

hyperspy/Release.py
```diff
@@ -25,7 +25,7 @@
 # When running setup.py the "+dev" string will be replaced (if possible)
 # by the output of "git describe" if git is available or the git
 # hash if .git is present.
-version = "1.2+dev"
+version = "2.0+dev"
```
francisco-dlp: Probably you didn't branch from the right branch, but it doesn't matter as RELEASE_next_minor and RELEASE_next_major are still identical except for the version. Could you set the version back to 1.2+dev?

hyperspy/_signals/eds.py
```diff
@@ -222,16 +224,107 @@ def rebin(self, new_shape):
...
+
+        """
+        Binning of the spectrum image by a non-integer pixel value.
+
+        Parameters
+        ----------
+        self: numpy.array
+            the original spectrum
+
+        scale: a list of floats for each dimension specify the new:old pixel
```
francisco-dlp: this docstring doesn't follow the [numpy specification](https://github.com/numpy/numpy/blob/master/doc/HOWTO_DOCUMENT.rst.txt).

hyperspy/_signals/eds.py
```diff
@@ -222,16 +224,107 @@ def rebin(self, new_shape):
...
+                    *= factors[i]
         if "Acquisition_instrument.TEM.Detector.EDS.live_time" in s.metadata:
-            for factor in factors:
+            for i, t in enumerate(s.axes_manager.navigation_axes):
                 s.metadata.Acquisition_instrument.TEM.Detector.EDS.live_time\
-                    *= factor
+                    *= factors[i]
         return s
 
+    def linear_bin(self, scale, crop='on'):
```
francisco-dlp: Wouldn't it be better to make `crop` take `True` and `False` instead?

francisco-dlp: Why adding this only to the EDS signal and not to `BaseSignal`? Actually, why not adding the functionality to the existing `rebin`?

hyperspy/misc/array_tools.py
```diff
@@ -96,11 +97,106 @@ def rebin(a, new_shape):
     new_shape = tuple(int(ns) for ns in new_shape)
     factor = np.asarray(shape) // np.asarray(new_shape)
     evList = ['a.reshape('] + \
-             ['new_shape[%d],factor[%d],' % (i, i) for i in range(lenShape)] + \
+             ['new_shape[%d],factor[%d],' % (i, i) for i in range(lenShape)] +\
              [')'] + ['.sum(%d)' % (i + 1) for i in range(lenShape)]
     return eval(''.join(evList))
 
 
+def _linear_bin(s, scale,
```
francisco-dlp: I think that it should be possible to improve the performance by i) rewriting it in C or Cython ii) vectorizing it. iii) using numba? Now that we distribute some cython code, i) doesn't look like a bad idea. I have no experience with numba, but, although it seems to be perfect for the job, it is probably not straightforward to install it in anything but anaconda. However, the easiest approach might be to "vectorize it". I think that it should be possible to write this functions as the standard `rebin` followed by a postprocessing step. For example, say that you want to rebin a list of 5 numbers, `a`, using 2.5 bins. You could do it like this:

```python

rebin_size = floor(scale)
a2rebin = a[0, 1, 3, 4]
result = rebin(a2rebin, ((2,)))
result[0] = (a[1] + a[2]) / 2
result[1] = (a[1] + a[2]) / 2

```

This is of course the simplest example that I've been able to come with, but I think that it should be possible to write it so that it works in the general case by automatically generating the index to obtain `a2rebin` and the interpolation step at the end. What do you think?

francisco-dlp (changes requested):
I think that this contribution would make many people happy. The main issue with this implementation is that it is slow because it iterates over the array using plain Python.

hyperspy/Release.py
k8macarthur: Fixed :)

hyperspy/_signals/eds.py
k8macarthur: Sure, that can be done.

k8macarthur: I think it is in the BaseSignal. I've added this function to EDS, base and EELS. The EDS and EELS functions  specifically adjust the dwell time as well to account for the new bin sizes.

hyperspy/misc/array_tools.py
k8macarthur: I haven't tried the code on anything larger than 600x600x2048 which seems to work in 2 seconds according to timeit:
import timeit
timeit.timeit(s.linear_bin(scale))
Out[22]:
2.048694275028538
But I guess that might depending on how much you're binning by. I agree faster would be better as I'm sure people would end up trying larger datasets than this. I would know where to start with optimising though. I'll try asking around the Jülich guys today. I would prefer to convert into C/Cython than vectorise if possible as I'm not sure I understand the vectorisation well enough to trust it's doing the exact correct thing. Currently the function is slow but 100% correct.

thomasaarholt:
In the interest of moving forwards iteratively, could I propose merging this (if the current incarnation is complete) and then creating a new pull request to move it to rebin/Cython/C?

francisco-dlp:
I have tried to optimize the code using plain numpy and I came up with this:

```python
import numpy as np
import math

def rebin_1d(array, new_size, axis):
    pad = (slice(None), ) * axis
    step =   array.shape[axis] / new_size
    new_shape = list(array.shape)
    new_shape[axis] = new_size
    result = np.zeros(new_shape)
    j = 0
    result_idx = pad + (j,) + (slice(None), ) * (array.ndim - 1 - axis)
    for i in range(0, array.shape[axis]):
        next_bin = step * (j + 1)
        new_floor = math.floor(next_bin)
        array_idx = pad + (i,) + (slice(None), ) * (array.ndim - 1 - axis)
        if i < new_floor:
            result[result_idx] += array[array_idx]
        else:
            new_fraction = next_bin - new_floor
            result[result_idx] += array[array_idx] * new_fraction    
            j += 1
            result_idx = pad + (j,) + (slice(None), ) * (array.ndim - 1 - axis)
            result[result_idx] += array[array_idx] * (1 - new_fraction)         
    return result


def rebin_nd(array, new_shape):
    for i, new_size in enumerate(new_shape):
        if new_size != array.shape[i]:
            array = rebin_1d(array, new_size, i)
    return array

```

Actually, despite the effort, this is only around 3x faster than the original one. The memory footprint is also lower. Probably there is room for further optimisation. Certainly that's the case for the memory footprint, although at the expense of performance unless we turn to Cython or Numba. As far as I can tell this code returns the same result as the original one with `crop=True`. Is there any case when one would like `crop=False`? Unless I am getting it wrong, `crop` is only necessary when the given factor results in a shape with fractional number of pixels. By providing the new shape instead of the factor, maybe there is no need for crop?

to266:
@francisco-dlp not sure if the function is meant to be used that way, but "subsampling" does not work (e.g. going from shape `(3,)` to shape `(7,)`) due to the assumption that the new size is always less than the original (the `(1 - fraction)` bit in particular).

Also, additional optimization can be done by reducing the loop length in half by noticing that the operation is always symmetric in terms of bins.

k8macarthur:
@francisco-dlp and @to266 The original function was designed either to reduced or increase the pixel width which is why I left the cropping that @thomasaarholt  suggested only as an option. Perhaps it should be explained slightly more clearly in the documentation but in theory using crop=FALSE and pixel sizes less than 1 should sub-sample the array. I haven't checked this recently since writing the pull request but in my original testing this did work. I'd be keen to keep this functionality if possible.

francisco-dlp:
@to266, good point, exploiting the symmetry should increase the performance. @to266 and @k8macarthur, wouldn't upsampling be equivalent to linear interpolation? Then we could use the faster `scipy.ndimage.zoom` instead for this case.

francisco-dlp:
I have tried exploiting the symmetry as proposed by @to266 but it didn't help with the performance. Here is the code:

```python

import numpy as np
import math

def rebin_1d_sym(array, new_size, axis):
    lpad = (slice(None), ) * axis
    rpad = (slice(None), ) * (array.ndim - 1 - axis)
    step =   array.shape[axis] / new_size
    new_shape = list(array.shape)
    new_shape[axis] = new_size
    result = np.zeros(new_shape)
    size = array.shape[axis]    
    j = 0
    result_idx = lpad + (j,) + rpad
    for i in range(0, size // 2):
        next_bin = step * (j + 1)
        new_floor = math.floor(next_bin)
        array_idx = lpad + (i,) + rpad
        array_idx_r = lpad + (size - i - 1,) + rpad
        result_idx_r = lpad + (new_size - j - 1,) + rpad
        if i < new_floor:
            result[result_idx] += array[array_idx]
            result[result_idx_r] += array[array_idx_r]
        else:
            new_fraction = next_bin - new_floor
            result[result_idx] += array[array_idx] * new_fraction
            result[result_idx_r] += array[array_idx_r] * new_fraction 
            j += 1
            result_idx = lpad + (j,) + rpad
            result_idx_r = lpad + (new_size - j - 1,) + rpad
            result[result_idx] += array[array_idx] * (1 - new_fraction)
            result[result_idx_r] += array[array_idx_r] * (1 - new_fraction)
    if size % 2 != 0:
        if new_size % 2 != 0:
            result[lpad + (new_size // 2,) + rpad] += array[lpad + (size // 2,) + rpad]
        else:
            result[lpad + (new_size // 2,) + rpad] += 0.5 * array[lpad + (size // 2,) + rpad]
            result[lpad + (new_size // 2 - 1,) + rpad] += 0.5 * array[lpad + (size // 2,) + rpad]
            
    return result


def rebin_nd_sym(array, new_shape):
    for i, new_size in enumerate(new_shape):
        if new_size != array.shape[i]:
            array = rebin_1d_sym(array, new_size, i)
    return array

```

francisco-dlp:
Some further performance enhancements:

```python

import numpy as np
import math

def rebin_1d_slice(array, new_size, axis):
    lpad = (slice(None), ) * axis
    rpad = (slice(None), ) * (array.ndim - 1 - axis)
    step =   array.shape[axis] / new_size
    new_shape = list(array.shape)
    new_shape[axis] = new_size
    result = np.zeros(new_shape)
    size = array.shape[axis]    
    j = 0
    result_idx = lpad + (j,) + rpad
    ls = 0
    for i in range(0, size):
        next_bin = step * (j + 1)
        new_floor = math.floor(next_bin)
        if i >= new_floor:  
            result[result_idx] += array[lpad + (slice(ls, i),) + rpad].sum(axis=axis)
            new_fraction = next_bin - new_floor
            array_idx = lpad + (i,) + rpad
            result[result_idx] += array[array_idx] * new_fraction
            j += 1
            result_idx = lpad + (j,) + rpad
            result[result_idx] += array[array_idx] * (1 - new_fraction)
            ls = i + 1
    result[result_idx] += array[lpad + (slice(ls, i + 1),) + rpad].sum(axis=axis)
            
    return result

def rebin_nd_slice(array, new_shape):
    factor = np.array(array.shape) / np.array(new_shape)
    sortidx = np.argsort(factor)[::-1]
    for i in sortidx:
        new_size = new_shape[i]
        if new_size != array.shape[i]:
            array = rebin_1d_slice(array, new_size, i)
    return array
```

k8macarthur:
My testing seems to agree that it works and it is faster, thank you for all your work trying to speed up the function. My only issue is that the new function is more complicated to apply to my work I I specifically want to tailor the pixel width. I think specifying the new shape rather than the new pixel dimension limits the number of binning options to a discrete number. I also wasn't aware that scipy.ndimage.zoom worked for multidimensional arrays. I still think it would be neater to have binning and subsampling in the same function. Is there anyway we can modify your suggested function to include both of these functionalities still? I have been looking at converting back to specifying a pixel size and I can't quite get it to run:

```python
import numpy as np
import math

def rebin_1d_slice(array, step, axis, crop=True):
    lpad = (slice(None), ) * axis
    rpad = (slice(None), ) * (array.ndim - 1 - axis)
    new_shape = list(array.shape)
    size = array.shape[axis] 
    if crop == True:
        new_shape[axis] = math.floor(array.shape[axis]/step)
    else:
        new_shape[axis] = math.ceil(array.shape[axis]/step)
    result = np.zeros(new_shape)
    j = 0
    result_idx = lpad + (j,) + rpad
    ls = 0
    for i in range(0, size):
        next_bin = step * (j + 1)
        new_floor = math.floor(next_bin)
        if i >= new_floor:  
            result[result_idx] += array[lpad + (slice(ls, i),) + rpad].sum(axis=axis)
            new_fraction = next_bin - new_floor
            array_idx = lpad + (i,) + rpad
            result[result_idx] += array[array_idx] * new_fraction
            j += 1
            result_idx = lpad + (j,) + rpad
            result[result_idx] += array[array_idx] * (1 - new_fraction)
            ls = i + 1
    result[result_idx] += array[lpad + (slice(ls, i + 1),) + rpad].sum(axis=axis)
            
    return result

def rebin_nd_slice(array, factor, crop=True):
    new_shape = list(array.shape)
    if crop == True:
        for i in range(len(new_shape)):
            new_shape[i] = math.floor(np.array(array.shape[i]) / factor[i])
    else:
        for i in range(len(new_shape)):
            new_shape[i] = math.ceil(np.array(array.shape[i]) / factor[i])
    sortidx = np.argsort(factor)[::-1]
    for i in sortidx:
        new_size = new_shape[i]
        if new_size != array.shape[i]:
            array = rebin_1d_slice(array, factor[i], i, crop)
    return array
```

thomasaarholt:
Kate, type ``` to before and after your code in the discussion field to format the code with the code font. If you type "python" immediately after the first triple, it will give it Python syntax highlighting as well.

to266:
[note that I edited @k8macarthur comment to display the code in a more readable format (as per @thomasaarholt )]

k8macarthur:
Ok thanks. :)

to266:
In order to combine the arbitrary (albeit still >1) step size, the function should:
 1. find out how many whole pixels contain the information that would be "cropped"
 2. store these separately and remove from the original array
 3. append to the end of the original array the last pixel form (2) that still gets included in the "whole" step, but multiplied by the relevant fraction
 4. rebin using the fast Francisco code
 5. if crop=False, append to the end the stored summed (if there were more than 1 in the binning dimension) pixels, multiplied by the leftover step/weight.

for `bin=1.5` and starting array `[1,1,1,1,1]` this would look something like
```
[1, 1, 1, 1, 1] #original
[1, 1, 1, 1] and [1] #split
[1, 1, 1, 1, 0.5] and [0.5] # multiply by weights and tack back on
[1.5, 1.5, 1.5] and [0.5] # rebin, no leftovers, so fast code works
[1.5, 1.5, 1.5, 0.5] # tack back on the leftover if crop=False
```

francisco-dlp:
I think that it'll be better to use `zoom` instead of adding the functionality to this function because `zoom` should be faster. Also it offers different option to deal with the boundaries. We could incorporate `zoom` in the same `method` e.g. `if step < 1 use zoom if > 1 use rebin else do nothing`.

Regarding factor vs bin discussion, I am not sure if accepting the factor instead or in addition to the shape is a good idea. The problem of rebinning as I understand it is to assign  the counts to a different set of bins of the same width spanning the same interval. Therefore, there cannot be fractional bins nor can the final interval be different. In @to266 example, if the result of binning `[1, 1, 1, 1, 1]` by 1.5 is `[1.5, 1.5, 1.5, 0.5]` it follows that the width of the last bin is 1/3 of the rest or that the interval is larger and that anything outside the original interval is assumed to be 0.

k8macarthur:
Hi All, With some help of a friend I have tried to speed up the original function this weekend. Let me know what you think.

Also, using a factor vs a bin give you the option of binning a spectrum in to pixel sizes of 1.2, 1.3 or 1.4 where the left over cropped pixel is would be a different size. At the limit of small spectrum there will only be a small number of available pixel sizes.

hyperspy/_signals/eels.py
```diff
@@ -1253,3 +1254,76 @@ def create_model(self, ll=None, auto_background=True, auto_add_edges=True,
...
+              dimension.
+         crop: when binning by a non-integer number of pixels it is likely that
+               the final row in each dimension contains less than the full
+               quota to fill one pixel.
+             e.g. 5*5 array binned by 2.1 will produce two rows containing 2.1
+             pixels and one row containing only 0.8 pixels worth. Selection of
+             crop = 'True' or crop = 'False' determines whether or not this
+             'black' line is cropped from the final binned array or not.
+
+        *Please note that if crop = 'False' is used:the final row in each
```
tjof2: typo, should be a comma not a colon?

hyperspy/_signals/eels.py
```diff
@@ -1253,3 +1254,76 @@ def create_model(self, ll=None, auto_background=True, auto_add_edges=True,
...
+        dimension may appear black, if a fractional number of pixels are left
+        over. It can be removed but has been left to preserve total counts
+        before and after binning.*
+
+        Return
+        ------
+        A new spectrum image with new dimensions width/scale for each
+        dimension in the data. The axes scales and dwell_time/exposure are also
+        corrected accordinly.
```
tjof2: another typo :-)

k8macarthur:
Hi @francisco-dlp , Just getting the time to come back to this again. Could you or someone remind me what the major issue is with it being included? Thanks Kate

francisco-dlp:
I think that this is a great addition. However, the current implementation is still slow. Unless somebody knows how to speed it up using plain numpy, I'll try to do it using numba if that's ok.

francisco-dlp:
k8macarthur/hyperspy#10

numba is amazing, the numbified version of ``linear_bin`` is as fast as the current ``rebin`` even though it is more general and, probably, there is still room for optimization.

@k8macarthur, although the function looks rather different, it's just a modified version of your algorithm so it should work exactly in the same way.

@hyperspy/developers, if this gets accepted, I suggest to add numba to extra_dependencies after modifying the code to make the import conditional. I've also tried Cython but numba is 10**10 sweeter!

francisco-dlp:
Some more thoughts on this PR:

* It'll be nice to get the functionality in ``rebin`` rather than adding an extra ``linear_bin`` method. ``rebin`` could take two incompatible parameters: ``new_shape`` and ``scales``. The current ``rebin`` could be used to process the axes with ``size % scale == 0`` and the rest with ``linear_bin`.
* When there is a mismatch between the "length" of the input and output (``crop=False``), the last item suffers from low intensity because we assume that the signal is 0 outside of the range of the original signal. Wouldn't it be more useful to multiply it by a factor. e.g. if the last item value corresponds to "exposing" 1/3 of its size, multiplying the current value by 3 would provide a better guess. We could control how we deal with the boundary with an extra keyword.
* Tests need to test also that the metadata is updated correctly and that the dtype is conserved when possible.
* @to266, any thoughts on lazifying this?
* Finally, although not obviously required for this PR, it should be possible to improve the memory requirements of this function by processing all dimensions at once. Currently it iterates over the axes. I think (but I am not sure) that the current ``rebin`` should be superior in this regard.

to266:
Its [possible](https://dask.readthedocs.io/en/latest/array-ghost.html) but not trivial. 

The (jitted) rebinning function should then be rewritten to not always operate on the first axis (simple) and be able to operate from someplace else than the end of the axis (more difficult). The location would be known though (using the `block_id`).

The actual function will have to calculate not only the shape of the new rebinned array, but also new chunks (in essence the shape of each block after the rebinning) before running.

k8macarthur:
I agree including it into the re_bin function would be much simpler for users than a separate function, then people wouldn't have to think about  which one they need to use. 

Personally, I do maps of nanoparticles to the extra fraction of a pixel at the edge of my image is normally vacuum. Therefore I hadn't given it too much thought. I guess multiplying it up might be a nice option to maintain as much information as possible. 

I'll have a think about some metadata tests. Currently I've only been checking the ones that should change (i.e. scaling and dwell_times) but I'm not sure if I wrote the actual tests for those.

hyperspy/misc/array_tools.py
```diff
@@ -115,6 +120,91 @@ def rebin(a, new_shape):
...
+                x1 += 1
+            if x2 != x1:
+                value+= data[math.floor(x1)]*(x2-x1)
+        else:
+            fx1 = math.floor(x1)
+            cx1 = math.ceil(x1)
+            if scale > (cx1 - x1) > 0:
+                value += data[fx1] * (cx1 - x1)
+                x1 = cx1
+            if x1 < (result.shape[0] + 1:
```
magnunor: Syntax error here

hyperspy/signal.py
```diff
@@ -2249,6 +2251,73 @@ def rebin(self, new_shape, out=None):
 
     rebin.__doc__ %= OUT_ARG
 
+    def linear_bin(self, scale, crop=True):
+        """
+        Binning of the spectrum image by a non-integer pixel value.
+
+        Parameters
+        ----------
+        scale: a list of floats for each dimension specify the new:old pixel
```
magnunor: This should be 'scale : ', same for all the other parameters

hyperspy/signal.py
```diff
@@ -2249,6 +2251,73 @@ def rebin(self, new_shape, out=None):
...
+        Return
+        ------
+        A new spectrum image with new dimensions width/scale for each
+        dimension in the data. The axes scales but not the dwell_time/exposure
+        have been corrected accordingly.
+
+        Examples
+        --------
+        Input:
+        spectrum = hs.signals.EDSTEMSpectrum(np.ones([4, 4, 10]))
```
magnunor: Should be >>> in front of all the inputs in the examples

magnunor: The example should be something like this:
```Python
>>> spectrum = hs.signals.EDSTEMSpectrum(np.ones([4, 4, 10]))
>>> spectrum.data[1, 2, 9] = 5
>>> print(spectrum)
<EDSTEMSpectrum, title: , dimensions: (4, 4|10)>
>>> print ('Sum = ', sum(sum(sum(spectrum.data))))
Sum =  164.0
>>> scale = [2, 2, 5]
test = spectrum.linear_bin(scale)
>>> print(test)
<EDSTEMSpectrum, title: , dimensions: (2, 2|2)>
>>> print('Sum = ', sum(sum(sum(test.data))))
Sum =  164.0
```
I haven't really tried the example itself, just commenting on the formatting

hyperspy/signal.py
```diff
@@ -2249,6 +2251,73 @@ def rebin(self, new_shape, out=None):
...
+             pixels and one row containing only 0.8 pixels worth. Selection of
+             crop = 'True' or crop = 'False' determines whether or not this
+             'black' line is cropped from the final binned array or not.
+
+        *Please note that if crop=False is used, the final row in each
+        dimension may appear black, if a fractional number of pixels are left
+        over. It can be removed but has been left to preserve total counts
+        before and after binning.*
+
+        Return
```
magnunor: Returns

hyperspy/misc/array_tools.py
```diff
@@ -115,6 +120,91 @@ def rebin(a, new_shape):
...
+         fill one pixel.
+         e.g. 5*5 array binned by 2.1 will produce two rows containing 2.1
+         pixels and one row containing only 0.8 pixels worth. Selection of
+         crop_str = 'True' or crop = 'False' determines whether or not this
+         'black' line is cropped from the final binned array or not.
+        *Please note that if crop=False is used, the final row in each
+    dimension may appear black, if a fractional number of pixels are left
+    over. It can be removed but has been left to preserve total counts
+    before and after binning.*
+    Return
```
magnunor: Returns

hyperspy/misc/array_tools.py
```diff
@@ -115,6 +120,91 @@ def rebin(a, new_shape):
...
+                x1 = cx1
+            if x1 < (result.shape[0] + 1:
+                value += data[math.floor(x1)]*(x2-x1)
+
+def _linear_bin(dat, scale, crop=True):
+    """
+    Binning of the spectrum image by a non-integer pixel value.
+    Parameters
+    ----------
+    originalSpectrum: numpy.array, or the s.data, where s is a signal array.
```
magnunor: originalSpectrum :

magnunor:
How the numpy docstring formatting should be: https://github.com/numpy/numpy/blob/master/doc/example.py#L38

magnunor:
The comments about the docstring in signal.linear_bin also has to be done for the other functions + the user guide examples.

hyperspy/misc/array_tools.py
k8macarthur: Already spotted this morning, I really shouldn't code last thing....

hyperspy/signal.py
k8macarthur: Not sure quite what you mean. This currently follows that same format as the rest of the function parameters in the file?

hyperspy/signal.py
k8macarthur: I guess that's the difference between using command line and notebook. Happy to follow and comparable format though.

k8macarthur:
So I think I've fixed all the suggestions from @magnunor but Travis still seems to be failing... There is an error saying that the Index is out of range but I can't reproduce on my computer. On my computer the tests run completely fine...Any suggestions would really be appreciated.

doc/user_guide/tools.rst
```diff
@@ -895,8 +895,39 @@ method that operates *in place*.
...
+The first uses the :py:meth:`~.signal.BaseSignal.rebin` method to rebin data,
+to a new dimensional size specified by the user. Please note the new shape must
+be a divisor of the original shape.
+
+.. code-block:: python
+
+    >>> s = hs.datasets.example_signals.EDS_SEM_Spectrum()
+    >>> print(s)
+    >>> print(s.rebin([512]))
+    <EDSSEMSpectrum, title: EDS SEM Signal1D, dimensions: (|1024)>
```
magnunor: This needs to be reordered (same as the docstrings), since I think this will fail with the new docstring/userguide tests.
```Python
>>> print(s)
<EDSSEMSpectrum, title: EDS SEM Signal1D, dimensions: (|1024)>
>>> print(s.rebin([512]))
<EDSSEMSpectrum, title: EDS SEM Signal1D, dimensions: (|512)>
```

doc/user_guide/tools.rst
```diff
@@ -895,8 +895,39 @@ method that operates *in place*.
...
+.. code-block:: python
+
+    >>> spectrum = hs.signals.EDSTEMSpectrum(np.ones([4, 4, 10]))
+    >>> spectrum.data[1, 2, 9] = 5
+    >>> print(spectrum)
+    >>> print ('Sum = ', sum(sum(sum(spectrum.data))))
+    >>> scale = [0.5, 0.5, 5]
+    >>> test = spectrum.linear_bin(step)
+    >>> print(test)
+    >>> print('Sum =', sum(sum(sum(test.data))))
```
magnunor: Same here

hyperspy/_signals/eds.py
```diff
@@ -191,13 +193,88 @@ def rebin(self, new_shape, out=None):
...
+
+            *Please note that if crop=False is used, the final row in each
+            dimension may appear black, if a fractional number of pixels are left
+            over. It can be removed but has been left to preserve total counts
+            before and after binning.*
+
+        Returns
+        ------
+        s : Signal subclass
+
```
magnunor: The ---- after the "headline" needs to be as many symbols as the headline, so add another - here.

hyperspy/_signals/eels.py
```diff
@@ -1331,6 +1332,71 @@ def create_model(self, ll=None, auto_background=True, auto_add_edges=True,
...
+             'black' line is cropped from the final binned array or not.
+
+            *Please note that if crop=False is used, the final row in each
+            dimension may appear black, if a fractional number of pixels are left
+            over. It can be removed but has been left to preserve total counts
+            before and after binning.*
+
+        Returns
+        ------
+        s : Signal subclass
```
magnunor: Same here, add another -

hyperspy/misc/array_tools.py
```diff
@@ -115,6 +120,97 @@ def rebin(a, new_shape):
...
+         'black' line is cropped from the final binned array or not.
+
+        *Please note that if crop=False is used, the final row in each
+        dimension may appear black, if a fractional number of pixels are left
+        over. It can be removed but has been left to preserve total counts
+        before and after binning.*
+
+    Returns
+    ------
+    np.array
```
magnunor: And here, another -

hyperspy/signal.py
```diff
@@ -2249,6 +2251,67 @@ def rebin(self, new_shape, out=None):
...
+             'black' line is cropped from the final binned array or not.
+
+            *Please note that if crop=False is used, the final row in each
+            dimension may appear black, if a fractional number of pixels are left
+            over. It can be removed but has been left to preserve total counts
+            before and after binning.*
+
+        Returns
+        ------
+        s : Signal subclass
```
magnunor: And here, another -

thomasaarholt:
Personally I would love to see some comments in the code explaining what the various bits do. Too often, in hyperspy code, there is little to no explanation :)

hyperspy/_signals/lazy.py
```diff
@@ -21,7 +21,7 @@
 
 import numpy as np
 import dask.array as da
-import dask.delayed as dd
+#import dask.delayed as dd
```
to266: Please remove this, as it breaks __many__ things

k8macarthur: Oops, sorry! Must have forgotten to exclude that from one of my more recent commits.

k8macarthur:
So it looks like travis is failing but not on anything to do with this pull request.

ericpre:
Except the small and straightforward conflict, this looks very good! I guess, it could make it in the next release (#1535)?!

k8macarthur:
Conflict fixed :) I think Travis and Appveyor are both failing because of tests unrelated to this function.

k8macarthur:
@francisco-dlp is there anything else you want from me before this can be merged?

thomasaarholt:
Can we rerun tests on this? (I don't know how)

thomasaarholt:
Nice one:) 

I think you're ok? Afaik only the tiff tests are failing?

francisco-dlp:
I still think that it'll be good to include this feature in the current ``rebin`` rather than adding a new method.

k8macarthur:
@francisco-dlp  I still think it's fine as two separate functions. However, in the interest of just getting this rolled in and off my desk so I can focus on other stuff I can make the changes if other people agree with you.

k8macarthur:
I've left the original rebin function in the signal files for now but added an if statement in the np.array function. Let me know what you think.

ericpre:
One function instead of two is more simple and less confusing for the end user, so people doesn't have to worry, which one they should be using.

hyperspy/tests/signal/test_eds_tem.py
```diff
@@ -327,3 +327,38 @@ def test_manual_remove_element(self):
         assert (
             sorted(s._xray_markers.keys()) ==
             ['Zn_Ka', 'Zn_La'])
+
+
+class Test_linear_bin:
+
+    def test_linear_bin1(self):
+        spectrum = EDSTEMSpectrum(np.ones([3, 3, 1]))
+        res = spectrum.linear_bin([1.5, 1.5, 1])
```
ericpre: Forgot to update linear_bin to rebin, the tests will not pass.

k8macarthur: Well spotted, thanks.

k8macarthur:
I've combined most of the rebin functions, but I don't think I understand how lazy works to modify that correctly.. I would be grateful if someone could look over those sections.

k8macarthur:
Apologies for the mass of commits but I think I've now finished merging the two functions. Could do with a check through. Hopefully it's nearly there now.

hyperspy/misc/array_tools.py
```diff
@@ -87,32 +108,156 @@ def rebin(a, new_shape):
...
+                return da.coarsen(np.sum, a, {i: int(f)
+                                              for i, f in enumerate(scale)})
+            # we provide slightly better error message in hypersy context
+            except ValueError:
+                raise ValueError("Rebinning does not allign with data dask chunks."
+                                 " Rebin fewer dimensions at a time to avoid this"
+                                 " error")
+
+
+def jit_ifnumba(func):
```
ericpre, line 176: Move jit_ifnumba decorator to `hyperspy.decorator`, since it could be used elsewhere?

hyperspy/tests/signal/test_tools.py
```diff
@@ -593,14 +593,14 @@ def test_valuemax(self):
 
     def test_rebin(self):
         s = self.s
-        new_shape = (3, 2, 1, 3)
+        scale = (1, 2, 1, 2)
         if self.s._lazy:
```
ericpre: This pytest.skip can be removed since setup.py says dask>=0.13?
The same for `TestOutArg.test_rebin`.

hyperspy/misc/array_tools.py
```diff
@@ -87,32 +108,156 @@ def rebin(a, new_shape):
...
+        dimension in the data.
+    """
+    if len(dat.shape) != len(scale):
+        raise ValueError(
+            'The list of bins must match the number of dimensions, including the\
+            energy dimension.\
+            In order to not bin in any of these dimensions specifically, \
+            simply set the value in shape to 1')
+
+    if not hasattr(_linear_bin_loop, "__numba__"):
```
ericpre: Move this warning to the decorator?

hyperspy/_signals/eds.py
```diff
@@ -177,25 +178,85 @@ def sum(self, axis=None, out=None):
...
+        if "Acquisition_instrument.TEM.Detector.EDS.real_time" in m.metadata:
+            for i, t in enumerate(m.axes_manager.navigation_axes):
+                m.metadata.Acquisition_instrument.TEM.Detector.EDS.real_time\
+                    *= scale[i]
+        if "Acquisition_instrument.TEM.Detector.EDS.live_time" in m.metadata:
+            for i, t in enumerate(m.axes_manager.navigation_axes):
+                m.metadata.Acquisition_instrument.TEM.Detector.EDS.live_time\
+                    *= scale[i]
+        if m.metadata.has_item('Signal,Noise_properties.variance'):
+            if isinstance(m.metadata.Signal.Noise_properties.variance,
```
ericpre: Wondering if the same should be done in the EELS case?

hyperspy/_signals/eds.py
```diff
@@ -177,25 +178,85 @@ def sum(self, axis=None, out=None):
...
+                          BaseSignal):
+                var = m.metadata.Signal.Noise_properties.variance
+                m.metadata.Signal.Noise_properties.variance = var.rebin(
+                    scale)#can't find where this function goes...
```
ericpre: I don't understand this comment. Is it a leftover from previous version?

ericpre (changes requested):
I made a PR to your branch to improve the docstrings/documentation, add tests and fix a few others things. Here, there are just a few minor comments.

The numba decorator works like a charm!

ericpre:
I have just log an issue on binning RGB images (#1617), which is not working. I mention it here _just for information_. Since this was already broken before this PR, I would say that it would be fair not to ask more on this PR. @k8macarthur: that being saying, if you strongly feel, that you could address this issue, this is obviously more than welcome! ;)

hyperspy/_signals/eds.py
thomasaarholt: Does this even need to be in eds.py? Couldn't it just as well have been under the standard signal method? It would make it easier to spot, should we be adding more signal types.

hyperspy/tests/signal/test_tools.py
k8macarthur: Don't know I was just leaving requirements like that to match the original code.

hyperspy/_signals/eds.py
k8macarthur: this check is also in the base signal but EDS has the additions of updating all the dwell time functions.

hyperspy/_signals/eds.py
k8macarthur: Yep gone.

k8macarthur:
@ericpre Just got back from my conference this week so started looking at this again. I would prefer to get this pull-request accepted before adjusting for RGB figures as I'm not sure exactly how the RGB figures are constructed.

hyperspy/misc/array_tools.py
k8macarthur: I have no preference over structural architecture, don't mind either way around.

k8macarthur:
test_plot_rectange_markers is currently what's failing appveyor and travis

ericpre:
This could be merged as it is. This is working nicely and the remaining comments are minor things, which could be addressed in another PR.  One PR will be needed anyway for RGB binning.

The failing test are plotting tests and it seems to be due to upstream changes... so not related to this PR.

ericpre:
The tests failing here are fixed in #1481, so the tests of this PR can be regarded as _passing_.

hyperspy/tests/signal/test_eds_tem.py
```diff
@@ -327,3 +327,48 @@ def test_manual_remove_element(self):
         assert (
             sorted(s._xray_markers.keys()) ==
             ['Zn_Ka', 'Zn_La'])
+
+
+@lazifyTestClass
```
to266: This decorator is not required and in fact may be harmful by tricking that the linear binning works lazily (it does not).

I suggest removing it and creating a new issue regarding the lazy implementation

ericpre: Ok, I did try it manually to load some data lazily and rebin the data and if I remember correctly, it did rebin the data, so I naively assumed that it was working... but I may have done something wrong or miss something!
Agree, better to remote it for now.

Are the tests still supposed to pass after the signals have been lazified inside the tests?

to266: The decorator looks for every test (i.e. a method that starts with `test...`), and creates its copy (which will be named `test_lazy...`) and prepends code that transforms all signals in the class as lazy. 

So these tests do not work for lazy because of two things:
 1. No `setup_method`, hence no signals to "lazify" to begin with.
 2. The original code is not adapted for lazy signals in general. I thought about it for around an hour at some point, and from what I can tell it is __much__ more difficult (but possible) to make it work lazily efficiently.

So currently it just runs the tests normally twice.

ericpre: Thanks for the explanations @to266!
1. I was missing the point of the `setup_method`, so yes, the exact same tests are run twice... But after defining the signals in `setup_method`, the tests run correctly, except for the one where some data are changed manually, which is normal.
2. I am not sure to understand here, but it looks to me that rebinning lazy signals works and after fixing the tests they also work. The following works:
```python
import numpy as np
import hyperspy.api as hs

s = hs.signals.Signal1D(np.arange(100).reshape((10, 10)))
sl = s.as_lazy()

# linear interpolation
sl2 = sl.rebin((3,4))
# fast integer rebin
sl3 = sl.rebin((2,2))
```

to266: It "works" in the sense that it does not crash for your small examples and gives a correct (non-lazy) answer.

It does not work as it's supposed to for lazy signals (which should be done both lazily and in blocks instead of just iterating dimensions and summing things). Instead it creates a numpy array to store the final and intermediate results, that essentially shrinks in every dimension as the algorithm iterates over them. So if you're not rebinning substantially in the first one (or maybe even up-scaling?) it "blows up" straight away.

Sorry for the short answer (I might go into detail at some point later this or next week if there is any need), but trust me - it does not do what it needs to for the lazy signals.

k8macarthur: Agreed, I have done nothing to adapt the function for lazy signals. I was trying to work out a way to still allow Lazy_rebin by integer pixels from the original rebin function but I'm not sure if I succeeded with that...

ericpre: OK, then rebinning of lazy signals will need some special care and will be for another day/PR!!
I think this is all good now?! (I feel like, it is not the first time I said this in this PR...)

thomasaarholt:
I tried upscaling a `(5, 5)` signal to `(10, 10)`. I do not have numba installed, and got the following result:
```python
>>> import hyperspy.api as hs
>>> import numpy as np
>>> ones = 100*np.ones((5,5))
>>> s = hs.signals.Signal1D(ones)
>>> s2 = s.rebin((10,10))
WARNING:hyperspy.misc.array_tools:Install numba to speed up the computation of `rebin`
>>> s2.data.shape
(0, 0)
```
The current warning is probably unrelated to the (0, 0) result. Any thoughts on why upscaling isn't happening?

I've tested with both Signal1D and 2D, with identical results.

k8macarthur:
To upscale you need to do s.rebin((0.5, 0.5))

thomasaarholt:
Of course.... You even reminded me as much, earlier :) Okay, that's fine then! 

But what about a warning saying that the signal has been binned into oblivion? There may be others who do the same mistake I do. (and yes, the should read the docs, but ~~I~~ people *ahem* never learn.)

k8macarthur:
Fair but would it be better to fail or bin to 1,1?

Best wishes,

Katherine MacArthur

thomasaarholt:
I would argue that it should fail. It does not give the user what they are
asking for.

k8macarthur:
@thomasaarholt Ok just finished my meeting. Try this. :) I've added an exception in both rebin functions (integer and linear) for me the exception appears in either:
s.rebin((10, 10)) raises an error
s.rebin((10, 10), crop=False) works perfectly
s.rebin((0.2, 10)) raises an error
s.rebin((10, 10), crop=False) works perfectly

francisco-dlp (approved):

thomasaarholt:
All good!

francisco-dlp:
One last thing: ``rebin`` should take both the old ``shape`` keyword argument (first) and the new ``scale`` argument: we cannot break the API in a minor release. Other than that it looks good to me.

thomasaarholt:
The docstring is a bit odd and sometimes incorrect. Here's a modified suggestion (starting from the current "rebin array"). I'm unsure the s.rebin() method gets it's docstring from a superclass or something, but perhaps it's best to just overwrite the %s?

```
Rebin the signal into a smaller or larger shape, based on a linear
interpolation. Specify **either** new_shape or scale. 

Scaling a dimension by 1 means no binning. A scale of less than one results in up-sampling.

Parameters
----------
new_shape : a list of floats or integer, default None
    For each dimension specify the new_shape. This will
    then be converted into a scale.
scale : a list of floats or integer, default None
    For each dimension specify the new:old pixel ratio, e.g. a ratio of 1
    is no binning and a ratio of 2 means that each pixel in the new
    spectrum is twice the size of the pixels in the old spectrum.
    The length of the list should match the dimension of the numpy array.
    ***Note : Only one of scale or new_shape should be specified otherwise
    the function will not run***
```

k8macarthur:
Agreed, I'm not sure if %S works or not. I've coped across the Docstrings for now. They can always be deleted and linked again as necessary...

thomasaarholt:
Multiple bugs here:
```python
s2 = s.rebin(scale = [2,2,2])
<ipython-input-8-657ee9d0f852> in <module>()
----> 1 s2 = s.rebin(scale = [2,2,2])

c:\users\thomasaar\hyperspy\hyperspy\_signals\eds.py in rebin(self, new_shape, scale, crop, out)
    203             scale = scale
    204 
--> 205         m = super().rebin(scale=scale, crop=crop, out=out)
    206         m = out or m
    207 

TypeError: rebin() got an unexpected keyword argument 'out'
```
and

```python
<ipython-input-11-7e5c8891e086> in <module>()
----> 1 s2 = s.rebin(new_shape= [100, 59,1024])

c:\users\thomasaar\hyperspy\hyperspy\_signals\eds.py in rebin(self, new_shape, scale, crop, out)
    194                             Not both.")
    195         elif new_shape != None:
--> 196             for axis in self.axes_manager._axis:
    197                 new_shape_in_array.append(
    198                     new_shape[axis.index_in_axis_manager])

AttributeError: 'AxesManager' object has no attribute '_axis'
```

k8macarthur:
So on my computer I've certainly fixed the second error. In the 1826ae0 commit. But Travis is failing and you're still getting the error. I'm not really sure what happened...

thomasaarholt:
I would remove "Rebin Array" in the very first line of the docstring as well. It doesn't fit with the style of the rest of the docstrings.

k8macarthur:
So the only thing failing now is:
    def test_rebin(self):
        self.signal.estimate_poissonian_noise_variance()
        new_s = self.signal.rebin(scale=(2, 2, 1))
        var = new_s.metadata.Signal.Noise_properties.variance
        assert new_s.data.shape == (1, 2, 6)
>       assert var.data.shape == (1, 2, 6)
E       assert (2, 2, 1) == (1, 2, 6)
E         At index 0 diff: 2 != 1
E         Use -v to get the full diff
but I'm not sure how the variance function works. Any ideas?

thomasaarholt:
Can't really look at it before the morning.

francisco-dlp:
``rebin`` should also rebin the ``variance`` in ``metadata`` if present. ``variance`` should have the same dimensions as the signal if it exists, so the operation consists only on running rebin with the same parameters on ``variace``.

francisco-dlp:
This is the line concerned https://github.com/hyperspy/hyperspy/blob/RELEASE_next_minor/hyperspy/signal.py#L2290

k8macarthur:
Ah, thanks, good spot. I missed that the rebin function calls itself...

francisco-dlp:
Why did you remove the lazy version?

k8macarthur:
Because as @to266  pointed out, the function no longer runs lazily. Therefore I thought it might be cleaner to remove it completely so that there's no confusion later on if someone wants to add this back in.

francisco-dlp:
Wasn't the idea to use the "old" rebin when applicable and the linear one otherwise? In that case, ``rebin`` could still run lazily when the problem falls in the scope of the old rebin. Being able to rebin big datasets lazily is not a feature that we would like to miss.

k8macarthur:
Ok, try this then... I don't understand enough about the Lazy stuff but I've put it back in with the same if statements infront. Might need some testing.

k8macarthur:
Tomorrow's a bank holiday weekend but I'll try and do everything I can for final tweaks on this function if you send them tonight.

francisco-dlp:
@k8macarthur, how is it going? Would you like a hand with this?

k8macarthur:
Yes please, it compiles and all the tests pass, I think. It's just the lazy that needs testing now.

Best wishes,

Katherine MacArthur

doc/user_guide/tools.rst
```diff
@@ -928,8 +927,61 @@ method that operates *in place*.
 Rebinning
 ^^^^^^^^^
 
-The :py:meth:`~.signal.BaseSignal.rebin` method rebins data in place down to a size
-determined by the user.
+Rebinning of the spectral data will be carried out by two methods.
+The fast method will run if the new_shape is a divisor of the original shape.
+Otherwise a slower linear interpolation method will be applied. Both methods are incorporated into the :py:meth:`~.signals.eds.rebin` function.
```
to266, line 932: Not groundbreaking, but it should probably not point to specifically `eds.rebin` method.

doc/user_guide/tools.rst
```diff
@@ -928,8 +927,61 @@ method that operates *in place*.
...
+
+Example of binning using linear interpolation:
+
+.. code-block:: python
+
+    >>> spectrum = hs.signals.EDSTEMSpectrum(np.ones([4, 4, 10]))
+    >>> spectrum.data[1, 2, 9] = 5
+    >>> print(spectrum)
+    <EDSTEMSpectrum, title: , dimensions: (4, 4|10)>
+    >>> print ('Sum = ', sum(sum(sum(spectrum.data))))
```
to266, line 961: Sum over all elements like that: `spectrum.data.sum()`

doc/user_guide/tools.rst
```diff
@@ -928,8 +927,61 @@ method that operates *in place*.
...
+    Sum =  164.0
+    >>> scale = [0.5, 0.5, 5]
+    >>> test = spectrum.rebin(scale=scale)
+    >>> print(test)
+    <EDSTEMSpectrum, title: , dimensions: (8, 8|2)>
+    >>> print('Sum =', sum(sum(sum(test.data))))
```
to266, line 967: Sum again

doc/user_guide/tools.rst
```diff
@@ -928,8 +927,61 @@ method that operates *in place*.
...
+    Sum =  164.0
+
+.. code-block:: python
+
+    >>> spectrum = hs.signals.EDSTEMSpectrum(np.ones([4, 4, 10]))
+    >>> spectrum.data[1, 2, 9] = 5
+    >>> print(spectrum)
+    <EDSTEMSpectrum, title: , dimensions: (4, 4|10)>
+    >>> print ('Sum = ', sum(sum(sum(spectrum.data))))
```
to266, line 976: sum

doc/user_guide/tools.rst
```diff
@@ -928,8 +927,61 @@ method that operates *in place*.
...
+    Sum =  164.0
+    >>> new_shape = [8, 8, 2]
+    >>> test = spectrum.rebin(new_shape)
+    >>> print(test)
+    <EDSTEMSpectrum, title: , dimensions: (8, 8|2)>
+    >>> print('Sum =', sum(sum(sum(test.data))))
```
to266, line 982: sum

hyperspy/_signals/eds.py
```diff
@@ -175,25 +177,61 @@ def sum(self, axis=None, out=None):
...
-    sum.__doc__ = Signal1D.sum.__doc__
+    def rebin(self, new_shape=None, scale=None, crop=True, out=None):
+        """
+        %s
+        """
+
+    #Series of if statements to check that only one out of new_shape or scale
+    #has been given. New_shape is then converted to scale. If both or neither
+    #are given the function raises and error and wont run.
+
```
to266, line 188: Throughout the function, comparisons if something is `None` should be done using `is` operator, like so:
```python
if new_shape is None and scale is not None:
    # do stuff
```

hyperspy/_signals/eds.py
```diff
@@ -175,25 +177,61 @@ def sum(self, axis=None, out=None):
...
+        if new_shape == None and scale == None:
+            raise ValueError("One of new_shape, or scale must be specified")
+        elif new_shape != None and scale != None:
+            raise ValueError("Only one out of new_shape or scale should be specified.\
+                            Not both.")
+        elif new_shape != None:
+            scale = []
+            for i, axis in enumerate(self.data.shape):
```
to266, line 196: This does not check that `new_shape` length is appropriate. 

I assume we require that full new shape is specified? I.e with signal shape `(5, 3, 2)` and `new_shape=[2, 1]` it fails? 
Alternatively we could assume that the untouched shapes are kept constant, or any surplus shape is ignored.

In the first case (fails) the code should be:
```python
#...
elif new_shape is not None:
    if len(new_shape) != len(self.data.shape):
        raise ValueError("Not enough dimensions in the new shape")
    scale = [datashape/newshape for datashape, newshape in zip(self.data.shape, new_shape)]
```

If we "pad" with ones and ignore the unnecessary, the code should be:
```python
elif new_shape is not None:
    lns = len(new_shape)
    lds = len(self.data.shape)
    if lns < lds:
        new_shape = tuple(new_shape) + self.data.shape[lns:]
    elif lns > lds:
        new_shape = new_shape[:lds]
    scale = [datashape/newshape for datashape, newshape in zip(self.data.shape, new_shape)]
```

hyperspy/_signals/eds.py
```diff
@@ -175,25 +177,61 @@ def sum(self, axis=None, out=None):
...
+                scale.append(self.data.shape[i]/new_shape[i])
+        else:
+            new_shape = new_shape
+            scale = scale
```
to266, line 200: Actually this piece of code gets repeated at least 3-4 times, and thus should be moved to a separate function. I would move it to `array_tools` module or some place similar (i.e. not within a class at all), and its signature should be something along the lines of (feel free to use a better function name):
```python
new_shape, scale = _rebin_preprocess(new_shape, scale, data_shape)
```
Here the parameters should be whatever the user entered plus the actual current data shape, and the results should be what you'd have at this point in the function where this comment is. The new function should have the necessary corrections according to my previous comments

to266 (changes requested):
I left a couple of comments regarding the code structure that should be addressed before merging. The changes should be fairly straightforward though

hyperspy/_signals/eds.py
francisco-dlp: I am working on this branch and I have addressed some of the code issues that you mention already. I'll push it soon, could you have a look? k8macarthur/hyperspy#12

francisco-dlp: pushed, thanks!

thomasaarholt:
Woohoo! Well done Kate and everyone :)

k8macarthur:
Excellent! Sorry I wasn't around for the final push.

PR #1311, opened 2016-11-01
Bugfix time format dm

thomasaarholt:
Fixed a bug where the different time formats in DigitalMicrograph files threw an error in Hyperspy. Thanks to @ericpre for tests and @magnunord for git help.

PR #1302, opened 2016-10-03
Fix #1293 (reading dm file with absent tags)

ericpre:
Fix #1293.
- Fix bug with dm plugin when 'Illumination_Mode' and/or 'Imaging_Mode' tags are not present in the 'Microscope_Info' tag tree.
- Add test with such file.

PR #1312, opened 2016-11-01
Bugfix time format dm v2

thomasaarholt:
This should be correct. Resubmit of #1311. @ericpre I had to manually copy out some code from your commit, as I wasn't sure what the debug stuff was for.


hyperspy/io_plugins/digital_micrograph.py
```diff
@@ -24,6 +24,7 @@
 import os
 import logging
 from datetime import datetime
```
magnunor: I think `datetime` can be removed?

hyperspy/io_plugins/digital_micrograph.py
```diff
@@ -785,11 +786,11 @@ def _get_mode(self, mode):
             return 'TEM'
 
     def _get_time(self, time):
-        dt = datetime.strptime(time, "%I:%M:%S %p")
+        dt = dateutil.parser.parse(time)
         return dt.time().isoformat()
```
magnunor: This solves the locale issue, but it will still throw an either `OverflowError` or `ValueError` if the `time` variable is wrong. So I think the parser should be in a `try` `except`, in case the `time` variable is something really weird. For example:

``` Python
try:
    dt = dateutil.parser.parse(time)
    return dt.time().isoformat()
except:
    _logger.warning("Time string, %s,  could not be parsed", time)
    return("") # Or whatever the default value should be
```

hyperspy/io_plugins/digital_micrograph.py
```diff
@@ -785,11 +786,11 @@ def _get_mode(self, mode):
...
 
     def _get_date(self, date):
-        dt = datetime.strptime(date, "%m/%d/%Y")
+        dt = dateutil.parser.parse(date)
         return dt.date().isoformat()
```
magnunor: Comment for `_get_time` also applies here.

magnunor:
Any progress on this? Really want to merge it since it affects a bunch of my datasets :)

I could implement it on my own based on your commits, but I'll let you do it if you want to @thomasaarholt.

hyperspy/io_plugins/digital_micrograph.py
ericpre: As mentioned in #1302, removing the `return` in the except loop will avoid setting a default to the metadata tree, in case this is not supported.

francisco-dlp:
What's the status of this one?

francisco-dlp:
Any comments regarding the status of this PR?

hyperspy/io_plugins/digital_micrograph.py
ericpre: I added a test in this [commit](https://github.com/hyperspy/hyperspy/pull/1302/commits/55a7d105a23debe51da6f0cf320e0d3cb0e767da) to confirm that returning `None` is working.

ericpre:
Once two magnunor's comments above are addressed, it should be ready to merge!

thomasaarholt:
Looking at it now!

thomasaarholt:
Done! Sorry this one took so long! I think this is ready to merge.

hyperspy/io_plugins/digital_micrograph.py
francisco-dlp: Don't you agree with @magnunor, can't `datetime` be removed?

magnunor:
I think `datetime` can be removed as an import, since we're using `dateutils` instead

thomasaarholt:
Done.

francisco-dlp:
Thanks @thomasaarholt for the contribution and @ericpre  and @magnunor for the review.

Issue #1301, opened 2016-10-02
align_zero_loss_peak single spectrum not working


magnunor:
`align_zero_loss_peak` does not work if the navigation dimensions is 0, ergo only a single spectrum.

``` Python
import hyperspy.api as hs
import numpy as np
s = hs.signals.EELSSpectrum(np.arange(0,1000))
s.align_zero_loss_peak() # Does not work

s = hs.signals.EELSSpectrum(np.arange(0,2000).reshape((2,1000)))
s.align_zero_loss_peak() # Works
```

Error message: 

``` Python
---------------------------------------------------------------------------
IndexError                                Traceback (most recent call last)
<ipython-input-20-f890f17f75fd> in <module>()
----> 1 s.align_zero_loss_peak()

/home/magnunor/Documents/HyperSpy/hyperspy/hyperspy/_signals/eels.py in align_zero_loss_peak(self, calibrate, also_align, print_stats, subpixel, mask, signal_range, show_progressbar, **kwargs)
    293             also_align=also_align,
    294             show_progressbar=show_progressbar,
--> 295             **kwargs)
    296         zlpc = self.estimate_zero_loss_peak_centre(mask=mask)
    297         if calibrate is True:

/home/magnunor/Documents/HyperSpy/hyperspy/hyperspy/_signals/signal1d.py in align1D(self, start, end, reference_indices, max_shift, interpolate, number_of_interpolation_points, interpolation_method, crop, expand, fill_value, also_align, mask, show_progressbar)
    650             number_of_interpolation_points=number_of_interpolation_points,
    651             mask=mask,
--> 652             show_progressbar=show_progressbar)
    653         for signal in also_align + [self]:
    654             signal.shift1D(shift_array=shift_array,

/home/magnunor/Documents/HyperSpy/hyperspy/hyperspy/_signals/signal1d.py in estimate_shift1D(self, start, end, reference_indices, max_shift, interpolate, number_of_interpolation_points, mask, show_progressbar)
    551                         dat = interpolate1D(ip, dat)
    552                     shift_array[indices] = np.argmax(
--> 553                         np.correlate(ref, dat, 'full')) - len(ref) + 1
    554                 pbar.update(1)
    555 

IndexError: too many indices for array
```

Reported at https://groups.google.com/forum/#!topic/hyperspy-users/VmgVelkCbiw

thomasaarholt:
[...]

PR #1313, opened 2016-11-01
Simple fix for ensuring align1D is not applied for single spectra

thomasaarholt:
Fixing #1301.

### DIFF
--- a/hyperspy/_signals/eels.py
+++ b/hyperspy/_signals/eels.py
@@ -287,12 +287,13 @@ class EELSSpectrum(Signal1D):
                 else self.axes_manager[-1].axis[0])
         right = (right if right < self.axes_manager[-1].axis[-1]
                  else self.axes_manager[-1].axis[-1])
-        self.align1D(
-            left,
-            right,
-            also_align=also_align,
-            show_progressbar=show_progressbar,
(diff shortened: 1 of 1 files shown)


hyperspy/_signals/eels.py
```diff
@@ -287,12 +287,13 @@ def estimate_zero_loss_peak_centre(s, mask, signal_range):
...
-            also_align=also_align,
-            show_progressbar=show_progressbar,
-            **kwargs)
+        if self.axes_manager.navigation_size > 1:
+            self.align1D(
+                left,
+                right,
+                also_align=also_align,
+                show_progressbar=show_progressbar,
+                **kwargs)
```
francisco-dlp, line 296: What about adding `else` write to the log that nothing was done because you can't align one single spectrum?

thomasaarholt: What does the rest of the group think? I feel like most people would expect `s.align_zero_loss_peak(True)` to set the zero loss peak at zero independently of the shape of the signal? I don't see a reason why this behaviour should _not_ be included.

francisco-dlp: Sorry, I misunderstood the point of your modification.

PR #1321, opened 2016-11-04
NEW: Linear fitting

thomasaarholt:
I've added support for both linear and non-negative linear fitting of components.
Example notebook here: https://nbviewer.jupyter.org/urls/dl.dropbox.com/s/t091ahzuduzrct6/Linear%20Fitting.ipynb

The fit from the normal non-linear `leastsq` is a little bit different, and in the notebook case the nonlinear one seems better. There was some discussion online about the tails of the gaussian making the fit unstable, and it being better to fit around the central part of the gaussians. Not sure how to go about doing that.
There's also a chance that there are better ways of doing this.

I'd like to get this working, and then talk about ways of having hyperspy be a bit smarter about which fit is chosen by default.

To Do:

- [ ] Create a system for checking if components are linear
- [ ] Optimise function
- [ ] Tests

### DIFF
--- a/hyperspy/model.py
+++ b/hyperspy/model.py
@@ -29,6 +29,8 @@ import scipy.odr as odr
 from scipy.optimize import (leastsq, least_squares,
                             minimize, differential_evolution)
 from scipy.linalg import svd
+from numpy.linalg import lstsq as linear
+
 from contextlib import contextmanager
 
 from hyperspy.external.progressbar import progressbar
@@ -937,10 +939,10 @@ class BaseModel(list):
 
             if method == 'ml':
                 weights = None
-                if fitter in ("leastsq", "odr", "mpfit"):
+                if fitter in ("leastsq", "odr", "mpfit", "linear"):
                     raise NotImplementedError(
                         "Maximum likelihood estimation is not supported "
-                        'for the "leastsq", "mpfit" or "odr" optimizers')
+                        'for the "leastsq", "mpfit", "odr" or "linear" optimizers')
             elif method == "ls":
                 metadata = self.signal.metadata
                 if "Signal.Noise_properties.variance" not in metadata:
@@ -1052,6 +1054,16 @@ class BaseModel(list):
                         (self._errfunc(self.p0, *args) ** 2).sum() /
                         (len(args[0]) - len(self.p0)))
                 self.fit_output = m
+
+            elif fitter == "linear":
+                signal_axis = self.axis.axis[np.where(self.channel_switches)]
+                component_data = np.array([component.function(signal_axis) /
+                                           for component in self if len(component.free_parameters) > 0])
+                output = linear(component_data.T, self.signal()[np.where(self.channel_switches)], **kwargs)
+
+                self.p0 = tuple([oldp0*factor for oldp0, factor in zip(self.p0, output[0])])
+                self.fit_output = output
+
             else:
                 # General optimizers
                 # Least squares or maximum likelihood
--- a/hyperspy/model.py
+++ b/hyperspy/model.py
@@ -838,6 +838,13 @@ class BaseModel(list):
                 "odr" performs the optimization using the orthogonal distance
                 regression algorithm. It does not support bounds.
 
+                "linear" performs least-squares fitting using linear regression. It does
+                not support bounds, but should be very fast.
+
+                "non_negative_linear" performs least-squares fitting using linear
+                regression (like "linear") under the constraint that none of the
+                solutions can be negative.
+
                 "Nelder-Mead", "Powell", "CG", "BFGS", "Newton-CG", "L-BFGS-B"
                 and "TNC" are wrappers for scipy.optimize.minimize(). Only
                 "L-BFGS-B" and "TNC" support bounds.
--- a/hyperspy/model.py
+++ b/hyperspy/model.py
@@ -30,6 +30,7 @@ from scipy.optimize import (leastsq, least_squares,
                             minimize, differential_evolution)
 from scipy.linalg import svd
 from numpy.linalg import lstsq as linear
+from scipy.optimize import nnls
 
 from contextlib import contextmanager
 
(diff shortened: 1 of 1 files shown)


hyperspy/model.py
```diff
@@ -29,6 +29,9 @@
 from scipy.optimize import (leastsq, least_squares,
                             minimize, differential_evolution)
 from scipy.linalg import svd
+from numpy.linalg import lstsq as linear
+from scipy.optimize import nnls
```
tjof2: this could go in line 29 as  `from scipy.optimize import (leastsq, least_squares, minimize, differential_evolution, nnls)`

thomasaarholt:
Right, I just tested the speed of linear fitting on a large dataset using numpy arrays rather than the hyperspy signals. In hyperspy, multifit with non-linear fitting takes 6 minutes. Multifit with the new linear fitting takes 1 minute and 30 sec. **Fitting using numpy linalg in one go takes less than one second**.

If we're serious about speed, it will be worth spending some time developing a multifit that let's everything happen in numpy and then construct the components again from the result. I'd like to ask for a brainstorming session to see if we can come up with some bright ways of doing this.

thomasaarholt:
I've just realised that I'm not taking into account any previous components in the model, when trying to fit using the linear fit. I need a way of subtracting the current fixed components from the signal to be fitted. Does a method for this already exist?

tjof2:
I'm sold on the idea of speed (to be expected I guess!) - a method for detecting linear models automatically would be good.

Can you qualify this statement a bit more?

> There was some discussion online about the tails of the gaussian making the fit unstable, and it being better to fit around the central part of the gaussians.

thomasaarholt:
By the way, I expect the nonlinear fitting to be much much faster as well when done through numpy/scipy. I would really appreciate some feedback on these ideas from those who came up with the fitting algorithms in the first place. Right now, this is my "go-to thesis distraction", for when writing becomes too much. But it would be nice if what I was doing was also sensible.

tjof2:
Some profiling might be needed to work out where most of that 6 minutes comes from.

thomasaarholt:
Could someone who knows how to do profiling reasonably quickly try it on a hyperspy multifit of some data?

tjof2:
https://github.com/rkern/line_profiler

francisco-dlp:
I think that the enhancement in performance is not surprising. Solving the LLS problem involves only matrix operations and performing all of then at once instead of iterating over all pixels is faster. LLS is naturally faster than NNLS and related non-linear optimisers. It would be possible to fit all the parameters at once using NNLS but, although I may be wrong, I don't think that this would boost performance as dramatically as in the case of LLS. Moreover, NNLS, unlike LLS, requires starting parameters in most cases which may be challenging to provide for the whole dataset (currently SAMFire addresses this issue when iterating over the pixels). That said, fitting the whole NNLS problem at once can be interesting for some applications. Another issue to consider is that fitting everything at once obviously requires more memory but, @to266, maybe dask solves this issue? In any case, I think that it makes sense to implement both global and per-pixel LLS. Automatising the detection of a linear model should be as simple as adding a linear attribute to the parameters except for the `Expression` component that would need a Python property instead to determine if the parameter is linear.

> I need a way of subtracting the current fixed components from the signal to be fitted. Does a method for this already exist?

No, it doesn't.

thomasaarholt:
Thanks @francisco-dlp!

to266:
While not for NLLS, dask does indeed have something to offer for large datasets (and, probably, few significantly fewer parameters than channels): [here](https://github.com/dask/dask/blob/master/dask/array/linalg.py#L711)

I have not tested this though... Should work like the numpy version, but out of core.

hyperspy/model.py
```diff
@@ -835,6 +837,13 @@ def fit(self, fitter=None, method='ls', grad=False,
                 "odr" performs the optimization using the orthogonal distance
                 regression algorithm. It does not support bounds.
 
+                "linear" performs least-squares fitting using linear regression. It does
+                not support bounds, but should be very fast.
+
+                "non_negative_linear" performs least-squares fitting using linear
```
tjof2, line 843: "performs least-squares fitting using linear regression" 

perhaps "performs linear least-squares fitting" would be better, since the distinction really is "LLS vs NLLS"

hyperspy/model.py
```diff
@@ -27,8 +27,10 @@
 import scipy
 import scipy.odr as odr
 from scipy.optimize import (leastsq, least_squares,
-                            minimize, differential_evolution)
+                            minimize, differential_evolution, nnls)
 from scipy.linalg import svd
+from numpy.linalg import lstsq as linear
```
tjof2, line 32: instead of `numpy.linalg.lstsq`, why not use SciPy's version which **does** support bounds? It would clean up the imports at the top of the file as well:

```
from scipy.optimize import (leastsq, least_squares,
                            minimize, differential_evolution, nnls, lsq_linear)
```

[Reference](https://docs.scipy.org/doc/scipy/reference/generated/scipy.optimize.lsq_linear.html).

tjof2: @thomasaarholt  See [lines 984-986](https://github.com/hyperspy/hyperspy/blob/RELEASE_next_minor/hyperspy/model.py#L984) in `model.py` for the approach to setting the bounds for `scipy.optimize.least_squares`, which is the same as here. Hopefully it isn't too complex a change.

tjof2:
```
ls_b = self.free_parameters_boundaries
ls_b = ([a if a is not None else -np.inf for a, b in ls_b],
        [b if b is not None else np.inf for a, b in ls_b])
```
and then `bounds = ls_b` as your argument, I think?

thomasaarholt:
I'm hoping to write up my thesis before christmas, so I won't be doing much (if any) more on this before then!

tjof2:
Anyone have ideas about checking for linearity?

tjof2:
Other than checking if `f(A + B) == f(A) + f(B)`?

thomasaarholt:
I think the idea was to introduce a `is_linear` == `True` or `False` tag to each component's parameters.

thomasaarholt:
And then before the fitting begins one would do something like
```python
if all_linear:
   # Fit using linear fitting
else:
   # Fit using nonlinear fitting

thomasaarholt:
Closed in favour of #1462.

PR #1322, opened 2016-11-07
NEW: HologramImage class, misc/holography package. Strong revision of #1152

woozey:
Finally, after long time, here is the revision of #1152: new PR for electron holography functionality also with contributions from @FWin22. Briefly about the revision:

- _reconstruct_phase_ method is split in several functions which operate on _nparrays_ which are placed in holography package (_misc.holography_)
- _HologramImage_ class contains methods which are warpers for the functions of holography package operating on the objects with _nav_dim>0_
- Reconstruction parameters are split in more feasible way: _sb_size, sb_position, sb_smooth_
- Improved: searching for sideband; smoothing of sideband filter.
- Added: possibility to input sideband filter size in physical units (i.e. mrad or 1/nm); possibility to change output shape.

@francisco-dlp Unfortunately, couldn't apply _map()_ since it works in place, whereas _reconstruct_phase_ method is able to change shape of the output. Also as we have discussed, I'll submit a PR to to hyperspy demos 

@CodeMonkeyJan is welcome to comment too))

Update: corresponding PR with demo https://github.com/hyperspy/hyperspy-demos/pull/29


thomasaarholt:
I too would love to see the demos for this :)

woozey:
Corresponding demo is in PR: https://github.com/hyperspy/hyperspy-demos/pull/29

hyperspy/_signals/electron_wave_image.py
```diff
@@ -0,0 +1,38 @@
...
+
+
+class ElectronWaveImage(ComplexSignal2D):
+
+    """ComplexSignal2D subclass for electron wave images."""
+
+    _signal_type = 'electron_wave'
+
+    @property
+    def reconstruction_parameters(self):
```
francisco-dlp: What's the use of this property? One could directly do the following:

``` python

im.reconstruct_phase(**metadata.Signal.holo_reconstruction_parameters.as_dictionary())

```

In any case, we try to avoid using tuples for this purpose because it is very easy to introduce a difficult to diagnose bug by getting the index of the value that you want wrong somewhere in the code. Wouldn't it be better to return a dictionary instead?

hyperspy/_signals/hologram_image.py
```diff
@@ -0,0 +1,274 @@
...
+        sb : str, optional
+            Chooses which sideband is taken. 'lower' or 'upper'
+
+        Returns
+        -------
+        Tuple of the sideband position (y, x), referred to the unshifted FFT.
+        """
+
+        if self.axes_manager.navigation_size:
+            self.unfold_navigation_space()
```
francisco-dlp: It would be better to use an `with` statement with `unfolded`, because if something goes wrong before folding back the signal will remain unfolded.

hyperspy/_signals/hologram_image.py
```diff
@@ -0,0 +1,274 @@
...
+            sb_position = np.zeros((self.axes_manager.navigation_size, 2), dtype='int64')
+            for i in range(self.axes_manager.navigation_size):
```
francisco-dlp: Ideally this should be performed with the `map` function, however, the `map` function does not yet support functions which return arrays with a different shape but the feature is implemented in #737. @to266, are you planning to resubmit #737? If yes, would it be reasonable to merge it into this branch?

In any case, why unfolding the navigation dimension for this operation? Wouldn't it be better that the navigation dimension of `sb_position` is equal to `self`'s?

hyperspy/_signals/hologram_image.py
```diff
@@ -0,0 +1,274 @@
...
+                sb_position[i] = find_sideband_position(self.inav[i].data, self.sampling, ap_cb_radius, sb)
+            self.fold()
+        else:
+            sb_position = find_sideband_position(self.data, self.sampling, ap_cb_radius, sb)
+        return sb_position
+
+    def find_sideband_size(self, sb_position):
+        """
+        Finds the position of the sideband and returns its position.
```
francisco-dlp: position -> size?

hyperspy/_signals/hologram_image.py
```diff
@@ -0,0 +1,274 @@
...
+
+        Parameters
+        ----------
+        sb_position : tuple
```
francisco-dlp: tuple -> array?

hyperspy/_signals/hologram_image.py
```diff
@@ -0,0 +1,274 @@
...
+            The sideband position (y, x), referred to the non-shifted FFT.
+
+        Returns
+        -------
+        Sideband size (y, x), referred to the unshifted FFT.
+        """
+        if isinstance(sb_position, BaseSignal):
```
francisco-dlp: If it also accepts `Signal`s this should be mentioned in the docstring.

hyperspy/_signals/hologram_image.py
```diff
@@ -0,0 +1,274 @@
...
+
+        Parameters
+        ----------
+        reference : ndarray, :class:`~hyperspy.signals.Signal2D
+            Vacuum reference hologram.
+        sb_size : float
+            Sideband radius of the aperture in corresponding unit (see 'sb_unit'). If None,
+            the radius of the aperture is set to 1/3 of the distance between sideband and
+            centerband.
+        sb_smooth : float, optional
```
francisco-dlp: Why isn't this called `smothness` instead?

hyperspy/_signals/hologram_image.py
```diff
@@ -0,0 +1,274 @@
...
+        Use wave.rec_param to extract reconstruction parameters, which can be used for batch
+        processing.
+        """
+
+        # Unfolding self:
+
+        # Parsing reference:
+        if not isinstance(reference, HologramImage):
+            if isinstance(reference, Signal2D):
+                warnings.warn('The reference image signal type is not HologramImage. It will be converted to '
```
francisco-dlp: We tend to add this sort of warnings to the logger instead.

hyperspy/_signals/hologram_image.py
```diff
@@ -0,0 +1,274 @@
...
+
+        return sb_size
+
+    def reconstruct_phase(self, reference=None, sb_size=None, sb_smooth=None, sb_unit=None,
+                          sb='lower', sb_position=None, output_shape=None, plotting=False):
+        """Reconstruct electron holograms.
+
+        Parameters
+        ----------
+        reference : ndarray, :class:`~hyperspy.signals.Signal2D
```
francisco-dlp: Is `None` a valid input too? If yes, it should be specified.

hyperspy/_signals/hologram_image.py
francisco-dlp: Are there any constraints in the dimensions of the reference?

hyperspy/_signals/hologram_image.py
```diff
@@ -0,0 +1,274 @@
...
+                sb_position = np.stack([sb_position[0]]*self.axes_manager.navigation_size)
+
+        if sb_size is None:  # Default value is 1/2 distance between sideband and central band
+            sb_size = self.find_sideband_size(sb_position)
+
+        folded = self.unfold_navigation_space()
+
+        # Standard edge smoothness of sideband aperture 5% of sb_size
+        if sb_smooth is None:
+            sb_smooth = sb_size * 0.05
```
francisco-dlp: A message with the assigned `sb_smooth` should be added to the log.

hyperspy/_signals/hologram_image.py
```diff
@@ -0,0 +1,274 @@
...
+
+        # Convert sideband size from 1/nm or mrad to pixels
+        if sb_unit == 'nm':
+            sb_size /= np.mean(self.f_sampling)
+            sb_smooth /= np.mean(self.f_sampling)
+        elif sb_unit == 'mrad':
+            try:
+                ht = self.metadata.Acquisition_instrument.TEM.beam_energy
+            except AttributeError:
```
francisco-dlp: This should instead raise an error explaining that the user have to define the beam energy first using `set_microscope_parameters`.

hyperspy/_signals/hologram_image.py
```diff
@@ -0,0 +1,274 @@
...
+                ht = int(input('Enter beam energy in kV: '))
+                self.metadata.add_node('Acquisition_instrument')
+                self.metadata.Acquisition_instrument.add_node('TEM')
+                self.metadata.Acquisition_instrument.TEM.add_node('beam_energy')
+                self.metadata.Acquisition_instrument.TEM.beam_energy = ht
+            wavelength = 1.239842447 / np.sqrt(ht * (1022 + ht))  # in nm
```
francisco-dlp: "magic numbers" should be avoided where possible or at least they should be accompanied by a comment explaining where they are coming from.

hyperspy/_signals/hologram_image.py
```diff
@@ -0,0 +1,274 @@
...
+            sb_size /= (1000 * wavelength * np.mean(self.f_sampling))
+            sb_smooth /= (1000 * wavelength * np.mean(self.f_sampling))
+
+        # Find output shape:
+        if output_shape is None:
+            output_shape = self.axes_manager.signal_shape
+
+        # ???
+        _logger.info('Sideband pos in pixels: {}'.format(sb_position))
```
francisco-dlp: These could be large arrays, do we want to add them to the log?

hyperspy/_signals/hologram_image.py
```diff
@@ -0,0 +1,274 @@
...
+        #     axs.scatter(sb_position[1], sb_position[0], s=20, color='red', marker='x')
+        #     axs.set_xlim(sb_position[1]-sb_size, sb_position[1]+sb_size)
+        #     axs.set_ylim(sb_position[0]-sb_size, sb_position[0]+sb_size)
+        #     plt.show()
+
+        # object electron wave:
+        if self.axes_manager.navigation_size:
+            wave_object = np.zeros((self.axes_manager.navigation_size, ) + output_shape, dtype='complex')
+            for i in range(self.axes_manager.navigation_size):
+                wave_object[i] = reconstruct(self.inav[i].data, holo_sampling=self.sampling,
```
francisco-dlp: Indexing using `.inav` and `.isig` should be avoided in loops because they have a big overhead.

hyperspy/_signals/hologram_image.py
```diff
@@ -0,0 +1,274 @@
...
+            elif sb_position.ndim == 2 and sb_position.shape[0] != self.axes_manager.navigation_size:
+                warnings.warn('The sideband position size does not match the size of the hologram image! '
+                              'The first values of sideband position will be used to reconstruct all slices'
+                              'of hologram image.')
+                sb_position = np.stack([sb_position[0]]*self.axes_manager.navigation_size)
+
+        if sb_size is None:  # Default value is 1/2 distance between sideband and central band
+            sb_size = self.find_sideband_size(sb_position)
+
+        folded = self.unfold_navigation_space()
```
francisco-dlp: As above regarding folding.

hyperspy/_signals/hologram_image.py
```diff
@@ -0,0 +1,274 @@
...
+                                             output_shape=output_shape, plotting=plotting)
+        else:
+            wave_object = reconstruct(self.data, holo_sampling=self.sampling,
+                                      sb_size=sb_size, sb_position=sb_position, sb_smoothness=sb_smooth,
+                                      output_shape=output_shape, plotting=plotting)
+        # Reconstructing reference wave and applying it (division):
+        folded_ref = False
+
+        if reference is None:
+            wave_reference = 1
```
francisco-dlp, line 347: Why not then adding this to the functions signature instead i.e. `(reference=1, ...)`?

hyperspy/_signals/hologram_image.py
```diff
@@ -0,0 +1,274 @@
...
+        else:
+            # reference electron wave
+            if reference.axes_manager.navigation_size:
+                folded_ref = reference.unfold_navigation_space()
```
francisco-dlp: The same issue with unfold mentioned above here.

hyperspy/_signals/hologram_image.py
```diff
@@ -0,0 +1,274 @@
...
+                wave_reference = np.zeros((reference.axes_manager.navigation_size, ) + output_shape, dtype='complex')
+                for i in range(reference.axes_manager.navigation_size):
+                    wave_reference[i] = reconstruct(reference.inav[i].data, holo_sampling=self.sampling,
+                                                    sb_size=sb_size[i], sb_position=sb_position[i], sb_smoothness=sb_smooth[i],
+                                                    output_shape=output_shape, plotting=plotting)
+
+                if reference.axes_manager.navigation_size != self.axes_manager.navigation_size:  # case when navdim of
+                    # reference is >0 and not equal to that of self:
+
+                    warnings.warn('The navigation size of the reference and the hologram do not match! Reference wave '
```
francisco-dlp: Wouldn't it be better to let the user decide in this case what to do? If the navigation sizes are different chances are that the user is doing something wrong but he may not realise it if he doesn't get an error.

hyperspy/misc/holography/reconstruct.py
```diff
@@ -0,0 +1,207 @@
...
+        sb_size : float
+            Size of the sideband filter in pixel.
+        sb_position : tuple
+            Sideband position in pixel.
+        sb_smoothness: float
+            Smoothness of the aperture in pixel.
+        output_shape: tuple, optional
+            New output shape.
+        plotting : boolean
+            Shows details of the reconstruction (i.e. SB selection).
```
francisco-dlp: I would be nice to describe in more detail what it is exactly plotting.

hyperspy/misc/holography/reconstruct.py
```diff
@@ -0,0 +1,207 @@
...
+            fig, axs = plt.subplots(1, 1, figsize=(4, 4))
+            axs.imshow(np.abs(fftshift(fft_shifted * aperture)), clim=(0, 0.1))
+            axs.scatter(sb_position[1], sb_position[0], s=10, color='red', marker='x')
+            axs.set_xlim(int(holo_size[0]/2) - sb_size/np.mean(f_sampling), int(holo_size[0]/2) +
+                         sb_size/np.mean(f_sampling))
+            axs.set_ylim(int(holo_size[1]/2) - sb_size/np.mean(f_sampling), int(holo_size[1]/2) +
+                         sb_size/np.mean(f_sampling))
+            plt.show()
+
+        fft_aperture = fft_shifted * aperture
```
francisco-dlp: What about moving this line above `if plotting` to enable using it inside the if statement?

hyperspy/misc/holography/reconstruct.py
```diff
@@ -0,0 +1,207 @@
...
+    """
+
+    h = np.array((np.asarray(sb_position) - np.asarray([0, 0]),
+                         np.asarray(sb_position) - np.asarray([0, holo_shape[1]]),
+                         np.asarray(sb_position) - np.asarray([holo_shape[0], 0]),
+                         np.asarray(sb_position) - np.asarray(holo_shape))) * sb_size_ratio
+    return np.min(np.linalg.norm(h, axis=1))
+
+
+def reconstruct(holo_data, holo_sampling, sb_size, sb_position, sb_smoothness, output_shape=None,
```
francisco-dlp, line 102: `find_sideband_size` and `find_sideband_position` iterate over `self` and are passed as arguments to `reconstruct` which also iterates over `self`. Wouldn't it be more efficient to include all three in the same loop?

francisco-dlp (changes requested):
Very nice contribution! The main issue with the code is that it has to workaround the current `map` limitation to output of the same dimension. However, as mentioned in one of the comments, there is a PR that relaxes this constraint. I would suggest (if @to266 agrees) to merge #737 into this branch and rewrite the code using the `map` function. I think that this should lead to a cleaner, more compact and faster code.

hyperspy/_signals/hologram_image.py
FWin22: Yes, it should be size.
``` python
    def find_sideband_size(self, sb_position):
        """
        Finds the size of the sideband and returns its position.
```

hyperspy/_signals/hologram_image.py
FWin22: Array makes sense.
```python
        sb_position : array
```

hyperspy/_signals/hologram_image.py
FWin22: How about:
```python
sb_position : array, :class:`~hyperspy.signals.BaseSignal
```

hyperspy/_signals/hologram_image.py
FWin22: Lets call it sb_smoothness.
```python
sb_smoothness : float, optional
```

hyperspy/_signals/hologram_image.py
FWin22: It was written below.
```python
_logger.info('Sideband aperture smoothness in pixels: {}'.format(sb_smoothness))
```

hyperspy/misc/holography/reconstruct.py
FWin22: At the moment it is plotting nothing, but we should think about what we want to plot.

FWin22: How about this:
```python
plotting : boolean
    Plots the masked sideband used for reconstruction.
```

hyperspy/_signals/hologram_image.py
FWin22: We could add the exact formula of relativistic electron wavelength using `scipy.constants`:

```python
momentum = 2 * constants.m_e * constants.elementary_charge * ht * 1000 \
                (1 + constants.elementary_charge * ht * 1000 / (
                2 * constants.m_e * constants.c ** 2))
wavelength = constants.h / np.sqrt(momentum)
```

hyperspy/_signals/electron_wave_image.py
FWin22: I think it is useful to have easy access to the properties, but having a dictionary would definitely be preferable. 

```python
@property
def reconstruction_parameters(self):
    assert self.metadata.Signal.has_item('holo_reconstruction_parameters'), \
        "No reconstruction parameters assigned to the wave"

    return self.metadata.Signal.holo_reconstruction_parameters.as_dictionary()
```

hyperspy/misc/holography/reconstruct.py
FWin22: Agreed.
```python
fft_aperture = fft_shifted * aperture
if plotting:
    fig, axs = plt.subplots(1, 1, figsize=(4, 4))
    axs.imshow(np.abs(fftshift(fft_aperture)), clim=(0, 0.1))
```

hyperspy/_signals/hologram_image.py
FWin22: Changed it to logger:
```python
_logger.warning('The reference image signal type is not HologramImage. It will '
                                'be converted to HologramImage automatically.')
```

woozey:
It seems that most of major changes that need to be done (including multiple loops and unfolding) depend on whether `map` will be used or not.
The #737 seem not been updated since release 1.1 was merged so obviously I've got conflicts trying to merge it in a new branch: woozey#10.
So I guess we have to wait for @to266 's response for a while, though I feel bit sad delaying this stuff...

francisco-dlp:
@woozey, hopefully @to266 will bring good news soon. I think that implementing it using `map` is really worth waiting a little bit: it should make the code more compact and robust and, furthermore, it'll benefit from all the multiprocessing and out of memory features coming soon to the `map` function.

to266:
@woozey I'll try to finish https://github.com/hyperspy/hyperspy/pull/1295 and then update https://github.com/hyperspy/hyperspy/pull/737 to be compatible (or probably depend on #1295) today. Once that's done, I'll try to have a look at this PR and suggest improvements to use these two previous PRs to simplify things and preferably make it run faster in some cases.

to266:
https://github.com/hyperspy/hyperspy/pull/1336

woozey:
The new version of Holography stuff is there. It uses `map` and all of the requests of @francisco-dlp should be now implemented.

hyperspy/_signals/hologram_image.py
```diff
@@ -0,0 +1,329 @@
...
+
+
+class HologramImage(Signal2D):
+    """Image subclass for holograms acquired via off-axis electron holography."""
+
+    _signal_type = 'hologram'
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.sampling = (self.axes_manager[0].scale, self.axes_manager[1].scale)
```
francisco-dlp: Doesn't this assume that the dataset contains just one single image? If yes, what about using `axes_manager.signal_axes` instead?

hyperspy/_signals/hologram_image.py
```diff
@@ -0,0 +1,329 @@
...
+        self.f_sampling = np.divide(1, [a * b for a, b in zip(self.data.shape, self.sampling)])
```
francisco-dlp: Again, you are assuming that this is a single image. You could access the size attribute of the signal axes instead .e.g `axes_manager.signal_axes[0].size`.

francisco-dlp: Beside, calculating this attributes in the constructor has an important drawback: if the user updates the calibration of the axes or simply crops the images, then this information is not updated. I would rather turn them into properties which are computed when accessed. These are very fast calculations, so it shouldn't affect the performance.

hyperspy/_signals/hologram_image.py
```diff
@@ -0,0 +1,329 @@
...
+            Chooses which sideband is taken. 'lower' or 'upper'
+        show_progressbar : boolean
+            Shows progressbar while iterating over different slices of the signal (passes the parameter to map method).
+
+        Returns
+        -------
+        Signal1D instance of sideband positions (y, x), referred to the unshifted FFT.
+        """
+
+        sb_position = self.deepcopy()
```
francisco-dlp: It's probably better (more memory efficient) to set the `in_place` kwarg of `map` to `False`.

hyperspy/_signals/hologram_image.py
```diff
@@ -0,0 +1,329 @@
...
+        sb_position.map(find_sideband_position, holo_sampling=self.sampling, central_band_mask_radius=ap_cb_radius,
+                        sb=sb, show_progressbar=show_progressbar)
+
+        # Workaround to a map disfunctionality:
```
francisco-dlp, line 116: What's the `map` dysfunctionality in question? We should try to fix it instead of working around it... In any case, most likely what you want is `sb_position.set_signal_type('')`

hyperspy/_signals/hologram_image.py
```diff
@@ -0,0 +1,329 @@
...
+        sb_position : :class:`~hyperspy.signals.BaseSignal
+            The sideband position (y, x), referred to the non-shifted FFT.
+        show_progressbar: boolean
+            Shows progressbar while iterating over different slices of the signal (passes the parameter to map method).
+
+        Returns
+        -------
+        Signal 1D instance with sideband size, referred to the unshifted FFT.
+        """
+        sb_size = sb_position.deepcopy()
```
francisco-dlp: As above.

hyperspy/_signals/hologram_image.py
```diff
@@ -0,0 +1,329 @@
...
+        if not isinstance(reference, HologramImage):
+            if isinstance(reference, Signal2D):
+                _logger.warning('The reference image signal type is not HologramImage. It will '
+                                'be converted to HologramImage automatically.')
+                reference.set_signal_type('hologram')
+            elif reference is not None:
+                reference = HologramImage(reference)
+
+        # Testing match of navigation axes of reference and self (exception: reference nav_dim=1):
+        if (reference and not reference.axes_manager.navigation_shape == self.axes_manager.navigation_shape
```
francisco-dlp, line 212: Maybe it's better to check this before changing the signal type automatically? The user may have provided the wrong signal...

hyperspy/_signals/hologram_image.py
```diff
@@ -0,0 +1,329 @@
...
+            Smoothness of the aperture in the same unit as sb_size.
+        sb_unit : str, None
+            Unit of the two sideband parameters 'sb_size' and 'sb_smoothness'.
+            Default: None - Sideband size given in pixels
+            'nm': Size and smoothness of the aperture are given in 1/nm.
+            'mrad': Size and smoothness of the aperture are given in mrad.
+        sb : str, None
+            Select which sideband is selected. 'upper' or 'lower'.
+        sb_position : tuple, :class:`~hyperspy.signals.Signal1D, None
+            Sideband position in pixel. If None, sideband is determined automatically from FFT.
```
francisco-dlp: It should mention the (y, x) convention.

hyperspy/_signals/hologram_image.py
```diff
@@ -0,0 +1,329 @@
...
+                          'cause wrong results.')
+            if reference is None:
+                sb_position = self.find_sideband_position(sb=sb)
+            else:
+                sb_position = reference.find_sideband_position(sb=sb)
+
+        else:
+            if not isinstance(sb_position, Signal1D):
+                sb_position = Signal1D(sb_position)
+
```
francisco-dlp, line 236: Maybe at this point it'll be good to check that `signal_dimension` is 1 and `signal_size` is 2?

hyperspy/_signals/hologram_image.py
```diff
@@ -0,0 +1,329 @@
...
+        if sb_position.axes_manager.navigation_size != self.axes_manager.navigation_size:
+            if sb_position.axes_manager.navigation_size:
+                warnings.warn('Sideband position dimensions do not match neither reference nor hologram dimensions.'
```
francisco-dlp: I would rather raise an error in this case as the input is incorrect.

hyperspy/_signals/hologram_image.py
```diff
@@ -0,0 +1,329 @@
...
+                if isinstance(sb_size, np.ndarray) and sb_size.size > 1:  # transpose if np.array of multiple instances
+                    sb_size = BaseSignal(sb_size).T
+                else:
+                    sb_size = BaseSignal(sb_size)
+
+        if sb_size.axes_manager.navigation_size != self.axes_manager.navigation_size:
+            if sb_size.axes_manager.navigation_size:
+                warnings.warn('Sideband size dimensions do not match neither reference nor hologram dimensions.'
+                              'The reconstruction will be performed with the first value.')
+                sb_size_temp = np.float64(sb_size.inav[0].data)
```
francisco-dlp: As above.

hyperspy/_signals/hologram_image.py
```diff
@@ -0,0 +1,329 @@
...
+
+        # Parsing sideband size
+        if sb_size is None:  # Default value is 1/2 distance between sideband and central band
+            if reference is None:
+                sb_size = self.find_sideband_size(sb_position)
+            else:
+                sb_size = reference.find_sideband_size(sb_position)
+        else:
+            if not isinstance(sb_size, BaseSignal):
+                if isinstance(sb_size, np.ndarray) and sb_size.size > 1:  # transpose if np.array of multiple instances
```
francisco-dlp, line 260: According to the docstring `sb_size` should not be an array. Here you consider the case when it is an array and assume that it needs transposing. Accepting only signals makes it easier to avoid issues arising from passing data with axes in the wrong order.

hyperspy/_signals/hologram_image.py
```diff
@@ -0,0 +1,329 @@
...
+            sb_size_temp = sb_size.deepcopy()
+        #
+
+        # Standard edge smoothness of sideband aperture 5% of sb_size
+        if sb_smoothness is None:
+            sb_smoothness = sb_size * 0.05
+        else:
+            if not isinstance(sb_smoothness, BaseSignal):
+                if isinstance(sb_smoothness, np.ndarray) and sb_smoothness.size > 1:
+                    sb_smoothness = BaseSignal(sb_smoothness).T
```
francisco-dlp, line 281: As above.

hyperspy/_signals/hologram_image.py
```diff
@@ -0,0 +1,329 @@
...
+                else:
+                    sb_smoothness = BaseSignal(sb_smoothness)
+
+        if sb_smoothness.axes_manager.navigation_size != self.axes_manager.navigation_size:
+            if sb_smoothness.axes_manager.navigation_size:
+                warnings.warn('Sideband smoothness dimensions do not match neither reference nor hologram dimensions.'
+                              'The reconstruction will be performed with the first value.')
+                sb_smoothness_temp = np.float64(sb_smoothness.inav[0].data)
```
francisco-dlp: As above.

hyperspy/_signals/hologram_image.py
```diff
@@ -0,0 +1,329 @@
...
+
+        # Convert sideband size from 1/nm or mrad to pixels
+        if sb_unit == 'nm':
+            sb_size_temp = sb_size_temp / np.mean(self.f_sampling)
+            sb_smoothness = sb_smoothness / np.mean(self.f_sampling)
+        elif sb_unit == 'mrad':
+            try:
+                ht = self.metadata.Acquisition_instrument.TEM.beam_energy
+            except AttributeError:
+                ht = int(input('Enter beam energy in kV: '))
```
francisco-dlp: I would rather raise an exception explaining that the information is missing and that it can be provided using `set_microscope_parameters`.

hyperspy/_signals/electron_wave_image.py
```diff
@@ -0,0 +1,36 @@
...
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with  HyperSpy.  If not, see <http://www.gnu.org/licenses/>.
+
+
+from hyperspy._signals.complex_signal2d import ComplexSignal2D
+
+
+class ElectronWaveImage(ComplexSignal2D):
```
francisco-dlp: Why adding it empty now then?

hyperspy/_signals/hologram_image.py
```diff
@@ -0,0 +1,329 @@
...
+                              'The reconstruction will be performed with the first value.')
+                sb_smoothness_temp = np.float64(sb_smoothness.inav[0].data)
+            else:  # sb_position navdim=0, therefore map function should not iterate it:
+                sb_smoothness_temp = np.float64(sb_smoothness.data)
+        else:
+            sb_smoothness_temp = sb_smoothness.deepcopy()
+
+        # Convert sideband size from 1/nm or mrad to pixels
+        if sb_unit == 'nm':
+            sb_size_temp = sb_size_temp / np.mean(self.f_sampling)
```
francisco-dlp: Is taking the mean of `f_sampling` the correct approach?

hyperspy/_signals/hologram_image.py
```diff
@@ -0,0 +1,329 @@
...
+        if output_shape is None:
+            # if sb_size.axes_manager.navigation_size > 0: #  Future improvement will give a possibility to choose
+            #     output_shape = (np.int(sb_size.inav[0].data*2), np.int(sb_size.inav[0].data*2))
+            # else:
+            #     output_shape = (np.int(sb_size.data*2), np.int(sb_size.data*2))
+            output_shape = self.axes_manager.signal_shape
+
+        # Logging the reconstruction parameters if appropriate:
+        if log_parameters:
+            _logger.info('Sideband position in pixels: {}'.format(sb_position))
```
francisco-dlp: Why is `log_parameters` optional? It doesn't cost much so why disabling it?

hyperspy/_signals/hologram_image.py
```diff
@@ -0,0 +1,329 @@
...
+            _logger.info('Sideband aperture radius in pixels: {}'.format(sb_size))
+            _logger.info('Sideband aperture smoothness in pixels: {}'.format(sb_smoothness))
+
+        # Shows the selected sideband and the position of the sideband
```
francisco-dlp: I would remove the commented out lines.

hyperspy/_signals/hologram_image.py
```diff
@@ -0,0 +1,329 @@
...
+        #     fft_holo = fft2(self.data) / np.prod(self.data.shape)
+        #     fig, axs = plt.subplots(1, 1, figsize=(4, 4))
+        #     axs.imshow(np.abs(fft_holo), clim=(0, 2.2))
+        #     axs.scatter(sb_position[1], sb_position[0], s=20, color='red', marker='x')
+        #     axs.set_xlim(sb_position[1]-sb_size, sb_position[1]+sb_size)
+        #     axs.set_ylim(sb_position[0]-sb_size, sb_position[0]+sb_size)
+        #     plt.show()
+
+        # Reconstructing object electron wave:
+        wave_object = self.deepcopy()
```
francisco-dlp: As above, use `in_place=False` instead.

hyperspy/_signals/hologram_image.py
```diff
@@ -0,0 +1,329 @@
...
+
+        # Reconstructing reference wave and applying it (division):
+        if reference is None:
+            wave_reference = 1
+        elif reference.axes_manager.navigation_size != self.axes_manager.navigation_size:  # case when reference is 1d
+            wave_reference = reference.deepcopy()
+
+            # Prepare parameters for reconstruction of the reference wave:
+
+            if reference.axes_manager.navigation_size != sb_position.axes_manager.navigation_size:  # 1d reference, but
```
francisco-dlp: I would rather check explicitly if nav_size is 0

hyperspy/_signals/hologram_image.py
```diff
@@ -0,0 +1,329 @@
...
+                # parameters are multidimensional
+                sb_position_ref = sb_position.inav[0].data
+            else:
+                sb_position_ref = sb_position_temp
+
+            if reference.axes_manager.navigation_size != sb_size.axes_manager.navigation_size:  # 1d reference, but
```
francisco-dlp: As above.

hyperspy/_signals/hologram_image.py
```diff
@@ -0,0 +1,329 @@
...
+                # parameters are multidimensional
+                sb_size_ref = np.float64(sb_size.inav[0].data)
+            else:
+                sb_size_ref = sb_size_temp
+
+            if reference.axes_manager.navigation_size != sb_smoothness.axes_manager.navigation_size:  # 1d reference, but
```
francisco-dlp: As above

hyperspy/_signals/hologram_image.py
```diff
@@ -0,0 +1,329 @@
...
+                                                       output_shape[0]
+        wave_image.axes_manager.signal_axes[1].scale = self.sampling[1] * self.axes_manager.signal_shape[1] / \
+                                                       output_shape[1]
+
+        # Reconstruction parameters are stored in holo_reconstruction_parameters:
+        rec_param_dict = OrderedDict([('sb_position', sb_position_temp), ('sb_size', sb_size_temp),
+                                      ('sb_units', sb_unit), ('sb_smoothness',
+                                                              sb_smoothness_temp)])
+
+        wave_image.metadata.Signal.add_node('holo_reconstruction_parameters')
```
francisco-dlp: I am not sure if it is a good idea to store the reconstruction parameters as they may be large. These days we tend to store only what is strictly needed for posterior analysis. I suppose that you want to store them for log purposes. We don't do this for any of the other methods to avoid adding clutter to the signal. Could at least this be optional with default `False`? Beside, what about storing it in `Signal.Holography.Reconstruction_parameters`? Notice that we capitalize node names.

hyperspy/tests/signal/test_hologram_image.py
```diff
@@ -0,0 +1,109 @@
...
+
+import numpy.testing as nt
+
+import hyperspy.api as hs
+from scipy.interpolate import interp2d
+
+
+class TestCaseHologramImage(object):
+
+    def setUp(self):
```
francisco-dlp, line 29: The tests concern only single images, could you modify it to test the multidimensional capabilities by e.g. reconstructing a (2, 3| x, y) nd-image?

hyperspy/_signals/hologram_image.py
```diff
@@ -0,0 +1,329 @@
...
+    """Image subclass for holograms acquired via off-axis electron holography."""
+
+    _signal_type = 'hologram'
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.sampling = (self.axes_manager[0].scale, self.axes_manager[1].scale)
+        self.f_sampling = np.divide(1, [a * b for a, b in zip(self.data.shape, self.sampling)])
+
+    def find_sideband_position(self, ap_cb_radius=None, sb='lower', show_progressbar=False):
```
francisco-dlp: What about `estimate_sideband_position` instead to emphasize the fact that this is just an estimation?

hyperspy/_signals/hologram_image.py
francisco-dlp: As above find -> estimate?

francisco-dlp (changes requested):
Other than the few comments in the review, there are a lot of calls to `deepcopy` and I think that not all of them are necessary. Calling `deepcopy` should be always avoided if possible.

hyperspy/_signals/hologram_image.py
woozey: Here I'd rather prefer keeping possibility to input arrays, therefore added corresponding changes in docstring. Transposition seem to be required in any case as `sb_size` signal dimension should be 0, and BaseSignal will always put all the dimensions of array into signal_dimension. The case when user messes up the navigation dimensions in the input array I'd leave as user's risk.

hyperspy/_signals/hologram_image.py
woozey: Here as well, would rather keep an option to use arrays...

hyperspy/_signals/hologram_image.py
woozey: @FWin22, could you check this please?

FWin22: In the way `f_sampling` is implemented, it gives the sampling along the image direction. As we only reconstruct with circular apertures and not ellipse-like apertures, there is no need to treat each direction by itself. Moreover, i cannot think of any example, where the sampling in both direction differ from each other. So we could either take the mean or just take the first entry of `f_sampling`

francisco-dlp: Are the images always square? If not, `f_sampling` would be different for each direction. If yes, then it'll probably be worth to check if that's the case and raise a warning otherwise.

By the way, does the user need to access the `sampling` and `f_sampling` attributes? If not, as they are only used by `reconstruct, wouldn't it be easier to simply calculate it inside this method?

FWin22: `sampling` is available via `axes_manager` anyway and `f_sampling` doesn't need to be accessed by the user, so we can move it within the reconstruct method.

hyperspy/_signals/hologram_image.py
woozey: The check is now made also before for the case if the input is HS signal. It is a little redundant in its current implementation but I think still less verbose than checking dimension of ndarray that can be an input as well.

hyperspy/_signals/hologram_image.py
woozey: Two reasons why I would leave it as it is at the moment:
1) It is faster for a user.
2) `set_microscope_parameters` doesn't seem to be a method available for all signals. In particular for me it seem it only works for `EDS_SEM`, even in `EELS` I get `NotImplementedError`. So it should be implemented in `HologramImage` first.

hyperspy/_signals/hologram_image.py
woozey: Initially that was done for the cases if larger stacks of holograms are reconstructed to have a possibility not to "spam" the log if it is needed. At the moment I did it the way you suggested.

hyperspy/_signals/hologram_image.py
francisco-dlp: That's a good point. We should probably have a `TEM` superclass from which all `TEM` signals inherit that includes `set_microscope_parameters`. It should work for EELS though. If you get a `NotImplementedError` is probably because you didn't initialize the backend before running the function e.g. `%matplotlib qt4`.

I don't think that it is a good idea to ask for input interactively. For example, [HyperSpyUI](http://hyperspy.org/hyperspyUI/) would struggle with it. If it raises an error, the user can then use `set_microscope_parameters` that includes a UI for HyperSpyUI. This is what we currently do for other methods that require missing, [here is an example](https://github.com/hyperspy/hyperspy/blob/RELEASE_next_minor/hyperspy/_signals/eels.py#L1050).

hyperspy/_signals/hologram_image.py
woozey: The intention here is that we like sometimes to check the feasibility of the parameters and to often to compare those across the datasets or different analyses trials. This actually needed as we most of the time use reconstruction method without providing parameters (the automatic routines seem to be very reliable). So will do both: make storing the parameters optional and inside node `.Holography`. In future defaults can be stored in global default of HS to provide users a possibility to enable or disable this feature for all analyses.

francisco-dlp: Sounds good. I am not sure if it is a good idea to add this to preferences. After some bad experiences with important parameters being set in `preferences`, I am currently of the opinion that we should not place there anything that affects the results or what's stored or not. (I know that this is not true for a number of parameters currently in preferences, but we should actually fix this.). I would rather leave it as an option with default to `True` if you think that it is important to keep this information, and add an entry to the log informing that it is storing this information.

hyperspy/_signals/hologram_image.py
woozey: I'm not sure what is wrong there with the `set_microsocpe_parameters` but here is what I've got either switching to qt4 backend or initialising my standard OS X backend:

> In[3]: %matplotlib qt4
In[4]: aa = hs.signals.EELSSpectrum(np.reshape(np.arange(16), (2, 2, 2, 2)))
In[5]: aa
Out[5]: <EELSSpectrum, title: , dimensions: (2, 2, 2|2)>
In[6]: aa.set_microscope_parameters(beam_energy=300)
  File "/Users/vadim/anaconda/lib/python3.5/site-packages/IPython/core/interactiveshell.py", line 2885, in run_code
  File "<ipython-input-6-2477a9aea25a>", line 1, in <module>
    aa.set_microscope_parameters(beam_energy=300)
  File "/Users/vadim/Work_Recent/Prog/_Py/hyperspy/hyperspy/_signals/eels.py", line 833, in set_microscope_parameters
    self._are_microscope_parameters_missing()
  File "/Users/vadim/Work_Recent/Prog/_Py/hyperspy/hyperspy/_signals/eels.py", line 793, in _are_microscope_parameters_missing
    is_ok = messagesui.information(par_str)
  File "/Users/vadim/Work_Recent/Prog/_Py/hyperspy/hyperspy/gui/messages.py", line 75, in information
    message.edit_traits()
  File "/Users/vadim/anaconda/lib/python3.5/site-packages/traits/has_traits.py", line 1859, in edit_traits
    handler, id, scrollable, args )
  File "/Users/vadim/anaconda/lib/python3.5/site-packages/traitsui/view.py", line 433, in ui
    ui.ui( parent, kind )
  File "/Users/vadim/anaconda/lib/python3.5/site-packages/traitsui/ui.py", line 222, in ui
    self.rebuild( self, parent )
  File "/Users/vadim/anaconda/lib/python3.5/site-packages/traitsui/toolkit.py", line 226, in ui_modal
    raise NotImplementedError
NotImplementedError
In[7]: aa.set_signal_type('EDS_SEM')
In[8]: aa
Out[8]: <EDSSEMSpectrum, title: , dimensions: (2, 2, 2|2)>
In[9]: aa.set_microscope_parameters(beam_energy=300)
In[10]: aa.metadata.Acquisition_instrument.SEM.beam_energy
Out[10]: 300

woozey: So what shall I do? I can simply copy a `set_microscope_parameters` method from `EDS_SEM` to `HologramImage` or can try creating `TEM` class and let `HologramImage` to inherit from it if there is a consensus about this...

woozey: Shall I post the above as an issue or is it a problem of my particular installation?

hyperspy/_signals/hologram_image.py
woozey: Actually the reconstruction is designed for square images only at the moment. I thought that it will deliver reasonable results for ( m, n | x, y) images where x=~y but it seems it doesn't really work. @FWin22 could you please check if `f_sampling` and other parts are capable for x!=y? Though I tend to restrict current version to square images only and do non-square once in next PR...

hyperspy/_signals/hologram_image.py
francisco-dlp: The exception is raised by traitsui, so, almost certainly it has to do not initializing the backend properly. Could try restarting the kernel and import hyperspy as follows (the order is important):

```python

%matplotlib qt4

import hyperspy.api as hs

```

Regarding `set_microscope_parameters`, for the moment you could simply copy the one in the EELS or SEM class and adapt it to your needs. We'll have to think on a more general solution to the problem, but it'll be probably better to leave this for a different PR.

woozey: This way it works, though with qt4 only...
Will copy one from `EDS_SEM` than.

hyperspy/_signals/hologram_image.py
FWin22: As far as i see it, there should be no error occuring when trying to reconstruct images with x != y, although the result will not be good. We have to adjust the aperture function to elliptic shape if we want to enable it. In addition `sb_smoothness` will be difficult to assess, as it then would depend on the direction. 
Can't we keep it as it is and either put it in the documentation that only squared images should be reconstructed or throw a warning when the image is non-square.

francisco-dlp: If I understand correctly, the options are i) to wait until the elliptic aperture is implemented ii) not to wait and live with the constraint. If you decide to go for two, I would suggest to raise a `NotImplementedError` explaining that the current code only works for square images / circular aperture and, of course, make this clear in the docstring and the User Guide.

hyperspy/_signals/electron_wave_image.py
woozey: Fulfilling yours and @FWin22 requests, added a method to display reconstruction parameters. More holography specific analyses will be released after this PR (not to overcomplicate this one), in particular calculation of charge density maps.

hyperspy/tests/signal/test_hologram_image.py
woozey: Done

hyperspy/_signals/hologram_image.py
woozey: I'v selected the 2nd option for now and started immediately to work on implementation of non-square images

woozey:
All the requested changes are made. I'd say it is ready for review again...

hyperspy/_signals/hologram_image.py
woozey: Actually found an issue, why it didn't work with non-square images. It seems that for `s` which is (|x, y) image

> s.axes_manager.signal_shape

results in `(x, y)`, but 

> s.data.shape

results in `(y, x)`. @francisco-dlp is it normal behaviour?

francisco-dlp: Yes, that's normal. HyperSpy uses x,y,... order unlike numpy, so the array axes are always in reverse order in respect to the signal axes.

hyperspy/_signals/electron_wave_image.py
```diff
@@ -0,0 +1,53 @@
...
+from hyperspy.signals import BaseSignal, Signal1D, ComplexSignal2D
+
+
+class ElectronWaveImage(ComplexSignal2D):
+
+    """ComplexSignal2D subclass for electron wave images."""
+
+    _signal_type = 'electron_wave'
+
+    def display_reconstruction_parameters(self):
```
francisco-dlp: I am happy to merge this PR except for this method that seems to me half cooked as it is not very useful for n-dimensional signals. I imagine that it would be more useful a method that displays the parameters using the markers on a multi-dimensional figure. I suggest that we leave this class (`ElectronWaveImage`) for another pull request. Notice that, as the reconstructed waves will have "electron_wave" signal type, they'll be assigned this class once it is ready to merge in another PR.

hyperspy/_signals/hologram_image.py
woozey: Actually, since aperture created using frequency array made with parameters of original data (in particular maintaining sampling in Fourier space), the aperture gets automatically elliptic in case of non-square images. This is how its top left part looks for (1024, 2048) image. The aspect ration looks good as well. @FWin22 do you expect any further troubles?
figure_1
In any case, I'll check the behaviour in more details and if it is the case will modify tests and documentation accordingly, so we should be able to release the reconstruction method with non-square image functionality in this PR.

hyperspy/_signals/electron_wave_image.py
woozey: ok

hyperspy/_signals/hologram_image.py
woozey: It now works perfectly for non-square images. There is a test reconstructing (3, 2 | 512, 768) hologram.

woozey:
It seem to be ready now) I've hopefully deserved some sleep))

doc/user_guide/electron_holography.rst
```diff
@@ -0,0 +1,92 @@
...
+(includes finding a side band in FFT, isolating and filtering it, recenter and calculate inverse Fourier transform)
+can be performed using :py:func:`~._signals.hologram_image.HologramImage.reconstruct_phase` method
+which returns a :py:class:`~._signals.electron_wave_image.ComplexImage2D` class and in future releases
+:py:class:`~._signals.electron_wave_image.ElectronWaveImage` class, containing the reconstructed
+electron wave. The `reconstruct_phase` method
+takes sideband position and size as parameters:
+
+.. code-block:: python
+
+    >>> wave_image = im.reconstruct_phase(sb_position=(<x>, <y>), sb_size=sb_radius)
```
francisco-dlp: Is it `(x, y)` or `(y, x)`? (The docstring suggests the latter.)

francisco-dlp:
I have tested  the reconstruction of non-square images and, although it seems to work, when comparing the result of cropping and reconstructing an image vs reconstructing and cropping the results are visually similar but the values are different. Moreover, the signal axes scales seem to be wrong in the crop and reconstruct approach xscale = 1.38, yscale=0.61 while they should be 0.92.

The following is adapted from the jupyter notebook demo above to reproduce the issues:

```python

import hyperspy.api as hs

holo0 = hs.load('./datasets/01_holo_Vbp_130V_0V_bin2_crop.hdf5', signal_type='hologram')
ref = hs.load('./datasets/00_ref_Vbp_130V_0V_bin2_crop.hdf5', signal_type='hologram')

rec_and_crop = holo0.reconstruct_phase(ref).isig[150:450, 150:350]
rec_and_crop.metadata.General.title = "R&C"
crop_and_rec = holo0.isig[150:450, 150:350].reconstruct_phase(ref.isig[150:450, 150:350])
crop_and_rec.metadata.General.title = "C&R"
difference = rec_and_crop - crop_and_rec
difference.metadata.General.title = "diff"
def interleave(lists):
    return [x for t in zip(*lists) for x in t]

images = [rec_and_crop, crop_and_rec, difference]
hs.plot.plot_images(interleave([(im.real, im.imag) for im in images]), cmap="RdBu_r", saturated_pixels=0.1, per_row=3, axes_decor="off", colorbar="single")

```
download

hyperspy/_signals/hologram_image.py
```diff
@@ -0,0 +1,392 @@
...
+
+                _logger.warning('The reference image signal type is not HologramImage. It will '
+                                'be converted to HologramImage automatically.')
+                reference.set_signal_type('hologram')
+            elif reference is not None:
+                reference = HologramImage(reference)
+
+        # Testing match of navigation axes of reference and self (exception: reference nav_dim=1):
+        if (reference and not reference.axes_manager.navigation_shape == self.axes_manager.navigation_shape
+            and reference.axes_manager.navigation_size):
```
francisco-dlp, line 213: Shouldn't we also check if the signal shape of the reference image matches the one of `self`?

woozey: Done!

woozey:
Regarding the test you made and the fact that difference is not 0, this is an expected behaviour to me. It arrises due to cropping affects sampling of Fourier space. Since the reconstruction involves  FFTing, filtering and IFFTing the result wouldn't be exactly the same. Though the artefacts are random numerical errors and as expected strongest deviations are at the edges.
Just in case I've made the same operations in Gatan's Digital Micrograph HoloWorks which can be accepted as a benchmark for hologram reconstruction. Here is an intensity of a similar difference which I got using DM. As one sees it is also not zero.
>Untitled: Minimum real      -8.54458 - 5.21992 i at (2,0)
Untitled: Minimum imaginary -0.172232 - 5.64608 i at (115,0)
Untitled: Maximum real      4.54949 + 0.451886 i at (127,0)
Untitled: Maximum imaginary -1.24891 + 3.2533 i at (0,506)`
screen shot 2017-01-04 at 0 23 03

tjof2:
@woozey "random numerical errors" or a result of floating point rounding errors? The two are different after all! :)

woozey:
> @woozey "random numerical errors" or a result of floating point rounding errors? The two are different after all! :)

The second one is definitely there, haven't thought it into deep to answer more precise. "random numerical errors" is probably too strong and undefined statement so I'd take it back.)) The solid statement would be that the Fourier space sampling of these two cases is different, which would be the main source of the mismatch. I assume that if one does the same comparison in Fourier space instead, the result should be the same once the sampling issue is fixed (I guess that would require interpolation of  `rec_and_crop`)...

francisco-dlp:
Thanks for the explanation and the fixes. The relative error is rather big even far from the borders (see images below). Nothing can be done about it?

In any case this is almost ready to merge. As far as I can tell, the only remaining issues are:

* [Making coveralls happy](https://coveralls.io/builds/9506930/source?filename=hyperspy%2F_signals%2Fhologram_image.py) (currently coverage is below 70%)
* Removing the `sampling` and `f_sampling` properties as discussed in the comments.

```python

(((crop_and_rec.amplitude - rec_and_crop.amplitude) / rec_and_crop.amplitude) * 100).isig[250.:350., 400.:500.].plot(cmap="RdBu_r", saturated_pixels=0.2)

```
amp



```python
(crop_and_rec.unwrapped_phase() - rec_and_crop.unwrapped_phase()).isig[250.:350., 400.:500.].plot(cmap="RdBu_r", saturated_pixels=0.5)
```
phase

```python

hs.plot.plot_images([crop_and_rec.amplitude.isig[250.:350., 400.:500.], rec_and_crop.amplitude.isig[250.:350., 400.:500.]], axes_decor="off", colorbar="single")

```
two_amp

woozey:
@francisco-dlp Coveralls seem to be happy now >90% and both sampling properties are removed.
I gave a thought to the issue with cropping and reconstruction. It doesn't seem to have any good solution to it. The two issues are change of Fourier space sampling and the second is the change of the filter shape once we use non-squre images. The fact that the reconstructed waves look similar says already something about consistency. I did some more checks and it looked not too bad taking into account that we are dealing with experimental data with strong noise. I started the same way as you didm but took square crops instead:
> rec_and_crop1 = holo0.reconstruct_phase(ref).isig[150:450, 150:450]
rec_and_crop1.metadata.General.title = "R&C"
crop_and_rec1 = holo0.isig[150:450, 150:450].reconstruct_phase(ref.isig[150:450, 150:450])
crop_and_rec1.metadata.General.title = "C&R"
difference1 = rec_and_crop1 - crop_and_rec1
difference1.metadata.General.title = "diff"

Than cropped even more to get rid of FFT-IFFT artefacts at the edges:
>In[13]: xstart = np.int((450-150)*0.1)
In[14]: xstop = np.int((450-150)*0.9)
In[15]: ystop = xstop
In[16]: ystart = xstart
In[17]: rec_n_crop_cp = rec_and_crop1.isig[xstart:xstop, ystart:ystop]
In[18]: crop_n_rec_cp = crop_and_rec1.isig[xstart:xstop, ystart:ystop]
In[20]: import numpy.testing as nt
In[23]: np.mean(rec_n_crop_cp.data)
Out[23]: (-0.24330612868092058+0.081784181981309378j)
In[24]: nt.assert_almost_equal(rec_n_crop_cp.data, crop_n_rec_cp.data, decimal=1)
In[25]: nt.assert_almost_equal(rec_n_crop_cp.data, crop_n_rec_cp.data, decimal=2)
...
AssertionError: 
Arrays are not almost equal to 2 decimals
(mismatch 23.99652777777777%)
 ...
In[26]: np.max(rec_n_crop_cp.data)
Out[26]: (1.6848738718898648+0.66547146165325688j)
In[27]: np.min(rec_n_crop_cp.data)
Out[27]: (-2.150320404829611+0.24154919648817227j)
In[28]: np.std(rec_n_crop_cp.data)
Out[28]: 0.72356959296812662

So the mean value of the wave is around 0, but it seem to vary a lot in +- range. The standard deviation is about 0.7, telling that being almost equal with 1 decimal is indeed not great result. Though for 2 decimal equality the mismatch is only 24% which I think isn't too bad. Again taking into account the noise in data and that no dud pixels were removed (bright spot in top left corner of your images, duds will be handled differently having different filter-pixel size relation, which is a consequence of different Fourier sampling)

Additionally I checked how similar is phase and got 0.5% mismatch for 1 decimal comparison, which seem decent to me as the phase noise of this particular experiment should be around 0.1 rad:
> nt.assert_almost_equal(rec_n_crop_cp.phase.data, crop_n_rec_cp.phase.data, decimal=1)
...
AssertionError: 
Arrays are not almost equal to 1 decimals
(mismatch 0.44097222222222854%)
...

So my conclusion, that there is not much we can do about it without doing interpolation in Fourier space while reconstructing. Removing dud pixels may slightly help (I have a function which can do the removal and I can submit it in next PR in Signal2D, I guess). Nevertheless the mismatch doesn't look dramatic to me at least the case of square images (the non-square once look also similar).

francisco-dlp (approved):

PR #1336, opened 2016-11-22
signal map with different shape outputs

to266:
Enables `s.map` to be used with functions that return differently-shaped arrays or arbitrary objects. (successor of https://github.com/hyperspy/hyperspy/pull/737, depends on https://github.com/hyperspy/hyperspy/pull/1295)

If outputs of the function are not of constant shapes throughout the navigation, the signal data is replaced by an array (navigation-shaped) of arrays (the results of the function, of any type) and signal axes are removed. As a result, most of the functions are not able to operate of this structure, and the data should be accessed directly. A convenient shortcut when slicing the navigation is also provided.

__Note:__ `map` still operates in place and replaces the data.

```python
In [1]: import scipy.ndimage

In [2]: import numpy as np

In [3]: import hyperspy.api as hs

In [4]: im = hs.signals.Signal2D(np.arange(0., 18).reshape((2,3,3)))

In [5]: angles = hs.signals.BaseSignal([0, 45])

In [6]: im.map(scipy.ndimage.rotate, angle=angles.T, reshape=True)
100%|██████████████████████████████████████████████████| 2/2 [00:00<00:00, 2884.67it/s]

In [7]: im.data.shape
Out[7]: (2,)

In [8]: im.data.dtype
Out[8]: dtype('O')

In [9]: im
Out[9]: <BaseSignal, title: , dimensions: (2|)>

In [10]: im.data
Out[10]: 
array([ array([[ -4.16333634e-16,   1.00000000e+00,   2.00000000e+00],
       [  3.00000000e+00,   4.00000000e+00,   5.00000000e+00],
       [  6.00000000e+00,   7.00000000e+00,   8.00000000e+00]]),
       array([[  0.        ,   0.        ,   0.        ,   0.        ],
       [  0.        ,  10.34834957,  13.88388348,   0.        ],
       [  0.        ,  12.11611652,  15.65165043,   0.        ],
       [  0.        ,   0.        ,   0.        ,   0.        ]])], dtype=object)

In [11]: # the shortcut
In [12]: im.inav[0].data
Out[12]: 
array([[ -4.16333634e-16,   1.00000000e+00,   2.00000000e+00],
       [  3.00000000e+00,   4.00000000e+00,   5.00000000e+00],
       [  6.00000000e+00,   7.00000000e+00,   8.00000000e+00]])

```


woozey:
It seems that the `axes_manager` doesn't get updated:

> In[41]: im = hs.signals.Signal2D(np.arange(0., 32).reshape((2,4,4)))
In[42]: im
Out[42]: <Signal2D, title: , dimensions: (2|4, 4)>
In[43]: im.map(np.reshape, newshape=(2, 8))
In[44]: im
Out[44]: <Signal2D, title: , dimensions: (2|4, 4)>
In[45]: im.data.shape
Out[45]: (2, 2, 8)

It should be an easy fix something like:

> im.axes_manager.signal_axes[0].size = im.inav[0].data.shape[0]
im.axes_manager.signal_axes[1].size = im.inav[0].data.shape[1]
im.axes_manager.signal_axes[0].scale = original_scale[0] * self.axes_manager.signal_shape[0] / \ im.inav[0].data.shape[0]
im.axes_manager.signal_axes[1].scale = original_scale[1] * self.axes_manager.signal_shape[1] / \
im.inav[1].data.shape[1]

where `original_scale[i] = im.axes_manager.signal_axes[i].scale` before map
The solution above will fail for `navdim=0 `, though...

francisco-dlp:
Is this WIP or ready for review?

to266:
Ready for review

to266:
only the usual samfire appveyor test fail left

woozey:
Philosophical question: what is the purpose of `map` method in general, i.e. why it operates in place? It would be nice to have a possibility to apply **any** function operating on nparrays using `map` even when signal dimensionality changes which to my understanding contradicts with `map` operating in place...
This is what I get trying to change signal dimensionality:

> In[2]: ss = np.arange(9).reshape((3,3))
In[3]: np.mean(ss, axis=1)
Out[3]: array([ 1.,  4.,  7.])
In[4]: sss = hs.signals.Signal2D(ss)
In[5]: sss
Out[5]: <Signal2D, title: , dimensions: (|3, 3)>
In[6]: sss.map(np.mean, axis=1)
  0%|          | 0/1 [00:00<?, ?it/s]
100%|██████████| 1/1 [00:00<00:00, 9258.95it/s]
  File "/Users/vadim/anaconda/lib/python3.5/site-packages/IPython/core/interactiveshell.py", line 2885, in run_code
  File "<ipython-input-6-d408abf397f8>", line 1, in <module>
    sss.map(np.mean, axis=1)
  File "/Users/vadim/Work_Recent/Prog/_Py/hyperspy/hyperspy/signal.py", line 3305, in map
    **kwargs)
  File "/Users/vadim/Work_Recent/Prog/_Py/hyperspy/hyperspy/signal.py", line 3455, in _map_iterate
    self.get_dimensions_from_data()
  File "/Users/vadim/Work_Recent/Prog/_Py/hyperspy/hyperspy/signal.py", line 2081, in get_dimensions_from_data
    axis.size = int(dc.shape[axis.index_in_array])

to266:
The in-place operation is somewhat historical and quite simple - to not require any more memory for the results. If I'm not mistaken, @pburdet was the original author of the function that eventually grew and became more complex. At that time he could essentially only have 2 copies of the data in the machine with maxed-out specs. So the workflow was complicated as it was (manually deleting and `gc`ing things).

Once this happened, a certain api was set, and now the default is to support it. In the previous iteration of this method (#737) I __did__ have an `out` argument, but I'm not sure it changes much - you can create a deepcopy of the signal before maping the function to get the same result.

I'll fix the leftover bug, thanks for spotting

francisco-dlp:
Is this ready for review? If yes, could you update it with the current RELEASE_next_minor to ease the task?

francisco-dlp:
By the way, the SAMFire test is failing more often than not... Any idea on how to fix it?

to266:
upstream merged. I'll try to fix the samfire thing in a separate PR..

doc/user_guide/tools.rst
```diff
@@ -784,6 +784,33 @@ arguments as in the following example.
   Rotation of images using :py:meth:`~.signal.BaseSignal.map` with different
   arguments for each image in the stack.
```
francisco-dlp, line 786: .. versionadded:: 1.2.0?

hyperspy/signal.py
```diff
@@ -3199,16 +3199,18 @@ def map(self, function,
             parallel=None, **kwargs):
         """Apply a function to the signal data at all the coordinates.
 
-        The function must operate on numpy arrays and the output *must have the
-        same dimensions as the input*. The function is applied to the data at
+        The function must operate on numpy arrays, but the output no longer has
```
francisco-dlp: I would cut it short: "The function must operate on numpy arrays." as new users are not interested in the history of the function.

hyperspy/signal.py
```diff
@@ -3199,16 +3199,18 @@ def map(self, function,
...
-        progress.
+        this method operates *in-place* and overwrites the data. Any extra
+        keyword argument is passed to the function. The keywords can take
+        different values at different coordinates. If the function takes an
+        `axis` or `axes` argument, the function is assumed to be vectorial and
+        the signal axes are assigned to `axis` or `axes`.  Otherwise, the
+        signal is iterated over the navigation axes and a progress bar is
+        displayed to monitor the progress.
+
+        Navigation dimensions and information is always preserved.
```
francisco-dlp: information? do you mean metadata?

francisco-dlp (changes requested):
Except for the minor comments this looks good to me. Following the discussion with @woozey above, I think that it would make sense to add an `inplace` keyword because, although one could use `deepcopy`, there are cases in which deepcopying requires more memory than internally callin `_deepcopy_with_new_data`.

hyperspy/signal.py
to266: I mean calibration and axes. Signal axes are not guaranteed to be preserved, and with odd examples (like the one @woozey suggested) they will get mixed up "incorrectly". Essentially if the number of signal axes changes, it's up to the user to figure out which is which, and the calibrations for them.

francisco-dlp: I see. What about "Navigation dimension and axes attributes are always preserved."?

francisco-dlp (approved):

francisco-dlp:
There is one more thing that I've just noticed that is easier to explain with a little example:

```python

import hyperspy.api as hs
import numpy as np
s = hs.signals.Signal1D(np.zeros((10, 10)))
def test(d, i):
    return d[:-i]
s.map(test, i=hs.signals.BaseSignal(np.arange(10)).T)

```
When indexing the navigation space I get:

```python
>>> s.inav[5]
<BaseSignal, title: , dimensions: (5|)>

```

But I would expect to get `<BaseSignal, title: , dimensions: (|5)>` instead.

to266:
Huh, that's a different bug than you expect, but a bug nevertheless.

 By design such slice will never have any signal axes (since the output shapes were different, and we can't know what would the calibrations be, hence we just remove signal axes).

How about such behavior (using your example)?
```python
In [6]: s.inav[5]
Out[6]: array([ 0.,  0.,  0.,  0.,  0.])

In [7]: s.inav[3]
Out[7]: array([ 0.,  0.,  0.,  0.,  0.,  0.,  0.])

```
Since we can't know anything about the axes and remove them (and the data in general could be any python object), we just return the thing directly. If it's a slice, the navigation information is still kept:

```python
In [3]: s = hs.signals.Signal1D(np.zeros((10,10)))

In [4]: the_sentence = 'This is some example text to slice. Does not matter what is inside'

In [10]: def test(d,i):
    ...:     return ' '.join(the_sentence.split()[:i])
In [11]: s.map(test, i=hs.signals.BaseSignal(np.arange(10)).T)
100%|██████████████████████████████████████████████████████████████████████████████| 10/10 [00:00<00:00, 12775.83it/s]
In [12]: s.data
Out[12]: 
array(['', 'This', 'This is', 'This is some', 'This is some example',
       'This is some example text', 'This is some example text to',
       'This is some example text to slice.',
       'This is some example text to slice. Does',
       'This is some example text to slice. Does not'], dtype=object)

In [13]: s
Out[13]: <BaseSignal, title: , dimensions: (10|)>

In [14]: s.inav[:4]
Out[14]: <BaseSignal, title: , dimensions: (4|)>

In [15]: s.inav[:4].data
Out[15]: array(['', 'This', 'This is', 'This is some'], dtype=object)

In [16]: s.inav[4]
Out[16]: 'This is some example'

**Comment by francisco-dlp** [comment]
:) That's a funny and clever example. I agree that what you propose is better for the case when what's in the array is not a numeric array. However, for the case in which what's inside is a numeric array (possibly more relevant to HyperSpy users), I think that it is useful to return a signal. Also, in this way, the navigation information would be preserved.

**Comment by to266** [comment]
But then we have to completely get rid of signal axes calibrations, etc, and just return a new `BaseSignal` as if it was just created from the array.

One example:
```python
In [31]: def test(d,i):
    ...:     return np.tile(d, (1,)*(i+1))
```
then the shapes are like so
```python
In [34]: [d.shape for d in s.data]
Out[34]: 
[(10,),
 (1, 10),
 (1, 1, 10),
 (1, 1, 1, 10),
 (1, 1, 1, 1, 10),
 (1, 1, 1, 1, 1, 10),
 (1, 1, 1, 1, 1, 1, 10),
 (1, 1, 1, 1, 1, 1, 1, 10),
 (1, 1, 1, 1, 1, 1, 1, 1, 10),
 (1, 1, 1, 1, 1, 1, 1, 1, 1, 10)]
```
and so each navigation pixel requires axes. 

Even if the number of axes stayed the same, the mapped function could have a transpose, meaning the axes get flipped/mixed up in general case.

francisco-dlp:
Wouldn't it be possible to create the right number of "default" signal axes when indexing to workaround the issue?

to266:
It is, but then in general there is no way to determine where to put the "original" axes, so might as well just replace all of them

francisco-dlp:
I fully agree.

to266:
Then this should do! :)

francisco-dlp:
Beautiful! A little test for the feature and it should be ready to merge.

to266:
On the way found another typo/bug ...

francisco-dlp:
It looks like travis is enjoying a break. I've tested it locally and it's fine, so I'll merge it.

woozey:
One more bug(
#1351

PR #1340, opened 2016-11-27
DOC: Indenting all non-major version changes except the latest

thomasaarholt:
I've had a go at tidying up the TOC of the documentation. I didn't spent long understanding the syntax of .rst files, but it's a minor change and I think it'll be alright, although it is currently "upside down", since headings normally are on top, but here we keep adding the newest version to the top. (So it is 1.2, 1.1, 1.0, rather than 1.0, 1.1, 1.2.)

The only intention here is to make the TOC more readable for newcomers, nothing else.

### DIFF
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -28,20 +28,20 @@ Enhancements
   supported file formats when available.
 
 v1.1
-====
+---
 
 This is a minor release. Follow the following links for details on all
 the `bugs fixed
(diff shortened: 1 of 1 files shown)


magnunor:
Sounds good! Did you run sphinx to make sure the end result looks nice?

thomasaarholt:
No. I really have no idea what I'm doing - I found an online "live" editor for rst, and pasted in there and it looked okay.

magnunor:
The webpage can be made using `make html` in the `hyperspy/doc` folder. But I'm not certain how to set this up in OS X. There is a sphinx package in Anaconda.

Issue #1347, opened 2016-12-02
Issues with SAMFire


thomasaarholt:
Trying to speed up fitting with parallelising with SAMFire and ipyparallel now. Ran into the following issues:

- `samf = m.create_samfire()` on a model with only a background: `AttributeError: 'DictionaryTreeBrowser' object has no attribute 'Noise_properties'`
Full error: https://gist.github.com/thomasaarholt/f2afa78e75327b0f43415ba5fa737944
- `samf = m.create_samfire()` on a model with a "homemade" gaussian added to many others: `ValueError: ('The length of the parameter must be ', 7)`. The error only appears when I add the "unknown" gaussian.
Full error: https://gist.github.com/thomasaarholt/70f6876923af7c2467b5f25e49cb1064

thomasaarholt:
ipyparallel is running fine in the terminal

to266:
I was aware of the first issue, however have not PRed a fix for this yet.

Just figured out the second problem as well, will all these fixes in a minute

thomasaarholt:
Now I get this guy:
https://gist.github.com/thomasaarholt/69fffed324d5e2e6fef575b26f76bc3c

This is after running `ipycluster start` in another terminal window (successful), estimating the noise on the modelled signal, and then fitting 10 different locations with `m.fit()`.

to266:
This is again not samfire, but model slicing bug. Can you reproduce it with the following?
```python
In [1]: import hyperspy.api as hs
QGtkStyle could not resolve GTK. Make sure you have installed the proper libraries.

In [2]: import numpy as np

In [3]: s = hs.signals.Signal1D(np.random.rand(10, 10, 100))

In [4]: s
Out[4]: <Signal1D, title: , dimensions: (10, 10|100)>

In [5]: m = s.create_model()

In [6]: m.append(hs.model.components1D.Lorentzian())

In [7]: m.append(hs.model.components1D.Lorentzian())

In [8]: m.print_current_values()
Components	Parameter	Value
Lorentzian
		A	1
		centre	0
		gamma	1
Lorentzian_0
		A	1
		centre	0
		gamma	1

In [9]: m[0].A.twin = m[1].A

In [10]: m.print_current_values()
Components	Parameter	Value
Lorentzian
		centre	0
		gamma	1
Lorentzian_0
		A	1
		centre	0
		gamma	1

In [11]: m1 = m.inav[0,0]
[...]

PR #1348, opened 2016-12-02
Various Samfire fixes

to266:
- solves both issues in https://github.com/hyperspy/hyperspy/issues/1347
 - a more robust plotting of histograms
 - enables updating parameter boundaries between samfire runs
 - enables updating GOF test between runs
 - removes things form public interface that should not have been there in the first place

### API change
`samfire_pool.update_optional_names` got renamed to `samfire_pool.update_parameters`. Also no longer updates just the optional names, but other worker parameters as well (e.g. parameter boundaries, goodness test)

### DIFF
--- a/hyperspy/model.py
+++ b/hyperspy/model.py
@@ -1722,12 +1722,12 @@ class ModelSpecialSlicers(object):
             slices,
             self.isNavigation)
         _signal = self.model.signal._slicer(slices, self.isNavigation)
-        if _signal.metadata.Signal.signal_type == 'EELS':
-            _model = _signal.create_model(
-                auto_background=False,
-                auto_add_edges=False)
-        else:
-            _model = _signal.create_model()
+        # let's not do anything automatically, since then it's difficult to keep
+        # track of the components:
+        import inspect
+        pars = inspect.signature(_signal.create_model).parameters
+        kwargs = {key: False for key in pars.keys() if key.startswith('auto_')}
+        _model = _signal.create_model(**kwargs)
 
         dims = (self.model.axes_manager.navigation_dimension,
                 self.model.axes_manager.signal_dimension)
--- a/hyperspy/model.py
+++ b/hyperspy/model.py
@@ -1722,8 +1722,9 @@ class ModelSpecialSlicers(object):
             slices,
             self.isNavigation)
         _signal = self.model.signal._slicer(slices, self.isNavigation)
-        # let's not do anything automatically, since then it's difficult to keep
-        # track of the components:
+        # TODO: for next major release, change model creation defaults to not
+        # automate anything. For now we explicitly look for "auto_" kwargs and
+        # disable them:
         import inspect
         pars = inspect.signature(_signal.create_model).parameters
         kwargs = {key: False for key in pars.keys() if key.startswith('auto_')}
--- a/hyperspy/samfire.py
+++ b/hyperspy/samfire.py
@@ -242,6 +242,9 @@ class Samfire:
             while True:
                 self._run_active_strategy()
                 self.plot()
+                if self.pixels_done == self.model.axes_manager.navigation_size:
+                    # all pixels are done, no need to go to the next strategy
+                    break
                 if self._active_strategy_ind == num_of_strat - 1:
                     # last one just finished running
                     break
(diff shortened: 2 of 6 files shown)


hyperspy/model.py
```diff
@@ -1722,12 +1722,12 @@ def __getitem__(self, slices):
             slices,
             self.isNavigation)
         _signal = self.model.signal._slicer(slices, self.isNavigation)
-        if _signal.metadata.Signal.signal_type == 'EELS':
-            _model = _signal.create_model(
-                auto_background=False,
-                auto_add_edges=False)
-        else:
-            _model = _signal.create_model()
+        # let's not do anything automatically, since then it's difficult to keep
```
francisco-dlp: I think that this is a dangerous approach that works with the code as it is at present but may break things in the future. I think that auto adding components is a bad idea anyway so, what about checking if it is an eds/eels model (the only two ones that create components so far) and if they are, set the auto keywords explicitly to `False`. In parallel we could open an issue to remove the `auto*` keys in favour of methods that the use have to run explicitly.

to266: I would rather move this to `next_minor` and at the same time remove autoadding anything. As you can see it used to be checking if it's an EELS model, etc., but then EDS model was added and hence the bug. In general I don't like adding exceptions where either a general approach or sane defaults should suffice...

In this case the "sane defaults" might include "sane kwarg names", making the current approach passable ;)

to266:
Also it seems I __may__ have finally fixed the appveyor samfire tests. Almost not surprisingly it seems to have been a matplotlib / traits / traitui error

hyperspy/model.py
francisco-dlp: I agree with the principle. However, to change the API we need to wait for `next_major` which may be a too long wait to get all of this fixes...

hyperspy/samfire.py
```diff
@@ -483,15 +491,26 @@ def _next_pixels(self, number):
...
-            d_ind = tuple([0 for _ in dic['chisq.data'].shape])
-        self.model.chisq.data[m_ind], dic['chisq.data'] = dic[
-            'chisq.data'].copy(), self.model.chisq.data[m_ind].copy()
-        self.model.dof.data[m_ind], dic['dof.data'] = dic[
-            'dof.data'].copy(), self.model.dof.data[m_ind].copy()
-
-        for comp_name, comp in dic['components'].items():
+            d_ind = tuple([0 for _ in dict_['dof.data'].shape])
+        m = self.model
+        for k in dict_.keys():
```
francisco-dlp, line 498: The commented out code below is more readable than the more compact and general new one. Is the extra generality needed? If yes, what about leaving a comment explaining what the code does instead or in addition to leaving the old code commented out below?

hyperspy/samfire_utils/samfire_pool.py
```diff
@@ -51,8 +51,8 @@ class SamfirePool(ParallelPool):
         given SAMFire object, populates the workers with the required
         information. In case of multiprocessing, starts worker listening to the
         queues.
-    update_optional_names
-        updates the optional names in all workers
+    update_parameters
```
francisco-dlp, line 54: This seems like an API change...

hyperspy/samfire.py
to266: I'll add an explanation.

In particular, I tried adding `corr` to a `Model2D` - a correlation measure to be used instead of `red_chisq`, and it was troublesome adding this to samfire.

In this case I coded the thing in the class (not included in this PR though), but generally it could be anything that gets computed on fitting a pixel, since there now is a `fitted` event that may trigger any arbitrary calculations that save to the model itself.

hyperspy/samfire_utils/samfire_pool.py
to266: Huh, true.... 
On the other hand, I doubt anyone else was aware of the function. On the yet another hand, this "bug" is actually __very__ bad if you do anything else that what I was used to - not only you can't change the parameter boundaries, but also the GOF test, other things... Your call anyway. I can PR to the major branch just as well

francisco-dlp: I am shamefully willing to turn a blind eye to this API change (hopefully @vidartf is not watching...)

francisco-dlp: However, it'll be good to describe in the header and possibly the changes list just in case somebody actually cares...

to266:
Added some comments, updated PR header, created an issue to remind us to make the appropriate API change for the next major version... Anything else?

PR #1353, opened 2016-12-13
Clustering

bm424:
Implements a set of clustering tools into `mva.py`, and a set of `plot_cluster_xyz` tools into `signal.py`.

Clustering is accomplished through the algorithms in `scikit-cmeans` (full disclosure: I've written this package, as I couldn't find any fuzzy c-means algorithms written in python that did all the things I needed it to). The default behaviour is to use the loadings from a previous decomposition to cluster the data, then reproject the results into the original dimensions, but both of these steps can be overriden.

Although I'm fairly confident in the robustness of the code, it may also be worth including some of `scikit-learn`'s hard clustering algorithms, in case users are worried.

Documentation and tests also included.


francisco-dlp:
Awesome. As this is not changing the existing API, could you resubmit it to RELEASE_next_minor instead?

By the way, have you tried contributing the algorithms in scikit-cmeans to sklearn?

bm424:
Whoops! No problem, just done.

I had a think about it, but although the basic API for my code is quite similar, in order to be flexible I've done things quite differently than sklearn in terms of the code structure. Theirs tends to use classes to hook directly into functions, but I wanted the algorithms to be easily plugged in to one another. I'm not sure sklearn would want code structured so differently.

thomasaarholt:
This looks very cool!

thomasaarholt:
One solution to handling noisy data is to include one massive cluster which gives the data points a probability of being part of. Are there any good ways of handling noise with fuzzy c-means?

tjof2:
@thomasaarholt depends on your noise distribution... :-)

bm424:
@thomasaarholt I agree with @tjof2, depends on exactly what you mean by noise. Normally for my applications, "noise" means the differences between data points that *cause* them to form clusters (rather than just single points), so the problem is dealt with inherently... do you have any examples?

thomasaarholt:
I may be trying to use the wrong tool for the job - I thought I'd try and see if I could use fuzzy clustering to count particles from my EDX data after getting intensity maps. Noise in this case refers to counts between the particles. Take a look:
nb_l

hyperspy/learn/mva.py
```diff
@@ -42,6 +44,12 @@
 _logger = logging.getLogger(__name__)
 
 
+class ProbabilisticGK(Probabilistic, GustafsonKesselMixin):
+    pass
+class PossibilisticGK(Possibilistic, GustafsonKesselMixin):
+    pass
+
+
 def centering_and_whitening(X):
```
dnjohnstone, line 52: Are these two functions supposed to be here and just passed?

dnjohnstone (commented):
Just a few points:

1. This still has lots of RELEASE_next_major stuff in. I think you need a quick rebase as that obviously can't be merged.

2. If this is going to happen it needs scikit-cmeans to be added to lists of dependencies. I think also part of the distribution stuff so that people get it installed directly - I guess you put scikit-cmeans on pypi already?

3. @francisco-dlp @vidartf @to266 There is a question about desirable external dependencies here and consistent treatment of that... Given that there exists quite a lot of clustering in scikit-learn that would seem like a natural home to me. I can understand not wanting to structure code to go in there for many reasons but writing a new package and immediately adding it as a dependency of hyperspy is potentially quite problematic. Inevitably there won't exist a wide developer base for a brand new package - not so long ago I remember discussions about whether scikit-image was an appropriate thing to depend on... Also what about scikit-fuzzy? (http://pythonhosted.org/scikit-fuzzy/)

thomasaarholt:
~~Something is wrong here :) Could someone help point out what commit it is that affects the 422 files?~~

ericpre (commented):
~Thanks @pquinn-dls for this PR. After review and merge of https://github.com/pquinn-dls/hyperspy/pull/4, this should be in very good shape!~

Edit: wrong PR...

dnjohnstone:
Is this now irrelevant as it just got superseeded/replicated in #2203 ?

tjof2:
Pretty certain this can be closed in favour of #2203 @francisco-dlp

Issue #1330, opened 2016-11-16
"line roi" is not working properly


CorentinLG:
If one does for instance:

line_roi = hs.roi.Line2DROI(2.5,1.5,8.5,1.5,0.3)
sr = line_roi.interactive(sc, color="yellow")

one gets : 
![line roi](https://cloud.githubusercontent.com/assets/23366094/20360954/187129d2-ac36-11e6-97ab-b36bcda4cfb2.png)

The width is not displayed as it should.

Then, the returned signal has a problem, it seems that the x,y coordinate of the line are not properly taken into account.
For instance, when using the line show above, the returned signal is 0 for many points, as if the line was actually running through the void (left of the image). The navigation and signal dimension seem alright though.
![line roi returned signal](https://cloud.githubusercontent.com/assets/23366094/20361419/c4ffea7a-ac37-11e6-82a0-53905594c5b1.png)

francisco-dlp:
Thank you for reporting this bug.

francisco-dlp:
This looks like a duplicate of #886, @to266, can you confirm this?

CorentinLG:
Yes, it does indeed cover the first part of the problem I raised, the line
width which is not displayed properly whe the windows open.

However it does not cover the second part, on the returned signal.

In any case, don't forget I am new to these things, so it could be me who
did not plot things correctly, or did not expect what I should have
expected to be plotted...
It requires someone with experience to check it...

Thanks.
Corentin

2016-11-17 15:28 GMT+01:00 Francisco de la Peña notifications@github.com:

francisco-dlp:
It seems to me that the drawing of the ROI is wrong on creation but, as reported in #886, moving it corrects the issue. The data in `sr` seems to be right i.e. in my test it corresponds with what it would be  expected if the ROI was properly drawn. I have slightly modified your example for the test as follows:

``` python
%matplotlib qt4
import hyperspy.api as hs
import numpy as np
im = hs.signals.Signal2D(np.zeros((100, 100)))
im.axes_manager[0].offset = 0.7
im.axes_manager[0].scale = 0.5
im.data[:50, 50:] = 1
im.plot()
line_roi = hs.roi.Line2DROI(10, 20,  30 ,40 ,2)
sr = line_roi.interactive(im, color="yellow")
sr.plot()

```

CorentinLG:
Not sure what is happening with my data then.
I tried your example, it returns a plot of the values, but not a image with
[...]

PR #1354, opened 2016-12-14
BUGFIX: Line2DROI in  #1330

thomasaarholt:
This is a bug which stems from the navigator dimension swapping between hyperspy signals and numpy data shapes.

nd.map_coordinates is taking x and y coordinates for the data shape, which is indexed y, x. This PR flips the data input, which I think is the right way around to do this.


francisco-dlp:
Looks good! It'll be nice to  add an unittest to check that the issue doesn't come back. The example in #1330 could serve as inspiration:

```python


import hyperspy.api as hs
import numpy as np
im = hs.signals.Signal1D(np.zeros((100, 100, 1024)))
im.axes_manager[0].offset = 0.7
im.axes_manager[0].scale = 0.5
im.data[:50, 50:, 500:600] = 1
line_roi = hs.roi.Line2DROI(10, 20,  30 ,40 ,2)

```

vidartf:
It seems there is already a unit test for this, which is the one that is currently failing. I think simply updating the expected values in that test should be enough (maybe adding offset and scale as suggested by @francisco-dlp)?

vidartf:
PS the test in question is here: https://github.com/thomasaarholt/hyperspy/blob/9c5e27fabcaa1070ef2debf0e522abd9933ce553/hyperspy/tests/utils/test_roi.py#L111

francisco-dlp:
This is a serious bug that this PR fixes, @thomasaarholt, are you available to modify the test as suggested above? If not, one of us could do it.

thomasaarholt:
I really am not. I've got a job interview on Thursday and am working late nights on my thesis. I would really appreciate it if someone else could modify the test.

francisco-dlp:
thomasaarholt/hyperspy#6

PR #1357, opened 2016-12-14
fix-doc: bcf warnings and add missing mapping of original_metadata

### LINKED ISSUE #1364: problems loading BCF file
I have a BCF file from Bruker Quantax software (attached).  When I try to open it with hyperspy, it produces an AttributeError: no such child: DX from line 533 of hyperspy\io_plugins\bcf.py, due to an attempt to assign semData.DX to self.image.x_res.

Here is my code for opening the file:

import hyperspy.api as hs
s=hs.load("Cathode_161007_70.bcf")

If I modify hyperspy\io_plugins\bcf.py file so that arbitrary values are assigned to self.image.x_res and self.image.y_res in place of semData.DX and semData.DY respectively, then I get a different error:
 
  File "hyperspy-1.1-py3.5-win-amd64.egg\hyperspy\io_plugins\bcf.py", line 598, in _set_sum_edx
    str(i))[0])

IndexError: list index out of range

[Cathode_161007_70.zip](https://github.com/hyperspy/hyperspy/files/671979/Cathode_161007_70.zip)

sem-geologist:
bcf can't save contain dead/live/real time
as also fwhm. Previous behavior was that
it was warning about missing dead time with every load of file.
This fix gets rid of repeated warning, and warns the user about shortcomings of
bcf format only once at the first loading of bcf

Tasks: 
- [x] add note to doc about missing dead/live/real times and fwhm
- [x] remove completely annoying warning from code, do not try to import dead time as it is never saved in bcf (there is logical explanation why, any producer which provide such value actually does great miss-service to his customers. The problem can be not actual for TEM, but for SEM it is very much important: Different mineral of radically diferent density will cause different counting rates so and different dead time, and there cant be universal dead time for whole image on SEM working on large scale heterogenoues samples)
- [x] add most of bruker saved parameters to be available as original_metadata
- [x] add some of parameters saved at the beginning of xml header (~~the problem: they can be missguiding, as it is possible to have completely different scanning averaging for imaging and mapping~~ the situation were investigated and it cleared up)
- [x] add tests
- [x] add calculated real_time to the metadata
- [x] fix hyperspy/hyperspy#1364


thomasaarholt:
Sounds like a good change.

sem-geologist:
hmm... I have problem with this 'warn only once'
It would do that easily with warnings.warn from standard library, but then tests goes nuts...
and logging have no option to emit the same message only once....

sem-geologist:
I couldn't succeed in making warning to appear only once, so I removed that pesky warning completely. Also I probably wanted to be too informative, but I just rethought that no other formats inform user that they are not capable of something.
So instead of warning I added the note in documentation.

Now it is ready to be reviewed if such approach sounds reasonable.

sem-geologist:
whohohoh... I see the error:
This is going to be pain in the...:

the problem is with my xml to dict thing with this:
```xml
<Something Type="TRTWhatever">
  <Type>XFlash | 30</Type>
</Something>
```

so the thing is if I put both: attributes and values into same dictionary if attributes name is the same as childs name, then it gets overwiritten...

francisco-dlp:
It looks good to me. However, it'll  be good to add a test for the new feature (`original_parameters`).

sem-geologist:
@francisco-dlp , the original_metadata is huge, shouldI test it all? or just selective few attributes? the tests is running already very long

francisco-dlp:
What do you mean by huge? If you are worried about making the test file too long, you could import the dictionary containing the expected original metadata from a separate file.

hyperspy/tests/io/test_bcf.py
```diff
@@ -112,7 +112,7 @@ def test_hyperspy_wrap():
                                                                       'azimuth_angle': 90.0}},
                                                  'stage_x': 62409.2,
                                                  'magnification': 131.1433,
-                                                 'tilt_stage': 0.0}},
```
sem-geologist, line 115: this was mistake and missconseption that bruker does not read the stage tilt value from microscope;
0.0 is tilt angle available to setup in separate Esprit window, stage tilt is actually 0.5 and is read from microscope.
P.S. this is horizontal position of German precision stage LOL

sem-geologist:
like with json?

francisco-dlp:
Sounds good.

ericpre:
Thanks @sem-geologist, this is useful!
Would it be worth adding the pixel time in the corresponding metadata entry? Would `real_time` be the best match?
```
├── Acquisition_instrument
│   ├── SEM
│   │   ├── Detector
│   │   │   └── EDS
│   │   │       └── real_time (s)
```

The [metadata specification](http://hyperspy.org/hyperspy-doc/current/user_guide/metadata_structure.html#eds) may need some refining (total time versus pixel time) for the case of mapping.

sem-geologist:
@ericpre, this I think should be discussed with @francisco-dlp, and thing is it is mapping, and it can be abrupt in the middle (i.e. lets say image hight is 512 pixels, it was doing its 3rd frame, when on the middle got stopped (it does so if it is real time setup as the main timer for the mapping in the Esprit); So in this case the pixels at upper part of image would have 1/3 more of real time than lower part. The only sensible way to calculate it is the sum time of whole frame.

francisco-dlp:
This information seems to be useful for analysis, so I am all for adding it to `metadata` as long as there is a reliable way to calculate it.

ericpre:
@sem-geologist, yes good point.
If I understand correctly you explanation in hyperspy/rosettasciio#239, only the total (real) time is reliable by computing `sum(original_metadata['LineCounter'])*original_metadata['DSP Configuration']['PixelAverage']*original_metadata['DSP Configuration']['LineAverage']*width`? If so, we shall go for this!

sem-geologist:
I tested throught all bcf files which I have in tests (some obtained not on our machines) and I see that to this equation actually 'PixelTime' should also be used, as on some machines its value is not 1 but 2 (one generated by Esprit Demo version). Knowing Bruker, they probably have in the wild few versions which differently sets up those parameters on some of microscopes and differently saves it in xml. I think to be on safe side all of them should be multiplied.

francisco-dlp:
@sem-geologist, is this ready for review. If yes, @ericpre, would you like to review it?

sem-geologist:
@francisco-dlp , @ericpre  , yes, generally it is ready

hyperspy/io_plugins/bcf.py
```diff
@@ -516,8 +576,7 @@ def __init__(self, xml_str):
         self._set_image(root)
         self.elements = {}
         self._set_elements(root)
-        self.line_counter = np.fromstring(str(root.LineCounter),
-                                          dtype=np.uint16, sep=',')
+        self.line_counter = interpret(root.LineCounter.text)
```
sem-geologist, line 579: 1. it is quite short and numpy is not effective at creating short arrays
2. if it would be numpy array, then whole dict tree would be imposible to dump with json straight forwardly (I mean original_metadata where this attribute lands in the end)

sem-geologist:
@francisco-dlp , this is ready for review

francisco-dlp:
I would rather leave the task to a Brucker user unless there is none available for this. @ericpre seems like the perfect human for the job as he's also knowledgeable in the metadata structure.

ericpre:
Yes, i can do it, just not now right ! ;)

doc/user_guide/io.rst
```diff
@@ -507,6 +507,9 @@ hybrid (virtual file system/container with xml and binary data, optionally compr
 Most bcf import functionality is implemented. Both high-resolution 16-bit SEM images
 and hyperspectral EDX data can be retrieved simultaneously.
 
+BCF can look as all inclusive format, however it does not save some key EDX parameters:
+any of dead/live/real times, FWHM at Mn_Ka line.
```
ericpre, line 540: Need a small change to the documentation since the real time is now parsed to metadata!

hyperspy/io_plugins/bcf.py
```diff
@@ -418,6 +418,47 @@ def get_file(self, path):
...
+            dictionary.update(d2)
+        if o.text is not None:
+            if len(dictionary) > 0:
+                dictionary.update({'value': o.pyval})
+            else:
+                return interpret(o.text)
+        if len(dictionary) > 0:
+            return dictionary
+
+
```
ericpre, line 461: Since this could possibly be used by other io plugins than the bcf plugin, move these two functions (`ObjectifyJSONEncoder` and `interpret`) to `hyperspy.hyperspy.misc.io.utils_readfile` or something similar?

hyperspy/io_plugins/bcf.py
```diff
@@ -1029,7 +1104,7 @@ def bcf_imagery(obj_bcf, instrument=None):
                      # possibilities to get such parameter from bruker
                      # or some SEM's'
                      'beam_energy': obj_bcf.header.sem.hv,
-                     'tilt_stage': obj_bcf.header.stage.tilt_angle,
+                     'tilt_stage': obj_bcf.header.stage.tilt_stage,
                      'stage_x': obj_bcf.header.stage.x,
                      'stage_y': obj_bcf.header.stage.y
```
ericpre: This following detail may be considered as a detail, but instead of passing a `metadata` dictionary, it may be better to use a `mapping` dictionnary for the io reader. It will avoid populating the metadata entry with `None` when for example the stage tilt or stage position is not available.
I can make a PR later for that, if you prefer.

hyperspy/io_plugins/bcf.py
```diff
@@ -678,6 +741,17 @@ def get_spectra_metadata(self, index=0):
...
+    def calc_real_time(self):
+        """calculate and return real time for whole hypermap
+        in seconds
+        """
+        line_cnt_sum = np.sum(self.line_counter)
+        line_avg = self.image.dsp_metadata['LineAverage']
+        pix_avg = self.image.dsp_metadata['PixelAverage']
+        pix_time = self.image.dsp_metadata['PixelTime']
+        width = self.image.width
+        return line_cnt_sum * line_avg * pix_avg * pix_time * width / 1000000.0
```
ericpre: The return value is `numpy.float64`, changing it to `float` should avoid displaying too many (and irrelevant) digits in the metadata tree (that happen with one of my files).

ericpre (changes requested):
Thanks for this PR, this looks good to me, except a few details!
I think an update of the metadata specification will be necessary to clarify what does the real time mean in case of the spectrum image (total time and not pixel time).

hyperspy/io_plugins/bcf.py
sem-geologist: This will work only with lxml, and is ment to work only with lxml, and lxml is used only with bcf, and lxml is not very welcome in hyperspy as bcf is ignored in tests if lxml is not present on the testing system. So no, unless some more libraries would use lxml (bcf is using lxml, because Brukers/Delphi xml implementation produces sometimes random junk, things not complying with xml spec, which makes it impossible of using it with sax or with native python dom implementation, also brukers xml are large,and lxml is fastest/ memory efficient at parsing.)

doc/user_guide/io.rst
sem-geologist: @ericpre is the addition bellow enought?

hyperspy/io_plugins/bcf.py
sem-geologist: What do you mean? could you give an example?

sem-geologist: I could simply not pass the None, I don't see the benefits getting one more abstraction layer. It is enought bcf is like onion... Actually I find it quite hard to follow the use of 'DictionaryTreeBrowser' in other format implementation. I understand the point I will do something about it, but I am not sure mapping is ok.

hyperspy/io_plugins/bcf.py
ericpre: Yes, if lxml is not pure python library (therefore a optional hyperspy dependency), then it may be better to leave it here. Other io plugin use xml and this functions look generic enough to be possibly used elsewhere in the future!
I was suggesting this because it happen to me previously to write some code that was already present in some of the io plugin or test files... and I find it out a couple of weeks/months later!!

hyperspy/io_plugins/bcf.py
ericpre: See here to know how  [mapping](https://github.com/hyperspy/hyperspy/blob/a9ce12db16d33952c05ec3e7b3d4de41f10af9d2/hyperspy/io.py#L391) works. Actually, in this case you may need to add more data to original_metadata, because it worked nicely only when you parse one parameter from the original_metadata dictionary! The [blockfile reader](https://github.com/hyperspy/hyperspy/blob/e130e34f347f236bfaf136add8a1629768468a4c/hyperspy/io_plugins/blockfile.py#L63) is simple and good example. In the dm plugin, there are a [couple of functions](https://github.com/ericpre/hyperspy/blob/f96a8e6c7d7aba43e254c598b488f5fab28a9d01/hyperspy/io_plugins/digital_micrograph.py#L795) to parse metadata that returns `None` and in this case the metadata tree isn't populated with entries.

Please do as you prefer and if you can avoid populating metadata with `None` value, it would be great! :) When testing on my files, I got a `None` for `stage_x`, `stage_y` and `stage_tilt`.

doc/user_guide/io.rst
ericpre: Yes!

sem-geologist:
@ericpre , I did the improvement, there should be no more `None` getting to metadata.
I moved the stage data (if available) to original_metadata. Also updated things to image metadata and added original metadata (only selection which is actual for images)

ericpre:
Thanks, this is ready to merge!
@sem-geologist, do you think that it is would be useful to have the x, y, z stage position available in the metadata? If so, then it may worth adding it to the [hyperspy metadata specification](http://hyperspy.org/hyperspy-doc/current/user_guide/metadata_structure.html) and then change accordingly this reader (and possibly a couple of other ones). This should be straightforward now that there are in the original_metadata, but it may be for another PR!

ericpre (approved):

sem-geologist:
@ericpre , not every esprit/bruker system saves the stage possition, special license is needed.

sem-geologist:
@ericpre , I just noticed hyperspy/hyperspy#1364. I have ready fix for that, could I add that to this PR? it is just few line changes. Looks some older bcf with some very primitive microscope configuration.

ericpre:
I guess so, it's still more or less in the scope of this PR and if it's convenient for you!

sem-geologist:
@to266 , @ericpre  changes are made, waiting for author tag can be removed. However looks travis got broken (or maybe conda environment, some used flags got deprecated), anyway it passes the tests on appveyor, so it should be ok.

sem-geologist:
I see, python `coverage` package got updated yesterday and something is going wrong with coverage (I tested on my linux machine and nosetests goes ok if coverage is not present, but as soon coverage is installed it fails same as travis)

sem-geologist:
The issue is already known: [link_of_coverage_on_bitbucket_issues](https://bitbucket.org/ned/coveragepy/issues/541/coverage-43-breaks-nosetest-with-coverage)

to266:
I'd merge this now, but ideally I'd like this to include the dataset you were asking for in hyperspy/hyperspy#1364. Wait for it?

hyperspy/io_plugins/bcf.py
```diff
@@ -593,8 +593,14 @@ def _set_sem(self, root):
...
-        self.image.x_res = float(semData.DX)  # in micrometers
-        self.image.y_res = float(semData.DY)  # in micrometers
+        try:
+            self.image.x_res = float(semData.DX)  # in micrometers
+            self.image.y_res = float(semData.DY)  # in micrometers
+            self.units = 'µm'
+        except AttributeError:
+            self.image.x_res = 1.0  # in pixels
+            self.image.y_res = 1.0  # in pixels
+            self.units = 'pix'
```
ericpre, line 603: Usually, when the units aren't defined, we use `Undefined` from the `traits.api` library.

sem-geologist:
no,it can come with seaparate PR

francisco-dlp:
@ericpre, thank you for the review. As pointed out by @to266, I think that it'll be good to wait until @mgsephton provides a file to test for one of the bugs (#1364) that this PR fixes. This may also make coveralls happy.

sem-geologist:
I just thought that maybe we dont need the full bcf file as only xml part is unique, so to save space we could copy to the test only extracted xml part of the file @mgsephton had shared in hyperspy/hyperspy#1364, this way we will not expose any scientificaly valuable information? @mgsephton is it ok?

mgsephton:
@sem-geologist, yes that's fine.  I was thinking of obtaining a BCF file for a stub we use to calibrate our system, consisting of pure Cu fixed onto an Al disc.  Can still supply this if it is useful.

sem-geologist:
@mgsephton, I can wait for separate file, I just have one request, if you could grab that mapping at very low resolution (i.e. image width 128) so it would be not very big. Thanks in advance.

mgsephton:
@sem-geologist, here is a BCF file I obtained using the Cu standard.
Does it meet your requirements? let me know if you need something different.
[Cu Hitachi TM3030Plus.zip](https://github.com/hyperspy/hyperspy/files/689028/Cu.Hitachi.TM3030Plus.zip)

sem-geologist:
Thanks @mgsephton.
Btw, you can check compression option on when saving bcf, our library in hyperspy will read it and decompress without a problem.

sem-geologist:
@francisco-dlp , file has been added and Tests updated. Coverage is still complaining, but it is rather from beiing some commits behind master

PR #761, opened 2015-11-11
ENH: EDAX I/O plugin

jat255:
Eventually will Resolve #756.

Submitting the PR early so people can comment along the way and prod me in the right direction....

TODO:
- [x] Figure out appropriate offsets and dtypes for reading binary formats
- [x] Allow for custom .ipr and .spc filenames
- [x] User guide documentation
- [x] Actual import of data
- [x] `file_reader` function
- [ ] Unit tests

PR #1358, opened 2016-12-15
NEW: EDAX I/O plugin

### LINKED ISSUE #756: Importing of EDAX EDS data
Having recently acquired an EDAX EDS/EBSD system, it has made me realize that there is no simple way to read the EDAX .spd (maps), .spc (spectra), and .ipr (image) files produced by EDAX's TEAM software in Hyperspy. 

I contacted EDAX applications and they provided me with the binary specifications for each of these files, so it shouldn't be too tricky to be able to implement at least read (and maybe even write) functionality for these formats. 

Apparently I cannot upload these documents without write permission to the repo, but they are available here: https://umd.box.com/s/nf4eqti4g07b76hh9zud25h6y1b6mmht

I am going to begin working on this in the next few days and will create a PR when I have some progress, but I wanted to put this issue here so we had a record of it. Any suggestions or tips would be be appreciated!

jat255:
Reopening of #761, with completed unit tests (I think).

Resolves #756 

TODO:

- [x] Figure out appropriate offsets and dtypes for reading binary formats
- [x] Allow for custom .ipr and .spc filenames
- [x] User guide documentation
- [x] Actual import of data
- [x] file_reader function
- [x] Unit tests


jat255:
@AndrewHerzing, could you try this out with a few of your EDAX files, if you get a chance? It might be good to have another pair of eyes confirming it's working

AndrewHerzing:
I've tried it on four different datasets and it works nicely.  The spatial calibration looks good and the metadata looks correct.  Great work.  The spectral calibrations seems to be somewhat off in one of the datasets I've looked at.  It appears that the offset is 0 but should have some value. This could be the fault of the SPD file though.  I'll look more carefully later.

jat255:
@AndrewHerzing thanks for checking it out. The spectral offset is contained in the `.spc` file, so if there is no corresponding `.spc` file with the same name as the `.spd` file (or provided with the `spc_fname` argument), I don't think it'll pick up the calibration.

Although, if it's picking up the eVperchan calibration, it should get the offset too, since those are in the same file...

AndrewHerzing:
That is not the case with the dataset in question as it has an SPC, SPD, and IPR file.  However, my guess is the calibration in TEAM was incorrect as the calibration is off by the same amount if I read it into DM using a plugin I wrote previously.  I'll double check when I can get onto our TEAM system to see if this is the case.

jat255:
Ok, thanks!

hyperspy/io_plugins/edax.py
```diff
@@ -0,0 +1,932 @@
...
+# Writing capabilities
+writes = False
+
+spd_extensions = ('spd', 'SPD')
+spc_extensions = ('spc', 'SPC')
+
+# read dictionary of atomic numbers from HyperSpy, and add the elements that
+# do not currently exist in the database (in case anyone is doing EDS on
+# Ununpentium...)
+atomic_num_dict = dict(((p.General_properties.Z, e) for (e, p) in elements_db))
```
to266: The inner brackets (generator construct) is not necessary here.

hyperspy/io_plugins/edax.py
```diff
@@ -0,0 +1,932 @@
...
+             'tilt_stage': spc_header['tilt']}
+    }
+
+    # Get elements stored in spectrum:
+    num_elem = spc_header['numElem']
+    if num_elem > 0:
+        element_list = sorted([atomic_num_dict[i] for
+                               i in spc_header['at'][:num_elem]])
+        metadata['Sample'] = {'elements': element_list}
+        print("Elemental information found in the spectral metadata " \
```
to266: `_logger.info` (or a more appropriate logger) please. We're trying to not print things any more :)

hyperspy/io_plugins/edax.py
```diff
@@ -0,0 +1,932 @@
...
+                                 count=1)
+
+        spc_dict = sarray2dict(spc_header)
+        original_metadata = {'spc_header': spc_dict}
+
+        nz = original_metadata['spc_header']['numPts']
+        data_offset = original_metadata['spc_header']['dataStart']
+
+        # Read data from file into a numpy memmap object
+        data = np.memmap(f, mode='c', offset=data_offset,
```
to266: The constant `mode` here is a trouble. In particular, if the file is larger than fits in memory, you are not able to open it with the `'c'` mode, as it tries to allocate enough memory to store the modifications. You also do not use / pass the `**kwargs` that you allow.

I think best would be to have something like this:
```python
mode = kwargs.pop('mode', 'c')
data = np.memmap(f, mode=mode, offset=data_offset, dtype='u4', shape=(1,nz), **kwargs).squeeze()
```

This allows passing different options (not only `mode`) to the crucial loading part while keeping the default behavior.

hyperspy/io_plugins/edax.py
```diff
@@ -0,0 +1,932 @@
...
+        ny = original_metadata['spd_header']['nLines']
+        nz = original_metadata['spd_header']['nChannels']
+        data_offset = original_metadata['spd_header']['dataOffset']
+        data_type = {'1': 'u1',
+                     '2': 'u2',
+                     '4': 'u4'}[str(original_metadata['spd_header'][
+                         'countBytes'])]
+
+        # Read data from file into a numpy memmap object
+        data = np.memmap(f, mode='c', offset=data_offset, dtype=data_type
```
to266: Same comment for `mode` applies here too

hyperspy/io_plugins/edax.py
```diff
@@ -0,0 +1,932 @@
...
+    # Get name of .spc file from the .spd map (if not explicitly given):
+    if not spc_fname:
+        spc_path = os.path.dirname(filename)
+        spc_basename = os.path.splitext(os.path.basename(filename))[
+            0] + '.spc'
+        spc_fname = os.path.join(spc_path, spc_basename)
+
+    read_spc = os.path.isfile(spc_fname)
+    if not read_spc:
+        print('Could not find .spc file named {}.\n' \
```
to266: `_logger.warning` or `_logger.info` here too

hyperspy/io_plugins/edax.py
```diff
@@ -0,0 +1,932 @@
...
+        ipr_basename = os.path.splitext(
+            os.path.basename(
+                original_metadata['spd_header'][
+                    'fName']))[0].decode() + '.ipr'
+        ipr_path = os.path.dirname(filename)
+        ipr_fname = os.path.join(ipr_path, ipr_basename)
+
+    read_ipr = os.path.isfile(ipr_fname)
+    if not read_ipr:
+        print('Could not find .ipr file named {}.\n' \
```
to266: Here as well (commenting only for the github review thing to work..)

to266: Also I feel the two `if`s can be obviously merged to `if/else`

hyperspy/io_plugins/edax.py
```diff
@@ -0,0 +1,932 @@
...
+        'offset': original_metadata['spc_header']['startEnergy'] if
+        read_spc else 1,
+        'units': 'keV' if read_spc else t.Undefined,
+    }
+
+    # Handle navigation units input:
+    scale = 1000 if nav_units == 'nm' else 1
+    if nav_units is not 'nm':
+        if nav_units not in [None, 'um']:
+            print("Did not understand nav_units input \"{}\". Defaulting to " \
```
to266: Logger here as well

hyperspy/tests/io/test_edax.py
```diff
@@ -0,0 +1,217 @@
...
+            my_path,
+            "edax_files",
+            test_files[1]))
+
+    def test_data(self):
+        nt.assert_equal(np.uint32, self.spc.data.dtype)     # test datatype
+        nt.assert_equal((4096,), self.spc.data.shape)       # test data shape
+        nt.assert_equal(
+            [0, 0, 0, 0, 0, 0, 1, 2, 3, 3, 10, 4, 10, 10, 45, 87, 146, 236,
+             312, 342], self.spc.data.tolist()[:20])   # test 1st 20 datapoints
```
to266: A minor thing, but I would swap the `tolist()` and slicing operations - no need to create a list with all 4096 elements

hyperspy/tests/io/test_edax.py
```diff
@@ -0,0 +1,217 @@
...
+        # Testing EDS parameters
+        nt.assert_almost_equal(0, eds_dict['azimuth_angle'])
+        nt.assert_almost_equal(34, eds_dict['elevation_angle'])
+        nt.assert_almost_equal(129.31299, eds_dict['energy_resolution_MnKa'],
+                               places=5)
+        nt.assert_almost_equal(50.000004, eds_dict['live_time'], places=6)
+
+        # Testing elements
+        nt.assert_equal(['Al', 'C', 'Ce', 'Cu', 'F', 'Ho', 'Mg', 'O'],
+                        sorted(elements))
```
to266: Aren't elements already sorted?

Depending on what you want to do specifically, I would:
 - Sort your manually typed list (just in case the sorting changes for some reason or someone swapped the order in the list by accident when adding something in the future)
 - If not important that the elements are in particular order, test sets instead of lists: `nt.assert_set_equal({'Al', 'C', ...}, set(elements))`

hyperspy/tests/io/test_edax.py
```diff
@@ -0,0 +1,217 @@
...
+                                     'scale': 0.01,
+                                     'size': 4096,
+                                     'units': 'keV'}}
+        nt.assert_dict_equal(spc_ax_manager,
+                             self.spc.axes_manager.as_dictionary())
+
+
+class TestSpdMap:
+
+    @classmethod
```
to266, line 76: ..are you monkey-patching the nosetests?

Could you explain why the usual `setUp` and `tearDown` do not work in your case?

hyperspy/tests/io/test_edax.py
```diff
@@ -0,0 +1,217 @@
...
+
+        # Testing EDS parameters
+        nt.assert_almost_equal(0, eds_dict['azimuth_angle'])
+        nt.assert_almost_equal(34, eds_dict['elevation_angle'])
+        nt.assert_almost_equal(126.60252, eds_dict['energy_resolution_MnKa'],
+                               places=5)
+        nt.assert_almost_equal(2621.4399, eds_dict['live_time'], places=4)
+
+        # Testing elements
+        nt.assert_equal(['Ce', 'Co', 'Cr', 'Fe', 'Gd', 'La', 'Mg', 'O', 'Sr'],
```
to266: Same `sorted`, etc comment applies here

to266 (changes requested):
Looks good, just a couple of things where I commented (and I trust it's obvious where/if I missed any similar things).

The most important bits are:
 - `kwargs` passing and `mode` arguments
 - logging instead of printing
 - the `setup` and `teardown` of the tests... It does not look like you need them to be classmethods. I may be wrong about this, but you'll have to explain it then :)

hyperspy/io_plugins/edax.py
jat255: This is a great point. Thanks for pointing it out.

hyperspy/tests/io/test_edax.py
jat255: I don't think it's particularly important that the elements are in a certain order, so I'll go with the set solution. Thanks.

hyperspy/tests/io/test_edax.py
jat255: Yeah, this looks weird, but it's nothing nefarious. The reason I did it this way was so I could upload a compressed test file. The uncompressed `.spd` map is 244M, compared to 12M for the compressed version. 

The default `setUp` and `tearDown` methods run before and after every test. If left to default, this would result in decompressing and deleting a 244M file five times (slowing down the tests by a significant amount). By making them class methods, they only run at the beginning and end of the overall test suite. I can change it if you want, but this seemed to be the most efficient way I could find of running things that didn't leave behind a 1/4 GB file on the user's system after testing.

On my machine, timing difference is 3.8s with `classmethods`, and 18.2s without them.

hyperspy/io_plugins/edax.py
```diff
@@ -0,0 +1,943 @@
...
+    # Read the .ipr header (if possible)
+    if read_ipr:
+        with open(ipr_fname, 'rb') as f:
+            ipr_header = np.fromfile(f,
+                                     dtype=get_ipr_dtype_list(endianess),
+                                     count=1)
+            original_metadata['ipr_header'] = sarray2dict(ipr_header)
+
+    # Read the .spc header (if possible)
+    if read_spc:
```
to266, line 835: As previously mentioned, I would merge the `if`s for readability, something along the lines of:
```python
# Figure out all the filenames in one go, and look for those files
if not abc_filename:
    # figure out the filename
if not xyz_filename:
    # ...
read_abc = os.path.isfile(abc_filename)
read_xyz = os.path.isfile(xyz_filename)

# then do the warnings/loadings depending if the file is there
if read_abc:
    # with open(...)
else:
    _logger.warning("...")
if read_xyz:
#...
```

It's OK later on (where you check if spc is there to add calibration to metadata), but at first it looks very strange with the 
```python
if not stuff:
    do_things()
if stuff:
    do_other_things()
``` 
and not
```python
if stuff:
    do_other_things()
else:
    do_things()
```

to266 (changes requested):
Just one last thing I'd like you to change, the strange `if not stuff` -> `if stuff` antipattern

hyperspy/io_plugins/edax.py
jat255: Yeah... I don't know why I did that. Hopefully the updated code makes more sense. Let me know if anything else needs changing.

to266 (approved):

PR #1359, opened 2016-12-15
ENH: scree plot improvements

jat255:
Re-opening of #607. Implemented requested changes, and should be ready for review.

**Todo**
- [x] Implement dictionary parsing for options
- [x] Highlighting by using different symbols (for gray figures).
- [x] Ability to pass the matplotlib axes (to plot several scree plots in the same axes)
- [x] Ability to pass the matplotlib figure (e.g. to customize the size).
- [x] Add options for horizontal line
- [x] Add options for axes titles
- [x] Documentation and user guide update


tjof2:
It would be good if you could, with a simple option flag, start the labelling on the x-axis from 1 rather than 0. Otherwise you can end up referring to the 0th principal component in the text, when really it's the 1st principal component.

I know why they're numbered that way with indexing in Python of course!

francisco-dlp:
@tjof2, I am not sure if that's a good idea as, if you're performing blind source separation or reconstruction the model after PCA you may like to specify the index of the components using Python's convention. I bet that, if we start counting the PCA components from one, many of us (starting with me) we'll get this step wrong more often than not which is more critical than referring to the 0th PC in a paper ;)

jat255:
I think I agree with @francisco-dlp. In the text of a paper, I think "First PC" is still clear, and saying something like "0-index" isn't too bad. I could make it a non-default option to provide flexibility, however.

tjof2:
Indeed, I don't mean by default, but as an option - the flexibility would be nice, since this all about flexibility for plotting scree plots for paper figures!

jat255:
@tjof2, how do you feel about https://github.com/hyperspy/hyperspy/pull/1359/commits/3c6465eb588c56776948c1f0d375abd8ae233bdf?

tjof2:
@jat255 exactly what I had in mind!

hyperspy/learn/mva.py
```diff
@@ -981,6 +981,9 @@ def scree_plot(self, n=50, log=True, cutoff=0.01, signal_num=0,
             (i.e. the signal components as opposed to noise).
             Only used if `cutoff` is None, will be automatically determined
             otherwise
+        start_at : int
+            Number to use as starting point for numbering of the principle
```
tjof2: "principal" :)

francisco-dlp:
Mmm, I am not sure if that's a good idea as it may be misleading to some. What about mixing the nomenclature e.g. "1st (index 0) Component "?

jat255:
@francisco-dlp you mean in the docstring, or in a paper?

francisco-dlp:
No, in the figure. You could use the `ordinal` function in `misc.utils`.

jat255:
Taking a gander at [google image results](https://www.google.com/search?q=scree+plots&source=lnms&tbm=isch&sa=X&ved=0ahUKEwifz_Gd_PjQAhUG82MKHch8D-EQ_AUICCgB&biw=2133&bih=1188&safe=active&ssui=on), we seem to be only ones starting our plot at 0, it appears.

tjof2:
@jat255 That's kind of what I was thinking. There are very good arguments for starting at 0 with Python using zero-indexing as @francisco-dlp points out, but for people who may be familiar with scree plots but not Python seeing the figure for the first time, I thought it would nice to have a simple way of starting at 1.

I, for one, was initially puzzled by this the very first time I did `plot_explained_variance_ratio` etc. in HyperSpy! Easily overcome, of course :-P

jat255:
If we labeled the x-axis by default as "number of principal components", that would be more clear, I think.

jat255:
> No, in the figure.

@francisco-dlp I'm not sure I understand where you mean to put the "1st (index 0) Component ". On the label?

francisco-dlp:
Forget about my two last comments, they make no sense: I was thinking on the index of the component as displayed when using `plot_decomposition_*`. And this is the issue, it has to be consistent. In the way that it is implemented, it is not currently possible for the `plot_decomposition_*` to use order instead of index.

I agree that for the scree plot it'll be good to have the option to start a 1, and I agree changing the x label to something like "order" or "number" when x represents order would prevent confusion. By the way, why not adding an `xaxis` kwarg that takes only "order" or "index" instead of the `start_at` argument that can take any number while only 0 and 1 make sense?

jat255:
Thoughts?

`xaxis='index'`:
image

`xaxis='number'`: 
image

francisco-dlp:
Looks great but, maybe it is the other way around i.e. 0 for index and 1th for order?

francisco-dlp:
I am obviously not a native speaker, so you'll both know better, but I prefer "order" to "number" for the arg value.

jat255:
> Looks great but, maybe it is the other way around i.e. 0 for index and 1th for order?

Yes, you're probably right. On second though, I'll make ordinal a flag as well, and just leave the default as cardinal numbering 

> I am obviously not a native speaker, so you'll both know better, but I prefer "order" to "number" for the arg value.

I've always thought of "order" and "index" as synonymous when it comes to programming, so I'd prefer number. "Number" also seems to be much more widely used in the scree plots I've seen online.

francisco-dlp:
I like that you use ordinal for "number" and not for index as it makes it harder to confuse what the xaxis represents. Would it make sense to make the ordinal optional only when plotting order?

(I can't believe that we're discussing this much about this tiny issue)

jat255:
Okay, more updates. Using a dict now for the axis, so 'index' vs. 'number' can be set independently from 'cardinal' vs. 'ordinal'

Here's the new default:


`xaxis = {'type':'number'}` (defaults labeling to ordinal)


`xaxis = {'type':'number', 'labeling':'cardinal'}` (parts can be controlled independently)



> (I can't believe that we're discussing this much about this tiny issue)

Well this is obviously the most important part of HyperSpy! party

thomasaarholt:
The "counting components" bit always confuses people when they first try PCA decomposition in Hyperspy. I'm a big fan of this. 👍

francisco-dlp:
I am too. My previous concerns were fully misguided, sorry for the noise.

hyperspy/learn/mva.py
```diff
@@ -958,6 +959,170 @@ def plot_explained_variance_ratio(self, n=50, log=True):
             ax.semilogy()
         return ax
 
+    def scree_plot(self, n=50, log=True, cutoff=0.01, signal_num=0,
```
francisco-dlp: I wouldn't change the name of the function. This is not exactly the scree plot as it doesn't plot the eigenvalues but the explained variance ratio. (And not even that when using poisson noise normalization...) So, why not simply replacing the old function?

francisco-dlp: I would set the `cutoff` to `None` by default as any uninformed guess will always be wrong and it can be misleading to the unaware user who may think that we have a clever way to do the job for him.

Talking about clever ways of estimating the number of PCs, maybe @tjof2 knows of any worth using? (I've tested the one in sklearn and it is so slow that I killed the computation before finding out if it was doing its job properly...)

francisco-dlp: You provide two features to highlight the signal components: different marker styles and the horizontal line. So, what about controlling them with the following keywords instead: `n_pcs=None`, `highlight_pcs=True`, `hline=True`?

hyperspy/learn/mva.py
```diff
@@ -958,6 +959,170 @@ def plot_explained_variance_ratio(self, n=50, log=True):
...
+            If True, the y axis uses a log scale.
+        cutoff : float or None
+            y-position at which to draw a line showing the cutoff between
+            signal and noise
+        signal_num : int
+            Number of components to highlight as important
+            (i.e. the signal components as opposed to noise).
+            Only used if `cutoff` is None, will be automatically determined
+            otherwise
+        xaxis : dict or None
```
francisco-dlp: Why not splitting this in two kwargs e.g. `xaxis_type`, `xaxis_labeling`? I think that it is more orthodox and user friendly this way.

hyperspy/learn/mva.py
```diff
@@ -958,6 +959,170 @@ def plot_explained_variance_ratio(self, n=50, log=True):
...
+            ``xaxis['labeling']`` should have str value ``'ordinal'`` or
+            ``'cardinal'``, and is used to determine the format of the tick
+            labels (i.e. "1st" or "1", respectively).
+        signal_fmt : dict
+            Dictionary of matplotlib formatting values for the signal
+            components
+        noise_fmt : dict
+            Dictionary of matplotlib formatting values for the noise
+            components
+        axes_titles: dict or None
```
francisco-dlp: It is very easy to use mpl to customize the axes labels, why providing another way?

From the zen of Python

francisco-dlp (changes requested):

hyperspy/learn/mva.py
tjof2: @francisco-dlp not really for PCA - the reason I used soft rather than hard-thresholding (i.e. PCA) of the singular values was because it's easier to automate! @bm424 might know more.

jat255: > So, why not simply replacing the old function?

Sure, I can do that. I originally didn't want to be presumptuous and replace code :)

jat255: > as any uninformed guess will always be wrong and it can be misleading to the unaware user

This is a good point. Should we try to provide a good estimate, or just leave that up to the user? I found the following tutorial, which reviews a few different ways to determine the "right" number of components: http://eric.univ-lyon2.fr/~ricco/tanagra/fichiers/en_Tanagra_Nb_Components_PCA.pdf

francisco-dlp: It's notoriously hard to make a good automatic estimation so let's at least leave it for another PR.

jat255:
> You provide two features to highlight the signal components: different marker styles and the horizontal line. So, what about controlling them with the following keywords instead: n_pcs=None, highlight_pcs=True, hline=True?

@francisco-dlp I prefer the ability to specify a specific cutoff value for the variance. If we do only "number of signal components", than the cutoff line will always go through one of the points. The point of the `cutoff` option (in my mind) is to be able to easily show "let's take everything that describes 99% of the data" (for example) and have the method figure out which components should be highlighted based off that. The `signal_num` option is provided as another way of defining the number of "interesting" components.

I'll add an `hline` control for drawing the cutoff line; that's important.

francisco-dlp:
I see, that's a good point indeed. But I still think that the other interface is more useful. If `signal_num >= 1` then it is number of components. Anything between (0, 1) is a percentage. This has the extra advantage that the line could be automatically plotted only when providing a percentage.

jat255:
Hmm, I like that idea of having one parameter for both, with different behavior for ints and a float. I think the name `signal_num` would be confusing for the (0,1) percentage option though. Perhaps something like `signal_thresh`, instead?

francisco-dlp:
What about simply `pcs` or `PCs`? Although normally we want to keep just "components" this may be a too restrictive usage case hence name.

jat255:
I think I like `signal_thresh` more, since it does not imply "number of components" like the others do.  I'll make sure to document the parameter well so it's clear that it has two different behaviors

francisco-dlp:
In that case, why not simply `threshold`?

francisco-dlp:
By the way, the number of components is sometimes defined in `signal.learning_results.output_dimension`, maybe you could check if that's the case to get the info automatically?

jat255:
> maybe you could check if that's the case

Yup, I realized that too when I found a bug a little bit ago if `n` was too large. It should be fixed in https://github.com/hyperspy/hyperspy/pull/1359/commits/82275fe14df3ebfba6de827355a7675408849143#diff-cd68ac14189365f87b724293fa3c9f38R1040

jat255:
Ok @francisco-dlp, docs are updated, so it should be ready for review once more

hyperspy/learn/mva.py
```diff
@@ -920,42 +919,217 @@ def get_explained_variance_ratio(self):
         s.axes_manager[-1].units = ''
         return s
 
-    def plot_explained_variance_ratio(self, n=50, log=True):
-        """Plot the decomposition explained variance ratio vs index number.
+    # def plot_explained_variance_ratio(self, n=50, log=True):
```
francisco-dlp: It'll be better to delete the old code.

hyperspy/learn/mva.py
```diff
@@ -920,42 +919,217 @@ def get_explained_variance_ratio(self):
...
 
         Parameters
         ----------
-        n : int
-            Number of components.
+        n : int or None
+            Number of components to plot. If None, all components will be plot
         log : bool
             If True, the y axis uses a log scale.
+        threshold : float or int
```
francisco-dlp, line 936: The functions doesn't behave as described here: it always plots the `hline` by default.  I personally like more the behaviour as described as it provides the information of whether the threshold was and int of perc visually. To implement it I bet that `hline` should default to "auto" instead of True.

jat255: Can you explain further? In my testing, the default of `True` does not draw a line if no `threshold` is given, due to https://github.com/hyperspy/hyperspy/pull/1359/files#diff-cd68ac14189365f87b724293fa3c9f38R1021

I can change the options to `'auto'` and `False`, if that makes more sense, but having it as `True` when no threshold is given does not make sense, since there will be no value given to determine where the line should be drawn.

francisco-dlp: Sorry, I misread the docstring i.e. I understood what I wanted to understand. What do you think about implementing the behaviour that I hallucinated?

jat255: If I understand correctly, `'auto'` should give the behavior that `True` currently has, and `False` should turn off the line entirely, correct? I don't think having a value of `True` makes sense in this situation.

francisco-dlp: * `True`: on
* `False`: off
* `auto`: `on if 0 < threshold < 1` `off if threshold >= 1`

jat255: Ok, in the case of `True`, if `threshold` isn't given, where should the line be drawn?

francisco-dlp: Sorry for that `True`: `on if threshold`

jat255: Gotcha. Sounds good.

francisco-dlp: The visualisation is so nice that I wish we had an automatic number of component estimator just to have it on by default ;)

jat255:
Ok, here's the updated behavior for `hline`. The title of each subplot shows the value of hline and whether the threshold value was a float, an int, or not given:

image

jat255:
> I wish we had an automatic number of component estimator just to have it on by default ;)

I agree! I think anything we do in that area would be misleading, however

francisco-dlp:
Is it ready for review?

jat255:
Should be!

hyperspy/learn/mva.py
```diff
@@ -1013,13 +1015,30 @@ def plot_explained_variance_ratio(self, n=None, log=True, threshold=0,
         # Determine right number of components for signal and cutoff value
         if isinstance(threshold, float):
```
francisco-dlp, line 1016: I would also check if `0 < t < 1`. If it is not, then it should raise a `ValueError`.

jat255: Ok, should be fixed. I also fixed an edge-case bug I found if the `threshold` value was too small (less than the minimum variance)

PR #1366, opened 2016-12-29
Correct some info in signal2d.rst

changhw01:
The info here should be for two-dimensional signals.


to266:
Hi, thanks. Could you PR this to the RELEASE_next_patch instead? Thanks

PR #1367, opened 2017-01-02
Add del_item method to DictionaryTreeBrowser class

ericpre:
Add a method and tests to delete a item of a DictionaryTreeBrowser object.
I may need it for #1283 to update date and time in the metadata.

I also moved two files (test_attrsetter.py and test_dictionary_tree_browser.py) to tests/misc to match their locations in the library (misc/utils.py).
Actually, I am wondering if the DictionaryTreeBrowser class would deserve its own file in the misc folder, because there are quite a lot of stuff in misc/utils.py. Would it be worth doing this?


francisco-dlp:
Why is this needed? Wouldn't you be able to accomplish the task using `del`?

```python
d = DictionaryTreeBrowser({'Branch':{'Leaf1':{'caterpillar':False, 'color':'green'}}})
del d.Branch.Leaf1.color

```

ericpre:
Yes, it does!

PR #1368, opened 2017-01-05
New Expression2D

francisco-dlp:
Add 2D capabilities to the `Expression` component. It can automatically add an `angle` parameter to rotate the component around its centre or the origin and automatically calculates the gradients.

The following example shows how to use it to create an elliptical gaussian2d with rotation capabilities equivalent to #864:

```python
import hyperspy.api as hs
import numpy as np

g = hs.model.components2D.Expression("exp(-((x-x0)**2 / (2 * sx ** 2) + (y-y0)**2 / (2 * sy ** 2)))",
                                     "gaussian2d", add_rotation=True, position=("x0", "y0"),
                                     module="numpy")

g.rotation_angle.value = np.radians(30)
g.sx.value = 0.5
g.sy.value = 1
g.x0.value = 1
g.y0.value = -0.5

l = np.linspace(-2, 2, 1000)
x, y = np.meshgrid(l, l)

f, axarr = plt.subplots(ncols=6)
axarr[0].imshow(g.function(x, y))
axarr[1].imshow(g.grad_sx(x, y))
axarr[2].imshow(g.grad_sy(x, y))
axarr[3].imshow(g.grad_x0(x, y))
axarr[4].imshow(g.grad_y0(x, y))
axarr[5].imshow(g.grad_rotation_angle(x, y))
for ax in axarr:
    ax.set_axis_off()

```
expression2d

The evaluation of the function takes 50 ms using `module="numpy"` and 8 ms (!) using `module="numexpr"` i.e. the performance using numpy is the same as with the component in #864 and with numexpr almost 10x faster. As it automatically generates the gradient, the performance when using it for fitting purposes should be better than 10x compared to #864.

TODO
--------
* [x] Test
* [x] Add rotation point argument (currently it uses `position` if defined, otherwise the centre)
* [x] User Guide


magnunor:
Hmmmmm, does this make https://github.com/hyperspy/hyperspy/pull/1241 redundant?

I think it is useful having component functions like `Gaussian2D`, but if the performance is that much better we might as well make these "premade" functions using the Expression2D.

francisco-dlp:
@magnunor, I agree that it'll be good to provide a library of 2D components in addition to the `Expression` component. I actually decided to work on this while reviewing #1241: I thought on contribution the gradients to your branch but I realised that it was easier to address the general problem with the `Expression` component.

francisco-dlp:
This should be ready to review.

doc/user_guide/model.rst
```diff
@@ -108,9 +108,11 @@ Specifying custom components
 
 .. versionadded:: 0.8.1 :py:class:`~._components.expression.Expression` component
 
+.. versionadded:: 1.1 :py:class:`~._components.expression.Expression` component can create 2D components.
+
```
dnjohnstone: Will this not be 1.2?

hyperspy/_components/expression.py
```diff
@@ -36,19 +45,28 @@ class docstring.
         expression: str
             Component function in SymPy text expression format. See the SymPy
             documentation for details. The only additional constraint is that
-            the variable must be `x`. Also, in `module` is "numexpr" the
+            the variable(s) must be `x` (for 1D components) or `x` and `y` for
+            2D components. Also, if `module` is "numexpr" the
             functions are limited to those that numexpr support. See its
```
dnjohnstone: Would remove brackets here:

"the variable(s) must be `x` for 1D components; or `x` and `y` for  2D components."

hyperspy/_components/expression.py
```diff
@@ -36,19 +45,28 @@ class docstring.
...
             documentation for details.
         name : str
             Name of the component.
         position: str, optional
             The parameter name that defines the position of the component if
             applicable. It enables adjusting the position of the component
-            interactively in a model.
+            interactively in a model. 2D components require passing a tuple
+            with the name of the two parameters e.g. `("x0", "y0")`.
         module: {"numpy", "numexpr"}, default "numpy"
```
dnjohnstone: Grammar:

"It enables interactive adjustment of the position of the component in the model. For 2D components, a tuple must be passed with the name of the two parameters e.g. `("x0", "y0")`.

hyperspy/_components/expression.py
```diff
@@ -36,19 +45,28 @@ class docstring.
...
             Module used to evaluate the function. numexpr is often faster but
-            it supports less functions.
-
+            it supports less functions and requires installing numexpr.
+        add_rotation: bool, default False
```
dnjohnstone: Grammar:

"it supports fewer functions"

dnjohnstone (commented):
Looks good to me and works for me. Some very minor comments made.

francisco-dlp:
@dnjohnstone, thank you for the comments, the language should be fixed now.

thomasaarholt:
@to266 removed the `needs review` tag on this. Is it cleared to be merged?

to266:
Not really, it just need authors attention first: both rebase / merge the appropriate branch and update this with the new changes (e.g. allowing substitutions in the expression).

Generally I think it's usually either "needs review" or "waiting for author", and not both. Once the changes are final, a (final) review can take place.

francisco-dlp:
Done!

francisco-dlp:
Fixed again.

hyperspy/_components/expression.py
```diff
@@ -118,14 +147,33 @@ class docstring.
...
+                roty = sympy.sympify(
+                    "{1} + (x - {0}) * sin(rotation_angle) + (y - {1}) *"
+                    "cos(rotation_angle)"
+                    .format(*position))
+            else:  # Rotate around the origin
+                rotx = sympy.sympify(
+                    "x * cos(rotation_angle) - y * sin(rotation_angle)")
+                roty = sympy.sympify(
+                    "x * sin(rotation_angle) + y * cos(rotation_angle)")
+            expr = expr.subs({"x": rotx, "y": roty}, simultaneous=False)
```
to266: I believe this if/else play is unneeded, and could be written as
```python
position = position or (0,0)
# proceed to use the "if position" code
```
when `sympy.sympify` parses the string, it also simplifies it - so the zeroes vanish anyway.

Also the default in the function is `False`, but it would never be so - always either a user-given value, or a `None` from `__init__` default is passed

hyperspy/tests/model/test_components2D.py
```diff
@@ -28,3 +28,136 @@ def test_values(self):
         assert_allclose(gt.max(), 0.0795774715459)
         assert_allclose(gt.argmax(axis=0)[0], 500)
         assert_allclose(gt.argmax(axis=1)[0], 500)
+
+
+class TestExpression2D:
+
+    def test_with_rotation(self):
+        g = hs.model.components2D.Expression(
+            "exp(-((x-x0)**2 / (2 * sx ** 2) + (y-y0)**2 / (2 * sy ** 2)))",
```
to266: How about moving the expression string to the class / make it global / fixture? Then there would be no need to check each expression that it's identical, and if there is something explicitly non-referenced, it's also clear.

May also be worth it making sure in at least one of the tests that setting the initial parameter values when initializing the component works.

to266 (changes requested):
A couple of things in addition to Duncans comment for 1.2 instead of 1.1

francisco-dlp:
@to266, thanks for the suggestions—all done!

to266 (approved):

Issue #1146, opened 2016-07-13
ENH: run tests on minimal environment


sem-geologist:
with this #1144 emergent fixes, I got the idea if it would not be clever to prevent such stuff in future by testing stuff also on minimalistic environment, without lxml, h5py, cython, compilers etx. present. The linux on travis is anyway undersaturated as now builds waits mostly for osx. So we could add one more linux environment to the travis matrix, which would test things on very minimalistic environment with bare minimum.

francisco-dlp:
Sounds like an excellent idea.

sem-geologist:
the feature got added, so closing

Issue #1338, opened 2016-11-25
bug-first time use: hyperspy outside of anaconda on fresh linuxsystem


sem-geologist:
hello,

I just tried to install and launch hyperspy on newly installed kde neon linux distro.
I however fail to use hyperspy, and I cant understand what is exact problem. scikit and sklearn is installed. I do as in manual:

```python
[1] %matplotlib qt

[2] import hyperspy.api as hs

---------------------------------------------------------------------------
ImportError                               Traceback (most recent call last)
<ipython-input-8-5e09b9d2af1c> in <module>()
----> 1 import hyperspy.api as hs

/home/empa/.local/lib/python3.5/site-packages/hyperspy/api.py in <module>()
     73 
     74 from hyperspy.Release import version as __version__
---> 75 from hyperspy import signals
     76 from hyperspy.io import load
     77 from hyperspy.defaults_parser import preferences

/home/empa/.local/lib/python3.5/site-packages/hyperspy/signals.py in <module>()
     38 
     39 # -*- coding: utf-8 -*-
---> 40 from hyperspy._signals.signal1d import Signal1D
     41 from hyperspy._signals.signal2d import Signal2D
     42 from hyperspy._signals.complex_signal import ComplexSignal

/home/empa/.local/lib/python3.5/site-packages/hyperspy/_signals/signal1d.py in <module>()
     20 import numpy as np
     21 
---> 22 from hyperspy.signal import BaseSignal
     23 from hyperspy._signals.common_signal1d import CommonSignal1D
     24 from hyperspy.gui.egerton_quantification import SpikesRemoval

/home/empa/.local/lib/python3.5/site-packages/hyperspy/signal.py in <module>()
     34 from hyperspy import io
     35 from hyperspy.drawing import mpl_hie, mpl_hse, mpl_he
---> 36 from hyperspy.learn.mva import MVA, LearningResults
     37 import hyperspy.misc.utils
     38 from hyperspy.misc.utils import DictionaryTreeBrowser

/home/empa/.local/lib/python3.5/site-packages/hyperspy/learn/mva.py in <module>()
     32 from hyperspy.misc.machine_learning import import_sklearn
     33 import hyperspy.misc.io.tools as io_tools
---> 34 from hyperspy.learn.svd_pca import svd_pca
     35 from hyperspy.learn.mlpca import mlpca
     36 from hyperspy.learn.rpca import rpca_godec, orpca

[...]

PR #1369, opened 2017-01-08
enh-test: bare min travis environment

sem-geologist:
This addresses: #1146

This adds additional travis linux environment
to the matrix which:
* does not install conda
* does not install cython, lxml, and anything which is not in setup dependency.
* uses available system `python3` and mimics most close the
  fresh installation using system available `pip3`.

Why it is needed?
on one side:
* hyperspy is available through pip
* documentation declares that it is one of the way to install it using pip.

on the other side:
* it does not work so perfectly if tried on fresh non-conda system
* some of the requirements are declared to be optional, but it's optionality is ignored in some
of tests and parts of hyperspy.

TASKS:

- [x] modify the `.travis.yml` to include one more matrix element, which tries to install hyperspy without conda.
- [x] fix the build/test/execution failures resulting from treating some libraries as (or not treating as) optional (contribution of fixes are more than welcome), either code should be fixed or, either the libraries should be included to dependencies list!

Things to fix:
- [x] `fast_svd` import error
- [ ] is `skimage` optional or not? lookign at signal class it is not, however it is missing in the dependency of setup.

### DIFF
--- a/hyperspy/misc/utils.py
+++ b/hyperspy/misc/utils.py
@@ -431,8 +431,9 @@ class DictionaryTreeBrowser(object):
         else:
             return False
 
-    def get_item(self, item_path):
-        """Given a path, return True if it exists.
+    def get_item(self, item_path, default=None):
+        """Given a path, return it's value if it exists, or default
+        value if missing.
 
         The nodes of the path are separated using periods.
 
@@ -441,6 +442,9 @@ class DictionaryTreeBrowser(object):
         item_path : Str
             A string describing the path with each item separated by
             full stops (periods)
+        default :
+            The value to return if the path does not exist.
+
 
         Examples
         --------
@@ -466,11 +470,11 @@ class DictionaryTreeBrowser(object):
             else:
                 item = self[attrib]
                 if isinstance(item, type(self)):
-                    return item.get_item(item_path)
+                    return item.get_item(item_path, default)
                 else:
-                    raise AttributeError("Item not in dictionary browser")
+                    return default
         else:
-            raise AttributeError("Item not in dictionary browser")
+            return default
 
     def __contains__(self, item):
         return self.has_item(item_path=item)
--- a/hyperspy/tests/test_dictionary_tree_browser.py
+++ b/hyperspy/tests/test_dictionary_tree_browser.py
@@ -126,3 +126,42 @@ class TestDictionaryBrowser:
         # time in the DictionaryBrowser
         dt_str = '2016-08-05T10:13:15.450580'
         self._test_date_time(dt_str)
+
+    def test_has_item(self):
+        # Check that it finds all actual items:
+        assert self.tree.has_item('Node1')
+        assert self.tree.has_item('Node1.leaf11')
+        assert self.tree.has_item('Node1.Node11')
+        assert self.tree.has_item('Node1.Node11.leaf111')
+        assert self.tree.has_item('Node2')
+        assert self.tree.has_item('Node2.leaf21')
+        assert self.tree.has_item('Node2.Node21')
+        assert self.tree.has_item('Node2.Node21.leaf211')
+
+        # Check that it doesn't find non-existant ones
+        assert not self.tree.has_item('Node3')
+        assert not self.tree.has_item('General')
+        assert not self.tree.has_item('Node1.leaf21')
+        assert not self.tree.has_item('')
+        assert not self.tree.has_item('.')
+        assert not self.tree.has_item('..Node1')
+
+    def test_get_item(self):
+        # Check that it gets all leaf nodes:
+        nt.assert_equal(self.tree.get_item('Node1.leaf11'), 11)
+        nt.assert_equal(self.tree.get_item('Node1.Node11.leaf111'), 111)
+        nt.assert_equal(self.tree.get_item('Node2.leaf21'), 21)
+        nt.assert_equal(self.tree.get_item('Node2.Node21.leaf211'), 211)
+
+        # Check that it gets all leaf nodes, also with given default:
+        nt.assert_equal(self.tree.get_item('Node1.leaf11', 44), 11)
+        nt.assert_equal(self.tree.get_item('Node1.Node11.leaf111', 44), 111)
+        nt.assert_equal(self.tree.get_item('Node2.leaf21', 44), 21)
+        nt.assert_equal(self.tree.get_item('Node2.Node21.leaf211', 44), 211)
+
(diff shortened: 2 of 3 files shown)


sem-geologist:
my first feedback to the first try:
it is failing!!!
* it is replicating the bug #1338 which is mystically resolved (so that is why it was closed) after installing/uninstaling sklearn/skimage, and closing and launching notebook again. This stinks, the failing stinking part should be identified and fixed, think about potential user, if he/she can't get it working from the start, probably after minute he/she is going to uninstall it... (and I run into it on fresh linux install #1338), btw the culprit is `fast_svd`.
* ~~dear maintainer of `digital_micrograph` plugin, your test are over ^ 2 verbose; leave debugging info out from travis, as (1) it is irrelevant, because it is even not the culprit i.e. of this failure – in other similar failure case it would hide the real place of failure by overflowing logs (2) debugging should not be done on the travis. In our case Travis built-in log viewer complains that it is too long, but even the downloaded log version is truncated at the end as it overflows 5MB. It also freezes travis tests at that 5MB overflowing event...~~ update: it looks hyperspy.api should set logging level, but this problem apperas as it does not manage to be loaded due to above...

I think this will need a lot of work, but mess should be cleaned.
I propose that things should be fixed in separate PR, in the meanwhile this PR will be updating to upstream

~~update:~~
~~it is possible that there are some other over-verbose plug-ins, just We cant see it as digital_micrograph is tested first and log is flooded by it. So generally:~~
~~* note should be added to doc for developers that "if plugin produces debug information, tests should be set to log them not".~~
~~* fix it in existing plugins.~~

sem-geologist:
should coveralls be also called after minimal environment testing finished?
I wish it would cumulate the coverage, then coverage would rise up, as some of stuff (especially those test skips in full tests, missing libraries conditions) were ignored previously.

sem-geologist:
@francisco-dlp , is skimage optional library? From failing tests it does not look like that. If it is intended to be optional, then something like lxml check inbcf should be added.

francisco-dlp:
Very nice initiative. I think that we should aim at decreasing the strong dependencies of HyperSpy to the bare minimum. So, you're right, skimage is currently a dependency but I think that it'll be sensible to  turn it into a recommend library.

sem-geologist:
skimage is used at least in two places: tiff io_plugin and _signal.complex;
first one treats it as optional, second one as obviously present.
so maybe it should be treated as dependency?

I think I should add scikit-image to setup.py install_req, shaving to the min should be other PR.
This PR is to setup workflow and tool which shows if install_req in setup.py is consistent with reality.

and if talking about shaving to min, I would start from completely other stuff like traistui and other thingies which have nothing to do with algorithms, but is used for (G)UI's - things where scripts does not need any of them.

francisco-dlp:
I fully agree: HyperSpy shouldn't need any plotting or gui package to run. Actually, fully getting rid of traitsui may be desirable as it is by far the package causing more installation troubles.

sem-geologist:
Nice!!! coverall cumulates coverage so now even loops/conditions, which is for skipping tests or imports, gets covered. +0.15% is not much but it is nice side effect :)

francisco-dlp:
Nice indeed. Is it ready for review?

By the way, why not submitting it to RELEASE_next_patch? You are even fixing a couple of bugs (as it always happens when adding unittests ;) )

sem-geologist:
it is ready for review, I think I mixed up the branches, now I am scratching my head... how to correct it?

francisco-dlp:
You could use [rebase or cherrypick](http://stackoverflow.com/questions/1994463/how-to-cherry-pick-a-range-of-commits-and-merge-into-another-branch) and, after pushing the changes, push the edit button at the right of the title and change the base to RELEASE_next_patch.

sem-geologist:
ups... I need to rebase first..

.travis.yml
```diff
@@ -3,42 +3,67 @@ language: python
...
-  - python setup.py install
+  - if [[ $MINIMAL_ENV == 'False' ]] ; then
+      DEPS="nose numpy scipy matplotlib ipython h5py sympy scikit-learn dill natsort setuptools scikit-image cython lxml ipyparallel";
+      conda create -n testenv --yes $DEPS;
+      source activate testenv;
+      conda install pip;
+      pip install coverage coveralls traits traitsui requests tqdm;
+      python -c "import sys; print(sys.executable)";
+      python setup.py install;
+    else pip3 install .;
```
vidartf: Why not move the coverage and coveralls installs for minimal case up here?

sem-geologist: Imho `pip install coverage coveralls...` should be  the same moved to `script` where it is needed. This way time is optimized, so if installation fails, no time is spent then for installing stuff it is not going to be used. I understand the urge to have it organized into travis present `before_install`, `install`, `script`... sections, but build/testing is all ready taking too long time, You can get nap while waiting for travis to finish... And sometimes you need to wait quite long only to get it to fail at installation/build step before even testing starts.

sem-geologist:
I tried todo rebase, but it does not work :/ I give up....

vidartf:
Run this command (assuming you have added `hyperspy` as the remote for the main repo):
```
git rebase --onto hyperspy/RELEASE_next_patch hyperspy/RELEASE_next_minor patch-14
```

It rebased without any conflicts for me.

sem-geologist:
yes it do same as my trial of doing it, it still gets whole list of commits between next_minor and next_patch branches...

sem-geologist:
Maybe i simply close this and do proper pull request... I have no idea why git is for me like thread for a cat...

vidartf:
Do you do `git push -f` afterwards?

sem-geologist:
Maybe i simply close this and do proper pull request... I have no idea why git is for me like thread for a cat...

sem-geologist:
I did `git pull` and `git push`

vidartf:
Do these steps instead:
```
git rebase --onto hyperspy/RELEASE_next_patch hyperspy/RELEASE_next_minor patch-14
git push -f
```

No pull after rebase, and the `-f` option is important for the push.

sem-geologist:
I get 
```
First, rewinding head to replay your work on top of it...
could not detach HEAD

sem-geologist:
ok, I think I did it...
I had already new branch ready with copy of changes in case it would not work...
git sometimes makes me wanting to detach my real head.
5 checks? is this normal (double appveyor and travis...)

sem-geologist:
oh... yeah.. it actually went completely wrong... ok, I am closing this, and reopening with other branch...

PR #1370, opened 2017-01-09
Add default value to DictionaryTreeBrowser.get_item

vidartf:
Follows same pattern as `dict.get()`.

### DIFF
--- a/hyperspy/misc/utils.py
+++ b/hyperspy/misc/utils.py
@@ -431,8 +431,9 @@ class DictionaryTreeBrowser(object):
         else:
             return False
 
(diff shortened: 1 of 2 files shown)


francisco-dlp:
It needs a test and submission to RELEASE_next_minor instead.

vidartf:
My thought was that this could be used in #1283 to improve the code there, which is why it was submitted here. Too optimistic? :)
Agree on adding tests.

vidartf:
Tests added (and passing locally). (first time I've encountered a label race-condition on GH :D )

PR #1371, opened 2017-01-09
Fix slashes in commit hooks for windows

### LINKED ISSUE #1327: git hooks fail on windows
It seems the script creating the git hooks get confused about back-slashes.
```
.git/hooks/post-checkout: line 4: c:usersvidarminiconda3python.exe: command not found
```
The line in the hook script looks like this: `c:\users\vidar\miniconda3\python.exe ....`, which is incorrect, as backslashes need to be escaped in shell scripts.

vidartf:
Fixes #1327.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -194,11 +194,14 @@ def find_post_checkout_cleanup_line():
 # after changing branches:
 if os.path.exists(git_dir) and (not os.path.exists(hook_ignorer)):
     exec_str = sys.executable
-    if os.name == 'nt':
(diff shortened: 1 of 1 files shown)


francisco-dlp:
It looks good to me, @sem-geologist, would you like to take a look at this?

sem-geologist (approved):
i am not windows expert, this should work.
However I am not very happy with hook approuch (there is issue where it is discussed) as it recompiles/ recytonizes code even ir nothing changed... for now it willwork, but I think I will have to find more elegant way than hooks... unfortunatelly nothing comes to mind.

PR #1373, opened 2017-01-10
enh-test-fix: add min_environment to travis

sem-geologist:
to replicate hyperspy installations
on minimal fresh linux system.
(duplicate of closed and broken (with git) #1369)


francisco-dlp:
@vidartf, is this ready to merge?

sem-geologist:
During review of #1369, @vidartf suggested to move `pip3 install coverage coveralls` from `script` section to `install` section.
I think there are some space to optimize the speed of travis, (which allready takes long) it is not possible to do so when everything works and all tests passes ok, but when it fails, I think installing tools should be in the `script` part so we can save time in failing tests. I did so for minimal env. But I think it would be also good idea to do so to major env?

sem-geologist:
also there is pip line which tries to install traits and traitsui. Why it is needed, it is allready in setup.py dependencies

sem-geologist:
argh... I am pulling my hair... it is failing on bcf_test, but the fix is already in, just in minor branch...

sem-geologist:
What will happen if I will add the fix to this branch? is there going to be conflict later in merging the branches?

francisco-dlp:
No, but how did a fix ended up in RELEASE_next_minor?

francisco-dlp:
You are right regarding traits and traitsui: it is unnecessary to install them "manually". The same can be said of requests and tqdm.

sem-geologist:
because #1357, bcf fixes (metadata, some bug fixes, test additions) I worked recently got merged to RELEASE_next_minor...

sem-geologist:
btw, why minor branch is default branch on github?

francisco-dlp:
Sorry for not spotting that it was pointing to the wrong branch before merging. Will you try rebasing?

Regarding the default branch...it's just a convention. Would you choose a different branch as default?

sem-geologist:
oh no... just not another rebasing... I already failed one half our ago...

hmm... I saw wrong branch is quite getting common for pull request tags. I don't know, I missed right branch already twice :D

sem-geologist:
and I get the same mess (just this time I made the mess on the copy of branch) with rebasing... I do exactly as @vidartf instructed. Somehow git is not cooperating with me.

tjof2:
The wrong branch thing will probably clear up once we're all used to it. It's like writing 2016 everywhere and not 2017...

Though making the relevant issue #1170 clearer/more prominent might help.

sem-geologist:
I guess I get rebasing mess due to that I work from tablet directly in guthub, and also on laptop, and commits where something is pulled from github to laptop (keeping update) brakes i.e. cherry-picking, so I guess the same thing makes rebase go nuts...

sem-geologist:
Oh, I remember why I aimed bcf fixes to RELEASE_next_minor – I wanted after the merge to start implementing lazy bcf loader for #1219.

francisco-dlp:
Normally everything that gets into RELEASE_next_patch is merged into next_minor and next_major. However, I couldn't find a way to make the process automatic so sometimes (as e.g. now) things get out of sync.

francisco-dlp:
By the way, do you know why [this cythonize test](https://circleci.com/gh/conda-forge/hyperspy-feedstock/35?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link) is failing in [conda-forge](https://github.com/conda-forge/hyperspy-feedstock/pull/6)? It doesn't fail in my local machine.

sem-geologist:
@francisco-dlp  to answer to your last question: at first sight I see some severe python version missmatch, conda py35_0 and later I see python 3.4 pyc files created...
particulary py35 conda build tools, and then python 3.4...

sem-geologist:
is #1374 ok?

francisco-dlp:
Yes, I've just merge it.

Are you sure that version mismatch is the issue? (It builds 2 versions p34 and py35).

sem-geologist:
It is ready for review and/or eventual merge (if my move of installing coverage and coverall from script to `install` section is acceptable)

francisco-dlp:
@vidartf?

vidartf:
LGTM, except that if we're adding scikit-image as an obligatory dependency it should probably be added several places except setup.py?

vidartf:
Or actually, it might look like that is already there for the other locations (I've lost track of all the places dependencies need to go).

sem-geologist:
what is in setup.py is the hard dependancy (and only dependency needed), there is no need to add it anywhere else. pip will download automatically. Also look, tests does not fail, so it works.

francisco-dlp:
@vidartf has a point, in User Guide / Install is listed as an optional dependency.

sem-geologist:
so should I add it to User Guide?
I dont see it possible to make it optional as _signal complex uses it

vidartf:
Maybe `h5py` should also be removed as optional dependency in user guide, since it is also now (although not from this PR) in `install_req` ?

francisco-dlp:
Ideally we should make it optional. It's as simple as raising an exception when not installed explaining that the feature requires installing skimage. This can be annoying but, as most users install using the bundle or conda forge and both install skimage, there is not difference for them. If hyperspy every ships with a linux dist, then most of them install the recommend (but optional) packages by default.

francisco-dlp:
Well spotted!

sem-geologist:
I can't understand how hdf5 can be optional? it is needed to save hyperspy native format, I understand we could do GUI things optional, but this should be hard dependancy

francisco-dlp:
Maybe, but [the fact is that at present it is not required](https://github.com/hyperspy/hyperspy/blob/RELEASE_next_minor/hyperspy/io_plugins/__init__.py#L45).

sem-geologist:
* hyperspy/signal.py
* hyperspy/_signals/complex_signal.py
* hyperspy/io_plugins/emd.py 
* hyperspy/io_plugins/hdf5.py 

VS

`hyperspy/io_plugins/__init__.py`  (which thinks that it is optional)

francisco-dlp:
Good point, it's not optional then ;)

sem-geologist:
It is one of the reason why this min_environment is needed, to really get clear picture what is optional and what is no way

sem-geologist:
So what should I do? edit the Docs?

francisco-dlp:
Yes, that should do it.

vidartf:
🎉

PR #1375, opened 2017-01-10
Fix nb interaction test

francisco-dlp:
The `notebook_interaction_test` is not passing with `ipykernel > 4.5` when not attached to a notebook. This fix skips the test when that's the case. See ipython/ipykernel#210.

### DIFF
--- a/hyperspy/tests/model/test_model.py
+++ b/hyperspy/tests/model/test_model.py
@@ -327,7 +327,7 @@ class TestModel1D:
         m.append(hs.model.components1D.Offset())
         m[0].notebook_interaction()
         m[0].offset.notebook_interaction()
-        
+
     def test_access_component_by_name(self):
         m = self.model
         g1 = hs.model.components1D.Gaussian()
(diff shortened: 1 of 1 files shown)


vidartf:
For clarity, under what conditions will it not be attached to a notebook?

francisco-dlp:
You are right, this only checks if it's running in a IP kernel, which hopefully it is enough to avoid the test failure. Running from a shell this test is always skipped.

francisco-dlp:
@vidartf, is this ok to merge?

vidartf:
Sure, although I'm not sure if I understand whether this is a temporary or permanent fix.

francisco-dlp:
It fixes a test that was failing when running from anything but an ipython kernel by skipping it. Ideally we should find a way to test the widgets code when running elsewhere but I don't know if that's possible. In other words, I don't know if this fix is permanent of temporary, but it's certainly a fix.

vidartf:
Then merge for now, and leave as a mental TODO for the future (i.e. leave it until somebody reports what it is supposed to be testing as broken) ;)

PR #1379, opened 2017-01-11
rm: pointless cythonization test

sem-geologist:
which brakes cicircle test as it does not launch tests in place.

- [x] pull all roots of this test


sem-geologist:
travis passes - this is ready to go

francisco-dlp:
Not yet

sem-geologist:
osx and windows builds and tests are so slow...

PR #1380, opened 2017-01-11
Fix msa no signal type

### LINKED ISSUE #1341: MSA file read Signal_type issue
In cases where the saved MSA file has a field Signal_type but it is left blank when the MSA is written, the MSA read step (io_plugin/msa.py) fails because quantity is not assigned. An additional case where quantity is assigned to a default or an empty string (not sure about down-stream consequences of this in full) might be useful.

francisco-dlp:
Fixes #1341.


francisco-dlp:
Is somebody available to have a look at this so that we can release today?

ericpre:
Yes, this is good to merge!

PR #1383, opened 2017-01-13
Doc fix holo docs

francisco-dlp:
Some small fixes to the holography chapter.

### DIFF
--- a/doc/api/hyperspy._signals.rst
+++ b/doc/api/hyperspy._signals.rst
@@ -84,6 +84,14 @@ hyperspy._signals.eels module
     :undoc-members:
     :show-inheritance:
 
+hyperspy._signals.hologram_image module
+---------------------------------------
+
+.. automodule:: hyperspy._signals.hologram_image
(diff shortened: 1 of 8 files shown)


francisco-dlp:
@woozey, could you have a look at this and let us know if it's ok?

By the way, what about adding an example image to `hs.datasets.example_signals` and use it for the example in the User Guide instead of showing the commands without applying them to a real dataset?

woozey:
@francisco-dlp The changes look good to me. Regarding putting an example, I can do that easily. Would you like me to contribute it to this PR or doing it as a separate one? I guess a dataset from demos having just one object and one reference hologram should suit here too?

francisco-dlp:
@woozey, thanks. I'll merge this and you can add the example in a separate PR. By the way, in [RtD](http://hyperspy.readthedocs.io/en/latest/) there is always an up-to-date version of the `RELEASE_next_major branch`. It will update with the changes here after merging.

PR #1384, opened 2017-01-16
Add ui to bundle

francisco-dlp:
Distribute HyperSpyUI with the Windows Bundle

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -101,7 +101,7 @@ before_deploy:
   - ps: Add-AppveyorMessage "Setting up WinPython environment..."
   - ps: Start-FileDownload ('http://freefr.dl.sourceforge.net/project/nsis/NSIS%202/2.46/nsis-2.46-strlen_8192.zip') ../nsis_patch.zip
   - ps: if ((Get-FileHash '../nsis_patch.zip').Hash -ne '3BA22DDC0F14DBD75A9487EB6C9BD85F535E7038927D251103B97E0AAD94EEAD') { exit(1) }
-  - "7z x ../nsis_patch.zip -o%NSIS_DIR -aoa"
+  - "7z x ../nsis_patch.zip -o%NSIS_DIR% -aoa"
   - ps: Start-FileDownload ('http://nsis.sourceforge.net/mediawiki/images/e/eb/Textreplace.zip') ../Textreplace.zip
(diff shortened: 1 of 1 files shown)


vidartf:
Hyperspyui has a dependency on hyperspy, so I think the UI install needs to come after the hyperspy install to get things in the right order. I'll try to test this further tomorrow.

francisco-dlp:
Done, thanks!

vidartf:
Will test it in action tomorrow.

appveyor.yml
```diff
@@ -116,6 +116,7 @@ before_deploy:
   - "%CMD_IN_ENV% python.exe -c \"import sys; print(sys.version)\""
   - "%CMD_IN_ENV% python.exe setup.py recythonize"
   - "%CMD_IN_ENV% python.exe setup.py install"
+  - "%CMD_IN_ENV% pip install --upgrade hyperspyui"
```
vidartf: It seems the `--upgrade` flag is causing some problems: https://ci.appveyor.com/project/vidartf/hyperspy/build/1.2.0.358/job/atqrbqwvka3b1k4c

francisco-dlp:
Fixed!

vidartf:
One more change needed, due to some strange errors with pip:

```diff
-  - "%CMD_IN_ENV% pip install hyperspyui"
+  # Try to run twice as workaround for permission error
+  - "%CMD_IN_ENV% pip install hyperspyui || pip install hyperspyui"
```

francisco-dlp:
Feel free to push this and any other required changes directly to this branch (no PR required).

vidartf:
I was originally a bit concerned about tag contamination, but with the new direct pushing, it should be safe (since tags only go up when you use `git push --tags`).

vidartf:
Changes pushed. For an appveyor run with these changes (and a tag), see https://ci.appveyor.com/project/vidartf/hyperspy/build/1.2.0.380 . The resulting bundles are available from the artifacts tab under the different jobs.

francisco-dlp:
I've just tried it. The desktop integration didn't work for me but running from terminal works and it's great!

vidartf:
Is the UI integration something that needs to be fixed in this package or the bundle package?

vidartf:
PS: Any idea why the `test_binary_string` test is failing on Travis? 😕

francisco-dlp:
Yes, it's due to dill failing to read functions across Python versions. I'm working on a fix.

You're right about the desktop integration, it should be fixed elsewhere.

PR #1385, opened 2017-01-16
Fix: Holography docstrings, added example datasets and fixed docstrings

woozey:
see #1383


woozey:
@francisco-dlp Hopefully, it is what you asked for)

doc/user_guide/electron_holography.rst
```diff
@@ -37,6 +37,8 @@ position and size as parameters:
 
 .. code-block:: python
 
+    >>> import hyperspy.api as hs
+    >>> im =  hs.datasets.example_signals.object_hologram()
     >>> wave_image = im.reconstruct_phase(sb_position=(<y>, <x>), sb_size=sb_radius)
```
francisco-dlp, line 43: What about also adding some plotting lines and figures? This could help users with e.g. plotting the unwrapped phase.

hyperspy/misc/example_signals_loading.py
```diff
@@ -45,3 +45,39 @@ def load_1D_EDS_TEM_spectrum():
     file_path = os.sep.join([os.path.dirname(__file__), 'eds',
                              'example_signals', '1D_EDS_TEM_Spectrum.hdf5'])
     return load(file_path)
+
+
+def load_object_hologram():
+    """
+    Load an object hologram image
+
```
francisco-dlp, line 57: This seems to be a NOTES sections (we follow [numpy's style](https://github.com/numpy/numpy/blob/master/doc/HOWTO_DOCUMENT.rst.txt).

hyperspy/misc/example_signals_loading.py
```diff
@@ -45,3 +45,39 @@ def load_1D_EDS_TEM_spectrum():
...
+    file_path = os.sep.join([os.path.dirname(__file__), 'holography',
+                             'example_signals', '01_holo_Vbp_130V_0V_bin2_crop.hdf5'])
+    return load(file_path, signal_type='hologram')
+
+
+def load_reference_hologram():
+    """
+    Load a reference hologram image
+
+    - Sample: Fe needle with YOx nanoparticle inclusions [Migunov, V. et al.
```
francisco-dlp, line 80: As above.

francisco-dlp:
Other than the small comments above, it looks good, thanks!

woozey:
It seem to be ready now)

francisco-dlp:
Looks good! One more thing that I didn't spot before (sorry for that): it may be good to add some extra [metadata](http://hyperspy.org/hyperspy-doc/current/user_guide/metadata_structure.html) to the files and maybe also a telling title? e.g.:


```python

im.metadata.set_item("Sample.description", "Fe needle with YOx nanoparticle inclusions")
im.metadata.set_item("General.authors", "Migunov, V. ")
im.metadata.set_item("General.doi", "10.1063/1.4916609")
```

woozey:
Done. Metadata added now)

> Out[48]: 
├── Acquisition_instrument
│   └── TEM
│       ├── Holography
│       │   └── Biprism_voltage = 130.0
│       ├── Tilt_alpha = 5.0
│       ├── acquisition_mode = Lorentz TEM holography
│       ├── beam_energy = 300.0
│       ├── dwell_time = 10.0
│       ├── microscope = FEI Titan G2 60-300 HOLO at ER-C
│       └── microscope_doi = 10.17815/jlsrf-2-70
├── General
│   ├── authors = Migunov, V. 
│   ├── doi = 10.1063/1.4916609
│   ├── original_filename = 01_holo_Vbp_130V_0V.dm3
│   └── title = Hologram of Fe needle
├── Sample
│   ├── credits = Prepared by London, A.
│   └── description = Fe needle with YOx nanoparticle inclusions
└── Signal
    ├── binned = False
    └── signal_type = hologram

francisco-dlp:
The new metadata must be added to the metadata specification. Some comments/suggestions:

* `Holography.Biprism_voltage`. What about simply `TEM.biprism_voltage`? Or `TEM.Biprism.voltage` if there may be more parameters associated to the biprism that we would like to add to the specification. Note that we capitalize the nodes but not the leaves.
* `Tilt_alpha`: currently we have `tilt_stage` instead. I agree that it would have been a better idea to have a `Stage` node with e.g. `x`, `y`, `z` coordinates and `alpha` and `beta` tilt angles. @ericpre, what do you think about it? For this PR, wouldn't `tilt_stage` be enough?
* `acquisition_mode` currently this has to be `TEM` or `STEM`.
* `dwell_time`: in hs md specifications there is not dwell time in the `TEM` node. I think that it is more consistent with the current structure to add `TEM.Detectors.Bright_field.dwell_time`—by the way, would "exposure" be better?
* `microscope_doi`: I think that it is unusual for a microscope to have a doi, what about adding this to ``General.notes``?
* `author`: we follow the latex syntax which requires the full first name (sorry, that's my mistake).
* `Sample.credits`: this is not part of the hspy md specifications and I think that it is a good idea to add it. @ericpre (and others), what do you think?

woozey:
> dwell_time: in hs md specifications there is not dwell time in the TEM node. I think that it is more consistent with the current structure to add TEM.Detectors.Bright_field.dwell_time—by the way, would "exposure" be better?

This part I actually got using `hs.load` on a dm3 file. Though I used the version before the fix and it also failed assigning `acquisition_mode`... Here more precise would be `TEM.Detector.CCD.exposure_time`. Being even more precise CCD would be just one of the TEM camera types as nowadays CMOS detectors getting very popular. So maybe even `TEM_camera` would be more appropriate, what do you think?

> microscope_doi: I think that it is unusual for a microscope to have a doi, what about adding this to General.notes?

Indeed, this is a new practice at ER-C, which made for evaluation of the facility usage.

> Sample.credits: this is not part of the hspy md specifications and I think that it is a good idea to add it. @ericpre (and others), what do you think?

I'll keep this implementation for the moment and will also list it in specs. But feel free to make another decision, I'll change it if needed.

ericpre:
+1 on stage node.
ideally +1 on `Detector.Camera` node, with its exposure time and possibly other specifications (camera type, etc.). However, moving the `exposure time` to another node will imply a small API break... and to be consistent, `dwell_time` (for STEM) should be also moved to a corresponding detector type... But this is maybe for another PR!
+1 on `Sample.credits`.

francisco-dlp:
If I got it right, this PR could just add the biprism voltage and sample credits and a separate PR should deal with the other improvements, sounds good!

woozey:
> If I got it right, this PR could just add the biprism voltage and sample credits and a separate PR should deal with the other improvements, sounds good!

What about `exposure`?

francisco-dlp:
We do have [TEM.exposure](http://hyperspy.readthedocs.io/en/latest/user_guide/metadata_structure.html), and hspy should have stored the information there instead of `TEM.dwell_time` in this case. So, this may be a bug to deal with in another PR. It won't be an issue for this file because we're going to increase the specification version number when fixing it and future versions of HyperSpy should put the info in the right place. We should open an issue with all the things that we have to fix discovered in this thread.

woozey:
I used some creativity putting my vision of how the metadata of camera and biprism should look like, see  changes in `doc.user_guide.metadata_structure`. Please let me know if I overdone it, I can put everything back)) Also please let me know if the changes should be reflected somewhere else, I'm still not very confident with structure of HS.

francisco-dlp:
@ericpre, what do you think of this?

ericpre:
I opened an issue (#1396) to discuss more specifically about the stage and camera metadata.
In order to make things more consistent and less confusing, I would suggest to leave the `camera` and `stage` metadata specification, as it was before this PR. If the location of the exposure time in the metadata is modified in this PR, it would need to be done for all io plugin, so it's better to do it in another PR.

doc/user_guide/metadata_structure.rst
```diff
@@ -302,9 +312,64 @@ live_time
     The time spent to record the spectrum in second, compensated for the
     dead time of the detector.
 
+TEM_camera
```
ericpre: I think it's better to leave the camera metadata spec to #1396.

woozey: No problem!

Issue #1307, opened 2016-10-14
Permanent markers


magnunor:
Currently when using `s.add_marker`, the marker will only be temporary: it will disappear if the plot is closed, then replotted.

It would be nice if these markers could be permanently added to a signal by adding them to the metadata, which would be checked when plotting.

In addition, the markers should be stored when saving the signal as a HDF5-file.

First step could be to add a `to_dictionary` function to the markers, so they could easily be saved in the HDF5.

thomasaarholt:
Would these sit in the `s.metadata` or elsewhere?

magnunor:
This was added in https://github.com/hyperspy/hyperspy/pull/1431

thomasaarholt:
I don't think that PR is what you think it is :)

magnunor:
You probably got the initial message, which linked to 143. Fixed now :)

thomasaarholt:
It was actually my ridiculously fast reflexes that opened the issue to see if the PR had been merged.

PR #1401, opened 2017-01-26
Implementing permanent markers

magnunor:
Currently when using `s.add_marker`, the marker will only be temporary: it will disappear if the plot is closed, then replotted. In addition, including many markers is really slow, due to each marker being plotted separately.

This implementation tries to improve this by:
- Possibility to add markers to a list in a signal, which is plotted by using `s.plot(plot_markers=True)`
- Plot the markers using matplotlib `Collections`, which are used to plot large amounts of objects at the same time.

https://github.com/hyperspy/hyperspy/issues/1307

Todo list:
- [ ] Documentation
- [ ] Unit tests
- [ ] Implement for all markers
- [x] Implement updating marker plotting when changing navigation position
- [ ] Ability to store markers in HDF5-file (i.e. save them in a signal)


magnunor:
Any thoughts of this way of implementing permanent markers? I'm breaking a little bit on how the markers currently work, by using matplotlib `Collections` to plot things. To do this, we got to gather all the plotting parameters in one place, and I currently do that by having a big function inside the `BaseSignal` class.

The `deepcopy` function does not seem to like me storing the `marker` data in `s.metadata`, so currently it is not working with 3D- or high dimensional data.

Simple example on how to use this:
```Python
import hyperspy.api as hs
import numpy as np

####################
s = hs.signals.Signal2D(np.random.random(size=(100, 100)))

s.metadata.set_item('Markers', [])
marker = hs.plot.markers.point(70, 20, color='red')
s.metadata.Markers.append(marker)

marker = hs.plot.markers.line_segment(50, 20, 80, 50, color='blue')
s.metadata.Markers.append(marker)

marker = hs.plot.markers.rectangle(10, 20, 40, 50, color='blue')
s.metadata.Markers.append(marker)
```

A lot of markers:
```Python
s1 = hs.signals.Signal2D(np.random.random(size=(100, 100)))
s1.metadata.set_item('Markers', [])
for i in range(5000):
    marker = hs.plot.markers.point(
            np.random.randint(0,100),
            np.random.randint(0,100),
            color='red')
    s1.metadata.Markers.append(marker)

for i in range(1000):
    marker = hs.plot.markers.line_segment(
            np.random.randint(0,100),
            np.random.randint(0,100),
            np.random.randint(0,100),
            np.random.randint(0,100),
            color='green',
            linewidth=2)
    s1.metadata.Markers.append(marker)

for i in range(100):
    x1 = np.random.randint(0,80)
    x2 = x1 + 3 
    y1 = np.random.randint(0,80)
    y2 = y1 + 3
    marker = hs.plot.markers.rectangle(
            x1, y1, x2, y2,
            color='yellow')
    s1.metadata.Markers.append(marker)
```

francisco-dlp:
In order to make the markers deepcopy-friendly they need to have a ``_to_dictionary`` method.

I'll give you feeback on the code later in the day.

magnunor:
This is now working for multidimensional datasets, with several thousand markers at the same time.

Still not certain if the current implementation is the best, but it is at least working!

Some things currently not working:
- The event which triggers this replotting of the markers when changing navigation positions is not disconnected when the plot is closed. What is the best way of doing this?
- How should the `horizontal_line` marker be plotted? The marker itself uses the matplotlib function `axvline`, but that function can only plot one line at a time. Which can be very slow if one wants to plot many lines in the same plot.

Script to test the implementation:
```Python
import hyperspy.api as hs
from numpy.random import randint

s = hs.signals.Signal2D(randint(0, 100, size=(10, 10, 100, 100)))
s.markers = []

for i in range(1000):
    marker = hs.plot.markers.point(
            randint(0, 100, size=(10, 10)),
            randint(0, 100, size=(10, 10)),
            color='red',
            linestyle='solid',
            linewidth=2)
    s.markers.append(marker)
for i in range(100):
    marker = hs.plot.markers.vertical_line_segment(
            randint(0, 100, size=(10, 10)),
            randint(0, 100, size=(10, 10)),
            randint(0, 100, size=(10, 10)),
            color='blue',
            linestyle='solid',
            linewidth=2)
    s.markers.append(marker)
for i in range(100):
    marker = hs.plot.markers.horizontal_line_segment(
            randint(0, 100, size=(10, 10)),
            randint(0, 100, size=(10, 10)),
            randint(0, 100, size=(10, 10)),
            color='yellow',
            linestyle='solid',
            linewidth=2)
    s.markers.append(marker)
for i in range(100):
    marker = hs.plot.markers.line_segment(
            randint(0, 100, size=(10, 10)),
            randint(0, 100, size=(10, 10)),
            randint(0, 100, size=(10, 10)),
            randint(0, 100, size=(10, 10)),
            color='green',
            linestyle='solid',
            linewidth=2)
    s.markers.append(marker)
```

magnunor:
I added a function for which adds permanent markers to a signal, which checks if the navigation dimensions are compatible between the marker and the signal.

I'm guessing this functionality can be added directly to the current `add_marker` function, where permanent or not can be toggled with `s.add_marker(marker, permanent=True)`.

vidartf:
You have merge conflicts (which unfortunately also prevents CI from running/reporting).

magnunor:
I merged RELEASE_next_minor into this branch, so CI is running again

magnunor:
Parts of this has been resubmitted in https://github.com/hyperspy/hyperspy/pull/1431.

The Matplotlib `Collections` functionality will be implemented in a future pull request, after https://github.com/hyperspy/hyperspy/pull/1431 is merged.

vidartf:
Time to start up the `Collections` bit again :)

magnunor:
I actually don't think we'll need it now, the only slow part is adding the markers to the signal. But that can easily be fixed by either manually populating the `metadata.Markers`, or to allowing `s.add_marker` to get a list of markers.

francisco-dlp:
multidimensional markers may still benefit from it, isn't it?

magnunor:
Not really, since the markers aren't readded when changing the nav-position, their position is simply changed. Making it very fast.

Try it:
```Python
import hyperspy.api as hs
import numpy as np
from numpy.random import random

s_points = hs.signals.Signal2D(np.random.random(size=(3, 100, 100)))
for i in range(800):
    marker = hs.markers.point(
            random((3))*100,
            random((3))*100,
            color='red')
    s_points.add_marker(marker, permanent=True, plot_marker=False)
```
Only really slow part is adding the markers

magnunor:
But I haven't tested this extensively for all markers. The point marker is at least fairly quick. I'll play around with this a bit more for my Atomap stuff, and see if we actually need to do any improvements. But I think all that can wait until after the 1.2 release (https://github.com/hyperspy/hyperspy/pull/1463).

francisco-dlp:
Yes, it seems fast enough for the purpose. Enabling passing an iterable of markers to ``add_marker`` sounds good.

magnunor:
I implemented faster adding here: https://github.com/hyperspy/hyperspy/pull/1467

PR #1406, opened 2017-01-27
Fix s.split()

thomasaarholt:
Fixes a small bug in the s.split() which prevented stacked signals from being split after saving.

I have not investigated the cause of why, in the metadata, ’s.metadata._HyperSpy.Stacking_history.step_sizes’ is read with type ‘np.int64’, rather than ‘int’


thomasaarholt:
Wrong branch...

Issue #1356, opened 2016-12-14
unbcf_fast library?


thomasaarholt:
So, now that Jupyter Notebook reports warnings inline on Mac, I'm aware of everything that can clutter the screen.
When I load hyperspy, I get the following warning:
```python
unbcf_fast library is not present...
Falling back to slow python only backend.
```

I couldn't find anything on what the unbcf format is, or why it didn't come with my git install of hyperspy. If it's not coming back, we should remove it/the warning.

to266:
It's there and alive, you probably just did not install it fully / correctly. It's a cython library that has to be compiled on each machine before use. 

I would try `python setup.py recythonize` in the hyperspy folder

thomasaarholt:
Aha! What does it do though?

thomasaarholt:
Installed cython, ran your command. It mentioned it was cythonizing the fast_unbcf, but then i get this error:
```python
(hs) wolf3947:hyperspy thomas$ python setup.py recythonize
/Users/thomas/miniconda3/envs/hs/lib/python3.5/site-packages/setuptools/dist.py:341: UserWarning: The version specified ('2.0-git-156-gc6850ed') is an invalid version, this may not work as expected with newer versions of setuptools, pip, and PyPI. Please see PEP 440 for more details.
  "details." % self.metadata.version
running recythonize
```

I updated to setuptools v31 without any change. I'm not sure what it is that is of version 2.0? Maybe my hyperspy branch?

francisco-dlp:
I wouldn't worry about the warning. That's because you're installing from the RELEASE_next_major branch and `setup.py` automatically adds the hashtag to the version number.

thomasaarholt:
Ah, I see. Thanks! Why am I installing from _major though, I wonder.

tjof2:
Tangentially related, why is it called `unbcf`? I can appreciate it might un-BCF something (whatever that means), but surely `bcf` would have been a clearer name - as @thomasaarholt notes, the `unbcf` format isn't a thing.

Furthermore, how do I stop this from recompiling every time I switch branches? Alternatively, can anything be done to make it compile quicker?

francisco-dlp:
It should only recompile when Cython code changes when switching branches. If it doesn't do that, then you've found a bug. Regarding the naming question, only @sem-geologist knows. I find the name enjoyable though.

sem-geologist:
[...]

PR #1409, opened 2017-01-28
FIX: unbcf_fast missing cython warning

sem-geologist:
to address the latest complain in #1356,
the warning message of unbcf_fast missing were redone:
1. at loading the library the message were moved from logger.warrning to logger.info
2. the more detailed warning is shown if unbcf_fast is not present only when real loading takes place


hyperspy/io_plugins/bcf.py
```diff
@@ -1135,6 +1137,14 @@ def bcf_hyperspectra(obj_bcf, index=0, downsample=None, cutoff_at_kV=None,
     """ Return hyperspy required list of dict with eds
     hyperspectra and metadata.
     """
+    global warn_once
+    if (fast_unbcf == False) and warn_once:
+        _logger.warning("""unbcf_fast library is not present...
+Parsing with slow python only backend.
```
tjof2: Parsing BCF with Python-only backend. If parsing is uncomfortably slow, first install cython, then reinstall hyperspy. For more information, check the 'Installing HyperSpy' section in the documentation.

PR #1412, opened 2017-01-30
Add filter_zero_loss_peak argument to spikes_removal_tool and fix one bug.

ericpre:
It should be PR to next_minor, but I initially check out from next_patch... I can change it to next_minor if necessary.
It may also be a good opportunity to add some tests (not for the ui) for the spikes_removal_tool, I guess it should be possible to test the different method/functions without using the ui?

- [x] Add filter_zero_loss_peak argument
- [x] Add documentation
- [x] Add tests
- [x] Fix bug with update pointer position


francisco-dlp:
Wouldn't this be equivalent?:


```python

s.isig[start:end].spikes_removal_tool()

```

ericpre:
Yes, it is! Thanks :) I will add it to the documentation!

Is there a clever way to set two intervals, that I would have miss? For example to ignore a zero-loss peak?

francisco-dlp:
You can pass a ``signal_mask`` to do that. That's simply a boolean matrix. ``Model1D`` has a sort of interactive signal mask editor (e.g. ``set_signal_range`` etc) that would be useful to turn into an independent function to create signal masks more conveniently.

ericpre:
Adding a `get_signal_range` method to `Signal1D`, that works in a similar manner than `set_signal_range`  and returns a mask?

francisco-dlp:
Something like that. Currently `Model1D` has:

* ``set_signal_range``
* ``remove_signal_range``
* ``reset_signal_range``

They enable editing a mask interactively. Because the mask is stored in the model as an attribute for internal consumption  there is no method to get the mask as you propose.

For the model it makes sense to store it internally, but does it make sense to store it internally for the signal? If not, it could be implemented as a ``Signal`` that includes the above methods e.g.

```python

mask = s.get_signal_mask() # This returns a mask masking nothing by default
mask.set_signal_range() # It remembers its parent mask and, if no signal range is passed, it plots s for interactive selection
mask.set_signal_range(signal=s2) # Interactively edit the mask with a different signal
s.spikes_removal_tool(signal_mask=mask)

```

hyperspy/_signals/eels.py
```diff
@@ -299,6 +299,31 @@ def estimate_zero_loss_peak_centre(s, mask, signal_range):
             substract_from_offset(without_nans(zlpc.data).mean(),
                                   also_align + [self])
 
+    def spikes_removal_tool(self, signal_mask=None,
+                            navigation_mask=None,
+                            signal_start=None,
+                            signal_end=None,
+                            filter_zero_loss=False,
+                            zero_loss_width=5.0):
+        if filter_zero_loss:
```
francisco-dlp: This option is really useful! Could implemented without the ``signal_start`` and ``signal_end`` kwargs?

hyperspy/_signals/eels.py
```diff
@@ -299,6 +299,31 @@ def estimate_zero_loss_peak_centre(s, mask, signal_range):
...
+            (signal_axis, ) = self.axes_manager[self.axes_manager.signal_axes]
+            axis = signal_axis.axis
+            mini_value = zlpc.data.mean() - zero_loss_width/2
+            maxi_value = zlpc.data.mean() + zero_loss_width/2
+            mask = -((mini_value <= axis) & (axis <= maxi_value))
+            if signal_mask:
+                signal_mask = mask & signal_mask
+            else:
+                signal_mask = mask
+            import matplotlib.pyplot as plt
```
francisco-dlp: We discourage plotting things outside of the signal plot as much as possible. An alternative would be to plot in a different way (maybe using a marker) (or simply not to plot as in model) the masked part.

hyperspy/_signals/eels.py
ericpre: I will make close this one and make a new PR.

hyperspy/_signals/eels.py
ericpre: actually, it is some left over of troubleshooting... but it would be good to find a way to display where is the "masked" area.

ericpre:
Ready for review.

ericpre:
Merged with upstream and resolved conflict.

doc/user_guide/signal1d.rst
```diff
@@ -121,8 +121,25 @@ Spike removal
 --------------
 
 :py:meth:`~._signals.signal1d.Signal1D.spikes_removal_tool` provides an user
-interface to remove spikes from spectra.
+interface to remove spikes from spectra. It is possible to use this tool
+on a specific interval of the data by :ref:`slicing the data
+<signal.indexing>`. For example, to keep the signal between indexes 8 and 17:
```
tjof2: I think "keep" is wrong, perhaps "to work only on the signal between indexes 8 and 17"

tjof2 (commented):
Minor doc change @ericpre otherwise looks good!

tjof2 (approved):

tjof2:
If you can resolve the conflicts this should be good to go?

tjof2:
@francisco-dlp can you check this over again since you initially reviewed it?

tjof2 (approved):
Subject to resolving conflicts.

Issue #1372, opened 2017-01-09
Migrate to py.test?


francisco-dlp:
It seems that `nose` is no longer under development and, although its being rewritten in a new project called [nose2](https://github.com/nose-devs/nose2),  most projects these days seem to use py.test. Other projects (e.g. matplotlib/matplotlib#6731) have started migrating to py.test, should we?

magnunor:
Switching away from `nose` sounds like a good idea, due to lack of development.

I've been using `unittest` for my own stuff, and it seems like the main advantage of py.test is less boilerplate, easier assertions and running the tests in parallel?

vidartf:
As far as I understand, `py.test` also allows for plugins :)

PR #1413, opened 2017-02-01
Migrate to pytest

francisco-dlp:
Migrate all tests to pytest. Note that some tests are not passing: those are real bugs that must be addressed in a separate pull request.

This PR also removes the instructions to install using Archlinux and Ubuntu because they were obsolete. As we lack tests for these instructions, I reckon that it's better to remove them as wrong instructions are worst than no instructions at all.

As you can see in the diffs, writing pytest tests is really easy if you're used to nose. Main differences:

* Fixtures: instead of ``def setUp(self)`` -> ``def setup_method(self, method)``
* No more ``nt.assert_x``. With pytest is as simple as ``assert ...``

Addresses #1372.


francisco-dlp:
The tests seem to be working fine (the failures are bugs). I suggest merging asap to avoid branching from current (obsolete) nose tests in RELEASE_next_minor, making PRs harder to merge.

francisco-dlp:
In other words, is anybody against merging this?

to266:
I'm for this, although I suppose I'll have to re-invent testing the lazy things... Though from what I've seen, it should be easier than the nosetest :)

ericpre:
Except CI, which is not working, this is good -the tests work nicely locally. I haven't had time to look in details to be considered as "reviewed", though, and I have to leave now.

Regarding the failing tests, you are talking about the FEI ones, right? If so, this is more or less a bug! ;)

francisco-dlp:
Thanks, I'll wait until tomorrow afternoon before merging in case that somebody thinks differently.

francisco-dlp:
Yes, I'm talking about the FEI ones. What do you mean with CI not working?

ericpre:
Sorry for the noise, I haven't seen your last commit: appveyor and travis are fine!

hyperspy/tests/io/test_protochips.py
```diff
@@ -75,47 +78,47 @@ def get_datetime(dt):
...
+        assert om.Start_time == datetime_gas_cell[2]
+        assert om.Holder_Pressure_units == 'Torr'
+        assert om.Tank1_Pressure_units == 'Torr'
+        assert om.Tank2_Pressure_units == 'Torr'
+        assert om.Vacuum_Tank_Pressure_units == 'Torr'
+        assert om.Time_units == 'Milliseconds'
+        assert om.User == 'eric'
 
 
 class test_ProtochipsGasCellCSVNoUser():
```
ericpre: This part haven't been updated and these tests are not collected (and run) by pytest, possibly because they need to be renamed to `TestProtochipsGasCellCSVNoUser`. There are a few other one, below.

stdeb.cfg
```diff
@@ -6,7 +6,7 @@ Depends:
     ipython (>= 2.0),
     python-traits,
     python-traitsui,
-    python-nose,
+    python-pytest,
     ipython-qtconsole,
     ipython-notebook,
     python-h5py,
```
ericpre, line 12: Just out of curiosity, why some package are python-something and others are python3-something?

ericpre (changes requested):
There is a mention of assert_almost_equal in the dev guide, that may need a small update.
With nose, appveyor runs 1526 tests and with pytest, it runs 1206 tests, same for travis.
The folder doc/package_lists and its content need to be removed since it is outdated since a while.

Otherwise, this looks good to go!

francisco-dlp:
Good catch!

Regarding stdeb, this is also broken! All of them should be python3-*.

francisco-dlp:
I've fixed the missing protochips tests. The mismatch between nose and pytest seems to be caused by the test generators in e.g. the dm3 tests. I'll try rewriting them with pytest parametrizations.

stdeb.cfg
sem-geologist: all of them should be python3? It is going to be so quite to the future on debian/ ubuntu and derivatives...

francisco-dlp:
@ericpre, I have rewritten the test generators using pytest parametrizations.

The total number of tests is now 1227 down from 1526 with nose. I think that the missing ones are due to compressing the three previous checks in dm3 and dm4 in one single test although I'm not 100% sure as comparing the output of pytest and nose is not straightforward.

francisco-dlp:
Actually we must still have missing tests, around ~179, because the changes to the dm3 tests have decreased the number by only 120, but there are 299 missing :(

ericpre:
Thanks for that, pytest seems to be a bit nicer to use than nose! If I can spot missing tests, I will let you know.

ericpre:
If I run `nosetests --collect-only ` on the `RELEASE_next_patch` branch, nose actually collects 1205 tests, while pytest collects 1225 tests (`pytest --collect-only`).

Travis ran [1325 tests](https://travis-ci.org/hyperspy/hyperspy/jobs/196550131) and same for [Appveyor](https://ci.appveyor.com/project/hyperspy/hyperspy/build/1.0.1763/job/ogj4nbh9t5dcs1er) on the `RELEASE_next_patch` branch.

Any idea about where could the difference come from? Do you observe the same thing?

The details of the difference between nose and pytest that I got on my system.
+10 in tests/io
+1 in tests/model
+1 in tests/mva
-1 in tests/samfire
+9 in tests/signal

francisco-dlp:
I think that the difference using ``--collect-only`` when running nose is
that in this mode it counts the test generators as just 1 test. pytest
counts the test generators properly with ``--collect-only``. We should
compare the ``pytest`` branch to ``RELEASE_next_minor`` as I have branched
from it to translate all the tests. Do you have a script to parse and
compare the output?

ericpre:
No I did it manually...

On RELEASE_next_minor, I still get only 1226 tests (with nose obviously).
OK, in RELEASE_next_minor, there are a couple of feature more, so more tests!

francisco-dlp:
So they match :) , although probably just by chance?

francisco-dlp:
I've just compared the output of ``nosetests -v`` and ``pytest -v`` using Python sets and they match  i.e. pytest is not missing any tests. @ericpre, you were right: with the ``--collect-only`` flag nose counts the tests properly, without it it claims to work more than it actually does.

@ericpre, should I proceed with the merge?

ericpre:
Yes, all the tests are alright, it's good to have made sure that this is the case.
Maybe a quick update in the "write tests and documentation" bit of the dev guide would be good?

Once, this is merged, I will come back to #1253 with pytest!

francisco-dlp:
Thanks! I'll push some a couple of commits to make it run in parallel and the dev guide update.

to266:
I'll soon PR a couple of improvements to this

to266:
https://github.com/francisco-dlp/hyperspy/pull/33

Issue #1415, opened 2017-02-02
Can not save to EMD without certain metadata


magnunor:
Making a minimal signal and saving it as an EMD file does not work:

```Python
import hyperspy.api as hs
import numpy as np

s = hs.signals.Signal2D(np.random.random(size=(100,100)))
s.save("test.emd")
```

Gives the error: `AttributeError: 'DictionaryTreeBrowser' object has no attribute 'General'`

magnunor:
Fixed with https://github.com/hyperspy/hyperspy/pull/1424

Issue #1398, opened 2017-01-22
BUG: O(n) growth of original_metadata with stacking


sem-geologist:
## The current hyperspy behaviour causing issue with memory and slowness
When loading sliced dataset and using `stack=True`, hyperspy concatenates all original metadata from all slices which is deepcopied and attached to new subsets/slices of hypercube.

## Known issues
1. Plotting with lines of EDS spectra is getting very slow if stack is big
2. Accessing slices is very slow if format contains very rich original_metadata (e.g. Zeiss tif files):
![original_metadata_time_dependancy](https://user-images.githubusercontent.com/12324068/45542698-e96da800-b812-11e8-9041-fd1a54cdbd58.png)
(for description of benchmark see below)

## Known hack/workaround 
after loading the dataset stack, 
```python
thingy = hs.load('FIBSLICE*.tif', stack=True)
```
override whole concatenated original metadata:
```python
from hyperspy.misc.utils import DictionaryTreeBrowser 
original_metadata = thingy.original_metadata  # optional, if you need something from it later
thingy.original_metadata = DictionaryTreeBrowser()
# enjoy de-crippled speed!!!
```

## Possible fixes
1. do not allow to accumulate the original metadata while stacking
2. do not copy original metadata when slicing
alternatively:
1. make original_metadata read-only and pass only the reference (proposed by @vidartf)

(see below, originally the third point was to prevent slicing when plotting EDS, but after finding that this bug affects not only EDS line plotting, but everything stacked, th point is no more considered, as points 1 or 2 would deal with this) 
**the original message was moved down.

sem-geologist:
Hello,
[...]

PR #1416, opened 2017-02-02
Fix bug when saving minimal signal in EMD-file format

magnunor:
Bug-fix for https://github.com/hyperspy/hyperspy/issues/1415

In this implementation I'm not checking `original_metadata`, due to discussion like the one in https://github.com/hyperspy/hyperspy/issues/1398#issuecomment-274500930. Do people think I should include look-up in the `original_metadata` as well (if it exist in the signal)?

- [ ] Add test for saving minimal signal

Should I wait for the `pytest` to be merged before making the test?


magnunor:
Sending this to RELEASE_next_minor instead (due to switch to pytest): https://github.com/hyperspy/hyperspy/pull/1424

Issue #1414, opened 2017-02-02
FEI tests issue


ericpre:
As mentioned in #1413, there are two questionable FEI tests about the import of linescans (1D navigation axis). More precisely, the issue seems to concern the name of the navigation axis.

In the current implementation the name of the navigation axis is `y`, mainly because of the way how linescans and spectrum images are imported. When I looked at it, I first thought that it was wrong, but since the linescan can be vertical, horizontal or oblique, we can't really use the same convention than for spectrum images (`x` and `y`).
Should we change the navigation axis name? Is it worth it?

francisco-dlp:
Maybe ``PositionX`` and Y can help determine if its "x", "y" or oblique?

ericpre:
Yes, but what name do we put if it's oblique?

francisco-dlp:
I would suggest "x". This shows that we may need to add a coordinates attribute to ``DataAxis`` for the cases like this one.

ericpre:
If we put `x` for a oblique linescan, it will be wrong and confusing because the convention (actually I don't know to what extend it's correct) is to have `x` as horizontal axis and `y` as vertical axis.
Maybe, something more generic like `Position`?

francisco-dlp:
Sounds good.

to266:
#1420

PR #1420, opened 2017-02-05
Fix fei tests

francisco-dlp:
Fix the name of single navigation axis that was sometimes set to ``y`` instead of ``x``. Note that this does not implement setting the axis name to ``Position`` for oblique scan as proposed by @ericpre  in #1414.

(This branches from #1413, for a clean diff see last commit.)

### DIFF
--- a/hyperspy/io_plugins/fei.py
+++ b/hyperspy/io_plugins/fei.py
@@ -485,7 +485,7 @@ def ser_reader(filename, objects=None, *args, **kwds):
     else:
         axes = []
         array_shape = [None, ] * int(ndim)
-        spatial_axes = ["x", "y"]
+        spatial_axes = ["x", "y"][:ndim]
         for i in range(ndim):
             idim = 1 + i if order == "C" else ndim - i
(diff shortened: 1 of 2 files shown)


ericpre:
Yes, it does the job, without complicating the code!

to266:
This needs a rebase/merge of the original pytest branch. Actually any reason why this is a patch, whereas #1413 goes to minor? Don't we want to patch the tests and then "trickle down" to minor/major?

francisco-dlp:
Sorry, I mislabeled it.

to266:
...Is there a way to force github to refresh the changes? Maybe you could rebase this on next_minor in order to show a clean diff?

francisco-dlp:
Only the label was wrong, I branched it from #1413. The diff will be cleaner was that one gets merged.

to266:
I merged #1413 before the last comment and hoped so too, but there we go.

francisco-dlp:
Merging Rnm into it did it.

PR #1424, opened 2017-02-06
Fix emd minimal saving 2

magnunor:
Bug-fix for #1415, resubmission of https://github.com/hyperspy/hyperspy/pull/1416

In this implementation I'm not checking `original_metadata`, due to discussion like the one in https://github.com/hyperspy/hyperspy/issues/1398#issuecomment-274500930. Do people think I should include look-up in the `original_metadata` as well (if it exist in the signal)?

(Sending this to RELEASE_next_minor, instead of RELEASE_next_patch due to the change to `pytest`. So I don't have to make a test for the old system first, then a new test for `pytest` when merging it into next_minor).

### DIFF
--- a/hyperspy/tests/io/test_emd.py
+++ b/hyperspy/tests/io/test_emd.py
@@ -67,6 +67,19 @@ def test_metadata():
     assert isinstance(signal, Signal2D)
 
 
+class TestMinimalSave():
+
+    def setup_method(self, method):
+        self.filename = 'testfile.emd'
+        self.signal = Signal1D([0, 1])
+
+    def test_minimal_save(self):
+        self.signal.save(self.filename)
+
+    def teardown_method(self, method):
+        remove(self.filename)
+
+
 class TestCaseSaveAndRead():
 
     def test_save_and_read(self):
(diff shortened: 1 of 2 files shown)


magnunor:
Travis and AppVeyor is failing due to the FEI plugin tests, the rest of the tests are passing.

francisco-dlp:
@magnunor, are happy with this?  If yes, why not merging it?

francisco-dlp:
(Sorry, the comment above was intended for #1432)

hyperspy/tests/io/test_emd.py
```diff
@@ -67,6 +67,19 @@ def test_metadata():
     assert isinstance(signal, Signal2D)
 
 
+class TestMinimalSave():
+
+    def setup_method(self, method):
+        self.filename = 'testfile.emd'
```
francisco-dlp: I would save the file in the tmp directory instead. In this way we may avoid future problems if we add more methods to this class and we run the tests in parallel. You could use e.g. `` with tempfile.TemporaryDirectory() as tmp:`` and this way there's no need for a tear down method.

magnunor: I changed it to:
```Python
with tempfile.TemporaryDirectory() as tmp:
    self.filename = tmp + '/testfile.emd'
```

PR #1425, opened 2017-02-06
Use sympy string expressions instead of functions for ``Parameter.twin_function``

### LINKED ISSUE #1397: Models using ``Parameter.twin_function`` may not load properly across Python versions
This was discussed identified and discussed in #1395. The cause is that we currently use dill to save ``Parameter`` "twin functions" to hdf5. However, there is not guarantee that dill will unpickle the functions properly across Python versions.

A solution could be using sympy instead. Sympy has the extra advantages that it would enable us to generate the gradient for twinned parameters automatically and, as the function is stored in a string, it's easier for humans to know what they do.

francisco-dlp:
Fixes #1397 by adding ``Parameter.twin_function_expr`` and ``Parameter.twin_inverse_function_expr``. Deprecates setting ``twin_function`` and ``twin_inverse_function``. When ``twin_function_expr`` is invertible its inverse is automatically set courtesy of sympy.

TODO
--------
* [x] Test the eels model with hartree-slater (not possible in unittests due to missing xs)

### DIFF
--- a/doc/user_guide/model.rst
+++ b/doc/user_guide/model.rst
@@ -523,15 +523,25 @@ For example:
             A	5.000000
             centre	0.000000
 
+.. deprecated:: 1.1.3
+    Setting the :py:attr:`~.component.Parameter.twin_function` and
+    :py:attr:`~.component.Parameter.twin_inverse_function` attributes. Set the
+    :py:attr:`~.component.Parameter.twin_function_expr` and
+    :py:attr:`~.component.Parameter.twin_inverse_function_expr` attributes
+    instead.
+
+.. versionadded:: 1.1.3
+    :py:attr:`~.component.Parameter.twin_function_expr` and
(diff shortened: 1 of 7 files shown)


francisco-dlp:

to266:
Why not include [_parse_substitutions](https://github.com/hyperspy/hyperspy/blob/RELEASE_next_minor/hyperspy/_components/expression.py#L22) from the `Expression`? I quite like how you used the `.format` of strings to pass things, but the substitutions also simplify the expression, etc. This would not subtract anything from the current implementation, just make hyperspy-sympy interface uniform.

Also the change is literally 
`sympy.sympify(value)` -> `_parse_substitutions(value)`, so quite simple :)

francisco-dlp:
That's a good idea, but ``_parse_substitutions`` is not in RNP. Could we merge this and add the new feature in RNm?

to266:
Oops, did not notice this is a patch. Sure

PR #1426, opened 2017-02-07
expression component substitutions

to266:
If the expression is inconvenient to write out in full (e.g. it's long and/or complicated), multiple substitutions can be given, separated by semi-columns. Both symbolic and numerical substitutions are allowed:

```python 
>>> expression = "h / sqrt(p2) ; p2 = 2 * m0 * e1 * x * brackets;"
>>> expression += "brackets = 1 + (e1 * x) / (2 * m0 * c * c) ;"
>>> expression += "m0 = 9.1e-31 ; c = 3e8; e1 = 1.6e-19 ; h = 6.6e-34"
>>> wavelength = hs.model.components1D.Expression(
 ... expression=expression,
 ... name="Electron wavelength with voltage")
```

### DIFF
--- a/doc/user_guide/model.rst
+++ b/doc/user_guide/model.rst
@@ -124,6 +124,19 @@ parameters for spectroscopy than the one that ships with HyperSpy:
     ... centre=0,
     ... module="numpy")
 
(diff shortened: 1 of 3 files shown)


francisco-dlp:
Cool!

PR #1429, opened 2017-02-07
Improve docstrings for EELSSpectrum and Signal1D

magnunor:
Tries to improve the docstrings so they pass the tests, however the GUI ones still fail due to no way of currently running those in the tests.

### DIFF
--- a/hyperspy/_signals/signal1d.py
+++ b/hyperspy/_signals/signal1d.py
@@ -98,18 +98,10 @@ def find_peaks_ohaver(y, x=None, slope_thresh=0., amp_thresh=None,
         contains position, height, and width of each peak
     Examples
     --------
-    >>> import numpy as np
     >>> x = np.arange(0,50,0.01)
     >>> y = np.cos(x)
-    >>> one_dim_findpeaks(y, x, 0, 0)
-    array([[  1.68144859e-05,   9.99999943e-01,   3.57487961e+00],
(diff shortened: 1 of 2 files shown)


hyperspy/_signals/signal1d.py
```diff
@@ -98,9 +98,10 @@ def find_peaks_ohaver(y, x=None, slope_thresh=0., amp_thresh=None,
         contains position, height, and width of each peak
     Examples
     --------
-    >>> x = arange(0,50,0.01)
-    >>> y = cos(x)
-    >>> one_dim_findpeaks(y, x,0,0)
+    >>> import numpy as np
```
francisco-dlp: There is no need to import numpy, the fixture does it and the User Guide explains that all the examples assume that plt, hs and np are in the namespace.

hyperspy/_signals/signal1d.py
```diff
@@ -98,9 +98,10 @@ def find_peaks_ohaver(y, x=None, slope_thresh=0., amp_thresh=None,
...
+    >>> x = np.arange(0,50,0.01)
+    >>> y = np.cos(x)
+    >>> one_dim_findpeaks(y, x, 0, 0)
```
francisco-dlp: The function is now called ``find_peaks_ohaver``

francisco-dlp:
Appending ``#doctest: +SKIP`` to the line that calls the GUI code should help with the GUI commands.

magnunor:
Fixed!

PR #1431, opened 2017-02-08
Add permanent markers

magnunor:
Resubmission of https://github.com/hyperspy/hyperspy/pull/1401/commits, without the Matplotlib `Collections`.

- [x] Ability to save markers as part of HDF5 HyperSpy file (as metadata?)
- [x] Tests
- [x] Add more extensive tests for saving/loading markers
- [x] Update user guide
- [x] Figure out what to do with s.transpose
- [x] Add test for when loading HDF5 with wrong marker data


hyperspy/drawing/marker.py
```diff
@@ -77,6 +77,23 @@ def marker_properties(self, kwargs):
             except:
                 pass
 
+    def _to_dictionary(self):
+        self_dict = self.__dict__
+        marker_dict = {}
+        marker_dict['_marker_properties'] = self_dict['_marker_properties']
```
vidartf: Doesn't seem to me like you really need to use `self_dict` here at all?

hyperspy/signal.py
```diff
@@ -1865,7 +1865,8 @@ def __call__(self, axes_manager=None):
         return np.atleast_1d(
             self.data.__getitem__(axes_manager._getitem_tuple))
 
-    def plot(self, navigator="auto", axes_manager=None, **kwargs):
+    def plot(self, navigator="auto", axes_manager=None, 
+            plot_markers=False, **kwargs):
```
vidartf: `plot_markers` need docstring entry (in the right place... `KWARGS_DOCSTRING` ?).

hyperspy/signal.py
```diff
@@ -4005,38 +4011,68 @@ def is_rgb(self):
...
-        >>> m = hs.plot.markers.rectangle(x1=150, y1=100, x2=400,
+        >>> m = hs.markers.rectangle(x1=150, y1=100, x2=400,
         >>>                                  y2=400, color='red')
         >>> im.add_marker(m)
 
+        Add permanent marker
+        >>> s = hs.signals.Signal2D(np.random.random(100, 100))
+        >>> marker = hs.markers.point(50, 60)
+        >>> s.add_marker(marker, permanent=True, plot_marker=False)
+        >>> s.plot(plot_markers=True) #doctest: +SKIP
```
vidartf: Is it possible to call this simply as `s.add_marker(marker, permanent=True, plot_marker=True)` ? The docstring + examples seem to indicate that this is not possible.

hyperspy/signal.py
```diff
@@ -4005,38 +4011,68 @@ def is_rgb(self):
...
+                self.plot()
+            if marker._plot_on_signal:
+                self._plot.signal_plot.add_marker(marker)
+            else:
+                self._plot.navigator_plot.add_marker(marker)
+            marker.plot()
+        if permanent:
+            if not hasattr(self, "markers"):
+                self.markers = []
+            self.markers.append(marker)
```
vidartf: It seems to me that if called with `permanent=False, plot_marker=False`, this will simply do nothing (except set `_plot_on_signal`)?

hyperspy/signal.py
```diff
@@ -4005,38 +4011,68 @@ def is_rgb(self):
...
+        >>> s.add_marker(marker, permanent=True, plot_marker=False)
+        >>> s.plot(plot_markers=True) #doctest: +SKIP
         """
-        if self._plot is None:
-            self.plot()
-        if plot_on_signal:
-            self._plot.signal_plot.add_marker(marker)
-        else:
-            self._plot.navigator_plot.add_marker(marker)
+        marker._plot_on_signal = plot_on_signal
```
vidartf: Is it possible for one marker to be plotted on several signals (that would be nice)? If so, won't this variable be overwritten when trying to plot for the second time? Either way, what will happen if you try to plot a marker for the second time? It should at the very least fail somewhat gracefully if it cannot be plotted several places.

Secondly, what happens if you try to add a marker twice to the same signal? Both these cases could maybe be test cases.

vidartf (commented):
Some initial comments based purely on the code. Will try to break by using once it gets more mature.

hyperspy/signal.py
magnunor: I'm thinking `BASE_PLOT_DOCSTRING` might be the right place? Since this doesn't go to the `imshow`?

hyperspy/signal.py
magnunor: Changed one of the examples to `s.add_marker(marker, permanent=True)`

hyperspy/signal.py
magnunor: It would be nice to be able to plot them in different signals, but I don't think I want to implement that in this pull request. I added a check to see if the markers are already added to another signal, and will throw an `Exception` if they are. Which exception would be the best to throw here?

hyperspy/signal.py
magnunor: Hmmm, I'm not really sure how to best implement this. Since in the old `s.add_marker` one would always run `s.plot()` when adding a marker. However with permanent markers one would not necessarily want to open a plot when adding a marker, and with `plot_marker=False` it seems strange to open a plot without plotting the marker itself, since the function we're calling is named `add_marker`.

But I'm guessing my current implementation changes how `add_marker` functions, so I shouldn't really do that in the RELEASE_next_minor branch... I guess the right way is adding an `s.add_marker(plot_signal=True)` argument.

magnunor:
I think there should be some checks to see if markers have valid data dimensions when adding them. For example:
```Python
s = hs.signals.Signal2D(np.zeros((3,10,10)))
marker0 = hs.markers.point(5, 6)
marker1 = hs.markers.point((5, 2), (6, 8))
marker2 = hs.markers.point((5, 2, 1), (6, 8, 7))
s.plot()
s.add_marker(marker0)
s.add_marker(marker1)
s.add_marker(marker2)
```
All these can be added, but navigating to the last navigation index will give `IndexError: index 2 is out of bounds for axis 0 with size 2` due to `marker1` having too few data points. So I think adding some checks to `s.add_marker` is the easiest, like here: https://github.com/hyperspy/hyperspy/pull/1401/files#diff-64e04530fdfa59f53ebfa39bd6b4fd7cR4078

hyperspy/signal.py
vidartf: Probably `ValueError` is a good bet ("an argument that has the right type but an inappropriate value").

magnunor:
I added a check for making sure the marker data dimensions match with the signal navigation dimensions. But I think actually understanding what is correct can be a bit hard to understand, so I will add more info about this in the `marker` or `add_marker` docstring.

hyperspy/signal.py
magnunor: I changed it from `Exception` to `ValueError`, and updated the tests

hyperspy/signal.py
vidartf: I would argue that previous implementation is a little unintuitive from a HyperSpy perspective. It seems the previous implementation is very MPL-centric. As far as I understand it, `add_marker` previously:
 - associated the marker with a specific MPL axes.
 - `marker.plot()` then plotted the marker on the associated axes object.
 - if the signal was replotted, marker kindof broke down (?).

A more intuitive approach can be added here:
 - always have `add_marker` plot the marker onto the axes (no matter the args), but maybe not cause the figure to be drawn (determined by arg?). This last point would allow for adding several markers to the figure without flicker, but would still not give the speedup of collections.
 - have `add_marker` add the marker to the signal (like you have done) when permanent

magnunor: I'm slightly tempted to simply make another function: `s.add_permanent_marker`, to avoid having to change how `s.add_marker` behaves.

hyperspy/signal.py
vidartf: Explicit is better than implicit: I would have it as `s.add_marker(marker, permanent=True, plot_markers=True)` (if we keep that keyword).

hyperspy/drawing/marker.py
vidartf: I would probably just use `self._marker_properties`? Is there a specific reason to use `__dict__` at all?

hyperspy/signal.py
vidartf: I haven't really used markers a lot, so I don't want to say anything here, but it kinda goes against the "one obvious way to do it" zen to have two such similarly named functions.

magnunor: > always have add_marker plot the marker onto the axes (no matter the args), but maybe not cause the figure to be drawn (determined by arg?). This last point would allow for adding several markers to the figure without flicker, but would still not give the speedup of collections.

I added a `plot_signal` argument for this, since there are some cases where one doesn't want to create matplotlib axes, for example when adding permanent markers.

I also think parts of the need for so many arguments here is that the "temporary marker" is possibly a bit strange as a public facing functionality, since I don't see that many cases for a regular user to use a marker which disappears when a plot window is closed. Of course, when writing plotting functions it is really useful.

So I think the "public" facing `s.add_marker` functionality should add permanent markers, without directly calling any plotting functions. Unless the plot window is open, then the marker can be added using the same plotting function which also draws the permanent markers.

magnunor:
Currently, the markers can be saved in a `signal`, however it will not be saved in a HDF5 file.

Should that be a part of this pull request? Or keep them separate, to make the pull requests a more manageable?

If they can be in separate pull requests this should be fairly close to being finished, it would be nice if someone else could test the code to see if it behaves in a reasonable way.

francisco-dlp:
I think that saving to hdf5 should be part of this PR as, for that format, we guarantee that you get everything back so, if they're stored in the ``metadata`` they must find their way to the hdf5 file.

magnunor:
Sounds good, I've got a `_to_dictionary` function in the `MarkerBase` class, and I'm guessing I need something else to actually make the `metadata.as_dictionary` function convert the marker object to a dict? Trying to save a signal with a marker in the metadata fails with `TypeError: Object dtype dtype('O') has no native HDF5 equivalent`, and using `s.metadata.as_dictionary` gives the marker as a marker object: `'Markers': [<hyperspy.drawing._markers.point.Point at 0x7ff8484861d0>],`

francisco-dlp:
``_to_dictionary`` should do, what's its output?

magnunor:
Using the test script
```Python
import hyperspy.api as hs
import numpy as np

marker = hs.markers.point(10, 10)
s = hs.signals.Signal2D(np.random.random(size=(15, 15)))

s.metadata.Markers = [marker]
s.metadata.as_dictionary()
s.save("test.hdf5")
```

The `as_dictionary` gives:
```Python
{'General': {'title': ''},
 'Markers': [<hyperspy.drawing._markers.point.Point at 0x7f6c98c1ab70>],
 'Signal': {'binned': False, 'signal_type': ''},
 '_HyperSpy': {'Folding': {'original_axes_manager': None,
   'original_shape': None,
   'signal_unfolded': False,
   'unfolded': False}}}
```
The saving:
```Python
ERROR:hyperspy.io_plugins.hdf5:The hdf5 writer could not write the following information in the file: 0 : <hyperspy.drawing._markers.point.Point object at 0x7f6c98c1ab70>
  File "/home/magnunor/Documents/HyperSpy/hyperspy/hyperspy/io_plugins/hdf5.py", line 439, in dict2hdfgroup
    group.attrs[key] = value
  File "h5py/_objects.pyx", line 54, in h5py._objects.with_phil.wrapper (/tmp/pip-at6d2npe-build/h5py/_objects.c:2684)
  File "h5py/_objects.pyx", line 55, in h5py._objects.with_phil.wrapper (/tmp/pip-at6d2npe-build/h5py/_objects.c:2642)
  File "/home/magnunor/.local/lib/python3.5/site-packages/h5py/_hl/attrs.py", line 93, in __setitem__
    self.create(name, data=value, dtype=base.guess_dtype(value))
  File "/home/magnunor/.local/lib/python3.5/site-packages/h5py/_hl/attrs.py", line 169, in create
    htype = h5t.py_create(original_dtype, logical=True)
  File "h5py/h5t.pyx", line 1450, in h5py.h5t.py_create (/tmp/pip-at6d2npe-build/h5py/h5t.c:16078)
  File "h5py/h5t.pyx", line 1470, in h5py.h5t.py_create (/tmp/pip-at6d2npe-build/h5py/h5t.c:15912)
  File "h5py/h5t.pyx", line 1525, in h5py.h5t.py_create (/tmp/pip-at6d2npe-build/h5py/h5t.c:15813)
TypeError: Object dtype dtype('O') has no native HDF5 equivalent
```

francisco-dlp:
The issue is that ``_to_dictionary`` must return a dictionary containing only things that hdf5 can write such as numpy arrays, strings, numbers...

magnunor:
Hmmm, trying to figure out where it goes wrong... Where should the `_to_dictionary` be called? When running `s.metadata.as_dictionary()`, or later?

magnunor:
I think the problem is in the `as_dictionary` function (https://github.com/hyperspy/hyperspy/blob/RELEASE_next_minor/hyperspy/misc/utils.py#L375) in `DictionaryTreeBrowser` (https://github.com/hyperspy/hyperspy/blob/RELEASE_next_minor/hyperspy/misc/utils.py#L167), since function checks the `_dtb_value_` for all the objects in the metadata, and for the marker this is ` [<hyperspy.drawing._markers.point.Point at 0x7fd5cb411710>]`.

francisco-dlp:
Yes, you're right, DTB is less automatic than I thought. Checking for a marker instance around [here](https://github.com/hyperspy/hyperspy/blob/RELEASE_next_minor/hyperspy/misc/utils.py#L388) and calling the relevant method in the marker class should do.

magnunor:
Would a general solution be to check if the object has a `_to_dictionary` function, and if it does, get a dictionary that way?

Maybe something like:
```Python
elif isinstance(hasattr(item_['_dtb_value_'], '_to_dictionary'):
    item = item_['_dtb_value_']._to_dictionary()
```
But I'm guessing that won't actually help in this case, since I'm saving the markers in a list

francisco-dlp:
Then maybe it has to be implemented [here](https://github.com/hyperspy/hyperspy/blob/RELEASE_next_minor/hyperspy/io_plugins/hdf5.py#L367).

However, we probably need to think carefully on how to do it. Wouldn't storing them in a dict instead of a list would make it more user friendly? In this way printing the ``metadata`` may look nicer and deleting individual markers may also be simpler e.g. `del metadata.Markers.marker12`. @to266, what do you think?

magnunor:
I agree, and I'm already implementing a dictionary based approach instead.

This also means we can name them more easily

magnunor:
I added the basics for saving in dictionaries, but still not completely working. Will finish it later.

magnunor:
The markers can now be saved in HDF5 files in the metadata, but the loading is not working at the moment.

Is the right way to parse the markers dicts into HyperSpy marker classes to have some parsing, and generate the marker objects somewhere in the HDF5 IO plugin? (https://github.com/hyperspy/hyperspy/blob/RELEASE_next_minor/hyperspy/io_plugins/hdf5.py#L246)

magnunor:
Test are passing now! And the markers work nicely when saved in the metadata.

Just need to make them loadable from a HDF5-file.

francisco-dlp:
Sounds great!

magnunor:
Loading markers with HyperSpy HDF5 files now work!

I should add some extra tests, to make sure all the properties are saved and loaded properly.

magnunor:
I'll fix the missing things tomorrow, which mostly consist of more thorough tests, improved docstring and writing about it in the user guide.

magnunor:
This mostly seems to work fine now, except doing `s.deepcopy()`. Which I guess needs some way to handle marker classes in the metadata.

magnunor:
`s.deepcopy` is now working as well. So slicing signals with markers will now work fine.

magnunor:
Ready for review now.

hyperspy/drawing/marker.py
```diff
@@ -77,6 +89,35 @@ def marker_properties(self, kwargs):
...
+
+    def _get_data_shape(self):
+        if self.data['x1'].flatten()[0].flatten()[0] is not None:
+            data_shape = np.array(self.data['x1'].item()).shape
+        elif self.data['x2'].flatten()[0].flatten()[0] is not None:
+            data_shape = np.array(self.data['x2'].item()).shape
+        elif self.data['y1'].flatten()[0].flatten()[0] is not None:
+            data_shape = np.array(self.data['y1'].item()).shape
+        elif self.data['y2'].flatten()[0].flatten()[0] is not None:
+            data_shape = np.array(self.data['y2'].item()).shape
```
to266: I believe `next(self.data[key][()].flat)` is faster than `self.data[key].flatten()[0].flatten()[0]` and gives the same output (but please check that it is the case).
I would also iterate over the keywords, not do the lookup twice.

Overall this should be equivalent:
```python
data_shape = None
for key in ('x1', 'x2', 'y1', 'y2):
    ar = self.data[key][()]
    if next(ar.flat) is not None:
        data_shape = ar.shape
        break
if data_shape is None:
    raise ValueError("None of the coordinates have value")
else:
    return data_shape
```

hyperspy/drawing/marker.py
```diff
@@ -77,6 +89,35 @@ def marker_properties(self, kwargs):
...
+        marker_dict['plot_on_signal'] = self._plot_on_signal
+
+        data_dict = {}
+        data_dict['x1'] = self.data['x1'].item().tolist()
+        data_dict['x2'] = self.data['x2'].item().tolist()
+        data_dict['y1'] = self.data['y1'].item().tolist()
+        data_dict['y2'] = self.data['y2'].item().tolist()
+        data_dict['text'] = self.data['text'].item().tolist()
+        data_dict['size'] = self.data['size'].item().tolist()
+        marker_dict['data'] = data_dict
```
to266: It's better to create a populated dictionary straight away, not create an empty one and then put k:v pairs one by one:
```python
marker_dict = {
    'marker_properties': self.marker_properties,
    'marker_type': self.__class__.__name__,
    'plot_on_signal': self._plot_on_signal,
    'data': { k: self.data[k][()].tolist() for k in ('x1', 'x2', 'y1', 'y2', 'text', 'size')}
}
```

hyperspy/io.py
```diff
@@ -425,6 +425,22 @@ def dict2signal(signal_dict, lazy=False):
                     value = function(value)
                 if value is not None:
                     signal.metadata.set_item(mpattr, value)
+    if "metadata" in signal_dict:
+        if "Markers" in mp:
+            markers_dict = {}
+            for marker_name in mp['Markers'].keys():
+                try:
+                    marker = dict2marker(
+                            mp['Markers'][marker_name], marker_name)
```
to266: why not like this?
```python
for marker_name, m_dict in mp['Markers']:
    try:
        marker = dict2marker(m_dict, marker_name)
# ... etc
```

hyperspy/io.py
```diff
@@ -425,6 +425,22 @@ def dict2signal(signal_dict, lazy=False):
                     value = function(value)
                 if value is not None:
                     signal.metadata.set_item(mpattr, value)
+    if "metadata" in signal_dict:
+        if "Markers" in mp:
```
to266: Could also combine the two `if`s:
```python
if 'metadata' in signal_dict and 'Markers' in mp:
    # ...
```

hyperspy/signal.py
```diff
@@ -3509,6 +3516,17 @@ def __deepcopy__(self, memo):
...
+        if dc.metadata.has_item('Markers'):
+            temp_marker_dict = dc.metadata.Markers.as_dictionary()
+            markers_dict = {}
+            for marker_name in temp_marker_dict.keys():
+                marker = dict2marker(
+                        temp_marker_dict[marker_name], marker_name)
+                if marker is not False:
+                    marker.axes_manager = dc.axes_manager
+                    markers_dict[marker_name] = marker
+            dc.metadata.Markers = markers_dict
```
to266: It seems this piece of code is repeated (also present in io.py). Why not move it to a separate function in utils or someplace appropriate?

hyperspy/signal.py
```diff
@@ -4039,39 +4057,132 @@ def is_rgb(self):
...
+                    self.plot()
+                self._plot.navigator_plot.add_marker(marker)
             marker.plot()
+        if permanent:
+            if not self.metadata.has_item('Markers'):
+                self.metadata.add_node('Markers')
+            for marker_tuple in list(self.metadata.Markers):
+                if marker is marker_tuple[1]:
+                    raise ValueError("Marker already added to signal")
+            name_list = list(self.metadata.Markers.as_dictionary().keys())
```
to266: no need for `as_dictionary()` (I think)

hyperspy/signal.py
```diff
@@ -4039,39 +4057,132 @@ def is_rgb(self):
...
+            name = marker.name
+            temp_name = name
+            for i in range(1, 100000):
```
to266: why not `while`?

hyperspy/signal.py
```diff
@@ -4039,39 +4057,132 @@ def is_rgb(self):
...
+        if (not (len(marker_data_shape) == 0)) and (
+                marker_data_shape != self.axes_manager.navigation_shape):
+            raise ValueError(
+                    "Navigation shape of the marker must be 0 or the "
+                    "same navigation shape as this signal.")
+        if (marker.signal is not None) and (marker.signal is not self):
+            raise ValueError("Markers can not be added to several signals")
+        marker._plot_on_signal = plot_on_signal
+        marker._plot_marker = plot_marker
+        if plot_signal:
```
to266: this `plot_signal` kwarg seems almost a duplicate if `plot_marker` - namely if `plot_marker=True`, then `plot_signal` does not matter, as `self.plot()` will be called. As `plot_signal` only matters when the marker is __not__ plotted, then why have the option to begin with? It seems to have very little to do with the markers.

hyperspy/signal.py
```diff
@@ -4039,39 +4057,132 @@ def is_rgb(self):
...
+                    break
+            marker.name = temp_name
+            self.metadata.Markers[name] = marker
+            marker.signal = self
+
+    def _plot_permanent_markers(self):
+        marker_dict_list = list(self.metadata.Markers.__dict__.values())
+        for marker_dict in marker_dict_list:
+            if marker_dict['_dtb_value_'] is not False:
+                marker = marker_dict['_dtb_value_']
```
to266: the key lookup could be done once - either in the loop, or when creating the the list to begin with:
```python
marker_dict_list = [v['_dtb_value_'] for v in self.metadata.Markers.__dict__.values()]
# OR ...
for marker_dict in marker_dict_list:
    marker = marker_dict['_dtb_value_']
    if marker is not False:
        # do things...
```

to266 (changes requested):
A couple of comments here and there. 

Also, I would test the markers `_to_dictionary()` method in a separate test (maybe I missed it?), and then use it in the hdf5 testing to make sure that the loaded markers are identical (since checking all the appropriate fields by hand is error-prone)

hyperspy/signal.py
magnunor: I didn't want to break the previous default behaviour, which was always to call `s.plot()` if `s._plot()` did not exist. https://github.com/hyperspy/hyperspy/blob/RELEASE_next_minor/hyperspy/signal.py#L4067

However, I think it is useful to suppress the calling of `s.plot()` for certain cases, for example when running on headless system or returning a signal with permanent marker (like in Atomap).

hyperspy/drawing/marker.py
magnunor: Fixed

hyperspy/drawing/marker.py
magnunor: Fixed (it worked fine)

hyperspy/io.py
magnunor: Fixed

hyperspy/io.py
magnunor: Fixed

hyperspy/signal.py
magnunor: I made a new function in drawing/markers.py: https://github.com/hyperspy/hyperspy/pull/1431/files#diff-25b050d21071c0d2c1d0b5524030835aR221 which both s.__deepcopy__ and io.dict2signal uses

hyperspy/signal.py
magnunor: Fixed

hyperspy/signal.py
magnunor: Fixed

hyperspy/signal.py
magnunor: Fixed (used the second suggestion)

magnunor:
>Also, I would test the markers _to_dictionary() method in a separate test (maybe I missed it?), and then use it in the hdf5 testing to make sure that the loaded markers are identical (since checking all the appropriate fields by hand is error-prone)

I agree this would be nice, but one problem is the dictionary values which are None, are not saved. So the comparison fails.

hyperspy/signal.py
to266: I agree that not always plotting is good, so I agree with the idea. Is then the new kwarg only added to preserve the default behaviour (which is to always plot the signal as the markers are added, even if the markers themselves are not plotted)?

If so, do we have to @francisco-dlp ? If we do, I'd add a comment / `#TODO` to remove such complications for next major release...

to266:
Well you can then sanitize the `None`, to return the required thing. Adapt as appropriate:
```python
In [2]: import hyperspy.api as hs

In [3]: m = hs.markers.point(3,2)

In [4]: d = m._to_dictionary()

In [5]: def sanitize(thing):
   ...:     new = {}
   ...:     for k, v in thing.items():
   ...:         if isinstance(v, dict):
   ...:             new[k] = sanitize(v)
   ...:         elif v is not None:
   ...:             new[k] = v
   ...:     return new
   ...: 

In [6]: d
Out[6]: 
{'data': {'size': 20, 'text': None, 'x1': 3, 'x2': None, 'y1': 2, 'y2': None},
 'marker_properties': {'color': 'black', 'linewidth': None},
 'marker_type': 'Point',
 'plot_on_signal': True}

In [7]: sanitize(d)
Out[7]: 
{'data': {'size': 20, 'x1': 3, 'y1': 2},
 'marker_properties': {'color': 'black'},
 'marker_type': 'Point',
 'plot_on_signal': True}
```

hyperspy/signal.py
francisco-dlp: If the previous behaviour was calling ``s.plot()`` even if ``plot_marker`` was ``False``, then I would be of the opinion that that's a bug. I don't think that we have to keep bugs in order not to break the previous behaviour ;) So, I agree with @to266: ``plot_signal`` does not add useful functionality (unless I am missing something).

magnunor: I removed the `plot_signal` argument, and changed the behaviour of `plot_marker` slightly: `plot_marker=False` will no longer open the figure. 
- `s.add_marker(marker, plot_marker=False)` will now do nothing. 
- `s.add_marker(marker, plot_marker=False, permanent=True)`, will not open a plot with the marker, but only add the marker to `s.metadata.Markers`.  
- `s.add_marker(marker, permanent=True)` will open a plot with the marker if `s._plot` does not already exist, and add the marker to `s.metadata.Markers`.

francisco-dlp: The option that does nothing can be confusing for users. Maybe it should raise a ``ValueError`` or at least print a warning in the log?

magnunor: Yeah, in my opinion the "temporary" markers shouldn't really be "public" facing. Meaning `add_marker` would add the markers to `s.metadata.Markers` where they can be removed or temporarily disabled from there.

I'll make a warning if both `permanent=False, plot_marker=False`.

francisco-dlp: No that you mention it, it'll be good to show in the User Guide how to delete markers from metadata.

magnunor: I added an example on how to delete a permanent marker from a signal

magnunor:
I added a test for `dict2marker` and a test where an HDF5 is saved with markers, then loaded. Then the markers are compared.

I think that should be all. 

Only thing which would be nice would be to temporarily disable plotting while adding the permanent markers, since I suspect this can increase the speed when plotting many markers at the same time. But I'm not sure how to do this.

hyperspy/drawing/_markers/point.py
```diff
@@ -87,7 +87,7 @@ def update(self):
                                  self.get_data_position('y1')])
         self.marker._sizes = [self.get_data_position('size')]
 
-    def plot(self):
+    def plot(self, fig_draw_animated=True):
```
francisco-dlp: What about simply ``draw``. Also, as this is part of the public API it should have a docstring.

magnunor: To be honest, I not 100% sure what `ax.hspy_fig._draw_animated()` does. I'm guessing it tries to draw every animated object which hasn't already been drawn?

hyperspy/tests/io/test_hdf5.py
```diff
@@ -365,6 +368,315 @@ def teardown_method(self, method):
...
+
+        assert len(list(s1.metadata.Markers)) == 8
+        assert sanitize_dict(m0._to_dictionary()) == m0_new_dict
+        assert sanitize_dict(m1._to_dictionary()) == m1_new_dict
+        assert sanitize_dict(m2._to_dictionary()) == m2_new_dict
+        assert sanitize_dict(m3._to_dictionary()) == m3_new_dict
+        assert sanitize_dict(m4._to_dictionary()) == m4_new_dict
+        assert sanitize_dict(m5._to_dictionary()) == m5_new_dict
+        assert sanitize_dict(m6._to_dictionary()) == m6_new_dict
+        assert sanitize_dict(m7._to_dictionary()) == m7_new_dict
```
to266: I can see that you like copy-paste'ing ^^ 

I'd format this as a loop / iterator of some kind, but I guess it does not matter in the end

magnunor: Yeah, not really sure what the best practice is for these kinds of things...

hyperspy/tests/io/test_hdf5.py
```diff
@@ -365,6 +368,315 @@ def teardown_method(self, method):
...
+        s.add_marker(m, permanent=True)
+        with tempfile.TemporaryDirectory() as tmp:
+            filename = tmp + '/test_save_horizontal_line_marker.hdf5'
+        s.save(filename)
+        s1 = load(filename)
+        marker1 = s1.metadata.Markers.get_item(name)
+        assert marker1.name == name
+        assert marker1.get_data_position('y1') == y
+        assert marker1.marker_properties['color'] == color
+        assert marker1.marker_properties['linewidth'] == linewidth
```
to266: Also in all these tests multiple `assert` statements could all just be replaced by the sanitized dictionary comparison (which was why I suggested it in the first place..) - and you'd not need to carry that all relevant attributes are listed manually.

hyperspy/tests/plot/test_plot_markers.py
```diff
@@ -105,6 +106,203 @@ def test_auto_update(self):
...
+        m0 = markers.point(5, 5)
+        m1 = markers.point((5, 10), (10, 15))
+        m2 = markers.point(np.zeros((3, )), np.zeros((3, )))
+        s.add_marker(m0)
+        with pytest.raises(ValueError):
+            s.add_marker(m1)
+        s.add_marker(m2)
+
+
+def sanitize_dict(dictionary):
```
to266: Function repeated in two test files. I think we had something like "test_utils.py", where this would fit well

hyperspy/drawing/_markers/point.py
francisco-dlp: It draws only the animated parts without redrawing the "background" to optimize speed.

magnunor: https://github.com/hyperspy/hyperspy/pull/1431/commits/99f296dae363f226ba844e6bf613822fceb0d74d really reduced the duration when closing a figure with many markers. With a figure with 500 markers, it went from 3 minutes to 1 second.

Similar improvement when opening a plot with 500 markers.

I'll rename the argument, and implement them for the other markers.

magnunor: I think some of the marker functions can be made private, since the average user will have no reason to call `update` and `plot`, and possible things like `get_data_position`, `add_data`. These could rather be added as properties like `marker.x`, `marker.y`, ...

 But I'm guessing making them private would change the API.

hyperspy/tests/io/test_hdf5.py
magnunor: Fixed. I replaced them with comparing the _to_dictionary outputs.

hyperspy/tests/plot/test_plot_markers.py
magnunor: Moved the function to misc.test_utils

hyperspy/tests/io/test_hdf5.py
magnunor: I replaced this with having them in list, and iterating over them.

hyperspy/drawing/_markers/point.py
magnunor: I added an `update_plot` argument to `marker.plot()` and `marker.close()`, if `update_plot=False` it does not call `hspy_fig.draw_animated()`.

to266 (approved):

to266:
LGTM

magnunor:
Everything should be good to go.

One thing which could be added is a way to add a large amount of markers at the same time. For example, adding 800 markers takes a very long time. But I guess that is a special case, which can be sorted later. A workaround is simply making the Markers dictionary manually and adding that to the metadata.

PR #1433, opened 2017-02-08
Remove tifffile

francisco-dlp:
tiffile ships with skimage that is a required dependency. Therefore, we don't need to ship it ourselves.


francisco-dlp:
@ericpre, do you know why the ``hyperspy.tests.io.test_tiff.test_write_read_intensity_axes_DM`` is not passing? (I've left two prints in there to help with debugging.)

ericpre:
The version of tifffile.py in the current skimage (0.12.3) doesn't support reading/writing some of the metadata, only the next release (0.13) should do, see #1043.

Shall we can wait a bit for the next scikit-image release to avoid dropping functionality? In this case, we will also need to have scikit-image >= 0.13.

francisco-dlp:
Other options would be:

* Leave the test xfail when skimage < 0.13
* Depend on tifffile directly. (It is now available in pypi and anaconda)

Do you like any of these options?

ericpre:
I like the second option!! And the more recent version of tifffile.py seems to offer better support for Tiff generated with EM-MENU software (Tietz camera).

Shall we depend on tifffile >= 0.10 until skimage releases 0.13? Hopefully, tifffile will be updated in the 0.13 release, see scikit-image/scikit-image#2497.

magnunor:
How will depending on tifffile directly affect the Debian integration? I'm guessing not so much, since python3-skimage (scikit-image) is already in Debian, and I'm guessing the new version of scikit-image will also be in the Debian repositories.

ericpre:
[scikit-image 0.13](https://github.com/scikit-image/scikit-image/releases/tag/v0.13.0) has been released, now it should possible to remove tifffile if we choose to depend on scikit-image > 0.13.

I noticed it, because there are a couple of tiff tests that fails, possibly because of scikit-image 0.13 :/

thomasaarholt:
When do we use tiffile in hyperspy? For loading and saving tiffs? In terms of reviewing this, what features should I be testing?

francisco-dlp:
I think that it'll be better to wait until #1563 is merged before reviewing this one.

ericpre (commented):
It may be good to add a test for lazy loading and saving of data, maybe in 'test_saving_loading_stack_no_scale'? I could have done it in my PR, but I did not thought about that at this moment!
Except this slightly off topic comments, this should be good to merge.

Issue #1436, opened 2017-02-09
EMD files with bytes string will not load


magnunor:
Some EMD files which were saved using a Python2 file writer have their metadata like axis names encoded in bytes, and they fail when trying to load them the HyperSpy EMD-reader: `TraitError: The 'name' trait of a DataAxis instance must be a string, but a value of b'scanY' <class 'bytes'> was specified.`.

This can be fixed by adding a check in the file reader, to see if the metadata strings are `bytes` or `str`. And if they are `bytes`, change them to `str` when loading.

francisco-dlp:
By the way, the [website of the EMD format specification](http://emdatasets.lbl.gov/) seems to have banished...

magnunor:
Still alive, but moved: https://emdatasets.com/

francisco-dlp:
#1439

PR #1439, opened 2017-02-09
EMD-reader: check and correct if metadata strings are bytes

magnunor:
Fix for https://github.com/hyperspy/hyperspy/issues/1436

Sent to RELEASE_next_minor since it is related to the work in https://github.com/hyperspy/hyperspy/pull/1424 (which was merged into RELEASE_next_minor).

It only checks `axis.name` and `axis.units` since these are trait objects, and will fail if given `bytes` instead of `str`.

- [x] Add test


francisco-dlp:
It would be great if you could provide a file and unittest for this.

magnunor:
I added an EMD/HDF5 file where the name and units in the `dim1` (data about the axis), is saved in bytes. And made a corresponding test.

PR #1453, opened 2017-02-15
Linearity attribute to parameters

thomasaarholt:
This is a stepping stone towards a smarter fitting method, where the fitter can recognise whether or not all parameters are linear and so use linear fitting (much much faster). Now all parameters have a `is_linear` boolean attribute.

I am uncertain about the correct approach to the expression component
(currently set any parameter’s `is_linear` to False).

I am also unsure about which parameters are linear for the `heaviside`,
`voigt`, `RC`, `PESCORELineShape` (with shirley) and `EELS_vignetting`, though I've given them my best guess.


tjof2:
I think you're right with the `voight` model - it's based on a convolution isn't it?

thomasaarholt:
It would be great if someone could take a look at the components mentioned above - I am rather unfamiliar with them.

thomasaarholt:
Right, I've had a play with the components, and as far as I can tell, the linearity attributes are set correctly. The components reported in #1583 don't work at all, so I'm less concerned about them anyway.

In any case, where I have been uncertain I have set the parameter not free, so that a nonlinear fitter will be defaulted.

I think this is ready to merge.

ericpre:
It breaks a couple of tests. What's the reason?

thomasaarholt:
You're right. I hadn't been paying attention to it because something else had been failing PR tests previously.

It's the expression component. I'm trying to figure out how I would initialise all parameters to have attribute `component.parameter.is_linear = False`, for them to be set later.

thomasaarholt:
Not sure why this commit created a conflicting file. Resolving it now.

thomasaarholt:
In the future, fitting tools should assume the fitting is not linear if the `is_linear` attribute is not present, as it could be on a manually created component.

hyperspy/_components/expression.py
```diff
@@ -118,6 +118,11 @@ class docstring.
             self.__doc__ = _CLASS_DOC % (
                 name, sympy.latex(_parse_substitutions(expression)))
 
+        for parameter in self._parameter_strings:
```
to266: why such a convoluted way? why not just
```python
for par in self.parameters:
    par.is_linear = False
```
at this point the component is close to just normal ones and populated.

to266:
Generally I would add default for every new parameter that it is not linear, and then remove all unnecessary explicit `is_linear=False`.

hyperspy/_components/expression.py
thomasaarholt: Thanks... I thought I had tried that initially and it didn't work. I must have confused myself somehow. Will fix tomorrow.

thomasaarholt: I had it that way initially, but had the code under the wrong method.

thomasaarholt:
Good point @to266. Won't need the expression is_linear at all then.

francisco-dlp:
It should be possible to determine automatically if a parameter is linear using sympy.

thomasaarholt:
Just leaving this here for later as I am on the move:
http://stackoverflow.com/questions/36283548/check-if-an-equation-is-linear-for-a-specific-set-of-variables/36290758

thomasaarholt:
There we go! Sympy support for checking expression linearity. Added tests.

Ready for review.

hyperspy/_components/expression.py
```diff
@@ -215,3 +218,14 @@ def compile_function(self, module="numpy", position=False):
                         self,
                         Expression)
                     )
+
+    def is_linear(self, expr, vars):
```
to266: I would move this function out of the class.

Also I don't think it needs to be as general as it is in the stackexchange example:
```python
def is_linear(expression, name):
    try:
        return sympy.Eq(sympy.diff(expression, name, 2), 0)
    except TypeError:
        return False

**Comment by thomasaarholt** [comment]
Agreed on both accounts. Your code suggestion needed a modification to work properly.

File: hyperspy/_components/expression.py [hunk 3]
```diff
@@ -215,3 +218,13 @@ def compile_function(self, module="numpy", position=False):
...
+
+def check_expression_is_linear(expression, name):
+    if name not in expression:
+        raise AttributeError("The parameter '" + name + "' is not in the expression.")
+    try:
+        if not sympy.Eq(sympy.diff(expression, name, 2), 0):
+            return False
+    except TypeError:
+        return False
+    return True
```
to266: I still think this needs a rewrite in the form I suggested previously (while obviously keeping the first check that the parameter is indeed in the expression).

Currently this reads:
```python
try:
    if not True:
        return False
    else:
        return True
except:
    return False
```

Which is clearly simpler this way:
```python
try:
    return [True/False] # the actual bool statement here. Add "not" if needed, which I don't think you do
except:
    return False
```

Oh, and I would add a test that explicitly checks that the `ValueError` is raised when a "bad" symbol is given to the function.

thomasaarholt:
I've realised the code I was using for checking linearity was checking the wrong thing. I've updated the tests.

@to266, since the "check" code is now entirely different, would you mind looking over it again?

francisco-dlp:
The current implementation of ``Expression.is_linear`` may still not be what we need. I'll show it with two examples:

```python

>>> e = hs.model.components1D.Expression("a * b * x", name="linear")
>>> e.a.is_linear
True
>>> e.b.is_linear
True

```

However lls can only solve a model with this component if ``a`` or ``b`` is constant (i.e. ``free is False``).

```python

>>> e = hs.model.components1D.Expression("a  * x + b", name="linear")
>>> e.a.is_linear
False

```
Here the component is a linear equation of the sort that lls can solve, but ``is_linear`` returns ``False``. This is correct, ``a`` is indeed not linear in this sense, but this is not the sort of linearity that matters in our case.

Let's forget about linearity for a moment and let's see what we need to know for running lls or not. As lls solves systems of linear equations of the form ``k1 * x1 + k2 * x2 + ... kn * xn = b1``, when dealing with an individual component we must check:

1. Is the component (not the parameters) linear i.e. are the **free** parameters jointly linear? In  general, the answer to this question would depend on which parameters are free at the time of fitting. I think that the stack overflow code of the original implementation of ``Expression.is_linear`` is a good way to check for this, where the ``vars`` are the free parameters.
2. Does the component contains a constant? If yes, it should be added to the constant of the lls model (``b1`` above). For example, in ``a * x +b ``, if ``b`` is not free, then the component has a constant term. As we are solving ``Y = A X + B`` where ``Y`` is the data, the fitting routine should extract the constants of all components, sum them and subtract them from ``Y``. So, probably it'll be good to add a ``constant_term`` property to the component.

Regarding all the other components, I would make the ``Parameter._is_linear`` attribute private and add a ``Component.is_linear`` property that checks if all free components have ``_is_linear == True``, which should be correct for most components. Also add a ``constant_term`` property that simply returns 0 and doesn't allow setting the value. However, if any component linearity is  "non-trivial", the component should redefine the ``is_linear`` property in order to take into account "non-joint-linearities" and the ``constant_term`` property.

Does it make sense?

thomasaarholt:
Absolutely. Thanks for such a comprehensive post. This is definitely turning into a bigger PR than I anticipated and I'm very glad to have people checking in on me!
Having a go at this this weekend.

thomasaarholt:
I've got a draft for the new approach. I think I've finished the expression component. Could someone please look through the code for it? It seems to work fine, but this is my first time working with `@property`, so I'd appreciate it if someone could give me some feedback.

For the remaining components I still need to check the `constant_term`, which for now is 0 for any new component.

francisco-dlp:
I just made a small test:


```python

>>> import hyperspy.api as hs
>>> e = hs.model.components1D.Expression("a * b * x + c", "test")
>>> e.is_linear
True
```

Shouldn't it return ``False`` when all parameters are free (default)?

thomasaarholt:
Right. One major change now is the introduction of the term "independent", to refer to parameters that are independent of the signal axes (that is, `x` in expressions, the energy axis in EDS/EELS, etc).

**`component.independent_parameters`** lists these parameters in a similar fashion to `component.parameters` and `component.free_parameters`:

```python
>>> comp = Expression("a * b * x + c*d", "test")
>>> comp.independent_parameters
[<Parameter c of test component>, <Parameter d of test component>]
```

**`component.independent_term`** returns the calculated value of the independent term in the component, substituing the independent part of the expression for the `parameter.value`s. For the expression `x + b`, it would return b.value. 

```python 
>>> comp.c.value = 20
>>> comp.d.value = 2
>>> comp.independent_term
40.00
```

It returns 0 for most components (as they do not have an independent term). In the linear_fitting PR, I am intending to include some form of `model.independent_term` to sum the value of the independent_term for all components in the given navigation indices.

thomasaarholt:
@francisco-dlp The cases you suggested are covered by tests - but do feel free to suggest more or poke other holes in the code!!!

hyperspy/_components/pes_core_line_shape.py
```diff
@@ -54,6 +54,9 @@ def __init__(self, A=1., FWHM=1., origin=0.):
         self.origin.grad = self.grad_origin
         self.ab.grad = self.grad_ab
 
+        # Linearity
+        self.A._is_linear = False # This is probably linear, but unsure
```
francisco-dlp: Yes, I think that this one is linear

hyperspy/component.py
```diff
@@ -1324,3 +1326,43 @@ def on_active_change(change):
                              ' a notebook')
             else:
                 raise
+    @property
+    def is_linear(self):
+        "Loops through the component's free parameters, checks that they are linear"
+        linear = True
+        if len(self.free_parameters) > 1:
```
francisco-dlp: This method can be simplified:

```python
if len(self.free_parameters) == 1:
    return self.free_parameters[0]._is_linear
else:
    return False
```

hyperspy/component.py
```diff
@@ -1324,3 +1326,43 @@ def on_active_change(change):
...
+
+    @property
+    def independent_term(self):
+        "Get value of the independent term of the component at current indices. Returns 0 for most components."
+        return self._independent_term()
+
+    def _independent_term(self):
+        return 0
+
+    def _check_only_one_linear_parameter(self):
```
francisco-dlp: Isn't this function unnecessary? The component can't be linear if the number of free parameters is > 1.

hyperspy/tests/component/test_independent.py
```diff
@@ -0,0 +1,60 @@
...
+        expression, 
+        name="test_constant",
+        a = 20.0,
+        b = 4.0,
+        c = 10.0,
+        d = 2.0,
+        e = 20.0)
+    assert g.independent_term == 10**5 + 20
+
+def test_independent_after_fit():
```
francisco-dlp: What's the point of this test?

francisco-dlp (changes requested):
Except for the small comments, it LGTM

hyperspy/component.py
thomasaarholt: This one was nice. Thanks!

hyperspy/tests/component/test_independent.py
thomasaarholt: You're right, and I actually wasn't using it anyway.

thomasaarholt:
Corrected your comments. Thanks @francisco-dlp .

hyperspy/component.py
thomasaarholt: You're right, and I actually wasn't using it anyway.

thomasaarholt:
Oops, accidentally deleted your test comment.  Hang on.

thomasaarholt:
Checking my email to see what you wrote about it:
> In hyperspy/tests/component/test_independent.py:
>         sigma = 1.0)
>     assert g.independent_term == 0
> 
> def test_independent_from_expression3():
>     expression = "a * b * x + 10**(c/d) + e"
>     g = Expression(
>         expression, 
>         name="test_constant",
>         a = 20.0,
>         b = 4.0,
>         c = 10.0,
>         d = 2.0,
>         e = 20.0)
> assert g.independent_term == 10**5 + 20
> 
> def test_independent_after_fit():
> What's the point of this test?
> 
Which test were you referring to?

test_independent_from_expression2 checks that the independent term indeed is zero for an expression where it should be zero. test_independent_from_expression3 ensures that one with multiple variables not depending on x are added correctly.

Right, looking at the last one, "test_independent_after_fit", I think I see what you mean. It's already handled by the previous test. Deleting, then we're good.

francisco-dlp:
As there is no time to get #1321 in for v1.3, I suggest to leave this one for v1.4 too as its raison d'être is  #1321

thomasaarholt:
Fully agreed.

thomasaarholt:
Closing this as it is merged into #1462.

PR #1466, opened 2017-02-20
Fix invalid notebook

vidartf:
One of the notebooks in the repo were no longer valid. I'm guessing it was the victim of a global search and replace.


francisco-dlp:
Thanks!

PR #1461, opened 2017-02-18
Finish testing plot

ericpre:
A few things to finish from #1253:

- [x] Finish migrating plotting test to pytest,
- [x] Update dev-guide regarding plot testing using pytest and pytest-mpl,
- [x] Remove notebook for testing plot and tidy up the plotting tests,
- [x] Support only plot testing with matplotlib 2.x (simplify the tests and matplotlib 2.0 is now the last stable release).
- [ ] Changes skipping tests to xfail, see @vidartf [comment](https://github.com/hyperspy/hyperspy/pull/1467#discussion_r102189853) in #1467.
- [ ] Find a way to get the plotting tests working on travis os x? If not, find a way not to run the plotting tests on travis os x, but still be able to run them on a mac (would be better than nothing!)?

Pytest is significantly nicer to use than nose!!! The plotting tests are also tidier.

PR #1467, opened 2017-02-20
Make s.add_marker accept a list of markers

magnunor:
This also includes several optimizations which greatly increases the speed of adding many markers at the same time: mostly avoiding calling `s.metadata`.

- [x] Add example in docstring
- [x] Add example in user guide

### DIFF
--- a/doc/user_guide/visualisation.rst
+++ b/doc/user_guide/visualisation.rst
@@ -955,8 +955,8 @@ This can be extended to 4 (or more) navigation dimensions:
     >>> marker = hs.markers.point(x=x, y=y, color='red')
     >>> s.add_marker(marker, permanent=True)
 
-Many markers can added by using a for loop. Note, adding many markers
-might lead to very slow plotting.
+Many markers can added by using a for loop. Note, adding many (50+) markers
+this way might lead to very slow plotting.
 
 .. code-block:: python
 
@@ -965,8 +965,8 @@ might lead to very slow plotting.
     >>>     marker = hs.markers.point(i, i, size=60)
     >>>     s.add_marker(marker, permanent=True)
 
-If you want to add a large amount of markers at the same time a list can be
-used:
+If you want to add a large amount of markers at the same time we advise
+to add them as an iterable (list, tuple, ...), which will be much faster:
 
 .. code-block:: python
 
@@ -974,7 +974,7 @@ used:
     >>> s = hs.signals.Signal2D(np.arange(300).reshape(3, 10, 10))
     >>> marker_list = []
     >>> for i in range(500):
-    >>>     marker = hs.markers.point(random()*10, random()*10, size=60)
+    >>>     marker = hs.markers.point(random()*10, random()*10, size=30)
     >>>     marker_list.append(marker)
     >>> s.add_marker(marker_list, permanent=True)
 
--- a/doc/user_guide/visualisation.rst
+++ b/doc/user_guide/visualisation.rst
@@ -955,16 +955,6 @@ This can be extended to 4 (or more) navigation dimensions:
     >>> marker = hs.markers.point(x=x, y=y, color='red')
     >>> s.add_marker(marker, permanent=True)
 
-Many markers can added by using a for loop. Note, adding many (50+) markers
-this way might lead to very slow plotting.
-
-.. code-block:: python
-
-    >>> s = hs.signals.Signal2D(np.arange(300).reshape(3, 10, 10))
-    >>> for i in range(10):
-    >>>     marker = hs.markers.point(i, i, size=60)
-    >>>     s.add_marker(marker, permanent=True)
-
 If you want to add a large amount of markers at the same time we advise
 to add them as an iterable (list, tuple, ...), which will be much faster:
 
@@ -972,11 +962,22 @@ to add them as an iterable (list, tuple, ...), which will be much faster:
 
     >>> from numpy.random import random
     >>> s = hs.signals.Signal2D(np.arange(300).reshape(3, 10, 10))
-    >>> marker_list = []
-    >>> for i in range(500):
-    >>>     marker = hs.markers.point(random()*10, random()*10, size=30)
-    >>>     marker_list.append(marker)
-    >>> s.add_marker(marker_list, permanent=True)
+    >>> markers = (hs.markers.point(random()*10, random()*10, size=30) for i in range(500))
+    >>> s.add_marker(markers, permanent=True)
+
+This can also be done using different types of markers
+
+.. code-block:: python
+
+    >>> from numpy.random import random
+    >>> s = hs.signals.Signal2D(np.arange(300).reshape(3, 10, 10))
+    >>> markers = []
+    >>> for i in range(200):
+    >>>     markers.append(hs.markers.horizontal_line(random()*10))
+    >>>     markers.append(hs.markers.vertical_line(random()*10))
+    >>>     markers.append(hs.markers.point(random()*10, random()*10))
+    >>>     markers.append(hs.markers.text(random()*10, random()*10, "sometext"))
+    >>> s.add_marker(markers, permanent=True)
 
 Permanent markers are stored in the HDF5 file if the signal is saved:
 
--- a/doc/user_guide/visualisation.rst
+++ b/doc/user_guide/visualisation.rst
@@ -965,6 +965,19 @@ might lead to very slow plotting.
     >>>     marker = hs.markers.point(i, i, size=60)
     >>>     s.add_marker(marker, permanent=True)
 
+If you want to add a large amount of markers at the same time a list can be
+used:
+
+.. code-block:: python
+
+    >>> from numpy.random import random
+    >>> s = hs.signals.Signal2D(np.arange(300).reshape(3, 10, 10))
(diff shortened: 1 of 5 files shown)


magnunor:
Adding 1000 point markers using `s.add_marker` takes about 1 second now

francisco-dlp:
Great!! Maybe it'll be good to point out the huge performance advantage of adding the markers in this way in the docs.

Beside, could it take a (more general) iterable instead of a list?

magnunor:
Any iterable should work fine. I'll change the docstring, and add some more about this in the user guide.

magnunor:
I changed the docstring and user guide a little bit

doc/user_guide/visualisation.rst
```diff
@@ -955,8 +955,8 @@ This can be extended to 4 (or more) navigation dimensions:
     >>> marker = hs.markers.point(x=x, y=y, color='red')
     >>> s.add_marker(marker, permanent=True)
 
-Many markers can added by using a for loop. Note, adding many markers
-might lead to very slow plotting.
+Many markers can added by using a for loop. Note, adding many (50+) markers
```
francisco-dlp: What about simply removing this from the User Guide: users in a hurry may miss the next paragraph which should be the preferred method in most cases.

doc/user_guide/visualisation.rst
```diff
@@ -965,16 +965,16 @@ might lead to very slow plotting.
...
-If you want to add a large amount of markers at the same time a list can be
-used:
+If you want to add a large amount of markers at the same time we advise
+to add them as an iterable (list, tuple, ...), which will be much faster:
 
 .. code-block:: python
 
     >>> from numpy.random import random
     >>> s = hs.signals.Signal2D(np.arange(300).reshape(3, 10, 10))
     >>> marker_list = []
```
francisco-dlp: What about using a generator instead?:

```python

markers = (hs.markers.point(random()*10, random()*10, size=30) for i in range(500))

```

hyperspy/signal.py
```diff
@@ -4111,45 +4114,71 @@ def add_marker(
...
-            if self._plot is None:
-                self.plot()
-            if marker._plot_on_signal:
-                self._plot.signal_plot.add_marker(marker)
-            else:
-                if self._plot.navigator_plot is None:
-                    self.plot()
-                self._plot.navigator_plot.add_marker(marker)
-            marker.plot()
+        if issubclass(marker.__class__, MarkerBase):
```
vidartf: Are markers iterable? If not, I would change this test to check for iterables instead (`from hyperspy.misc.utils import isiterable`). Then we can still ducktype markers without explicitly having to inherit `MarkerBase`.

doc/user_guide/visualisation.rst
magnunor: Hmmm, doesn't work. Since I use the last marker to update the plot. `TypeError: 'generator' object is not subscriptable`. I'll try to avoid this by calling the `_draw_animated` function at the end of the `add_marker` function instead.

hyperspy/signal.py
magnunor: Fixed! But I'm using the `issubclass(marker.__class__, MarkerBase)` when plotting the the permanent markers, since other stuff can (I guess) fairly easily sneak into the metadata.Markers.

doc/user_guide/visualisation.rst
magnunor: Fixed

doc/user_guide/visualisation.rst
magnunor: Fixed, and I added another example for adding different types of markers at the same time

magnunor:
I also added some more robustness with the updating of the plot. Now both the navigator and signal plot will be updated if they exist.

magnunor:
This should be ready to be merged if people are happy with it.

francisco-dlp:
@vidartf, are you happy with it?

hyperspy/signal.py
vidartf: I'm confused about where/when/why things can sneak into metadata, and not entirely sure who's responsibility it is to deal with it (ZoP: "Errors should never pass silently.").

vidartf:
What's the deal with CI on Mac?

francisco-dlp:
Is the question: why are some plotting tests skipped in Mac? I imagine that it has something to do with different fonts causing test failures. @ericpre should know. This was introduced in #1253 which I merged somehow prematurely thinking that the sooner it gets merged the less plotting bugs will be introduced inadvertently.

ericpre:
Travis does not support emulating on os x. If I remember correctly, even if we run pytest without the mpl plugin on travis os x, the tests will return an error. That's why, they are skipped on os x.
I think, this is mentioned in the dev_guide of the #1461.

francisco-dlp:
Why do the tests need screen emulation at all?

ericpre:
It looks to me, that to have the attribute `_background`, you need to create a figure, which is not possible on os x. I think this [function](https://github.com/hyperspy/hyperspy/blob/RELEASE_next_minor/hyperspy/drawing/utils.py#L103) will not work on travis os x.

hyperspy/signal.py
magnunor: I removed the `issubclass` `MarkerBase` check, removed the `MarkerBase` import and did some minor improvements to the `s._plot_permanent_markers` function.

francisco-dlp:
@ericpre, I have the impression that matplotlib itself does not use screen simulation for their tests, so I was just wondering if a different approach is possible. In any case, maybe we can continue the discussion in #1461?

ericpre:
yes!

magnunor:
This pull request should be good to merge, unless anyone has any additional feedback

francisco-dlp:
@vidartf, you're the reviewer.

vidartf:
I'm still confused about the CI: You say that it is skipped, but for me it says 27 Failures, and it is causing Travis to fail. The question for me is then: Are the skips marked wrong? Should any of the tests of this PR be skipped for Mac? If so, that needs to be corrected before this can be merged.

vidartf:
Also, I would mark them as expected failures, so that we can pick up on it in the future if this suddenly starts working.

magnunor:
All the MacOS tests fail with the same error:

```Python
self = <hyperspy.drawing.signal1d.Signal1DFigure object at 0x13bcbe780>
    def _draw_animated(self):
        if self.ax.figure:
            canvas = self.ax.figure.canvas
>           canvas.restore_region(self._background)
E           AttributeError: 'Signal1DFigure' object has no attribute '_background'

hyperspy/drawing/figure.py:28: AttributeError
```

So I'm guessing its related to that the MacOS environment doesn't create the canvas when its supposed to (i.e. when calling `s.plot()`)?

magnunor:
Oooh, I guess this is due to me not marking the tests with skip for darwin

magnunor:
I added skips for all the tests which involve calling s.add_marker(plot_marker=True), I guess this is less then ideal but currently all the other tests involving plotting skips darwin as well

ericpre:
Yes, this is to expect and to be more precise, there are two different things here:
- the plotting tests are skipped on os x and this is working fine. Maybe this can be improved later.
- the tests introduce in this PR seems to call the `create_figure` or a similar function, which as it not now (and as it was before) can not work on travis os x!
However, could it be possible to run these tests without calling this function?

hyperspy/tests/io/test_hdf5.py
```diff
@@ -371,6 +371,7 @@ def teardown_method(self, method):
 
 class Test_permanent_markers_io:
 
+    @pytest.mark.skipif("sys.platform == 'darwin'")
```
vidartf, line 374: I would prefer xfail instead (http://doc.pytest.org/en/latest/skipping.html#mark-a-test-function-as-expected-to-fail), which would allow mac users to run full test suite locally, and would also notify us if Travis setup changes..

magnunor:
The `s.add_marker` function relies on the create_figure function in drawing/signal1d.Signal1DFigure https://github.com/magnunor/hyperspy/blob/443ac12098d57ebc10966df14eab4fc5a34f418b/hyperspy/drawing/signal1d.py#L58 

> the tests introduce in this PR seems to call the create_figure or a similar function, which as it not now (and as it was before) can not work on os x!

Does this mean `create_figure` in `Signal1DFigure` doesn't work at all in macOS, or just the tests?

hyperspy/tests/io/test_hdf5.py
magnunor: I guess this should be done for all the current plotting tests? If so, I think that belongs in a separate pull request.

ericpre: yeah, I will it in #1461.

ericpre:
@magnunor: just the tests on travis osx.

magnunor:
Test are passing now that I skip the ones which require plotting in the macOS testing.

With regards to changing to xfail instead of skipif: since this is also something which should be done for all the currently "skipif" tests, this should be done in a different pull request? For example https://github.com/hyperspy/hyperspy/pull/1461 ?

vidartf:
Leaving it to another PR is ok, for me :)

PR #1471, opened 2017-02-22
Bugfix lazy memmap

to266:
A quick fix to change default blockfile `mmap_mode` depending on `lazy`.

Previously default was `c`, which requires as much memory as the data to be allocated, thus to load a file lazily minimal command had to be `hs.load(filename, lazy=True, mmap_mode='r')`.

Now default is `None` and gets assigned to `c` if `lazy=False` and `r` if `True`. Any user-given `mmap_mode` value is still preserved and behavior unchanged, but loading lazily is simpler:
`hs.load(filename, lazy=True)`.

### DIFF
--- a/hyperspy/io_plugins/blockfile.py
+++ b/hyperspy/io_plugins/blockfile.py
@@ -140,10 +140,12 @@ def get_header_from_signal(signal, endianess='<'):
     return header, note
 
 
-def file_reader(filename, endianess='<', mmap_mode='c',
+def file_reader(filename, endianess='<', mmap_mode=None,
                 lazy=False, **kwds):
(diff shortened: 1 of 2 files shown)


francisco-dlp:
This has introduced a randomly failing test (see Rnm min env travis failure)

francisco-dlp:
This is what was in ``dask.keys`` when it failied locally for me:

```python
dict_keys([('array-e9f1437a98f2ea0b6963e592f729ea7d', 0, 0, 0, 0), 'array-original-e9f1437a98f2ea0b6963e592f729ea7d'])
```

francisco-dlp:
Hopefully fixed in #1476

Issue #1464, opened 2017-02-20
Dead links


vidartf:
There are some dead links around:
```
C:\dev\hyperspy [RELEASE_next_minor ≡ ]> pip install pytest-check-links
C:\dev\hyperspy [RELEASE_next_minor ≡ ]> py.test --check-links -k .html
============================= test session starts =============================
platform win32 -- Python 3.5.2, pytest-3.0.6, py-1.4.32, pluggy-0.4.0
rootdir: C:\dev\hyperspy, inifile: pytest.ini
plugins: nbval-0.4.1, tornado-0.4.5, timeout-1.2.0, cov-2.4.0, check-links-0.0.2
collected 1643 items

doc\_static\articles_using_hyperspy.html .....F..F.FF...
hyperspy\tests\drawing\test_plot_image.ipynb .

================================== FAILURES ===================================
 C:\dev\hyperspy\doc\_static\articles_using_hyperspy.html: http://link.aip.org/link/?APL/93/112505/1
http://link.aip.org/link/?APL/93/112505/1: <urlopen error [Errno 11001] getaddrinfo failed>
 C:\dev\hyperspy\doc\_static\articles_using_hyperspy.html: 16/j.ultramic.2010.10.001
16/j.ultramic.2010.10.001: No such file: C:\dev\hyperspy\doc\_static\16\j.ultramic.2010.10.001
 C:\dev\hyperspy\doc\_static\articles_using_hyperspy.html: http://dx.doi.org/10.1088/0957-4484/21/14/145606
http://dx.doi.org/10.1088/0957-4484/21/14/145606: HTTP Error 403: Forbidden
 C:\dev\hyperspy\doc\_static\articles_using_hyperspy.html: http://iopscience.iop.org/0957-4484/21/14/145606/
http://iopscience.iop.org/0957-4484/21/14/145606/: HTTP Error 403: Forbidden
============================ 1627 tests deselected ============================
============ 4 failed, 12 passed, 1627 deselected in 29.03 seconds ============
```

Sphinx also has a linkcheck command:
```
cd docs
make linkcheck   (alternatively makefile.bat linkcheck on Windows)
```
gives a multitude of broken links and permanent redirects.

francisco-dlp:
Thanks!

vidartf:
Closed in #1475

PR #1475, opened 2017-02-22
Fix most broken links

francisco-dlp:
Most dead links fixed.

Some remain dead because they are in ``external`` or they point to resource that are no longer available but we should still credit them with the link. Beside, some may be available in webarchive, so keeping the link is useful.

This also removes the anaconda environment as it has been deprecated for our purposes by conda-forge.


francisco-dlp:
#1464

PR #1477, opened 2017-02-22
Parameter arguments for gaussian and polynomial

thomasaarholt:
**Edit: The premise for this PR was a bit flawed, as the advantage of leastsq fitting is being able to rely on the fit of the previous pixel. Therefore it may even be counter-productive to leastsq fit an entire dataset at once.**

**However, I think I'll leave this PR open as the functionality may come in handy at a later point.**

For "whole-dataset-leastsq" fitting, I've found it necessary to call the `component.function()` method a lot for the parameter values in many pixels (for every iteration of the fitter). In order to speed things up, I've rewritten the method to take an optional parameter `parameters`, which calls the component function on those new parameters. For the gaussian component:
```python
    def function(self, x, parameters = None):
        if parameters is not None:
            A, s, c = parameters
        else:
            A = self.A.value
            s = self.sigma.value
            c = self.centre.value
        return A * (1 / (s * sqrt2pi)) * np.exp(-(x - c)**2 / (2 * s**2))
```
(The nice thing is then being able to use `np.apply_along_axis(func=component.function, axis=0, arr=parameters_array)` after constructing an array of parameters, to build the contribution to the model from each component. This should be much faster than looping through the pixels and updating p0)

I haven't done this to more components than the background and gaussian, but I'm happy to quickly add this to the rest as well. Or if there are better ways of doing things, then I'm all ears :)

### DIFF
--- a/hyperspy/_components/gaussian.py
+++ b/hyperspy/_components/gaussian.py
@@ -116,10 +116,13 @@ class Gaussian(Component):
         self.sigma.grad = self.grad_sigma
         self.centre.grad = self.grad_centre
 
-    def function(self, x):
-        A = self.A.value
-        s = self.sigma.value
-        c = self.centre.value
+    def function(self, x, parameters = None):
+        if parameters is not None:
+            A, s, c = parameters
+        else:
+            A = self.A.value
+            s = self.sigma.value
+            c = self.centre.value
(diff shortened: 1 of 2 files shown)


ericpre:
Closing this since this is now implemented with the `function_nd` method of the `Expression` (and a few others) component, see #2032.

PR #1481, opened 2017-02-23
Fix #1382: fix labelling of plot_explaned_variance_ratio and add tests for mva plotting functions.

### LINKED ISSUE #1382: s.plot_explained_variance_ratio(small_int) axis label bug
The x-axis has duplicates of the integer values for small ints fed for n components to the method.
screenshot 2017-01-12 17 42 27

ericpre:
Fix #1382.

- [x] Fix bug with xticklabels in plot_explained_variance_ratio
- [x] Add tests (with plots ;))
- [x] Set the default of n to 30
- [x] Add tests for the plot_* functions (plot_decomposition_factors, plot_decomposition_factors, plot_bss_loadings, plot_bss_factors)
- [x] Add title to figures generated with plot_* 
- [x] Make legend animated and more compact in figures generated with plot_*
- [x] Merge with upstream after #1510 is merged and remove skipping tests on osx.
- [x] Update the baseline images: for some reason, there are very small changes in matplotlib 2.0.1 and they make the plotting tests fail. Moreover, the tests are now done with the v2 style (the default in pytest-mpl is still v1).


hyperspy/tests/mva/test_decomposition.py
```diff
@@ -229,3 +253,38 @@ def test_decomposition_supported_return_false(self):
                 algorithm=algorithm,
                 return_info=False,
                 output_dimension=1) is None
+
+
+@pytest.mark.skipif("sys.platform == 'darwin'")
+class TestPlotExplainedVarianceRatio:
+
+    def setup_method(self, method):
+        MY_PATH = os.path.dirname(__file__)
```
francisco-dlp: It would be better to generate a dataset for this (even if just with a seeded ``np.random``) because otherwise, if the FEI readers gets broken, it'll seem that this plotting function is broken too.

francisco-dlp:
Other than the comment above LGTM.

hyperspy/tests/mva/test_decomposition.py
ericpre: yes, it did this initially, but the plotting test failed because the data (and therefore the explained variance ratios) are not exactly the same!! And it needs a big enough dataset to have enough components.
Would it be better if I open the data from a numpy file? Or is there a way to always get the same explained variance ratios from a generated data?

francisco-dlp: If you use e.g. ``np.random.seed(1)`` before creating the random dataset then it should always be the same. For example, the following may do:

```python

import numpy as np
import hyperspy.api as hs
np.random.seed(1)
s = hs.signals.Signal1D(np.dot(np.random.random((100, 5)), np.random.exponential(size=(5,100))))

```

ericpre: Thanks, this is doing the job nicely!

ericpre:
The tests using random.seed() are failing, while there were passing previously using the FEI data set. Could it be that the random data are not "exactly" the same on different environment (numpy version, C compiler, python version, etc.) ?
This is quite surprising because this is used quite a lot in hyperspy testing and there is no problem with it at all...

francisco-dlp:
Sorry, that's my mistake. As random generators are called twice...we need to seed before each call:

```python
np.seed(1)
sources = np.random.exponential(size=(5,100))
np.seed(1)
mixmat = np.dot(np.random.random((100, 5))
self.s = signals.Signal1D(np.dot(mixmat, sources))

```

francisco-dlp:
Now, I don't know why it is failing, it passes all the tests in my machine.

By the way, wouldn't it be better (for consistency) to move the test to ``test/drawing/``?

francisco-dlp:
Sorry, no it doesn't pass the tests in my machine either.

ericpre:
Yeah, I can move it to `tests/drawing`.
Please let me know, if you think that there is some hope to find where the discrepancy come from. If not, I would suggest to use a saved numpy array for these tests.

francisco-dlp:
We use random in many tests and it works across python versions and OSs. Are you sure that you've generated the base images using the same random seed?

ericpre:
Yes, I am sure that the images have been generated with the current code of this PR and they run fine on my system. After you mention random.seed, I have noticed that it is use a lot and works fine...

francisco-dlp:
[travis testing with files created in my linux computer](https://travis-ci.org/francisco-dlp/hyperspy/builds/204666961)

francisco-dlp:
Travis seems happy, I'll send a PR.

ericpre:
Happy travis makes me happy! :)

On 23 Feb 2017 5:32 p.m., "Francisco de la Peña" <notifications@github.com>
wrote:

Travis seems happy, I'll send a PR.

—
You are receiving this because you authored the thread.
Reply to this email directly, view it on GitHub
<https://github.com/hyperspy/hyperspy/pull/1481#issuecomment-282062650>,
or mute
the thread

francisco-dlp:
It still failed for the min env. I think that the issue is that, as the data has no noise, the "residual" components may be too sensitive to machine differences. I've added gaussian noise hoping that this will fix it for good...

ericpre:
Yes, you are right, the different in the [plot was in the 10e-31](https://github.com/ericpre/hyperspy/pull/4/commits/e63953ce16ef06b4348adfa4a1059a8e63fbc572#diff-99498ead917c0f8bf5d9b915a7a8e0d6)... Now, it looks much better.

francisco-dlp:
I've forgotten merging this one :(

ericpre:
No worries, the last release seemed to have been tedious enough!!!
Once #1510 is merge, I can update this PR, the main thing is to remove skipping tests on travis osx.

ericpre:
The current issue with plotting tests failing (see for example #1306) is related to small matplotlib changes. This is confirmed by rebuilding a _working_ commit: the plotting tests were originally [passing](https://travis-ci.org/hyperspy/hyperspy/builds/220274603) and now this is not the case [anymore](https://travis-ci.org/ericpre/hyperspy/builds/228405570)...

Updating the baseline images should fixed it. I did it in this PR, because I was sure that there was a reliable commit. The baseline images had to be update to the matplotlib 2 style anyway, which is not the default in pytest-mpl.

ericpre:
Since this is fixing the plotting test which are currently failing (see last point of the first post of this PR), I would suggest to merge this ASAP!

doc/user_guide/mva.rst
```diff
@@ -337,7 +337,7 @@ results. All the other methods display each component in its own window. For 2
 and 3 it is wise to provide the number of factors or loadings you wish to
 visualise, since the default is to plot all of them. For BSS, the default is
 the number you included when running the :py:meth:`~.learn.mva.MVA.blind_source_separation`
-method.
+method. In case of one dimensional factors or loadings, the latters can be toggled on and off by clicking on the legended line.
```
francisco-dlp: typo: legended

Maybe it reads better like this: "One-dimensional factors/loadings can be toggled on and off by clicking on their corresponding line in the legend."?

hyperspy/drawing/signal.py
```diff
@@ -23,6 +23,8 @@
 from mpl_toolkits.axes_grid1 import make_axes_locatable
 from traits.api import Undefined
 
+from hyperspy.drawing.utils import set_axes_decor
+
 
 def _plot_quiver_scatter_overlay(image, axes_manager,
```
francisco-dlp: Actually, it'll be a good idea to remove this function rather than modifying it as it is not used anywhere.

hyperspy/signal.py
```diff
@@ -741,10 +758,11 @@ def plot_decomposition_factors(self,
                                    comp_ids,
                                    calibrate=True,
                                    same_window=None,
-                                   comp_label='Decomposition factor',
```
francisco-dlp, line 744: We cannot break the API in a minor release.

hyperspy/signal.py
```diff
@@ -792,19 +808,24 @@ def plot_decomposition_factors(self,
...
                                             comp_ids=comp_ids,
                                             calibrate=calibrate,
                                             same_window=same_window,
-                                            comp_label=comp_label,
+                                            comp_label=title,
                                             cmap=cmap,
                                             per_row=per_row)
 
     def plot_bss_factors(self, comp_ids=None, calibrate=True,
-                         same_window=None, comp_label='BSS factor',
```
francisco-dlp, line 805: As above, we can't break the API.

hyperspy/signal.py
```diff
@@ -853,23 +871,29 @@ def plot_bss_factors(self, comp_ids=None, calibrate=True,
...
                                             same_window=same_window,
-                                            comp_label=comp_label,
+                                            comp_label=title,
                                             per_row=per_row)
 
     def plot_decomposition_loadings(self,
                                     comp_ids,
                                     calibrate=True,
                                     same_window=None,
-                                    comp_label='Decomposition loading',
```
francisco-dlp, line 867: As above, we can't break the API.

hyperspy/signal.py
```diff
@@ -932,22 +959,29 @@ def plot_decomposition_loadings(self,
...
-            comp_label=comp_label,
+            comp_label=title,
             cmap=cmap,
             no_nans=no_nans,
-            per_row=per_row)
+            per_row=per_row,
+            axes_decor=axes_decor)
 
     def plot_bss_loadings(self, comp_ids=None, calibrate=True,
-                          same_window=None, comp_label='BSS loading',
```
francisco-dlp, line 947: As above, we can't break the API.

hyperspy/signal.py
```diff
@@ -1013,10 +1054,20 @@ def plot_bss_loadings(self, comp_ids=None, calibrate=True,
             with_factors=with_factors,
             factors=factors,
             same_window=same_window,
-            comp_label=comp_label,
```
francisco-dlp, line 1016: As above, we can't break the API.

francisco-dlp (changes requested):
Just some minor comments, so it's almost ready to merge.

hyperspy/drawing/signal.py
ericpre: Yes, it would be good to remove this function and possibly also `_plot_factors_or_pchars`, which is in the same case.
Is there a way to make this find-able easily from the history if somebody want to use it later? Maybe making an issue that link to it?

hyperspy/signal.py
ericpre: yes... I will leave both and put a `DeprecationWarning` for this version? Should I then make a PR to next_release_major to remove both the `comp_label` argument and the `DeprecationWarning`?

hyperspy/signal.py
```diff
@@ -244,145 +244,6 @@ def __getitem__(self, name):
 class MVATools(object):
     # TODO: All of the plotting methods here should move to drawing
```
ericpre: While we are at tidying things up, should I move all plotting functions to a `MVADrawing` class in `drawing.signal`?

PR #1485, opened 2017-02-25
hyperspy hdf5 extension hdf5 -> hspy

francisco-dlp:
Change hyperspy hdf5 file extension to hspy.

# Advantages:

* hyperspy cannot load any hdf5 file, only those recorded using its own standard. Adding the extension would make this clearer.
* HyperSpyUI could associate itself only with hspy files, allowing other programs to associate themselves with general hdf5 files.
* Anybody receiving a hspy file will have an easier time at finding the right software to open it.

# Disadvantages
* It is less straightforward to realize that the hspy file can be opened with any hdf5 reader.

If everybody agrees, maybe this could make it into the 1.2 release?

### DIFF
--- a/doc/user_guide/io.rst
+++ b/doc/user_guide/io.rst
@@ -211,16 +211,40 @@ of arbitrary dimensions. It is based on the `HDF5 open standard
 applications
 <http://www.hdfgroup.org/products/hdf5_tools/SWSummarybyName.htm>`_.
 
-Note that only HDF5 files written by HyperSpy are supported
+Only loading of HDF5 files following the HyperSpy specification are supported.
+
+.. versionadded:: 1.2
+    Saving hdf5 files with extension ``.hspy``.
+
+The default extension is ``.hdf5`` but, in order to make it explicit that the
+HDF5 follows the HyperSpy specification, it is possible to save it with the
+``.hspy`` extension as follows.
+
+
+.. code-block:: python
+
+    >>> s = hs.signals.BaseSignal([0])
+    >>> s.save('test.hspy')
+
+It is possible to use the ``.hspy`` extension by default by changing the
+value is ``preferences`` using the GUI or programatically:
+
+.. code-block:: python
+
+    >>> hs.preferences.General.hspy_extension = True
+    >>> hs.save() # make the changes permanent
+
+From HyperSpy 1.3 ``.hspy`` will be the default extension.
 
 .. versionadded:: 0.8
+    Saving list, tuples and signals present in `:py:attr:`~.metadata``.
 
-It is also possible to save more complex structures (i.e. lists, tuples and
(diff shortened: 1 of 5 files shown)


magnunor:
Sounds good! 

Should we make a "HDF5 identifier" function in the IO system? So that if HyperSpy is given a file with the .hdf5 extension, it will try to load it with the appropriate file reader (.hspy or .emd reader)?

francisco-dlp:
Sounds good. Maybe it would also be handy to implement a more general hdf5 support. For example, if it is not a known hdf5 format it could also load all the datasets in the file into signal instances.

thomasaarholt:
@francisco-dlp That would be nice. Would make it much easier for people using hdf5 for non-EM applications.

francisco-dlp:
This is ready for review. Could somebody take a look so that we can finally release today?

hyperspy/tests/io/test_hdf5.py
```diff
@@ -185,10 +186,47 @@ def test_binary_string(self):
...
+        return ".hspy"
+    else:
+        return ".hdf5"
+
+
+def test_hspy_extension(tmpfilepath):
+    # test_marker_point_y2_data_deleted.hdf5 has 5 markers,
+    # where one of them is missing the y2 value, however the
+    # the point marker only needs the x1 and y1 value to work
+    # so this should load
```
to266: The comments seem to be unnecessary

hyperspy/tests/io/test_hdf5.py
```diff
@@ -185,10 +186,47 @@ def test_binary_string(self):
...
+        assert os.path.exists(tmpfilepath + ".hspy")
+    finally:
+        preferences.General.hspy_extension = hspy_extension
+
+
+def test_hdf5_extension(tmpfilepath):
+    # test_marker_point_y2_data_deleted.hdf5 has 5 markers,
+    # where one of them is missing the y2 value, however the
+    # the point marker only needs the x1 and y1 value to work
+    # so this should load
```
to266: Here as well

doc/user_guide/io.rst
```diff
@@ -211,16 +211,40 @@ of arbitrary dimensions. It is based on the `HDF5 open standard
...
-are converted to numpy arrays when saved. This procedure homogenizes any types
-of the objects inside, most notably casting numbers as strings if any other
-strings are present:
+It is possible to use the ``.hspy`` extension by default by changing the
+value is ``preferences`` using the GUI or programatically:
+
+.. code-block:: python
+
+    >>> hs.preferences.General.hspy_extension = True
+    >>> hs.save() # make the changes permanent
```
to266: Does this save the preferences..? I'm not familiar

to266 (commented):
Just a couple of quick cleanup things and a question

PR #1491, opened 2017-02-26
Digital Micrograph annotation loading

magnunor:
With the permanent markers having been included, we can now load the Digital Micrograph (and other file format) annotations as permanent markers.

- [x] Finish implementation
- [x] Tests
- [ ] User guide

https://github.com/hyperspy/rosettasciio/issues/22


thomasaarholt:
Woohoo!

magnunor:
The basic functionality works now, at least for the files I generate. 

However, the DM-files which I make are for some reason fairly large: 10x10 bool datasets takes about 600 kB. So if I want to include several of them to test the different annotation types, it would use needlessly much space.

thomasaarholt:
Here's a 1x2 pixel binary with a line, circle and square on it. Comes out as 52kb as a dm4 or 4.19 kb as a zip. Perhaps it's possible to keep it as a zip as well?
[3anno.zip](https://github.com/hyperspy/hyperspy/files/1230374/3anno.zip)

thomasaarholt:
That said, do not spend time figuring out how to read a zip-file into Hyperspy. :smile:  It's not worth the effort right now. A 50 kb test file should be fine.

Also, I guess the code for the various annotations is very similar, so you don't need to test all of them, hence my 3-annotation file suggestion.

francisco-dlp:
[Here is an example](https://github.com/hyperspy/hyperspy/blob/RELEASE_next_minor/hyperspy/tests/io/test_edax.py#L23) of how to unzip files for testing.

magnunor:
@francisco-dlp, nice! That reduces 7 MB down to ~60 kB.

thomasaarholt:
Well done :)

magnunor:
The implementation should be ready for review now!

I wasn't sure if (or where) I should mention this in the user guide. In the markers page: http://hyperspy.org/hyperspy-doc/current/user_guide/visualisation.html#markers or the IO file format page: http://hyperspy.org/hyperspy-doc/current/user_guide/io.html#supported-formats

thomasaarholt:
It looks great!

A couple of questions:
There is currently no obvious way of *not* plotting with the annotation. Should there be?

What would be the expected behaviour of doing `s.T.T.plot()` if `s.plot()` shows a plot with an annotation? Currently the annotation is lost.

Should a marker (in the metadata) have the possibility of being switched on and off without being deleted?

magnunor:
To not plot the markers: `s.plot(plot_markers=False)`.

Any kind of slicing and transposing means all the annotations are lost. This is due to how the permanent markers themselves are implemented.

Individual markers can be toggled using `s.metadata.Markers.LineSegment10.plot_marker = False`. This is actually not documented in http://hyperspy.org/hyperspy-doc/current/user_guide/visualisation.html#markers, which it should be...

thomasaarholt:
It would be nice if the plot_markers was in the docstring for signal2D.plot(). Shift + Tab doesn't reveal it to me, as I would expect it.

Otherwise, this is great and looks ready to go :)

magnunor:
For me, in the ipython qtconsole, it shows up in the docstring for both Signal1D and Signal2D. However, it is only present in the signature for Signal1D, not in the signature for Signal2D.

hyperspy/io_plugins/digital_micrograph.py
```diff
@@ -1011,6 +1123,27 @@ def file_reader(filename, record_by=None, order=None, lazy=False):
...
+                    scale_x = axis['scale']
+                    offset_x = axis['offset']
+            try:
+                markers_dict = _get_markers_dict(
+                        dm.tags_dict, offset_x, offset_y, scale_x, scale_y)
+            except Exception as err:
+                _logger.warning(
+                        "Markers could not be loaded from the file"
+                        "due to: {0}".format(err))
+                marker_dict = {}
```
ericpre: Typo here: `markers_dict` instead of `marker_dict`.

hyperspy/io_plugins/digital_micrograph.py
```diff
@@ -969,6 +969,118 @@ def get_mapping(self):
...
+                        'x1': position[1]*scale_x+offset_x,
+                        'y2': position[2]*scale_y+offset_y,
+                        'x2': position[3]*scale_x+offset_x,
+                        'size': 20,
+                        'text': marker_text,
+                        }
+            temp_dict['marker_properties'] = marker_properties
+            name = temp_dict['marker_type'] + str(annotation['UniqueID'])
+            markers_dict[name] = temp_dict
+    return(markers_dict)
```
ericpre: Any particular reason to use brackets with the return statement?

ericpre (commented):
Would the code be a bit neater if the `_get_markers_dict` function would be a method of the `DigitalMicrographReader` class?
I have tried with a stack of images and the marker position does not seem to be correct.

Another thing but possibly out of the scope of this PR would be to add the option to draw the pointer widget -taken from another signal plot and provided as option of the plot- within the spectrum image area.

hyperspy/io_plugins/digital_micrograph.py
magnunor: Old bad habit... I'll change it

magnunor: Fixed!

hyperspy/io_plugins/digital_micrograph.py
magnunor: Fixed!

magnunor:
@ericpre, thanks for the feedback!

I'll see if I can make `_get_markers_dict` a method instead.

Do you have an image stack with annotations I could use?

Having the possibility to use a pointer in a subset of an image to change the positions in another signal, similarly to how you can in DM with spectrum images, would be pretty nice. However, as you say, that is outside the scope of this pull request.

ericpre:
@magnunor, please find [here](https://github.com/hyperspy/hyperspy/files/1257868/test.zip) the offending stack of images!

If I extract a single plane in DM and add an annotation, then the annotation is loaded properly as a marker in HyperSpy... I put both files in the zip.

magnunor:
@ericpre, I moved `get_markers_dict` into the `ImageObject` class, and abstracted the code somewhat.

I also fixed the bug with the image stack, which was due to the scale from the "z"-direction being used.

So everything should be good to go, as soon as the tests pass.

ericpre:
The roi line and the marker line does not behave exactly in the same: when a roi is loaded, the end positions of the line are in the middle of the pixel, while for marker line, it seems to be at the end of the pixel. With the current reference files, it is not easy to see, but if you use an image made of incremented values, then you can see clearly the difference. Do you think this can be improved?

magnunor:
I've seen the same thing for my tests, but wasn't really sure how to approach the issue. Since I think normally these annotations would be mostly used for SI survey images, which are typically 512x512 or higher number of pixels. Which means the shift of half a pixel wouldn't really be noticeable.

- The most elegant way of solving this would be to add some kind of "pixel alignment" functionality to the markers, where we could specify either "top", "bottom" or "centre" for the vertical alignment, and "left", "right" and "centre" for the horizontal alignment. 
- The not so elegant way is to add half a pixel value to the marker positions for the ROI annotations, but I don't know if that is a very good way, since it essentially modifies the coordinates... I'm also not certain where I would add these? Is it a simple -0.5x and -0.5y for every position?

Another thing which is lacking now how the markers are scaled/offset. If the offset or scale in a signal changes, they will move. However, the DM annotations will not move when changing the scale/offset inside DM. So there should be a "scaled" attribute in each marker, where if True (or False?) it should always stay in the same pixel position, even if the offset/scale is changed.

jat255:
@magnunor Just came across another need for this. Is there anything holding this back, currently? (given the 2.5 year gap...)

magnunor:
@jat255, there will probably be some merge conflicts, but that shouldn't be too tricky to work out.

In addition to that, there are two potential blockers:

* The markers will be half a pixel offset in relation to how they're visualized in DM
* The markers will not follow changes in offset and scale

Both of these will require some more extensive changes in the marker functionality, which I do not have time to implement.

While it would be nice to have these things implemented as well, I think they are minor enough that the pull request can be merged. Further improvements can be done in future, if people need them.

jlaehne:
+1 on merging this pretty soon.

While playing with this, I realized that when a new file without annotations is read, replacing the same object, the annotations from the original object are kept. I don't know if that should be seen as a bug? Probably it is a rather unusual use case to not create a new object for a new load.

**Edit:** If the newly loaded file also includes annotations, they are replaced. The problem is only when replacing an object with annotations by one without.

jlaehne:
Can we get this in soon, to at least have the functionality available, just realized again that hardly any other program reading DM4 files can display the annotations. It seems to be tested quite intensively. Any tweaks could be added later and we could open an issue with nice to have features.

ericpre:
We can merge even if we know that there are issues with these markers. Usually, I am not keen to merge changes with known issues because it is very rare that it is followed up and fixed...

magnunor:
Yeah, it is a bit of a dilemma. Getting the full functionality working properly will require adding several things to HyperSpy's marker implementation. I'll be happy to wrap up this pull request, but I don't have time to implement the additional marker functionality.

ericpre:
There are two failures in the test suite.

magnunor:
A bit busy at the moment (exam period...), but I'll see if I can get this done in a couple of weeks. Feel free to ping me if I haven't fixed this over new-years.

jlaehne:
Any progress?

magnunor:
@jlaehne, unfortunately not. Most of my development time has been taken up by the various lazy processing/plotting improvements. The most recent one being reviewing the `.map` improvements.

This is on my todo list, and I'd like to do it, but it has shifted a bit down in the priority list due to other things coming along.

nem1234:
Any progress?
Should I wait for it to be finished?

I'm trying to continue this PR.
May be I need to add 

- arrow and elipsoid into  drawing._markers,
- (-0.5,-0.5) offset for marker positions,
- font settings, 
- text background colors.

Should I create a new PR?

jlaehne:
@magnunor can you merge in RnM?

magnunor:
I got this synced up to Rmn, some tests failing.

jlaehne:
Does  hyperspy/hyperspy#2871 bring this one further towards completion?

magnunor:
I'm not sure actually, when I had a look at this a month ago there were some things not working.

I'll do a check of this once https://github.com/hyperspy/hyperspy/pull/2871 is merged.

magnunor:
There is a number of things failing, including strange behaviour in the marker functionality. So I probably won't have time to address this in quite a while.
I'm also a bit nervous about touching I/O functionality this close to a release, since messing up things relating to file loading is really bad.

So I will not have time to finish this for the 1.7 release, but if anyone else wants to give this a shot, feel free!

nem1234:
Just a comment.
For me, restarting from the current RELEASE_next_minor seems easier than merging many patches from RELEASE_next_minor into current branch.
And also, for me, the most strange behaviour of markers is the default z-ordering. Alphabetical order.....

jlaehne:
If continued, needs to be based on and adapted to hyperspy/hyperspy#2972?

ericpre:
Closing as this should be dealt with in https://github.com/hyperspy/rosettasciio.

PR #1495, opened 2017-02-28
Improve Error when using s.plot() with unsupported signal dimensions

magnunor:


hyperspy/signal.py
```diff
@@ -1864,7 +1864,11 @@ def plot(self, navigator="auto", axes_manager=None,
         elif axes_manager.signal_dimension == 2:
             self._plot = mpl_hie.MPL_HyperImage_Explorer()
         else:
-            raise ValueError('Plotting is not supported for this view')
+            raise ValueError(
+                    "Plotting is not supported for this view. "
+                    "Try 's.transpose(signal_axes=1).plot()' for plotting as "
```
francisco-dlp: Maybe "Try e.g. 's.transpose[...]" to stress that there is a choice to be made?

magnunor: Changed it to "Try e.g. 's.transpose(signal_axes=1).plot()' for ..."

francisco-dlp:
Thanks!

PR #1496, opened 2017-02-28
Update digital_micrograph.py

erh3cq:
Add support for microscope name stored under Microscope_Info.Microscope
Add support for aperture_size when EELS.aperture_size is '2mm' or '2 mm'

### DIFF
--- a/hyperspy/io_plugins/digital_micrograph.py
+++ b/hyperspy/io_plugins/digital_micrograph.py
@@ -924,7 +924,7 @@ class ImageObject(object):
                     None),
                 "ImageList.TagGroup0.ImageTags.EELS_Spectrometer.Aperture_label": (
                     "Acquisition_instrument.TEM.Detector.EELS.aperture_size",
-                    lambda string: float(string.replace('mm', '').replace(' ',''))),
+                    lambda string: float(string.replace('mm', ''))),
                 "ImageList.TagGroup0.ImageTags.EELS Spectrometer.Instrument name": (
                     "Acquisition_instrument.TEM.Detector.EELS.spectrometer",
                     None),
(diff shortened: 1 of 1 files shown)


hyperspy/io_plugins/digital_micrograph.py
```diff
@@ -922,7 +924,7 @@ def get_mapping(self):
                     None),
                 "ImageList.TagGroup0.ImageTags.EELS_Spectrometer.Aperture_label": (
                     "Acquisition_instrument.TEM.Detector.EELS.aperture_size",
-                    lambda string: float(string.replace(' mm', ''))),
+                    lambda string: float(string.replace('mm', '').replace(' ',''))),
```
francisco-dlp: Maybe just the following would do?

```python

lambda string: float(string.replace('mm', '')),

```

francisco-dlp:
If you are quick at replying to the little comment above this may make it into v1.2 (due today). By the way, that will add you to the authors in our zenodo doi but, unless you update your name and affiliation in your github profile now, you'll appear as erh3cq.

hyperspy/io_plugins/digital_micrograph.py
erh3cq: then it leaves the string as '2 ' and i think it needs '2'?

erh3cq:
Updated. Thank you!

hyperspy/io_plugins/digital_micrograph.py
francisco-dlp: I think that ``float`` wouldn't care about spaces.

erh3cq: yes you are correct.  How do I update this pull (just starting with gitHub)

PR #1497, opened 2017-02-28
NEW_find_peaks_2D

dnjohnstone:
A replacement for #1292 to move things along.

This uses map() so that it's compatible with new lazy_signal operations.


hyperspy/_signals/signal2d.py
```diff
@@ -650,6 +649,74 @@ def add_ramp(self, ramp_x, ramp_y, offset=0):
         ramp += ramp_y * yy
         self.data += ramp
 
+    def find_peaks2D(self, method='skimage', *args, **kwargs):
```
to266: This needs the usual `show_progressbar` and `parallel` kwargs that come with `map` methods.

Also, tests should probably parametrized for both `parallel` and `lazy`

hyperspy/tests/signal/test_2D_tools.py
```diff
@@ -140,6 +142,84 @@ def test_add_ramp_lazy():
...
+ref = hs.signals.Signal2D(ref)
+ans = np.empty((1,), dtype=object)
+ans[0] = np.array([[xref, yref]])
+
+
+datasets = [dense, sparse0d, sparse1d, sparse2d]
+
+@pytest.mark.parametrize('method', peak_methods)
+@pytest.mark.parametrize('dataset', datasets)
+def test_creates_signal(method, dataset):
```
to266: Example parametrization for both `lazy` and `parallel` (only need to run parallel when not lazy, so three options):

```python
@pytest.mark.parametrize('method', peak_methods)
@pytest.mark.parametrize('dataset', datasets)
@pytest.mark.parametrize('lazy,parallel', [(True, None), (False, True), (False, False)])
def test_creates_signal(method, dataset, lazy, parallel):
    if lazy:
        dataset = dataset.as_lazy()
    peaks = dataset.find_peaks2D(method=method, parallel=parallel)
    assert isinstance(peaks, hs.signals.BaseSignal)

Comment (to266) [reply hyperspy/tests/signal/test_2D_tools.py]: Same for other tests

**Review by to266 — CHANGES_REQUESTED** [review]
I think some of the tests/functions should be made "optional" via pytest to not fail in minimal testing environments.
In addition, I think it might be useful to move the peakfinding tests to a new test file (simply copy-pasting).

File: hyperspy/tests/signal/test_2D_tools.py (hunk 2 above)
Comment (to266) [reply hyperspy/tests/signal/test_2D_tools.py]: Also, to skip the `'stat'` method without `sklearn`, add this in the test:
```python
#@pytest.mark.parametrize(....)
def test_whatever_name(method, ....):
    if method=='stat':
        _ = pytest.importorskip("sklearn")
    # continue with the test
```
Could not find a nice way to do that with `parametrize`, but that should suffice

magnunor:
I just noticed this through the pycrystem project, and this would be _really_ nice for finding atomic columns for Atomap.

What is needed for getting this merged?

francisco-dlp:
In addition to @to266 comments, the ipywidgets should be moved to hyperspy/hyperspy_gui_ipywidgets.

ericpre:
I have solved the merge conflicts and updated this branch to address the comments of the review.
- [x] Add traitsui widget (https://github.com/hyperspy/hyperspy_gui_traitsui/pull/20),
- [x] add ipywidget widget (https://github.com/hyperspy/hyperspy_gui_ipywidgets/pull/20),
- [x] fix a few issue with ragged array,
- [x] add more tests.

francisco-dlp:
It seems that parts of all of this PR ended up in pyxem, see e.g. https://github.com/pyxem/pyxem/blob/6b23f629a48ec966553c785eba177e343e90d311/pyxem/utils/peakfinders2D.py . It's probably not a bad idea (given its generality) to incorporate it in HyperSpy as originally planned.

What do you think about it @bm424  and @dmlundeby?

dmlundeby:
Me? I don't have any strong opinion on this, I guess you're referring to someone else 😅

francisco-dlp:
Sorry, I meant @dnjohnstone

ericpre:
There is a similar `find_peaks` method in pixstem: https://gitlab.com/pixstem/pixstem/blob/master/pixstem/pixelated_stem_class.py#L776
I haven't look at atomap but I would imagine that it could be useful for initial peaks position measurement.

@dnjohnstone and @magnunor do you have comments on this PR?

This tool could also be used to accommodate more generic blob detection methods in order to measure nanoparticles, defects, etc. We could think having sets of parameters for specific case, which will select an algorithm and provide initial parameter values suitable to specific application.

dnjohnstone:
This looks fine to me, thanks for tidying it up @ericpre - in general I agree with @francisco-dlp that it's probably a better fit in the main hyperspy repo than in pyxem.

There are probably a few things like this in pyxem that could be upstreamed to hyperspy and maybe we should do that at some point.

dnjohnstone:
@ericpre I've merged this up, made some docstring updates and added a cross-correlation method that we had in pyxem but did not have in this PR.

It would be great if we could get it in to hyperspy-1.5.0 so that we can address pyxem/pyxem#354 in a reasonable timeframe.

dnjohnstone:
The failure here is in the HologramImage tests so I don't think it has anything to do with this PR.

@ericpre / @francisco-dlp do you need anything more doing on this one?

hyperspy/utils/peakfinders2D.py
```diff
@@ -0,0 +1,455 @@
...
+            m_peaks = len(peaks)
+        peak_centers = np.array(
+            [np.mean(peak, axis=0) for peak in peaks])  # 7
+        return peak_centers
+
+    return clean_peaks(np.round(stat_peak_finder(z)).astype(int))
+
+
+def find_peaks_dog(z, min_sigma=1., max_sigma=50., sigma_ratio=1.6,
+                   threshold=0.2, overlap=0.5, exclude_border=False):
```
ericpre, line 453: Why adding `exclude_border` if this is not used?

hyperspy/utils/peakfinders2D.py
```diff
@@ -0,0 +1,455 @@
...
+        if len(np.intersect1d(center, (0, 1) + z.shape + tuple(
+                        c - 1 for c in z.shape))) > 0:
+            continue
+        clean_centers.append(center)
+    return np.array(clean_centers)
+
+
+def find_peaks_log(z, min_sigma=1., max_sigma=50., num_sigma=10,
+                   threshold=0.2, overlap=0.5, log_scale=False,
+                   exclude_border=False):
```
ericpre, line 501: Same as above.

ericpre (commented):
@dnjohnstone, I wrote two comments for your recent commits.

This is one major issue I forgot to mention and I want to discuss, when used in interactive mode with the GUI, it not possible to get the output because the peaks positions are saved as an attribute of the `PeaksFinder2D`! The usual pattern with GUIs in HyperSpy is to do the operation in place, but obviously this can not work here, so we need to find another approach.

ericpre:
> It would be great if we could get it in to hyperspy-1.5.0 so that we can address pyxem/pyxem#354 in a reasonable timeframe.

I think there are still a few things not right and we may be too short in time to get this PR merged for 1.5.

hyperspy/utils/peakfinders2D.py
dnjohnstone: Thanks - it was supposed to be used as a kwarg for the skimage blob_dog method - I've made it actually use this and updated the skimage dependency to >=0.15 as that is when this kwarg was introduced.

hyperspy/utils/peakfinders2D.py
dnjohnstone: Fixed as above

dnjohnstone:
> @dnjohnstone, I wrote two comments for your recent commits.
> 
> This is one major issue I forgot to mention and I want to discuss, when used in interactive mode with the GUI, it not possible to get the output because the peaks positions are saved as an attribute of the `PeaksFinder2D`! The usual pattern with GUIs in HyperSpy is to do the operation in place, but obviously this can not work here, so we need to find another approach.

Thanks @ericpre for the review! I've dealt with the exclude border thing I think.

Regarding the GUI - we never really intended that to be a way to get the output, just a way to tweak parameters a bit to find something suitable before running the function properly in a notebook or script. I felt like that was a good balance in terms of getting the obvious benefits of being able to see what was being found with different parameters and not being too time consuming to code up.

Would it be possible to find a similar compromise here to move forward?

My opinion is that a working set of peak finders is useful progress and I also know that, at least personally, I'm not going to find time to work on convenience GUI features in this area any time soon.

ericpre:
I am wondering if we should split the `find_peaks2D` method into a `find_blob` and a `find_peaks2D`? They will both be doing the same thing, simply having different default parameter.

dnjohnstone:
> I am wondering if we should split the `find_peaks2D` method into a `find_blob` and a `find_peaks2D`? They will both be doing the same thing, simply having different default parameter.

Hmmmm - I can see why that's a good idea, but also think it's a slightly strange code duplication. Ultimately, I'm pretty flexible on this topic, I think I'd rather keep it as one function and settle on some defaults, but up to you.

dnjohnstone:
> I am wondering if we should split the `find_peaks2D` method into a `find_blob` and a `find_peaks2D`? They will both be doing the same thing, simply having different default parameter.

Having thought about this a bit more, I really think it would be helpful if we could initially get one implementation of this in and then worry about relatively minor things like default parameters later...

magnunor:
@dnjohnstone, very nice! I haven't had time to go through the implementation in detail, but I would definitely utilize this in `Atomap`.

I might also use it for `pixStem` as well, depending on the performance. If the performance is slower, I might make a pull request with performance improvements in the future.

With regards to adding extra features to the pull request, I suggest merging with the current feature set, as i) it would be very useful even without the extra features, ii) we can add more in the future in other pull requests.

dnjohnstone:
> @dnjohnstone, very nice! I haven't had time to go through the implementation in detail, but I would definitely utilize this in `Atomap`.
> 
> I might also use it for `pixStem` as well, depending on the performance. If the performance is slower, I might make a pull request with performance improvements in the future.
> 
> With regards to adding extra features to the pull request, I suggest merging with the current feature set, as i) it would be very useful even without the extra features, ii) we can add more in the future in other pull requests.

Great! Do we think anything further needs doing on this then or can we get it done?

dnjohnstone:
So the test failure here is to in test_hologram_image - does that mean we can basically get this in?

The updates to map_result_construction would be very useful for us to have in ASAP as it is a bug fix for ragged arrays which affects quite a lot of what we do in pyxem. Could it be quicker to have a 1.5.3 with that bug fix in it?

dnjohnstone:
@ericpre what remains to be done on this? I'm happy to do some of it, especially if it would get this in faster

doc/user_guide/signal2d.rst
```diff
@@ -76,3 +76,92 @@ while the offset is determined by the `offset` parameter. The fulcrum of the
...
+    >>> s.find_peaks2D(method='laplacian_of_gaussians')
+    >>> s.find_peaks2D(method='difference_of_gaussians')
+
+These methods are essentially wrappers around the
+`scikit-image <http://scikit-image
+.org/docs/dev/auto_examples/plot_blob.html>`_ Laplacian
+of Gaussian and Difference of Gaussian methods, based on stacking the
+Laplacian/difference of images convolved with Gaussian kernels of various
+standard deviations. Both are very rapid and relatively robust, given
+appropriate parameters.
```
tjof2: Can you define "appropriate parameters" any better?

hyperspy/_signals/signal2d.py
```diff
@@ -732,6 +740,116 @@ def add_ramp(self, ramp_x, ramp_y, offset=0):
...
+
+                 'laplacian_of_gaussian' - a blob finder implemented in
+                 `scikit-image` which uses the laplacian of Gaussian
+                 matrices approach.
+
+                 'difference_of_gaussian' - a blob finder implemented in
+                 `scikit-image` which uses the difference of Gaussian
+                 matrices approach.
+
+                 xc' - A cross correlation peakfinder
```
tjof2: missing "'" - and is there any more detail you can give on what it does/where it comes from?

hyperspy/_signals/signal2d.py
```diff
@@ -732,6 +740,116 @@ def add_ramp(self, ramp_x, ramp_y, offset=0):
...
+        """
+        method_dict = {
+            'local_max': peak_local_max,
+            'max': find_peaks_max,
+            'minmax': find_peaks_minmax,
+            'zaefferer': find_peaks_zaefferer,
+            'stat': find_peaks_stat,
+            'laplacian_of_gaussian':  find_peaks_log,
+            'difference_of_gaussian': find_peaks_dog,
+            'cross_correlation' : find_peaks_xc,
```
tjof2: does this match the "'xc'" in the docstring?

hyperspy/signal.py
```diff
@@ -3885,7 +3885,8 @@ def map(self, function,
         ragged : None or bool
             Indicates if the results for each navigation pixel are of identical
             shape (and/or numpy arrays to begin with). If ``None``,
-            the appropriate choice is made while processing. Note: ``None``
+            the appropriate choice is made while processing.If True in case of lazy signal,
```
tjof2: space after "."

dnjohnstone:
Just while I remember - we should make sure the performance improvements from @tjof2 that were made here: https://github.com/pyxem/pyxem/pull/582 make it in to this version!

And some other changes here: https://github.com/pyxem/pyxem/pull/581/files

ericpre:
I am afraid I will not have time to do much more on this PR, some else should continue.

tjof2:
@dnjohnstone would it perhaps be any easier if you just copied over the block of code that works (?) in pyxem along with the tests in a new PR in one go?

dnjohnstone:
> @dnjohnstone would it perhaps be any easier if you just copied over the block of code that works (?) in pyxem along with the tests in a new PR in one go?

I think unfortunately the pxyem/hyperspy versions of this have diverged a bit - the version in this PR includes some updates to how map() works for lazy signals and does some GUI stuff more properly, whereas the pyxem one has some updates to the individual peak finding methods (mostly yours). We probably want all of these things and for that reason I think it's better to try to get everything in to this PR.

But certainly happy to copy over the code block with the actual peak finders.

tjof2:
Ah ok makes sense, thanks

dnjohnstone:
Is there any chance of getting this in for v1.6 @francisco-dlp ? I think this PR is working and has been reviewed. I can spend a bit of time on this towards the end of the week if it will get it in...

francisco-dlp:
I haven't reviewed it, so that's a question for @ericpre and/or @tjof2.

ericpre:
I will not be able to help on this, this week. From my perspective, it is in good shape, if we ignore one bug with the mark position when changing navigation index through the GUI widgets.
I have merged the conflict but there is a test failure on travis, which is most likely related to different dependency versions, travis uses more recent versions.
@dnjohnstone and @tjof2, it would be great if you could finish it!

This PR needs the GUI widgets, see https://github.com/hyperspy/hyperspy/pull/1497#issuecomment-495654049:

> I have solved the merge conflicts and updated this branch to address the comments of the review.
> 
> * [x]  Add traitsui widget ([hyperspy/hyperspy_gui_traitsui#20](https://github.com/hyperspy/hyperspy_gui_traitsui/pull/20)),
> * [x]  add ipywidget widget ([hyperspy/hyperspy_gui_ipywidgets#20](https://github.com/hyperspy/hyperspy_gui_ipywidgets/pull/20)),
> * [x]  fix a few issue with ragged array,
> * [x]  add more tests.

tjof2:
@dnjohnstone if you are going to work on it, I'm happy to review (and have the time to)

dnjohnstone:
Great, that sounds like a plan. I'll try to get to this tomorrow.

dnjohnstone:
@tjof2 do you have any ideas on this travis failure? The issue is that the order of the peaks has changed - I guess perhaps due to travis using scikit-image-0.17.2 where azure pipelines is using scikit-image-0.16.2 . I'm not sure that the order actually matters, so long as it is consistent, but I guess that might only be resolved by fixing dependencies to newer versions and going with that. I'm fine with setting dependencies to new versions, but others might not want that?

tjof2:
@dnjohnstone why not sort it internally to avoid it:

```
ind = np.lexsort((a[:,0], a[:,1]))    
return a[ind]
```
And so always return a sorted array of 2D pairs

ericpre:
> @dnjohnstone why not sort it internally to avoid it:
> 
> ```
> ind = np.lexsort((a[:,0], a[:,1]))    
> return a[ind]
> ```
> 
> And so always return a sorted array of 2D pairs

If it decreases performance for a large number of peaks (or large navigation dimension), it may be worth doing only for testing?

tjof2:
> If it decreases performance for a large number of peaks (or large navigation dimension), it may be worth doing only for testing?

```
import numpy as np    

def sort_pairs(a):
    ind = np.lexsort((a[:,1],a[:,0]))    
    return a[ind]

a = np.random.randn(1000, 2)
%timeit sort_pairs(a)
# 89 µs ± 1.06 µs per loop (mean ± std. dev. of 7 runs, 10000 loops each)

a = np.random.randn(10000, 2)
%timeit sort_pairs(a)
# 1.34 ms ± 9.49 µs per loop (mean ± std. dev. of 7 runs, 1000 loops each)
```
@ericpre  I don't think that's too bad a performance hit to do as a final bit of post-processing. But good point.

tjof2:
Hmmm why didn't that work?

hyperspy/utils/peakfinders2D.py
```diff
@@ -0,0 +1,509 @@
...
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with  HyperSpy.  If not, see <http://www.gnu.org/licenses/>.
+
+import numpy as np
+import scipy.ndimage as ndi
+from numba import njit
```
tjof2: Numba is currently an optional dependency in HyperSpy. There's a `jit_ifnumba` decorator that would be better. https://github.com/hyperspy/hyperspy/blob/30d28d23f074bbccc92ff6a537addb22566d2d1d/hyperspy/decorators.py#L111

`njit` is just an alias for `jit(nopython=True)`, and passing `cache=True` will work fine here as well.

hyperspy/utils/peakfinders2D.py
```diff
@@ -0,0 +1,534 @@
...
+    """JIT-compiled standard deviation of array.
+    Parameters
+    ----------
+    X : numpy.ndarray
+        Input array.
+    Returns
+    -------
+    float
+        Standard deviation of X.
+    Notes
```
tjof2, line 67: Line breaks between sections needed here and the function above

tjof2:
Just a couple of comments above, plus resolving the sorted array issue.

dnjohnstone:
> Just a couple of comments above, plus resolving the sorted array issue.

Thanks - I've also just broken (semi-intentionally) the stat method, I think it had a bug in it though and now has a different issue.

dnjohnstone:
> Hmmm why didn't that work?

That's a very good question - short answer is I have absolutely no idea and it's very irritating right now ;)

tjof2:
I guess (a) verify my snippet with lexsort actually works and (b) check that it is actually calling and using the sort function when expected :) perhaps there's a code branch where it doesn't get called.

hyperspy/_signals/signal2d.py
```diff
@@ -778,6 +786,126 @@ def add_ramp(self, ramp_x, ramp_y, offset=0):
...
+            pf2D = PeaksFinder2D(self, method=method, peaks=peaks, **kwargs)
+            pf2D.gui(display=display, toolkit=toolkit)
+        elif current_index:
+            peaks = method_func(self.__call__(), **kwargs)
+        else:
+            peaks = self.map(method_func, show_progressbar=show_progressbar,
+                             parallel=parallel, inplace=False,
+                             ragged=True, **kwargs)
+
+        return peaks
```
tjof2, line 906: @dnjohnstone why not call the sort here to be certain it's the last thing that runs before it returns?

dnjohnstone:
> I guess (a) verify my snippet with lexsort actually works and (b) check that it is actually calling and using the sort function when expected :) perhaps there's a code branch where it doesn't get called.

Yeah I think it's the code branch where clean_peaks doesn't get called option. The snippet is good. And I'm pretty sure when the currently failing test runs clean_peaks is not being used because I made clean_peaks return some garbage array and that didn't break it. It's more that I can't find that code branch...

hyperspy/_signals/signal2d.py
tjof2: Just in case, for example, there's anything amiss with `.map()` and multiple threads...though it fails for `parallel=False` too I guess.

dnjohnstone:
Ah! Got it - peak_local_max is called directly from skimage so it doesn't go through clean peaks... will fix that.

dnjohnstone:
The test_close_find_peaks test that is now failing turns out to fail for most methods (but not all). Will have a look at that tomorrow.

doc/user_guide/signal2d.rst
```diff
@@ -115,3 +115,136 @@ while the offset is determined by the `offset` parameter. The fulcrum of the
...
 :py:class:`~.axes.AxesManager` of the signal.
+
+
+Peak finding
+------------
+
+.. versionadded:: 1.6
+
+The :py:meth:`~._signals.signal2d.Signal2D.find_peaks` method provides access
+to a number of algorithms for that achieve peak finding in two dimensional
```
tjof2: remove "that"

ericpre:
Great, all green now! :)

doc/user_guide/signal2d.rst
```diff
@@ -115,3 +115,136 @@ while the offset is determined by the `offset` parameter. The fulcrum of the
...
+signals. The methods available are as follows:
```
tjof2: no need for "as follows"

doc/user_guide/signal2d.rst
```diff
@@ -115,3 +115,136 @@ while the offset is determined by the `offset` parameter. The fulcrum of the
...
+.. code-block:: python
+
+    >>> s.find_peaks(method='local_max')
+    >>> s.find_peaks(method='max')
+    >>> s.find_peaks(method='minmax')
+
+These methods are searching peaks using maximum (and mimimum) values in the
+image. There all have a ``distance`` parameter to set the minimum distance
+between the peaks.
+- the ``'local_max'`` method uses the :py:func:`skimage.feature.peak_local_max`
```
tjof2, line 142: need a line break here before the listfor the docs to render correctly

doc/user_guide/signal2d.rst
```diff
@@ -115,3 +115,136 @@ while the offset is determined by the `offset` parameter. The fulcrum of the
...
+  the maximum and minimum filtered images.
+
+Zaeferrer peak finder
+^^^^^^^^^^^^^^^^^^^^^
+
+.. code-block:: python
+
+    >>> s.find_peaks(method='zaefferer')
+
+This algorithm was developed by Zaefferer [1]_ and the
```
tjof2: References tend to go in `user_guide/bibliography.rst` I think

doc/user_guide/signal2d.rst
```diff
@@ -115,3 +115,136 @@ while the offset is determined by the `offset` parameter. The fulcrum of the
...
+
+Ball statistical peak finder
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+.. code-block:: python
+
+    >>> s.find_peaks(method='stat')
+
+Developed by Gordon Ball, and described in the Ph.D. thesis of Thomas A.
+White, this method is based on finding points which have a statistically
```
tjof2: Worth referencing the thesis properly at least? It's mentioned more than once.

doc/user_guide/signal2d.rst
```diff
@@ -115,3 +115,136 @@ while the offset is determined by the `offset` parameter. The fulcrum of the
...
+higher value than the surrounding areas, then iterating between smoothing and
+binarising until the number of peaks has converged. This method is slow, but
```
tjof2: It _can be slow_. Perhaps mention that if you have `numba` installed, it will automatically run much faster?

doc/user_guide/signal2d.rst
```diff
@@ -115,3 +115,136 @@ while the offset is determined by the `offset` parameter. The fulcrum of the
...
+
+.. code-block:: python
+
+    >>> x, y = np.meshgrid(np.arange(-2, 2.5, 0.5), np.arange(-2, 2.5, 0.5))
+    >>> template = hs.model.components2D.Gaussian2D().function(x, y)
+    >>> s.find_peaks(method='template_matching', template=template)
+
+This method locate peaks in the cross correlation between the image and a
+template followed by the :py:func:`~.utils.peakfinders2D.find_peaks_minmax`
+function. See
```
tjof2: See... ? Hanging sentence.

doc/user_guide/signal2d.rst
```diff
@@ -115,3 +115,136 @@ while the offset is determined by the `offset` parameter. The fulcrum of the
...
+  the new peaks.
+
+.. note:: Some methods take significantly longer than others, particularly
+   where there are a large number of peaks to be found. The plotting window
+   may be inactive during this time.
+
+References
+----------
+
+.. [1] S. Zaefferer, “New developments of computer-aided
```
tjof2: Move to `user_guide/bibliography.rst`?

hyperspy/_signals/signal2d.py
```diff
@@ -778,6 +785,126 @@ def add_ramp(self, ramp_x, ramp_y, offset=0):
...
+
+    find_peaks.__doc__ %= (SHOW_PROGRESSBAR_ARG, PARALLEL_ARG,
+                           DISPLAY_DT, TOOLKIT_DT)
```
tjof2, line 909: You should probably include `https://github.com/hyperspy/hyperspy/blob/30d28d23f074bbccc92ff6a537addb22566d2d1d/hyperspy/docstrings/signal.py#L75` for the docstring for `max_workers`.

hyperspy/_signals/signal2d.py
```diff
@@ -778,6 +785,126 @@ def add_ramp(self, ramp_x, ramp_y, offset=0):
...
+            peaks = BaseSignal(np.empty(self.axes_manager.navigation_shape),
+                               axes=axes_dict)
+            pf2D = PeaksFinder2D(self, method=method, peaks=peaks, **kwargs)
+            pf2D.gui(display=display, toolkit=toolkit)
+        elif current_index:
+            peaks = method_func(self.__call__(), **kwargs)
+        else:
+            peaks = self.map(method_func, show_progressbar=show_progressbar,
+                             parallel=parallel, inplace=False,
+                             ragged=True, **kwargs)
```
tjof2: pass `max_workers=...`

hyperspy/_signals/signal2d.py
```diff
@@ -778,6 +785,126 @@ def add_ramp(self, ramp_x, ramp_y, offset=0):
         ramp += ramp_y * yy
         self.data += ramp
 
+    def find_peaks(self, method='local_max', interactive=True,
+                   current_index=False, show_progressbar=None,
+                   parallel=None, display=True, toolkit=None, **kwargs):
```
tjof2: `parallel=None, max_workers=None, ...` for consistency with other methods like `s.align()`. See notes below for the docstring as well.

hyperspy/_signals/signal2d.py
```diff
@@ -778,6 +785,126 @@ def add_ramp(self, ramp_x, ramp_y, offset=0):
...
+        'threshold' argument, which will be map to the 'min_distance' and
+        'threshold_abs' of the :py:func:`skimage.feature.peak_local_max`
+        function.
+
+        Returns
+        -------
+        peaks : :py:class:`~hyperspy.signal.BaseSignal` or numpy.ndarray if current_index=True
+            Array of shape `_navigation_shape_in_array` in which each cell
+            contains an array with dimensions (npeaks, 2) that contains
+            the x, y pixel coordinates of peaks found in each image.
```
tjof2: Mention that it is a _sorted_ array here (first along `y` then `x` I think)?

hyperspy/utils/peakfinders2D.py
```diff
@@ -0,0 +1,572 @@
...
+
+    Parameters
+    ----------
+    peaks : numpy.ndarray
+        Array of found peaks.
+
+    Returns
+    -------
+    peaks : numpy.ndarray
+        Sorted array of found peaks.
```
tjof2: Mention the sort order perhaps.

hyperspy/utils/peakfinders2D.py
```diff
@@ -0,0 +1,572 @@
...
+    if len(peaks) == 0:
+        return NO_PEAKS
+    else:
+        ind = np.lexsort((peaks[:,0], peaks[:,1]))
+        return peaks[ind]
+
+
+def find_local_max(z, min_distance=1, threshold_abs=None, threshold_rel=None,
+                   exclude_border=True, indices=True, num_peaks=np.inf,
+                   footprint=None, labels=None, num_peaks_per_label=np.inf):
```
tjof2: Can you clarify why you don't just pass `**kwargs` here?

tjof2 (changes requested):
A few small comments @dnjohnstone, otherwise looks good.

hyperspy/utils/peakfinders2D.py
dnjohnstone: I did initially but then changed it to this for consistency with the log/dog methods already here. One advantage of explicitly stating these is that it's locally documented a bit better. I'm ambivalent about what is "right" though...

tjof2: OK. I'll defer to @francisco-dlp on the preferred approach here.

francisco-dlp: @dnjohnstone's approach is certainly better in terms of usability, but it is brittle because any changes in the skimage function will break it or make it incomplete, increasing the maintenance burden. Hence, +1 for ``**kwargs``.

dnjohnstone: @ericpre is there any reason I'm missing why you hadn't used **kwargs in other places already? e.g. does it cause some issue with the GUI?

doc/user_guide/signal2d.rst
```diff
@@ -135,9 +135,10 @@ Maximum based peak finder
     >>> s.find_peaks(method='max')
     >>> s.find_peaks(method='minmax')
 
-These methods are searching peaks using maximum (and mimimum) values in the
+These methods search for peaks using maximum (and mimimum) values in the
```
tjof2: Typo (minimum)

hyperspy/_signals/signal2d.py
```diff
@@ -858,7 +859,8 @@ def find_peaks(self, method='local_max', interactive=True,
         peaks : :py:class:`~hyperspy.signal.BaseSignal` or numpy.ndarray if current_index=True
             Array of shape `_navigation_shape_in_array` in which each cell
             contains an array with dimensions (npeaks, 2) that contains
-            the x, y pixel coordinates of peaks found in each image.
+            the `x, y` pixel coordinates of peaks found in each image sorted
+            first along y and then along `x`.
```
tjof2: Backticks around y?

dnjohnstone:
I've got some issue with the DOC converting to string with MAX_WORKERS_ARG in there - not sure what I've done wrong there?

tjof2:
> I've got some issue with the DOC converting to string with MAX_WORKERS_ARG in there - not sure what I've done wrong there?

You need an extra line with  `%s` in the docstring I think such that it gets replaced.

hyperspy/_signals/signal2d.py
```diff
@@ -778,6 +785,128 @@ def add_ramp(self, ramp_x, ramp_y, offset=0):
...
+
+        interactive : bool
+            If True, the method parameter can be adjusted interactively.
+            If False, the results will be returned.
+        current_index : bool
+            if True, the computation will be performed for the current index.
+        %s
+        %s
+        %s
+        %s
```
tjof2, line 844: Another `%s` here for `MAX_WORKERS_ARG`

tjof2:
Last two typos above and the **kwargs and then LGTM.

dnjohnstone:
> Last two typos above and the **kwargs and then LGTM.

Great, thanks! Think we should be good to go then.

tjof2:
The docs build failure looks unrelated (it was working up to now). Let's wait for the CI to finish

dnjohnstone:
> The docs build failure looks unrelated (it was working up to now). Let's wait for the CI to finish

Yeah, that was a surprise failure given I'd only tweaked a couple of typos there... But certainly CI should finish

dnjohnstone:
Thanks very much to everyone for all their work on this - especially @ericpre and @tjof2

tjof2:
Well done for getting it over the line @dnjohnstone!

PR #1498, opened 2017-02-28
Update signal.py

erh3cq:
Added an indexmin() and valuemin() similar to indexmax() and valuemax().
This file does not contain the new marker import feature. It kept raising errors on my build.


francisco-dlp:
``valuemin`` is duplicated and ``indexmin`` is not present. Beside it requires adding unittests [like this one](https://github.com/hyperspy/hyperspy/blob/RELEASE_next_minor/hyperspy/tests/signal/test_tools.py#L196).

francisco-dlp:
Looks good, thanks!

PR #1501, opened 2017-03-01
Add Py36 environments to travis

francisco-dlp:
This also uses miniconda for the minimum environment instead of the system python as it was before to make the code more compact.

This is WIP and is currently broken. travis is having issues today and the log is currently unavailable, so I don't know why it's broken. Any help would be appreciated.


sem-geologist:
Do we really need such huge matrix?
min env should not go with miniconda, it kills completely the purpoise of minimal python. Are we moving completely to py36? Did conda moved there? debian testing is still on 35.

.travis.yml
```diff
@@ -2,82 +2,61 @@ language: python
...
-      if [[ $TRAVIS_OS_NAME == 'linux' ]]; then
-        pip uninstall -y matplotlib;
-        pip install matplotlib;
-      fi;
-      python setup.py install;
-      pip install pytest pytest-xdist;
-    else pip install pytest;
-    pip3 install .;
+      DEPS="pip numpy scipy matplotlib ipython h5py sympy scikit-learn dill natsort setuptools scikit-image cython lxml ipyparallel dask traits traitsui";
+    else DEPS="pip ipython numpy scipy matplotlib h5py sympy scikit-image";
```
sem-geologist, line 31: this make no sense, why we need to install `ipython numpy scipy matplotlib h5py` in the minimal environment?
We should not do that – instead of it let to pull the things which are declared in setup.py by installation script.
Else we are only pretending that we test it on minimal environment. We want to emulate exact environment as the customer with his/her python3 version (be it conda or provided with system), if we decide not to test minimal environment then we can get rid of half of this logic.
And if it is tests which fails, then it is tests which have to be fixed.

.travis.yml
```diff
@@ -2,82 +2,61 @@ language: python
...
 
 after_success:
 - coveralls
 - if [[ $MINIMAL_ENV == 'False' ]]; then
-    if [[ $TRAVIS_OS_NAME == 'osx' ]]; then
-      python setup.py bdist_wheel;
-    else python setup.py sdist;
-    fi;
+    python setup.py bdist_wheel;
+  else python setup.py sdist;
```
sem-geologist, line 55: No!, why we would build sdist from min environment??? it would have no cython, and we will get no C code prepared for installation from source. This line should go away completely, or be inserted somethere else

.travis.yml
```diff
@@ -2,82 +2,61 @@ language: python
...
   fi
 
 before_deploy:
-- if [[ $TRAVIS_OS_NAME == 'osx' ]]; then
-    export DISTRIB=$(ls ./dist/*.whl);
-  else export DISTRIB=$(ls ./dist/*.tar.gz);
-  fi
+- export DISTRIB=$(ls ./dist/*.whl);
```
sem-geologist, line 59: no more sdist?

.travis.yml
```diff
@@ -2,82 +2,61 @@ language: python
...
-  - python: 3.5
-    env: export FAIL_ON_EXTERNAL_DEPRECATION='False'; MINIMAL_ENV='True'
-  - os: osx
+  - env: export PYTHON=3.5; FAIL_ON_EXTERNAL_DEPRECATION='False'; MINIMAL_ENV='False'
+  - env: export PYTHON=3.5; FAIL_ON_EXTERNAL_DEPRECATION='False'; MINIMAL_ENV='False'; MPLBACKEND="agg"
+    os: osx
+    language: generic
+  - env: export PYTHON=3.6; FAIL_ON_EXTERNAL_DEPRECATION='False'; MINIMAL_ENV='False'
+  - env: export PYTHON=3.6; FAIL_ON_EXTERNAL_DEPRECATION='True'; MINIMAL_ENV='False'
+  - env: export PYTHON=3.6; FAIL_ON_EXTERNAL_DEPRECATION='False'; MINIMAL_ENV='True'
```
sem-geologist, line 11: if we use miniconda it kill completely the sense of minimal environment. miniconda is however ('spam', 'spam', 'spam'...) not only 'eggs'

sem-geologist:
by using miniconda for minimal environment we shave few lines however:
1. prolong the minimal env testing 100%; instead with stock python we could test in the same time both min py36 and min py35 (which is still used on most of the stable distros or LTR releases)
2. by not using stock python we no more emulate pip hyperspy installation on brand new system

francisco-dlp:
I think that installing in miniconda is "stricter" than installing in Ubuntu's system Python. Ubuntu comes with more python packages pre-installed than miniconda so, if it installs in miniconda it has to install in Ubuntu. The only issue is that the system pip version may be too old in some systems, but that's not something up to us to fix.

We can test also Py35 using miniconda if we want to (we do it for the standard test), I just thought that it may not be necessary for the min env.

By the way, the min env was an excellent idea!

sem-geologist:
but in this shape as it is it is broken, it is pretending to be minimal.
At least remove `DEPS="pip ipython numpy scipy matplotlib h5py sympy scikit-image";`
 this should not be installed through pip, because then we are no more testing what is minimal environment. Those which is minimal should be declared in setup.

sem-geologist:
also do not create sdist with minimal environment especialy that the artifacts are not collected in current state of the code anymore

francisco-dlp:
I fully agree, maybe we could add an extra command to setup.py to return the requirements so that we can pass them to conda? (We don't want to compile numpy, scipy etc ourselves...)

I modified this in a rush for the HyperSpy 1.2 release and this has already been sort-of merged due to a mistake caused by working from local and remote terminals at the same time: I added the v1.2 tag to a commit in this branch and pushed to upstream thinking that I was in RELEASE_v1.2 :( Anyway, the tests work and we can carry on improving from here.

francisco-dlp:
I'll merge it as it has already been effectively merged.

sem-geologist:
I probably missed something, but you don't need to compile anything anymore, binary wheels are used in linux already too. So pip does not need to recompile stuff.

francisco-dlp:
We have linux wheels in the GH release, but I couldn't upload them to pypi, I got a platform not supported error...

sem-geologist:
Not talking about hyperspy wheels, I am talking abou numpy, scipy, ....
wheels, and that it needs no compiling, and so conda is not needed. Even if
it was compiling, previously minimal env was doing everything in half of
time than now with conda and not compiling

francisco-dlp:
You are right, pypi has wheels for all those requirements for linux. Would
you like to send a PR turning back to installing all from pip? By the way,
I don't know why pypi complains when I try to upload hyperspy's linux
wheels.

sem-geologist:
I see one very obvious thing - scipy, numpy are bdist `manylinux1`, and
hyperspy is `linux`, thats why it is incompatible.

sem-geologist:
It actually is the case, I am trying to find how to bdist  to `manylinux1`,
is there any code which push hyperspy bdist artefacts to pypi?

francisco-dlp:
We push them using twine. It is not done automatically as it relies on a
password.

sem-geologist:
btw there is the  pep 513 dedicated for the problem [https://www.python.org/dev/peps/pep-0513/#the-manylinux1-policy](https://www.python.org/dev/peps/pep-0513/#the-manylinux1-policy)

sem-geologist:
there is auditwheel tool which can be used to create manylinux1 from linux1 (if it is compliant) https://github.com/pypa/auditwheel

PR #1507, opened 2017-03-02
Fail when not specifying number of decomposition components to plot

mwalls:
to stop it exploding when you forget to put a number

### DIFF
--- a/hyperspy/signal.py
+++ b/hyperspy/signal.py
@@ -374,7 +374,7 @@ class MVATools(object):
         else:
             return f
 
-    def _plot_loadings(self, loadings, comp_ids=None, calibrate=True,
+    def _plot_loadings(self, loadings, comp_ids, calibrate=True,
                        same_window=None, comp_label=None,
                        with_factors=False, factors=None,
                        cmap=plt.cm.gray, no_nans=False, per_row=3):
@@ -736,7 +736,7 @@ class MVATools(object):
                     s.save(filename)
 
     def plot_decomposition_factors(self,
-                                   comp_ids=None,
+                                   comp_ids,
                                    calibrate=True,
                                    same_window=None,
                                    comp_label='Decomposition factor',
@@ -859,7 +859,7 @@ class MVATools(object):
                                             per_row=per_row)
 
     def plot_decomposition_loadings(self,
-                                    comp_ids=None,
+                                    comp_ids,
                                     calibrate=True,
                                     same_window=None,
                                     comp_label='Decomposition loading',
--- a/hyperspy/signal.py
+++ b/hyperspy/signal.py
@@ -256,7 +256,7 @@ class MVATools(object):
         ----------
 
         comp_ids : None, int, or list of ints
-            if None, returns maps of 3 components.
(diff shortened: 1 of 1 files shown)


francisco-dlp:
Wouldn't it have been easier to set the number in [plot_decomposition_factors](https://github.com/hyperspy/hyperspy/blob/RELEASE_next_minor/hyperspy/signal.py#L740) and plot_decomposition_loading](https://github.com/hyperspy/hyperspy/blob/RELEASE_next_minor/hyperspy/signal.py#L863)? e.g. 

```python

    def plot_decomposition_loadings(self,
                                    comp_ids=3,
                                    calibrate=True,
                                    same_window=None,
                                    comp_label='Decomposition loading',
                                    with_factors=False,
                                    cmap=plt.cm.gray,
                                    no_nans=False,
                                    per_row=3):
```

mwalls:
Yes of course. As you know, my programming style is to change random lines of code till it does what I want.  But I have two questions:
1) Does your way mean that the default is changed to 3 but if someone (for what reason I don't know) actually enters "None" they will still get all of them?
2) Ok we'll use your method but I have to do it, right? (If I do the pull request I've got to make the changes?). Still a bit new to this...

mwalls:
Also, if anyone has strong feelings about it being something other than 3 components, I'm not going to fight about it.

francisco-dlp:
I am actually wondering if it is not better not to set any default and let the function fail if the number of components is not specified. to do that simply remove the ``None`` from the function keyword argument.

You are not obliged to do it, but, since it's your idea and you've already at it...

The easiest way to proceed is to revert the changes, commit the new version and push e.g.

```bash

git revert 1535cd6

### change the code ###

git commit -a -m "Set default number of components for plotting"
git push

```

ericpre:
+1 on letting fail the function if the number of components is not specified. I don't think that there is a sensible default.
If there is one thing that I remember of my early days of HyperSpy is the frustration just after realising that I forgot to specify the number of components... Now, I am very well vaccinated with that one! :D

mwalls:
Ive just tried to do it with the error message but can't tell if you're all seeing it. Sorry.

hyperspy/signal.py
```diff
@@ -374,7 +374,7 @@ def _plot_factors_or_pchars(self, factors, comp_ids=None,
         else:
             return f
 
-    def _plot_loadings(self, loadings, comp_ids=None, calibrate=True,
```
ericpre, line 377: This was not necessary to change the definition of this function, but it does not have any consequences. It just makes the code a bit less let consistence with the definition of `_plot_factors_or_pchars`.

ericpre:
Yes, this is good and it could be merged: there is just a minor comment, but it is not really important.
The tests failing in travis are not related to this PR.

mwalls:
ok by me. Can't remember if I'm allowed to close one. Will try now.

mwalls:
Looks like I shouldn't have done that.

ericpre:
Yes, please just wait that somebody merges it!

Issue #1506, opened 2017-03-02
Signal scaled extent


magnunor:
When plotting publication figures from HyperSpy signals using Matplotlib, I usually do something like this:

```Python
ax.imshow(
        s.data,
        extent=[
            s.axes_manager[0].low_value,
            s.axes_manager[0].high_value,
            s.axes_manager[1].low_value,
            s.axes_manager[1].high_value])
```
I feel the part defining the extent could be more compact, by for example having something like `s.axes_manager.signal_extent`.

```Python
ax.imshow(s.data, extent=s.axes_manager.signal_extent)
```

magnunor:
Fixed in https://github.com/hyperspy/hyperspy/pull/1508

PR #1508, opened 2017-03-02
Add s.axes_manager.signal_extent and s.axes_manager.navigation_extent

magnunor:
https://github.com/hyperspy/hyperspy/issues/1506

### DIFF
--- a/hyperspy/tests/axes/test_axes_manager.py
+++ b/hyperspy/tests/axes/test_axes_manager.py
@@ -79,6 +79,46 @@ class TestAxesManager:
         assert am2[3].size != am[3].size
 
 
+class TestAxesManagerScaleOffset:
+
+    def test_low_high_value(self):
+        data = arange(11)
+        s = BaseSignal(data)
+        axes = s.axes_manager[0]
+        assert axes.low_value == data[0]
+        assert axes.high_value == data[-1]
+
+    def test_change_scale(self):
+        data = arange(132)
+        s = BaseSignal(data)
+        axes = s.axes_manager[0]
+        scale_value_list = [0.07, 76, 1]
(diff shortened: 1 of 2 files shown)


magnunor:
Should the `axes_manager._update_attributes` be called when changing the scale or offset?

magnunor:
Maybe this should go to RELEASE_next_patch?

hyperspy/tests/axes/test_axes_manager.py
```diff
@@ -75,3 +77,102 @@ def test_update_from(self):
...
+                signal_axis0.low_value, signal_axis0.high_value,
+                signal_axis1.low_value, signal_axis1.high_value,
+                )
+        assert signal_extent == s.axes_manager.signal_extent
+        assert len(s.axes_manager.navigation_extent) == 2
+        nav_axis = s.axes_manager.navigation_axes[0]
+        nav_extent = (nav_axis.low_value, nav_axis.high_value)
+        assert nav_extent == s.axes_manager.navigation_extent
+
+    def test_changing_scale_offset(self):
```
to266, line 196: All looks good, but this last test. It only ever references `high_value` and `low_value` of each axis, and does not actually test that changing scale or offset changes this.

I think it best if instead of this you'd include a general `DataAxis` test that checks if `low_value` and `high_value` are updated appropriately when changing `offset` and `scale`

magnunor: I added some tests for checking this: https://github.com/hyperspy/hyperspy/pull/1508/files#diff-c962fa54b5ae9b1da4d5667347ccd04cR82

francisco-dlp:
@to266, is this ready to merge?

PR #1510, opened 2017-03-04
Finish plot testing

ericpre:
- [x] tidy up the way the matplotlib backend is set
- [x] set the backend to 'agg' for testing on all platforms
- [x] get plot testing working on travis os x
- [x] reduce the memory used by the figure created during plot testing: close/delete the figure more properly
- [x] Fix a test failing in appveyor, because of an unncessary plt.show()
- [x] update dev guide


hyperspy/conftest.py
```diff
@@ -1,12 +1,17 @@
...
 from traits.etsconfig.api import ETSConfig
 ETSConfig.toolkit = "null"
-import matplotlib
-matplotlib.use("Agg")
+
+# pytest-mpl 0.7 already import pyplot this line is called, so setting the 
+# matplotlib backend to 'agg' as early as we can will be useless.
+# However, resetting the rcParams to matplotlib default does the job.
+import matplotlib.pyplot as plt
+plt.rcParams.clear()
```
ericpre: As documented, this bit resets the matplotlib backend to 'agg'. Do we want the tests to be run locally with 'agg'? Or would it better to test other backends, which will be the default in the system? The latter could be good to test more stuff but CI will not be necessarily consistent with the tests run locally and it could be a bit confusing if not aware of this detail.

francisco-dlp: I think that locally we should test using the default mpl backend in the system. Otherwise [this test](https://github.com/hyperspy/hyperspy/blob/RELEASE_next_minor/hyperspy/tests/drawing/test_utils.py#L33) will always be skipped.

ericpre:
I m quite happy with the status of plot testing now, there are a couple of things, which have been significantly improved in this PR.
This is ready to review and once this is merged, I will add a few others tests in another PR.

hyperspy/conftest.py
ericpre: Yes and this tests is actually failing using the `TkAgg` on my system... :(
The error says "AssertionError: Expected 'mock' to be called once. Called 11 times."...

francisco-dlp: That sounds like a bug...

hyperspy/samfire_utils/samfire_worker.py
```diff
@@ -23,9 +23,6 @@
 from queue import Empty
 import dill
 import numpy as np
-import matplotlib
-
-matplotlib.rcParams['backend'] = 'Agg'
```
ericpre: @to266, I removed this when I tidied up the way the matplotlib backend is set. Could you please confirm that there is no particular reason to have this here?

francisco-dlp: @to266 will know better, but I think that it'll be good to keep things as they were in this case because the worker will certainly never need interactive plotting and, as far as I remember, this was preventing a backend related crash.

hyperspy/conftest.py
```diff
@@ -1,12 +1,14 @@
...
-matplotlib.use("Agg")
+
+# pytest-mpl 0.7 already import pyplot, so setting the matplotlib backend to 
+# 'agg' as early as we can is useless for testing.
+import matplotlib.pyplot as plt
 
 import pytest
 import numpy as np
-import matplotlib.pyplot as plt
+import matplotlib
```
francisco-dlp, line 11: Is this needed?

hyperspy/tests/drawing/test_utils.py
```diff
@@ -15,23 +15,15 @@
...
 import hyperspy.drawing.utils as utils
-from hyperspy.misc.test_utils import assert_warns
 
 
-@pytest.mark.skipif(sys.platform == 'darwin',
-                    reason="Plot testing not supported on osx by travis-ci")
-@cleanup
 def test_create_figure():
-    dummy_warning = 'dummy_function have been called after closing windows'
     if matplotlib.get_backend() not in ("GTKAgg", "WXAgg", "TkAgg", "Qt4Agg"):
```
francisco-dlp: Maybe this is not needed anymore?

hyperspy/tests/drawing/test_mpl_testing_setup.py
```diff
@@ -16,14 +16,35 @@
...
-import pytest
 from distutils.version import LooseVersion
+import pytest
+
+from hyperspy.misc.test_utils import check_running_tests_in_CI
 
 
-def test_mlp_agg_for_testing():
-    assert matplotlib.get_backend() == 'agg'
+def test_mlp_agg_for_CI_testing():
```
francisco-dlp, line 25: Why is this test needed? If everything works with something other than the agg backend there is nothing wrong about it...

francisco-dlp (commented):
Just some small comments.

hyperspy/conftest.py
ericpre: What do you mean exactly?

francisco-dlp: importing matplotlib

hyperspy/tests/drawing/test_utils.py
ericpre: I would say, this test is still useful, although this is not tested in CI, we spotted a bug recently with this one!

francisco-dlp: Sorry for not explaining myself: I was questioning if we should not let the test run for all backends since all of them must be supported now that you are using the right mpl event for closing the figure.

hyperspy/tests/drawing/test_mpl_testing_setup.py
ericpre: For CI, some tests will work and other will hold the test.
For example, this was the case with [one test in appveyor](https://ci.appveyor.com/project/hyperspy/hyperspy/build/1.0.2106), which was on hold, because it was waiting for a figure to be closed, or something similar! This happen because there was a `plt.show()` call in one function somewhere.
The question is then: where is the bug? Is it the `plt.show()` call, or is it the test?
By having this test, I was thinking, that it would be easier to find out what is wrong.

hyperspy/samfire_utils/samfire_worker.py
ericpre: Ok, let wait for @to266 answer!

hyperspy/conftest.py
ericpre: It is used in the `mpl_cleanup` pytest fixture, am i missing something obvious?!

hyperspy/tests/drawing/test_utils.py
ericpre: yes!

hyperspy/samfire_utils/samfire_worker.py
ericpre: I reverted this change.

hyperspy/tests/drawing/test_mpl_testing_setup.py
```diff
@@ -16,14 +16,40 @@
...
     # for simplicity, only matplotlib 2.x is supported for testing
     assert LooseVersion(matplotlib.__version__) >= LooseVersion('2.0.0')
+
+
+# Skip if mpl plugin is not called, because it will not failed (no image comparison)
+@pytest.mark.skipif(not pytest.config.getvalue("mpl"),
+                    reason="'mpl' plugin not call.")
+@pytest.mark.xfail(reason="Check if plotting tests are working: if this failed,"
+                   "it means that the image comparison of the plotting test are"
+                   " not working.",
```
to266: Here the `reason` and commented explanations do not match. I assume it's just the wording of the reason that could be improved?

Also, `skipif` reason should have `"...call"` -> `"...called"`

ericpre: yes, the reason is wrong...

thomasaarholt:
At least one other PR (#1382) relies on this. I'm unsure how to test this, so could someone else take a look?

Issue #1502, opened 2017-03-01
Brucker BCF reader fails to load some files


francisco-dlp:
This is the error:

```python

In [1]: import hyperspy.api as hs

In [2]: hs.load("map2-data.bcf")
---------------------------------------------------------------------------
KeyError                                  Traceback (most recent call last)
<ipython-input-2-2cb6879a9d88> in <module>()
----> 1 hs.load("map2-data.bcf")

/home/fjd29/Python/hyperspy3/hyperspy/io.py in load(filenames, signal_type, stack, stack_axis, new_axis_name, lazy, **kwds)
    229             objects = [load_single_file(filename, lazy=lazy,
    230                                         **kwds)
--> 231                        for filename in filenames]
    232 
    233         if preferences.Plot.plot_on_load:

/home/fjd29/Python/hyperspy3/hyperspy/io.py in <listcomp>(.0)
    229             objects = [load_single_file(filename, lazy=lazy,
    230                                         **kwds)
--> 231                        for filename in filenames]
    232 
    233         if preferences.Plot.plot_on_load:

/home/fjd29/Python/hyperspy3/hyperspy/io.py in load_single_file(filename, signal_type, **kwds)
    275                                 reader=reader,
    276                                 signal_type=signal_type,
--> 277                                 **kwds)
    278 
    279 

/home/fjd29/Python/hyperspy3/hyperspy/io.py in load_with_reader(filename, reader, signal_type, **kwds)
    284     lazy = kwds.get('lazy', False)
    285     file_data_list = reader.file_reader(filename,
--> 286                                         **kwds)
    287     objects = []
    288 

/home/fjd29/Python/hyperspy3/hyperspy/io_plugins/bcf.py in file_reader(filename, select_type, index, downsample, cutoff_at_kV, instrument, lazy)
   1105 
   1106     # objectified bcf file:
-> 1107     obj_bcf = BCF_reader(filename)
   1108     if select_type == 'image':
   1109         return bcf_imagery(obj_bcf)

/home/fjd29/Python/hyperspy3/hyperspy/io_plugins/bcf.py in __init__(self, filename)
    801         header_file = self.get_file('EDSDatabase/HeaderData')
    802         header_byte_str = header_file.get_as_BytesIO_string().getvalue()
--> 803         self.header = HyperHeader(header_byte_str)
    804         self.hypermap = {}
    805 

[...]

PR #1512, opened 2017-03-04
Fix-enh bcf v2

sem-geologist:
This PR fixes bugs which were found with loading version 2 bcf #1502.
The first issue were previously some statically getting the xml nodes from the bcf header part,
this patch changed that to use xpath which allows getting the failed nodes independently from the order.
The second issue is the hypercube loading failing (python pure parser) or segfaulting (cython implementation), it appeared that with previously tested files flag 0 never were met and were not implemented, causing the failure (previously only 1, 2, 3 flags were observed)
the 0 flag marked pixels are packed in 16 bit. So in the end it can be said that flag 0 - 16bit packed impulses, 1 - 12bit packed impulses, 2, 3+? - instructively packed arrays.

Finally, v2 bcf contains mapping overview, which would be handy to return in the signal list. The Overview image contains rectangle information of mapping area which allows to reconstruct the marker.

Tasks:

- [x] fix the xml parser
- [x] fix the datacube pure python reading implementation
- [x] fix the datacube cython reading implementation
- [x] add loading of overview image (v2 only)
- [x] add rectangular marker to overview image marking the region of hypermap (v2 only)
- [x] add the docstring of bcf-only parameters to main hspy load function
- [x] do not try reading overview if there is no overview (on v2, thanks to @ericpre )


sem-geologist:
@francisco-dlp , I think WIP flag can be removed, This should work

sem-geologist:
damn... I see I again pushed that to wrong branch...

sem-geologist:
This is ready for review

francisco-dlp:
I get a segfault when loading the dataset.

sem-geologist:
did you removed c files and recythonized? and build_ext --inplace ?

francisco-dlp:
Yes.

francisco-dlp:
Before merging this we also need a smaller file for the test. I'll see if somebody here can provide it.

sem-geologist:
best would be if the image itself (overview) would not exceed 100pixels, in that way the file will be really small

sem-geologist:
I think setup.py and handling recythonization is a bit broken, I see that it does not remove old c file, I also was getting segfault, until I manually removed C file in io_plugins. I will have to deal with this problem in separate PR. @francisco-dlp  if you remove c file manualy and do recythonize, does it still segfaults?

francisco-dlp:
You may be right. After deleting it manually, recythonizing and building in place now it doesn't segfault, but it doesn't use the cython version either :(

sem-geologist:
that is weird... are you sure there is no *.so files produced?

francisco-dlp:
The so file is there, but somehow it fails to import it.

francisco-dlp:
This is the error:
```
ImportError: /home/fjd29/Python/hyperspy3/hyperspy/io_plugins/unbcf_fast.cpython-35m-x86_64-linux-gnu.so: undefined symbol: PyFPE_jbuf
```

But I think that this is might be a particular problem of my installation. So I would say that this is good to merge, but we have to wait for a test file.

sem-geologist:
are you sure you use hyperspy in development mode? maybe it pushes the library to some global folder, so it would have no extension there. only built in place. Try building without `--inplace`, or check if hyperspy when imported is using the same patch where is the source.

sem-geologist:
https://github.com/scikit-learn/scikit-learn/issues/7542

and this is why I don't use (and am against) conda on linux

sem-geologist:
@francisco-dlp , to long to wait for those images, so I added the overview image loading and rectangular marker :) to this PR. Now I need to wait for small file of bcf v2 to add it to test.

sem-geologist:
Good this still not landed to the main. There were some additional bugs found, 1.st the cython implementation is wrong and gives some gibberish for 16bit packed pixels.
2nd. I feel urge to share excerpt from the bcf xml part (pay attention to the 3rd line):
```xml
<ClassInstance Type="TRTSpectrumRegion" Name="Au">
    <TRTBase>
       <Description>Au-MAB</Description>
    </TRTBase>
    <Element>79</Element>
    <Line>MAB</Line>
    <Energy>2.149462777</Energy>
    <Width>1.644548913E-1</Width>
    <ProcWidth>8.7E-1</ProcWidth>
    <Color>16711680</Color>
  </ClassInstance>
```
this `Au-MAB` bugger hit me unexpected. What a shitty, lazy Bruker'ish way to save the space.

How I should fix it?

1. because it is the region of close lines, I could chose to return only one: I could check if line representing string length is >2 then crop it to 2 leaving i.e. MA.
2. check if it is >2 length and generate two lines Ma and Mb.

(to implement 1 is very easy, 2 is more complicated and will make things more cluttered)

sem-geologist:
Everything is fixed now.

@francisco-dlp , I should mention that I was really skeptical about leaving pure python bcf binary parser function as it is ugly and long. But this issue just proved how wrong I was, it was much easier to find the problem and fix it in python parser, and to spot and troubleshoot the new problem introduced into cython code.

sem-geologist:
@francisco-dlp , it is ready for review and merge, v2 bcf was added for tests, failures of travis and appveyor is due to `interpolate` changes in scipy, there are no traces in the travis and appveyor logs of  any bcf failures. The coverage of `io_plugins/bcf.py` with two new bcf samples got up to 93% from 89%.

hyperspy/io.py
```diff
@@ -112,6 +112,23 @@ def load(filenames=None,
         For SEMPER unf- and EMD (Berkley)-files, if True (default is False)
         additional information read during loading is printed for a quick
         overview.
+        
+    downsample : int (1–4095)
+        For Bruker bcf files, if set to integer (>=2) (default 1)
```
francisco-dlp: The language might need some improvement. Would any native speaker like to help (e.g. @tjof2)?

If not, I can help making it more spanglish...

sem-geologist: well... I tried it to sound in the same style as doc "For SEMPER unf- and EMD...." line above

to266:
LGTM. I'd merge if not the @francisco-dlp worry about English

sem-geologist:
It is written in the same style as SEMPER unf- and EMD part in the same docstring, I guess it can be beautified later?

francisco-dlp:
Here is a suggestion:

```
downsample : int (1–4095)
        For Bruker bcf files, if set to integer (>=2) (default 1)
        bcf is parsed into down-sampled size array by given integer factor,
        multiple values from original bcf pixels are summed forming downsampled
        pixel. This allows to improve signal and conserve the memory with the
        cost of lower resolution.
    cutoff_at_kV : {None, int, float}
       For Bruker bcf files, if set to numerical (default is None)
       bcf is parsed into array with depth cutoff at coresponding given energy.
       This allows to conserve the memory, with cutting-off unused spectra's
       tail, or force enlargement of the spectra size.
    select_type: {'spectrum', 'image', None}
       For Bruker bcf files, if one of 'spectrum' or 'image' (default is None)
       the loader returns either only hypermap or only SEM/TEM electron images.
```

francisco-dlp:
By the way, wouldn't ``bin`` be a better name for the parameter as it is binning, not downsampling, what it does? Also, the docstring does not clarify which axes are binned (the spatial axes only?). Also, is there any constraint of the factor (i.e. it must be an integer divisor or the spatial axes size?). And, why not adding the option to bin each axis independently (including the spectral axis) using the same syntax as in ``BaseSignal.rebin``?

sem-geologist:
@francisco-dlp yes it is spatial axes only, why you would want to downsample x and y independently?
There is no implementation in downsampling/binning of signal axe, as that would bring significant penalty in reading speed. Downsample as implemented comes with no cost as it uses the technical nuance of way the bcf is read -- pixel by pixel. Additionally it saves the memory, as pixels are filled into smaller size of array without creating the full size array. It does something like `skimage.measure.block_reduce()` with `sum` as the function, just differently than that it costs no memory and speed. It is useful when bcf mapping is gathered with high spatial resolution (which max is 4096x4096 pixels x >=1024 channels x >=8bit) where a lot of machines with less than 32GB memory would have trouble of opening such file.
Also it is useful for simultanious aquisition of high spatial resolution images (SEI, BSE, etc...) and hypermap (which at that resolution is noisy, but when downsampled is good enough).

I dont know if binning is better term than downsampling?

francisco-dlp:
That makes sense. As I understanding it, binning involves summing the several pixels into a single one while sownsampling involves selecting 1 of every n pixels. Both results in a dataset of the same size but through different methods and outcomes.

What about my question regarding the binning factor?

Regarding the binning axes, the main point was that it is not crystal clear (although it can be inferred) that it only applies to the spatial axes.

By the way, I've just noticed that I've pasted the wrong docstring.

francisco-dlp:
This is the suggested spanglish version:

```python

    bin : int (1–4095)
        For Bruker bcf files, if set to integer (>=2) (default 1)
        the spatial axes are binned by the given integer factor.  For
        example, if `bin` is 3, 3 consecutive pixels from the x-axis in the file are
        summed to obtain each x-axis pixel in the loaded data. Binning enhances the
        signal to noise ratio and decreases the size of the data in memory at
        the expense of spatial resolution.
    cutoff_at_kV : {None, int, float}
       For Bruker bcf files, if a number (default is None)
       the spectral data is cropped to the given `cutoff_at_kV` value in kV
       if below the axis upper limit, of padded up to this value otherwise.
       This is useful to reduce the size of the data in memory when the
       spectral range above a given value is not required.
    select_type: {'spectrum', 'image', None}
       For Bruker bcf files, if 'spectrum'/'image' (default is None)
       the load only multi-dimensional spectra/images.
```

francisco-dlp:
By the way, shouldn't it be keV instead of kV?

sem-geologist:
yes it should be `keV`,  I will need to change it in all the files (same as `downsample` to `bin`.
The factor constrain is > = 1. Anything more than size of spatial axes size will give the same effect as with max spatial size so it is even not checked as the only thing it would  bring the reading speed penalty. So where is practically no upper bound. Nevertheless it is suggested `int (1-4096)` as those are maximum pixel values which can be set in Bruker Esprit.

francisco-dlp:
So, what happens when it is not a divisor of the axis size e.g. binning 3 when the size is 10?

sem-geologist:
and btw, it is both, x and y axes binned:
with `bin=3`: (or >3)
[[3, 3, 3],[3,3,3],[3,3,3]] would be binned to [[27]]
with `bin=2`:
it would be binned to [[12,6],[6,3]];
nothing is done with the border, it is left for user to decide.

francisco-dlp:
@sem-geologist, we're thinking on releasing soon. Would you have time to finish this (it's almost ready except for some minor things and the merge conflicts).

francisco-dlp:
I've fixed the conflicts so merging.

PR #1503, opened 2017-03-01
Update Volume Plasmon

erh3cq:
Gave the dampening energy a more descriptive name.
Also, allowed for introduction of initial parameters during component
creation.

Also, Let me know if I am doing this whole push/pull thing correctly

PR #1513, opened 2017-03-05
Update Volume Plasmon Drude

erh3cq:
Added initial conditions to make component creation cleaner


erh3cq:
I have started a new branch as requested in #1503. @francisco-dlp I have reset my branch RELEASE_next_minor to the same as hyperspy's, branched from that for editing volume plasmon.

Thank you for all the help, and great guide on hyperspy now! [Here](https://github.com/emesene/emesene/wiki/GitHowTo) is another guide that may help for situations like I was in.  I ended up having so many merge/comits in all the wrong places. Resetting made everything run smooth after.  At least we will see...

Hopefully I can start contributing more now!

francisco-dlp:
Thanks!

PR #1504, opened 2017-03-01
Eels lldecon

erh3cq:
Added a function for deconvolution of two low-loss spectra.

PR #1514, opened 2017-03-05
EELS ll fourier deconvolution function

erh3cq:
Added a function that can Fourier ratio deconvolute two low-loss
spectra. ZLP spectra can be provided to avoid the need for ZLP
approximation.

This is helpful for VEELS analysis of surface and bulk plasmons when an interface is involved.


erh3cq:
As mentioned in #1513 I reset my base fork and have followed the correct procedure (I think) to allow for separate pull requests.  So, this should be much cleaner than #1504.  Also, @magnunor this pull already has the fixes from the previous pull.  Let me know if I need to fix anything else on it.

erh3cq:
What do I need to fix for coverage/coveralls to pass?

francisco-dlp:
You need to add tests. However, I would wait until this gets a full review before writing the test.

erh3cq:
Sure thing. I will wait for a full review. I assume the tests are like the lazy test I did for Volume Plasmon before

hyperspy/_signals/eels.py
```diff
@@ -751,6 +751,111 @@ def fourier_ratio_deconvolution(self, ll,
                 self.tmp_parameters.filename +
                 'after_fourier_ratio_deconvolution')
         return cl
+    
+    def fourier_ratio_deconvolution_llspectra(self, kernel, zl=None,
```
thomasaarholt: zl should be zlp, which is the common notation for the zero loss peak in the rest of hyperspy

hyperspy/_signals/eels.py
```diff
@@ -751,6 +751,111 @@ def fourier_ratio_deconvolution(self, ll,
...
+                                        fwhm=None,
+                                        threshold=None):
+        """Performs Fourier-ratio deconvolution of two similar spectrum.
+        
+        Parameters
+        ----------
+        kernel : EELSSpectrum
+            The spectrum to be deconvolved from parent spectrum
+        zl : ZLP spectrum to avoid Gaussian approximation of ZLP
```
thomasaarholt: zl -> zlp

hyperspy/_signals/eels.py
```diff
@@ -751,6 +751,111 @@ def fourier_ratio_deconvolution(self, ll,
...
+        orig_spectrum_size = self.axes_manager.signal_axes[0].size
+    
+        if zl is None:
+            zl_provided = False
+            if threshold is None:
+                threshold = kernel.estimate_elastic_scattering_threshold()
+        else:
+            zl_provided = True
+    
+        spectrum = self.deepcopy()
```
thomasaarholt: @francisco-dlp any suggestions for something other than deepcopy here?

hyperspy/_signals/eels.py
```diff
@@ -751,6 +751,111 @@ def fourier_ratio_deconvolution(self, ll,
...
+        size = kernel_size + spectrum_size - 1
+        # Increase to the closest multiple of two to enhance the FFT
+        # performance
+        size = int(2 ** np.ceil(np.log2(size)))
+        
+        axis = kernel.axes_manager.signal_axes[0]
+        js = np.fft.rfft(spectrum.data, n=size, axis=axis.index_in_array)
+        jk = np.fft.rfft(kernel.data, n=size, axis=axis.index_in_array)
+    
+        if not zl_provided:
```
thomasaarholt: zlp_provided

hyperspy/_signals/eels.py
```diff
@@ -751,6 +751,111 @@ def fourier_ratio_deconvolution(self, ll,
...
+        Wang, Feng, Ray Egerton, and Marek Malac. “Fourier-Ratio Deconvolution
+        Techniques for Electron Energy-Loss Spectroscopy (EELS).”
+        Ultramicroscopy 109, no. 10 (September 2009): 1245–49.
+        doi:10.1016/j.ultramic.2009.05.011.
+        """
+        
+        self._check_signal_dimension_equals_one()
+        orig_spectrum_size = self.axes_manager.signal_axes[0].size
+    
+        if zl is None:
```
thomasaarholt: zlp, same with zl_provided below

hyperspy/_signals/eels.py
```diff
@@ -751,6 +751,111 @@ def fourier_ratio_deconvolution(self, ll,
...
+            
+            I0 = kernel.estimate_elastic_scattering_intensity(threshold=threshold)
+            I0 = I0.data
+            if kernel.axes_manager.navigation_size > 0:
+                I0_shape = list(I0.shape)
+                I0_shape.insert(axis.index_in_array, 1)
+                I0 = I0.reshape(I0_shape)
+            
+            from hyperspy.components1d import Gaussian
+            g = Gaussian()
```
thomasaarholt: define the sigma, A, centre in the Gaussian() function instead of below?

thomasaarholt (commented):
Just typos and a few cleanups really. Just looked through the code, not run it.

erh3cq:
@thomasaarholt thank you. Fixed all of the zl-> zlp changes and the Gaussian().
For deepcopy() : since it returns a new signal at the end, would it be best to replace all of the spectrum variables with self? Then at the end return self. Or would this update the original signal rather than returning a new one?  What we change here can also be applied to `def fourier_ratio_deconvolution()`, may decrease memory usage.

thomasaarholt:
If you replaced all instances of spectrum with self, it would both update the original signal and return it (the former of which I don't think we want to do). If you've copied the same approach as the other deconvolution methods, then that's absolutely fine. I was just thinking of ways to reduce the amount of memory used, say if one opens a very large EELS SI. But I think this is fine.

thomasaarholt:
For completion/clarity, you may want to change the `zl` in `zl = g.function(` to something like zlp, but that's not anything that the user sees, so it's up to you :)

thomasaarholt:
It looks like all the other methods use `deepcopy()`, so that's fine :)

erh3cq:
Your call on the zl to zlp (you are more attune to hyperspy ways).  I kind of liked the zl 1)because it is how the other fourier ratio was set up (so it is consistent), and 2) that section is  approximating the zlp (so zl creates a differentiation between actual and approximated). But, this is all in my head and it has to make since to others.  @thomasaarholt let me know :).

As far as `deepcopy()`, sounds good.  I will try to keep this in mind.  It is what other methods use, but I also agree that if wee can figure out an alternative, then it could help with big files (memory hogs)

thomasaarholt:
Sorry it took me long to respond! The `zlp` and `zl` are fine as they stand! If you could come up with some tests, then that would be great!

erh3cq:
Will do. May be a few days. So for tests, do I just feed a random array and parameters?
I know it works for my spectra.

thomasaarholt:
If you feed a specific array and parameters to it, and then test to make sure the result is what you would be expecting, then that's great. You can probably use any pre-existing tests for the other algorithms.

Testing is really great for helping spot a problem when changing the other parts of hyperspy :) Whilst your code currently works, it might break in the future if we change something else!

thomasaarholt:
Woohoo, well done. Currently you have also added the "fix setup" commit from vidar? Was that on purpose?

erh3cq:
Haha well yes and no. I meant to upload it, because I can't switch to the this branch without his fix. However, I wasn't thinking about this pull automatically updating (oops). I can delete it.

erh3cq:
Why is this failing? It passed the tests on my comp, and my build is up to date.

thomasaarholt:
Don't worry, pretty sure something else is failing.

erh3cq:
@thomasaarholt sounds great. Thanks :smile:

thomasaarholt:
Yes. If you follow appveyor (scroll to the bottom and then keep scrolling), you'll see:

`hyperspy\tests\signal\test_1D_tools.py:230: AssertionError`. That's what is failing.

erh3cq:
Yeah. It says a few errors are occurring, and none of them are associated with files I worked with.  Most of them are for SEM.  I am not sure what the errors mean either....

thomasaarholt:
Did you mean to delete this?

erh3cq:
Yes. I was having issues on GitHub with conflicts between remote/local.
Even after fixing the conflicts it would not allow me to merge. The
conflicts I had for the pull were due to the same conflicts between my
remote and local.  It would not let me update by branches, so I deleted and
re-forked.

PR #1516, opened 2017-03-06
attempt to use ipywidgets progressbar option in prefereces

to266:


francisco-dlp:
Maybe this could make it into Rnp too?

to266:
Rebased

thomasaarholt:
To review this, I sorta need to know what it is supposed to do :) Or at least how this changes from previous behaviour.

to266:
As per the title, this adds "attempt to use ipywidgets progressbar" option in prefereces. If it's ticked, it preserves the current behaviour, if not - it uses the "text" format that is compatible with terminals or if ipywidgets are not installed.

PR #1519, opened 2017-03-07
hanning with lazy signals as well

to266:
To allow easier lazy implementation of https://github.com/hyperspy/hyperspy/pull/1514


erh3cq:
@to266 what do I need to do with this?

to266:
Nothing with this. In #1514, however, you should ideally adapt the code to also run on lazy signals. It includes using lazy Hanning tapering, which this PR implements.

I guess this would mainly mean using `da.fft.*` instead of `np.fft.*` methods when any of the input signals are lazy.

I'm not familiar enough with the use-case to advice in any great detail

PR #1520, opened 2017-03-09
s.add_marker: greatly increase speed of adding many markers

magnunor:
This greatly improves the speed of adding many markers using a list, adding a list of ~2000 markers went from ~5 minutes to ~5 seconds.


thomasaarholt:
How? How did that small code change make it so much faster?

magnunor:
When adding markers, the `add_marker` function makes sure that each marker has a unique name. If the marker name already exist, a number is added to the end of the marker name.

So when adding many (2000+) markers via a list, the function would have to iterate from 0 to whatever number of markers which already had been added. Now it simply starts iterating from the last known number suffix, making it much faster.

PR #1527, opened 2017-03-14
Edge labels when plotting EELS signals

thomasaarholt:
I've added support for plotting EELS edges, working off the code for plotting X-ray lines for EDS signals.

I've modified the code a bit, to make it a bit easier to read, and I've also changed some of the default functionality. For instance, `s.plot(True)` will always try to plot any elements present in `s.metadata.Sample.elements`, because I feel that there will be very few cases where that is not the case. If people feel the opposite is true, then I'm happy to change that.

I've made s.plot(EELS_edges) take either EELS_edges as a boolean argument, as a list of elements, edge families (say "Zr_L") or specific edges (say "Zr_L3"), or as a combination of those. It can also take the argument of a single edge/element as string.

I've decided not to use a method like the EDS `s.add_lines()`, as one tends to work with fewer markers in EELS, and code above allows for easy checking of whether additional lines are correct or not (personally I've never used `s.add_lines()` much)

The user-case I am imagining is one where one has 3-4 elements that are known, whilst there is one unlabelled peak. Then the following makes sense to try:

```python
s = some_EELS_signal
s.add_elements(["Zr", "Fe", "O"])
s.plot("Cr_L3")
```

screen shot 2017-03-14 at 13 51 29

I'm going to add some tests to ensure that the input is taken correctly. Thoughts on the PR?


francisco-dlp:
Wouldn't it make more sense for EELS to have axvlines instead of line segments and, maybe horizontal labels?

thomasaarholt:
I tried horizontal labels on the line segments, and it overlapped too much, but with axvlines it might work as I can specify the positions independently of the edge height. Not sure how to prevent overlap, though. Maybe if I alternate the height of the labels.

thomasaarholt:
What about this? Since EELS signals decay rapidly, I've made edges on the left half of the plot offset from the bottom axis and edges on the right offset from the top. I was trying to find a way to place the axvline under the EELS graph, but `zorder=-1000` doesn't put it under. Any ideas?

screen shot 2017-03-15 at 15 23 11

francisco-dlp:
The left/right divide is a nice idea. However, when using ``enable_plot_components`` the bottom side of the figure may get cluttered. What about offsetting always from the top with the following rule:

* First subshell of "set", e.g. L3, at top position
* Following subshells increase offset e.g. ``L2_offset = L3_offset + k``, ``L1_offset = L2_offset + k``
* Following first "subshell" of set resets the offset, unless closest edge to the left has offset 0, in which case increases offset by k + 2 in order to avoid overlapping with previous and next subshell of edge to the left.


Also, instead of fixing the labels to given coordinates, it would be good to define their position relative to the axes using [matplotlib transforms](http://matplotlib.org/users/transforms_tutorial.html). In this way, when zooming the label will remain visible.

Finally, since @ericpre implemented plot testing, adding unittests would be nice.

thomasaarholt:
I actually tried using transforms first! I converted from axes coordinates (say 0.3) to data coordinates. The problem with that is that 0.3 on the current plot in data units will not be 0.3 when moving through a spectrum image (say, for a thinner region). 

What I've done is to calculate min and max for the data on each spectrum image, and then offset the labels as a fraction of this:
```python
datamax = self.data.max(self.axes_manager.signal_indices_in_array)
datamin = self.data.min(self.axes_manager.signal_indices_in_array)
delta = datamax - datamin
(...)
y_position = datamin + (n*0.1*delta + 0.05*delta)
```
I might try and see if I can plot the edge labels ordered by onset_energy. Then I can avoid the overlap problem by the method you describe above. 

Edit: If you have a better idea involving transforms, then do please suggest it!

thomasaarholt:
Right, I forwent what I suggested above, and did what you suggested instead. There's a problem with edges like the Cr_L1 and Fe_L2 seen in the image below.
I'm tempted to order them by energy and then offset them consecutively instead, by k = 1, 2, 3, 1, 2,3. But that might not look great either.

screen shot 2017-03-16 at 18 09 33

thomasaarholt:
Hm, Cr_L1 isn't correctly placed actually. Will take a quick look.

francisco-dlp:
Looks good! What about:

* Making the label box background transparent. Or, maybe getting rid of the box all together?
* Using matplotlib's colorcyle to using different colors for different elements

ericpre:
I was thinking the same for the colors of the lines and the labels. I guess it would be good also to properly write the subscript instead of having an underscore.

Have you tried to stop the line up to the curve? In case it looks a bit less intrusive and nicer?

For the plotting tests, I guess it would be good to have #1510 merged before adding some more, because at the moment, it may be a bit confusing... So if somebody have time/wants to review it, that would be great! ;)

thomasaarholt:
@ericpre @francisco-dlp I tried using `vertical_line_segment` up to the curve now, both where `y1=None` and `y1=datamin` where `datamin` is an array of the minimum for each pixel. For both cases the `ylim` of the image is expanded at the bottom like in the first very image in this PR.
Using it from the top and down to the curve is fine. :/

thomasaarholt:
This is what it looks like with `edge = markers.vertical_line_segment(x=line_energy[i], y1=27000, y2=intensity[i])`:
screen shot 2017-03-16 at 18 50 40

This is immediately after calling s.plot(True). When calling it using `vertical_line` `ylim` is taken from the minimum of the curve to its maximum, and there is no unnecessary whitespace under the curve.

Bug?

thomasaarholt:
I wonder if it somehow could be to do with line 73 in `vertical_line_segment.py`, where it gets initialised to 0, 0, 1. Doesn't explain why ylim[0] goes below zero though.
`self.marker = self.ax.vlines(0, 0, 1, **self.marker_properties)`

francisco-dlp:
From top down (like a flag) sounds good!

thomasaarholt:
Top-down, coming right up.

thomasaarholt:
Closing this, since #2418 is merged. Well done @ptim0626 :)

PR #1528, opened 2017-03-14
Added desktop.ini to .gitignore

jan-car:


jan-car:
Sorry, I did not mean to push that. I use icons for my project folders and Git kept bugging me that the configuration file was not versioned :-P...

PR #1529, opened 2017-03-15
create_model fails with some lazy signals

francisco-dlp:
This PR includes fixes for the following two bugs:

* Most lazy signals were not instances of their standard counterparts. This was breaking model creation from lazy signals.
* ``hs.signals`` included lazy signals.

It also lazifies selected model tests.

### DIFF
--- a/hyperspy/io_plugins/edax.py
+++ b/hyperspy/io_plugins/edax.py
@@ -869,7 +869,7 @@ def spd_reader(filename,
         if nav_units not in [None, 'um']:
             _logger.warning("Did not understand nav_units input \"{}\". "
                             "Defaulting to microns.\n".format(nav_units))
-        nav_units = '$\mu m$'
+        nav_units = r'$\mu m$'
 
     # Create navigation axes dictionaries:
     x_axis = {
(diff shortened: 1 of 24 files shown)


hyperspy/io_plugins/edax.py
```diff
@@ -869,7 +869,7 @@ def spd_reader(filename,
         if nav_units not in [None, 'um']:
             _logger.warning("Did not understand nav_units input \"{}\". "
                             "Defaulting to microns.\n".format(nav_units))
-        nav_units = '$\mu m$'
+        nav_units = r'$\mu m$'
```
sem-geologist, line 872: Why we simply can't use unicode? It is python3 in the end.

sem-geologist (commented):
@francisco-dlp , you probably didnot noticed my message on gitter, the failures in travis are caused by new numpy version

hyperspy/io_plugins/edax.py
francisco-dlp: That's a good point, but maybe something to fix in a different PR. Here I was just fixing a py36 warning.

PR #1534, opened 2017-03-16
roi.interactive is not honouring the axes kwarg 

francisco-dlp:
The argument is passed to the internal ``interactive`` call but not to ``add_widget``. This PR fixes it.


vidartf:
Note that the docstring is now not matching the actual behavior (the docstring says that the `kwargs` are passed to the function, not to `add_widget`).

PR #1538, opened 2017-03-17
Fix interpolate test

francisco-dlp:
The ``interpolate_in_between`` test is failing with SciPy 0.19. I couldn't find the reason in the scipy changes list. Nevertheless, this PR increases the tolerance so that the test passes with scipy <=0.19.


to266:
This does the model component name thing as well (one commit too many, technically)

francisco-dlp:
Sorry for that. As the other one has already been merged, I guess that it doesn't matter...

PR #1539, opened 2017-03-17
Fix restore edx model

### LINKED ISSUE #1526: Restore a model without all x ray lines (EDS)
Hello, 
i think an issue occur when i try to create a model with only few x ray lines (not all that are contained in my metadata and not all the family lines), the multifit give good results. So i would like to save and restore my model (big data). But the problem occur here, restore the model is impossible, hyperspy try to load parameters (from dictionary i guess) for all xray lines even the x ray lines removed (from components before the multifit) previously. And for xray lines that have been removed hyperspy doesn't find, of course, any parameters.
I try to save the dictionary after the fit and to load the model with this new one but the same error occur. 
So maybe when i try to restore my model a new one is created with the argument auto_add_lines=True without option to change that. I don't know. 
And thank you for your work !

francisco-dlp:
Fixes #1526

### DIFF
--- a/hyperspy/models/edsmodel.py
+++ b/hyperspy/models/edsmodel.py
@@ -866,3 +866,16 @@ class EDSModel(Model1D):
         if plot_result and img.axes_manager.signal_dimension != 0:
             utils.plot.plot_signals(intensities, **kwargs)
         return intensities
+
+    def remove(self, thing):
+        thing = self._get_component(thing)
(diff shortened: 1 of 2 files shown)


thomasaarholt:
This PR affects a lot of files. Is that correct?

francisco-dlp:
Now it looks better (it was pointing to the wrong branch). Thanks for spotting it!

PR #1540, opened 2017-03-20
Multiple stacking bugfixes

to266:
- A bug was present where stacking of signals with masked data was not possible due to dask not having masked arrays itself.
 - stacking with non-default axis was neither tested properly nor worked, now does.

Now we replace masked elements with `np.nan` (as we did for the circular ROI in particular) and are able to lazify and stack such signals.

The behaviour is different from the default numpy, though, which may have been an older bug in numpy/our stacking:

Previously, stacking two arrays resulted in the mask disappearing (this is also how our stacking behaved):
```python
>>> ar = np.ma.masked_array(data=[1,2,3], mask=[0,1,0])
>>> ar
masked_array(data = [1 -- 3],
             mask = [False  True False],
       fill_value = 999999)
>>> np.stack([ar,ar])
masked_array(data =
 [[1 2 3]
 [1 2 3]],
             mask =
 False,
       fill_value = 999999)
```

Now, effectively, this happens:
```python
>>> np.stack([np.where(x.mask, np.nan, x) for x in [ar, ar]])
array([[  1.,  nan,   3.],
       [  1.,  nan,   3.]])
>>> res = hs.stack([hs.signals.BaseSignal(x) for x in [ar, ar]])
>>> res
<Signal1D, title: Stack of , dimensions: (2|3)>
>>> res.data
array([[  1.,  nan,   3.],
       [  1.,  nan,   3.]])
```
Which actually preserves the mask / gets rid of the data that was masked.

### DIFF
--- a/hyperspy/tests/utils/test_stack.py
+++ b/hyperspy/tests/utils/test_stack.py
@@ -41,17 +41,23 @@ class TestUtilsStack:
 
     def test_stack_not_default(self):
         s = self.signal
-        s1 = s.deepcopy() + 1
-        s2 = s.deepcopy() * 4
+        s1 = s.inav[:, :-1] + 1
+        s2 = s.inav[:, ::2] * 4
         result_signal = utils.stack([s, s1, s2], axis=1)
         axis_size = s.axes_manager[1].size
+        axs1 = s1.axes_manager[1].size
+        axs2 = s2.axes_manager[1].size
         result_list = result_signal.split()
         assert len(result_list) == 3
-        np.testing.assert_array_almost_equal(
-            result_list[0].data, result_signal.inav[:, :axis_size].data)
-        result_signal = utils.stack([s, s1, s2], axis='y')
-        np.testing.assert_array_almost_equal(
-            result_list[0].data, result_signal.inav[:, :axis_size].data)
+        for rs in [result_signal, utils.stack([s, s1, s2], axis='y')]:
+            np.testing.assert_array_almost_equal(
+                result_list[0].data, rs.inav[:, :axis_size].data)
+            np.testing.assert_array_almost_equal(
+                s.data, rs.inav[:, :axis_size].data)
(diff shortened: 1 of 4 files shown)


hyperspy/tests/utils/test_stack.py
```diff
@@ -41,17 +41,23 @@ def test_stack_of_stack(self):
...
+        axs2 = s2.axes_manager[1].size
         result_list = result_signal.split()
         assert len(result_list) == 3
-        np.testing.assert_array_almost_equal(
-            result_list[0].data, result_signal.inav[:, :axis_size].data)
-        result_signal = utils.stack([s, s1, s2], axis='y')
-        np.testing.assert_array_almost_equal(
-            result_list[0].data, result_signal.inav[:, :axis_size].data)
+        for rs in [result_signal, utils.stack([s, s1, s2], axis='y')]:
+            np.testing.assert_array_almost_equal(
```
thomasaarholt, line 53: Speaking as a relative testing noob here, but is this a good way to perform multiple tests? If the test fails, is it clear which of the expressions below that fail?

thomasaarholt:
Code looks good, aside from my single comment.

hyperspy/tests/utils/test_stack.py
to266: In this particular case - I think so. Both stacks should be identical, and the multiple asserts just check that each of the stacked pieces are actually inside (so just testing the same function, but different parts of its result)

thomasaarholt: Cool, that's fine then! 👍

ialxn:
This fix solves my problem with stacking as reported on the mailing list. So fine with me.

thomasaarholt:
This seems to work nicely. I suggest we add some more documentation as well, but for now we can perhaps merge this, so as to move forward towards v1.3?

PR #1544, opened 2017-03-21
IPywidgets UI

francisco-dlp:
In this PR

* New IPywidgets User Interface
* New UI for ROIs
* traitsui and ipywidgets are now optional
* New register mechanism to register user interface elements
* ipywidgets are fully convered by unittests
* Multiple UI issues fixed.
* Creating an EELSModel with a spectrum without the required parameters no longer raises a GUI element
* Deprecate ``AxesManager.show`` and ``Signal1D.integrate_in_range``.
* All functions that take a signal range can now take a ``roi.SpanRoi``.
* Fix the nbagg backend sluggishness.
* Improveme interactive plotting speed for all mpl backends.
* All functions that take a signal range argument now accept a ``SpanRoi``.
* New signal slicing using ROIs
* New ``hs.link_directional`` and ``hs.link_bidirectional`` functions to link traits, traitlets and traits to traitlets.
* Split the GUI code into two separate packages: https://github.com/hyperspy/hyperspy_gui_traitsui and https://github.com/hyperspy/hyperspy_gui_ipywidget

# To test it

1. Install ipywidgets and enable them:

```bash
jupyter nbextension enable --py --sys-prefix widgetsnbextension
```

2. Install https://github.com/hyperspy/hyperspy_gui_traitsui and/or https://github.com/hyperspy/hyperspy_gui_ipywidgets and  https://github.com/hyperspy/link_traits .

3. Then start hyperspy as usual using **any** matplotlib backend except the inline backend. This PR makes the nbagg backend usable. To use try it:

```python

%matplotlib nbagg
import hyperspy.api as hs

```

# TODO

## Widgets
* [x] Add ROI gui doc
* [x] Fix busy nbagg again
* [x] preferences
* [x] Spikes removal tool
* [x] Signal navigation sliders
* [x]  Axes manager gui
* [x] Data axis gui
* [x] Model signal range
* Smoothers
    * [x] Savitzky Golay
    * [x] Lowess
    * [x] TV
* [x] Background removal
* [x] Fit component
* [x] Crop
* [x] Signal calibration: s.calibrate()
* [x] SEM/TEM parameter: s.set_microscope_parameters()
*  Load: hs.load() [No equivalent widget -> not implementing]
* [x] ImageContrastEditor: press h on an image, I actually never used this one...
*  Integrade signal: s.integrate_in_range() (enable use ROI instead)
* [x] Add progress bar to spikesremovaltool
* [x] set value slider step
* [x] Add GUI to set_microscope_parameters
* [x] Add non-continuous update
* [x] Use ``gui`` instead of ``get_gui`` where possible
* [x] Disable left/right traitsui
* [x] Add ROIs widgets
* [x] Add display to methods that call gui internally
* [x] Fix EDS missing parameters message

Deprecate:
* [x] AxesManager.show
* [x] Integrate in range
* [x] Crop
## Other tasks
* [x] Fix setting widget value out of axis bounds
* [x] Fix remove_background error out of axis limits
* [x] Add link and dlink to api
* [x] Add transform to link
* [x] All gui starts by gui
* [x] Remove interactive as it is no longer necessary
* [x] slice signal with SpanRoi and Rectangle2D
* [x] Tests
    * [x] Axes
    * [x] micro parameters
    * [x] model
    * [x] tools
    * [x] preferences
    * [x] roi
* [x] Fix existing non bidirectional widgets
* [x] Fix traitsui color still breaking things
* [x] Add toolkit option
* [x] Add preferences option to disable individual toolkits
* [x] Unlink on close
* [x] Add register
* [x] Split traitsui into a separate package
* [x] Fix long labels
* signal_range with ROIs
    * [x] Implementation
    * [x] Documentation
    * [x] Test
* [x] Documentation
* Manual testing
    * Preferences
        * [x] ipywidgets
        * [x] traitsui
    * AxesManager
        * [x] ipywidgets
        * [x] traitsui
    * Sliders
        * [x] ipywidgets
        * [x] traitsui
    * Smooth lowess
        * [x] ipywidgets
        * [x] traitsui
    * Smooth savitzky
        * [x] ipywidgets
        * [x] traitsui
    * Smooth TV
        * [x] ipywidgets
        * [x] traitsui
    * Background
        * [x] ipywidgets
        * [x] traitsui
    * Spikes removal tool
        * [x] ipywidgets
        * [x] traitsui
    * Calibrate
        * [x] ipywidgets
        * [x] traitsui
    * Image contrast
        * [x] ipywidgets
        * [x] traitsui
    * Model
        * [x] ipywidgets
    * Fit component
        * [x] ipywidgets
        * [x] traitsui
    * Parameter
        * [x] ipywidgets
    * Component
        * [x] ipywidgets
    * EELSCL
        * [x] ipywidgets
    * Scalable Fixed
        * [x] ipywidgets
    * Interactive range
        * [x] ipywidgets
        * [x] traitsui
    * Load
        * [x] traitsui
    * EELS parameters
        * [x] ipywidgets
        * [x] traitsui
    * EDS TEM parameter
        * [x] ipywidgets
        * [x] traitsui
    * EDS SEM parameter
        * [x] ipywidgets
        * [x] traitsui
    * Integrate in rage
        * [x] traitsui

(am I missing something else?)


ericpre:
A few more:
- Signal calibration: s.calibrate()
- SEM/TEM parameter: s.set_microscope_parameters()
- Load: hs.load()
- ImageContrastEditor: press h on an image, I actually never used this one...
- Integrade signal: s.integrate_in_range()

If you could please add a bit of documentation to the dev guide, this will help us to give a hand with this! ;)

francisco-dlp:
Thanks!

magnunor:
I'm not very familiar with IPywidgets: can they work outside a notebook? In other words: without a web browser?

francisco-dlp:
No, they're HTML widgets. The nice things is that we're not replacing/removing traitsui, we are adding a "jupyter notebook interface".

jat255:
This is super cool. Is it expected behavior right now that the traits gui also opens when you call the methods?

francisco-dlp:
This is WIP so we still have to decide those sort of things. What I would suggest is:

* traitsui becomes optional
* ipywidgets mandatory (?)
* Whether traitsui is used or not is decided in preferences (?)
* Whether ipywidgets is used or not is decided in preferences (?)

Opinions more than welcome!

hyperspy/gui/__init__.py
```diff
@@ -38,7 +38,7 @@ def set_ets_toolkit(toolkit):
 _logger.debug('Current MPL backend: %s', backend)
 if "WX" in backend:
     set_ets_toolkit("wx")
-elif "Qt" in backend:
+elif "Qt4" in backend:
```
ericpre: Is it fixing #1559? If so, does it mean that instead of using `%matplotlib qt`, we should recommend using `%matplotlib qt4`, before importing hyperspy?

francisco-dlp: No, it's not fixing #1559. This fixes using that qt5 backend albeit without the traitsui gui as it does not yet support qt5.

The ipywidgets give us a lot more flexibility when choosing the mpl backend as the ipywidgets are obviously backend independent. Therefore this PR should make hspy usable with e.g. tk or nbagg.

ericpre: Ok. Forgot my previous comment, it does not make sense...

francisco-dlp:
This is ready for review and beta testing.

I will add tests and documentation over the next few days but I would like to get feedback at this stage in case some major changes are required.

To try it just install ipywidgets, enable them (``jupyter nbextension enable --py widgetsnbextension --sys-prefix``) and run hyperspy in the jupyter notebook.

ericpre:
With pyqt5

```python
data = np.arange(0, 100, 1).reshape(2, 5, 10)
s = hs.signals.Signal1D(data)
s.plot(navigator='slider')

NotImplementedError                       Traceback (most recent call last)
<ipython-input-4-81e26e0d9a7f> in <module>()
----> 1 s.plot(navigator='slider')

/home/eric/Python_dev/hyperspy/hyperspy/signal.py in plot(self, navigator, axes_manager, plot_markers, **kwargs)
   1968                     " \"slider\", None, a Signal instance")
   1969 
-> 1970         self._plot.plot(**kwargs)
   1971         self.events.data_changed.connect(self.update_plot, [])
   1972         if self._plot.signal_plot:

/home/eric/Python_dev/hyperspy/hyperspy/drawing/mpl_he.py in plot(self, **kwargs)
    152                 self.pointer.color = 'red'
    153                 self.pointer.connect_navigate()
--> 154             self.plot_navigator(**kwargs.pop('navigator_kwds', {}))
    155         self.plot_signal(**kwargs)
    156 

/home/eric/Python_dev/hyperspy/hyperspy/drawing/mpl_he.py in plot_navigator(self, colorbar, scalebar, scalebar_color, axes_ticks, saturated_pixels, vmin, vmax, no_nans, centre_colormap, title, **kwds)
     67             get_gui(self=self.axes_manager.navigation_axes,
     68                     toolkey="navigation_sliders",
---> 69                     title=self.signal_title + " navigation sliders")
     70             return
     71         title = title or self.signal_title + " Navigator" if self.signal_title else ""

/home/eric/Python_dev/hyperspy/hyperspy/ui_registry.py in get_gui(self, toolkey, display, toolkit, **kwargs)
     97         if toolkit in toolkits:
     98             used_toolkits.add(toolkit)
---> 99             thisw = f(obj=self, display=display, **kwargs)
    100             if not display:
    101                 widgets[toolkit] = thisw

/home/eric/Python_dev/hyperspy/hyperspy/gui_traitsui/utils.py in wrapper(*args, **kwargs)
     14         obj, kwargs = f(*args, **kwargs)
     15         if display:
---> 16             obj.edit_traits(**kwargs)
     17         else:
     18             return obj

/opt/anaconda3/lib/python3.6/site-packages/traits/has_traits.py in edit_traits(self, view, parent, kind, context, handler, id, scrollable, **args)
   1868 
   1869         return view.ui( context, parent, kind, self.trait_view_elements(),
-> 1870                         handler, id, scrollable, args )
   1871 
   1872     #---------------------------------------------------------------------------

/opt/anaconda3/lib/python3.6/site-packages/traitsui/view.py in ui(self, context, parent, kind, view_elements, handler, id, scrollable, args)
    431             kind = self.kind
    432 
--> 433         ui.ui( parent, kind )
    434 
    435         return ui

/opt/anaconda3/lib/python3.6/site-packages/traitsui/ui.py in ui(self, parent, kind)
    219                                    dispatch = 'ui' )
    220         self.rebuild = getattr( toolkit(), 'ui_' + kind )
--> 221         self.rebuild( self, parent )
    222 
    223     #---------------------------------------------------------------------------

/opt/anaconda3/lib/python3.6/site-packages/traitsui/toolkit.py in ui_live(self, ui, parent)
    218             interface using information from the specified UI object.
    219         """
--> 220         raise NotImplementedError
    221 
    222     def ui_modal ( self, ui, parent ):

NotImplementedError: 
```

Replace this error with a warning and keep executing the rest of the code?

Edit: Fix done!

francisco-dlp:
Yes, thanks!

ericpre:
Regarding the choice of the ui: currently we select the ui that we want to use in the preference. Would it be worth, to replace this by a dropdown list to choose the 'preferred' ui, when both can be used at the same time? In case the selected ui can't used (ipywidget in the qtconsole or traitsui with qt5), fallback to the other one?

francisco-dlp:
It should be fixed now. 

Regarding the choice of toolkit in preferences I think that what you propose cannot be implemented because the ipython kernel does not know which interface you're using to interact with it. With the current implementation.

ericpre:
Ah yes, we can't know if we are running in the qtconsole or the notebook... I will try to see if there is sensible workaround. Alternatively, it could be solved by leaving it as it is and recommending the users to add a line at the beginning of a notebook to disable the unwanted ui.

francisco-dlp:
Documenting it explicitly is definitely a good idea. Now both traitui and ipywidgets are optional dependencies, so the issue will only arise for users who decide installing both.

francisco-dlp:
For v2 we could consider splitting the ``gui_*`` into separate packages, which should make installation more straightforward: for example, installing the hyperspy_gui_traitsui package could downgrade pyqt if still needed while there is no way to do that with the current monolithic approach.

ericpre:
> Documenting it explicitly is definitely a good idea. Now both traitui and ipywidgets are optional dependencies, so the issue will only arise for users who decide installing both.

Yes, good point! :)
For the next release, the windows installer bundle will have both ui installed?

francisco-dlp:
Yes, it will because we distribute it with traitsui.

francisco-dlp:
Sorry, I meant HyperSpyUI.

thomasaarholt:
Minor things:
**Spikes removal**
- Add some newlines to the help text, preferably before each numbered line.
- Start with Advanced and Help closed rather than open (it's quite large in the browser).
- No "completion" message when one has gone through the whole signal. Unsure what the best implementation is. 

**Preferences**
- Add a close button to the window, like Spikes removal.

**Signal navigation sliders (´s.plot(navigator="slider")´)**
- If possible, after the navigation slider, add two boxes, Index and Value, showing their value (pixel number and scaled value) instead of two sliders.

thomasaarholt:
Bugs:
[Spikes removal going past axis limits of signal.](https://gist.github.com/thomasaarholt/fd7df9a69751fafbdaeafb268381514e) - I can't seem to replicate it. Low priority for now.
[Pressing Remove at the start raises a TypeError. (Maybe an older bug)](https://gist.github.com/thomasaarholt/88652ac74ca4eaab7c52c9edac842f95)

francisco-dlp:
@thomasaarholt, thanks for your suggestions. I have implemented the changes that you propose in the in nav sliders and fixed the spikes removal typeerror bug. @ericpre has fixed the spikes removal help text layout.

I haven't added a close button to the ipywidgets preferences because there is already a close cross at the top left.

I agree that starting the spikes removal tool with the accordions closed is better, but I couldn't find how to do it (does anybody know)?

There is no completion message because it feels weird in the notebook. There is a progress bar instead.

thomasaarholt:
Any particular reason for including 14 decimals on the value of the slider? :)

francisco-dlp:
Ask the ipywidgets devs, it's their default. There is room for it, so why not?

francisco-dlp:
I can't actually reproduce the 14 decimals issue. It seems to be smart enough to crop the digits most of the time, but not always probably due to rounding error.

to266:
Oh, I remember ipywidgets sliders having issues with very small values. Is the behaviour/bug still there? In essence there was a relatively large (of the order of `1e-5`?) smallest step that the sliders could manage. If it's still there, we might want to think about patching it with offset/scale similarly like we do in `DataAxis`.

francisco-dlp:
Yes, it seems to be still there.

francisco-dlp:
They've moved https://github.com/jupyter-widgets/ipywidgets/issues/259 to their backlog milestone :(

francisco-dlp:
This is (finally!) ready for review.

thomasaarholt:
Trying to figure out the smoothest way of checking this out.

Since I'm missing the hyperspy_gui_ipywidgets package, just checking out the new PR isn't enough. I tried cloning the hyperspy_gui_ipywidgets repo and installing via `pip install -e ./`, but then it complains that my version of hyperspy isn't compatible (#1597).

For now I'm just locally manually editing the setup.py to remove the hyperspy requirement.

thomasaarholt:
Okay, got it installed. 

In my opinion, in `hs.preferences.gui()`, the choice for which UI to use should be a list, not two chechboxes. What if a user accidentally turned them both off? For me (who is updating), both UIs were on to start with, which made it a bit confusing.

In any case, the options are not next two one another, which made me think that I couldn't turn the traitsUI one off. 

screenshot 2017-05-17 22 11 05

thomasaarholt:
In the instructions of this PR, add that the hyperspy/link_traits package needs to be installed.

thomasaarholt:
Nitpicking:
**Spikes Removal**
Remove the newline (turned into a `<br>` in the html code), before "To remove spikes".
screen shot 2017-05-17 at 22 40 29

In general we might want to reduce css padding to reduce whitespace in the UI.

thomasaarholt:
On Windows, running %matplotlib nbagg and importing hyperspy yields the following error. I'm understand why the warning is being raised, but not why the error occurs.:

```python
ModuleNotFoundError                       Traceback (most recent call last)
c:\users\thomasaar\hyperspy\hyperspy\api.py in <module>()
     13     # Register traitui UI elements by importing the module
---> 14     import hyperspy_gui_traitsui
     15 except ImportError:

ModuleNotFoundError: No module named 'hyperspy_gui_traitsui'

<ipython-input-1-8f77aec737e5> in <module>()
      1 get_ipython().magic('matplotlib nbagg')
----> 2 import hyperspy.api as hs

c:\users\thomasaar\hyperspy\hyperspy\api.py in <module>()
     14     import hyperspy_gui_traitsui
     15 except ImportError:
---> 16     _logger.warning(
     17         "The traitsui GUI elements are noot available, probably because the "
     18         "hyperspy_gui_traitui package is not installed.")

NameError: name '_logger' is not defined
```

**Correction**: There is some bug with trying to run this without the hyperspy traits gui package installed. The error does not show after installing it.

It would be nice if hyperspy automatically realised that the traits or ipywidgets backend was not available and defaulted to the available one.

francisco-dlp:
That's a bug, thanks!

francisco-dlp:
Fixed, thanks!

thomasaarholt:
I've got a bunch of small stuff to upload when I have Internet. Turns out they don't have Internet on ferries in Norway. 

Take a look at m.set_signal_range. When I select range and press apply, I get an "obj" error. That's crept up twice so far.

magnunor:
Some minor points:
- The images should be possible to display horizontally (I think Thomas mentioned this somewhere), since having them vertically means it can be hard to show a signal and navigation window at the same time.
- The exception when using `hs.load()` should say that this functionality is not implemented for this GUI, currently it says: `ImportError: No toolkit installed. Install ipywidgets or traitsui to enable GUI elements.`

thomasaarholt:
I'm working my way down the list of widgets in the top post on this PR:

**`m.set_signal_range`**
The final sentence in `m.set_signal_range`' docstring is now incorrect, as calling it without parameters no longer resets it, but calls it interactively.

In addition, selecting a range with it and pressing Apply raises an error:
```python
UnboundLocalError                         Traceback (most recent call last)
C:\ProgramData\Miniconda3\lib\site-packages\hyperspy_gui_ipywidgets\tools.py in on_apply_clicked(b)
     43 
     44     def on_apply_clicked(b):
---> 45         obj = obj
     46         if obj.ss_left_value != obj.ss_right_value:
     47             obj.span_selector_switch(False)

UnboundLocalError: local variable 'obj' referenced before assignment
```

**`Fit_component`**
The text isn't fully displayed:
`Click on the signal figure and drag to the right to s....`

**`s.crop_signal1D()`**
Same error as the one from `m.set_signal_range` with the "obj".

**`s.calibrate()`**
Perhaps separate issue, but it should raise an error if the user doesn't update the "New Left" and "New Right" values before pressing apply. I noticed it because I was being stupid.

**`hs.load()`**
Explain why it doesn't work in the error message. Atm it suggests installing traitsui or ipywidgets.

hyperspy/_signals/signal1d.py
```diff
@@ -902,54 +906,97 @@ def smooth_lowess(self,
                      show_progressbar=show_progressbar,
                      ragged=False,
                      parallel=parallel)
+    smooth_lowess.__doc__ = \
+        """Lowess data smoothing in place.
```
vidartf: I think the docstring would look nicer if it was indented to an appropriate level. Common indenting should be automatically removed as per [PEP257](https://www.python.org/dev/peps/pep-0257/#handling-docstring-indentation).

francisco-dlp: Good point, thanks!

francisco-dlp:
Thanks @magnunor and @thomasaarholt!. Regarding displaying the figures side by side will be a lot easier with mpl 2.1 where the nbagg figures are going to be ipywidgets, so I suggest to leave it until mpl 2.1 is released.

thomasaarholt:
@francisco-dlp, where did you learn that nbagg figures will be ipywidgets?

francisco-dlp:
https://github.com/matplotlib/matplotlib/pull/5754

thomasaarholt:
In your first post here, by "Add ROIs widgets" are you referring to `roi.gui()`?

thomasaarholt:
`ImageContrastEditor`
Not working on nbagg. Pressing h does nothing. Otherwise works fine.

That's all I could test manually I think.

francisco-dlp:
Yes, all the ROIs have now a ``gui`` method e.g.:

```python

sr = hs.roi.SpanROI(10, 20)
sr.gui()

```

I cannot reproduce the issue with ``ImageContrastEditor`` in Linux and Windows using Chrome.

thomasaarholt:
I'm also in Chrome, Windows.  You are able to change the contrast? Strange.

vidartf:
Will MPL-widgets work in nbagg? 😕

francisco-dlp:
@thomasaarholt, you're right: I was speaking from memory as I remember that we've discussed this before and at the time it was working for me. Now it doesn't work for me either :(

francisco-dlp:
@thomasaarholt, @magnunor  and @vidartf, thank you for reporting the issues. I think that I have fixed them all except for the image contrast editor that, I suspect, is a mpl issue that will get solved in mpl 2.1 when the canvas becomes an ipywidget. Is anybody against merging this?

francisco-dlp:
Thanks @ericpre, @thomasaarholt  and @vidartf for reviewing and beta testing this.

carmenmakepeace:
I have tried my best to follow the steps as clearly as possible but I am still getting 

/anaconda3/lib/python3.6/site-packages/h5py/__init__.py:36: FutureWarning: Conversion of the second argument of issubdtype from `float` to `np.floating` is deprecated. In future, it will be treated as `np.float64 == np.dtype(float).type`.
  from ._conv import register_converters as _register_converters
WARNING:hyperspy.api:The traitsui GUI elements are not available, probably because the hyperspy_gui_traitui package is not installed.

after trying to 
%matplotlib qt
import hyperspy.api as hs 
import numpy as np


I downloaded the github files onto my mac os and then did the pip installs of the hyperspy-gui-traitsui and the ipywidgets

I have installed hyperspy on windows in the past and the process seems a lot simpler. Any help would be greatly appreciated. I am to some extent still very new to hyperspy so I really dont mind even what seems like an obvious answer.

francisco-dlp:
@carmenmakepeace, this is an old pull request that has been merged. For installing the latest HyperSpy and hyperspy_gui_ipywidgets you can follow the instruction in [the User Guide](http://hyperspy.org/hyperspy-doc/current/user_guide/install.html#quick-instructions-to-install-hyperspy-using-anaconda-linux-macos-windows) and the [hyperspy_gui_ipywidgets repository](https://github.com/hyperspy/hyperspy_gui_ipywidgets). Due to recent updates to external libraries, things may not go as smoothly as they should. If you experience any issue you are welcome to ask for help in [hyperspy's gitter chat](https://gitter.im/hyperspy/hyperspy).

PR #1546, opened 2017-03-22
Model print current values

### LINKED ISSUE #1543: micro bug fix for model.print_current_values()
At present `model.print_current_values()` only prints components that have no `__iter__` attribute.

`import hyperspy.api as hs`
`s = hs.load('wobbly_peak.hdf5')`
`sin = hs.model.components1D.Expression('A*sin(b*x + c)', name='sin', A=100, b=0.1, c=10)`
`gaus = hs.model.components1D.Gaussian(centre=500)`
`poly = hs.model.components1D.Polynomial(2)`
`m = s.create_model()`
`m.extend([sin, gaus, poly])`
`m.fit()`
`m.print_current_values()`

Output before patch:
Note, no output for Polynomial.

`Components      Parameter       Value`
`sin`
`                A       3.11476`
`                b       0.105981`
`                c       7.80408`
`Gaussian`
`                A       1`
`                centre  500`
`                sigma   1`
`Polynomial`

after patch:
Naming of the components could be fixed to comply with official style (did not find anything)

`Components      Parameter       Value`
`sin`
`                A       3.11476`
`                b       0.105981`
`                c       7.80408`
`Gaussian`
`                A       1`
`                centre  500`
`                sigma   1`
`Polynomial`
`                a[0]    -0.00117882`
`                a[1]    0.413538`
`                a[2]    -4.8575`

patch
`--- model_old.py	2017-03-21 15:21:32.577933994 +0100`
`+++ model.py	2017-03-21 15:22:55.813400171 +0100`
`@@ -1483,6 +1483,9 @@`
`                     if not hasattr(parameter.value, '__iter__'):`
`                         print("\t\t%s\t%g" % (`
`                             parameter.name, parameter.value))`
`+                    else:`
`+                        for parid in range(len(parameter.value)):`
`+                            print("\t\ta[%d]\t%g" % (parid, parameter.value[parid]))`
 
     def set_parameters_not_free(self, component_list=None,
                                 parameter_name_list=None):

ialxn:
Fixes #1543. Actually, it is probably also an enhancement

### DIFF
--- a/hyperspy/model.py
+++ b/hyperspy/model.py
@@ -1460,7 +1460,7 @@ class BaseModel(list):
             if only_active is False or component.active:
                 component.plot(only_free=only_free)
 
-    def print_current_values(self, only_free=True):
+    def print_current_values(self, only_free=True, skip_multi=False):
         """Print the value of each parameter of the model.
 
         Parameters
@@ -1468,6 +1468,9 @@ class BaseModel(list):
         only_free : bool
             If True, only the value of the parameters that are free will
              be printed.
+        skip_multi : bool
+             If True, parameters that have the attribute `__iter__` set,
+             such as e.g. tuples, lists, arrays, are not printed.
 
         """
         print("Components\tParameter\tValue")
@@ -1481,8 +1484,9 @@ class BaseModel(list):
                     else component.parameters
                 for parameter in parameters:
                     if hasattr(parameter.value, '__iter__'):
-                        for idx in range(len(parameter.value)):
-                            print("\t\ta[%d]\t%g" % (idx, parameter.value[idx]))
+                        if skip_multi:
+                            for idx in range(len(parameter.value)):
+                                print("\t\ta[%d]\t%g" % (idx, parameter.value[idx]))
                     else:
                         print("\t\t%s\t%g" % (
                             parameter.name, parameter.value))
--- a/hyperspy/model.py
+++ b/hyperspy/model.py
@@ -1480,7 +1480,10 @@ class BaseModel(list):
                 parameters = component.free_parameters if only_free \
                     else component.parameters
                 for parameter in parameters:
-                    if not hasattr(parameter.value, '__iter__'):
+                    if hasattr(parameter.value, '__iter__'):
+                        for idx in range(len(parameter.value)):
+                            print("\t\ta[%d]\t%g" % (idx, parameter.value[idx]))
+                    else:
                         print("\t\t%s\t%g" % (
                             parameter.name, parameter.value))
 


francisco-dlp:
I would make it optional by e.g. adding a ``print_multivalue`` (I couldn't come up with a better name, maybe you could?) keyword argument to the function as there are some parameters with very "long" values.

hyperspy/model.py
```diff
@@ -1480,7 +1483,11 @@ def print_current_values(self, only_free=True):
                 parameters = component.free_parameters if only_free \
                     else component.parameters
                 for parameter in parameters:
-                    if not hasattr(parameter.value, '__iter__'):
+                    if hasattr(parameter.value, '__iter__'):
+                        if skip_multi:
+                            for idx in range(len(parameter.value)):
+                                print("\t\ta[%d]\t%g" % (idx, parameter.value[idx]))
+                    else:
```
francisco-dlp, line 1490: This else block should be indented furhter to mach the previous ``if``.

ialxn: No, this should be correct. The `else:` block concerns parameters that do not have the attribute `__iter__` set. The `if skip_multi:` is only applied to iterable parameters.

francisco-dlp: You are right! Maybe the issue is that it should be ``if not skip_multi``?

francisco-dlp: Also, wouldn't it be better to print the parameter name instead of ``a``? This is what I get when using ``skip_multi=True`` (should be ``False``) on an EELS model:

```python


Components	Parameter	Value
PowerLaw
		A	40
		r	-0
Fe_M3
		a[0]	0
		a[1]	0
		a[2]	0
		a[3]	0
		a[4]	0
		a[5]	0
		a[6]	0
		a[7]	0
		a[8]	0
		a[9]	0
		a[10]	0
		a[11]	0
		a[12]	0
		a[13]	0
		a[14]	0
		a[15]	0
		a[16]	0
		a[17]	0
		a[18]	0
		a[19]	0
		a[20]	0
		a[21]	0
		a[22]	0
		a[23]	0
		a[24]	0
		intensity	1
```

ialxn: I'll check to get the parameter name printed and the `skip_multi` issue. But I guess, the parameter values (all zeros) are correct.

ialxn: Yes, of course, it should be `if not skip_multi`.

Otherwise I am a bit at a loss. I only realized now that the `__iter__`able parameters can be more complex, like dictionaries as in the EELS model. Is there a generic way to iterate through the parameters that catch tuples, arrays, ..., dictionaries? Do I have to deal with things like dictionaries of dictionaries? Or a list that contains a dictionary and a list?

francisco-dlp: ``value`` can only contain numbers or a tuple of numbers, so the way you did it should be fine.

francisco-dlp:
#1549

Issue #1557, opened 2017-03-28
ORPCA failing when normalize_poissonian_noise=True


thomasaarholt:
ORPCA is failing on my spectrum images when the argument `normalize_poissonian_noise` is set to `True` (which is the case a lot of the time in Electron Microscopy)
It is reporting a matrix multiplication shape mismatch. @tjof2 any thoughts?

A bit of a bodged hyperspy test as a working example:

```python
import hyperspy.api as hs
from hyperspy.tests.mva.test_rpca import TestORPCA
test = TestORPCA()
test.setup_method("ml") # why the setup_methods all take a "method" argument, I don't know
test.A.shape
(256, 1024)
s = hs.signals.Signal1D(test.A)
s.data[s.data < 0] = 0
s.decomposition(True, algorithm="ORPCA", output_dimension=5)

WARNING:hyperspy.learn.rpca:No method specified. Defaulting to 'CF' (closed-form solver)
WARNING:hyperspy.learn.rpca:No initialization specified. Defaulting to 'qr' initialization
WARNING:hyperspy.learn.rpca:Number of training samples for 'qr' method not specified. Defaulting to 10 samples
WARNING:hyperspy.learn.rpca:Nuclear norm regularization parameter is set to default: 1 / sqrt(nfeatures)
WARNING:hyperspy.learn.rpca:Sparse regularization parameter is set to default: 1 / sqrt(nfeatures)
/Users/thomas/miniconda3/envs/hs/lib/python3.5/site-packages/tqdm/_tqdm.py:65: DeprecationWarning: sys.getcheckinterval() and sys.setcheckinterval() are deprecated.  Use sys.setswitchinterval() instead.
  sys.setcheckinterval(100)
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
<ipython-input-38-1e32f72a728b> in <module>()
      6 s = hs.signals.Signal1D(A.A)
      7 s.data[s.data < 0] = 0
----> 8 s.decomposition(True, algorithm="ORPCA", output_dimension=5)

/Users/thomas/Dropbox/0_Git/hyperspy/hyperspy/learn/mva.py in decomposition(self, normalize_poissonian_noise, algorithm, output_dimension, centre, auto_transpose, navigation_mask, signal_mask, var_array, var_func, polyfit, reproject, return_info, **kwargs)
    434             # Rescale the results if the noise was normalized
    435             if normalize_poissonian_noise is True:
[...]

PR #1558, opened 2017-03-28
orpca now consistent with old implementation

to266:
a quick fix for https://github.com/hyperspy/hyperspy/issues/1557

No tests for now.. But the class / function should probably be made consistent in some other PR anyway (currently class takes the transpose of the function. The function is left purely to keep the API, no other reason actually..), so I'll leave the tests for that.

### DIFF
--- a/hyperspy/learn/rpca.py
+++ b/hyperspy/learn/rpca.py
@@ -550,6 +550,7 @@ def orpca(X, rank, fast=False,
        algorithms", arXiv:1609.04747, (2016), http://arxiv.org/abs/1609.04747.
 
     """
+    X = X.T
     _orpca = ORPCA(rank, fast=fast, lambda1=lambda1,
                    lambda2=lambda2, method=method,
                    learning_rate=learning_rate, init=init,
@@ -557,4 +558,5 @@ def orpca(X, rank, fast=False,
                    momentum=momentum)
     _orpca._setup(X, normalize=True)
     _orpca.fit(X)
-    return _orpca.finish()
+    Xhat, Ehat, U, S, V = _orpca.finish()
+    return Xhat.T, Ehat, U, S, V
--- a/hyperspy/tests/mva/test_rpca.py
+++ b/hyperspy/tests/mva/test_rpca.py
@@ -9,7 +9,7 @@ class TestRPCA:
     def setup_method(self, method):
         # Define shape etc.
         m = 256  # Dimensionality
-        n = 256  # Number of samples
+        n = 250  # Number of samples
         r = 3
         s = 0.01
 
@@ -106,6 +106,15 @@ class TestORPCA:
         # Check the low-rank component MSE
         normX = np.linalg.norm(X - self.A) / (self.m * self.n)
         assert normX < self.tol
+
+    def test_fast(self):
+        X, E, U, S, V = orpca(self.X, rank=self.rank, fast=True)
+        # Only check shapes
+        assert X.shape == (self.m, self.n)
+        assert E.shape == (self.m, self.n)
+        assert U.shape == (self.m, self.rank)
+        assert S.shape == (self.rank,)
+        assert V.shape == (self.n, self.rank)
 
     def test_method_BCD(self):
         X, E, U, S, V = orpca(self.X, rank=self.rank, method='BCD')


thomasaarholt:
Thanks @to266, that was quick! :D

thomasaarholt:
@to266 ORPCA is still not working for me. I think you have fixed one bug but exposed another? If I do 
```python
s.decomposition(True, "ORPCA", 20)
s.get_decomposition_model(5)
s.plot()
```
then every pixel has the same factor, but in varying intensity.

thomasaarholt:
One more thing - As the decomposition progresses, I see the ram being used increasing steadily, from about 3.5 GB (which is just the signal loaded into memory) to 6 GB. At the end it drops down to 3.9 GB. I haven't watched this before, but is that supposed to be happening?

to266:
Yes, it's supposed to happen.

The ORPCA does not work for you because you probably did not read the documentation:
```python
s = hs.signals.Signal1D(np.random.rand(100,99,98)).as_lazy()
s.decomposition(7, algorithm='ORPCA', normalize_poissonian_noise=True)
```

thomasaarholt:
I was comparing non-lazy signals when doing the above. Sorry about the confusion with the other issue. The ram-usage was what made me investigate the decomposition of lazy signals. (At work I've been downgraded to a PC with 8 GB of ram, which is the upper limit for working with several GB spectrum images).

I've double and triple checked now, and the following code produces a spectrum image `sc` with only one component, on my computer:

```python
%matplotlib qt4

import hyperspy.api as hs
import os

path = r"C:\Users\thomasaar\Documents"
name = "EELS Spectrum Image 0eV"
ext = ".dm3"

path = os.path.join(path, "")

s = hs.load(path + name + ext)
s.align_zero_loss_peak(True)
s.decomposition(normalize_poissonian_noise=True, algorithm="ORPCA", output_dimension=7)

sc = s.get_decomposition_model()
sc.plot()
```

to266:
I think this is just a peculiarity of ORPCA. 
In the quick example (both lazy and normal) that I had before it seems that identical bug is present from the first sight too, but actually the procedure to reconstruct the model for any of the non-bss decompositions is exactly the same. Thus if you get "reasonable" factors and loadings from `get_decomposition_factors` and its friends, the model will be equally reasonable.

The issue is that with `normalize_poissonian_noise=True` ORPCA only picks up one significant component, and even reconstructing the full data model with all components looks almost identical to just the first one. The simplest way to verify this is by looking at the loading coefficients. In my case the first one has max value of 1.9e-4, next is 1e-9, followed by 1e-20, etc. I guess to both normalize and get reasonable components you'll have to tweak the ORPCA parameters.

thomasaarholt:
Okay, that's good to be clear on. I'm a bit surprised to be honest, as I can't remember this being the case when I was using ORPCA to noise reduce some EELS Line SIs, but I am very likely mistaken.

I too see the same decreasing loading coefficients. In any case, don't bother with this now, I am happy using the other algorithms for now :) You write your thesis ;)

thomasaarholt:
(For later)
Something has certainly changed. I just installed hyperspy 1.0.1, and there ORPCA works in the manner I would expect. Running exactly the same notebook, I get a far better decomposition model back with the older version. This is probably due to a change of defaults "under the hood", but it would be good to highlight them and, unless you have a good reason, maybe change them back?

The warnings on default settings are the same, but in different order (except for what I assume is renaming of nsamples and nfeatures.

```python
WARNING:hyperspy.learn.rpca:No method specified. Defaulting to 'CF' (closed-form solver)
WARNING:hyperspy.learn.rpca:Nuclear norm regularization parameter is set to default: 1 / sqrt(nsamples)
WARNING:hyperspy.learn.rpca:Sparse regularization parameter is set to default: 1 / sqrt(nsamples)
WARNING:hyperspy.learn.rpca:No initialization specified. Defaulting to 'qr' initialization
WARNING:hyperspy.learn.rpca:Number of training samples for 'qr' method not specified. Defaulting to 10 samples
```

and the new:

```python
WARNING:hyperspy.learn.rpca:No method specified. Defaulting to 'CF' (closed-form solver)
WARNING:hyperspy.learn.rpca:No initialization specified. Defaulting to 'qr' initialization
WARNING:hyperspy.learn.rpca:Number of training samples for 'qr' method not specified. Defaulting to 10 samples
WARNING:hyperspy.learn.rpca:Nuclear norm regularization parameter is set to default: 1 / sqrt(nfeatures)
WARNING:hyperspy.learn.rpca:Sparse regularization parameter is set to default: 1 / sqrt(nfeatures)
```

tjof2:
@thomasaarholt  Notice the difference between:

Old: `Sparse regularization parameter is set to default: 1 / sqrt(nsamples)`
**and** 
New: `Sparse regularization parameter is set to default: 1 / sqrt(nfeatures)`

The default parameter of `1/sqrt(...)` should use the number of features, not the number of samples (since this is in theory unknown when doing an online decomposition of course!), but at some point (probably related to the transposed signal bug...) I got them the wrong way round.

@to266 is right of course, parameter choice is crucial for ORPCA, you can't just set a number of components and expect a good result like PCA, due to the nature of the problem being solved. If the matrix is scaled between 0 and 1, then it just so happens `1/sqrt(nfeatures)` is a very good default choice. 

The normalization for Poisson noise will change the range of values so it's no longer [0,1], and this will affect the parameter choice. It would be nice to know in advance how much the normalization has scaled the matrix by, so you could scale the regularization parameters under the hood and avoid this issue, but it's not clear how you normalize for Poisson noise in an online manner.

Lastly, please try the SGD method (or Momentum version) with the values suggested in the docs, and let me know how well they work in comparison. The CF solver is actually a bit terrible in practice due to the matrix inverse involved, but it provides a nice exact answer for theoretical analysis of the type one might write in a thesis...

francisco-dlp:
@to266  and @thomasaarholt, is this ready?

thomasaarholt:
Afaik, this works as it should.

to266:
I believe so

thomasaarholt:
Merge ahead.

Issue #1560, opened 2017-03-30
transpose on lazy signal does not rechunk


francisco-dlp:
When transposing, signal operations can require big amounts of memory if the lazy data is not rechunked. Issue reported by @magnunor in gitter using the following example:

```python

>>> s = hs.load("large_dataset.hdf5", lazy=True)
>>> s.data
dask.array<array, shape=(256, 256, 256, 256), dtype=>u2, chunksize=(2, 2, 256, 256)>
>>> s1 = s.T
>>> s1.data
dask.array<transpose, shape=(256, 256, 256, 256), dtype=>u2, chunksize=(256, 256, 2, 2)>
>>> s.plot(navigator='slider') # works fine
>>> s1.plot(navigator='slider') # crashes my computer

```

PR #1561, opened 2017-03-31
Lazy chunking improvements

to266:
Fixes https://github.com/hyperspy/hyperspy/issues/1560 

Also improves `change_dtype` behaviour to chunk according to the larger dtype before actually casting to not crash the memory. e.g. `uint8` with `(8, 8, 16, 16)` to `float` will first get rechunked to `(1, 1, 16, 16)` and only then re-cast as `float`.

### DIFF
--- a/hyperspy/_signals/lazy.py
+++ b/hyperspy/_signals/lazy.py
@@ -814,6 +814,11 @@ class LazySignal(BaseSignal):
         target.explained_variance = explained_variance
         target.explained_variance_ratio = explained_variance_ratio
 
+    def transpose(self, *args, **kwargs):
+        res = super().transpose(*args, **kwargs)
+        res._make_lazy(rechunk=True)
+        return res
+    transpose.__doc__ = BaseSignal.transpose.__doc__
 
 def _reshuffle_mixed_blocks(array, ndim, sshape, nav_chunks):
     """Reshuffles dask block-shuffled array
--- a/hyperspy/_signals/lazy.py
+++ b/hyperspy/_signals/lazy.py
@@ -93,7 +93,7 @@ class LazySignal(BaseSignal):
         self._lazy = False
         self._assign_subclass()
 
-    def _get_dask_chunks(self, axis=None):
+    def _get_dask_chunks(self, axis=None, dtype=None):
         """Returns dask chunks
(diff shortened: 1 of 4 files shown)


francisco-dlp:
LGTM but I have a question. Currently the chunks seem to "expand" all axes in order to fulfill the 100MB rule. This means that when iterating following the usual scanning zigzag pattern (e.g. to aligh signals using the cascade method) spectra that are not needed for that particular step of the operations are loaded to memory. Wouldn't it be better to chunk following the zigzag pattern then? Also chunks bigger than a individual signal are useful assuming that the operation is going to go through all the dataset and that order doesn't matter, but there are occasions when order matters (as in the previous example) and we may not want to process the whole dataset. Wouldn't then be good to provide an easy way to rechunk for these situtations?

to266:
Could you explain your second example? I don't get it

francisco-dlp:
Say you want to sum over a selected number of non-contiguous pixels in an SI, wouldn't it be more efficient to have small chunks of 1 individual signal in this case?

to266:
Just creating the rechunking may be too slow:
```python
In [47]: %%time
    ...: chunks = (1,1,1000)
    ...: data = da.random.random((500,500,1000), chunks=chunks)
    ...: inds = np.random.randint(500, size=(100, 2))
    ...: res = da.stack([data[i[0], i[1], :] for i in inds])
    ...: _ = res.sum(0).compute()
    ...: 
CPU times: user 13.6 s, sys: 360 ms, total: 13.9 s
Wall time: 14.2 s

In [48]: %%time
    ...: chunks = (50,125,1000)
    ...: data = da.random.random((500,500,1000), chunks=chunks)
    ...: inds = np.random.randint(500, size=(100, 2))
    ...: res = da.stack([data[i[0], i[1], :] for i in inds])
    ...: _ = res.sum(0).compute()
    ...: 

CPU times: user 3.77 s, sys: 163 ms, total: 3.93 s
Wall time: 1.45 s
```
Also, rechunking does not replace the chunks in the DAG - instead just adds another operation that combines them or splits further, so loading in large "squares" would still happen, thus I guess these examples are in fact better-performing than I would expect.

Issue #1396, opened 2017-01-19
improve camera and stage metadata


ericpre:
As mentioned in #1385, it would be good to add more camera and stage metadata.
Currently the exposure time in TEM is stored in `Acquisition_instrument.TEM.exposure` and for STEM, there is `Acquisition_instrument.TEM.dwell_time`.

If we add a `camera` node, I guess it would be good to move the `Acquisition_instrument.TEM.exposure` leaf to `Acquisition_instrument.TEM.Detector.camera`. In this case, it would be considered as a "detector node" similar to EELS and EDS ones.
Regarding STEM mode, should the `Acquisition_instrument.TEM.dwell_time` leaf stay as it is? Is there any io plugins that import metadata about STEM detector? As far as I know, the DM and FEI don't support it and we don't even know the kind of detector.

If we are moving the `Acquisition_instrument.TEM.exposure` leaf, it will break the API, so I guess the PR would need to be made to the `RELEASE_next_major` branch?

ericpre:
See this [commit](https://github.com/hyperspy/hyperspy/pull/1385/commits/ffe95289899db257ac7009cd2ca5648de06b151d) for camera metadata.

woozey:
> If we add a camera node, I guess it would be good to move the Acquisition_instrument.TEM.exposure leaf to Acquisition_instrument.TEM.Detector.camera. In this case, it would be considered as a "detector node" similar to EELS and EDS ones.

That makes the most sense to me, as camera is a detector too. In the https://github.com/hyperspy/hyperspy/pull/1385/commits/ffe95289899db257ac7009cd2ca5648de06b151d mentioned above my first ideas about what necessary camera metadata may look like.

Regarding stage metadata it could contain quite advanced information (though it is rarely saved together with images at the moment...). For example:
-  Position: x,y,z, alpha, beta, azimuth
-  Stage type (holder type in this case)
-  Temperature
-  Applied voltage, current and magnetic field (?) Actually this part may even need another branch in `TEM` called `insitu`.
-  Also I dream about the day when all lens and other optical element currents will be saved together with data...

sem-geologist:
IMHO stage data could also be handy not only for TEM but also for SEM

what is the parameters alpha, beta and azimuth? for possition?

maybe for stage position also rotation could be added, or is this the azimuth parameter?

woozey:
> what is the parameters alpha, beta and azimuth? for possition?

[...]

PR #1563, opened 2017-04-02
Fix reading metadata tif issue with scikit image 0.13

ericpre:
Tifffile.py have been updated in the last release of scikit-image and this PR fixes parsing scale and units of tiff files.

- [x] Fix parsing scale and units with last version of scikit-image.
- [x] Parse more metadata for SEM.
- [x] Add a stage node (tilt_a, tilt_b, rotation, x, y, z) to metadata specification (See discussion in #1396).
- [x] Get a TVIPS camera test file and parse the metadata.

In this PR, the metadata (including scale and units) would not be parsed with older version of scikit-image (<0.13) or tifffile.py. The file will be still loaded correctly, though. The main reason is to make more simple. I would suggest to also complete #1433 after merging this one.


ericpre:
It would be great if this PR could be reviewed and merged asap, to get CI green again... and then do #1565 and #1433!

francisco-dlp:
Why not simply requiring skimage >= 0.13?

ericpre:
Agree.

francisco-dlp:
Just a couple of things:
* When changing the metadata structure (in this case replacing ``tilt_stage`` by the stage node) we must change the version in ``io_plugins/hdf5.py``. In this case, we need to increase the major index so it'll be v3.0. Now that you are at it, you could add the version to the metadata structure chapter, as this is actually the version of our hdf5 specification.
* HyperSpy must be able to read all the information from previous hdf5 specifications, so  ``io_plugins/hdf5.py`` requires modification to look for ``tilt_stage`` and place the information in the new node. A test to check that it does it properly is also required.

ericpre:
I think some of it is in #1565. The two PR go together, sorry for the confusion, I made a separate one that update the metadata in all io plugins. I didn't wanted to make this one too complicated and focused here on tiff only.
Should we move the discussion regarding hdf5 in #1565?

francisco-dlp:
The same applies to the camera node in #1565. Why not e.g. merging #1565 here and closing it?

francisco-dlp:
Sorry to come up with a question after merging: why not renaming ``tilt_a`` and ``tilt_b`` to the more familiar ``tilt_alpha`` and ``tilt_beta``?

ericpre:
No particular reason, can change it to `tilt_alpha` and `tilt_beta`.

Issue #1511, opened 2017-03-04
Warn when plotting signal that contains markers


francisco-dlp:
A signal can contain permanent markers but they are not plotted by default. Although it is possible to know if it contains markers by checking the metadata, wouldn't it be useful to add a warning to the logger when plotting a signal that contains markers when ``plot_markers`` is ``False``?

magnunor:
I don't think this is a good idea, since markers will be added to signals automatically (for example https://github.com/hyperspy/hyperspy/pull/1491). This means that at some point, when more of the file loaders also adds markers, pretty much every `s.plot(plot_markers=False)` will produce a warning.

Semi-related: we could make `plot_markers=True` the default

ericpre:
It looks like users don't like to see warning... maybe it is best to avoid it, if not very important.
+1 on making plot_markers=True, the default.
This default could be set in the preferences. It could also be signal dependant.

thomasaarholt:
+1 for the plot_markers=True as well. @magnunor's argument about constant warnings is true.

sem-geologist:
+1 for `plot_marker=True`; or it should go to `logger.info`, not `warnning`. Bcf v2 is going to have marker too (only one rectangle only on overview image,but it would be nice if it would be plotted automatically)

ericpre:
Fixed by #1564.

PR #1564, opened 2017-04-03
Change default for s.plot(plot_markers=False) to plot_markers=True

magnunor:
This pull request changes the default behavior for s.plot: the default for plot_markers is changed from False to True. So permanent markers will be plotted by default.

Some more discussion here: https://github.com/hyperspy/hyperspy/issues/1511


magnunor:
The failing travis tests are related to FEI loader:
```Python
    def test_read_FEI_SEM_scale_metadata_8bits():
        fname = os.path.join(MY_PATH2, 'FEI-Helios-Ebeam-8bits.tif')
        s = hs.load(fname)
>       assert s.axes_manager[0].units == 'm'
E       AssertionError: assert <undefined> == 'm'
E        +  where <undefined> = <width axis, size: 512>.units
```

Which I don't think is related to my changes

magnunor:
Any thoughts on this?

ericpre:
Should be fixed in #1563: update your branch with upstream.

PR #1566, opened 2017-04-04
Fix post checkout githook

vidartf:
Fix the post checkout git hook, such that all paths for `rm` gets quoted. Without this, something could go wrong if e.g. the repository is in the path `C:\windows development files\hyperspy` (will try to delete `C:\windows`, `development` and `hyperspy/hyperspy/io_plugins/unbcf_fast.c`).

Note: That command should not actually do anything to `c:\Windows` since `rm` should not delete folders, but you get my drift.

An alternative is of course to remove the post-checkout code from the setup.py code, and make it an opt-in feature instead. I would argue that installing hooks automatically is a little confusing (and possibly rude).

Someone will have to confirm if this quoting (double quotes) will work safely on other OSes than Windows.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -111,7 +111,6 @@ for leftover in raw_extensions:
                     path +
                     '.cpython-*.so'))
 
-
 def count_c_extensions(extensions):
     c_num = 0
     for extension in extensions:
@@ -201,7 +200,7 @@ def find_post_checkout_cleanup_line():
 # after changing branches:
 if os.path.exists(git_dir) and (not os.path.exists(hook_ignorer)):
     exec_str = sys.executable
-    recythonize_str = ' '.join([exec_str,
+    recythonize_str = ' '.join(['"%s"'%exec_str,'"%s"'%
                                 os.path.join(setup_path, 'setup.py'),
                                 'clean --all build_ext --inplace\n'])
     if os.name == 'nt':
@@ -224,7 +223,7 @@ if os.path.exists(git_dir) and (not os.path.exists(hook_ignorer)):
             line_n = find_post_checkout_cleanup_line()
             if line_n is not None:
                 hook_lines[line_n] = 'rm ' + \
-                    ' '.join([i for i in cleanup_list]) + '\n'
+                    ' '.join(['"%s"' % i for i in cleanup_list]) + '\n'
                 hook_lines[line_n + 1] = recythonize_str
             else:
                 hook_lines.append('\n#cleanup_cythonized_and_compiled:\n')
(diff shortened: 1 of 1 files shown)


erh3cq:
Still having the issue
"""
~\Documents\GitHub\hyperspy [Mule ≡]> git checkout RELEASE_next_minor
Your branch is up-to-date with 'origin/RELEASE_next_minor'.
Switched to branch 'RELEASE_next_minor'
rm: cannot remove 'C:/Users/Eric Hoglund/Documents/GitHub/hyperspy/hyperspy/io_plugins/unbcf_fast.c*': No such file or directory
rm: cannot remove 'C:/Users/Eric Hoglund/Documents/GitHub/hyperspy/hyperspy/io_plugins/unbcf_fast.cpython-*.pyd': No such file or directory
.git/hooks/post-checkout: line 4: C:/Program: No such file or directory
"""

vidartf:
Well, those are OK if the files actually do not exist.

vidartf:
I.e. if they were not previously built.

erh3cq:
the files are named uncb_fast.c and uncb_fast.pyx
So the first remove should work?
The second should be pyx not .pyd?
The last remove seems to be attempting to remove the file from C:/Program not .git/hooks

vidartf:
The last line is actually not a removal command, but I guess that should get quotes as well.

sem-geologist:
@vidartf  I agree that all this githook is terrible mess, I were thinking to remove it completely, but still not thought up the replacement... Maybe `.hook_ignore` file should be created by default in the source tree, then by default that post checkout hook would not work.
But maybe instead of small hack It should  completely be ditched from `setup.py`---it complicates it too much.

erh3cq (changes requested):
Can we make 185&186 have quotes? On windows it is searching for C:\Program Files, but that contains spaces.
227 needs quotes?

I am wondering, would it be better to use .replace(' ','%20') rather than quotes on all of these?
-setup_path.replace(' ','%20') near line 47
-git_dir.replace(' ','%20')

erh3cq:
I have edited the setup file, and it is working.  How do I upload it to here?

vidartf:
The best solution is to make a PR to my branch ([vidartf:fix-githook](https://github.com/vidartf/hyperspy/tree/fix-githook)).

vidartf:
Whoops, some commits from minor got included, I'll rebase the branch to sort it out.

francisco-dlp:
Don't worry about it, the next release will be a minor.

vidartf:
There might still be some issues left, but this PR should at least be a strict improvement.

PR #1567, opened 2017-04-04
Make ripple format a bit more tolerant

francisco-dlp:
The ripple format stipulates that the separator between keys and values is a single tab. However, it is not uncommon to find ripple files with spaces and/or a mixture of tabs and spaces in the wild. This let's HyperSpy load these mildly illegal files.

### DIFF
--- a/hyperspy/io_plugins/ripple.py
+++ b/hyperspy/io_plugins/ripple.py
@@ -162,7 +162,7 @@ def parse_ripple(fp):
                 err = 'Separator in line "%s" is wrong, ' % line
                 err += 'it should be a <TAB> ("\\t")'
                 raise IOError(err)
-            line = line.split(sep)  # now it's a list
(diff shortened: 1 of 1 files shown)


thomasaarholt:
I was literally going to create an issue on this 30 minutes ago, but then I had to run to the train.

thomasaarholt:
Will test this on my file and see if it now works

thomasaarholt:
EDIT: Hold on this. I wasn't using the correct branch on my computer. New PC.

Hm, it still doesn't work on my file, which has up to 9 spaces between the key and value.

[Base(5).zip](https://github.com/hyperspy/hyperspy/files/896159/Base.5.zip)

thomasaarholt:
It works! Looks good, @francisco-dlp!

francisco-dlp:
Thanks @thomasaarholt, I'll merge it then.

PR #1571, opened 2017-04-06
Warn calibration

### LINKED ISSUE #1570: Bug (?) in load_data()
I am loading EDS data with `load_data()`. If I read the data as a tuple

`a, b = hs.load(["1.msa", "2.msa"], signal_type='EDS_TEM')`

the data are correctly loaded independent of whether they have the same shape or the same calibration. If I load them as a stack

`s = hs.load(["1.msa", "2.msa"], signal_type='EDS_TEM', stack=True)`

I do get an error (`cannot broadcast`), which is correct, if they do not have the same shape. If the data have the same shape (same number of channels) but are differently calibrated (different values of `offset`, `scale`, and `units`) the data are read without problems but the calibration of the first spectrum read is silently applied to all subsequently read spectra.
We should probably issue an error too in this case.

ialxn:
This fixes #1570 for me but I only checked with EDS (msa) files (no multiple signals per file). And it is ugly and not really pythonic. So please comment/suggest.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -131,10 +131,10 @@ before_deploy:
   - "%CMD_IN_ENV% %WP_INSTDIR%/scripts/env.bat"
   # Give info about python vesion and compiler used to compile the python
   - "%CMD_IN_ENV% python.exe -c \"import sys; print(sys.version)\""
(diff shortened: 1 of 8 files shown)


ialxn:
Restructured code and submitted new pull request (V2).

PR #1572, opened 2017-04-07
Warn calibration v2

ialxn:
This fixes Issue #1570 . Signals not matching calibration data `scale`, `offset`, `units` of first signal are omitted and a warning is issued.

V2: restructured code. Fixed logic (`if ... elif` -> `if ... else`).

### DIFF
--- a/hyperspy/io.py
+++ b/hyperspy/io.py
@@ -245,6 +245,39 @@ def load(filenames=None,
     >>> s = hs.load('file*.blo', lazy=True, stack=True)
 
     """
+    def calibration_is_ok(obj, calibration):
+        """
+        Check if calibration data of `obj` matches `calibration`
+
+        Parameters
+        ----------
+        obj : signal
+            The signal to be checked
+        calibration : tuple
+            calibration data (offset, scale, units)
+
+        Returns
+        -------
+        True :
+            calbration data matches
+        False :
+            otherwise
+        """
+        this_axis0 = obj.axes_manager.as_dictionary()["axis-0"]
+        if calibration[0] != this_axis0["scale"] or \
+           calibration[1] != this_axis0["offset"] or \
+           calibration[2] != this_axis0["units"]:
+            return False
+        else:
+            return True
+
+    def get_calibration(obj):
+        """
+        Returns calibration data (`scale`, `offset`, `units`) of signal `obj`
+        """
+        a = obj.axes_manager.as_dictionary()["axis-0"]
+        return (a["scale"], a["offset"], a["units"])
+
     deprecated = ['mmap_dir', 'load_to_memory']
     warn_str = "'{}' argument is deprecated, please use 'lazy' instead"
     for k in deprecated:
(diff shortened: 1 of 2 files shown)


francisco-dlp:
As pointed out in #1570, warning the user when the calibration doesn't match is certainly a good idea. This is a first step, but it needs to be more general. In my mind all axes should obey some rules regarding calibration. For example, if we have to line spectra ``(10| 100)`` and ``(10| 100)``, if we stack them like this ``(20| 100)`` the calibration of the signal axis must be the same, but also the scale and units of the navigation axis, and the offset should follow ``offset2 = offset1 + scale1 * 10``. On the other hand, stacking like this ``(2, 10| 100)`` the nav and sig axes should have the same calibration. Also, when stacking signals of higher signal dimension (e.g. images) all signal axes should be checked.

ialxn:
I see. I was wondering how to deal with the issue in a more general way. I could imagine that some people would want to stack `(10|100)` and `(10|100)`  to `(20|100)` as a time series, or just a collection of signals. So enforcing the same calibration or not ... I do not see a "correct" way to do this without an additional keyword, which is also clumsy. So maybe the fix should really be just a clear statement in the documentation.

On a second thought. Maybe there is a heuristics that catches most real problems, such as just a single calibration that differs.

On a third thought. Maybe the easiest solution is just to warn the user but still load the data.

to266:
For now I would go the warning and thorough documentation route.

At some point later, if/when we think it's worth spending time implementing, we could add a `kwarg` to most functions that takes into account calibrations (such as `calibrated=True`) - which would require non-stacked axes to be identical / would crop them to be identical, and then stack (or perform any other operation?) the signals in the logical (in terms of the axis) way.

tjof2:
Will tidy this up plus related issue (#1570) such that its documented and logs a clear warning.

tjof2:
OK so here's an example:

```python
import os
import tempfile
import hyperspy.api as hs

s1 = hs.signals.Signal1D([1.0, 2.0, 3.0])
s2 = hs.signals.Signal1D([1.0, 2.0, 3.0])

s1.axes_manager[0].scale = 1.0
s1.axes_manager[0].offset = 1.0
s1.axes_manager[0].units = "cm"

s2.axes_manager[0].scale = 2.0
s2.axes_manager[0].offset = 1.0 + 1e-5
s2.axes_manager[0].units = "mm"

with tempfile.TemporaryDirectory() as tmpdir:
    f1 = os.path.join(tmpdir, "s1.hspy")
    f2 = os.path.join(tmpdir, "s2.hspy")

    s1.save(f1, overwrite=True)
    s2.save(f2, overwrite=True)

    _ = hs.load([f1, f2], stack=True)

WARNING:hyperspy.io:Mismatched calibration when stacking signals: scale (1.0 != 2.0), offset (1.0 != 1.00001), units (cm != mm). The calibration of the first signal will be applied to all signals
```

As @francisco-dlp mentions above, ideally this should be much more general. However, printing a warning is a good step for now.

tjof2:
Actually, this might not be the best place for it. Will close and open a new PR that's cleaner.

PR #1580, opened 2017-04-08
Fix #1578, improve the figure size of the navigation plot.

### LINKED ISSUE #1578: Inconsistent rcparams for sig and nav plots
Setting `matplotlib.rcParams['figure.figsize'] = np.array([6.4, 4.8])/2`* changes the size of new signal plot windows, but not the navigator plot windows.

This implies that the navigator window figure size might be hard coded. It would be nice if it read out the default figure size.

This will become relevant for the `%matplotlib notebook` backend with the new ipywidget UI, as the interactive plot is inline. Huge images will make it difficult to use the controls.

*The list contains default parameters. Instead of `matplotlib`, `plt` can be used.

A way to set the same figure size is to change `matplotlib.rcParams['figure.dpi'] = 80`.

ericpre:
Normalise the figure size with the parameters set in matplotlib.rcParams['figure.figsize'].

Fix #1578.


thomasaarholt:
It works great. Thanks @ericpre!

thomasaarholt:
I think you did something you didn't mean to do @ericpre. You are now modifying 73 files.

ericpre:
No, this is expected; read the comments of the commit b0ec9c6.

thomasaarholt:
Wow, ok. Was on mobile, and just spotted the large change. Good work.

ericpre:
What about this one?
The tests failure are the _usual suspects_, i.e. the scikit-image v0.13 issues.
I had to change the baseline images, because the default style in pytest-mpl is 'classic' and there was issue with the style changes (default figure size and dpi are not the same in mpl 1.x and 2.x)

francisco-dlp:
I would suggest a different approach: instead of calculating a factor, simply set the ``max_size`` to the ``max(rcParams["figsize"])``.

ericpre:
Yes, it works well and it's more simple! Thanks!
While I am looking at this kind of things, I am going to try to tweak the aspect ratio of the figure. In case of high aspect ratio image, there is a lot of empty space. Not sure I can get much improvement but I will give a go.

On related topic: in a image plot, is there a reason to plot the axes labels when we have the scalebar? Shall we set the default to remove it? In the navigation plot, is worth keeping the colorbar?

thomasaarholt:
Eric, can you create a new PR for that? We have a lot of PRs that keep getting larger and larger - nice to break things down! :) But yes please, that would be awesome!

francisco-dlp:
@ericpre, you are right, the axes and the scalebar should be anticorrelated. At some point they were, so if they are not now, you've found a but.

I am for keeping the colobar in the navigator: it doesn't hurt when it is not needed, and not having it can lead to misinterpretation when e.g. vmin is not 0 or is negative.

francisco-dlp:
I think that this is ready to merge except for 857628a that I think should be reverted.

ericpre:
Yes, I had a go at trying to fix a few bugs with the aspect ratio of the figure, but it needs some work and I will not have time to do this soon. So, I reverted 857628af92fa15ea469ca77c974f07c639b884cc, to finished here and I will make another PR later.

ericpre:
Same than #1608, thanks @francisco-dlp!

PR #1585, opened 2017-04-10
Fix #1575, test_compilers.c location

### LINKED ISSUE #1575: test_compilers.obj is installed in the wrong directory
When I install hyperspy from Rnm I get an install error, as it can't find test_compilers.obj. If my hyperspy path is:
`C:\Users\thomasaar\Downloads\hyperspy\`
Then it is installed in 
`C:\Users\thomasaar\Downloads\hyperspy\Users\thomasaar\Downloads\hyperspy\hyperspy\misc\etc`
It is easily spotted by the creation of a "Users" folder in the hyperspy directory.

This appears to only be an issue on Windows?

thomasaarholt:
On Mac and Windows, the directory `.../hyperspy/Users...` is created when setup.py is ran because the compiler took an unnecessarily long path. Since I don't have a linux machine to test on, it would be nice if someone else could confirm that it works there.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -26,6 +26,7 @@ import itertools
 import subprocess
 import os
 import warnings
+from tempfile import TemporaryDirectory
 from setuptools import setup, Extension, Command
 import sys
 
@@ -174,8 +175,9 @@ compiler = distutils.ccompiler.new_compiler()
 assert isinstance(compiler, distutils.ccompiler.CCompiler)
 distutils.sysconfig.customize_compiler(compiler)
 try:
-    compiler.compile([os.path.join(setup_path, 'hyperspy', 'misc', 'etc',
-                                   'test_compilers.c')])
+    with TemporaryDirectory() as tmpdir:
+        compiler.compile([os.path.join(setup_path, 'hyperspy', 'misc', 'etc',
+                                   'test_compilers.c')], output_dir=tmpdir)
 except (CompileError, DistutilsPlatformError):
     warnings.warn("""WARNING: C compiler can't be found.
 Only slow pure python alternative functions will be available.


francisco-dlp:
Are you sure that the issue is the long path? I would rather be suspicious of the ``/`` linux directory separator, although I thought that ``join`` would handle them correctly in windows.

Regarding the fix, I am not sure if it is a good idea turning the absolute path into a relative one as if somebody runs ``setup.py`` from elsewhere it won't work.

thomasaarholt:
I forgot about the windows case. Added that now. You have a point though, calling `python C:\Users\thomasaar\hyperspy\setup.py install` from another directory does not work with this PR.

I think I am missing something with how ccompile works. With master, on Mac and Windows (showcased for Mac here: ), when I checkout a branch the `.../hyperspy/Users...` directory is created, and the `test_compilers.o` file is not created in the correct directory. 

To be clear, it is not in `/Users/thomas/hyperspy/hyperspy/misc/etc/test_compilers.o`, but instead in `/Users/thomas/hyperspy/Users/thomas/hyperspy/hyperspy/misc/etc/test_compilers.o`. See screenshot below.

screenshot 2017-04-17 19 11 06

francisco-dlp:
Does #1723 fix this?

thomasaarholt:
It does not. 
image

I tried checking out latest master, deleting the path, then checking it out again, and the file and path is still created.

sem-geologist:
I am trying to remember why I needed that hack for fast_bcf parser. I think the error message was different on different os if compiler was not present. I should look into this and try fixing the mess which originated from my bcf implementation.

sem-geologist:
Oh I see, that is a dummy file which is only tried to be compiled to check if compiler is present (and is not used for anything). I am wondering if the file could not be assembled (as it is very short) in some temp directory instead of being hard coded in the separate dummy file. Then it could be tried to be compiled with separate python function (so that created object/dll file would also be created in tmp folder). There is python routine for handling tmp folders and no more `os.path.join` will be needed in user/ hyperspy folder hierarchy.

ericpre:
Indeed, the following should be better:
```python
with TemporaryDirectory() as tmpdir:
    compile.compile()
```

thomasaarholt:
Waiting for checks to finish. I've tested on my machine (Windows), and it seems to do the trick. Thanks to @ericpre for the suggestion.

PR #1588, opened 2017-04-10
Update signal1d.rst

mwalls:
In the Documentation, a number of links to the detailed command descriptions are broken. I don't know if this is a known issue. For example, in this section (Signal1D tools) only the spikes removal tool link actually works. I think it may be due to the presence of a thing called "Signal1DTools" which no longer exists. Singal1D does exist so I'm trying it here. If I'm right I'll do the others but I don't know how to test links within the documentation before I do a pull request, so I need confirmation.
Thanks

### DIFF
--- a/doc/user_guide/signal1d.rst
+++ b/doc/user_guide/signal1d.rst
@@ -14,7 +14,7 @@ In addition to cropping using the powerful and compact :ref:`Signal indexing
 <signal.indexing>` syntax the following method is available to crop spectra
 using a GUI:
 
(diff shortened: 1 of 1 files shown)


thomasaarholt:
Are you building the docs with sphinx? In any case, I think this is much appreciated :)

mwalls:
No, I don't know what Sphinx is. I'm just doing it by hand.
Thanks for your encouragement!
(But do you know if I'm right?)

francisco-dlp:
Thanks!

thomasaarholt:
Needless to say, you're right :) please do continue to pick up on these
things!

PR #1593, opened 2017-04-11
Fix a typo in docs - components1D.Expression

thomasaarholt:
Since #1590 was merged into Rnm and not Rnp and this modifies the same file, I can either point this here to Rnm, or figure out how to copy @mwalls commit into my PR and point it at Rnp.


francisco-dlp:
We don't worry about Rnp and Rnm for this realease as it'll be a minor.

francisco-dlp:
Thanks!

PR #1594, opened 2017-04-11
Run tests in windows bundle

ericpre:
- [x] Upgrade scikit-image to v0.13 in WinPython Windows bundle (#1563 needs scikit-image v0.13 to keep the code simple).
- [x] Run tests in WinPython.

After a couple of tedious iteration with appveyor, I can confirm this is still [building correctly](https://ci.appveyor.com/project/ericpre/hyperspy/build/1.0.40)!
This PR also includes #1563 to pass the tests.


francisco-dlp:
Waiting for #1563 to be ready.

vidartf:
Given that the hyperspy tests normally run at around ~45 minutes, this would have the effect of having the re-run overshoot the maximum AppVeyor time (1 hour). This is the reason why it wasn't included in the first place.

ericpre:
The tests in the WinPython bundle are run in the 'before_deploy' part and this is done only when a tag is added, so it will be only for releases and for other PRs, there will be no change.

vidartf:
Yes, but when there is a tag, it will try to run the tests twice, which will cause it to take over 1 hour, right?

francisco-dlp:
@ericpre, have you tested it? If yes, do you run into the issue described by @vidartf?

ericpre:
Yes, I have tested it, see the [link](https://ci.appveyor.com/project/ericpre/hyperspy/build/1.0.40) in the first post.

ericpre:
Actually, it does take more than 1h, but it works. The build, which are failed, are with Python 3.6, which I assumed it is not important, before the WinPython bundle is Python 3.5?!

francisco-dlp:
Yes, we're stuck with Py35 because winpython 3.6 does not come with pyqt4

vidartf:
@ericpre Sorry, my mistake: I was mixing the total run-times of the entire job, and the partial run-times (which are typically around 10 min now!). Good job on getting the test times shorter, whoever did it :D

francisco-dlp:
The tests weren't the issue, the issue was that the bundle was packaged with all the pyc files. After removing then the running time decreased dramatically...

PR #1598, opened 2017-04-17
Model and component print_current_values() methods

thomasaarholt:
### Description of the change
I wanted a `print_current_values()` method for components that was similar to the model method. In notebooks, using the fancy `component.gui()` is a bit much if you just want a quick look.

It's implemented both a html view as well as a terminal-style view with an improved look.

Since I've created a class that can be looped to produce either html or text views of a component, I've also added this for the model class, for printing all components, with filters for whether they are active or have free parameters. One also has the option of displaying it in the old style. I can add a preferences option for this if desired. With this method I overwrite the old `m.print_current_values()` method.

[Here it is in action showing the values of the EDS Spectrum model](https://nbviewer.jupyter.org/urls/dl.dropbox.com/s/gwcl31sl0u8ofrt/Component_print_current_values3.ipynb).

### Progress of the PR
- [x] Component HTML view implemented
- [x] Component terminal view implemented
- [x] Model view of both implemented
- [x] add tests,
- [x] ready for review.

```python
%matplotlib inline
import hyperspy.api as hs

m = s.create_model()
m.fit()

back, g1, g2, *_ = m 
back.print_current_values(fancy=True)
g1.print_current_values(fancy=True, only_free=True)
g2.print_current_values(fancy=False, only_free=False)
m.print_current_values(only_free=True)
m.print_current_values(fancy=False)
```


thomasaarholt:
Annoyingly, qtconsole does no longer support html printing, so this command would not work there.

I'll leave it to you guys to decide whether we want a text-based print similar to the model method or stick with this html table view. I prefer the latter (which is why I made it), but if we have many users of qtconsole, then I understand.

Looking at the code of the s.axes_manager class, I can't tell how code distinguishes between the qtconsole and the notebook. Does anyone have any advice?

francisco-dlp:
There is a solution to the problem. Adapted from the ipython docs:

```python
class PritMe:
    def __init__(self, text):
        self.text = text

    def __repr__(self):
        return self.text
    
    def _repr_html_(self):
        return "<h1>" + self.text + "</h1>"
```
If your method returns an instance of this class it will be printed properly in the notebook and the terminal. We should take this approach for all methods printing tabular data.

francisco-dlp:
Ideally, we would like to have a ``PrettyTable`` class that takes tabular data with headers and prints it properly in the notebook and the terminal. It'll be worth checking if there is something like this out there not to reinvent the wheel—I'll be surprised if not.

to266:
[this](http://jupyter-notebook.readthedocs.io/en/latest/changelog.html?highlight=table#table-style) might be of interest.

Also, have a look how `axes_manager` prints these things and either use it or improve upon it at the same time to make all our tables look similar

thomasaarholt:
~~The table output on that page (thanks Tomas!) comes from pandas - I'll see
if there is code we can borrow from it. Hyperspy does not depend on pandas,
right? (Otherwise I would use it directly).~~
It does not. it is part of IPython-

francisco-dlp:
No, hspy doesn't depend on pandas. I think that, at least there is a small pure python package to perform the task, it'll be better to just implement our own. Using @to266's ``AxesManager`` one everywhere sounds good!

thomasaarholt:
Right, I think this is ready for review. I am using a class rather than a method in order to use the `__repr__` and `_repr_html_` class functionality.

[Example video](https://streamable.com/0icms)

thomasaarholt:
Hang on, just need to update push access rights on my pc. The final commit (from pc) didn't send.

thomasaarholt:
There we go. All good, please check it out.

francisco-dlp:
Looks good except for:

* It needs documentation in the User Guide
* What about adding the std if available?

francisco-dlp:
By the way, how did you get the awesome collapsable metadata in your notebook example?

thomasaarholt:
I used html5. Doesn't work in Microsoft browsers (the arrows just disappear though, so it doesn't break).
```html
<details>
  <summary>See More</summary>
  This text will be hidden if your browser supports it.
</details>
```
However, I must have removed it here for some reason. I'm a bit ill, so will look at this when I get better.

thomasaarholt:
I've just added docs and the standard deviation to this! I think this is ready to merge!

thomasaarholt:
I'm intending to add the same to the m.print_current_values (possibly creating a m.current_values as well as a m.print_current_values). s.metadata will get the same treatment, hopefully including the details / summary html5.

thomasaarholt:
**updated top post with example**

I've realised I can just use the `IPython.display.display` method to "print" both text and html in console and notebook respectively. Renamed things to `print_current_values`.
Also introduced print_current_values for model as well as individual components.

Do people prefer the old or the new style:
`m.print_current_values(fancy=False)` or `m.print_current_values_old()`. The latter is the previous one.

francisco-dlp:
@thomasaarholt, is this ready for review?

thomasaarholt:
Yes :)

thomasaarholt:
I've made the code a bit more elegant by reusing it between the model and component value printing. Also added a better filter for active/non-free components/parameters, better tests and improved and corrected the documentation.

thomasaarholt:
Rebased the commit history as it was rather noisy. Also fixed a bug in an example in the docs (1e5 is a float and the numpy random `size` takes an int)

It would be nice if I could get feedback on whether people prefer the html view or the terminal view on print_current_values().

Currently the old and new views look like this:
**Old:**
```
Components	Parameter	Value
Gaussian
		A	100102
		centre	9.99969
		sigma	0.0100337
```
**New terminal view:**
```
Model1D:  histogram
Gaussian: Gaussian
Active: True
Parameter Name |  Free |      Value |        Std |        Min |        Max
============== | ===== | ========== | ========== | ========== | ==========
             A |  True | 100005.863 | 242.212919 |        0.0 |       None
         sigma |  True | 0.00999744 | 2.79595677 |       None |       None
        centre |  True | 9.99979784 | 2.79595557 |        7.0 |       14.0
```
**New fancy view:**
image

ericpre:
I find there are too many digits and the value are not easy to read, what about using a formatting such as `.6g`?

```python
>>> a = 43249.324
>>> f"{a:.6g}"
'43249.3'

>>> a = 0.0000043249324
>>> f"{a:.6g}"
'4.32493e-06'

>>> a = 0.43249324
>>> f"{a:.6g}"
'0.432493'

>>> a = 43249324
>>> f"{a:.6g}"
'4.32493e+07'
```

Other than that, it looks good to me, I have merged with upstream to solved the conflict.

thomasaarholt:
That's great, thanks for doing that. I agree with you on the formatting, and .6g looks like a good choice.

thomasaarholt:
I've refactored some of the code to make it more legible and compact, and implemented your suggested `"{:6g}"` format.

thomasaarholt:
Note that the `"{:6g}"` formatting is only implemented for the "fancy" view (where it was definitely needed), and not for the terminal view, where the spacing is fixed and therefore more difficult to adjust.

This is what it looks like now (by modifying the example to include some random "extreme" values):
image

PR #1608, opened 2017-05-02
Improve vmin and vmax argument of plot_images

ericpre:
This PR allows to pass a list of 'vmin' and 'vmax' values to set individually the limit value of each image when plotting multiple images with plot_images.


hyperspy/drawing/utils.py
```diff
@@ -724,10 +749,6 @@ def plot_images(images,
     ax_im_list = [0] * len(isrgb)
     # Loop through each image, adding subplot for each one
     for i, ims in enumerate(images):
-        # Get handles for the signal axes and axes_manager
-        axes_manager = ims.axes_manager
```
ericpre: I removed these few lines, because I think these are doing nothing! If case I am missing something, please yell at me and I will put it back! ;)

to266: The purpose of these lines is to create new `axes_manager` for iterating over the signal when there is more than one navigation element. In this case it's done by simply creating an identical signal pointing to the same data.

The purpose of __that__ is to avoid triggering most (maybe even all) of the events when iterating the original `axes_manager`, avoiding:
 - Changing the navigation position (`im.axes_manager.indices`) of the plotted signal. Iterating would leave it in the last index (this is easy)
 - If the original signal is plotted, such iteration would also "flip through" it, thus significantly slowing down everything. (because the original axes_manager has events connected to plot updating) (the main reason)

ericpre: Thanks for the explanation @to266. When I have got a bit of time I will come back to this.

francisco-dlp:
@ericpre, I have reverted the issue spotted by @to266 and merged Rnm, should I go ahead with merging this?

francisco-dlp:
Merging as it is passing the tests and it LGTM.

ericpre:
Thanks @francisco-dlp! Sorry, I just came back yesterday evening and started to look at it on my way back but I got stuck with an issue with matplotlib and the qt backend on my laptop... Happy to see that CI does not have this issue! :)

francisco-dlp:
Thanks @ericpre and sorry for the rush. I have a small window of opportunity to release v1.3 today...

PR #1609, opened 2017-05-03
Fix Kramers-Kronig Analysis bug

thomasaarholt:
@SlytherinEELS showed that the kramers kronig analysis failed on single spectra. That should be fixed now, but I would prefer if we could test it + write a test.


francisco-dlp:
I've labeled it as WIP because the test is missing.

thomasaarholt:
I'm going to need someone else to help write a test - I've not done any kramers-kronig before.

francisco-dlp:
The fix looks good. I'll merge it without the test and open an issue about the missing test.

francisco-dlp:
Here is the test issue #1633

PR #1640, opened 2017-05-25
Fix figure aspect

francisco-dlp:
#1580 introduced a bug that was leading to 0-size figures in some cases. This takes #1580 back to a previous iteration that doesn't present the issue.


francisco-dlp:
Merging so that tests pass again.

Issue #1392, opened 2017-01-19
Test docstrings


francisco-dlp:
Currently the docstrings are not tested i.e. they can't be trusted.

magnunor:
Is there an easy way to run them as tests?

to266:
https://docs.python.org/3.5/library/doctest.html

ericpre:
Some docstrings will need a bit of formating, but in principle it should work with [nose and the `--with-doctest` option](http://nose.readthedocs.io/en/latest/plugins/doctests.html) or [pytest and the `--doctest-modules` option](http://doc.pytest.org/en/latest/doctest.html) (in case we move to py.test #1372).

PR #1647, opened 2017-05-26
Fix example code in user guide

ericpre:
Since it is possibly a bit late to setup testing documentation (see #1392) for the next release, I checked manually some part of the user guide and fixed the issues.


francisco-dlp:
Wonderful, thanks!

PR #1648, opened 2017-05-26
Avoid spikes_removal_tool raising an error when ipywidgets_gui is not selected

ericpre:
When ipywidgets_gui is not selected in preferences, the test raises an error. I create a decorator, because it could possibly be used for other tests.


francisco-dlp:
I think that it shouldn't raise an error when ipywidgets is not enabled to start with.

ericpre:
Yes, I was assuming that the spikes removal tool needs a gui, but this is actually not the case. I changed accordingly and now it catches the error and only raise a warning.
The test was broken only after importing `hyperspy.api`, otherwise it was working fine. Is it normal?

hyperspy/signal_tools.py
```diff
@@ -930,6 +930,14 @@ def find(self, back=False):
                 # This is only available for traitsui, ipywidgets has a
                 # progress bar instead.
                 pass
+            except ValueError as error:
```
francisco-dlp, line 933: I would simply always add the error message to the logger as this check is weak: if we ever change the text of the error message it'll get broken.

PR #1662, opened 2017-05-31
FEI EMD reader

TomSlater:
### Description of the change
Add support to read FEI emd files from Velox.
An initial check is performed in the file_reader function to distinguish FEI emd files. If the 'Velox' tag is found then proceeds to load file as FEI format through the new FeiEMDReader class.

### Progress of the PR
- [x] Reads spectra, images and spectrum images,
- [x] fast reading of SI using numba,
- [x] read each individual EDS detector,
- [x] support reading individual frame or range of frames,
- [x] support energy rebin during loading,
- [x] add documentation,
- [x] tidy up,
- [x] ready for review.

### For another PR
- add minimum example file for different file version -to be provided by @robindegen?
- improve reading speed by reading spectrum stream in parallel?
- improve reading speed by using "FrameLocationTable" dataset? 
- add lazy support.


ericpre:
Fix #698.

hyperspy/io_plugins/emd.py
```diff
@@ -404,14 +421,426 @@ def log_info(self):
...
+        for data_sub_group in _get_keys_from_group(image_group):
+            self.dictionaries.append(
+                self._read_image(image_group[data_sub_group]))
+
+    def _read_image(self, data_sub_group):
+        """ Return a dictionary ready to parse of return to io module"""
+
+        read_stack = (self.read_SI_image_stack or self.im_type == 'Image')
+        if read_stack:
+            data = np.rollaxis(np.array(data_sub_group['Data']), axis=2)
```
to266: can instead use `np.transpose`, which has the dask equivalent:
```python
if lazy:
    transpose = da.transpose
    data = da.from_array(data_sub_group['Data'], ) # any other required arguments? 
    # Probably the dask (lazy) array will be created at some point before this anyway
else:
    transpose = np.transpose
    data = data_sub_group['Data']
data = transpose(data, (2,0,1)
```

ericpre:
I have merged RELEASE_next_patch into this branch to check if the tests are passing. It makes the diff is a bit more complicated but I guess RELEASE_next_patch will be merged into RELEASE_next_minor sooner or  later, so it should not be a problem?

francisco-dlp:
I just updated Rnm. Thanks for the reminder.

francisco-dlp:
I think that you'll need to merge the current Rnm in again to simplify the diff.

ericpre:
Yes, now the diff looks better.

to266 (commented):
Just a quick look (actually I wrote the comment some time ago, but forgot to "finish" the review....

This may move it a bit closer towards lazy loading. I will try to have a better look what's missing for that relatively soon, if this is not rushed. Otherwise it could come as a separate PR

k8macarthur:
Does this function extract the spectrumstream? As in each frame of spectra separately? I'm trying to understand from FEI how to extract this part but the guy helping me seems to have gone quiet...

ericpre:
Yes, it does but we forgot to update the docstring accordingly... so it is not very obvious. Have a look at the `__init__` method of the `FeiEMDReader` class. Also make sure, you have pint>=0.8 when you checkout this branch (during a checkout the dependency are not checked if I remember correctly).

k8macarthur:
So a first attempt at trying to use this, this afternoon results in the error: 
AssertionError                            Traceback (most recent call last)
<ipython-input-6-ea48ec9f40b5> in <module>()
----> 1 s = hs.load()

c:\program files\anaconda3\lib\site-packages\hyperspy\io.py in load(filenames, signal_type, stack, stack_axis, new_axis_name, lazy, **kwds)
    241             objects = [load_single_file(filename, lazy=lazy,
    242                                         **kwds)
--> 243                        for filename in filenames]
    244 
    245         if len(objects) == 1:

c:\program files\anaconda3\lib\site-packages\hyperspy\io.py in <listcomp>(.0)
    241             objects = [load_single_file(filename, lazy=lazy,
    242                                         **kwds)
--> 243                        for filename in filenames]
    244 
    245         if len(objects) == 1:

c:\program files\anaconda3\lib\site-packages\hyperspy\io.py in load_single_file(filename, signal_type, **kwds)
    284                                 reader=reader,
    285                                 signal_type=signal_type,
--> 286                                 **kwds)
    287 
    288 

c:\program files\anaconda3\lib\site-packages\hyperspy\io.py in load_with_reader(filename, reader, signal_type, **kwds)
    293     lazy = kwds.get('lazy', False)
    294     file_data_list = reader.file_reader(filename,
--> 295                                         **kwds)
    296     objects = []
    297 

c:\program files\anaconda3\lib\site-packages\hyperspy\io_plugins\emd.py in file_reader(filename, log_info, lazy, **kwds)
    407 def file_reader(filename, log_info=False,
    408                 lazy=False, **kwds):
--> 409     emd = EMD.load_from_emd(filename, lazy)
    410     if log_info:
    411         emd.log_info()

c:\program files\anaconda3\lib\site-packages\hyperspy\io_plugins\emd.py in load_from_emd(cls, filename, lazy)
    321         # One node should remain, the data node (named 'data', 'signals',
    322         # 'experiments', ...)!
--> 323         assert len(node_list) == 1, 'Dataset location is ambiguous!'
    324         data_group = emd_file.get(node_list[0])
    325         if data_group is not None:

AssertionError: Dataset location is ambiguous!

k8macarthur:
What version of Velox are you trying to code this for? My data was taken on version 4.

k8macarthur:
For reference my data structure looks like this: 
image

ericpre:
Have you checked out the correct branch?! ;)

k8macarthur:
I don't know, where's the correct branch?

ericpre:
https://github.com/TomSlater/hyperspy/tree/fei_emd_reader

You can find the link at the bottom of the messages, just above the "checks box".

k8macarthur:
That's the one I looked at.

k8macarthur:
Ok, so I tested this on my data this afteroon. All three signals seem to have extracted well  and in the correct shape. However, only the first frame in the time series appears to be in the data (based on the signal to noise). The names are also off, but I think this is a minor point.
```
s = hs.load()
s

[<Signal2D, title: SI SuperX-HAADF-DF4 1834, dimensions: (|1024, 1024)>,
 <Signal2D, title: SI SuperX-HAADF-DF4 1834, dimensions: (|1024, 1024)>,
 <EDSTEMSpectrum, title: SI SuperX-HAADF-DF4 1834, dimensions: (1024, 1024|4096)>]

**Comment by ericpre** [comment]
By default, it should be returning the sum of all frames and as it is now, you can read a frame range using the argument: `first_frame` and `last_frame`. It may be worth adding another argument to load each frame individually.

To avoid making unecessary big dataset, you can also used the argument `energy_rebin` and `SI_dtype`. The documentation needs some tweaks!!

**Comment by k8macarthur** [comment]
I agree, integration should probably be the default. But at least for my purposes it would be very helpful to have an option to import the whole time frame series (obviously not as default).

**Comment by k8macarthur** [comment]
So if I understand correctly, s = hs.load(first_frame=10, last_frame=20) should give me sum of 10 frames? Currently, the sum of the resulting data cube is the same as that for hs.load()

**Comment by k8macarthur** [comment]
Great progress @ericpre however, I guess when moving on  importing each frame individually this will really need to be lazified to make the data actually manageable? Maybe add a warning on data sizes if this option is selected until then?

**Comment by ericpre** [comment]
Indeed, supporting lazy import will be good... however, because of the nature of the spectrum stream, I suspect it will need quite a bit of refactoring, so I would leave it for now.
Regarding data size, I updated the documentation. The user will need to explicitly set some options to get large amount of data, so I would say a warning is not really necessary.

**Comment by phyxie** [comment]
Hi All,
I'm pretty new to hyperspy, but I've been following this thread as I'm running a Titan STEM using Velox and I'd like to utilise hyperspy for more advanced data analysis (particularly EDX spectral images).
We're recently been updated to the latest version (2.1.0.1226-19132be, build date 2017-07-20), and I'm having issues opening .emd spectral images and spectra.
I thought I'd check and see which version the current .emd reader is associated with, and if there's anything I can contribute in the way of sample data as we have regular software updates from FEI.
I'm getting an error similar to the one that k8macarthur reported, i can post it if it will help.

Regards,
Ash

**Comment by ericpre** [comment]
@phyxie, please make sure you have pulled the last changes of this branch. If you still have issues, the  simplest is to paste the error traceback or to share the reluctant files!

File: hyperspy/io_plugins/emd.py [hunk 2]
```diff
@@ -404,14 +421,581 @@ def log_info(self):
...
+                self.dictionaries.append({'data': stream.spectrum_image,
+                                          'axes': axes,
+                                          'metadata': md,
+                                          'original_metadata': stream.original_metadata,
+                                          'mapping': self._get_mapping()})
+
+    def _get_dispersion_offset(self, original_metadata):
+        for detectorname, detector in original_metadata['Detectors'].items():
+            _logger.debug('Detector: %s' % detector['DetectorName'])
+            if detector['DetectorName'] == 'SuperXG21' or detector['DetectorName'] == 'SuperXG11':
```
robindegen: For various reasons it's better to check for SuperXG2, since this detectorname may not always be there depending on your microscope configuration and settings.

hyperspy/io_plugins/emd.py
```diff
@@ -404,14 +421,581 @@ def log_info(self):
...
+    def get_pixelsize_offset_unit(self, stream_index=0):
+        om_br = self.stream_list[stream_index].original_metadata['BinaryResult']
+        return om_br['PixelSize'], om_br['Offset'], om_br['PixelUnitX']
+
+
+class FeiSpectrumStream(object):
+
+    """
+    Below some information we have got from FEI:
+    'The SI data is stored as a spectrum stream, ‘65535’ means next pixel,
```
robindegen: We call these markers "Gate pulses"

hyperspy/io_plugins/emd.py
robindegen: You shouldn't assume that SuperXG21 is available as a detector name for 2 reasons:
1. In case of the SuperXG2 Lite detector (2 segment), these are called SuperXG2Lite3 and SuperXG2Lite4.
2. Segment 1 may be missing as a detector all together if it was (temporarily) disabled in the microscope configuration.

It's better to use ```BinaryResult.Detector``` to determine the key here. At least for G2, the name of the segment used to acquire this stream is in that field. So if you have independently acquired segments, each stream will have a unique segment name here.

hyperspy/io_plugins/emd.py
```diff
@@ -404,14 +421,581 @@ def log_info(self):
...
+        # each of them
+        for image_sub_group_key in _get_keys_from_group(image_group):
+            self.dictionaries.append(
+                self._read_image(image_group, image_sub_group_key))
+
+    def _read_image(self, image_group, image_sub_group_key):
+        """ Return a dictionary ready to parse of return to io module"""
+        image_sub_group = image_group[image_sub_group_key]
+        original_metadata = _parse_metadata(image_group, image_sub_group_key)
+        self.detector_name = original_metadata['BinaryResult']['Detector']
```
robindegen: With the upcoming release of Velox 2.2, we also save Elemental Maps in images. This means that every image here will have SuperX as name.

hyperspy/io_plugins/emd.py
```diff
@@ -404,14 +421,581 @@ def log_info(self):
...
+    def parse_acquisition_settings(self, acquisition_settings_group):
+        acquisition_settings = json.loads(
+            acquisition_settings_group.value[0].decode('utf-8'))
+        self.bin_count = int(acquisition_settings['bincount'])
+        if self.data_dtype is None:
+            self.data_dtype = acquisition_settings['StreamEncoding']
+        return acquisition_settings
+
+    def get_spectrum_image(self):
+        stream = self.stream_data
```
robindegen: Starting from Velox 2.1, in the same group as the stream, we also save a "FrameLocationTable" dataset. This contains the offset of the beginning of each frame within the spectrum stream. This can used to quickly load subset range of frames from the stream.

hyperspy/tests/io/test_emd.py
```diff
@@ -159,7 +177,159 @@ def teardown_method(self, method):
         remove(os.path.join(my_path, 'emd_files', 'example_temp.emd'))
 
 
-if __name__ == '__main__':
-
-    import pytest
-    pytest.main(__name__)
+class TestFeiEMD():
```
robindegen, line 180: Let me know if you need small test files for different configurations like SuperXG2Lite and DualX

ericpre:
@robindegen, this is great that you can help on this and thanks for your comments!

Yes, it would be good to have a few files (not too large!) for different detector configuration. In particular, an example of the bare minimum required to write an velox emd file -in case we want to add support for exporting emd file compatible with velox.

robindegen:
@ericpre No problem. I'll see if i can generate a few small files from our simulation environment. They will be in the upcoming Velox 2.2 file format (version 6). 

From what I can see, I don't expect breaking changes here, since most of the changes are related to storage of overlays and annotations.

hyperspy/io_plugins/emd.py
ericpre: Very good that you tell us this, since it will be very useful to load the data lazily!!

francisco-dlp:
Is this still WIP?

ericpre:
I have updated the first post of this thread to summarise the status of this PR. I think, this is in fairly good shape, we may need to tidy up the code abit but this should already be working nicely. I put a few question marks on future improvements (lazy support and speed) we could possibly do in a separate PR.

hyperspy/io_plugins/emd.py
robindegen: With upcoming release of Velox 2.3, due to technical reasons, the BinaryResult:Detector field is no longer guaranteed to be filled in all cases. If it's not filled in, there will be only 1 detector under "Detectors"; which is the used detector for the acquisition.

francisco-dlp: Thanks for the info. Do you a document describing the format specification?

robindegen: Unfortunately no. The format is still in flux while we're trying to stabilise it more and more. Once we feel it's at a stable point, we will update documentation. Feel free to ask me any questions though; I know all the little details about the EMD file format.

francisco-dlp: Great, thanks!

ericpre: > With the upcoming release of Velox 2.2, we also save Elemental Maps in images. This means that every image here will have SuperX as name.

@robindegen, where is the name of each elemental map saved?

robindegen: @ericpre There's a few ways you can implement this. I'll share the simplest method, because the "real" way requires you to parse almost everything in the file (json text arrays). If you're interested in that approach we should probably do a chat/mail session at some point.

Anyway:
Under the ```/Presentation/Displays/ImageDisplay/``` group there are a whole bunch of text arrays containing Json. Other than the color mix display, there will be a display for each element that was selected in the periodic table, plus for each selected STEM detector. 

The Json looks like this (example):
```
{
    "display": {
        "id": "Co",
        "label": "Co",
        "overlays": "",
        "priority": "27",
        "closable": "false",
        "visible": "true"
    },
    "dataPath": "\/Data\/Image\/9a522c7075e94dcbbd781e55abf9e36d",
    "seriesIndex": "0",
    "zoom": "1",
    "pan": {
        "x": "0",
        "y": "0"
    },
    "angle": "0",
    "cropMode": "Off"
}
```

The ```dataPath``` will point to one of the images, and the ```display.label``` will be the name of the element or STEM detector respectively. 

Note that label is a URL encoded string, which can contain UTF-8 characters. For example it can contain %C2%B5, which translates to µ (Micro).

Again, be aware that this is a little bit of a trick, as this is not really how Velox interprets this data. This should however work for pretty much all images, even for DPC or Ceta experiments. There will always be a somewhat descriptive title under "Label".

Hope this helps.

ps.
Interestingly enough I noticed that ```display.priority``` seems to be equal to the element number. Don't use this however, as this is more of a coincidence than anything and is very likely to change.

ericpre: Thanks for your quick reply. I have seen this but I was a bit disappointed to notice that it doesn't tell us what energy line it is.... Is it necessary to parse all the other metadata (it seems that there is table of all elements with the corresponding energy line used for mapping) to just get the energy line?

robindegen: @ericpre You can use the ImageQuantificationRecord dataset for that.  There should be one at ```/Operations/ImageQuantificationOperation/(guid)```. This is again a text array that contains Json. In there you will find ```elementProperties```.

There are element properties (and thus line selections) saved for all elements that Velox supports.
There can only be 1 image per element, and line selections are tied to that element. You can't have a seperate elemental map image for each line for example (Velox doesn't currently support this).

So for a specific element, you can look here at what the line selections are. The selected lines for a specific element are saved as xray transitions (iupac style).

Excerpt from this portion of the Json:
```
  "elementProperties": [
        {
            "element": "1",
            "xrayTransitions": [
                {
                    "L3": "K"
                }
            ],
            "atomicShellFamily": "K",
            "quantify": "true",
            "includeInAutoPeakId": "true",
            "defaultLineSelect": "true"
        },
        {
            "element": "2",
            "xrayTransitions": [
                {
                    "L3": "K"
                }
            ],
            "atomicShellFamily": "K",
            "quantify": "true",
            "includeInAutoPeakId": "true",
            "defaultLineSelect": "true"
        },
        {
            "element": "3",
            "xrayTransitions": [
                {
                    "L3": "K"
                }
            ],
            "atomicShellFamily": "K",
            "quantify": "true",
            "includeInAutoPeakId": "true",
            "defaultLineSelect": "true"
        },
```

Hope this helps.

hyperspy/io_plugins/emd.py
```diff
@@ -404,14 +432,707 @@ def log_info(self):
...
-    emd = EMD.load_from_emd(filename, lazy)
-    if log_info:
-        emd.log_info()
     dictionaries = []
-    for signal in emd.signals.values():
-        dictionaries.append(signal._to_dictionary())
+    if fei_check(filename) == True:
+        _logger.debug('EMD is FEI format')
+        if lazy:
+            raise ValueError('Lazy loading is not supported for FEI EMD file.')
```
francisco-dlp: better to raise `NotImplementedError` in this case

hyperspy/io_plugins/emd.py
```diff
@@ -404,14 +432,707 @@ def log_info(self):
...
+            self._check_im_type()
+            self._parse_metadata_group(f.get('Operations'), 'Operations')
+            if self.im_type == 'SpectrumStream':
+                self.p_grp = f.get('Presentation')
+                self._parse_image_display()
+            self._read_data(load)
+
+    def _read_data(self, load):
+        self.load_images = self.load_SI = self.load_spectrums = True
+        if load == 'spectrums':
```
francisco-dlp: Maybe "single_spectra" instead of "spectrums"?

hyperspy/io_plugins/emd.py
```diff
@@ -404,14 +432,707 @@ def log_info(self):
...
+                    v = json.loads(
+                        subgroup[subgroup_key].value[0].decode('utf-8'))
+                    sub_dict[subgroup_key] = v
+            else:
+                sub_dict = json.loads(subgroup.value[0].decode('utf-8'))
+            d[group_key] = sub_dict
+        self.original_metadata.update({group_name: d})
+
+    def _read_spectrum_image(self):
+        if not self.load_SI:
```
francisco-dlp: Why this redundant line? To me it is strange: we tell the function to read the SI and it decides internally and silently that it won't do it...

hyperspy/io_plugins/emd.py
```diff
@@ -404,14 +432,707 @@ def log_info(self):
...
+        if self.individual_frame:
+            frame_time = float(original_metadata['Scan']['FrameTime'])
+            axes.append({'index_in_array': i,
+                         'name': 'Time',
+                         'offset': 0,
+                         'scale': frame_time,
+                         'size': spectrum_image_shape[i],
+                         'units': 's',
+                         'navigate': True})
+            i += 1
```
francisco-dlp: As this is not iterating, `i = 1` may be clearer

hyperspy/io_plugins/emd.py
```diff
@@ -404,14 +432,707 @@ def log_info(self):
...
+                      'offset': offset,
+                      'scale': dispersion,
+                      'size': spectrum_image_shape[i + 2],
+                      'units': 'keV',
+                      'navigate': False}])
+
+        md = self._get_metadata_dict()
+        md['Signal']['signal_type'] = 'EDS_TEM'
+
+        if self.individual_detector is False:
```
francisco-dlp: I think that `sum_detectors_data` may be clearer that `individual_detector`

hyperspy/io_plugins/emd.py
```diff
@@ -404,14 +432,707 @@ def log_info(self):
...
+    def _read_individual_stream(self, stream_data, key):
+        stream = FeiSpectrumStream(stream_data, **self.kwargs)
+
+        acquisition_key = '{}/AcquisitionSettings'.format(key)
+        acquisition_settings_group = self.spectrum_stream_group[acquisition_key]
+        stream.parse_acquisition_settings(acquisition_settings_group)
+
+        metadata_group = self.spectrum_stream_group[key]
+        stream.parse_metadata(metadata_group)
+#        stream._memory_check()
```
francisco-dlp: ?

hyperspy/io_plugins/emd.py
```diff
@@ -404,14 +432,707 @@ def log_info(self):
...
+        self.kwargs = {'shape': self.shape,
+                       'energy_rebin': self.energy_rebin,
+                       'first_frame': self.first_frame,
+                       'last_frame': self.last_frame,
+                       'individual_frame': self.individual_frame,
+                       'data_dtype': self.data_dtype}
+
+        if individual_detector:
+            self.streams = self._read_all_streams(subgroup_keys)
+            #self.frame_number = self.streams.frame_number
```
francisco-dlp: ?

hyperspy/io_plugins/emd.py
```diff
@@ -404,14 +432,707 @@ def log_info(self):
...
+        else:
+            stream_data_list = [
+                self.spectrum_stream_group['{}/Data'.format(key)][:].T[0]
+                for key in subgroup_keys]
+            # Read the first stream
+            self.streams = [self._read_individual_stream(stream_data_list[0],
+                                                         subgroup_keys[0])]
+            self.sum_spectrum_image = self.streams[0].spectrum_image
```
francisco-dlp: Maybe `summed_spectrum_image` or `total_spectrum_image`? Otherwise it looks like a method or an attribute.

hyperspy/io_plugins/emd.py
```diff
@@ -404,14 +432,707 @@ def log_info(self):
...
+    def get_spectrum_image(self):
+        stream = self.stream_data
+        shape = self.shape
+        if self.last_frame is None:
+            self.last_frame = int(
+                np.ceil((stream == 65535).sum() / (shape[0] * shape[1])))
+        if self.bin_count % self.energy_rebin != 0:
+            raise ValueError('The `energy_rebin` needs to be a divisor of the',
+                             ' total number of channels.')
+        self.frame_number = self.last_frame - self.first_frame
```
francisco-dlp: Maybe `number_of_frames` instead?

hyperspy/misc/utils.py
```diff
@@ -159,7 +159,7 @@ def slugify(value, valid_variable_name=False):
     value = value.translate(None, _slugify_strip_re_data).decode().strip()
     value = value.replace(' ', '_')
     if valid_variable_name is True:
-        if value[:1].isdigit():
+        if value.isdigit():
```
francisco-dlp: I don't agree with this "1a" is not a digit but is not a valid variable name

doc/user_guide/io.rst
```diff
@@ -625,15 +626,65 @@ as navigator in the plotting.
...
+.. warning::
+
+   This format is still not stable and files generated with the most recent version of Velox may not be supported.
+
+.. _Extra-loading-arguments-fei-emd:
+
+Extra loading arguments
++++++++++++++++++++++++
+
+- `load` : one of {'all', 'images', 'spectrums', 'spectrum_image'} (default is 'all')
```
francisco-dlp: I would provide more details on what each of these keyword arguments do. At the moment the purpose of some parameters are describe in the intro, that of others after the list and that of some others not described at all.

doc/user_guide/io.rst
```diff
@@ -625,15 +626,65 @@ as navigator in the plotting.
...
+- `first_frame` : integer (default is 0)
+- `last_frame` : integer (default is None)
+- `individual_frame` : boolean (default is False)
+- `individual_detector` : boolean (default is False)
+- `energy_rebin` : integer (default is 1)
```
francisco-dlp: Maybe `rebin_energy`?

doc/user_guide/io.rst
```diff
@@ -625,15 +626,65 @@ as navigator in the plotting.
...
+.. _Extra-loading-arguments-fei-emd:
+
+Extra loading arguments
++++++++++++++++++++++++
+
+- `load` : one of {'all', 'images', 'spectrums', 'spectrum_image'} (default is 'all')
+- `first_frame` : integer (default is 0)
+- `last_frame` : integer (default is None)
+- `individual_frame` : boolean (default is False)
+- `individual_detector` : boolean (default is False)
```
francisco-dlp: Maybe `sum_detectors` instead?

doc/user_guide/io.rst
```diff
@@ -625,15 +626,65 @@ as navigator in the plotting.
...
+- `energy_rebin` : integer (default is 1)
+- `SI_dtype` : numpy dtype (default is None)
```
francisco-dlp: What does `None` do?

doc/user_guide/io.rst
```diff
@@ -625,15 +626,65 @@ as navigator in the plotting.
...
+- `read_SI_image_stack` : boolean (default is False)
+
+The `load` specifies the type of data to load. The `SI_dtype` and `energy_rebin` are particularly useful in combination with `individual_frame=True` to reduce the data size when one want to read the individual frame.
```
francisco-dlp: `load` *parameter*

doc/user_guide/io.rst
francisco-dlp: Also explain that individual detectors are returned as individual signals and individual frames as an extra signal dimension

hyperspy/io_plugins/emd.py
```diff
@@ -404,14 +432,707 @@ def log_info(self):
...
+                      'scale': dispersion,
+                      'size': spectrum_image_shape[i + 2],
+                      'units': 'keV',
+                      'navigate': False}])
+
+        md = self._get_metadata_dict()
+        md['Signal']['signal_type'] = 'EDS_TEM'
+
+        if self.individual_detector is False:
+            self.dictionaries.append({'data': streams.sum_spectrum_image,
```
francisco-dlp: Can you have at the same time individual streams and individual detectors?

hyperspy/io_plugins/emd.py
```diff
@@ -404,14 +432,707 @@ def log_info(self):
...
+
+        acquisition_key = '{}/AcquisitionSettings'.format(key)
+        acquisition_settings_group = self.spectrum_stream_group[acquisition_key]
+        stream.parse_acquisition_settings(acquisition_settings_group)
+
+        metadata_group = self.spectrum_stream_group[key]
+        stream.parse_metadata(metadata_group)
+#        stream._memory_check()
+
+        stream.get_spectrum_image()
```
francisco-dlp: As the method doesn't return anything, what about `compute_spectrum_image` instead?

francisco-dlp (commented):
In addition to the mainly minor style comments, I think that it would be good to update the `hs.load` docstring. Other than that, LGTM :)

doc/user_guide/io.rst
```diff
@@ -633,15 +634,92 @@ as navigator in the plotting.
...
+
+.. _Extra-loading-arguments-fei-emd:
+
+Extra loading arguments
++++++++++++++++++++++++
+
+- `select_type` : one of {None, 'image', 'single_spectrum', 'spectrum_image'} (default is None). 
+- `first_frame` : integer (default is 0).
+- `last_frame` : integer (default is None)
+- `individual_frame` : boolean (default is False)
```
francisco-dlp: Would `sum_frames` be clearer?

hyperspy/misc/elements.py
```diff
@@ -3581,3 +3581,14 @@
...
+# read dictionary of atomic numbers from HyperSpy, and add the elements that
+# do not currently exist in the database (in case anyone is doing EDS on
+# Ununpentium...)
+atomic_number2name = dict((p.General_properties.Z, e) for (e, p) in elements_db)
+atomic_number2name.update({93: 'Np', 94: 'Pu', 95: 'Am', 96: 'Cm', 97: 'Bk',
+                           98: 'Cf', 99: 'Es', 100: 'Fm', 101: 'Md', 102: 'No',
+                           103: 'Lr', 104: 'Rf', 105: 'Db', 106: 'Sg',
+                           107:'Bh', 108: 'Hs', 109: 'Mt', 110: 'Ds', 
+                           111: 'Rg', 112: 'Cp', 113: 'Uut', 114: 'Uuq',
+                           115: 'Uup', 116: 'Uuh', 117: 'Uus', 118: 'Uuo'})
```
robindegen: Highest element that the EMD file format supports is 119: Uue.

robindegen:
General note for Velox 2.3 and up: For Spectrum Integration files (STEM+EDS) there may be EMD files that do not contain any Spectrum Stream data at all, but only a Spectrum Image (which is a Thermo Fisher proprietary format). This can happen under 2 circumstances:

1. The file is not written directly by Velox, but other Thermo Fisher software that also supports the EMD format.
2. The file was "pruned". This is an upcoming feature for the EMD format that allows for stripping the file down to only the bare essential data, which can reduce the filesize from multiple gigabytes, to a few megabytes. This is very useful for sharing files over the internet etc, but it completely removes all spectrum stream data.

francisco-dlp:
Today I've given implementing lazy loading for this file format a try by creating a data wrapper with numpy style indexing support. It seems to work fine but I won't have time to tidy it up into a PR until Thursday. Would the authors of this PR prefer that I send the PR to this branch or should I send it to the main repository once this one gets merged?

ericpre:
@robindegen: thanks for the info! We don't use the Spectrum Image data, what is in there exactly?

@francisco-dlp: It is up to you, personally I would suggest to merge this one first.

robindegen:
@ericpre A spectrum image is a proprietary data format that we use for loading SI/STEM EDS files quickly. We do not require the spectrum streams to load the file. Unfortunately I can not share any information about it's internal structure. However it might be the case that somebody has a SI/STEM EDS file that does not contain any spectrum streams in the future, though this is not the default option to not write them, I expect it to be relatively rare.

francisco-dlp:
@robindegen, thanks for the info. @ericpre, what about adding a message to the log when loading a file that has a spectrum image but not the corresponding stream explaining that the SI data is in a proprietary format that hspy cannot load and to suggest saving the file without pruning it in Velox to workaround the issue?

hyperspy/tests/io/test_emd.py
```diff
@@ -159,7 +177,252 @@ def teardown_method(self, method):
...
+
+        s = load(os.path.join(self.fei_files_path,
+                              'fei_SI_SuperX-HAADF_10frames_10x50.emd'),
+                 sum_frames=False,
+                 last_frame=5,
+                 SI_dtype=np.uint8,
+                 rebin_energy=256)
+        signal = s[1]
+        assert isinstance(signal, EDSTEMSpectrum)
+        assert signal.axes_manager.navigation_shape == (10, 50, 5)
```
ericpre, line 315: The navigation axes are in the following order (x, y, time). Would it be better to have it as (time, x, y) instead?

francisco-dlp: Personally I like it as it is because it keeps the different x coordinates nearby in memory and the different frames far apart—why do you think that it'll be better the other way around?

ericpre: I have no preference, this is why I am mentioning/asking it!

ericpre:
Thanks @to266, @robindegen and @francisco-dlp for all the useful comments!!! :)

PR #1663, opened 2017-05-31
Fix fit_component multifit and add test

### LINKED ISSUE #1661: fit_component only_current=True not working
fit_component does not multifit when I set `only_current = False`. Has the method changed in any way? Would be nice to write a test to catch this in the future.

thomasaarholt:
Fixes #1661. @francisco-dlp could you take a look and comment on why the code was the way it was set-up? Seems odd.

### DIFF
--- a/hyperspy/models/model1d.py
+++ b/hyperspy/models/model1d.py
@@ -38,8 +38,6 @@ from hyperspy.misc.utils import signal_range_from_roi
 @add_gui_method(toolkey="Model1D.fit_component")
 class ComponentFit(SpanSelectorInSignal1D):
 
-    only_current = t.Bool(True)
-
     def __init__(self, model, component, signal_range=None,
                  estimate_parameters=True, fit_independent=False,
                  only_current=True, **kwargs):
@@ -56,6 +54,7 @@ class ComponentFit(SpanSelectorInSignal1D):
         self.estimate_parameters = estimate_parameters
         self.fit_independent = fit_independent
         self.fit_kwargs = kwargs
+        self.only_current = only_current
         if signal_range == "interactive":
             if not hasattr(self.model, '_plot'):
                 self.model.plot()
--- a/hyperspy/tests/model/test_fit_component.py
+++ b/hyperspy/tests/model/test_fit_component.py
@@ -153,3 +153,25 @@ class TestFitSeveralComponent:
                                    m(),
                                    rtol=self.rtol,
                                    atol=10e-3)
+
+class TestFitSI:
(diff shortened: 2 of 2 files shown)


thomasaarholt:
I do not understand why the deconvolution tests are failing in Rnp? I certainly haven't changed them. They are failing in my local Rnm as well.

ericpre:
This seems to be related to the version of pytest because they are passing with pytest=3.1.0 and are failing with 3.1.1.

thomasaarholt:
`3.0.7` here, and it fails. Windows.

thomasaarholt:
The code runs fine manually, albeit with some warnings:

```python
class TestFourierRatioDeconvolution:

    def test_running(self):
        s = hs.signals.EELSSpectrum(np.arange(200))
        s_ll = hs.signals.EELSSpectrum(np.zeros(100))
        s_ll.data[18:23] = [3, 10, 20, 10, 3]
        s_ll.axes_manager[0].offset = -20
        s.fourier_ratio_deconvolution(s_ll)
        
a = TestFourierRatioDeconvolution()
a.test_running()

/Users/thomas/hyperspy/hyperspy/_components/power_law.py:131: RuntimeWarning: divide by zero encountered in true_divide
  r = 2 * log(I1 / I2) / log(x2 / x1)
/Users/thomas/hyperspy/hyperspy/_components/power_law.py:131: RuntimeWarning: invalid value encountered in log
  r = 2 * log(I1 / I2) / log(x2 / x1)
```

Running it like this works fine in the test:
```python
def test_thomas(self):
        s.set_signal_type("EELS")
        s.fourier_ratio_deconvolution(s)
```

ericpre:
This is surprising that it also fails with pytest 3.0.7, because I can reproduce nicely and this looks consistent with CI over different commits. Appveyor with pytest 3.0.7 is passing in this PR, while travis with pytest 3.1.1 is failing. In another PR (#1665), both travis and appveyor are failing with pytest 3.1.1.

ericpre:
Yes, that is it. This is because of the `RuntimeWarning` as for the other failing tests, however in this case the error traceback is wrong and this is why it was confusing.

thomasaarholt:
I didn't really understand your last comment - anything we should do?

ericpre:
The warnings make the tests failing and this is why `test_running` is failing and `test_thomas` is not. This needs to be sorted out in another PR. Maybe also report the wrong error traceback to pytest.

thomasaarholt:
Understood. thanks!

PR #1668, opened 2017-06-13
Resize pointer in plot

ericpre:
~~This is an attempt and work in progress to implement resizing the pointer when plotting signal. There are a few things broken (that could be fixed) but before going forward I would like to know if this is the right approach.~~
In the current version, the widget is changed to a resizeable one and the `__call__` of the signal class method is changed to take the sum of the ROI in a way that no new signal is created (to keep it fast). The widget is not properly connected, which mean that the signal get updated only when the axes_manager is changed (basically only when the top left pixel is changed).

~~Initially, I didn't wanted to change the `AxesManager` and the `DataAxis`, but now I think that adding an `width` attribute to the `DataAxis` would be possibly be neater and I suspect this is also be better for slicing and for the slider.~~

Please let me know what do you think!

### Progress of the PR
- [x] use resizeable rectangle as pointer for 2D navigation,
- [x] use horizontal resizeable spaner as pointer for 1D navigation: ~~only working with `navigator='spectrum'`,~~
- [x] add `pointer_operation` parameter to `plot` method with a default set to `np.sum`,
- [ ] update navigation sliders,
- [x] update docstring,
- [ ] update user guide,
- [x] add tests,
- [x] ready for review.

### To discuss
- [x] ~~use vertical resizeable spaner for 1D navigation and 1D signal? It is not straightforward to get it work in this case, but I may come back to it...~~ done in #2010.
- [x] set appropriate behaviour for model plotting? For model plotting, the resizeable pointer is disable.


hyperspy/signal.py
```diff
@@ -1924,8 +1924,18 @@ def _get_undefined_axes_list(self):
...
         if axes_manager is None:
             axes_manager = self.axes_manager
-        return np.atleast_1d(
-            self.data.__getitem__(axes_manager._getitem_tuple))
+        if getattr(self._plot, 'pointer', None) is not None:
+            indices = self._plot.pointer.indices
+            size = self._plot.pointer.get_size_in_indices()
+            array_slices = [slice(None, None, None)] * \
+                self.axes_manager.navigation_dimension
+            for i, axis in enumerate(self.axes_manager.navigation_axes[::-1]):
```
to266: You should not iterate over all navigation axes if the pointer is only defined for the two. I suggest getting the axes from the pointer itself (and store them there if they are not currently).

Additionally, no need to create a `[slice(None)]*nav_dim` list which is then completely replaced. You could, for example, use the `list(axes_manager._getitem_tuple)` and replace just the indices the pointer changes.

ericpre: Thanks for looking at this!
Do you think is worth trying to add a `width` attribute to `DataAxis` and get the slice from the axes directly? For example, in this attempt, when you close the navigation plot and create it again, the pointer is not properly restored because the size of the pointer is not saved.

to266: I'm not sure what you mean by the `width`. The width of the full axis? Or just the width of the pointer that is currently slicing?

Now that I think about it, I quite like the idea of storing (instead of just the indices of the current "position") spans/slices of the selection independently for each axis. This would allow visually interactively slicing 3-4-5-nD navigation spaces. For example, plot as `<1,2| 3,4,5>`, set the regions of interest for (1,2), then plot as `<2,3|1,4,5>` - the width from (2) is still preserved, and you can adjust (3,) to match whatever you want, and then continue onwards. I quite like such proposal.. 

The immediate drawback seems to be not slicing too many axes when recreating the default behavior, but that will have to be resolved with sane defaults.

@francisco-dlp what do you think?

ericpre: I mean the width of the pointer and therefore the slice in the corresponding axis (I think!). Yes, what your describe is more or less what I was thinking of, but I am not familiar enough with the details of the axes_manager and axis to know if this will have issue or if it does not fit the design of all this.
This is why I am asking about this!

francisco-dlp:
This is a feature that has been requested several times, hence I am sure that it is welcome by many. However, I am not sure that the best way to implement this is to change the behaviour of the pointer.

The fact that the pointer size can be changed with the +/- keys confuses many users into thinking that it performs integration when it doesn't. This addresses the issue by making it behave as expected. However, I think that implementing it in this way adds extra complexity (e.g. ``s()`` now changes meaning and functionality that depends on this may not work as expected when the pointer becomes a ROI) and misses the opportunity of implementing the feature in a more flexible way.

At the end of the day, this is a shortcut for something like this:

```python
%matplotlib tk
import hyperspy.api as hs
import numpy as np
s = hs.signals.Signal1D(np.random.random((10, 32,32,100)))
s.plot()
roi = hs.roi.RectangularROI(0,0,30,30)
w = roi.add_widget(s)
sr = hs.interactive(roi, signal=s, event=roi.events.changed )
ss = hs.interactive(sr.sum, axis=sr.axes_manager.navigation_axes[:2])
ss.plot()
```

And this is indeed a lot of code for what is a very common operation. So, I think that offering a shortcut is very useful. But, what about adding a roi option to ``sum``, ``std`` etc to perform the operation? For example

```python

ss = s.sum(axis=(0, 1), roi=True) # Automatically get an appropriate ROI or set of ROIs and display the figures. A RectangularROI in this case.
ss2 = ss.sum(axis=(0, 1), roi=(hs.roi.SpanROI(0, 10), hs.roi.SpanROI(0, 10)) # Customise the ROIs

```

ericpre:
Yes, adding a roi option to `sum` and others is very good idea, but for another PR!! ;)

Nevertheless, having a re-sizeable pointer by default (without having to write one single line of code) will be very handy and convenient. I think, that it is really worth it and I am sure we all have seen the disappointed face of new hyperspy user after he or she noticed that the interaction with the pointer is quite poor.
The way to properly address this, is not straightforward, thought... if not, I guess it would have been already done!!!

The main question is then: do we want to have the default pointer to be a ROI? If so, what should be the expected behaviour (i.e. the default)?

Regarding the flexibility, the behaviour change and the added complexity, I guess that these are (more or less) details which can be sorted out in the implementation. For example, could the `__call__` method take an option to define what operation to perform on the pointer?
I am not very happy with the current approach and I suspect that adding a `width` attribute to `DataAxis` could be better.

francisco-dlp:
> Nevertheless, having a re-sizeable pointer by default (without having to write one single line of code) will be very handy and convenient.

Good point!

> The way to properly address this, is not straightforward, thought... if not, I guess it would have been already done!!!

This was true until @vidartf contributed the ROIs. Now it's just a matter of discussing about it and reaching consensus.

My gut feeling is that changing the default behaviour of ``__call__`` can lead unexpected issues. Adding a ``width`` attribute to ``DataAxis`` would result in a sort of "virtual binning" feature and it'll require to change the "virtual size" of the navigation space when iterating over a signal. One can see the appeal of an interactive virtual binning feature, but it would only be useful for small signals as it'll be a lot slower than the standard binning (which, by the way, can be performed lazily). So, I would rather go for your suggestion of adding an option to ``__call__`` and ``get_current_signal`` with default to the standard behaviour and not support this virtual binning for integration. Following @to266's idea, also  the navigator pointer could be resizable and the sliders could be replaced by range sliders in order to support summing over all navigation axes.

vidartf:
When implementing ROIs, having integrated/mean functionality was one of the main motivations. I won't comment on what the default behavior should be, but if it decided that an ROI should be the default, then care need to be taken such that the implementation only affects the default signal plot:
 - It does not make any sense to have an ROI for a model plot (what does the model fit?).
 - It should not affect the signal that is returned when iterating over the signal, as this will affect a lot of other functionality.

ericpre:
I tried to do most of the things suggested above.
- the default of signal plotting is using the resizeable pointer, when possible.
- in the plot, it is possible to pass an option to choose the operation (numpy math operation) to perform over the pointer.
- since the default of the `__call__` and the `get_current_signal` is set to return values for single pixel (as before), it should not have any consequences on anything else if it is not explicitly coded.
- the model plotting is not changed. In the current version, this is possible to use a resizeable pointer for plotting only the corresponding option is explicitly passed to the plot method (not sure, it does make sense but it does not cost much for sure!). It could be interesting to initialise some fitting parameter by averaging over specific area, but this is possibly for another PR. Along the same line, it could be very interesting to fit the background by averaging over neighbouring pixel, this is not something new and it would be a very nice feature to have!

There are quite a few files changed in drawing but I have tried to make small commits with detailed commit comments, so you can more or less follow what I was doing.

@francisco-dlp, could you please give me some guidance on how to change the navigation sliders to range sliders? I am a bit lost with how this is done...

hyperspy/drawing/signal1d.py
```diff
@@ -282,28 +294,31 @@ def color(self, color):
...
         if self.line is not None:
             self.line.remove()
         self.line, = self.ax.plot(self.axis.axis, data,
                                   **self.line_properties)
         self.line.set_animated(self.ax.figure.canvas.supports_blit)
-        self.axes_manager.events.indices_changed.connect(self.update, [])
-        self.events.closed.connect(
-            lambda: self.axes_manager.events.indices_changed.disconnect(
-                self.update), [])
+#        self.axes_manager.events.indices_changed.connect(self.update, [])
```
ericpre: I have commented this, because in `Signal1DFigure`, the line(s) is/are already been updated when indices of the `axes_manager` are changed. This connection is then not necessary and it was actually redrawing twice the same thing. I think, that this makes sense but I would prefer to make sure that I haven't miss something!

francisco-dlp:
@ericpre, could you merge Rnm into this.

Also, wouldn't it be good for consistency to have a resizeable pointer for line spectra too i.e. turning the current line into a line span?

ericpre:
I have solve the merge conflicts and update the first comment of this PR to summarise the changes and what is left to do/consider.

ericpre:
This is ready to review.

As mentioned above having range sliders (ipywidgets and traitsui) would be very useful, but with the current implementation, I can't find a way to link the widgets value with the actual range. It should be easy to do with adding a `width` and/or `range` attribute to the `DataAxis` class, though. I am happy to do this because I think it would be a very convenient feature.

ericpre:
Related discussion on gitter on using ROI as pointer:
https://gitter.im/hyperspy/hyperspy?at=5ddd1bf7f65fec1c8e726d55

PR #1673, opened 2017-06-21
Release next patch

jan-car:


jan-car:
Oh, I didn't even notice that I did a pull request, I originally just wanted to push the newest HyperSpy version to my branch and must have clicked one too many times... Just as an explanation for the minimal difference and the lack of a description :-) ...

PR #1674, opened 2017-06-22
Add roi argument to math operation

ericpre:
As discussed in #1668, this PR add a roi option to most of the mathematical operations.

- [x] add roi argument to most of the signal operation:
  - sum, mean, max, min, std, var (and their nan* conuterpart), indexmin, indexmax and valuemax,
  - default roi=None, same behaviour as before,
  - roi=True, add a roi on the plot, the signal need to be plotted, raise an error if not. If no axis provided, default is to add a roi on the navigation plot.
  - roi=hyperspy.roi.BaseROI instance or list of such instances: for command line use, no plot or interactivity needed.
  - if several rois are provided, return the operation over all the rois. 
- [x] operation can now take the 'signal' or 'navigation' string for the axis argument to perform the operation on the signal or navigation only
  - in combinaison with roi=True, add a roi on the signal or the navigation plot
- [x] update docstring
- [x] ~~this is still a bit too slow, need to tweak when the events are triggered to minimise computation/redraw events~~: ~~remove the `Signal1DLine` update connection, since the `Signal1DFigure` update is already doing the job~~ The `Signal1DLine` update connection is necessary to update the line of the right pointer: put it in a if loop. (plotting speed improved by ~1.5)
- [x] add tests


francisco-dlp:
@ericpre, could you merge Rnm into this?

ericpre:
I have solved the merge conflicts. The lazy tests should fail but before fixing these, it would be worth checking if the approach is OK.

PR #1675, opened 2017-06-24
Fix tests failing after a few new dependency releases

ericpre:
After new numpy 1.13, dask 0.15 and pytest 3.1.1, a few tests were failing.
- ~dask is pinned at 0.14.3, not great but at least for the moment the tests are passing, so still an minor improvement! I guess this is better to leave this one to @to266?! ;) Apparently, they have a few API changes.~
- fix some tests, which were raising a warning and pytest is regarding these as failure.
- increase the maximal number of open matplotlib figure from 20 (default) 25 to avoid displaying a warning in pytest log. Some tests are actually opening such a large amount of figure, but there are properly closed afterwards, so this is fine. If the message appear again, then the plotting tests need to be checked.


ericpre:
Anybody to look at this? I think that it would be good to have the tests working with the most recent releases of the dependencies, so that travis and appveyor can pass!
Since the last push to this PR, sympy 1.1 have been released and there is a test failing with the expression component if used with substitution (see [this travis log](https://travis-ci.org/ericpre/hyperspy/builds/246502632)). I have opened issue #1686 ~~and I guess that it could be fixed in another PR, because I don't know how to fix it correctly.~~

to266:
@ericpre could you point to tests that fail due to the new dask version? I can't find it, and would rather not pin it -- they introduced some excellent and very useful features

ericpre:
I reverted the commit, which pinned dask to 0.14.3, so it should be clear what is going on.

ericpre:
hyperspy/tests/signal/test_eels.py is failing, but not on windows!

to266:
https://github.com/ericpre/hyperspy/pull/5

hyperspy/tests/mva/test_bss.py
```diff
@@ -120,10 +120,14 @@ def test_diff_axes_string_with_mask(self):
...
         self.s.learning_results.factors[5, :] = np.nan
         factors = self.s.get_decomposition_factors().inav[:3]
+        mask.change_dtype("float")
+        mask_diff = mask.diff(axis="x", order=1)
+        mask.change_dtype("bool")
+        mask_diff.change_dtype("bool")
         self.s.blind_source_separation(
             3, diff_order=0, fun="exp", on_loadings=False,
             factors=factors.diff(axis="x", order=1),
-            mask=mask.diff(axis="x", order=1))
```
to266: It seems to work for me.. Are you sure this is needed?
```python
In [1]: import numpy as np

In [2]: np.__version__
Out[2]: '1.13.1'

In [3]: ar = np.ones((3,2), dtype=bool)

In [4]: ar
Out[4]: 
array([[ True,  True],
       [ True,  True],
       [ True,  True]], dtype=bool)

In [5]: np.diff(ar, axis=0)
Out[5]: 
array([[False, False],
       [False, False]], dtype=bool)

In [6]: np.diff(ar.astype(float), axis=0).astype(bool)
Out[6]: 
array([[False, False],
       [False, False]], dtype=bool)

```

ericpre: Yes, it works with 1.13.1, but not 1.13.0... Considering numpy can be annoying to install from source, should we leave as it is or exclude 1.13.0 but keep the code more simple?

ericpre: I reverted this change and exclude numpy=1.13.0 in setup.py. Since the time between 1.13.0 and 1.13.1 was fairly short, this is very unlikely that someone got stuck with numpy 1.13.0!

francisco-dlp:
LGTM, @to266?

PR #1689, opened 2017-07-12
save(overwrite=False) now saves if no existing file

hentr:


hyperspy/io.py
```diff
@@ -478,6 +478,9 @@ def save(filename, signal, overwrite=None, **kwds):
         ensure_directory(filename)
         if overwrite is None:
             overwrite = overwrite_method(filename)
+        if overwrite is False:
```
to266: This "fails" (as in does not overwrite) silently, and also might confuse future developers.

I would prefer this to be formatted along the lines:
```python
if overwrite is False and os.path.isfile(filename):
    raise ValueError("some appropriate text")
# If we get here, either we overwrite or the file does not exist
# Put the code that in your PR starts with line #485 
# (i.e. without the `if overwrite is True`)
```

to266 (changes requested):
Also a simple test is needed

thomasaarholt:
@hentr Would you be kind enough to add @to266's suggestion?

hentr:
will do, but little time, maybe next week

hyperspy/io.py
hentr: I am a bit confused, are you suggesting replacing lines 481-483 with the above? 

When I test, this PR works correctly for all four combinations: 
1) overwrite = True, file exists --> file is overwritten
2) overwrite = True, file doesn't exist --> new file is written
3) overwrite = False, file exists --> do nothing
4) overwrite = False, file doesn't exist --> new file is written

I am not able to interpret the Travis CI and AppVeyor reports, but if the problem is not doing anything at 3), then raising an error is no better. I was using `save(overwrite=False)` because I wanted no interruptions when looping a process over many files creating multiple output files for each, not knowing whether I had made one of output files before. 

Better would be (starting at line 481): 
```
if overwrite is False:
    if os.path.isfile(filename):
        pass
    else: 
        overwrite = True
```
or?

hentr: ah I understand what you mean, then: 
```
if overwrite is False and os.path.isfile(filename):
   pass
else:
   write file... (line 485 onwards) 
```

ericpre: Travis and appveyor are failing because of other tests, which are now fixed in the upstream branch. You can merge upstream/RELEASE_next_patch into your branch if you want to get travis and appveyor passing, but you don't need to worry about this for now!

to266: @hentr yes, essentially, but not just `pass`. At the very least you should `_logger.warning(f"file {filename} exists, did not overwrite")`, or similar/better. We try to avoid silently passing things.

hentr: bah, was still not thorough enough, as inputting `overwrite='whatever'` was the same as `overwrite=True`, so now starting at line 479: 
```python
        if not(overwrite is True or overwrite is False):
            overwrite = overwrite_method(filename)
        if overwrite is False and os.path.isfile(filename):
            _logger.warning(f"file {filename} exists, did not overwrite")
        else:
            writer.file_writer(filename, signal, **kwds)
            ...
```

hentr:
is this waiting for me? I cannot find any way to review/do anything with the "Changes requested", which are now implemented.

ericpre:
Yes, I guess @to266 was also asking for a test!

hentr:
what/how test? I did at least test the four input logic combinations to work back in July, don't know what more to test?

ericpre:
We have a test suite, see the [dev guide](http://hyperspy.org/hyperspy-doc/current/dev_guide.html#write-tests).
You need to write a small function which does similarly things that you did manually earlier, but the test will do it automatically to check that it does get broken in the future by running the test suite. You can read about unit testing and if you want to know more.
For example, in this case, I would put this test in a new file (for example `hyperspy.tests.io.test_io.py`) and name it `test_io_overwriting()`.

ericpre:
I have simplified the logic and added tests.

hyperspy/tests/io/test_io.py
```diff
@@ -0,0 +1,79 @@
...
+
+import os
+import numpy as np
+import pytest
+from unittest.mock import patch
+
+from hyperspy.signals import Signal1D
+from hyperspy.io import load
+
+dirpath = os.path.dirname(__file__)
```
francisco-dlp: PEP8: module variables should be uppercase.

hyperspy/tests/io/test_io.py
```diff
@@ -0,0 +1,79 @@
...
+        self.new_s = Signal1D(np.ones(5))
+        # make sure we start from a clean state
+        self._clean_file()
+        self.s.save(fullfilename)
+
+    def _clean_file(self):
+        if os.path.exists(fullfilename):
+            os.remove(fullfilename)
+
+    def _check_file_is_written(self, signal, filename):
```
francisco-dlp: Checking in this way is a bit weak and may get broken for other reasons. What about using the hashlib instead to computer e.g. sha256 or simply verify the the modification date has changed in respect to the original?

francisco-dlp (commented):
Other than the little comments above, LGTM.

PR #1695, opened 2017-07-18
Reverse plot_spectra legend order

### LINKED ISSUE #1694: Ordering of legend in plot_spectra
When using the `plot_spectra` function with a padding > 0, the plots are plotted from the bottom up. When creating the legend from a list of labels, the items are added to the matplotlib legend from top down, leading to a non-intuitive labeling (especially if colors in the legend are repeated).

```python
matplotlib.style.use('seaborn-deep')
sigs = [hs.signals.Signal1D(np.arange(20)*(i/10)) for i in range(1,6)]
for s in sigs:
    s.axes_manager[0].name = 'X'
    s.axes_manager[0].units = 'arb.'
hs.plot.plot_spectra(sigs, style='cascade',
                     legend=['{}'.format(i) for i in range(1,6)])
plt.tight_layout()
```
image

I would propose that we either plot "top-down" in `plot_spectra`, or switch the ordering of the labels in the legend (likely easier).

jat255:
Added code to reverse the order of the legend when plotting with `plot_spectra`. Resolves #1694. I think this is a more intuitive default setting.

_Note_: I couldn't find a more succinct way to reverse the order of the legend than getting the handles and text values manually. I tried using `handles, labels = ax.get_legend_handles_labels()` as recommended [here](https://stackoverflow.com/questions/34576059/reverse-the-order-of-legend), but for some reason that was returning empty lists for the handles and labels, so the method that I implemented was necessary. I'm happy to take any suggestions as to a better way to do this.

Turns this:
image

Into this:
image


jat255:
Thinking about this more, the same could be achieved by setting `padding` to `-1.0`, but I think the current "bottom-up" ordering would make most sense to users by default.

jat255:
Builds are failing because the test plots need to be updated.

francisco-dlp:
Looks good to me, but updating the test images is required. @magnunor, @ericpre, what do you think?

jat255:
Ok, updated the reference plots. All the plotting tests passed on my local machine, so I think they're okay.

hyperspy/drawing/utils.py
```diff
@@ -1081,6 +1081,10 @@ def plot_spectra(
                            line_style=line_style,)
         if legend is not None:
             plt.legend(legend, loc=legend_loc)
+            l = ax.get_legend()
```
ericpre: Put this in an function with a comment explaining what is it doing would make the code slightly more readable?

ericpre:
Reserving the plot order (top-bottom) by using the `padding` solution has the advantage not to introduce more code and to be consistent with the matplotlib default. However, if one want to manually change the order, the legend will not be reversed... so this is not very good neither. I agree that the current implementation is better.

hyperspy/drawing/utils.py
jat255: Yup, makes sense. Just changed it to do this.

jat255:
@ericpre an alternative could be a flag option in `plot_spectra` that allows one to activate this code only when desired. Yet another option for the plotting method, but makes it more flexible.

ericpre:
Great, thanks @jat255! To my opinion, this is good as it is.

ericpre:
Link to the corresponding (closed) issue matplotlib/matplotlib#333 on this topic.

francisco-dlp:
@ericpre, why not merging it then?

PR #1698, opened 2017-07-20
ENH: Add "zero_fill" option to the background removal tool

jat255:
Add a new option `zero_fill` to `remove_background` that is off by default. If turned on, it will set the value of all channels prior to the lower bound of the fitting region to zero (like how DigitalMicrograph does it).


jat255:
Ok, this should be ready for review. It is working as expected (for me) on both normal and lazy signals, but some testing by others is certainly in order.

Goes along with the GUI parts in hyperspy/hyperspy_gui_ipywidgets#1 and hyperspy/hyperspy_gui_traitsui#1

hyperspy/_signals/signal1d.py
```diff
@@ -1046,6 +1056,7 @@ def remove_background(
                 signal_range=signal_range,
                 background_estimator=background_estimator,
                 fast=fast,
+                zero_fill=zero_fill,
```
ericpre, line 1090: At line 1038: similarly to here, I would also add the possibility to pass the "zero_fill" (also the "background_estimator" and the "fast", while you are at it!), in order to be able to preset the option from the notebook or the qtconsole.

ericpre (commented):
Looks all good (including the two other GUI PRs). I have just added one suggestion on passing argument.

hyperspy/_signals/signal1d.py
jat255: I'm not sure I understand what you mean. Do you mean in the call to the `BackgroundRemoval` constructor?

ericpre:
On a slightly different topic, do you also have the issue with the windows disappearing the first time a traitsui instance is used with a spanselector? I remember that similar issues have been reported before, but if I remember correctly, it was intermittent... now it looks to me that it is all the time, but only the first time!

jat255:
Yup, still happens when using the traits UI. I spent almost a whole day trying to figure out why it happened, but never did solve it.

ericpre:
Yes, I also spent a couple of hours on it... very annoying.

hyperspy/_signals/signal1d.py
ericpre: Yes! The idea is to be able to do, for example:
```python
s.remove_background(fast=False, zero_fill=True)
```
and get the option accordingly set in the UI.

jat255:
Somehow this fell by the wayside, but I've updated it to the latest RNminor, and added the class attribute, so this should be ready for review @ericpre

jat255:
The AppVeyor test failure again was because of HDF5. The remove_background tests seemed to run fine.

jat255:
@ericpre any chance you could take a look over this quickly?

francisco-dlp:
@ericpre, is this good to merge now?

ericpre:
@jat255, can you please solve the conflicts?

jat255:
Okay, should be GTG. Tests passed on my Linux VM. I can't run any tests on my Windows system, for some reason, but that's not related to this PR.

jat255:
https://github.com/hyperspy/hyperspy_gui_ipywidgets/pull/1 and https://github.com/hyperspy/hyperspy_gui_traitsui/pull/1 have been updated, too.

jat255:
@ericpre Can you take a look at the GUI PRs too? Those need to be merged in at the same time as this, preferably.

ericpre:
I can't merge them, I don't have write access to those repos.

jat255:
Makes sense! @francisco-dlp, help?

Issue #1693, opened 2017-07-14
Rectangle roi not working with non-square pixel


ericpre:
The following is not working:

```python
import hyperspy.api as hs
import numpy as np

s = hs.signals.Signal1D(np.arange(2 * 4 * 6).reshape(2, 4, 6))
s.axes_manager[0].scale = 0.5
#s.axes_manager[1].scale = 0.5

roi_nav = hs.roi.RectangularROI(0, 0, 1, 0.5)`
s.plot()
w = roi_nav.add_widget(s)
```
and produces the following error:

```python
Traceback (most recent call last):

  File "<ipython-input-1-83892adf2fe2>", line 12, in <module>
    w = roi_nav.add_widget(s)

  File "/home/eric/Python_dev/hyperspy/hyperspy/roi.py", line 447, in add_widget
    widget.set_mpl_ax(ax)

  File "/home/eric/Python_dev/hyperspy/hyperspy/drawing/widget.py", line 168, in set_mpl_ax
    self.select()

  File "/home/eric/Python_dev/hyperspy/hyperspy/drawing/widget.py", line 182, in select
    canvas.pick_event(mouseevent, self.patch[0])

  File "/opt/anaconda3/lib/python3.6/site-packages/matplotlib/backend_bases.py", line 1883, in pick_event
    self.callbacks.process(s, event)

  File "/opt/anaconda3/lib/python3.6/site-packages/matplotlib/cbook.py", line 554, in process
    proxy(*args, **kwargs)

  File "/opt/anaconda3/lib/python3.6/site-packages/matplotlib/cbook.py", line 416, in __call__
    return mtd(*args, **kwargs)

  File "/home/eric/Python_dev/hyperspy/hyperspy/drawing/widget.py", line 671, in onpick
    super(ResizableDraggableWidgetBase, self).onpick(event)

  File "/home/eric/Python_dev/hyperspy/hyperspy/drawing/widget.py", line 438, in onpick
    super(DraggableWidgetBase, self).onpick(event)

  File "/home/eric/Python_dev/hyperspy/hyperspy/drawing/widget.py", line 958, in onpick
    self.pick_offset = (x - self._pos[0], y - self._pos[1])

TypeError: unsupported operand type(s) for -: 'NoneType' and 'float'
```

If you uncomment the line:

```python
s.axes_manager[1].scale = 0.5
```
Then, this is working.

ericpre:
Slightly different issue:

```python
import hyperspy.api as hs
[...]

PR #1699, opened 2017-07-25
Fix a few bugs with rectangle and span roi and other tweaks.

ericpre:
Initially, I wanted to fix a bug with rectangle roi initialisation but at the end I did a bit more!

- [x] Fix bug initialisation roi (it was possible to add a roi up to the last index),
- [x] range widget stays within bounds (similarly to other widgets),
- [x] fix bug in `set_bounds` of the range widget (was using index instead of value),
- [x] replace `ValueError` in `set_bounds` by warning and limit the set value to the bound when out of the bounds,
- [x] add tests.


ericpre:
Fix the second issue in #1693 (see [second comment](https://github.com/hyperspy/hyperspy/issues/1693#issuecomment-315430059)).

francisco-dlp:
I think that the warnings can become annoying in some useful cases. For example consider the following:

```python
%matplotlib qt

import hyperspy.api as hs

import numpy as np

im1 = hs.signals.Signal2D(np.random.random((256, 256)))
im2 = hs.signals.Signal2D(np.random.random((512, 512)))
im1.plot()
im2.plot()

r = hs.roi.RectangularROI(0, 0, 128, 128)
w1 = r.add_widget(im1)
w2 = r.add_widget(im2)

```

When moving the widgets of `im2` out of the range of `im1` when get warnings every time that the widget parameters change. Wouldn't it be better to print the warning just the first time the ROI is updated to a position that is out of range for ``im1`` and print another one when it gets back to the valid range if ever?

francisco-dlp:
I like the idea of decreasing the logging level one notch for this as unnecessary warnings are know to unnerve some users. Would it be possible to add the id of the widget to the message to make it more informative?

Now, when taking ``w2`` out of ``im1`` range, ``w1`` size decreases to eventually surround one single pixel. I think that the previous behaviour (staying at the outer most possible coordinates for a given size) was better because:
1. For high-res images a rectangle surrounding a single pixel may be hard to see.
2. When using the resized rectangle to move the widget in ``im1`` the size stays reduced and it automatically resizes the size of the other widget.

ericpre:
Yes, it needs some more fiddling!
1: I am thinking of making a PR to add a handle when a widget is below a specific size. I have seen this in the last version of Velox and I think this is just a simple and good idea.
2: I am not sure if fully understand this point but indeed moving the widget should not change its size...

francisco-dlp:
Adding a handle when the widget is below a given size (in pixels, using mpl's display coordinate) is an excellent idea! Among other benefits it'll make the current rectangular pointer more intuitive. Currently it sorrounds a pixel and, when it's too little to handle, we use the ``+`` key to make it bigger, what misleads new users into thinking that it integrates in the ROI. Another solution is taking the DM approach of using a fixed-size-in-pixels handle instead, but this has the disadvantage of masking what's behind it. Automatically switching from one to the other solves the issue without drawbacks :) !

Regarding point 2, to get the point you can perform the experiment with the code in my comment above.

PR #1700, opened 2017-07-25
Fix dm reader when microscope info missing

francisco-dlp:
HyperSpy was failing to read dm file not containing the microscope info. This fixes it.

```python

>>> hs.load("1 CL.dm3")
~/Python/hyperspy3/hyperspy/io_plugins/digital_micrograph.py in _get_microscope_name(self, ImageTags)
    806         try:
--> 807             if ImageTags.Session_Info.Microscope != "[]":
    808                 return ImageTags.Session_Info.Microscope

~/Python/hyperspy3/hyperspy/misc/utils.py in __getattribute__(self, name)
    337         name = slugify(name, valid_variable_name=True)
--> 338         item = super(DictionaryTreeBrowser, self).__getattribute__(name)
    339         if isinstance(item, dict) and '_dtb_value_' in item and "key" in item:

AttributeError: 'DictionaryTreeBrowser' object has no attribute 'Session_Info'

<ipython-input-2-cb2b289e826d> in <module>()
----> 1 hs.load("1 CL.dm3")

~/Python/hyperspy3/hyperspy/io.py in load(filenames, signal_type, stack, stack_axis, new_axis_name, lazy, **kwds)
    241             objects = [load_single_file(filename, lazy=lazy,
    242                                         **kwds)
--> 243                        for filename in filenames]
    244 
    245         if len(objects) == 1:

~/Python/hyperspy3/hyperspy/io.py in <listcomp>(.0)
    241             objects = [load_single_file(filename, lazy=lazy,
    242                                         **kwds)
--> 243                        for filename in filenames]
    244 
    245         if len(objects) == 1:

~/Python/hyperspy3/hyperspy/io.py in load_single_file(filename, signal_type, **kwds)
    284                                 reader=reader,
    285                                 signal_type=signal_type,
--> 286                                 **kwds)
    287 
    288 

~/Python/hyperspy3/hyperspy/io.py in load_with_reader(filename, reader, signal_type, **kwds)
    302             if signal_type is not None:
    303                 signal_dict['metadata']["Signal"]['signal_type'] = signal_type
--> 304             objects.append(dict2signal(signal_dict, lazy=lazy))
    305             folder, filename = os.path.split(os.path.abspath(filename))
    306             filename, extension = os.path.splitext(filename)

~/Python/hyperspy3/hyperspy/io.py in dict2signal(signal_dict, lazy)
    432                 value = signal.original_metadata.get_item(opattr)
    433                 if function is not None:
--> 434                     value = function(value)
    435                 if value is not None:
    436                     signal.metadata.set_item(mpattr, value)

~/Python/hyperspy3/hyperspy/io_plugins/digital_micrograph.py in _get_microscope_name(self, ImageTags)
    808                 return ImageTags.Session_Info.Microscope
    809         except AttributeError:
--> 810             if 'Name' in ImageTags['Microscope_Info'].keys():
    811                 return ImageTags.Microscope_Info.Name
    812             elif 'Microscope' in ImageTags['Microscope_Info'].keys():

~/Python/hyperspy3/hyperspy/misc/utils.py in __getitem__(self, key)
    327 
    328     def __getitem__(self, key):
--> 329         return self.__getattribute__(key)
    330 
    331     def __setitem__(self, key, value):

~/Python/hyperspy3/hyperspy/misc/utils.py in __getattribute__(self, name)
    336             name = name.decode()
    337         name = slugify(name, valid_variable_name=True)
--> 338         item = super(DictionaryTreeBrowser, self).__getattribute__(name)
    339         if isinstance(item, dict) and '_dtb_value_' in item and "key" in item:
    340             return item['_dtb_value_']

AttributeError: 'DictionaryTreeBrowser' object has no attribute 'Microscope_Info'


```


hyperspy/io_plugins/digital_micrograph.py
```diff
@@ -807,10 +807,14 @@ def _get_microscope_name(self, ImageTags):
...
-            elif 'Microscope' in ImageTags['Microscope_Info'].keys():
-                return ImageTags.Microscope_Info.Microscope
+            try:
+                if 'Name' in ImageTags['Microscope_Info'].keys():
+                    return ImageTags.Microscope_Info.Name
+                elif 'Microscope' in ImageTags['Microscope_Info'].keys():
+                    return ImageTags.Microscope_Info.Microscope
+            except:
+                _logger.info("Microscope name not present")
+                return None
```
ericpre: Is it worth specifying?
```python 
return None
```
Isn't python doing it by default?

ericpre:
Maybe adding a test to check loading a file with a minimum set the metadata?

hyperspy/io_plugins/digital_micrograph.py
thomasaarholt: @ericpre is correct.

thomasaarholt:
For a test file, would it be enough to create a completely empty signal with nothing in it, or does it need to be a specific dm file? (like, a stem image or eels spectrum)

I remember creating an empty dm file with lots of ROIs for @magnunor, and that seemed to load fine?

ericpre:
No, I don't think that a completely empty file can work, because it possibly need to have a few required tags to trigger the bug.

thomasaarholt:
I can try saving a small EELS spectra, deleting some of the tags and loading that.

hyperspy/io_plugins/digital_micrograph.py
ericpre: What's about changing this to:
```python
def _get_microscope_name(self, ImageTags):
    # This function is called only when the 'Microscope_Info' tag is present
    if "Session_Info" in ImageTags.keys() and len(ImageTags.Session_Info.Microscope) > 0:
        return ImageTags.Session_Info.Microscope
    if 'Name' in ImageTags['Microscope_Info'].keys():
        return ImageTags.Microscope_Info.Name
    if 'Microscope' in ImageTags['Microscope_Info'].keys():
        return ImageTags.Microscope_Info.Microscope
    # in case none of them above works add this message to the logger.
    _logger.info("Microscope name not present.")
```
I find it easier to read.

francisco-dlp: I would suggest the following (non-equivalent code) instead:

```python
def _get_microscope_name(self, ImageTags):
    # This function is called only when the 'Microscope_Info' tag is present
    if "Session_Info" in ImageTags.keys() andImageTags.Session_Info.Microscope:
        return ImageTags.Session_Info.Microscope
    elif 'Name' in ImageTags['Microscope_Info'].keys() and ImageTags.Microscope_Info.Name:
        return ImageTags.Microscope_Info.Name
    elif 'Microscope' in ImageTags['Microscope_Info'].keys() and ImageTags.Microscope_Info.Microscope:
        return ImageTags.Microscope_Info.Microscope
    else:
        # in case none of them above works add this message to the logger.
        _logger.info("Microscope name not present.")
        return None
```

``return None`` is unnecessary, but it is easier to understand for somebody who doesn't know that Python does it for you, so I would leave it to the author's taste.

ericpre: Yes, I agree, that writing `return None` make the code more clear.
Regarding the `if "Session_Info" in ImageTags.keys() and ImageTags.Session_Info.Microscope:`, it can exist and be empty, which is the case by default. AFAIK it needs to be set by the user in the global info of DM. However, in case, this is an empty list or string it is worth reading the microscope name from the 'Microscope_Info' which has a default value when connected to a microscope (for example, something like "FEI Tecnai" or "FEI Tecnai Remote" on any FEI microscope including Titan and other.)
Actually, I am wondering if this is worth parsing `ImageTags.Microscope_Info.Microscope` since most of the time it will set to a wrong default value...

francisco-dlp:
Ready for review again.

hyperspy/io_plugins/digital_micrograph.py
```diff
@@ -806,14 +806,17 @@ def _get_date(self, date):
...
-            elif 'Microscope' in ImageTags['Microscope_Info'].keys():
-                return ImageTags.Microscope_Info.Microscope
+        locations = (
+            "Session_Info.Microscope",
+            "Microscope_Info.Name",
+            "Microscope_Info.Microscope",
+        )
+        for loc in locations:
+            mic = ImageTags.get_item(loc)
+            if mic not in (None, "[]"):
```
ericpre: This looks much better now.
I think that `if mic is not None and len(mic) > 0:` would be better to check that this is not an empty string or an empty list. This `"[]"` looks wrong to me, I have checked a few files (from different DM version) and I can't find any having.

francisco-dlp: You're right, I forgot the empty string.

Regarding "[]", there must be a reason for it. It would have been better to add a comment explaining the purpose of `"[]"`, but, as the chances of this becoming a microscope name are remote, leaving as is is harmless. I suggest to leave it just in case a particular DM version / DM setup marked this field as empty using [].

ericpre: Yes and it doesn't hear much anyway!
I have a file where I am getting the following (I possibly empty manually the tag in DM):
`└── microscope = []`
and adding an empty list in your tuple should fix it.

francisco-dlp: Merge it then?

ericpre: Sorry, I haven't been very clear. To avoid the unwanted behaviour mentioned above `└── microscope = []`, it needs to be changed to:
```python
if mic not in (None, "[]", "", []):
```

francisco-dlp: Curious, thanks! It should be fixed now.

hyperspy/io_plugins/digital_micrograph.py
```diff
@@ -813,7 +813,7 @@ def _get_microscope_name(self, ImageTags):
         )
         for loc in locations:
             mic = ImageTags.get_item(loc)
-            if mic not in (None, "[]", ""):
+            if mic and mic != "[]":
```
ericpre: Good to know that it can also work that way, nice.

PR #1701, opened 2017-07-26
Make warnings about missing GUI features optional

jan-car:
Added `preferences.GUIs.warn_if_guis_are_missing` option, which can be used to toggle warnings about missing installations of `hyperspy_gui_ipywidgets` or `hyperspy_gui_traitsui`.

This is just a small suggestions if someone decides to not install one of the gui interfaces and is annoyed by the warning each time hyperspy is imported.

Maybe it is better to still warn if none of the options are installed?

If there are arguments against this change, feel free to close this PR :-)...

### DIFF
--- a/doc/user_guide/getting_started.rst
+++ b/doc/user_guide/getting_started.rst
@@ -54,7 +54,15 @@ and the
 
 By default, HyperSpy warns the user if one of the GUI packages is not installed.
 These warnings can be turned off using the :py:class:`~.defaults_parser.Preferences`
-class (see :ref:`here <configuring-hyperspy-label>` for more information).
+class: (see :ref:`here <configuring-hyperspy-label>` for more information), or
+directly via:
+
+    .. code-block:: python
+
+       >>> import hyperspy.api as hs
+       >>> hs.preferences.GUIs.warn_if_guis_are_missing = False
+       >>> hs.preferences.save()
+
 
 Now you are ready to load
 your data (see below).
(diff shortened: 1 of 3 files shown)


magnunor:
I think this is a good idea, since as you say, the warning gets a bit annoying over time.

francisco-dlp:
Looks good. I would only suggest to add a little line to the user guide explaining that these warning can be disabled in preferences.

jan-car:
Hm, apparently, my markup additions do not result in nice cross references, any idea what I did wrong?

ericpre:
I have checked the generated html (with `make html`) and it looks alright.
Would it be worth explicitly mentioning the line that needs to be called?
`hs.preferences.GUIs.warn_if_guis_are_missing = False`
Even if you use the auto-completion, it is not straightforward to find!

jan-car:
I added a small code block to showcase how to turn off the warnings in a script.

doc/user_guide/getting_started.rst
```diff
@@ -52,6 +52,18 @@ that you have installed at least one of HyperSpy's GUI packages:
 and the
 `traitsui GUI <https://github.com/hyperspy/hyperspy_gui_traitsui>`_.
 
+By default, HyperSpy warns the user if one of the GUI packages is not installed.
+These warnings can be turned off using the :py:class:`~.defaults_parser.Preferences`
```
francisco-dlp: What about:

```rest

These warnings can be turned off using the :py:class:`~.defaults_parser.Preferences` GUI (see :ref:`here <configuring-hyperspy-label>` for more information) or programmatically as follows:

```

jan-car: Yes, that's better!

francisco-dlp:
LGTM, @ericpre if it LGTY too, just merge it.

PR #1702, opened 2017-07-26
Fixed small bug in holographic reconstruction during side-band search (now uses amplitudes of the Fourier transform instead of real part)

jan-car:
The sideband position is now determined from the amplitudes of the Fourier transform, rather than from its real part. This caused an error for the estimation of the sideband position for very low dose data.

Additionally added the option to search the left and right half (in addition to lower and upper half).
@FWin22  and @woozey, you should have a look at this PR!

TODOS:
- [x] Add a test for the extra feature (added by @woozey).
- [ ] Add a test that triggers the bug that this PR is fixing (reasonable?).


jan-car:
I did not cover the added functionality by the tests, maybe @woozey can point me where I should best place the according testfunctions?

francisco-dlp:
I have added it to Rnm because it contains a new feature/enhancement. If you would like the bug fix part to make it to Rnp, just split the last commit into a separate PR.

I have labeled it WIP as it requires:
* [ ] a test for the extra feature (as mentioned above)
* [ ] a test that triggers the bug that this PR is fixing

It'll also be good to either provide a better description of the bug that this is fixing in the header of this PR or open a separate issue to report the bug.

jan-car:
Thanks for the feedback, can I just change the target branch to `release_next_minor`? 
Concerning the test, I would like @woozey to help me out with that, I initially made this PR on @FWin22's behalf (maybe he could provide a short description of the bug, as well?).

FWin22:
The sideband position is now determined from the amplitudes of the Fourier transform, rather than from its real part. This caused an error for the estimation of the sideband position for very low dose data.

jan-car:
Thanks @FWin22 , I updated the titel and the first comment accordingly!
@woozey, maybe you can just send a pull request to my fork with the tests? Because you've written the last tests, I think you are most familiar with them.

woozey:
@francisco-dlp :
>I have labeled it WIP as it requires:
 a test for the extra feature (as mentioned above)

Done!
> a test that triggers the bug that this PR is fixing
It'll also be good to either provide a better description of the bug that this is fixing in the header of this PR or open a separate issue to report the bug.

Strictly speaking it isn't a bug, it is rather mathematically wrong implementation of determination of the sideband centre... The issue can be described in the following way:
The centre of sideband is the maximum of the **amplitude** of FFT of the hologram in the sideband region. The original implementation was assigning the centre of sideband to  the maximum of **real** part of FFT, which usually coincides with the maximum of the amplitude. Nevertheless there are some unusual conditions when the selection of the maximum of real part of FFT leads to wrong determination of the sideband centre.

Having in mind that it is not a classical bug, it is actually very difficult to reproduce the conditions in the test in which  the centre of sideband would be determined differently before and after the fix. So far I didn't manage to reproduce it by hologram simulation as it done in `test_hologram_image`. Honestly, I would rather leave this fix without additional test, since we know that this is mathematically correct way of doing it, and since the only way of testing it I can imaging so far is to use one of 4K images that we have acquired as the test sample. This option doesn't seem feasible to me due to space concerns. In principle I can of-course synthesis an artificial FFT image (**which won't be an FFT of any meaningful hologram**) where the maximum of real part will not coincide with maximum of amplitude. So  @francisco-dlp, @CodeMonkeyJan, shall we leave this without additional test or do you have any suggestion?

francisco-dlp:
Sounds reasonable as it is then, thanks for the explanation.

PR #1710, opened 2017-08-18
Trigger events only once when settings axes manager indices

### LINKED ISSUE #1708: Setting coordinates is currently inefficient
For example:

```python

s.axes_manager.indices = (20, 30)

```
would first set the index of the first nav axis to 20, trigger all the events and then set the index of the second nav axis and trigger all the events again. A better approach would be to trigger the common events just once.

ericpre:
Fix #1708.

### DIFF
--- a/hyperspy/tests/axes/test_axes_manager.py
+++ b/hyperspy/tests/axes/test_axes_manager.py
@@ -19,7 +19,7 @@
 
 from unittest import mock
 
-from hyperspy.axes import DataAxis, AxesManager
+from hyperspy.axes import AxesManager
 from hyperspy.signals import BaseSignal, Signal1D, Signal2D
 from numpy import arange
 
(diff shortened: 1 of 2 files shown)


to266:
Looks good, only needs a test

hyperspy/tests/axes/test_axes_manager.py
```diff
@@ -216,3 +216,39 @@ def test_changing_scale_offset(self):
             signal_axis1.low_value, signal_axis1.high_value,
         )
         assert signal_extent == s.axes_manager.signal_extent
+
+
+class CallCounter:
```
to266: While this is simple, I'd rather stick to already in-built methods to do this:
```python
In [1]: from unittest import mock

In [2]: m = mock.Mock()

In [3]: m()
Out[3]: <Mock name='mock()' id='140657786956880'>

In [4]: m.call_count
Out[4]: 1
```

hyperspy/tests/axes/test_axes_manager.py
```diff
@@ -216,3 +216,39 @@ def test_changing_scale_offset(self):
...
+
+    # both indices changed again, call only once
+    s.axes_manager.indices = (2, 3)
+    assert s.axes_manager.indices == (2, 3)
+    assert call_count.count == 2
+
+    # single index changed, call only once
+    s.axes_manager.indices = (2, 2)
+    assert s.axes_manager.indices == (2, 2)
+    assert call_count.count == 3
```
to266: Please also check setting coordinates as well

hyperspy/tests/axes/test_axes_manager.py
ericpre: Yes, indeed!

Issue #1276, opened 2016-08-27
General documentation suggestions


thomasaarholt:
I was looking through the documentation today and thinking of more general things we can do to improve it.
- We currently have quite a few "New in version X.X.X" text fields in the documentation. Quite a few are from v0.7 and v0.8. At what point should one remove these and let features be an assumed part of Hyperspy?
- In the table of contents, the number of version changelogs is becoming quite long and will just keep on growing. Do we want to either shorten that up to 0.5, 0.6, 0.7... (rather than 0.5, 0.5.1 etc), or remove them from the table of contents altogether (leaving "What's New" or changing it to "Changelog")?
- We still haven't got an example notebook database (#1003). It would be nice to create one and add it somewhere sensible in the documentation.

Are there any other suggestions?

magnunor:
> We currently have quite a few "New in version X.X.X" text fields in the documentation. Quite a few are from v0.7 and v0.8. At what point should one remove these and let features be an assumed part of Hyperspy?

Agreed. I guess we can only keep the N most recent versions?

> In the table of contents, the number of version changelogs is becoming quite long and will just keep on growing. Do we want to either shorten that up to 0.5, 0.6, 0.7... (rather than 0.5, 0.5.1 etc), or remove them from the table of contents altogether (leaving "What's New" or changing it to "Changelog")?

Yeah, I think it takes up too much space in the table of contents. In my opinion the "Getting started" headline should be visible without having to scroll, to make it a little bit easier for new users to find the starting information.

thomasaarholt:
Some more suggestions:

There is no section that describes how one should modify the documentation, and I think quite a few people new to developing hyperspy (and to developing in general) find it a bit scary.

Having just had to reinstall sphinx again, here are the prerequisites for building hyperspy docs

```
pip install sphinx
pip install sphinxcontrib.napoleon
pip install sphinx_rtd_theme
```

In order to build the docs, one navigates to hyperspy/docs in a local repository, and type `make html`. This builds the docs into the subfolder `hyperspy/docs/_build/html` and can be seen by opening `index.html`. A lot of red text does appear, and that is because we haven't been very thorough in fixing warnings, not because a new user has done something wrong.

ericpre:
For installing doc dependencies, this should be fixed in #1819.

thomasaarholt:
That's great! 

[...]

PR #1715, opened 2017-08-19
DOC, CHANGES.txt: reduce section level for older releases

magnunor:
Discussed here: https://github.com/hyperspy/hyperspy/issues/1276

Currently there are way too many versions under the What's new header in the user guide, and this takes up quite a large amount of vertical space. This means that the install and actual guide gets pushed far down the screen, making it harder for new users to find what they're looking for.

This pull request pushes the section level one step down for all versions older than 1.0.0, and sorts them under a new "Older versions" heading.

### DIFF
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -305,9 +305,12 @@ API changes
   classes deprecated the old `Spectrum` `Image` and `Signal` classes.
 
 
+Older versions
+==============
+
 
 v0.8.5
-======
+------
 
 
 This is a maintenance release. Follow the following links for details on all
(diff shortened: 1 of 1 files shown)


magnunor:
Any thoughts on this?

ericpre:
Yes, indeed it is now a bit long!
What's about keeping only the last version in the `What's new` section and adding a `Changelog` section at the same level than the `What's new` section, so that when there is a new release, we only need to move the content of the old `What's new` to the `Changelog` section without having to change the section heading.

```
What's new
**********
v1.3
====
Some new changes.

Changelog
**********
v1.2
====
Some old changes.
```

magnunor:
That is easier to maintain, since one doesn't have to update each header manually. I guess we would place the "Changelog" a bit further down in the index? Alternatively "hide" some of the version headers somehow?

ericpre:
Yes, I guess just below the `What's new` will be good enough and will be keep the code simple: just retrieve the text from the changes.txt file, as before. If you want to put it somewhere else, you may need to find a more sophisticated way to parse the text from the changes.txt file?
I think this is fine if the changelog is long in its own separate page, so no need to try to hide headers.

ericpre:
Fixed in #1798.

PR #1716, opened 2017-08-19
MRCZ io_plugin

robbmcleod:
Couple points to look at:

* both the mrc and mrcz plugins support the '.mrc' file extension. I'm guessing you pick the plugin by the 
  extension somewhere?
* The signal axes are correctly ordered but the mixed C/Fortran ordering is a little confusing.
* Why no `__getitem__` and `__setitem__` methods for the metadata attribute?


robbmcleod:
That's strange that TravisCI can't find `mrcz` for MacOSX.  It's a pure-python repo.

.travis.yml
```diff
@@ -33,7 +33,7 @@ before_install:
 install:
 
   - if [[ $MINIMAL_ENV == 'False' ]] ; then
-      DEPS="pip numpy scipy matplotlib>=2.0.2 ipython h5py sympy scikit-learn dill natsort setuptools scikit-image cython lxml ipyparallel dask traits traitsui";
+      DEPS="pip numpy scipy matplotlib>=2.0.2 ipython h5py sympy scikit-learn dill natsort setuptools scikit-image cython lxml ipyparallel dask traits traitsui blosc mrcz";
```
ericpre: These are the dependencies installed from the conda-forge channel. I cann't find any repo for the mrcz, so I guess it would be better to remove mrcz from here and leave pip install it automatically from pypi when hyperspy is installed.

appveyor.yml
```diff
@@ -139,7 +139,7 @@ before_deploy:
   - "ECHO %SCIKIT_IMAGE%"
   # - "%CMD_IN_ENV% wget http://www.lfd.uci.edu/~gohlke/pythonlibs/tuoh5y4k/%SCIKIT_IMAGE% --header User-Agent:Chrome/23.0.1271.97"
   - "%CMD_IN_ENV% pip install %SCIKIT_IMAGE%"
-  - "%CMD_IN_ENV% pip install --upgrade tqdm notebook cython ipython configobj start_jupyter_cm ipywidgets ipyparallel sympy pytest"
+  - "%CMD_IN_ENV% pip install --upgrade tqdm notebook cython ipython configobj start_jupyter_cm ipywidgets ipyparallel sympy pytest blosc mrcz"
```
ericpre: Same as for .travis.yml, install blosc from the conda-forge channel with conda and install mrcz with pip from pypi.

hyperspy/tests/io/test_mrcz.py
```diff
@@ -0,0 +1,150 @@
...
+#==============================================================================
+# MRCZ Test
+# 
+# Internal python-only test. Build a random image and save and re-load it.
+#==============================================================================
+tmpDir = tempfile.gettempdir()
+mrcName = os.path.join( tmpDir, "testMage.mrcz" )
+MAX_ASYNC_TIME = 2.0
+
+class PythonMrczTests:
```
ericpre: This test class is not discovered by pytest, the name needs to start with "Test".

ericpre (changes requested):
Just a few comments to get the tests working. I can come back to the rest later.

robbmcleod:
Ok, getting closer, JSON version problem now.

robbmcleod:
Looks good to me.

ericpre:
Indeed, it is working fine!
There are two more things worth looking at:
-  for lazy signal, hyperspy use dask. As it is currently, it is working, except for saving signal (the data are stored as a `dask.Array` instance). Do you think it would possible to support writing mrcz from `dask.Array` in `python-mrcz`?
-  this PR introduce a non pure python dependency (blosc), which can possibly brake hyperspy installation on some system. For example, I could not install blosc out of the box (https://github.com/Blosc/python-blosc/issues/127). I am not sure if this issue is solved.

I have tested the Gohlke's blosc binary in the Winpython HyperSpy bundle and it works.

robbmcleod:
I'll look into writing from a `dask.Array`, it's probably just a matter of calling `numpy.asarray()`.

For `blosc`, perhaps we can ask Matthew Brett for help.  I'll ask Francesc if he would like to move `blosc` into having `manywheels` support.

robbmcleod:
Also, quick point: MRCZ doesn't list blosc as a requirement because the import is inside a `try` block.  If `blosc` isn't present, compression is disabled, but it won't break anyone's install.

ericpre:
Ok great, then we will able to merge this PR even if the blosc wheels/binaries are not ready yet.
Regarding writing from `dask.Array`, I am not sure if this is relevant here but it may be worth at having a quick look at #1743.

robbmcleod:
I added support for saving from `dask.array.core.Array` in `python-mrcz 0.2.5`.  My tests suggest the ideal way would be to pull off the chunks one-by-one and compress them but `blosc` doesn't really support that. `c-blosc2` might but it's not Python-aware.  

    import dask.array as da
    import numpy as np
    from time import perf_counter
    import mrcz

    t0 = perf_counter()
    foo = da.ones( (50,2048,2048), chunks=(1,2048,2048) )
    t1 = perf_counter()
    np.array(foo)
    t2 = perf_counter()
    foo.__array__()
    t3 = perf_counter()
    for I in np.arange(foo.shape[0]):
        foo[I,:,:].__array__()
    t4 = perf_counter()
    mrcz.asyncWriteMRC( foo, 'foo.mrcz', compressor='lz4' )
    t5 = perf_counter()

    print( f"Create da time: {t1-t0:.2e} s" )
    print( f"Convert to array: {t2-t1:.3f} s" )
    print( f"Call __array__: {t3-t2:.3f} s" )
    print( f"Sliced: {t4-t3:.3f} s" )
    print( f"MRCZ save: {t5-t4:.2e} s" )

    >>> Create da time: 3.96e-04 s
    >>> Convert to array: 1.458 s
    >>> Call __array__: 0.888 s
    >>> Sliced: 0.497 s
    >>> MRCZ save: 1.16e-03 s

I'm not sure it's a good idea to use `asyncWriteMRC` because Dask is pure-python, but it might help if the user calls some GIL-releasing function afterward.

ericpre (approved):

ericpre:
This is great, thanks @robbmcleod for this nice contribution! I just added the tests for the lazy case and move blosc to `extras-require` to make it optional. Once appveyor and travis tests are passed, we can merge.

ericpre:
Looking at this [travis log](https://travis-ci.org/hyperspy/hyperspy/jobs/281405156), `mrcz` still depends on blosc.

robbmcleod:
Quick status update: we're in the process of getting `multiwheels` support for `blosc`.  Hopefully that will be ready by the end of the week.

ericpre:
@robbmcleod, since the master branch of mrcz now raises an `ImportError` when blosc is not installed, I have fixed the tests accordingly in this PR. If you release a new version of mrcz, then we can merge this PR. Do you have an idea when you will make a new release?

robbmcleod:
The master branch of `mrcz 0.3.4` does not raise an `ImportError` if `blosc` is missing, it swallows it and disables the compression tools.  The `mrcz.test()` unit tests do generate an error, so is that what you mean? 

I just tested it with a fresh conda install, and then installed `numpy` and then `mrcz` via pip and it seems to work as intended.

ericpre:
I am talking about [this version](https://github.com/em-MRCZ/python-mrcz/commit/ce67e4982b03f059ed7101e6aba400474d923e8b), which raises an `ImportError` (only when compression is needed). This is good because it informs the user to install blocs when it is required. If you test this branch with https://github.com/em-MRCZ/python-mrcz/commit/ce67e4982b03f059ed7101e6aba400474d923e8b, then it all works nicely: the tests catch the error when it is needed and blosc can be an optional dependency.

robbmcleod:
Oh, right, I forgot about that.  Let me clean up the docstrings so they meet NumPy standards and I'll release 0.3.5.

robbmcleod:
Eric,

The 0.3.5 release is live now:

https://github.com/em-MRCZ/python-mrcz/commit/3cd4ffe82ee2e4635a42d1eef448abbf16398227

https://pypi.python.org/pypi/mrcz

robbmcleod:
I see you had a random test failure on the asynchronous tests.  You might want to consider re-triggering those builds.  I haven't had trouble with them, but weird things can happen between when Python thinks the file write is finished, and when the OS thinks the file write is finished.  If you encounter this problem more than once, the only fix I could propose would be a small sleep in-between getting the result from the Future and reading.  That or disabling that test.

ericpre:
Yes, it did happen a few times, so I added (#1852) a sleep, as you recommended and so far it seems to do the job.

Issue #1066, opened 2016-06-09
Use matplotlib scalebar?


magnunor:
Currently HyperSpy has its own implementation of a scalebar https://github.com/hyperspy/hyperspy/blob/496e3234e916cefed72c21907b5e676f2c247539/hyperspy/drawing/_widgets/scalebar.py#L23.

Matplotlib has its own version (called a sizebar), which I improved a while ago to be more consistent with how scalebars usually look in electron microscopy publications. Would it be an idea to use the matplotlib implementation? https://github.com/matplotlib/matplotlib/blob/196f3446a3d5178c58144cee796fa8e8aa8d2917/lib/mpl_toolkits/axes_grid1/anchored_artists.py#L229

I'm not certain how the matplotlib sizebar would handle the interactive plots, but this could be nice for the functions which make "static" images.

vidartf:
Without looking to deeply into this: An alternative might be to simplify the Hyperspy scalebar widget to use the MPL sizebar as its internal patch. It should also then be fairly easy (?) to make it draggable.

ericpre:
An alternative would be to consider https://github.com/ppinard/matplotlib-scalebar.

hakonanes:
We use `mpl_toolkits.axes_grid1.anchored_artists.AnchoredSizeBar` (https://matplotlib.org/3.1.0/api/_as_gen/mpl_toolkits.axes_grid1.anchored_artists.AnchoredSizeBar.html) in `orix` (https://github.com/pyxem/orix/blob/master/orix/plot/crystal_map_plot.py#L207).

ericpre:
This is the approach https://github.com/hyperspy/hyperspy/pull/1718 is taking but _if I remember correctly_ matplotlib-scalebar has advantages for resizing the scalebar automatically when zooming interactively and should save us implementing logic for setting the size, etc.

jlaehne:
Done in #2561

PR #1718, opened 2017-08-20
Replace scalebar with matplotlib sizebar

magnunor:
https://github.com/hyperspy/hyperspy/issues/1066

Currently HyperSpy has its own scalebar implementation. It would be better to use the upstream matplotlib scalebar instead.

- [ ] Finish implementation
- [ ] Write tests


jlaehne:
Should be obsolete after implementation of #2561.

PR #1719, opened 2017-08-22
Fixed MRC plugin after NumPy update

### LINKED ISSUE #1705: Numpy Update Breaks MRC Reader Plugin
Changes to Numpy appear to be causing a TypeError to be thrown when using the MRC Reader plugin.  The error is: TypeError: only integer scalar arrays can be converted to a scalar index

This occurred somewhere between v1.11.3 and v1.13.1.  To fix this, a user can downgrade to NumPy 1.11.3.

The problem arises whenever the code tries to reference the variables NX , NY , NZ , MODE, and NEXT, because each of these is stored as a NumPy array in the std_header dictionary.  Simply adding a '[0]' after each makes the code work again.

askorikov:
Fixed #1705
Starting from some recent version of NumPy it is an error to use 1D arrays for indexing -
 so we need to extract scalars here. Info: https://stackoverflow.com/a/42444003

### DIFF
--- a/hyperspy/io_plugins/mrc.py
+++ b/hyperspy/io_plugins/mrc.py
@@ -159,8 +159,8 @@ def file_reader(filename, endianess='<', **kwds):
     if lazy:
         mmap_mode = 'r'
     data = np.memmap(f, mode=mmap_mode, offset=f.tell(),
-                     dtype=get_data_type(std_header['MODE'], endianess)
-                     ).squeeze().reshape((NX, NY, NZ), order='F').T
+                     dtype=get_data_type(std_header['MODE'][0], endianess)
+                     ).squeeze().reshape((NX[0], NY[0], NZ[0]), order='F').T
 
     original_metadata = {'std_header': sarray2dict(std_header)}
     # Convert bytes to unicode


francisco-dlp:
Thanks!

Note: I've realised that this bug-fix was sent to the wrong branch (Rnm instead of Rnp) after merging, and I've fixed by cherry-picking the commit.)

AndrewHerzing:
I seem to still be seeing this bug with version 1.7.3.  Here is the trace:

```python
File ~/anaconda3/envs/tomo/lib/python3.9/site-packages/hyperspy/io_plugins/mrc.py:155, in file_reader(filename, endianess, **kwds)
    153 else:
    154     _logger.warning("There was a problem reading the extended header")
--> 155     f.seek(1024 + std_header['NEXT'])
    156     fei_header = None
    157 NX, NY, NZ = std_header['NX'], std_header['NY'], std_header['NZ']

TypeError: only integer scalar arrays can be converted to a scalar index
```

ericpre:
Can you provide a file to reproduce the issue?

AndrewHerzing:
I can get a test dataset for you tomorrow if need be, but I think I figured out the source of the trouble.  The problem file was generated using SerialEM which uses its own format for the mrc files.  Details are here:

[IMOD MRC format](https://bio3d.colorado.edu/imod/doc/mrc_format.txt)

but it should still work with the reader.  I did some more digging and it looks like the problem is similar to the original issue and arose due to a change in NumPy.  The value of `std_header['NEXT']` is a `numpy.ndarray`.  So, an error is returned in line 155:

`f.seek(1024 + std_header['NEXT'])`

because `f.seek()` expects an integer and not an array. 

This problem was fixed in other parts of the code during the original commit, but this line remained the same.  I guess most people are using FEI style MRC files and did not have any issue since line 155 is ignored in that case.

If line 155 is simply changed to:

`f.seek(1024 + std_header['NEXT'][0])`

then it seems to work.

ericpre:
Yes, providing a small file would be useful to add to the test suite. The development of the IO code is taking place at https://github.com/hyperspy/rosettasciio, but it may be worth making a patch release until RosettaSciIO is used in a released version of hyperspy.

AndrewHerzing:
I am fine if we move the issue to rosettasciio.  I have modified my forked repo to account for the I/O issue as well as the scale issue already on rosettasciio, so I am in no hurry.  Here is a link to test dataset which will reproduce the error:

https://drive.google.com/file/d/1zv1gaa3YYe8Sg5kbaUsPx3qV5yfcFsyX/view?usp=sharing

It is a short tilt series consisting of 9 images which are 512x512 pixels each with 'int16' data type.

PR #1723, opened 2017-08-29
Fix issue with installing the dev version in a windows anaconda

### LINKED ISSUE #1704: Install dev version on windows
@k8macarthur reported the following issue in gitter:

```
Obtaining file:///C:/Users/macark/Documents/GitHub/hyperspy
    Complete output from command python setup.py egg_info:
    C:\Users\macark\Documents\GitHub\hyperspy\setup.py:180: UserWarning: WARNING: C compiler can't be found.
    Only slow pure python alternative functions will be available.
    To use fast implementation of some functions writen in cython/c either:
    a) check that you have compiler (EXACTLY SAME as your python
    distribution was compiled with) installed,
    b) use binary distribution of hyperspy (i.e. wheels, egg, (only osx and win)).
    Installation will continue in 5 sec...
      Installation will continue in 5 sec...""")
      File "<string>", line 1, in <module>
      File "C:\Users\macark\Documents\GitHub\hyperspy\setup.py", line 298, in <module>
        with update_version_when_dev() as version:
      File "C:\Users\macark\Documents\GitHub\hyperspy\setup.py", line 275, in __enter__
        stdout=subprocess.PIPE)
      File "c:\program files\anaconda3\lib\subprocess.py", line 947, in __init__
        restore_signals, start_new_session)
      File "c:\program files\anaconda3\lib\subprocess.py", line 1224, in _execute_child
        startupinfo)
    FileNotFoundError: [WinError 2] The system cannot find the file specified

    ----------------------------------------
Command "python setup.py egg_info" failed with error code 1 in C:\Users\macark\Documents\GitHub\hyperspy\
```

I could reproduce the installation error in an anaconda distribution on Windows 64 bits. Note that there is first the warning related to the compilation of the Cpython part and then the actual error.

Adding `shell=True` as an argument of subprocess.Popen in [setup.py#L275](https://github.com/hyperspy/hyperspy/blob/RELEASE_next_minor/setup.py#L275) seems to be a way to fix... However, I am not sure if this is the proper way.

ericpre:
### Description of the change
Fix #1704 issue reported by @k8macarthur on gitter when installing the dev version in a windows anaconda. I could reproduce locally in a fresh install. I don't know why appveyor does not have the issue but this PR seems to do the job.

### Progress of the PR
- [x] Add shell=True argument to subprocess.Popen in setup.py
- [x] Fix and tidy up path in setup.py.
- [x] ready for review.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -272,7 +272,8 @@ class update_version_when_dev:
         if self.release_version.endswith(".dev"):
             p = subprocess.Popen(["git", "describe",
                                   "--tags", "--dirty", "--always"],
-                                 stdout=subprocess.PIPE)
+                                 stdout=subprocess.PIPE,
+                                 shell=True)
             stdout = p.communicate()[0]
             if p.returncode != 0:
                 # Git is not available, we keep the version as is
--- a/setup.py
+++ b/setup.py
@@ -92,26 +92,19 @@ def update_version(version):
 
 # Extensions. Add your extension here:
 raw_extensions = [Extension("hyperspy.io_plugins.unbcf_fast",
-                            ['hyperspy/io_plugins/unbcf_fast.pyx']),
+                            [os.path.join('hyperspy', 'io_plugins', 'unbcf_fast.pyx')]),
                   ]
 
 cleanup_list = []
 for leftover in raw_extensions:
     path, ext = os.path.splitext(leftover.sources[0])
     if ext in ('.pyx', '.py'):
-        cleanup_list.append(os.path.join(setup_path, path + '.c*'))
+        cleanup_list.append(''.join([os.path.join(setup_path, path), '.c*']))
         if os.name == 'nt':
-            cleanup_list.append(
-                os.path.join(
-                    setup_path,
-                    path +
-                    '.cpython-*.pyd'))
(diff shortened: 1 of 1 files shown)


ericpre:
May be worth checking if this PR fixes #1585 and #1575.

setup.py
```diff
@@ -92,26 +92,19 @@ def update_version(version):
...
-                            ['hyperspy/io_plugins/unbcf_fast.pyx']),
+                            [os.path.join('hyperspy', 'io_plugins', 'unbcf_fast.pyx')]),
                   ]
 
 cleanup_list = []
 for leftover in raw_extensions:
     path, ext = os.path.splitext(leftover.sources[0])
     if ext in ('.pyx', '.py'):
-        cleanup_list.append(os.path.join(setup_path, path + '.c*'))
+        cleanup_list.append(''.join([os.path.join(setup_path, path), '.c*']))
```
thomasaarholt, line 102: @francisco-dlp Why did you change from `path + 'c*'` to `"".join` in this case?

francisco-dlp: Sorry, I can't remember doing this. The main difference is appending the extension after computing the join path which, in the case of the path to a file, should be the same.

PR #1725, opened 2017-08-31
NEW: fft() and ifft()

woozey:
Revised code of #817 with the following changes:
- added option `shifted` in both methods, which calculates FFT with origin shifted in the middle
- FFT is now instance of `ComplexSignal`

_Note: original implementation of IFFT is kept, meaning it returns real part only_ 

FYI: @CodeMonkeyJan, @FWin22


hyperspy/signal.py
```diff
@@ -3169,6 +3169,109 @@ def integrate_simpson(self, axis, out=None):
...
+
+        if shifted:
+            im_fft = self._deepcopy_with_new_data(np.fft.fftshift(
+                np.fft.fftn(self.data, axes=axes), axes=axes))
+        else:
+            im_fft = self._deepcopy_with_new_data(np.fft.fftn(self.data, axes=axes))
+
+        im_fft.change_dtype("complex")
+        shape_fft = self.axes_manager.shape
+        im_fft.metadata.General.title = 'FFT of ' + \
```
ericpre: `'FFT of {}'.format(im_fft.metadata.General.title)` is a recommended string handling practise.

hyperspy/signal.py
```diff
@@ -3169,6 +3169,109 @@ def integrate_simpson(self, axis, out=None):
...
+            im_fft.metadata.General.title
+        for ax, dim in zip(axes, shape_fft):
+            axis = im_fft.axes_manager[ax]
+            axis.scale = 1. / dim / self.axes_manager[ax].scale
+            axis.units = str(self.axes_manager[ax].units) + '$^{-1}$'
```
ericpre: May be worth using the pint library here? In case something turn up with a signal having a physical dimension different from 1... ;)

hyperspy/signal.py
```diff
@@ -3169,6 +3169,109 @@ def integrate_simpson(self, axis, out=None):
...
+
+        if shifted:
+            im_ifft = self._deepcopy_with_new_data(np.fft.ifftn(np.fft.ifftshift(
+                self.data, axes=axes), axes=axes))
+        else:
+            im_ifft = self._deepcopy_with_new_data(np.fft.ifftn(
+                self.data, axes=axes))
+
+        im_ifft.metadata.General.title = 'iFFT of ' + \
+            im_ifft.metadata.General.title
```
ericpre: same as above regarding string formatting.

hyperspy/signal.py
```diff
@@ -3169,6 +3169,109 @@ def integrate_simpson(self, axis, out=None):
...
+
+        Notes
+        -----
+        For further information see the documentation of numpy.fft.fftn
+        """
+
+        if self.axes_manager.signal_dimension == 0:
+            raise AttributeError("Signal dimension must be at least one.")
+        ax = self.axes_manager
+        axes = np.arange(ax.signal_dimension) + ax.navigation_dimension
```
ericpre: I am not sure if your get the correct array indices order that way but this may be irrelevant for what it is used for. Replace it with `self.axes_manager.signal_indices_in_array` would be more simple?

hyperspy/signal.py
```diff
@@ -3169,6 +3169,109 @@ def integrate_simpson(self, axis, out=None):
...
+            im_fft = self._deepcopy_with_new_data(np.fft.fftshift(
+                np.fft.fftn(self.data, axes=axes), axes=axes))
+        else:
+            im_fft = self._deepcopy_with_new_data(np.fft.fftn(self.data, axes=axes))
+
+        im_fft.change_dtype("complex")
+        shape_fft = self.axes_manager.shape
+        im_fft.metadata.General.title = 'FFT of ' + \
+            im_fft.metadata.General.title
+        for ax, dim in zip(axes, shape_fft):
```
```python
for axis in self.axes_manager.signal_axes:
    axis.scale = 1. / dim / axis.scale
    ...
```
or something similar will be easier to read and less error-prone.

hyperspy/signal.py
```diff
@@ -3169,6 +3169,109 @@ def integrate_simpson(self, axis, out=None):
...
+        Notes
+        -----
+        For further information see the documentation of numpy.fft.ifftn
+
+        """
+
+        if self.axes_manager.signal_dimension == 0:
+            raise AttributeError("Signal dimension must be at least one.")
+        ax = self.axes_manager
+        axes = np.arange(ax.signal_dimension) + ax.navigation_dimension
```
ericpre: Same as above regarding array indices.

hyperspy/signal.py
```diff
@@ -3169,6 +3169,109 @@ def integrate_simpson(self, axis, out=None):
...
+            im_ifft = self._deepcopy_with_new_data(np.fft.ifftn(np.fft.ifftshift(
+                self.data, axes=axes), axes=axes))
+        else:
+            im_ifft = self._deepcopy_with_new_data(np.fft.ifftn(
+                self.data, axes=axes))
+
+        im_ifft.metadata.General.title = 'iFFT of ' + \
+            im_ifft.metadata.General.title
+        im_ifft = im_ifft.real
+        for ax, dim in zip(axes, self.axes_manager.shape):
```
ericpre: Same as above.

hyperspy/tests/signal/test_fourier_transform.py
```diff
@@ -0,0 +1,98 @@
...
+
+@pytest.mark.parametrize('lazy', [True, False])
+def test_fft_signal2d(lazy):
+    im = Signal2D(np.random.random((2, 3, 4, 5)))
+    if lazy:
+        im = im.as_lazy()
+    im.axes_manager.signal_axes[0].units = 'nm'
+    im.axes_manager.signal_axes[1].units = 'nm'
+    im_fft = im.fft()
+    assert im_fft.axes_manager.signal_axes[0].units == 'nm$^{-1}$'
```
ericpre: Check also the scale?

hyperspy/signal.py
```diff
@@ -3169,6 +3169,109 @@ def integrate_simpson(self, axis, out=None):
...
+            axis = im_ifft.axes_manager[ax]
+            axis.scale = 1. / dim / self.axes_manager[ax].scale
+            if str(self.axes_manager[ax].units).endswith('$^{-1}$'):
```
ericpre: Same here as for the units comment above.

ericpre (commented):
Thanks @woozey for taking this forward!
Two things:
- I am wondering if it would be worth also adding "FFT of " in `signal_type`? Or store somewhere else that the data are  in frequency space?
- what about adding a `power_spectrum` method... just to avoid people being disappointing with the results of FFT, when there are looking for the log of the power spectrum... I am thinking about something along the line of `s.power_spectrum(log=True)` with `log=True` as default for convenience?

Otherwise, this looks good, the comments are just a few suggestions to improve the style and make the code easier to read.

hyperspy/signal.py
```diff
@@ -3169,6 +3169,109 @@ def integrate_simpson(self, axis, out=None):
             return s
     integrate_simpson.__doc__ %= (ONE_AXIS_PARAMETER, OUT_ARG)
 
+    def fft(self, shifted=True):
```
ericpre: add `**kwargs` to pass to `np.fft.fftn`?

hyperspy/signal.py
```diff
@@ -3169,6 +3169,109 @@ def integrate_simpson(self, axis, out=None):
...
+        im_fft.metadata.General.title = 'FFT of ' + \
+            im_fft.metadata.General.title
+        for ax, dim in zip(axes, shape_fft):
+            axis = im_fft.axes_manager[ax]
+            axis.scale = 1. / dim / self.axes_manager[ax].scale
+            axis.units = str(self.axes_manager[ax].units) + '$^{-1}$'
+            axis.offset = -axis.high_value / 2.
+        return im_fft
+
+    def ifft(self, shifted=True):
```
ericpre: add `**kwargs` to pass to `np.fft.ifftn`?

woozey:
@ericpre 
> I am wondering if it would be worth also adding "FFT of " in signal_type? Or store somewhere else that the data are in frequency space?

Good idea, will add it.

> what about adding a power_spectrum method... just to avoid people being disappointing with the results of FFT, when there are looking for the log of the power spectrum... 

Yes, the standard in EM is plotting log of power spectrum... My vision is that FFT should actually be a separate signal with log of power spectrum as default plot. What about actually a plotting function method rather than a method returning log of power spectrum?  Something like `s.plot_fft(power_spectrum=True, log=True)` with both defaults set to 'True'?

ericpre:
Yes! You are right, there is no need for a `power_spectrum` method, just adding options to the `plot` method should do exactly what is needed. 

```python
s = hs.signals.Signal2D(np.arange(100*100).reshape(100, 100))
s_fft = s.fft()
s_fft.plot(True) # plot the log of the power spectrum
s_fft.plot() # plot the real and imaginary part on a linear scale
```
With `power_spectrum` (False by default), it is possible to write just `True` in the bracket, so it is still compact and convenient. For the `log` option, it could be None as default, which would set True in case of power spectrum and False otherwise. This approach is a bit debatable because people can get confused when the "same" method of different kind of signal have different option order...
Although this sounds great in principle, the power_spectrum will need to be calculate on the fly, to avoid using too much memory (in case of large dataset, we don't want to have the power spectrum of the whole dataset in memory...) and nothing is currently implemented to perform a "live" operation when plotting a signal but #1668 should solve this issue.
Maybe this could be done in a separate PR...

FWin22:
Regarding the fftshift option:
I would only use it as an option for plotting (with default = False), but not for any numerical operation. All `numpy.fft` functions work with the "unshifted" FFT. It will be too confusing using `np.fft.fftshift` all the time and there is absolutely no need for this (except for plotting maybe).

woozey:
@ericpre, @FWin22 :
> I would only use it as an option for plotting (with default = False), but not for any numerical operation. All numpy.fft functions work with the "unshifted" FFT. It will be too confusing using np.fft.fftshift all the time and there is absolutely no need for this (except for plotting maybe).

I tend to agree with that, possibility to use other FFT tools is a good argument not to shift the FFT and I can't think of any other reason of shifting it excepting. So I will change default for `shifted=False` if there are no objections and implement plotting function

doc/user_guide/tools.rst
```diff
@@ -1047,6 +1047,26 @@ to reverse the :py:func:`~.utils.stack` function:
...
+.. code-block:: python
+
+    >>> im = hs.signals.Signal2D(scipy.misc.lena())
+    >>> im.fft().plot()
+
+.. figure::  images/lena_fft.png
+  :align:   center
+  :width:   400
+
+By default both methods calculate FFT and IFFT with origin shifted in the centre. Use `shifted=False` option to
```
ericpre: Update doc regarding `shifted`.

hyperspy/signal.py
```diff
@@ -3169,6 +3170,115 @@ def integrate_simpson(self, axis, out=None):
...
+        Compute the inverse discrete Fourier Transform.
+
+        This function computes real part of the inverse of the discrete
+        Fourier Transform over the signal axes by
+        means of the Fast Fourier Transform (FFT).
+
+        Parameters
+        ----------
+        shifted : bool, optional
+            If True, the origin of FFT will be shifted in the centre (Default: False).
```
ericpre: Same as above regarding **kwargs.

hyperspy/signal.py
```diff
@@ -3169,6 +3170,115 @@ def integrate_simpson(self, axis, out=None):
...
+        """Compute the discrete Fourier Transform.
+
+        This function computes the discrete Fourier Transform over the signal
+        axes by means of the Fast Fourier Transform (FFT) as implemented in
+        numpy.
+
+        Parameters
+        ----------
+        shifted : bool, optional
+            If True, the origin of FFT will be shifted in the centre (Default: False).
```
ericpre, line 3184: Mention that **kwargs goes to `np.fft.fftn`.

hyperspy/signal.py
```diff
@@ -3169,6 +3170,115 @@ def integrate_simpson(self, axis, out=None):
...
+        im_ifft.metadata.Signal.signal_type = 'iFFT of {}'.format(im_ifft.metadata.Signal.signal_type)
+        im_ifft = im_ifft.real
+        ureg = UnitRegistry()
+        for axis in im_ifft.axes_manager.signal_axes:
+            axis.scale = 1. / axis.size / axis.scale
+            try:
+                units = ureg.parse_expression(str(axis.units)) ** (-1)
+                axis.units = '{:~}'.format(units.units)
+            except UndefinedUnitError:
+                _logger.warning('Units are not set or cannot be recognized')
```
ericpre, line 3312: If we get 'UndefinedUnitError', should the units be set to `t.Undefined`?

woozey:
@ericpre, with `plot(True)` stuff above were you referring to modification of the corresponding method of `ComplexSignal?`

hyperspy/tests/signal/test_fourier_transform.py
```diff
@@ -0,0 +1,117 @@
...
+# along with  Hyperspy.  If not, see <http://www.gnu.org/licenses/>.
+
+import pytest
+import numpy as np
+from numpy.testing import assert_allclose
+
+from hyperspy.signals import Signal1D, Signal2D, ComplexSignal1D, ComplexSignal2D, BaseSignal
+
+
+@pytest.mark.parametrize('lazy', [True, False])
```
ericpre, line 25: Actually, as far I understood, this does not test properly lazy signals (although the object tested is an instance of `LazySignal`...), because there is no implementation of the `fft` and `ifft` methods in the `LazySignal` class. @to266 can you please confirm?

If you want to add support lazy signals, you could try to use the `map` method, which is properly implemented in the `LazySignal` class.

ericpre (commented):
A few more comments on things that I missed previously, sorry...

ericpre:
Regarding the plotting, the plotting code is quite convoluted... so I have started a [branch](https://github.com/ericpre/hyperspy/tree/intensity_log_scale_plot), which should be not far from being ready to be reviewed.

francisco-dlp:
Just a couple of comments for now (I'll go through the code once I find some time):

* I think that we shouldn't change the signal type as it is the same signal, just in the different space. That said, the sort of analysis that are applied to FFTs tend to be very different from those applied to real signals so, I wonder if we shouldn't add a boolean ``metadata.Signal.FFT``. Then we could have a ``BaseFFT`` class and specialised e.g. ``HRTEM_FFT`` subclasses. What do you think?

* ``fftshift`` just returns a view of the data, which is cheap. However, if we use it just for plotting, how will ROIs deal with FFTs?

hyperspy/tests/signal/test_fourier_transform.py
francisco-dlp: I think that in the latest version of dask supported numpy functions operating of dask arrays return dask arrays. Hence, if the fft functions are supported, then it should work out of the box. If not, we should create the lazy counterparts.

woozey:
@francisco-dlp:
>  ...I wonder if we shouldn't add a boolean metadata.Signal.FFT. Then we could have a BaseFFT class and specialised e.g. HRTEM_FFT subclasses.

I actually strongly support idea creating BaseFFT class. It is wise strategically and will simplify dealing with plotting as far as i can see... I can refactor the code accordingly and create new PR. @ericpre, what would you say about that?

woozey:
> `fftshift` just returns a view of the data, which is cheap. However, if we use it just for plotting, how will ROIs deal with FFTs?

Aren't ROIs setup to take into account `.axes_manager.signal_axes.offset` value? If not that is certainly the way to go. At the moment `.fft()` takes care of the offset.

hyperspy/tests/signal/test_fourier_transform.py
woozey: I'm not familiar with the stated above... what would be the way finding it out?

hyperspy/signal.py
woozey: I actually think that keeping original units is better approach... For example if someones units are nanometers written with typo, doing fft and than ifft in that case will result in the same units rather than loosing them completely. I'm also not sure how good is units coverage of pint package.

hyperspy/tests/signal/test_fourier_transform.py
francisco-dlp: I can confirm that it does not work out of the box. To enable lazy fft the function must simply check if ``data`` is a ``dask.array`` and, if it is, use ``dask.array.fft`` etc instead of the numpy functions.

woozey: Does it mean that one has to use`map` with `mask.array.Array` instances? Wouldn't be better to use `map` in general i.e. for usual and lazy signals? Does `map` handles lazy signals properly?

francisco-dlp: No, there is no need to use ``map``, it is as simply as using the fft functions from ``dask.array.fft`` instead of numpy.

ericpre: Since [dask 0.15.0](http://dask.pydata.org/en/latest/changelog.html#id3), `fft`, `fftshift` and co are implemented in dask. The dask array minimum version in setup.py will need to be updated.

Yes, normally, it should be as simple as:
```python
import dask.array as da
if isinstance(self.data, da.Array):
    fft = da.fft
else:
    fft = np.fft

# and then use `fft.fftn` or `fft.fftshift`
```

hyperspy/signal.py
ericpre: Pint is quite good, it will deal with units (and potential encoding related issue, that we can get with um, angstrom...) much better than we will do!
If there is a typo, the warning will first point it out and removing the unit may force the user to correct it (where the actual error is)! ;) 

In this specific case, the benefit of using pint are quite marginal thought... so maybe it is not worth it...

hyperspy/tests/signal/test_fourier_transform.py
woozey: It isn't that straight forward as that((( You will see tests failing. The point is that `dask` performs `fft` and `ifft` only on the axis that has a single chunk. Meaning one should check the chunking of the lazy signal data before fft and probably rechunk it in the case if the requirement is not met. In addition to that `fftshift` seem to rechunk the data for some reason, so ifft doesn't work out of the box that cause the test failing( I can of course rechunk it after performing `fftshift`, but not sure what would be the most elegant way of doing it and how to do it without screwing up something... So any help would be appreciated.

francisco-dlp: CC @to266

woozey: @to266 doesn't seem to answer, any idea how to proceed here?

francisco-dlp: By default hyperspy guarantees that the chunk contains at least 1 signal. Therefore, hopefully the fft chunking constraint shouldn't be a problem (?) Regarding the ``fftshift`` chunking issue, are you sure that it rechunks the data? Could you provide an example of the issue using data "chunked" by hyperspy? 

By the way, as ``fftshift`` just provides a different view of the data (at least in the numpy case), what about keeping the original in ``data`` and adding a ``data_fftshifted`` attribute to ``BaseFFTSignal``? The ``data_fftshifted`` can be used for plotting and slicing. After slicing, there could be a mechanism to generate the new ``data`` from the new ``fftshifted_data``.

ericpre: Would shifting be useful for anything else than plotting? If not, we can just shift the data when we plot it -see [#1727](https://github.com/hyperspy/hyperspy/pull/1727/files#diff-64e04530fdfa59f53ebfa39bd6b4fd7cR1924).

francisco-dlp: Yes, it's useful also for slicing and ROI selection.

woozey: @francisco-dlp here is an example:

```python
In[1]: import numpy as np
import hyperspy.api as hs
import dask.array as da
In [3]: im = hs.signals.Signal2D(np.random.random((2, 3, 4, 5)))
In [4]: im = im.as_lazy()
In [5]: im
Out[5]: <LazySignal2D, title: , dimensions: (3, 2|5, 4)>
In [6]: isinstance(im.data, da.Array)
Out[6]: True
In [7]: im.data.chunks
Out[7]: ((2,), (3,), (4,), (5,))
In [8]: axes = im.axes_manager.signal_indices_in_array
In [9]: axes
Out[9]: (3, 2)
In [15]: im_fft = im._deepcopy_with_new_data(da.fft.fftshift(
                    da.fft.fftn(im.data, axes=axes), axes=axes))
In [18]: im_fft.change_dtype(complex)
In [19]: im_fft
Out[19]: <LazyComplexSignal2D, title: , dimensions: (3, 2|5, 4)>
In [22]: im_ifft = im_fft._deepcopy_with_new_data(da.fft.ifftn(da.fft.ifftshift(
                    im_fft.data, axes=axes), axes=axes))
> ValueError                                Traceback (most recent call last)
<ipython-input-22-9c6b54261673> in <module>()
      1 im_ifft = im_fft._deepcopy_with_new_data(da.fft.ifftn(da.fft.fftshift(
----> 2                     im_fft.data, axes=axes), axes=axes))

> ~/anaconda/lib/python3.5/site-packages/dask/array/fft.py in func(a, s, axes)
    166         for each_axis in axes:
    167             if len(a.chunks[each_axis]) != 1:
--> 168                 raise ValueError(chunk_error % (each_axis, a.chunks[each_axis]))
    169 
    170         chunks = out_chunk_fn(a, s, axes)

> ValueError: Dask array only supports taking an FFT along an axis that 
has a single chunk. An FFT operation was tried on axis 3 
which has chunks (2, 3). To change the array's chunks use dask.Array.rechunk.

The error arises since the da.ifftshift() has rechecked the Array, see bellow:

> In [25]: im_fft.data.chunks
Out[25]: ((2,), (3,), (4,), (5,))
In [26]: da.fft.ifftshift(im_fft.data, axes=axes).chunks
Out[26]: ((2,), (3,), (2, 2), (3, 2))
```

woozey:
@francisco-dlp, any idea why AppVeyor complaining about `dark.array.fft.fftn` and Travis CI not? They both seem to use the latest version of `dask`...

woozey:
So concerning the FFT with Lazy signals using dask and getting rechunked data after `fftshift` that caused `ifft` failing:
Issue that I have posted in the dask caused a PR fixing `da.fftshift` in order to maintain the original chunking: https://github.com/dask/dask/pull/2733. The last one was closed than due to the fact that rechunking is inefficient operation. So the conclusion was that the users should be aware of rechunking and handle it by themselves.

Based on stated above, the rechunking of shifted signals should be kept as it is now and in the sake of efficiency using shifted FFT with lazy signals should be avoided.

woozey:
Now user guide suggests to plot power spectrum rather than FFT itself and shows a simple way to do it.
@thomasaarholt - FYI

thomasaarholt:
@woozey This is really nice. Two comments:

How does s.ifft() know whether the fft is`shifted` or not? I couldn't find anything in the code.

~~Could I suggest adding another convenience argument of `power=True` to s.fft()? One that would take the absolute value of the square of the signal, returning a power spectrum? Then s.fft(True, True) would return a real, calibrated power spectrum showing the frequencies present in a given signal.~~

Actually, I only just spotted s.amplitude. That's probably enough! :)
`fft = np.square(s.fft(True).amplitude)` does it nicely.

woozey:
> How does s.ifft() know whether the fft isshifted or not? I couldn't find anything in the code.

It actually doesn't. This is done manually by user with parameter `shifted=True` or `False`.

> Actually, I only just spotted s.amplitude. That's probably enough! :)
fft = np.square(s.fft(True).amplitude) does it nicely.

I personally would use `np.log()` in front... I think it would make more sense implementing the power spectrum in plotting of FFTs, but as far as I understand that would ideally require creating a separate `FFTSignal` class, isn't it? Alternatively one can think of adding a function `plot_power_spectrum()` or `plot_fft(log=True)`, but this is subject to another PR in my opinion, which I can make right after merging this one.

thomasaarholt:
All good. This is working nicely. The only tests failing are plot-related tests, external to this PR.

ericpre:
I think we still need to sort out the shifting bit in this PR. Similar thing is where/how to save the information about the signal being in Fourier space? Actually, I would agree with @francisco-dlp that we should not change the `signal_type` since this is the same signal but in different space...
dask/dask#2733 have been solved, can it be useful here?

hyperspy/signal.py
```diff
@@ -3169,6 +3171,142 @@ def integrate_simpson(self, axis, out=None):
...
+
+        Return
+        ------
+        s : ComplexSignal
+
+        Examples
+        --------
+        >>> im = hs.signals.Signal2D(scipy.misc.ascent())
+        >>> im.fft()
+        <Image, title: FFT of , dimensions: (|512, 512)>
```
magnunor: This should be `<ComplexSignal2D, title: FFT of , dimensions: (|512, 512)>`

hyperspy/signal.py
```diff
@@ -3169,6 +3171,142 @@ def integrate_simpson(self, axis, out=None):
...
+
```
magnunor: Could you add the power spectrum example from the user guide to the docstring example as well? Since most people using this function will probably expect the same output as they get in DM.

hyperspy/signal.py
```diff
@@ -3169,6 +3171,142 @@ def integrate_simpson(self, axis, out=None):
...
+        ------
+        s : Signal
+
+        Examples
+        --------
+        >>> import scipy
+        >>> im = hs.signals.Signal2D(scipy.misc.ascent())
+        >>> imfft = im.fft()
+        >>> imfft.ifft()
+        <Image, title: iFFT of FFT of , dimensions: (|512, 512)>
```
magnunor: This should be `<Signal2D, title: real(iFFT of FFT of ), dimensions: (|512, 512)>`

magnunor:
Does `s.fft` and `s.ifft` handle signals with navigation dimensions, for example 4D-datasets? If not, it should raise a `NotImplementedError`(?) for these types of signals.

hyperspy/signal.py
woozey: Both my fault. I kept whatever was implemented before, rather than going carefully line by line. Thanks for picking it up. All three are done now.

woozey:
@ericpre 
> I think we still need to sort out the shifting bit in this PR. Similar thing is where/how to save the information about the signal being in Fourier space? Actually, I would agree with @francisco-dlp that we should not change the signal_type since this is the same signal but in different space...

I will remove everything to do with changing `signal_type`, since indeed we don't really change it. What do you think is the best way storing the information about the FFT? @francisco-dlp, how about if I add boolean entry `metadata.FFT.shifted` which will be deleted together with FFT node upon performing IFFT?

> dask/dask#2733 have been solved, can it be useful here?

Thanks for pointing this out. I will check how it performs and will likely remove part of code that checks chunking.

woozey:
@magnunor 
Yes, it does work for nav_dim>0. The tests are actually using 4D data. It also works with lazy signals.

francisco-dlp:
@woozey, adding a ``FFT`` node sounds good.

francisco-dlp:
By the way, just in case somebody (like me) has forgotten about, @ericpre implemented log image and power spectrum plotting in #1727

woozey:
@ericpre 
>> dask/dask#2733 have been solved, can it be useful here?

>Thanks for pointing this out. I will check how it performs and will likely remove part of code that checks chunking.

The patch mentioned above is released and works well. Using it will set the requirements for `dask>=0.16.1` and the code will break on the systems that have later versions. On the other hand if I keep it as it is, the code will `try` to performing  the `ifft` without rechunking which will work with new `dask`, otherwise if `dask` is outdated it will rechunk and do `ifft` which is presumably due to rechunking will be slower than using newest `dask` version.

So the question is it ok to increase requirements for `dask`, or it is better to keep it more robust and working also for older versions? @francisco-dlp what would you suggest?

francisco-dlp:
I think that it is fine to set the requirement to the latest version of dask as long as all tests pass with it.

hyperspy/signal.py
```diff
@@ -3169,6 +3171,141 @@ def integrate_simpson(self, axis, out=None):
...
+        Compute the inverse discrete Fourier Transform.
+
+        This function computes real part of the inverse of the discrete
+        Fourier Transform over the signal axes by means of the
+        Fast Fourier Transform (FFT) as implemented in
+        numpy.
+
+        Parameters
+        ----------
+        shifted : bool, optional
```
ericpre: Since we now have the node `Signal.FFT.shifted`, ifft can know if the ifft need to be shifted or not. What's about having `shifted=None` to let decide if the ifft should be shifted or not (based on the `Signal.FFT.shifted` node)?

ericpre (commented):
I think this looks good now!! Just one very small comment.

hyperspy/signal.py
woozey: Done. The way I implemented it is that by default is `shifted=None` and it checks the `Signal.FFT.shifted` node, and if it doesn't exist it sets `shifted=False`. One can still pass boolean values instead.

woozey:
Although I promised to @thomasaarholt that introducing apodization (which is used for plotting in order to remove streaking) is straight forward, I have underestimated complexity introduced by lazy signal compatibility and natural multidimensionality of Hyperspy. I do work on the implementation now but it going to take a while... Do we want to wait until I implement it, or shall I put it in a separate PR. This one seems to be checked from several people and ready (or almost ready) to be merged.

francisco-dlp:
I would go for leaving that for another PR. @ericpre, this LGTM, what do you think?

thomasaarholt:
Agreed on waiting for another PR. LGTM!

woozey:
LGTM-?

thomasaarholt:
LGTM = Looks good to me. :)

ericpre:
Thanks @woozey!

Issue #543, opened 2015-05-07
Log scale on Y-axis


rgpdx:
Currently in Hyperspy v0.8 there is no way to plot with Y-axis in log-scale. Can this be incorporated in a future version? Spectral data in linear scale is not always useful, and a semi-log plot is critical in many instances to clearly discern peaks and visualize/characterize the data for decision-making.

francisco-dlp:
Currently pressing the "l" key with the spectrum windows on focus toggles the y axis log scale. Do you have in mind a more convenient way to perform the task? For example, something like this:

```python
s.plot(yscale="log")
```

rgpdx:
My use case with Hyperspy will be to automatically take a bunch of
emsa files, plot them out and do some basic analysis on the files for
smoothing and peak detection. I want to put the plots into a pdf or in
a pptx file (using python-pptx) for automated reports. The kinds of
plots I look at are just easier to understand in semi-log scale and
therefore I'd like the feature to be something like what you stated
below -


s.plot(yscale="log")


is perfectly acceptable. Thank you for considering it.

francisco-dlp:
Ok, we'll leave it in our wish-list waiting for some attention. In the meantime you could perform the task less conveniently as follows:

```python

import matplotlib.pyplot as plt
ax = plt.gca()
ax.set_yscale("log")
ax.figure.canvas.draw()

```

rgpdx:
Francisco

Can you give me a py file I can use to try this out? I tried the following
code and it doesn't work for me. Am I missing something?

import hyperspy as hsp
import matplotlib.pyplot as plt
ax = plt.gca()
ax.set_yscale("log")
ax.figure.canvas.draw()
s = load('test_emsa.emsa')
s.plot()

francisco-dlp:
Your script doesn't work because `ax.set_yscale("log")` modifies the y scale of the axes in the active figure, but you create the figure after setting the scale. The following works for me:

```python
import hyperspy.hspy as hs
import numpy as np
import matplotlib.pyplot as plt
%matplotlib qt


s= hs.signals.Spectrum(np.random.rand(100))
s.plot()
ax = plt.gca()
ax.set_yscale("log")
ax.figure.canvas.draw()

```
![example_log_scale](https://cloud.githubusercontent.com/assets/989719/7608741/2b893658-f964-11e4-9b1a-d423d9f46c94.png)

rgpdx:
Yes, that worked. Thanks!

PR #1727, opened 2017-09-10
Intensity log scale plot

ericpre:
Following #1725, this PR add support for plotting signal on a log scale with the `intensity_scale` argument of the `plot`, which can be set to 'linear', 'log' or None.:
- [x] add `intensity_scale` to the `plot` method of the `BaseSignal` class,
- [x] scale can be toggled between linear and logarithmic scale by pressing the _l_ key,
- [x] add `power_spectrum` and `shifted` parameters to the `plot` method of complex signals,
- [x] plot power spectum with zero frequency component shifted to the center for FFT by default,
- [x] add tests,
- [x] update doc,
- [x] ready for review.

By default if the signal to plot has the `Signal.FFT` node in the metadata, the power spectrum will be plotted on a log scale (the complex FFT can still be plotted by setting `power_spectrum=False`).


ericpre:
Fix #543.

ericpre:
@woozey, in case you fancy having a look! ;)

thomasaarholt:
I've been playing around with the log plotting. Works nicely!

One minor error: My (EELS) spectrum image has some negative values in it, but certainly not all of the values are negative:

```python
>>> eels
<EELSSpectrum, title: EELS Spectrum Image, dimensions: (24, 56|2048)>
>>> eels.T.plot(intensity_scale='log')
ValueError: Data has no positive values, and therefore can not be log-scaled.
```

thomasaarholt:
`fft.plot(shifted=True/False)` seems to have no effect on my fft signal.

image

ericpre:
Does the ```ValueError``` comes from matplotlib? Can you give the traceback, to make sure?
I don't think that there is anything sensible to do from our side.

thomasaarholt:
Based on a quick google search I think you are correct. I can double check tomorrow.

Any thought on the `shifted`?

thomasaarholt:
I haven't got time right now to look for the file with negative counts. I'm pretty sure that this was matplotlib, not you.

woozey:
Nice PR! It works fine for me. I'll try some more use cases and will give more feedback. Some comments that I have now:
- Plotting shifted FFT by default is good for non-experienced users who just want to see their FFT as they are used to, though for advanced users plotting may give an impression that the FFT itself is shifted. Would it be better warning a user that the display shows shifted FFT while the signal itself is not?
- Another one with regards to shifting:
``` python
holo.fft(shifted=True).plot()
holo.fft(shifted=False).plot()
```
Would  result in:
screen shot 2018-04-16 at 14 46 41
screen shot 2018-04-16 at 14 46 59
whereas following the idea of your implementation it probably has to look the same. Would it be better to check the `shifted` node in metadata to decide if the display needs to be shifted or not?
- And the last one, having in mind that you have done so much work figuring all of that out, it would be nice to have a possibility to plot any other signal on a log scale, something like `s.plot(log_scale = True)`. Would you be able to implement that (probably in different PR) without a lot of extra efforts?

ericpre:
Thanks for the feedback! There are indeed quite a few things which need to be discussed here.

I did look at using the "shifted" node of the metadata, but I ended up not being happy about it.
Regarding the defaults, we could consider having different defaults for different kind of signal, which may become particularly relevant after the split of hyperspy/hyperspy-extension

This PR implements the `intensity_scale` option to the plot, which should allow you to plot any signal on a log scale. If not, this is a bug! ;)

woozey:
> I did look at using the "shifted" node of the metadata, but I ended up not being happy about it.

What is the reason? Is there anything that should be improved there?

> Regarding the defaults, we could consider having different defaults for different kind of signal, which may become particularly relevant after the split of hyperspy/hyperspy-extension

I wouldn't say my worry are defaults, as I'm happy to have it shifted by default. My real worry is that it is confusing for a user getting the plots shifted although the actual data is not. So I still would strongly advice to put a warning or something in order to inform a user.

> This PR implements the intensity_scale option to the plot, which should allow you to plot any signal on a log scale. If not, this is a bug! ;)

Indeed! I have overseen that for no reason! It works excellent! Thanks!

ericpre:
I don't remember exactly, but I don't think there was any particular reason, I just wanted to keep it simple and also hear what other people would expect as default behaviour before going forward.
Maybe it would be good to change the argument name from `shifted` to `centered` in the `plot` method? So that there is no confusion with the `shifted` node of the metadata, because this argument is only about having the plot with the (0,0) component in the centre regardless of the actual data.

The problem with warning is that people doesn't like it... maybe we could log this at the "info" level.

Normally, they should be a shortcut "l" (as L) to toggle from linear to log scale.

woozey:
> I don't remember exactly...
Maybe it would be good to change the argument name from shifted to centered in the plot method?

This sounds reasonable to me. The name does not really matter but that certainly will help avoiding confusion! Another option could be doing similar to `matplotlib.pyplot.imshow()`, i.e. calling the parameter `origin` and passing string to it, but this is up to you.

> The problem with warning is that people doesn't like it... maybe we could log this at the "info" level.

Perhaps... I don't know what is the best practice in such cases.

> Normally, they should be a shortcut "l" (as L) to toggle from linear to log scale.

This one works too!

thomasaarholt:
I think a combination of logging to info + clear documentation is fine.

francisco-dlp:
Could you merge Rnm into this branch?

ericpre:
@woozey: renaming `shifted` to `origin` with `['center'|'upper' | 'lower']` as possible parameter to mimic matplotlib sounds good to me.

francisco-dlp:
The implementation of ``shifted`` in this PR has the limitation that ROIs are not supported. The combination of FFT and ROIs is typical in EM e.g. FFT + ROI (for masking) + IFFT. Therefore I think that it'll be good to find a workaround to the issue.

What about the following alternative implementation: add ``shifted`` and ``power_spectrum`` properties to the complex signal. In this way

```python
im.plot(shifted=True, power_spectrum=True)
```
would be equivalent to
```python
im.shifted.power_spectrum.plot(intensity="log")
```

The obvious disadvantage is that, just for plotting,  ``shifted`` and ``power_spectrum`` will need to process the whole dataset, while currently only the current signal is processed. A workaround would be to call ``as_lazy`` on the way:

```python
im.as_lazy().shifted.power_spectrum.plot()
```

What do you think?

woozey:
@francisco-dlp with respect to using ROI for masking, how would this work? Is there anything like `roi_to_mask()` exists? In case if it doesn't work out of the shelf, I wouldn't even bother to make it compatible...
`shifted` exists as a kwarg of `fft()` method, so the following could be used:
``` python
holo = hs.datasets.example_signals.object_hologram()
fft = holo.fft(shifted=True)
roi = hs.roi.CircleROI(cx=0.134844, cy=-0.255886, r=0.100868)
fft.plot(power_spectrum=True, shifted=False)
roi.interactive(fft)
```
screen shot 2018-07-11 at 14 33 15

Note that I had to switch off `shifted` for the plotting of `power_spectrum` as default is `True`... I think I mentioned this problem somewhere above.  So if something like `roi_to_mask()` would exist, the following would do the job:
``` python
(fft*roi.to_mask()).ifft()
```

francisco-dlp:
Masking a ROI is already possible. It's a bit harder with the ``CircleROI`` because it uses mask arrays, but it should work:

```python
%matplotlib qt

import hyperspy.api as hs

holo = hs.datasets.example_signals.object_hologram()
fft = holo.fft(shifted=True)
roi = hs.roi.CircleROI(cx=-0.00106177, cy=-0.00743237, r=0.100868, r_inner=0)
fft.plot(power_spectrum=True, shifted=False)
holo_roi = roi.interactive(fft)
holo_roi.data.data[~holo_roi.data.mask] = 0
fft.ifft.plot(()
```

woozey:
@francisco-dlp 
The example above doesn't seem to work the way I'd expect, the main reason is that one would apply mask to the `fft` itself, rather than `holo_roi`. But the mask is different size in such a case. Also another minor remark, the mask shouldn't be inverted as it is `True` outside the ROI.

francisco-dlp:
Sorry that the name of the variables were misleading: the mask is applied to the fft in the example. Also, the mask is inverted with ``~``. The same example with better variable names and a typo fixed:

```python

%matplotlib qt

import hyperspy.api as hs

holo = hs.datasets.example_signals.object_hologram()
fft = holo.fft(shifted=True)
roi = hs.roi.CircleROI(cx=-0.00106177, cy=-0.00743237, r=0.100868, r_inner=0)
fft.plot(power_spectrum=True, shifted=False)
fft_roi = roi.interactive(fft)
fft_roi.data.data[~fft_roi.data.mask] = 0
fft.ifft().plot()

```

francisco-dlp:
I read too quickly your remark regarding inverting the mask: you're right, it's `True` outside the mask, and that's why I invert it: to replace what's inside the mask with zeros.

ericpre:
> The implementation of shifted in this PR has the limitation that ROIs are not supported. The combination of FFT and ROIs is typical in EM e.g. FFT + ROI (for masking) + IFFT. Therefore I think that it'll be good to find a workaround to the issue.

Yes, supporting roi with "shifted" plot would possibly be a pain... however to do bragg filtering we don't need it _providing_ that the fft have been calculated "shifted". 
We still need a function to mask the FFT outside the roi and I don't see how we can do this in hyperspy currently, but it should not be a very complicated! Then it should be fairly straighforward to make a method which does bragg filtering. If I am not missing anything, it looks to me that the most annoying bits are already in place.

francisco-dlp:
Reading @ericpre comment I now understand what @woozey meant which is not the same I meant in my first comment. @woozey, does the following work as you expected?

```python
%matplotlib qt

import hyperspy.api as hs

holo = hs.datasets.example_signals.object_hologram()
fft = holo.fft(shifted=True)

roi = hs.roi.CircleROI(cx=-0.00106177, cy=-0.00743237, r=0.100868, r_inner=0)
fft.plot(power_spectrum=True, shifted=False)
fft_roi = roi.interactive(fft)

def filter(roi, fft):
    fft_mask = fft._deepcopy_with_new_data(np.ones_like(fft.data, dtype="bool"))
    mask_roi = roi(fft_mask)
    mask_roi.data.data[~mask_roi.data.mask] = False
    fft = fft.deepcopy()
    fft.data[fft_mask.data] = 0
    return fft.ifft()

filter(roi=roi, fft=fft).plot()

# Change ROI parameters and filter again
roi.set(cx=0.141215, cy=-0.272874, r=0.0307913, r_inner=0)
filter(roi=roi, fft=fft).plot()
```

ericpre:
Yes! I guess that we should leave this for another PR?
I am tempted to rename the metadata `FFT.shifted` = [True | False] to `FFT.origin` = ["top" | "centre"] and also change the argument of the plot method accordingly. What do you think?

Also, I am wondering if the fft should be shifted by default?

woozey:
@francisco-dlp your suggestion seem to work nicely, thanks.
@ericpre
> I am tempted to rename the metadata FFT.shifted = [True | False] to FFT.origin = ["top" | "centre"] 

I would be fine with either.

 I still have some doubts about the following:
- `holo.fft(shifted=Ture).plot()` would do shifting twice and plot shows unshifted FFT. As far as I understood you want to display FFT with origin in the middle. So I would suggest to use the tag in metadata in order to decide if it needs shifting before plotting or not.
- I really want to warn you agains shifting FFT in the middle by default in plotting and especially in FFT method. In the last one, the main reason is that if you want to process FFT further all functions in `numpy.fft` package work on UNSHIFTED FFTs. Essentially shifting FFT is typically  done only for better visualisation. 
- Having `shifted=True` as default in plotting can be very misleading. For instance, if plotting is done in order to check how the FFT looks like prior further processing, the user will get an impression that `s.fft()` returns shifted FFT by default which isn't the case. In case  if final decision is to keep the default as it is, I would strongly recommend to write in title of the plot "**Shifted** power spectrum of FFT of Signal" (or something like that), that user has a chance to see immediately what is happening.

ericpre:
To avoid confusing (which is very easy in this PR!!) I try to summarise my understanding:

Ideally, a sensible default would be:
- compute fft not shifted
- plot not shifted (= origin not centred)
- use roi, with shifted=True for plotting if desired.

The issue that we currently have here is that we can't shift the fft only for plotting and use the roi at the same time. I am afraid that getting this to work will complicate the roi implementation quite a bit (it is already not much fun)... If someone find some clever tricks to get work nicely, it is more than welcome.

Assuming that most people using the ftt would also want to plot and it use the roi for filtering, in this case, they will need to:
- compute fft shifted
- plot "shifted" (= origin centred) but without any shifting operation during plotting
- roi work fine and fft has the origin centred which makes the user happy!

It looks to me that setting `shifted=True` as default in the fft operation would be convenient for the neophyte and it will be the behaviour that would expect most users (same argument works for power spectrum versus actual fft). I am not convinced that this is something that we want to enforce, because writing `im.fft(True)` instead of `im.fft()` should be manageable for the users! They may be surprised the first time (if they don't use shifted=True to compute the fft) but if they are well behaving users, they will read the documentation, learn something and work it out! :)

Using "origin" in the metadata and as plotting argument should be less misleading than "shifted".

woozey:
@ericpre , I would agree with the last statements. My preferred set of defaults would be:
* `s.fft(shifted=False)`
* `fft.plot(shifted (or origin)=False, power_spectrum=True)`

francisco-dlp:
Some suggestions:

Personally I would prefer to keep the name of the kwarg as it is because i) it is consistent with numpy ii) when applied to spectra ``origin="top"`` doesn't make much sense.

I agree with @woozey that the default for ``shifted`` should be ``False``. I also agree with him in that it should honour the ``FFT.shifted`` metadata i.e.:

```python

if shifted is FFT.shifted:
    do nothing
elif not shifted and FFT.shifted:
   unshift
else:
   shift
```

I also think that it would be good to remove the ``None`` option to automatically shift when ``power_spectrum`` is ``True``

Regarding the ``power_spectrum``, I think that it would be better to set it to ``False`` too. In this way we make sure that the users always know what they're doing and looking at. I know that this is not how some EM software behave, and that's probably why in our community some believe that the FFT and log(PSD) are the same thing.

francisco-dlp:
I forgot to mention that for the power spectrum of a 1D signal y-axis to be in power spectral density units, I think that we must multiply it by ``signal_axis.scale / signal_axis.size``. Also, wouldn't it be better to display the units too (quantity units ** 2 * signal_axis.units I think)? Otherwise we could simply set the units to arbitrary, but this may annoy people from other fields.

ericpre:
Yes, indeed, "origin" got its drawback too... I did as you suggested and did two versions for the name of the parameter:

1. with "shifted"
2. with "shift"

I prefer the second one because it is clear what it does regardless if the FFT is shifted or not. Moreover, I find it is more convenient to use with ROI:
```python
import hyperspy.api as hs
holo = hs.datasets.example_signals.object_hologram()
fft = holo.fft(True)
fft.plot(True) # True is for the power spectrum
```

With the version with "shifted":
```python
import hyperspy.api as hs
holo = hs.datasets.example_signals.object_hologram()
fft = holo.fft(True)
fft.plot(True, shifted=True) # True is for the power spectrum, but we also need to specify `shifted=True` if not the FFT is shifted back...
```

woozey:
Now I see the point for shift and shifted! Yes, I think shift would be better than. I'll test the latest version in a while.

woozey:
Just gave it a try. Looks good to me!

hyperspy/drawing/image.py
```diff
@@ -361,15 +363,29 @@ def format_coord(x, y):
...
+        if self.intensity_scale == 'log':
+            from matplotlib.colors import LogNorm
+            norm = LogNorm(vmin=self.vmin, vmax=self.vmax)
+        else:
+            if self.intensity_scale not in ['linear', None]:
+                _logger.warning('The provided `intensity_scale` argument is '
+                                'not supported, falling back to the default '
+                                '(linear).')
+            # if the `norm` kwargs is passed to matplotlib, we use it
+            norm = kwargs.pop('norm', None)
```
francisco-dlp: I would be good to add an example to the docstring and the user guide of how to use mpl's ``norm`` to e.g. display images with negative numbers using ``SymLogNorm``

doc/user_guide/tools.rst
```diff
@@ -1073,18 +1073,18 @@ The Fast Fourier transform and its inverse can be applied on a signal with the :
...
   :width:   400
 
-Note that for visual inspection of FFT it is common to plot logarithm of amplitude rather than FFT itself as it is done
-    in the example above.
+Note that for visual inspection of FFT, it is common to plot the power spectrum (absolute value of the complex signal) on a logarithmic scale rather than the FFT itself as it is done in the example above.
+By default, in case of FFT, HyperSpy plots the power spectrum and shifts the zero frequency component to the center of the signal. This can be changed 
+by setting ``power_spectrum=False`` and ``shifted=False`` parameters of the plot method.
 
-By default both methods calculate FFT and IFFT with origin at (0, 0) (not in the centre of FFT). Use `shifted=True` option to
+By default, both methods calculate FFT and IFFT with origin at (0, 0) (not in the centre of FFT). Use ``shifted=True`` option to
```
francisco-dlp: I think that it would be good to mention here that ROIs don't work on "unshifted" FFTs shifted to plot.

hyperspy/docstrings/plot.py
```diff
@@ -53,6 +53,10 @@
         plot_markers : bool, default True
             Plot markers added using s.add_marker(marker, permanent=True).
             Note, a large number of markers might lead to very slow plotting.
+        intensity_scale : {None, 'linear', 'log'}, default is None.
```
francisco-dlp: What about using ``"auto"`` instead of ``None``. The advantage is that it's more explicit. I know that in HyperSpy ``None`` often means ``"auto"``, maybe something else to fix?

hyperspy/drawing/image.py
```diff
@@ -361,15 +363,29 @@ def format_coord(x, y):
             vmin, vmax = utils.centre_colormap_values(self.vmin, self.vmax)
         else:
             vmin, vmax = self.vmin, self.vmax
-        if ims:
+
+        # set the image normalisation
+        if self.intensity_scale == 'log':
+            from matplotlib.colors import LogNorm
+            norm = LogNorm(vmin=self.vmin, vmax=self.vmax)
```
francisco-dlp: What happens if ``norm`` is in ``kwargs`` and ``intensity_scale == "log"``?

Wouldn't it be clearer if we use ``norm`` instead of ``intensity_scale`` as the name for this kwarg. If it is a mpl norm, we use it directly, if ``"linear"  or "log" or "auto"`` we do as it currently does for ``intensity_scale``.

hyperspy/drawing/image.py
francisco-dlp: If the norm is an mpl norm it could be an instance or a class (currently it assumes that it is an instance). If it is a class, it could create the instance passing ``vmin`` and ``max`` to the constructor.

hyperspy/_signals/complex_signal2d.py
```diff
@@ -57,6 +57,9 @@ def add_phase_ramp(self, ramp_x, ramp_y, offset=0):
         self.phase = phase
 
     def plot(self,
+             power_spectrum=False,
+             intensity_scale=None,
+             shift=False,
```
francisco-dlp: What about calling this ``shift_fft`` instead?

hyperspy/signal.py
```diff
@@ -1973,12 +1980,17 @@ def plot(self, navigator="auto", axes_manager=None, plot_markers=True,
...
         self._plot.axes_manager = axes_manager
         self._plot.signal_data_function = self.__call__
+
+        if self.metadata.has_item("Signal.quantity"):
+            self._plot.quantity_label = self.metadata.Signal.quantity
         if self.metadata.General.title:
-            self._plot.signal_title = self.metadata.General.title
+            title = self.metadata.General.title
+            if kwargs.get('power_spectrum', False):
+                title = title.replace('FFT', 'Power spectrum')
```
francisco-dlp: This assumes that FFT is contained in the title. It may not be the case because the user has changed it or the data comes from an external source.

hyperspy/signal.py
```diff
@@ -1973,12 +1980,17 @@ def plot(self, navigator="auto", axes_manager=None, plot_markers=True,
...
+                self._plot.quantity_label = 'Power spectral density'
```
francisco-dlp: What about the units?

hyperspy/signal.py
```diff
@@ -3256,7 +3268,7 @@ def fft(self, shifted=False, **kwargs):
 
         Parameters
         ----------
-        shifted : bool, optional
+        shift : bool, optional
             If True, the origin of FFT will be shifted in the centre (Default: False).
```
francisco-dlp: to the centre?

hyperspy/signal.py
```diff
@@ -3301,21 +3313,24 @@ def fft(self, shifted=False, **kwargs):
         im_fft.change_dtype("complex")
         im_fft.metadata.General.title = 'FFT of {}'.format(
             im_fft.metadata.General.title)
-        im_fft.metadata.set_item('Signal.FFT.shifted', shifted)
+        im_fft.metadata.set_item('Signal.FFT.shifted', shift)
+        if hasattr(self.metadata.Signal, 'quantity'):
+            self.metadata.Signal.__delattr__('quantity')
```
francisco-dlp, line 3315: Why removing the quantity? Couldn't we compute the right one with pint? Same for units.

francisco-dlp (changes requested):
In addition to the small comments I wonder if it wouldn't be better to replace the ``fft`` ``shift`` kwarg with a ``shift_fft`` or maybe better (for discovery) ``fft_shift`` method. The main advantage is that it'll enable going from unshifted to shifted and back to unshifted. It'll also make be more explicit that the operation doubles the size in memory of the data temporarily.

francisco-dlp (changes requested):
In addition to the comments above I think that it would be better to replace the ``shift`` kwarg of ``fft`` with a ``fft_shift`` method. In this way:
1. It'll be possible to "unshift" a "shifted" fft
2. It'll be more explicit that the operations requires temporarily doubling the size in memory.

hyperspy/signal.py
ericpre: I am not sure if pint can do it properly.

ericpre:
> In addition to the comments above I think that it would be better to replace the shift kwarg of fft with a fft_shift method. In this way:
> 1. It'll be possible to "unshift" a "shifted" fft

This is already the possible, isn't
> 2. It'll be more explicit that the operations requires temporarily doubling the size in memory.

I don't see what is doubling the memory?

francisco-dlp:
I may be wrong, but I think that ``fftshift`` returns a copy of the data. Therefore, it requires 2x the memory compared to an "unshifted" fft.

francisco-dlp:
@ericpre, is this ready for review?

ericpre:
> I may be wrong, but I think that fftshift returns a copy of the data. Therefore, it requires 2x the memory compared to an "unshifted" fft.

Even if it's return a copy (I haven't checked), for plotting it is computed only using the current index, so this is possibly not a big deal in term of memory issue?

Other than that, yes, it is ready for review.

francisco-dlp:
Sorry my point wasn't fully clear: I meant the ``shifted`` kwarg in ``fft``. If you prefer, I merge this and I can submit a new PR to turn ``shfit`` into a ``fft_shift`` method.

ericpre:
Yes, let's do this in another PR.

Issue #1726, opened 2017-09-08
Scaling Issues in .bcf files and beyond


k8macarthur:
The .bcf import function automatically imports the spectrum image with the scale in micro. These leads to two problems:

- [ ] The scale bar appears in the ridiculous scale. Could this be forced to be greater than 1 for example?

- [ ] The EDX quantification function assumes the pixel size is in units of nm but doesn't actually check for this. Therefore a check for units needs to be added.

 I would also suggest that we could added a check to the import function to automatically convert the scale to smaller units if necessary?

francisco-dlp:
Good points. @ericpre did some work on this using pint but I can't remember how far he got into solving the issue.

k8macarthur:
This came up during the workshop. In the short term, fixing the quantification function should be relatively easy. I'll try and get to that asap. I think @magnunor  had some thoughts about how to update the scale bar function.

magnunor:
I've been thinking about updating the scale bar function, as mentioned in https://github.com/hyperspy/hyperspy/issues/1066. But I guess the scaling issue discussed here is a separate thing, since https://github.com/hyperspy/hyperspy/issues/1066 only deals with the visualization, not the numbers or text.

sem-geologist:
at this moment it is in µm as:
1. bcf internally are in µm
2. µm is universal unit in between cm and nm orders.
E.g. in geology (which was my main aim) lots of mapping are done in cm scale (with image width up to 2 cm) nm scale would make no sense, µm is just right.

The best would be to cast all units into SI and then use some library to show appropriate order.
The moment I wrote the library I saw that all io-pluggins returned hard coded units.

ericpre:
The units issue is fixed in #1730.

PR #1729, opened 2017-09-13
Fix quant scaling error

k8macarthur:
Fix to the quantification function so that the units of the scaling are checked before determining electron dose. Otherwise it raises a ValueError.


k8macarthur:
This is written to fix part of #1726 .

francisco-dlp:
Should this go to Rnp instead?

Wouldn't it be better to use pint to fix this for good i.e. support all length units?

ericpre:
I am finishing the conversion of scale and units using pint, so, once this is done, it would be easy to convert an axis to the desired units.

ericpre:
If we wait for #1730 to be merged into next_minor, this one should also go to next_minor!

k8macarthur:
Ok, I'll wait on your pull then @ericpre  assuming you would be able to have a function that simply converts everything to 'nm'?

ericpre:
Yes, if you wait for #1730, you will have a method of `DataAxis` to convert to the unit of your choice!

k8macarthur:
I'll revisit this when #1730 is accepted as I expect this correction wont fit anymore...

ericpre:
#1730 has been merged.

ericpre:
@k8macarthur, you can revisit this PR, now!

k8macarthur:
@ericpre  Great! I had a look at this today. I tried using s.axes_manager[0].convert_to_units()  but it says this function doesn't exist. Am I using this right?

ericpre:
The following works fine for me:
```python
import hyperspy.api as hs
s.axes_manager

<Axes manager, axes: (12|400)>
  Probe position |     12 |      0 |       0 |       1 |     nm 
Electron energy  |    400 |        |   4e+02 |       1 |     eV 

s.axes_manager[0].convert_to_units('µm')
s.axes_manager

<Axes manager, axes: (12|400)>
  Probe position |     12 |      0 |       0 |   0.001 |     um 
Electron energy  |    400 |        |   4e+02 |       1 |     eV 
```

ericpre:
You need to have hyperspy >= 1.4, other than that I don't see why it would not work.

k8macarthur:
I think my update of local branches didn't work for some reason... fixed now :)

ericpre:
Superseded by #2090.

PR #1730, opened 2017-09-13
Scale/units conversion 

ericpre:
This PR supersedes #1106.
- add the `convert_to_units` method to `DataAxis` (can be inplace or not)
- add the `convert_units` method to `AxisManager`
- support slicing with units
- by default, the units are converted automatically when loading
- add `scale_as_quantity` and `offset_as_quantity` properties.

Examples:
```python
>>> import hyperspy.api as hs
>>> axis = hs.hyperspy.axes.DataAxis(size=10, scale=0.1, offset=10, units='mm')
>>> print("scale: {}, offset: {}, units: {}".format(axis.scale, axis.offset, axis.units))
>>> axis.convert_to_units('um')
>>> print("scale: {}, offset: {}, units: {}".format(axis.scale, axis.offset, axis.units))
scale: 0.1, offset: 10.0, units: mm
scale: 100.0, offset: 10000.0, units: µm
>>> axis.convert_to_units('cm', inplace=False)
(100.00000000000001, 10000.0, 'um')
```

```python
>>> import numpy as np
>>> s = hs.signals.Signal1D(np.random.random((50, 50, 2048)))
>>> am = s.axes_manager
>>> for axis in am.navigation_axes:
>>>     axis.units = 'm'
>>>     axis.scale = 3e-9
>>> for axis in am.signal_axes:
>>>     axis.units = 'eV'
>>>     axis.scale = 1.e-3
>>> am2 = am.deepcopy()
>>> am
<Axes manager, axes: (50, 50|2048)>
     <undefined> |     50 |      0 |       0 |   3e-09 |      m 
     <undefined> |     50 |      0 |       0 |   3e-09 |      m 
     <undefined> |   2048 |        |       0 |  0.0015 |     eV 

>>> am.convert_units()
>>> am
<Axes manager, axes: (50, 50|2048)>
     <undefined> |     50 |      0 |       0 |       3 |     nm 
     <undefined> |     50 |      0 |       0 |       3 |     nm 
     <undefined> |   2048 |        |       0 |     1.5 |    meV 

>>> am2.convert_units('signal')
<Axes manager, axes: (50, 50|2048)>
     <undefined> |     50 |      0 |       0 |   3e-09 |      m 
     <undefined> |     50 |      0 |       0 |   3e-09 |      m 
     <undefined> |   2048 |        |       0 |     1.5 |    meV 

>>> am.convert_units('navigation', 'um')
>>> am
<Axes manager, axes: (50, 50|2048)>
     <undefined> |     50 |      0 |       0 |   0.003 |     µm 
     <undefined> |     50 |      0 |       0 |   0.003 |     µm 
     <undefined> |   2048 |        |       0 |  0.0015 |     eV 
```

``` python
>>> s.axes_manager[0].units = 'µm'
>>> s.isig[:'2000 nm'].data
array([0, 1, 2, 3])
```

``` python
>>> s.axes_manager[0].scale_as_quantity = '2.5 µm'
>>> s.axes_manager
<Axes manager, axes: (|10)>
     <undefined> |     10 |        |       0 |     2.5 |     µm 
>>> s.axes_manager[0].offset_as_quantity = '2.5 nm'
<Axes manager, axes: (|10)>
     <undefined> |     10 |        |     2.5 | 2.5e+03 |     nm 
```


ericpre:
I initially tried to convert the units only when plotting (without changing the `axes_manager` of the signal) and I was doing it by copying the `axes_manager` at the creation of the plot. For obvious reason it was not possible anymore to change the position of the pointer from the signal `axes_manager`, since both axes_manager were different. It could be possible to find workaround this, but I think that this more simple the way how it is now: converting automatically at loading or cropping.

francisco-dlp:
Very cool! Could you add an example showing how these new features work in practice to the original comment? I think that this is good practice as it makes it easier to understand what the PR does for users and reviewers (we should add it to the Dev Guide).

I think that automatically changing the units is dangerous. For example, to perform math operation on the data the units have to be consistent and it's going to be difficult to track when/where the units have changed if hspy does it automatically. We could check the units when performing math operations on signals but it is also common to perform math operations directly using the numpy array, so even that wouldn't fix the issue.

hyperspy/signal.py
```diff
@@ -1928,7 +1928,7 @@ def __call__(self, axes_manager=None):
             self.data.__getitem__(axes_manager._getitem_tuple))
 
     def plot(self, navigator="auto", axes_manager=None,
-             plot_markers=True, **kwargs):
+             convert_units=True, plot_markers=True, **kwargs):
```
ericpre: Actually, if the units are converted during loading, there is no point to set `convert_units=True` as default.

ericpre:
> Could you add an example showing how these new features work in practice to the original comment? I think that this is good practice as it makes it easier to understand what the PR does for users and reviewers (we should add it to the Dev Guide).

Maybe, we should set a template for issue and pull request? See [https://github.com/blog/2111-issue-and-pull-request-templates](https://github.com/blog/2111-issue-and-pull-request-templates).

francisco-dlp:
Sounds good for PRs. For issues it's a bit harder because people use issues to report bugs, ask for help with a certain problem, ask for features and having to fill a general template that doesn't match with the user topic might work as an (unwanted) deterrent.

doc/user_guide/tools.rst
```diff
@@ -829,7 +832,7 @@ using an external function can be more easily accomplished using the
 
   Rotation of images by the same amount using :py:meth:`~.signal.BaseSignal.map`.
 
-The :py:meth:`~.signal.BaseSignal.map` method can also take variable
+The :py:meth:`~.signal.Bfancy inaseSignal.map` method can also take variable
```
to266: Not required

hyperspy/axes.py
```diff
@@ -825,6 +990,98 @@ def _on_scale_changed(self):
...
+
+        if axes is None:
+            axes = self.navigation_axes + self.signal_axes
+        elif axes == 'navigation':
+            axes = self.navigation_axes
+            convert_signal = False
+        elif axes == 'signal':
+            axes = self.signal_axes
+            convert_navigation = False
+        elif isinstance(axes, DataAxis) or type(axes) is int or type(axes) is str:
```
to266: `isinstance(axes, (DataAxis, int, str))` should do the same

hyperspy/axes.py
```diff
@@ -825,6 +990,98 @@ def _on_scale_changed(self):
...
+            if not isinstance(axes, DataAxis):
+                axes = self[axes]
+            axes = (axes, )
+            convert_navigation = axes[0].navigate
+            convert_signal = not axes[0].navigate
```
to266: I would prefer
`conver_signal = not convert_navigation`
it's both faster (no `axes[0]` lookup) and clearer that the two are opposite

hyperspy/axes.py
```diff
@@ -825,6 +990,98 @@ def _on_scale_changed(self):
...
+        else:
+            raise TypeError(
+                'Axes type `{}` is not correct.'.format(type(axes)))
+
+        if type(units) is str or units is None:
```
to266: `isinstance(units, str)`

hyperspy/axes.py
```diff
@@ -825,6 +990,98 @@ def _on_scale_changed(self):
...
+            units = [units] * len(axes)
+        elif type(units) is list:
```
to266: `if isinstance(units, list) and len(units) != len(axes):`

i.e. combine the two conditionals

to266: Alternatively you could just loop the last one if not enough units are given:
```python
units = ['um', 'nm']
axes = [1,2,3]
units = units + units[-1:]*(len(axes) - len(units))
```

hyperspy/axes.py
```diff
@@ -825,6 +990,98 @@ def _on_scale_changed(self):
...
+        for axis, unit in zip(axes, units):
+            axis.convert_to_units(unit, factor)
+            if current_units is not None and current_units:
+                if current_units != axis.units:
+                    # take units along first axis and use that one
+                    for axis in axes:
+                        axis.convert_to_units(axes[0].units, factor)
+                    return
+            else:
+                current_units = axis.units  # initialisation first iteration
```
to266: If you only intend to keep the first `is not None and len(unit)` unit, why not find it first from the list, and then loop over the axes just once? 
If this seems acceptable, how about this:
```python
first_good_unit = next(filter(lambda x: x is not None and x, units))
# Change `micron` to `um` if you want here as well
for axis in filter(lambda ax: ax.unit != first_good_unit, axes):
    axis.convert_to_units(first_good_unit, factor)
```

to266 (changes requested):
A quick look through the code. Once corrected, will need another check

hyperspy/axes.py
ericpre: Combining both conditionals is not working here, because it will skip the case when units is a list with the correct length and it will end up raising an error in the `else` block.

I am a bit sceptical about extrapolating the list of units, for the case when navigation and signals axis are passed. I would prefer to force the user to specify the correct units.

hyperspy/axes.py
ericpre: I simplified here, actually it was very simple and really unnecessary complicated!!

ericpre:
@to266, would be possible for you to review this again?

hyperspy/axes.py
```diff
@@ -56,13 +61,142 @@ def generate_axis(offset, scale, size, offset_index=0):
...
+            warnings.warn('Unit "{}" not supported for conversion. Nothing '
+                          'done.'.format(units),
+                          UserWarning)
+            return True
+        return False
+
+    def _convert_compact_units(self, factor=0.25, inplace=True):
+        """ Convert units to "human-readable" units, which means with a 
+            convenient prefix. `factor` is a adjustage factor used to 
+            determine the prefix of the units (in combinaison with the size of
```
k8macarthur: *combination

hyperspy/axes.py
```diff
@@ -56,13 +61,142 @@ def generate_axis(offset, scale, size, offset_index=0):
...
+            self.offset = offset
+            self.units = units
+        else:
+            return scale, offset, units
+
+    def convert_to_units(self, units=None, inplace=True, factor=0.25):
+        """ Convert the scale and the units of the current axis. If the units
+        is not supported by the pint library, the scale and units are not
+        changed.
+
```
k8macarthur: If the unit of measure is not supported....

hyperspy/axes.py
```diff
@@ -56,13 +61,142 @@ def generate_axis(offset, scale, size, offset_index=0):
...
+        Parameters
+        ----------
+        units : {str | None}
+            Default = None
+            If str, the axis will converted to the provided units.
+            If `None`, the scale and the units are converted to the appropriate
```
k8macarthur: If str, the axis will be converted to the provided units.

hyperspy/axes.py
```diff
@@ -56,13 +61,142 @@ def generate_axis(offset, scale, size, offset_index=0):
...
+            Default = None
+            If str, the axis will converted to the provided units.
+            If `None`, the scale and the units are converted to the appropriate 
+            scale and units to avoid displaying scalebar with >3 digits or too 
+            small number. This can be tweaked by the `factor` argument.
+        inplace : bool
+            If `True`, convert the axis in place. if `False` return the  
+            `scale`, `offset` and `units`.
+        factor : float
+            `factor` is a adjustage factor used to determine the prefix of the
```
k8macarthur: 'factor' an adjustable value used to determine the prefix of the units.

hyperspy/axes.py
```diff
@@ -56,13 +61,142 @@ def generate_axis(offset, scale, size, offset_index=0):
...
+            If str, the axis will converted to the provided units.
+            If `None`, the scale and the units are converted to the appropriate 
+            scale and units to avoid displaying scalebar with >3 digits or too 
+            small number. This can be tweaked by the `factor` argument.
+        inplace : bool
+            If `True`, convert the axis in place. if `False` return the  
+            `scale`, `offset` and `units`.
+        factor : float
+            `factor` is a adjustage factor used to determine the prefix of the 
+            units (in combinaison with the size of the axis).
```
k8macarthur: *combination

hyperspy/axes.py
```diff
@@ -825,6 +997,103 @@ def _on_scale_changed(self):
     def _on_offset_changed(self):
         self.events.any_axis_changed.trigger(obj=self)
 
+    def convert_units(self, axes=None, units=None, same_units=True,
+                      factor=0.25):
+        """ Convert the scale and the units of the selected axes. If the units
+        is not supported by the pint library, the scale and units are not
```
k8macarthur: If the unit of measure is not supported....

hyperspy/axes.py
```diff
@@ -825,6 +997,103 @@ def _on_scale_changed(self):
...
+        changed.
+
+        Parameters
+        ----------
+        axes : {int | string | iterable of `DataAxis` | None}
+            Default = None
+            Convert to the convenient scale and units on the specified axis.
```
k8macarthur: Convert to a convenient scale

hyperspy/axes.py
```diff
@@ -825,6 +997,103 @@ def _on_scale_changed(self):
...
+            If int, the axis can be specified using the index of the
+            axis in `axes_manager`.
+            If string, argument can be `navigation` or `signal` to select the
+            navigation or signal axes. The axis name can also be provided.
+            If `None`, convert all axes.
+        units : {list of string of the same length than axes | str | None}
+            Default = None
+            If list, the selected axes will be converted to the provided units.
+            If str, the navigation or signal axes will converted to the
```
k8macarthur: will be converted

hyperspy/axes.py
```diff
@@ -825,6 +997,103 @@ def _on_scale_changed(self):
...
+            scale and units to avoid displaying scalebar with >3 digits or too 
+            small number. This can be tweaked by the `factor` argument.
+        same_units : bool
+            If `True`, force to keep the same units if the units of
+            the axes differs. It only applies for the same kind of axis, 
+            `navigation` or `signal`. By default the converted units of the 
+            first axis is used for all axes. If `False`, convert all axes 
+            individually.
+        factor : float
+            `factor` is a adjustage factor used to determine the prefix of the
```
k8macarthur: 'factor' an adjustable value used to determine the prefix of the units.

hyperspy/io_plugins/blockfile.py
```diff
@@ -107,23 +107,29 @@ def get_header_from_signal(signal, endianess='<'):
...
     else:
         note = ''
-    if signal.axes_manager.navigation_dimension == 2:
-        NX, NY = signal.axes_manager.navigation_shape
-        SX = signal.axes_manager.navigation_axes[0].scale
-        SY = signal.axes_manager.navigation_axes[1].scale
-    elif signal.axes_manager.navigation_dimension == 1:
-        NX = signal.axes_manager.navigation_shape[0]
+    # The navigation and signal units are 'nm' and 'cm', respectively, so we
+    # convert the units according before saving the signal
```
k8macarthur: accordingly

k8macarthur:
Found a few typos in the docstrings.

francisco-dlp:
@ericpre, could you merge Rnm into this?

hyperspy/axes.py
```diff
@@ -57,13 +63,142 @@ def generate_axis(offset, scale, size, offset_index=0):
...
+            self.scale = scale
+            self.offset = offset
+            self.units = units
+        else:
+            return scale, offset, units
+
+    def convert_to_units(self, units=None, inplace=True, factor=0.25):
+        """ Convert the scale and the units of the current axis. If the unit 
+        of measure is not supported by the pint library, the scale and units 
+        are not changed.
```
francisco-dlp: changed -> modified?

hyperspy/axes.py
```diff
@@ -57,13 +63,142 @@ def generate_axis(offset, scale, size, offset_index=0):
...
+
+        Parameters
+        ----------
+        units : {str | None}
+            Default = None
+            If str, the axis will be converted to the provided units.
+            If `None`, the scale and the units are converted to the appropriate 
+            scale and units to avoid displaying scalebar with >3 digits or too
```
francisco-dlp: What about changing ``None`` to ``"auto"``?

Simpler docstring: If "auto", automatically determine the optimal units too big or too small numbers.

hyperspy/axes.py
```diff
@@ -57,13 +63,142 @@ def generate_axis(offset, scale, size, offset_index=0):
...
+        units : {str | None}
+            Default = None
+            If str, the axis will be converted to the provided units.
+            If `None`, the scale and the units are converted to the appropriate 
+            scale and units to avoid displaying scalebar with >3 digits or too 
+            small number. This can be tweaked by the `factor` argument.
+        inplace : bool
+            If `True`, convert the axis in place. if `False` return the  
+            `scale`, `offset` and `units`.
+        factor : float
```
francisco-dlp: It would be good to give more details on how ``factor`` affects the automatic units conversion.

hyperspy/axes.py
```diff
@@ -57,13 +63,142 @@ def generate_axis(offset, scale, size, offset_index=0):
...
+        return self._units
+
+    @units.setter
+    def units(self, s):
+        if s == t.Undefined:
+            self._units = s
+        elif s == '':
+            self._units = t.Undefined
+        else:
+            self._units = s.replace('um', 'µm').replace(' ', '')
```
francisco-dlp: I think that it'll be better to avoid preprocessing the string—if that was a good idea pint would do it.

hyperspy/axes.py
```diff
@@ -57,13 +63,142 @@ def generate_axis(offset, scale, size, offset_index=0):
     Numpy array
 
     """
+
     return np.linspace(offset - offset_index * scale,
                        offset + scale * (size - 1 - offset_index),
                        size)
 
 
+class UnitConversion(object):
```
francisco-dlp: ``object`` not needed in Python3

hyperspy/axes.py
```diff
@@ -462,6 +606,35 @@ def update_from(self, axis, attributes=["scale", "offset", "units"]):
             any_changes = True
         return any_changes
 
+    @property
+    def scale_as_quantity(self):
```
francisco-dlp, line 615: These new properties are not documented in the User Guide.

hyperspy/axes.py
```diff
@@ -462,6 +606,35 @@ def update_from(self, axis, attributes=["scale", "offset", "units"]):
...
+        return self._units
+
+    @units.setter
+    def units(self, s):
+        if s == t.Undefined:
+            self._units = s
+        elif s == '':
+            self._units = t.Undefined
+        else:
+            self._units = s.replace('um', 'µm').replace(' ', '')
```
francisco-dlp: Same as above.

hyperspy/axes.py
```diff
@@ -826,6 +999,103 @@ def _on_scale_changed(self):
...
+            If `None`, the scale and the units are converted to the appropriate 
+            scale and units to avoid displaying scalebar with >3 digits or too 
+            small number. This can be tweaked by the `factor` argument.
+        same_units : bool
+            If `True`, force to keep the same units if the units of
+            the axes differs. It only applies for the same kind of axis, 
+            `navigation` or `signal`. By default the converted units of the 
+            first axis is used for all axes. If `False`, convert all axes 
+            individually.
+        factor : float
```
francisco-dlp: As ``factor`` appears in many docstrings, it could be extracted into a single string and used where needed.

hyperspy/io_plugins/blockfile.py
```diff
@@ -38,7 +38,7 @@
 # Recognised file extension
 file_extensions = ['blo', 'BLO']
 default_extension = 0
-
+auto_convert_units = True
```
francisco-dlp: This does not appear to be used anywhere

hyperspy/io.py
```diff
@@ -105,13 +106,16 @@ def load(filenames=None,
         until it finds a name that is not yet in use.
     lazy : {None, bool}
         Open the data lazily - i.e. without actually reading the data from the
-        disk until required. Allows opening arbitrary-sized datasets. default
+        disk until required. Allows opening arbitrary-sized datasets. The default
         is `False`.
+    convert_units : {None, bool}
```
francisco-dlp: I am not sure if it is a good idea to let the reader decide if to convert units or not. I would rather default to ``False`` and, if the user finds that the units are not convenient, he can always call the ``AxesManager.convert_units`` manually.

hyperspy/io_plugins/README
```diff
@@ -13,6 +13,7 @@ All the read/write plugins must provide a python file containing:
         # Recognised file extension
         file_extensions = <Tuple of string>
         default_extension = <Int>	# Index of the extension that will be used by default
+        auto_convert_units = <Bool> # Convert automatically the units
```
francisco-dlp: As above, I think that it'll be better not to hardwire this option in the reader.

hyperspy/signal.py
```diff
@@ -1971,6 +1973,7 @@ def plot(self, navigator="auto", axes_manager=None,
                 "'s.transpose(signal_axes=(1,2)).plot()' "
                 "for plotting as a 2D signal.")
 
+        self._plot.auto_convert_units = convert_units
```
francisco-dlp: What's the use of this?

hyperspy/signal.py
```diff
@@ -1951,6 +1951,8 @@ def plot(self, navigator="auto", axes_manager=None,
 
         if axes_manager is None:
             axes_manager = self.axes_manager
+        if convert_units:
```
francisco-dlp: I don't think that it is a good idea to add this option to ``plot``: the user may not be aware that this is actually changing the AxesManager units in-place. If that's what the user wants to do, he could call ``AxesManager.concert_units`` manually before calling plot.

It would be possible to change the units only for the plot axes_manager, but that would lead to issues with the ROIs.

doc/user_guide/tools.rst
```diff
@@ -453,6 +453,7 @@ features differ from numpy):
```
francisco-dlp, line 453: I suggest to add a section to this chapter of the User Guide explaining how AxesManager manages the units and the new ``convert_units`` ``AxesManager`` and ``DataAxis`` methods.

francisco-dlp (changes requested):
This is a very welcome new feature. I've left some few comments that need to be addressed before merging.

francisco-dlp:
On a side note: would it be possible (in future pull requests) to:

* Force some objects (e.g. ROIs) to always operate on SI units regardless of the AxesManager units to make it possible to operate with the same ROI on axes with different units
* Add units in this way to all metadata numbers and method/function parameters and convert to the right units internally before performing the operations?

hyperspy/axes.py
ericpre: I have opened an issue in the [pint repo](https://github.com/hgrecco/pint/issues/666) to ask. Pint seems to have a fairly good unicode support, so I don't see, why it would not work. Maybe, to avoid issue with python < 2.6?

hyperspy/signal.py
ericpre: Yes, it is removed now.

hyperspy/io.py
ericpre: This is actually one of the motivation of this PR: for example, the fei reader convert scale and offset systematically to "nm" for convenience, because the file is saved with the units in "m".  Same for the bruker reader, the units are saved in "um" in the file, which is not great for nanometer scale EDS mapping. 
Allowing to have a reader parameter to determine if the units needs to be converted or not is an improvement to the current situation. Only the reader, which present this issue have `auto_convert_units=True`.

ericpre:
I have addressed all comments.

> On a side note: would it be possible (in future pull requests) to:
> 
> - Force some objects (e.g. ROIs) to always operate on SI units regardless of the AxesManager units to make it possible to operate with the same ROI on axes with different units
> - Add units in this way to all metadata numbers and method/function parameters and convert to the right units internally before performing the operations?

Yes, it is sensible and since this PR already changes many file, better to do in another one, as you suggested!

hyperspy/axes.py
francisco-dlp: Personally I would prefer not to perform this translation in HyperSpy as it can lead to unexpected issues. For example, somebody may define an unit that contains the "um" chain. Also, it is not so hard to write μ even I just did it!

hyperspy/io.py
francisco-dlp: I still think that it is better to default to "don't autoconvert units" and not to hardwire any different behaviour as this is not transparent for the user. Units auto conversion is convenient, but it is somehow unpredictable: it can change with different hyperspy and pint versions and and files stored in the same format could return different units depending on the data stored—too much unpredictability IMO.

Personally I think that the units in the file must be respected i.e. if FEI writes everything in S.I. we should read the data in S.I. by default. I know that this is not the current situations, and for me this fixes it as long as it respects the file units by default. The reason why some file readers automatically convert units is that there was no easy way of converting units before you implemented this feature.

hyperspy/axes.py
ericpre: Yes, indeed, this is not right, I revert it.

hyperspy/io.py
ericpre: Since it is only the FEI reader and some tiff files which can generate inappropriate units, I removed the auto_convert_units reader parameter and I reverted to the previous conversion in the FEI reader, since I would be annoyed to have to write `convert_units=True` each time I want to load an emi/ser file.

doc/user_guide/io.rst
```diff
@@ -75,6 +75,13 @@ Almost all file readers support accessing the data without reading it to memory
 analysing large files. To load a file without loading it to memory simply set
 ``lazy`` to ``True`` e.g.:
 
+The units of the navigation and signal axes can be converted automatically 
+during loading using the ``convert_units`` parameter. If `True`, the 
+``convert_to_units`` method of the ``axes_manager`` will be used for the conversion.
+If `None` (default), the default setting of the file reader will be used to
```
francisco-dlp: I think that this part of the documentation needs updating to take into account the latest changes.

francisco-dlp (commented):
Will LGTM as soon as the little doc issue is fixed and Rnm merged in.

francisco-dlp:
unittests are failing

ericpre:
Sorry, usually I checked but I haven't this time... it should be fine now!

PR #1733, opened 2017-09-18
Add 'line_width' in the plot_spectra function

MartialD:
Update utils.py
Add the line_width command in the plot_spectra function and sub function

### DIFF
--- a/hyperspy/drawing/utils.py
+++ b/hyperspy/drawing/utils.py
@@ -338,7 +338,7 @@ def _make_heatmap_subplot(spectra):
     return im._plot.signal_plot.ax
 
 
-def _make_overlap_plot(spectra, ax, color="blue", line_style='-'):
+def _make_overlap_plot(spectra, ax, color="blue", line_style='-', line_width=2.0):
     if isinstance(color, str):
         color = [color] * len(spectra)
     if isinstance(line_style, str):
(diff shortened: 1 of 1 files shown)


francisco-dlp:
The tests are not passing because this changes the default linewidth from mpl default (1) to 2. I think that it'll be better to keep the current default. Also adding a little test to check that the new arguments is well taken into account would be a good idea.

ericpre:
What's about adding a `**kwargs_plot` to pass to `matplotlib.axes.Axes.plot`? In the `plot_spectra` function, there is already a `**kwargs`, which is passed to `matplotlib.pyplot.figure` or `matplotlib.pyplot.subplots`.

MartialD:
@ericpre : It is a good idea, but  **kwargs (acting on matplotlib.pyplot.figure or subplot) already have a 'linewidth' parameters (acting on the frame), how can we differentiate 'linewidth' in **kwargs and **kwargs_plot?

francisco-dlp:
Actually, I think that for a function that plots spectra it makes more sense to pass the extra kwargs to plot rather than figure. This would be a API change i.e. it'll go to hyperspy 2.0, which is due as soon as we split hyperspy (hopefully in the next 6 months). 

So, what about:

* ``**kwargs`` go to mpl plot
* new ``fig_kwargs`` takes a dictionary and goes to mpl fig

ericpre:
Yes, indeed, my previous comment was not very pertinent, since it is not possible to have two **kwargs at the same time... but a pattern similar to [signal plot](https://github.com/hyperspy/hyperspy/blob/RELEASE_next_minor/hyperspy/drawing/mpl_he.py#L162), as @francisco-dlp suggested, would do the job!

PR #1740, opened 2017-09-27
Fourier log cannot cast fix

erh3cq:
Durring fourier_log deconvolution an error occurs:
`Cannot cast ufunc multiply output from dtype('float64') to dtype('int32') with casting rule 'same_kind'`
This is meant to fix the issue by allowing cast='unsafe'.  The issue is that a spectrum is dtype(int), but the hanning filter applied is of dtype(float).  I am not sure if this is the best route to fix the issue, or if it is better to make the hanning filter return integers (since the deconvolved spectra should be dtype(int)). Either way, this allows for the code to execute.


francisco-dlp:
Good point. At present, HyperSpy is not good at preserving the int dtype i.e. some routines assume that the dtype is `float` and won't run if it is `int`. This is one of them. I think that better than making it run without complaining (and unexpectedly returning float data) is to make it raise an informative error that explains that deconvolution requires float input. What do you think?

din14970:
experienced this issue as well (as a casual user). What worked for me is simply turning the data into floats -> s.data = s.data.astype("float").

francisco-dlp:
That's the proper way to proceed. The actual bug is not raising a more informative error message suggesting to do just that ;)

thomasaarholt:
Just to raise awareness of it, we have the signal method `s.change_dtype(float)`.

erh3cq:
Sorry, I didn't get an update about this post. @francisco-dlp I agree, the bug is not raising an error of how to fix the issue which is fixed by @thomasaarholt and @din14970 solutions.  Possibly an improved bug message?

thomasaarholt:
Improved message sounds good :)

erh3cq:
I have improved the message. Sorry for the long list of commits...I had to reset the branch.

ericpre:
I don't know what it is going on with the git history but it doesn't look right. Either you sort it out or you cherry pick your the useful commits and make a new PR, because this will be impossible to review properly.

erh3cq:
@ericpre Done in #1879

PR #1741, opened 2017-09-27
Add hack for BCF used MSXML decimal non-compliance 

sem-geologist:
There is many bugs in Esprit and its file formats.
The non-compliance to XML standard is well know (one of the reason behind using lxml, is often inappropriate formatted XML).
Suddenly another non-compliance were found. Some (not all) of XML values are saved in the localized form.
The XML bugs are introduced by MSXML implementation wrapped by standard delphi libraries, which is used by Bruker in Esprit. Not to mention that fixing it (MSXML) is not possible as it is there for nearly 2 decades.
Locale converted numbers creeps only into limited XML nodes, only single numeric value bearing nodes. Arrays are not affected, they are serialised into XML complying the standard.
No problem is caused, while locale on Windows where Esprit is installed stays en_US; however setting it to any locale with decimal separator other than dot (`.`) makes evaluation of non-compliant decimal, XML serialized, values to fail, leading to recognition of the  decimal value as list of length=2, leading to TraitError.

~~This PR adds a short note to io.py if TraitError is raised during bcf loading,
and it adds the possibility to temporary enable other number format checking, which is set by an added function.~~
The check and fix can ~~'t~~ be automated, but this is rather a hack around---not the bug-fix of--- the Bruker's->delphi->MSXML bug.  

~~- [x] updated io.py to throw information when the TraitError is raised, and how to get around it~~
~~- [x] bcf.py - inserted the function and updated other to set decimal and thousand separators, using it for decimal search and  normalization of it when parsing XML.~~
- [x] bcf.py - injected regular expression to find and replace decimal commas before parsing XML string
- [x] ~~tests? test for hack around?~~ test the compiled regular expression feeding positive and negative XML snippets
- [x] ready for review (if no tests is needed, it should be ready)

### DIFF
--- a/hyperspy/tests/io/test_bcf.py
+++ b/hyperspy/tests/io/test_bcf.py
@@ -198,8 +198,8 @@ def test_wrong_file():
 def test_decimal_regex():
     lxml = pytest.importorskip("lxml")
     from hyperspy.io_plugins.bcf import fix_dec_patterns
-    dummy_xml_positive = b'<dummy_tag>85,658<\\dummy_tag>'
-    dummy_xml_negative = b'<dum_tag>12,25,23,45,56,12,45<\\dum_tag>'
+    dummy_xml_positive = b'<dummy_tag>85,658</dummy_tag>'
+    dummy_xml_negative = b'<dum_tag>12,25,23,45,56,12,45</dum_tag>'
     assert b'85.658' in fix_dec_patterns.sub(b'\\1.\\2', dummy_xml_positive)
     assert b'.' not in fix_dec_patterns.sub(b'\\1.\\2', dummy_xml_negative)
     
--- a/hyperspy/tests/io/test_bcf.py
+++ b/hyperspy/tests/io/test_bcf.py
@@ -198,8 +198,13 @@ def test_wrong_file():
 def test_decimal_regex():
     lxml = pytest.importorskip("lxml")
     from hyperspy.io_plugins.bcf import fix_dec_patterns
-    dummy_xml_positive = b'<dummy_tag>85,658</dummy_tag>'
-    dummy_xml_negative = b'<dum_tag>12,25,23,45,56,12,45</dum_tag>'
-    assert b'85.658' in fix_dec_patterns.sub(b'\\1.\\2', dummy_xml_positive)
-    assert b'.' not in fix_dec_patterns.sub(b'\\1.\\2', dummy_xml_negative)
+    dummy_xml_positive = [b'<dummy_tag>85,658</dummy_tag>',
+                          b'<dummy_tag>85,658E-8</dummy_tag>',
+                          b'<dummy_tag>-85,658E-8</dummy_tag>']
+    dummy_xml_negative = [b'<dum_tag>12,25,23,45,56,12,45</dum_tag>',
+                          b'<dum_tag>12e1,23,-24E-5</dum_tag>']
+    for i in dummy_xml_positive:
+        assert b'85.658' in fix_dec_patterns.sub(b'\\1.\\2', i)
+    for j in dummy_xml_negative:
+        assert b'.' not in fix_dec_patterns.sub(b'\\1.\\2', j)
     
--- a/hyperspy/tests/io/test_bcf.py
+++ b/hyperspy/tests/io/test_bcf.py
@@ -200,6 +200,6 @@ def test_decimal_regex():
     from hyperspy.io_plugins.bcf import fix_dec_patterns
     dummy_xml_positive = b'<dummy_tag>85,658<\\dummy_tag>'
     dummy_xml_negative = b'<dum_tag>12,25,23,45,56,12,45<\\dum_tag>'
-    assert '85.658' in fix_dec_patterns.sub(b'\\1.\\2', dummy_xml_positive)
-    assert '.' not in fix_dec_patterns.sub(b'\\1.\\2', dummy_xml_negative)
+    assert b'85.658' in fix_dec_patterns.sub(b'\\1.\\2', dummy_xml_positive)
+    assert b'.' not in fix_dec_patterns.sub(b'\\1.\\2', dummy_xml_negative)
     
--- a/hyperspy/tests/io/test_bcf.py
+++ b/hyperspy/tests/io/test_bcf.py
@@ -193,7 +193,7 @@ def test_wrong_file():
     lxml = pytest.importorskip("lxml")
     filename = os.path.join(my_path, 'bcf_data', 'Nope.bcf')
     with pytest.raises(TypeError):
-        load(filename)fix_dec_patterns = re.compile(b'(>-?\d+),(\d*<)')
+        load(filename)
 
 def test_decimal_regex():
     lxml = pytest.importorskip("lxml")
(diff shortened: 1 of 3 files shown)


sem-geologist:
@thomasaarholt , could you test this PR, as you had previously reported the problem in gitter?
What locale OS is set where yours Esprit 1.9 is installed?

thomasaarholt:
I guarantee that there'll be a mix of locale's on our system(s) at work. Norwegian (,) and English (.).

I'll have a go tomorrow!

sem-geologist:
I was just digging deeper trying to find out where the problem orginates.
[http://www.codenewsfast.com/cnf/thread/0/permalink.thr-ng1922q1099](http://www.codenewsfast.com/cnf/thread/0/permalink.thr-ng1922q1099)

Looks it is actually windows xml implementation to blame, as embarco(TM,(R)) delfi does not ship its own xml engine but uses that crap.
I am still not sure about thousand separators, I am not sure it is saved into xml...

thomasaarholt:
How does esprit itself differentiate between the locales? Surely there is some indicator within the bcf file?

sem-geologist:
Esprit saves no information about locale, I think Bruker could be even not aware about the bug, which is so very Window'ish 
[http://www.devsuperpage.com/search/Articles.asp?ArtID=506968](http://www.devsuperpage.com/search/Articles.asp?ArtID=506968)

The delphi/pascal Xml libraries used by bruker, wrapps MSXML implementation---one huge crap.
I am not sure but it is very possible that some of the values could be not read in between few different windows locales/esprit.
I will test it tomorrow.

sem-geologist:
Ok I have tested comma bearing file on esprit on en_US, it opens correctly. I guess Bruker is aware of the MSXML shortcomming and uses some workarround

ericpre:
Could you please add a test to check that it replaces correctly the string?
Just create a few dummy minimal strings for the different cases and run your regular expression substitution on it and check that the results is what you expect.
I would also suggest to describe what is doing the regular expression, it is not easily readable and it can be an easy source of mistakes.

sem-geologist:
i think the used regular expression is the most elegant way.

The problem: 
1. str.replace() can't be used globaly, as xml contains also lists.
2. str.replace() could be used on the single xml nodes however it would cost to check every string with re, to find comma decimals, that would be done during objectifying the xml.

Solution: intercept the xml string after extracting it from bcf file, just before objectifying it, and search and replace comma decimal separators globaly.

What should pass:
```xml
<some_tag>digits,digits</some_tag>
```
yes, that could also be list of two elements, but luckily there are none in Esprit format.

what should not pass:
```xml
<some_tag>digits,digits,digits,digits</some_tag> #lists with three or more elements
<some_tag>digits,letters</some_tag>
```
we are interested only in commas, not whole strings.
to get only comma we could use look-behind and look-forward:
```python
import re
reg_expression = re.compile(r'(?<=\d),(?=\d))
```
`reg_expression.findall(xml)` would return only commas, however it would get commas from lists too.
Too exclude commas from lists, it could be improved by expanding `\d` with `*` and adding tag marks: `>`at look behind and `<` to look-forward:
```python
import re
reg_expression = re.compile(r'(?<=>\d*),(?=\d*<))
```
That however fails in python, as look-behind have to be fixed length.
The working alternative solution is to use [capturing group](https://www.regular-expressions.info/brackets.html), which can be then referenced with \diggit (e.g. \1, \2) notation.
So instead of look-behind and look forward:
```python
reg_expression = re.compile(r'(>\d*),(\d*<))
```
those becomes the groups and can be restored with `reg_expression.sub(r'\1.\2', xml_str)`.
Finally, possible negative numbers have to be addressed and `r''` changed into `b''` as xml_str is binary string.
So the final implemented version is like this:
```python
import re
reg_expression = re.compile(b'(>-?\d*),(\d*<))
reg_expression.sub(b'\\1.\\2', xml_str)
```

~~I don know how to explain the regular expression shortly and if it is a good idea to explain it in the code. There are two groups of people: one understand regular expressions, other don't.~~

sem-geologist:
@thomasaarholt , I changed the implementation significantly, now it should automatically check for decimal commas. Could you check your Norwegian locale bcf files?

thomasaarholt:
Hi @sem-geologist I'm afraid it still does not work, with the same error being reported.
Here is the file I'm testing. There is a chance there are multiple bugs that are confusing us.

[8.zip](https://github.com/hyperspy/hyperspy/files/1366831/8.zip)

sem-geologist:
@thomasaarholt , thank you very much for your file sample, thanks to it the fix is coming.
The problem is that I never tested on our microscopes (polish locale, thus comma as decimal separator) nano scale hypersmaps, which your file got covered. I was not aware that bruker would use  `d.ddE-d` notation, particularly that I saw values (>1) such as 404231,412345 not 4,04231412345E5. Maybe it depends from particular Esprit version.
Anyway, improved regular expression will take care.

sem-geologist:
```python
reg_expression = re.compile(b'(>-?\d*),(\d*([Ee]-?\d*)?<)')
```
captures the troublesome scientific notation. I could limit use to only uppercase E, but I think using upper and lower `[Ee]` is safer, as that could be influenced by locale (I guess).

thomasaarholt:
Well done! I'm looking forward to testing this now!

ericpre:
There is no doubt that regular expression would be an elegant way to fix the issue. And having explanation in the code is very important, as it helps to make the code sustainable. If you or anyone else need to work on this part of the code in a couple of years time, I am sure that the explanation will make our life easier! :)
Maybe it is worth adding a link to this PR in the code?

sem-geologist:
@ericpre , the 6 line explanation before definition of regex is not enought? i dont think we should rewrite `re` manual. this is very basic re pattern.

hyperspy/tests/io/test_bcf.py
```diff
@@ -194,3 +194,17 @@ def test_wrong_file():
...
+    dummy_xml_positive = [b'<dummy_tag>85,658</dummy_tag>',
+                          b'<dummy_tag>85,658E-8</dummy_tag>',
+                          b'<dummy_tag>-85,658E-8</dummy_tag>']
+    dummy_xml_negative = [b'<dum_tag>12,25,23,45,56,12,45</dum_tag>',
+                          b'<dum_tag>12e1,23,-24E-5</dum_tag>']
+    for i in dummy_xml_positive:
+        assert b'85.658' in fix_dec_patterns.sub(b'\\1.\\2', i)
+    for j in dummy_xml_negative:
+        assert b'.' not in fix_dec_patterns.sub(b'\\1.\\2', j)
+
```
ericpre, line 212: Add a "negative control", i.e. a test to check that a decimal number originally formatted as `85.658` stays as `85.658`?

sem-geologist: Negative control had been added

ericpre:
@sem-geologist, can you confirm that the failure of the bcf tests in [python 3.5 on linux](https://travis-ci.org/hyperspy/hyperspy/jobs/285813232) is related to the lxml library?

sem-geologist:
Yes, lxml were took out from the default conda library list.
or something else (faulty linking of lxml library?)..., anyway, it is clear that lxml is buggy in the installation.
That mobilised me to do #1750

Issue #1739, opened 2017-09-27
Improve reading/writing speed of hspy files


blazzter:
Is there a way how to make hdf5 faster to read/write? If I convert my data (>2GB) to h5py, it just takes much much longer to load it compared to simple binary read.

robindegen:
I've noticed it really depends on how you write it. For example if your dataset is expandable it's slower than if it's assigned it's required size up front. Also the chunk size matters greatly in terms of peformance.

magnunor:
I've seen the same thing myself, with binary files loading much faster compared to HDF5-files.

One trick is using `s = hs.load("afile.hdf5", lazy=True)`, which means the data is not loaded into memory, unless you need it to be. See http://hyperspy.org/hyperspy-doc/current/user_guide/big_data.html#working-with-big-data for more info.

I've especially found the plotting example to be very useful:
```python
# for fastest results, just pick one signal space pixel
nav = s.isig[256, 256].T
# Alternatively, sum as per default behaviour
nav = s.sum(s.axes_manager.signal_axes).T
nav.compute()

s.plot(navigator=nav)
```

ericpre:
Yes, there may be room for improvement of reading/writing speed of hspy files; it needs to be checked. I rename the title of the issue to more it more clear.

magnunor:
Making things read faster might be difficult, since our read speed is limited by how `h5py` does it. The fact that binary files or the same data saved as a flat NumPy files loads faster, might be due to various overheads in the `h5py` file-reader. This _might_ be improved by making it run in parallel mode (http://docs.h5py.org/en/latest/mpi.html), but I haven't tried this.

Another important factor is how we save the data, as mentioned by @robindegen the chunking size and other settings can have very large effects on the time it takes to load the data. So we might be able to improve the loading time of data saved in HyperSpy, by changing some of these settings.

ericpre:
[...]

PR #1743, opened 2017-09-28
Faster saving hspy format

to266:
Avoids converting numpy array to dask array (which is slow for large signals because it has to build a large DAG) when saving.

If people from https://github.com/hyperspy/hyperspy/issues/1739 could verify that saving is indeed faster, that would be good.

### DIFF
--- a/hyperspy/io_plugins/hspy.py
+++ b/hyperspy/io_plugins/hspy.py
@@ -558,8 +558,10 @@ def overwrite_dataset(group, data, key, signal_axes=None, **kwds):
     else:
         if isinstance(data, da.Array):
             da.store(data.rechunk(dset.chunks), dset)
+        elif data.flags.c_contiguous:
+            dset.write_direct(data)
         else:
-            da.store(da.from_array(data, chunks=dset.chunks), dset)
+            dset[:] = data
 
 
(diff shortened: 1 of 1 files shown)


ericpre:
Yes, this is exactly what we need to pay attention to: using properly the different libraries together!!

Would it be worth having a function to benchmark writing and reading files, so that we do this kind of thing, we can monitor the improvement and document it in the code itself?

thomasaarholt:
Well said, @ericpre. And smart thinking, @to266.

I really like the idea of benchmarking speeds of things, as it would make improvements really obvious.

For instance, benchmarking the time of `import hyperspy.api` would be nice.

francisco-dlp:
This LGTM, @thomasaarholt. @ericpre?

thomasaarholt:
~~I did not experience a speed-up on saving a 350 MB file. Am I missing something?
Note that I changed branches and restarted the kernel between the two tries.~~
Ignore. Had the wrong branch.

The following image uses the correct branch. There is a slight increase in speed ~5%.
image

Otherwise looks good.

thomasaarholt:
I just ran it again with a larger file size (2GB) and saw a 9% speedup.
This is nice. 

LGTM.

PR #1749, opened 2017-10-05
Holography: set high_cf default to True in `reconstruct` and `estimate_sideband_position` methods 

woozey:
### Description of the changes
Low pass filtering  (new feature in current state of the branch) in `misc.holography.reconstruct.estimate_sideband_position()` regulated by `high_cf` parameter caused holography demo delivering wrong results. See https://github.com/hyperspy/hyperspy-demos/issues/40. The problem is default value of `high_cf` which was set to `False`. After discussion with @FWin22 and @CodeMonkeyJan who implemented the low pass filtering we have decided that the default value should be set to `True` which would be more natural for users.
- Changed parameter `high_cf default value to `True` in `HologramImage` and in `misc.holography`
- Cleaned `misc.holography.reconstruct.reconstruct` from unused `high_cf` parameter

### Progress of the PR
- [x] Changes implemented,
- [x] update docstring,
- [x] ready for review.

### Example
```python
>>> import hyperspy.api as hs
>>> sb_position = ref.estimate_sideband_position(sb='upper')
>>> sb_position.data
array([390,  58])
>>> # Correct output. Current state would deliver [451,  29] instead.
```


jan-car:
Just to emphasize: The new default value `high_cf=True` will be the more generous case, which lets the side-band-finding method search a wider range of frequencies, which is good.
If problems arise with a side-band unexpectedly found in the high frequency range, the user can then re-run the side-band-finder with excluded high frequencies by setting `high_cf` to `False`.
Sounds good to me!

EDIT: The CI problems seem to stem from something unrelated (PIL and the bcf reader?). Not sure what the problem is...

ericpre:
Thanks!
Yes, travis and appveyor are not very fit these days... ;)

PR #1750, opened 2017-10-05
ENH: BCF xml parsing: drop lxml, use xml.etree

sem-geologist:
### This PR drops usage of lxml and adopts built-in xml.etree.ElementTree
- it is not a first time when travis ci and appveyor fail/skips tests due to lxml geting inculded in/out from default anaconda library list.
- xml.etree is faster at parsing
- reduces installation complexity (so that bcf would be installed even with minimal installation)
- reduces dependency (as etree is built-in and distributed with python)

### Progress of the PR
- [x] getting working replacement of bcf.py
  - [x]  adopt complicated xpath definitions to significantly narrowed down etree xpath subset (fix overview image reading)
- [x] remove lxml checks in other files (`io/__init__.py`),
- [x] update relevant parts in documentations
- [x] update installation scripts dropping lxml dependency (setup.py, .travis.yml, apveyor.yml ...)
- [x] check tests and do some adjustment if needed to pass all of them, (original metadata structure is not changed ~~going to be very minory changed, adding the XML attributes with '@attribute' notation~~)
- [x] ~~find faulty xml example (I believe I encountered one once), and adopt regex code for its fixing prior-parsing, (containing numbers in the tags: e.g. ```<1>good day</1>```)~~ can't find such faulty bcf file. I have only bruker eds `*.spx` and project files with messy xml tags. Without working bug example this will wait until I or somebody will find such buggy bcf file.
- [x] ready for review.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -33,7 +33,7 @@ before_install:
 install:
 
   - if [[ $MINIMAL_ENV == 'False' ]] ; then
-      DEPS="pip numpy scipy matplotlib>=2.0.2 ipython h5py sympy scikit-learn dill natsort setuptools scikit-image cython lxml ipyparallel dask traits traitsui";
+      DEPS="pip numpy scipy matplotlib>=2.0.2 ipython h5py sympy scikit-learn dill natsort setuptools scikit-image cython ipyparallel dask traits traitsui";
     else DEPS="pip ipython numpy scipy matplotlib>=2.0.2 h5py sympy scikit-image";
     fi
   - conda create -n testenv --yes python=$PYTHON;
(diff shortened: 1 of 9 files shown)


sem-geologist:
CI checks fail not due to bcf re-implementation, but due to plotting test failures.

ericpre:
This look alright. Can you please merge with `hyperspy:RELEASE_next_minor`, so that we can make sure that the bcf tests are also passing on appveyor? Since matplotlib 2.1 the plotting tests are failing... but this is another story.

ericpre:
Thanks @sem-geologist!

PR #1752, opened 2017-10-10
fix: bcf hitachi; enh: multi-hypermap index

sem-geologist:
The fix to #1751 and more future-proof.

PR #1755, opened 2017-10-11
Fix bcf hitachi: doc addendum

sem-geologist:
some additional docstrings and minimal addition to documentation concerning PR #1752.

The major change of PR #1752 is handling of dataset internal indexes in bcf.

Previously it was using xml tag which tells how many dataset there is in the bcf file, then for that range (0, 1...) the parser was looking for xml tags with those indexes.
However as some bugs with Hitachi bcf showed, that indexes not always starts from 0, but can be arbitrary numbers.

Current implementation, does check for indexes outside of XML, looking inside the single file syste (SFS) container file tree and extracting indexes from available binary data cube filenames, which directly correspond to spectra tags in XML.

Additionally it changed the default value of index= kwarg to None, which result in loading single, lowest index bearing dataset, and accepts additional value of 'all' to load all datasets.
It is future proof, as I have seen no such multi dataset files, except single file, with second dataset being empty. 

- [x]  ready for review

### DIFF
--- a/doc/user_guide/io.rst
+++ b/doc/user_guide/io.rst
@@ -575,14 +575,14 @@ currently supported by HyperSpy.
 
 Extra loading arguments
 ^^^^^^^^^^^^^^^^^^^^^^^
-select_type: One of ('spectrum', 'image'). If specified just selected type of
-data is returned. (default None)
+select_type: one of (None, 'spectrum', 'image'). If specified, only the corresponding
+type of data, either spectrum or image, is returned. By default (None), all data are loaded.
 
-index: One of (None, int, "all"). Index of dataset in bcf file, which can hold few datasets.
-Default None value result in first or the only dataset returned (the lowest index available).
-When set to 'all' it result in loading all available datasets.
+index: one of (None, int, "all"). Allow to select the index of the dataset in the bcf file,
+which can contains several datasets. Default None value result in loading the first dataset.
+When set to 'all', all available datasets will be loaded and returned as separate signals.
 
-downsample: the downsample ratio of hyperspectral array (hight and width only),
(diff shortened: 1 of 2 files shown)


doc/user_guide/io.rst
```diff
@@ -578,15 +578,17 @@ Extra loading arguments
 select_type: One of ('spectrum', 'image'). If specified just selected type of
 data is returned. (default None)
```
ericpre: Improvement suggestion: 
select_type: one of (None, 'spectrum', 'image'). If specified, only the corresponding type of data, either spectrum or image, is returned. By default (None), all data are loaded."

doc/user_guide/io.rst
```diff
@@ -578,15 +578,17 @@ Extra loading arguments
...
 
-index: index of dataset in bcf v2 files, which can hold few datasets (delaut 0)
+index: One of (None, int, "all"). Index of dataset in bcf file, which can hold few datasets.
+Default None value result in first or the only dataset returned (the lowest index available).
+When set to 'all' it result in loading all available datasets.
```
ericpre: Improvement suggestion:
index: one of (None, int, "all"). Allow to select the index of the dataset in the bcf file, which can contains several datasets. Default None value result in loading the first dataset. When set to 'all', all available datasets will be loaded and returned as separate signals.

doc/user_guide/io.rst
```diff
@@ -578,15 +578,17 @@ Extra loading arguments
...
 
 downsample: the downsample ratio of hyperspectral array (hight and width only),
```
ericpre: Typo: "hight" instead of "height".

ericpre (commented):
Just two improvement suggestions of the wording.

PR #1757, opened 2017-10-13
Fix filenames with square brackets not loading

### LINKED ISSUE #1325: hs.load() fails on a filename with closed square brackets
A signal with filename `"signal[200-300]eV.hdf5"` is successfully saved with hyperspy, but it fails to load.  I think the square brackets are not being escaped properly.

<ipython-input-40-e948282fe202> in <module>()
----> 1 b = hs.load(path + 'signal[200-300]eV.hdf5')

/Users/thomas/Dropbox/0_Git/GitHub/hyperspy_aar/hyperspy/io.py in load(filenames, signal_type, stack, stack_axis, new_axis_name, mmap, mmap_dir, **kwds)
    146                                if os.path.isfile(f)])
    147         if not filenames:
--> 148             raise ValueError('No file name matches this pattern')
    149     elif not isinstance(filenames, (list, tuple)):
    150         raise ValueError(

ValueError: No file name matches this pattern

thomasaarholt:
Fixes #1325. This way I won't rediscover the bug next year.

### DIFF
--- a/hyperspy/io.py
+++ b/hyperspy/io.py
@@ -173,7 +173,7 @@ def load(filenames=None,
             raise ValueError("No file provided to reader")
 
     if isinstance(filenames, str):
-        filenames = natsorted([f for f in glob.glob(filenames)
+        filenames = natsorted([f for f in glob.glob(glob.escape(filenames))
                                if os.path.isfile(f)])
         if not filenames:
             raise ValueError('No file name matches this pattern')


ericpre (changes requested):
It breaks the following:
```python
import hyperspy,api as hs
hs.load('*.hspy')
```

thomasaarholt:
Interesting, and good find. I’m a bit busy at the moment but I’ll see if I can find the time to try and find a solution.

thomasaarholt:
I didn't realise previously that the asterix in ericpre's case referred to a wildcard used with glob. Writing this comment here to remind myself next time I look at it!

tjof2:
@thomasaarholt perhaps try the following instead of `glob.escape` to keep it working with the star wildcard while escaping the brackets?

```python
import re

def escape_square_brackets(text):
    rep = dict((re.escape(k), v) for k, v in {"[": "[[]", "]": "[]]"}.items())
    pattern = re.compile("|".join(rep.keys()))
    return pattern.sub(lambda m: rep[re.escape(m.group(0))], text)
```
For example, say I have two files in `/home/data/` like this:
```
/home/data/afile[1x1].txt
/home/data/afile[1x2].txt
```
Then:
```python
>>> path = "/home/data/afile[*].txt"
>>> glob.glob(path)
[]
>>> glob.glob(escape_square_brackets(path))
['/home/data/afile[1x2].txt', '/home/data/afile[1x1].txt']
```
Works OK I think @ericpre?

tjof2:
That said, maybe you'd want to be able to turn the escaping on/off - what if you want to use `"["` and `"]"` for regex in the path too? @francisco-dlp

PR #1759, opened 2017-10-13
Fix progressbar showing when it shouldn't

### LINKED ISSUE #1758: m.as_signal progressbar always shows
On eels spectra, s.remove_background(show_progressbar=False) still shows the progressbar.

thomasaarholt:
Fixes #1758.


ericpre (dismissed):
This doesn't work for the default case (`show_progressbar=None`).

thomasaarholt:
That should do it. Thanks for noticing!

ericpre:
Have you tested it? I am asking this, because when I test it, I get an error.

thomasaarholt:
Sorry, I stupidly assumed it would be fine.

The error lies with how the `_as_signal_iter` method is written. `progressbar` takes `show_progressbar` as an argument for the `position` variable. That's what confused me. I'll have another look tomorrow.

PR #1760, opened 2017-10-14
Fix plotting tests for matplotlib v.2.1 and other bugs

ericpre:
### Description of the change
Matplotlib 2.1 introduces some changes, which breaks our plotting tests but at the same time helps/force us to find a few bugs and to fix it!
- we had a recursive loop when drawing plot, [this matplotlib change](https://matplotlib.org/api/api_changes.html#matplotlib-cbook-callbackregistry-process-suppresses-exceptions-by-default) now raise an error -now fixed by disconnecting and re-connecting the `draw_event` when drawing. I was expecting that it will fix the bug mentioned in [#1698](https://github.com/hyperspy/hyperspy/pull/1698#issuecomment-325693283) and #726, but this is not the case...
- change the matplotlib installation in appevyor: use freetype=2.6 from conde-forge instead of the matplotlib wheel from pypi.

### Progress of the PR
- [x] fix recursive loop with `draw_event`,
- [x] avoid create new axis, when not necessary -see [this API change in matplotlib 2.1](https://matplotlib.org/api/api_changes.html#axes-collision-in-figure-add-axes),
- [x] update two baseline images.
- [x] update dev guide regarding the required freetype version for plotting tests,
- [x] ready for review.


hyperspy/drawing/figure.py
```diff
@@ -51,12 +76,18 @@ def _draw_animated(self):
...
                      a.get_animated() is True]
                 except AttributeError:
-                     # The method was called before draw. This is a quick
-                     # fix. Properly fixing the issue involves avoiding
-                     # calling this method too early in the code.
+                    # The method was called before draw. This is a quick
+                    # fix. Properly fixing the issue involves avoiding
+                    # calling this method too early in the code.
+                    _logger.warning('Please report this warning to the '
+                                    'developpers: `_draw_animated` was called '
```
magnunor: Typo: developpers -> developers

magnunor:
Checking to see if this solves my "recursive loop" plotting issue.

magnunor:
This fixes the bug! I suggest we get this merged asap, and release a new patch version (1.3.1). Since I'm guessing anyone currently installing HyperSpy via either PyPI or Anaconda will experience the same bug?

ericpre:
It sounds to me that you have just committed yourself to review this one! ;)
This issue occurs only with matplotlib 2.1 and two workarounds are to downgrade matplotlib to 2.0 or to use another backend, so it is not really like HyperSpy is broken.

magnunor:
Seems like the failing travis test is due to lxml missing from the Anaconda repos, which should be fixed in RELEASE_next_minor (so I guess we can ignore that failing test).

I'm not sure if I feel qualified enough to comment on the changes in the plotting part, since I'm not that familiar with it... However, it seems to work.

The workarounds are fine, but the bug itself is fairly non-intuitive for non-experienced users. So the majority will probably just give up if they encounter it.

hyperspy/drawing/figure.py
```diff
@@ -51,12 +76,18 @@ def _draw_animated(self):
...
+                                    '`draw`.')
                     pass
             if canvas.supports_blit:
-                canvas.blit(self.figure.bbox)
+                if hasattr(canvas, 'update'):
+                    canvas.update()
```
vidartf: I cannot find any official documentation for this function. Could you link it? When will it be there, and what does it do differently than the blit?

to266: according to [this](https://github.com/matplotlib/matplotlib/pull/9505), is the `update` or `blit` even necessary? 

The relevant bit:
```
 +    In most backends callbacks subscribed to this callback will be
 +    fired after the rendering is complete but before the screen is
 +    updated.  Any extra artists drawn to the canvas's renderer will
 +    be reflected without an explicit call to ``blit``.
 +
 +    .. warning ::
 +
 +       Calling ``canvas.draw`` and ``canvas.blit`` in these callbacks may
 +       not be safe with all backends and may cause infinite recursion.
```

ericpre: This is the `update` method from QtWidget. For some reason, the `blit(self.figure.bbox)` is not working nicely anymore with the qt4 backend (I haven't tested qt5): it does not draw completely the plot but it does it after a draw event is triggered... the `update` method does the job as before though!

francisco-dlp: As this has worked properly until recently, I that that you may have trying to workaround an external bug here. What is triggering the bug, the version of qt4 or matplotlib? Maybe the issue has been fixed externally in the mean time? In any case, would it be possible to simply skip the problematic version in ``setup.py``? If not,  I would suggest to  change the ``if hasattr(canvas, 'update')`` for checking if the backend is qt4 in order not to trigger the ``update`` unnecessarily for other backends.

ericpre: Yes, this was more or less a workaround and not a very good one, quite a few changes have been made for the qt backends in the last few matplotlib releases -not sure this is completely finished! I have not followed what's going on but there are some relevant information in matplotlib/matplotlib#9505 and matplotlib/matplotlib#9199.

francisco-dlp:
I have just added some comments to the line were you flush the events. Other than that it LGTM so...

PR #1774, opened 2017-11-10
Update mva.py

mwalls:
### Requirements

### Description of the change
- Change A.
- Change B.
- etc.

### Progress of the PR
- [ ] add tests,
- [ ] ready for review.

```python
>>> import hyperspy.api as hs
>>> import numpy as np
>>> # Your new feature...
```

### DIFF
--- a/hyperspy/learn/mva.py
+++ b/hyperspy/learn/mva.py
@@ -789,8 +789,8 @@ class MVA():
     def _auto_reverse_bss_component(self, target):
         n_components = target.bss_factors.shape[1]
         for i in range(n_components):
-            minimum = np.nanmin(target.bss_loadings[:, i])
-            maximum = np.nanmax(target.bss_loadings[:, i])
+            minimum = np.nanmin(target.bss_factors[:, i])
+            maximum = np.nanmax(target.bss_factors[:, i])
             if minimum < 0 and -minimum > maximum:
                 self.reverse_bss_component(i)
                 _logger.info("IC %i reversed" % i)


mwalls (commented):
I'm trying to comment on my own change. I never know what I'm doing here. The suggested change is just to make the component reversal work properly in bss. It seems to work for me now

mwalls:
In passing, I also don't know the reason for what seems to me an overly elaborate criterion. Why not just use:
 if maximum< 0:
                 self.reverse_bss_component(i)

or even use the mean instead of the maximum, to make the components mostly positive?

thomasaarholt:
Just to clarify, you're referring to line 794 in mva.py.
This PR should probably target the RELEASE_next_patch branch, as it fixes a bug, right?

Nice find, btw. This has probably been there forever.

mwalls:
Thanks Thomas. Yes I think it should be done for the next patch if it's correct. But I don't know how to put labels on these things. Am I supposed to do it? Or is that for the administrators?

Yes in my second comment, about the criterion, I'm referring to 794. But I'm guessing that Francisco (I think he wrote it) must have had a reason...

thomasaarholt:
You can try changing the target branch in the top of this PR here on github. That will probably try to push all the changes from RELEASE_next_minor however, which we don't want. 

If the above does try to push a ton of changes, I would recommend closing this PR (having noted the changes you wanted to make), go to the front page of the hyperspy repo, change branch to the RELEASE_next_patch branch and then edit the code here in github (since there's only one file being changed). That'll push changes to the _patch branch.

mwalls:
ok I'll try the second option. The first one sounds scary.

mwalls:
I tried the second option (or thought I did) and it wanted to merge 51 things. I think that's what you feared in your first suggestions. So I've closed that PR and re-opened this one.

mwalls:
I went to 
hyperspy/hyperspy/edit/RELEASE_next_patch/hyperspy/learn/mva.py
and did the changes and proposed the change. Then at the top, the branch had changed back to RELEASE_next_minor, so I changed it again to RELEASE_next_patch and it said they cant be merged automatically, but that I could still make the PR, which I did, but then it said I wanted to merge 51 things...
Is it obvious what I'm doing wrong?

thomasaarholt:
That's interesting - can I urge you to try once more? I tried what you said and got the following screenshot, which looks correct. The base fork and the head fork are set correctly and in the right order - and were set automatically this way.

image

thomasaarholt:
I think what you did was somehow select edit before changing branch? Go to the front page, change branch to _patch, then navigate to hyperspy/learn/mva.py and press the edit button.

mwalls:
I should maybe point out that I'm not authorised to edit directly the code in Github. It always makes a new branch in my fork instead. Is that the problem?

mwalls:
ok I'll try it

thomasaarholt:
I'm not authorised either :) When you start editing the first time, it creates a fork of hyperspy on your user, and clones the branch you were trying to edit, so that you have access to the full code. Once you've edited the branch, you'll create a request to have hyperspy *pull* your code to it.

That's at least the way I've got git lingo in my head 😆

PR #1777, opened 2017-11-22
Patch 6

mwalls:
This is a test. Not sure I'm sending this to the right place.


mwalls:
This was a mistake. It looks like I'va acidentally done the first option you suggested, Thomas.

PR #1778, opened 2017-11-22
Fix bug in bss component reversal

mwalls:
The bss components are sometimes completely negative. I think it's because the code looks at the min and max in the loadings and not the factors. The idea is to fix this.
Incidentally, why is the criterion in line 794 not just a simpler

if maximum < 0 :
                 self.reverse_bss_component(i)

(or even the mean instead of the max, to make components mostly positive...)


mwalls:
Ok Thomas I think it worked this time. Maybe I did what you said somehow. Thanks for the info.

thomasaarholt:
Glad it worked :)

ericpre:
Would it be sensible to add an option to allow the user to choose between the loadings and the factors to determine if a component needs to be reversed or not?

Regarding the two conditions for reversing a component: you suggesting with the mean sounds tempting because it is simpler but usually the code is written for a reason, maybe there is a good one here... I can't think of any though!

mwalls:
For the option, why not?  (Not sure I know how to do it).  With regard to the criterion, yes I said the same thing in the first version of this PR (which I deleted because I sent it to the wrong branch or something). I haven't changed that line because I'm guessing Francisco (probably) did it like that for a reason.

ericpre:
something along the line of:
```python
def _auto_reverse_bss_component(self, target, use_loading=True):
    ...
    for i in range(n_components):
         if use_loading:
              values = target.bss_loadings
         else:
              values = target.bss_factors
         minimum = np.nanmin(values[:, i])
         maximum = np.nanmax(values[:, i])
    ...
```
then adding an option to `blind_source_separation` or setting appropriately the value (not idea if it would make sense to reuse the value of `on_loading`) to pass to `_auto_reverse_bss_component` line 676 ?

mwalls:
Sounds god but I will leave it to someone else to implement. (Or am I the only person who can change code in this PR?)

ericpre:
@mwalls, I made the changes as discussed, do you want to give a try to see if it does what you wanted? I am not very pleased with the name of the parameter `reverse_component_criterion`, does someone have a better name?

francisco-dlp:
This LGTM but it needs a test.

Issue #1773, opened 2017-11-10
BUG: BCF-dask memory leak


sem-geologist:
in issue #1720 (which is closed) @jat255 had reported additional observation that deletion of bcf obtained signal does not free memory. I have bumped into this issue after upgrading from 1.1 to 1.3 version which had broke my workflow of dealing with thousands of bcfs.

It is clear that xml part is not involved and something had changed in hypercube reading.
I shortly tested with dummy rpl files same workflow on the same machine and it is clear that only bcf are affected. It is possible that same mem leak is affecting  hyperspy-ui (which I had observed previously with bcfs).

I will work in finding the leak, but suggestions how to do that is welcome.

**locating the source of the leak:**

- [x] Does "pure_python" (calling method `BCF_reader.py_parse_hypermap`) implementation generates destroyable array? YES, it can be deleted instantaneously (no `gc.collect()` is needed – `del` takes care)
- [x] Is cython parsed array (function `unbcf_fast.parse_to_numpy`) destroyable? YES, arrays can be deleted with `del`.
- [x] Is the array destroyable when called through method `BCF_reader.parse_hypermap`? **NO**, both ( when using python or cython) implementations results in indestructible array when abstracted by this function.
- [x] What have changed in that method in between versions 1.1 to 1.3 ? Lazy/dask wrappers were added at 1.2.
- [x] What exactly is leaking? It have nothing to do with any hyperspy code.

sem-geologist:
I tested unbcf_fast.pyx parsing function, the array is destroyable.
I narrowed down the problem - it is lazy loading code!
"pure python" loading is not affected as it missed the lazy-loading code injection.
@to266 , any insight to this problem?

sem-geologist:
I think this can be connected:
https://github.com/dask/dask/issues/2625

btw, I tested above bug on python 3.5 on KDE neon (based on Ubuntu 16.04)

sem-geologist:
oh and btw, my test on rpl files missed the point as it do not use dask.

Could anybody test heavily other formats to see if dask problem persists there too.

ericpre:
Would it be possible to write a test, which checks if the memory have been released after deleting the object? And run this test for all io readers?

sem-geologist:
@ericpre , no need for that. I can replicate the leak without hyperspy. Writing test is the last thing needed in this case.
So what is the problem? it is `dask_delayed`.
[...]

PR #1779, opened 2017-11-22
Partial fix of bcf memory leak

sem-geologist:
### Description of the changes:
- unify cythonish and pure pythonic parsing function interfaces
- exclude dask from wrapping the numpy array with `.compute()` method if `lazy=False`

### Detailed description:
To begin with, #1773 `dask` array bug makes the library of little use at this moment, compared with pre-v1.2 hyperspy versions.
The bug could be partly neglected by excluding `dask` involvement when `lazy=False`.
Inspection of the code made me realize that dask/lazy loading loop was implemented in current way due to complicated bcf library codebase.
The bcf code included in hyperspy were (still is) planned to be used with some other custom software of mine,  where some functions in hyperspy are stripped and the code in some parts looks weird (e.g. class with only `__init__` method where more proper implementation would be  a dictionary.
The `if lazy` part is complicated too, as pure python and cython parsing functions have different interfaces. The pure python parsing method was written explicitly for hyperspy and differs from cython implementation where it is a function. Furthermore some classes are pointless in bcf for hyperspy, as io anyway accepts only dictionary. Blown-up instances of classes particularly are dangerous with dask delayed. So...

### Progress of the PR
- [x] get rid of HyperMap class, ~~same can be achieved with `dict`~~; better - don't do persistant parsing, get rid of that method, don't create abstract atributes/ dicts, define relevant stuff straight into final dict.
- [x] move the `.py_parse_hypermap` method out of class, refactor it into function
- [x] unify parsing function interfaces (python/cython), relocate shape, dtype generation code from those functions into single code. 
- [x] refactor the `if lazy` logic, which will get more simple with simple interface
- [x] check tests,
- [x] ready for review.


sem-geologist:
@francisco-dlp , maybe this could go to next_patch?

francisco-dlp:
No, it cannot because you have branched out for  `next_minor`. However, you could fix that and resubmit it to `next_patch`.

hyperspy/io_plugins/bcf.py
```diff
@@ -946,246 +908,217 @@ def parse_hypermap(self, index=None,
...
         if type(cutoff_at_kV) in (int, float):
-            eds = self.header.get_spectra_metadata(index)
-            cutoff_chan = eds.energy_to_channel(cutoff_at_kV)
+            eds = self.header.spectra_data[index]
+            max_chan = eds.energy_to_channel(cutoff_at_kV)
         else:
-            cutoff_chan = None
-
+            max_chan = self.header.estimate_map_channels(index=index)
+        shape = (-(-self.header.image.height // downsample),
```
francisco-dlp: Why are you using the negative sign in `(-(-`? If it is to avoid getting a `0` I would suggest the more explicit:

```python


shape = (self.header.image.height // downsample, self.header.image.width // downsample, max_chan)
shape = [ss if ss >= 1 else 1 for ss in shape]
```

hyperspy/io_plugins/bcf.py
```diff
@@ -946,246 +908,217 @@ def parse_hypermap(self, index=None,
...
+            #    >1 - instructively packed spectra;
+            # value which sometimes shows the size of packed data (uint16);
+            # number of pulses if pulse data are present (uint16) or
+            #      additional pulses to the instructively packed data;
+            # packed data size (32bit) (without additional pulses) \
+            #       next header is after that amount of bytes;
+            x_pix, chan1, chan2, dummy1, flag, dummy_size1, n_of_pulses,\
+                data_size2 = strct_unp('<IHHIHHHI',
+                                       buffer1[offset:offset + 22])
+            pix_idx = (x_pix // dwn_factor) + ((-(-width // dwn_factor)) *
```
francisco-dlp: As above

hyperspy/io_plugins/bcf.py
```diff
@@ -946,246 +908,217 @@ def parse_hypermap(self, index=None,
...
+            # use assigment, which is ~4 times faster, than inplace add
+            if max_chan < chan1:  # if pixel have more channels than we need
+                chan1 = max_chan
+            if (dwn_factor == 1):
+                vfa[max_chan * pix_idx:chan1 + max_chan * pix_idx] =\
+                    pixel[:chan1]
+            else:
+                vfa[max_chan * pix_idx:chan1 + max_chan * pix_idx] +=\
+                    pixel[:chan1]
+    vfa.resize((-(-height // dwn_factor),
```
francisco-dlp: As above

francisco-dlp (commented):
Other than the small style comment, it looks good to me.

hyperspy/io_plugins/bcf.py
sem-geologist: there is such math function: floor, what I need here is roof, that is the trick

sem-geologist: i.e. `5 // 2` would give 2, but I need 3, but I also need that `4 // 2` would give 2 instead of returning 2 in both cases.

sem-geologist: I could add comment that it is roof, maybe it would be clear then what `-(-` does.

sem-geologist:
aaa I dont know how it is doing that, but it is pulling some stuff from next_minor. I need to find how to drop those files...

sem-geologist:
also I see in tests that some numpy deprecation appeared (`fromstring` deprecated, it should be replaced with `frombuffer`)

hyperspy/io_plugins/bcf.py
francisco-dlp: Then, maybe what you need is ``np.ceil`` or ``math.ceil``

francisco-dlp:
Regarding ``fromstring``, @ericpre fixed it in #1799 & #1806

hyperspy/io_plugins/bcf.py
sem-geologist: I think math.ceil is more approprate as it returns integer. I wonder how did I missed that simple solution...

sem-geologist:
I can't understand how #1806 could pass, as replacing `fromstring` into `frombuffer` is not enough. thanks to new numpy's way in too-fast, non-discussed rushed changes. The main weakness of this deprecation is that frombuffer is not direct replacement of `fromstring` as first returns immutable array where e.g. byteswich() is then not working, while the other returns a new array. The direct replacement should be like this: `np.frombuffer(a, *args).copy()`. so i added `.copy()` where appropriate to comply with newer numpy and pass all tests.

@francisco-dlp , the PR was reuploaded removing traces of next-major. I replaced all `-(-a//b)` constructions into `ceil(a / b)` (using `math.ceil`). I would like this to be merged, so I could start to clean up way for #1801 .  I would like to have it separate from this PR.

PR #1781, opened 2017-11-28
Fix #1360: optimized copy warning

### LINKED ISSUE #1360: Optimized copy warning
Whenever I load any DM signal, I get the message:
`WARNING:hyperspy.signal:<EDSTEMSpectrum, title: EDS Spectrum Image 0eV, dimensions: (150, 279|2048)> data is replaced by its optimized copy`
This can be a bit scary for new users, as was evidenced during our workshop in Oxford last week. Is there a particular need to emphasise that the data has been replaced? As a "normal" user, I don`t even know what the implication is.

The documentation talks about `optimize` in conjunction with `hs.transpose`. Perhaps loading a DM file involves calling this function?

ericpre:
### Description of the change
Fix #1360 about the confusing warning "data is replaced by its optimized copy" when spectrum images are loaded from dm3/dm4 files.

### Progress of the PR
- [x] Add argument to the dm reader to set the `optimize` parameter of the `transpose` method when loading data,
- [x] tweak message and change the logger level from warning to info,
- [x] update docstring,
- [x] update user guide,
- [x] ready for review.

### Minimal example
```python
>>> import hyperspy.api as hs
>>> s = hs.load('some_spectrum_image.dm3', optimize=True)
>>> s = hs.load('some_spectrum_image.dm3', optimize=False)
```
In both cases, no message will be displayed, except if the logger level is set to "info".


hyperspy/docstrings/signal.py
```diff
@@ -28,3 +28,10 @@
 
        -------------------- {0} --------------------
         {1} """
+
+OPTIMIZE_ARG = \
+    """optimize : bool, default is False
+            If the data should be re-ordered in memory, most likely making a
```
francisco-dlp: What about something like this:

If `True`, the location of the data in memory is optimised for the fastest iteration over the navigation axes. This operation can cause a peak of memory usage and require considerable procession times for large datasets and/or low specification hardware.

hyperspy/_signals/common_signal1d.py
```diff
@@ -17,21 +17,23 @@
...
 
 class CommonSignal1D(object):
 
     """Common functions for 1-dimensional signals."""
 
-    def to_signal2D(self):
+    def to_signal2D(self, optimize=True):
         """Returns the one dimensional signal as a two dimensional signal.
 
         Always ensures the data is stored optimally, hence often making a copy
```
francisco-dlp: Always -> By default

hyperspy/_signals/lazy.py
```diff
@@ -258,7 +258,7 @@ def __array__(self, dtype=None):
     def _unfold(self, *args):
         raise lazyerror
 
-    def _make_sure_data_is_contiguous(self, log=None):
+    def _make_sure_data_is_contiguous(self):
```
francisco-dlp: Even though ``log`` is not used here, we must put it back as it is used in the non-lazy function and call to the method may use the arg now or in the future.

hyperspy/signal.py
```diff
@@ -4039,7 +4038,7 @@ def __len__(self):
         nitem = nitem if nitem > 0 else 1
         return nitem
 
-    def as_signal1D(self, spectral_axis, out=None):
+    def as_signal1D(self, spectral_axis, out=None, optimize=True):
         """Return the Signal as a spectrum.
 
         The chosen spectral axis is moved to the last index in the
```
francisco-dlp, line 4046: As before, mention that this is now only the default behaviour.

hyperspy/signal.py
```diff
@@ -4565,14 +4563,15 @@ def transpose(self, signal_axes=None,
                                     optimize=optimize)
                 res.metadata.set_item('Signal.Noise_properties.variance', var)
         if optimize:
-            res._make_sure_data_is_contiguous(log=True)
```
francisco-dlp: As before, I think that we shouldn't remove the ``log`` arg.

francisco-dlp (changes requested):
Just some little requests. Other than that LGTM

hyperspy/_signals/lazy.py
ericpre: I don't see where it is used in his non-lazy counterpart, the logger is used and not the log argument. Could it be a legacy from the time, when there was no logger? Or am I missing something?

francisco-dlp: Here https://github.com/hyperspy/hyperspy/blob/7932b6c44fea40991883a97a9a42a5ae301b8b1b/hyperspy/signal.py#L2690

ericpre: Yes, sorry I was looking in the wrong branch... reverted it back now!

PR #1789, opened 2017-12-16
Add new artificial_data module

magnunor:
This pull request adds a new module which contains functions for generating artificial data, for use in things like docstrings or for people to test HyperSpy functionalities.

Currently it contains a functions to generated an artificial Mn-L32 core loss edge signal + model, and an atomic resolution STEM image.

### Todo
- [x] Add core loss signal function
- [x] Add atomic resolution Signal2D function
- [x] Add unit tests
- [x] Update user guide
- [x] Ready for review.

### To test this:
```python
>>> import hyperspy.api as hs
>>> s = hs.datasets.artificial_data.get_core_loss_eels_signal()
>>> s.plot()
```


francisco-dlp:
Great initiative!

magnunor:
I added functions to generate a single + line scan EELS low loss spectrum, and line scan core loss spectrum.

The plan is to have easily accessible artificial data for the various docstrings, so i) people can easily test them, ii) we can run them as unit tests.

hyperspy/datasets/artificial_data.py
```diff
@@ -0,0 +1,239 @@
+import numpy as np
+from hyperspy import components1d, components2d
+from hyperspy.signals import EELSSpectrum, Signal2D
+
+
+def get_low_loss_eel_signal():
```
ericpre: "eels" instead of "eel", to be consistent with the convention in hyperspy (EELSSpectrum, etc.)?

ericpre (commented):
Looks nice. Is it worth updating `doc/api/hyperspy.datasets.rst` and mentioning it in the dev guide also?

hyperspy/datasets/artificial_data.py
magnunor: Done! (Electron energy loss spectroscopy spectrum?)

magnunor:
I added `datasets.artificial_data` to the API doc, and to the dev guide.

hyperspy/datasets/artificial_data.py
ericpre: Yes, I know and it is redundant and not correct... I tried myself to use EEL spectrum and I gave up because it does look a bit weird since almost everyone used EELS spectrum. The important thing is to be consistent!

PR #1793, opened 2018-01-05
ENH: add colormap iterating to plot_images

jat255:
This is something that I've wanted on my own plots for a while, so rather than just coding it up as a one-off, I figured it would be useful to add back to HyperSpy. It meets my requirements, but I'd be happy to get some suggestions of how it could work better.

### Description of the change
Often when plotting images, it can be desired to plot images with different colormaps for each signal (such as when correlating a set of spectra plotted with `plot_spectra`). This PR does that with the following changes:
- Primary change is that the `cmap` option of `plot_images` can now be a list, much like the `color` option of `plot_spectra`
- To go along with this, the `centre_colormap` option has also been changed to be iterable
- An additional option of `mpl_colors` was added to `cmap`, which will loop through the default `matplotlib` colors (useful when comparing line intensities or PCA components/loadings)
- Additionally added a helper utility function called `make_cmap`, which generates a `matplotlib` colormap from a list of colors (this function is used when using the `mpl_colors` option described above

### Progress of the PR
- [x] Change implemented 
- [x] update docstring 
- [x] add tests - **There are no plotting tests currently for `plot_images`, but we should add some**
- [x] ready for review

### Script I was using locally for testing
```python
import matplotlib as mpl
mpl.use('Qt4Agg')

import hyperspy.api as hs
from hyperspy.drawing.utils import make_cmap
import scipy

plt.ioff()

image = hs.signals.Signal2D([scipy.misc.ascent()]*12)
image2 = hs.signals.Signal2D([scipy.misc.ascent()]*12)
angles = hs.signals.BaseSignal(range(0,120,10))
angles2 = hs.signals.BaseSignal(range(120,0,-10))
image.map(scipy.ndimage.rotate, angle=angles.T, 
	      show_progressbar=False, reshape=False)
image2.map(scipy.ndimage.rotate, angle=angles2.T, 
	       show_progressbar=False, reshape=False)

hs.plot.plot_images(image, per_row=4, 
					suptitle='MPL Colors',
                    cmap='mpl_colors', axes_decor='off',
                    colorbar='multi', tight_layout=True)

hs.plot.plot_images(image, per_row=4, 
					suptitle='Default',
                    cmap=None, axes_decor='off',
                    colorbar='multi', tight_layout=True)

hs.plot.plot_images(image, per_row=4,
					suptitle='One string', 
                    cmap='viridis', axes_decor='off',
                    colorbar='multi', tight_layout=True)

hs.plot.plot_images(image, per_row=4,
					suptitle='Short list', 
                    cmap=['viridis', 'PuBu_r', 'coolwarm',
                    	  'inferno', 'nipy_spectral'], 
                    axes_decor='off',
                    colorbar='multi', tight_layout=True)

hs.plot.plot_images([image, image2], per_row=6,
					suptitle='Two signals', 
                    cmap='mpl_colors', 
                    axes_decor='off',
                    colorbar='multi', tight_layout=True)

hs.plot.plot_images(image, per_row=4,
					suptitle='make_cmap example', 
                    cmap=make_cmap(['#ffbb3e', (0.5,0,0)], 'Custom_cmap'), 
                    axes_decor='off',
                    colorbar='multi', tight_layout=True)

plt.show()
```


jat255:
How it looks when using the script above:

1-mplcolors
2-default
3-onestring
4-shortlist
5-twosignals
6-makecmap

ericpre:
For info:

- there are plot testing for `plot_images` in [test_plot_signal2d.py](https://github.com/hyperspy/hyperspy/blob/RELEASE_next_minor/hyperspy/tests/drawing/test_plot_signal2d.py#L118)
- the plot testing in CI are failing because at some point the freetype version changed.

jat255:
> there are plot testing for plot_images in test_plot_signal2d.py

@ericpre Thanks, I moved my tests into that file

This should be ready for review. All the tests are passing on my machine.

francisco-dlp:
This is awesome!

@ericpre, have you reviewed this or would you like me to do it?

ericpre:
I can review it a bit later, in case you want to focus on other PRs waiting
for review! ;)

hyperspy/drawing/utils.py
```diff
@@ -420,8 +420,16 @@ def plot_images(images,
...
-        cmap : matplotlib colormap, optional
-            The colormap used for the images, by default read from pyplot
+        cmap : matplotlib colormap, list, or ``'mpl_colors'``, *optional*
+            The colormap used for the images, by default read from ``pyplot``.
+            A list of colormaps can also be provided, and the images will
+            cycle through them. Optionally, the value ``'mpl_colors'`` will
+            cause the cmap to loop through the default ``matplotlib``
+            colors (to match with the default output of the
+            :py:func:`~.drawing.utils.plot_spectra` method.
+            Note: if using more than one colormap, using the ``'single'``
```
ericpre, line 430: I think that we should not allow to use different colormaps and a single colorbar together. This will quite confusing and I can't think of any situation where it would be useful.

hyperspy/drawing/utils.py
```diff
@@ -580,6 +577,46 @@ def plot_images(images,
...
+    if cmap is None:
+        cmap = [plt.get_cmap().name]
+    elif cmap == 'mpl_colors':
+        for n_color, c in enumerate(mpl.rcParams['axes.prop_cycle']):
+            make_cmap(colors=['#000000', c['color']],
+                      name='mpl{}'.format(n_color))
+        cmap = ['mpl{}'.format(i) for i in
+                    range(len(mpl.rcParams['axes.prop_cycle']))]
+    # cmap is list, tuple, or something else iterable (but not string):
+    elif hasattr(cmap, '__iter__') and not isinstance(cmap, str):
```
ericpre: You could pass many thing in there and get an error later in the code. I would rather consider all the expected input and finished in the `else` by raising a ValueError when the parameter is not correct to provide a clear error message.

doc/user_guide/visualisation.rst
```diff
@@ -548,8 +548,92 @@ which is used to call subplots_adjust method of matplotlib
...
+  :width:   500
+
+  Using :py:func:`~.drawing.utils.plot_images` with ``cmap='mpl_colors'``
+  together with :py:func:`~.drawing.utils.plot_spectra` to visualize the
+  output of a non-negative matrix factorization of the EDS data.
+
+
+.. NOTE::
+
+    If a list or other iterable type is given for the ``cmap`` argument
```
ericpre: See comments below.

ericpre (changes requested):
Looks good, just a few comments.
I suspect you also need to merge with upstream to update the other baselines images since we have to change to freetype 2.8...

jat255:
Ok, sorry for the delay, but this has been updated as requested. I rebased onto RNMinor, updated to freetype 2.8 locally, and regenerated the associated plots, so I think this should be ready to go.

jat255:
Actually, hold off on this. It looks like I might have made a license violation by using the ``make_cmap`` code I found online. It's licensed under a [CC BY-NC-ND 3.0](https://creativecommons.org/licenses/by-nc-nd/3.0/deed.en_US) which is not open-source, and does not allow derivative code, so I don't think it would be compatible with HyperSpy. I've reached out to the author of that code for an exemption, and I'll update here when he gets back to me.

jat255:
The license thing is no longer an issue. The author got back to me, and he has [published the code](https://github.com/CSlocumWX/custom_colormap/blob/master/custom_colormaps.py) under a BSD-style license, so we should not have a problem using it.

jat255:
Alright, should be good to go and ready for review once more. Per @francisco-dlp's suggestion, I updated the code to raise an error when incompatible arguments are given, and the tests have been updated to check for this.

jat255:
@ericpre any chance you could look this over if you have some time?

hyperspy/drawing/utils.py
```diff
@@ -580,6 +582,52 @@ def plot_images(images,
...
+        cmap = ['mpl{}'.format(i) for i in
+                range(len(mpl.rcParams['axes.prop_cycle']))]
+        colorbar = __check_single_colorbar(colorbar)
+    # cmap is list, tuple, or something else iterable (but not string):
+    elif hasattr(cmap, '__iter__') and not isinstance(cmap, str):
+        try:
+            cmap = [c.name for c in cmap]  # convert colormap to string
+        except AttributeError:
+            cmap = [c for c in cmap]   # c should be string if not colormap
+        colorbar = __check_single_colorbar(colorbar)
```
ericpre: same as above.

hyperspy/drawing/utils.py
```diff
@@ -580,6 +582,52 @@ def plot_images(images,
...
+    # If no cmap given, get default colormap from pyplot:
+    if cmap is None:
+        cmap = [plt.get_cmap().name]
+    elif cmap == 'mpl_colors':
+        for n_color, c in enumerate(mpl.rcParams['axes.prop_cycle']):
+            make_cmap(colors=['#000000', c['color']],
+                      name='mpl{}'.format(n_color))
+        cmap = ['mpl{}'.format(i) for i in
+                range(len(mpl.rcParams['axes.prop_cycle']))]
+        colorbar = __check_single_colorbar(colorbar)
```
ericpre: `__check_single_colorbar(colorbar)` return `None` in any case and it then removes the colorbar from the plot, by setting `colorbar=None`!!! I would just do the check and not the assign the result of the function to `colorbar`.

ericpre (commented):
This looks good, just one small bug to fix!

ericpre (approved):

ericpre:
I have fixed the issue with colorbar and updated the baseline images accordingly.

jat255:
@ericpre thanks for fixing that bug. I was out of town on travel for a few days and so couldn't get to it. Glad to see this merged in!

PR #1807, opened 2018-01-19
Fancy metadata

thomasaarholt:
### Description of the change
I've enhanced the metadata with a html view to make it look a bit nicer. It works nicely in Firefox and Chrome. Microsoft browers display a less fancy version, although it is fine for use.

### Progress of the PR
- [x] ready for review.

```python
%matplotlib inline
import hyperspy.api as hs
s.metadata
```
image


francisco-dlp:
Could you add a test? Other than that, LGTM!

dnjohnstone:
@thomasaarholt could this get a test please so it can be merged?

Issue #1796, opened 2018-01-10
Units unchanged in metadata when using "hs.material.atomic_to_weight() on EDS quantification


CorentinLG:
The output of the s.quantification() function is a list and each has a title in its metadata of the type : "atomic percent of Al_Ka" for instance.

The function hs.material.atomic_to_weight(quantification) converts units, but the "title" of the resulting list remains unchanged and indicates atomic instead of weight%.

It can be confusing.

k8macarthur:
Should be fixed in the pull request #1820

ericpre:
Fixed in #1827.

PR #1820, opened 2018-02-05
Fix to atomic-to-weight function

k8macarthur:
Fix to issue #1796


thomasaarholt:
Should the file "ysk1a5qt.4xt.txt" be merged with this?

k8macarthur:
Well spotted! Oops...

ericpre:
I think this should go to next_patch. Also having all these merges with upstream make the history log is bit messy. Could you please checkout from next_patch and open another PR?

k8macarthur:
I'm happy to try that but it seems like the release_next_patch is really behind the release_next_minor

PR #1825, opened 2018-02-07
Parse more metadada from FEI emd

ericpre:
### Progress of the PR
- [x] parse elevation angle of the EDS detector.
- [x] parse live and real time for each EDS detector (only for compatible hardware with individual EDS detector readout),
- [x] parse gain and offset when available,
- [x] add tests,
- [x] ready for review.


ericpre:
I deleted the branch unintentionally...

Issue #1780, opened 2017-11-27
Improving signal management for EDS quantification


CorentinLG:
The four issues below are related to the post-modelling treatment of EDS data, for quantification purposes. 

1) after a model has been fitted, we use "result = m.get_lines_intensity()" to extract the peak heights. However, this function returns a list (ex : result[0] is the signal for the first x-ray line). For further quantification operation, it may be better to have a dictionnary instead. One could manipulate "result" as one single objet and call it for each specific xray_line.

2) As for "m.get_lines_intensity()", the "s.quantification()" function returns a list whereas it would be probably more convenient if it returned a dictionnary node (within the model ?).

3) "m.get_line_intensity()" returns the intensity each line separately (say Fe_Ka). However, Fe_Kb is also measured and fitted, and is always proportional to Fe_Ka. I think that there should be an option for "m.get_lines_intensity()" to return (Fe_Ka + Fe_Kb), since it is the value that is used for quantification afterwards.

4) To use k- or zeta-factors for quantification, "s.quantification() takes "result" and "kfactors" as arguments. The k-factors list must have the same length as the "result" list, which length is the number of considered x-ray lines. In practice, one has to order the kfactor list in the same fashion as the result list. This is annoying and leads to systematic mistakes. K-factors being specific to each microscope, it would be convenient if one could enter them within the (hyperspectral signal?) metadata, associated to their respective xray_lines. Then, s.quantification() would take only "result" as argument and respecting the ordering of each list would no longer be an issue.

francisco-dlp:
I think that these are excellent usability enhancements that requires an API change. I've labelled it "next major".

vidartf:
> Fe_Kb is also measured and fitted, and is always proportional to Fe_Ka

Note: This is not true if `m.free_sub_xray_lines_weight()` has been called.

CorentinLG:
You are right. We need caution here. I think by default Ka and Kb should be summed directly, but we should keep our options open for when we want to work on each of them separately.

k8macarthur:
I like the ideas in general. Personally I use my own dictionary to store all my EDX quantification library and quickly write that into a list. I'm not sure I quite understand how being able to store the cross-sections/z-factors etc into the metadata would speed things up. Otherwise being able to sum lines could be beneficial.

CorentinLG:
[...]

PR #1827, opened 2018-02-08
Fix metadata in newly creates maps.

k8macarthur:


k8macarthur:
@ericpre  Try this.

ericpre:
I would suggest to merge this as a temporary fix for #1796 but I think that the units should be stored in `Signal.quantity` and not in the title. #1730 should help significantly with this. Also, this kind of improvement could be done when the EDS quantification workflow gets improved (see discussion #1780).

k8macarthur:
Agreed, there are definitely some long-term plans for EDX which would probably change this function. I just thought I’d quickly fix the present error in the mean time.

Best wishes,

Katherine MacArthur

PR #1830, opened 2018-02-10
Fix deprecation warning invalid escape sequence

ericpre:
### Description of the change
Fix deprecation warning due to invalid escape sequence. I used a [tool from the numpy](https://github.com/numpy/numpy/blob/master/tools/find_deprecated_escaped_characters.py) repository to find them all.

### Progress of the PR
- [x] set docstring to raw string,
- [x] improve docstring (latex formatting),
- [x] set string use in regular expression to raw string,
- [x] update windows path with 2 backslashes,
- [x] ready for review.

### DIFF
--- a/hyperspy/_components/lorentzian.py
+++ b/hyperspy/_components/lorentzian.py
@@ -23,22 +23,22 @@ from hyperspy.component import Component
 
 class Lorentzian(Component):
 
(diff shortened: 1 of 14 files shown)


ericpre:
I label this PR as a bug because it makes pytest fails the winpython build.

PR #1831, opened 2018-02-12
Fix an issue with reading TEM images from a velox emd file.

ericpre:
This PR fixes an issue with parsing datetime of TEM images in velox emd file, since the `AcquisitionStartDatetime` used to parse the date and time is not present in the metadata of TEM images acquired with a CETA camera.


hyperspy/io_plugins/emd.py
```diff
@@ -828,13 +828,18 @@ def _get_metadata_dict(self, om):
...
-        meta_gen['time'] = time
+        try:
+            # We have only one entry in the original_metadata, so we can't use the
+            # mapping of the original_metadata to set the date and time in the
+            # metadata: need to set manually here
+            unix_time = om['Acquisition']['AcquisitionStartDatetime']['DateTime']
+            date, time = self._convert_datetime(unix_time).split('T')
+            meta_gen['date'] = date
+            meta_gen['time'] = time
+        except KeyError:
```
ericpre: @robindegen, if the `AcquisitionStartDatetime` key is not present in the metadata of TEM images, do you know how we can parse the date and time?

robindegen: @ericpre As far as I can tell, AcquisitionStartDatetime should have been filled in for TEM images. It might however be dependent on the TEM server version that Velox is running on. I'd have to discuss internally why these values could be missing. Do you have any EMD files that you could share that have this issue?

What you can do however is  use ```CustomProperties``` which has a bunch of key-value pairs. One of them is ```Detectors[BM-Ceta].TimeStamp``` which should contain the timestamp for that particular frame. The only downside is is that this is not the time the acquisition was started, but rather then that particular frame was stored.

ps. Do you happen to know on which TEM server version (and possibly microscope model) these values are missing?

ericpre: Thanks for the quick reply. I have shared two files on [dropbox](https://www.dropbox.com/sh/50mopwcbtk1ea0k/AADZsKCD3zaQVhBB-zTUHgqUa), corresponding to different versions of velox and TEM server. None of them has got the `AcquisitionStartDatetime` metadata but the most recent one has `Detectors[BM-Ceta].TimeStamp`. Sorry, I don't know the TEM server version on top of my head ;), but I can check later.

francisco-dlp:
Coveralls is not happy.  Would it be possible to add a file that presents the issue to the tests? Also (although not necessarily for this PR), why not taking this approach for all non-essential metadata in the file (i.e. continuing reading if the key is not found).

robindegen:
@francisco-dlp @ericpre That's probably a good idea. Most metadata we write into the file is considered "optional" internally and it can vary between server type (Titan, Talos etc.), version, and detector type. The keys should always be the same, but they may be missing completely or just have 0 written to them.

ericpre:
Yes, indeed this is supposed to be the case for most of the metadata but clearly not for this specific one!! I am going to check this.
@robindegen, would it be possible to get a file with the minimum set of metadata?

robindegen:
@ericpre I can try and see if that's possible. The thing is it varies per microscope configuration, so it's hard to say what the "base set" is. I'll let you know if I can get something.

robindegen:
@ericpre I've created an internal issue in our bug tracker to investigate this missing field.

ericpre:
I have added two test files (replace the actually data with much smaller data to keep the file small) and make all possible import optional.
Now, since I have added a couple of `try: except:` blocks and I can't test these since I don't have such file, coreralls complains even more than before!!! :/
I would suggest to process this PR and if @robindegen can provide file with minimum set of metadata then let's do it in another PR.

francisco-dlp:
Could you merge Rnm in?

hyperspy/io_plugins/emd.py
```diff
@@ -745,28 +746,46 @@ def _read_image(self, image_group, image_sub_group_key):
...
+        try:
+            image_display_group = self.p_grp.get('Displays/ImageDisplay')
+            key_list = _get_keys_from_group(image_display_group)
+            self.map_label_dict = {}
+            for key in key_list:
+                v = json.loads(
+                    image_display_group[key].value[0].decode('utf-8'))
+                data_key = v['dataPath'].split('/')[-1]  # key in data group
+                self.map_label_dict[data_key] = v['display']['label']
+        except:
```
francisco-dlp: I would specify the exception here.

hyperspy/io_plugins/emd.py
```diff
@@ -745,28 +746,46 @@ def _read_image(self, image_group, image_sub_group_key):
...
+                if hasattr(subgroup, 'keys'):
+                    sub_dict = {}
+                    for subgroup_key in _get_keys_from_group(subgroup):
+                        v = json.loads(
+                            subgroup[subgroup_key].value[0].decode('utf-8'))
+                        sub_dict[subgroup_key] = v
+                else:
+                    sub_dict = json.loads(subgroup.value[0].decode('utf-8'))
+                d[group_key] = sub_dict
+        except:
```
francisco-dlp: and here

francisco-dlp (commented):
Except for a couple of minor comments, it LGTM.

ericpre:
All done.

PR #1832, opened 2018-02-13
Document HyperSpy extensions

francisco-dlp:
Add a new "HyperSpy extensions" section to the Introduction chapter of the User Guide.

### DIFF
--- a/doc/user_guide/intro.rst
+++ b/doc/user_guide/intro.rst
@@ -29,8 +29,8 @@ There are multiple HyperSpy extensions. For a list of extensions hosted
 publicacly in GitHub search for the GitHub topic `hyperspy-extension <https://github.com/topics/hyperspy-extension>`_
 
 .. note::
-    From version 2.0, HyperSpy will be splitted into a core package (HyperSpy)
-    that will provide the common infraestructure and a number of HyperSpy
+    From version 2.0, HyperSpy will be split into a core package (HyperSpy)
+    that will provide the common infrastructure and a number of HyperSpy
     extensions.
      
 Our vision
--- a/doc/user_guide/intro.rst
+++ b/doc/user_guide/intro.rst
@@ -17,6 +17,22 @@ of the dataset.
 Its modular structure makes it easy to add features to analyze different kinds
 of signals.
 
+HyperSpy extensions
+-------------------
+
+In this document we refer to external programs that build on HyperSpy as
+"HyperSpy extensions". Those programs can e.g. provide extra data analysis tools
+file formats, add graphical user interfaces, provide
+extra blind source separation algorithms etc.
+
+There are multiple HyperSpy extensions. For a list of extensions hosted
+publicacly in GitHub search for the GitHub topic `hyperspy-extension <https://github.com/topics/hyperspy-extension>`_
+
+.. note::
+    From version 2.0, HyperSpy will be splitted into a core package (HyperSpy)
+    that will provide the common infraestructure and a number of HyperSpy
+    extensions.
+     
 Our vision
 ----------
 
@@ -50,7 +66,7 @@ that sets its character:
   and the
   `traitsui GUI <https://github.com/hyperspy/hyperspy_gui_traitsui>`_. If you
   need a full, standalone GUI, `HyperSpyUI <http://hyperspy.org/hyperspyUI/>`_
-  is for you. 
+  is for you.
 * We see HyperSpy as a collaborative project, and therefore we care
   about making it easy for others to contribute to it. In other words,
   we want to minimise the “user becomes developer” threshold. To achieve this


francisco-dlp:
Of course everything is up for discussion. For example, what do people think about the term "hyperspy-extension"? An alternative that we've discussed before is "hyperspykit", but personally I find "hyperspy-extension" more intuitive and more fit for the purpose: making it easy to find projects that build on hyperspy and promoting the creation of new projects instead of contributing to the main repository when the functionality is specific to a particular kind of signal. Can anybody think or a better one?

Extensions developers, if you would like your project to appear in the list, simply add the hyperspy-extension label to your project's repository once we agree on the label. For demo purposes I have added the ``hyperspy-extension`` label to the GUIs, see https://github.com/topics/hyperspy-extension .

magnunor:
I feel hyperspy-extension is better, simply due to be being a lot more descriptive.

What kind of projects would be "extensions"? Are Atomap and fpd_data_processing extensions?

ericpre:
"hyperspy-extension" is the best terminology I have heard so far and somehow sounds fairly natural, so I think it is very good!

Although using Github topic is a nice way to generate a list of other project, it only works with projects hosted on github. Maybe it would be worth also having a list of hyperspy-extension on the hyperspy website?

Add what you said in this PR to the doc and the github topic, since it summarises nicely the idea?
>  making it easy to find projects that build on hyperspy and promoting the creation of new projects instead of contributing to the main repository when the functionality is specific to a particular kind of signal

@magnunor, I guess this is up to the project itself to consider whether it can be regarded as hyperspy-extension or not. One good thing of the hyperspy-extension would be to clarify for the user what the different libraries do. In that sense, the atomap and fpd_data_processing libraries would fit very well! :)

francisco-dlp:
@ericpre, I fully agree with you in that Github topics suffer from the major drawback of not considering packages hosted elsewhere. I only see one way around the issue: keeping a list of extensions in the website. Anybody wanting to add his project to the list could simply send a PR to the website webpage (we'll have to document this in the dev guide). However, I would suggest to also use the GH topic for its convenience. We could add a link to the GH topic to the website list and encourage extension authors to also list their packages directly in the link for better visibility. Does it sound good?

francisco-dlp:
@magnunor, in my mind a hyperspy-extension is any external package that adds extra features to hyperspy. In this sense, a good fraction of what today we call HyperSpy will become multiple hyperspy-extensions with v2.0. To be more specific, an external package is not a hyperspy extension unless it adds features to HyperSpy. For example, a package that imports hyperspy internally to process some data shouldn't be called a hyperspy extension. The current gui-* packages are hyperspy extensions because they add features (GUI elements) to HyperSpy.

I think that in order to promote the creation of extensions, we should make it easy for external programs to provide the following objects:

1. Signals
2. Components
3. GUI elements
4. File readers and writers
5. BSS and decomposition algorithms
6. Plotting backends

—Anybody would like to add or remove something from this list?

I think that providing the infrastructure for 1-4 should be relatively straightforward. 4 would require defining an API for file readers and writers, which is something that we should have done a long ago. Before working towards 5 I think that we should rewrite the whole decomposition and BSS code. As for 6, that's probably the most complex task and I would rather suggest to leave it until somebody seriously considers writing a hyperspy extension to use something other than mpl for plotting.

Once the infrastructure is in-place, installing a hyperspy-extension that provides e.g. a new signal should result in the signal being available in ``hyperspy.signals``.

I am glad that you both like the name "extension", or rather the adjective, as the packages themselves can be called whatever. I think that it's good not to suggest any special naming conventions for the extensions (unlike other examples of the same sort of ecosystems such as scipy's scikits). Instead, just  mentioning in the documentation of the extension package that it can be considered a hyperspy extension should help the users find the documentation for the parts of the objects that are common to all hyperspy objects (and relief the developer from the task of documenting what's already documented in hyperspy).

Does it sound good?

francisco-dlp:
I have just clarified a little bit the last paragraph of the previous post that I wrote in a rush before entering a meeting (I am now entering another one...)

dnjohnstone:
Hi, our project pyXem: https://github.com/pyxem/pyxem/ is probably a reasonable example of an "extension" package that's sort of semi-developed.

In this case we define specialised signals for storing things associated with electron diffraction data and components for fitting electron diffraction data. We also add a number of "generator" classes that do things like convert signals from one type to another, perform simulations, and compare simulations with signals.

The main issue at present is that hyperspy contains a load of things that read the components and signals from hs.signals and hs.components and this does not include those that are defined in an extension package. The result is a load of strange behaviors like the signal type being changed by certain functions that shouldn't because hyperspy assigns it back to one of the signals it knows about. 

At present we bodge this by reassigning back to pyxem classes when we need to - but it's ugly.

What is needed is some way of reading in signals and  components from extension packages automatically - not really sure how that should be done.

dnjohnstone:
Regarding naming hyperspy-extension --> It sounds a little like you expect people to actually name their packages e.g. hyperspy-diffraction. I don't think that's going to happen, or would be appropriate. In the case of pyXem for example the package depends pretty much as heavily on pymatgen as it does hyperspy, so what is an extension of what?

dnjohnstone:
Another question might be to ask whether we really need to have static lists of 'allowed' signals and components in hyperspy at all. It feels a bit as though this is going to always end up creating a maintenance nightmare as more 'extension' packages are added.

I think there is an architectural attitude to hyperspy that comes from an idea that hyperspy is an 'end-user package' - whilst this is partly true in its current implementation it's likely that many extensions will do things like process data with hyperspy tools and then get physical insight by some kind of comparison to a physical simulation, which may well depend on some other substantial packages.

The question is - do you want to build a package where everyone has to work in some overarching hyperspy dogma, or do you want to build a package which provides this fantastic signal class and function mapping and model fitting, but that simply makes it easy to subclass Signal and Component as a platform for other work?

francisco-dlp:
No, if you read carefully the previous messages you'll see that I am not proposing to name the packages in any particular way, nor to even recommended it. That said, I think that it is good practice to refer (not name) to an external package as a "HyperSpy extension" when the package provides extra specialised HyperSpy objects such as signals, components etc and, more generally, follows HyperSpy's conventions. I think that most developers and certainly most users would appreciate the advantage of sticking to HyperSpy's conventions and referring to a given package as a "HyperSpy extension" when the goal is to add extra functionalities not available in HyperSpy itself. But then again, that's absolutely optional and obviously each developer team knows best how to refer to their package and address the needs of their users.

francisco-dlp:
I am not getting the idea behind your last post. What do you mean by "static list of allowed signals"?

dnjohnstone:
All I'm really saying is that it sounds like the plan at the moment might be to have packages that add to hyperspy rather than packages that build on hyperspy. Maybe that is how you see it, maybe not, but perhaps the structure is different depending on how this is seen. Or maybe there are choices that support both kinds of development.

E.g. It sounds like the current plan is to go to a place where having installed a hyperspy-extension that provides a new signal the user would use that signal with an interface like:

hs.signals.Signal1D()
hs.signals.EDSTEMSpectrum()
hs.signals.ElectronDiffraction()

Whilst that might make sense if the goal of the extension package is just to provide the data processing tools that you apply to the new signal this is only one way to write something that builds on hyperspy.

In our current implementation it's written more in a way that imagines the user not worrying too much about the fact that hyperspy is helping with things under the hood for processing data. That's because we liked an architecture where there's more balanced emphasis on simulation. Writing everything as a method of the new signal class just didn't work very well when we tried it. So right now, for example, have both:

pxm.ElectronDiffraction()

and,

pxm.IndexationGenerator()

In other words, I might prefer to write a package with a rather different API to hyperspy, so in this sense is not really an "extension", but a package that depends on hyperspy and sublasses BaseSignal or whatever in various ways, rather than than a package that adds to hyperspy. Perhaps then the emphasis is then on making subclassing easy in external packages, which is perhaps a slightly different question - or maybe not.

It's not that I want it to not be seen as an extension of hyperspy - it's just that it's not clear to me where things like our generator classes would fit in that framework.

francisco-dlp:
In your example ``IndexationGenerator`` seems to be a class that is initialised with an indexation library and a signal instance. I think that following HyperSpy's current conventions, this feature could have been implemented as a ``Signal method``  instead. For example, instead of:

```python
indexer = IndexationGenerator(dp,library)
match_results_A = indexer.correlate()

```

```python
match_results_A = dp.correlate(library)
```
because in HyperSpy the convention is to only use functions for operations that take several signals as inputs such a ``stack``. Have you tried this? If yes, why didn't it work well for you?

Sticking to current HyperSpy's conventions (or changing them if needed) has the advantage that users dealing with multiple "extensions" can easily get the structure of all of them because it is common. One of the goals of the extensions concept for HyperSpy itself is to encourage the development of more packages extending HyperSpy currently available objects without directly contributing to HyperSpy itself. If this happens, better common conventions should emerge from interacting with a broader community and stretching HyperSpy usage beyond its current limits.

Assuming that a given package extends hyperspy and, in addition, adds objects that don't fit (and cannot fit) in HyperSpy's structure, a possibility could be splitting it in two: a package that extends and another one that simply uses those features internally. Of course another alternative is to create a monolithic package instead and not even mention that the package extends HyperSpy in any way. However, I don't think that this good practice because that package, if installed, will inject objects into HyperSpy itself that will be available even when not importing the package.

To avoid confusion, packages that use HyperSpy internally (such as HyperSpyUI) are not concerned by this discussion/PR. This is just a discussion about how to add the feature to HyperSpy to register external objects as first class citizens (so that, e.g. a new external signal is properly loaded from file for example). This has nothing to do with encouraging/discouraging packages that build on top of HyperSpy (vs extending it), nor has anything to do with any "overarching hyperspy dogma" other than developing and promoting conventions for the benefit of the developers and users concerned.

francisco-dlp:
(I have just updated my obviously wrong indexer example)

dnjohnstone:
Perhaps the IndexationGenerator wasn't the best place to start - because you're probably right in that case seen in isolation - indeed initially, when we started, it was implemented the way you describe. So why did it end up how it is....

What happened was we first made a LibraryGenerator() which takes a crystal structure and makes a diffraction library (there are no Signals at this point). We then had a get_indexation() method to match that library to a signal - which is ok, but - later it felt more natural for the package in its own right to have anything that takes something simulated and matches it with a Signal, or does a simulation, to be in our Generator type of class and we restructured to have essentially 3 types of objects - Signals, Components and Generators. This let us move towards a common API for similar tasks from an applications perspective that are achieved in very different ways from a code perspective e.g. getting an orientation map (by pattern matching) and getting a strain map (essentially using hyperspy model fitting), both could be done using our generators. It also mimics the code structure in pymatgen which is also drawn on heavily for the crystal structure manipulation.

An advantage is that you can set up the generator with minimal computational cost and then run it later using some other computer.

The positive of this is that you could take what's in the ElectronDiffraction class and have it as a stand alone extension to hyperspy, the methods left there are much more like normal hyperspy methods, but you'd lose things that relate to simulations. Although that could be a separate package - so that's one solution as you describe.

I guess it's harder for me to see a place for a LibraryGenerator() accessed as an extension to hyperspy. Of course if the general concept of Generator objects was appealing more widely, then that could become a part of hyperspy-core and then you could add external generators too - these would be objects that set up the conditions to do a simulation or to match a simulation to a signal.

Perhaps it would be a good thing for hyperspy more generally because I guess it is a general principle that we should be doing simulations and matching them to data.

thomasaarholt:
Could I suggest that you make another issue about the pyxem issues with hyperspy, @dnjohnstone? I think it's excellent that you're pointing the weaknesses in extending hyperspy, and it's particularly well timed considering that we are splitting hyperspy into smaller bite-sized packages.

thomasaarholt:
The following are my two cents. The example is just a quick draft.

I think one can distinguish between extensions that can be used within hyperspy, like the guis and learning methods, and the applications that have been built *upon* hyperspy, like Atomap and pyxem. 

Personally, I don't think we need to be particularly formal about this. I think the primary purpose should be to advertise the possibility of using the apps that have been created, to the users. For the larger packages, we just refer to them to advertise themselves.

Keep it short and sweet, something ala:

### Hyperspy extensions
*Some sentence about how hyperspy is very flexible.*
Hyperspy can be extended further with the following extensions:
- pyqt GUI: *some statement*
- ipywidgets GUI: GUI built into the jupyter notebook
- learning: Several machine learning blah blah blah

There are also several apps that build upon hyperspy and take it into new domains:
- **atomap**: One-liner about atomap
- **pyxem**: One-liner about pyxem
- etc.

dnjohnstone:
Ok.... buckle up for a long one! I'm trying to express options rather than what I think is best, because frankly I don't know what I think is best, just agreed with @francisco-dlp to have the debate in public here.

I don't think there's much point in a separate discussion about pyxem - the package doesn't really matter it's just an example. It seems well timed in relation to splitting hyperspy because @francisco-dlp  and I have been talking about the whole thing in that context for a few years now. Initially (about a year ago) the code that was the start of pyxem was part of a PR to hyperspy - but anticipating the splitting of hyperspy it seemed more sensible to just do it outside hyperspy and discover issues the hard way that arise when building a package to extend hyperspy. So the point is to share what we have learnt about that to inform the general discussion - if the consensus is that what we've done with pyxem is wrong then we will probably just change it to fit the consensus.

The issues are general points of discussion about defining a Signal or a Component or an io_plugin that outside of hyperspy that in so doing "extend hyperspy".

At the moment the primary issue is that there are a number of functions in hyperspy that essentially involve checking against the components and signals it knows about or reassigning the data to one of the signals that it knows about. The solution proposed at present is to make hyperspy register extension packages so that the components and signals it knows about get longer. I think there may be an alternative, which is to look at the methods that do that and see if we can stop them referring to the components and signals in hyperspy - but this is potentially some quite involved rewriting of core hyperspy features (it may not be at all feasible). On the other hand it may make writing 'builds on hyperspy' packages easier, which some people will still want to do.

An example is that there is a DummyComponentContainer() class used in model fitting - at present a component outside of hyperspy doesn't get read into the dictionaries there (because hyperspy doesn't know about it) - one of the questions I'm trying to raise is whether we try to make the container know about the new component (what I see as the current proposal) or rather explore ways to avoid needing the container any more. Again the latter may make 'builds on hyperspy' easier.

Why is this so long? Well I should be clear that really I'm fully in agreement with @francisco-dlp that it would be lovely if it were possible for people to go to the hyperspy website download a distribution with a set of extensions and get one very nice coherent package with a consistent API that enables e.g. (taking EM as an obvious example) EDX, EELS, Diffraction, Holography analyses to be done. This is a bit like how LAMMPS works for atomistic simulations and I think this kind of "hyperspy superdistribution" is both what many electron microscopy groups would want (if it were there already, I would certainly be using it) and a sensible way to get good exposure of new developments.

For that to work though we need to figure out how to have a sufficiently general hyperspy-core to allow new analyses e.g. involving simulation (as in pyxem) to also be extended. We also need to be as clear as we can be about how to make that contribution so that it can ultimately become a part of the "hyperspy superdistribution". Perhaps what I'm saying with my pyxem comments is that I currently don't think the hyperspy-core is sufficiently general for the addition of more involved analyses requiring e.g. simulation.

Having thought about it a bit since my previous comment, I guess what I'm suggesting is that adding a "Generator" type of object as the extendable part of the hyperspy-core to work with simulations and some kinds of analyses may make this viable. Certainly this would mean that from a pyxem perspective I would just make our Generators subclass the hyperspy-core Generator.

francisco-dlp:
> I think there may be an alternative, which is to look at the methods that do that and see if we can stop them referring to the components and signals in hyperspy - but this is potentially some quite involved rewriting of core hyperspy features (it may not be at all feasible). On the other hand it may make writing 'builds on hyperspy' packages easier, which some people will still want to do.

Personally I can't think on an alternative to creating a mechanism to register external objects. If HyperSpy doesn't know about an object, how can it return it e.g. on loading? Note that this is not a HyperSpy peculiarity. Say that you subclass the numpy array, then save it using ``np.save`` and load it. You'll get a numpy array back, not your subclass.

When registering objects there are several alternative:

1. Automatically searching and registering extensions on loading
2. Manually registering the package after installation so that it always gets loaded when importing HyperSpy. A registered package could be disabled (without unregistering it) in preferences.
3. Requiring importing the package before importing HyperSpy in order to use their objects

In addition, any external object could contain a provider attribute with the name of the package that provides it. In this way HyperSpy could try to automatically import the object when e.g. loading a dataset that uses it and could print a warning explaining that proper loading of the object needs to install an external package.

Personally I don't like 1 much because it may slow down HyperSpy's startup time and it is only possible to disable an extension by uninstalling it. 3 is the most explicit (which is good) but also more verbose. 2 may be a good middle ground, although the need to register manually every extension may annoy some. My vote actually goes for providing 2 and 3 at the same time so that the users can choose.

@dnjohnstone, regarding features not fitting current HyperSpy's structure (e.g. simulations), I think that that's an excellent point. I think that we all consider HyperSpy and its ecosystem as a community effort where reinventing the wheel is among the biggest sins. Hence, it would be good to provide a place to discuss features that a given external package provides and are missing in HyperSpy such as your generator. Those features could grow inside the external package and, if other external packages ever require them, moving them to core HyperSpy could be considered. The key for this to work is that external packages developers don't forget to list and discuss those features in e.g. issues living in a new milestone created for the purpose in the hyperspy repository.

Thoughts?

dnjohnstone:
Regarding hs.load I agree - I think for people that decide to 'build on hyperspy' wrapping or re-writing the load function is what will have to happen, and I agree that this is normal in python. Of course if it's a registered extension hyperspy will handle that.

But there are some Signal methods (I think those involving transposes) that are inherited in a subclass and for some reason return a hyperspy signal in one of the hs.signals classes rather than in the original signal class defined in the external package. This behaviour is the real problem in my opinion and I thing that may be solvable by re-writing methods rather than (or as well as) registering external packages. I will try to write down specific functions that do this.... (watch this space). I think this happens because sometimes functions obviously don't return the same signal type. Perhaps there is some kind of 'on the fly registration' option so that if a function is applied to a signal in a given class there is always the possibility for that class to be returned.

Similarly for the external Components - ~90% of methods are fine, it's just things like as_signal() that throw issues with the DummyComponentContainer().

For registering - I agree 2 & 3 sound good and a choice is always nice.

I like your proposal about features that don't currently fit in HyperSpy's structure in principle - I'm just not sure how that will fit with having a consistent hyperspy API with external packages. E.g. would you be imagining (continuing the pyxem example) something like:

dp = hs.signals.ElectronDiffraction(data)
library = pxm.LibraryGenerator(structure, orientations)
results = dp.correlate(library)

To me, that's fine, but a little unintuitive. In that case, will we always be starting work using such an extension by importing both libraries? I guess that might be a bit weird with a manual registration (i.e. your option 2)?

magnunor:
Some random thoughts on this, as I've made projects built upon HyperSpy: Atomap and fpd_data_processing.

Atomap is a bit special, since it mostly deals with vector data, as opposed to HyperSpy which mostly deals with raster data. So it uses its own classes which does not inherit any of the HyperSpy signals, however it follows the HyperSpy "convention" with for example `sublattice.plot()`, and having most of the functionality accessible through class methods. The output data is HyperSpy signals.

fpd_data_processing is more in line with HyperSpy's structure, where it defines two new classes `PixelatedSTEM` and `DPCSignal`, which both are based on `Signal1D`, `Signal2D`, `LazySignal1D` or `LazySignal2D`. Here, most of the functionality is implemented using class methods, but the functions doing the actual processing is contained in their own modules which (mostly) accepts NumPy arrays (just so that other people could in theory access the processing functionality without having to use HyperSpy signals). In fpd_data_processing I've implemented my own `load` function, which (mostly) uses the HyperSpy `load` function to actually get the data.

So I think this discussion has two different branches: the more "formal" architecture, and the webpage/documentation/advertisement

Architecture:
- I think the easiest is for the different libraries to wrap `hs.load`, or make their own load functions.
- General functionality in the wider ecosystem should be "upstreamed" to the "core" HyperSpy, if it is useful for other projects.
- Some functions (such as `s.T`) returns a `Signal2D` object, instead of a class object (for example PixelatedSTEM). I'm guessing this should be possible to change.

Webpage:
- Having a list of projects built upon HyperSpy would be nice, for example like @thomasaarholt suggested.

ericpre:
I think it could be good to consider a bit more the usability from the user perspective. I remember to have complaint myself a while ago about having to add a `%matplotlib qt` before importing `hyperspy`... after I understood the reason behind it, I moved on to explain it to a couple of people which were also complaining about it!!
I guess, that users would appreciate not to have to worry about importing 4-5 different library (in a specific order, possibly) before starting doing anything. Even if we expect most of the users to use the notebook, I personally use quite often the ipython qtconsole...
I would clearly prefer the option 2, although I don't see the need of disabling an extension (maybe to avoid populating the hyperspy namespace with too much objects) and the registration could be done automatically as it is for GUIs extensions at the moment.
If option 3 means that all extensions need to be imported manually before registration, it could be quite tedious, but the user could choose to opt for 2 or 3 with a preference setting.

It sounds to me that most of the technicalities mentioned here should be possible to sort out. The registration mechanism for GUIs (traitsui and ipywidgets) works just fine and is fairly transparent to the user. For _foreign_ hyperspy objects coming from extensions,  it should be possible to add these objects to the hyperspy namespace at the time of the registration to the extension, basically importing the extension api into hyperspy (maybe an extension api specific to hyperspy?). The hyperspy user guide would document the existence of these objects accordingly and refer/links to the corresponding extension user guide, so it should be clear to the user "what's what".
Regarding the method returning a signal other than the class object, if this method is override by the child class it should be possible to make it return what you want. For the `load` function, maybe it could work with different io plugins for different extensions? If some signals need to be tied up together, maybe it would be an example of good synergy if it pushes hyperspy to rethink the how different signals work together?

ericpre:
I am for merging this PR for the following reasons:
- the (small) proposed changes _only_ introduce the concept of hyperspy-extensions and there is a consensus on this.
- taking small step as this one allows time for discussion, so starting to document what is hyperspy-extensions is useful.
- I think this is a good first step and we will refine details in future PRs.

doc/user_guide/intro.rst
```diff
@@ -17,6 +17,22 @@ of the dataset.
...
+In this document we refer to external programs that build on HyperSpy as
+"HyperSpy extensions". Those programs can e.g. provide extra data analysis tools
+file formats, add graphical user interfaces, provide
+extra blind source separation algorithms etc.
+
+There are multiple HyperSpy extensions. For a list of extensions hosted
+publicacly in GitHub search for the GitHub topic `hyperspy-extension <https://github.com/topics/hyperspy-extension>`_
+
+.. note::
+    From version 2.0, HyperSpy will be splitted into a core package (HyperSpy)
```
dnjohnstone: Split not splitted -- the latter is not a word.

doc/user_guide/intro.rst
```diff
@@ -17,6 +17,22 @@ of the dataset.
...
+    that will provide the common infraestructure and a number of HyperSpy
```
dnjohnstone: Typo -- infrastructure.

francisco-dlp:
Fixed, thanks @dnjohnstone

PR #1833, opened 2018-02-13
FIX-ENH BCF: (fix #1801) extend sfs file tree reading

### LINKED ISSUE #1801: Error loading Bruker EDS data (BCF files)
I've acquired a Bruker EDS map in BCF format from a ROI in an SEM image. Originally the file contained both EBSD data, however I've just saved the EDS component to keep the file size down.

When I try and load the file
```python
import hyperspy.api as hs
bk = hs.load('eds.bcf')
```

I get the following error message
```python
~/Documents/hyperspy/hyperspy/io_plugins/bcf.py in _parse_image(self, xml_node, overview)
    671             if any(array1):
    672                 item = self.gen_hspy_item_dict_basic()
--> 673                 data = array1.reshape((image.height, image.width))
    674                 detector_name = str(img.find('./Description').text)
    675                 item['data'] = data

ValueError: cannot reshape array of size 114688 into shape (448,512)
```
The full error can be found at https://gist.github.com/muddywellyboots/b4a47283e873641ab6dd5219bea6231d

The file can be found at [eds.bcf.zip](https://github.com/hyperspy/hyperspy/files/1639757/eds.bcf.zip)

sem-geologist:
Long time ago I cropped up the sfs tree reader function as the internal file tree from bcf for EDX hypermaps is never exceeding single data chunk. However this prevents reading EDX hypermaps from EBSD-EDX combined bruker bcf files, as those contains large file trees always exceeding the single data chunk.
By bringing back extended sfs tree reader, such files can be read, and EDX part can be imported into hyperspy without a problem.
Additionally, the full reading functionality will allow to go forward with EBSD data reading implementation (if anyone would wish to implement it) as it allows to read and extract all the files from the bcf container (using sfs methods in bcf library).

With making Rectangle permanent marker as optional for overview image 
~~This partly~~ completely fixes the issue reported in #1801 .

### Description of the change
* Changes the file tree reading part in the io_plugins/bcf.py
* checks if the permanent rectangle metadata is actually saved when overview image is present

### Progress of the PR
- [x] SFS reading fix,
- [x] check presence of overview rectangle,
- [x] append test,
- [x] ready for review.

### Impact to source size:
+1.1MB due to new bcf(ebsd-bearing) test file


magnunor:
You could maybe reduce the size of the binary files by zipping them: https://github.com/hyperspy/hyperspy/pull/1491/files#diff-371868400621ecb704a1d7ecadc552e0R294

To avoid increasing the repository size, you would have to rebase away the original uncompressed binary file.

sem-geologist:
@magnunor , this looks useful, but... @francisco-dlp , would not be it a good idea to do this for all tests? instead of organizing zipping/unzipping per format?

1. All tests could be zipped into single archive; unzipped once before running tests, removed temp after tests.
2. Do that per format, but move zipping/unzipping, temp creating-removing logic into misc, so it could be reused per format, instead of code repetition.
Both should be probably done as separate PR?

ericpre:
We need to be able to run tests individually, so for that, separate archive would be better I guess.
Also I would keep as simple as possible because there is no need to make it sophisticated. Personally, I find the use of the `setup_class` and `teardown_class` fairly simple. In case of the velox emd file tests, the zip is unpack in the io/emd_files folder, so you don't need to go your temp folder, which I find also convenient if you want to have a look at the data.

sem-geologist:
> ...`setup_class` and `teardown_class`...

...huh what?

sem-geologist:
@ericpre , I mean is there some other way than @magnunor proposed?

magnunor:
For the `setup_...` and `teardown_...` functionality: https://docs.pytest.org/en/2.7.3/xunit_setup.html

ericpre:
@sem-geologist, sorry I forgot to put the link to what I was referring:
https://github.com/hyperspy/hyperspy/blob/RELEASE_next_minor/hyperspy/tests/io/test_emd.py#L180

In short, yes same idea, slightly different way to do it.

sem-geologist:
Should I do it in this PR, can't I do that in the future? (in PR aimed  to minor, not patch version)

ericpre:
I think @magnunor suggested to do it now and remove the uncompressed 1Mo file from the git history by doing a rebase in order to reduce the total repository size. So, it may be worth doing it in this PR if you do the rebase, if not this is not a big deal.

sem-geologist:
I don't get Why I should do rebase? This is bug fix of #1801, it rather should go to patch. should not it?

ericpre:
Yes, this is the correct branch, the suggestion is to remove commit a4edf17 from the history to avoid increasing the repo size when unnecessary.

sem-geologist:
But that file increases coverage. I can do separate PR where I will zip **all** bcf test files (I would rather address it in next_minor)

sem-geologist:
It (The test file) increases coverage not only for this PR, but also restores part of coverage lost with #1779 . 1MB is not much when comparing with total weight of all test files.

ericpre:
Sure, we want to keep the test file!
Sorry, if it was not clear: what we mean is to remove the 1Mo file form the history but add this file to the zip file. Anyway, the compression only reduces the size by about 25% so it is not massive but it is still an improvement.
In any case, since this is not a big deal, we can leave it as it is for now.

sem-geologist:
I think this situation needs some broader disscussion, I am going to make sepparate issue.

PR #1835, opened 2018-02-15
EDS Bremsstrahlung

ZanettaPM:
### Description of the change

A background component has been added to enhance what we obtain with Polynomial background. 
This Background is based on Kramer emission law  [Kramer 1923] (other emission models are also available) and convoluted with absorption model from Love and Scott [Sewell. et al, 1985] and the detector efficiency curve of the TEM.

Key words: Kramer's Law, Mass absorption mixture function, quantification, better fit

### Progress of the PR
- [x] Change implemented:
The component can be added to a model
The component is fitted with other components (gaussians)
The fit works with datasets with  1,2 or 3 dimensions

- [x] update docstring,
- [x] update user guide 
- [x] ready for review.

SEM example

```python
s=hs.datasets.example_signals.EDS_SEM_Spectrum()
s.add_lines()

m=s.create_model(auto_background=False)
m.add_physical_background(emission_model='Lifshin',correct_for_backscatterring=False)
m.components.Bremsstrahlung.initialize() 
m.fit_background(bounded=True)
m.fit()
m.plot(plot_components=True)
```

TEM example
```python
s = hs.datasets.example_signals.EDS_TEM_Spectrum()
s=s.isig[1.18:]
s.add_elements(['Ga','Cu','Si'])
s.add_lines()

# This part is an example of the quantification effect on the background modeling
kfactors = [1.77,1.450226,1,5.075602,1] #For Fe Ka and Pt La
bw = s.estimate_background_windows(line_width=[5.0, 2.0])
intensities = s.get_lines_intensity(background_windows=bw)
Weight_percent = s.quantification(intensities, method='CL', factors=kfactors,composition_units='weight' )
#

m=s.create_model(auto_background=False)
m.add_physical_background(quantification=Weight_percent) #here the detector is not the good one which explain the deviation at low energy
m.components.Bremsstrahlung.initialize()
m.fit_background(bounded=True) 
m.fit(bounded=True)
m.plot(plot_components=True)
```


jeinsle:
this is very interesting and very relevant to some of the tasks we have been working on here for SEM-EDS experiments.  Have you tested this for quantification?  and can we pull it down here for testing?

francisco-dlp:
Thanks for this contribution.

Is this a known method or is it innovative science-wise? If the answer is yes, we won't be able to merge it until the science is published. Of course we don't need to wait until  then for reviewing the code (reviewing the science is up to the scientific reviewers of course).

ZanettaPM:
Thanks for your comments ! 
@jeinsle for the moment I didn't use this code for SEM-EDS quantification but I know that @CorentinLG will  use it in future weeks in order to enhance is TEM quantification. 

Of course you can download my github branch (Bremmsstrahlung) to test this function. Some modifications/corrections still have to be implemented but, of course, you can. (Don't hesitate to send me your detector efficiency data if you wanna that I pull it in hyperspy). There is a LPSC abstract [here](https://www.hou.usra.edu/meetings/lpsc2018/pdf/2232.pdf) which describe a part of this work and a paper that I've started to write (if you can refer to it when it will be published). 

@francisco-dlp this part is only the implementation of known methods. I've deleted parts which are unpublished yet. So you can already merge it !

hyperspy/_components/physical_background.py
```diff
@@ -0,0 +1,205 @@
...
+                    u[i] =w[i].data
+                elif "La" in w [i].metadata.General.title:
+                    u[i] =w[i].data*2.5
+                elif "La" in w [i].metadata.General.title:
+                    u[i] =w[i].data*2.8        
+            g=np.ones([len(model._signal.metadata.Sample.elements)] )
+            t=u.sum() 
+            for i in range (0,len(u)):
+                g[i] =u[i] /t*100
+    elif type(quantification) is np.ndarray: # if quantification is already an array of weight percent, directly keep the array
```
ZanettaPM: This condition doesn't work very well. It would be easier if lists were also admited. The problem is that the result of s.quantification is also a list.

.gitignore
```diff
@@ -42,3 +42,21 @@ hyperspy/io_plugins/unbcf_fast.c
 hyperspy/misc/etc/test_compilers.obj
 /.cache/
 /.spyproject/
+test/test background.ipynb
```
thomasaarholt: A bunch of the files in your gitignore are being merged. You should remove .gitignore from the PR

hyperspy/_components/physical_background.py
```diff
@@ -0,0 +1,205 @@
...
+        self.Window.free=False
+        self.quanti.free=False
+        
+        self.isbackground=True
+
+        # Boundaries
+        self.coefficients.bmin=0
+        self.coefficients.bmax=None
+
+    def initialyze(self): # this function is necessary to initialyze the quant map
```
thomasaarholt: Spelling: it's initialise or initialize, not initialyze

hyperspy/_components/physical_background.py
```diff
@@ -0,0 +1,205 @@
...
+            self.quanti.value=Wpercent(self.model,E0,self._whitelist['quanti'])
+        else: 
+            self.quanti.map['values'][:] = Wpercent(self.model,E0,self._whitelist['quanti'])
+            self.quanti.map['is_set'][:] = True
+        
+        self.Window._number_of_elements=len(self.model._signal.axes_manager.signal_axes[-1].axis)
+        self.Window._create_array()
+        self.Window.value=Windowabsorption(self.model,self._whitelist['detector'])
+        
+        return {'Quant map has been created'}
```
thomasaarholt: Is there a reason your method is returning a `set` object?

hyperspy/_components/physical_background.py
```diff
@@ -0,0 +1,205 @@
...
+
+        Mu=Mucoef(self.model,self.quanti.value)
+        
+        Mu=np.array(Mu,dtype=float)
+        Mu=Mu[self.model.channel_switches]
+        
+        Window=np.array(self.Window.value,dtype=float)
+        Window=Window[self.model.channel_switches]
+            
+        return np.where((x>0.17) & (x<(E0)),((a*100*((E0-x)/x))*((1-np.exp(-2*Mu*b*10**-5 ))/((2*Mu*b*10**-5)))*Window),0) #implement choice for coating correction
```
thomasaarholt: I'd recommend tidying this return statement (separate it into variables) so that the code is more readable.

hyperspy/_components/physical_background.py
```diff
@@ -0,0 +1,205 @@
...
+                elif "La" in w [i].metadata.General.title:
+                    u[i] =w[i].data*2.5
+                elif "La" in w [i].metadata.General.title:
+                    u[i] =w[i].data*2.8        
+            g=np.ones([len(model._signal.metadata.Sample.elements)] )
+            t=u.sum() 
+            for i in range (0,len(u)):
+                g[i] =u[i] /t*100
+    elif type(quantification) is np.ndarray: # if quantification is already an array of weight percent, directly keep the array
+        g=quantification
```
thomasaarholt: Use a more readable variable name than `g`? Same goes for `w`

thomasaarholt (commented):
First off, this is cool and great!

I've added some comments. In general, I'd like to docstrings for each of your methods, to make it easier to understand each step. In some cases you can just change the comment you have after the method into a docstring.

I have no idea what `Wpercent` is supposed to do, as an example.

Finally, you'll need to update the (website) documentation with an explanation, as well as refer to the appropriate references for the model.

hyperspy/_components/physical_background.py
ZanettaPM: Hmm not really it's more a validation step. 
The quant map can take some time to be created if the dataset is substantial. 
But obviously it's a useless term, just to reassure the user.

.gitignore
ZanettaPM: Oh yes thanks. 
And thank you for all your comments !

sem-geologist:
@ZanettaPM , you probably acidentally deleted content of and pushed `.gitignore` in commit https://github.com/hyperspy/hyperspy/pull/1835/commits/258f30d552c742709ee5b6cf28b3096f3b15919d

ZanettaPM:
@sem-geologist yes i've tried to delete my .gitignore file but i think i'm doing bad things

.gitignore
ZanettaPM: Honestly I don't know how to delete undesirable files as .gitignore. 

Should I remove those files in my local repository and start a new PR ? 

Sorry for this !

.gitignore
```diff
@@ -42,3 +43,10 @@ hyperspy/io_plugins/unbcf_fast.c
 hyperspy/misc/etc/test_compilers.obj
 /.cache/
 /.spyproject/
+
+### Code ###
+# Visual Studio Code - https://code.visualstudio.com/
```
sem-geologist: what is this?

sem-geologist:
For this physical background to work correctly, detector window characteristics is needed? I see in your documentation these det window curves. Is it possible to calculate those from known thikness and composition of the detector window (bruker i.e. contains det window composition and layer thinknesses, also it contain detector composition and thiknesses, but no such curves are present)?

ZanettaPM:
Sure ! In fact i started with this method, i've changed recently. 
I could easily create a function to build a detector efficiency curve based on the brucker information. 
These informations are contained in metadata ?

.gitignore
ZanettaPM: i reset my gitignore to resolve a merging conflict

sem-geologist:
it is in original_metada

sem-geologist:
all goodies about detector is under `eds.original_metadata.Detector`.
```python
> eds.original_metadata.Detector.WindowType
'slew AP3.3'
> eds.original_metadata.Detector.WindowLayers
├── Layer0
│   ├── Atom = 5
│   └── Thickness = 0.013
├── Layer1
│   ├── Atom = 6
│   └── Thickness = 0.145
├── Layer2
│   ├── Atom = 7
│   └── Thickness = 0.045
├── Layer3
│   ├── Atom = 8
│   └── Thickness = 0.085
├── Layer4
│   ├── Atom = 13
│   └── Thickness = 0.035
└── Layer5
    ├── Atom = 14
    ├── RelativeArea = 0.23
    └── Thickness = 380.0
```

ZanettaPM:
An idea for the condition to look at original_metadata ? 
something like: 
if data is brucker file: 
thickness=self.original_metadata.Detector.WindowLayers
?

sem-geologist:
I highly advice to add unified required information api/form for metadata, so that it could be mapped correctly if available

sem-geologist:
I mean your function would require some exact form in metadata, and io_plugins should make sure it is correctly mapped from original_metadata. I don't think it is good idea to nest "if/else this or that format/vendor" inside your methods.

ZanettaPM:
Usually we need the composition of the layer. But i guess in that case we can set 100% of each atom on different thickness on different layer 
Hope this will correspond to the reality

ZanettaPM:
> I mean your function would require some exact form in metadata, and io_plugins should make sure it is correctly mapped from original_metadata. I don't think it is good idea to nest "if/else this or that format" inside your methods.

Oh ok ! But i'm not sure to be able to modify the Api part

thomasaarholt:
As written in gitter, I suggest you merge the latest hyperspy code so that any potential errors that have been fixed in the most recent version do not appear when testing here.

thomasaarholt:
**Had another go at this below. You can mostly ignore this comment**
I just had a go at testing this branch, and get the following error after the following code, both before and after pulling upstream:

```python
import hyperspy.api as hs
s = hs.datasets.example_signals.EDS_SEM_Spectrum().isig[3.:9.]
# Create a spectrum image
nav = hs.signals.Signal2D([[1,0.8],[0.4,0.1]]).T
S = s * nav
m = S.create_model(auto_background=False)
m.add_physical_background()
m.plot()
```
Error `TypeError: object of type 'float' has no len()`:
https://gist.github.com/thomasaarholt/9ce38698ef6cc47ae3899f63adff7506

thomasaarholt:
I had a second go following the instructions in `m.add_physical_background`.
The following works
```
import hyperspy.api as hs

s = hs.datasets.example_signals.EDS_TEM_Spectrum()
s.add_lines()
kfactors = [1.450226, 5.075602] #For Fe Ka and Pt La
bw = s.estimate_background_windows(line_width=[5.0, 2.0])
intensities = s.get_lines_intensity(background_windows=bw)
Weight_percent = s.quantification(intensities, method='CL', factors=kfactors,composition_units='weight' )

m=s.create_model(auto_background=False)
m.add_physical_background(quantification=Weight_percent)
m.components.Bremsstrahlung.initialize() # Carefull it's a necessary step
m.fit_background(bounded=True)
m.fit()
m.plot()
```
Although the fit isn't terrific, but that's maybe because it is an example signal?

I observe that the fit doesn't work if s is a spectrum image.

ZanettaPM:
For your first error @thomasaarholt i think it's because you have to initialize the component before plot
In that way the quanti map will be created and it is necessary to calculate the whole signal of the background component

ZanettaPM:
for the second one, for me it has worked every time i had used spectrum images so i'll need an example to understand. However, i know that there is an error if the spectrum image is just a line of pixels, is that your case ? 
Lastly concerning the TEM example, indeed the background in TEM is really weak so the detector efficiency curve is really important. The problem is that i don't know the detector of this example. You can approach it by using the SuperX detector curve because it's a TEM detector (totally different of the PolymerC which is a SEM detector) but still, it's not the appropriate detector efficiency curve

ZanettaPM:
m.add_physical_background(detector='SuperX' , quantification=Weight_percent)

thomasaarholt:
@ZanettaPM, the code to turn it into a SI:

```
import hyperspy.api as hs
nav = hs.signals.Signal2D([[1.,0.8], [0.6, 0.4]]).T # define navigation space
s = hs.datasets.example_signals.EDS_TEM_Spectrum()
s = s * nav # just multiply it. The new signal inherits the signal type of the first signal in the equation

s.add_lines()
kfactors = [1.450226, 5.075602] #For Fe Ka and Pt La
bw = s.estimate_background_windows(line_width=[5.0, 2.0])
intensities = s.get_lines_intensity(background_windows=bw)
Weight_percent = s.quantification(intensities, method='CL', factors=kfactors,composition_units='weight' )

m=s.create_model(auto_background=False)
m.add_physical_background(quantification=Weight_percent)
m.components.Bremsstrahlung.initialize() # Carefull it's a necessary step
m.fit_background(bounded=True)
m.fit()
m.plot()
```

ZanettaPM:
Okay i'll try this tomorrow and let you know

ZanettaPM:
> @ZanettaPM, the code to turn it into a SI:
I clearly don't understand what happen... it works for every other SI that i've used.. i'll try to find the issue

ZanettaPM:
Ahah ! @thomasaarholt  you'll laugh ! 
The problem come from the m.fit function... we forgot to write "kind='multi'' and m.multifit()

CorentinLG:
I have a case for which the background fit is failing when there is no signal intensity at a given pixel. 
So if you have a SI of a sample with holes in it, the fit won't run. You could mask the holes before running the fit of course, but it is not ideal.

Here is the error I got:
'''python

<ipython-input-10-f5e234a90cff> in <module>()
----> 1 m.fit_background(kind = 'multi', bounded = True, windows_sigma=(3, 3)) #mask = ~M1.data,

C://Users/Corentin/Documents/Github/hyperspy/hyperspy\models\edsmodel.py in fit_background(self, start_energy, end_energy, windows_sigma, kind, **kwargs)
    448                 self.fit(**kwargs)
    449             if kind == 'multi':
--> 450                 self.multifit(**kwargs)
    451             self.reset_signal_range()
    452         self.fix_background()

C://Users/Corentin/Documents/Github/hyperspy/hyperspy\model.py in multifit(self, mask, fetch_only_fixed, autosave, autosave_every, show_progressbar, interactive_plot, **kwargs)
   1514                                 self.fetch_stored_values(
   1515                                     only_fixed=fetch_only_fixed)
-> 1516                                 self.fit(**kwargs)
   1517                                 i += 1
   1518                                 pbar.update(1)

C://Users/Corentin/Documents/Github/hyperspy/hyperspy\model.py in fit(self, fitter, method, grad, bounded, ext_bounding, update_plot, **kwargs)
   1287                     output = \
   1288                         least_squares(self._errfunc, self.p0[:],
-> 1289                                       args=args, bounds=ls_b, **kwargs)
   1290                     self.p0 = output.x
   1291 

C:\Anaconda3\envs\HyperspyDev\lib\site-packages\scipy\optimize\_lsq\least_squares.py in least_squares(fun, x0, jac, bounds, method, ftol, xtol, gtol, x_scale, loss, f_scale, diff_step, tr_solver, tr_options, jac_sparsity, max_nfev, verbose, args, kwargs)
    803 
    804     if not np.all(np.isfinite(f0)):
--> 805         raise ValueError("Residuals are not finite in the initial point.")
    806 
    807     n = x0.size

ValueError: Residuals are not finite in the initial point.

ZanettaPM:
Yes @CorentinLG i had this problem few weeks ago. 
The new comits should fix it ! 
I also added the possibility to put a phase map in the attribute of the function. The objective is to calculate a finite number of mass absorption coefficient curves instead of calculate a new curve in each point. Of course you still can calculate (as before) a mass absorption curve in each point without the quanti array but this process takes more time. 

I have to update the documentation about this...

jlaehne:
Closed in favor of #2836

PR #1839, opened 2018-02-17
bcf_fix_addendum

sem-geologist:
### Some more fixes for bcf

which accidentally missed the #1833. 
This PR fixes the main error addressed at the beginning of #1801.

Up till this PR, all SEM images contained inside bcf xml header were thought to be statically of 16bit dtype.
The new sample files shed the light that <ItemSize> tag holds dtype information (it is saved as integer where 1 = uint8; 2 = uint16 and so on..).

Additionally, some microscopes/OEM with EBSD bcf have no description for some of SEM signal/images, which is addressed too in this PR.

### DIFF
--- a/hyperspy/io_plugins/bcf.py
+++ b/hyperspy/io_plugins/bcf.py
@@ -677,7 +677,7 @@ class HyperHeader(object):
         image = Container()
         image.width = int(xml_node.find('./Width').text)  # in pixels
         image.height = int(xml_node.find('./Height').text)  # in pixels
-        image.dtype = 'u' + xml_node.find('./ItemSize').text 
+        image.dtype = 'u' + xml_node.find('./ItemSize').text  # in bytes ('u1','u2','u4') 
         image.plane_count = int(xml_node.find('./PlaneCount').text)
         image.images = []
         for i in range(image.plane_count):
--- a/hyperspy/io_plugins/bcf.py
+++ b/hyperspy/io_plugins/bcf.py
@@ -688,14 +688,12 @@ class HyperHeader(object):
                 item = self.gen_hspy_item_dict_basic()
                 data = array1.reshape((image.height, image.width))
                 desc = img.find('./Description')
-                #detector_name = str(dscrp.text) if  else ''
                 item['data'] = data
                 item['axes'][0]['size'] = image.height
                 item['axes'][1]['size'] = image.width
                 item['metadata']['Signal'] = {'record_by': 'image'}
                 item['metadata']['General'] = {}
                 if desc is not None:
-                    item['metadata']['Signal']['signal_type'] = str(desc.text)
                     item['metadata']['General']['title'] = str(desc.text)
                 if overview and (rect_node is not None):
                     item['metadata']['Markers'] = {'overview': over_dict}
--- a/hyperspy/io_plugins/bcf.py
+++ b/hyperspy/io_plugins/bcf.py
@@ -677,22 +677,26 @@ class HyperHeader(object):
         image = Container()
         image.width = int(xml_node.find('./Width').text)  # in pixels
         image.height = int(xml_node.find('./Height').text)  # in pixels
+        image.dtype = 'u' + xml_node.find('./ItemSize').text 
         image.plane_count = int(xml_node.find('./PlaneCount').text)
         image.images = []
         for i in range(image.plane_count):
             img = xml_node.find("./Plane" + str(i))
             raw = codecs.decode((img.find('./Data').text).encode('ascii'),'base64')
-            array1 = np.frombuffer(raw, dtype=np.uint16)
+            array1 = np.frombuffer(raw, dtype=image.dtype)
             if any(array1):
                 item = self.gen_hspy_item_dict_basic()
                 data = array1.reshape((image.height, image.width))
-                detector_name = str(img.find('./Description').text)
+                desc = img.find('./Description')
+                #detector_name = str(dscrp.text) if  else ''
                 item['data'] = data
                 item['axes'][0]['size'] = image.height
                 item['axes'][1]['size'] = image.width
-                item['metadata']['General'] = {'title': detector_name}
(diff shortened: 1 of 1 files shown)


hyperspy/io_plugins/bcf.py
```diff
@@ -677,22 +677,26 @@ def _parse_image(self, xml_node, overview=False):
...
                 item['data'] = data
                 item['axes'][0]['size'] = image.height
                 item['axes'][1]['size'] = image.width
-                item['metadata']['General'] = {'title': detector_name}
-                item['metadata']['Signal'] = {'signal_type': detector_name,
-                                              'record_by': 'image'}
+                item['metadata']['Signal'] = {'record_by': 'image'}
+                item['metadata']['General'] = {}
+                if desc is not None:
+                    item['metadata']['Signal']['signal_type'] = str(desc.text)
```
ericpre: The `signal_type` is not supposed to be set with this. I think if you don't have the correct metadata to put there, then just ignore it. The information is already in the title just below, so this is enough.

ericpre (commented):
The file "bcf-edx-ebsd.bcf" was reading fine before this PR? Is the issue happening with other files than this one, then?
Is it normal that the EDS spectrum image from "bcf-edx-ebsd.bcf" mainly made of zeros signals, except in the middle?

Maybe worth adding a test to check the data type and some of the metadata are correct (in addition to just read the file)?

hyperspy/io_plugins/bcf.py
sem-geologist: Esprit (Bruker's software) behaves differently on different microscopes. It have two channels for getting SEM images (getting images by ext scan using the SEM detectors). Some instruments have intelligent link of Esprit, and it is able to get the signal name correctly; other instruments have limited link with SEM, and have no way in knowing what type of detectors is set for channel 1 and channel 2. For such a case, Esprit gui allows to provide it manually under `image` setup tab. If user is not aware of that, the signals are by default named as `Ch 1` and `Ch 2` (or `signal 1` and `signal 2`, or other linguistic variants, which is). In many cases User sets those names accordingly (e.g. SEI, BSE). In some cases it can get out of sync (it is up to user to keep it named rightly, if system is not intelligent). However there can be more than 2 images saved in bcf, and normally signals of 3rd and more images are named automatically (e.g. `Argus` detector when EBSD detector is inserted). All of these use tag <Description> and looking to other libraries and examples I thought that that should be mapped to `metadata.Signal.signal_type`. But I relooked the Metadata structure in documentation and you are right. SEI, BSE, Argus... should not go there, only EDS_TEM and EDS_SEM; I don't know how I missed this.

Generally I see that `Metadata structure`  in docs got quite much updated (there is such thing as Stage) I think it will be good idea to update bcf spec and map some of available metadata into hyperspy signals metadata.

sem-geologist:
yes "bcf-edx-ebsd.bcf" (which is the first minimal file example `bcf3.bcf` provided by @muddywellyboots in #1801 (look at middle of post)) is read correctly before this PR, that's why original bug had slipped out from my mind. However the file (huge file) shared at top of #1801 is read only after this PR. No minimal example can replicate the data type behavior (nobody knows why). maybe it is microscope specific? Anyway the  Tests for data type is not needed as it is very straight forward and logical. In case of float/double type of data (e.g. quantification results), Bruker uses XML structure and plain text, these however is not currently read and returned into hyperspy. Only `uint*` kind of images are saved with flattened/serialized stream coded with base64 algorithm.

The file does not contains zeroes (I mean it is not saved as zeros, it is saved as empty pixels, the empty line takes exactly 32bits). It looks a bit weird. But this is OEM specific way of gathering mapping for small selected rectangle. There is two known for me ways:
1. Esprit takes separate overview image, and during mapping it registers microscope images (of same kind as in overview) and saves it. In version 2 bcf it puts rectangle marker on overview image marking where smaller mapping (and images) were acquired. (This type of behavior I am well aware as our Zeiss-Bruker microscope systems behave like that)
2. Esprit takes overview image, but during mapping it registers no microscope image(s) for selected region. It does not attach the rectangle marker to overview image. Instead it saves whole size of array as the original overview image, with empty pixels outside of rectangle. (As I dont have system with such behavior, It is not obvious for me how it should map into hyperspy, as I dont know how it is represent in Esprit)

for "bcf-edx-ebsd.bcf" there is no need for another test, as meta data is not much different from any other casual bcf (with the exception there one of images misses <Describtion> tag, which probably comes from FIB imaging, and is OEM specific). As tag reading is now optional there is not much what to test.

Maybe for 2nd way it could be possible to emulate the 1st behaviour. I think I will need to review this format and decide what to implement. I will open new issue and will list what features of bcf is still not implemented in hyperspy, people then can vote what they would like to prioritize. In last half year I found some interesting things in this format, and updated hyperspy metadata structure made some of them possible to be read into sensibly.

For this small PR, which is bug-fix, I am only going to tidy up  `signal_type` metadata field.

muddywellyboots:
@sem-geologist Nice work! I've tried loading a selection of files from our kit and there are no issues now.

hyperspy/io_plugins/bcf.py
```diff
@@ -677,22 +677,24 @@ def _parse_image(self, xml_node, overview=False):
         image = Container()
         image.width = int(xml_node.find('./Width').text)  # in pixels
         image.height = int(xml_node.find('./Height').text)  # in pixels
+        image.dtype = 'u' + xml_node.find('./ItemSize').text
```
sem-geologist: maybe I should add the coment what is going here?
`<ItemSize>` tag holds the pixel size in bytes, which is read here as a string then converted into numpy-able dtype notation, by prepending letter 'u', resulting into e.g. `u1`,`u2`,`u4`. I really hope there is no `u3` (24-bits), as it would be not so straightforward to read into numpy.

ericpre: Yes, adding a comment doesn't cost much!!

sem-geologist:
So can this get merged?

Issue #1842, opened 2018-02-21
Stdout is polluted with progress bars


uellue:
When working on analyzing the scaling behavior of excessive dask overheads (Issue #1840) as a function of the input data size, I came across two surprising and annoying behaviors of the compute() method. http://hyperspy.org/hyperspy-doc/current/api/hyperspy._signals.html?highlight=lazysignal#hyperspy._signals.lazy.LazySignal.compute

Current behavior: The compute() method displays a progress bar by default. The progress bar is written to stdout. This output got mixed with the actual benchmarking data that I was writing to stdout.

Expected behavior: The compute() method is quiet by default, at least when running in a script. The progress bar is written to stderr.

francisco-dlp:
The issue is that ``compute`` can not know if it is running in a script or in an interactive session. Typical HyperSpy usage involves an interactive session eventually followed by running a script once the code is refined. I think that the progress bar is useful in an interactive session. If it is an issue for a particular use case, it can always be disabled for a given method or, more generally, for all methods via the ``prefereces`` e.g.

```python
import hyperspy.api as hs
hs.preferences.General.show_progressbar = False
```
This will only affect the current session. For permanently disabling it for a given use:

```python
import hyperspy.api as hs
hs.preferences.General.show_progressbar = False
hs.preferences.save()
```
All of this is fine, however, you've identified a bug because ``LazySignal``'s ``compute`` doesn't honour the default set in ``preferences``, so I'm marking this as an easy to fix bug.

uellue:
@francisco-dlp At least on Unix you can easily get this information from `sys.stdout.isatty()` and `sys.stderr.isatty()` -- generally the `isatty()` method of a file object. 

Regarding stdout vs. stderr there are discussions, but it seems that stderr is better suited. https://unix.stackexchange.com/questions/331611/do-progress-reports-logging-information-belong-on-stderr-or-stdout

Update: To be precise, it doesn't tell you if running interactively, but it DOES tell you if stdout is redirected to a file. Another indicator is this one: https://stackoverflow.com/questions/2356399/tell-if-python-is-in-interactive-mode

magnunor:
It is also possible to manually disable the progressbar when calling `s.compute` by using `progressbar=False`: `s.compute(progressbar=False)`.

HyperSpy uses `tqdm` for these progressbars, which should default to `sys.stderr`

sk1p:
[...]

PR #1843, opened 2018-02-21
Changed default behavior: Activate progressbar only if running interactively

uellue:
Otherwise deactivate progress by default. Fix for issue #1842 

Currently dask uses stdout for progress bars. A fix to switch to stderr by default and give users control
is in preparation.

### DIFF
--- a/hyperspy/_signals/lazy.py
+++ b/hyperspy/_signals/lazy.py
@@ -21,7 +21,7 @@ from functools import partial
 
 import numpy as np
 import math as math
-import sys
+import __main__ as main
 import dask.array as da
 import dask.delayed as dd
 from dask import threaded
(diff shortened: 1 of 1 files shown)


francisco-dlp:
I am not sure if this fixes the issue because running from a script in a terminal still prints a progress bar. Only when the stdout of the script is redirected elsewhere the progress bar is not printed. This may sound like a good idea anyway but it can backfire. For example, we lose the progress bar in the jupyter notebook because the jupyter notebook is not a tty.

uellue:
@francisco-dlp funny that Jupyter doesn't emulate a terminal. The second option is
```
import __main__ as main
if hasattr(main, '__file__'):
    ...
```

thomasaarholt:
Just as a possiblity, one can check for iPython/Jupyter simply by `get_ipython` ([from here](https://stackoverflow.com/questions/44100477/how-to-check-if-you-are-in-a-jupyter-notebook)).

So something like this. When in the notebook or qtconsole, no import statement is necessary:
```python
try:
    get_ipython
    is_ipython = True
except:
    is_ipython = False

tjof2:
Is this still an issue/a desired change?

tjof2:
Replaced by #2548

PR #1851, opened 2018-02-28
Fix plot component

### LINKED ISSUE #1847: EELS model components scaled incorrectly due to convolution with low-loss
When creating an EELS model, if I convolve the high-loss signal with the low-loss signal, the model components are scaled incorrectly. I tried to normalize the low-loss, but the problem persists.

In [this](https://nbviewer.jupyter.org/github/theodoho/Gaussian_Amplitude/blob/master/EELS%20fine%20structure%20analysis%20Mn-L%20Sum%20and%20Ratio%203.ipynb) notebook, I do not convolve the high-loss with the low-loss, **cell 14**, and the model components are consistent with the actual model and data, as can be seen in **cell 38**.

In [this](https://nbviewer.jupyter.org/github/theodoho/Gaussian_Amplitude/blob/master/EELS%20fine%20structure%20analysis%20Mn-L%20Sum%20and%20Ratio%20Convolved.ipynb) notebook, I convolve the high-loss with the low-loss, **cell 14**, and the model components are clearly not consistent with the actual model and data, as can be seen in **cell 38**. They are way too low.

In both cases I plot the normalized white-line sum (**cell 45**) and the white-line ratio (**cell 46**). The qualitative trends are the same, but the absolute values change; the change is slight for the white-line ratio and very big for the normalized white-line sum.

ericpre:
This PR resolves #1847.

### Progress of the PR
- [x] Plot convolved component,
- [x] add a note to the user guide to explain the small fitted value when the component are convolved with low loss spectrum,
- [x] add tests,
- [x] ready for review.

```python
>>> import hyperspy.api as hs
>>> s = hs.datasets.eelsdb(title="Hexagonal Boron Nitride", spectrum_type="coreloss")[0]
>>> ll = hs.datasets.eelsdb(title="Hexagonal Boron Nitride", spectrum_type="lowloss")[0]
>>> m = s.create_model(ll=ll)
>>> m.enable_fine_structure()
>>> m.smart_fit()
>>> m.plot(plot_components=True)
```


thomasaarholt:
THIS IS EXCITING

Right. Going to check in the morning.

thomasaarholt:
What's up with the tests failing with pngs?

thomasaarholt:
This is extremely nice!
Seems to work well, both with and without ll convolution.
screen shot 2018-03-08 at 15 51 54

hyperspy/component.py
```diff
@@ -1027,8 +1020,6 @@ def _component2plot(self, axes_manager, out_of_range2nans=True):
         s = self.__call__()
         if not self.active:
             s.fill(np.nan)
-        if self.model.signal.metadata.Signal.binned is True:
-            s *= self.model.signal.axes_manager.signal_axes[0].scale
```
thomasaarholt: This seems like an important change? Was this relevant just for the convolution plot or plots in general?

thomasaarholt:
Thanks for tidying up so much of the code, that sort of spring cleaning is important too!

hyperspy/component.py
ericpre: What do you mean? Regarding binned data, there is no change in this PR.

thomasaarholt: Then why does the PR remove lines 1030 and 1031?

ericpre: Plotting component has been reworked and these lines needed to be removed to be correct. If you read the code, you will understand.

thomasaarholt: Gotcha! Was just making sure you hadn't accidentally removed something :)

hyperspy/component.py
```diff
@@ -1013,10 +1009,7 @@ def __call__(self):
         -------
         numpy array
         """
-
-        axis = self.model.axis.axis[self.model.channel_switches]
-        component_array = self.function(axis)
-        return component_array
+        return self.model.__call__(component_list=[self])
```
francisco-dlp: I think that this is a dangerous change:

*  ``__call__`` should always return the unconvolved component as it was doing previously as the output should not depend on the model and its particular ``convolve`` setting.
* ``__call__`` should work also when ``self.model`` is not defined.

hyperspy/models/model1d.py
```diff
@@ -395,35 +396,33 @@ def __call__(self, non_convolved=False, onlyactive=False):
         numpy array
         """
 
+        if component_list is None:
+            component_list = [component for component in self]
```
francisco-dlp: ``component_list = self``?

hyperspy/models/model1d.py
```diff
@@ -395,35 +396,33 @@ def __call__(self, non_convolved=False, onlyactive=False):
...
+        if isinstance(component_list, (list, tuple)):
```
francisco-dlp:
```python
elif not isinstance(component_list, (list, tuple)):
    raise ValueError(
```

doc/user_guide/eels.rst
```diff
@@ -149,14 +149,18 @@ Define the chemical composition of the sample
     >>> s.add_elements(('B', 'N'))
 
 
-In order to include the effect of plural scattering we provide the low-loss
-spectrum to :py:meth:`~._signals.eels.EELSSpectrum_mixin.create_model`:
+In order to include the effect of plural scattering, the component of the model is convolved with the loss loss spectrum in which case the low loss spectrum needs to be provided to :py:meth:`~._signals.eels.EELSSpectrum_mixin.create_model`:
```
francisco-dlp: the component of the model is convolved  -> the model is convolved

doc/user_guide/eels.rst
```diff
@@ -149,14 +149,18 @@ Define the chemical composition of the sample
...
 
 .. code-block:: python
 
     >>> m = s.create_model(ll=ll)
 
 
+.. NOTE::
+
+    In case of the convolution with the low loss spectrum, the fitted value of some of the parameter may seem to be unexpectedly small which can be misleading. Those paramater values are actually correct since the values of the low loss spectrum are usually much larger than 1. Indeed, the `convolution <https://en.wikipedia.org/wiki/Convolution>`_ gives the integral of the pointwise multiplication of the component function with the low-loss spectrum shifted over the component and it results that the function values of the convolved component are much larger than in the non-convolved case, which means that the fitted parameter values need to be very small to fit the data.
```
francisco-dlp: I would remove this note because the reason why the numbers are small is that hyperspy assumes the dwell time of the ll and cl to be the same. We should rather fix that...

francisco-dlp (changes requested):
It looks mainly good to me except for the few comments.

hyperspy/component.py
ericpre: Yes, reverted to previous version and move the `model.__call__()` to `_component2plot`.

doc/user_guide/eels.rst
ericpre: Removed it. Possibly better to leave this for another PR. Now, we should be parsing the time correctly from DM files, so it should be fine to fix it indeed.

hyperspy/component.py
```diff
@@ -1009,15 +1009,20 @@ def __call__(self):
...
+        return component_array
 
     def _component2plot(self, axes_manager, out_of_range2nans=True):
         old_axes_manager = None
         if axes_manager is not self.model.axes_manager:
             old_axes_manager = self.model.axes_manager
             self.model.axes_manager = axes_manager
             self.fetch_stored_values()
-        s = self.__call__()
+        if self.model:
```
francisco-dlp: I think that it is unnecessary here to check if model is defined because this method is private and designed to be called only from model. The rest of the code assumes that model is defined.

francisco-dlp:
Except for the minor comment, LGTM!

hyperspy/component.py
ericpre: Yes and I think this was already changed in my last commit https://github.com/hyperspy/hyperspy/pull/1851/commits/e884931abfc1a956c789ed534172bf22497fdc75!

PR #1853, opened 2018-03-04
Add lazy reading support for FEI EMD format

francisco-dlp:
Lazy reading for all objects (spectra, images and SIs) stored in FEI EMD files.

### Progress of the PR
- [x] add tests,
- [x] ready for review.


setup.py
```diff
@@ -63,6 +63,7 @@
                'pint>=0.8',
                'statsmodels',
                'mrcz>0.3.4',
+               'sparse',
```
ericpre: `sparse>=0.2`, which has a dependency on `scipy>=0.19` (penultimate release, March 2017) and could potentially limit the installation of hyperspy in older python environment. Are we ok with this? I think this is fine and it is not worth adding the lazy support of velox emd to "extras".

francisco-dlp:
All tests related to this PR are passing, so this is ready for review.

doc/user_guide/io.rst
```diff
@@ -703,20 +703,18 @@ Electron Microscopy (NCEM). See http://emdatasets.com/ for more information.
...
-spectrum stream, FEI emd file also contains a spectrum image dataset which is 
-a proprietary format used by Velox and is not supported by HyperSpy. It is possible 
-that a file has been pruned when saving with Velox in order to reduce its size 
-(not default), which means that the spectrum stream is lost and can't be recovered.
-Loading a spectrum image is slow if `numba <http://numba.pydata.org/>`_ is 
-not installed.
+This is a non-compliant variant of the standard EMD format developed by FEI.
+HyperSpy supports importing images, EDS spectrum and EDS 
+spectrum streams (spectrum images stored in a sparse format). For spectrum
+streams, there several loading options (described below) to control the frames
```
ericpre: typo: there *are* several

doc/user_guide/io.rst
```diff
@@ -703,20 +703,18 @@ Electron Microscopy (NCEM). See http://emdatasets.com/ for more information.
...
+and detectors to load and if to sum them on loading.  The default is
+to import the sum over all frames and over all detectors in order to decrease
+the data size in memory.
+
+Note that pruned FEI EMD files contain the spectrum image. Therefore,
```
ericpre: typo: Note that pruned FEI EMD files doesn't contain the spectrum stream(s)

hyperspy/io_plugins/emd.py
```diff
@@ -621,25 +668,37 @@ def _read_image(self, image_group, image_sub_group_key):
...
-        if read_stack:
-            data = np.rollaxis(np.array(image_sub_group['Data']), axis=2)
-            # Get the scanning area shape of the SI from the images
-            self.SI_shape = data.shape[1:]
+        h5data = image_sub_group['Data']
+        # Get the scanning area shape of the SI from the images
+        self.spatial_shape = h5data.shape[:-1]
+        # Set the axes in frame, y, x order
+        if self.lazy:
+            if self.lazy:
```
ericpre: I think one `if self.lazy:` is enough!

hyperspy/io_plugins/emd.py
```diff
@@ -621,25 +668,37 @@ def _read_image(self, image_group, image_sub_group_key):
...
-        if read_stack and data.shape[0] > 1:
+        if not read_stack:
+            data = data[0:1, ...]
+
+        if data.shape[0] == 1:
+            # Squeeze
+            data = data[0, ...]
+            i = 0
+        else:
             frame_time = original_metadata['Scan']['FrameTime']
```
ericpre, line 701: Set `frame_time` to a default value in case the metadata is not present?

hyperspy/io_plugins/emd.py
```diff
@@ -718,30 +772,74 @@ def _parse_metadata_group(self, group, group_name):
...
+                            "Velox.")
+            return
+
+        def _read_stream(key):
+            stream = FeiSpectrumStream(spectrum_stream_group[key], self)
+            return stream
+
+        subgroup_keys = _get_keys_from_group(spectrum_stream_group)
+        if len(subgroup_keys) == 1:
+            _logger.warning("The file contains only one spectrum stream")
```
ericpre: It should raise a warning only when `sum_detectors=False`. Add "." at the end of the warning message. Maybe move this warning in the `else` below.

hyperspy/io_plugins/emd.py
```diff
@@ -718,30 +772,74 @@ def _parse_metadata_group(self, group, group_name):
...
+        if self.sum_EDS_detectors:
+            # Read the first stream
+            s0 = _read_stream(subgroup_keys[0])
+            streams = [s0]
+            # add other stream streams
+            if len(subgroup_keys) > 1:
+                for key in subgroup_keys[1:]:
+                    stream_data = spectrum_stream_group["key"]['Data'][:].T[0]
```
ericpre: `"key"` with "? This is not covered by tests: if I don't forgot, I will generate a few small files in the next few days to increase the tests coverage and make a PR to this branch.

hyperspy/decorators.py
```diff
@@ -107,3 +105,11 @@ def wrapper(self, *args, **kwargs):
         else:
             cm(self, *args, **kwargs)
     return wrapper
+
+
+def jit_ifnumba(*args, **kwargs):
```
ericpre, line 110: Is it worth setting the default of `nopython=True` to make sure it is not falling back to object mode? This is currently this issue with rebin (fixed in #1821) and we haven't pay attention to it because it was burried in the long list of warnings...

hyperspy/_signals/lazy.py
```diff
@@ -84,17 +84,46 @@ class LazySignal(BaseSignal):
...
+
+        """
+        arrkey = None
+        for key in self.data.dask.keys():
+            if "array-original" in key:
+                arrkey = key
+                break
+        if arrkey:
+            try:
+                self.data.dask[arrkey].file.close()
```
ericpre, line 123: The tests with `lazy=True` raise an error: `AttributeError: 'COO' object has no attribute 'file'`

hyperspy/misc/io/fei_stream_readers.py
```diff
@@ -0,0 +1,320 @@
...
+
+@jit_ifnumba
+def _stream_to_sparse_COO_array_sum_frames(
+        stream_data, last_frame, shape, channels, rebin_energy=1, first_frame=0):
+    navigation_index = 0
+    frame_number = 0
+    ysize, xsize = shape
+    frame_size = xsize * ysize
+    data_list = []
+    coords = []
```
ericpre: Rename `coords` to `coords_list` to be consistent with `data_list`. I am being a bit picky here... ;)

hyperspy/misc/io/fei_stream_readers.py
```diff
@@ -0,0 +1,320 @@
...
+            if frame_number == last_frame:
+                break
+        # if different of ‘65535’, add a count to the corresponding channel
+        if value != 65535:  # Same spectrum
+            if data:
+                if value == count_channel:  # Same channel, add a count
+                    data += 1
+                else:  # a new channel, same spectrum—requires new coord
+                    # Store previous channel
+                    coords.append((
```
ericpre, line 125: Append data to a list is known to be inefficient: would it be better to expand an array regularly, every thousand element for example?

hyperspy/io_plugins/emd.py
```diff
@@ -621,25 +668,37 @@ def _read_image(self, image_group, image_sub_group_key):
...
+                data = da.transpose(
+                    da.from_array(
+                        h5data,
+                        chunks=h5data.chunks),
+                    axes=[2, 0, 1])
         else:
-            data = image_sub_group['Data'][:, :, 0]
-            # Get the scanning area shape of the SI from the images
-            self.SI_shape = data.shape
+            data = np.rollaxis(np.array(h5data), axis=2)
```
ericpre, line 682: Accessing the hdf5 dataset is slow, possibly because there are as many chunks as images in the stack.
For the case of not loading the SI image stack, I would read only the first chunk and change accordingly below, since it is much faster (more than 10x):
``` python
if not read_stack:
    data = h5data[:, :, 0]
else:
    data = np.rollaxis(np.array(h5data), axis=2)
```

ericpre (changes requested):
This is a nice refactoring of the velox emd loader, which is less messy than before! But it needs a few fixes and tweaks.
I have tried it on ~1Go dataset and loading spectrum stream is ~20% faster but loading images is much slower.
For lazy loading, I get an `IndexError` when I compute the data.

ericpre:
This is the `IndexError` when trying to compute the lazy signal:

``` python

  File "<ipython-input-2-5e4db31e54a6>", line 1, in <module>
    s[-1].compute()

  File "/home/eric/Dev/hyperspy/hyperspy/_signals/lazy.py", line 102, in compute
    data = da.compute()

  File "/opt/anaconda3/lib/python3.6/site-packages/dask/base.py", line 143, in compute
    (result,) = compute(self, traverse=False, **kwargs)

  File "/opt/anaconda3/lib/python3.6/site-packages/dask/base.py", line 392, in compute
    results = get(dsk, keys, **kwargs)

  File "/opt/anaconda3/lib/python3.6/site-packages/dask/threaded.py", line 75, in get
    pack_exception=pack_exception, **kwargs)

  File "/opt/anaconda3/lib/python3.6/site-packages/dask/local.py", line 521, in get_async
    raise_exception(exc, tb)

  File "/opt/anaconda3/lib/python3.6/site-packages/dask/compatibility.py", line 67, in reraise
    raise exc

  File "/opt/anaconda3/lib/python3.6/site-packages/dask/local.py", line 290, in execute_task
    result = _execute_task(task, data)

  File "/opt/anaconda3/lib/python3.6/site-packages/dask/local.py", line 271, in _execute_task
    return func(*args2)

  File "/opt/anaconda3/lib/python3.6/site-packages/dask/array/core.py", line 76, in getter
    c = np.asarray(c)

  File "/opt/anaconda3/lib/python3.6/site-packages/numpy/core/numeric.py", line 492, in asarray
    return array(a, dtype, copy=False, order=order)

  File "/opt/anaconda3/lib/python3.6/site-packages/sparse/coo.py", line 1192, in __array__
    x = self.todense()

  File "/opt/anaconda3/lib/python3.6/site-packages/sparse/coo.py", line 350, in todense
    x[coords] = data

IndexError: index 65280 is out of bounds for axis 0 with size 21
```

francisco-dlp:
@ericpre, thank you for the thorough review. I am now working on fixing the issues that you've highlighted. However, I haven't been able to reproduce the ``IndexError`` exception that you report above with my single big real dataset and with synthetic ones that I have created for the purpose. Could you share the file or investigate the issue on your own if you prefer?

setup.py
francisco-dlp: Thinking about it, this can also be an issue to install in Debian. Also, sparse is not yet in conda forge https://github.com/pydata/sparse/issues/115 .Hence, I'll make lazy optional.

hyperspy/_signals/lazy.py
francisco-dlp: Yes, that's normal because the COO object is indeed not a file.

hyperspy/misc/io/fei_stream_readers.py
francisco-dlp: Does it matter when using numba? It may actually decrease the performance in this case. Also, I am not sure of where you are proposing to store the coordinates in between the expansion steps.

hyperspy/io_plugins/emd.py
francisco-dlp: Actually, I wanted to discuss with you this feature. Wouldn't it be better to remove ``read_stack`` all together? If the user is not loading lazily and the SI fits in memory, it's likely that the image stack fits too. If loading lazily, it doesn't matter at all. What do you think?

francisco-dlp:
This should be ready for review again.

ericpre:
I have uploaded two files with different version of velox, both of them raise the `IndexError` during the `compute()`:
https://www.dropbox.com/sh/dn86yst7atqkgv5/AACv6GszG-djUjna95ezgD8ua?dl=0

I forgot to generate small files to test the several streams configuration this morning, but I will do it this evening, if I don't forget again...

hyperspy/_signals/lazy.py
ericpre: ok, but the file should be closed somewhere?

francisco-dlp:
Thanks, I've fixed it in the last commit.

hyperspy/misc/io/fei_stream_readers.py
ericpre: yeah, maybe I guess this is already good enough for now!

hyperspy/_signals/lazy.py
francisco-dlp: The file is already closed because the data  is the stream data itself that is not loaded lazily. What is computed lazily here is the conversion between the sparse array and the dense array.

hyperspy/io_plugins/emd.py
ericpre: I think this is still useful. We don't want to always load the stack of images and there are cases, where users may want to load the stack of images but not the stack of SI: for example, to check if there were damage/contamination/drift visible on the image.
However, it would be good to tweak the default: when we load the stack of SI, the default of `load_SI_image_stack` should be `True` and otherwise it should be 'False'.

hyperspy/io_plugins/emd.py
ericpre: What's about this one?

hyperspy/_signals/lazy.py
ericpre: From line 543, it looks like the file is not closed:
``` python
finally:
    if not self.lazy:
        f.close()
```
and as it is now, we should not closed it, because it needs to stay open to access the spectrum and the images.
Should the `close_file` take into account this particular case and not add the error to the logger, since it is misleading?

francisco-dlp: You are right, we shouldn't close it because otherwise the lazy spectra and images will get broken. And actually, calling close doesn't actually close the file for the reason above. We could not attempt to close the file when it is a SI in the tests. However, by leaving it as is we are actually testing that the code is ``except`` is working fine. So maybe it's not that bad after all?

A side note, the addition of this feature makes only sense in windows, where, for some reason, the files are not closed after the object is deleted even after running the garbage collector. So, this is a workaround for a bug (or maybe a limitation) elsewhere and most users won't actually need to use it. It is useful here to make the tests pass in appveyor. The alternative is not removing the files as it was previously done.

hyperspy/io_plugins/emd.py
francisco-dlp: What about not loading any image of the image stack (instead of just the first one) when ``load_SI_image_stack`` is ``False``? (Note that we can get the shape without actually loading the image. Also, we maybe should get the shape from elsewhere to consider the case of a SI acquired without recording the image.)

We could also rename the kwarg to ``load_SI_image``.

hyperspy/_signals/lazy.py
ericpre: Although I would prefer that this tests has his own in the unit tests of the lazy signal (because it may look confusing to see this error when testing emd file), if you put a comment explaining this, then it is fine.

hyperspy/io_plugins/emd.py
ericpre: If loading the whole stack was as fast as loading the first image, I could prefer your suggestion, also renaming to `load_SI_image` would be more clear with a default to `True`. But it can take a couple of seconds to load the whole stack so I find it a bit annoying.

ericpre:
Even if there are still room for improvement, I would be fine with merging this for now and we can take a bit more time later to try to improve it.
It seems that there is an issue with the `jit_ifnumba` decorator in travis and appveyor.

francisco-dlp:
Yes, I'll fix that later this afternoon.

What we should get definitely right before releasing is the keyword arguments, so I think that it'll be good to converge on this before merging this. The reason why I don't like the current behaviour of ``load_SI_image_stack`` is that it always returns you a ``Signal2D`` i.e. the user cannot choose not to read any image from the stack. Also it seems rude to me that after telling to ``load`` not to load an image stack it actually goes and load the first image of the stack. I would rather cope with waiting a couple of seconds (or nothing if lazy) than to let ``load`` behave this way ;)

ericpre:
Yes, the original keywords are not great, indeed... There is also the `select_type` keyword, so an alternative could be:

- `select_type=None, load_SI_image=False`: load everything, which means all maps, first image of the SI images stack and spectrum image.
- `select_type=None, load_SI_image=True`: same as above but load the SI images stack.
- `select_type='image', load_SI_image=False`: same as first but without spectrum images.
- etc.

francisco-dlp:
I've now fixed the decorator. I propose to merge this one as is and leave the keyword argument modification for another PR as the issue is not related to this PR.

PR #1854, opened 2018-03-04
add: Brukers spx reader

sem-geologist:
### Description of the change
This PR adds to bcf plugin the Bruker's spx (single spectra) file reading capabilities:

### Progress of the PR
- [x] initial spx reader,
- [x] add bells and whistles of spx, (element selection->metadata, quanti results to original_metadata)
- [x] update user guide,
- [x] rename the bcf.py into `bruker.py` and change references to it
- [x] add tests,
- [x] ready for review.

### Things to keep back?
~~which could be changed at `next_major`?:
rename `bcf.py` into `bruker.py` -- would it brake some production scripts?~~


ericpre:
The io plugins are not in the public API, so I would say that it could go to next minor.

sem-geologist:
@ericpre , @francisco-dlp , this PR can be reviewed

hyperspy/tests/io/test_bruker.py
```diff
@@ -205,3 +207,16 @@ def test_decimal_regex():
...
         assert b'.' not in fix_dec_patterns.sub(b'\\1.\\2', j)
+
+def test_all_spx_loads():
+    for spxfile in spx_files:
+        filename = os.path.join(my_path, 'bruker_data', spxfile)
+        s = load(filename)
+        assert s.data.dtype == np.uint64
+        assert s.metadata.Signal.signal_type == 'EDS_SEM'
+
+def test_stand_alone_spx():
```
ericpre, line 218: Check the value of some more metadata? tilt alpha, beam energy, data, time, etc.

hyperspy/io_plugins/bruker.py
```diff
@@ -967,6 +953,59 @@ def add_filename_to_general(self, item):
...
+                 mode: {'Detector':
+                            gen_detector_node(spectrum),
+                         'beam_energy': spectrum.hv}
+               },
+                'General': {'original_filename': filename.split('/')[-1],
+                            'title': 'EDX',
+                            'date': spectrum.date,
+                             'time': spectrum.time},
+                 'Sample': {'name': name,
+                            #'elements': TODO
```
ericpre: Some left over I guess?

ericpre (changes requested):
This looks good, just one thing needed and two minor comments:
- the tests data folder needs to be update in https://github.com/hyperspy/hyperspy/blob/RELEASE_next_minor/setup.py#L378

hyperspy/tests/io/test_bruker.py
sem-geologist: there is no need to do that as it is already extensively tested with bcf readers (which uses exactly the same code for metadata loading/transfering into hyperspy metadata dict from xml). this test tests only features which are always present in spx but not in bcf (such as dead/live times...).

hyperspy/io_plugins/bruker.py
sem-geologist: yeah, forgot to remove, the code below reads elements into dictionary if the xml node is present (if selection was made in Esprit, else there is no node).

ericpre (approved):

PR #1855, opened 2018-03-05
New lazy close file

francisco-dlp:
Currently when loading a file lazily the file remains open, usually until the signal is deleted. However, in e.g. Windows, it seems that the file is not released even after the signal is deleted, causing issues with e.g. removing test files from our unittests. This PR adds a ``LazySignal.close_file`` method and a ``LazySignal.compute`` ``close_file`` argument to close the file that works only when the file is a hdf5 file.

This is already included in #1853 in order to make the appveyor tests pass, but I am submitting it separately for clarity.

### Progress of the PR
- [x] add tests,
- [x] ready for review.

### DIFF
--- a/hyperspy/_signals/lazy.py
+++ b/hyperspy/_signals/lazy.py
@@ -84,16 +84,45 @@ class LazySignal(BaseSignal):
     """
     _lazy = True
 
-    def compute(self, progressbar=True):
-        """Attempt to store the full signal in memory.."""
+    def compute(self, progressbar=True, close_file=False):
+        """Attempt to store the full signal in memory.
+
+        close_file: bool
+            If True, attemp to close the file associated with the dask
+            array data if any. Note that closing the file will make all other
+            associated lazy signals inoperative.
+
+        """
         if progressbar:
             cm = ProgressBar
         else:
             cm = dummy_context_manager
         with cm():
-            self.data = self.data.compute()
+            da = self.data
+            data = da.compute()
+            if close_file:
+                self.close_file()
+            self.data = data
         self._lazy = False
         self._assign_subclass()
+
+    def close_file(self):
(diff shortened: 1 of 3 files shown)


hyperspy/tests/io/test_hdf5.py
```diff
@@ -355,12 +355,13 @@ def test_oom_loading(self):
         assert self.shape == s.data.shape
         assert isinstance(s.data, da.Array)
         assert s._lazy
+        s.close_file()
 
     def teardown_method(self, method):
         gc.collect()        # Make sure any memmaps are closed first!
         try:
```
ericpre, line 362: What is the reason to let the tests pass if we can't remove the file? If an error is raised, it is most likely because something is wrong and we need to fix it?

hyperspy/_signals/lazy.py
```diff
@@ -84,17 +84,46 @@ class LazySignal(BaseSignal):
...
+
+        """
         if progressbar:
             cm = ProgressBar
         else:
             cm = dummy_context_manager
         with cm():
-            self.data = self.data.compute()
+            da = self.data
+            data = da.compute()
```
ericpre, line 102: Is there a specific reason to write this in 2 lines and not just `data = self.data.compute()`? If so, add an explanation in comment.

ericpre (commented):
I am not convinced that the test cover the error this PR is supposed to fix. I can't reproduce the error seen in appveyor in #1853 by playing with the tests of this PR. If I understand correctly, the error in #1853 happen because the file is not released/closed and therefore it can't be deleted. But if I don't close the file in `TestLoadingOOMReadOnly`, the file can be still removed and the `teardown` goes fine...

ericpre:
Done in #1853.

Issue #1786, opened 2017-12-11
Power law background removal bug?


mwalls:
When running the demo by Magnus about EELS quantification. the one that is here:
https://github.com/hyperspy/hyperspy-demos/blob/master/electron_microscopy/EELS/EELS_analysis.ipynb

the data can sometimes disappear partially as indicated in the figures below.
![quantify_problem_signal](https://user-images.githubusercontent.com/1022064/33839347-39b4a15c-de92-11e7-9433-5bf5c8c380b5.png)
![quantify_problem_nav](https://user-images.githubusercontent.com/1022064/33839348-39cba5be-de92-11e7-836b-126449dd7e0a.png)

The problem seems only to occur when not using the 'fast' option. it is not 100% reproducible. Different parts of the data disappear in different tests, possibly as a function of the exact fitting window.

mwalls:
Here is another phenomenon that can occur. In this case the window went from 438 to 448 eV. The last spectrum has some kind of oscillating artefact signal
![quantify_problem_signal3](https://user-images.githubusercontent.com/1022064/33840422-2f3b1dfc-de95-11e7-9529-c739fdfef1dd.png)
![quantify_problem_nav3](https://user-images.githubusercontent.com/1022064/33840423-32019822-de95-11e7-847e-8010b8cd0c4d.png)

 added:

mwalls:
It is also almost completely negative...

mwalls:
In fact the background subtraction seems not to be reproducible. If I run the following code:
`for x in range(1,10):
    sbr=s.remove_background(signal_range=(437,449.0),fast=False)
    sbr.plot()`

I get 10 different backgound removals. Here are 3 of the images:
![figure_eels_spectrum_image_ high-loss _navigator-3](https://user-images.githubusercontent.com/1022064/33886885-44bcb4c0-df48-11e7-9938-1488b323b59b.png)
![figure_eels_spectrum_image_ high-loss _navigator-2](https://user-images.githubusercontent.com/1022064/33886889-46c486a8-df48-11e7-8746-611b3a98e691.png)
![figure_eels_spectrum_image_ high-loss _navigator-1](https://user-images.githubusercontent.com/1022064/33886894-48a10f1e-df48-11e7-9f8a-940dc473e321.png)

mwalls:
Another attempt to show the code:
`for x in range(1,10):
    sbr=s.remove_background(signal_range=(437,449.0),fast=False)
    sbr.plot()`

[...]

Issue #1728, opened 2017-09-10
s.remove_background(fast=False) bad fitting


magnunor:
Sometimes when using `s.remove_background(fast=False)` the fitting leads to several NaN values in the data.

```Python
import numpy as np        
import hyperspy.api as hs 

x = np.arange(370, 570)   
powerlaw = hs.model.components1D.PowerLaw(A=1000000000, r=2, origin=300)                                 

data_list = []            
for i in range(20):       
    data_list.append(powerlaw.function(x) + np.random.random(size=len(x))*10000)                         

data = np.array(data_list)                          
s = hs.signals.Signal1D(data_list)                  
s.axes_manager[-1].offset = 370                     

for i in range(100):      
    s1 = s.remove_background(signal_range=(379., 446.), fast=False)                                      
    if not np.invert(np.isnan(s1.data)).all():      
        print("Bad fit")  
        break
```

One example of this: 
![remove_background_bug](https://user-images.githubusercontent.com/1690979/30250658-4b10a526-964a-11e7-8847-ffbc0b59db77.png)

It only happens occasionally, and if it happens the "bad" region is at different navigation positions.

It does not happen when using `fast=True`, which means something is going wrong here: https://github.com/hyperspy/hyperspy/blob/RELEASE_next_minor/hyperspy/_signals/signal1d.py#L1013

Also happens for "real" datasets, for example early in the [EELS notebook](https://github.com/hyperspy/hyperspy-demos/blob/master/electron_microscopy/EELS/EELS_analysis.ipynb).

ericpre:
It look to me that this issue has been fixed by #1860?

magnunor:
Maybe, I'll check if the bug still appears.

magnunor:
I ran the above script 1000 times, and bug not not occur even once. So this seems to be fixed!

PR #1860, opened 2018-03-09
fix problem with EELS background subraction

mwalls:
There is currently a problem when fast =False. The signal range has to be put back to the full range to create the model across the whole range.

### DIFF
--- a/hyperspy/_signals/signal1d.py
+++ b/hyperspy/_signals/signal1d.py
@@ -1027,6 +1027,7 @@ _spikes_diagnosis,
         if not fast:
             model.set_signal_range(signal_range[0], signal_range[1])
             model.multifit(show_progressbar=show_progressbar)
+            model.reset_signal_range()
         return self - model.as_signal(show_progressbar=show_progressbar)
 
     def remove_background(


thomasaarholt:
Could you include a minimum working example? It isn't entirely clear to me when fast=False is a problem.

mwalls:
I gave some in issue #1786 a few months ago. The background to subtract was sometimes not extrapolated beyond the fitting region when fast = False

mwalls:
Magnus had also come across it before. I noticed it first when trying to run his EELS analysis tutorial from the site. For a while we thought the fitting routine was starting from crazy values of A and r, but it turns out it's the subtraction of the background that is not working properly. I think it's that when you create the signal to be subtracted using `model.as_signal(..)` (line 1030 in the original code) you should have set all the "channel switches" to True (reset_signal_range seems to do this) . Otherwise the model is only converted to a signal for those which are true to begin with, and those are just the ones that are included in `set_signal_range`. Why the current code actually works randomly for some lines in the line-spectrum is still a mystery to me. 
Hope this is a bit clearer. Thanks for any comments!

magnunor:
Yeah, I often get this during workshops, and it is really annoying. 

See https://github.com/hyperspy/hyperspy/issues/1728 for a way of replicating it.

thomasaarholt:
Looking at the file change you've done, I realise the problem.

Maybe a better solution would be to get the signal range the model currently has, and then set the signal range to those values after fitting, instead of resetting it?

**Edit:**
Looking at the [method](https://github.com/hyperspy/hyperspy/blob/9e939d848da387c0415d8543fa5966a6a36268f2/hyperspy/models/model1d.py#L461), it would take a bit more effort to implement.

There exists no record of the signal_range previously set, so even if one created a `m.get_signal_range()`, there's currently nowhere to get it from.

Two options I see:
(A): leave it as it is. 
(B): create a property `m._signal_range` which holds the current signal range (default `(None, None)`), then update `m.reset_signal_range()` and `m.set_signal_range` to respect it. Finally create a `m.get_signal_range()`, and change PR to:

```python
if not fast:
    current_signal_range = model.get_signal_range()
    model.set_signal_range(signal_range[0], signal_range[1])
    model.multifit(show_progressbar=show_progressbar)
    model.set_signal_range(current_signal_range)
```

mwalls:
Not sure I follow you. My understanding is that line 1028 sets the range over which to fit the background. This is of course a limited range before the edge(s). So far so good. The problem is exactly that the signal range used to make the new signal to be subtracted is then only that short range. The rest of the new signal created by the command 'model.as_signal' is filled with nans, not the extrapolated part of the fitted power law function, which is what we really want.  We need all the channels to be included in the new signal, not just the fitted region ones?

thomasaarholt:
Okay, then I misunderstood the issue.

francisco-dlp:
LGTM, thanks!

francisco-dlp:
Merged manually to Rnp because it branched out from Rnm.

PR #1863, opened 2018-03-15
New component for models based on the skew normal distribution

jlaehne:
### Description of the change
Added a new component based on the skew normal distribution, which is related to the normal distribution (Gaussian) and can be used to fit asymmetric peaks. For definition see https://en.wikipedia.org/wiki/Skew_normal_distribution. Component includes the derivatives of the function. Tested by fitting a spectral map with a model including a background Gaussian + a Skew Normal peak.

- new file _components/skew_normal.py was created
- components1d.py was changed to include the new component

### Progress of the PR
- [x] added test function,
- [x] ready for review.

```python
>>> import hyperspy.api as hs
>>> import numpy as np
>>> m = s.create_model()
>>> g = hs.model.components1D.SkewNormal()
>>> m.append(g)
>>> m.fit()
```


francisco-dlp:
Would it be possible to write it as a ``Expression`` for the reasons discussed in #1864?

jlaehne:
As the function definition includes the error function (scipy.erf), I
did not see a way to write it as an expression and used the Error
Function component as a starting point. Concerning performance, I was
positively surprised that the fits converge rather fast (e.g. in the
case of asymmetric peaks much faster than for a Gaussian component).

tjof2:
I don't know much about this distribution beyond having just read the wiki link, but what advantages/disadvantages are there compared to any of the following - all of which seem a little more widely-used than the skew-normal:

- [Log-normal distribution](https://en.wikipedia.org/wiki/Log-normal_distribution)
- [(Generalized) gamma distribution](https://en.wikipedia.org/wiki/Gamma_distribution)
- [Beta distribution](https://en.wikipedia.org/wiki/Beta_distribution)
- [Chi-squared distribution](https://en.wikipedia.org/wiki/Chi-squared_distribution)

In particular, the top two.

Is it just for convenience - i.e. you've seen some asymmetric peaks and this just so happens to fit?

jlaehne:
@tjof2: An important difference to the distributions you mention is that a negative shape parameter directly mirrors the distribution along the vertical axis, which adds flexibility. Also it is derived from the normal distribution - so it fits very well for peaks that resemble a Gaussian, but have a certain asymmetry, which is quite common in spectroscopic data. So I would say that it is actually not mere convenience that it fits better than the other asymmetric distributions and could be quite useful for a variety of applications. Still it is true that it remains a more phenomenological approach than having a real physical background. 

To give a specific example, I implemented the function for the case of (cathodo)luminescence spectra where I had to disentangle a broad, asymmetric substrate luminescence from superimposed (Gaussian) luminescence peaks. In this sense, it could for example be also a quite helpful function for background removal procedures. I was pointed to this specific distribution by a senior colleague who has regularly used this function for spectral data as well.

An interesting read on top of the wiki article is the following introduction and other pages on that website: http://azzalini.stat.unipd.it/SN/Intro/intro.html

francisco-dlp:
This looks good except for my comment above: it would be good to rewrite it using the ``Expression`` component. The error functions does not seem to be an issue:

```python

>>> import hyperspy.api as hs
>>> import numpy as np
>>> err = hs.model.components1D.Expression("erf(x)", name="erf")
>>> err.function(np.arange(10))
array([0.        , 0.84270079, 0.99532227, 0.99997791, 0.99999998,
       1.        , 1.        , 1.        , 1.        , 1.        ])
```

francisco-dlp:
For an example on how to do it you can check the ``GaussianHF`` code.

ericpre:
@jlaehne: I have recently rewrote the `Gaussian` component to use `Expression` in #2032, you can see in commit https://github.com/hyperspy/hyperspy/pull/2032/commits/fe3e8e46408a21d8182dfd73e5b2eabf81066959 how much simpler it is using `Expression`! The speed improvement using numexpr is not negligible too...

jlaehne:
@francisco-dlp: When I try your minimal example for the expression from Aug 24, I get an error!? I have also tried to adapt the example to the gaussian function, same error:


```
<ipython-input-11-d32db80aab20> in <module>
      3 err = hs.model.components1D.Expression("erf(x)", name="erf")
      4 gauss = hs.model.components1D.Expression("gaussianhf(x)", name="gaussianhf")
----> 5 err.function(np.arange(10))

~/xtended/software/hyperspy/hyperspy/_components/expression.py in f(x)
    220                 x, y, *[p.value for p in self.parameters])
    221         else:
--> 222             def f(x): return self._f(x, *[p.value for p in self.parameters])
    223         setattr(self, "function", f)
    224         parnames = [symbol.name for symbol in parameters]

<lambdifygenerated-18> in _lambdifygenerated(x)
      1 def _lambdifygenerated(x):
----> 2     return (erf(x))

NameError: name 'erf' is not defined
```

jlaehne:
@francisco-dlp : You seem to have a numpy derivate that includes the error function: https://github.com/numpy/numpy/issues/12515

It is not contained in the standard numpy package so we should probably not write components that rely on it. Instead the standard implementation of `erf` resides in `scipy.special` (as used also for the error function component). Is there any way to use this function in the definition of an expression?

ericpre:
Since sympy 1.3, the scipy module is supported in sympy -see https://github.com/sympy/sympy/wiki/Release-Notes-for-1.3 and https://github.com/sympy/sympy/pull/14992- and the following should work fine:

```python
import hyperspy.api as hs
import numpy as np
err = hs.model.components1D.Expression("erf(x)", name="erf", module="scipy")
err.function(np.arange(10))
```

jlaehne:
Thanks @ericpre. With the dependency of sympy>=v1.3 this component is now defined as expression and works! Test functions are implemented, and I successfully tested it on a quite complex fitting procedure where this component was used for background removal, so should be finally ready.

hyperspy/_components/skew_normal.py
```diff
@@ -0,0 +1,108 @@
...
+    .. math::
+
+        f(x) &= 2 A \phi(x) \Phi(x) \\
+        \phi(x) &= \frac{1}{\sqrt{2\pi}}\mathrm{exp}{\left[-\frac{t(x)^2}{2}\right]} \\
+        \Phi(x) &= \frac{1}{2}\left[1 + \mathrm{erf}\left(\frac{\alpha~t(x)}{\sqrt{2}}\right)\right] \\
+        t(x) &= \frac{x-x_0}{\omega}
+    
+    
+    ============== =============
+    Variable        Parameter
```
ericpre: It would be better to stick to the already existing internal convention for the table by using parameter and attribute for now and leave the table formatting changes to #2148.

hyperspy/_components/skew_normal.py
```diff
@@ -0,0 +1,108 @@
...
+    ============== =============
+    :math:`A`       A
+    :math:`x_0`     x0
+    :math:`\omega`  scale
+    :math:`\alpha`  shape
+    ============== =============
+    
+    Parameters
+    -----------
+        A : float
```
ericpre: Fix indentation:
```
    Parameters
    ----------
    A : float
        Height parameter of the peak.
    x0 : float
        ....
```

hyperspy/_components/skew_normal.py
```diff
@@ -0,0 +1,108 @@
...
+            Height parameter of the peak.
+        x0 : float
+            Location of the peak position (not maximum, which is given by the `mode` property).
```
ericpre: Please keep the length line shorter than 79 characters when possible.
https://www.python.org/dev/peps/pep-0008/#maximum-line-length

hyperspy/_components/skew_normal.py
```diff
@@ -0,0 +1,108 @@
...
+            Width (sigma) parameter.
+        shape: float 
+            Skewness (asymmetry) parameter. For shape=0, the normal distribution (Gaussian) is obtained. The distribution is right skewed (longer tail to the right) if shape>0 and is left skewed if shape<0.
+            
+    The properties `mean` (position), `variance`, `skewness` and `mode` (=position of maximum) are defined for convenience.
+    """
+
+    def __init__(self, x0=0., A=1., scale=1., shape=0., module="scipy",
+                 **kwargs):
+        super(SkewNormal, self).__init__(
```
ericpre, line 83: Add the following to make sure that sympy >= 1.3 and raise an understandable message to the user:

```python
        if LooseVersion(sympy.__version__) < LooseVersion("1.3"):
            raise ImportError("The `SkewNormal component requires SymPy >= ",
                              "1.3")
```

ericpre (commented):
Just a couple of comments, mainly about formatting.

hyperspy/_components/skew_normal.py
jlaehne: done

hyperspy/_components/skew_normal.py
jlaehne: done

hyperspy/_components/skew_normal.py
jlaehne: done

hyperspy/_components/skew_normal.py
jlaehne: done

Issue #1864, opened 2018-03-16
model.as_signal() does not work for a model including GaussianHF component


jlaehne:
If working on an object that includes multidimensional data, the model.as_signal() method fails when the model includes a component of GaussianHF component. It works, when this is replaced by Gaussian.
For a single spectrum (navigation dimensions 0), the method works also for GaussianHF.

The reasons is that the component GaussianHF is defined as an expression in the __init__ statement of the class and not as a separate function (defined through expression instead of component class of hyperspy).

I am working on a redefinition of GaussianHF as component.

** Code Snippet:
>>> model = S.create_model()
>>> g = hs.model.components1D.GaussianHF()
>>> model.append(g)
>>> model.multifit
>>> model.as_signal()

** Error message
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
<ipython-input-27-c32c84cbe301> in <module>()
----> 1 A = model.as_signal()

~/medien/pdi/git/hyperspy/hyperspy/model.py in as_signal(self, component_list, out_of_range_to_nan, show_progressbar, out, parallel)
    490             for sp, csm in zip(splits, np.cumsum(splits)):
    491                 slices[ind] = slice(csm - sp, csm)
--> 492                 models.append(self.inav[tuple(slices)])
    493                 array_slices = self.signal._get_array_slices(tuple(slices),
    494                                                              True)

~/medien/pdi/git/hyperspy/hyperspy/model.py in __getitem__(self, slices)
   1814                 if 'init' in parse_flag_string(flags_str):
   1815                     init_args[k] = value
-> 1816             _model.append(getattr(components, comp._id_name)(**init_args))
   1817         copy_slice_from_whitelist(self.model,
   1818                                   _model,

TypeError: __init__() got an unexpected keyword argument 'name'

francisco-dlp:
Good catch!

Regarding the proposed fix #1868, I think that it is more of a workaround than a fix. The issue is that the components created by ``Expression`` should work like any other components and, in this case, it doesn't. Writing a component as a expression has performance advantages and also makes the code more compact and readable, hence easier to maintain.

ericpre:
Fixed in #1909.

PR #1865, opened 2018-03-16
Issue #1864: Changed definition of GaussianHF from Expression to Component

jlaehne:
### Description of the change
gaussianhf.py was changed to
- a definition of the function through component function instead of expression
- include gradients

As described in issue #1864, model.as_signal() did not work with GaussianHF. This is fixed through the change in how the function is defined (not as an expression any longer).

### Progress of the PR
- [X ] Change implemented (can be split into several points),
- [not needed] update docstring (if appropriate),
- [not needed] update user guide (if appropriate),
- [ ] add tests,
- [ ] ready for review.

```python
>>> import hyperspy.api as hs
>>> import numpy as np
>>> model = s.create_model()
>>> g1 = hs.model.components1D.GaussianHF()
>>> model.append(g1)
>>> model.multifit()
```

### DIFF
--- a/hyperspy/_components/gaussianhf.py
+++ b/hyperspy/_components/gaussianhf.py
@@ -17,15 +17,17 @@
 # along with  HyperSpy.  If not, see <http://www.gnu.org/licenses/>.
 
 import math
+import numpy as np
 
-from hyperspy._components.expression import Expression
+from hyperspy.component import Component
+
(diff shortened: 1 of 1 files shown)


ericpre:
As mentioned in #1864 the fix in #1909 is better.

ericpre:
Thanks for the spotting the bug and making a PR anyway! ;)

Issue #1866, opened 2018-03-16
Left arrow resets view when navigating SI plots


thomasaarholt:
When zoomed in on an area in a plot, it is nice to be able to move forwards and around to compare neighbouring pixels using the arrow keys. When moving backwards (using the left arrow) however, this _also_ triggers the original behaviour of the left arrow in matplotlib displays, namely to reset the view to the previous view.

It would be nice to fully overwrite this so that left arrow only moves to the left, and does not reset the view.

francisco-dlp:
A workaround for this is to use numbers (for convenience the ones in the numeric pad instead). However, it'll be nice to provide a real solution for this bug.

thomasaarholt:
A possible solution is here: https://stackoverflow.com/questions/35624183/disable-matplotlibs-default-arrow-key-bindings. We should probably remove the forward (right arrow) behaviour as well as it acts as the opposite of the left arrow.

The suggested solution is rather brute force, but can be modified to be for a per axes basis:
```python
import matplotlib as mpl
mpl.rcParams['keymap.back'].remove('left')
mpl.rcParams['keymap.forward'].remove('right')
```
Edit: I've checked and the above does solve the problem.

thomasaarholt:
Since lots of people use hyperspy on their laptop, I'd suggest adding *additional* support for the numpad.

francisco-dlp:
Changing the default mpl keymap for the user's session after importing hyperspy sounds indeed like a brute force approach. What about doing it only when registering hspy keymap (i.e. on creating the multi dimensional signal figure) and restoring them when closing the figure?

thomasaarholt:
Yes, that sounds good. I found people doing the following for temporarily setting the grid on, but I'm not sure how to remove things temporarily from a list:
```python
with plt.rc_context({"axes.grid": True, "grid.linewidth": 0.75}):
    f, ax = plt.subplots()  # Will make a figure with a grid
    ax.plot(x, y)
```
Could we do something along the lines of
```python
with plt.rc_context({"keymap.back": [],}):
   plot_stuff()
```

Edit: Nope, the above does not work.

francisco-dlp:
We can't use a context manager here because the user starts interacting with the plot after the plot is finished.

I am wondering if it wouldn't be better to simply change hspy's keymap to ``Ctrl + Arrow``. 

[...]

PR #1868, opened 2018-03-17
Add support for hotkeys in higher navigation dimensions

thomasaarholt:
### Description of the change
Before this PR, for higher dimension (>2) navigators, the hotkeys (left, right, 4, 6 etc...) did not work.
The PR adds additional support for moving in X and Y with **Ctrl** + arrow keys / 2468 to avoid the error reported in #1866.

Also, `Shift + (arrows or 2468)` now controls dimensions `(2, 3)` and` Ctrl + Alt + (arrows or 2468)` now controls dimensions `(4, 5)` for a total of hotkey control of six dimensions of navigation space.

Note: I originally went for Ctrl + Shift to control 2,3, but Ctrl + Shift + Numpad keys does not work on Windows in matplotlib (for some unknown reason).

If people are happy with these changes I will update the user guide.

*Did you know? You can use PageUp/PageDown to increase the number of steps taken by the hotkeys!*

### Progress of the PR
- [ ] update user guide
- [ ] ready for review.

### DIFF
--- a/hyperspy/axes.py
+++ b/hyperspy/axes.py
@@ -931,9 +931,11 @@ class AxesManager(t.HasTraits):
             return
         x = self.navigation_axes[0]
         try:
-            if event.key == "right" or event.key == "ctrl+right" or event.key == "6" or event.key == "ctrl+6":
+            if event.key == "right" or event.key == "ctrl+right" or \
+                    event.key == "6" or event.key == "ctrl+6":
                 x.index += self._step
-            elif event.key == "left" or event.key == "ctrl+left" or event.key == "4" or event.key == "ctrl+4":
+            elif event.key == "left" or event.key == "ctrl+left" or \
+                    event.key == "4" or event.key == "ctrl+4":
                 x.index -= self._step
             elif event.key == "pageup":
                 self._step += 1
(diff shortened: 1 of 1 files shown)


ericpre:
It needs unit tests.

@thomasaarholt, please try to commit in a sensible way to keep the history readable and useful. For example, there are quite a few of your commits with just one line of comments or one very small change. Numerous open source projects squash the PR commits into a single one before merging to make the history easily readable. HyperSpy doesn't do that, but it would be good that we keep the history as tidy as possible.

thomasaarholt:
Thanks for the feedback. I fully agree that I should be way better at that. I've been taught to commit often, but this is perhaps too often, and without good reason. 

Do you have any suggestions for a sensible guide on when to commit?

I feel a bit like I'm torn between committing for my own purposes and for the general public.

Could it be smart to commit often, then tidy up the commits before pushing? What if I'm working on multiple computers (which I am)?

thomasaarholt:
Since I realised there was a much better way of doing this (using a dictionary of `event.key : (axes, direction)`), I'm closing this PR and writing a new one, taking into account Eric's comments above.

In case this gets dug up later: There are issues with using the NumPad with modifier keys (shift, etc.). For instance, Shift + 4 becomes `left arrow` on my work keyboard. For now, I'm sticking with arrow keys, and people smarter than I am can implement the NumPad.

ericpre:
Obviously, it is a bit subjective, depends on the people and other things.
I could tell you what I try to do myself:

- I try to commit things which are in working state including tests, something complete or at least "self-containing". The idea is that you read the changes, you can understand the reasoning, very useful for review... Also, it is very good to write the tests as the same time that you implement a new feature or fixes a bug. In fact, this is very convenient to run the tests to check the actual changes, even if it looks tedious at the beginning. This works particularly well for small changes.
- I will do the doc at the end, when I have clear idea of how it will be.
- Quite often, if I notice that I forgot something, I amend the last commit instead of creating a new one.
- I try to finish what I have started not to waste time, so I avoid to have to stop in a middle of a task. If I can see that I will not have time then I do something else...
- to work between branches, you can stash your changes.

To my opinion working on different computers is a very annoying, I don't do it. Particularly that nowadays laptop are very powerful, still a bit pricey though... If I need to use another computer, I connect remotely to it.
I don't think that the git history has been designed to be use as backup for yourself or to share file between computers, this is significantly different from dropbox.
Rewriting the git history is not very convenient and it can also mess up the review comments on github if they are several iterations. I think this is a better practise and more simple to do it right from the beginning.

ericpre:
Done in #1870.

PR #1870, opened 2018-03-19
Add support for hotkeys in higher navigation dimensions

thomasaarholt:
### Description of the change
Before this PR, for higher dimension (>2) navigators, the hotkeys (left, right etc...) did not work.
The PR adds additional support for moving in X and Y with **Ctrl** + arrow keys to avoid the error reported in #1866.

Also by default, `Shift + arrows` now controls dimensions `(2, 3)` and` Ctrl + Alt + arrows` now controls dimensions `(4, 5)` for a total of hotkey control of six dimensions of navigation space.

Finally, all of these changes are now customisable in the preferences gui.

Note: The modifier functionality has been added to the arrow keys, but not to the numpad numbers. Shift + Numpad keys and Alt + Numpad keys does not work on Windows in matplotlib (for some unknown reason).

*Did you know? You can use PageUp/PageDown to increase the number of steps taken by the hotkeys!*

This PR replaces #1868, both due to poor commit history and realising that there was a better way to code this.

### Progress of the PR
- [x] Added test, - Tests need updating to reflect the correct hotkeys
- [x] update user guide - still need to take francisco's suggestions into account.
- [x] ready for review.


thomasaarholt:
@ericpre I would appreciate feedback on the test, in particular. I will try to be better about clearer commits in the future.

thomasaarholt:
I've added documentation. This is ready for review.

If desired, I can change the current `**dict.fromkeys` style to a regular dict style. It was helpful when I had many key combinations with the same value, but since I had to remove the `modifier + keypad` combinations it isn't as helpful and possibly more confusing.

francisco-dlp:
I think that a more standard way of setting the dict would be nice.

``Ctrl + Alt`` does not work in Linux because it is used by the GNOME. What about ``Shift + Arrows`` and ``Alt + Arrows`` instead?
Also, I think that VIM users may appreciate using ``h`` ``j`` ``k`` ``l`` for navigation too (including modifiers). Adding hese extra shortcuts has the advantage that it is unlikely that any OS will assign ``Shift+`` and ``Alt+`` those keys to anything else.

Finally, what about adding the option to customise the keys (and modifiers) to preferences? I think that this is important because it'll be difficult to find a set of keys and modifiers that works for all platforms (including personal customisations) and locales at present and is future-proof.

hyperspy/tests/axes/test_axes_manager.py
```diff
@@ -263,3 +263,28 @@ def test_setting_indices_coordinates():
     s.axes_manager.indices = (2, 2)
     assert s.axes_manager.indices == (2, 2)
     assert m.call_count == 6
+
+class TestAxesHotkeys:
+
+    def setup_method(self, method):
+        s = Signal1D(zeros(7 * (5,)))
```
francisco-dlp, line 272: ``np.zeros`` after importing numpy using ``import numpy as np``?

doc/user_guide/visualisation.rst
```diff
@@ -50,12 +50,21 @@ the x-axis if 1D:
...
-on and the spectrum or navigator figure is selected**.When using the numpad
+on and the spectrum or navigator figure is selected**. When using the numpad
 arrows the PageUp and PageDown keys change the size of the step.
 
 An extra cursor can be added by pressing the ``e`` key. Pressing ``e`` once
 more will disable the extra cursor:
 
+In matplotlib, left and right arrow keys are by default set to navigate the 
+"zoom" history. To avoid the problem of changing zoom while navigating, 
+Ctrl + arrows can be used instead.
```
francisco-dlp: You could also add that we'll deprecate arrow keys (without modifiers) in 2.0.

doc/user_guide/visualisation.rst
```diff
@@ -50,12 +50,21 @@ the x-axis if 1D:
 To change the current coordinates, click on the pointer (which will be a line
 or a square depending on the dimensions of the data) and drag it around. It is
 also possible to move the pointer by using the numpad arrows **when numlock is
-on and the spectrum or navigator figure is selected**.When using the numpad
+on and the spectrum or navigator figure is selected**. When using the numpad
```
francisco-dlp: When using the *keyboard*?

doc/user_guide/visualisation.rst
```diff
@@ -50,12 +50,21 @@ the x-axis if 1D:
...
 arrows the PageUp and PageDown keys change the size of the step.
```
francisco-dlp: step size?

doc/user_guide/visualisation.rst
```diff
@@ -50,12 +50,21 @@ the x-axis if 1D:
...
 
 An extra cursor can be added by pressing the ``e`` key. Pressing ``e`` once
 more will disable the extra cursor:
 
+In matplotlib, left and right arrow keys are by default set to navigate the 
+"zoom" history. To avoid the problem of changing zoom while navigating, 
+Ctrl + arrows can be used instead. 
+
+To navigate navigation dimensions larger than 2, modifier keys can be used.
+Ctrl + Shift + left/right, Ctrl + Shift + up/down, Ctrl + Alt + left/right
```
francisco-dlp: Ctrl + Shift + left/right and Ctrl + Shift + up/down (Ctrl + Alt + left/right and Ctrl + Alt + up/down) navigates dimensions 2 and 3 (4 and 5) respectively?

doc/user_guide/visualisation.rst
```diff
@@ -69,17 +78,20 @@ can be too small to be dragged or even seen. It
...
-PageDown    Decrease step size
-``+``           Increase pointer size
-``-``           Decrease pointer size
-``h``       Launch the contrast adjustment tool (only for Signal2D)
-=========   =============================
+=======================   =============================
+key                       function
+=======================   =============================
+e                         Switch second pointer on/off
+Arrows                    Change coordinates for dimensions 0 and 1 (typically x and y)
```
francisco-dlp: If we're going to deprecate them, we could remove them from this table.

doc/user_guide/visualisation.rst
```diff
@@ -69,17 +78,20 @@ can be too small to be dragged or even seen. It
...
+Ctrl + Arrows             As above without also changing zoom
```
francisco-dlp: And remove the mention to the zoom here.

doc/user_guide/visualisation.rst
thomasaarholt: Yes, I like that better.

thomasaarholt:
Okay, I managed add basic preferences support and rebase the commits. @francisco-dlp In order to display text fields in the preferences GUI, I am having to use` t.Directory` rather than `t.CStr` or `t.String,` since the latters fail. 

I have tried adding a new tab in the preferences window for "hotkeys", but can't figure it out. I attempted to duplicate "GUIs" into a "GUIs2", but after making duplicates of all the entries where I find "GUIs", it still does not appear when I call `hs.preferences.gui()`. Ideas?

francisco-dlp:
https://github.com/thomasaarholt/hyperspy/pull/11

francisco-dlp:
I've moved this to the 1.4 milestone because it requires a hspy_guis releases that we won't have time to make before releasing v1.3.1.

thomasaarholt:
Great, thanks Fransisco. 1.4 milestone is fine.

thomasaarholt:
That worked great, @francisco-dlp, and looks much more sensible. I may add some more possibilities to the Enums.

I see now that the ipywidgets gui needs a definition of what the String should be represented as. I'll take a look at it now.

hyperspy/tests/axes/test_axes_manager.py
thomasaarholt: Added import numpy at top of test.

doc/user_guide/visualisation.rst
thomasaarholt: Done.

thomasaarholt:
@francisco-dlp I think this is actually ready now. I've modified the test to take into account the preferences-based modifier/hotkeys, updated the docs, and I've submitted hyperspy/hyperspy_gui_ipywidgets#3 and hyperspy/hyperspy_gui_traitsui#10 to support the preferences tab/string-field support.

I'm still happy for this to be delayed until 1.4, unless you think it's ready for 1.3.1.

PR #1873, opened 2018-03-20
Fix Real-Time Vs Live_Time Error in EDX Quantifcation

k8macarthur:


k8macarthur:
@francisco-dlp Just spotted this bug whilst I was sorting my notebooks for next week. Would be helpful to include if you haven't already done the build for next week?

ericpre:
@k8macarthur, the test suite fails.

k8macarthur:
Sorry, didn't spend the time checking I got every incidence of it. Should be good now.

k8macarthur:
Essentially the azimuth_angle = None step in the set up for the tem_quantification tests was causing one of the tests to fail when s.microscope_set_parameters was used. Having azimuth_angle=None in the metadata caused it to try and open a GUI.

hyperspy/tests/signal/test_eds_tem.py
```diff
@@ -177,7 +177,6 @@ def test_metadata(self):
         np.testing.assert_approx_equal(
             TEM_md.Detector.EDS.energy_resolution_MnKa, 130)
 
-        self.signal.set_microscope_parameters(real_time=3.1)
```
ericpre: Either this is no longer needed and the implementation is removed from the code or it needs to be tested. I would suggest to keep it.

hyperspy/_signals/eds_sem.py
```diff
@@ -177,11 +177,10 @@ def set_microscope_parameters(self,
                                   elevation_angle=None,
                                   energy_resolution_MnKa=None,
                                   display=True, toolkit=None):
-        if {beam_energy, live_time, tilt_stage, azimuth_angle,
-                elevation_angle, energy_resolution_MnKa} == {None}:
+        if set([beam_energy, live_time, tilt_stage, azimuth_angle,
+                elevation_angle, energy_resolution_MnKa]) == {None}:
             tem_par = EDSSEMParametersUI(self)
             return tem_par.gui(toolkit=toolkit, display=display)
-            return self._set_microscope_parameters_gui()
```
ericpre, line 184: Quite useful to remove this method which doesn't seem to exist anymore.

ericpre (commented):
Please have a look at the comment on the test of the real time. Other than that, this looks good to merge.

k8macarthur:
Agreed, I've added real_time and a check that it's set, same as probe_area. No longer needed for quantification but users should still be able to set it properly. Thanks for the help Eric.

ericpre:
Great thanks!

PR #1878, opened 2018-03-21
Fix lazy decomposition

### LINKED ISSUE #1858: rPCA inconsistency?
Hi all,

recently I may have spotted an inconsistency in Hyperspy's implementation of robust PCA. Subsequently I tried to narrow it down in an idealized test case and keep getting the same strange behavior. So the test case consists of a matrix ```A``` full of ones, except for one element, which is set to two. In my understanding, this should be decomposed into a matrix ```L``` filled with ones and a matrix ```S``` filled with zeros and a one in the element corresponding to the two in ```A```, such that approximately ```L+S = A```. I checked this with another implementation (```r_pca``` from https://github.com/dganguli/robust-pca), which gave the expected result, but Hyperspy disagreed completely. Is this a bug or am I using Hyperspy's rPCA routine incorrectly? I'm using Hyperspy version 1.3. Here's my code:

```python
import numpy as np

import importlib
importlib.import_module('r_pca')
from r_pca import R_pca

import hyperspy.api as hs

n=5
m=10
A = np.ones((n,m))
A[int(n/2),int(m/2)] = 2.0

rpca = R_pca(A)
L, S = rpca.fit(max_iter=1000, iter_print=100)

hsdata = hs.signals.Signal1D(A)
Lhs,Shs = hsdata.decomposition(algorithm='RPCA_GoDec', return_info=True, output_dimension=min(np.shape(A)), power=3)

print('Ganguli-Candes: L= ')
print(L)

print('Hyperspy: L= ')
print(Lhs)

print('Ganguli-Candes: S= ')
print(S)

print('Hyperspy: S= ')
print(Shs)

print('Ganguli-Candes: L+S= ')
print(L+S)

print('Hyperspy: L+S= ')
print(Lhs+Shs)

print('Done.')

import hyperspy
print('Hyperspy version '+str(hyperspy.__version__))
```

And here's the corresponding output:
```
iteration: 1, error: 18.36113064260148
iteration: 8, error: 1.3991493881412786e-06
WARNING:hyperspy.learn.rpca:Threshold 'lambda1' is set to default: 1 / sqrt(nsamples)
WARNING:hyperspy.learn.rpca:Convergence tolerance not specifed. Defaulting to 1e-3
WARNING:hyperspy.learn.rpca:Maximum iterations not specified. Defaulting to 1e3
Ganguli-Candes: L=
[[1.00010184 1.00010184 1.00010184 1.00010184 1.00010184 1.00019043
  1.00010184 1.00010184 1.00010184 1.00010184]
 [1.00010184 1.00010184 1.00010184 1.00010184 1.00010184 1.00019043
  1.00010184 1.00010184 1.00010184 1.00010184]
 [1.00031282 1.00031282 1.00031282 1.00031282 1.00031282 1.00040142
  1.00031282 1.00031282 1.00031282 1.00031282]
 [1.00010184 1.00010184 1.00010184 1.00010184 1.00010184 1.00019043
  1.00010184 1.00010184 1.00010184 1.00010184]
 [1.00010184 1.00010184 1.00010184 1.00010184 1.00010184 1.00019043
  1.00010184 1.00010184 1.00010184 1.00010184]]
Hyperspy: L=
[[1. 1. 1. 1. 1. 1. 1. 1. 1. 1.]
 [1. 1. 1. 1. 1. 1. 1. 1. 1. 1.]
 [1. 1. 1. 1. 1. 2. 1. 1. 1. 1.]
 [1. 1. 1. 1. 1. 1. 1. 1. 1. 1.]
 [1. 1. 1. 1. 1. 1. 1. 1. 1. 1.]]
Ganguli-Candes: S=
[[0.         0.         0.         0.         0.         0.
  0.         0.         0.         0.        ]
 [0.         0.         0.         0.         0.         0.
  0.         0.         0.         0.        ]
 [0.         0.         0.         0.         0.         0.99959858
  0.         0.         0.         0.        ]
 [0.         0.         0.         0.         0.         0.
  0.         0.         0.         0.        ]
 [0.         0.         0.         0.         0.         0.
  0.         0.         0.         0.        ]]
Hyperspy: S=
[[1. 1. 1. 1. 1. 1. 1. 1. 1. 1.]
 [1. 1. 1. 1. 1. 1. 1. 1. 1. 1.]
 [1. 1. 1. 1. 1. 1. 1. 1. 1. 1.]
 [1. 1. 1. 1. 1. 1. 1. 1. 1. 1.]
 [1. 1. 1. 1. 1. 1. 1. 1. 1. 1.]]
Ganguli-Candes: L+S=
[[1.00010184 1.00010184 1.00010184 1.00010184 1.00010184 1.00019043
  1.00010184 1.00010184 1.00010184 1.00010184]
 [1.00010184 1.00010184 1.00010184 1.00010184 1.00010184 1.00019043
  1.00010184 1.00010184 1.00010184 1.00010184]
 [1.00031282 1.00031282 1.00031282 1.00031282 1.00031282 2.
  1.00031282 1.00031282 1.00031282 1.00031282]
 [1.00010184 1.00010184 1.00010184 1.00010184 1.00010184 1.00019043
  1.00010184 1.00010184 1.00010184 1.00010184]
 [1.00010184 1.00010184 1.00010184 1.00010184 1.00010184 1.00019043
  1.00010184 1.00010184 1.00010184 1.00010184]]
Hyperspy: L+S=
[[2. 2. 2. 2. 2. 2. 2. 2. 2. 2.]
 [2. 2. 2. 2. 2. 2. 2. 2. 2. 2.]
 [2. 2. 2. 2. 2. 3. 2. 2. 2. 2.]
 [2. 2. 2. 2. 2. 2. 2. 2. 2. 2.]
 [2. 2. 2. 2. 2. 2. 2. 2. 2. 2.]]
Done.
Hyperspy version 1.3
```

Just out of curiosity I also tried the same with larger dimensions and with the transpose of ```A```, but the result was equivalent, as it should be.

francisco-dlp:
Fixes #1858 

### Progress of the PR
- [ ] add tests,
- [ ] ready for review.

### DIFF
--- a/hyperspy/_signals/lazy.py
+++ b/hyperspy/_signals/lazy.py
@@ -255,9 +255,6 @@ class LazySignal(BaseSignal):
     def __array__(self, dtype=None):
         return self.data.__array__(dtype=dtype)
 
-    def _unfold(self, *args):
-        raise lazyerror
-
     def _make_sure_data_is_contiguous(self, log=None):
         self._make_lazy(rechunk=True)
 
--- a/hyperspy/_signals/lazy.py
+++ b/hyperspy/_signals/lazy.py
@@ -771,14 +771,17 @@ class LazySignal(BaseSignal):
             if reproject:
                 if algorithm == 'PCA':
                     method = obj.transform
-                    post = lambda a: np.concatenate(a, axis=0)
+
+                    def post(a): return np.concatenate(a, axis=0)
                 elif algorithm == 'ORPCA':
                     method = obj.project
                     obj.R = []
-                    post = lambda a: obj.finish()[4]
+
+                    def post(a): return obj.finish()[4]
                 elif algorithm == 'ONMF':
                     method = obj.project
-                    post = lambda a: np.concatenate(a, axis=1).T
+
+                    def post(a): return np.concatenate(a, axis=1).T
 
                 _map = map(lambda thing: method(thing),
                            self._block_iterator(
--- a/hyperspy/_signals/lazy.py
+++ b/hyperspy/_signals/lazy.py
@@ -255,9 +255,6 @@ class LazySignal(BaseSignal):
     def __array__(self, dtype=None):
         return self.data.__array__(dtype=dtype)
 
-    # def _unfold(self, *args):
-    #     raise lazyerror
-
     def _make_sure_data_is_contiguous(self, log=None):
         self._make_lazy(rechunk=True)
 
--- a/hyperspy/_signals/lazy.py
+++ b/hyperspy/_signals/lazy.py
@@ -255,6 +255,9 @@ class LazySignal(BaseSignal):
     def __array__(self, dtype=None):
         return self.data.__array__(dtype=dtype)
 
+    # def _unfold(self, *args):
+    #     raise lazyerror
+
     def _make_sure_data_is_contiguous(self, log=None):
         self._make_lazy(rechunk=True)
 
@@ -771,17 +774,14 @@ class LazySignal(BaseSignal):
             if reproject:
                 if algorithm == 'PCA':
                     method = obj.transform
-
-                    def post(a): return np.concatenate(a, axis=0)
+                    post = lambda a: np.concatenate(a, axis=0)
                 elif algorithm == 'ORPCA':
                     method = obj.project
                     obj.R = []
-
-                    def post(a): return obj.finish()[4]
+                    post = lambda a: obj.finish()[4]
                 elif algorithm == 'ONMF':
                     method = obj.project
-
-                    def post(a): return np.concatenate(a, axis=1).T
+                    post = lambda a: np.concatenate(a, axis=1).T
 
                 _map = map(lambda thing: method(thing),
                            self._block_iterator(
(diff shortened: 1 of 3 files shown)


tjof2:
@francisco-dlp if you're dropping RPCA scaling (as per @Konrad0 raising the issue here: https://github.com/hyperspy/hyperspy/issues/1858) then can you update the default values for `lambda1` and `lambda2`? Instead of `1.0 / sqrt(nsamples)` (or `nfeatures` or whatever way round it is) - can you change the numerator to be the mean of the matrix?

The choice of 1.0 as the numerator relies on the matrix being approximately scaled to `[0,1]` range - but if we're doing away with scaling, we should have some sense of the data range fed into the parameters. I think any of the mean/median/max will be fine.

I'm away from a PC to do this myself.

francisco-dlp:
@tjof2, sure, I wanted to ask you about this before proceeding, but there has been no need. What would you think about not providing default values for the parameters at all? In this way the importance of setting the right values for those parameters is more apparent to the user. We could also add an "auto" value for the parameters to compute them as you mention, with a big warning explaining that ``auto`` may not always lead to optimal results.

tjof2:
@francisco-dlp that suits me fine - good plan. Don't know if the docs would need updating to match?

francisco-dlp:
Closing in favour of #1918

PR #1879, opened 2018-03-22
Fourier log improve message

erh3cq:
During fourier_log deconvolution an error occurs:
Cannot cast ufunc multiply output from dtype('float64') to dtype('int32') with casting rule 'same_kind'.
 This is meant to improve the error message so that people know how to fix the issue.

### DIFF
--- a/hyperspy/_signals/signal1d.py
+++ b/hyperspy/_signals/signal1d.py
@@ -1198,33 +1198,38 @@ _spikes_diagnosis,
                 nav_chunks = dc.chunks[:-1]
             zeros = da.zeros(nav_shape + (offset,),
                              chunks=nav_chunks + ((offset,),))
-        if side == 'left' or side == 'both':
-            if self._lazy:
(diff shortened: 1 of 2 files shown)


hyperspy/_signals/signal1d.py
```diff
@@ -1207,33 +1207,38 @@ def hanning_taper(self, side='both', channels=None, offset=0):
...
-                thelist = [therest, tapered]
-                if offset != 0:
-                    thelist.append(zeros)
-                dc = da.concatenate(thelist, axis=-1)
-            else:
-                dc[..., -channels - offset:rl] *= (
-                    np.hanning(2 * channels)[-channels:])
-                if offset != 0:
-                    dc[..., -offset:] *= 0.
+        try:
```
francisco-dlp: What about checking if the dtype is float instead of using the ``try except`` approach?

ericpre:
This PR is now rebased on RELEASE_next_patch, changed to check the dtype and comes with corresponding tests.

PR #1885, opened 2018-03-26
BUGFIX: Remove illegal filename characters from window titles

### LINKED ISSUE #1884: Cannot save the result of get_lines_intensity(plot_result=True)
Using the latest `RELEASE_next_minor` and the following versions:

```
Python 3.6.4 | packaged by conda-forge
hyperspy (1.4.dev0+git.553.g3ea779ae, c:\users\jat\git_repos\hyperspy)
hyperspy-gui-ipywidgets (1.0.0, c:\users\jat\git_repos\hyperspy_gui_ipywidgets)
hyperspy-gui-traitsui (1.0.3, c:\users\jat\git_repos\hyperspy_gui_traitsui)
pyqt                      4.11.4                   py36_2    conda-forge
matplotlib                2.0.0               np112py36_2    conda-forge
```

The output of `s.get_lines_intensity(plot_result=True)` cannot be saved using the save figure icon in the matplotlib figure. There's no output in the console or in the notebook when clicking the button.

jat255:
Fixes #1884. 

This is a bit of a workaround, but will prevent issues like #1884 from popping up in the future. Ideally, the backends of `matplotlib` would handle this more elegantly, but instead we can be careful about what filenames we pass in.

`matplotlib` automatically uses the current window title to create a suggested filename when you click the "Save figure" button.  The only place we can control this is in the window title. This PR replaces illegal non-standard characters in the window title with underscores. We lose a little bit of "prettiness" in the opened windows, for the benefit of being able to save those figures interactively.

### DIFF
--- a/hyperspy/drawing/utils.py
+++ b/hyperspy/drawing/utils.py
@@ -28,6 +28,7 @@ import matplotlib as mpl
 import hyperspy as hs
 from distutils.version import LooseVersion
 import logging
+import re
 
 from hyperspy.defaults_parser import preferences
 
@@ -125,6 +126,8 @@ def create_figure(window_title=None,
     """
     fig = plt.figure(**kwargs)
     if window_title is not None:
+        # remove non-alphanumeric characters to prevent file saving problems
+        window_title = re.sub('[^\w\-_\. ]', '_', window_title)
         fig.canvas.set_window_title(window_title)
     if _on_figure_window_close is not None:
         on_figure_window_close(fig, _on_figure_window_close)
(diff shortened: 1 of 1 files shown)


francisco-dlp:
It seems to me that the reason why this has not been fixed yet in matplotlib is that it is not straightforward. Sanitizing the filename sounds good for Windows users, but not for other OSs. For example, my Linux OS handles most of the characters that windows doesn't without issue. If that's the case of MacOS too, what about sanitizing only when running in windows?

jat255:
Personally, I'm of the opinion that it doesn't really matter, since the user can name the file whatever they want, and what we provide is just a suggestion. Personally, even on Linux I would never use a `:` in a filename, because someday I might want to open that file on a windows box, or send it to a colleague or something. I don't see any harm in taking those characters out of the suggestion. Just my $0.02

francisco-dlp:
That's a good point!

PR #1890, opened 2018-04-04
Fix blit

### LINKED ISSUE #1874: Qt repaint issue with recent version of matplotlib > 2.1.2
Hi there,

I've been a hyperspy user since early '16, so I take the chance to deeply thank you for your work :]

I tried installing hyperspy on a new machine today and cannot visualise data with the qt backend. I get an infinite loop of :
QWidget::repaint: Recursive repaint detected

kernel crash follows. The data display normally on inline backend and on a hyperspy version installed on an other computer. Are you aware of the problem? Here is the my anaconda environment config:

pyqt 5.6.0
qt 5.6.2
matplotlib 2.2.2
numpy 1.11.3
hyperspy 1.3

I use anaconda on windows 7

Cheers,

Nicolas

francisco-dlp:
Fixes #1874. The code that this fixes was so broken that what is difficult to understand is why it was working at all with mpl < 2.2.0.

### Progress of the PR
- [x] add tests,
- [x] ready for review.

### DIFF
--- a/hyperspy/drawing/figure.py
+++ b/hyperspy/drawing/figure.py
@@ -29,6 +29,7 @@ _logger = logging.getLogger(__name__)
 class BlittedFigure(object):
 
     def __init__(self):
+        self._draw_event_cid = None
         self._background = None
         self.events = Events()
         self.events.closed = Event("""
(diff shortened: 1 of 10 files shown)


thomasaarholt:
Code looks good to me, and plotting still works ^^

hyperspy/drawing/figure.py
```diff
@@ -39,52 +40,29 @@ def __init__(self):
...
-            canvas = self.figure.canvas
-            canvas.mpl_disconnect(self.draw_event_cid)
-            if canvas.supports_blit:
-                self._set_background()
-                self._draw_animated()
-            else:
-                canvas.draw_idle()
-            self.draw_event_cid = canvas.mpl_connect(
-                'draw_event', self._on_draw)
+    def _update_background(self, *args):
```
ericpre: The name of this method doesn't describe very well what it does and it is a bit confusing. With `_update_background`, you mean "cache background" but then you draw all the animated artists. I would either rename this method and take `_draw_animated` out or find a name more suitable.

hyperspy/drawing/signal1d.py
```diff
@@ -60,8 +60,9 @@ def create_figure(self):
             window_title="Figure " + self.title if self.title
```
ericpre: I would move the `create_figure` method to `BlittedFigure` because this part is repeated in `ImagePlot` and `HistogramTilePlot`. Also it makes the code easier to read since the callback is a method of `BlittedFigure`.

hyperspy/drawing/signal1d.py
```diff
@@ -349,7 +350,7 @@ def update(self, force_replot=False):
         if self.plot_indices is True:
             self.text.set_text(self.axes_manager.indices)
         if self.ax.figure.canvas.supports_blit:
```
ericpre, line 343: For the same reason as above, I would put it in a method and to move it to `BlittedFigure` and call the same method from `ImagePlot` and `HistogramTilePlot`.

hyperspy/drawing/image.py
```diff
@@ -371,7 +371,7 @@ def format_coord(x, y):
             else:
                 ims[0].changed()
             if self.figure.canvas.supports_blit:
-                self._draw_animated()
+                self._update_animated()
                 # It seems that nans they're simply not drawn, so simply replacing
                 # the data does not update the value of the nan pixels to the
                 # background color. We redraw everything as a workaround.
```
ericpre: I can't reproduce this!

ericpre (commented):
This is much more clear than before! I would suggest to add a few explaination on what it is doing to make our life easier in the future! ;)
I have added a few comments suggesting to move a few things to `BlittedFigure` in order to separate it from the plot "configuration" part in `Signal1DFigure` and `ImagePlot`.

hyperspy/drawing/figure.py
francisco-dlp: Done. Drawing the animated elements has to be done "manually" here, so I can't take it out from the method.

hyperspy/drawing/signal1d.py
francisco-dlp: Done

hyperspy/drawing/signal1d.py
francisco-dlp: Done

hyperspy/drawing/image.py
francisco-dlp: Yes, that was probably a workaround for another bug (internal or internal). I've removed it.

francisco-dlp:
This is again ready for review.

ericpre:
OK, we are almost there... CI is now complaining about the `spikes_removal_tool` and when I run the tests on my system, I actually get more failures with the `test_eds_tem.py:Test_eds_markers`! I was wondering why we would need to flush_events for the `ImagePlot` and not `Signal1DFigure`. It looks to me that it would make more sense to put:
```python
self.ax.figure.canvas.draw_idle()
try:
    # make sure that the GUI framework has a chance to run its event loop
    # and clear any GUI events.  This needs to be in a try/except block
    # because the default implementation of this method is to raise
    # NotImplementedError
    self.ax.figure.canvas.flush_events()
except NotImplementedError:
    pass
```
in the method of `BlittedFigure` and call it at the end of the `plot` method of  `ImagePlot` and not `Signal1DFigure`.
This does fix the failure with the `test_eds_tem.py:Test_eds_markers` but not with `test_tools.py:test_spikes_removal_tool`...

francisco-dlp:
The spikes removal tool error (and a few others) came with  https://github.com/francisco-dlp/hyperspy/pull/40 that I have now reverted.

francisco-dlp:
@ericpre, since  francisco-dlp#40 was actually unrelated to this PR and the tests are now passing again, won't you consider merging this and fixing the duplicated call to update in a new PR?

ericpre:
Yes, I am waiting for appveyor to finish!

hyperspy/drawing/figure.py
```diff
@@ -39,52 +41,57 @@ def __init__(self):
...
-                                    'before `draw`.')
-                    pass
-            if canvas.supports_blit:
-                canvas.blit(self.figure.bbox)
+        """Draw animated plot elements
+
+        """
+        for ax in self.figure.axes:
+            # Create a list of animated artitst and draw them.
+            artists = []
```
tacaswell, line 77: You can probably use `ax.get_children()` here.

Issue #1894, opened 2018-04-05
add_*_noise keeps data type when it shouldn't


francisco-dlp:
For example:

```python
>>> import hyperspy.api as hs
>>> import numpy as np
>>> s = hs.signals.Signal1D([0, 1])
>>> s.data.dtype
dtype('int64')
```
The dtype should be ``float64`` instead as the noise.

```python
>>> s.add_gaussian_noise(std=1)
>>> s.data.dtype
dtype('int64')
>>> s.data
array([-1,  0])
>>> s = hs.signals.Signal1D([0.5, 1])
>>> s.add_poissonian_noise()
>>> s.data.dtype
dtype('float64')
>>> s.data
array([0., 0.])
```
The dtype should always be ``int`` for poisson "noised" data.

ericpre:
Fixed in #1895.

PR #1895, opened 2018-04-05
Fix lazy noise (#1886)

### LINKED ISSUE #1886: add_*_noise does not operate lazily for lazy signals
To reproduce the issue:

```python
>>> import hyperspy.api as hs
>>> s = hs.signals.Signal1D([1,2,3]).as_lazy()
>>> s.add_poissonian_noise()
>>> s.data
array([1, 2, 2])
```

francisco-dlp:
### Description of the change

Fixes #1886 and #1894 

### Progress of the PR
- [x] add tests,
- [x] ready for review.


hyperspy/tests/signal/test_tools.py
```diff
@@ -184,6 +185,42 @@ def test_numpy_func(self):
...
+            kwargs["chunks"] = s.data.chunks
+        else:
+            data = s.data.copy()
+            from numpy.random import seed, normal
+        seed(1)
+        s.add_gaussian_noise(std=1.0)
+        seed(1)
+        if s._lazy:
+            s.compute()
+        np.testing.assert_array_almost_equal(
```
ericpre, line 209: If you do `s.compute` for lazy signal, then `s` become a non lazy signal and in the tests there is no need to use the `normal` function from dask?

hyperspy/tests/signal/test_tools.py
```diff
@@ -184,6 +185,42 @@ def test_numpy_func(self):
...
+            kwargs["chunks"] = s.data.chunks
+        else:
+            data = s.data.copy()
+            from numpy.random import seed, poisson
+        seed(1)
+        s.add_poissonian_noise()
+        seed(1)
+        if s._lazy:
+            s.compute()
+        np.testing.assert_array_almost_equal(
```
ericpre: Same as above?

hyperspy/tests/signal/test_tools.py
francisco-dlp: I also thought so, but it turns out that the random numbers generated by dask and numpy with the same seed differ.

ericpre: The tests are failing with the same systematic error. Could it be because of the seed issue?

francisco-dlp:
Thanks @ericpre. Now it should be fixed. I have added a ``keep_dtype`` kwarg to ``add_poissonian_noise`` for convenience. I haven't added it to the gaussian noise one because it is not possible to do it in a safe way, so it's better to let numpy raise an exception rather than having a way to introduce difficult to detect bugs.

PR #1896, opened 2018-04-05
Plot imaginary part of real numbers for ComplexSignal.

### LINKED ISSUE #1893: Imaginary part of real numbers is not plotted for ComplexSignal
I have experienced that the imaginary part of a complex signal sometimes misses when plotted. It seems like it is consequently when the indices of the signal corresponds to a zero imaginary part. A reconstructed situation is added below where index 0 results in a missing imaginary part but index 1 does not.

I believe the bug is within the` plot() ` function in the `ComplexSignal1D` class, but I cannot see a clear cause.

```python
data = (np.ones((500,2)) + 1j*np.ones((500,2))*np.array([0,1])).T

signal = hs.signals.ComplexSignal1D(data)

# Imaginary part missing
#signal.axes_manager.indices = (0,)
#signal.plot()

# Imaginary part appear
signal.axes_manager.indices = (1,)
signal.plot()
```

ericpre:
Fixes #1893.

### Progress of the PR
- [x] ready for review.

### DIFF
--- a/hyperspy/drawing/mpl_hse.py
+++ b/hyperspy/drawing/mpl_hse.py
@@ -112,7 +112,7 @@ class MPL_HyperSignal1D_Explorer(MPL_HyperExplorer):
             sl.autoscale = True
             sl.data_function = self.signal_data_function
             sl.plot_coordinates = True
-            sl.plot_imag = True
+            sl._plot_imag = True
             sl.set_line_properties(color="blue", type='step')
             # Add extra line to the figure
             sf.add_line(sl)
--- a/hyperspy/drawing/mpl_hse.py
+++ b/hyperspy/drawing/mpl_hse.py
@@ -93,8 +93,8 @@ class MPL_HyperSignal1D_Explorer(MPL_HyperExplorer):
         sf.axes_manager = self.axes_manager
         self.signal_plot = sf
         # Create a line to the left axis with the default indices
-        is_complex = np.iscomplex(self.signal_data_function()).any()
         sl = signal1d.Signal1DLine()
+        is_complex = np.iscomplexobj(self.signal_data_function())
         sl.autoscale = True if not is_complex else False
         sl.data_function = self.signal_data_function
         sl.plot_indices = True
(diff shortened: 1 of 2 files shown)


francisco-dlp:
Does this actually fix something? (The code seems to be roughly equivalent to the original one). Also, I wouldn't change name of the attribute for a patch release as it may break external packages.

hyperspy/drawing/mpl_hse.py
```diff
@@ -93,8 +93,8 @@ def plot_signal(self):
         sf.axes_manager = self.axes_manager
         self.signal_plot = sf
         # Create a line to the left axis with the default indices
-        is_complex = np.iscomplex(self.signal_data_function()).any()
         sl = signal1d.Signal1DLine()
+        is_complex = np.iscomplexobj(self.signal_data_function())
```
francisco-dlp, line 97: Good catch! The original line was working AFAIK, but it was brittle.

francisco-dlp:
Sorry for the previous (too) early comment. Yes, this fixes the issue by changing ``np.iscomplex`` to ``np.iscomplexobj``. I am happy to merge this, except for the ``get_complex`` to ``plot_complex`` rename. I agree that the new name is better, but we shouldn't change the public API in patch releases. Maybe you could pull request the rename part of this PR to RnM?

ericpre:
What's about making `plot_imag` private, add a deprecated property `get_complex` which return `self._plot_imag` and remove it in hyperspy 2.0?
I can't think of any case when this attribute can be used really, so it should be private.

francisco-dlp:
:+1:

ericpre:
Done.

hyperspy/drawing/signal1d.py
```diff
@@ -201,6 +203,13 @@ def __init__(self):
         self._line_properties = {}
         self.type = "line"
 
+    @property
+    def get_complex(self):
+        warnings.warn("The `get_complex` attribute is deprecated and will be"
+              "removed in 2.0, please use `_plot_image` instead.",
```
thomasaarholt: minor typo - _plot_image -> _plot_imag

ericpre: Done.

PR #1900, opened 2018-04-08
Fix redundant drawing connection

ericpre:
### Progress of the PR
- [x] Remove redundant drawing connection to avoid unnecessary re-draw and increase plotting speed,
- [x] adjust accordingly the `spikes_removal_tool`,
- [x] fix an issue with the pointer which was not updating when removing spikes,
- [x] ready for review.

### DIFF
--- a/hyperspy/drawing/figure.py
+++ b/hyperspy/drawing/figure.py
@@ -58,22 +58,6 @@ class BlittedFigure(object):
             self._draw_event_cid = self.figure.canvas.mpl_connect(
                 'draw_event', self._on_blit_draw)
 
-    def _init_draw(self):
-        """ Initialise the drawing of the figure"""
-        # ask the canvas to re-draw itself the next time it
-        # has a chance.
-        # For most of the GUI backends this adds an event to the queue
-        # of the GUI frameworks event loop.
-        self.ax.figure.canvas.draw_idle()
-        try:
-            # make sure that the GUI framework has a chance to run its event loop
-            # and clear any GUI events.  This needs to be in a try/except block
-            # because the default implementation of this method is to raise
-            # NotImplementedError
-            self.ax.figure.canvas.flush_events()
-        except NotImplementedError:
-            pass
-
     def _on_blit_draw(self, *args):
         fig = self.figure
         # As draw doesn't draw animated elements, in its current state the
--- a/hyperspy/drawing/figure.py
+++ b/hyperspy/drawing/figure.py
@@ -58,6 +58,22 @@ class BlittedFigure(object):
             self._draw_event_cid = self.figure.canvas.mpl_connect(
                 'draw_event', self._on_blit_draw)
 
+    def _init_draw(self):
+        """ Initialise the drawing of the figure"""
+        # ask the canvas to re-draw itself the next time it
+        # has a chance.
+        # For most of the GUI backends this adds an event to the queue
+        # of the GUI frameworks event loop.
+        self.ax.figure.canvas.draw_idle()
+        try:
+            # make sure that the GUI framework has a chance to run its event loop
+            # and clear any GUI events.  This needs to be in a try/except block
+            # because the default implementation of this method is to raise
+            # NotImplementedError
+            self.ax.figure.canvas.flush_events()
+        except NotImplementedError:
+            pass
+
     def _on_blit_draw(self, *args):
         fig = self.figure
         # As draw doesn't draw animated elements, in its current state the
--- a/hyperspy/drawing/image.py
+++ b/hyperspy/drawing/image.py
@@ -292,19 +292,7 @@ class ImagePlot(BlittedFigure):
                 pass
 
         self.connect()
-        # ask the canvas to re-draw itself the next time it
-        # has a chance.
-        # For most of the GUI backends this adds an event to the queue
-        # of the GUI frameworks event loop.
-        self.figure.canvas.draw_idle()
-        try:
-            # make sure that the GUI framework has a chance to run its event loop
-            # and clear any GUI events.  This needs to be in a try/except block
-            # because the default implementation of this method is to raise
-            # NotImplementedError
-            self.figure.canvas.flush_events()
-        except NotImplementedError:
-            pass
+        self._init_draw()
 
     def update(self, **kwargs):
         ims = self.ax.images
--- a/hyperspy/drawing/image.py
+++ b/hyperspy/drawing/image.py
@@ -292,7 +292,7 @@ class ImagePlot(BlittedFigure):
                 pass
 
         self.connect()
-        self._init_draw()
+        self.figure.canvas.draw()
 
     def update(self, **kwargs):
         ims = self.ax.images
--- a/hyperspy/drawing/signal1d.py
+++ b/hyperspy/drawing/signal1d.py
@@ -87,6 +87,10 @@ class Signal1DFigure(BlittedFigure):
             line.sf_lines = self.right_ax_lines
             if line.axes_manager is None:
                 line.axes_manager = self.right_axes_manager
+        line.axes_manager.events.indices_changed.connect(line.update, [])
+        self.events.closed.connect(
+            lambda: line.axes_manager.events.indices_changed.disconnect(
+                line.update), [])
         line.axis = self.axis
         # Automatically asign the color if not defined
         if line.color is None:
@@ -112,11 +116,6 @@ class Signal1DFigure(BlittedFigure):
         for marker in self.ax_markers:
             marker.plot()
         plt.xlim(np.min(x_axis_lower_lims), np.max(x_axis_upper_lims))
-        # To be discussed
-        self.axes_manager.events.indices_changed.connect(self.update, [])
-        self.events.closed.connect(
-            lambda: self.axes_manager.events.indices_changed.disconnect(
-                self.update), [])
         self.ax.figure.canvas.draw_idle()
         if hasattr(self.figure, 'tight_layout'):
             try:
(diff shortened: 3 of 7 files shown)


hyperspy/drawing/image.py
```diff
@@ -292,19 +292,7 @@ def plot(self, **kwargs):
...
-        self.figure.canvas.draw_idle()
-        try:
-            # make sure that the GUI framework has a chance to run its event loop
-            # and clear any GUI events.  This needs to be in a try/except block
-            # because the default implementation of this method is to raise
-            # NotImplementedError
-            self.figure.canvas.flush_events()
-        except NotImplementedError:
-            pass
+        self._init_draw()
```
francisco-dlp: Wouldn't it be better to simply call ``self.figure.canvas.draw()`` instead?

hyperspy/drawing/signal1d.py
```diff
@@ -125,6 +125,7 @@ def plot(self):
                 # tight_layout is a bit brittle, we do this just in case it
                 # complains
                 pass
+        self._init_draw()
```
francisco-dlp: Wouldn't it be better to simply call ``self.figure.canvas.draw()`` instead?

hyperspy/drawing/figure.py
```diff
@@ -58,6 +58,22 @@ def create_figure(self, **kwargs):
             self._draw_event_cid = self.figure.canvas.mpl_connect(
                 'draw_event', self._on_blit_draw)
 
+    def _init_draw(self):
```
francisco-dlp: This may be unnecessary, see below.

hyperspy/signal_tools.py
```diff
@@ -873,10 +873,12 @@ def __init__(self, signal, navigation_mask=None, signal_mask=None):
                 self.noise_type = "shot noise"
         else:
             self.noise_type = "shot noise"
+        self.find()
```
francisco-dlp: Why calling ``find`` here?

hyperspy/signal_tools.py
```diff
@@ -873,10 +873,12 @@ def __init__(self, signal, navigation_mask=None, signal_mask=None):
...
 
     def _threshold_changed(self, old, new):
         self.index = 0
         self.update_plot()
+        self.find()
```
francisco-dlp: and here?

hyperspy/signal_tools.py
```diff
@@ -950,6 +952,8 @@ def find(self, back=False):
                     minimum),
                 self.signal.axes_manager.signal_axes[0].index2value(
                     maximum))
+            self.signal.axes_manager.indices = self.coordinates[self.index]
```
francisco-dlp: Why are these two new lines necessay?

hyperspy/tests/signal/test_tools.py
```diff
@@ -869,11 +869,8 @@ def test_spikes_removal_tool():
 
     sr = SpikesRemoval(s)
     sr.threshold = 1.5
-    sr.find()
```
francisco-dlp: It seems that the spikes removal tool is now behaving differently, when it shouldn't.

hyperspy/signal_tools.py
```diff
@@ -1005,9 +1009,11 @@ def create_interpolation_line(self):
         self.interpolated_line.set_line_properties(
             color='blue',
             type='line')
-        self.signal._plot.signal_plot.add_line(self.interpolated_line)
+        self.signal._plot.signal_plot.add_line(self.interpolated_line,
```
francisco-dlp: why shouldn't this line be updated automatically?

hyperspy/drawing/signal1d.py
```diff
@@ -87,6 +87,11 @@ def add_line(self, line, ax='left'):
             line.sf_lines = self.right_ax_lines
             if line.axes_manager is None:
                 line.axes_manager = self.right_axes_manager
+        if connect_event:
```
francisco-dlp: If this is finally not necessary for this PR (to be discussed) I would suggest to remove it.

ericpre: There are situations, where you don't want to connect the second line (spikes_removal_tool, model line , etc, need to check) to avoid redrawing a second time, the same thing.

hyperspy/drawing/signal1d.py
ericpre: Yes, indeed.

hyperspy/signal_tools.py
ericpre: I am to going it to revert it because it may not be desirable if the threshold value is too high in case of large dataset since it can take too long to perform... The idea is to save the user a click on the "find" button at the start of the spikes_removal_tool, but it may be a wrong good idea! Is it what you are concerned about?

hyperspy/signal_tools.py
ericpre: Yes one of the two is not necessary, some left over... I moved the second line at a more appropriate spot in the `update_plot()` method and this one is indeed necessary, because the pointer was not updating! Possibly since the split of the UI.

hyperspy/drawing/signal1d.py
ericpre: Do you think it is worth to look for places, when this may be the case and fixes it accordingly?

francisco-dlp: I think that it's better to stick to the topic of this particular PR. I think that this is a good idea in general as it decreases the time needed to review individual PRs and it makes it easier to find a short title that describes the changes.

hyperspy/tests/signal/test_tools.py
```diff
@@ -872,7 +872,6 @@ def test_spikes_removal_tool():
     sr.find()
     assert s.axes_manager.indices == (0, 1)
     sr.threshold = 0.5
-    assert s.axes_manager.indices == (0, 0)
```
francisco-dlp: This was testing that changing the threshold sets the indices back to zero, why removing it?

hyperspy/drawing/signal1d.py
```diff
@@ -310,7 +307,6 @@ def plot(self, data=1):
                                      fontsize=12,
                                      color=self.line.get_color(),
                                      animated=self.ax.figure.canvas.supports_blit)
-        self.ax.figure.canvas.draw_idle()
```
francisco-dlp: Doesn't removing this imply that the line is not drawn when calling the ``plot`` method? If yes, why would that be a good idea?

hyperspy/signal_tools.py
```diff
@@ -1005,9 +1005,11 @@ def create_interpolation_line(self):
...
+                                               connect_event=False)
         self.interpolated_line.autoscale = False
         self.interpolated_line.plot()
+        self.interpolated_line.update()
```
francisco-dlp: Why do we need to call ``update`` here if the line data haven't changed?

hyperspy/drawing/signal1d.py
francisco-dlp: If this feature is gonna stay I would suggest to rename the keyword to something more explicit such as ``update_on_navigation`` or something of this sort.

ericpre:
Ready for review again: indeed, there were a few things not very sensible!

hyperspy/signal_tools.py
```diff
@@ -960,6 +959,7 @@ def update_plot(self):
         self.reset_span_selector()
         self.update_spectrum_line()
         if len(self.coordinates) > 1:
+            self.signal._plot.pointer.indices = self.coordinates[self.index]
```
francisco-dlp: This is actually a workaround for what may be a minor bug: when ``_auto_update_plot`` is switched from ``False`` to ``True`` the pointer's indices should be set  to the current ones and the patch redrawn, probably around https://github.com/hyperspy/hyperspy/blob/RELEASE_next_minor/hyperspy/drawing/mpl_hse.py#L55

hyperspy/signal_tools.py
```diff
@@ -1005,7 +1005,8 @@ def create_interpolation_line(self):
         self.interpolated_line.set_line_properties(
             color='blue',
             type='line')
-        self.signal._plot.signal_plot.add_line(self.interpolated_line)
+        self.signal._plot.signal_plot.add_line(self.interpolated_line,
+                                               update_on_navigation=False)
```
francisco-dlp: I think that ``update_on_navigation`` may not be needed as setting its ``auto_update`` attribute to ``False`` achieves the same effect and, ideally, there should only be one way of achieving the effect.

hyperspy/drawing/signal1d.py
```diff
@@ -87,6 +87,11 @@ def add_line(self, line, ax='left'):
             line.sf_lines = self.right_ax_lines
             if line.axes_manager is None:
                 line.axes_manager = self.right_axes_manager
+        if update_on_navigation:
```
francisco-dlp: I think that this may not be needed, see below.

hyperspy/signal_tools.py
ericpre: Yes, `auto_update_plot=True` will update the pointer position and redraw the pointer nicely but it will do it for each index and slows down significantly finding spikes, since a indices change of the axes manager will redraw the patch position! So it seems that it is working as expected but not what we want.
`_update_patch_position()` is also working fine but since the actual pointer indices haven't changed, the same thing is redrawn. The only way that I see here is setting the pointer indices manually and changing `_update_patch_position()` to `_on_navigate(self.signal.axes_manager)` does the job here. Actually, just using the pointer position or indices properties would also do, since it trigger `_pos_changed` event and call `_update_patch_position`.

hyperspy/signal_tools.py
ericpre: Yes, this is better, with a caveat that we need to ignore the `auto_update=False` on the `span_selector_changed` callback.

hyperspy/drawing/signal1d.py
ericpre: Done.

PR #1901, opened 2018-04-08
Parse more metadata dm

### LINKED ISSUE #1859: DM EELS spectra do not read exposure from original_metadata
I'm observing that for EELS spectra imported from dm3 files, the exposure (or dwell time or "integration time" depending on names) is not correctly placed in the metadata.

This is valid for dm3 files recorded in Oxford and Oslo on separate instruments.

I noticed this in connection with #1851 as my lowloss and coreloss had not been acquired with similar exposure times. I then realised that it appears that the eelsmodel convolved with lowloss does not take into account exposure times or the number of frames, even if they are provided.

ericpre:
Partially fix #1859. It seems that the integration time is used for single spectrum while exposure is used for mapping. When using the exposure metadata, it needs to be multiplied by the number of frames: integration time = exposure * number of frames.

### Progress of the PR
- [x] parse the EELS dwell time of DM files to metadata,
- [x] add beta tilt,
- [x] ready for review.

### DIFF
--- a/hyperspy/io_plugins/digital_micrograph.py
+++ b/hyperspy/io_plugins/digital_micrograph.py
@@ -854,7 +854,9 @@ class ImageObject(object):
             "{}.Microscope Info.Voltage".format(tags_path): (
                 "Acquisition_instrument.TEM.beam_energy", lambda x: x / 1e3),
             "{}.Microscope Info.Stage Position.Stage Alpha".format(tags_path): (
-                "Acquisition_instrument.TEM.Stage.tilt_alpha", None),
+                "Acquisition_instrument.TEM.Stage.tilt_alpha", lambda x: round(x, 2)),
+            "{}.Microscope Info.Stage Position.Stage Beta".format(tags_path): (
+                "Acquisition_instrument.TEM.Stage.tilt_beta", lambda x: round(x, 2)),
             "{}.Microscope Info.Stage Position.Stage X".format(tags_path): (
                 "Acquisition_instrument.TEM.Stage.x", lambda x: x * 1e-3),
             "{}.Microscope Info.Stage Position.Stage Y".format(tags_path): (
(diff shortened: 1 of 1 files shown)


hyperspy/io_plugins/digital_micrograph.py
```diff
@@ -821,6 +821,21 @@ def _parse_string(self, tag):
...
+        # compute exposure * number of frames
+        try:
+            if 'Integration_time_s' in tags.keys():
+                return float(tags["Integration time (s)"])
+            else:
+                frame_number = 1
+                if "Number_of_frames" in tags.keys():
+                    frame_number = float(tags["Number_of_frames"])
+                return float(tags["Exposure_s"]) * frame_number
+        except AttributeError:
```
francisco-dlp: How is this going to raise an ``AttributeError``? Wouldn't ``if, elif, else`` be preferable in this case?

hyperspy/io_plugins/digital_micrograph.py
```diff
@@ -839,7 +853,9 @@ def get_mapping(self):
             "{}.Microscope Info.Voltage".format(tags_path): (
                 "Acquisition_instrument.TEM.beam_energy", lambda x: x / 1e3),
             "{}.Microscope Info.Stage Position.Stage Alpha".format(tags_path): (
-                "Acquisition_instrument.TEM.Stage.tilt_alpha", None),
+                "Acquisition_instrument.TEM.Stage.tilt_alpha", lambda x: round(x, 2)),
```
francisco-dlp: Why using ``round`` here? Even if that's the current typical accuracy, this may change. Also, it would be good to list this enhancement in the PR description.

ericpre: Yeah, I though about it, but this already allows a precision of a 1/100 of a degree! It is just to avoid to display value with nonsense digits, but then, there is the question of why 2 and not 3 decimal digits...

francisco-dlp: I would suggest not to round it to stay of the safe side...

ericpre: Done.

PR #1904, opened 2018-04-10
Use `get_children` to generate animated artists list when using blit

francisco-dlp:
### Description of the change

The list of artists to draw when using ``blit`` is hard-wired. This uses matplotlib's ``get_children`` to generate the list instead. Thanks @tacaswell for suggesting this fix in #1890.

### Progress of the PR
- [x] ready for review.


hyperspy/drawing/figure.py
```diff
@@ -74,16 +74,10 @@ def _draw_animated(self):
...
-            artists.extend(ax.patches)
-            artists.extend(ax.lines)
-            artists.extend(ax.texts)
-            artists.extend(ax.artists)
-            artists.append(ax.get_yaxis())
-            artists.append(ax.get_xaxis())
-            [ax.draw_artist(a) for a in artists if a.get_animated()]
+            for artist in ax.get_children():
+                # It turns out that not all artists have an animated attribute
+                if hasattr(artist, "animated") and artist.animated:
```
ericpre: It looks to me that `get_children()` is making a list with a specific order and the image is drawn on on the widgets, I would suggest:
```python
# Create a list of animated artitst and draw them.
artists = ax.get_children()
# need to put the images at the beginning of the list
if len(ax.images) > 0:
    for image in ax.images:
        artists.remove(image)
        artists.insert(0, image)
for artist in artists:
    if artist.get_animated():
        ax.draw_artist(artist)
```

ericpre (changes requested):

hyperspy/drawing/figure.py
tacaswell: You probably should sort by z-order `artists = sorted(artists, key=attrgetter('zorder'))` rather than hard-coding images at the front.

francisco-dlp:
@ericpre, you're totally right. I have now fixed the code, but it still doesn't work. For example, when plotting a spectrum image the rectangle pointer is missing, even though the ``get_children()`` returns the patch and ``get_animated()`` is ``True``, hence ``draw_artist`` is executed. I thought that it could be a ``zorder`` issue, but setting the rectangle patch's zorder to 10 didn't help. Maybe the order of execution of ``draw_artist`` for each individual artist matters?

francisco-dlp:
Plotting as follows:

```python
%matplotlib qt
import hyperspy.api as hs
import numpy as np

s = hs.signals.Signal1D(np.zeros((10, 10, 100)))
s.plot()
```

This is the execution order of the animated artists with this code:

```
Line2D(_line0)
Text(1,8.9,'$2\\,$<undefined>')
Rectangle(xy=(-0.5, -0.5), width=1, height=1, angle=0)
AxesImage(34.5556,59.359;465.278x465.278)
YAxis(528.913194,23.466953)
Line2D(_line0)
Text(-0.1,1.05,'(0, 0)')
XAxis(66.972222,58.277778)
YAxis(66.972222,58.277778)
```

And this is the execution order with the hard-wired list:

```
AxesImage(34.5556,59.359;465.278x465.278)
Line2D(_line0)
Text(1,8.9,'$2\\,$<undefined>')
Rectangle(xy=(-0.5, -0.5), width=1, height=1, angle=0)
YAxis(528.913194,23.466953)
Line2D(_line0)
Text(-0.1,1.05,'(0, 0)')
YAxis(66.972222,58.277778)
XAxis(66.972222,58.277778)
```

The ``zorder`` of the image is 0 and that of the rectangle 1, so the latter should be painted above the image.

ericpre:
The current version of this branch doesn't set properly the zorder as suggested by @tacaswell, have you forgot to push some commits? I would expect to read something along the line of:
```python
for ax in self.figure.axes:
    # Create a list of animated artitst and draw them.
    for artist in sorted(ax.get_children(), key=attrgetter('zorder')):
        if artist.get_animated():
            ax.draw_artist(artist)
```

francisco-dlp:
No, I just failed to notice @tacaswell comment this morning when I set to implement the changes that you've suggested yesterday. It should be fixed now.

francisco-dlp:
I've edited the title and description and removed the bug label as this is finally not fixing any plotting issue.

ericpre (approved):

ericpre:
This looks good to me.

francisco-dlp:
merge it then?

PR #1909, opened 2018-04-12
Fix 1864 and add kwargs to GaussianHF's constructor

francisco-dlp:
When using the ``GaussianHF`` component several operations raised errors because the component is created using the ``Expression`` component, that was adding extra kwargs to the whitelist. This fixes the issue.

This also adds ``**kwargs`` to ``GaussianHF``'s constructor to enable passing keyword arguments to ``Expression``. See the example below.

### Progress of the PR
- [x] add tests,
- [x] ready for review.

The following example (completed from 1864) no longer raises an error.

```python
%matplotlib qt

import numpy as np
import hyperspy.api as hs

s = hs.signals.Signal1D(np.random.random((200, 200, 100)))

model = s.create_model()
g = hs.model.components1D.GaussianHF(module="numexpr")
model.append(g)
sm = model.as_signal()
```

### DIFF
--- a/hyperspy/_components/expression.py
+++ b/hyperspy/_components/expression.py
@@ -196,9 +196,10 @@ class Expression(Component):
                            modules=module, dummify=False)
 
         if self._is2D:
-            f = lambda x, y: self._f(x, y, *[p.value for p in self.parameters])
+            def f(x, y): return self._f(
+                x, y, *[p.value for p in self.parameters])
         else:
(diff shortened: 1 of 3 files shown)


francisco-dlp:
I am submitting this to discuss the fix approach. After we reach agreement, I'll add a test.

francisco-dlp:
I have added a test and improved the fix. This is ready for review.

PR #1910, opened 2018-04-13
New roi line angle

woozey:
### Description of the change
New feature for calculating angle of `roi.Line2DROI` with respect to axis. 
- Adds new method `roi.Line2DROI.angle()`

FYI: @jan-car, @FWin22 

### Progress of the PR
- [x] Change implemented,
- [x] update docstring,
- [x] update user guide,
- [x] add tests,
- [x] ready for review.

```python
>>> import hyperspy.api as hs
>>> import scipy
>>> holo = hs.datasets.example_signals.object_hologram()
>>> roi = hs.roi.Line2DROI(x1=465.577, y1=445.15, x2=169.4, y2=387.731, linewidth=0)
>>> holo.plot()
>>> ss = roi.interactive(holo)
```
screen shot 2018-04-13 at 10 23 21

```python
>>> roi
Line2DROI(x1=465.577, y1=445.15, x2=169.4, y2=387.731, linewidth=0)
>>> roi.angle(axis='y')
-100.97166759025453
>>> roi.angle(axis='vertical', units='radians')
-1.7622880506791903
>>> holo.map(scipy.ndimage.rotate, angle=roi.angle(axis='x'), inplace=False).plot()
```
screen shot 2018-04-13 at 10 26 32


francisco-dlp:
Excellent idea and great example of usage!

woozey:
User guide updated, ready for review))

hyperspy/roi.py
```diff
@@ -1040,6 +1040,50 @@ def length(self):
...
+            conversation = 1.
+        else:
+            raise ValueError("Units are not recognized. Use  either 'degrees' or 'radians'.")
+
+        if any(axis == a for a in ['horizontal', 'x']):
+            return np.arctan2(y, x) * conversation
+        elif any(axis == a for a in ['vertical', 'y']):
+            return np.arctan2(x, y) * conversation
+        else:
+            raise ValueError("Axis is not recognized."
```
ericpre: Add a space after the point.

hyperspy/roi.py
```diff
@@ -1040,6 +1040,50 @@ def length(self):
...
+            'x' is alias to 'horizontal' and 'y' is 'vertical'
+            (Default: 'horizontal')
+        units : str, {'degrees', 'radians'}
+            Select units in which the angle to be calculated
+            (Default: 'degrees')
+
+        Returns
+        -------
+        angle : float
+            Angle in radians or degrees in range [-pi, pi]
```
ericpre: typo: "degrees in range [-pi, pi]", actually this is already documented above, so this is a bit redundant.

hyperspy/roi.py
```diff
@@ -1040,6 +1040,50 @@ def length(self):
         d_row, d_col = p1 - p0
         return np.hypot(d_row, d_col)
 
+    def angle(self, axis='horizontal', units='degrees'):
+        """"Angle between ROI line and selected axis
+
+        Parameters
+        ----------
+        axis : str, {'horizontal', 'vertical', 'x', 'y'}, optional
+            Select axis with respect to which teh angle to be measured.
```
ericpre: "Select axis against which the angle of the ROI line is measured"?

hyperspy/roi.py
```diff
@@ -1040,6 +1040,50 @@ def length(self):
...
+            'x' is alias to 'horizontal' and 'y' is 'vertical'
+            (Default: 'horizontal')
+        units : str, {'degrees', 'radians'}
+            Select units in which the angle to be calculated
```
ericpre: Simplify: "The units of the output."?

hyperspy/roi.py
```diff
@@ -1040,6 +1040,50 @@ def length(self):
         d_row, d_col = p1 - p0
         return np.hypot(d_row, d_col)
 
+    def angle(self, axis='horizontal', units='degrees'):
+        """"Angle between ROI line and selected axis
+
+        Parameters
+        ----------
+        axis : str, {'horizontal', 'vertical', 'x', 'y'}, optional
```
ericpre: 'x' and 'y' can be names of axis and not necessarily the horizontal or vertical one of the signal axis. Maybe it could useful to be able to provide the name of the axis and remove these alias to avoid confusion?

ericpre (commented):
A few comments, mainly on wording/typos.

francisco-dlp:
@woozey, will you be able to address the minor comments by @ericpre above today so that we can include your contribution in v1.4?

hyperspy/roi.py
woozey: I'm not sure if I fully understood you here. Do you expect to provide name of the axis and 'horizontal'/'vertical'? I find 'horizontal/ 'vertical' convenient for using the method to align images with respect to image axis. (I also usually would not bother naming the image axes... )))

ericpre: Most reader will provide a name axis. I am suggesting to be able to use the name of the axis (the one from the axes_manager) as an alternative to 'horizontal' and 'vertical' and not to use 'x' and 'y' for the reason mentioned above.

hyperspy/roi.py
```diff
@@ -1046,17 +1046,17 @@ def angle(self, axis='horizontal', units='degrees'):
...
         units : str, {'degrees', 'radians'}
-            Select units in which the angle to be calculated
+            The angle units of the output
             (Default: 'degrees')
 
         Returns
         -------
         angle : float
-            Angle in radians or degrees in range [-pi, pi]
+            Angle in a range of [-pi, pi]
```
ericpre: I would remove this because this is already explain above and if you use degree, then this is wrong.

hyperspy/roi.py
```diff
@@ -1081,7 +1081,7 @@ def angle(self, axis='horizontal', units='degrees'):
         elif any(axis == a for a in ['vertical', 'y']):
             return np.arctan2(x, y) * conversation
         else:
-            raise ValueError("Axis is not recognized."
+            raise ValueError("Axis is not recognized. "
```
ericpre: If you change the `axis` parameter, this would need to change accordingly.

woozey:
@francisco-dlp Yep, I will have enough time today.

francisco-dlp:
Great, thanks!

hyperspy/roi.py
woozey: Removing 'x' and 'y' alias is no problem, and I'll do it to avoid confusion. Providing the axis name is actually a problem as `roi` objects have no 'axes_manager' and are not really connected to any signal as far as I understand? Am I right here or you have some suggestion how to implement it?

ericpre: The widgets should be `axes` aware. But if this is a bit too much of a pain then don't worry.

francisco-dlp:
Just waiting for this to release ;)

hyperspy/roi.py
woozey: I just have removed the alias. I would have to do quite a research to get something from widgets and that still this won't feel to me as the way to go, since the widgets are only used with `interactive()` (if I understand correctly)...

woozey:
@francisco-dlp, @ericpre if the issue with axis seems to be resolved, it should be ready to go

francisco-dlp:
Thanks @woozey, I'll wait to hear from @ericpre .

PR #1913, opened 2018-04-15
Remove bundle build from appveyor (move to hyperspy-bundle repo)

ericpre:
### Progress of the PR
- [x] move to the bundle build part of `appveyor.yml` to https://github.com/hyperspy/hyperspy-bundle, as dicussed below.
- [x] tidy up `appveyor.yml`
- [x] ready for review.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -5,7 +5,6 @@ environment:
 
   global:
     TEST_DEPS: "pytest pytest-cov wheel"
-    NSIS_DIR: "%PROGRAMFILES(x86)%/NSIS"
     MPLBACKEND: "agg"
 
 
@@ -18,9 +17,6 @@ environment:
        PYTHON_MAJOR: 3
        PYTHON_ARCH: "32"
        CONDA_PY: "35"
-       CONDA_NPY: "19"
-       WP_URL: 'https://github.com/winpython/winpython/releases/download/1.3.20160209/WinPython-32bit-3.5.1.2.exe'
-       WP_CRC: '172d19a743ccfaf55af779d15f29f67fca83a46f08b0af855dfaf809b4184c0d'
        DEPS: "numpy scipy matplotlib ipython h5py sympy scikit-learn dill setuptools natsort scikit-image cython ipyparallel dask"
 
      - PYTHON: "C:\\Miniconda35-x64"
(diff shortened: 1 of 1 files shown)


francisco-dlp:
Is this ready for review?

ericpre:
Yes.

francisco-dlp:
It looks good, except for targetting HyperSpyUI master instead of the current stable. Wouldn't it be better to release HyperSpyUI instead?

ericpre:
Agree, shall we wait for HyperSpyUI to be released (and available in pypi) then?

francisco-dlp:
If HyperSpyUI will be released in 1-2 days, we can wait. Otherwise, we must either:

1. release HyperSpy without the bundle
2. release HyperSpy with HyperSpyUI master.

This is because the plotting is currently broken for users using the latest mpl.

I am not too in favour of 2 because the reason why @vidartf doesn't release HyperSpyUI yet is that it suffers from a critical bug affecting window users.

francisco-dlp:
If we go for 2. (and even if we don't), what about moving the WP building code to https://github.com/hyperspy/hyperspy-bundle where it belongs? As discussed elsewhere, decoupling the bundle releases from HyperSpy's is advantageous. Actually, the bundle releases should happen automatically triggered by winpython releases and hyperspy releases. Only when things go wrong or we want to improve the bundle could we should release the bundle "manually". Thoughts?

ericpre:
Yes, I am for moving the bundle build to hyperspy-bundle, so I am going to remove all the bundle build code from hyperspy in this PR.
For the bundle release, we should definitely wait for the fixes in HyperSpyUI.

francisco-dlp:
In the interest of keeping the history tidy I think that it'll be better to revert the last commit, merge this and send another PR to remove the code from this repo.

ericpre:
Yes, most of the commits are now obsolete, would it better to close this PR and just make a new one to remove the relevant part?

francisco-dlp:
Sounds good!

ericpre:
I squashed the commits.

francisco-dlp:
Looks good, but does more than described. It'll be good to add a little description of the changes unrelated to moving WP to the bundle repo before merging.

ericpre:
Do you mean the `on_failure` part? I copied and paste the content of the appveyor from a upstream branch, I possibly took next minor by mistake. Shoujld I removed it and squash it to keep things simple?

francisco-dlp:
Yes, that's what I mean. Sounds good.

PR #1914, opened 2018-04-17
Support all arguments of plot_images in case of single image.

### LINKED ISSUE #1883: saturated_pixels doesn't work as expected in plot_images
I'm having issues with changing the contrast of an image when plotting using this line of code:
`hs.plot.plot_images(s, saturated_pixels=1)`
However when using: 
`s.plot(saturated_pixels=1)`
the plotted image reflects the correct pixel saturation

ericpre:
Fixes #1883.

### Progress of the PR
- [x] don't use the `plot` method of the signal for single image to support all arguments,
- [x] `saturated_pixels` follow similar pattern than vmin and vmax,
- [x] improve sanity check with the type of `images`.
- [x] ready for review.


francisco-dlp:
LGTM, except that a test would be nice.

ericpre:
When writing the tests, I did a bit more and I have updated the description of the PR accordingly.

PR #1915, opened 2018-04-18
Update user guide regarding traitsui supporting qt5 

ericpre:
Add a note in the user guide explaining that as of traitsui v6.0, qt5 is supported and the conflict issue in anaconda is resolved. 

### Progress of the PR
- [x] update user guide,
- [x] ready for review.


francisco-dlp:
I think that it is better to simply remove the note as this discussion does not belong to HyperSpy any more.

Issue #1677, opened 2017-06-28
.dm3/.dm4 reader should be advertised on website


thomasaarholt:
I've just discovered the existing dm-plugin for reading and writing hdf5: https://github.com/niermann/gms_plugin_hdf5/releases

It has previously been discussed here: #793 

It works quite well, but I believe it is only compiled for 64-bit DM at the moment. It would be nice if it could be advertised on the hyperspy website.

magnunor:
It should be compiled for 32-bit as well, judging by the files present in the release zip: 
- GMS-1.X (32 bit): hdf5_GMS1X_x86.dll
- GMS-2.X (32 bit): hdf5_GMS2X_x86.dll
- GMS-2.X (64 bit): hdf5_GMS2X_amd64.dll

Might want to hear what @niermann thinks about this.

thomasaarholt:
Ah, I didn't spot that part in the readme. I'll attempt again.

Edit: Yeah, works! Still doesn't read the dimensions in in the right order, but still pretty good!

ericpre:
If I remember correctly, you just need to transpose (`s.T` or something similar) the data before saving as hdf5/hspy file. DM can not transpose large dataset, it just crashes...

niermann:
As already discussed the binaries are provided in the release zip. Currently the plugin is just intended to provide methods (accessible by DM scripting) to query HDF5 files, and read datasets and attributes. Also a DM script is provided to read my own internal custom datasets. What's missing here is a DM script, which interprets the data/metadata stored in hyperspy files in a way DM will understand. I'm not (yet) using hyperspy, and also try to avoid DM scripting, so I have no intention to write such a script myself in the next future. However, I'm happy to help anyone, who is willing to write such a script.

thomasaarholt:
@niermann @francisco-dlp 
I've had a bit of a play, and I can confirm that the plugin works great with hyperspy. My previous errors were due to the order of the data dimensions and the fact that my data was in float64.

There are two requirements for the file to import correctly:

1. The data type must not be float64, but float32 (`s.change_dtype("float32")`)
2. The data must be transposed before saving (`s.T.save(filename)` works great)

The first point is only necessary if one wants to use the spectrum picker tool, for EELS and EDS, as far as I can tell.

Finally, I have modified a single string in the import script, to default to importing "*.hspy" files, rather than "*.hdf5". In case we want to put it on the website.
[...]

Issue #793, opened 2015-12-12
HDF5 reader plugin for DM


magnunor:
As far as I know there is no way to import HDF5-files directly into Digital Micrograph. It is possible to do it indirectly by converting it to raw binary or using the RPL-plugin https://github.com/hyperspy/ImportRPL . 

However, it would be nice if there was some way to directly load HDF5 into DM. 

Has anyone tried this? I'm guessing one way could be to compile the official HDF5 C++ reader into a DLL compatible with DM using the DM SDK.

Not certain if this is the right place to post this, but guessing it is relevant since it could reduce the barrier of using HyperSpy a little bit. An extra bonus would be making the NCEM EMD format loadable in DM as well...

thomasaarholt:
I'd definitely use this. Rpl is a bit annoying. Would also be nice since hdf5 is the standard we use for Hyperspy.

niermann:
Hi there!

As I'm also using HDF5 to store most of my python processed TEM data, I created a plugin for DigitalMicrograph some time ago. I've stumbled over this feature proposal here, thus I uploaded the plugin's source to GitHub (after polishing the plugin a bit):
https://github.com/niermann/gms_plugin_hdf5

The plugin registers a set of functions to DM's scripting engine, which provide access the HDF5 files from scripts. Within the repository in the tests/ directory there is a DM script, which can be installed into the DM menu to load HDF5 files. However, this script currently only interprets my own internal metadata, but might be easily adapted to be used with other HDF5 based formats.

Best regards,
Tore

niermann:
You're absolutely right, especially when you need a different visual studio environment for different versions of DM. Currently the DLLs are not publically available. I have compiled ones for DM versions 1.8 and 2.X (the latter 32 and 64 bit). As far as I'm informed the DM 2.X ones should also work for 3.X.  I will look into ways of making them somewhere downloadable...

magnunor:
How big are they? You could probably include them in your https://github.com/niermann/gms_plugin_hdf5 repository.

ericpre:
That's great! Looking forward the DLL! :)
Currently the GMS rpl plugin has a limitation: it can't import large dataset (> ~1 Go) and several Go datasets are becoming very common...

niermann:
I created a release in GitHub repository, which contains a zip file with all the binaries. Enjoy them!
https://github.com/niermann/gms_plugin_hdf5/releases

magnunor:
[...]

PR #1920, opened 2018-04-19
Move read scripts from web to UG

francisco-dlp:
Move http://hyperspy.org/download.html#related-software to User Guide. The intention is to remove the download section from the website and point users directly to the install section of the UG.


ericpre (commented):
If we remove the download section of the website, I think the installation section of the user guide will need to be rework a bit because on the website there is not as much information and it is more clear.

Also add the hdf5 plugin for DM mentioned in #793 and #1677?
Since raw/rpl doesn't work for large file (~>1 Go), DM crashes when transposing data, maybe it would be good to mention that this hf5 reader can be useful to transfer big file to DM. When I used it in the past I needed to have the data saved transposed: s.T.save('file.hspy') to have the data in DM with the correct shape.

PR #1924, opened 2018-04-20
Remove advice about the bundle

mwalls:
Since this is no longer the way to get the latest version (if I've understood) I don't think telling people to use the bundle is a good idea.
Incidentally on the "News" page, the links to the bug fixes and enhancements seem to be wrong. They imply that none of either have occurred in this release. I wanted to change that but could not find the file about release 1.3.1.
Maybe I'm jumping the gun and this is all in hand, My apologies if so...

### Requirements

### Description of the change
- Change A.
- Change B.
- etc.

### Progress of the PR
- [ ] add tests,
- [ ] ready for review.

```python
>>> import hyperspy.api as hs
>>> import numpy as np
>>> # Your new feature...
```

### DIFF
--- a/doc/user_guide/install.rst
+++ b/doc/user_guide/install.rst
@@ -4,8 +4,7 @@
 Installing HyperSpy
 ===================
 
-The easiest way to install HyperSpy in Microsoft Windows is installing the
-:ref:`HyperSpy Bundle <hyperspy-bundle>`.
+
 
 For quick instructions on how to install HyperSpy in Linux, MacOs or Windows
 using the `Anaconda Python distribution <http://docs.continuum.io/anaconda/>`_


ericpre:
They will a bundle released very soon, it needs one last tweak... and yes indeed the link in the website is broken! :/ Thanks for spotting the mistake.

mwalls:
Ok, but then I'm not sure I understand the statement: "Starting with this version, the HyperSpy WinPython Bundle distribution is no longer released in sync with HyperSpy." I may have missed a discussion about this. Will the bundle still always exist, but just a bit later? Is there some reason why the bundled version will now systematically be released after the unbundled? If so, the text on the "install.rst" page  should have at least a remark to that effect?

ericpre:
This is related to the hyperspy split and the hyperspy-extensions. The idea would be to release a bundle when it is needed: changes in the python scientific stack, hyperspy, hyperspyUI, or any other hyperspy-extension may required to release a bundle without requiring a hyperspy release.

mwalls:
OK I'm way out of date. But I think the text could be clearer for new and/or non-expert users. Just not sure I'm the person to write it. Maybe I'll have a talk to Francisco...

ericpre:
The bundle have been released, see:
https://github.com/hyperspy/hyperspy-bundle/releases

ericpre:
Closing since this have been clarified. Reopen if this is still not clear.

PR #1925, opened 2018-04-21
New Argand diagram for complex signals

woozey:
### Description of the change
Added `ComplexSignal.argand_diagram` method that generates Argand plot for complex data.
### Progress of the PR
- [x] Change implemented,
- [x] update docstring,
- [x] update user guide,
- [x] add tests,
- [x] ready for review.

```python
>>> import hyperspy.api as hs
>>> holo = hs.datasets.example_signals.object_hologram()
>>> w = holo.reconstruct_phase(ref)
>>> w.argand_diagram(display_range=[-3, 3]).plot()
```
screen shot 2018-04-21 at 14 38 28


hyperspy/_signals/complex_signal.py
```diff
@@ -20,8 +20,10 @@
 
 import numpy as np
 import dask.array as da
+import matplotlib.pyplot as plt
```
thomasaarholt: You don't currently need this import :)

woozey:
@thomasaarholt seem to be finally ready for the revision))

doc/user_guide/tools.rst
```diff
@@ -1660,6 +1660,25 @@ underlying method is :py:func:`~skimage.restoration.unwrap`, which uses the
 algorithm described in :ref:`[Herraez] <Herraez>`.
 
 
+.. _complex.argand:
+
+Calculate and display Argand diagram
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+Sometimes it is convenient to visualize a complex signal as a plot of its
+imaginary part versus real one. In this case so called Argand diagram can be
```
thomasaarholt: spelling: "so called Argand diagrams" (plural diagrams)

doc/user_guide/tools.rst
```diff
@@ -1660,6 +1660,25 @@ underlying method is :py:func:`~skimage.restoration.unwrap`, which uses the
...
+calculated using :py:func:`~hyperspy.signals.ComplexSignal.argand_diagram`
```
thomasaarholt: ComplexSignal2D (see next comment)

hyperspy/_signals/complex_signal.py
```diff
@@ -262,15 +262,12 @@ def angle(self, deg=False):
         return super().angle(angle, deg=deg)
     angle.__doc__ = ComplexSignal_mixin.angle.__doc__
 
-    def argand_diagram(self, plot=True, size=[256, 256], display_range=None):
+    def argand_diagram(self, size=[256, 256], display_range=None):
```
thomasaarholt: I think this should be a ComplexSignal2D method and not a general complexsignal? Or does it work for 1D signals as well? (2D and 1D ComplexSignals inherit the general ComplexSignal methods)

hyperspy/_signals/complex_signal.py
```diff
@@ -260,6 +261,59 @@ def angle(self, deg=False):
...
+        re = self.real.data.ravel()
+
+        if display_range:
+            if np.asarray(display_range).shape == (2, ):
+                display_range = [[display_range[0], display_range[1]],
+                                 [display_range[0], display_range[1]]]
+            elif np.asarray(display_range).shape != (2, 2):
+                raise ValueError('display_range should be array_like, shape(2,2) or shape(2,).')
+
+        ap = np.histogram2d(re, im, bins=size, range=display_range)
```
thomasaarholt: I'd recommend turning this into `ap, xedges, yedges = np.histogram2d(re, im, bins=size, range=display_range)` and then in the next line use `sap = Signal2D(ap.T)`. That makes it clearer that the object in `ap[0]` isn't of the same kind as `ap[1]` etc. What does `ap` stand for? You might want to give it a better name!

hyperspy/_signals/complex_signal.py
```diff
@@ -260,6 +261,59 @@ def angle(self, deg=False):
...
+        sap = Signal2D(ap[0].T)
+        sap.metadata = self.metadata.deepcopy()
+        sap.metadata.General.title = 'Argand diagram of {}'.format(self.metadata.General.title)
+        sap.metadata.Signal.signal_type = 'signal2d'
```
thomasaarholt: This should be blank, I think. I think signal_type is kindof deprecated since we have class names that cover the same use. @ericpre @francisco-dlp do you have thoughts on this?

hyperspy/_signals/complex_signal.py
```diff
@@ -260,6 +261,59 @@ def angle(self, deg=False):
...
+        sap.axes_manager.signal_axes[0].name = 'Real'
+        sap.axes_manager.signal_axes[0].offset = ap[1][0]
```
thomasaarholt: If you take into account the change suggested above, then change ap[1] here and in the next line to xedges

hyperspy/_signals/complex_signal.py
```diff
@@ -260,6 +261,59 @@ def angle(self, deg=False):
...
+        sap.axes_manager.signal_axes[0].scale = np.abs(ap[1][0] - ap[1][1])
+        sap.axes_manager.signal_axes[1].name = 'Imaginary'
+        sap.axes_manager.signal_axes[1].offset = ap[2][0]
```
thomasaarholt: And the same here and next line to yedges

hyperspy/_signals/complex_signal.py
ericpre: From a quick look, it seems that this function works fine for 1D and 2D signal but it does not take into account the navigation dimension.

hyperspy/_signals/complex_signal.py
ericpre: `real_axis_edges` and `imag_axis_edges` (or something along these line) would be clearer than `xedges` and `yedges`, which are a bit confusing since once can be mislead with `x` and `y` referring to the x, y axes of the original image.

hyperspy/_signals/complex_signal.py
ericpre: Yes, I don't think this should be used here.

hyperspy/_signals/complex_signal.py
ericpre: Set the units too when they are available?

hyperspy/_signals/complex_signal.py
```diff
@@ -260,6 +261,59 @@ def angle(self, deg=False):
         return super().angle(angle, deg=deg)
     angle.__doc__ = ComplexSignal_mixin.angle.__doc__
 
+    def argand_diagram(self, size=[256, 256], display_range=None):
```
ericpre: Instead of `display_range`, I would suggest to use `range`. Using _display_ implies that this argument is intented to be used to adjust the display of something, while here it actually defines how the output signal is calculated. And this would make it consistent with the corresponding one of `np.histogram2d`.

hyperspy/_signals/complex_signal.py
woozey: @thomasaarholt  It does work for signals with any dimensionality as it is just plotting of  data on Im(Re) axis.
@ericpre This is also correct it creates a diagram for the whole dataset at once rather than for slices in navigation dimensions.

ericpre (commented):
Other than a few comments (navigation dimension, units and naming), this looks good to me. Nice documentation and tests!

hyperspy/_signals/complex_signal.py
ericpre: Ok, I would have expected some array reshaping but it may not be necessary, if everything is ordered properly when creating the `Signal2D` of the _argand_ signal.
If this is the case, it would be good to add a test checking the navigation axes (and their attributes) are correct, etc.
Complex signal can take a bit of space into memory, is it worth adding lazy processing?

hyperspy/_signals/complex_signal.py
thomasaarholt: Agreed with ericpre. I just copied the numpy docstring syntax.

hyperspy/_signals/complex_signal.py
woozey: @ericpre Perhaps it would make sense adding possibility to apply the function to the whole dataset as well as to iterate it along navigation dimensions, what would you say?

Indeed, Lazy is missing. Will add it asap.

hyperspy/_signals/complex_signal.py
woozey: Agreed in principle, but I don't like using `range` since it shadows python built-in function. Do you think it is fine?

hyperspy/_signals/complex_signal.py
woozey: @ericpre With units, I'm not sure. These are just _Im_ and _Re_ parts of complex numbers which are kind of unitless by definition. On the other hand one may think of applying the same units as the quantity of original signal, though I'm not sure if that is correct neither. Moreover I can't remember any complex value that would also have units as having an imaginary part would mean that it is something that doesn't exist physically. Correct me if I'm wrong. 
Regardless your answer, I will try to ask smart people around me to get better understanding.

ericpre: I don't know/can't think of a case where it would be used but I would imagine that it would exist!
In any case, from a purely generic point of view, we could extract the units from `s.real.metadata.Signal.quantity` and `s.imag.metadata.Signal.quantity`, since it should in bracket of this metadata.
Similarly, the name of the axis could be pull from there, when available.

hyperspy/_signals/complex_signal.py
ericpre: I think this is fine as it is (providing that the tests with navigation dimension are OK). If this is not necessary, I don't see why we add the option to iterate on navigation axis.

hyperspy/_signals/complex_signal.py
ericpre: As far as I know, this is not a problem and https://docs.scipy.org/doc/numpy/reference/generated/numpy.histogram.html uses `range` as argument, in my opinion, it would be logical to follow a similar convention here.

thomasaarholt: I'm tempted to suggest "window_limits" or "limits" instead, since `range` hints at a different type of iterable.

hyperspy/_signals/complex_signal.py
woozey: Sounds good! I had a thought about complex number in physics and the most obvious one is the electrical impedance which have units of Ohm. In this case the units for each of Re and Im parts are also Ohms, which is logical since the complex number is a sum of these two. So your suggestion should do.

woozey: @ericpre It appeared to be more complex (since we are dealing with complex numbers))) than I thought. The issue is the absence of a standard for using `quantity` attribute. I approached it voluntarily by parsing the `quantity` if it is available which isn't a robust solution against non-trivial use cases.  Please have a look and let me know if it makes sense or if you would have better suggestions. I'll add corresponding test once it is decided how this part will look like.

woozey: e.g.
```python
>>> from hyperspy.misc.utils import parse_quantity
>>> parse_quantity('Resistance (Ohm)')
('Resistance ', 'Ohm')

hyperspy/_signals/complex_signal.py
woozey: Sorry for the delay. It isn't straight forward to implement it for `dask` arrays. I'm still looking at how it can be done, but not very certain if I will manage

ericpre: No worries, it was a suggestion, we can merge it without lazy support if you prefer.
Indeed, there is no histogram2d, nor histogramdd in dask, but this https://github.com/dask/dask/issues/2911#issuecomment-422401799 may be useful.

hyperspy/_signals/complex_signal.py
ericpre: Just to clarify: `signal_type` is NOT deprecated but setting it is not necessary here. `signal_type` is actually quite important to define signal, see http://hyperspy.org/hyperspy-doc/current/dev_guide/writing_extensions.html#registering-a-new-basesignal-subclass.

thomasaarholt: Gotcha. That's actually quite clear. I remember being a bit confused about how `signal_type` and the class type seemed to do the same thing, but I see here that `signal_type` tells Hyperspy which class to use. Thanks.

hyperspy/_signals/complex_signal.py
ericpre: I open a PR (https://github.com/woozey/hyperspy/pull/35) to this branch to tweak the parsing of the quantity metadata and I think the approach should be good now. Please let me know what you think.

hyperspy/_signals/complex_signal.py
woozey: @ericpre Lazy now would rise `NotImplementedError`, and the test would checking it

hyperspy/_signals/complex_signal.py
woozey: @ericpre Added test for units assignment. Should be ready now

Issue #1712, opened 2017-08-18
Strange behaviour of `map` with `np.log`


woozey:
Simple example:
>sss = hs.signals.Signal2D(np.arange(9).reshape(3,3))
sss.map(np.log)

>---------------------------------------------------------------------------
>ValueError                                Traceback (most recent call last)
<ipython-input-12-00b966ccb30a> in <module>()
----> 1 sss.map(np.log)

>/usr/local/Library/Frameworks/Python.framework/Versions/3.6/lib/python3.6/site-packages/hyperspy-1.3-py3.6-macosx-10.10-x86_64.egg/hyperspy/signal.py in map(self, function, show_progressbar, parallel, inplace, ragged, **kwargs)
   3506         # iterate over the coordinates.
   3507         try:
-> 3508             fargs = inspect.signature(function).parameters.keys()
   3509         except TypeError:
   3510             # This is probably a Cython function that is not supported by

>/usr/local/Library/Frameworks/Python.framework/Versions/3.6/lib/python3.6/inspect.py in signature(obj, follow_wrapped)
   2999 def signature(obj, *, follow_wrapped=True):
   3000     """Get a signature object for the passed callable."""
-> 3001     return Signature.from_callable(obj, follow_wrapped=follow_wrapped)
   3002 
   3003 

>/usr/local/Library/Frameworks/Python.framework/Versions/3.6/lib/python3.6/inspect.py in from_callable(cls, obj, follow_wrapped)
   2749         """Constructs Signature for the given callable object."""
   2750         return _signature_from_callable(obj, sigcls=cls,
-> 2751                                         follow_wrapper_chains=follow_wrapped)
   2752 
   2753     @property

>/usr/local/Library/Frameworks/Python.framework/Versions/3.6/lib/python3.6/inspect.py in _signature_from_callable(obj, follow_wrapper_chains, skip_bound_arg, sigcls)
   2336         raise ValueError(msg)
   2337 
-> 2338     raise ValueError('callable {!r} is not supported by signature'.format(obj))
   2339 
   2340 

>ValueError: callable <ufunc 'log'> is not supported by signature

The same behaviour I see also for `np.abs`.
Running freshly installed:
>hyperspy                  1.3                       (pip)
hyperspy-gui-ipywidgets   1.0.0                     (pip)

under MacOSX in Python 3.6.1
Any ideas what's wrong?

francisco-dlp:
[...]

PR #1933, opened 2018-04-27
New `BaseSignal.apodization()`

woozey:
### Description of the change
Introduces apodization for all signals. Also apodization will be an option wen FFT is computed, which allows removing streaking.
FYI: @thomasaarholt, @jan-car 

- Add `apodization` method in `BaseSignal`
- Add optional `apodization` argument to `BaseSignal.fft()`

### Progress of the PR
- [x] Change implemented:
   - [x] `apodization` method in `BaseSignal`:
      - [x] Hann or hanning window;
      - [x] Hamming window;
      - [x] Tukey window;
      - [x] high order Hann window;
   - [x] `apodization` argument in `fft()`.
- [x] update docstring,
- [x] add tests,
- [x] ready for review.

```python
>>> import hyperspy.api as hs
>>> import numpy as np
>>> fft_pp = np.log(ref.fft(shifted=True).amplitude)
>>> fft_pp_a = np.log(ref.apply_apodization().fft(shifted=True).amplitude)
>>> hs.plot.plot_images([fft_pp, fft_pp_a], tight_layout=True)
```
screen shot 2018-04-27 at 23 49 39


hyperspy/misc/math_tools.py
```diff
@@ -69,3 +69,42 @@ def anyfloatin(things):
...
+        number of points in window
+    order : int
+        Filter order
+
+    Return
+    ------
+    window : array
+        window
+    """
+    sin_arg = np.pi * (m - 1.) / m
```
thomasaarholt, line 109: I suggest adding an assert for checking m and order are ints

hyperspy/misc/math_tools.py
```diff
@@ -69,3 +69,42 @@ def anyfloatin(things):
...
+    cos_arg = 2. * np.pi / (m - 1.) * (np.arange(m))
+    window = 0.
```
thomasaarholt: Just a suggestion, feel free to ignore if you think it makes things unclear:
You can do away with the window = 0 and the loop by using
`window = sum([(-1) ** i / i * np.sin(i * sin_arg) * (np.cos(i * cos_arg) - 1) for i in range(1, order+1)])`

hyperspy/misc/math_tools.py
```diff
@@ -69,3 +69,42 @@ def anyfloatin(things):
...
+
+
+def hann_window_nth_order(m, order):
+    """
+    Calculates 1D Hann window of nth order
+
+    Parameter
+    ---------
+    m : int
+        number of points in window
```
thomasaarholt: ... of points in window (typically the length of a signal)

hyperspy/signal.py
```diff
@@ -3213,20 +3218,27 @@ def fft(self, shifted=False, **kwargs):
 
         if self.axes_manager.signal_dimension == 0:
             raise AttributeError("Signal dimension must be at least one.")
+        if apodization==True:
+            apodization = 'hann'
+
+        if apodization:
+            im_fft = self.deepcopy().apply_apodization(type=apodization)
+        else:
+            im_fft = self.deepcopy()
```
thomasaarholt: `im_fft = self` - no need to make an extra copy of a (potentially) large signal into memory here. It's already copied further down. The apodization=True case is correct though.

hyperspy/signal.py
```diff
@@ -4794,6 +4806,78 @@ def T(self):
...
+    def apply_apodization(self, type='hann', hann_order=None, tukey_alpha=0.5, inplace=False):
+        """
+        Apply apodization window to a signal either in place or return a new signal.
+
+        Parameters
+        ----------
+        type : string, optional
+            Select between 'hann', 'hamming', 'tukey'
+            (Default: 'hann')
+        hann_order : None or int, optional
```
thomasaarholt, line 4944: What does hann_order=None mean if hann window is used? 

Also, add a line of "only used if `window='hann'`

hyperspy/signal.py
```diff
@@ -4794,6 +4806,78 @@ def T(self):
         """
         return self.transpose()
 
+    def apply_apodization(self, type='hann', hann_order=None, tukey_alpha=0.5, inplace=False):
```
thomasaarholt: Maybe change `type` to `window` or `window_type`? `type` is typically reserved.

hyperspy/signal.py
```diff
@@ -4794,6 +4806,78 @@ def T(self):
...
+        elif type == 'tukey':
+            window_function = lambda m: sp.signal.tukey(m, tukey_alpha)
+        else:
+            raise ValueError('Wrong type parameter value.')
+
+        windows_1d = []
+
+        axes = np.array(self.axes_manager.signal_indices_in_array)
+
+        for axis, axis_index in zip(self.axes_manager.signal_axes, axes):
```
thomasaarholt, line 4987: I can't comment on this part. It would be great if someone else could give feedback on dask.

hyperspy/tests/signal/test_apodiztion.py
```diff
@@ -0,0 +1,63 @@
+# Copyright 2007-2016 The Hyperspy developers
```
thomasaarholt: typo in file name of this file

hyperspy/signal.py
```diff
@@ -4794,6 +4806,78 @@ def T(self):
...
+                windows_1d.append(window_da)
+            else:
+                windows_1d.append(window_function(axis.size))
+
+        window_nd = outer_nd(*windows_1d)
+
+        if inplace:
+            slice_w = []
+
+            for i in range(self.data.ndim):
```
thomasaarholt, line 5003: some comment on what exactly is happening here would be useful for the future

hyperspy/tests/signal/test_apodiztion.py
```diff
@@ -0,0 +1,63 @@
...
+    parameters = []
+    for lazy in [False, True]:
+        for window_type in ['hann', 'hamming']:
+            parameters.append([lazy,
+                               window_type])
+    return parameters
+
+
+@pytest.mark.parametrize('lazy, window_type', _generate_parameters())
+def test_apodization(lazy, window_type):
```
thomasaarholt: test looks good. Can we add something similar for the fft case? Using the holo dataset

thomasaarholt (commented):
I've added some comments.

woozey:
Ready for review))

hyperspy/signal.py
```diff
@@ -4920,6 +4932,84 @@ def T(self):
...
+        elif window == 'hamming':
+            window_function = lambda m: np.hamming(m)
+        elif window == 'tukey':
+            window_function = lambda m: sp.signal.tukey(m, tukey_alpha)
+        else:
+            raise ValueError('Wrong type parameter value.')
+
+        windows_1d = []
+
+        axes = np.array(self.axes_manager.signal_indices_in_array)
```
ericpre, line 4985: Would it be more sensible to use the `map` method to do the computation?

ericpre (commented):
Sorry for coming back to this after a long time... I have one comment regarding the use of the `map` method to perform the computation but other than that it looks good to me.

hyperspy/signal.py
woozey: I don't think it is applicable here... This part constructs nd window function from 1d, which involves creating list of 1d window functions corresponding to the length of each signal axis and then doing outer product. 

The multiplication of each slice with the window is done later at the end (lines 4999 to 5012), and this is actually the part where map functions could have been used... I don't remember the exact reason I've done it manually, perhaps wanted to reduce possible overheads that map introduces with changing dimensionality and etc... Let me know If you would prefer using map there. If needed I can even try benchmarking both options in terms of performance.

woozey: I suspect this one could have been actual reason: #1712

hyperspy/signal.py
```diff
@@ -3299,19 +3304,26 @@ def fft(self, shift=False, **kwargs):
 
         if self.axes_manager.signal_dimension == 0:
             raise AttributeError("Signal dimension must be at least one.")
+        if apodization==True:
+            apodization = 'hann'
+
+        if apodization:
+            im_fft = self.deepcopy().apply_apodization(window=apodization)
```
```suggestion
            im_fft = self.apply_apodization(window=apodization, inplace=False)
```
No need for a `deepcopy` here if you don't use `inplace=True`.

hyperspy/signal.py
```diff
@@ -4920,6 +4932,84 @@ def T(self):
...
+
+            # Iterate over all dimensions of the data
+            for i in range(self.data.ndim):
+                if any(i == axes):  # If current dimension represents one of signal axis, all elements in window
+                    # along current axis to be subscribed
+                    slice_w.append(slice(None))
+                else:  # If current dimension is navigation one, new axis is absent in window and should be created
+                    slice_w.append(None)
+
+            self.data = self.data * window_nd[tuple(slice_w)]
```
ericpre, line 5010:
```suggestion
            self.data = self.data * window_nd[tuple(slice_w)]
            self.events.data_changed.trigger(obj=self)
```
Trigger an event when data is changed in place.

hyperspy/signal.py
```diff
@@ -4920,6 +4932,84 @@ def T(self):
...
+        else:
+            return self.deepcopy() * window_nd
```
ericpre: I don't think it needs a `deepcopy` here.

hyperspy/signal.py
ericpre: Yes, I meant the computation using `self.data` not for the window. I suggested to use `map` because if should be easier to maintain. I don't know if there will be any speed penalty, but normally if it goes through `_map_all` there should not be any significant overheads. But the calculation is fairly simple, so it is not a big deal not to use map.
#1712 is for numpy ufuncs, which operate element wise on the inputs which should not be the case here.

woozey: > 

> #1712 is for numpy ufuncs, which operate element wise on the inputs which should not be the case here.

It turns out that `np.multiply` is also ufunc:
>np.multiply(*args, **kwargs)
Type:            ufunc
String form:     <ufunc 'multiply'>

So I got the same error...

hyperspy/signal.py
woozey: indeed, will change it in a moment

hyperspy/signal.py
ericpre: I was more thinking of defining a function which takes an axes parameter and pass this function to map. But that's ok, we can leave it as it is.

woozey: If I understood you correctly, in this case one would have to recalculate the window each time, which can be quite inefficient.

ericpre: The windows could be a parameter of the function but looking at it now, I agree that using numpy slicing is more simple than going through the `map` hoops! :)
I will wait for CI to finish to check there was not side effect from the last changes and I will merge it.

Issue #1599, opened 2017-04-20
Splitting HyperSpy progress


francisco-dlp:
Issue to track HyperSpy splitting progress. Original discussion @  #821.

* [x] GUI toolkit register - ipywidgets [WIP in #1544] 
    * [x] GUI - ipywidgets #1544]
    * [x] GUI - traitsui #1544
* [x] Split the IO features into their own package, RosettaScIO. See for the implementation ~#2174~ #2972 and #1978 for the discussion.
* [x] Register external components, models and signals. See #2180.
* [ ] Split the electron microscopy features into their own package. ~(Waiting for #2180 to be ready.)~
* [ ] Merge as many pending PRs as possible before the split.

francisco-dlp:
Splitting the signals into packages simply requires a mechanism to register new signals and components similar to the one in #1544 that registers UI toolkit. Once such a method is ready, we could reorganise the code of the specialise signals into separate folders in the hyperspy root folder for the 1.x series, so that releasing v2.0 would simply consist on moving those folder to their own repositories inside the hyperspy organisation umbrella.

thomasaarholt:
Could someone propose how one would install hyperspy after it has been split? I'm wondering how user-friendly conda will be with this.

Say, if I wanted (hypothetically) the basic hyperspy module, the imaging module, the ipywidgets gui and the eels module. How would I install this?

francisco-dlp:
Does the following look reasonable?

```bash

conda install hyperspy-core hyperspy-visualization hyperspy-gui-ipywidgets hyperspy-tem-eels

```

thomasaarholt:
That looks good. I assume hyperspy-visualization (etc) would depend on hyperspy-core.

Are there any examples of other code that is split in this sense? There may be lessons worth learning from other libraries. I guess jupyter is a good example of this.

thomasaarholt:
I'm also missing an idea of how the io-plugin register will be handled. Do we have a clear plan of how this should be implemented?

- A method for installing the io plugins separately
- Preferences support for turning plugins on and off (?)

I assume creating individual modules for each plugin is the way forward, although this has the potential to get pretty messy?

Is there a way of creating a "package deal" for instance for installing hyperspy-core, -visualization, -tem-eels and all the io plugins? I fear that having to type out the `conda install` for each plugin would create a tedious installation procedure and a barrier to future users.

magnunor:
[...]

PR #1943, opened 2018-05-10
Reading of binary Merlin files

magnunor:
The Medipix3 is a fairly new type of detector (at least in the TEM world), and the software used to interface with the detector (Merlin) output Merlin binary files (.mib).

Previously, we've been using `fpd.fpd_file` to convert the data from binary to HDF5. However using `np.memmap`, reshape and slicing it turned out to be pretty easy to read the binary files directly.

This pull request implements a reader for this filetype.

### Progress of the PR
- [ ]  Make it possible to use without specifying `probe_x` and `probe_y`: it should load the data with 1 navigation dimension
- [ ]  Add support for non-lazy loading
- [ ]  Add option for specifying flyback pixels
- [ ]  Add tests (need to get some datasets, which can be zipped to greatly reduce the file size)
- [ ]  Add info to user guide
- [ ]  Ready for review

To test this:
```python
>>> import hyperspy.api as hs
>>> from hyperspy.io_plugins.merlin import file_reader
>>> file_dict = file_reader("dataset.mib", probe_x=64, probe_y=64, lazy=True)[0]
>>> s = hs.signals.Signal2D(**file_dict).as_lazy()
```

### DIFF
--- a/hyperspy/io_plugins/merlin.py
+++ b/hyperspy/io_plugins/merlin.py
@@ -1,0 +1,114 @@
+# -*- coding: utf-8 -*-
+# Copyright 2007-2018 The HyperSpy developers
+#
+# This file is part of  HyperSpy.
+#
+#  HyperSpy is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+#  HyperSpy is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with  HyperSpy.  If not, see <http://www.gnu.org/licenses/>.
+
+import logging
+import numpy as np
+import dask.array as da
+
+
+_logger = logging.getLogger(__name__)
+# Plugin characteristics
+# ----------------------
+format_name = 'Merlin binary'
+description = 'Read data from Merlin binary files'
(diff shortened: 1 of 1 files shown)


M0hsend:
Hi Magnus, 
Just FYI we have a lazy mib importer jointly put together with @dnjohnstone and now part of pyxem (https://github.com/pyxem/pyxem/blob/master/pyxem/io_plugins/mib.py). It would probably make sense to have a unified mib importer for hyperspy and pyxem. 
What do you guys think?

francisco-dlp:
This is the main challenge of splitting HyperSpy: how to avoid that features (like the mib reader in pyxem) that would be useful to general HyperSpy users are developed as part of other packages based on HyperSpy and not in HyperSpy itself. As the extensions community will naturally be more segregated, the danger of reinventing the wheel (such in this case) and losing feedback from a wider community are non-negligible.

For the particular case of IO features, a solution may be to create a package to provide support for all the EM file formats. What do you think? Until the split happens, I think that it'll make sense to merge this effort with pyxem's reader as @M0hsend suggests.

magnunor:
I was not aware of the mib reader in pyxem! Luckily I didn't spend too much time on this pull request :)

It would be nice to "upstream" that functionality to either HyperSpy or some kind of EM file IO library. This library could include HyperSpy's io_plugins + the mib one. Ideally, it would only return numpy objects and dicts (as they currently do), which would mean potentially other libraries could also use the functionality.

tjof2:
Closing for now since the pyxem reader is described in https://github.com/hyperspy/hyperspy/issues/1599#issuecomment-407363252 as being "more complete" and the consensus seems to be that it should eventually go in the separate IO library anyway.

dnjohnstone:
Agree - we should probably chat to @matkraj about the best solution here at some point too.

PR #1944, opened 2018-05-12
Fix compatibility with dask 0.17.4

ericpre:
Tests are failing since last dask release (0.17.4). See PR #https://github.com/dask/dask/pull/3429 for the detailed changes.

### Progress of the PR
- [x] builtins has been removed from dask.numpy_compat: use builtins `abs` instead of calling it throught dask,
- [x] dask bumped numpy version to 1.11 so we need to do the same,
- [x] ready for review.

### DIFF
--- a/hyperspy/_components/gaussian.py
+++ b/hyperspy/_components/gaussian.py
@@ -49,7 +49,7 @@ def _estimate_gaussian_parameters(signal, x1, x2, only_current):
     if isinstance(data, da.Array):
         _sum = da.sum
         _sqrt = da.sqrt
-        _abs = da.numpy_compat.builtins.abs
+        _abs = abs
     else:
         _sum = np.sum
         _sqrt = np.sqrt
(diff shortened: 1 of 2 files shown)


setup.py
```diff
@@ -48,7 +47,7 @@
 
 install_req = ['scipy>=0.15',
                'matplotlib>=1.2, !=2.1.0, !=2.1.1',
-               'numpy>=1.10, !=1.13.0',
+               'numpy>=1.11, !=1.13.0',
```
francisco-dlp: Why is this needed? It's ``dask`` who requires v1.11 not HyperSpy.

francisco-dlp:
Why not sending this to  Rnp?

setup.py
ericpre: Yes, indeed!

ericpre:
Since the next release is going to be minor, I was thinking it would more simple to work on one branch only and to avoid merging next patch to next minor regularly? If you think it is still worth sticking to the normal workflow I will rebase on the patch branch.

francisco-dlp:
Releases not always goes to plan e.g. we may find fix a serious issue before v1.4 is ready and decide to make a patch release, so I think that it'll be better resubmit to Rnp.

PR #1945, opened 2018-05-12
OneView_IS Reader

M0hsend:
Adding a function to io.py to read Gatan OneView_IS datasets.

OneView_IS (in-situ) datasets are saved as nested folders with the following format:
`Example_OneView_IS/Hour_00/Minute_00/Second_00/Example_OneView_IS_Hour_00_Minute_00_Second_00_Frame_0000.dm4`
In each `Second_xx` folder as many frames taken per second are populated. The camera captures continuously in this mode. 

files changed:
```
hyperspy/api_nogui.py
hyperspy/io.py
```

This is a first and very basic attempt. `oneview_is_reader` gets the folder location of the dataset and loads the data as a `Signal2D` image stack lazily. As these datasets are generally large only lazy loading is performed.

Sharing to get feedback form watchers as to where to take this.

### Progress of the PR
- [x] Initial changes implemented,
- [ ] Accommodate K2-IS datasets,
- [ ] update docstring,
- [ ] add tests,
- [ ] ready for review.

```python
>>> import hyperspy.api as hs
>>> s = hs.oneview_is_reader('/Documents/data/OneView_IS_example')

### DIFF
--- a/hyperspy/io.py
+++ b/hyperspy/io.py
@@ -266,7 +266,8 @@ def load(filenames=None,
                        for filename in filenames]
 
         if len(objects) == 1:
-            objects = objects[0]
+            objects = objects[0]    
+            
     return objects
 
 
@@ -337,6 +338,50 @@ def load_with_reader(filename,
     if len(objects) == 1:
         objects = objects[0]
     return objects
+
+
+def oneview_is_reader(folder_path):
+    """
+    Imports OneView_IS datasets.
+    These datasets are saved as nested folders with the format:
+    'Dataset/Hour_00/Minute_00/Second_00/*.dm4' etc.
+    Function parameters:
+     
+    folder_path (string): the directory path of the Dataset folder
+    ----------
+    Retuns:
+    imageStack: Hypespy Signal2D lazy stack of the frames, ordered sequentially
+                    The image axes parameters are copied over from first frame.
+                    The thrid axis is defines as 't', scaled with the camera
+                    exposure time, read from the first frame metadata.
+                    OneView_IS mode captures the frames continuously.
+    """
+     
+    os.chdir(folder_path)
+     
+    filenames_sort = []
+    for root, dirs, files in os.walk(folder_path):
+        for file in files:
+            if file.endswith('.dm4') or file.endswith('.dm3'):
+                filenames_sort.append(os.path.join(root, file))
+    
+    filenames_sort.sort()
+
+    #This works due to the naming format of these datasets.
+    #May need changing if to cover other formats
+     
+    imageStack = load(filenames_sort, lazy = True, stack = True)
+     
+    imageStack.axes_manager[0].name = 't'
+    imageStack.axes_manager[0].units = 'sec'
+    imageStack.axes_manager[0].scale = imageStack.inav[0].metadata.\
+            Acquisition_instrument.TEM.Camera.exposure
+     
+    imageStack.axes_manager['t'].index = 0 
+    #For some reason the index is changed to 211! Resettting to 0 above.
+     
+    return imageStack
+
 
 
 def assign_signal_subclass(dtype,
--- a/hyperspy/api_nogui.py
+++ b/hyperspy/api_nogui.py
@@ -74,6 +74,7 @@ del docstrings
 from hyperspy.Release import version as __version__
 from hyperspy import signals
 from hyperspy.io import load
+from hyperspy.io import oneview_is_reader
 from hyperspy.defaults_parser import preferences
 from hyperspy.utils import *
 from hyperspy import datasets

### CODE REVIEW

**Comment by magnunor** [comment]
Very nice!

Normally the things related to loading specific file formats go in the [`io_plugins`](https://github.com/hyperspy/hyperspy/tree/RELEASE_next_minor/hyperspy/io_plugins) folder. However, I'm not exactly certain how to handle this "format" since it is basically a folder with DM-files...

So I think it should go somewhere in that folder, possibly in [`digital_micrograph.py`](https://github.com/hyperspy/hyperspy/blob/RELEASE_next_minor/hyperspy/io_plugins/digital_micrograph.py).

The docstring should follow the NumPy docstring format, for example as seen in [`hs.load`](https://github.com/hyperspy/hyperspy/blob/RELEASE_next_minor/hyperspy/io.py#L54)

**Comment by M0hsend** [comment]
Many thanks @magnunor !
I had a brief chat with @francisco-dlp on where this should go to. He was also agreeing that `io_plugins` is not the right place, since this is not a new file format. Maybe writing a more general "structured folder" reader would be more suitable? Are there any other dataset types like this besides OneView_IS and K2_IS?

File: hyperspy/io.py [hunk 1]
```diff
@@ -339,6 +340,50 @@ def load_with_reader(filename,
...
+                filenames_sort.append(os.path.join(root, file))
+    
+    filenames_sort.sort()
+
+    #This works due to the naming format of these datasets.
+    #May need changing if to cover other formats
+     
+    imageStack = load(filenames_sort, lazy = True, stack = True)
+     
+    imageStack.axes_manager[0].name = 't'
```
ericpre, line 375: Name axis explicitly as `time`, there is no need for abbreviation

hyperspy/io.py
```diff
@@ -339,6 +340,50 @@ def load_with_reader(filename,
...
+    imageStack.axes_manager[0].units = 'sec'
```
ericpre, line 376: I would prefer `s` rather than `sec` to be consistent with the formatting of the pint library, which we use elsewhere.

hyperspy/io.py
```diff
@@ -339,6 +340,50 @@ def load_with_reader(filename,
...
+    imageStack.axes_manager[0].scale = imageStack.inav[0].metadata.\
```
ericpre, line 377: Is there a need for `inav[0]`? Is it (always) correct that the scale of this axis matches the exposure time?

hyperspy/io.py
```diff
@@ -339,6 +340,50 @@ def load_with_reader(filename,
...
+        for file in files:
+            if file.endswith('.dm4') or file.endswith('.dm3'):
+                filenames_sort.append(os.path.join(root, file))
+    
+    filenames_sort.sort()
+
+    #This works due to the naming format of these datasets.
+    #May need changing if to cover other formats
+     
+    imageStack = load(filenames_sort, lazy = True, stack = True)
```
ericpre, line 373: `lazy=True` should be an option with a default to `True` and not hard coded, to allow the user to load the data to memory.
Also since the default in hyperspy is to load the data to memory, I would suggest to inform the user using the logger at the info level (`_logger.info()`).

ericpre (commented):
Maybe it would be better to put this in the dm reader, because making it more generic for other type of data which follow a well defined structure may be making it too complicated.
Instead of creating a new function, I would just pass a folder to the `load` function, such as `hs.load("folder")`.
Loading a folder is currently not supported in HyperSpy, so with a bit of sanity check, the `load` function should be able to understand that it needs to load an Gatan IS type data set?

Regarding tests, your could for example, copy and paste an already existing file from the test suite in the correct file structure and check that the stack contains all the data.

francisco-dlp:
Some quick comments:

1. In its current implementation, wouldn't 
    ```python
    >>> s = hs.oneview_is_reader('/Documents/data/OneView_IS_example')
    ```
    
    be roughly equivalent to
    
    ```python
    >>> s = hs.load('/Documents/data/OneView_IS_example/*/*/*/*.dm4', stack=True, lazy=True)
    ```
    except for the axis calibration step?
    If yes, is it worth adding a function for this (vs e.g. documenting how to perform the task in the UG)?

2. What about parsing the time information from the file name in order to generate the axis parameters from it? This would prevent issues arising from assumptions that may not always be right for this or future versions of the software. Other source of issues could be users who perform analysis on the data using DM and store the files inside the structured folder. In its current implementation the function incorrectly loads those files.

3. Regarding adding this to ``load`` vs a separate function, I agree with @ericpre in that it is more convenient to add it to ``load``. Still, I would leave the implementation as a separate function that can be called from inside ``load``.

4. Instead of a ``lazy`` and ``stack`` keyword arguments, what about adding **kwarg to be passed to ``load``?

woozey:
@M0hsend Great that you implement the reader for the structured folders! We were missing it here in ER-C. I have a Matlab code which does similar things so if you'd like to have a look, I can send it to you. Though it isn't anything very specific about it...

I've tried your code with one of my K2-IS datasets which actually should work directly, but since I actually export it to RAW files rather than .dm3 which work faster the code broke. I wonder if RAW files  is an option for OneView IS as well? Anyway I'll try to export the K2 data in dm3 and 4 and to check if the reader does what it should. Up to now I have following comments:

1. As mentioned above it would be nice to have it working with (Gatan) RAW files as well.
2. The same as Francisco's comment, I got an error when I had some analyses  data stored together with the original data in the root folder. 
3. It was mentioned somewhere already, that one should be able to load only part of the files either specifying start/stop frame or time interval.

M0hsend:
Thank you all for your detailed comments. I shall submit a modified PR once back from my current high on euphoria and meconium!

tjof2:
What's the status of this @M0hsend?

M0hsend:
Hi @tjof2, Unfortunately I have not had a chance to progress much on this.

jlaehne:
If continued, needs to be rebased on and adapted to #2972?

ericpre:
Closing as this should be dealt with in https://github.com/hyperspy/rosettasciio

PR #1947, opened 2018-05-14
BUGFIX create cyclic map for iterating_kwargs

AEljarrat:
Dear all,
I've encountered an issue with the signal.map method. When trying to map an operation into an input hyperspectral signal (e.g. spectrum image) using a single spectrum as parameter, AttributeError is raised. 

### Bug description 
A minimum (not) working example would be;

```
sM = hs.signals.Signal1D(np.random.rand(10, 1024))
vM = hs.signals.Signal1D(np.random.rand(10, 1024))
v1 = hs.signals.Signal1D(np.random.rand(1024))
def foo(bar, val=v1):
    result = bar + val
    return result

# This works
rM = sM.map(foo, val=vM, inplace=False)
# This does not work
r1 = sM.map(foo, val=v1, inplace=False, parallel=False)

```
This affects other methods that use map for more complicated calculations. For example - the reason I found this bug - when using the eels.richardson_lucy_deconvolution method to deconvolve a single ZLP from EELS spectra, the AtributeError error is again raised. 

### Workaround
A quick workaround to this issue is to repeat the single spectrum so that the navigation shape matches that of the input signal as follows,

```
# This one fails when the navigation dimensions of spc and zlp are larger than 1 and 1, respectively
rld = spc.richardson_lucy_deconvolution(zlp) 
# This one below works
psf = hs.signals.Signal1D(np.repeat(zlp.data[None,:], spc.axes_manager.navigation_size, 0))
rld = spc.richardson_lucy_deconvolution(psf) 

```
### Bug fix
I've implemented what I believe is a better solution by making the utils.create_map_objects create a cyclic iterator for the iterating_kwargs (see commit). This effectively lets the signal iterator for the parameter signal repeat until the input signal reaches the end.

If someone tries this I guess it would be also easy for me to provide some tests for this capability.

### Progress of the PR
- [x] Change implemented
- [x] ready for review
- [x] add tests
- [x] User guide update / example


hyperspy/misc/utils.py
```diff
@@ -944,9 +944,9 @@ def create_map_objects(function, nav_size, iterating_kwargs, **kwargs):
     Moved to a separate method to reduce code duplication.
     """
     from hyperspy.signal import BaseSignal
-    from itertools import repeat
+    from itertools import repeat, cycle
 
-    iterators = tuple(signal[1]._iterate_signal()
+    iterators = tuple(cycle(signal[1]._iterate_signal())
```
to266: This indeed meets the requirements, but also might require significant additional memory (since the in-built `cycle` just stores the yielded results).

Have you considered adding a new method, for example `_cycle_signal` to do exactly this, but without the drawback? It could be very similar to the `_iterate_signal`, but with a different end condition - for one would stop, and the other would start at the beginning. In fact, I would hope the two methods to be almost identical, and thus it might be possible to refactor them in a very pleasant manner :)

hyperspy/tests/utils/test_map.py
```diff
@@ -0,0 +1,106 @@
...
+        # Done
+        np.testing.assert_array_almost_equal( res_simple.data, res_convol.data )
+
+class TestMapSignal:
+    """
+    Test map onto signal with signal parameter(s). Proceed by applying a
+    very simple function and parameters that have to produce zero as result.
+    """
+    def setup_method(self, method):
+        def foo(bar, val=0.):
```
to266: I don't think this has to be contained in a `setup_method` - I would expect it to be possible to write the `foo` directly in the test class. (not all methods have to be tests!)

hyperspy/tests/utils/test_map.py
```diff
@@ -0,0 +1,106 @@
...
+    Gaussian filter to all the images in an image dataset using variable sigma
+    parameter(s)
+    """
+    def setup_method(self, method):
+        self.sigma  = 2.5
+        self.sigmas = hssig(np.linspace(2,5,10)).T
+        self.image  = hsimg(np.random.random((10, 64, 64)))
+        self.image.metadata.General.title = 'test'
+
+    def test_A(self):
```
to266: Better test names should be given - when the test fails, you should aim for it to be immediately clear what was broken. `test_A` only makes sense if you either know exactly what it does, or open the test to begin with, and the goal is to avoid having to do so.

to266 (commented):
A couple of things to note in addition to my comments in the code:
 - A user guide update / example is currently missing
 - `hyperspy/hyperspy/tests/signal/test_map_method.py` already exists, and it might be better to keep similar tests in one file.

AEljarrat:
Hi to266,
thanks for reviewing the commits! Following your suggestion, I have added a new `_cycle_signal` method that is also now used by the `create_map_objects` method. Is this close to what you were thinking?

I completely agree on your comments on the tests. I have overlooked the already existing tests for the map method. I will work towards integrating my tests into this file, as it seems some duplicity might exist.

francisco-dlp:
@AEljarrat, are you planning to add an example to the user guide?

AEljarrat:
@francisco-dlp, I've never added an example to the UG before. Had to do it in 2 commits, because the first change I tried did not turn out good when I looked at it in the web. Next time, I should find a way to review UG changes locally, sorry.

francisco-dlp:
Great, thanks.

I've added the "regression" label because, for the case of RLD this used to work before it was rewritten using ``map``.

@to266, is this good to merge?

Issue #1957, opened 2018-05-23
BUG?: `rebin` requires dwell time for EELS signals


woozey:
I got my EELS map loaded from FEI's .emi file. Trying to bin it down got following:
``` python
>>> eels_before.rebin(scale=(1., 1., 8.))

---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
<ipython-input-119-395d823fc946> in <module>()
----> 1 eels_before.rebin(scale=(1., 1., 8.))

~/anaconda/lib/python3.5/site-packages/hyperspy/_signals/eels.py in rebin(self, new_shape, scale, crop, out)
   1349         time_factor = np.prod([factors[axis.index_in_array]
   1350                                for axis in m.axes_manager.navigation_axes])
-> 1351         mdeels = m.metadata.Acquisition_instrument.TEM.Detector.EELS
   1352         m.get_dimensions_from_data()
   1353         if "Acquisition_instrument.TEM.Detector.EELS.dwell_time" in m.metadata:

~/anaconda/lib/python3.5/site-packages/hyperspy/misc/utils.py in __getattribute__(self, name)
    336             name = name.decode()
    337         name = slugify(name, valid_variable_name=True)
--> 338         item = super(DictionaryTreeBrowser, self).__getattribute__(name)
    339         if isinstance(item, dict) and '_dtb_value_' in item and "key" in item:
    340             return item['_dtb_value_']

AttributeError: 'DictionaryTreeBrowser' object has no attribute 'EELS'
```
It seems from the code that the functions trying to access "Acquisition_instrument.TEM.Detector.EELS.dwell_time" in metadata, which is not there for my signal:
``` python
>>> eels_before.metadata

├── Acquisition_instrument
│   └── TEM
│       ├── Detector
│       │   └── Camera
│       │       └── exposure = 0.000004
│       ├── Stage
│       │   ├── tilt_alpha = 0.0
│       │   ├── tilt_beta = 0.0
│       │   ├── x = -0.163507
│       │   ├── y = -0.01909
│       │   └── z = 0.081091
│       ├── acquisition_mode = STEM
│       ├── beam_energy = 300.0
│       ├── camera_length = 77.0
│       ├── magnification = 115000.0
│       └── microscope = Titan 300 kV D3096 SuperTwin
├── General
│   ├── authors = Service
│   ├── date = 2018-02-20
│   ├── original_filename = 13.39.50 Drift corrected spectrum image_before_1.ser
[...]

Issue #1923, opened 2018-04-20
Rebin Error for Linear Binning


k8macarthur:
```---------------------------------------------------------------------------
TypingError                               Traceback (most recent call last)
<ipython-input-16-d31d93845a2c> in <module>()
----> 1 s.rebin(scale=(3,3,1))

c:\users\macark\documents\github\hyperspy\hyperspy\_signals\eds.py in rebin(self, new_shape, scale, crop, out)
    182             new_shape=new_shape,
    183             scale=scale,)
--> 184         m = super().rebin(new_shape=new_shape, scale=scale, crop=crop, out=out)
    185         m = out or m
    186         time_factor = np.prod([factors[axis.index_in_array]

c:\users\macark\documents\github\hyperspy\hyperspy\signal.py in rebin(self, new_shape, scale, crop, out)
   2368         s = out or self._deepcopy_with_new_data(None, copy_variance=True)
   2369         data = hyperspy.misc.array_tools.rebin(
-> 2370             self.data, scale=factors, crop=crop)
   2371 
   2372         if out:

c:\users\macark\documents\github\hyperspy\hyperspy\misc\array_tools.py in rebin(a, new_shape, scale, crop)
    154     if _requires_linear_rebin(arr=a, scale=scale):
    155         _logger.debug("Using linear_bin")
--> 156         return _linear_bin(a, scale, crop)
    157     else:
    158         _logger.debug("Using standard rebin with lazy support")

c:\users\macark\documents\github\hyperspy\hyperspy\misc\array_tools.py in _linear_bin(dat, scale, crop)
    295         result = np.zeros((dim,) + dat.shape[1:], dtype=dtype)
    296         # Carry out binning over axis[0]
--> 297         _linear_bin_loop(result=result, data=dat, scale=s)
    298         # Swap axis[0] back to the original axis location.
    299         result = result.swapaxes(0, axis)

C:\ProgramData\Anaconda3\lib\site-packages\numba\dispatcher.py in _compile_for_args(self, *args, **kws)
    328                                 for i, err in failed_args))
    329                 e.patch_message(msg)
--> 330             raise e
    331 
    332     def inspect_llvm(self, signature=None):

C:\ProgramData\Anaconda3\lib\site-packages\numba\dispatcher.py in _compile_for_args(self, *args, **kws)
    305                 argtypes.append(self.typeof_pyval(a))
    306         try:
--> 307             return self.compile(tuple(argtypes))
    308         except errors.TypingError as e:
[...]

Issue #1846, opened 2018-02-26
``rebin`` does not modify the rebinned axes offsets


francisco-dlp:
For example, currently:

```python

>>> import hyperspy.api as hs
>>> import numpy as np
>>> s = hs.signals.Signal1D(np.arange((10)))
>>> sb = s.rebin((5,))
>>> sb.metadata.Signal.binned
False
>>> sb.axes_manager.signal_axes[0].offset
0.0
>>> s.axes_manager.signal_axes[0].offset
0.0
```

The expected behaviour is:

```python

>>> import hyperspy.api as hs
>>> import numpy as np
>>> s = hs.signals.Signal1D(np.arange((10)))
>>> sb = s.rebin((5,))
TypeError: The signal is not binned, hence it cannot be rebinned
>>> sb.metadata.Signal.binned = True
>>> sb = s.rebin((5,))
>>> sb.axes_manager.signal_axes[0].offset
0.5
>>> s.axes_manager.signal_axes[0].offset
0.0
```

thomasaarholt:
Nice catch with not being able to rebin a non-binned signal.

k8macarthur:
So if I get this error correctly the offset should be changed so that each value sits in the middle of the bin?

k8macarthur:
I'm trying to work on the rebin function at the moment but I don't quite understand the problem here...
Could anyone explain more clearly about the offset? Also I'm guessing the 'binned' search should go in                                         each of the signal classes and not in the base function right?

francisco-dlp:
Imagine that you are binning a 3 channels spectrum image into just one channel. If the first pixel energy is centred at 0 eV, second at 1 eV and third at 2 eV, the result of binning should be a single 3x larger pixel centred at 1 eV. However, currently we get a 3x larger pixel centred at 0 eV. The counts and scale are correctly computed, only the offset is wrong.

I think that both the "binned" and "offset" bugs should be fixed in the parent method.

k8macarthur:
Ok I've updated #1958 with these two corrections as well.

k8macarthur:
Assuming the new offset is (n-1)*x/2 where n is the rebin scale and x is the original scale.

ericpre:
Fixed in #1965.

PR #1958, opened 2018-05-23
Update to Rebin Function Fixing #1957 and #1923

k8macarthur:
### Description of the change
- Function now checks for dwell_time or live_time and doesn't break if not found in the metadata but does raise a warning to clarify that it has not been changed. #1957 
- Checks for dtype of array if using the linear_bin method and converts to float if found to be an integer. #1923 
- Checks dtype of scale or new_shape, if using linear_bin method and raises Error if integer. #1923 

Still needs
- [x] Update to fix checking whether or not spectrum is binned (and so can be rebinned).
- [x] Update to fix rebinned axes offsets #1846


k8macarthur:
I think the three mentioned issues are the only one relating to the rebin function but please let me know if I have missed any.

francisco-dlp:
Since this is a bug-fix, could you submit this to Rnp instead?

francisco-dlp:
Just changing the base branch in GH is not enough because you've branched out from Rnm. You'll probably need to rebase or cherry pick the commits into a new branch e.g.

```bash
git checkout -b FIX_rebin RELEASE_next_patch
git cherry-pick 1521b19aa502eade2a5fe9d4f3d602fc8cee1096..433d0680bfde4579723e5a13eba19afde9c2e78b
```

PR #1959, opened 2018-05-23
Enhancement: select and replot one axis in plot_images

AEljarrat:
### Description
This enhancement proposal adds the possibility to select and replot and axis in plot_images. This is done by clicking into this axis. Each click triggers a plot event, in which the selected axis alone is presented in a figure. 

This helps navigating through a panel with many figures by enlarging some of them and allowing comfortable zooming. I use this trick a lot and thought about contributing it. 

### Progress of the PR
- [x] Change implemented 
- [x] update docstring
- [x] ready for review
- [x] add tests

### Minimal example of new feature
```python
>>> import hyperspy.api as hs
>>> import numpy as np
>>> imgdata = np.random.rand(20, 512, 512)
>>> imgs = hs.signals.Signal2D(imgdata)
>>> hs.plot.plot_images(imgs, axes_decor='off', axes_replot=True)
>>> # Double click on one axis and get a new plot
```

### DIFF
--- a/hyperspy/tests/drawing/test_plot_signal2d.py
+++ b/hyperspy/tests/drawing/test_plot_signal2d.py
@@ -353,7 +353,8 @@ def test_plot_images_multi_signal_w_axes_replot(mpl_cleanup):
     img_list = [imgs, imgs.inav[:2], imgs.inav[0]]
     subplots=hs.plot.plot_images(img_list, axes_replot=True)
     f = plt.gcf()
-    plt.pause(0.01) # this pause is important somehow
+    f.canvas.draw()
+    f.canvas.flush_events()
 
     tests = []
     for axi in subplots:
--- a/hyperspy/tests/drawing/test_plot_signal2d.py
+++ b/hyperspy/tests/drawing/test_plot_signal2d.py
@@ -351,7 +351,7 @@ def test_plot_images_multi_signal_w_axes_replot(mpl_cleanup):
     imdata = np.random.rand(3, 5, 5)
     imgs = hs.signals.Signal2D(imdata)
     img_list = [imgs, imgs.inav[:2], imgs.inav[0]]
-    subplots=hs.plot.plot_images(img_lists, axes_decor=None)
+    subplots=hs.plot.plot_images(img_list, axes_decor=None)
     f = plt.gcf()
     f.canvas.draw()
     f.canvas.flush_events()
--- a/hyperspy/tests/drawing/test_plot_signal2d.py
+++ b/hyperspy/tests/drawing/test_plot_signal2d.py
@@ -347,6 +347,27 @@ def test_plot_images_single_image(mpl_cleanup):
     ax = hs.plot.plot_images(image0, saturated_pixels=0.1)
     return ax[0].figure
 
+def test_plot_images_multi_signal_w_axes_replot(mpl_cleanup):
+    imdata = np.random.rand(3, 5, 5)
+    imgs = hs.signals.Signal2D(imdata)
+    img_list = [imgs, imgs.inav[:2], imgs.inav[0]]
+    subplots=hs.plot.plot_images(img_list, axes_replot=True)
+    f = plt.gcf()
+    plt.pause(0.01) # this pause is important somehow
+
+    tests = []
+    for axi in subplots:
+        imi = axi.images[0].get_array()
+        x, y = axi.transData.transform((2, 2))
+        # Calling base class method because of backends
+        plt.matplotlib.backends.backend_agg.FigureCanvasBase.button_press_event(
+                                                  f.canvas, x, y, 'left',  True)
+        fn = plt.gcf()
+        tests.append(
+            np.allclose(imi, fn.axes[0].images[0].get_array().data) )
+        plt.close(fn)
+    assert np.alltrue(tests)
+    return f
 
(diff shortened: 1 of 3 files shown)


francisco-dlp:
I find this feature very useful, nice idea! I have some concerns about the implementation though: why doing it the hacky way when it can be done more easily by registering an function that simply plots the image (maybe even using ``Signal.plot``) on a double-clicking event for each individual axis ?

AEljarrat:
Hi @francisco-dlp , I have tried to re-implement the same idea using `Signal.plot`, storing in the process a reference to the list of images that was passed to the `plot_images` method. I don't know exactly, but I guess this is very wasteful memory-wise... The good thing is that all calibrations, labels, etc... are stored and reproduced in the re-plot window.

In my opinion, the good thing about the former implementation is that it was making an exact copy of the already existing axis, therefore having minimum memory impact.

Perhaps a better way would be to salvage the calibration information from the `mpl.axes.Subplot` object, and use this to build an appropriate `Signal2d` for re-plotting?

AEljarrat:
Extending on my previous message, here's an *minimum example* of the current behaviour;
```python
imgdata = np.random.rand(12, 512, 512)
imgs = hs.signals.Signal2D(imgdata)
hs.plot.plot_images([imgs, imgs.inav[:6]*2, imgs.inav[0]], axes_replot=True)
```
In this example, the event manager stores references to the list elements we have passed to plot_images. Since `imgs` was already instantiated, the reference to the first element is to the already existing object. However, the second and third elements have new references, and I guess are stored in memory as separate signals as long as the image panel is open. The way I have checked this is using the `id` builtin function.

francisco-dlp:
I think that keeping a reference to the image is not wasteful memory-wise because, even if the ``Signal2D`` object has a different ``id``, its data is just a view of the original.
Regarding the implementation, wouldn't it be possible to perform the registration inside the ``plot_images`` main for loop instead of searching for the right image clicked afterwards?

AEljarrat:
Hi @francisco-dlp, I was not sure if storing from the main loop would create new persisting objects. I have re-checked (see code below) again and it seems that at least the data stored is a view of the original, so you are right and its a clean method :)

I could now write a note in the UG explaining this new functionality, but I am really not sure how to write a test for it. Any ideas?

Here's the code I used to check (should print a bunch of True);
```python
imgdata = np.random.rand(12, 512, 512)
imgs = hs.signals.Signal2D(imgdata)

datal1 = []
for ims in imgs:
    for im in ims:
        datal1.append(im.data)
        print(im.data.base is imgs.data)

datal2 = []
for ims in imgs:
    for im in ims:
        datal2.append(im.data)
        print(im.data.base is imgs.data)
        
for d1, d2 in zip(datal1, datal2):
    print(np.may_share_memory(d1, d2))
```

francisco-dlp:
I suppose that testing this feature should involve triggering manually the in-axes double click event and comparing the resulting imaging with a stored one using the usual method. I have no experience with manually triggering matplotlib events, but it seems that the events are stored in a dictionary in ``f.canvas.callbacks.callbacks``. Probably it's possible to trigger them manually by passing an event dictionary.

AEljarrat:
I have prepared a simple test that;
- creates a `plot_images` figure with several panels 
- triggers the double click replot event for each panel
- compares the image data in both images
- All of this using a list of iterating signals

Additionally, I have added a note to the UG.

hyperspy/tests/drawing/test_plot_signal2d.py
```diff
@@ -347,6 +347,28 @@ def test_plot_images_single_image(mpl_cleanup):
     ax = hs.plot.plot_images(image0, saturated_pixels=0.1)
     return ax[0].figure
 
+def test_plot_images_multi_signal_w_axes_replot(mpl_cleanup):
+    imdata = np.random.rand(3, 5, 5)
+    imgs = hs.signals.Signal2D(imdata)
+    img_list = [imgs, imgs.inav[:2], imgs.inav[0]]
+    subplots=hs.plot.plot_images(img_list, axes_replot=True)
```
ericpre: Just a detail to keep things tidy: adding `axes_decor=None` would avoid sending a warning to the logger.

ericpre (commented):
A few comments, other than that, it looks good:
- when reading the PR, I initially understood that double click would "replot" the image (in the same axis). Maybe this could be made more clear by saying that it would plot the image in a new figure?
- is it necessary to add an argument to disable this? The list of argument for `plot_images` is already quite long, I think that it would be fine to have it set by default. It would also slightly simplify the code by removing the if loops.

AEljarrat:
@ericpre I agree completely. Thanks for the reviewing!

hyperspy/tests/drawing/test_plot_signal2d.py
```diff
@@ -351,7 +351,7 @@ def test_plot_images_multi_signal_w_axes_replot(mpl_cleanup):
     imdata = np.random.rand(3, 5, 5)
     imgs = hs.signals.Signal2D(imdata)
     img_list = [imgs, imgs.inav[:2], imgs.inav[0]]
-    subplots=hs.plot.plot_images(img_list, axes_replot=True)
+    subplots=hs.plot.plot_images(img_lists, axes_decor=None)
```
ericpre: This typo makes the test fail.

ericpre:
It looks good to me. @francisco-dlp?

PR #1965, opened 2018-05-28
Fix rebin

k8macarthur:
### Description of the change
- Function now checks for dwell_time or live_time and doesn't break if not found in the metadata but does raise a warning to clarify that it has not been changed. #1957 
- Checks for dtype of array if using the linear_bin method and converts to float if found to be an integer. #1923 
- Checks dtype of scale or new_shape, if using linear_bin method and raises Error if integer. #1923 

Still needs
- [x] Update to fix checking whether or not spectrum is binned (and so can be rebinned).
- [x] Update to fix rebinned axes offsets #1846

### DIFF
--- a/hyperspy/_signals/eels.py
+++ b/hyperspy/_signals/eels.py
@@ -1352,10 +1352,12 @@ class EELSSpectrum_mixin:
         m.get_dimensions_from_data()
         if "Acquisition_instrument.TEM.Detector.EELS.dwell_time" in m.metadata:
             mdeels.EELS.dwell_time *= time_factor
-        if "Acquisition_instrument.TEM.Detector.EELS.exposure" in m.metadata:
+        elif "Acquisition_instrument.TEM.Detector.EELS.exposure" in m.metadata:
             mdeels.EELS.exposure *= time_factor
-        if "Acquisition_instrument.TEM.Detector.Camera.exposure" in m.metadata:
+        elif "Acquisition_instrument.TEM.Detector.Camera.exposure" in m.metadata:
             mdeels.Camera.exposure *= time_factor
+        else:
+            warnings.warn('No dwell_time could be found in the metadata so this has not been updated.') 
         if out is None:
             return m
         else:
--- a/hyperspy/_signals/eels.py
+++ b/hyperspy/_signals/eels.py
@@ -1357,7 +1357,7 @@ class EELSSpectrum_mixin:
         elif "Acquisition_instrument.TEM.Detector.Camera.exposure" in m.metadata:
             mdeels.Camera.exposure *= time_factor
         else:
-            _logger.info('No dwell_time could be found in the metadata so this has not been updated.') 
+            _logger.info('No dwell_time could be found in the metadata so this has not been updated.')
         if out is None:
             return m
         else:
--- a/hyperspy/_signals/eels.py
+++ b/hyperspy/_signals/eels.py
@@ -1348,14 +1348,20 @@ class EELSSpectrum_mixin:
         m = out or m
         time_factor = np.prod([factors[axis.index_in_array]
                                for axis in m.axes_manager.navigation_axes])
-        mdeels = m.metadata.Acquisition_instrument.TEM.Detector
+        mdeels = m.metadata
         m.get_dimensions_from_data()
-        if "Acquisition_instrument.TEM.Detector.EELS.dwell_time" in m.metadata:
-            mdeels.EELS.dwell_time *= time_factor
-        elif "Acquisition_instrument.TEM.Detector.EELS.exposure" in m.metadata:
-            mdeels.EELS.exposure *= time_factor
-        elif "Acquisition_instrument.TEM.Detector.Camera.exposure" in m.metadata:
-            mdeels.Camera.exposure *= time_factor
+        if "Acquisition_instrument" in m.metadata:
+            mdeels = m.metadat.Acquisition_instrument
+            if "Acquisition_instrument.TEM.Detector.EELS.dwell_time" in m.metadata:
+                mdeels.TEM.Detector.EELS.dwell_time *= time_factor
+            elif "Acquisition_instrument.TEM.Detector.EELS.exposure" in m.metadata:
+                mdeels.TEM.Detector.EELS.exposure *= time_factor
+            elif "Acquisition_instrument.TEM.Detector.Camera.exposure" in m.metadata:
+                mdeels.TEM.Detector.Camera.exposure *= time_factor
+        elif "dwell_time" in m.metadata:
+            m.metadata.dwell_time *= time_factor
+        elif "exposure" in m.metadata:
+            m.metadata.exposure *= time_factor
         else:
             _logger.info('No dwell_time could be found in the metadata so this has not been updated.')
         if out is None:
--- a/hyperspy/_signals/eels.py
+++ b/hyperspy/_signals/eels.py
@@ -1357,7 +1357,7 @@ class EELSSpectrum_mixin:
         elif "Acquisition_instrument.TEM.Detector.Camera.exposure" in m.metadata:
             mdeels.Camera.exposure *= time_factor
         else:
-            warnings.warn('No dwell_time could be found in the metadata so this has not been updated.') 
+            _logger.info('No dwell_time could be found in the metadata so this has not been updated.') 
         if out is None:
             return m
         else:
--- a/hyperspy/_signals/eels.py
+++ b/hyperspy/_signals/eels.py
@@ -1350,20 +1350,15 @@ class EELSSpectrum_mixin:
                                for axis in m.axes_manager.navigation_axes])
         mdeels = m.metadata
         m.get_dimensions_from_data()
-        if "Acquisition_instrument" in m.metadata:
-            mdeels = m.metadat.Acquisition_instrument
-            if "Acquisition_instrument.TEM.Detector.EELS.dwell_time" in m.metadata:
-                mdeels.TEM.Detector.EELS.dwell_time *= time_factor
-            elif "Acquisition_instrument.TEM.Detector.EELS.exposure" in m.metadata:
-                mdeels.TEM.Detector.EELS.exposure *= time_factor
-            elif "Acquisition_instrument.TEM.Detector.Camera.exposure" in m.metadata:
-                mdeels.TEM.Detector.Camera.exposure *= time_factor
-        elif "dwell_time" in m.metadata:
-            m.metadata.dwell_time *= time_factor
-        elif "exposure" in m.metadata:
-            m.metadata.exposure *= time_factor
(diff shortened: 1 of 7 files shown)


k8macarthur:
Does anyone have any tips on how I raise a warning without causing the tests to fail? I could simply add real_time in the metadata for the test examples but I feel like this is just a patch.

francisco-dlp:
The only warning that don't wake the test fail are hspy's deprecation warnings. Why not adding a warning to the log instead?

hyperspy/misc/array_tools.py
```diff
@@ -301,6 +308,9 @@ def _linear_bin(dat, scale, crop=True):
         # The new dimension size is old_size/step, this is rounded down normally
         # but if crop is switched off it is rounded up to the nearest whole
         # number.
+        if type(s) != float:
```
thomasaarholt: I don't think this is necessary? Python 3 doesn't do Python 2's "floor division" if the divisor is an int. aka in Python 3, 2/3 = 0.666, not 0.

hyperspy/_signals/eels.py
```diff
@@ -1348,12 +1348,16 @@ def rebin(self, new_shape=None, scale=None, crop=True, out=None):
         m = out or m
         time_factor = np.prod([factors[axis.index_in_array]
                                for axis in m.axes_manager.navigation_axes])
-        mdeels = m.metadata.Acquisition_instrument.TEM.Detector.EELS
+        mdeels = m.metadata.Acquisition_instrument.TEM.Detector
```
thomasaarholt: I wonder if we should wrap all of this (from here until the end of the else statement below) in an `if "Acquisition_instrument" in m.metadata`. 

I've done a few simulations of EELS spectra, where there is no acquisition_instrument. There I need to convert the dataset into Signal1D, rebin, and then back to EELS for this to work.

hyperspy/signal.py
```diff
@@ -2360,6 +2360,9 @@ def rebin(self, new_shape=None, scale=None, crop=True, out=None):
         Sum =  164.0
 
         """
+        if self.metadata.Signal.binned == False:
+            raise TypeError('The signal is not binned, hence cannot be rebinned.\
```
ericpre: What is the reason to prevent rebinning not binned data?

hyperspy/misc/array_tools.py
ericpre: This may be necessary to avoid issue with numba and it would be better to use `not np.issubdtype(s, float)` because `type(np.float32) != 'float'` return `True`.

hyperspy/misc/array_tools.py
```diff
@@ -163,6 +166,10 @@ def rebin(a, new_shape=None, scale=None, crop=True):
     # check whether or not interpolation is needed.
     if _requires_linear_rebin(arr=a, scale=scale):
         _logger.debug("Using linear_bin")
+        if np.issubdtype(a.dtype, np.integer):
+            # The operations below require a float dtype with the default
+            # numpy casting rule ('same_kind')
+            a.astype("float")
```
ericpre: It needs to be `a.astype("float", casting="safe", copy=False)`. Setting the argument casting to "safe" should make sure that the value don't change during casting (the default is "unsafe" which doesn't check nothing).

hyperspy/misc/array_tools.py
```diff
@@ -163,6 +166,10 @@ def rebin(a, new_shape=None, scale=None, crop=True):
     # check whether or not interpolation is needed.
     if _requires_linear_rebin(arr=a, scale=scale):
         _logger.debug("Using linear_bin")
+        if np.issubdtype(a.dtype, np.integer):
+            # The operations below require a float dtype with the default
```
ericpre: I don't understand this comment: what operation are you talking about?

ericpre (changes requested):
The tests are failing because of the "binned=True" check in the metadata. The issue mentioned in #1923 is not covered by the tests.

ericpre:
Just to check that the following is the intended behaviour:
```python
import hyperspy.api as hs
import numpy as np

s = hs.signals.Signal1D(np.ones(10))
s2 = s.rebin(scale=(3, ))
assert s.data.sum() == s2.data.sum()
# sum of s2 and s are different in this case
```
which doesn't conserve the sum but setting `crop=False` does (even if the sampling of the last bin is not correct, but worrying starts to be overkilled...).

hyperspy/signal.py
k8macarthur: This was done in response to @francisco-dlp  comments in #1846. However I think it breaks too many other workflows so I'll remove it again/

hyperspy/misc/array_tools.py
k8macarthur: Good spot.

hyperspy/misc/array_tools.py
k8macarthur: This refers to the _linear_bin function.

hyperspy/misc/array_tools.py
k8macarthur: This was necessary to fix a numpy error where the default is to keep same kind.

hyperspy/_signals/eels.py
k8macarthur: What's the structure of your metadata?

k8macarthur: You'd think if it was all Gatan there would be some consistancy....

k8macarthur:
@ericpre Thanks for the comments, I think the next commit should respond to all you've suggested. With regards to the default behaviour, that is correct. This is the default behaviour the linear interpolation function had before and this pull request does not change that. The reason why crop is default is so that any subsequent quantification will still be accurate on all remaining pixels.

@thomasaarholt any advice you can give on updating dwell time which will work with the majority of EELS data would be very helpful as I almost never do EELS.

hyperspy/misc/array_tools.py
```diff
@@ -301,6 +308,9 @@ def _linear_bin(dat, scale, crop=True):
         # The new dimension size is old_size/step, this is rounded down normally
         # but if crop is switched off it is rounded up to the nearest whole
         # number.
+        if not np.issubtype(s, float):
```
k8macarthur: So Travis is failing because of: 
```
FutureWarning: Conversion of the second argument of issubdtype from `float` to `np.floating` is deprecated. In future, it will be treated as `np.float64 == np.dtype(float).type`.
```
Does that mean converting to np.issubdtype(s, np.float64) or np.issubdtype(s, np.dtype(float).type)?

thomasaarholt:
Assuming the function currently does what you want it to, you should just replace 
```
float 
```
With
```
np.floating
```

hyperspy/misc/array_tools.py
```diff
@@ -163,6 +166,10 @@ def rebin(a, new_shape=None, scale=None, crop=True):
     # check whether or not interpolation is needed.
     if _requires_linear_rebin(arr=a, scale=scale):
         _logger.debug("Using linear_bin")
+        if np.issubdtype(a.dtype, np.integer):
+            # The _linear_bin function below requires a float dtype
+            # because of the default numpy casting rule ('same_kind').
+            a.astype("float", castin="safe", copy=False)
```
ericpre: Typo: it should be `casting`  and it needs a test.

hyperspy/_signals/eels.py
```diff
@@ -1348,12 +1348,16 @@ def rebin(self, new_shape=None, scale=None, crop=True, out=None):
...
         m.get_dimensions_from_data()
         if "Acquisition_instrument.TEM.Detector.EELS.dwell_time" in m.metadata:
-            mdeels.dwell_time *= time_factor
-        if "Acquisition_instrument.TEM.Detector.EELS.exposure" in m.metadata:
-            mdeels.exposure *= time_factor
+            mdeels.EELS.dwell_time *= time_factor
```
ericpre: Add a test for these.

hyperspy/signal.py
```diff
@@ -2376,7 +2376,7 @@ def rebin(self, new_shape=None, scale=None, crop=True, out=None):
             s.data = data
         s.get_dimensions_from_data()
         for i, factor in enumerate(factors):
-            s.axes_manager[i].offset = (factor - 1)/2*s.axes_manager[i].scale
+            s.axes_manager[i].offset = (factor - 1)/2*s.axes_manager[i].scale + s.axes_manager[i].offset
```
ericpre: Check this value in the tests?

hyperspy/misc/array_tools.py
```diff
@@ -163,6 +166,10 @@ def rebin(a, new_shape=None, scale=None, crop=True):
     # check whether or not interpolation is needed.
     if _requires_linear_rebin(arr=a, scale=scale):
         _logger.debug("Using linear_bin")
+        if np.issubdtype(a.dtype, np.integer):
+            # The _linear_bin function below requires a float dtype
+            # because of the default numpy casting rule ('same_kind').
```
ericpre, line 171: Where exactly does the casting take place? I am not sure if the default is actually `same_kind`.

ericpre (commented):
The following still gives an issue with numba:
```python
import hyperspy.api as hs
import numpy as np

s = hs.signals.Signal1D(np.arange(1000).reshape(10, 10, 10))
s.change_dtype(np.uint8)
s2 = s.rebin(scale=(3, 3, 1))
```

hyperspy/misc/array_tools.py
k8macarthur: Typo fixed but I'm not entirely sure how to write the test for this?

hyperspy/_signals/eels.py
k8macarthur: I can add a test for sum but I'm not sure how to add the dictionary tree to the metadata such that I can check all of them?

hyperspy/signal.py
k8macarthur: Done

hyperspy/misc/array_tools.py
k8macarthur: I'm not sure how well I understand the error having not looked at this in a while. However it was maintaining the dtype during linear binning which was causing the resin error that came up with weird results. A floating dtype is needed to preserve counts where we're binning in such a way to split up pixels.

k8macarthur:
So I've finally had time to look at this today. I added quite a lot more tests and fixed two errors in the code in the process. The only things missing now are:
-Tests for the casting error.
-Tests for all the EELS metadata as I'm not sure how to added all the dictionary tree elements.

hyperspy/_signals/eels.py
k8macarthur: I've removed this now. I realised that when you create an EELS spectrum with signals.EELSSPECTRUM there is no acquisition instrument in the data tree. Figured it shouldn't break on internally created EELS spectrum. :)

ericpre (approved):

PR #1969, opened 2018-05-31
ADD note to UG, for users that cannot install dependencies from source

### LINKED ISSUE #1950: problem plotting after clean installation
Dear all, 
its been a while since i've been able to make an installation of hyperspy with working plot capabilities from scratch. Following the steps in the documentation, I install hyperspy in a clean conda environment (python 3, cython) using pip and the source code from git. Typically, I get a PyQT error when trying to use plots. Finally, I've found this error is solved if I additionally install pyside2; `conda install pyside2 -c conda-forge`.

I don't know if there is a reason why this package is not automatically installed as a dependency of hyperspy. Anyhow, I thought it would maybe be helpful to add this info to the documentation. I am not sure about this, but I guess it does not hurt to report it, so we can discuss further.

Cheers!
Alberto.

AEljarrat:
### Description of the change
Fixes #1950 by adding a note to the UG for users that install from source but are not able to get the required libraries installed. For this reason I have added the note to the "Installing the required libraries" section.

### Progress of the PR
- [x] update user guide
- [x] ready for review.

### DIFF
--- a/doc/user_guide/install.rst
+++ b/doc/user_guide/install.rst
@@ -31,7 +31,7 @@ Bundle. This is a customised `WinPython <http://winpython.github.io/>`_
 distribution that includes HyperSpy, all its dependencies and many other
 scientific Python packages.
 
-For details and download links go to https://github.com/hyperspy/hyperspy-bundle 
+For details and download links go to https://github.com/hyperspy/hyperspy-bundle
 
 .. _quick-anaconda-install:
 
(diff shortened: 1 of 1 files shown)


ericpre (commented):
Although this is the right approach, this actually doesn't behaviour as it is supposed to do with the current hyperspy conda-forge recipe, because it has hyperspy-gui-ipywidgets as dependency, which obviously has hyperspy as dependency... so hyperspy get installed when it should not.
This issue is that the hyperspy-gui-ipywidgets should not be a hyperspy dependency but at the moment, this is the case because it is clearly more convenient for the user. I guess this approach should be revisited when hyperpsy is split.
Either we leave it as it is and wait for the split or a step to uninstall hyperspy with conda need to be added.

AEljarrat:
You are right! I have changed the note to the UG in case we decide to include it. I think it would not hurt to include it since it is a pretty fail-safe way of getting the dependencies right. I guess it would not work if the source dependencies would have changes from the conda-forge dependencies, which would be strange.

PR #1973, opened 2018-06-05
BUGFIX: explicitly set axis limits in plot_spectra

### LINKED ISSUE #1972: BUG: plot_spectra() does not correctly plot a decreasing axis
If the scale of the x-axis is negative (common in IR spectroscopy), using the `plot_spectra` function will plot the data in increasing order:

Example plotting using `s.plot()`:
image

Example using `plot_spectra()`:
image

This is because matplotlib defaults to an increasing axis, and there is no explicit setting of axes limits like there is in `signal.plot()` (see https://github.com/hyperspy/hyperspy/blob/RELEASE_next_minor/hyperspy/drawing/signal1d.py#L113-L116).

This shouldn't be too hard to fix by copying those lines or similar into `plot_spectra`

jat255:
### Description of the change
Resolves #1972. In that bug, a signal with a decreasing axis is plotted as increasing when using the helper function `plot_spectra()`. This is because this method never explicitly changes the figure's x limits like the normal `plot()` method does. This PR fixes that.

### Progress of the PR
- [X] add tests
- [X] ready for review.

Compare this example before and after the fix:

```python
import scipy
s = hs.signals.Signal1D(scipy.misc.ascent()[100:160:10])
s.axes_manager[1].offset = 512
s.axes_manager[1].scale = -1
hs.plot.plot_spectra(s)
```


ericpre (commented):
This looks good.
Some of the baseline doesn't need to be changed, can you please remove them by amending your last commit? If you add a new commit, the images will be still added to the repo.
I think it is a good practise to update the baseline images only when required to avoid the git repo to unnecessarily grow in size.

jat255:
Ok, should be fixed. I couldn't figure out how to generate just those new ones using `py.test`, but I guess I could have just discarded the ones that didn't need to be changed. Either way, I removed them and force pushed, so this should be GTG.

ericpre:
I don't think that there is a proper way to do it. Usually, I discard the one which doesn't need to be kept, which is useful to check if everything is as it is supposed to be.
The tests are failing, so you may have deleted too many baseline images or you introduced unexpected changes! ;)

jat255:
So the reason this failed is because of (I think) an inconsistency in our previously default settings for the `'mosaic'` style compared with all the other plots. If you look at any other plotted spectrum (such as https://github.com/hyperspy/hyperspy/blob/RELEASE_next_minor/hyperspy/tests/drawing/plot_signal1d/test_plot_spectra_default.png), the data fully spans the axis, and there is no gap between the plotted spectrum and the left/right borders of the axis. For some reason, this happens in the current implementation of the mosaic style (see the baseline image below). This PR explicitly sets the axis limits in `_plot_spectrum()`, which is the reason the mosaic style has changed. I would argue that we should do it this way, that way all the styles are consistent in spreading the spectrum over the full extent of the axes. I am going to update the baseline images for the mosaic style as well so it tests out alright.

Baseline (existing):


From this PR:


Diff:

jat255:
Only one of the four AppVeyor runs failed, and it was in `test_hdf5`, which is totally unrelated to this PR, so I'm guessing it was just a random one-off failure. Otherwise everything looks correct.

jat255:
I just rebased onto the latest RNP

PR #1977, opened 2018-06-11
Bug kramerskronig single spectrum

AEljarrat:
### Description 
Kramers Kronig analysis (KKA) with a single spectrum seems to be broken (see MEW below). I have looked into the issue, and it seems it is due to a couple of matrices (K and te) being cast to numbers in case single spectrum KKA was performed. I propose some small changes to fix this bug, by preserving the array nature of K and te.

### Progress of the PR
- [x] Change implemented.
- [x] add tests,
- [x] ready for review.

### Minimal example of the bug
```python
>>> Ne = 512
>>> Eax = {'size':Ne, 'name':'E', 'units':'eV', 'scale':0.05}
>>> axl = [Eax,]
>>> s = hs.signals.EELSSpectrum(np.random.rand(Ne), axis=axl)
>>> z = hs.signals.EELSSpectrum(np.random.rand(Ne), axis=axl)
>>> s.set_microscope_parameters(10., 10., 10.)
>>> cdf, out = s.kramers_kronig_analysis(z, 1, n=1.,full_output=True)

/home/aeljarrat/hyperspy/hyperspy/_signals/eels.py:1208: RuntimeWarning: divide by zero encountered in double_scalars
  K = (K / (np.pi / 2) / (1 - 1. / n ** 2)).reshape(

<ipython-input-4-41bd1f068e33> in <module>()
      6 s.set_microscope_parameters(10., 10., 10.)
      7 
----> 8 cdf, out = s.kramers_kronig_analysis(z, 1, n=1.,full_output=True)

~/hyperspy/hyperspy/_signals/eels.py in kramers_kronig_analysis(self, zlp, iterations, n, t, delta, full_output)
   1207                 K = (Im / eaxis).sum(axis=axis.index_in_array) * axis.scale
   1208                 K = (K / (np.pi / 2) / (1 - 1. / n ** 2)).reshape(
-> 1209                     np.insert(K.shape, axis.index_in_array, 1))
   1210                 # Calculate the thickness only if possible and required
   1211                 if zlp is not None and (full_output is True or

TypeError: 'numpy.float64' object cannot be interpreted as an integer

```

### DIFF
--- a/hyperspy/_signals/dielectric_function.py
+++ b/hyperspy/_signals/dielectric_function.py
@@ -122,6 +122,7 @@ class DielectricFunction_mixin:
         data = ((-1 / self.data).imag * eels_constant(self, zlp, t).data *
                 self.axes_manager.signal_axes[0].scale)
         s = self._deepcopy_with_new_data(data)
+        s.data = s.data.real
         s.set_signal_type("EELS")
         s.metadata.General.title = ("EELS calculated from " +
                                     self.metadata.General.title)
--- a/hyperspy/_signals/eels.py
+++ b/hyperspy/_signals/eels.py
@@ -1288,19 +1288,13 @@ class EELSSpectrum_mixin:
         if 'thickness' in output:
             # As above,prevent errors if the signal is a single spectrum
             if len(te) != 1:
-                thickness = eps._get_navigation_signal(
-                    data=te[self.axes_manager._get_data_slice(
-                        [(axis.index_in_array, 0)])])
-                thickness.metadata.General.title = (
-                    self.metadata.General.title + ' thickness '
-                    '(calculated using Kramers-Kronig analysis)')
-                output['thickness'] = thickness
-            else:
-                thickness = eps._get_navigation_signal(data=te)
-                thickness.metadata.General.title = (
-                    self.metadata.General.title + ' thickness '
-                    '(calculated using Kramers-Kronig analysis)')
-                output['thickness'] = thickness
+                te = te[self.axes_manager._get_data_slice(
+                        [(axis.index_in_array, 0)])]
+            thickness = eps._get_navigation_signal(data=te)
+            thickness.metadata.General.title = (
+                self.metadata.General.title + ' thickness '
+                '(calculated using Kramers-Kronig analysis)')
+            output['thickness'] = thickness
         if full_output is False:
             return eps
         else:
--- a/hyperspy/_signals/eels.py
+++ b/hyperspy/_signals/eels.py
@@ -1295,6 +1295,12 @@ class EELSSpectrum_mixin:
                     self.metadata.General.title + ' thickness '
                     '(calculated using Kramers-Kronig analysis)')
                 output['thickness'] = thickness
+            else:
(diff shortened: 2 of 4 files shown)


hyperspy/_signals/dielectric_function.py
```diff
@@ -122,6 +122,7 @@ def get_electron_energy_loss_spectrum(self, zlp, t):
         data = ((-1 / self.data).imag * eels_constant(self, zlp, t).data *
                 self.axes_manager.signal_axes[0].scale)
         s = self._deepcopy_with_new_data(data)
+        s.data = s.data.real
```
francisco-dlp, line 125: This is fixing yet another bug, maybe mention it in the PR?

AEljarrat: Sorry, I forgot to mention this in the PR, although it is mentioned in the commit. I run accross the issue while I was writing the test. The deepcopy creates a new dielectric type signal, but we want an EELS spectrum.

AEljarrat:
As mentioned above, I ran across an additional (mini) bug while I was writing the test. When estimating the inelastic scattering from a dielectric function using the `get_electron_energy_loss_spectrum` method, the result was another dielectric type although it should be EELS type. The issue is deepcopy creates a new dielectric type signal, but we want an EELS spectrum.

francisco-dlp:
It looks good to me except for one thing: shouldn't the thickness always be a ``Signal`` for consistency?

AEljarrat:
It is easy to make it a signal, and I suppose it has the advantage that metadata is preserved.

@francisco-dlp, Thank you for (almost instantaneously) reviewing! Lately I've been trying to fix bugs instead or doing hacky workarounds. I know I'm quite inexperienced, so please feel free to tell me if I'm missing any important step in the process.

hyperspy/_signals/eels.py
```diff
@@ -1295,6 +1295,12 @@ def kramers_kronig_analysis(self,
                     self.metadata.General.title + ' thickness '
                     '(calculated using Kramers-Kronig analysis)')
                 output['thickness'] = thickness
+            else:
+                thickness = eps._get_navigation_signal(data=te)
+                thickness.metadata.General.title = (
+                    self.metadata.General.title + ' thickness '
+                    '(calculated using Kramers-Kronig analysis)')
+                output['thickness'] = thickness
```
francisco-dlp: A minor suggestion: what about the following more compact version instead?

```python
            if len(te) != 1:
                te = te[self.axes_manager._get_data_slice(
                        [(axis.index_in_array, 0)])]
            thickness = eps._get_navigation_signal(data=te)
            thickness.metadata.General.title = (
                self.metadata.General.title + ' thickness '
                '(calculated using Kramers-Kronig analysis)')

francisco-dlp:
No need to apologise: there is almost no PR that passes the review process without modification. No need to be modest either concerning your experience: you're one of HyperSpy's olders contributors and it's great to have you back!

AEljarrat:
@francisco-dlp thanks again for the reviewing, and the good words. It's nice to be back!

PR #1981, opened 2018-06-20
BUGFIX reversed line order in animated legend in plot_spectra

AEljarrat:
### Description of the change
When using `hs.plot.plot_spectra` the animated legend would not work properly. Although the legend lines correspond to the plot data, when clicked, the wrong lines would disappear. This minimalist PR fixes this issue. Indeed, I have only tried this in my system, using qt and nbagg backends. Is it possible this bug only appears for me?

### Progress of the PR
- [x] Change implemented 
- [x] add tests,
- [x] ready for review.

```python
>>> x = np.linspace(0., 2., 512)
>>> n = np.arange(1, 5)
>>> x_pow_n = x[None,:]**n[:,None]
>>> s = hs.signals.Signal1D(x_pow_n)
>>> my_legend = [r'x^'+str(io) for io in n]
>>> hs.plot.plot_spectra(s, legend=my_legend)
### wrong line dissapears when clicking the legend
```

### DIFF
--- a/hyperspy/drawing/utils.py
+++ b/hyperspy/drawing/utils.py
@@ -1317,7 +1317,7 @@ def animate_legend(figure='last'):
         ax = plt.gca()
     else:
         ax = figure.axes[0]
-    lines = ax.lines
+    lines = ax.lines[::-1]
     lined = dict()
     leg = ax.get_legend()
     for legline, origline in zip(leg.get_lines(), lines):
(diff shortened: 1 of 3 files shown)


francisco-dlp:
I can reproduce the bug with MPL 2.2.2 and this fixes it for me. Good catch! 
I've downgraded my mpl to indentify the onset of the issue that appears to be matplotlib v2.0. Therefore I suggest to add to this PR a commit to set the minimum matplotlib version to 2.0.0 in setup.py.

francisco-dlp:
Thanks, are you planning to add a test for this?

AEljarrat:
Yes, I think I will be able to do a small test in the near future, using the MEW I posted as base. I am not sure where I should put it, do you think `tests/drawing/test_plot_signal1d.py` would be a good place?

francisco-dlp:
Yes, there you'll fine the ``TestPlotSpectra`` class.

AEljarrat:
The test has been added, checking that the appropriate lines are set to non-visible.

PR #1986, opened 2018-06-29
Fix excess memory usage when performing reduce operations

francisco-dlp:
Performing reduce operations on **lazy** HyperSpy signals (e.g. ``s.sum()``) can lead to very high memory consumption.

I have identified the issues and fixed it in this pull request. The issue is that, prior to performing the operation, HyperSpy automatically rechunks the data, in this case wrongly.

### Description of the change

 This pull request fixes the issue by optimizing the re-chunking for the operation.

### Progress of the PR
- [x] add tests,
- [x] ready for review.

Running the following script uses 8GB of memory with HyperSpy v1.3.1 and just ~100MB with this patch.

```python
import dask.array as da
import hyperspy.api as hs
from hyperspy._signals.signal1d import LazySignal1D
data = da.random.random((512, 512, 4096), chunks=(4, 4, 4096))
s = LazySignal1D(data).as_lazy()
suma = s.sum()
suma.compute()
```


francisco-dlp:
I'll write a test for this once I get back feedback on the approach.

magnunor:
I've been working with `dask` quite a bit in pixStem, I'll have a look at this on Sunday.

Also, I have several functions and improvements which probably can be "upstreamed" to HyperSpy.

francisco-dlp:
Thanks @magnunor, I'll be looking forward to your feedback.

magnunor:
I think the better approach would be to utilize the already in-built `dask.array` functions. For example: https://dask.pydata.org/en/latest/array-api.html#dask.array.sum

francisco-dlp:
Thanks for the feedback. I think that in this case it is equivalent, dask casts the numpy functions to dask functions.

magnunor:
It is probably equivalent here, but in general using the dask.array functions directly means we do not need to manually handle removal of extra axes (and other "housekeeping" steps), since the dask.array functions does this for us.

Also, one way of getting multiprocessing for non-lazy signals could be to make the numpy array into a dask.array, and use the inbuilt dask functions. Since this is automatically multiprocessed. However, this will be very dependant on the type of processing.

magnunor:
It is probably equivalent here, but in general using the dask.array functions directly means we do not need to manually handle removal of extra axes (and other "housekeeping" steps), since the dask.array functions does this for us.

Also, one way of getting multiprocessing for non-lazy signals could be to make the numpy array into a dask.array, and use the inbuilt dask functions. Since this is automatically multiprocessed. However, this will be very dependant on the type of processing.

francisco-dlp:
Thanks @magnunor. As we have to keep everything working equivalently for numpy and dask arrays, I suggest to leave it as it is at least it is causing an issue.

What do you think about the fix in this PR?

magnunor:
Yeah, sounds like a good idea, especially for a Rnp release.

The fix looks fine, but I haven't had time to test it myself.

francisco-dlp:
Maybe somebody else could test it (@ericpre ?)

ericpre:
Yes, this fix is an improvement of the current implementation (from 8Gb to 100Mb...).
Other improvements suggested here will need significantly more work, so it is worth getting this merged and the rest can be done later in other PRs.

@francisco-dlp: can you add a test to check the memory usage?

francisco-dlp:
Thank @ericpre. I am not sure about how to write this test. I would like to avoid checking memory usage, preferring checking chunks instead. However, the chunks are currently internal. I'll probably split the chunk calculation into a separate method and check that it is doing the right thing.

francisco-dlp:
Regarding @magnunor comment concerning using ``dask`` for multiprocessing, I think that this is currently possible by simply calling ``sl = s.as_lazy()`` and operating on ``sl``.

francisco-dlp:
I have changed the rechunking approach from using HyperSpy's at least 1 signal per chunk approach to dask's automatic chunk guessing. For reduce operations there is no need to keep at least one signal per chunk, and as this can lead to bottlenecks when performing reduce operations on files with certain chunk configurations, I think than dask's approach is better in this situation.

francisco-dlp:
I have also added a test.

ericpre:
If you run the example of the first post of this PR, there is an error with the `da.rechunk` function which can't take `"auto"` for the `chunks` parameter.

francisco-dlp:
Actually it works for me. Probably it has something to do with the dask version.

francisco-dlp:
Which may be critical for the bundle, do you know which dask version is in the bundle?

ericpre:
Yes, it doesn't work with dask==0.17.4 but with the last version it is fine.

The dask version in the bundle is 0.15.4, see the [changelog](https://github.com/winpython/winpython/blob/master/changelogs/WinPythonQt5-64bit-3.6.3.0_History.md). An update of the dependency requirement may be good enough for now.

francisco-dlp:
Yes, it seems that the feature was introduced in v0.18, 18 days ago! I'll update the dep req. Good catch @ericpre, thanks!

ericpre:
Can you also call one of these function in the test? Because currently, the tests are not failing with an old version of dask.

francisco-dlp:
That's curious, because the code should be executed in [this line in the tests](https://github.com/hyperspy/hyperspy/pull/1986/files#diff-4ab6befca99bc4c5d36b3178549db2a2R924).

francisco-dlp:
I can confirm that this is the case (the test fails in my computer when downgrading dask).

ericpre:
Yes, my mistake!

Issue #1602, opened 2017-04-23
Adding a zero-order polynomial fails


thomasaarholt:
Adding a constant background to a model by adding a zero-order polynomial background fails when trying to look at it:

```python
m = s.create_model(auto_background=False)
m.add_polynomial_background(0)
m[-1]
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
/Users/thomas/miniconda3/envs/hs/lib/python3.5/site-packages/IPython/core/formatters.py in __call__(self, obj)
    670                 type_pprinters=self.type_printers,
    671                 deferred_pprinters=self.deferred_printers)
--> 672             printer.pretty(obj)
    673             printer.flush()
    674             return stream.getvalue()

/Users/thomas/miniconda3/envs/hs/lib/python3.5/site-packages/IPython/lib/pretty.py in pretty(self, obj)
    381                             if callable(meth):
    382                                 return meth(obj, self, cycle)
--> 383             return _default_pprint(obj, self, cycle)
    384         finally:
    385             self.end_group()

/Users/thomas/miniconda3/envs/hs/lib/python3.5/site-packages/IPython/lib/pretty.py in _default_pprint(obj, p, cycle)
    501     if _safe_getattr(klass, '__repr__', None) not in _baseclass_reprs:
    502         # A user-provided repr. Find newlines and replace them with p.break_()
--> 503         _repr_pprint(obj, p, cycle)
    504         return
    505     p.begin_group(1, '<')

/Users/thomas/miniconda3/envs/hs/lib/python3.5/site-packages/IPython/lib/pretty.py in _repr_pprint(obj, p, cycle)
    699     """A pprint that just redirects to the normal repr function."""
    700     # Find newlines and replace them with p.break_()
--> 701     output = repr(obj)
    702     for idx,output_line in enumerate(output.splitlines()):
    703         if idx:

/Users/thomas/hyperspy/hyperspy/_components/polynomial.py in __repr__(self)
     65     def __repr__(self):
     66         text = "%s order Polynomial component" % ordinal(
---> 67             self.get_polynomial_order())
     68         if self.name:
     69             text = "%s (%s)" % (self.name, text)

/Users/thomas/hyperspy/hyperspy/_components/polynomial.py in get_polynomial_order(self)
     48 
     49     def get_polynomial_order(self):
---> 50         return len(self.coefficients.value) - 1
     51 
     52     def function(self, x):

TypeError: object of type 'float' has no len()

```

[...]

Issue #1834, opened 2018-02-15
Different boundaries for parameters with number of elements >1


ZanettaPM:
Boundaries for parameter with number of elements > 1 (as `coefficients` in polynomial background for example) are just repeated for the different elements.  

here:  https://github.com/hyperspy/hyperspy/blob/137cfdadd2a7212f5104adb20fdd8c68a99af19a/hyperspy/component.py#L426

It could be important to be able to set different boundaries for the different elements, Notably for background components.

ericpre:
Closing since the Polynomial component have been rewritten and this is now possible to bound the coefficient values independently.

Issue #1667, opened 2017-06-13
`Polynomial` component parameters cannot be independently bounded


bm424:
In `Component`, parameters with multiple values (like polynomial coefficients) set the same lower bound for each one when `bmin` is set. Although this can be fixed by explicitly setting `_bounds` for the component, this isn't expected behaviour, I think. An example use-case is defining a line whose slope is known to be negative but whose y-intercept should be positive. Possible  syntax: `Polynomial.bmin = (-1, 1)`

thomasaarholt:
Good call. In the meantime, users could also define their own polynomial using `Expression` with independently bounded coefficients, but that's of course not ideal.

to266:
When someone attempts to fix this, I would rewrite the polynomial component to have as many `Parameter`s as there are actual parameters - and not have one tuple that is difficult to work with. The tuple is trivial to re-create when calling the `.function` method anyway.

ericpre:
Fixed in #1989.

PR #1989, opened 2018-06-30
Make Polynomial an Expression + bugfixes

thomasaarholt:
### Description of the change
I've rewritten the Polynomial component as an Expression, with the primary purpose of changing its parameter from a `coefficients` array to `order + 1` number of parameters, a, b, c etc. as needed.
I've also edited the Expression component to allow a case where the string expression can be just a parameter, i.e. `Expression("pi*r**2", name="Area")` instead of something that must involve `x`. I simultaneously raise a better error for the case when someone tries to force an Expression to be a numeric constant, i.e. "5"

This fixes:
#1602 
#1667
#1834 (because only Polynomial has `number_of_elements` > 1) 

### Progress of the PR
- [x] Rewrite Polynomial as Expression
- [x] Allow Expression to take non-x string expressions
- [x] update docstring,
- [x] update user guide,
- [x] add tests,
- [x] ready for review.

```python
>>> from hyperspy._components.polynomial import Polynomial
>>> offset = Polynomial(0, a=5)
>>> linear = Polynomial(1, a=1, b=2)
>>> quadratic = Polynomial(2, a=1, b=2, c=5)
```


hyperspy/_components/expression.py
```diff
@@ -156,9 +156,18 @@ class docstring.
     def compile_function(self, module="numpy", position=False):
         import sympy
         from sympy.utilities.lambdify import lambdify
+        try:# Expression is just a constant
+            int(self._str_expression)
```
francisco-dlp: Wouldn't ``float(`` be better?

hyperspy/_components/polynomial.py
```diff
@@ -16,58 +16,41 @@
...
 
     """n-order polynomial component.
 
-    Polynomial component defined by the coefficients parameters which is an
-    array of len the order of the polynomial.
-
-    For example, the [1,2,3] coefficients define the following 3rd order
-    polynomial: f(x) = 1x² + 2x + 3
+    Polynomial component consisting of order + 1 parameters.
+    The highest order parameter is "a", followed by the rest of the alphabet, i.e.
```
francisco-dlp: Looks nice. However, I think that it'll be better to use the other usual convention for naming the coefficients: a2 x² + a1 x + a0.

The advantages are:

* The number of letters of the alphabet is finite, numbers are not
* It's a lot easier to know the name of the parameter corresponding to a given power of x.

For easy sorting, it'll be good to use zero-padding e.g. ``a10 x**10 + a09 x**9 ...``

hyperspy/_components/polynomial.py
```diff
@@ -16,58 +16,41 @@
...
-
-from hyperspy.component import Component
+from hyperspy._components.expression import Expression
 from hyperspy.misc.utils import ordinal
+import string
+import numpy as np
 
-
-class Polynomial(Component):
+class Polynomial(Expression):
```
francisco-dlp, line 31: We cannot break the API in a minor release. What about calling the new component ``Polynomial2`` and keep the old ``Polynomial`` with a deprecation warning explaining the ``Polynomial`` is deprecated and it'll be replaced by ``Polynomial2`` in HyperSpy 2.0?

francisco-dlp (changes requested):
Just some minor comments to address before mergin. Nice works!

ericpre:
This is ready for review, below is a summary of what has been done to address the previous review comments and other bug fixes.

- [x] merge with upstream and solve conflicts,
- [x] use `a0`, `a1`, `a2`, etc. and zero padding when necessary,
- [x] restore `Polynomial` component and create Expression based `Polynomial2` component
- [x] Add deprecation warning, filter these in the test suite
- [x] use `Polynomial2` in `remove_background` to fix `function_nd` and enable corresponding test
- [x] remove support for zero order polynomial, not supported by sympy, see https://github.com/sympy/sympy/issues/5642, raise `ValueError` when appropriate and add tests

thomasaarholt:
LGTM! Thank you for finishing this!

francisco-dlp (approved):

francisco-dlp:
This looks good. However, I think that, since it is deprecating a component, this PR should be accompanied of another one that implements a mechanism to load hspy files with old Polynomial components with HyperSpy 2.0 by transforming them into the new Polynomial component.

ericpre:
> This looks good. However, I think that, since it is deprecating a component, this PR should be accompanied of another one that implements a mechanism to load hspy files with old Polynomial components with HyperSpy 2.0 by transforming them into the new Polynomial component.

Good point, I think it should be good now.

francisco-dlp:
I think that it'll be better to move that code to a new PR to RnM. In this way there'll be no need to check hyperspy's version. What should be checked instead is the version of hyperspy used to create the file. This information is stored in the hspy file but probably not in the component dictionary. I think that this is a bug, each component dictionary in the file should contain the version of the package where it is coming from because  a model can be stored in a hspy file using e.g v1.2, then read with hspy 1.4 and saved again without recreating the model and its components. Therefore the hspy version in the file may not be the one used to create the components.

By the way, any ``Polynomial2`` component should also be converted to ``Polynomial``.

An alternative would be to send this PR to RnM. In this way we avoid introducing a component, ``Polynomial2``, that will be removed in the next version.

ericpre:
It still fixes quite a few bugs, so it will be useful to merge it in `RELEASE_next_patch`.

Yes, the dictionary conversion is only relevant for `RELEASE_next_major` and once the commits of this PR land there, I will make a PR to sort out the deprecation, fix the issue with loading `Polynomial2` and take into account the case of `Polynomial` created with HyperSpy >=2.0 (no need for dictionary conversion).

francisco-dlp:
Personally I prefer not to merge this until we have a solution in place for loading obsolete components, including recording all required information in the hspy file. Otherwise we may end up releasing v1.4.2 without the features required to generate hspy files that can be properly read in future versions. I think that this is the first time that we change the API of a component, so it is the first time that we need to consider this issue. This is timely because with #2180 also external components may need to change their API and we need to provide a mechanism to do it.

ericpre:
@francisco-dlp: there is the attribute [`_id_version`](https://github.com/hyperspy/hyperspy/blob/1a8ce90794a494bda8e7dd18b959a90076438081/hyperspy/component.py#L748) but it is not used anywhere and it is not in the `_whitelist`.
In addition to `_id_version`, shall we add the name and the version of the library generating the component and add all these three attributes to the `_whitelist`?

ericpre:
Close #2129.

francisco-dlp:
I've just sent a PR https://github.com/thomasaarholt/hyperspy/pull/12 to merge this with #2180.

ericpre:
@francisco-dlp, if we want to get this finished today (which will be good), it may better to push your changes of https://github.com/thomasaarholt/hyperspy/pull/12 directly to this branch and I will review it.

francisco-dlp:
Done!

ericpre:
Thanks, I was wondering why there were still so many commits but the base branch was `RELEASE_next_patch`!

thomasaarholt:
Let me know if I need to do anything to help you guys. I appreciate you finishing this!

hyperspy/_components/polynomial_deprecated.py
```diff
@@ -0,0 +1,170 @@
...
+
+            Component.__init__(self, ['coefficients', ])
+            self._whitelist['order'] = ('init', order)
+            self.coefficients._number_of_elements = order + 1
+            self.coefficients.value = np.zeros((order + 1,))
+            self.coefficients.grad = self.grad_coefficients
+        else:
+            from hyperspy._components.polynomial import Polynomial
+            self.__class__ = Polynomial
+            print(kwargs)
```
ericpre: Remove this `print` call?

hyperspy/_signals/signal1d.py
```diff
@@ -55,6 +55,8 @@
 from hyperspy._signals.lazy import LazySignal
 from hyperspy.docstrings.signal1d import CROP_PARAMETER_DOC
 from hyperspy.docstrings.signal import SHOW_PROGRESSBAR_ARG, PARALLEL_ARG
+from hyperspy.misc.test_utils import ignore_warning
```
ericpre: This is not used anymore and it can be removed.

hyperspy/models/edsmodel.py
```diff
@@ -284,7 +286,8 @@ def add_polynomial_background(self, order=6):
         order: int
             The order of the polynomial
         """
-        background = create_component.Polynomial(order=order)
+        with ignore_warning(message="The definition of the `Polynomial`"):
+            background = create_component.Polynomial(order=order)
```
ericpre: By default, it will use the deprecated polynomial and it should use the new one by setting `legacy=False`.

hyperspy/_components/polynomial_deprecated.py
```diff
@@ -0,0 +1,170 @@
...
+        """
+        if legacy:
+            from hyperspy.misc.utils import deprecation_warning
+            msg = (
+                "The API of the `Polynomial` component will change in v2.0."
+                "To use the new API set `legacy=False`.")
+            deprecation_warning(msg)
+
+            Component.__init__(self, ['coefficients', ])
+            self._whitelist['order'] = ('init', order)
```
ericpre, line 75: I don't think this is necessary to add `legacy` to the whitelist, but I am mentioning it in case it has been overseen and I would have miss something.

hyperspy/tests/component/test_components.py
```diff
@@ -178,6 +178,7 @@ def test_function_nd(self):
         assert_allclose(o.function_nd(axis.axis), s.data)
 
 
+@pytest.mark.filterwarnings("ignore:The definition of the `Polynomial`")
 class TestPolynomial:
```
ericpre: Rename this to `TestDeprecatedPolynomial`.

hyperspy/tests/component/test_components.py
```diff
@@ -236,14 +238,107 @@ def test_3d_signal(self):
...
+        poly2 = hs.model.components1D.Polynomial(order=2, legacy=False)
+        _ = poly2._load_dictionary(poly2_dict)
+        assert poly2.a2.value == 1
+        assert poly2.a2._bounds == (None, None)
+        assert poly2.a1.value == 2
+        assert poly2.a1._bounds == (10, 0.0)
+        assert poly2.a0.value == 3
+
+
+class TestPolynomial2:
```
ericpre: Rename this to `TestPolynomial`.

hyperspy/tests/component/test_components.py
```diff
@@ -178,6 +178,7 @@ def test_function_nd(self):
         assert_allclose(o.function_nd(axis.axis), s.data)
 
 
+@pytest.mark.filterwarnings("ignore:The definition of the `Polynomial`")
```
ericpre:
```suggestion
@pytest.mark.filterwarnings("ignore:The API of the `Polynomial`")
```

hyperspy/tests/drawing/test_figure.py
```diff
@@ -63,6 +63,7 @@ def test_close_figure_with_plotted_marker(self):
         s._plot.close()
         self._assert_figure_state_after_close(s._plot.signal_plot)
 
+    @pytest.mark.filterwarnings("ignore:The definition of the `Polynomial`")
```
ericpre:
```suggestion
    @pytest.mark.filterwarnings("ignore:The API of the `Polynomial`")
```

hyperspy/tests/model/test_model.py
```diff
@@ -883,6 +894,7 @@ def test_std1_red_chisq_in_range(self):
         np.testing.assert_allclose(self.m.red_chisq.data, 0.86206965)
 
 
+@pytest.mark.filterwarnings("ignore:The definition of the `Polynomial`")
```
ericpre:
```suggestion
@pytest.mark.filterwarnings("ignore:The API of the `Polynomial`")
```

hyperspy/signal_tools.py
```diff
@@ -540,127 +540,32 @@ class Load(t.HasTraits):
 
 @add_gui_method(toolkey="hyperspy.Signal1D.contrast_editor")
 class ImageContrastEditor(t.HasTraits):
-    mpl_help = "See the matplotlib SymLogNorm for more information."
     ss_left_value = t.Float()
     ss_right_value = t.Float()
```
ericpre: The resolution of the merge conflict is not correct.

hyperspy/_components/polynomial.py
```diff
@@ -17,61 +17,59 @@
...
+    """
 
-    def grad_coefficients(self, x):
-        return np.vstack([self.grad_one_coefficient(x, i) for i in
-                          range(self.coefficients._number_of_elements)])
+    def __init__(self, order=2, module="numexpr", **kwargs):
+        # Not to break scripts once we remove the legacy Polynomial
+        if "legacy" in kwargs:
+            del kwargs["legacy"]
+        from hyperspy.misc.utils import deprecation_warning
```
ericpre: This import is not used.

ericpre (changes requested):
This is a much approach than the previous one! There is a merge conflict which hasn't been resolved correctly but the rest is mostly tidying up the code.

ericpre:
@thomasaarholt, maybe you could follow up with the review of your PR in https://github.com/hyperspy/hyperspyUI/pull/187.

hyperspy/_components/polynomial_deprecated.py
francisco-dlp: I agree.

francisco-dlp:
All fixed, thanks @ericpre!

ericpre (approved):

francisco-dlp:
@ericpre, thanks for fixing that. For some reason, ``meld`` is taking the wrong decision on make conflicts lately.

ericpre:
It look to me that there was no conflict in this part of the file (I don't think that the contrast editor tool was changed at all in this branch), so indeed there should be something weird going on!

PR #1995, opened 2018-07-02
Optional rechunk keyword argument for many lazy operations

### LINKED ISSUE #1982: Transpose and lazy signals
Transposing lazy signals currently leads to very non-optimal chunking:

```python
import dask.array as da
from hyperspy._signals.signal2d import LazySignal2D

data = da.random.random((50, 50, 100, 100), chunks=(10, 10, 20, 20))
s = LazySignal2D(data)
s.data
dask.array<da.random.random_sample, shape=(50, 50, 100, 100), dtype=float64, chunksize=(10, 10, 20, 20)>

s.T.data
dask.array<rechunk-merge, shape=(100, 100, 50, 50), dtype=float64, chunksize=(50, 100, 50, 50)>

s1 = s.deepcopy()
s1.data = da.swapaxes(da.swapaxes(s.data, 0, 2), 1, 3)
s1.data
dask.array<swapaxes, shape=(100, 100, 50, 50), dtype=float64, chunksize=(20, 20, 10, 10)>
```

In this example using `s.T` leads to the dataset being split into just two chunks, which for large datasets means that working with this new transposed signal becomes more difficult.

However, using `da.swapaxes` the original chunking structure is preserved, potentially making it easier to work with.

```python
s1 = s.deepcopy()
s1.data = da.swapaxes(da.swapaxes(s.data, 0, 2), 1, 3)
s1.data
dask.array<swapaxes, shape=(100, 100, 50, 50), dtype=float64, chunksize=(20, 20, 10, 10)>
```

One possibility could be to have `rechunk` as an argument in `LazySignal2D` (https://github.com/hyperspy/hyperspy/blob/RELEASE_next_minor/hyperspy/_signals/lazy.py#L874), but I'm not certain what the default behaviour should be.

francisco-dlp:
Fixes #1982

### Description of the change

* Add ``chunks`` argument for ``hspy`` format saving.
* Add ``rechunk`` argument to all reduce operations and ``change_dtype``
* Make ``transpose`` honor the ``optimize`` keyword. Previously this keyword was ignored. My interpretation to fix this is that ``rechunk`` is the equivalent of ``optimize`` for lazy signals.
* Add ``optimize`` keyword arguments to ``roll_axis`` and ``swap_axes`` for consistency with ``transpose``.
* Enable ``swap_axes`` for lazy signals.
  
### Progress of the PR
- [x] add tests,
- [x] ready for review.


ericpre:
@francisco-dlp, can you please solve the conflict?

francisco-dlp:
Done, I'll add a few more tests and it'll be done.

hyperspy/signal.py
```diff
@@ -50,7 +50,7 @@
 from hyperspy.misc.slicing import SpecialSlicers, FancySlicing
 from hyperspy.misc.utils import slugify
 from hyperspy.docstrings.signal import (
-    ONE_AXIS_PARAMETER, MANY_AXIS_PARAMETER, OUT_ARG, NAN_FUNC, OPTIMIZE_ARG)
+    ONE_AXIS_PARAMETER, MANY_AXIS_PARAMETER, OUT_ARG, NAN_FUNC, OPTIMIZE_ARG, RECHUNK_ARG)
```
ericpre, line 53: The `RECHUNK_ARG` also needs to be added to `diff`, `derivative`, `integrate_simpson`, `integrate1D`, `valuemax`, `valuemin` and `get_histogram`.

Also update the docstring of `change_dtype` and possibly other.

francisco-dlp:
I have now added tests for all the methods that go the new ``rechunk`` keyword.

Does it look good @ericpre ?

francisco-dlp:
Sorry, I forgot to add a test for ``change_dtype``, I've just uploaded it.

hyperspy/tests/signal/test_tools.py
```diff
@@ -920,6 +918,40 @@ def test_lazy_changetype_rechunk():
...
+    s = signals.Signal2D(ar).as_lazy()
+    s._make_lazy(rechunk=True)
+    assert s.data.dtype is np.dtype('uint8')
+    chunks_old = s.data.chunks
+    s.change_dtype('float', rechunk=False)
+    assert s.data.dtype is np.dtype('float')
+    assert chunks_old == s.data.chunks
+
+
+def test_lazy_reduce_rechunk():
```
ericpre, line 932: Use @pytest.mark.parametrize here for the list of method?

hyperspy/tests/signal/test_tools.py
```diff
@@ -158,6 +158,12 @@ def test_print_summary(self):
         # Just test if it doesn't raise an exception
         self.signal._print_summary()
 
+    def test_print_summary_statistics(self):
+        # Just test if it doesn't raise an exception
+        self.signal.print_summary_statistics()
+        if self.signal._lazy:
+            self.signal.print_summary_statistics(rechunk=False)
```
ericpre, line 165: This method uses a lot of memory when I guess it shouldn't.

ericpre (commented):
A few quick and easy comments but I could not look at it in depth... It look sensible overall.

hyperspy/tests/signal/test_tools.py
francisco-dlp: That can happen when the chunking is not optimal (unsurprising since we don't rechunk) but, are you sure? The signal is rather small.

hyperspy/tests/signal/test_tools.py
francisco-dlp: I agree that this would lead to clearer test failures, but it's not straightforward here since the signal is created internally. As it does the job and there is not much time left, I suggest to leave it as it is for this release.

hyperspy/signal.py
francisco-dlp: ok, working on it.

francisco-dlp: Done

francisco-dlp:
Travis is happy. If this looks good, I suggest to merge it as Appveyor seems to take ages to launch a build today and it'll be nevertheless tested in conda-forge and the hspy_bundle before making a release.

hyperspy/tests/signal/test_tools.py
ericpre: I have tested it using the example in #1986/

ericpre:
Ok, merging this, considering that we will need to improve it once we have a clearer idea on how to do this in a better way.

hyperspy/tests/signal/test_tools.py
francisco-dlp: I've just run it and memory consumption doesn't seem aberrant to me. This method is very simply, and if there is an issue with it I guess that it must be unrelated to HyperSpy

PR #1999, opened 2018-07-03
Change logger level from "warning" to "info" when data are made contiguous

ericpre:
Change the logger level to avoid getting a warning when loading DM files, which can confuse users. I suggest to try to get in the 1.3.2 release if this is fine as it is, if not we can leave it for latter. I submit it now, because I keep forgetting making this PR and I just came across it while reviewing #1995.


francisco-dlp:
OK, I'll wait for travis to finish.

ericpre:
Appveyor is fine:
https://ci.appveyor.com/project/ericpre/hyperspy/build/1.0.458

PR #2003, opened 2018-07-06
plot remainder in remove_background

AEljarrat:
Hi everyone!
this is a proposal to add an additional line to `Signal1d.remove_background` previewing the result of the operation (not unlike to a similar tool in a famous electron microscopy software). This PR is a naive implementation of this idea, which I hope could be useful to many. If you like the idea, I am open to suggestions how to improve it and willing to extend the PR with some tests and updating the UG.

### Description of the change
This PR adds a line to the `remove_background` plot that shows the signal - background result. This line is obtained from the data that is already generated by this tool, so that its as fast as possible. The new line is in green color.

### Progress of the PR
- [x] Change implemented 
- [x] update docstring
- [x] update UG
- [x] add tests,
- [x] ready for review

```python
axd = {'name':'awesome',
       'units':'^_^',
       'scale':0.1,
       'offset':-10.,
       'size':512}
s = hs.signals.Signal1D(np.random.rand(axd['size']), axes=[axd,])
s = hs.signals.BaseSignal(np.ones((2,2))).T * s

m = s.create_model()
vpm = hs.model.components1D.VolumePlasmonDrude()
zlp = hs.model.components1D.Gaussian(1000.)
m.append(vpm)
m.append(zlp)
s = m.as_signal(show_progressbar=False)

s.remove_background(plot_remainder=True)
```
The green line showing the remainder should appear when a spectral range is selected.


magnunor:
Nice! Could you include an image of how it looks like?

jat255:
This is pretty slick!

I just tried it out with the example data (which has the background already removed, so it looks funny), but it appears to be working as advertised:

image

AEljarrat:
@magnunor, this is how the MWE looks after a spectral range is selected and some zoom is applied:
figure__from_fitted_model_signal

ericpre (commented):
It needs a plotting test, other than that, it looks good.

thomasaarholt:
@AEljarrat I just wanted to say that that was a really cool way of creating a multidimensional hyperspy object - by multiplying a signal with shape (2,2|) with one with (|512)!

AEljarrat:
@ericpre I have finally added some plot test and also updated the docs. The PR should be now ready. Thanks for reviewing!
@thomasaarholt thanks for the nice words!

francisco-dlp:
@ericpre, is this good to merge now?

hyperspy/tests/signal/test_remove_background.py
```diff
@@ -59,3 +60,54 @@ def test_background_remove_pl_int(self):
             background_type='PowerLaw',
             show_progressbar=None)
         assert np.allclose(s1.data, np.zeros(len(s1.data)), atol=60)
+
+
+def test_BackgroundRemoval_gui():
```
ericpre: I find this test a bit too convolved and complicated.

ericpre:
I have one comments on the test but I added a plotting test. Other than that, it is fine to merge.

francisco-dlp:
In that case, let's wait to see if @AEljarrat wants to address the test issue. If we don't hear from him before the release and there are no further issues, I suggest merging it for v.1.4 as I bet users will enjoy this new feature.

hyperspy/tests/signal/test_remove_background.py
AEljarrat: You are probably right. It seems you have added a better test module. Should the old one be removed then?

ericpre: Yes, I would suggest be to remove it because it can happen that we spend more time to understand what does the tests than fixing the actual problem!!

AEljarrat: I have deleted the filed and pushed. The commit shows in my remote repository but not in this PR. I am lost. Any idea what could be happening?

AEljarrat: OK change pushed

AEljarrat:
Hi @francisco-dlp, I think the test issue is now solved.

francisco-dlp:
I'll leave @ericpre decide if that's the case and merge.

ericpre:
Thanks, this is nice. The y-autoscale will need to be improved but this out of the scope of this PR.

PR #1994, opened 2018-07-02
Fix issue with figure rendering for signal1D.

ericpre:
Fixes #1988.

PR #2006, opened 2018-07-11
Add tests for #1994

ericpre:
- [x] Add plotting test to check that the plot gets updated when data is changed.
- [x] Set matplotlib rcParams "interactive" to False when running the tests.
- [x] Ready for review.


francisco-dlp:
Setting interactive to False decreases the covering, why is that a good idea?

ericpre:
I think that coveralls may be confused: I have just checked it locally and setting `interactive=False` doesn't change the coverage.

If interactive is set to True in the rcParams of matplotlib, then the plotting tests are very slow, which is the reason to set it to False when running the tests.

PR #2008, opened 2018-07-13
Fix lazy FEI EMD SI

### LINKED ISSUE #18: True lazy reading to open EMD files
Currently lazy reading of EDX spectra in EMD files is not truly lazy. The data is stored in a compressed format in the file. HyperSpy reads the compressed data in memory and uncompresses it lazily. 

This lead to error messages if the number of frame is consequent.

Implementing pure lazy reading could permit to work with big datasets (i.e. cartography with too much frames)

### LINKED ISSUE #1966: True lazy reading to open EMD files
Currently lazy reading of EDX spectra in EMD files is not truly lazy. The data is stored in a compressed format in the file. HyperSpy reads the compressed data in memory and uncompresses it lazily. 

This lead to error messages if the number of frame is consequent.

Implementing pure lazy reading could permit to work with big datasets (i.e. cartography with too much frames)

francisco-dlp:
Partially fixes hyperspy/rosettasciio#18 in the sense that there may be no need to implement the feature lazily after all.

The Signal class was actually loading the data to memory because it didn't know how to deal with the sparse array. This commits wraps the sparse array in a dask array before it gets into the Signal constructor.

### DIFF
--- a/hyperspy/misc/io/fei_stream_readers.py
+++ b/hyperspy/misc/io/fei_stream_readers.py
@@ -1,4 +1,5 @@
 import numpy as np
+import dask.array as da
 
 from hyperspy.decorators import jit_ifnumba
 
@@ -202,7 +203,9 @@ def stream_to_sparse_COO_array(
             first_frame=first_frame,
             last_frame=last_frame,
         )
-    return DenseSliceCOO(coords=coords, data=data, shape=shape)
+    dense_sparse = DenseSliceCOO(coords=coords, data=data, shape=shape)
+    dask_sparse = da.from_array(dense_sparse, chunks="auto")
+    return dask_sparse
 
 
 @jit_ifnumba()


ericpre:
If I understand correctly, the sparse array was kept in memory (which is not a big problem) but the memory error happen when the sparse array was converted to a dense array at a later stage after all the data was imported?

francisco-dlp:
Yes, that's correct.

hyperspy/misc/io/fei_stream_readers.py
```diff
@@ -202,7 +203,9 @@ def stream_to_sparse_COO_array(
             first_frame=first_frame,
             last_frame=last_frame,
         )
-    return DenseSliceCOO(coords=coords, data=data, shape=shape)
+    dense_sparse = DenseSliceCOO(coords=coords, data=data, shape=shape)
+    dask_sparse = da.from_array(dense_sparse, chunks="auto")
```
ericpre, line 207: Following #1995, would it be sensible to start passing `chunks` parameters from the load function down to reader, in this case, here?

ericpre (commented):
Ok, looks good, I put one comment which I am not sure if we should address in this PR or in a separate one.

hyperspy/misc/io/fei_stream_readers.py
francisco-dlp: Yes, but I suggest to leave it for another PR.

PR #2009, opened 2018-07-13
Update warning for decomposition of non-float types

jat255:
Per [discussion](https://gitter.im/hyperspy/hyperspy?at=5b48cba9c2d95c60f4d7777a) on gitter chat... change the warning for decomposition to a logger error, since in this situation HyperSpy does something different than what the user expects.


francisco-dlp:
Personally I had in mind a proper error for this e.g. ``raise TypeError``

jat255:
Agreed and changed.

francisco-dlp:
Thanks. Could you add a test for this? A simple one, just testing that the right dtype does not raise an error while the wrong one does.

jat255:
Sure thing; done

PR #2010, opened 2018-07-17
Widget enhancement

ericpre:
### Description of the change
Add a few enhancements to the widgets required for #1668.

### Progress of the PR
- [x] Support blit for range widgets,
- [x] support vertical widgets on image (for #1668)
- [x] add color parameters to widgets,
- [x] add alpha parameters to widgets,
- [x] add tests,
- [x] ready for review.

PR #2011, opened 2018-07-18
Fix roi widget direction

francisco-dlp:
### Description of the change

Calling ``SpanROI.add_widget`` with axes a vertical axis wrongly added a horizontal span widget. This fixes this issue.

In addition, this adds ``**kwargs`` to the ``BaseROI.add_widget`` method that are passed to the widget constructor and add tests for all ROI widgets.

I am sending this to Rnm because it is based on #2010.

### Progress of the PR
- [x] add tests,
- [x] ready for review.

```python
%matplotlib qt

import hyperspy.api as hs
import numpy as np

s = hs.signals.Signal2D(np.random.random((100, 1000)))
s.plot()
roi = hs.roi.SpanROI(10, 50)
w = roi.add_widget(s, axes=[s.axes_manager.signal_axes[1]])
```

Before:
spanroi_widget_broken

With this PR:
spanroi_widget_fixed


hyperspy/roi.py
```diff
@@ -516,26 +542,13 @@ def _apply_roi2widget(self, widget):
...
-                return widgets.HorizontalLineWidget
-        elif plotdim == 1:  # It is a spectrum
+        direction = guess_vertical_or_horizontal(axes=axes, signal=signal)
+        if direction == "vertical":
             return widgets.VerticalLineWidget
+        elif direction == "horizontal":
+            return widgets.HorizontalLineWidget
         else:
-            raise ValueError("Could not find valid widget type")
+            raise ValueError("direction must be either horizontal or vertical")
```
ericpre, line 554: This error message implies that the user is providing the parameters `direction` wrongly, however the user is not providing the `direction` parameter but at best, it is providing `axes`, which by default is set to `None`. Would it be better to say:
"The direction of the widget can't be guessed, the `axes` parameter needa to be changed."

hyperspy/roi.py
```diff
@@ -628,7 +641,13 @@ def _apply_roi2widget(self, widget):
...
 
     def _get_widget_type(self, axes, signal):
-        return widgets.RangeWidget
+        direction = guess_vertical_or_horizontal(axes=axes, signal=signal)
+        if direction == "vertical":
+            return partial(widgets.RangeWidget, direction="horizontal")
+        elif direction == "horizontal":
+            return partial(widgets.RangeWidget, direction="vertical")
+        else:
+            raise ValueError("direction must be either horizontal or vertical")
```
ericpre, line 653: Same as above regarding error message.

hyperspy/roi.py
```diff
@@ -486,6 +490,28 @@ def __call__(self, signal, out=None, axes=None):
...
+    if plotdim == 2:  # Plot is an image
+        # axdim == 1 and plotdim == 2 indicates "spectrum stack"
+        if idx == 0 and axdim != 1:    # Axis is horizontal
+            return "vertical"
+        else:  # Axis is vertical
+            return "horizontal"
+    elif plotdim == 1:  # It is a spectrum
+        return "vertical"
+    else:
+        raise ValueError("Could not find valid widget type")
```
ericpre: Similar to below: more clear error message mentioning that the axes provided doesn't allow to determine the appropriate widget?

hyperspy/roi.py
```diff
@@ -486,6 +490,28 @@ def __call__(self, signal, out=None, axes=None):
         return s
 
 
+def guess_vertical_or_horizontal(axes, signal):
+    # Figure out whether to use horizontal or veritcal line:
```
ericpre, line 495: To be more consistent with RangeWidget, I would suggest that this function determine the "movement direction" of the widget rather than the line direction.

hyperspy/roi.py
ericpre: It looks to me that this error message will never be called, because the error will occur before in the `guess_vertical_or_horizontal` function.

hyperspy/roi.py
```diff
@@ -628,7 +641,13 @@ def _apply_roi2widget(self, widget):
         widget.set_bounds(left=self.left, right=self.right)
 
     def _get_widget_type(self, axes, signal):
-        return widgets.RangeWidget
+        direction = guess_vertical_or_horizontal(axes=axes, signal=signal)
```
ericpre, line 647: If the return of the `guess_vertical_or_horizontal` function is changed appropriately as mentioned above, this could be simplified to:
```python
return partial(widgets.RangeWidget, direction=guess_vertical_or_horizontal(axes=axes, signal=signal))
```

ericpre (commented):
A few suggestion to make more consistent and simple.

hyperspy/roi.py
francisco-dlp: That's true, but I think that it is good practice to leave the code as-is. In this way, if something changes elsewhere and the method is called with a wrong value it'll raise a meaningful error.

hyperspy/roi.py
francisco-dlp: As above ;)

hyperspy/roi.py
francisco-dlp: Done

hyperspy/roi.py
francisco-dlp: I don't get it, shouldn't it be the same?

hyperspy/roi.py
francisco-dlp: For the same reason as above, I would prefer to leave it as-is.

francisco-dlp:
I have now added tests.

hyperspy/roi.py
ericpre: The following code should illustrate what I mean.
```python
import hyperspy.api as hs
import numpy as np

s = hs.signals.Signal2D(np.random.random((100, 1000)))
s.plot()

roi = hs.roi.SpanROI(10, 50)
w = roi.add_widget(s, axes=[s.axes_manager.signal_axes[1]])
# `guess_vertical_or_horizontal` returns `horizontal`
# RangeWidget is instantiated with direction=`vertical`
assert w.span.direction == 'vertical'

roi2 = hs.roi.Point1DROI(5)
w2 = roi2.add_widget(s, axes=[s.axes_manager.signal_axes[1]])
# `guess_vertical_or_horizontal` returns `horizontal`
# HorizontalLineWidget is instantiated
assert isinstance(w2, hs.hyperspy.drawing.widgets.HorizontalLineWidget)
```
The current definition of direction doesn't match the one of `SpanROI`, which follows the one of matplotlib `SpanSelector`.

francisco-dlp: I see, however I don't see the issue as the code works as it should.

hyperspy/tests/drawing/test_plot_roi_widgets.py
```diff
@@ -0,0 +1,143 @@
...
+class TestPlotROI():
+
+    def setup_method(self, method):
+        # Create test image 100x100 pixels:
+        self.im = Signal2D(np.arange(50000).reshape([10, 50, 100]))
+        self.im.axes_manager[0].scale = 1e-1
+        self.im.axes_manager[1].scale = 1e-2
+        self.im.axes_manager[2].scale = 1e-3
+
+    @pytest.mark.mpl_image_compare(BASELINE_DIR=BASELINE_DIR,
```
ericpre: The image comparison tests are not working because `BASELINE_DIR=BASELINE_DIR` should not be capitalised (`baseline_dir=BASELINE_DIR`).

francisco-dlp: I've just fixed it: I've trusted too much the refactor feature of VSCode.

hyperspy/roi.py
ericpre: Yes, very true and this is why this is just a suggestion to make it more consistent! This is not really important, because the widgets are not directly exposed in the public API, however someone  instantiating directly this kind of widgets may be surprised about this -at least I was!

francisco-dlp: That's true. Maybe we could consider changing the definition of ``direction`` in the range widget to be consistent with that of Line in another PR?

ericpre: I would prefer the other way around, because the RangeWidget wraps a matplotlib SpanSelector, so if we need to choose between one the two definitions, I would prefer to stay consistent with the matplotlib one.

francisco-dlp: But that would result in a horizontal ``VerticalLineWidget`` isn't it?

ericpre: Yes, indeed, a _horizontally moving_ `VerticalLineWidget`! :D

PR #2014, opened 2018-07-19
Friendler testing guide to new users

thomasaarholt:
- Replaced py.test with pytest, as the website only uses the latter nowadays (and I was confused initially a to whether the hs docs were speaking about one or two different packages)
- Added install commands for pytest and pytest-mpl (the latter of which pytest will fail without). 
- Also added a reference to pytest-html for easier reading of the test results.

### DIFF
--- a/doc/dev_guide.rst
+++ b/doc/dev_guide.rst
@@ -164,18 +164,26 @@ outputs when coding. Writing tests can seem laborious but you'll probably
 soon find that they're very important as they force you to sanity check all
 you do.
 
-HyperSpy uses the `py.test <http://doc.pytest.org/>`_ library for testing. The
-tests reside in the ``hyperspy.tests`` module. To run them:
+HyperSpy uses the `pytest <http://doc.pytest.org/>`_ library for testing. The
+tests reside in the ``hyperspy.tests`` module. 
+
+First ensure pytest and its plugins are installed:
 
 .. code:: bash
 
-   py.test --pyargs hyperspy
+   conda install pytest pytest-mpl freetype 
+
+To run them:
+
+.. code:: bash
+
+   pytest --pyargs hyperspy
 
 Or, from HyperSpy's project folder simply:
 
 .. code:: bash
 
-   py.test
+   pytest
 
 
 Useful hints on testing:
@@ -207,7 +215,7 @@ Useful hints on testing:
   PR page. This service can help you to find how well your code is being tested
   and exactly which part is not currently tested.
 * `pytest-sugar <https://pypi.python.org/pypi/pytest-sugar>`_ can be installed
-  to have a nicer look and feel of py.test in the console (encoding issue have
+  to have a nicer look and feel of pytest in the console (encoding issue have
   been reported in the Windows console).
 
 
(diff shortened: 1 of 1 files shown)


doc/dev_guide.rst
```diff
@@ -164,18 +164,26 @@ outputs when coding. Writing tests can seem laborious but you'll probably
...
-tests reside in the ``hyperspy.tests`` module. To run them:
+HyperSpy uses the `pytest <http://doc.pytest.org/>`_ library for testing. The
+tests reside in the ``hyperspy.tests`` module. 
+
+First ensure pytest and its plugins are installed:
+
+.. code:: bash
+
+   conda install pytest pytest-mpl freetype 
+
```
francisco-dlp, line 180: I would add the alternative way of installing the deps with pip

```bash
pip install hyperspy[test]
```

thomasaarholt:
Done!

doc/dev_guide.rst
```diff
@@ -167,10 +167,13 @@ you do.
...
-First ensure pytest and its plugins are installed:
+First ensure pytest and its plugins are installed by:
 
 .. code:: bash
-
+   
+   # Either
+   pip install hyperspy[test]
+   # Or
    conda install pytest pytest-mpl freetype
```
ericpre: freetype should be already installed and it should not be required here?

doc/dev_guide.rst
```diff
@@ -167,10 +167,13 @@ you do.
...
 tests reside in the ``hyperspy.tests`` module. 
 
-First ensure pytest and its plugins are installed:
+First ensure pytest and its plugins are installed by:
 
 .. code:: bash
-
+   
+   # Either
+   pip install hyperspy[test]
```
ericpre, line 175: Not sure if this would install the release version, but most likely, the developer want to install with the pip development mode:
```
pip install .[test]
```
Need to be checked if this work nicely.

doc/dev_guide.rst
thomasaarholt: It was specificed in the [dev_guide](https://github.com/hyperspy/hyperspy/blob/57800eab365201c81309f76fe18b18b161c1b375/doc/dev_guide.rst#plot-testing) as requiring minimum version 2.8. I don't know what it is that requires freetype, and since it isn't mentioned anywhere else I thought to include it. I'll remove it from this PR.

doc/dev_guide.rst
thomasaarholt: I think you mean `pip install -e .[test]`? Although the dependency check will be the same for the [test] part.

I think `pip install .[test]` would overwrite the any current hyperspy develop install with a non-develop one.

thomasaarholt: I can confirm that `pip install -e .[test]` installs hyperspy as a dev install and also installs the dependencies.

thomasaarholt:
I've taken into account the suggestions.

doc/dev_guide.rst
francisco-dlp: Freetype may not be installed in miniconda, so I think that it's better to leave it here.

LGTM, @ericpre, if you agree with the above, could you merge this?

ericpre: freetype is a dependency of matplotlib, I don't think it is necessary to put it here and running `conda install freetype` after installing hyperspy doesn't do nothing.

doc/dev_guide.rst
```diff
@@ -164,18 +164,31 @@ outputs when coding. Writing tests can seem laborious but you'll probably
...
+   pip install -e .[test]
+   # Or just installing the dependencies using conda
+   conda install pytest pytest-mpl freetype
+
+To run them:
 
-   py.test --pyargs hyperspy
+.. code:: bash
+
+   pytest --pyargs hyperspy
```
ericpre: pytest --mpl --pyargs hyperspy

thomasaarholt:
Done. Merge?

PR #2015, opened 2018-07-21
Fix coveralls badge pointer to the wrong branch.

ericpre:
The coveralls badge is pointing to the wrong branch -the master branch which doesn't exist anymore. The proper fix would change the default branch in coveralls, but it seems that there are issues with doing that.
See https://github.com/lemurheavy/coveralls-public/issues/862 and https://github.com/lemurheavy/coveralls-public/issues/36.

### Progress of the PR
- [x] Update coveralls badge, now pointing to RELEASE_next_minor,
- [x] ready for review.


francisco-dlp:
Shouldn't this go to Rnm? And shouldn't we change the Rnp and RnM badges to point to the right branches too?

ericpre:
Initially, I thought to keep it simple but it does make that each branch point to the correct page. I did the other branches in two PRs and I update appveyor and travis too.

PR #2016, opened 2018-07-23
Add add_powerlaw argument to get_core_loss_eels_signal/model

magnunor:
The `artificial_data.get_core_loss_eels_signal` and `artificial_data.get_core_loss_eels_model` currently returns an artificial core less EEL spectrum without a power law background. This is fine for some cases, but for others (like https://github.com/hyperspy/hyperspy/pull/2003#issuecomment-403091405) having it would be nice.
 
So this pull requests add a new `add_powerlaw` argument to the `artificial_data.get_core_loss_eels_signal` and `artificial_data.get_core_loss_eels_model` functions.

### Progress of the PR
- [x] Add `add_powerlaw` argument to both functions
- [x] Update docstring
- [ ] Update user guide
- [x] add tests,
- [ ] ready for review.

### Minimal example
```python
import hyperspy.api as hs
s = hs.datasets.artificial_data.get_core_loss_eels_signal(add_powerlaw=True)
s.plot()
```

with_powerlaw

Without the power law

```python
import hyperspy.api as hs
s = hs.datasets.artificial_data.get_core_loss_eels_signal(add_powerlaw=True)
s.plot()
```

without_powerlaw

### DIFF
--- a/hyperspy/datasets/artificial_data.py
+++ b/hyperspy/datasets/artificial_data.py
@@ -49,6 +49,9 @@ def get_core_loss_eels_signal(add_powerlaw=False):
 
     Similar to a Mn-L32 edge from a perovskite oxide.
 
+    Some random noise is also added to the spectrum, to simulate
+    experimental noise.
+
     Parameters
     ----------
     add_powerlaw : bool
@@ -69,6 +72,16 @@ def get_core_loss_eels_signal(add_powerlaw=False):
 
     >>> s = ad.get_core_loss_eels_signal(add_powerlaw=True)
     >>> s.plot()
+
+    To make the noise the same for multiple spectra, which can
+    be useful for testing fitting routines
+
+    >>> np.random.seed(seed=10)
+    >>> s1 = ad.get_core_loss_eels_signal()
+    >>> np.random.seed(seed=10)
+    >>> s2 = ad.get_core_loss_eels_signal()
+    >>> (s1.data == s2.data).all()
+    True
 
     See also
     --------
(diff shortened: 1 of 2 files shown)


magnunor:
I'm not really sure where to place this in the user guide, since the only place either `example_signals` or `artificial_data` is documented is in [Getting started](https://github.com/hyperspy/hyperspy/blob/RELEASE_next_minor/doc/user_guide/getting_started.rst#loading-example-data-and-data-from-online-databases), which I think shouldn't be filled up with a lot of extra details like this.

I guess the smart thing would be to add a new section in the user guide, about the various datasets like these which are available in HyperSpy, but that is a bit out of scope of this pull request (and my time schedule)...

thomasaarholt:
Would it be possible to add a "random=True" argument and code to the various artificial signals to allow them to be reproducible? I tried using them for testing puproses, but the fitted values vary slightly, which makes it less than ideal to work with for that purpose.

ericpre:
@thomasaarholt: you need to set the seed in your test.

ericpre (approved):
If you want to document it, you could mention it in the dev guide but other that it looks fine to me.

thomasaarholt:
@ericpre Thanks! I didn't realise I could just set that.

@magnunor maybe add a quick docstring reference to setting `np.random.seed(seed=some_number)` in order to make it reproducible.

magnunor:
@thomasaarholt, good idea! I added this in the docstring, and also added a unit test to check that two spectra generated using the same random seed really are exactly the same.

PR #2019, opened 2018-07-26
Small documentation updates

thomasaarholt:
### Description of the change
- sphinxcontrib-napoleon is incorporated into sphinx as of v1.3 (current is 1.7), so I have removed it from the instructions and install files, and updated conf.py to get it from the correct module.
- Add doc-install commands to developer instructions
- Explain what parameter.map is, for model fitting purposes

### Progress of the PR
- [x] tested changes,
- [x] ready for review.


setup.py
```diff
@@ -69,7 +69,7 @@
     "gui-jupyter": ["hyperspy_gui_ipywidgets"],
     "gui-traitsui": ["hyperspy_gui_traitsui"],
     "test": ["pytest>=3", "pytest-mpl", "matplotlib>=2.0.2"],
-    "doc": ["sphinx", "numpydoc", "sphinxcontrib-napoleon", "sphinx_rtd_theme"],
+    "doc": ["sphinx", "numpydoc", "sphinx_rtd_theme"],
```
francisco-dlp: ``sphinx>=1.3``?

doc/dev_guide.rst
```diff
@@ -266,15 +266,30 @@ specification <https://github.com/numpy/numpy/blob/master/doc/HOWTO_DOCUMENT
...
 
-User-guide Documentation -- A description of the functionality of the code and
+User-guide documentation -- A description of the functionality of the code and
 how to use it with examples and links to the relevant code.
 
-Build the documentation -- To check the output of what you wrote, you can build
+Build user-guide documentation
+^^^^^^^^^^^^^^^^^^^
+
+To check the output of what you wrote, you can build
```
francisco-dlp: What about replacing all this with:

```bash

pip install hyperspy[doc]
```

The advantage is that we only have to update the dependencies in setup.py, decreasing the risk of getting the docs out-of-sync.

doc/user_guide/model.rst
```diff
@@ -420,12 +420,22 @@ Getting and setting parameter values and attributes
...
+pixel indices in ``m.axes_manager.indices`` or as calibrated coordinates in
+``m.axes_manager.coordinates``.
+
 :py:attr:`~.component.Component.parameters` contains a list of the parameters
 of a component and :py:attr:`~.component.Component.free_parameters` lists only
 the free parameters.
 
-The value of a particular parameter can be accessed in the
-:py:attr:`~.component.Parameter.value`.
+The value of a particular parameter in the current coordinates can be
```
francisco-dlp, line 432: Wouldn't it be better to document ``Parameter.as_signal`` in the UG instead and move this to the source as comments for the developers? ``as_signal`` returns the same arrays but in a more user-friendly way. Users needing direct access to ``Parameter.map`` should be able to find their way by inspecting the code.

``map`` is not private because when I wrote this part of the code it hadn't yet crossed my mind that anybody else would use it, so I didn't follow any convention.

doc/dev_guide.rst
ericpre: Done and rename `docs` to `build-doc` to avoid a confusion: user may think that it will install the doc and the doc build tools.

doc/user_guide/model.rst
ericpre: I think it is still worth leaving this in the user guide, because I have seen users using map without going through the code...

PR #2021, opened 2018-07-27
Add width and range attribute to DataAxis to use range sliders.

ericpre:
As mentioned in #1668, this is an attempt (and work in progress) to support range sliders. I am making a separate PR, to keep things simple. Because of the way we need to links the attributes of the different objects, I could not find a better approach than adding a `width` and `range` attribute to `DataAxis`.

It would be good if I could get feedback on this and get this PR finished before working further on #1668, so I can adjust the last one accordingly.

You can test it using the following [hyperspy_gui_ipywidgets](https://github.com/ericpre/hyperspy_gui_ipywidgets/tree/axes_range_widget) branch and in the notebook:

```python
import hyperspy.api as hs
from hyperspy.axes import DataAxis, get_gui
hs.preferences.GUIs.enable_ipywidgets_gui = True
hs.preferences.GUIs.enable_traitsui_gui = False

axis = DataAxis(100, scale=0.1)
get_gui([axis], use_width=True, use_range=False, toolkey="navigation_sliders")

```

### DIFF
--- a/hyperspy/axes.py
+++ b/hyperspy/axes.py
@@ -77,6 +77,7 @@ class DataAxis(t.HasTraits):
     slice = t.Instance(slice)
     navigate = t.Bool(t.Undefined)
     index = t.Range('low_index', 'high_index')
+    width = t.Range('low_index', 'high_index')
     axis = t.Array()
     continuous_value = t.Bool(False)
 
@@ -112,6 +113,14 @@ class DataAxis(t.HasTraits):
             obj : The DataAxis that the event belongs to.
             value : The new value
             """, arguments=["obj", 'value'])
+        self.events.width_changed = Event("""
+            Event that triggers when the width of the `DataAxis` changes
+
+            Arguments:
+            ---------
+            obj : The DataAxis that the event belongs to.
+            width : The new width
+            """, arguments=["obj", 'width'])
         self._suppress_value_changed_trigger = False
         self._suppress_update_value = False
         self.name = name
@@ -122,6 +131,7 @@ class DataAxis(t.HasTraits):
         self.high_index = self.size - 1
         self.low_index = 0
         self.index = 0
+        self.width = 1
         self.update_axis()
         self.navigate = navigate
         self.axes_manager = None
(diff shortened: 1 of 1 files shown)


thomasaarholt:
It would have been really nice if the central part of the range slider was draggable. I did some digging, and it's been referred to here but not implemented: jupyter-widgets/ipywidgets#1351

A couple of things:
- Setting width to some number, and then increasing index to max currently resets width to 0.
- Raise a better error when setting both range and width to True
```python
>>> get_gui([axis], use_width=True, use_range=True, toolkey="navigation_sliders")
UnboundLocalError                         Traceback (most recent call last)
<ipython-input-2-05dbe2c86816> in <module>()
      6 
      7 axis = DataAxis(100, scale=0.1)
----> 8 get_gui([axis], use_width=True, use_range=True, toolkey="navigation_sliders")

~/hyperspy/hyperspy/ui_registry.py in get_gui(self, toolkey, display, toolkit, **kwargs)
    152         if toolkit in toolkits:
    153             used_toolkits.add(toolkit)
--> 154             thisw = f(obj=self, display=display, **kwargs)
    155             if not display:
    156                 widgets[toolkit] = thisw

~/hyperspy_gui_ipywidgets/hyperspy_gui_ipywidgets/utils.py in wrapper(*args, **kwargs)
     77     def wrapper(*args, **kwargs):
     78         display = kwargs.pop("display", True)
---> 79         wdict = f(*args, **kwargs)
     80         if display:
     81             IPython.display.display(wdict["widget"])

~/hyperspy_gui_ipywidgets/hyperspy_gui_ipywidgets/axes.py in ipy_navigation_sliders(obj, use_width, use_range, **kwargs)
     48             def update_bound(change):
     49                 width_widget.max = axis.size - change.new - 1
---> 50             index_widget.observe(update_bound, "value")
     51             link((axis, "width"), (width_widget, "value"))
     52 

UnboundLocalError: local variable 'index_widget' referenced before assignment
```

ericpre:
This is WIP to discuss the approach, there are quite a few things broken...

thomasaarholt:
I see. I didn't realise from your post that you were asking for feedback on the approach. I was tired after a long flight though, but was interested so I had a go.

The first post in #1668 isn't very clear after it's edits. To a newcomer it was a bit difficult to set myself into what was being attempted achieved, a that confusion probably bled over into my answer.

PR #1767, opened 2017-10-22
New set edge onset cli

magnunor:
Same as https://github.com/hyperspy/hyperspy/pull/1271, but without the interactive parts (since I don't have the time to make it work). Related: https://github.com/hyperspy/hyperspy/pull/490, and https://github.com/hyperspy/hyperspy/issues/487.

This implements a way to set the edge onsets of core loss edges, as outlined in https://doi.org/10.1016/j.ultramic.2016.07.004 (arXiv: https://arxiv.org/abs/1608.07814)

- [ ] Update user guide
- [ ] Add more tests
- [ ] Ready for review

PR #1271, opened 2016-08-26
New get edge onset

magnunor:
Resubmitted version of https://github.com/hyperspy/hyperspy/pull/490.

Need to update it to the current newest RELEASE_next_minor.

Also, the article which explains how the fitting works + example on how to use it has been published: http://authors.elsevier.com/a/1TR8E15DbnCdAj , especially Figure 2. Is it ok to include a link to the article in the docstring and user guide? I'm also planning on putting the preprint version on arxiv, so everyone can access it.
- [ ] Update to newest branch
- [ ] Improve docstrings
- [ ] Improve old code

PR #2022, opened 2018-07-29
EELSModel.estimate_and_set_coreloss_edge_onset

magnunor:
As discussed a long time ago in https://github.com/hyperspy/hyperspy/issues/487, there is currently no good way of setting the onsets of the `EELSCLEdge` components.

This pull requests implements the method of doing this as outlined in https://doi.org/10.1016/j.ultramic.2016.07.004 (arXiv: https://arxiv.org/abs/1608.07814).

The method itself is explained here: https://github.com/hyperspy/hyperspy/pull/490#issue-32186703

### Changes

- Adds `misc.eels.tools.get_edge_onset`: to find onset using the method explained above
- Adds `misc.utils.find_nearest_index` and `misc.utils.find_nearest_index_from_right`: which are utility functions for `misc.eels.tools.get_edge_onset`
- `EELSModel.set_coreloss_edge_onset`: which is the main function people will use, which calls either the UI function in `signal_tools` or `EELSModel._set_coreloss_edge_onset`.
- `signal_tools.SetCorelossEdgeOnset`: which handles the UI parts of the implementation
- New widget in `hyperspy_gui_traitsui`: https://github.com/hyperspy/hyperspy_gui_traitsui/pull/12
- New widget in `hyperspy_gui_ipywidgets`: https://github.com/hyperspy/hyperspy_gui_ipywidgets/pull/7

### Progress of the PR
- [x] Implemenation
- [x] Add docstrings
- [x] update user guide
- [x] add tests
- [x] ready for review

```python
import hyperspy.api as hs
s.set_microscope_parameters(
        beam_energy=200, convergence_angle=26, collection_angle=20)
s.add_elements(['Mn', ])
m = s.create_model(auto_background=False, GOS='hydrogenic')
m.set_coreloss_edge_onset(m.components.Mn_L3)
```

set_coreloss_edge_onset

Old pull requests: https://github.com/hyperspy/hyperspy/pull/1767, https://github.com/hyperspy/hyperspy/pull/1271, https://github.com/hyperspy/hyperspy/pull/490


hyperspy/misc/eels/tools.py
```diff
@@ -307,3 +309,63 @@ def eels_constant(s, zlp, t):
...
+    >>> s = hs.datasets.artificial_data.get_core_loss_eels_signal()
+    >>> data, x_axis = s.data, s.axes_manager[0].axis
+    >>> onset = get_edge_onset(data, 600, 700, x_axis, 0.1)
+
+    """
+    if not ((percent_position >= 0.) and (percent_position <= 1.)):
+        raise ValueError(
+                "percent position ({0}) must be between 0 and 1".format(
+                    percent_position))
+    start_i = utils.find_nearest_index(x_axis, start)
```
francisco-dlp: Why not using ``value2index`` instead? Avoiding adding extra functions whenever possible should help easing the task of maintaining the code.

hyperspy/misc/eels/tools.py
```diff
@@ -307,3 +309,63 @@ def eels_constant(s, zlp, t):
...
+    end_i = utils.find_nearest_index(x_axis, end)
```
francisco-dlp: As above.

hyperspy/misc/eels/tools.py
```diff
@@ -307,3 +309,63 @@ def eels_constant(s, zlp, t):
...
+
+    data = data[start_i:end_i]
+    x_axis = x_axis[start_i:end_i]
+
+    data_max = data.max()
+    data_min = data.min()
+    data_onset_value = (data_max - data_min) * percent_position + data_min
+
+    interpolate = interp1d(x_axis, data, fill_value='extrapolate')
+    x_axis_interp = np.arange(x_axis[0], x_axis[-1], 0.0001)
```
francisco-dlp: Wouldn't it be possible to use ``spline.roots`` for this instead of computing the interpolated spectrum with the step hardwired?

Beside, what's the point of using interpolation to estimate the onset position using a method that is only qualitative?

I suggest to make interpolation optional. This would have the advantage that the algorithm could be vectorized boosting performance.

hyperspy/misc/eels/tools.py
```diff
@@ -307,3 +309,63 @@ def eels_constant(s, zlp, t):
...
+    data_max_i = data_interpolated.argmax()
+    data_min_i = data_interpolated.argmin()
+
+    if data_max_i < data_min_i:
+        temp_i = data_max_i
+        data_max_i = data_min_i
+        data_min_i = temp_i
+
+    threshold = 0.0001 * (data_max - data_min)
+    data_onset_i = utils.find_nearest_index_from_right(
```
francisco-dlp: isn't it possible to use ``value2index`` here too?

hyperspy/models/eelsmodel.py
```diff
@@ -976,3 +978,82 @@ def resume_auto_fine_structure_width(self, update=True):
                 self.resolve_fine_structure()
         else:
             warnings.warn("Not suspended, nothing to resume.")
+
+    def set_coreloss_edge_onset(
```
francisco-dlp: What about calling this ``estimate_and_set_coreloss_edge_onset``?

francisco-dlp (changes requested):
Nice and polished contribution!

hyperspy/misc/eels/tools.py
magnunor: This doesn't work on `EELSSpectrum` but directly on numpy arrays, so `value2index` can't be called here. The reasoning for that is trying to have a bit cleaner separation between the HyperSpy signal parts, and the "calculation" parts.

magnunor:
I reimplemented most of the `get_edge_onset` function, so it no longer interpolates the whole spectrum. Now it finds the index in the original spectrum, and uses a spline between that index and the next, to get subpixel accuracy. The `find_index...` functions were also removed, since they were no longer necessary.

hyperspy/models/eelsmodel.py
magnunor: Done!

hyperspy/misc/eels/tools.py
magnunor: This is greatly improved now, so no interpolation is done. A spline is used to get subpixel accuracy between two values.

hyperspy/misc/eels/tools.py
```diff
@@ -307,3 +308,61 @@ def eels_constant(s, zlp, t):
...
+    >>> onset = get_edge_onset(data, 600, 700, x_axis, 0.1)
+
+    """
+    if not ((percent_position >= 0.) and (percent_position <= 1.)):
+        raise ValueError(
+                "percent position ({0}) must be between 0 and 1".format(
+                    percent_position))
+    start_i = np.argmax(x_axis > start)
+    end_i = np.argmax(x_axis > end)
+    if end_i == 0:
```
francisco-dlp: Woulndn't it be better to raise a ``ValueError`` here?

hyperspy/misc/eels/tools.py
```diff
@@ -307,3 +308,61 @@ def eels_constant(s, zlp, t):
...
+        end_i = len(x_axis)
+
+    data = data[start_i:end_i]
+    x_axis = x_axis[start_i:end_i]
+
+    data_max = data.max()
+    data_min = data.min()
+    data_max_i = data.argmax()
+    data_min_i = data.argmin()
```
francisco-dlp: I would add here a sanity check: ``if data_max_i < data_min_i``. This may happen when analysing multi-dimensional datasets if the edge is not present at a given location. Actually, this sanity check could be used to assign a ``nan`` for the onset at those pixels.

hyperspy/misc/eels/tools.py
```diff
@@ -307,3 +308,61 @@ def eels_constant(s, zlp, t):
...
+
+    """
+    if not ((percent_position >= 0.) and (percent_position <= 1.)):
+        raise ValueError(
+                "percent position ({0}) must be between 0 and 1".format(
+                    percent_position))
+    start_i = np.argmax(x_axis > start)
+    end_i = np.argmax(x_axis > end)
+    if end_i == 0:
+        end_i = len(x_axis)
```
francisco-dlp, line 361: Wouldn't it be better to raise a ``ValueError`` here?

hyperspy/misc/eels/tools.py
```diff
@@ -307,3 +308,61 @@ def eels_constant(s, zlp, t):
...
+    data_onset_value = (data_max - data_min) * percent_position + data_min
+
+    data_crop = data[data_min_i:data_max_i][::-1]
+    data_onset_i = np.argmax(data_crop < data_onset_value)
+    data_onset_i = len(data_crop) - data_onset_i
+
+    onset_slice = slice(data_min_i + data_onset_i - 1,
+                        data_min_i + data_onset_i + 1)
+
+    spline = InterpolatedUnivariateSpline(
```
francisco-dlp: I still think that spline interpolation is an overkill feature here that breaks the "vectority" of the algorithm. Actually, here a spline is used just for linear interpolation. Why not writing the linear interpolator vectorially instead?

hyperspy/models/eelsmodel.py
```diff
@@ -976,3 +978,83 @@ def resume_auto_fine_structure_width(self, update=True):
...
+            percent_position=0.1):
+        x_axis = self.axes_manager.signal_axes[0].axis
+
+        if only_current:
+            edge_onset_energy = get_edge_onset(
+                self.signal(), signal_range[0], signal_range[1],
+                x_axis, percent_position=percent_position)
+            component.onset_energy.value = edge_onset_energy
+            component.onset_energy.store_current_value_in_array()
+        else:
```
francisco-dlp, line 1054: If the algorithm is written in vectorial form, this iteration becomes unnecessary.

hyperspy/models/eelsmodel.py
```diff
@@ -976,3 +978,83 @@ def resume_auto_fine_structure_width(self, update=True):
                 self.resolve_fine_structure()
         else:
             warnings.warn("Not suspended, nothing to resume.")
+
+    def estimate_and_set_coreloss_edge_onset(
```
francisco-dlp, line 982: Wouldn't it be good to add a ``mask`` kwarg to mask positions where the edge is not present?

francisco-dlp (changes requested):
Other than the comments above, the User Guide / docstring suggestions from my previous review has not been addressed. A clear discussion of the limits of the method could prevent misuse by people who may otherwise blindly believe the values returned by the algorithm.

magnunor:
For some reason your previous comments about user guide/docstrings changes doesn't show up for me.

Was this the comment about this?

>  A clear discussion of the limits of the method could prevent misuse by people who may otherwise blindly believe the values returned by the algorithm.

magnunor:
I replaced the spline with a simple linear calculation.

However, the `get_edge_onset` function still only works on single spectra, and I don't have time to improve that aspect of it before October.

hyperspy/misc/eels/tools.py
magnunor: Fixed by adding several `if`s at the beginning of the function, to check if `start`/`end`/+++ is out of bounds of x-axis.

hyperspy/misc/eels/tools.py
magnunor: Fixed by adding several ifs at the beginning of the function, to check if start/end/+++ is out of bounds of x-axis.

hyperspy/misc/eels/tools.py
magnunor: Good idea! However, I'm not sure how to add a `np.nan` to a `EELSCLEdge` component.
```python
Mn_L3.onset_energy.value = np.nan
```
Gives the error
```python
hyperspy/hyperspy/misc/eels/base_gos.py in get_qaxis_and_gos(self, ienergy, qmin, qmax)
     57         else:
     58             index = self.qaxis.searchsorted(qmax)
---> 59             g1, g2 = qgosi[index - 1:index + 1]
     60             q1, q2 = self.qaxis[index - 1: index + 1]
     61             gosqmax = get_linear_interpolation((q1, g1), (q2, g2), qmax)

ValueError: not enough values to unpack (expected 2, got 1)
```

magnunor:
With the EELS functionality having been moved to eXSpy, I'm guessing this pull request should be moved there as well?

If yes, one question would be where the GUI functionality should be placed? Should they be in `hyperspy_gui_ipywidget` and `hyperspy_gui_traitsui`?

ericpre:
They can go in either: `eXSpy` or the `hyperspy_gui_*` - there are already quite a few in `hyperspy_gui_*`: https://github.com/hyperspy/hyperspy_gui_ipywidgets/blob/main/hyperspy_gui_ipywidgets/hyperspy_extension.yaml
Since the existing ones are already in `hyperspy_gui_*` it would make sense to keep following this pattern?

PR #2025, opened 2018-07-31
Fix numpy indexing deprecation warning. Add lazy test for tiff io.

ericpre:
### Progress of the PR
- [x] Fixes numpy deprecation warning which makes io tiff tests fail,
- [x] add lazy test for tiff io,
- [x] ready for review.


thomasaarholt:
That's great. Thanks for fixing!

Code looks fine.

PR #2026, opened 2018-08-03
Fix for #1937: title not properly imported for some dm files.

### LINKED ISSUE #1937: Stacking set of dm4-files: Metadata-Title is no String
Hello again,
i hope that you guys are able to help again.
I have a problem getting around the problem shown below.

`im = hs.load('*.dm4', stack=True)`

`TypeError: must be str, not list`

Am i able to change the metadata in the files before loading them into hyperspy as a stack or perhaps specifying some sort of signal_type-list during loading? I guess its not possible and i have to change the files metadata before loading them as a stack, but i thought asking would not hurt.
Here a single image is loaded and the base info is given to:
`im` results in
`<Signal2D, title: [], dimensions: (|1920, 1792)>`
As you can see, the title of the files seems not to be a string but a list actually as the error states.
Why that is, i dont know, because i am not the experimentalist working on our TEM but just getting the data.
though these are just images, i will have to operate on images with a subset of EELS spectra.
Perhaps i missed some part in the documentary again, though i RTFM.

Error message (using anaconda): 
**TypeError Traceback (most recent call last)
<ipython-input-38-745a65e4cb4b> in <module>()
      1 im = hs.load("*.dm4", stack=True)
      2 im

/Apps/anaconda3/envs/HS/lib/python3.6/site-packages/hyperspy/io.py in load(filenames, signal_type, stack, stack_axis, new_axis_name, lazy, **kwds)
    231                 signal = stack_method(
    232                     signal, axis=stack_axis, new_axis_name=new_axis_name,
    233                     lazy=lazy)
    234                 signal.metadata.General.title = os.path.split(
    235                     os.path.split(os.path.abspath(filenames[0]))[0])[1]

/Apps/anaconda3/envs/HS/lib/python3.6/site-packages/hyperspy/misc/utils.py in stack(signal_list, axis, new_axis_name, lazy, **kwargs)
    864             # Get the title from 1st object
    865             signal.metadata.General.title = (
    866                 "Stack of " + first.metadata.General.title)
    867             signal.original_metadata = DictionaryTreeBrowser({})
    868         else:

TypeError: must be str, not list**

Thanks for taking the time to read this paragraph!

Sincerely,
Maximilian

PS: I heard that there is a possibility to save the files correctly with "tag-specification" in digital micrograph but do you guys know another workaround?

ericpre:
This PR fixes #1937.

- [x] make the title is a string for dm files,
- [x] ready for review.

### DIFF
--- a/hyperspy/io_plugins/digital_micrograph.py
+++ b/hyperspy/io_plugins/digital_micrograph.py
@@ -548,15 +548,8 @@ class ImageObject(object):
 
     @property
     def title(self):
-        title = self.imdict.get_item("Name")
-        if isinstance(title, list) and len(title) >0:
-            # take the first item of the list hoping for the best...
-            title = title[0]
-        if not isinstance(title, str):
-            # We only want a string, if this is not the case, set to default
-            _logger.info("Title can't be imported.")
-            title = ''
-        return title
+        title = self.imdict.get_item("Name", "")
+        return title if title else ""
 
     @property
     def record_by(self):
--- a/hyperspy/io_plugins/digital_micrograph.py
+++ b/hyperspy/io_plugins/digital_micrograph.py
@@ -549,6 +549,9 @@ class ImageObject(object):
     @property
     def title(self):
         title = self.imdict.get_item("Name", "")
+        # ``if title else ""`` below is there to account for when Name
+        # contains an empty list.
+        # See https://github.com/hyperspy/hyperspy/issues/1937
         return title if title else ""
 
     @property
--- a/hyperspy/io_plugins/digital_micrograph.py
+++ b/hyperspy/io_plugins/digital_micrograph.py
@@ -548,10 +548,15 @@ class ImageObject(object):
 
     @property
     def title(self):
-        if "Name" in self.imdict:
-            return self.imdict.Name
-        else:
-            return ''
+        title = self.imdict.get_item("Name")
+        if isinstance(title, list) and len(title) >0:
+            # take the first item of the list hoping for the best...
+            title = title[0]
+        if not isinstance(title, str):
+            # We only want a string, if this is not the case, set to default
+            _logger.info("Title can't be imported.")
+            title = ''
+        return title
 
     @property
     def record_by(self):


francisco-dlp:
Maybe @Bejoscha could explain why sometimes the title is a list to help us decide if this is a good fix for the issue?

Bejoscha:
@francisco-dlp I'll gladly will try to help out, but am somehow missing the context of this question. (just got it by email) Is there some previous conversionation I can read to catch up on, or could you give me a short summary of what the issue is? I don't quite understand "Sometimes the title is a list" at all right now.

ericpre:
@Bejoscha, please have a look at the issue #1937 for the context. There is one dataset acquired with a K2 on GMS2 and when HyperSpy parse the title, it seems that the corresponding tag is a list!

Bejoscha:
@ericpre Can't download that file from the #1937 anymore. Can anyone provide a fresher link?
But it sounds omnious to me. AFAIK there is nothing fancy about filenames with DM3/DM4 files at all. What is the file called? Is the parser itself running into some "," maybe? The only "tricky" part with DM TagStructures (but not filenames!) are colons (:) and square brackets [ ], because colons signify a "branch" in the tag, while the [ ] are used to indicate indexed items in TagLists. (TagLists are a subclass of TagGroups, but that shouldn't concern you.) Of course, I don't know what exactly the parser doese neither ;c)

Bejoscha:
What I would recommend anyhow is, that you inspect the strange file in DigitalMicrograph itself (any version), reading it as TagGroup rather than an image. I'm attaching a little **DM-script** to do that:

_string path = GetApplicationDirectory( "open_save", 0 )
if (!OpenDialog(NULL,"Select file", path, path ) ) exit(0)
taggroup tg = NewTagGroup()
if ( !tg.taggrouploadfromfile(path) ) Throw( "Could not load:\n"+path )
tg.TagGroupOpenBrowserWindow( path, 1 )_

ericpre:
[Here](https://www.dropbox.com/s/6a5t2vdniceuntl/cg_nitisdbeameffect1_hour_00_minute_00_second_02_frame_0000.dm4?dl=0) is a link to download the file.

Bejoscha:
Thanks. I could open the file just fine in DM (GMS 3) and with the script. (see below). What seems to be the case is, that the "name" field in the tag-structure itself is actually empty.
How was this file saved? With a regular DM or was it "exported" by some other software?
vis

Bejoscha:
Just resaving the image from DM and then opening it again might be what fixes all issues??
Anyway, be aware in any DM file browser that the file contains multiple images! There is the image data itself and there usually is a thumbnail image (the one you see as icon on Windows systems).
f.e. after resaving the file (AA.dm4)
 the file as tagstructure looks like this:
vis2

The first image in the list of images is the thumbnail.

Bejoscha:
As for the fix - I don't know the Python code at all - but in case of "read title failed" wouldn't it make more sense to use the actual FILENAME instead of the "" then?
Note, that this would also fix the issue when somebody manually edited the file's name. Currenlty, this would still use the internally held (old=incorrect) name. The reason the name is in the tags at all is just because DM files may contain an arbitrary number of images, so the file-name is not the "image name". (It only is for one-image files.)

ericpre:
Thanks for looking at this!

> How was this file saved? With a regular DM or was it "exported" by some other software?

@mgrove-wwu should be able to answer this question. In #1937, he said that the image was saved with DM. You seems to implies that this is not normal DM behaviour (at least not a "feature" defined in a conventional way ;)) and the image may have been exported using a script? Even if this was the case, it should be using DM functions...

Bejoscha:
Na, I'm not implying anything for saving. If it was saved from a DM script, then you're right it should work identical as saving with the UI.
However, similar to Hyperspy, other people have created their own reader/writers in the past which partly work, or only work for specific versions. 

And, of course, there is always the possibility that a certain GMS version has had a bug, but that that bug would not affect DM itself.

The binary structure of the DM files (The Taggroup) is pretty straight forward and hence easy to retro-engeneer. However, the structure names/meaning of the tags themselves are purely for "internal" use by DM and may hence vary from version to version. It's just Gatan's job to make sure that DM itself can deal with the files ;c)

This is pretty much the reason why the DM file format is not "open" - noone wants to deal with inconsistency problems outside DM, if things change from version to version.

However, people who do write their own readers/writes may always use the little "trick" of loading the file as TagFile to see what is going on and adapt, when necessary. 

As for the issue with thumbnails: I don't know when code changes with that respect, but I'm pretty sure somewhere in the GMS 2.x series as this was when windows vista and win7 came into play. I can easily imagine changes being made with this respect between different GMS 2 versions.
It is only the latest _GMS 2.3_ version, which (theoretically) should guarantee the most bug-free behavior.

hyperspy/io_plugins/digital_micrograph.py
```diff
@@ -555,7 +555,7 @@ def title(self):
         if not isinstance(title, str):
             # We only want a string, if this is not the case, set to default
             _logger.info("Title can't be imported.")
-            title = ''
+            title = os.path.splitext(os.path.split(self.filename)[1])[0]
```
francisco-dlp: I don't think that we need to do this. HyperSpy stores the filename in ``s.tmp_metadata.filename`` and uses it for the signal title if ``metadata.General.title`` is not defined.

ericpre:
I have updated the default title to be consistent with DM convention of setting the filename as title. From @Bejoscha comments, it seems that there is no obvious reason why with this file has a "TagLists" as title and this is not expected for DM to do this, so falling back to the filename should be fine.

hyperspy/io_plugins/digital_micrograph.py
ericpre: Indeed, I thought this was the case, but it doesn't work. I would expect this to be happen in [`load_with_reader`](https://github.com/hyperspy/hyperspy/blob/RELEASE_next_patch/hyperspy/io.py#L298-L308), but this is actually not!

francisco-dlp: I've just tested it with a dm3 file and it works for me.

francisco-dlp:
```python
>>> s = hs.load("/home/abc/EELS_1.0nA_CoreLoss_binned2.dm3")
>>> s.tmp_parameters
├── extension = dm3
├── filename = EELS_1.0nA_CoreLoss_binned2
└── folder = /home/abc
```

ericpre: yes, but does it set to `s.metadata.General.title`?

francisco-dlp: No, but I think that that's fine because the file name and the title are two different things.

ericpre: Yes, shall stick to hyperspy default (empty string) as for new data, then?
```python
>>> s = hs.signals.Signal1D([0,1])
>>> s.metadata
├── General
│   └── title = 
└── Signal
    ├── binned = False
    └── signal_type = 
```

francisco-dlp: Sounds good.

ericpre: Done, reset to previous commit.

hyperspy/io_plugins/digital_micrograph.py
```diff
@@ -548,10 +548,15 @@ def names(self):
 
     @property
     def title(self):
-        if "Name" in self.imdict:
-            return self.imdict.Name
-        else:
-            return ''
+        title = self.imdict.get_item("Name")
```
francisco-dlp: What about the following instead:

```python
title = self.imdict.get_item("Name", default="")
return title if title else ""
```

This takes into account all known possibilities (no Name attr, Name is a str and Name is empty list). If in some file Name contains something else it'll lead to a new bug that we can investigate and fix appropriately.

ericpre: Done.

PR #2029, opened 2018-08-04
Flake8 configuration and flake8 fixes

magnunor:
Currently HyperSpy is not following the python style guide (PEP8), which aims to keep the code consistent. One way of checking the code is by using `flake8`, which also checks for other things such as undefined variables and redefined variables. (I actually found a bug in the `Arctan` component while doing this). 

Ideally we want to have the code pass `flake8` without any errors, so the continuous integration system can check if new contributions add any new errors. I've been very happy with this setup in `atomap` and `pixstem`, since it really improves the code quality and several times it has helped me catch bugs before including them into the main branch.

### Changes

- Add a configuration file for flake8: `.flake8`
- Try to fix all the issues flagged by `flake8`

The last one is a pretty big task, and potentially a bit risky since it is possible to subtly break functionality.

Some files will (currently) have to be excluded through the `flake8` configuration file:

- If they simply import functions or classes
- If they contain very long lines, for example `misc/elements.py`

In addition, I've been fixing minor docstring typos or styie issue when I see them.

### Progress of the PR
- [ ] Finish config file
- [ ] Fix all flake8 issues
- [ ] Check that nothing broke
- [ ] Ready for review.

### To test flake8

Run from the HyperSpy root directory

```bash
python3 -m flake8
```

### DIFF
--- a/.flake8
+++ b/.flake8
@@ -14,8 +14,30 @@ exclude =
     hyperspy/utils/plot.py
     hyperspy/utils/roi.py
     hyperspy/utils/eds.py
+    hyperspy/datasets/example_signals.py
+    hyperspy/misc/machine_learning/import_sklearn.py
+    hyperspy/samfire_utils/fit_tests.py
     # Ignore files which contains constants or physical values
     hyperspy/misc/eds/ffast_mac.py
     hyperspy/misc/elements.py
-    hyperspy/datasets/example_signals.py
-    hyperspy/misc/machine_learning/import_sklearn.py
+    # Files with unresolved issues
+    hyperspy/misc/eels/tools.py
+    hyperspy/misc/eels/eelsdb.py
+    hyperspy/misc/io/tools.py
+    hyperspy/misc/example_signals_loading.py
+    hyperspy/misc/utils.py
+    hyperspy/misc/material.py
+    hyperspy/_components/lorentzian.py
+    hyperspy/_components/gaussianhf.py
+    hyperspy/_components/expression.py
+    hyperspy/_components/gaussian.py
+    hyperspy/io_plugins/protochips.py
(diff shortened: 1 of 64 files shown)


magnunor:
I regret starting this...

Anyone knows how to include automatic flake8 testing in the continuous integration system?

sk1p:
In LiberTEM we are using tox to run tests, have a look here:

https://github.com/LiberTEM/LiberTEM/blob/master/.travis.yml
https://github.com/LiberTEM/LiberTEM/blob/master/tox.ini

We set TOXENV=flake8 variable in the testing matrix entry to select flake8, but maybe it can be replaced with something like `script: pip install flake8 && flake8 ./hyperspy/`, so that it works without tox?

Maybe it would be a better idea to have a plot of pep8 failures instead of hard-failing the build though, and to merge pep8 fixes bit-by-bit. If you do it in one large PR, I can imagine you will get quite a lot of conflicts as the main branch changes underneath...

magnunor:
You're probably right, having it all in one pull request makes it i) a real mess to sort out merge conflicts, ii) really hard to review everything properly.

I guess the easiest would be as you suggest: add the `.flake8` config file, possible with all the offending files being ignored. And then slowly sort out one file at a time. At least the files which are currently passing (any?) won't get worse...

ericpre:
@magnunor: if you think that this is a big task, maybe it would be more manageable to do this after the split, rather than now? There are already _quite a few_ PRs to finish before the split and it would be useful to get it done!

magnunor:
@ericpre, yeah. That sounds like a better plan. 

The reason for starting this was me having a couple of hours of no WIFI on an airplane :) However as I've now seen doing all this in one chunk would probably take several days.

PR #2032, opened 2018-08-14
Speed up fast background removal

ericpre:
### Description of the change
Add a method to a few "background type" model component to compute the value of the component significantly quicker than it is currently done. This is particularly useful in combination of `estimate_parameters` of this components, for example for background removal of EELS data set.

### Progress of the PR
- [x] Add the `function_nd` method to the `Expression` and (a few others) components to compute the component value without iteration over navigation axis,
- [x] add tests,
- [x] fix a bug in the `estimate_parameters` of the `GaussianHF` component,
- [x] simplify tests and increase their coverage for a few components,
- [x] move component tests from the tests/model folder to the tests/component folder,
- [x] ready for review.


hyperspy/tests/component/test_components.py
```diff
@@ -1,12 +1,20 @@
...
-
 from numpy.testing import assert_allclose
+import pytest
 
 import hyperspy.api as hs
 from hyperspy.models.model1d import Model1D
 from hyperspy.misc.test_utils import ignore_warning
 
 
+def _generate_parameters():
```
francisco-dlp: To make the code more readable you could turn the ``_generate_parameters`` function into a global variable instead e.g:

```python
import itertools

TRUE_FALSE_2_TUPLE = [p for p in itertools.product((True, False), repeat=2)]
```

hyperspy/tests/component/test_gaussianhf.py
```diff
@@ -19,9 +19,12 @@
 
 import numpy as np
 from numpy.testing import assert_allclose
+import pytest
 
 from hyperspy.components1d import GaussianHF
 from hyperspy.signals import Signal1D
+from hyperspy.tests.component.test_components import _generate_parameters
```
francisco-dlp: I think that it'll be better for readability to redefine the suggested ``TRUE_FALSE_2_TUPLE `` variable Instead of importing the ``_generate_parameters`` function from a different module.

hyperspy/_components/power_law.py
```diff
@@ -152,3 +155,21 @@ def estimate_parameters(self, signal, x1, x2, only_current=False,
...
+        """%s
+
+        """
+        x = axis[np.newaxis, :]
+        A = self.A.map['values'][..., np.newaxis]
+        r = self.r.map['values'][..., np.newaxis]
+        o = self.origin.map['values'][..., np.newaxis]
+        # Since no parameter can be used to normalise for binned data, we do 
+        # it here.
+        if self.binned:
```
francisco-dlp: Why is multiplying by a factor needed here? ``integrate1D`` in ``estimate_parameters`` should take care of this.

hyperspy/_components/gaussian.py
```diff
@@ -209,3 +212,15 @@ def fwhm(self):
     @fwhm.setter
     def fwhm(self, value):
         self.sigma.value = value / sigma2fwhm
+
+    def array(self, axis):
```
francisco-dlp: Wouldn't it be better to make this function private since it is not available for all components? Also, wouldn't ``nd_function`` or ``function_nd`` be a more explicit name?

francisco-dlp (changes requested):

francisco-dlp:
On a more general note, what do you think about generalizing and simplifying the approach as follows:

* Implement a ``nd_function`` method for ``Expression``
* Rewrite the functions the functions that have an ``estimate_parameters`` method (the ones available in ``remove_background``, hence the only ones modified in his PR) using ``Expression``.

The advantage would be a more compact and readable (hence easier to maintain code). Also, ``Expression`` can provide extra speed improvement by passing ``modules="numexpr"``.

hyperspy/_signals/signal1d.py
```diff
@@ -1033,20 +1033,27 @@ def _remove_background_cli(
             signal_range[0],
             signal_range[1],
             only_current=False)
-        if not fast:
+        if fast:
```
francisco-dlp: The multi-dimensional approach is actually specially well suited for lazy computation. This would bring the extra advantages of parallelization and avoid the memory issue. So, I wonder if it wouldn't be possible to implement this using dask as follows (pseudocode):

```python
if lazy:
    # Get map/x chunks from the signal navigation/signal chunks
    # Create lazy parameters and ``x`` arrays (hence  ``nd_function`` output) using these chunks.
else:
   # Create lazy parameters and ``x`` arrays (hence  ``nd_function`` output) using dask's new auto chunking.
   # Call ``compute`` at the end.
```
The lazy computation could be controlled via a kwarg with default to True e.g. ``remove_background(parallel=True)``.

francisco-dlp:
``EELSModel`` could also benefit from the speed-up, see https://github.com/hyperspy/hyperspy/blob/RELEASE_next_minor/hyperspy/models/eelsmodel.py#L483

ericpre:
@francisco-dlp: thanks for the very good idea, I did most of the things you suggested and it is significantly better: more clear and a faster using numexpr! 
I was not happy with the name of the method `array` but I could not think of any better name... so I changed it to `function_nd` as you suggested, which is much better!

Regarding EELSModel, I don't see where they could a improvement in speed, because this doesn't improve the speed of the `estimate_parameters` method.

hyperspy/_components/offset.py
```diff
@@ -92,9 +95,19 @@ def estimate_parameters(self, signal, x1, x2, only_current=False):
...
-            self.offset.map['values'][:] = dc[gi].mean(axis.index_in_array)
-            if binned is True:
+            self.offset.map['values'][:] = dc[tuple(gi)].mean(axis.index_in_array)
+            if self.binned:
                 self.offset.map['values'] /= axis.scale
             self.offset.map['is_set'][:] = True
             self.fetch_stored_values()
             return True
+
+    def function_nd(self, axis):
```
ericpre, line 105: I haven't rewrote this component to an `Expression`, because it looks to me that the expression string needs to have a free parameter, which is not the case here! Anyway, it seems unlikely (see documentation available online) that there will be a speed improvement using numexpr in this specific case.

hyperspy/_signals/signal1d.py
ericpre: Actually, this should in the `as_signal` method of the model and check if the `function_nd` is available for all components and use this approach if necessary.

francisco-dlp: I fully agree.

hyperspy/_components/polynomial.py
```diff
@@ -119,8 +122,18 @@ def estimate_parameters(self, signal, x1, x2, only_current=False):
...
                 cmap_shape = nav_shape + (self.get_polynomial_order() + 1, )
                 self.coefficients.map['values'][:] = cmaps.reshape(cmap_shape)
-                if binned is True:
+                if self.binned:
                     self.coefficients.map["values"] /= axis.scale
                 self.coefficients.map['is_set'][:] = True
             self.fetch_stored_values()
             return True
+
+    def function_nd(self, axis):
```
ericpre, line 131: Because the `coefficients` parameter is defined as an array of several parameters, I am not sure if rewriting the polynomial as an Expression will be a good way to improve it. There may be better approach.

hyperspy/_signals/signal1d.py
francisco-dlp: Currently ``as_signal`` operates in parallel but pixel by pixel, what should be much slower than performing the same operation using dask vectorially. This is very interesting because it should be possible to write all but the ``EELSCLEdge`` component using ``Expression``.

hyperspy/_components/polynomial.py
francisco-dlp: Yes, that's the other problematic component that I've forgotten. However, see #1989

hyperspy/_signals/signal1d.py
ericpre: Yes, we needs to do this.
I had a look at this and I could not get it to work in the time given. I will not have time to do this properly now, so I will open an issue to keep it for later. In the mean time, since this is an improvement of the current implementation, I would suggest to keep this approach for now.

hyperspy/_components/polynomial.py
ericpre: Happy to see polynomial expression being done in #1989!! ;)

hyperspy/_components/power_law.py
```diff
@@ -109,12 +109,14 @@ def estimate_parameters(self, signal, x1, x2, only_current=False,
...
             s = signal
-        I1 = s.isig[i1:i3].integrate1D(2j).data.astype("float")
-        I2 = s.isig[i3:i2].integrate1D(2j).data.astype("float")
         if s._lazy:
             import dask.array as da
             log = da.log
+            I1 = s.isig[i1:i3].integrate1D(2j).data.astype("float")
+            I2 = s.isig[i3:i2].integrate1D(2j).data.astype("float")
         else:
+            I1 = s.data[..., i1:i3].sum(-1)
```
ericpre: This is an speed improvement of 100x. Using `integrate1D` give slightly different results (~2e-3) for binned signal, so for an estimation this is fine!

francisco-dlp: I don't get it: it should give exactly the same result for binned signals and be a worst approximation for unbinned signals.

francisco-dlp: After looking at the tests, that's probably what you meant.

hyperspy/_components/power_law.py
```diff
@@ -109,12 +109,14 @@ def estimate_parameters(self, signal, x1, x2, only_current=False,
             s = signal.get_current_signal()
         else:
             s = signal
-        I1 = s.isig[i1:i3].integrate1D(2j).data.astype("float")
-        I2 = s.isig[i3:i2].integrate1D(2j).data.astype("float")
         if s._lazy:
             import dask.array as da
             log = da.log
+            I1 = s.isig[i1:i3].integrate1D(2j).data.astype("float")
```
francisco-dlp: I would rather keep this code the same for ``lazy`` and standard signals since they should both give the same result.

Although proper integration vs simply summing doesn't make much difference in your example, it can be very different in other cases for unbinned signals e.g. for non-linear axes. I think that it'll be better to leave it as it was.

francisco-dlp (commented):
Other than the above, LGTM!

hyperspy/_signals/signal1d.py
```diff
@@ -1045,20 +1059,24 @@ def remove_background(
             background_type='Power Law',
             polynomial_order=2,
             fast=True,
+            module=None,
             show_progressbar=None, display=True, toolkit=None):
-        signal_range = signal_range_from_roi(signal_range)
+
+        if module is None:
```
francisco-dlp: What about moving the auto numexpr code to the individual components written with ``Expression``. Ideally we could move it to ``Expression`` itself but ``numexpr`` does not support as many functions as numpy.

In this way:
* The speed improvement would be available everywhere where the components are used.
* We can reduce the number of kwargs of ``remove_background``.

francisco-dlp: By the way, we could add numexpr to the requirements: it just depends on numpy and is pypi and conda installable.

hyperspy/_components/power_law.py
ericpre: Shall we just do the simple sum for binned signal?

francisco-dlp: That's what integrate1D does, so I think that the code is simpler just using it.

ericpre: Yes, but it does it ~50-100 times slower! I have changed to use the `out` parameter but it is still not as fast as the faster alternative (possibly a factor 10 slower). I may look at further improvement but for now, I guess this is good enough.

francisco-dlp: That's interesting and it reminds me of the important overhead involved in deepcopying a signal because of the ``metadata``. Something to take seriously elsewhere.

francisco-dlp:
@ericpre, is this ready for review?

ericpre:
Yes.

francisco-dlp:
LGTM, will you finally keep ``integrate1D`` with ``out``?

francisco-dlp:
Oh, and it'll be good to add (as suggeted above) a little discussion about ``fft_shift`` and markers to the User Guide.

ericpre:
> Oh, and it'll be good to add (as suggeted above) a little discussion about fft_shift and markers to the User Guide.

I guess this was for #1727?

> LGTM, will you finally keep integrate1D with out?

Yes, I think so, even if it should be possible to do speed it up further, this is already an improvement.

PR #2037, opened 2018-08-23
Fix wrong baseline images for model plot testing

ericpre:
It seems that the model doesn't plot properly (experimental data plotted with dots are duplicated) with matplotlib <=2.2.2 and matplotlib > ?. This is not fixed in matplotlib 2.2.3 and this PR update the baseline images for model plot testing accordingly.
The difference is fairly subtle, you need to zoom in to see it and this can explain why we haven't noticed it before until the plotting tests start to fail following matplotlib 2.2.3 release. Below is an example of the `test_fit_EELS_convolved` test.

figure_convolved_-_true_signal


francisco-dlp:
Shouldn't we require ``matplotlib >= 2.2.3`` then?

ericpre:
Well it is not really a big deal, so I am not sure if it is worth updating matplotlib requirement to last version.

francisco-dlp:
If we don't, the tests will not pass with an older version. The only drawback of depending on new version is that it may not yet be available in all linux distributions. However, as HyperSpy is not available in any Linux distribution yet AFAIK, that's not an issue.

ericpre:
True, for testing it needs to be updated for sure.
For the user, I am reluctant to force to update to the last version unless this is really necessary. Updating matplotlib to last version may also update dependencies and in some case it may be a source of issue if there are some dependencies conflict.

francisco-dlp:
In general I agree with your cautiousness but in this case, in practice HyperSpy already requires mpl > 2.1.1. That means that we'll only be excluding 1 extra minor mpl release (2.1.2). That's statistically unlikely to cause any issues.

ericpre:
Ok, done.

PR #2040, opened 2018-08-24
Add indices control to documentation

jat255:
This PR adds a few statements in the User Guide about how one can directly modify the `inidices` of an `AxesManager` to programmatically set the "current position". While in the relevant files, I also fixed a number of intersphinx links that weren't connecting correctly.


jat255:
Here are the new bits:
https://github.com/hyperspy/hyperspy/pull/2040/files#diff-0ed6148bc415112a91b14b804cb6b2fdR400
https://github.com/hyperspy/hyperspy/pull/2040/files#diff-4717924c9d836c8feccb1246064a38fbR914

PR #2043, opened 2018-08-29
Addition of mib importer

chrisallen80:
Importer for .mib files from Merlin/Medipix direct electron detectors. This is a hack of the raw/rpl importer. mib files are always saved in mib/hdr pairs. When loading into hyperspy it is the hdr file which must be loaded.

- Added mib to io_plugins __init__.py
- Added mib.py 

- [x] Read meta-data from .hdr file
- [x] Import various bit depth data (1,6, 12,24)
- [x] split inter-frame headers from data
- [x] lazy loading of data
- [ ] get frame time from inter-frame headers
- [ ] update docstring 
- [ ] update user guide 
- [ ] add tests
- [x] ready for review.

```python
>>> import hyperspy.api as hs
>>> import numpy as np
>>> s = hs.load("filename.hdr")

### DIFF
--- a/hyperspy/io_plugins/mib.py
+++ b/hyperspy/io_plugins/mib.py
@@ -1,0 +1,739 @@
+# -*- coding: utf-8 -*-
+# Copyright 2007-2016 The HyperSpy developers
+#
+# This file is part of  HyperSpy.
+#
+#  HyperSpy is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+#  HyperSpy is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with  HyperSpy.  If not, see <http://www.gnu.org/licenses/>.
+
+#  for more information on the hdr/RAW format, see
+#  http://www.nist.gov/lispix/
+#  and
+#  http://www.nist.gov/lispix/doc/image-file-formats /raw-file-format.htm
+
+import codecs
+import os.path
+from io import StringIO
+import logging
+
+import numpy as np
+
+from hyperspy.misc.io.utils_readfile import *
+from hyperspy import Release
+from hyperspy.misc.utils import DictionaryTreeBrowser
+
+_logger = logging.getLogger(__name__)
+
+
+# Plugin characteristics
+# ----------------------
+format_name = 'hdr'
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Comment by magnunor** [comment]
Nice!

This is the same code which pyXem uses?

In the example you use the header file to load the data, does the reader itself require the header file? I think all the data you really need is in the binary files.

**Comment by chrisallen80** [comment]
@magnunor it is similar to the pyXem code, with some updates to deal will all bit depths.

You are correct that all the relevant information is in the inter-frame headers in the binary files, but prior to attempting to open the binary file you don't know the bit depth. As the hdr file is always there and states the bit-depth of the mib file I felt this was simpler. I'm happy to try to get it working without using the hdr files if this is preferable.

**Comment by magnunor** [comment]
I do it without the header file in my simple reader in `pixStem`, feel free to use that code: https://gitlab.com/pixstem/pixstem/blob/master/pixstem/io_tools.py#L14

See https://gitlab.com/pixstem/pixstem/blob/master/pixstem/io_tools.py#L76 for how to grab just the header part in the binary file.

It might also be nice to be able to specify what the probe dimensions should be, and removal of things like flyback pixels. This can be done as optional parameters. Possible implementation of that: https://gitlab.com/pixstem/pixstem/blob/master/pixstem/io_tools.py#L85

No worries if you don't have time to do all these things in this pull request. The basic functionality can be added first, and subsequent pull requests can expand upon the functionality later.

File: hyperspy/io_plugins/mib.py [hunk 1]
```diff
@@ -0,0 +1,739 @@
...
+
+import codecs
+import os.path
+from io import StringIO
+import logging
+
+import numpy as np
+
+from hyperspy.misc.io.utils_readfile import *
+from hyperspy import Release
```
sem-geologist, line 32: these two lines.
Please review the dependency, it should not depend from hyperspy (exception is `hyperspy.misc.io.utils_readfile` as it will go together with io_plugins to splited library).
It would be great idea not to use `DictionaryTreeBrowser` in io library for metadata tree building. It is better to expose all metadata as `original_metadata` and use mapping dictionary to build proper hyperspy tree.

hyperspy/io_plugins/mib.py
```diff
@@ -0,0 +1,739 @@
...
+    else:
+        idepth, iheight, iwidth = 0, 1, 2
+        names = ['depth', 'height', 'width']
+
+    scales = [1, 1, 1]
+    origins = [0, 0, 0]
+    units = ['', '', '']
+    sizes = [hdr_info[names[i]] for i in range(3)]
+
+    if 'date' not in hdr_info:
```
sem-geologist, line 443: all these `if`s could be simplified with using mapping

sem-geologist (commented):
It is not review, just want to point a few caveats

dnjohnstone:
I think this is now out of scope until after the split of hyperspy - can we close it and re-visit after a split?

chrisallen80:
Agreed, closing.

PR #2039, opened 2018-08-24
Release v1.4

francisco-dlp:

PR #2044, opened 2018-08-29
Fix missing folder in package_data

ericpre:
Following the addition of a the plotting test in #2003, the corresponding baseline image folder is missing in the package_data.


francisco-dlp:
I've missed this PR and I've fixed this in #2039. I'll close it to avoid the burden of merging it.

Issue #2048, opened 2018-09-03
The IO test files are too large


francisco-dlp:
The file size of the ``IO`` test folder is 88.2 MB. As a consequence, the file size of HyperSpy v1.4 source distribution is 67.7 MB and hits the PyPI's file size limit, see https://github.com/pypa/warehouse/issues/4659. This is currently holding the v1.4 release.

The bigger offenders are:

* ``edax_files.zip`` 36.5 MB
* ``bruker_data/`` 12.8 MB
* ``FEI_new/`` 10.3 MB
* ``emd_files/`` 9.3 MB
* ``dm4_1D_data/`` 4.5 MB
* ``tiff_files/`` 3.5 MB

Hopefully the PyPI maintainers will increase the file size limit of our project so that we can go on with the release. However, regardless, we should reduce the file size of the test files when possible. Otherwise this is going to remain an issue for the planned separate IO library #1978.

magnunor:
From a longer term perspective, one solution could be to compress all the files, and use the "decompress" trick in the test themselves. Simply compressing the IO-test folder reduces it from 88 MB to 57 MB.

We could also try using files which only have zero values, which would make the compression much more efficient. For example, in `fpd` 33 MB of Merlin binary files (mib), is reduced to 80 kB, due to the files being only zeros.

ericpre:
A short term workaround is to remove some of the large files and skip the
corresponding tests just for the release... Not great but this would do for
this release.

ericpre:
I meant remove from packaging only, not from the source code.

>

francisco-dlp:
@magnunor, thanks for the suggestion. Unfortunately compressing the files doesn't help with this because the source distribution is itself compressed. I've tried compressing all the files with zip and creating a new sdist and I get exactly the same file size. Using files that only have zero values should help, but we won't be able to cover all our test cases with zero-filled files.

@ericpre, we could remove files and skip tests, but as pypi is the primary archiving repository of Python packages, I would consider that an action of last resort (i.e. something we could reluctantly do if they don't increase the file size limit of our project as long as nobody comes with a better alternative). So far we haven't heard from PyPA.

sem-geologist:
@francisco-dlp , I think I had already discussed previously against double and triple compressions as that actually 1. increase the size and 2. puts overhead. And I am looking to you EDAX_test files.

[...]

PR #2049, opened 2018-09-03
Remove edax_files.zip from distribution

francisco-dlp:
The ``edax_files.zip`` test files are too large and take hyperpy's distribution beyond PyPI's file size limit of 60MB. See #2048. 

This PR removes the file from the distribution. Instead, the test files are downloaded from GitHub. The tests are skipped if the download fails.

### DIFF
--- a/hyperspy/tests/io/test_edax.py
+++ b/hyperspy/tests/io/test_edax.py
@@ -32,13 +32,15 @@ if not TEST_FILES_OK:
     try:
         r = requests.get(
             "https://github.com/hyperspy/hyperspy/blob/e7a323a3bb9b237c24bd9267d2cc4fcb31bb99f3/hyperspy/tests/io/edax_files.zip?raw=true")
-        with open(ZIPF, 'wb') as f:
-            SHA256SUM_GOT = hashlib.sha256(r.content).hexdigest()
-            if SHA256SUM_GOT == SHA256SUM:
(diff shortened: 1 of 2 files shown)


jat255:
This is the solution I would have suggested, so LGTM

hyperspy/tests/io/test_edax.py
```diff
@@ -2,27 +2,48 @@
...
+# As a temporary solution, we attempt to download the test files from GitHub
+# and skip the tests if the download fails.
+if not TEST_FILES_OK:
+    try:
+        r = requests.get(
+            "https://github.com/hyperspy/hyperspy/blob/e7a323a3bb9b237c24bd9267d2cc4fcb31bb99f3/hyperspy/tests/io/edax_files.zip?raw=true")
+        with open(ZIPF, 'wb') as f:
+            f.write(r.content)
+            TEST_FILES_OK = os.path.isfile(ZIPF)
+    except BaseException:
```
jat255: Would it make sense to do a hash check here for the zip file?

francisco-dlp: Good point, I'll fix that in a minute.

francisco-dlp:
Thanks @jat255, it should be fine now.

If nobody is against it, if this passed the tests in conda-forge, I'll merge it and release.

PR #2054, opened 2018-09-09
Fix cyclic import, thanks @dmlundeby

### LINKED ISSUE #2053: Can't import BaseSignal through hyperspy.signal
The following line raises an ImportError in HyperSpy v1.4

`import hyperspy.signal`

This was possible in HyperSpy v1.3. This is possibly what causes HyperSpyUI not to launch with HyperSpy v1.4, see https://github.com/hyperspy/hyperspyUI/issues/170

francisco-dlp:
Fixes #2053


ericpre:
To avoid similar issue in the future, it would be good to add tests to check that all files can be imported, at least for the API? @francisco-dlp, do you want to do this in this PR?

PR #2058, opened 2018-09-18
Bug  in powerlaw when numexpr is not installed

AEljarrat:
The PowerLaw component (found in `_components/power_law.py`) cannot be imported without having the`numexpr` package installed. This package is a new requirement of hyperspy, however, it is stated in the docstring that the default and safer behavior of the function is to use `numpy`. I have changed the default option to `numpy` as corresponds to this rationale.

### Description of the change
- Changed the default option of power law component to "numpy"

### Progress of the PR
- [x] Change implemented.
- [x] ready for review.

### Minimal example of the bug fix:
```python
## If numexpr is not installed, this fails
s = hs.signals.EELSSpectrum(np.random.random(1000))
s.power_law_extrapolation(window_size=20, 
                          extrapolation_size=24)
```

### DIFF
--- a/hyperspy/_components/power_law.py
+++ b/hyperspy/_components/power_law.py
@@ -39,7 +39,7 @@ class PowerLaw(Expression):
 
     """
 
-    def __init__(self, A=10e5, r=3., origin=0., module="numexpr", **kwargs):
+    def __init__(self, A=10e5, r=3., origin=0., module="numpy", **kwargs):
         super(PowerLaw, self).__init__(
             expression="A *( x - origin) ** -r",
             name="PowerLaw",


thomasaarholt:
I think @ericpre set numexpr as default in order to speed up sympy (as well as adding it to Gaussian). He also added it to setup.py in order for it to be a requirement for installation. The issue is that when people `git pull` to update their development versions, python doesn't have a mechanism to realise that there have been new packages added.

I note that `Expression `still has numpy as default. @ericpre is that because numexpr does not support as many math functions as numpy does?

ericpre:
Yes, following #2032, numexpr is a hyperspy dependency and as @thomasaarholt explained above, you need to install it manually.

AEljarrat:
I already knew about the dependency problem. However, I was confused between the docstring in `Expression` and `PowerLaw`. So it seems numexpr is a sane choice here, and its better to leave it that way. Sorry for opening this useless issue, I will close it now!

ericpre:
This is true: the docstring is confusing and not consistent... this will need to be improved!

PR #2062, opened 2018-09-23
Obtain X-ray lines from incoming arguments rather than metadata

### LINKED ISSUE #2061: TEM-EDS quantification fails when specifying X-ray lines
The `s.quantification` method does not use the X-ray lines specified in `s.get_lines_intensity` to display the composition, but rather the X-ray lines in the metadata.

This caused an `IndexError: list index out of range` when my signal contained both K_a and K_b lines.

I think the issue is [this line](https://github.com/hyperspy/hyperspy/blob/RELEASE_next_minor/hyperspy/_signals/eds_tem.py#L360), which obtains X-rays from metadata and not from the intensities.

dmlundeby:
### Description of the change
Obtain X-ray lines from the intensity signals, rather than from metadata.

### Progress of the PR
- [x] Change implemented
- [x] add tests
- [x] ready for review

```python
>>> import hyperspy.api as hs
>>> s = hs.load(some_eds_signal, signal_type='EDS_TEM')
>>> s.set_elements(['Al', 'Zn'])
>>> s.add_lines(only_one=False)
>>> s.quantification(s.get_lines_intensity(xray_lines=['Al_Ka', 'Zn_Ka'], method='CL', factors=[1, 1])
```
This does no longer raise an exception.

My first attempt on bugfix here, hope it works :-)

Should fix #2061

### DIFF
--- a/hyperspy/_signals/eds_tem.py
+++ b/hyperspy/_signals/eds_tem.py
@@ -357,7 +357,7 @@ class EDSTEM_mixin:
             navigation_mask = self.vacuum_mask(navigation_mask, closing).data
         elif navigation_mask is not None:
             navigation_mask = navigation_mask.data
-        xray_lines = self.metadata.Sample.xray_lines
+        xray_lines = [intensity.metadata.Sample.xray_lines[0] for intensity in intensities]
         composition = utils.stack(intensities, lazy=False)
         if method == 'CL':
             composition.data = utils_eds.quantification_cliff_lorimer(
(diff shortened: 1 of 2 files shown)


k8macarthur:
Good spot! This fix looks good to me. It needs a test added. A simple test in the quantification section of test_eds_tem.py should suffice. Check that the x-ray lines used for quantification are those in the intensities list.

ericpre:
Thanks @dmlundeby.
Your name on github will be used to make the list of contributor for the next release, so you may want to also add your last name in order to have your full name on the list. This is also used by zenodo when making a DOI.

PR #2063, opened 2018-09-24
Set hyperspy-*-gui requirement to last version

ericpre:
Set hyperspy-*-gui requirement to last version to have the plot tab in preferences, as documented in the user guide.
Even if this is not strictly required, this is fairly convenient and doesn't cost nothing in term of dependency. It may also be worth updating the hyperspy conda recipe.

### Progress of the PR
- [x] Change implemented,
- [x] ready for review.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -68,10 +68,10 @@ install_req = ['scipy>=0.15',
 
 extras_require = {
     "learning": ['scikit-learn'],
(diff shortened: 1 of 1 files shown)


francisco-dlp:
Yes, not setting that requirement was required because hyperspy v1.4 was released before the guis, but now that they've been released, this makes sense.

PR #2065, opened 2018-09-30
Fix tests and tidy up CI

ericpre:
### Description of the change
Travis and appveyor were failing because of a few reasons; This PR fixes it, removes a few warnings generated during the tests and also tidies up CI.

### Progress of the PR
- [ ] ~~Fix decorator issue with `matplotlib` 3.0 (I could not find a way to use the public API to do this),~~
- [ ] ~~`matplotlib` 3.0 fixes a bug in the layout, so the baseline image needs to be update,~~
- [ ] ~~set requirements to `matplotlib` >= 3.0 for testings only,~~
- [x] pin matplotlib to 2.2.3 for testing,
- [x] fix `numba` issue when creating array from empty list and remove requirement `numba` < 0.39,
- [x] `sparse` is not an optional dependency, tidy up code accordingly,
- [x] fix a plotting warning in the tests of the spikes removal tool,
- [x] fix a warning during reading tests of protochips files,
- [x] appveyor: install dependencies from conda-forge channel to be consistent with the installation of hyperspy with conda (and also travis),
- [x] ready for review.


hyperspy/misc/io/fei_stream_readers.py
```diff
@@ -1,25 +1,20 @@
 import numpy as np
 import dask.array as da
+import sparse
 
 from hyperspy.decorators import jit_ifnumba
 
 
-try:
-    import sparse
```
sem-geologist, line 8: so `sparse` is going to be hard dependency ?

ericpre: It is already the case: https://github.com/hyperspy/hyperspy/blob/24619445e501cbc541d1b4179db31337d4959a8f/setup.py#L49-L67

ericpre:
The plotting tests crash in the bundle. If we want to release fairly soon, it may worth pinning matplotlib to 2.2.3 and fixing the other tests in a separate PR.

ericpre:
Also fixes the dependency requirement conflict following the last sparse release as mentioned in #2068.

hyperspy/misc/io/fei_stream_readers.py
```diff
@@ -193,24 +193,33 @@ def stream_to_sparse_COO_array(
...
-            stream_data=stream_data,
-            shape=spatial_shape,
-            channels=channels,
-            rebin_energy=rebin_energy,
-            first_frame=first_frame,
-            last_frame=last_frame,
-        )
+    # workaround for empty stream, numba "doesn't support" empty list, see
+    # https://github.com/numba/numba/pull/2184
+    if np.allclose(stream_data, np.full_like(stream_data, 65535)):
```
francisco-dlp: Wouldn't it be possible to use a workaround such as the one in https://github.com/numba/numba/issues/2152#issuecomment-254520288 instead? This one is fine when the array is indeed empty, but when it is not (the most common and useful case) it requires going twice through the data.

hyperspy/tests/io/protochips_data/invalid_protochips_file.csv
```diff
@@ -12,4 +12,4 @@ Time, Notes, Channel A Current, Channel A Voltage, Channel A Resistance, Channel
 196, , 0.0179984, 39.9278, 2218.40830296026, 0, 0, NaN
 296, , 0.0179992, 39.7102, 2206.22027645673, 0, 0, NaN
 396, , Test, 39.5232, 2195.79432762021, 0, 0, NaN
-
+497, , 0.0179994, 39.3669, 2187.1229040968, 0, 0, NaN
```
francisco-dlp: Why was the absence of this line raising a warning?

francisco-dlp (commented):
Other than the comment above, LGTM.

hyperspy/tests/io/protochips_data/invalid_protochips_file.csv
ericpre: For this line: https://github.com/hyperspy/hyperspy/blob/24619445e501cbc541d1b4179db31337d4959a8f/hyperspy/io_plugins/protochips.py#L176

Actually, there are two bugs here... for some files, the first line of data is skipped... and the slicing should be [1:-1] to avoid taking the average with the edge values.

francisco-dlp: It is worth then opening a separate issue to report the bug.

hyperspy/misc/io/fei_stream_readers.py
ericpre: Indeed, this doubles the memory requirement when loading data... it should be fixed now.

hyperspy/tests/io/protochips_data/invalid_protochips_file.csv
ericpre: Fixed.

setup.py
```diff
@@ -73,7 +73,7 @@
     "mrcz": ["blosc>=1.5", 'mrcz>=0.3.6'],
     "test": ["pytest>=3", "pytest-mpl", "matplotlib>=2.0.2"],
     "doc": ["sphinx>=1.7", "sphinx_rtd_theme"],
-    "speed": ["numba<0.39"],
+    "speed": ["numba"],
 
 }
 extras_require["all"] = list(itertools.chain(*list(extras_require.values())))
```
ericpre, line 79: I think that we should remove `"test"` from `"all"`, because there is no reason to install pytest and pytest-mpl for an user install.

francisco-dlp: I agree.

appveyor.yml
```diff
@@ -60,23 +60,19 @@ install:
...
+  - "pip install .[all]"
 
 build: false  # Not a C# project, build stuff at the test step instead.
 
 test_script:
   # Run the project tests
   - ps: Add-AppveyorMessage "Running tests..."
   - "python setup.py build_ext --inplace"
-  - "py.test --mpl"
+  - "pytest --mpl"
```
francisco-dlp: May this be the reason why pytest fails in appveyor for 3.5?

ericpre: Non, I have tried... I will revert all the changes that I made on CI, because at the moment conda doesn't resolve nicely the dependencies with python 3.5. I guess the main reason is that a few packages aren't build anymore for python 3.5 in conda-forge.
When we add support for python 3.7, we should revisit tidying up CI and possibly also adding testing with numpy and matplotlib development.

francisco-dlp:
In the interest of finishing this PR asap, what about dropping py35 from the tests and adding py37? For that we should not build the bundle for py35 and add (if possible) a py37 version. It seems that WP dropped py35 in Feb 2018.

In any case, it seems reasonable to only support (i.e. test) the two latest minor releases of Python.

sem-geologist:
if You plan to drop py35 please maybe rebase this to `NEXT_minor`, as this would be quite huge change. py35 is still on ubuntu (and derivatives) 16.04 LTS; I understand that ubuntu 18.04 is here, but some people are still keeping 16.04 (it is LTS in the end, it is supported for 5 years) as 18.04.1 still have some unsolved issues (or unsolvable issues).
Bear also in mind that you can run into some issues with py37, as some packages are still missing.
It also should be emphasized in main documentation of hyperspy that the only supported way to install hyperspy is through conda distribution (be it with `conda install` or `pip install` in conda environment). So with drop of py35, hyperspy would drop support of installing hyperspy in system-python (<3.6).
P.S. Generally, as one of those last users, which was using hyperspy on system-python in linux, I had moved to conda, after being not able to upgrade from 16.04 to 18.04 (custom pyQt version). It is much better idea to work, develop and experiment leaving system-python and its libraries intact.

sem-geologist:
@francisco-dlp , whois WP?

ericpre:
WP is WinPython.
Maybe, we can drop python 3.5 for win32 only, which is a the most annoying one... For python 3.7, we may need to wait that the [rebuilt is completed on conda-forge](https://conda-forge.org/status/).

francisco-dlp:
@sem-geologist, I share some of your concerns and I have some extra ones. However, I think that restricting the number of supported (meaning tested) versions to 2 makes sense because we must keep the burden of maintaining HyperSpy within reasonable limits. Other projects have already started not testing with py35 and, as @ericpre pointed out, some may no longer be packaging for py35 (for an example of a software package more critical for general Linux users and with a fat bigger user base see https://github.com/borgbackup/borg/issues/3919). Therefore, it's just a matter of time that some of HyperSpy's requirements wont be available in wheel form for py35.

It is fair to point out that HyperSpy will remain compatible with py35 for as long as the following two conditions as met:

* We don't use any of the new features introduced in later release (this is unlikely to happen in HyperSpy anytime soon even if we don't enforce it)
* HyperSpy requirements don't strictly require Py > 35

However, I think that, as the majority of are users are using either anaconda or the bundle we should make sure that we align with anaconda policy on this and it seems that the policy so far is to continue building for py35.

Summary: probably it is not a good idea to drop py35 support until anaconda does it.

appveyor.yml
```diff
@@ -60,15 +60,11 @@ install:
 
   # Install the dependencies of the project.
   - ps: Add-AppveyorMessage "Installing conda packages..."
-  - "%CMD_IN_ENV% conda install -yq %TEST_DEPS%"
-  - "%CMD_IN_ENV% conda install -yq %DEPS%"
-  - "pip install pytest-mpl"
-
-  # Having 'sip' folder on path confuses import of `sip`.
-  #- "%CMD_IN_ENV% conda install -yq pip"
+  - "conda install -yq -c conda-forge %TEST_DEPS%"
```
francisco-dlp, line 63: Why should we use conda-forge to install all the deps? In a standard installation most people will have downloaded the full anaconda (vs miniconda here), hence most of the dependencies would be already installed and coming from the standard anaconda repos.

setup.py
francisco-dlp: Why has this gone? I thought that it was a nice idea.

francisco-dlp (commented):
As this seems to fix an issue with the protochips reader, it would be good to open a separate issue describing the bug. Other than that an the tiny comments, LGTM.

setup.py
ericpre: This is what was breaking the installation in python 3.5. I couldn't figure out why and what is happening.... I would not have time to finish this, so this is why I reverted it.

appveyor.yml
ericpre: This was part of removing the "test" optional dependency installation, so this should be removed now but pytest-cov still need to be installed.

appveyor.yml
```diff
@@ -60,11 +59,11 @@ install:
 
   # Install the dependencies of the project.
   - ps: Add-AppveyorMessage "Installing conda packages..."
-  - "conda install -yq -c conda-forge %TEST_DEPS%"
   - "conda install -yq -c conda-forge %DEPS%"
```
ericpre, line 64: For win32, this should pull matploltib=2.2.3 from the default channel since this is not in the conda-forge channel, but it doesn't work...

ericpre:
I reverted last commit too because for python 3.5 on windows it looks to me that there is something broken depending on how the dependencies are installed.

appveyor.yml
francisco-dlp: My point was more about why conda-forge instead of the standard anaconda repository.

ericpre: pytest-mpl is not in defaults.

francisco-dlp: OK

PR #2064, opened 2018-09-28
bug-fix: Zeiss time, date, scale, new tifffile dep, tiff.py simplification

sem-geologist:
### Description of the change
This  PR aims to fix #2057 and #2056, where more recent tifffile.py (~patched for~ with Date and Time exposure for Zeiss Sem format ~in this PR~) is needed than skimage is providing.
Additionally, the structure of `io_plugins/tiff.py` was partially refactored.
Changes:
- most of encoding, decoding workarounds and functions got deleted as they are not needed with recent `tifffile.py`
- Zeiss Date and Time is correctly read into metadata. (this fixes #2057)
- New universal (for different versions of zeiss sem tifs) scale parsing code replaces previous code applicable only with 1k images (this replaces PR #2056 )
- `_parse_scale_unit` in `tiff.py` got its parts swapped into such order: `force_read_resolution and 'ResolutionUnit' in op`, then fei, zeiss or tvips, and then DM and imagej, which does not go through multiple `if`'s used for DM, when reading other vendor images, or forcing the resolution.
- most recent version of tifffile.py returns correct dtype (8bit instead of 16) for zeiss sem images using less RAM after loading, and use less of RAM-CPU bandwidth during loading (8bit instead of 48bit - 6 times less) - that speeds up loading.
- Metadata class was refactored, as it had only a single method for the other input than `self`. encapsulated mapping dictionaries were moved out from class into global dictionaries (as they present non dynamical data) so only single copy would be used. As additional result `DataTreeBrowser` was dropped from imports.

### Progress of the PR
- [x] majority of changes.
- [x] fix bugs produced by this PR in the tests (~writing  needs to be fixed~).
- [x] adjust tests.
- [x] ready for review.

PR #2079, opened 2018-10-22
CZ-tiff fixes and tifffile inclusion. re-upload of #2064

sem-geologist:
re-upload of #2064, as it was not able to be reopened.


ericpre:
Closing since #2064 is merged.

PR #2088, opened 2018-11-03
Use imageio instead of deprecated scipy functions to read and write images.

ericpre:
See https://docs.scipy.org/doc/scipy/reference/release.1.0.0.html#deprecated-features
Reading and saving images will break with the coming scipy 1.2 release.

### Progress of the PR
- [x] replace imread and imsave functions from scipy with imageio imread and imwrite,
- [x] add imageio dependency,
- [x] add support for saving color image,
- [x] add tests,
- [x] ready for review.

### DIFF
--- a/hyperspy/io_plugins/image.py
+++ b/hyperspy/io_plugins/image.py
@@ -18,9 +18,10 @@
 
 import os
 
-from scipy.misc import imread, imsave
+from imageio import imread, imwrite
 
-from hyperspy.misc.rgb_tools import regular_array2rgbx
+
+from hyperspy.misc import rgb_tools
 
 # Plugin characteristics
 # ----------------------
@@ -47,7 +48,10 @@ def file_writer(filename, signal, file_format='png', **kwds):
             The fileformat defined by its extension that is any one supported by
             PIL.
     """
-    imsave(filename, signal.data)
+    data = signal.data
+    if rgb_tools.is_rgbx(data):
+        data = rgb_tools.rgbx2regular_array(data)
+    imwrite(filename, data)
 
 
 def file_reader(filename, **kwds):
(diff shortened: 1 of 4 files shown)


sem-geologist:
so `imageio` is getting to be used in hyperspy... As I can understand it is used only for non-electron microscopy formats?

ericpre:
Yes, because this is the most simple and what scipy recommend. See https://github.com/scipy/scipy/pull/7869 for more details on the motivations.

For information on transitioning from scipy’s imread, this can be relevant:
https://imageio.readthedocs.io/en/stable/scipy.html

sem-geologist:
would not be better to make imageio optional library instead of hard requirement? The dependancy list of hyperspy is already so long. And imageio does not add any microscopy formats.`

ericpre:
> would not be better to make imageio optional library instead of hard requirement? The dependancy list of hyperspy is already so long. And imageio does not add any microscopy formats.

`imageio` add quite a [few formats](https://github.com/hyperspy/hyperspy/blob/fc934bc66b8ece5df564ac0996db5e28b70b00f2/hyperspy/io_plugins/image.py#L31) and I personally use hyperspy very often to convert data to jpg so that I can browse my data very easily. `imageio` is a pure python library, so it is installed without any problem.

sem-geologist:
@ericpre , you just identified the problem I have with hyperspy using `matplotlib` instead of i.e. `pyqtgraph`.  And that is sad, that we need one another library just to browse the data (as matplotlib is slow as hell) out from hyperspy.

francisco-dlp:
LGTM.

Regarding the dependency debate, in Rnp we should leave things as they are. I agree with @ericpre that adding this dependency costs very little as it's going to be a requirement of multiple packages that we depend on. However, it's true that it is not necessary to run hyperspy itself, so, after the split, this should be an optional dependency.

@sem-geologist, regarding matplotlib being slow at browsing images, as you probably know, the situation improves when disabling the axes, scale bar etc. But, anyway, it's certainly still slow for big images and it'll be good to implement a different alternative backend even if only for images.

PR #2090, opened 2018-11-15
Fix quant scaling error

k8macarthur:
### Description of the change
The dose calculation function used for cross section quantification previously assumed that the units of the navigation axis are in 'nm'. Therefore I've added a check to the function and convert_to_units for when the units are not 'nm'.  

Test to check whether or not different units produces the same quantification results was also added. 

As this is a bug fix no update to the user guide is needed.


k8macarthur:
This change will not change the units in the original spectrum image, just locally within the function for the correct electron dose to be calculated.

ericpre:
@k8macarthur I have rebased your branch on Rnp to fix some left over of a merge conflict and the merge commits. You will need to checkout this branch.

hyperspy/_signals/eds_tem.py
```diff
@@ -627,6 +627,10 @@ def _get_dose(self, method, beam_current='auto', live_time='auto',
                 if probe_area in parameters:
                     area = parameters.TEM.probe_area
                 else:
+                    if self.axes_manager[0].units != 'nm':
+                        self.axes_manager[0].convert_to_units('nm')
+                    if self.axes_manager[1].units != 'nm':
+                        self.axes_manager[1].convert_to_units('nm')
```
ericpre: This is changing the units in place. The following should work:
```python
pixel_size = []
# in case there are more than 2 navigation axes.
for axis in self.axes_manager.navigation_axes[:2]:                    
    pixel_size.append(axis.convert_to_units('nm', inplace=False)[0])
```
In particularly, for a dataset with navigation dimension > 2, how do we know if we pick the correct navigation axes? Here, I assume that the first two items of the `axes_manager.navigation_axes` are the "x" and "y" spatial dimension but it could something else!
Depending on how this is implemented, this should be documented.

hyperspy/_signals/eds_tem.py
```diff
@@ -627,6 +627,10 @@ def _get_dose(self, method, beam_current='auto', live_time='auto',
...
                     pixel1 = self.axes_manager[0].scale
                     pixel2 = self.axes_manager[1].scale
```
ericpre: By the method to get the dose will be wrong for linescan: `self.axes_manager[1].scale` could be a signal axis..

hyperspy/tests/signal/test_eds_tem.py
```diff
@@ -225,6 +227,24 @@ def test_quant_zeta(self):
...
+        s.axes_manager[1].scale = 0.5/1000
+
+        method = 'cross_section'
+        factors = [3, 5]
+        intensities = s.get_lines_intensity()
+        res = s.quantification(intensities, method, factors)
+        res2 = s2.quantification(intensities, method, factors)
+        np.testing.assert_allclose(res[0][0].data, res2[0][0].data)
+        # Check that the quantification doesn't change the units of the signal
+        assert s.axes_manager[0].units == 'µm'
```
ericpre, line 245: This is this assertion which need to be true. At the moment, the units are changed inplace.

ericpre (changes requested):
Two things need to be fixed:
- not to change the units in place,
- currently `get_dose("cross section")` works only for spectrum image.

See more detailed comments below.

ericpre (approved):

ericpre:
I have fixed the comments of my review.

francisco-dlp:
LGTM

PR #2092, opened 2018-11-16
direct dict2dict mapping (moving toward independence of io_plugins)

sem-geologist:
### Description of the change
1. This PR adds more efficient dictionary mapping, which is used in io_plugins. This PR aims removing any usage of DictionaryTreeBrowser (DTB) from io_plugins. Many io_plugins are creating and destroying DTB only to map one dictionary structure to the another, which is highly inefficient and makes io_plugins to depend directly to hyperspy. New `map_dict_to_dict` is going to be drop-in replacement, and needs no changes to currently used io_plugin mapping dictionaries.

2. looks that some other plugins uses xml sparingly. The MSXML bug-workaround methods and dictionarization of etree is moved to separate file accessible by all io_plugins (`dict_tools.py`). The path of this collection of tools should be discussed. (currently it is `misc/io/dict_tools.py`)

3. Bruker.py is refactored – that is getting rid of step-by-step building of `metadata` dictionaries (often using if-else) into parsing only `original_metadata`, and moving the `metadata` generation to be done completely by `mapping` mechanism. That simplifies code, and removes duplicate code (i.e. separate (bcf and spx) dict building.

4. **to be filled later...**

### Progress of the PR
- [x] New `map_dict_to_dict` function implemented
- [ ] refactor/simplyfy bruker.py heavy using mapping, making it more universal and showcase the simplicity of the mapping.
- [ ] refactor currently used DTB for mapping in `io_plugins` and replace with new function.
- [ ] update docstring (if appropriate), (?)
- [ ] update user guide (if appropriate), (?)
- [ ] update tests,
- [ ] ready for review.

### DIFF
--- a/hyperspy/io.py
+++ b/hyperspy/io.py
@@ -32,6 +32,7 @@ from hyperspy.misc.utils import stack as stack_method
 from hyperspy.io_plugins import io_plugins, default_write_ext
 from hyperspy.exceptions import VisibleDeprecationWarning
 from hyperspy.ui_registry import get_gui
+from hyperspy.misc.io.dict_tools import map_dict_to_dict
 
 _logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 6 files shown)


sem-geologist:
I am closing this PR, as (1) I see many new io_plugins were created and many depends on different parts of hyperspy (additionally to DictionaryTreeBrowser). Separation of io from hyperspy looks ongoing walk to steeper and steeper mountain. (2) Work on #1978 looks like had completely stalled. (3) There was completely no interest in this PR. Looks that everyone are content how things (slowishly) works. I think it is possible that no one see io_plugin separation from hyperspy anymore beneficial, as anyway everyone would use hyperspy due to signals and its convenience features. (4) this is based on very old version, implementation should anyway be done as new PR based of current Hyperspy codebase, but making it useful for all io_plugins is rather dubious.

Issue #2093, opened 2018-11-16
Issue when loading learning results


dmlundeby:
When learning results are loaded from a .npz file, then some attributes are stored as 0D numpy arrays instead of its original datatype. For example, the attribute `decomposition_algorithm` is stored as an 0D numpy array instead of a string.

This causes an error when loading learning results and then storing the signal.

ericpre:
Fixed in #2094.

PR #2094, opened 2018-11-16
Fix issue when loading decomposition results

dmlundeby:
Fix bug causing attributes to be saved as 0D numpy arrays.

Fix by converting the attributes to their original Python datatype.

### Progress of the PR
- [x] Change implemented
- [x] Add tests
- [x] Ready for review

```python
>>> import hyperspy.api as hs
>>> s = hs.load(multidimensional_eds_signal)
>>> s.change_dtype('float64')
>>> s.decomposition()
>>> s.learning_results.save('results.npz')
>>> s.learning_results.load('results.npz')
>>> s.save('out.hspy')
```

This does no longer cause an error.

### DIFF
--- a/hyperspy/learn/mva.py
+++ b/hyperspy/learn/mva.py
@@ -1281,6 +1281,9 @@ class LearningResults(object):
         for key, value in decomposition.items():
             if value.dtype == np.dtype('object'):
                 value = None
+            # Unwrap values stored as 0D numpy arrays to raw datatypes
+            if isinstance(value, np.ndarray) and value.shape == ():
+                value = value.item()
             setattr(self, key, value)
         _logger.info("\n%s loaded correctly" % filename)
         # For compatibility with old version ##################
@@ -1318,11 +1321,6 @@ class LearningResults(object):
         if hasattr(self, 'ica_factors'):
             self.bss_factors = self.ica_factors
             del self.ica_factors
-        #
-        # Output_dimension is an array after loading, convert it to int
-        if hasattr(self, 'output_dimension') and self.output_dimension \
-                is not None:
-            self.output_dimension = int(self.output_dimension)
         _logger.info(self._summary())
 
     def summary(self):
--- a/hyperspy/learn/mva.py
+++ b/hyperspy/learn/mva.py
@@ -1282,7 +1282,7 @@ class LearningResults(object):
             if value.dtype == np.dtype('object'):
                 value = None
             # Unwrap values stored as 0D numpy arrays to raw datatypes
-            if isinstance(value, np.ndarray) and value.shape == ():
+            if isinstance(value, np.ndarray) and value.ndim == 0:
                 value = value.item()
             setattr(self, key, value)
         _logger.info("\n%s loaded correctly" % filename)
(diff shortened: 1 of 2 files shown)


hyperspy/learn/mva.py
```diff
@@ -1281,6 +1281,8 @@ def load(self, filename):
         for key, value in decomposition.items():
             if value.dtype == np.dtype('object'):
                 value = None
+            if isinstance(value, np.ndarray) and value.shape == ():
```
ericpre: Add a comment to explain why we need to do this here.
Also, I would suggest to use `value.size == 1` instead of `value.shape == ()`

ericpre (changes requested):
See one comment above and a test needs to be added, since this is not tested at all (https://coveralls.io/builds/20147432/source?filename=hyperspy%2Flearn%2Fmva.py#L1258). Your example is good enough, it only needs small adjustments to add it the test suite. For example, to generate a dummy signal, simply use `s = hs.signals.Signal1D(np.random.random((5, 10))) ` and to check if the file save correctly, you can use:

```python
import tempfile
with tempfile.TemporaryDirectory() as tmpdir:
    fname2 = os.path.join(tmpdir, "test_save_load_cycle_learning_results.hspy")
```

ericpre:
Also since this is a bug fix, it should go to the `RELEASE_next_patch` branch instead of the `RELEASE_next_minor`. You may need to rebase and possibly force push your commits to this branch.

ericpre:
Fix #2093.

dmlundeby:
Thank you for feedback! I'll do the fixes after end of semester. 
I did start a test case, but didn't know how to save and load test files properly. I'll add the test using `tempfile`.

hyperspy/learn/mva.py
dmlundeby: I think it is best to retain `value.shape == ()` because the issue is only 0D numpy arrays and not higher-dimensional numpy arrrays.

hyperspy/learn/mva.py
```diff
@@ -1318,11 +1321,6 @@ def load(self, filename):
         if hasattr(self, 'ica_factors'):
             self.bss_factors = self.ica_factors
             del self.ica_factors
-        #
-        # Output_dimension is an array after loading, convert it to int
-        if hasattr(self, 'output_dimension') and self.output_dimension \
```
dmlundeby, line 1323: Removed this check as it is included in the check [above](https://github.com/hyperspy/hyperspy/pull/2094/files#diff-cd68ac14189365f87b724293fa3c9f38R1285).

francisco-dlp:
The tests are failing due to #2120

hyperspy/learn/mva.py
ericpre: Ok, then would it be more specific (and clear) to use `value.ndim == 0`?

dmlundeby: Yes indeed. I've updated the branch now.

ericpre (approved):

ericpre:
Thanks!

PR #2099, opened 2018-11-23
Bugfix subplots legend plot spectra

AEljarrat:
Dear all, 

I like to use "hs.plot.plot_spectra" to do quick comparisons, especially since the animated legend allows me to set the visibility of the lines. It is also possible to use this command to fill several subplots in the same figure. However, in this case the animated legend cannot by used. 

### Description of the change
- This small PR fixes this bug. 

### Progress of the PR
- [x] Change implemented.
- [x] ready for review.

```python
>>> %matplotlib qt
>>> import hyperspy.api as hs
>>> import numpy as np
>>> 
>>> Ns = 5
>>> Nc = 256
>>> 
>>> data = np.random.rand(Ns, Nc)
>>> name = np.arange(Ns).astype('str')
>>> s    = hs.signals.Signal1D(data)
>>> 
>>> # This works
>>> hs.plot.plot_spectra(s.inav[:2], legend=name[:2])
>>> hs.plot.plot_spectra(s.inav[2:], legend=name[2:])
>>> 
>>> # This does not work
>>> f, ax = plt.subplots(1, 2)
>>> hs.plot.plot_spectra(s.inav[:2], legend=name[:2], fig=f, ax=ax[0])
>>> hs.plot.plot_spectra(s.inav[2:], legend=name[2:], fig=f, ax=ax[1])
```

### DIFF
--- a/hyperspy/drawing/utils.py
+++ b/hyperspy/drawing/utils.py
@@ -1319,7 +1319,7 @@ def plot_spectra(
             ax.legend(legend, loc=legend_loc)
             _reverse_legend(ax, legend_loc)
             if legend_picking is True:
-                animate_legend(figure=fig)
+                animate_legend(fig=fig, ax=ax)
     elif style == 'cascade':
         if fig is None:
             fig = plt.figure(**kwargs)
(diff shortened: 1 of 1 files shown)


ericpre:
Nice thanks!

PR #2102, opened 2018-12-01
Disable a couple of CI builds

ericpre:
At the moment, the appveyor build takes more 2h to complete, because we are testing 6 different builds and there are run sequentially. The situation is significantly with travis, which is running 5 builds at the same but I would also suggest to disable the python 3.5 and python 3.6 mac osx build to save CI time.

The idea is to continuously test python 3.5 and python 3.6 support with linux and to other disable build for a release when the build will be enabled to generate the wheels. This should be a good compromise between testing across different platform and python version and saving CI time.

On a side node, conda-forge is not packaging for python 3.5 anymore (only python 3.6 and python 3.7).

### Progress of the PR
- [x] Disable python 3.5 build in appveyor,
- [x] enable one win32 build,
- [x] disable python 3.5 and python 3.6 on mac osx,
- [x] ready for review.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,13 +18,14 @@ matrix:
   - env: export PYTHON=3.7; FAIL_ON_EXTERNAL_DEPRECATION='False'; MINIMAL_ENV='False'
     os: osx
     language: generic
-    # Uncomment these build to generate wheels
-  # - env: export PYTHON=3.6; FAIL_ON_EXTERNAL_DEPRECATION='False'; MINIMAL_ENV='False'
-  #  os: osx
-  #  language: generic
-  #- env: export PYTHON=3.5; FAIL_ON_EXTERNAL_DEPRECATION='False'; MINIMAL_ENV='False'
-  #  os: osx
-  #  language: generic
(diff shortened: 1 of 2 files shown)


francisco-dlp:
Isn't it possible to run the disabled builds only on release instead of having to comments and uncomment the code manually?

ericpre:
I did look at it and could not a way to do but a 2nd iteration on this may be working! ;)

For reference:
https://www.appveyor.com/blog/2018/09/11/how-to-skip-build-matrix-jobs-conditionally/

francisco-dlp:
Could you create a tag in your repo (if we do it in this one it'll create a DOI in zenodo) to test if it's really working?

ericpre:
Yes, it works:
- with tag: https://ci.appveyor.com/project/ericpre/hyperspy/builds/20792589
- without tag: https://ci.appveyor.com/project/ericpre/hyperspy/builds/20792300

francisco-dlp:
Isn't it possible to do the same in travis?

ericpre:
Done for travis:
https://travis-ci.org/ericpre/hyperspy/builds/463927066
https://travis-ci.org/ericpre/hyperspy/builds/463925979

PR #2111, opened 2018-12-07
Fix axis argument in call to sum() when scaling SI to normalize Poisson noise during lazy decomp

hakonanes:
### Progress of the PR
- [ ] Change implemented,
- [ ] add tests,
- [ ] ready for review.

Running
```python
>>> import hyperspy.api as hs
>>> s = hs.load(diffraction_patterns, lazy=True)
>>> s.change_dtype('float')
>>> s.decomposition(algorithm='PCA', output_dimension=200, normalize_poissonian_noise=True)
```
does no longer throw
```python

  File "<ipython-input-1-88d7158fb358>", line 32, in <module>
    normalize_poissonian_noise=poisson_noise)

  File "/home/hakon/miniconda3/envs/pxm/lib/python3.6/site-packages/hyperspy/_signals/lazy.py", line 764, in decomposition
    data.sum(axis=range(ndim)),
[...]
  File "/home/hakon/miniconda3/envs/pxm/lib/python3.6/site-packages/dask/array/utils.py", line 144, in validate_axis
    raise TypeError("Axis value must be an integer, got %s" % axis)

TypeError: Axis value must be an integer, got range(0, 2)
```
since the it now reads `tuple(range(ndim))` instead of just `range(ndim)`.

### DIFF
--- a/hyperspy/_signals/lazy.py
+++ b/hyperspy/_signals/lazy.py
@@ -761,8 +761,8 @@ class LazySignal(BaseSignal):
                 ndim = self.axes_manager.navigation_dimension
                 sdim = self.axes_manager.signal_dimension
                 bH, aG = da.compute(
-                    data.sum(axis=range(ndim)),
-                    data.sum(axis=range(ndim, ndim + sdim)))
+                    data.sum(axis=tuple(range(ndim))),
+                    data.sum(axis=tuple(range(ndim, ndim + sdim))))
                 bH = da.where(sm, bH, 1)
                 aG = da.where(nm, aG, 1)
 


francisco-dlp:
Looks good except that you should have branched it from Release next patch. Don't worry, I'll fix that, but next time think on branching from Rnp when you are submitting a bug fix.

Also, it'll be good to add a test. Would you like to do that? If not, don't worry, we can take care of it since, unfortunately, there are not tests for this feature that you can easily modify, what explains why this got broken without anybody noticing it.

hakonanes:
Oh, sorry, thanks for letting me know. I could write a test, however, I do not know where to start (tests/signal/test_lazy.py? tests/mva/test_decomposition.py?) since I have not done it before... so perhaps it is best if you do it.

francisco-dlp:
OK, I'll merge this and I'll open an issue about the missing test. Thanks for your contribuion!

Issue #2104, opened 2018-12-04
s.remove_background non-interactive: lose metadata and axes_manager info


magnunor:
The output signal from `s.remove_background(signal_range=(..., ...))` does not retain either the metadata or axes_manager info.

Does occurs both in `RELEASE_next_patch`, the most recent HyperSpy bundle, and most recent Anaconda release.

The immediate consequence is that the EELS notebook is currently broken for the workshop next week.

```python
import numpy as np
import hyperspy.api as hs

s = hs.signals.Signal1D(np.random.random(size=(10, 500)))
s.axes_manager[0].name = 'axis0'
s.axes_manager[1].name = 'axis1'
s.axes_manager[0].units = 'units0'
s.axes_manager[1].units = 'units1'
s.axes_manager[0].scale = 0.9
s.axes_manager[1].scale = 1.1
s.metadata.General.title = "atitle"

s_r = s.remove_background(signal_range=(1, 100))
```

`s.axes_manager` output:
```
<Axes manager, axes: (10|500)>
            Name |   size |  index |  offset |   scale |  units
================ | ====== | ====== | ======= | ======= | ======
           axis0 |     10 |      0 |       0 |     0.9 | units0
---------------- | ------ | ------ | ------- | ------- | ------
           axis1 |    500 |        |       0 |     1.1 | units1
```

`s_r.axes_manager` output:
```
<Axes manager, axes: (10|500)>
            Name |   size |  index |  offset |   scale |  units
================ | ====== | ====== | ======= | ======= | ======
     <undefined> |     10 |      0 |       0 |       1 | <undefined>
---------------- | ------ | ------ | ------- | ------- | ------
     <undefined> |    500 |        |       0 |       1 | <undefined>
```

`s.metadata.General.title` output:
```
'atitle'
```

`s_r.metadata.General.title` output:
```
''
```

thomasaarholt:
Any idea on how this came about? This can't have been the case all along?

francisco-dlp:
It's amazing how buggy such a simple function can be. I rarely use this functionality but today I've used it and I found that with fast enabled the background was not properly subtracted, so one more bug.

ericpre:
Fixed in #2156.

PR #2114, opened 2018-12-09
Fix bug in non-interactive s.remove_background(fast=True)

magnunor:
Fixes bug in `s.remove_background(signal_range=(..., ...), fast=True)`, where the metadata and axes_manager information was not copied over to the result.

More info: https://github.com/hyperspy/hyperspy/issues/2104

- [x] Fix bug
- [ ] Add tests
- [ ] ready for review.

### DIFF
--- a/hyperspy/_signals/signal1d.py
+++ b/hyperspy/_signals/signal1d.py
@@ -1036,8 +1036,8 @@ _spikes_diagnosis,
         if fast and not self._lazy:
             try:
                 axis = self.axes_manager.signal_axes[0].axis
-                result = Signal1D(self.data -
-                                  background_estimator.function_nd(axis))
+                result = self - Signal1D(
+                        background_estimator.function_nd(axis))
             except MemoryError:
                 result = self - model.as_signal(
                     show_progressbar=show_progressbar)
--- a/hyperspy/_signals/signal1d.py
+++ b/hyperspy/_signals/signal1d.py
@@ -1036,8 +1036,7 @@ _spikes_diagnosis,
         if fast and not self._lazy:
             try:
                 axis = self.axes_manager.signal_axes[0].axis
-                result = self - Signal1D(
-                        background_estimator.function_nd(axis))
+                result = self - background_estimator.function_nd(axis)
             except MemoryError:
                 result = self - model.as_signal(
                     show_progressbar=show_progressbar)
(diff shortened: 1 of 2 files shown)


magnunor:
@ericpre mentioned that this fix might lead to a slowdown when using `s.remove_background` interactively. Having compared this pull request with the current RELEASE_next_patch, there doesn't seem to be any difference. However, this might be due to my specific computer setup.

Could some else see if this fix reduces the responsiveness when running `s.remove_background` interactively?

hyperspy/_signals/signal1d.py
```diff
@@ -1036,8 +1036,8 @@ def _remove_background_cli(
         if fast and not self._lazy:
             try:
                 axis = self.axes_manager.signal_axes[0].axis
-                result = Signal1D(self.data -
-                                  background_estimator.function_nd(axis))
+                result = self - Signal1D(
+                        background_estimator.function_nd(axis))
```
ericpre: Actually, `Signal1D` is not necessary here and by removing it we can avoid the overhead of the creation of a signal, which make it 10% faster, not very significant but still something.

ericpre (changes requested):
Actually, I was mistaken with the slow down in the [power law estimation](https://github.com/ericpre/hyperspy/blob/c314c8266c89dcacc04b9fe6f87668376e8abdd7/hyperspy/_components/power_law.py#L128) as discussed in #2032. The slow down here is only of ~10%.

hyperspy/_signals/signal1d.py
magnunor: Good point! Fixed.

magnunor:
I added a bunch of tests to reduce the odds of `s.remove_background` silently breaking again.

magnunor:
I also think I discovered some new bugs through these tests:

```python
s = hs.signals.Signal1D(np.arange(10, 100)[::-1])
s_r = s.remove_background(signal_range=(20, 50), background_type='Polynomial', fast=True)
print(s.data.shape)
(90,)
print(s_r.data.shape)
(3, 90)
```

Similar for `background_type='Offset'` gives `(90,)` and `(1, 90)`.

magnunor:
Closing this, as it has been superseded by https://github.com/hyperspy/hyperspy/pull/2156

PR #2130, opened 2019-02-13
Make plot EDS maps easier

ericpre:
When using `hs.plot.plot_images` to plot a list of EDS maps generated from `s.get_lines_intensities`, the maps need to be transposed, if not an error is raised. This PR make the `hs.plot.plot_images` clever enough, so that it can figure out when to transpose the maps and do it when appropriate.

Same apply for 1D EDS profile.

### Progress of the PR
- [x] `hs.plot.plot_images` and `hs.plot.plot_spectra` transpose the data when appropriate,
- [x] add tests,
- [x] ready for review.


jat255 (commented):
Would it be worth checking in here if the signal type is EDS? Are there any other cases where the signal dimension could be zero and this could do something unexpected?

ericpre:
I don't see why it would need to be EDS signal actually, even if currently it would be useful only for EDS because there are no functions similar to `get_lines_intensities` to get EELS maps from an `EELSModel`. I guess it would be a convenient to have one.

francisco-dlp:
LGTM. Only documenting this in the docstrings and possibly in the User Guide is left.

ericpre:
> LGTM. Only documenting this in the docstrings and possibly in the User Guide is left.

Done.

hyperspy/drawing/utils.py
```diff
@@ -458,137 +470,141 @@ def plot_images(images,
...
-        `tight_layout` is known to be quite brittle, so an option is provided
-        to disable it. Turn this option off if output is not as expected,
-        or try adjusting `label`, `labelwrap`, or `per_row`
+    Extra keyword arguments are passed to `matplotlib.figure`.
+
+    Parameters
+    ----------
+    images : list of Signal2D or BaseSignal
+        `images` should be a list of Signals to plot. For `BaseSignal` with
+        navigation dimensions of 2, the signal will be tranposed, which is
```
francisco-dlp: What about: "for signals with navigation dimension 2 and signal dimension 0...".

I would remove the remark regarding EDS maps since this feature is useful in many other cases.

hyperspy/drawing/utils.py
```diff
@@ -1186,9 +1206,11 @@ def plot_spectra(
 
     Parameters
     ----------
-    spectra : iterable object
-        Ordered spectra list to plot. If `style` is "cascade" or "mosaic"
-        the spectra can have different size and axes.
+    spectra : list of Signal1D or BaseSignal
+        Ordered spectra list of signal to plot. If `style` is "cascade" or
```
francisco-dlp, line 1210: As above.

francisco-dlp (commented):
Other than those small comments, I suggest to add a ``new in version`` note to the plotting section of the User Guide describing the feature and when it was introduced.

Issue #1400, opened 2017-01-23
msa locale settings


sem-geologist:
I wanted to save my spectra to msa format however I got this:

```python
> max_pixel_spectrum
<EDSSEMSpectrum, title: Stack of EDX, dimensions: (|1347)>
> max_pixel_spectrum.save(filename='max_pix_spectra.msa',
                        extension='msa')

---------------------------------------------------------------------------
Error                                     Traceback (most recent call last)
<ipython-input-16-f43a23994200> in <module>()
      1 max_pixel_spectrum.save(filename='max_pix_spectra.msa',
----> 2                         extension='msa')

/mnt/bulk-data/development/git/hyperspy/hyperspy/signal.py in save(self, filename, overwrite, extension, **kwds)
   2027             basename, ext = os.path.splitext(filename)
   2028             filename = basename + '.' + extension
-> 2029         io.save(filename, self, overwrite=overwrite, **kwds)
   2030 
   2031     def _replot(self):

/mnt/bulk-data/development/git/hyperspy/hyperspy/io.py in save(filename, signal, overwrite, **kwds)
    464             overwrite = overwrite_method(filename)
    465         if overwrite is True:
--> 466             writer.file_writer(filename, signal, **kwds)
    467             _logger.info('The %s file was created' % filename)
    468             folder, filename = os.path.split(os.path.abspath(filename))

/mnt/bulk-data/development/git/hyperspy/hyperspy/io_plugins/msa.py in file_writer(filename, signal, format, separator, encoding)
    327             loc = locale.getlocale(locale.LC_TIME)
    328             if os_name == 'posix':
--> 329                 locale.setlocale(locale.LC_TIME, ('en_US', 'latin-1'))
    330             elif os_name == 'windows':
    331                 locale.setlocale(locale.LC_TIME, 'english')

/usr/lib/python3.5/locale.py in setlocale(category, locale)
    592         # convert to string
    593         locale = normalize(_build_localename(locale))
--> 594     return _setlocale(category, locale)
    595 
    596 def resetlocale(category=LC_ALL):

Error: unsupported locale setting
```

I am using debian testing. Why this is trying to use locale?

francisco-dlp:
This format stores the time information using us locales, so they're required. We should document this better. In most OS they're installed by default, but in my Debian I had to install them manually too. Probably we should document this in the User Guide.

sem-geologist:
[...]

PR #2132, opened 2019-02-26
Fix saving msa on linux

thomasaarholt:
The latin-1 locale did not exist on my version of Ubuntu (by default, at least). This raised an error when saving .msa files. If someone else running Linux or Mac could try the code snippet below to confirm that this is a common issue, that would be great.

```python
s.original_metadata.FORMAT = None
s.metadata.General.date = "2019-02-11"
s.save('test.msa', overwrite=True)
```

### DIFF
--- a/hyperspy/io_plugins/msa.py
+++ b/hyperspy/io_plugins/msa.py
@@ -325,7 +325,7 @@ def file_writer(filename, signal, format=None, separator=', ',
             # their name depends on library versions, platform etc.
             loc = locale.getlocale(locale.LC_TIME)
             if os_name == 'posix':
-                locale.setlocale(locale.LC_TIME, ('en_US', 'latin-1'))
+                locale.setlocale(locale.LC_TIME, ('en_US', 'utf8'))
             elif os_name == 'windows':
                 locale.setlocale(locale.LC_TIME, 'english')
             try:


hyperspy/io_plugins/msa.py
```diff
@@ -325,7 +325,7 @@ def file_writer(filename, signal, format=None, separator=', ',
             # their name depends on library versions, platform etc.
             loc = locale.getlocale(locale.LC_TIME)
             if os_name == 'posix':
-                locale.setlocale(locale.LC_TIME, ('en_US', 'latin-1'))
+                locale.setlocale(locale.LC_TIME, ('en_US', 'utf8'))
```
ericpre, line 328: Actually, I don't think this is necessary to specify the encoding.

sem-geologist:
This is wrong on so many levels. The msa plugin should not change system locale settings whatever - it really is very bad and dangerous thing. Why msa plugin is not formating dates using python built-in `datetime, date, time` utilities?

thomasaarholt:
@sem-geologist I didn't want to make big changes, just get it working. But I agree, this should be done properly. 

@ericpre, do you remember why you implemented it this way? I'm happy to rewrite that section - we should use `hyperspy.misc.date_time_tools` in this case.

ericpre:
> This is wrong on so many levels. The msa plugin should not change system locale settings whatever - it really is very bad and dangerous thing. Why msa plugin is not formating dates using python built-in `datetime, date, time` utilities?

This is already using standard python modules and in particular the `datetime` module. The actual issue is taht msa files requires US formatting of date and the standard python `datetime` module write the date depending on the local... which is why the locale need to be changed for the time the date is to be written correctly.
I agree that this is not a great solution and if I remember correctly the only alternative would be to depend on a non-standard library, such as pytz or other...
This is not a good approach but this is not a big deal neither because it changes the locale temporary (revert it back after writing the date) and only in this is specific python process, it doesn't affect any other processes.

> @ericpre, do you remember why you implemented it this way? I'm happy to rewrite that section - we should use hyperspy.misc.date_time_tools in this case.

This plugin has been mainly written by others and I think I just did a few tweaks. My comment above should summarise the reason to use this approach.
I don't see how `hyperspy.misc.date_time_tools` would be useful here?

thomasaarholt:
I don't understand the what you mean by US formatting. In the [spec](http://www.amc.anl.gov/ANLSoftwareLibrary/02-MMSLib/XEDS/EMMFF/EMMFF.IBM/Emmff.Total) (search "#DATE") says it dates and time ala "11-Feb-2012" and 12:00. How is that US formatting?

Why not just something like this?
```python
>>> s = hs.datasets.example_signals.EDS_TEM_Spectrum()
>>> s.metadata.General.date
'14.10.2014'
>>> parser.parse(s.metadata.General.date).strftime("%d-%b-%Y")
'14-Oct-2014'

```

> I don't see how hyperspy.misc.date_time_tools would be useful here?

It was partly meant as a reminder to future me if I ended up working on this, that we already have a module for working with dates.

thomasaarholt:
[Ah, I now see.](https://docs.python.org/2/library/datetime.html#strftime-and-strptime-behavior)

We could just write a converter from `month_in_number_form` to `month_in_three_letter_name`.

ericpre:
xref https://github.com/hyperspy/hyperspy/issues/1400.

ericpre:
Fixed in #2171.

PR #2134, opened 2019-03-05
Make Lorentzian an Expression

jlaehne:
### Description of the change
Lorentzian component rewritten as expression. The parameter `gamma` is renamed `_gamma` to avoid a conflict with an existing function in simpy. Additional setter/property functions allow for backward compatibility with the gamma parameter. The docstring was improved for better consistency with other components.

### Progress of the PR
- [X] add simple test
- [X] ready for review.

```python
>>> import hyperspy.api as hs
>>> lorentz1 = hs.model.components1D.Lorentzian2(A=2., Gamma=0.5, centre=1.)
>>> x = np.arange(lorentz1.centre.value-lorentz1.Gamma.value*5,lorentz1.centre.value+lorentz1.Gamma.value*5,step=0.01)
>>> s = hs.signals.Signal1D(lorentz1.function(x))
>>> s.axes_manager.signal_axes[0].axis = x
>>> m = s.create_model()
>>> m.append(lorentz1)
>>> m.fit()
>>> m.plot()
>>> m.print_current_values()
```


hyperspy/_components/lorentzian2.py
```diff
@@ -0,0 +1,74 @@
...
+    +---------------------+-----------+
+    |      :math:`x_0`    |  centre   |
+    +---------------------+-----------+
+    
+    For convenience the `fwhm` attribute can be used to get and set
+    the full-with-half-maximum.
+
+    """
+
+    def __init__(self, A=1., Gamma=1., centre=0., module="numexpr", **kwargs):
```
ericpre: Use `gamma` for the API and `_gamma` internally (in the expression) and set both parameters as being identical, so there is no need to create a new `Lorentzian2`?

hyperspy/_components/lorentzian2.py
```diff
@@ -0,0 +1,74 @@
...
+        self.A.bmin = 0.
+        self.A.bmax = None
+
+        self.Gamma.bmin = 0.
+        self.Gamma.bmax = None
+
+        self.isbackground = False
+        self.convolved = True
+
+
```
ericpre: Add `gamma` property and setter to make `_gamma` and `gamma` identical.

ericpre (changes requested):
As explained in the comments, it is possible to keep the same API and there is no need to create a new `Lorentzian2`: to workaround the `gamma` name issue with sympy, one solution is to use `_gamma` internally with adding a property and a setter to make `gamma` == `_gamma`.
It also needs a test to check that the component can be created without any errors. `tests/component/test_components.py` should be a good place for this test.

jlaehne:
I have tried the solution of using either `_gamma` or `Gamma` internally with appropriate setter/property functions. However, in that case you access the value of the parameter (for an instance `lorentz` of the component) through `lorentz.gamma` instead of through `lorentz.gamma.value`. For full backward compatibility, how would I implement the setter/property function so that I can also access `gamma.value`? and not only `_gamma.value`?

ericpre:
> For full backward compatibility, how would I implement the setter/property function so that I can also access `gamma.value`? and not only `_gamma.value`?

The following should work:
```python
@property
def gamma(self):
    return self._gamma

@gamma.setter
def gamma(self, value):
    self._gamma = value
```

jlaehne:
Thanks, my fault. I tried to set program the setter for gamma.value instead just gamma. 

Thus, the newest commit throws out the Lorentzian2 and instead uses the new expression definition directly for the Lorentzian.

Docstring improved for consistency with other components.

hyperspy/_components/lorentzian.py
```diff
@@ -16,76 +16,78 @@
...
+from hyperspy._components.expression import Expression
 
-from hyperspy.component import Component
+class Lorentzian2(Expression):
 
-
-class Lorentzian(Component):
-
-    r"""Cauchy-Lorentz distribution (a.k.a. Lorentzian function) component
+    r"""Cauchy-Lorentz distribution (a.k.a. Lorentzian function) component.
```
ericpre, line 24: Keep `Lorentzian` (and not `Lorentzian2`)!

hyperspy/_components/lorentzian2.py
```diff
@@ -0,0 +1,93 @@
+# -*- coding: utf-8 -*-
```
ericpre: This is not needed anymore!

hyperspy/_components/lorentzian.py
```diff
@@ -16,76 +16,78 @@
...
+        A : float
+            Height parameter, where :math:`A/\gamma` is the maximum of the peak.
+        gamma : float
+            Scale parameter corresponding to the half-width-at-half-maximum of the peak, which corresponds to the interquartile spread.
+        centre : float
+            Location of the peak maximum.
+        **kwargs
+            Extra keyword arguments are passed to the ``Expression`` component.
+            An useful keyword argument that can be used to speed up the
+            component is `module`. See the ``Expression`` component
```
ericpre: I would remove this sentence: "An useful keyword argument that can be used to speed up the component is `module`." because `numexpr` is now already the default, there is currently no room left for speed improvement...
Until recently, `numexpr` was an optional dependency and `numpy` was used by default, so I guess this sentence was mentioned in a few places so that people are informed and they can install and use `numexpr` if there can/wish.

ericpre (changes requested):
Still a few things to tidy up/sort out -see comments.
I would also suggest to add a simple test to check that the component can be created without any error (for example, here without the `_gamma` workaround the function will not "compile"). Actually when you are at it, it would be good to do it for all expression-based components.

jlaehne:
OK, @ericpre, I have created a set of test functions for Lorentzian. Can you have a look. If that is OK, I will create test functions for all the other components already defined as expressions. I have also worked on consistency in the docstrings for all expression based components.

hyperspy/_components/lorentzian.py
```diff
@@ -16,76 +16,75 @@
...
-    +---------------------+-----------+
-    |      :math:`a`      |     A     |
-    +---------------------+-----------+
-    |    :math:`\gamma`   |   gamma   |
-    +---------------------+-----------+
-    |      :math:`x_0`    |  centre   |
-    +---------------------+-----------+
+        f(x)=\frac{A}{\pi}\left[\frac{\gamma}{\left(x-x_{0}\right)^{2}+\gamma^{2}}\right]
+
+    ================= ===========
```
ericpre: What is the reason to change the header of the table? I would leave this kind of changes to be discussed in another PR (#2148 should be a better place?) so it can be done consistently across all components.

jlaehne: See discussion at #2148

hyperspy/_components/lorentzian2.py
jlaehne: It still seems to be part of all hyperspy files I checked.

ericpre: Sorry not to have been clear, I meant the file `lorentzian2` is not needed anymore!

hyperspy/_components/lorentzian.py
ericpre: Similarly to other similar PRs, please use the convention already in place for the table and propose your suggestion in #2148.

jlaehne: Done, also for  #2149 (logistic function).

jlaehne (commented):
good point

hyperspy/_components/lorentzian.py
```diff
@@ -16,76 +16,82 @@
...
+
+    @fwhm.setter
+    def fwhm(self, value):
+        self._gamma.value = value / 2
+
+    @property
+    def gamma(self):
+        return self._gamma
+
+    @gamma.setter
```
francisco-dlp: I think that this property setter is unnecessary because users don't need to set the gamma parameter.

hyperspy/tests/component/test_lorentzian.py
```diff
@@ -0,0 +1,46 @@
...
+    g = Lorentzian()
+    g.A.value = 1.5*np.pi
+    g.gamma.value = 1
+    g.centre.value = 2
+    assert_allclose(g.function(2), 1.5)
+    assert_allclose(g.function(4), 0.3)
+
+def test_util_gamma_getset():
+    g1 = Lorentzian()
+    g1.gamma.value = 1.0
```
francisco-dlp: 1.0 is the default value, maybe it'll be good to set it to a different value?

francisco-dlp (commented):
Other than those 2 comments, LGTM.

hyperspy/_components/lorentzian.py
ericpre: We need the setter to avoid changing the API.

francisco-dlp: Can you provide more details?

ericpre: See https://github.com/hyperspy/hyperspy/pull/2134#issuecomment-477182152 and related discussion.
In short: `gamma` can't be used as a parameter because `sympy` already has it defined as a function. @jlaehne used `_gamma` as workaround and I suggested the property/setter to keep the same API.

hyperspy/tests/component/test_lorentzian.py
jlaehne: Good idea. Done. Added also a getset-test for fwhm.

hyperspy/_components/lorentzian.py
francisco-dlp: I see, but I am not proposing to remove the property, just the property setter because I don't see why somebody would like to replace the ``gamma`` parameter by something else.

jlaehne: Well, using the property setter is a possible way to set starting parameters prior to fitting.

francisco-dlp: I think that it is not necessary, because the setter is to set the ``gamma`` attribute itself. To set the starting parameters you use the ``value`` attribute of the ``gamma`` ``Parameter``.

ericpre: > I see, but I am not proposing to remove the property, just the property setter because I don't see why somebody would like to replace the `gamma` parameter by something else.

Yes, I misread your first comment. In this case, they should not be a need to set `gamma` since during the initialisation of the component, `_gamma` is being set!

jlaehne: OK, your right. I tested again and the `@property` alone does the job, the setter is superfluous, so I removed it.

ericpre (approved):

francisco-dlp:
This is not yet ready, there are 3 failing tests.

jlaehne:
Can you be more specific? All the tests in `test_lorentzian.py` run without error on my system. I also ran some code adding noise to a Lorentzian and then doing a fit to it.

francisco-dlp:
See e.g. https://travis-ci.org/hyperspy/hyperspy/jobs/533209673#L1111

hyperspy/_components/lorentzian.py
```diff
@@ -16,76 +16,79 @@
...
+        return self._gamma.value * 2
+
+    @fwhm.setter
+    def fwhm(self, value):
+        self._gamma.value = value / 2
+
+    @property
+    def gamma(self):
+        return self._gamma
+
```
ericpre, line 90: Add property to access the grad of `_gamma` which is made in `compile_function`.

```suggestion

    @property
    def grad_gamma(self):
        return self.grad__gamma
```

ericpre:
Related to the current tests failure, I am not sure if this is a bug or just initial parameter values issue, the following works fine:

```python
from hyperspy.signals import Signal1D
from hyperspy.components1d import Lorentzian

l = Lorentzian(centre=50, gamma=5)
s = Signal1D(l.function(np.arange(100)))
m = s.create_model()
m.append(Lorentzian())

m[0].centre.value = 1
m.fit()
m.plot()
```

while changing the initial value of centre from 1 to 0 doesn't fit anymore:
```python
l = Lorentzian(centre=50, gamma=5)
s = Signal1D(l.function(np.arange(100)))
m = s.create_model()
m.append(Lorentzian())

m[0].centre.value = 0
m.fit()
m.plot()
```

hyperspy/_components/lorentzian.py
jlaehne: done

jlaehne:
@francisco-dlp: I removed the line `self._gamma.name = 'gamma'`, which solved the problem with the failing `test_model_selection_criteria` tests in my local notebook.

jlaehne:
@ericpre: Could be a bug, as it does fit fine with the old Lorentzian component even for `m[0].centre.value = 0`.

However, you can replace `Lorentzian` by either `Gaussian` or `GaussianHF` and you have the same problem. If the centre is far from `0`, but you use this starting value, it fails to fit. Thus, it looks it would be a more general bug of `expression` and not of this specific component.

jlaehne:
... but now `test_samfire` fails

francisco-dlp:
I have just sent you a pull request (https://github.com/jlaehne/hyperspy/pull/1) that uses a different approach (see #2173) to workaround the ``gamma`` issue.

ericpre:
@jlaehne could you please check and merge https://github.com/jlaehne/hyperspy/pull/1?  Once https://github.com/jlaehne/hyperspy/pull/1 is merged, this one can be merge.

jlaehne:
Indeed, looks like a better solution. Tests worked locally.

However, the problem with `m[0].centre.value = 0` (or even a value close to 0) pointed out by @eripre persists. In fact, it now not only leads to a failed fit, but to an error message:
```
RuntimeWarning                            Traceback (most recent call last)
<ipython-input-36-ee95f1ce2384> in <module>
      6 
      7 m[0].centre.value = 0
----> 8 m.fit()
      9 m.plot()

~/medien/pdi/git/hyperspy/hyperspy/model.py in fit(self, fitter, method, grad, bounded, ext_bounding, update_plot, **kwargs)
   1128                              (len(args[0]) - len(self.p0)))
   1129 
-> 1130                     self.p_std = np.sqrt(np.diag(pcov))
   1131                 self.fit_output = output
   1132 

RuntimeWarning: invalid value encountered in sqrt
```

ericpre:
> However, the problem with m[0].centre.value = 0 (or even a value close to 0) pointed out by @eripre persists. In fact, it now not only leads to a failed fit, but to an error message:

This bug is not related to this PR, so I think this is good to merge now.
If setting the initial parameters helps to workaround this issue, it may be possible to add an `estimate_parameters` method to the `Lorentzian` component, similarly as it is done for the `Gaussian` one.

PR #2136, opened 2019-03-06
Rewritten Logistic component as Expression.

jlaehne:
### Description of the change
Logistic component rewritten as Expression. Extended docstring.

I don't really see the point of the b parameter in this definition of the logistic function, but did not want to touch the definition of the function.

### Progress of the PR
- [X] ready for review.

### DIFF
--- a/hyperspy/_components/logistic.py
+++ b/hyperspy/_components/logistic.py
@@ -26,12 +26,12 @@ class Logistic(Expression):
 
     .. math::
     
-        f(x) = a / (1 + b * exp(-c * (x - origin)))
+        f(x) = \\frac{a}{1 + b \\mathrm{exp}\\left[-c \left((x - origin\\right)\\right]}
 
     Parameters
     -----------
     a : Float
-        The curve's maximum y-value, lim(x->infty)=a
+        The curve's maximum y-value, \\mathrm{lim}(x\\to\\infty) = a
     b : Float
         Additional parameter: 
         b>1 shifts origin to larger values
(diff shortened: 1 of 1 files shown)


hyperspy/_components/logistic.py
```diff
@@ -17,87 +17,58 @@
...
+            b=b,
+            c=c,
+            origin=origin,
+            position="origin",
+            module=module,
+            autodoc=False,
+            **kwargs)
+
+        # Boundaries
+        self.a.bmin = None
```
ericpre: Is there any reason to set explicitly `bmin`, `bmax` to `None` for the different parameters? `None` should be already the default?

hyperspy/_components/logistic.py
```diff
@@ -17,87 +17,58 @@
...
 
-class Logistic(Component):
+class Logistic(Expression):
 
-    """Logistic function component
+    """Logistic function (sigmoid/s-shaped curve) component
 
-    f(x) = a/(1+b*exp(-c*(x-origin)))
+    .. math::
+
```
ericpre, line 28: Use latex to display nicely the formula (see for example https://github.com/hyperspy/hyperspy/blob/RELEASE_next_minor/hyperspy/_components/lorentzian.py#L30)?

ericpre (changes requested):
A few comments.

jlaehne:
Removed the unneccessary default parameter settings.

jlaehne:
Thanks Eric, I implemented the proposed changes.

hyperspy/_components/logistic.py
jlaehne: No, so removed.

hyperspy/_components/logistic.py
jlaehne: Good idea. I also improved the docstring further for consistency.

PR #2137, opened 2019-03-06
Rewritten Bleasdale componenet as Expression

jlaehne:
### Description of the change
Changed component Bleasdale to be written as expression. Uses `numpy` as `where` function does not work using `numexpr` (although it should according to the docs).

### Progress of the PR
- [X] Added test function,
- [X] ready for review.


hyperspy/_components/bleasdale.py
```diff
@@ -18,62 +18,34 @@
 
 import numpy as np
```
ericpre, line 19: This import is not necessary anymore.

hyperspy/_components/bleasdale.py
```diff
@@ -18,62 +18,34 @@
...
-
-class Bleasdale(Component):
+class Bleasdale(Expression):
 
     """Bleasdale function component.
 
-    f(x) = (a+b*x)^(-1/c)
+    .. math::
+    
+        f(x) = (a+b*x)^(-1/c)
```
ericpre: Typo: `(a+b*x)^{(-1/c)}`

hyperspy/_components/bleasdale.py
```diff
@@ -18,62 +18,34 @@
...
 import numpy as np
 
-from hyperspy.component import Component, Parameter
+from hyperspy._components.expression import Expression
 
-
-class Bleasdale(Component):
+class Bleasdale(Expression):
 
     """Bleasdale function component.
```
ericpre: Add "also called the Bleasdale-Nelder function"?

hyperspy/_components/bleasdale.py
```diff
@@ -18,62 +18,34 @@
...
+    
+        f(x) = (a+b*x)^(-1/c)
 
     Attributes
     ----------
     a : Float
     b : Float
     c : Float
+    
+    For (a+b*x)<=0, the component will be set to `nan`.
```
ericpre: Is it going to set to `nan` or to 0?

ericpre (changes requested):
Not sure where does the sympy error comes from when using `where` but `numexpr` support the numpy `where` function, so in principle it should work.

```python
import numpy as np
import numexpr as ne

a = np.linspace(-10, 10, 100)

ne.evaluate('where(a<0, 0, a)')
```
Maybe a sympy bug?

I would suggest to use numexpr by using the same approach as in the [PowerLaw](https://github.com/hyperspy/hyperspy/blob/RELEASE_next_minor/hyperspy/_components/power_law.py#L66) component.

jlaehne:
Numexpr works definitely as you say. So it has to be a sympy bug (see below).

I have updated with `numexpr` following the example of `PowerLaw`. However, this sets the values for which the function is not defined to `nan`, even if I explicitly write `0` in the `where` statement. 

The implementation with `numpy` instead of `numexpr` correctly sets the value to `0`. In general, setting nan should not be a problem, but for backwards compatibility it might be preferable to use the implementation with `numpy`? What do you think @ericpre

This decision would also influence the rewriting of several other components.

hyperspy/_components/bleasdale.py
jlaehne: see comment below

jlaehne:
An further alternative is setting the module to 'scipy'. This solution allows for the where statement to be integrated in the expression instead of in an extra function. Also it correctly interprets the where statement and sets not defined values to zero (though it then returns a complex number).

jlaehne:
This is the traceback, when trying to evaluate the expression with numexpr (without the extra function workaround). It seems to be `sympy.lambdify` where the problem originates.

```

  File "/usr/lib/python3.7/site-packages/IPython/core/interactiveshell.py", line 3291, in run_code

  File "<ipython-input-2-8a3f2683928c>", line 1, in <module>
    g = Bleasdale()

  File "/home/jonas/medien/pdi/git/hyperspy/hyperspy/_components/bleasdale.py", line 59, in __init__
    **kwargs)

  File "/home/jonas/medien/pdi/git/hyperspy/hyperspy/_components/expression.py", line 144, in __init__
    self.compile_function(module=module, position=position)

  File "/home/jonas/medien/pdi/git/hyperspy/hyperspy/_components/expression.py", line 234, in compile_function
    dummify=False)

  File "/usr/lib/python3.7/site-packages/sympy/utilities/lambdify.py", line 773, in lambdify
    c = compile(funcstr, filename, 'exec')

  File "<lambdifygenerated-2>", line 2
    return (evaluate('  # Not supported in Python:
                                                  ^
SyntaxError: EOL while scanning string literal
```

jlaehne:
Looking closer, above error with `numexpr` occurs when creating the gradients by using `lambdify` on the differentiated expression.

Lambdify works fine on a `numexpr`, even when the `where` statement is included:
```
import numexpr as ne
from sympy import lambdify,symbols,diff
x = symbols('x')
expr = "where(x>0,x,0)"
f = lambdify(x, expr, 'numexpr')
a = np.arange(-2,10)
f(a)
```

However, when you differentiate the expression and lambdify, you get the Syntax error:
```
grad_expr = diff(expr,x)
df = lambdify(x, grad_expr, 'numexpr')
```

Using `numpy` instead of `numexpr` does not raise this error, but calling the gradient function `df` afterwards raises an error.

jlaehne:
However, the workaround using an extra `def function` borrowed from the `PowerLaw` component does not work either, as the following minimal example shows:

```
import numpy as np
from hyperspy.components1d import Expression

class Test(Expression):
    def __init__(self, module="numexpr", **kwargs):
        super(Test, self).__init__(
            expression="x",
            name="Test",
            module=module,
            autodoc=False,
            **kwargs)        
    def function(self,x):
       return np.where(x>0,super().function(x),0)

g = Test()
x = np.arange(-3,10)
g.function(x)
```

The `where` function is not evaluated. Basically, it is as if you implement the component without a where statement. For this reason, a function that is not defined over the whole range or that goes to infinity returns `nan` for that range.

jlaehne:
Once #2186 is merged, this should be ready to merge as well.

jlaehne:
Adapted according to #2186, ready for review.

hyperspy/_components/bleasdale.py
```diff
@@ -42,18 +42,47 @@ class Bleasdale(Expression):
...
         
     
-    For :math:`(a+b\cdot x)\leq0`, the component will be set to nan.
-
+    For :math:`(a+b\cdot x)\leq0`, the component will be set to 0.
     """
     
-    def __init__(self, a=1., b=1., c=1., module="numexpr", **kwargs):
+    def __init__(self, a=1., b=1., c=1., module="numexpr",
+                 compute_gradients=False, **kwargs):
```
francisco-dlp: Since sympy cannot compute the gradients in this case, wouldn't it be better not to provide the option to enable them here?

francisco-dlp:
Except for the small comment, it looks good to me. However, now that you are at it, would it be possible to fix the gradients? They were wrong in the original implementation because they should be 0  to the left of the cutoff value.

hyperspy/_components/bleasdale.py
jlaehne: done

jlaehne:
Added where statements to the gradients and removed option to set compute_gradients.

jlaehne:
Gradient functions still raise errors for certain values, e.g. `g.grad_a(-0.5)` for `a,b,c = 1,2,2`. However, he should not even compute the function for these values due to the where statement. Could that be an upstream bug of np.where?

francisco-dlp:
No, it's not an upstream bug, it's a HyperSpy bug that affects all components that use ``where``. The first argument of ``where`` must be an array. Good catch!

The following works:
```python
>>> g.grad_a(np.asarray(-0.5))
array(0.)
```

francisco-dlp:
@ericpre, this LGTM. If you agree, feel free to merge.

ericpre (approved):

PR #2138, opened 2019-03-06
Rewritten RC component as Expression

jlaehne:
### Description of the change
RC component rewritten as Expression. Extended docstring.

### Progress of the PR
- [x] Add test function
- [x] ready for review.

### DIFF
--- a/hyperspy/_components/rc.py
+++ b/hyperspy/_components/rc.py
@@ -29,13 +29,16 @@ class RC(Expression):
         f(x) = V_\mathrm{0} + V_\mathrm{max} \left[1 - \mathrm{exp}\left(
             -\frac{x}{\tau}\right)\right]
     
-    ====================== ===========
-    Variable                Parameter 
-    ====================== ===========
-    :math:`V_\mathrm{0}`    V0     
-    :math:`V_\mathrm{max}`  Vmax    
-    :math:`\tau`            tau    
-    ====================== ===========
+    +-----------------------+-----------+
+    | Parameter             | Attribute |
+    +-----------------------+-----------+
+    +-----------------------+-----------+
+    | :math:`V_\mathrm{0}`  |  V0       |
+    +-----------------------+-----------+
+    | :math:`V_\mathrm{max}`|  Vmax     |
+    +-----------------------+-----------+
+    | :math:`\tau`          |  tau      |
+    +-----------------------+-----------+
     
     
     Parameters
@@ -64,4 +67,4 @@ class RC(Expression):
             **kwargs,
         )
         
-        self.isbackground = True
+        self.isbackground = False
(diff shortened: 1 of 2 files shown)


hyperspy/_components/rc.py
```diff
@@ -16,24 +16,38 @@
...
     """
-    """
+    RC function component (based on the time-domain capacitor voltage response of an RC-circuit)
+    
+    .. math::
 
-    def __init__(self, V=1, V0=0, tau=1.):
-        Component.__init__(self, ('Vmax', 'V0', 'tau'))
-        self.Vmax.value, self.V0.value, self.tau.value = V, V0, tau
+        f(x) = V0 + Vmax * (1 - exp(-x / \tau))
```
ericpre: Use latex to display nicely the formula (see for example https://github.com/hyperspy/hyperspy/blob/RELEASE_next_minor/hyperspy/_components/lorentzian.py#L30)?

Add a parameter/attribute mapping table as in https://github.com/hyperspy/hyperspy/blob/RELEASE_next_minor/hyperspy/_components/lorentzian.py#L32 to make it easier to read?

ericpre (changes requested):
Other than the comments below, this looks good to me.

jlaehne:
Added LaTeX formatting and parameter table.

hyperspy/_components/rc.py
jlaehne: done. see #2148 for discussion on table format

hyperspy/_components/rc.py
```diff
@@ -16,24 +16,52 @@
...
+    
+    .. math::
 
-class RC(Component):
-
-    """
+        f(x) = V_\mathrm{0} + V_\mathrm{max} \left[1 - \mathrm{exp}\left(
+            -\frac{x}{\tau}\right)\right]
+    
+    ====================== ===========
```
ericpre: Please use the internal convention for the table for now and use #2148 to discuss these changes.

hyperspy/_components/rc.py
```diff
@@ -16,24 +16,52 @@
...
+            name="RC",
+            Vmax=Vmax,
+            V0=V0,
+            tau=tau,
+            module=module,
+            autodoc=False,
+            **kwargs,
+        )
+        
+        self.isbackground = True
```
ericpre: Out of curiosity, what is the reason to consider this component as a background?

hyperspy/_components/rc.py
jlaehne: done

hyperspy/_components/rc.py
jlaehne: Good point, must have sneaked in from a copy-paste. Does not make much sense for an RC function (whereas it would indeed make sense for the skew-normal).

ericpre: This is only used for EDS and EELS model, but I guess this is better to remove it completely.

jlaehne: Now set to False.

ericpre (approved):

Issue #2141, opened 2019-03-15
Update plot after align_zero_loss_peak


magnunor:
If a plot is open, and `s.align_zero_loss_peak` is called, the plot is not updated with the align spectrum. To update it, `s.plot()` has be called again manually.

Example:
```python
import numpy as np
import hyperspy.api as hs
g = hs.model.components1D.Gaussian(A=1, sigma=2, centre=20)
s = hs.signals.EELSSpectrum(g.function(np.arange(100)))
s.axes_manager.signal_axes[0].offset = -10
s.plot()
s.align_zero_loss_peak() # Plot does not update
s.plot() # Plot updates
```

Since many of the signal methods automatically update the plot, `align_zero_loss_peak` not doing this can lead to confusion, with the user thinking that the method didn't work correctly.

Ideally `align_zero_loss_peak` should update plots associated with it.

Before replot:
![align_zlp_before_replot](https://user-images.githubusercontent.com/1690979/54457906-8d042a80-475a-11e9-939e-a054b61ae884.jpg)

After replot:
![align_zlp_after_replot](https://user-images.githubusercontent.com/1690979/54457920-942b3880-475a-11e9-8163-b4e6fbe086f5.jpg)

ericpre:
#2144 is merged.

PR #2144, opened 2019-03-16
Open plots will now be updated when using s.align_zero_loss_peak

magnunor:
Previously, when using align_zero_loss_peak with the plot of the
signal open, it would not update. This could lead to the user
thinking that the function wasn't working correctly.

This change also updates other signals specified with the
also_align parameter in align_zero_loss_peak.

To test this:

```python
import numpy as np
import hyperspy.api as hs
g = hs.model.components1D.Gaussian(A=1, sigma=2, centre=20)
s = hs.signals.EELSSpectrum(g.function(np.arange(100)))
s.axes_manager.signal_axes[0].offset = -10
s1 = hs.signals.EELSSpectrum(np.arange(100))

s.plot()
s1.plot()
s.align_zero_loss_peak(also_align=[s1]) # Without this fix: it does not update the plots. With this fix: both plots update.
```

https://github.com/hyperspy/hyperspy/issues/2141

### DIFF
--- a/hyperspy/_signals/eels.py
+++ b/hyperspy/_signals/eels.py
@@ -277,6 +277,7 @@ class EELSSpectrum_mixin:
                 value = value.compute()
             for signal in signals:
                 signal.axes_manager[-1].offset -= value
+                signal.events.data_changed.trigger(signal)
 
         def estimate_zero_loss_peak_centre(s, mask, signal_range):
             if signal_range:


magnunor:
appveyor failing for some unrelated reason (not finding 'conda'?)

ericpre:
> appveyor failing for some unrelated reason (not finding 'conda'?)

The conda issue was fixed in #2128.

Issue #2142, opened 2019-03-15
Lazy compute return signal type


magnunor:
[pixStem](https://pixstem.org/) implements much of its functionality by creating two new classes, `PixelatedStem` and `LazyPixelatedStem`, which inherit either `Signal2D` or `LazySignal2D`. Most the inherited functions return the same signal type, which makes this approach of extending HyperSpy very practical.

However, `s.compute()` returns only HyperSpy signals, not any of the extended ones in `pixStem`. Example of this:

```python
import dask.array as da
import hyperspy.api as hs
from hyperspy._signals.signal2d import LazySignal2D, Signal2D

class LazyTestSignal(LazySignal2D):

    pass

s = LazyTestSignal(da.random.random((100, 100)))
print(s) # <LazyTestSignal, title: , dimensions: (|100, 100)>
s.compute()
print(s) # <Signal2D, title: , dimensions: (|100, 100)>
```

It would be nice to be able to specify which type of signal class should be returned, in the signal itself. This could be made by overloading the `compute` (and `as_lazy`) methods. For example:

```python
import dask.array as da
import hyperspy.api as hs
from hyperspy._signals.signal2d import LazySignal2D, Signal2D

class LazyTestSignal(LazySignal2D):

    def __init__(self, *args, **kwargs):
        self._non_lazy_version = TestSignal
        super().__init__(*args, **kwargs)

    def compute(self, *args, **kwargs):
        super().compute(*args, **kwargs)
        self.__class__ = self._non_lazy_version
        self.__init__(**self._to_dictionary())


class TestSignal(Signal2D):

    def __init__(self, *args, **kwargs):
        self._lazy_version = LazyTestSignal
        super().__init__(*args, **kwargs)

    def as_lazy(self, *args, **kwargs):
        res = super().as_lazy(*args, **kwargs)
        res.__class__ = self._lazy_version
        res.__init__(**res._to_dictionary())
        return res

s = LazyTestSignal(da.random.random((100, 100)))
print(s) # <LazyTestSignal, title: , dimensions: (|100, 100)>
s.compute()
print(s) # <TestSignal, title: , dimensions: (|100, 100)>
s1 = s.as_lazy()
print(s1) # <LazyTestSignal, title: , dimensions: (|100, 100)>
```

However, it would be better if HyperSpy had support for this directly, for example by the `compute` method checking for a `self._non_lazy_version` attribute, and if it exist use that class instead of one found by `_assign_subclass`.

[...]

PR #2145, opened 2019-03-16
Improve compute and as_lazy to make extending HyperSpy easier

magnunor:
pixStem implements much of its functionality by creating two new classes, `PixelatedStem` and `LazyPixelatedStem`, which inherit either `Signal2D` or `LazySignal2D`. Most the inherited functions return the same signal type, which makes this approach of extending HyperSpy very practical.

However, s.compute() returns only HyperSpy signals, not any of the extended ones in pixStem. This pull request adds new attributes to `BaseSignal` and `LazySignal`, adds a new parameter to `s._assign_subclass`, so that both `s.compute` and `s.as_lazy` can specify which signal class the output signal from these functions should be.

So currently this happens:
```python
import dask.array as da
import hyperspy.api as hs
from hyperspy._signals.signal2d import LazySignal2D, Signal2D

class LazyTestSignal(LazySignal2D):

    pass

s = LazyTestSignal(da.random.random((100, 100)))
print(s) # <LazyTestSignal, title: , dimensions: (|100, 100)>
s.compute()
print(s) # <Signal2D, title: , dimensions: (|100, 100)>
```

With the new implementation
```python
import dask.array as da
import hyperspy.api as hs
from hyperspy._signals.signal2d import LazySignal2D, Signal2D

class LazyTestSignal(LazySignal2D):

    def __init__(self, *args, **kwargs):
        self._non_lazy_signal_class = TestSignal
        super().__init__(*args, **kwargs)


class TestSignal(Signal2D):

    def __init__(self, *args, **kwargs):
        self._lazy_signal_class = LazyTestSignal
        super().__init__(*args, **kwargs)

s = LazyTestSignal(da.random.random((100, 100)))
print(s) # <LazyTestSignal, title: , dimensions: (|100, 100)>
s.compute()
print(s) # <TestSignal, title: , dimensions: (|100, 100)>
s1 = s.as_lazy()
print(s1) # <LazyTestSignal, title: , dimensions: (|100, 100)>
```

## Todos

- [ ] Finish implementation
- [ ] Write tests
- [ ] Add info to developer guide about this

https://github.com/hyperspy/hyperspy/issues/2142

### DIFF
--- a/hyperspy/_signals/lazy.py
+++ b/hyperspy/_signals/lazy.py
@@ -82,6 +82,7 @@ class LazySignal(BaseSignal):
     (assuming storing the full result of computation in memory is not feasible)
     """
     _lazy = True
+    _non_lazy_signal_class = None
 
     def compute(self, progressbar=True, close_file=False):
         """Attempt to store the full signal in memory.
@@ -103,7 +104,8 @@ class LazySignal(BaseSignal):
                 self.close_file()
             self.data = data
         self._lazy = False
-        self._assign_subclass()
+        self._assign_subclass(
+                signal_class=self._non_lazy_signal_class)
 
     def close_file(self):
         """Closes the associated data file if any.
--- a/hyperspy/signal.py
+++ b/hyperspy/signal.py
@@ -1560,6 +1560,7 @@ class BaseSignal(FancySlicing,
     _signal_dimension = -1
     _signal_type = ""
     _lazy = False
+    _lazy_signal_class = None
     _alias_signal_types = []
     _additional_slicing_targets = [
         "metadata.Signal.Noise_properties.variance",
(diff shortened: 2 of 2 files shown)


magnunor:
I'm guessing the question here is if adding the private variables `_non_lazy_signal_class` and `_lazy_signal_class` is the way to go, or if there are some better ways of "telling" HyperSpy which class to return.

ericpre:
Does it work if the signal changes from `LazySignal2D` to `Signal1D` during the `compute` step?

One better way would be to use `_signal_type` attribute of your object and when a lazy signal "is computed", it should create an instance of the correct class automatically. This would need to have a (dynamic) registry of signals for the different libraries, which is not implemented at the moment...
This is a discussion for the split really. I copy a link to the setuptools documentation, because I find it relevant to design a registry:
https://setuptools.readthedocs.io/en/latest/setuptools.html#dynamic-discovery-of-services-and-plugins

In the mean time, one better way to do would be to override the `compute` method in your `LazyTestSignal` and set the appropriate class type there.

magnunor:
I ended up overloading the `compute` and `as_lazy` functions: https://gitlab.com/pixstem/pixstem/commit/4eb92904e13cec407be337cbdcdf39c82bc903df

ericpre:
Implemented in #2180.

PR #2148, opened 2019-03-27
Component Exponential, GaussianHF, Gaussian and PowerLaw documentation

jlaehne:
### Description of the change
Docstring for exponential component formatted using latex for math and a parameter table 
(using simple table layout, which works best in conjunction with latex) to create consistency among components.

Also some corrections in the docstring for GaussianHF (typo, parameter table, extra latex).

Added test functions where missing.

### Progress of the PR
- [X] Add test function for exponential, gaussian and power_law,
- [X] ready for review.

### DIFF
--- a/hyperspy/_components/exponential.py
+++ b/hyperspy/_components/exponential.py
@@ -27,7 +27,7 @@ class Exponential(Expression):
         f(x) = A\\cdot\\exp\\left(-\\frac{x}{\\tau}\\right)
 
     ================= ===========
-      Parameter        Attribute 
+     Variable          Parameter 
     ================= ===========
      :math:`A`         A     
      :math:`\\tau`     tau    
--- a/hyperspy/_components/exponential.py
+++ b/hyperspy/_components/exponential.py
@@ -33,14 +33,15 @@ class Exponential(Expression):
     :math:`\tau`   tau    
     ============= =============
 
+    
     Parameters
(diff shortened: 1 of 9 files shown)


jlaehne:
@ericpre; OK, lets discuss the formatting of the docstring at this point (and not in #2134).

My original motivation for the simple table (http://www.sphinx-doc.org/en/master/usage/restructuredtext/basics.html) was that I was not using the raw string mode (r"""...) and then sphinx could not handle the math symbols in the grid table. However, I also find it looks better to have a table with the column headers in bold face instead of having a whole extra line offsetting the header from the body of the table. On top, the simple table layout takes less space in the code. So I would argue to go for the "simple table".

I guess it makes sense to use raw string mode consistently in all docstrings? So far it was also mixed (even when using math mode). Is there an overall convention for hyerspy?

Additionally, for the table header, I would use the convention "Variable" and "Parameter", instead of "Parameter" and "Attribute". Most other docstrings use "Parameters" for the detailed specs on what parameters the function takes and it is confusing if this is called "Attribute" in the table.

Any other comments on the general format for component docstrings as implemented in this PR? For example, should a `Returns` section be implemented?

thomasaarholt (approved):
LGTM

francisco-dlp:
Nice, thanks!

PR #2149, opened 2019-03-27
Component Logistic documentation + test

jlaehne:
### Description of the change
Docstring made consistent with other expression components.

### Progress of the PR
- [x] Add test function,
- [x] ready for review.


thomasaarholt:
Looks good to me!

jlaehne:
Docstring now as proposed in #2148

hyperspy/_components/logistic.py
```diff
@@ -22,27 +22,44 @@
...
 class Logistic(Expression):
 
-    """Logistic function (sigmoid or s-shaped curve) component.
+    r"""Logistic function (sigmoid or s-shaped curve) component.
 
     .. math::
     
-        f(x) = \\frac{a}{1 + b\\cdot \\mathrm{exp}\\left[-c \left(x - origin\\right)\\right]}
+        f(x) = \frac{a}{1 + b\cdot \mathrm{exp}\left[-c 
+            \left((x - x_0\right)\right]}
```
ericpre, line 29: I don't see any benefit to introduce `x_0` here and find it rather confusing?

jlaehne: Well, also GaussianHF uses `h`, `c`, `W` in the math formula for example to avoid full parameter names, which I think looks nicer in the math display in the end. We can of course call this also `o`, but I usually find the convention of `x-x_0` for the center/origin of distribution functions quite sensible.

PR #2150, opened 2019-03-29
Add tests to check creation of 1D components and fix a few bugs.

ericpre:
### Progress of the PR
- [x] Add tests to check the creation of 1D components,
- [x] fix a few issues,
- [x] ready for review.

### DIFF
--- a/hyperspy/tests/component/test_components.py
+++ b/hyperspy/tests/component/test_components.py
@@ -24,9 +24,37 @@ import pytest
 import hyperspy.api as hs
 from hyperspy.models.model1d import Model1D
 from hyperspy.misc.test_utils import ignore_warning
+from hyperspy import components1d
 
 
 TRUE_FALSE_2_TUPLE = [p for p in itertools.product((True, False), repeat=2)]
+
+
+def get_components1d_list():
+    components1d_list = [c_name for c_name in
+                         dir(components1d) if '_' not in c_name]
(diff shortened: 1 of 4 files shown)


hyperspy/tests/component/test_components.py
```diff
@@ -24,11 +24,39 @@
...
 from hyperspy.misc.test_utils import ignore_warning
+from hyperspy import components1d
 
 
 TRUE_FALSE_2_TUPLE = [p for p in itertools.product((True, False), repeat=2)]
 
 
+def get_components1d_list():
+    components1d_list = [c_name for c_name in
+                         dir(components1d) if '_' not in c_name]
```
thomasaarholt: The following will still do all tests if a component name ends up containing an underscore:
```suggestion
                         dir(components1d) if not cname.startswith('_')]
```

thomasaarholt (commented):
Only one comment, otherwise LGTM!

hyperspy/tests/component/test_components.py
francisco-dlp: @ericpre, are you planning to take into account @thomasaarholt comment?

ericpre: Following the current convention, the name of the components should not contain underscore. Anyway this was not a good way to do it so I changed it to be more reliable.

PR #2154, opened 2019-04-17
enhancement/pca estimate of no of signifcant components

pquinn-dls:
I've added a method to estimate the no of signficant components in PCA - enhancement on the current threshold method.
This performs a simple measurements of where the elbow position is and stores it. 
This calculation is performed in the decomposition method and when using plot_explained_variance_ratio  setting vline=True will plot the 
vertical line for this estimated position. 

screeplot3

A property no_significant_components has also been added to LearningResults. Storing the value is to help with automated scripts and later cluster analysis commits.
Small change also added to user guide


jat255:
@pquinn-dls thanks for the contribution! At first glance, this looks pretty good, and very useful! Is there a reference that you could include in the documentation to using this method? It looks like a fairly obvious method once I read the comments in the code, but I hadn't seen it used before. 

A couple of requests:
- [x] Can you please add a few test cases and ensure that they pass? You can look in the `test` folder for examples using the `pytest` framework. I would create a few dummy signals with a known number of significant components, and then ensure that the method returns the expected result.
- [x] If it's not too much trouble, it might be nice to have the "triangle" visualized at least in the documentation to make it clear how the point is determined. Here's a horrible quick example of what I mean using a screenshot tool:
   - image
- [x] It seems to me that in your example, the 4th component is the furthest from the "triangle line", but the `vline` appears to be drawn through the 5th component. Is this an error, or am I just not understanding how the method works correctly?

pquinn-dls:
Hi,

Yes you're correct the elbow is the 4th but the no of significant pca components was set to elbow + 1
I realised that I was plotting the no of components - not the index so I've adjusted that. 

I've added a reference, some additional docs and an example image of the method.

Ville Satopää, Jeannie Albrecht, David Irwin, Barath Raghavan. Finding a "Kneedle" in a Haystack: Detecting Knee Points in System Behavior. 31st International Conference on Distributed Computing Systems Workshops, pp. 166-171, Minneapolis, Minnesota, USA, June 2011.

I've also added a test but I'm not sure if I've done this correctly - need to read a bit more on this!

doc/user_guide/mva.rst
```diff
@@ -104,8 +104,16 @@ the total variance specified, and the components above this value will be
 styled distinctly from the remaining components to show which are considered
 signal, as opposed to noise. Alternatively, by providing an integer value
 for ``threshold``, the line will be drawn at the specified component (see
-below).  The "elbow" position can be auto-estimated and a vertical line drawn to
-represent the no of signicant pca components by specifying "vline" as True. 
+below).  The no of significant components can be estimated and a vertical line drawn to
```
jat255: `no` --> `number`

doc/user_guide/mva.rst
```diff
@@ -104,8 +104,16 @@ the total variance specified, and the components above this value will be
...
+represent this by specifying "vline" as True. In this case the elbow or knee is found
```
jat255: "vline" --> \`\`vline\`\`

jat255: "In this case the elbow" --> "In this case, the elbow"

doc/user_guide/mva.rst
```diff
@@ -104,8 +104,16 @@ the total variance specified, and the components above this value will be
 styled distinctly from the remaining components to show which are considered
 signal, as opposed to noise. Alternatively, by providing an integer value
 for ``threshold``, the line will be drawn at the specified component (see
-below).  The "elbow" position can be auto-estimated and a vertical line drawn to
-represent the no of signicant pca components by specifying "vline" as True. 
+below).  The no of significant components can be estimated and a vertical line drawn to
+represent this by specifying "vline" as True. In this case the elbow or knee is found 
+in the variance plot by estimating the distance from each point in the variance plot 
+to a line joining the first and last points of the plot.  The no of significant PCA components
```
jat255: "no" --> "number"

jat255 (changes requested):

doc/user_guide/mva.rst
pquinn-dls: changed

doc/user_guide/mva.rst
pquinn-dls: fixed

doc/user_guide/mva.rst
pquinn-dls: fixed

doc/user_guide/mva.rst
```diff
@@ -104,11 +104,11 @@ the total variance specified, and the components above this value will be
 styled distinctly from the remaining components to show which are considered
 signal, as opposed to noise. Alternatively, by providing an integer value
 for ``threshold``, the line will be drawn at the specified component (see
-below).  The no of significant components can be estimated and a vertical line drawn to
-represent this by specifying "vline" as True. In this case the elbow or knee is found 
+below).  The number of significant components can be estimated and a vertical line drawn to
+represent this by specifying ''vline'' as True. In this case, the elbow or knee is found
```
jat255: Sorry, can you change this to use backticks (\`) instead of single quotes (like \`\`threshold\`\` above it)? That will cause the value to be formatted in monospace in the generated documentation.

jat255 (changes requested):

jat255:
Also, there were some merges into `RELEASE_next_minor`  yesterday. Can you please rebase your branch onto an updated copy of `RELEASE_next_minor`?

pquinn-dls:
Rebased to RELEASE_next_minor and fixed the vline documentation.

jat255:
I made some suggested changes in https://github.com/pquinn-dls/hyperspy/pull/2, and it looks like @ericpre did so as well in https://github.com/pquinn-dls/hyperspy/pull/1. Can you please review these and incorporate them onto your branch? After that I think this PR will be good to go.

hyperspy/learn/mva.py
```diff
@@ -1222,13 +1248,54 @@ def undo_treatments(self):
...
+        # The elbow should be the point which is the
+        # furthest distance from this line
+        #
+        y2 = np.log(error_estimate[maxpoints])
+        x2 = maxpoints
+        y1 = np.log(error_estimate[0])
+        x1 = 0
+        # loop 
+        distance = np.zeros((maxpoints))
+        for i in range(maxpoints):
```
jat255, line 1283: I'll add here that it would be better to do this with a vector calculation (https://stackoverflow.com/a/46470401/1435788), rather than looping through the array, but since it's such a small one-dimensional array, there'e probably no harm in leaving it as a loop.

doc/user_guide/mva.rst
```diff
@@ -104,7 +104,21 @@ the total variance specified, and the components above this value will be
...
+the elbow or knee is found in the variance plot by estimating the distance from
+each point in the variance plot to a line joining the first and last points of
+the plot.  As the index starts at zero, the number of significant PCA components
+is the elbow index position + 1.
+
+.. figure::  images/screeplot_elbow_method.png
+   :align:   center
+   :width:   500
+
+More details about the elbow or knee finding technique can be found in
```
tjof2, line 120: I think more details are needed here rather than just referring back to the paper:

> In this case, the elbow or knee is found in the variance plot by estimating the distance from each point in the variance plot to a line joining the first and last points of the plot.

We should be clear about e.g. choosing between multiple candidates.

pquinn-dls: Updated the description and commented on multiple candidates.

hyperspy/learn/mva.py
pquinn-dls: If it's ok I'll leave it as is for now as I think it's negligible computationally compared to the decomposition process in the majority of cases.

pquinn-dls: I've merged and edited the changes and comments from @jat255  and @ericpre

jat255:
This looks pretty good to me, but will have to wait for @ericpre or @francisco-dlp to merge, since they have write access.

pquinn-dls:
Great, thanks for your help

ericpre (commented):
It looks good to me too. I have one comment: rename `num_significant_components` to `number_significant_components` to follow [PEP8](https://www.python.org/dev/peps/pep-0008/)? 
Even if this can be regarded as a matter of taste, we usually use full words as naming conventions and using `number` instead of `num` doesn't make it significantly longer. I am so used to it now that I was mislead by the original `no_significant_components`.

ericpre:
The failures are related to the releases of upstream libraries (numpy and mrcz).

ericpre:
Thanks @pquinn-dls.

PR #2157, opened 2019-04-18
Gaussian2D component  rewritten as Expression

jlaehne:
### Description of the change
Gaussian2D function component rewritten as Expression. Extended docstring.

### Progress of the PR
- [x] Add test function
- [x] ready for review.


hyperspy/_components/gaussian2d.py
```diff
@@ -55,14 +55,13 @@ class Gaussian2D(Expression):
         Location of the Gaussian maximum (peak position) in `x` direction.
     centre_x : float
         Location of the Gaussian maximum (peak position) in `y` direction.
-    
-    
+
```
jlaehne, line 58: The double line break is needed here. Otherwise, the following comment will be formatted as if it is another Parameter!

Issue #2133, opened 2019-03-01
Opening of incomplete spectrum images


Divitini:
Hi,

I recently started encountering problems when loading .emi/.ser EDX datasets. Some datasets do not load; I suspect this is due to the datasets being incomplete scans. Traceback pasted below. As far as I can tell incomplete datasets used to be opened without issues in older versions of Hyperspy.
I am not sure whether this is a bug or one needs a flag in the load() function - can anyone have a look?

I'm using Hyperspy 1.4.1 from the Windows installer bundle (tried both with upgrade from older version and clean re-install).

Example dataset (incomplete SI that doesn't load) here:
https://drive.google.com/drive/folders/1KdRpj0I1o0LGyTqKmPgTfD4BJ9TunUYm?usp=sharing

Thanks,
Giorgio

---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
<ipython-input-10-e25301263415> in <module>
----> 1 s1 = hs.load("12.37.58 Drift corrected spectrum image.emi")

~\HyperSpy Bundle\python-3.7.0.amd64\lib\site-packages\hyperspy\io.py in load(filenames, signal_type, stack, stack_axis, new_axis_name, lazy, convert_units, **kwds)
    271             objects = [load_single_file(filename, lazy=lazy,
    272                                         **kwds)
--> 273                        for filename in filenames]
    274 
    275         if len(objects) == 1:

~\HyperSpy Bundle\python-3.7.0.amd64\lib\site-packages\hyperspy\io.py in <listcomp>(.0)
    271             objects = [load_single_file(filename, lazy=lazy,
    272                                         **kwds)
--> 273                        for filename in filenames]
    274 
    275         if len(objects) == 1:

~\HyperSpy Bundle\python-3.7.0.amd64\lib\site-packages\hyperspy\io.py in load_single_file(filename, **kwds)
    308     else:
    309         reader = io_plugins[i]
--> 310         return load_with_reader(filename=filename, reader=reader, **kwds)
    311 
    312 

~\HyperSpy Bundle\python-3.7.0.amd64\lib\site-packages\hyperspy\io.py in load_with_reader(filename, reader, signal_type, convert_units, **kwds)
    315     lazy = kwds.get('lazy', False)
    316     file_data_list = reader.file_reader(filename,
--> 317                                         **kwds)
    318     objects = []
    319 

[...]

Issue #1928, opened 2018-04-25
fei.py cant load some SER files


ercius:
Im using hyperspy 1.3.1 freshly installed from conda today and python 3.5. 

I receive this error when using hs.load() on a recently acquired data set:
```python
~\Anaconda3\envs\hyperspy3\lib\site-packages\hyperspy\io_plugins\fei.py in load_ser_file(filename)
--> 336             data[i] = np.fromfile(f,dtype=np.dtype(data_dtype_list + tag_dtype_list),count=1)
    337             f.seek(offset)
    338         _logger.info("Data info:")

ValueError: setting an array element with a sequence.
```
I did some digging in the source to find out the issue. The issue has to do with the difference between the number of elements in a series file and the __valid__ number of elements. Our TIA software always has a series size of 10 for an EELS acquisition (I have no idea why or how to fix this). So, the `data_offset_array.shape = 10` and has a valid dataset at data_offset_array[0]. Most of the other values in the data_offset_array are actually set to 0. This causes all kinds of issues.

The problem lies in fei.py starting with line 332: https://github.com/hyperspy/hyperspy/blob/79b3bf78b3b5c294feae6f23fa2ddf5fcd4e08fa/hyperspy/io_plugins/fei.py#L332
Here is the full code block:
```python
        for i, offset in enumerate(data_offset_array):
            data[i] = np.fromfile(f,
                                  dtype=np.dtype(
                                      data_dtype_list + tag_dtype_list),
                                  count=1)
            f.seek(offset)
```
This will fail because `f.seek(offset)` when offset == 0 gives the error shown at the beginning.

I added an if statement to check whether the offset is 0 and this fixed the issue.
```python
        for i, offset in enumerate(data_offset_array):
            print('i = {}, offset = {}'.format(i,offset))
            if(offset > 0):
                data[i] = np.fromfile(f,dtype=np.dtype(data_dtype_list + tag_dtype_list),count=1)
            f.seek(offset)
```
Using `header['ValidNumberElements']` and skipping non valid data is another way to do this.

thomasaarholt:
Just in case this explains why TIA is always acquiring a series in EELS:
In TIA, in the PEELS menu (you can find this in the bottom right menu list if its not visible on screen) there's a flap-out triangle. 

Open this and for the various acquisition modes (Acquire, Cumulative etc - can't remember exact names right now) you can choose the number of series and the number of passes.

[...]

PR #2160, opened 2019-04-25
Use ValidNumberElements instead of TotalNumberElements

ercius:
### Description of the change
This patch fixes loading FEI SER data where the acquisition was stopped before it was finished. SER files indicate this by the value in header['ValidNumberElements'] which was not previously used. Instead, the code used header['TotalNumberElements'] which may not match the file size. This is documented in  #1928 and some other related issues such as #2133 .

### A few sentences and/or a bulleted list to describe and motivate the change:
- Use header['ValidNumberElements'] for the dataOffsetArray size.
- Use header['ValidNumberElements'] for the data array size
- Crop series data (e.g. image series) instead of setting all values not acquired to 0 or NaN

### Progress of the PR
**Note: Need to remove printing statements, but they are useful for review.**
- [x] Add tests
- [X] Ready for review.

```python
>>> import hyperspy.api as hs
>>> from pathlib import Path
>>> imPath = Path('./hyperspy/tests/io/FEI_old/03_Scanning Preview_1.ser')
>>> im = hs.load(imPath.as_posix())
>>> print('image series data shape = {} and should be == (5, 128, 128)'.format(im.data.shape))
>>> sPath = Path('./hyperspy/tests/io/FEI_old/X - Au NP EELS_2.ser')
>>> s = hs.load(sPath.as_posix())
>>> print('spectra series data shape = {} and should be == (1, 2048)'.format(s.data.shape))
```


ericpre:
@ercius, could you please review https://github.com/ercius/hyperspy/pull/1 and merge it if you are happy with the changes?

ercius:
This looks good to me. Ill merge it.

francisco-dlp:
This looks good to me except for the new way of dealing with the missing data. Previously it kept the original dimensions and filled the array with nans or zeros depending on the dtype. Now the array is cropped and/or squeezed. I think that this can lead to issues, for example, when trying to load and stack a number of files that should have the same dimensions. If they contain invalid data because something went wrong during acquisition that stacking operations will fail. Wouldn't it be better to add this new behaviour as an option that could be ``False`` by default?

ericpre:
> This looks good to me except for the new way of dealing with the missing data. Previously it kept the original dimensions and filled the array with nans or zeros depending on the dtype. Now the array is cropped and/or squeezed. I think that this can lead to issues, for example, when trying to load and stack a number of files that should have the same dimensions. If they contain invalid data because something went wrong during acquisition that stacking operations will fail. Wouldn't it be better to add this new behaviour as an option that could be False by default?

Yes, it makes sense, this is now done with a default behaviour set to load only the valid data. I think this is the most common use case, particularly when acquiring data in "preview" mode it can save creating unnecessarily large dataset.

ercius:
I agree with this. Many users acquire data in 'preview' mode with a large
series size. There should be the ability to turn it off, but only loading
valid data should the default behavior.

francisco-dlp:
@ercius, the thing is that we cannot change the default behaviour in a patch or minor version. That's way I suggest to keep the default as it was.

ericpre:
OK, default value of `only_valid_data` is changed to False with a mention that this default value will be changed to True in 2.0.

PR #1927, opened 2018-04-24
Support for non-uniform axes

ericpre:
### Description of the change
Continuation of #438 to add support for non linear axis.

### Progress of the PR
- [x] Add basis functionality to create non linear axis,
- [x] split `DataAxis` into different class as suggested by @francisco-dlp,
- [ ] check that functionality is supported by non linear axis, raise an appropriate error when required,
- [x] Fix plotting,
- [ ] Increase the hspy version number when saving non-linear axes. HyperSpy < ? won't be able to read these files.
- [ ] tidy up,
- [ ] update user guide,
- [ ] add tests,
- [ ] ready for review.

### Minimal example of the bug fix or the new feature
```python
import hyperspy.api as hs
import numpy as np
dict0 = {'axis': np.arange(10)**5, 'name':'Axis0', 'units':'A', 'navigate':True}
dict1 = {'size': 100, 'name':'Axis1', 'units':'A', 'scale':0.2, 'offset':1, 'navigate':False}
s = hs.signals.Signal1D(np.random.random((10, 100)), axes=[dict0, dict1])
s.plot()
s.axes_manager

<Axes manager, axes: (10|100)>
            Name |   size |  index |  offset |   scale |  units 
================ | ====== | ====== | ======= | ======= | ====== 
           Axis0 |     10 |      0 |  non-linear axis  |      A 
---------------- | ------ | ------ | ------- | ------- | ------ 
           Axis1 |    100 |      0 |       1 |     0.2 |      A 

```

Plot non linear signal
```python
s2 = s.T
s2.plot()
```

Create functional axis:
```python
from hyperspy.axes import FunctionalDataAxis
expression = "a * x**2 + b"
axis = FunctionalDataAxis(size=10, expression=expression,  a=0.1, b=10)
axis.axis
# output: array([ 10. ,  10.1,  10.4,  10.9,  11.6,  12.5,  13.6,  14.9,  16.4,  18.1])
```

PR #2165, opened 2019-05-01
Adding plugin to read and write to Universal Spectroscopy and Imaging Data HDF5 files

ssomnath:
### Description of the change
- Added new io plugin for USID
- Added thorough unit tests

### Progress of the PR
- [x] no need to update docstrings
- [x] already updated user guide,
- [x] already added tests,
- [x] ready for review.

```python
>>> import hyperspy.api as hs
>>> import numpy as np
>>> # Save signal to USID compliant HDF5 file:
>>> s.save('USID.h5')
>>> # Load signal(s) from USID HDF5 files
>>> new_signal = hs.load('USID.h5')
```


doc/user_guide/io.rst
```diff
@@ -212,6 +212,8 @@ HyperSpy. The "lazy" column specifies if lazy evaluation is supported.
     +--------------------+--------+--------+--------+
     | EDAX .spc and .spd |    Yes |    No  |    Yes |
     +--------------------+--------+--------+--------+
+    | h5USID .h5         |    Yes |    No  |    No  |
```
ericpre: It seems that it can write too!

hyperspy/io_plugins/usid_hdf5.py
```diff
@@ -0,0 +1,447 @@
...
+        A HyperSpy signal
+    """
+    if not isinstance(filename, str):
+        raise TypeError('filename should be a string')
+    if os.path.exists(filename):
+        raise FileExistsError('Appending is not yet supported. A file already exists at: {}. '
+                              'Please delete the file at the location or specify a '
+                              'different path for the file'.format(filename))
+    """
+    # CANNOT import BaseSignal! Probably circular import
```
ericpre: Add it here, in this function and it should work fine.
```python
from hyperspy.signal import BaseSignal
```

hyperspy/io_plugins/usid_hdf5.py
```diff
@@ -0,0 +1,447 @@
...
+    if not isinstance(object2save, BaseSignal):
+        raise TypeError('object2save should be a valid hyperspy.signals.BaseSignal object')
+    """
+
+    # Not sure how to safely ignore spurious / additional dimensions
+    if len(object2save.axes_manager.shape) != len(object2save.data.shape):
```
ericpre: It should not be necessary to check this and this will actually not work with "ragged" array. That being said I don't think that there is any format in hyperspy supporting ragged array without workaround.

setup.py
```diff
@@ -73,6 +73,7 @@
     "gui-traitsui": ["hyperspy_gui_traitsui>=1.1.0"],
     "mrcz": ["blosc>=1.5", 'mrcz>=0.3.6'],
     "speed": ["numba"],
+    "usid": ["pyUSID"],
```
ericpre: I should be fine to add `pyUSID` as required dependency since it is available from the conda-forge channel and on pypi. The only major drawback is that it will increase our numpy version requirement to >1.13.

doc/user_guide/io.rst
```diff
@@ -881,6 +883,126 @@ good approximation. Further release of HyperSpy will read the time axis more
...
+In order to prevent accidental misinterpretation of information downstream, the keyword argument
+``ignore_non_linear_dims`` can be set to ``False`` which will result in a ``ValueError`` instead.
+
+.. code-block:: python
+
+    >>> hs.load("sample.h5")
+    ValueError: Cannot load provided dataset. Parameter: Bias was varied non-linearly.
+    Supply keyword argument "ignore_non_linear_dims=True" to ignore this error
+
+Lazy reads will be made available in an upcoming version of pyUSID and HyperSpy.
```
ericpre: What is the reason not to support lazy processing?

hyperspy/io_plugins/usid_hdf5.py
```diff
@@ -0,0 +1,447 @@
...
+
+    Returns
+    -------
+
+    """
+
+    sig = {'data': ndim_form,
+           'axes': dim_dict_list,
+           'metadata': {
+               'Signal':{'quantity': quantity, 'units': units, 'signal_type': sig_type},
```
ericpre: `units` is not a metadata as defined in the metadata structure: http://hyperspy.org/hyperspy-doc/current/user_guide/metadata_structure.html

hyperspy/io_plugins/usid_hdf5.py
```diff
@@ -0,0 +1,447 @@
...
+               'General': {'original_filename': h5_path, 'dataset_path': h5_dset_path,
+                           'original_file_type': 'USID HDF5', 'pyUSID_version': usid.__version__}
```
ericpre: same as above for `original_file_type`.

hyperspy/io_plugins/usid_hdf5.py
```diff
@@ -0,0 +1,447 @@
...
+                              h5_main.get_pos_values,
+                              ignore_non_linear_dims=ignore_non_linear_dims)
+    spec_dict = __get_dim_dict(h5_main.spec_dim_labels,
+                               usid.hdf_utils.get_attr(h5_main.h5_spec_inds, 'units'),
+                               h5_main.get_spec_values,
+                               ignore_non_linear_dims=ignore_non_linear_dims)
+
+    num_spec_dims = len(spec_dict)
+    num_pos_dims = len(pos_dict)
+    if verbose:
```
ericpre: Use the logger instead of the `print` function?

hyperspy/io_plugins/usid_hdf5.py
```diff
@@ -0,0 +1,447 @@
...
+                                                 dim.offset + dim_size * dim.scale,
+                                                 dim.scale)))
+    if len(dim_list) == 0:
+        return usid.Dimension('Arb', 'a. u.', 1)
+    return dim_list[::-1]
+
+# ######################################################################################################################
+
+
+def file_reader(filename, dset_path='', verbose=False, ignore_non_linear_dims=True, **kwds):
```
ericpre: Set `dset_path` default to None? Usually, we load most of the data as a list of signals.

Add parameters specific to this reader to the `load` function: https://github.com/hyperspy/hyperspy/blob/145003e38598d8244625390bc3de368de7db985d/hyperspy/io.py#L46

ericpre (commented):
A few very quick comments for now:
- At some point (fairly soon?), support for non linear axis will be added in hyperspy (#1927), it would be nice if you could add some comments to what would be necessary to change when the non linear axis branch is merged.
- it would be useful to support lazy loading with dask.

Formatting:
- out of curiosity, why using two underscore `__` instead of only one `_` which seems to be the convention in python.
- Limit all lines to a maximum of 79 characters, https://www.python.org/dev/peps/pep-0008/#maximum-line-length

doc/user_guide/io.rst
ssomnath: > * At some point (fairly soon?), support for non linear axis will be added in hyperspy (#1927), it would be nice if you could add some comments to what would be necessary to change when the non linear axis branch is merged.

OK. I added notes in _get_dim_dict()

> * it would be useful to support lazy loading with dask.

Indeed. This is being tested and will be pushed in the next release of pyUSID soon.

> * out of curiosity, why using two underscore `__` instead of only one `_` which seems to be the convention in python.

double underscore prevents those objects from being used outside the module in which they are declared. I typically use these for functions which generally have no purpose outside the module. I typically use single underscore for those internal functions that may be useful for those who know what they are doing. 

> * Limit all lines to a maximum of 79 characters, https://www.python.org/dev/peps/pep-0008/#maximum-line-length

OK. Done.

ssomnath: > It seems that it can write too!

Thanks for catching this typo. I fixed it

hyperspy/io_plugins/usid_hdf5.py
ssomnath: Turns out that I don't need to import the BaseSignal class after all so this is fine for now I guess. One question I have is how do I return a dataset (as a dictionary) which has N navigation axes and M signal axes? At the moment, this information is lost in the translation. When I [originally developed an ingestor](https://gist.github.com/ssomnath/097d2bb7c1468709ce2899512deb5b6d) - see lines 97 - 124, I was able to move the signal axes... I am not sure how to express this in the dictionary that I return now

setup.py
ssomnath: We use only one feature of Numpy 1.13 (can't remember which one) which is why we had to bump up our minimum version number. I can see how we can work around this issue. 

I currently listed pyUSID only as an optional requirement. Did you mean to include it as a core requirement? Is there a reason why bumping numpy requirement to 1.13 would be a major drawback / show-stopper?

hyperspy/io_plugins/usid_hdf5.py
ssomnath: OK. I moved all non-standard metadata to original_metadata

hyperspy/io_plugins/usid_hdf5.py
ssomnath: Done

hyperspy/io_plugins/usid_hdf5.py
ssomnath: I don't think it would be a good idea to load dozens of datasets into memory (numpy). We can change the default value to None once lazy reading is implemented. Until then, I would recommend keeping the default to reading a single dataset.

ssomnath:
I think I have addressed nearly all your comments in my fork. Let me know if you can think of any other changes.

hyperspy/io_plugins/usid_hdf5.py
ssomnath: Now, ``dset_path=None`` is the default given that lazy loading is allowed.

setup.py
ssomnath: The latest version of pyUSID (0.0.6.1) now only needs numpy 1.10. This should alleviate your concern

doc/user_guide/io.rst
ssomnath: Lazy loading and saving now supported

ericpre:
@ssomnath it would more convenient to keep this PR open and not to open another one in order to keep the discussion in the same place and not have it scattered over several PRs. For the review, it is also more useful when comparing changes. If necessary, "Force-push" to your branch is fine.
Can you please reopen this PR?

hyperspy/io_plugins/usid_hdf5.py
```diff
@@ -0,0 +1,521 @@
...
+        return _usidataset_to_signal(h5_dset,
+                                     ignore_non_linear_dims=
+                                     ignore_non_linear_dims, **kwds)
+
+    # At least close the file handle if not lazy load
+    if not kwds.get('lazy', True):
+        h5_f.close()
+
+
+def file_writer(filename, object2save, *args, **kwds):
```
ericpre: Remove `*args` since it is not used.

ericpre:
@ssomnath: sorry to coming back very late to this...
I have solve the merge conflict and it seems that recent changes in pyUSID make the USID tests fail. pyUSID 0.0.6.3 and master fails but 0.0.6.2 work fine (it seems that it started from https://github.com/pycroscopy/pyUSID/commit/8ef95004345f1fbb4636b795283d967d0d4e383b).

What the changes between pyUSID 0.0.6.2 and 0.0.6.3, which would make the USID tests of this PR fail?
What would be better to do? Update this PR and use pyUSID>=0.0.6.3 or fix pyUSID and exclude pyUSID 0.0.6.3?

ssomnath:
@ericpre - bug fixes completed. Would you mind revisiting this again?

hyperspy/io_plugins/usid_hdf5.py
```diff
@@ -0,0 +1,519 @@
...
+        Dictionary whose keys are flattened to a single level
+    Notes
+    -----
+    Taken from https://stackoverflow.com/questions/6027558/flatten-nested-
+    dictionaries-compressing-keys
+    """
+    items = []
+    for k, v in nested_dict.items():
+        new_key = parent_key + sep + k if parent_key else k
+        if isinstance(v, collections.MutableMapping):
```
ericpre: Use `collections.abc.MutableMapping` to avoid a warning.

doc/user_guide/io.rst
```diff
@@ -904,6 +906,127 @@ good approximation. Further release of HyperSpy will read the time axis more
...
+    [<Signal2D, title: HAADF, dimensions: (|128, 128)>,
+    <Signal1D, title: EELS, dimensions: (|64, 64, 1024)>]
+
+We can load a specific dataset using the ``dset_path`` keyword argument. setting it to the
+absolute path of the desired dataset will cause the single dataset to be loaded.
+
+.. code-block:: python
+
+    >>> # Loading a specific dataset
+    >>> hs.load("sample.h5", dset_path='/Measurement_004/Channel_003')
```
ericpre: Add `dset_path` to the docstring of [`load`](https://github.com/hyperspy/hyperspy/blob/1b32a1bbe82dc089e1517619fe7b53128a1c1e21/hyperspy/io.py#L56)?

ericpre (commented):
Thanks @ssomnath for the coming back quickly on this!
A few more things and it should be good to merge:
- see the few comments,
- a deprecation warning with `collections` needs to be fixed - see comment,
- https://github.com/conda-forge/pyusid-feedstock/pull/12 needs to be merged to make CI happy (the tests pass nicely locally with pyusid 0.0.7),
- fix the following userwarning when running the test suite:
```
hyperspy/tests/io/test_usid.py::TestHS2USIDallKnown::test_n_pos_0_spec
hyperspy/tests/io/test_usid.py::TestHS2USIDallKnown::test_0_pos_n_spec
hyperspy/tests/io/test_usid.py::TestHS2USIDallKnown::test_n_pos_m_spec
hyperspy/tests/io/test_usid.py::TestHS2USIDallKnown::test_append_to_existing_file
hyperspy/tests/io/test_usid.py::TestHS2USIDallKnown::test_append_to_existing_file
hyperspy/tests/io/test_usid.py::TestHS2USIDlazy::test_base_nd
hyperspy/tests/io/test_usid.py::TestHS2USIDedgeAxes::test_no_axes
hyperspy/tests/io/test_usid.py::TestHS2USIDedgeAxes::test_incorrect_axes
hyperspy/tests/io/test_usid.py::TestUSID2HSbase::test_n_pos_0_spec
hyperspy/tests/io/test_usid.py::TestUSID2HSbase::test_0_pos_n_spec
hyperspy/tests/io/test_usid.py::TestUSID2HSbase::test_n_pos_m_spec
hyperspy/tests/io/test_usid.py::TestUSID2HSbase::test_lazy_load
hyperspy/tests/io/test_usid.py::TestUSID2HSbase::test_lazy_load
hyperspy/tests/io/test_usid.py::TestUSID2HSdtype::test_complex
hyperspy/tests/io/test_usid.py::TestUSID2HSdtype::test_compound
hyperspy/tests/io/test_usid.py::TestUSID2HSdtype::test_non_linear_dimension
hyperspy/tests/io/test_usid.py::TestUSID2HSdtype::test_non_linear_dimension
hyperspy/tests/io/test_usid.py::TestUSID2HSmultiDsets::test_pick_specific
hyperspy/tests/io/test_usid.py::TestUSID2HSmultiDsets::test_read_all_by_default
  /home/eric/Dev/pyUSID/pyUSID/io/usi_data.py:151: UserWarning: This dataset does not have an N-dimensional form
    warn('This dataset does not have an N-dimensional form')

```

PR #2173, opened 2019-05-17
New `rename_pars` keyword argument for ``Expression`` component

francisco-dlp:
### Description of the change

New ``rename_pars`` keyword for ``Expression`` ``Component`` to enable using different names for the parameters in the ``expression`` and the ``Component``. This can be useful e.g. to avoid name clashes with scientific functions.

### Progress of the PR
- [x] add tests,
- [x] ready for review.

```python
>>> e = hs.model.components1D.Expression("A*x", "test", rename_pars={"A": "B"})
>>> e.parameters
(<Parameter B of test component>,)
```


ericpre:
Much better way to deal with it than the workaround in #2134!

PR #2178, opened 2019-05-18
Remove "added in version X" from docs up to 1.1 + minor docs changes

### LINKED ISSUE #1276: General documentation suggestions

thomasaarholt:
### Description of the change
- Removed "Added in version 0.7, 0.8, etc up to and including 1.0"
- Replaced aforementioned with a title where necessary
- Reworked some text, most importantly included text-references to `parameter.bmin` and `.bmax`.
- Fixed a bunch of minor warnings

This fixes #1276.
### Progress of the PR
- [x] Change implemented
- [x] ready for review.

### DIFF
--- a/doc/user_guide/dielectric_function.rst
+++ b/doc/user_guide/dielectric_function.rst
@@ -1,8 +1,6 @@
 Dielectric function tools
 -------------------------
 
-.. versionadded:: 0.7
-
 The :py:class:`~._signals.dielectric_function.DielectricFunction` class
 inherits from :py:class:`~._signals.complex_signal.ComplexSignal` and can
 thus access complex properties. To convert a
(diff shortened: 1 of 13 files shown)


francisco-dlp:
I agree that it's a lot more pleasant to read the documentation without all those labels. It is safe to assume that 3 years after the release of Hyperspy 1.0, most users should be familiar with new features and API changes introduced by them. Also, if somebody misses them, they can always go back to previous versions of the documentation.

In summary, LGTM!

PR #2180, opened 2019-05-21
New register extensions

francisco-dlp:
### Description of the change

Add "extension" mechanism to register components, signals and models provided by external projects.

In this implementation registering HyperSpy extensions involves:

1. Creating a Python package that includes new signals, models or components.
2. Declaring the objects in a ``hyperspy_extension.yaml`` file locate in the Python package root. For an example [check this](https://github.com/hyperspy/hyperspy_sample_extension/blob/master/hspy_ext/hyperspy_extension.yaml).
3. Installing the Python package that provides the extensions
4. Enabling the extensions by adding ``entry_points`` to ``setup.py``

This is part of the splitting HyperSpy effort (see #821 and #1599).

### Progress of the PR
- [x] Change implemented (can be split into several points)
    - [x] External components
    - [x] External models
    - [x] External signals
    - [x] External GUIs
        - [x] Implement for  ``hyperspy_gui_ipywidgets``
        - [x] Implement for  ``hyperspy_gui_traitsui``
    - [x] ~~Create extension register script (currently the extension must be registered manually, see below)~~ Automatically register extensions on installation
    - [x] Store info about the object provider to raise meaningful errors when attempting to load a file that requires installing an external extension.
    - [x] Add comments to example specs file (thanks @thomasaarholt)
    - [x] Add Component2D example
    - [x] Implement informative error when external package not installed
- [x] update docstring and User Guide (if appropriate),
- [x] Update Dev Guide
    - [x] Document how to write and register extensions
    - [x] Slit Dev Guide into Chapters
- [x] add tests,
- [x] ready for review.

Note that this PR adds 2 new requirements:

* [PTable](https://pypi.org/project/PTable/) to display tables in both pretty txt and html versions
* [sphinxcontrib-mermaid](https://github.com/mgaitan/sphinxcontrib-mermaid) to write flowcharts using markdown-like syntax (only required to build the docs).

### How to test it

1. Install / checkout hyperspy from this branch
2. Install https://github.com/hyperspy/hyperspy_sample_extension/

```python
>>> import hyperspy.api as hs
>>> s = hs.signals.BaseSignal([0])
>>> hs.print_known_signal_types()
    +--------------------+---------------------+--------------------+----------+
    |    signal_type     |       aliases       |     class name     | package  |
    +--------------------+---------------------+--------------------+----------+
    | DielectricFunction | dielectric function | DielectricFunction | hyperspy |
    |      EDS_SEM       |                     |   EDSSEMSpectrum   | hyperspy |
    |      EDS_TEM       |                     |   EDSTEMSpectrum   | hyperspy |
    |        EELS        |       TEM EELS      |    EELSSpectrum    | hyperspy |
    |      hologram      |                     |   HologramImage    | hyperspy |
    |      MySignal      |                     |      MySignal      | hspy_ext |
    +--------------------+---------------------+--------------------+----------+
>>> s.set_signal_type("MySignal")
>>> print(s)
<MySignal, title: , dimensions: (|1)>
>>> m = s.create_model()
>>> print(m)
<MyModel>
>>> from hspy_ext.component import MyComponent
>>> m.append(MyComponent())
>>> m.save("test_extension")
>>> sr = hs.load("test_extension.hspy")
>>> m = sr.models.restore("a")
>>> print(m)
<MyModel>
>>> print(m.components)
   # |      Attribute Name |      Component Name |      Component Type
---- | ------------------- | ------------------- | -------------------
   0 |         MyComponent |         MyComponent |         MyComponent
```


francisco-dlp:
@magnunor and @dnjohnstone, I am very interested to know your opinion on this. It'll be great if you could test it with the packages that you maintain.

thomasaarholt:
Great to see progress on this!
Please add some comments to the yaml file with possible options (particularly for the dtype). 

Also, is it necessary that fields like `signal_dimension` and `signal_type` are repeated between the yaml and signal.py?

francisco-dlp:
> Please add some comments to the yaml file with possible options (particularly for the dtype).

Good point (added to TODO list above).

Yes, the duplication is necessary in order to avoid having to import the external objects to register them. This new approach of registering objects without importing them should also boost hyperspy's importing time once we redesign the API for 2.0 (currently it imports them anyway even if it is not longer necessary.)

dnjohnstone:
> @magnunor and @dnjohnstone, I am very interested to know your opinion on this. It'll be great if you could test it with the packages that you maintain.

Basically looks good to me, thanks! @pc494 or @shogas might have useful opinions on this too.

What are you thinking for the "extension register script"? I feel like a short/medium term solution (i.e. whilst the number of extension packages is small) to make this easy for users could be for the script to just look for atomap and pyxem and register it if it's there? Maybe during setup...

Is the plan for this to be included in hyperspy before M&M this year? If so I'll add "registering with hyperspy" as a pyxem-0.9.0 milestone, which we aim to release end of July for the workshop at M&M. Will make a branch of pyxem to test this at some point later today or maybe tomorrow.

magnunor:
Very cool!

I tested this with pixStem, as currently using this for Atomap won't really be useful. However, with this functionality new being possible, I might do a small redesign of Atomap.

For pixStem I made a `hyperspy_extension.yaml`: https://gitlab.com/pixstem/pixstem/blob/hyperspy_extensions/pixstem/hyperspy_extension.yaml

It seems to work pretty well:

```python
>>> import numpy as np
>>> import hyperspy.api as hs
>>> s = hs.signals.Signal2D(np.random.random((10, 10, 20, 20)))
>>> s.set_signal_type('PixelatedSTEM')
>>> s
<PixelatedSTEM, title: , dimensions: (10, 10|20, 20)>
```

Also for lazy signal:
```python
>>> import numpy as np
>>> import hyperspy.api as hs
>>> s = hs.signals.Signal2D(np.random.random((10, 10, 20, 20)))
>>> s1 = s.as_lazy()
>>> s1.set_signal_type('PixelatedSTEM')
>>> s1
<LazyPixelatedSTEM, title: , dimensions: (10, 10|20, 20)>
```

### Some possible improvements:

There should be an error if you ask for a signal type which does not exist.

So for example:

```python
>>> import numpy as np
>>> import hyperspy.api as hs
>>> s = hs.signals.Signal2D(np.random.random((2, 20, 20)))
>>> s.set_signal_type('stuff')
>>> s
<Signal2D, title: , dimensions: (2|20, 20)>
```

Should raise some kind of error.

Similarly, if you ask for a Lazy signal, from a non-lazy signal, it should give an error message. Currently it returns the default signal for that signal dimension:

```python
>>> import numpy as np
>>> import hyperspy.api as hs
>>> s = hs.signals.Signal2D(np.random.random((10, 10, 20, 20)))
>>> s.set_signal_type('LazyPixelatedSTEM')
>>> s
<Signal2D, title: , dimensions: (10, 10|20, 20)>
```

ericpre:
> 4. Enabling the extensions by adding the name of the module to the ~/.hyperspy/hspy_extensions.yaml file. This can be automated with conda but not with pip.

We should avoid running post installation script when not necessary.
I made a [PR](https://github.com/francisco-dlp/hyperspy/pull/42) to this branch to suggest a fix for the configuration file. The approach has been tested with https://github.com/ericpre/hyperspy_sample_extension/tree/use_data_file_setup and it works fine.

The naming convention may need some tweaks.

ericpre:
Regarding documentation:
- main developer documentation in https://github.com/hyperspy/hyperspy_sample_extension 
- add a short section in the developer guide guiding to https://github.com/hyperspy/hyperspy_sample_extension?
- mention the existence of extension in the `README.md` file? and user guide?

francisco-dlp:
Thanks all for the feedback.

@magnunor, although it is possible to get a lazy signal by setting the signal type to a lazy one, it is recommended to start from the standard signal as call ``as_lazy()``. Regarding raising an error when the requested signal type is not available, the thing is that ``set_signal_type`` has never behaved like that. The reason is that the ``signal_type`` is metadata that describes the type of signal. Even if HyperSpy does not provides a signal subclass for a particular signal type, users may want to set this metadata nevertheless. Whether this was a good idea is of course debatable. I agree that it may be confusing that nothing happens after setting the ``signal_type`` to an "unsupported`` type. What about simply adding a warning to the log?

@dnjohnstone and @ericpre, I have just modified the way to register extensions. Now it happens automatically on installation by using ``entry_points``. For this to work, the package that provides the extension must have a line like [this](https://github.com/hyperspy/hyperspy_sample_extension/blob/master/setup.py#L187) in ``setup.py``. I think that this method is cleaner than the one in https://github.com/francisco-dlp/hyperspy/pull/42 . Do you agree @ericpre ?

@dnjohnstone, I think that releasing this on time for M&M is feasible.

ericpre:
> @dnjohnstone and @ericpre, I have just modified the way to register extensions. Now it happens automatically on installation by using entry_points. For this to work, the package that provides the extension must have a line like this in setup.py. I think that this method is cleaner than the one in francisco-dlp#42 . Do you agree @ericpre ?

Yes, Indeed, this is much better!

hyperspy/extensions.py
```diff
@@ -0,0 +1,52 @@
...
+            if "signals" in _ext_ext:
+                ALL_EXTENSIONS["signals"].update(_ext_ext["signals"])
+            if "components1D" in _ext_ext:
+                ALL_EXTENSIONS["components1D"].update(
+                    _ext_ext["components1D"])
+            if "components2D" in _ext_ext:
+                ALL_EXTENSIONS["components2D"].update(
+                    _ext_ext["components2D"])
+            if "guis" in _ext_ext:
+                if "toolkey" in _ext_ext:
```
ericpre:
```suggestion
                if "toolkeys" in _ext_ext["guis"]:
```

hyperspy/extensions.py
```diff
@@ -0,0 +1,52 @@
...
+                    ALL_EXTENSIONS["guis"]["toolkeys"].update(
```
ericpre:
```suggestion
                    ALL_EXTENSIONS["guis"]["toolkeys"].extend(
```

hyperspy/extensions.py
```diff
@@ -0,0 +1,52 @@
...
+                        _ext_ext["guis"["toolkeys"]])
```
ericpre:
```suggestion
                        _ext_ext["guis"]["toolkeys"])
```

dnjohnstone:
Thanks for this!

I made a branch of pyxem for this, see here: https://github.com/pyxem/pyxem/pull/409

A couple of points that came up, which are probably the start of a "HyperSpy Extension Guide":

1. I think we should define a community convention for the format of signal_type - many signals that are in the main hyperspy repo have signal_types with underscores in them e.g. "EDS_SEM" but both @magnunor and I seemed to do this natural as a camel name e.g. "ElectronDiffraction". Anyone have opinions on this?

2. signal_dimension is a problem for some of the signals we have in pyxem. E.g. DiffractionVectors is typically a ragged signal class. A number of others e.g. CrystallographicMap currently inherit direct from BaseSignal and we haven't defined a signal_dimension although we probably could. Thoughts on the best way to be dealing with these cases?

ericpre:
> 1. I think we should define a community convention for the format of signal_type - many signals that are in the main hyperspy repo have signal_types with underscores in them e.g. "EDS_SEM" but both @magnunor and I seemed to do this natural as a camel name e.g. "ElectronDiffraction". Anyone have opinions on this?

We should follow the python convention but I am not aware of any in this case. If there is no reason to change it, I would suggest to recommand to the way it is currently done in hyperspy, because this is what would make less deprecation. I don't think there is a need for a strict convention in the spelling but I agree it would be better to be consistent.
+1 on a HyperSpy Extension Guide.

> 2. signal_dimension is a problem for some of the signals we have in pyxem. E.g. DiffractionVectors is typically a ragged signal class. A number of others e.g. CrystallographicMap currently inherit direct from BaseSignal and we haven't defined a signal_dimension although we probably could. Thoughts on the best way to be dealing with these cases?

As far I understand, signals with ragged array are BaseSignal and by definition doesn't have signal dimension. I guess this is better to leave this consideration for another PR.

francisco-dlp:
Adding widgets through extensions is now implemented, see [the ipywidgets specs](https://github.com/hyperspy/hyperspy_gui_ipywidgets/pull/19/files#diff-dc4013263399a5767c101d476b8f1686) for an example.

francisco-dlp:
@dnjohnstone, I think that we should indeed define a convention for the format of ``signal_dtype``. Moreover, we have to consider that, now that we open the door to any package adding signals to HyperSpy, we have to decide on how to deal with ``signal_type``s clashes. A way to avoid them would be to add a prefix to the signal_type e.g. ``"pyXemElectronDiffraction`` or ``"pyXem_ElectronDiffraction``. In the particular case of pyXem, if everybody agrees that its ``ElectronDiffraction`` is the standard signal in the HyperSpy ecosystem, then maybe the prefix can be skipped and, if another package decides to create their own, theirs should have the prefix. However, there are some signals in pyXem (like ``TemplateMatchingResults``, ``VectorMatchingResults`` etc) that I think should carry the prefix regardless, because their signal_type doesn't mean much without proving the context (pyXem).

I think that this is an important discussion, so it'll be good to hear from as many people as possible.

dnjohnstone:
Yeah that's a complicated discussion....

I think the prefixing thing is pretty ugly but I do see your point and that it might be necessary if we're not going to have agreement around an individual package that covers some domain.Personally, I think the whole point of collaborative open-source working is to avoid re-inventing the wheel and it is therefore no bad thing to push people to work on packages together when the names would be the same because the functions are similar. For things like TemplateMatchingResults and VectorMatchingResults - I'd prefer to prefix them with DiffractionVectorMatchingResults and DiffractionTemplateMatching results than with the package name...

Another facet to that is that in the two examples discussed in this PR PixelatedSTEM (pixstem) and ElectronDiffraction (pyxem) are essentially the same thing with different names, which is also not good!

To my mind the point is that the packages should really merge to consolidate efforts and be more in the spirit of collaborative open-source work. However, we can all think of examples where that hasn't happened for various (sometimes good, sometimes not) reasons.

ericpre:
> However, there are some signals in pyXem (like TemplateMatchingResults, VectorMatchingResults etc) that I think should carry the prefix regardless, because their signal_type doesn't mean much without proving the context (pyXem).

Users are supposed to get this information from the metadata, may be we could do something along the line of: 

```python
├── General
│   └── title = 
├── History
│   └── creation
│       ├── date = 2019-05-23
│       ├── library = pyxem 1.0
│       ├── reader = io_reader_name 1.0
│       └── time = 15:00:00
└── Signal
    ├── binned = True
    └── signal_type = ElectronDiffraction
```

Alternatively, we could add an attribute containing the _name_ and the _version_ of the library used to make the signal and display it in the metadata.

Two related comments on this PR:
- Check for duplicate `signal_type` when loading the extensions?
- add a convenient function to generate a list of all available `signal_type`? Or structure it in a dictionary with name of the library as keys? Update accordingly the docstring of `set_signal_type`.

magnunor:
Naming conventions are always tricky, especially when we start combining camel case with abbreviations. I basically use camel case, since that is the python naming conventions for classes.

> Another facet to that is that in the two examples discussed in this PR PixelatedSTEM (pixstem) and ElectronDiffraction (pyxem) are essentially the same thing with different names, which is also not good!

I agree this duplication is not ideal, although there are some big differences between the libraries. pixStem is focused on non-precession pixelated STEM/4-D STEM data, which also includes things likes magnetic DPC imaging, fluctuation electron microscopy, and HOLZ analysis using the dask library. There is some overlap, which is not good as it causes duplication of work. I think the first step for improving that is combining the common data crunching functions into one shared "processing" library. However, this is a much bigger discussion which should not be done in this pull request :)

pc494:
In terms of signal types that already exist within Hyperpsy I think this is an excellent development. The `ElectronDiffraction` problem isn't going to go anywhere, but @ericpre suggestion could easily be implemented in individual ElectronDiffraction repos.

eg) 
```
dp.save(fp) #sets some metadata that says this is a pyxem.ElectronDiffraction
hs.load(fp) #throws a warning (and then loads a Signal2D as it does currently)
pxm.load(fp) #uses hs.load() internally, but suppresses the warning. Uses the metadata to convert to the correct class and carry metadata through
```

In a more general sense, I think the signal register should only contain signals defined within split hyperspy packages (and the above should be used for other organisations packages) - one of these split packages might one day be `hyperspy-ElectronDiffraction`, but in the short term this would get the ball rolling on splitting.

francisco-dlp:
Thanks for the feedback @pc494. Some comments:

The advantage of letting HyperSpy know about e.g. pyXem, is that users will be able to use hyperspy's ``load`` function to load pyXem objects if pyXem is installed.

```python
>>> hs.load("file_containing_pyXem_DP.hspy")
<ElectronDiffraction(...)>
```
If pyXem is not installed it should print an useful warning (not yet implemented):

```python
>>> dp = hs.load("file_containing_pyXem_DP.hspy")
warning: this file contains a ElectronDiffraction signal provided by pyXem, but pyXem is not installed. The signal will be loaded into a generic Signal2D class. Please install pyXem to properly load this dataset by following the instructions in https://pyxem.github.io/pyxem/#installation to install this package.
>>> m = dp.restore_model("a")
ImportError: this model contains a component provided by the Python package pyXem. However, pyXem is not installed. Please follow the instructions in https://pyxem.github.io/pyxem/#installation to install this package.
```

Regarding the content of the signal register, that will be defined by the extensions installed, nothing will be hardwired into hyperspy.

Regarding name clashes, I think that the best approach is to avoid them, otherwise it will get very confusing for users.

@dnjohnstone, I fully agree with your comments above: there shouldn't be two packages providing the same features in the hyperspy ecosystem unless there are very good reasons for such a thing to happen. Regarding pixSTEM and pyXem, my non-expert view is that, while it is true that both packages provide different features, it is also true that the domain of pixSTEM seems to include that of pyXem (while the opposite probably is not true). I think that this is confusing for users, since it is not straightforward to know which package they have to use for a certain "pixelated detector" task. Therefore, I think it'll be great if the pyXem and pixSTEM teams could join efforts. I must say that it doesn't help that the names of the packages are very similar. As I am not a regular user of any of the those packages, when people ask me about what's the right package to perform a certain task, I often give the wrong advice simply because I can't remember which is which. What brings me to the issue of naming conventions. While it is perfectly possible to write packages that provides new hyperspy objects without complying with any naming convention (and that's a good thing), I think that something like hyperspy-eels is a lot more descriptive and easier to remember than any sexier name that we can come up wtih. So, I suggest that, in the interest of the users (existing and future), we stick to the convention of  ``hyperspy-<very descriptive word>`` when deciding the name of the packages that will result from splitting HyperSpy.

pc494:
It's worth noting here that there are at least 4 packages doing various SED thing that depend on hyperspy, eg) [problematic](https://github.com/stefsmeets/problematic), [py4DSTEM](https://github.com/py4dstem/py4DSTEM) + pyxem and pixSTEM

ericpre:
This PR should also fix #2142: the compute method of lazy signal returns the correct signal.

ericpre:
This PR may be relevant for @CSSFrancis, @bsavitzky too and they could possibly have some comments/feedback on this.

francisco-dlp:
For anybody interested in trying the traitsui widgets too https://github.com/hyperspy/hyperspy_gui_traitsui/pull/21

francisco-dlp:
This is now ready for review.

ericpre:
@francisco-dlp, there are commit from #2137, would it be worth finishing #2137 first in order to simplify (not much but still) the review of this PR and avoid any annoyance with merge conflict?

doc/dev_guide/speeding_up_code.rst
```diff
@@ -0,0 +1,63 @@
...
+
+If you need to improve the speed of a given part of the code your first choice
+should be `Numba <https://numba.pydata.org/>`_. This is because Numba code is
+very similar (when not identical) to Python code, and, therefore, it is a lot
+easier to maintain than Cython code.
+
+Writing Cython code
+-------------------
+
+Cython code will only if:
```
ericpre: Something missing here.

doc/dev_guide/writing_extensions.rst
```diff
@@ -0,0 +1,275 @@
...
+`the HyperSpy Sample Extension package
+<https://github.com/hyperspy/hyperspy_sample_extension>`_.
+
+
+Creating new HyperSpy BaseSignal subclasses
+-------------------------------------------
+
+When and where create a new ``BaseSignal`` subclass
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+I are bad.
```
ericpre: Needs cleaning up, I guess?!

hyperspy/api_nogui.py
```diff
@@ -70,16 +78,9 @@
...
 
-
-from hyperspy.Release import version as __version__
-from hyperspy import signals
-from hyperspy.io import load
-from hyperspy.defaults_parser import preferences
-from hyperspy.utils import *
-from hyperspy import datasets
-from hyperspy.logger import set_log_level
 set_log_level(preferences.General.logging_level)
```
ericpre: This should come as early as possible in order get the logger set to debug during import. It seems that autopep8 is moving this around: see https://github.com/hyperspy/hyperspy/commit/e6a86a3bdc0dc06ccce23c4651f7442bfbf24506#diff-3c6be422290c5d00f0017314dd0f690e.

doc/dev_guide/writing_extensions.rst
```diff
@@ -0,0 +1,275 @@
...
+can declare new toolkeys and widgets. For example, the `hyperspy-gui-traitsui
+<https://github.com/hyperspy/hyperspy_gui_traitsui>`_ and
+`hyperspy-gui-ipywidgets
+<https://github.com/hyperspy/hyperspy_gui_ipywidgets>`_ provide widgets for
+toolkeys declared in HyperSpy.
+
+Registering toolkeys
+^^^^^^^^^^^^^^^^^^^^
+
+Typically new toolkeys are declared using the
```
ericpre: Mention `get_gui`?

hyperspy/misc/utils.py
```diff
@@ -1053,3 +1059,41 @@ def add_scalar_axis(signal):
...
+
+    Returns
+    -------
+    dic: dict
+        Dictionary containing ``package`` and ``package_version`` (if available)
+    """
+    dic = {}
+    # Note that the following can be "__main__" if the component was user
+    # defined
+    dic["package"] = obj.__module__.split(".")[0]
```
ericpre, line 1074: Is `obj` always supposed to have a `__module__` attribute?

hyperspy/misc/utils.py
```diff
@@ -1053,3 +1059,41 @@ def add_scalar_axis(signal):
...
+    if dic["package"] != "__main__":
+        try:
+            dic["package_version"] = importlib.import_module(
+                dic["package"]).__version__
+        except AttributeError:
+            dic["package_version"] = ""
+            _logger.warning(
+                "The package {package} does not set its version in " +
+                "{package}.version. Please report this issue to the " +
```
ericpre:
```suggestion
                "{package}.__version__. Please report this issue to the " +
```

francisco-dlp:
Good point, I've just updated the branch.

hyperspy/io_plugins/bruker.py
```diff
@@ -26,6 +26,21 @@
 
 # Plugin characteristics
 # ----------------------
+from os.path import splitext, basename
```
ericpre, line 29: I guess autopep8 move these imports: move `#Plugin characteristics` after the imports.

hyperspy/misc/utils.py
```diff
@@ -1053,3 +1059,41 @@ def add_scalar_axis(signal):
                     offset=0,
                     name="Scalar",
                     navigate=False)
+
+
+def get_object_package_info(obj):
+    """Get info about object package
```
ericpre, line 1064: Add a test for this function?

hyperspy/hyperspy_extension.yaml
```diff
@@ -0,0 +1,229 @@
...
+  LazyDielectricFunction:
+    signal_type: DielectricFunction
+    signal_type_aliases:
+      - dielectric function
+    signal_dimension: 1
+    dtype: complex
+    lazy: True
+    module: hyperspy._signals.dielectric_function
+
+components1D:
```
ericpre, line 143: Use `components` instead of `components1D` or `components2D`?

hyperspy/components1d.py
```diff
@@ -31,29 +31,18 @@
...
+from hyperspy.extensions import EXTENSIONS as _EXTENSIONS
+import importlib
+
+_g = globals()
+for _component, _specs in _EXTENSIONS["components1D"].items():
+    _g[_component] = getattr(
+        importlib.import_module(
+            _specs["module"]), _component)
+
+del importlib
```
ericpre: Tidy up a bit more the namespace?

hyperspy/ui_registry.py
```diff
@@ -13,16 +13,27 @@
...
+if "widgets" in ALL_EXTENSIONS["GUI"] and ALL_EXTENSIONS["GUI"]["widgets"]:
+    for toolkit, widgets in ALL_EXTENSIONS["GUI"]["widgets"].items():
+        TOOLKIT_REGISTRY.add(toolkit)
+        for toolkey, specs in widgets.items():
+            if not toolkey in UI_REGISTRY:
+                raise NameError("%s is not a registered toolkey" % toolkey)
+            UI_REGISTRY[toolkey][toolkit] = specs
+
+
 def register_widget(toolkit, toolkey):
```
ericpre: Is `register_widget` still necessary? There is an import in `hyperspy_gui_traitsui`, which is not necessary anymore, so this could be removed.

hyperspy/extensions.py
```diff
@@ -0,0 +1,59 @@
...
+    os.path.abspath(os.path.dirname(__file__)), "hyperspy_extension.yaml")
+with open(_ext_f, 'r') as stream:
+    EXTENSIONS = yaml.safe_load(stream)
+EXTENSIONS["GUI"]["widgets"] = {}
+
+# External extensions are not integrated into the API and not
+# import unless needed
+ALL_EXTENSIONS = copy.deepcopy(EXTENSIONS)
+
+_ext_extensions = [
```
ericpre: Rename to `_external_extensions` to make it easier to read.

hyperspy/extensions.py
```diff
@@ -0,0 +1,59 @@
...
+    entry_point.module_name
+    for entry_point in pkg_resources.iter_entry_points('hyperspy.extensions')]
+
+for _ext_ext_mod in _ext_extensions:
```
ericpre: Rename to `_ext_ext_module` to make it easier to read.

ericpre (changes requested):
This looks very good to me, thanks @francisco-dlp for this very good job!
I put a couple of comments which should fairly quick to address.

I feel it would be good to explain a bit more about how it is designed to be used for both the user and the developer, in particular to clarify:
- for the user: the extensions objects are not added to the hyperspy api and the creating of the extension object can be done through the extension api (or by using `set_signal_type`, as it is already explained).
- for the developer: when calling generic function from hyperspy, the `signal_type` is key to keep the same kind of object (through call of `set_signal_type` internally), if a new object needs to be created, for example.

hyperspy/misc/utils.py
francisco-dlp: Yes, as explained in the comment above the line of code.

hyperspy/hyperspy_extension.yaml
francisco-dlp: Good point. In principle that could be done, as we can filter the components later on. However, filtering them at the yaml level i) saves some code ii) makes more explicit what sort of component the extension is adding.

I am happy with both, so I let you decide. Do let me know.

francisco-dlp:
Thanks for the great review. I've fixed all the issues that you've spotted except for the ones that you've fixed yourself in https://github.com/francisco-dlp/hyperspy/pull/44 and the components1D/2D discussion above.

hyperspy/hyperspy_extension.yaml
ericpre: I think I would prefer to have:
```yml
signals:
  MySignal:
    signal_type: "MySignal"
    signal_dimension: 1
    dtype: real
    lazy: False
    module: my_package.signal
components:
  MyComponentClass:
    signal_dimension: 1
    module: my_package.components
```
Because components and signals would at the same level and it would be "symmetrical".
We could also consider changing the API and have all the HyperSpy components in `hs.components`, but this is a discussions for another time! ;)

francisco-dlp: After thinking on this a bit more, I think that it is better to keep it as it is because i) it is less verbose ii) the components are stored "pre-filtered" into 1D and 2D components.

francisco-dlp:
I've just added a new ``class`` field to the components definition so that we set ``_id_name`` to an UUID instead of the class name. This is because there is a high chance that there'll be name clashes if we don't take that approach. Also, UUIDs makes it easy to find the object by simply searching in the web. That could be useful if e.g. a component is relocated.

This is ready for review again.

hyperspy/misc/utils.py
ericpre: If it does not have `__module__` attribute, an error will be raised and it should be caught.

doc/dev_guide/writing_extensions.rst
```diff
@@ -212,18 +212,35 @@ example:
...
+      # UUIDs. The same UUID must be stored in the components ``_id_name`` attribute.
+      2ffbe0b5-a991-4fc5-a089-d2818a80a7e0:
         # The module where the component is located.
         module: my_package.components
+        class: MyComponent2DClass
+
+.. note::
+
+  HyperSpy's legacy components use their class name instead of an UUID as
+  ``id_name``. This is for compatibility with old versions of the software. We
```
ericpre: Typo: ``_id_name``.

doc/dev_guide/writing_extensions.rst
```diff
@@ -212,18 +212,35 @@ example:
...
+  encourage all extension developers to use UUIDs instead in order to i) avoid
+  name clashes ii) make it easy to find the component online if e.g. the package
```
ericpre, line 242: Above in the yaml description: it says "_id_name of the component. It must be an UUID4", while here it encourage the use of uuid. I would stick to "it must" to keep simple and consistent (even if it still works without using uuid).

ericpre (commented):
There are still a few unanswered comments left from my previous review.

A few other comments regarding UUID:
- shall we strongly suggest that a component API changes, its UUID should changes too?
- it is not clear to me what we will not UUID for everything? Using UUID avoids name clashes but it does not enforce avoiding the implementation of the same thing in different library (if it has the same name, it is most likely that it does very similar kind of thing!). Anyway, I am not convinced that name clashes will be efficient to avoid different implementation of the same thing.

ericpre:
update hyperspy_sample_extention to use uuid: https://github.com/hyperspy/hyperspy_sample_extension/pull/1

hyperspy/misc/utils.py
francisco-dlp: Currently this is only used internally for components and signals, all of which have a ``__module__`` attribute. If the function ever gets an object without it, I think that it is good that it raises the exception that it generates i.e. I think that there is no need to catch it.

doc/dev_guide/writing_extensions.rst
francisco-dlp: Good point, done.

francisco-dlp:
>  shall we strongly suggest that a component API changes, its UUID should changes too?

I don't have a strong view on this, it depends on the circumstances. For example, in  https://github.com/thomasaarholt/hyperspy/pull/12 I changed the `_id_name` because I thought that it was good to take the opportunity to "upgrade it" to UUID.

> it is not clear to me what we will not UUID for everything?

That's because HyperSpy automatically switches classes when e.g. using ``transpose``. It does it based on the ``signal_type``. So if ``signal_type`` is ``"EELS"`` and ``signal_dimension`` is 1, it'll assign the ``EELSSpectrum`` class. If ``signal_dimension`` is 2, it assigns the ``Signal2D`` class. If there was an ``EFTEM`` class it would assign that one instead. So, in short, the assignment of classes depends on the ones available. One we split HyperSpy, and ``EELS`` dataset can be assigned a ``Signal1D`` class in the EELS package is not installed. So, unlike in the case of components, where there is only 1 class that can recreate a particular component, for signals there are many options.


> Using UUID avoids name clashes but it does not enforce avoiding the implementation of the same thing in different library (if it has the same name, it is most likely that it does very similar kind of thing!). Anyway, I am not convinced that name clashes will be efficient to avoid different implementation of the same thing.

I agree, but I think that there are cases when different packages may like to have their own implementation of a given component. As the number of extensions grow, it won't be easy to check whether a given component name has been taken. UUIDs avoid the worst consequences of something like that happening such as hspy files that cannot be loaded. In any case, we cannot (and I think that we should not even if we could) force external package developers not to implement the same features in different packages, even if almost always that's a bad idea.

hyperspy/docstrings/plot.py
```diff
@@ -10,87 +10,69 @@
...
-        the "navigator", with a cursor to navigate that data is
-        raised. In any case it is possible to navigate the data using
-        the sliders. Currently only signals with signal_dimension equal to
-        0, 1 and 2 can be plotted.
-
-        Parameters
-        ----------
-        %s""" % BASE_PLOT_DOCSTRING_PARAMETERS
-
-PLOT1D_DOCSTRING = \
```
ericpre: This merge conflict haven't been resolved correctly.

ericpre:
This look good to me, once the merge conflict is fixed (as pointed out by the test suite failure), it can be merged.

francisco-dlp:
Done, thanks. I'll merge once the tests pass.

ericpre (approved):

PR #2182, opened 2019-05-21
Roi manager

ericpre:
### Description of the change

Currently, adding ROI requires the initialisation of ROI with some values and then adding it to the signal using the `add_widget` method. This PR aims to create ROI more easily and interactively and also add support for managing several ROIs together.
This could be useful to define multiple windows for fitting background for example. This is also a different approach to #1785.

It would be good to have some feedback on this approach and if we agree on the design then I can work on the rest...

### Progress of the PR
- [ ] Change implemented
  - [x] add ROIsManager
  - [x] create ROI interactively for SpanROI
  - [ ] create ROI interactively for all possible ROI
  - [ ] add tool to the matplotlib toolbar to create SpanROI interactively
  - [ ] add tool to the matplotlib toolbar for all possible ROI depending if the figure is 1D or 2D
  - [ ] add shorcut to delete ROI from the figure
  - [ ] add shortcut to create ROI interactively?
  - [ ] save ROI as dictionary
  - [ ] create ROI from dictionary
  - [ ] add GUI to ROIsManager
  - [ ] use multiple SpanROI to define the windows to fit the background of EDS or EELS spectrum
  - [ ] save background windows as dictionary in the metadata for EDS fitting?
- [ ] update docstring,
- [ ] update user guide,
- [ ] add tests,
- [ ] ready for review.

Before to create a span selector:
```python
import hyperspy.api as hs
import numpy as np
s = hs.signals.Signal1D(np.arange(100))
s.plot()
span_roi = hs.roi.SpanROI(left=20, right=50)
span_roi.add_widgets(s)
```

With this PR:
```python
import hyperspy.api as hs
import numpy as np
s = hs.signals.Signal1D(np.arange(100))
s.plot()
# Click on the button "SpanROI" of the matplotlib toolbar to add the span selector interactively
span_selector = s.rois_manager[0]
```

the more complicated version:
```python
import hyperspy.api as hs
import numpy as np
s = hs.signals.Signal1D(np.arange(100))
s.plot()
s.rois_manager.create_ROI_interactively(s._plot.signal_plot)
# Click on the matplotlib figure to add the span selector
span_selector = s.rois_manager[0]
```

### DIFF
--- a/hyperspy/__init__.py
+++ b/hyperspy/__init__.py
@@ -4,6 +4,9 @@ import logging
 
 _logger = logging.getLogger(__name__)
 
+import matplotlib
+# Necessary to add tool to the toolbar
+matplotlib.rcParams["toolbar"] = "toolmanager"
 
 from hyperspy import docstrings
 
--- a/hyperspy/drawing/_widgets/range.py
+++ b/hyperspy/drawing/_widgets/range.py
@@ -76,9 +76,10 @@ class RangeWidget(ResizableDraggableWidgetBase):
                 self.ax = None
         self.__is_on = value
 
-    def _add_patch_to(self, ax):
+    def _add_patch_to(self, ax, set_initial_value=True):
         self.span = ModifiableSpanSelector(ax, **self._SpanSelector_kwargs)
-        self.span.set_initial(self._get_range())
+        if set_initial_value:
+            self.span.set_initial(self._get_range())
         self.span.bounds_check = True
         self.span.snap_position = self.snap_position
         self.span.snap_size = self.snap_size
@@ -412,7 +413,7 @@ class ModifiableSpanSelector(SpanSelector):
             return False, {}
         x_pt = self._get_point_size_in_data_units()
         hit = self._range[0] - x_pt, self._range[1] + x_pt
-        if hit[0] < self._get_mouse_position < hit[1]:
+        if hit[0] < self._get_mouse_position(mouseevent) < hit[1]:
             return True, {}
         return False, {}
 
--- a/hyperspy/drawing/figure.py
+++ b/hyperspy/drawing/figure.py
@@ -27,7 +27,7 @@ from hyperspy.drawing import utils
 _logger = logging.getLogger(__name__)
 
 
-class BlittedFigure(object):
+class BlittedFigure:
 
     def __init__(self):
         self._draw_event_cid = None
@@ -59,6 +59,9 @@ class BlittedFigure(object):
         if self.figure.canvas.supports_blit:
             self._draw_event_cid = self.figure.canvas.mpl_connect(
                 'draw_event', self._on_blit_draw)
+
+    def add_ROI_interactively(self, event=None):
+        self.rois_manager.create_ROI_interactively(self)
 
     def _on_blit_draw(self, *args):
         fig = self.figure
--- a/hyperspy/drawing/image.py
+++ b/hyperspy/drawing/image.py
@@ -59,7 +59,7 @@ class ImagePlot(BlittedFigure):
     """ % PLOT2D_DOCSTRING
 
     def __init__(self):
-        super(ImagePlot, self).__init__()
+        super().__init__()
         self.data_function = None
         # Args to pass to `__call__`
         self.data_function_kwargs = {}
--- a/hyperspy/drawing/mpl_he.py
+++ b/hyperspy/drawing/mpl_he.py
@@ -90,6 +90,7 @@ class MPL_HyperExplorer(object):
             sf.ylabel = r'$\Sigma\mathrm{data\,over\,all\,other\,axes}$'
             sf.axis = axis
             sf.axes_manager = self.axes_manager
+            sf.rois_manager = self.rois_manager
             self.navigator_plot = sf
             # Create a line to the left axis with the default
             # indices
@@ -112,6 +113,7 @@ class MPL_HyperExplorer(object):
         elif len(self.navigator_data_function().shape) >= 2:
             imf = image.ImagePlot()
             imf.data_function = self.navigator_data_function
+            imf.rois_manager = self.rois_manager
             imf.colorbar = colorbar
             imf.scalebar = scalebar
             imf.scalebar_color = scalebar_color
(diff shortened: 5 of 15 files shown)


thomasaarholt:
I was getting no matplotlib SpanROI toolbar option, so I tried commenting out the try-except from line 70 in `hyperspy/drawing/signal1d.py`. That let me see the error:

```python
<ipython-input-1-6f79b9b87d5e> in <module>
      3 import numpy as np
      4 s = hs.signals.Signal1D(np.arange(100))
----> 5 s.plot()

(...)

~/hyperspy/hyperspy/drawing/signal1d.py in __init__(self, title)
     57         self.create_figure()
     58 
---> 59         self._init_toolbar()
     60         self.create_axis()
     61 

~/hyperspy/hyperspy/drawing/signal1d.py in _init_toolbar(self)
     72             warnings.simplefilter("ignore")
     73             toolmanager = self.figure.canvas.manager.toolmanager
---> 74             toolmanager.add_tool("SpanROI", SpanROITool)
     75             self.figure.canvas.manager.toolbar.add_tool(
     76                     toolmanager.get_tool("SpanROI"), "toolgroup")

AttributeError: 'NoneType' object has no attribute 'add_tool'
```

On further inspection, it turns out that the error is from me using the notebook/widget backends rather than qt5. It works on the latter, but not on the former. To reproduce, in jupyter notebook call `%matplotlib notebook` before running your above example.

Edit: I think that adding GUI functionality like what you are suggesting as part of this PR, is super cool.

ericpre:
It look to me that the widget backend (for jupyterlab) is still fairly experimental (from what I have tested if), so this is not a big surprise that it doesn't support all matplotlib features. The repo of this backend haven't received any commit since July last year... :(
The notebook backend works fine, but as we know it is not compatible with jupyterlab.

francisco-dlp:
For some reason this doesn't work for me using the ``notebook`` backend in the notebook either, but it does work well using the qt backend.

On the approach itself I have mixed feelings. On the one hand, it is cool to be able to add a ROI just by clicking a button. On the other hand,  this requires adding a ROI register to ``BaseSignal``, something I find counter-intuitive since ROIs are designed to be independent from signals (I would prefer a general register, but see below).  Also, just like any other UI actions in HyperSpy, it makes the notebooks less reproducible because in HyperSpy UI actions don't generate the code required to perform the same task by scripting.

The thing is that HyperSpyUI solves both issues nicely: i) it provides a general register for HyperSpy objects ii) it has a macro recorder that can generate the code to execute all GUI actions by scripting. I don't use HyperSpyUI much because I prefer scripting to UIs (that was actually my main motivation to start working in what today is HyperSpy). However, one could imagine a blend of HyperSpy & HyperSpyUI that could offer the best of both worlds and that I can imagine using as my default HyperSpy interface:

i) Runs in a Jupyter Notebook but HyperSpyUI runs alongside (maybe activated using a %hyperspyui magic like the QtConsole). Alternatively, the Jupyter Notebook could be embedded in the HyperSpyUI GUI, although I am partial to the other way around solution.
ii) Uses HyperSpyUI's (automatic) register in parallel to the standard IPython variables register in the notebook.
iii) When using the notebook, figures are displayed optionally in HyperSpyUI. (This solves the issue of having tens of figures floating around.)
iv) Uses HyperSpyUI's macro recorder to automatically  inject code into the Jupyter Notebook to execute the tasks that the user has performed using the standard HyperSpy widgets.
v) Actions like creating ROIs can be performed using HyperSpyUI's toolbar.

Of course, the drawback of this approach is that it requires the HyperSpyUI GUI to run alongside the notebook, what may not always be possible. An alternative would be to recreate HyperSpyUI's toolbar inside the notebook and/or the mpl toolbar and keep on embedding the figures in the notebook, losing advantage iii.

What do you think? It'll be good to hear also from @vidartf .

ericpre:
There are several things mixed in this PRs, which is possibly not a very good thing, but my aim is mainly:
1. be able to interactively handle several ROIs and adjust the ROIs for model fitting, typically to fit the background spectra (EDS, EELS, Raman, etc),
2. add/remove ROIs or widgets interactively for 1.
3. Store the ROI in the signal for 1. for example when reopening the fitted data.

The, while working on this, I realised that adding ROIs interactively could solved a few others issues and this approach could be generalised.

> On the other hand, this requires adding a ROI register to BaseSignal, something I find counter-intuitive since ROIs are designed to be independent from signals (I would prefer a general register, but see below).

Yes, indeed, it is not clear to me, if the ROI manager should be independent of the signal or not.

Clearly connecting a notebook to the console of hyperspyui would be interesting but I think it would fit a different purpose of this PR.

vidartf:
I haven't looked at this code for a few years, so this is mostly my gut reactions to some of the points :)

1. I haven't looked at the code in detail, but having a way to initialize a ROI interactively on a figure is a useful feature (especially since it can be a bit tricky). Where such code belongs, I have no idea (if it works as a stand-alone function, that is probably best, but I don't know if that is possible).

2. I'm not sure if it makes sense to have buttons for adding ROIs on the general plot widgets (UI-initiated creation seems counter-intuitive from a scripting base). Having a function/method to add such a button on request might make sense for certain use cases (e.g. if you are building an app-like interface). Although it might be argued that functionality should best exist elsewhere.

3. Having a manager to organize ROIs across multiple signals/widgets might also make sense (probably orthogonal to the other issues though?). Immediately, this seems like it would be a new object that keeps references to widgets/signals and ROIs (and that widgets/signals/ROIs are agnostic about managers).

vidartf:
Just to clarify the difference between pt 1 and 2:
- Having a script that shows a figure and blocks on a user inputting an ROI would be point 1.
- Having a plot with a button that can create new ROIs at an arbitrary point in time would be point 2.

ericpre:
Discussion on gitter on using ROI for the plot pointer, which may be relevant to this topic:
https://gitter.im/hyperspy/hyperspy?at=5ddd1bf7f65fec1c8e726d55

A ROI register may be useful to store the ROIs used in a signal plot and would make easier to fetch them for reuse, for example. to add them on a compatible signal.

Issue #2176, opened 2019-05-18
Ambiguity on "Power Law" vs "Powerlaw"


thomasaarholt:
`s.remove_background` takes the `background_type="Power Law"` argument and default value. However, its docstring suggests that "PowerLaw" is correct, and the _model_ docs ([model](http://hyperspy.org/hyperspy-doc/current/user_guide/model.html) and [eels](http://hyperspy.org/hyperspy-doc/current/user_guide/eels.html)) refer to `PowerLaw` not `Power Law` in both component and attribute names.

This is confusing, particularly so for new users, so i suggest we change it to be consistent. "Power Law" is the only component that includes a space in its name, so unless anyone feels strongly about it, I suggest we make it "PowerLaw", not "Power Law".

francisco-dlp:
+1

jlaehne:
The component's correct name is `PowerLaw` indeed. The only place where `Power Law` was used was in the `remove_background()` function. To follow the suggestion by @thomasaarhodlt, I created #2184 to implement the change.

thomasaarholt:
@francisco-dlp this does technically break the API. What do you suggest?

francisco-dlp:
I think that it doesn't because ``remove_background`` still access ``"Power Law"``, regardless of if it is a good idea or not. Probably we should deprecate it this alias and remove it in 2.0,

jlaehne:
I made `PowerLaw` the default naming, but left the lines where `Power Law` is identified as alias. So as @francisco-dlp says, it should not break the API. Whether, that alias should be removed at some point, is another question.

thomasaarholt:
Neat! I withdraw my previous statement!

Yes, @jlaehne if you could add a `if background_type=="Power Law": raise_warning` line, then that would be great.

ericpre:
This issue has changed with #2184. Now the `remove_background` method accept all version of `Power Law` with space and capitalise:
https://github.com/hyperspy/hyperspy/blob/03cb77f1417dc349c0184b3489812f6c9af7f71c/hyperspy/signal_tools.py#L1198

I think this issue can be closed, because the possible values of the `background` argument of the `remove_background` method doesn't need to be consistent the actual component name.

thomasaarholt:
Agreed

PR #2184, opened 2019-05-22
Powerlaw naming

jlaehne:
Implements change in Naming of PowerLaw component within background removal tool suggested in #2176. Removed space in the name when referring to the PowerLaw component. Checks for the two variations `PowerLaw` and `Power Law` are left for backwards compatibility.

### Progress of the PR
- [X] ready for review.


ericpre:
This is a detail but I think prefer to have "Power law" in the GUI rather than "PowerLaw":

image

I guess it should be possible to set `PowerLaw` as parameter of `remove_background`. What is displayed in the GUI is defined in the [`t.Enum`](https://github.com/hyperspy/hyperspy/blob/4cd8e81d95a286c587bff4d12568f088b3dd5481/hyperspy/signal_tools.py#L647).

francisco-dlp:
@ericpre, I propose to merge this since fixing the name displayed in the widgets is a different issue. Personally, I think that, even though "Power law" does look better, complexifying the code just for that may not be worth it.

jlaehne:
To avoid confusions, as pointed out by @thomasaarholt, I think it would be best to also use the real component name in the GUI. But, I guess, in the end, what is displayed is a matter of taste.

ericpre:
This is a simple change (2 lines with a if statement) to get the spelling correct, see https://github.com/ericpre/hyperspy/commit/919484476ba282d454ff768fca65d99f3b883409. Users using the GUIs and not familiar with hyperpsy components may just think this is a typo.

jlaehne:
Well, we have a parallel discussion between here and #2176. 

What @ericpre suggests here is to use `Power law` (by the way introducing a third spelling to the `Power Law` used so far) in the function and having `PowerLaw` as an alias for the command line.

In #2176, @thomasaarholt and @francisco-dlp were aiming at completely of making `Power Law` an alias and completely removing it in future versions, at the expense of the spelling in the GUI.

jlaehne:
As no one else commented, I implemented the changes suggested by @ericpre. 

While the background removal GUI now displays 'Power law', inline also 'Power Law' and 'PowerLaw' are accepted. That should make it fool proof in the usage.

jlaehne:
Fixed merge conflicts and extended the Name handling concept to `SkewNormal` component.

tjof2:
What's the status of this?

jlaehne:
Should be ready, can you have a look @thomasaarholt and provide a review as you proposed this originally.

ericpre (changes requested):
A more simple approach as done in #2307 (see https://github.com/ericpre/hyperspy/blob/82eed9b8374e7699e6f1103dff1d4c6c782a18c2/hyperspy/signal_tools.py#L1173) would be more robust and simple.

ericpre:
Done in #2307.

jlaehne:
This solution differed/went beyond #2307 in three aspects:
1. 'Power law' would be the correct spelling in the GUI not 'Power Law'
2. The naming should be consistent also for the new components SkewNormal and SplitVoigt
3. This implementation accepted the three aliases 'Power law', 'Power Law' and 'PowerLaw' in the CLI call of the `remove_background` function.

I have created a new PR #2465 that adds these three points (with #2307 this is much cleaner than the original) and also updates the documentation.

PR #2185, opened 2019-05-24
Estimate parameters for Lorentzian component

jlaehne:
Add an estimate_parameter function for the Lorentzian component based on the median (=centre) and half interquartile distance (=gamma). Responds to error for bad starting parameters identified in https://github.com/hyperspy/hyperspy/pull/2134#issuecomment-494968395.

Also added get/set functions for height parameter both for Lorentzian and Gaussian.

### Progress of the PR
- [X] Added get/set functions for height in Lorentzian and Gaussian,
- [X] Change implemented for only_current=True,
- [X] make it work for only_current=False 
- [X] add tests,
- [X] ready for review.

```g = hs.model.components1D.Lorentzian()
x = np.arange(-10, 10, 0.01)
data = np.zeros((32, 32, 2000))
data[:] = g.function(x).reshape((1, 1, 2000))
s = hs.signals.Signal1D(data)
s.axes_manager._axes[-1].offset = -10
s.axes_manager._axes[-1].scale = 0.01
g.estimate_parameters(s, -10, 10, True)
```


francisco-dlp:
A side-effect of this PR is that, once done, you can easily add ``Lorentzian`` to the ``remove_background`` options.

PR #2186, opened 2019-05-24
Fix definition condition for expression component

ericpre:
Fix the issue with using `where` to define definition range in expression component. See https://github.com/hyperspy/hyperspy/pull/2137#issuecomment-484787027 for more details.

@jlaehne, do you want to give a try with the expression base component in #2137? I haven't tried but normally, now it should work.

- [x] Add `definition_condition` parameter to the Expression component,
- [x] fix `PowerLaw` component, which was not behaving as expected (`left_cutoff` was ignored),
- [x] add tests,
- [x] ready for review.


jlaehne:
I used this to make #2137 work and checked that it works correctly for the `left_cutoff` in `PowerLaw`. 

So LGTM.

francisco-dlp:
I agree that this is a workaround for #2137. However, I think that this approach suffers from some issues:
1. It is not general enough because it only fixes the issue when there is a single ``where`` in the expression.
2. Probably, the reason why sympy does not differentiate expressions that contain ``where`` is that the derivative is not defined at the end points of the intervals defined by ``where``. Therefore, properly fixing the issue would involve adding nans at those points. However, the consumers of the ``grad_*`` functions are scipy  optimzer that will probaly get upset if they get ``nans``.
3. Enabling the differentiation of functions defined by parts is something to contribute to sympy not hyperspy.

I simpler way to solve this issue would be not to create the gradient methods when an expression contains ``where``.

ericpre:
Yes, this is a not very nice workaround... I reverted it and changed the code to exclude the automatic calculation of gradient for expression containing "where" condition and also added back the original gradient for the power law.
The gradients are not using numexpr but this could be improved in another PR.

francisco-dlp:
The issue with the approach is that if they ever add support for function defined by parts, then HyperSpy won't compute the gradients regardless. Furthermore, there can be other scenarios in which sympy's lambdify derivatives cannot be computed or the output is invalid.
So, what about taking a simpler approach:

* We could add a ``gradients`` keyword to the component constructor that defaults to ``True``.  In the docstring we can explain that computing the gradient of functions that are not derivable everywhere can fail (raise an exception when creating the ``Expression`` or when calling the gradient). In those cases setting ``gradients`` to ``False`` can help.
* The above may be enough and it has the merit of being the simplest and most transparent solution. However, if we want to minimize the number of exception that users are exposed to, we could catch the exception when computing the gradient and send it to the logger and try calling the gradient function before setting the gradient attribute and, if it raises an exception, catch it and send it to the logger.

ericpre:
There is a test of the `PowerLaw` component which should failed if sympy add support for derivative of expression containing a "where" condition.

PR #2190, opened 2019-05-31
Fix plotting tests with matplotlib 3.1

ericpre:
The main changes are related to the bounding box of the axis, which requires updating the baseline images, see commits comments for other details.

### Progress of the PR
- [x] Fix tests,
- [x] Fix deprecation of `SpanSelector.buttonDown` which actually raise an error...
- [x] ready for review.


thomasaarholt:
I've looked through the code and tested the various cases of mva (plot_explained, threshold) and the two cases in #2195. Looks good to me!

PS: Well done fixing a bug 48 hrs before it was first reported!

francisco-dlp:
Shouldn't we update the mpl version i[n this line of setup.py](https://github.com/hyperspy/hyperspy/blob/RELEASE_next_minor/setup.py#L78).

Other than that, LGTM.

ericpre:
> Shouldn't we update the mpl version in this line of setup.py.

Yes, done.

PR #2202, opened 2019-06-05
Raise TypeError when plotting float16 image

### LINKED ISSUE #1417: Add support for plotting float16?
Using ROI with certain dtypes throws an exception:

```Python
import hyperspy.api as hs
import numpy as np
s = hs.signals.Signal2D(np.random.random(size=(100,100)))
line_roi = hs.roi.Line2DROI(1., 1., 50., 50., 1)
s.plot()
roi1D = line_roi.interactive(s, color="yellow") # This works

s = hs.signals.Signal2D(np.random.random(size=(100,100)).astype('float16'))
line_roi = hs.roi.Line2DROI(1., 1., 50., 50., 1)
s.plot()
roi1D = line_roi.interactive(s, color="yellow") # This throws an exception, but adds the line roi
```

The exception: https://gist.github.com/magnunor/bb66dafad715e225a4292693e9beb2db

ericpre:
Since this is not supported by matplotlib, we raise a TypeError suggesting to change the dtype.

Fix #1417.

### DIFF
--- a/doc/user_guide/visualisation.rst
+++ b/doc/user_guide/visualisation.rst
@@ -125,6 +125,13 @@ To close all the figures run the following command:
     learn how to pan/zoom and more  `in the matplotlib documentation
     <http://matplotlib.sourceforge.net/users/navigation_toolbar.html>`_
 
+
+.. NOTE::
+
+    Plotting ``float16`` image is not supported by matplotlib; however, it is possible to
+    to convert the type of the data by using the ``change_dtype`` method, e.g. 
+    `s.change_dtype('float32')`.
+
 Multidimensional image data
 ===========================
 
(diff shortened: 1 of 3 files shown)


francisco-dlp:
Although the error implemented in this PR is more helpful than the [current one](https://gist.github.com/magnunor/bb66dafad715e225a4292693e9beb2db) , the current one shows more explicitly where is the issue. Also, I think that it is not up to hyperspy to raise an error when an external library (in this case it seems scipy.ndimage) doesn't support a certain dtype because, if scipy.ndimage and matplotlib start supporting float16 dtype at some point in the future, it'll be hyperspy along who will be the issue.

ericpre:
Yes, I agree, I removed the error message and left the note in the user guide.

PR #2205, opened 2019-06-10
Snip background removal for 1D spectra

pquinn-dls:
### Description of the change
SNIP is a standard background removal tool for XRF/EDS spectra but can also be applied to any 1D spectra, e.g. 1D XRD patterns. 

I've created a snip component, added it to BackgroundRemoval in signal_tools 
and added it to remove_background in signal1D.

It works pretty well in command line (see below). The interactive GUI works but I couldn't get the component parameters to appear in the GUI so I'm not setting the traits correctly. 
In GUI mode you should select a peak - the peak width is used to refine the filter.

It's more of a statistical filter than a fitter so I'm not sure if it works as a component so I'm putting the pull request up for people to look at first.

### Progress of the PR
- [ ] Get comments on outline and fix issues
- [ ] update docstrings (if appropriate),
- [ ] add tests,
- [ ] ready for review.

```python
>>> import hyperspy.api as hs
>>> s = hs.datasets.example_signals.EDS_TEM_Spectrum()
>>>sb=sc.remove_background(signal_range=(0.,20.),background_type='Snip',fast=True)
>>>sb.plot()
```
image


hyperspy/_signals/signal1d.py
```diff
@@ -1068,75 +1069,24 @@ def remove_background(
             signal_range='interactive',
             background_type='Power Law',
             polynomial_order=2,
+            snip_width=10,
+            snip_iterations=16,
             fast=True,
             zero_fill=False,
             plot_remainder=True,
             show_progressbar=None, display=True, toolkit=None):
-        """
```
thomasaarholt, line 1077: It looks like you accidentally removed the docstring from the `remove_background` method. When you add it back, you could also add "Snip" to the background_type as well as document the snip_width and snip_iterations.

hyperspy/misc/math/filter/snip.py
```diff
@@ -0,0 +1,75 @@
...
+
+    .. [1] C.G. Ryan etc, "SNIP, a statistics-sensitive background
+           treatment for the quantitative analysis of PIXE spectra in
+           geoscience applications", Nuclear Instruments and Methods in
+           Physics Research Section B, vol. 34, 1998.
+    """
+
+    background = np.array(spectrum)
+    spectra_size = len(background)
+    window_p = int(width)
```
thomasaarholt, line 37: It would be better to raise an error if width is a float. At the moment people can specify a float, and the code will automatically convert this. `raise ValueError('The window_p argument must be an integer')`

thomasaarholt: also, what does the p in window_p stand for?

hyperspy/_signals/signal1d.py
thomasaarholt: The disappearance of the docstring is also what is causing the tests to fail.

thomasaarholt:
Very nice to see a new component added, and good work! I haven't been able to run the code from where I currently am, but I had a read through, see comments above.

We need to add a test to test both the component as well as the SNIP code itself. I suggest you draw inspiration from the [Lorentzian](https://github.com/hyperspy/hyperspy/blob/RELEASE_next_minor/hyperspy/tests/component/test_lorentzian.py) and [Gaussian](https://github.com/hyperspy/hyperspy/blob/RELEASE_next_minor/hyperspy/tests/component/test_gaussian.py) test cases. (The gaussian case is rather thorough, you probably don't need that much).

thomasaarholt:
Also, apologies for not getting to this before now.

PR #2216, opened 2019-06-25
Add gamma adjustment and fix contrast editor tool

ericpre:
### Progress of the PR
- [x] Use `SymLogNorm` matplotlib norm to plot log of negative,
- [x] add gamma adjustment to plot signal2D,
- [x] speed up histogram calculation used in the contrast editor tool,
- [x] fix plotting histogram in the contrast editor tool,
- [x] fix range selection in the contrast editor tool,
- [x] add gamma curve to the histogram,
- [x] add saturated slider to the contrast editor tool,
- [x] add tests,
- [x] ready for review.

It requires https://github.com/hyperspy/hyperspy_gui_traitsui/pull/22
```python
import hyperspy.api as hs
import numpy as np
s = hs.signals.Signal2D(np.random.random(size=(10, 10, 100, 100)))*1000
s += hs.signals.Signal2D(np.arange(10*10*100*100).reshape((10, 10, 100, 100)))
s.plot()
s._plot.signal_plot.gui_adjust_contrast() # or press `h` on the figure.
```


francisco-dlp:
This PR hugely improves the contrast editor tool, great job!

Some comments/suggestions:

* For adjusting the contrast it is preferable to use the mpl's colormap normalizations because in this way the value of the intensity of the pixels (the one that we can get by hovering over the image) does not change. Therefore, I think that, for gamma adjustment, it will be preferable to use mpl's ``PowerNorm`` to skimage's ``adjust_gamma``.
* gamma adjustment is a sort of normalization and, therefore, it should be incompatible with e.g. log normalization. Instead, it could be added to the list of normalization choices as e.g. ``power``.
* It'll be nice to add the possibility to compute the number of bins automatically using the algorithms in ``get_histogram``.
* ``SymLog`` is a nice addition. However, I think that it'll be more useful if:
  1. It uses a divergent colormap to visually separate the + and - data
  2. Its parameters are adjustable via keywords and GUI widgets instead of hardwired
* It'll be cool to be able to change the norm using the GUI
* Now that you are at this, if would be nice to use mpl's colorbar ``extend`` keyword to make it explicit that the range has been clipped.
* Would it be possible to make the step of ``saturated_pixels`` smaller?

ericpre:
Thanks for the review and the good suggestions. I changed accordingly, except for:

> * Now that you are at this, if would be nice to use mpl's colorbar extend keyword to make it explicit that the range has been clipped.

There are some issues with extend and the log scale. I need to investigate a bit more and ask/reports to matplotlib.

> * Would it be possible to make the step of saturated_pixels smaller?

I can't find how to do it. I reduce the max to 10 instead as a workaround.

I am really sure about the following comments:

> * SymLog is a nice addition. However, I think that it'll be more useful if:
        i. It uses a divergent colormap to visually separate the + and - data

I implemented the suggestion but I am actually not really convinced about the behaviour when changing the norm with the GUIs.

francisco-dlp:
Some comments:

* I think that ``norm=auto`` doesn't make much sense in the GUI. Instead, if the ``norm="auto"` keyword has been passed, the GUI should display the current colormap as selected automatically.
* Updating the number of bins doesn't work unless we press `Apply`. Can the update happen automatically as for the other parameters?
* The ``auto`` number of bins could be implemented in a more user friendly way. For example, it could be a tick box that, when enabled, it automatically computes the number of bins onf selection and every time ``Apply`` is pressed. On its side there could be a dropdown widgets to select the number of bins estimation method. The number of bins computed could be displayed in read-only mode when ``auto`` is checked. By the way, using the ones available in ``get_histogram`` would be good for consistency, with the exceptions of ``blocks`` that requires a non-uniform axis. When ``auto`` is False, the number of bins box can then become editable and work interactively as discussed above.
* ``update_histogram`` call internally ``get_data``. That's "cheap" when working with data in memory, but it forces re-reading the data from disk in lazy mode, resulting in very slow histogram update.
*  When selecting ``log`` in the GUI, symlognorm is correctly used instead of log, but the colormap doesn't change to a divergent one, nor ones gets the symlognorm parameter controls in the GUI.
* When passing the ``norm="log"``, if the data is negative, a divergent colormap is used to display the image. However, when using the GUI to change norm to e.g. "linear", the colormap doesn't change back to a linear colormap.
* I think that the ``RdBu_r`` colormap is more suited for the purpose than ``coolwarm``, because the former is white at zero while the latter isn't.
* If we'll be checking if the current image has negative values, then it'll be good to change automatically to ``RdBu_r`` when that's the case with the linear norm too (unless the user selects a particular colormap).
* I feel slightly uncomfortable about the automatic selection of the ``SymLogNorm`` because I don't think that it is transparent to the user. Moreover, there is an alternative to it that is setting ``vmin>0``. Currently, the norm is set automatically to ``SymLogNorm`` even when ``vmin>0``. Wouldn't it be good not to switch automatically (adding ``SymLogNorm`` to the options) instead?

hyperspy/drawing/image.py
```diff
@@ -354,33 +357,53 @@ def format_coord(x, y):
...
 
         norm = copy.copy(self.norm)
-        if norm == 'log':
-            norm = LogNorm(vmin=self.vmin, vmax=self.vmax)
+        if norm == 'power' or (norm == 'auto' and self.gamma != 1.0):
+            # with auto norm, we use the power norm when gamma differs from its
+            # default value.
+            norm = PowerNorm(self.gamma, vmin=vmin, vmax=vmax)
+        elif norm == 'log':
+            if data.min() <= 0:
```
francisco-dlp: ``nanmin`` instead?

hyperspy/signal_tools.py
```diff
@@ -569,40 +663,226 @@ def update_span_selector_traits(self, *args, **kwargs):
...
+            values = ((xaxis-cmin)/(cmax-cmin)) ** self.gamma * max_hist
 
-    def update_histogram(self):
-        for patch in self.patches:
-            self.ax.patches.remove(patch)
-        self.plot_histogram()
+        return xaxis, values
+
+    @property
+    def negative_values_displayed(self):
```
francisco-dlp: Maybe make it private?

hyperspy/drawing/image.py
```diff
@@ -306,6 +303,8 @@ def plot(self, data_function_kwargs={}, **kwargs):
         self.figure.canvas.draw()
 
     def _add_colorbar(self):
+        # Bug extend='min' or extend='both' and power law norm
```
francisco-dlp, line 309: What's exactly the bug?

ericpre:
> * I think that norm=auto doesn't make much sense in the GUI. Instead, if the ``norm="auto"` keyword has been passed, the GUI should display the current colormap as selected automatically.

Yes, initially, I did it to match the `norm` parameter of the plot.

> * Updating the number of bins doesn't work unless we press Apply. Can the update happen automatically as for the other parameters?

I could not find a way to update it with pressing *enter* with `traitsui`, updating on typing the number I find it not great, particularly if the histogram takes time to calculate.

> * The auto number of bins could be implemented in a more user friendly way. For example, it could be a tick box that, when enabled, it automatically computes the number of bins onf selection and every time Apply is pressed. On its side there could be a dropdown widgets to select the number of bins estimation method. The number of bins computed could be displayed in read-only mode when auto is checked. By the way, using the ones available in get_histogram would be good for consistency, with the exceptions of blocks that requires a non-uniform axis. When auto is False, the number of bins box can then become editable and work interactively as discussed above.

Yes, I will see if I have enough time to do this, not really sure...

> * update_histogram call internally get_data. That's "cheap" when working with data in memory, but it forces re-reading the data from disk in lazy mode, resulting in very slow histogram update.

This is only from the displayed image, so this is already in memory.

> * I feel slightly uncomfortable about the automatic selection of the SymLogNorm because I don't think that it is transparent to the user. Moreover, there is an alternative to it that is setting vmin>0. Currently, the norm is set automatically to SymLogNorm even when vmin>0. Wouldn't it be good not to switch automatically (adding SymLogNorm to the options) instead?

Yes, good idea. I set it to the minimum positive values.

Regarding the colormap, I can not find a consistent behaviour which would be useful and not biased, so I would prefer not to set any colormap at least in this PR.

hyperspy/drawing/image.py
ericpre: It is difficult to reproduce in a minimal example, but it you use the `extend='both'` in the colorbar and changes gamma in the GUI, the ticks of the colorbar are going to be behave weirdly.

francisco-dlp:
> This is only from the displayed image, so this is already in memory.

I tested it with lazy data and it was much slower, so, maybe I got the diagnosis wrong, but I think that symptoms were right...

ericpre:
> > This is only from the displayed image, so this is already in memory.
> 
> I tested it with lazy data and it was much slower, so, maybe I got the diagnosis wrong, but I think that symptoms were right...

I don't know but I changed it so it makes a local copy of the displayed data.

francisco-dlp:
Looks super good to me!

ericpre:
Thanks for the review, the couple of iterations was very good and I am happy too with the outcome!

PR #2217, opened 2019-06-26
Fix #1712 and raise a warning when using map of ufuncs

### LINKED ISSUE #1712: Strange behaviour of `map` with `np.log`

ericpre:
- [x] Fix #1712,
- [x] raise a warning when using map of ufuncs,
- [x] ready for review.


ericpre:
I am not sure if it would be sensible to raise an error when calling map with a numpy ufunc because it would be more simple to apply the ufunc to the signal directly.

dnjohnstone (approved):
I think a warning is appropriate if it is only silly and doesn't break things (as this PR will cause to be true for numpy ufuncs). Whilst it might be pythonically simpler to use numpy directly here it might not be simpler to hyperspy users with limited python background. Approved despite failing tests on the basis that the Travis failure is in python 3.5 which will no longer be supported and the AppVeyor failure is not to do with this PR.

hyperspy/signal.py
```diff
@@ -3905,12 +3905,20 @@ def map(self, function,
         # If the function has an axis argument and the signal dimension is 1,
         # we suppose that it can operate on the full array and we don't
         # iterate over the coordinates.
+        fargs = []
```
francisco-dlp, line 3908: I would also print the exception in the log. Also, it would be good to add a test. Finally, shouldn't this go to Rnp?

ericpre: Should be all done now!

PR #2220, opened 2019-07-11
Drop python 3.5 support

ericpre:
Since a couple of weeks, hyperspy installation fails on appveyor and travis.
https://ci.appveyor.com/project/hyperspy/hyperspy/builds/25878228
https://travis-ci.org/hyperspy/hyperspy/builds/557181438

We could possibly find a workaround but I don't think it is worth it because it is going to be every time more complicated. Conda-forge is not building python 3.5 packaging since about 9 months.

It has been mentioned a couple of times that dropping python 3.5 will have the advantage to allow using f-string, which make the code mode readable, see for example https://github.com/hyperspy/start_jupyter_cm/pull/1#discussion_r301549351.

### Progress of the PR
- [x] Update setup.py and CI,
- [x] ready for review.


francisco-dlp:
Looks good. Just maybe add a line to CHANGES explaining that we're dropping Python <3.6 support.

PR #2222, opened 2019-07-12
Explain possible warning when importing hyperspy in the user guide.

ericpre:
Clarify the different warning which can be raised when importing hyperspy. They may be some more I would have missed to mentioned explicitly.

### Progress of the PR
- [x] udpate image of preferences GUI,
- [x] ready for review.


francisco-dlp (commented):
LGTM, I've just made some small suggestions.

francisco-dlp:
@ericpre, if you agree with my suggestions, feel free to merge.

PR #2240, opened 2019-08-06
Serpentine scan pattern for multifit

thomasaarholt:
# Description of the change
A common problem is that multifit fitting fails going from the end of one row to the beginning of the next (where the spectrum changes abruptly). While [SAMFIRE ](http://hyperspy.org/hyperspy-doc/current/user_guide/model.html#smart-adaptive-multi-dimensional-fitting-samfire)exists, a simpler solution to this problem is to iterate through the signal indices in a serpentine pattern, as seen on the image below. 

- Added a ["serpentine" generator equivalent of np.ndindex](https://stackoverflow.com/questions/57366966/np-ndindex-but-where-every-new-index-touches-the-previous).
- Added a iterpath option to the axes manager `s.axes_manager._iterpath` (default `serpentine`, alternative `flyback`), so that (`for index in s.axes_manager...`) can operate in serpentine or the old flyback manner.
- Added a multifit `iterpath` option to let multifit take use of the new serpentine iterpath

The image shows the scanning direction with increasing grey value. The code supports ndimensional arrays with navigation dimensions larger than 2 as well.
image

### Progress of the PR
- [x] serpentine generator implemented
- [x] axes_manager can now choose `serpentine` or `flyback` iterpaths
- [x] multifit argument support
- [x] add tests,
- [x] ready for review.

I've [created some noisy test data here](https://gist.github.com/thomasaarholt/5d373d94f1350e66083582dbf2d5842e), which results in the following result when fitted by the normal multifit routine as well as the serpentine (formerly "zigzag") one:
image


jat255:
I have not had a chance to review the code yet, but in my experience, I've seen this scan pattern in electron microscopes usually called "serpentine", rather than "zigzag". I don't have much of a preference and this may be a US/UK language distinction, perhaps. Unless there are other objections, I would probably recommend we use the term "serpentine" though, since it is what they typically use when talking about EM scan generators that do the same pattern.

thomasaarholt:
Perfect! I struggled to find a way to describe the scan pattern when I was searching for a way to generate the indices. I actually referred to it as "snake-style" in the stackoverflow link above. Serpentine it is!

I'll change the label in the PR in the next couple of days, current at a conference (not M&M).

jat255:
Great! I'm also thinking it might be better to generalize the argument in the event we might want to support future alternate path strategies. So rather than `zigzag=True`, something like `iterpath='serpentine'` or some other terminology. This would prevent an API change if we needed to add additional paths later on. I would call the default `'flyback'` (which is also the term used by scan generators).

It may also be worth making the serpentine strategy the default. I'm having a hard time envisioning a case where it would perform worse than the current default.

magnunor:
Good idea @thomasaarholt! Making it the default also sounds good.

In addition, it would be nice having something like your test signal in `hs.datasets.artificial_data`.

thomasaarholt:
@magnunor Excellent suggestion! I'll open a PR for that. I'll try to add a background to it as well.

I will follow @jat255 suggestion of an `iterpath='serpentine'` argument and change the underlying axes_manager properties accordingly.

thomasaarholt:
I have implemented @jat255 and @magnunor's comments, but I'm waiting with changing the docs in case we don't want serpentine is default.

thomasaarholt:
The two failing tests appear unrelated. It is a bit odd that nothing failed yesterday, though.

hyperspy/model.py
```diff
@@ -1251,7 +1251,7 @@ def fit(self, fitter="leastsq", method='ls', grad=False,
 
     def multifit(self, mask=None, fetch_only_fixed=False,
                  autosave=False, autosave_every=10, show_progressbar=None,
-                 interactive_plot=False, **kwargs):
+                 interactive_plot=False, iterpath='serpentine', **kwargs):
```
francisco-dlp: We can only make ``serpentine`` the default ``iterpath`` method in 2.0. Otherwise, the result of multifitting would be unexpectedly different when using the next point release of HyperSpy.

francisco-dlp: Now that you are at this, it would be cool if ``iterpath`` could take an iterator in addition to ``serpentine`` and ``flyback``.

doc/user_guide/model.rst
```diff
@@ -1051,6 +1051,21 @@ reaching the end of the dataset.
     :py:meth:`~.model.BaseModel.multifit` to perform the fit over the entire
     spectrum image.
 
+A common problem is that the fitting fails going from the end of one row to the
```
francisco-dlp: This would make more sense to the reader if the other "scanning pattern" was also described. Also it would be good to add a ``..versionadded::`` directive.

thomasaarholt:
@francisco-dlp I agree that custom iterpaths would be cool, but I also recognise that it would mean a bunch more work. I also believe that serpentine would be a significant improvement to the fitting algorithm on its own.

I'm happy to revisit this in the future, but I would _much_ prefer doing PRs in bite-sized chunks. Are you happy with this PR if I leave `flyback` as  the default approach and fully document both iterpaths? I will also include a comment about making serpentine the default iterpath for 2.0.

francisco-dlp:
Sounds good! In addition to the comment you suggest you could add a deprecation warning that triggers when using ``multifit`` with the default ``flyback`` argument, explaining that the default will change in 2.0.

thomasaarholt:
I've added the changes, just letting tests run now.

thomasaarholt:
Should be good to go. My code doesn't touch the failing tests.

ericpre:
From a quick look, the 2 test failures are related to a change in the behaviour of iterating over the axes manager.
It is always to see that some tests are catching unexpected consequences! This is caught by the plotting test, which means that the coverage of the test of the axes manager is not good enough.

thomasaarholt:
You were right, @ericpre. I'm really curious how you were able to see that so quickly. Since Travis showed the tests passing on the pip-build and failing on the conda build, I was feeling confident it was due to some bug introduced through a new numpy/mpl version.

The test failure was due to differing behaviour between flyback and serpentine when the axes_manager was in "serpentine mode". I had missed a line of code when I set flyback to be default again.

When `hs.plot.plot_images` takes a list `of lists` of images, (as it does in [line 150 in the test](https://github.com/hyperspy/hyperspy/blob/55b233759e2acbd80875379affcb5e1030ed7c11/hyperspy/tests/drawing/test_plot_signal2d.py#L150)), the plot function iterates through the axes manager and plots them with their indices as name. Since the iteration is different, the images end up in different order.

The failing test is fixed by my latest commit. If we set serpentine as default, we can consider either updating the tests or updating the figures to match the new axes_manager iterpath.

That said, we may want to keep the flyback iterpath on the axes_manager as standard, and only change it for the multifit. What do you think?

thomasaarholt:
Ready for new review.

ericpre:
This looks fine to me, @francisco-dlp, are you happy with this PR? Shall we leave https://github.com/hyperspy/hyperspy/pull/2240#discussion_r320255930 for later?

dnjohnstone:
> This looks fine to me, @francisco-dlp, are you happy with this PR? Shall we leave [#2240 (comment)](https://github.com/hyperspy/hyperspy/pull/2240#discussion_r320255930) for later?

I think the user guide needs to be updated to account for the earlier review from @francisco-dlp - to recap - it needs a .._version_added flag and the text needs updating to explain the "conventional" iterpath and why this is better for hyperspectral mapping data. @thomasaarholt can you do that?

Other than that - I agree that this looks good and that we should merge this as a clear and self contained step forwards.

Can we then also raise an issue at the same time for a "feature request" to "add custom iterpaths" to account for that suggestion from @francisco-dlp and ensure it is not forgotten.

thomasaarholt:
Sure, will do. @francisco-dlp what is the next version (that this will be merged into). 1.5.3? 1.6.0?

dnjohnstone:
> Sure, will do. @francisco-dlp what is the next version (that this will be merged into). 1.5.3? 1.6.0?

Semantic versioning will require this to be 1.6.0 https://semver.org/

The minor version "MUST be incremented if new, backwards compatible functionality is introduced to the public API."

thomasaarholt:
I've added the changes. Should be good to go now.

ericpre:
Thanks @thomasaarholt.

PR #2245, opened 2019-08-30
bug-fix_bcf_v2_1

sem-geologist:
### Description of the change
bcf files sometimes have automatic or fixed energy width mode.
Prior code implementation was assuming that bcf XML part always have
width window of xray line energy in the XML. That is causing error when bcf is
set to use automatic width function, and energy Width is absent from xml.
Moreover, energy width was not exposed to hyperspy metadata (only to original_metadata).
So there is no loss with getting rid of that one line, and it is most simple solution to the problem.

### Progress of the PR
- [x] Change implemented,
- [x] check tests,
- [x] ready for review.

### DIFF
--- a/hyperspy/io_plugins/bruker.py
+++ b/hyperspy/io_plugins/bruker.py
@@ -727,8 +727,7 @@ class HyperHeader(object):
                     "./ClassInstance[@Type='TRTSpectrumRegion']"):
                 tmp_d = dictionarize(j)
                 self.elements[tmp_d['XmlClassName']] = {'line': tmp_d['Line'],
-                                                        'energy': tmp_d['Energy'],
-                                                        'width': tmp_d['Width']}
+                                                        'energy': tmp_d['Energy']}
         except AttributeError:
             _logger.info('no element selection present in the spectra..')
 


ericpre:
> bcf files sometimes have automatic or fixed energy width mode.

What does it mean more precisely?
Is it not worth still trying to import this information in `original_metadata`, when possible?

The tests suite have been fixed.

francisco-dlp:
@sem-geologist, we are planning to release v1.5.2 very soon. We can wait for this if you think that it can be ready in a few days.

sem-geologist:
@ericpre , I don't think it is important to get it into metadata (it anyway was not pulling everything). What is it exactly? in bcf xml part there is node with element selection - it is actually not metadata at all, as element selection does not influence the hyperspectra anyhow (differently than i.e. acceleration voltage, beam current, counting times and etc. - those indeed are the real metadata). Element selection (and corresponding fraction of the spectra to represent the element xray counts) can be done and redone after hyperspectra acquisition. It can be completely ignored. The only usability from that is convenience feature, so that hyperspy would be able to use the selected x-ray lines (or elements).  The element line XML nodes additionally to the atom number and line, contains other information such as an energy (which is pointless to store in the first place - it is supposed to be universal), window width or node noting that it is in Auto mode (how this auto mode works? it is black box). So for convenience it is enough to get from that node only atom number and x-ray line ( if available). That then is mapped to hyperspy metadata. for width hyperspy have its own mechanisms.

PR #2246, opened 2019-09-02
Problem to run the latest release

MartialD:
Hi all,

### Description of the change
- Remove the 'f' in front of text comments on these files to be able to run hyperspy package.
- 'from scipy import sparse' instead of 'import sparse'
- 'sparse.coo_matrix' instead of 'sparse.COO'

### DIFF
--- a/hyperspy/drawing/image.py
+++ b/hyperspy/drawing/image.py
@@ -375,8 +375,8 @@ class ImagePlot(BlittedFigure):
                 if col >= 0 and row >= 0:
                     z = data[row, col]
                     if np.isfinite(z):
-                        return f'x={x:1.4g}, y={y:1.4g}, intensity={z:1.4g}'
-                return f'x={x:1.4g}, y={y:1.4g}'
+                        return 'x={x:1.4g}, y={y:1.4g}, intensity={z:1.4g}'
+                return 'x={x:1.4g}, y={y:1.4g}'
             self.ax.format_coord = format_coord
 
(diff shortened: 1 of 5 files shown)


ericpre:
@MartialD: these are dependencies update issues:

> - Remove the 'f' in front of text comments on these files to be able to run hyperspy package.

HyperSpy does not support python 3.5 since the last release and it means that we started to use [f-string](https://docs.python.org/3/whatsnew/3.6.html#whatsnew36-pep498), which has been introduced in python 3.6.


> - 'from scipy import sparse' instead of 'import sparse'
> - 'sparse.coo_matrix' instead of 'sparse.COO'

HyperSpy uses the [`sparse` library](https://sparse.pydata.org/) instead of the `sparse` submodule of `scipy` and this should be working. (By the way, the last release of sparse seems to break )

Normally, if hyperspy have been installed using pip or conda, you should not have these issues and I would guess that you are running hyperspy [in a development environment](http://hyperspy.org/hyperspy-doc/current/user_guide/install.html#development-version). Unfortunately, under this condition, there is no straightforward way to check for dependency update when checking out a more recent development branch, the only way is to re-run the `pip install -e .`. 
Can you please confirm this is the case? If not, it seems that there is something wrong somewhere.

That being said, since a few weeks, recent release of upstream libraries (numpy>=1.17, in particular...) breaks some functionalities of hyperspy but if you installed it using conda or the hyperspy bundle, it should be fine.

ericpre:
Please reopen if there does not solve the issue.

Issue #2244, opened 2019-08-30
errors loading BCF files


jeinsle:
I have been trying to import a series of EDS maps generated with Esprit version 2.1.  however, when I try to load this file (https://www.dropbox.com/s/iqmclr7fn871pek/Mapping_%281%2C1%29.bcf?dl=0) I get this error:
 ```python
KeyError                                  Traceback (most recent call last)
<ipython-input-6-ad1ad3391a64> in <module>
     ----> 1 b=hs.load('20190723_bon_accord_BCF_data/Mapping_(1,1).bcf')
     ~\AppData\Local\conda\conda\envs\hyperspy\lib\site-packages\hyperspy\io.py in 
 load(filenames, signal_type, stack, stack_axis, new_axis_name, lazy, convert_units, **kwds)
     279             objects = [load_single_file(filename, lazy=lazy,
    280                                         **kwds)
 --> 281                        for filename in filenames]
     282 
    283         if len(objects) == 1:`

 ~\AppData\Local\conda\conda\envs\hyperspy\lib\site-packages\hyperspy\io.py in 
 <listcomp>(.0)
 279             objects = [load_single_file(filename, lazy=lazy,
 280                                         **kwds)
 --> 281                        for filename in filenames]
     282 
     283         if len(objects) == 1:
 
 ~\AppData\Local\conda\conda\envs\hyperspy\lib\site-packages\hyperspy\io.py in 
 load_single_file(filename, **kwds)
     316     else:
     317         reader = io_plugins[i]
 --> 318         return load_with_reader(filename=filename, reader=reader, **kwds)
     319 
     320 
 
 ~\AppData\Local\conda\conda\envs\hyperspy\lib\site-packages\hyperspy\io.py in 
 load_with_reader(filename, reader, signal_type, convert_units, **kwds)
     323     lazy = kwds.get('lazy', False)
     324     file_data_list = reader.file_reader(filename,
 --> 325                                         **kwds)
     326     objects = []
     327 
 
 ~\AppData\Local\conda\conda\envs\hyperspy\lib\site- 
    packages\hyperspy\io_plugins\bruker.py in file_reader(filename, *args, **kwds)
        1189     ext = splitext(filename)[1][1:]
        1190     if ext == 'bcf':
     -> 1191         return bcf_reader(filename, *args, **kwds)
    1192     elif ext == 'spx':
   1193         return spx_reader(filename, *args, **kwds)
 
[...]

PR #2250, opened 2019-09-05
Fix real time bcf for large spectrum image on windows

ericpre:
Fix the second part of #2244: with the numpy overflow warning. I could reproduce the issue on windows using the file provided by @jeinsle (https://www.dropbox.com/s/bqi8l4idppg45cl/E17_017.bcf?dl=0). The issue occurs on windows but not linux system.

### Progress of the PR
- [x] change dtype to float64,
- [x] ready for review.

### DIFF
--- a/hyperspy/io_plugins/bruker.py
+++ b/hyperspy/io_plugins/bruker.py
@@ -821,7 +821,7 @@ class HyperHeader(object):
         """calculate and return real time for whole hypermap
         in seconds
         """
-        line_cnt_sum = np.sum(self.line_counter).astype(np.float64)
+        line_cnt_sum = np.sum(self.line_counter, dtype=np.float64)
         line_avg = self.dsp_metadata['LineAverage']
         pix_avg = self.dsp_metadata['PixelAverage']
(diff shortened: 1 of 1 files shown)


hyperspy/io_plugins/bruker.py
```diff
@@ -821,12 +821,12 @@ def calc_real_time(self):
         """calculate and return real time for whole hypermap
         in seconds
         """
-        line_cnt_sum = np.sum(self.line_counter)
+        line_cnt_sum = np.sum(self.line_counter).astype(np.float64)
```
francisco-dlp: I assume that ``astype`` is needed because ``line_counter`` was float32 to start with. Wouldn't it be better to use ``np.sum(self.line_counter, dtype=np.float64)`` instead then?

ericpre: Yes, indeed!
The original type was int32 and the reason why it worked on linux and not on windows is the following: 

on linux
```python
>>> np.array([0]).dtype
dtype('int64')
```

on windows:
```python
>>> np.array([0]).dtype
dtype('int32')
```

PR #2254, opened 2019-09-10
Add colormap setting to preferences

ericpre:
### Progress of the PR
- [x] Add colormap setting for navigator and signal plot to preferences,
- [x] remove setting current matplotlib rcParams `image.cmap` when importing hyperspy,
- [x] add tests,
- [x] ready for review.

```python
import hyperspy.api as hs
import numpy as np
s = hs.signals.Signal1D(np.arange(10*10*10).reshape(10, 10, 10))
hs.preferences.Plot.cmap_navigator = 'viridis'
s.plot()
```


francisco-dlp:
It looks good. Just a small less-is-more suggestion: Wouldn't it be better to add just a single colormap item to preference that affects all the colormaps (including those in ``plot_images``)? That would make it easier to switch to a given colormap for all purposes (most likely the typical case). If somebody wants a different colormap for the navigator, he can then use the relevant ``plot`` keyword argument.

ericpre:
Since the navigator and the signal dimension display "different kind" of images, I still think that it would be sensible to have different default value for the colormap.

jlaehne:
Nice feature, LGTM.

dnjohnstone:
I agree with @ericpre that most of the time I would personally want a different cmap for navigation and signal. E.g. for scanning diffraction data I often like to see colourful diffraction patterns but greyscale images for some reason.

Using the preferences for plot_images also could easily be handled in a separate PR, but also I feel like this feature is much more useful for the normal plot and with a user hat on I'm much more happy to set it manually in cases when I use plot_image.

@francisco-dlp is there any reason why this therefore can't be merged as is?

hyperspy/drawing/utils.py
```diff
@@ -483,11 +483,11 @@ def plot_images(images,
         image.
         If any signal shape is not suitable, a ValueError will be raised.
     cmap : matplotlib colormap, list, or ``'mpl_colors'``, *optional*
-        The colormap used for the images, by default read from ``pyplot``.
-        A list of colormaps can also be provided, and the images will
-        cycle through them. Optionally, the value ``'mpl_colors'`` will
-        cause the cmap to loop through the default ``matplotlib``
-        colors (to match with the default output of the
+        The colormap used for the images, by default reuse the preferences 
+        setting ``color map signal``. A list of colormaps can also be provided,
```
tjof2: Rewrite it to e.g.:

> The colormap used for the images. Default is to use the preferences setting ``color map signal``.

Is "color map signal" correct - i.e. with no underscores, and not "cmap_signal"?

ericpre: With "color map signal", I am referring to the name of the setting as displayed in the GUI. I have tried to reworded it slightly to make it more clear.

tjof2 (approved):

hyperspy/drawing/utils.py
tjof2: Looks better!

ericpre:
I tagged this PR as bug fix because currently hyperspy is setting the matplotlib 'image.cmap' rcParams to "gray" while it should not be the case.

Issue #2251, opened 2019-09-06
Release v1.6


francisco-dlp:
I think that it is time to start planning the next minor release that is going to be the last one of the 1 series i.e. the last one before the split.

The split will make it more difficult to contribute code that you have started developing based on the current HyperSpy structure because chances are that at least part of your code will then belong to a different package. Therefore, I suggest we focus our efforts on finishing and merging as many pending contributions as possible, rather than in making brand new contributions.

Since we will be breaking the API, this is also a good time to streamline the syntax by e.g. pruning methods are no longer relevant, changing the default value of arguments when that leads to a better outcome, renaming functions etc. All those API changes should come with a deprecation warning in v1.6 so that users have time to adapt their scripts. I propose to list and discuss in this issue those API changes in a checklist.

# PRs

- [ ] ~#1927~ 
- [x] #2035
- [x] #1497
- [x] #2352 
- [x] #2358
- [x] #2184
- [x] #2262
- [x] #2272 (and its gui counterparts)
- [x] #2289
- [x] #2307 (and its gui counterparts)
- [x] #2350
- [x] #2351
- [ ] ~#1462~
- [x] #2203 
- [ ] ~#2172~
- [x] #1159
- [x] #2383

jlaehne:
I think the support for non linear axis #1927 is a clear candidate for v1.6. I'll be happy to contribute to the implementation.

jlaehne:
The following `Components` are so far not implemented as `Expression`. For which does it make sense to aim at a conversion, is it feasible or even desirable for all (some are quite complex, see the mentioned requirements)?
- `arctan` (ifelse) -> PR #2395
- `eels_cl_edge` (complex!)
- `eels_vignetting` (complex)
- `heaviside` (where x2) -> PR #2262
- `offset` (prevented by https://github.com/sympy/sympy/issues/5642)
- `pes_core_lineshape` (complex!)
- `pes_see` (ifelse, convolve, where)
- `scalable_fixed_pattern` (if-else x2)
- `voigt` (complex - see proposal below) -> merged

It would also solve the issue that  'function_nd' is not defined for some of these functions, see #2225

francisco-dlp:
Good point, indeed the best way to fix #2225 is to implement as many components as possible using ``Expression``.

[...]

PR #1791, opened 2017-12-19
Interactive calibration for Signal2D

magnunor:
This pull requests implements a interactive way of calibrating the signal dimensions in a Signal2D object. It is based on the calibration functionality for Signal1D.

Needs to be used with https://github.com/hyperspy/hyperspy_gui_traitsui/pull/3, or https://github.com/hyperspy/hyperspy_gui_ipywidgets/pull/2

### Progress of the PR
- [x] Initial implementation
- [x] Improve implementation
- [x] Add GUI elements to hyperspy_gui_ipywidgets
- [x] Add docstring to signal2d.calibrate
- [x] Update user guide
- [x] Add tests
- [x] Ready for review.

To test this, grab this branch + the traitsui pull request: https://github.com/hyperspy/hyperspy_gui_traitsui/pull/3.

```python
>>> import hyperspy.api as hs
>>> import numpy as np
>>> s = hs.signals.Signal2D(np.random.random((100, 100)))
>>> s.calibrate(toolkit='traitsui')
```
(This traitsui GUI image is slightly outdated now, since I did some minor changes in the most recent commit).
calibrate_signal2d

```python
>>> import hyperspy.api as hs
>>> import numpy as np
>>> s = hs.signals.Signal2D(np.random.random((100, 100)))
>>> s.calibrate(toolkit='ipywidgets')
```

ipywidget_calibrate2d

PR #2256, opened 2019-09-19
Interactive crop_image

thomasaarholt:
### Description of the change
I've written a method for interactive cropping that plots a signal (unless it already is plotted), and then displays a rectangle centered a quarter of the way into the image. I find it rather cumbersome if I just want to get a slightly smaller image to have to plot it, and then read off the coordinates number by number and then crop using `s.isig` or `s.crop_image`.
image

We already have two methods that are related:
`s.crop()` and `s.crop_image`. I don't really see why we have these, since .isig and.inav do the same thing in a general manner, unless their intention is to save memory in some manner.

### Progress of the PR
- [x] Change implemented
- [x] update docstring
- [ ] update user guide
- [ ] add tests,
- [ ] ready for review.

```python
#%matplotlib notebook | widget | qt # your favourite backend here
import hyperspy.api as hs
s = hs.datasets.example_signals.object_hologram()
s.crop_image_interactive()
```

### DIFF
--- a/hyperspy/_signals/signal2d.py
+++ b/hyperspy/_signals/signal2d.py
@@ -28,6 +28,8 @@ from skimage.feature.register_translation import _upsampled_dft
 from hyperspy.defaults_parser import preferences
 from hyperspy.external.progressbar import progressbar
 from hyperspy.misc.math_tools import symmetrize, antisymmetrize
+from hyperspy.misc.signal_tools import _get_crop_range
+from hyperspy.roi import RectangularROI
 from hyperspy.signal import BaseSignal
 from hyperspy._signals.lazy import LazySignal
 from hyperspy._signals.common_signal2d import CommonSignal2D
@@ -703,7 +705,7 @@ class Signal2D(BaseSignal, CommonSignal2D):
         See also:
         ---------
         crop
-
+        crop_image_interactive
         """
         self._check_signal_dimension_equals_two()
         self.crop(self.axes_manager.signal_axes[1].index_in_axes_manager,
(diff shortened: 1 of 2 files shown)


francisco-dlp:
> I don't really see why we have these, since .isig and.inav do the same thing in a general manner, unless their intention is to save memory in some manner.

The reason is that the `crop` method predates `isig` i.e. there is no good reason at all for this. Maybe something to add to the to-deprecate list in #2251?

jlaehne:
Probably it is a question of how my gui is set up, but I do not get a button to apply the cropping.

Also, is it restricted to certain types of signals? If I load an arbitrary signal image, I get `AttributeError: 'Signal1D' object has no attribute 'crop_image_interactive'`

thomasaarholt:
@jlaehne This is a 2D signal crop method which will only work on the signal dimension. If you're doing this on a spectrum image, try doing (`s.T.crop_image_interactive()`).

jlaehne:
OK, but wouldn't it make sense to implement such a interactive cropping tool in a more general way that could work also on e.g. the navigation image of a 1D signal spectral map, ...?

thomasaarholt:
Yes, absolutely. I initially stepped away from this because A) `crop` was taken and B) I just wanted a crop_image method.

I can rewrite it to handle 1D and 2D signals, as well as try to handle signal/navigation as input parameters. I guess I'll then plot `s.T.sum()` and then use the respective 1D/2D crop feature, and then pass those coordinates to `isig`/`inav`. 

I'll then rename it `s.interactice_crop()`. 

Sound OK?

ericpre:
I think before considering adding such feature, we should first implement adding a roi interactively, as mentioned in other PRs (for example https://github.com/hyperspy/hyperspy/pull/1785). There are a few PRs which requires this (https://github.com/hyperspy/hyperspy/pull/1791, https://github.com/hyperspy/hyperspy/pull/1785) and once this is done properly, it should be easy to implement the feature of this PR.
There is some discussion on this in https://github.com/hyperspy/hyperspy/pull/2182.

Issue #2089, opened 2018-11-06
The take off angle function does not work for double tilt sample holder


CorentinLG:
this function was implemented for single tilt holder only.
Requires clarification of how each tilt angle is documented in the metadata and an update of the geometrical calculation.

thomasaarholt:
For future reference, would you clarify for what sort of signal and calculation you are referring to?

CorentinLG:
in utils.py, there is a function called 'take_off_angle', which is used to calculate the angle at which X-rays are leaving the sample going to the EDS detector. It is a function of the elevation angle of the detector, which depends on each microscope, but also of the sample holder tilt.

The take_off_angle function is used by other functions, and at present the tilt angle value is taken from the metadata. 
However, it works only for single tilt holder.  For double tilt holder, the metadata is structured into sub section (tilt alpha and tilt beta), and the take_off_angle function then raises an error.

PR #2258, opened 2019-09-23
Absorption correction for kfactors, z-factors and cross sections

k8macarthur:
## Description of the change
This pull request aims to add basics absorption correction (assuming one detector and planar thin film sample) to all three quantification methods. This is with the hope that once the framework is in place it can be expanded to more intelligent methods of absorption correction in the future. 
- etc.

### Still requiring
- [x] Tests
- [x] Documentation update
- [x] Maybe add some sort of counter so people can see the number of iterations being carried out, at least to show that the programe hasn't frozen.

For both zeta_factors and cross_sections this addition required minimal change to the functions. For kfactors the use must now provide either a number or 2D thickness profile of the sample such that a mass thickness can be calculated. 

zeta factor and cross section outputs remain unchanged. The kfactor quantification will also output the calculated mass_thickness if absorption correction is included.

My default absorption correction is false so the new functionality will only be used if it is specified. 

I have also added a chatter line to specify how many interations have been used but I'm not sure if this is useful or not. 

NB: I know this is not a very well polished function but I'm at the stage where a second opinion would be useful, hence the pull request. #800


k8macarthur:
Also I am aware that my binary files have been included in this pull request. i'll remove them as soon as I work out how.

k8macarthur:
@dnjohnstone  @siriagus  @CorrentinLG might all have interest in this.

k8macarthur:
I don't understand the one remaining tests error, I would be grateful if anyone could help suggest where this one is coming from. I've traced it down to the kfactor functions but can't work out the mistake...
```
..\..\_signals\eds_tem.py:437: in quantification
    absorption_correction=abs_corr)
..\..\misc\eds\utils.py:393: in quantification_cliff_lorimer
    ref_index, ref_index2)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

self = dask.array<reshape, shape=(2, 4), dtype=float64, chunksize=(1, 4)>
key = (slice(None, None, None), 0), value = array([0.7729221, 0.2270779])

    def __setitem__(self, key, value):
        from .routines import where
        if isinstance(key, Array):
            if isinstance(value, Array) and value.ndim > 1:
                raise ValueError('boolean index array should have 1 dimension')
            y = where(key, value, self)
            self.dtype = y.dtype
            self.dask = y.dask
            self.name = y.name
            self._chunks = y.chunks
            return self
        else:
            raise NotImplementedError("Item assignment with %s not supported"
>                                     % type(key))
E           NotImplementedError: Item assignment with <class 'tuple'> not supported
```

hyperspy/_signals/eds_tem.py
```diff
@@ -353,37 +380,159 @@ def quantification(self,
...
-                composition.data, kfactors=factors,
-                mask=navigation_mask) * 100.
+            int_stack = utils.stack(intensities)
+
+            comp_old = utils.stack(intensities)
+            comp_old.data = np.zeros_like(comp_old.data)
+
+            abs_corr = None # initial
+
+            while True:
```
thomasaarholt: You can use tqdm in a while loop for a progress bar. See here for instructions: https://stackoverflow.com/questions/45808140/using-tqdm-progress-bar-in-a-while-loop

k8macarthur: Thanks Tom, I've tried to add that now. Let me know if what you think.

thomasaarholt:
Do you have a minimum working example I can use to test the code?

k8macarthur:
I've been using the default signal set up in the test suite. 

```
s = hs.signals.EDSTEMSpectrum(np.ones([2, 2, 1024]))
energy_axis = s.axes_manager.signal_axes[0]
energy_axis.scale = 1e-2
energy_axis.units = 'keV'
energy_axis.name = "Energy"
s.set_microscope_parameters(beam_energy=200,
                            live_time=2.5, tilt_stage=0.0,
                            azimuth_angle=0, elevation_angle=35,
                            energy_resolution_MnKa=130,
                            beam_current=0.05)
elements = ['Al', 'Zn']
xray_lines = ['Al_Ka', 'Zn_Ka']
intensities = [300, 500]
for i, xray_line in enumerate(xray_lines):
    gauss = Gaussian()
    line_energy, FWHM = s._get_line_energy(xray_line, FWHM_MnKa='auto')
    gauss.centre.value = line_energy
    gauss.A.value = intensities[i]
    gauss.sigma.value = FWHM
    s.data[:] += gauss.function(energy_axis.axis)

s.set_elements(elements)
s.add_lines(xray_lines)
s.axes_manager[0].scale = 0.5
s.axes_manager[1].scale = 0.5
s.axes_manager[0].units = 'nm'
s.axes_manager[1].units = 'nm'

#All the above is set up.
#A mimum example is:
inten = s.get_lines_intensity()
results = s.quantification(inten, 'CL', 
                            [1, 2.0009344042], 
                            composition_units = 'weight',
                            absorption_correction=True, 
                            thickness = 100
                           )
```

hyperspy/_signals/eds_tem.py
```diff
@@ -31,6 +31,7 @@
 from hyperspy.misc.eds import utils as utils_eds
 from hyperspy.ui_registry import add_gui_method, DISPLAY_DT, TOOLKIT_DT
 from hyperspy.misc.elements import elements as elements_db
+from tqdm import tqdm
```
ericpre: Please change to `from hyperspy.external.progressbar import progressbar` from https://github.com/hyperspy/hyperspy/blob/74d2ffe97b3a4873536701b27383670dc5f7c0eb/hyperspy/external/progressbar.py#L23

hyperspy/_signals/eds_tem.py
```diff
@@ -387,6 +388,7 @@ def quantification(self,
 
         xray_lines = [intensity.metadata.Sample.xray_lines[0] for intensity in intensities]
         it = 0
+        pbar = tqdm(total=max_iterations+1)
```
ericpre: Following the change above, replace this line by `pbar = progressbar(total=max_iterations+1)`

hyperspy/_signals/eds_tem.py
```diff
@@ -353,37 +381,166 @@ def quantification(self,
...
+
         composition = utils.stack(intensities, lazy=False)
+        if take_off_angle == 'auto':
+            toa = self.get_take_off_angle()
+        else:
+            toa = take_off_angle
+
+        #determining illumination area for cross sections quantification.
+        if method == 'cross_section':
+            if probe_area == 'auto':
```
ericpre, line 410: Would it be more sensible to make a function to calculate the `probe_area` and use it in `_get_dose` too?

hyperspy/_signals/eds_tem.py
```diff
@@ -353,37 +381,166 @@ def quantification(self,
...
+                parameters = self.metadata.Acquisition_instrument.TEM
+                if probe_area in parameters:
+                    area = parameters.TEM.probe_area
+                else:
+                    if (self.axes_manager.navigation_dimension > 2 and
+                            navigation_axes is None):
```
ericpre: `navigation_axes` is not defined yet in this function.

hyperspy/_signals/eds_tem.py
```diff
@@ -287,30 +289,53 @@ def quantification(self,
...
         factors: list of float
             The list of kfactors, zeta-factors or cross sections in same order
             as intensities. Note that intensities provided by Hyperspy are
             sorted by the alphabetical order of the X-ray lines.
             eg. factors =[0.982, 1.32, 1.60] for ['Al_Ka', 'Cr_Ka', 'Ni_Ka'].
-        composition_units: 'weight' or 'atomic'
+        composition_units: {'atomic', 'atomic'}
             The quantification returns the composition in atomic percent by
             default, but can also return weight percent if specified.
+        absorption_correction: {'False', True} bool
```
ericpre: `absorption_correction: bool` is enough.

hyperspy/_signals/eds_tem.py
```diff
@@ -353,37 +381,166 @@ def quantification(self,
...
+
+            abs_corr = None # initial
+
+            while True:
+                results = utils_eds.quantification_cliff_lorimer(
+                    int_stack.data,
+                    kfactors=factors,
+                    absorption_correction=abs_corr)
+                composition.data = results * 100.
+                if absorption_correction != False:
```
ericpre: `absorption_corrrection` is supposed to be a boolean so `if absorption_corrrection:` should be enough.

hyperspy/_signals/eds_tem.py
```diff
@@ -287,30 +289,53 @@ def quantification(self,
...
-                       composition_units='atomic',
-                       navigation_mask=1.0,
-                       closing=True,
-                       plot_result=False,
+                       composition_units = 'atomic',
+                       absorption_correction = False,
+                       tilt_stage = 0,
+                       take_off_angle = 'auto',
+                       thickness = None,
+                       conv_crit = 0.5,
```
ericpre: `convergence_criterion` to make it explicit?

hyperspy/_signals/eds_tem.py
```diff
@@ -287,30 +289,53 @@ def quantification(self,
...
+                       navigation_mask = 1.0,
+                       closing = True,
+                       plot_result = False,
+                       probe_area = 'auto',
+                       max_iterations = 30,
```
ericpre, line 304: Add `mass_thickness='auto'`.

hyperspy/_signals/eds_tem.py
```diff
@@ -353,37 +381,166 @@ def quantification(self,
...
+                    absorption_correction=abs_corr)
+                composition.data = results * 100.
+                if absorption_correction != False:
+                    if thickness is not None:
+                        mass_thickness = intensities[0].deepcopy()
+                        mass_thickness.data = self.CL_get_mass_thickness(composition.split(),
+                                                                    thickness)
+                        mass_thickness.metadata.General.title = 'Mass thickness'
+                    else:
+                        warnings.warn('Thickness is required for absorption'\
```
ericpre: Need a single space before the end of the string and the "\" is not necessary:
`'Thickness is required for absorption '`

hyperspy/_signals/eds_tem.py
```diff
@@ -353,37 +381,166 @@ def quantification(self,
...
+                                    'will contain no correction for absorption.')
+
+                res_max = np.max((composition - comp_old).data)
+                it += 1
+                pbar.update(1)
+                if not absorption_correction or res_max < (conv_crit/100):
+                    break
+                    pbar.close()
+                elif it >= max_iterations:
+                    raise Exception('Absorption correction failed as solution '
```
ericpre: Since the last few release, we started to use f-string, which is more readable (note the f at the beginning of the string):
```python
                    raise Exception('Absorption correction failed as solution '
                                    f'did not converge after {max_iterations} '
                                    'iterations')
```

hyperspy/_signals/eds_tem.py
```diff
@@ -353,37 +381,166 @@ def quantification(self,
...
+                                                       toa, probe_area)
+
             composition.data = results[0] * 100
             number_of_atoms = composition._deepcopy_with_new_data(results[1])
             number_of_atoms = number_of_atoms.split()
         else:
-            raise ValueError("Please specify method for quantification, "
-                             "as 'CL', 'zeta' or 'cross_section'.")
+            raise ValueError('Please specify method for quantification,'
+                             'as \'CL\', \'zeta\' or \'cross_section\'')
```
ericpre: I would prefer to avoid using `\` when not necessary because I find it easier to read (and to maintain).

hyperspy/_signals/eds_tem.py
```diff
@@ -287,30 +289,53 @@ def quantification(self,
...
+        method: {'CL', 'zeta', 'cross_section'}
             Set the quantification method: Cliff-Lorimer, zeta-factor, or
             ionization cross sections.
         factors: list of float
             The list of kfactors, zeta-factors or cross sections in same order
             as intensities. Note that intensities provided by Hyperspy are
             sorted by the alphabetical order of the X-ray lines.
             eg. factors =[0.982, 1.32, 1.60] for ['Al_Ka', 'Cr_Ka', 'Ni_Ka'].
-        composition_units: 'weight' or 'atomic'
+        composition_units: {'atomic', 'atomic'}
```
ericpre: `composition_units: {'atomic', 'weight'}`
Specify the default is `'atomic'`.

hyperspy/_signals/eds_tem.py
```diff
@@ -287,30 +289,53 @@ def quantification(self,
...
+        absorption_correction: {'False', True} bool
+            Specify whether or not an absorption correction should be applied.
+            'False' by default so absorption will not be applied unless
+            specfied.
+        tilt_stage : the sample holder tilt stage, used to calculate the
+            absorption pathways (in degrees).
+        mass_thickness: {'auto'}
+            mass_thickness in ***find out units*** (can be a single value or
+            have the same navigation dimension as the signal).
+            NB: Must be specified for 'CL' method, for 'zeta' or 'cross_section'
```
ericpre: Not clear what do you mean with "must be specify". It already has a default value, which means that it does not need to be specify? The tests with absorption correction do not specify with values.

hyperspy/_signals/eds_tem.py
```diff
@@ -287,30 +289,53 @@ def quantification(self,
...
+            methods, first quantification step provides a mass_thickness
+            internally during quantification.
+        conv_crit: The convergence criterium defined as the percentage
+            difference between 2 successive iterations. 0.5% by default.
         navigation_mask : None or float or signal
             The navigation locations marked as True are not used in the
             quantification. If int is given the vacuum_mask method is used to
```
ericpre: Just notice a typo in the doc: it should be: `If float is` and the other one below.

hyperspy/_signals/eds_tem.py
```diff
@@ -414,16 +572,24 @@ def quantification(self,
                     "Sample.xray_lines", ([xray_line]))
         if plot_result and composition[i].axes_manager.navigation_size != 1:
             utils.plot.plot_signals(composition, **kwargs)
+
+        if absorption_correction == 'True':
```
ericpre: If absorption_correction is boolean, this will not work. Simply replace to `if absorption_correction:`

hyperspy/_signals/eds_tem.py
```diff
@@ -414,16 +572,24 @@ def quantification(self,
...
+            print('Conversion found after {0} interations.'.format(str(it)))
```
ericpre: Use the logger (and f-strings): `_logger.inf(f'Conversion found after {it} interations.')`

hyperspy/_signals/eds_tem.py
```diff
@@ -414,16 +572,24 @@ def quantification(self,
...
+
         if method == 'zeta':
             self.metadata.set_item("Sample.mass_thickness", mass_thickness)
             return composition, mass_thickness
         elif method == 'cross_section':
             return composition, number_of_atoms
         elif method == 'CL':
-            return composition
+            if absorption_correction != False:
```
ericpre: `if absorption_correction:`

hyperspy/_signals/eds_tem.py
```diff
@@ -662,7 +828,47 @@ def _get_dose(self, method, beam_current='auto', live_time='auto',
...
+        composition: stack of compositions as determined from an initial k_factor
+            quantification.
+            Probe current in nA
+        thickness: float or array
+            Either a float value for thickness in nm or an array equal to the size
+            of the EDX map with thickness at each position of the sample.
+
+        Returns
+        --------
+        Mass_thickness as an array in kg/m^2.
```
ericpre: Use unicode: `kg/m²`

hyperspy/_signals/eds_tem.py
```diff
@@ -353,37 +381,166 @@ def quantification(self,
...
+                                      'function will still run. However if '
+                                      '1 nm² is not correct, please read the '
+                                      'user documentations for how to set '
+                                      'this properly.')
+
         if method == 'CL':
-            composition.data = utils_eds.quantification_cliff_lorimer(
-                composition.data, kfactors=factors,
-                mask=navigation_mask) * 100.
+            int_stack = utils.stack(intensities)
```
ericpre: It looks to me that this if/elif/else block could simplify by simply replace the quantification function and setting the argument accordingly:

```python
if method == 'CL':
    quantification_method = utils_eds.quantification_cliff_lorimer
    kwargs = {...}
elif method == 'zeta':
    quantification_method = utils_eds.quantification_zeta_factor
    kwargs = {...}

elif method == 'cross_section':
    quantification_method = utils_eds.quantification_cross_section
    kwargs = {...}
else:
    raise ValueError("Please specify method for quantification, "
                     "as 'CL', 'zeta' or 'cross_section'")
...
results = quantification_method(**kwargs)
```

ericpre (commented):
I have sent a PR to this branch to fix the failure of the test suite and I mainly commented small things which can tweaked. I haven't look at the calculation and I can't tell if other absorption correction calculation would fit nicely using this approach simply because I don't know them! I am sure @TomSlater will have genuine deep feeling about them and his expertise and views on the matter will be appreciated! :)

TomSlater:
It's great that this is being implemented, thanks @k8macarthur .

Considering each method is essentially putting different parameters in to the same absorption correction equation, would it make sense to have a simple function to call once you've calculated the attenuation coefficients and mass thickness in each method? Or maybe before this point? This could have an argument that is 'sample geometry' in the future that could be varied between 'thin film' and 'nanoparticle' for example. It might make it easier to add different sample morphologies in the future.

I'm also not sure I like calculating absorption with a thickness map. Obviously you're not truly calculating the X-ray path length through the sample at any point. I'd prefer just to say that samples need to have uniform thickness in order to perform accurate absorption correction if using the equation for thin films. i.e. input only one thickness value.

I don't feel particularly strongly about either point so happy to change my mind if others think differently.

CorentinLG:
Regarding thickness map: I think it is usefull to keep this possibility.
Imagine for instance a FIB section of a heterogeneous sample containing various phases of different densities. You could first obtain a map of the mass thickness using STEM-EELS (t/l) and reinject it for EDX quantification.

ericpre:
Thanks @TomSlater and @CorentinLG your comments.

Having the option to provide a thickness map would be convenient, but it can be also be fairly inaccurate and possibly misleading non expert user, in a sense that the user may think that the absorption correction is clever enough to take into account thickness variation away from the current pixel and not suffer from such issues. FIB specimen prepared with very thick edges are typically a source of such issues.

So long the documentation explain what does the absorption correction and warn the user (in the docstring or user guide or both) about this issue and there are a few references to good books and papers, this should good and useful.

hyperspy/_signals/eds_tem.py
k8macarthur: fixed

hyperspy/_signals/eds_tem.py
k8macarthur: fixed

hyperspy/_signals/eds_tem.py
k8macarthur: fixed

hyperspy/_signals/eds_tem.py
k8macarthur: fixed

hyperspy/_signals/eds_tem.py
k8macarthur: fixed

k8macarthur:
Thanks all for the comments. My main thinking at the moment is to have a basic form of absorption correction function available that can then be subsequently expanded on and improved. My reason for thinking of allowing thickness maps is because the zeta and cross section methods can automatically produce a mass-thickness map and therefore do some quantification on a per pixel basis. I agree this leads to a not so accurate representation of the beam path, but so do the assumptions of uniform composition and uniform sample thickness.

I think if it can be carefully explained in the user documentation it might be ok. But obviously needs to be a generally agreed census that this is going to be helpful.

hyperspy/_signals/eds_tem.py
k8macarthur: fixed

hyperspy/_signals/eds_tem.py
k8macarthur: fixed

hyperspy/_signals/eds_tem.py
k8macarthur: It needs to be provided for CL. for the other two methods an internally calculated one from the initial quantification is used.

ericpre:
@k8macarthur 
> I think if it can be carefully explained in the user documentation it might be ok. But obviously needs to be a generally agreed census that this is going to be helpful.

Sounds like a good plan to me! And the CIs are green now.

@TomSlater 
>  This could have an argument that is 'sample geometry' in the future that could be varied between 'thin film' and 'nanoparticle' for example. It might make it easier to add different sample morphologies in the future.

Maybe this argument could added when the absorption correction for nanoparticle will be added later?

k8macarthur:
Ok I'm now confident the function runs and does everything I intended it to. My aim was tp produce a basic absorption correction calculation for all three methods so that it is at least implemented in hyperspy and then it should be easier to extend and improve it in the future. I'll update the user docs today. 

I would personally (if my maths were good enough) like the zeta factor and cross section quantification methods to have detector azimuthal and height information and then quantify path length based on that. This would be the ideal scenario as mass-thickness information is immediately supplied for these methods. However this is probably a lot of development involved in something like this.

k8macarthur:
@TomSlater if you want to make direct changes to the branch and we can work together to add your ideas to the function now rather than risk putting something out which is less than ideal, if you'd prefer?

k8macarthur:
@TomSlater In answer to your comments above. The Cliff-Lorimer function accepts a thickness information which could used as a sample geometry input. Then the zeta and CL use both the same mass thickness function because they are in mass units and the cross sections has a different mass thickness calculation because it has atomic units. These would be the three functions to change:
get_abs_corr_cross_section
get_abs_corr_zeta
CL_get_mass_thickness(self, weight_percent, thickness)
to improve the EDX quantification. The rest of the quantification sub functions just call on these two the calculate an absorption correction factor each time it is needed.

For the k-factor approach, either we just replace thickness, with sample geometry or we could create a thickness_map function which creates a map for the require geometry? Or is that still perpetuating the pixel-by-pixel problem?

TomSlater:
I think for the moment what you've got works and I wouldn't want to add complications at this stage. Saying that, I will go through and see if I can make simplifications.

In terms of calculating true X-ray paths, there's no way of getting something exact from 2 spatial dimensions. You have to make assumptions about the variations in composition and density along the projected direction and about the surface topography (which you may get a good idea from via thickness but might not be exactly right). Exactly how much these influence composition is debatable and will be sample dependent, probably assumptions will give you something accurate to a few percent.

I'll see if I can go through asap and may make changes but if it works in terms of standard absorption correction for parallel-sided samples then that's great for the majority of cases.

hyperspy/_signals/eds_tem.py
k8macarthur: Done.

hyperspy/_signals/eds_tem.py
k8macarthur: I have thickness=auto do we need mass_thickness as well?

hyperspy/_signals/eds_tem.py
k8macarthur: I've added the spaces but not sure what you need by the quotes not being necessary. Is this removing all the quote marks in lines 452-454 or just leaving first and last in?

k8macarthur:
@TomSlater I've added an if statement which assumes beta_tilt is zero if it is not in the metadata. Maybe there should also be a warning...
It would be really helpful if you could write some tests that check your modifications to the tilt function produce the correct tilt.

k8macarthur:
@TomSlater I've added an if statement which assumes beta_tilt is zero if it is not in the metadata. Maybe there should also be a warning...
It would be really great if you could write some test functions for your new version of stage tilt. Happy to help with it but would need help determining what the real answers should be.

hyperspy/_signals/eds_tem.py
ericpre: No. I wrongly suggested this at the time of the review because there was a `mass_thickness` parameter in the docstring (which was not correct), but it has fixed by correcting the docstring.

hyperspy/_signals/eds_tem.py
ericpre: I suspect I meant the `\` character!
Generally, I prefer to avoid using the `\` character when it is not necessary (as in this case) to make the code more readable and avoid unexpected behaviour/mistake since `\` can be used to escape other characters in string. This is a bit of matter of taste, so up to you, but the following should work fine:
```python
                        warnings.warn('Thickness is required for absorption '
                                      'correction with K-Factor Method. Result '
                                      'will contain no correction for absorption.')
```

Use "k-factor" instead of "K-Factor"?

There is still a space missing in the first line.

hyperspy/_signals/eds_tem.py
```diff
@@ -353,47 +381,133 @@ def quantification(self,
...
+
+        elif method =='cross_section':
+            quantification_method = utils_eds.quantification_cross_section
+            kwargs = {"intensities" : int_stack.data,
+                    "cross_sections" : factors,
+                    "dose" : self._get_dose(method, **kwargs),
+                    "absorption_correction" : abs_corr_factor}
+
+        else:
+            raise ValueError('Please specify method for quantification,'
```
ericpre: A missing space have been introduced after the `,`.

TomSlater:
Sorry, on my to do list to add tests this week and to check everything works fine.

hyperspy/_signals/eds_tem.py
```diff
@@ -287,48 +289,74 @@ def quantification(self,
...
+            generate a mask with the float value as threhsold.
             Else provides a signal with the navigation shape.
         closing: bool
             If true, applied a morphologic closing to the mask obtained by
             vacuum_mask.
         plot_result : bool
             If True, plot the calculated composition. If the current
             object is a single spectrum it prints the result instead.
+        max_iterations : int
+            An upper limit to the number of calculations for absorption correction.
```
ericpre, line 359: The `probe_area` argument is not documented in the docstring.

hyperspy/_signals/eds_tem.py
```diff
@@ -353,47 +381,133 @@ def quantification(self,
...
+            toa = take_off_angle
+
+        #determining illumination area for cross sections quantification.
+        if method == 'cross_section':
+            if probe_area == 'auto':
+                parameters = self.metadata.Acquisition_instrument.TEM
+                if probe_area in parameters:
+                    probe_area = parameters.TEM.probe_area
+                else:
+                    probe_area = self._get_probe_area(navigation_axes=None)
```
ericpre: In case of `navigation dimension > 2`, this is going to lead to a confusing error message asking the user to specify the `navigation_axes` parameter but the `quantification` function does not take this argument. No sure, what is the best, maybe tweak the error message in `_get_probe_area`?

hyperspy/_signals/eds_tem.py
```diff
@@ -562,6 +689,36 @@ def create_model(self, auto_background=True, auto_add_lines=True,
...
 
+
+    def _get_probe_area(self, navigation_axes=None):
+        """Calculates a pixel area which can be approximated to probe area,
+        when the beam is larger than or equal to pixel size.
+        """
+        if (self.axes_manager.navigation_dimension > 2 and
+                navigation_axes is None):
+            raise ValueError("With `probe_area='auto' and "
+                             "navigation dimension > 2, you need "
```
ericpre: Change to: "The navigation axes corresponding of the probe are ambiguous, you need to specify the `navigation_axes` parameter."?

Even in the case of `navigation_dim == 2`, it assumes that these two navigation axes correspond to these of the probe, which will be wrong in case of stack of line scan, as in case of a time serie of line scan, for example.

k8macarthur: Fair point I hadn't thought of that. Is there a way around this or just a question of documenting it?

ericpre: I have tried to fix this in https://github.com/k8macarthur/hyperspy/pull/17.

ericpre:
@TomSlater and @k8macarthur, on similar topic, while you are doing angle calculations, is it worth fixing https://github.com/hyperspy/hyperspy/issues/2089 by taking into account the beta tilt in the take off angle calculation?

TomSlater:
I have updated the take_off_angle function in misc/eds/utils to take beta tilt in to account, so this should now resolve #2089 . This should mean that alpha tilt, beta tilt, azimuthal and elevation angles are all taken in to account when calculating absorption.

This is all for a parallel sided thin film only.

Tests in https://github.com/k8macarthur/hyperspy/pull/18

ericpre:
> I have updated the take_off_angle function in misc/eds/utils to take beta tilt in to account, so this should now resolve #2089 . This should mean that alpha tilt, beta tilt, azimuthal and elevation angles are all taken in to account when calculating absorption.

Indeed, sorry, it seems that I confused myself! ;)

This PR looks good to me now, anything else to add?

TomSlater:
I think it's pretty much complete for thin film samples, certainly looks good to me.

In case someone wanted to check the equation I used for adding beta-tilt, I've attached a picture of the derivation. I couldn't find a reference for doing this although I'm sure there is one somewhere.

IMG_20191115_132017030

k8macarthur:
@TomSlater agreed that we only needed tests for the angle derivation, the fact that it is called correctly and all the quantification tests run is good enough for me. I fixed the error when it didn't find a beta tilt anyway. Which could be the case if a single tilt holder was used.

Yes it is only the thin film approximation but I think the user documentation makes that clear and my hope is more complex geometries can follow later.
I think it's done.

hyperspy/_signals/eds.py
```diff
@@ -710,9 +706,13 @@ def get_take_off_angle(self):
         tilt_stage = mp.Stage.tilt_alpha
         azimuth_angle = mp.Detector.EDS.azimuth_angle
         elevation_angle = mp.Detector.EDS.elevation_angle
+        if 'beta_tilt' not in mp:
```
ZanettaPM, line 709: This test does not seems to work, it always return 0. 
I think it should be 'tilt_beta' but maybe it's only the case for emd files ?

ZanettaPM: And replace mp by mp.Stage ?

ZanettaPM: A last comment, an error shows up when value are not integers when i try this function. 
With all these modifications it works perfectly for me. Nice work !

ericpre: Do you want to open a PR to fix it?

PR #2259, opened 2019-09-27
Add lorentzian remove background

ericpre:
Continuation of #2185.

### Progress of the PR
- [x] Fix lazy `estimate_parameters` of the `Lorentzian` component,
- [x] add `Lorentzian` component to `remove_background`,
- [x] add tests,
- [x] ready for review.

```python
import hyperspy.api as hs
import numpy as np

lorentzian = hs.model.components1D.Lorentzian()
lorentzian.A.value = 10
lorentzian.centre.value = 10
lorentzian.gamma.value = 1
s = hs.signals.Signal1D(lorentzian.function(np.arange(0, 20, 0.01)))
s.axes_manager[0].scale = 0.01
s1 = s.remove_background(signal_range=(None, None),
                         background_type='Lorentzian',
                         show_progressbar=None,
                         fast=False)

```

### DIFF
--- a/hyperspy/_components/lorentzian.py
+++ b/hyperspy/_components/lorentzian.py
@@ -21,6 +21,7 @@ import dask.array as da
 
 from hyperspy._components.expression import Expression
 
+
 def _estimate_lorentzian_parameters(signal, x1, x2, only_current):
     axis = signal.axes_manager.signal_axes[0]
     i1, i2 = axis.value_range_to_indices(x1, x2)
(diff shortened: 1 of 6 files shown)


jlaehne:
I tested the new background removal function on some real and model datasets. In principle it works. However, for the sake of background removal in noisy datasets, it is not very convenient that the estimators for both the Gaussian and the Lorentzian components use the maximum of the dataset for the height parameter.

It becomes clear when adding the line 's.add_gaussian_noise(std=lorentzian.A.value*0.05)' to the minimal example.

Instead, we should probably take the value at the estimated centre position and average over a few values around this point - as the range for averaging one could for example choose a certain percentage of datapoints within the width of the estimated function.

tjof2:
> it is not very convenient that the estimators for both the Gaussian and the Lorentzian components use the maximum of the dataset for the height parameter.

Why not use some upper percentile of the dataset instead (if it's simpler than using the estimated centre position), e.g. 90%ile

ericpre:
Thanks @jlaehne and @tjof2 for your comments! I agree that the estimation of the parameter for the Lorentzian is not great.

@jlaehne, would you like to open a separate PR or send a PR to this branch to improve it?
If the estimation of the parameters are not very accurate, we may want to raise a warning when removing the background using `fast=True`.

francisco-dlp:
This LGTM and, unless somebody disagrees, I suggest merging it.

Regarding the accuracy of the estimation method, I think that it is good to keep in mind that the original purpose of the ``estimate_parameters`` method is to provide a fast and rough estimation of the parameter values that will later be refined by NLLS fitting. Therefore, I don't think that it is a big issue that the estimation is not very accurate when the SNR is poor, as long as it is good enough for the NNLS algorithm. When using it in the ``remove_background`` context, the user should simply be aware that the estimation will be rough for poor SNR data unless he goes for the slower curve fitting option.

That said, if there is a better way of estimating the parameters, then we should implement it when possible, as long as there is no drawback e.g. in terms of speed or robustness.

In any case, I agree with @ericpre that improving the centre estimation method is better suited for a separate PR.

jlaehne:
+1 for merge

If I come up with a good and simple way to estimate the height parameter, I will open a separate PR.

PR #2261, opened 2019-10-08
Skew normal estimate parameters

jlaehne:
### Description of the change
Implemented parameter estimation for skew normal component based on the methods of moments. Added tests.

### Progress of the PR
- [X] Added parameter estimation,
- [X] Added tests,
- [x] Fix issue in determining `A` (height),
- [x] Make it work for lazy signal (dask array),
- [x] Add to background removal,
- [x] update user guide,
- [x] add further tests,
- [x] ready for review.

```python
>>> g = hs.model.components1D.SkewNormal()
>>> x = np.arange(-10, 10, 0.01)
>>> data = np.zeros((32, 32, 2000))
>>> data[:] = g.function(x).reshape((1, 1, 2000))
>>> s = hs.signals.Signal1D(data)
>>> s.axes_manager._axes[-1].offset = -10
>>> s.axes_manager._axes[-1].scale = 0.01
>>> g.estimate_parameters(s, -10, 10, False)
```

### DIFF
--- a/doc/user_guide/signal1d.rst
+++ b/doc/user_guide/signal1d.rst
@@ -44,7 +44,6 @@ It is possible to crop interactively using :ref:`roi-label`. For example:
 
    Interactive spectrum cropping using a ROI.
 
-
 .. _signal1D.remove_background:
 
 Background removal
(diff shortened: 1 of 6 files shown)


jlaehne:
Dask Array does not seem to support numpy's fancy indexing capabilities used in:
`height = data[np.arange(signal.axes_manager.navigation_size), iheight]`
and in 
`height = data[(*np.indices(signal.axes_manager.navigation_shape), iheight)]`

In the meantime, I found a workaround using the dask array function vindex.

ericpre:
Thanks!

PR #2264, opened 2019-10-21
Fix fast background subtraction for binned signals

askorikov:
### Description of the change
Fixed incorrect background subtraction using "fast" option due to the missing scaling for binned signals.

### Progress of the PR
- [x] Change implemented,
- [x] ready for review.

```python
>>> from hyperspy.components1d import PowerLaw
>>> from hyperspy.signals import Signal1D
>>> import numpy as np

# Generate a power law spectrum
>>> s = Signal1D(np.empty((100,)))
>>> axis = s.axes_manager.signal_axes[0]
>>> axis.scale = 0.02
>>> axis.offset = 1
>>> g1 = PowerLaw(50015.156, 1.2)
>>> s.data = g1.function(axis.axis)
>>> s.metadata.Signal.binned = True

# Subtract background using "fast" option
>>> s2 = s.remove_background(signal_range=[1.0, 2.0], background_type='PowerLaw', fast=True)

# Assert that the result is relatively close to zero
>>> assert(np.linalg.norm(s2.data) / np.linalg.norm(s.data) < 0.05)
```


ericpre:
Thanks @askorikov. It looks to me that this bug have been introduced in https://github.com/hyperspy/hyperspy/commit/3facbc1df3bee8c186cb0efecaef1c6e3fc7d0a4 from the PR https://github.com/hyperspy/hyperspy/pull/2032 and a most appropriate place to fix it would be in the `estimate_parameters` method of the `PowerLaw` component.

There should be something wrong in the test too because it doesn't catch this bug...

Since this is a bug fix, this PR should go to the `RELEASE_next_patch` branch instead of `RELEASE_next_minor`. Your branch is ahead of `RELEASE_next_patch`, so you will have to rebase your commit onto `RELEASE_next_patch`.

askorikov:
Oh indeed, I messed up the branch. What is the preferred solution: changing the base branch in this request or starting a new one?

I think `estimate_parameters` of the `PowerLaw` works correctly, as it properly takes into account the scale of the axis (through using `integrate1D`). However, the scaling is not reverted when calculating the values of the estimated background. The other background types (offset, Gaussian, etc.) seem to fail in the same way. This could've probably be fixed in `function_nd` of the `Expression` class, but currently it's agnostic of the binning of the signal.

I also dug into the tests a bit, and it seems that components' tests take the scaling into account explicitly, and the test for `remove_background` only uses binned signals. So this issue turned out to not be covered.

francisco-dlp:
Good catch @askorikov! I think that your fix is very good as is. ``function_nd`` is indeed agnostic of the binning of the signal as it should be, so I wouldn't touch that. The only thing that I would suggest is adding tests to avoid re-introducing this bug in the future.

Regarding the wrong branch issue, the way to fix it is up to you as long as it works.

ericpre:
As for #2265, it would be good to rebase this branch on RELEASE_next_patch or create a new branch.

askorikov:
Sure, I will do that after implementing new tests to cover this issue.

PR #2269, opened 2019-10-30
Voigt component as expression

jlaehne:
### Description of the change
Implemented Voigt lineshape component as expression based on the implementation of @AEljarrat  mentioned in #2251. As the existing `Voigt` function had specific features for photoemission spectroscopy, this has been moved to `PESVoigt`, but similar to the solution for the `Polynomial` component, this is still available as legacy `Voigt` component (with a deprecation warning) and the new component is at the moment only chosen for `legacy=False`.

### Progress of the PR
- [X] Change implemented,
- [X] update docstring,
- [x] update user guide,
- [x] add tests,
- [x] ready for review.

```python
>>> import hyperspy.api as hs
>>> import numpy as np
>>> voigt = hs.model.components1D.Voigt(legacy=False)
>>> 
```


hyperspy/_components/pes_voigt.py
```diff
@@ -0,0 +1,291 @@
...
+    """Voigt lineshape.
+
+    The voigt peak is the convolution of a Lorentz peak with a Gaussian peak.
+
+    The formula used to calculate this is::
+
+        z(x) = (x + 1j gamma) / (sqrt(2) sigma)
+        w(z) = exp(-z**2) erfc(-1j z) / (sqrt(2 pi) sigma)
+
+        V(x) = scale Re(w(z(x-center)))
```
ericpre: typo: `scale * Re(w(z(x-center)))`

hyperspy/_components/pes_voigt.py
```diff
@@ -0,0 +1,291 @@
...
+    return scale * V.real
+
+
+class Voigt(Component):
+    # Legacy class to be removed in v2.0
+
+    """This is the legacy Voigt profile component dedicated to photoemission 
+    spectroscopy data analysis that will renamed to `PESVoigt` v2.0. To use
+    the new Voigt lineshape component set `legacy=False`. 
+
```
ericpre, line 93: Add:
```
See the documentation of ``hyperspy._components.voigt.Voigt`` to use the new definition of the Voigt component. 
```

hyperspy/_components/voigt.py
```diff
@@ -16,136 +16,99 @@
...
-    """Voigt lineshape.
+class Voigt(Expression):
 
-    The voigt peak is the convolution of a Lorentz peak with a Gaussian peak.
+    r"""Voigt component.
 
-    The formula used to calculate this is::
+    Symmetric peak shape based on the convolution of a Lorentzian and Normal 
+    (Gaussian) distribution. In this case using an approximate formula by David 
+    (see Notes).
```
ericpre: Is it worth mentioning the difference with a pseudo-Voigt? Linear combination of Gaussian and Lorentzian versus this approximation for the convolution of Gaussian and Lorentzian?

hyperspy/_components/voigt.py
```diff
@@ -171,58 +134,54 @@ def estimate_parameters(self, signal, E1, E2, only_current=False):
...
-        >>> s = hs.signals.Signal1D({'data' : data})
-        >>> s.axes_manager.axes[-1].offset = -10
-        >>> s.axes_manager.axes[-1].scale = 0.01
-        >>> g.estimate_parameters(s, -10,10, False)
+        >>> g = hs.model.components1D.Voigt(legacy=False)
+        >>> x = np.arange(-10, 10, 0.01)
+        >>> data = np.zeros((32, 32, 2000))
+        >>> data[:] = g.function(x).reshape((1, 1, 2000))
+        >>> s = hs.signals.Signal1D(data)
+        >>> s.axes_manager._axes[-1].offset = -10
```
ericpre: Why using `_axes` instead of `axes`?

hyperspy/tests/component/test_pesvoigt.py
```diff
@@ -0,0 +1,65 @@
...
+
+@pytest.mark.parametrize(("lazy"), (True, False))
+@pytest.mark.parametrize(("only_current", "binned"), TRUE_FALSE_2_TUPLE)
+def test_estimate_parameters_binned(only_current, binned, lazy):
+    s = Signal1D(np.empty((200,)))
+    s.metadata.Signal.binned = binned
+    axis = s.axes_manager.signal_axes[0]
+    axis.scale = .05
+    axis.offset = -5
+    g1 = PESVoigt()
```
ericpre, line 53: Add test for `Voigt(legacy=True)`?

hyperspy/_components/voigt.py
```diff
@@ -16,136 +16,99 @@
...
-def voigt(x, FWHM=1, gamma=1, center=0, scale=1):
-    """Voigt lineshape.
+class Voigt(Expression):
 
-    The voigt peak is the convolution of a Lorentz peak with a Gaussian peak.
+    r"""Voigt component.
 
-    The formula used to calculate this is::
+    Symmetric peak shape based on the convolution of a Lorentzian and Normal 
+    (Gaussian) distribution. In this case using an approximate formula by David
```
ericpre: Because the formula to calculate the Voigt component looks complicated at first read, I would first give the definition of the Voigt function with the convolution (as a reference) and then I would give the formula used for the calculation.

hyperspy/_components/voigt.py
```diff
@@ -16,136 +16,99 @@
...
-    Ref: W.I.F. David, J. Appl. Cryst. (1986). 19, 63-64
 
-    adjusted to use stddev and HWHM rather than FWHM parameters
+    Parameters
+    -----------
+    centre : float
+        Location of the maximum of the peak.
+    area : float
+        Intensity below the peak.
+    lwidth : float
```
ericpre: Any reason to use `lwidth` and `gwidth` instead of `gamma` and `sigma`?

ericpre (commented):
Nice PR, thanks @jlaehne. I made a couple of suggestion to clarify the docstring.
It would be good to add the new Voigt component to the `remove_background` tool. Other than that, this PR looks good to me.

@francisco-dlp, could you have a quick look at the API change and the legacy component, it looks fine to me, but a second pair of eyes would be too good to make sure this is done correctly.

hyperspy/_components/pes_voigt.py
jlaehne: I changed it to math typesetting like in the new Voigt.

hyperspy/_components/pes_voigt.py
jlaehne: done

hyperspy/_components/voigt.py
jlaehne: I added an explanation based on the abstract of David et al.

hyperspy/tests/component/test_pesvoigt.py
jlaehne: added test

hyperspy/_components/voigt.py
jlaehne: It was copy pasted from `Gaussian`. I also removed it in `Gaussian`, `GaussianHF`, and `Lorentzian`.

hyperspy/_components/voigt.py
jlaehne: good idea. done.

jlaehne:
Concerning the use of the `lwidth` and `gwidth` parameters instead of `gamma` and `sigma`:
I took it from the component of @AEljarrat. 

However, I also find it more intuitive to work with a width parameter than with `sigma`. Of course, `gamma=HWHM=lwidth`, but then I thought it more consistent to use a similar keyword for both distributions. 

Would you prefer `gamma` and `FWHM` like in the PESVoigt?

Maybe even better would be `gamma` and `fwhm` to be consistent with `Lorentzian` and `GaussianHF` components?

@ericpre, @francisco-dlp - what is your opinion on the naming?

(PS: Also added Voigt to `remove_background`)

hyperspy/_components/voigt.py
jlaehne: see comment  below

hyperspy/_components/voigt.py
ericpre: Sorry for this wrong suggestion, there is no `axes` attribute to the axes manager! I made a PR to this branch to fix it: https://github.com/jlaehne/hyperspy/pull/2.
I guess, this is one of these where one writes a bit too quickly and the brain is following up properly... ;)

ericpre:
Regarding `fwhm` versus `sigma`, I don't have any particular view and indeed trying to make consistent with other function would be sensible!
- `Gaussian` has `A`, `sigma` and `centre`
- `GaussianHF` has `height`, `FWHM` and `centre`
- `Lorentzian` has `A`, `gamma` and `centre`
- `PSEVoigt` has `area`, `FWHM`, `gamma` and `centre`

To be consistent with `Gaussian` and `Lorentzian` I would say that `sigma` and `gamma` should be used in `Voigt`.

Other opinions?

hyperspy/signal_tools.py
```diff
@@ -1003,12 +1005,15 @@ def set_background_estimator(self):
...
-            self.bg_line_range = 'full'
+        elif self.background_type == 'Power Law':
+            self.background_estimator = components1d.PowerLaw()
+            self.bg_line_range = 'from_left_range'
         elif self.background_type == 'SkewNormal':
             self.background_estimator = components1d.SkewNormal()
             self.bg_line_range = 'full'
+        elif self.background_type == 'Voigt':
+            with ignore_warning(message="The API of the `Voigt` component"):
+                self.background_estimator = components1d.Voigt()
```
ericpre:
```suggestion
                self.background_estimator = components1d.Voigt(legacy=False)
```
to be consistent with the `remove_background` method.

jlaehne: Thanks. I changed that.

hyperspy/_components/pes_voigt.py
```diff
@@ -0,0 +1,331 @@
...
+    centre: Parameter
+    FWHM : Parameter
+    gamma : Parameter
+    resolution : Parameter
+    shirley_background : Parameter
+    non_isochromaticity : Parameter
+    transmission_function : Parameter
+    spin_orbit_splitting : Bool
+    spin_orbit_branching_ratio : float
+    spin_orbit_splitting_energy : float
```
dnjohnstone: I realise that this is due to become legacy, but shouldn't the attributes that are specific to PESVoigt not be in this more general Voigt docstring?

hyperspy/_components/pes_voigt.py
```diff
@@ -0,0 +1,331 @@
...
+        FWHM=:math:`2 \sigma \sqrt(2 \log(2))` of the Gaussian distribution.
+    gamma : Parameter
+        :math:`\gamma` of the Lorentzian distribution.
+    resolution : Parameter
+    shirley_background : Parameter
+    non_isochromaticity : Parameter
+    transmission_function : Parameter
+    spin_orbit_splitting : Bool
+    spin_orbit_branching_ratio : float
+    spin_orbit_splitting_energy : float
```
dnjohnstone, line 191: I think these parameters need some explanation.

hyperspy/_components/pes_voigt.py
```diff
@@ -0,0 +1,331 @@
...
+from hyperspy._components.gaussian import _estimate_gaussian_parameters
+
+
+sqrt2pi = math.sqrt(2 * math.pi)
+sigma2fwhm = 2 * math.sqrt(2 * math.log(2))
+
+
+def voigt(x, FWHM=1, gamma=1, center=0, scale=1):
+    r"""Voigt lineshape.
+
```
dnjohnstone, line 32: stray 'r' here

hyperspy/_components/pes_voigt.py
```diff
@@ -0,0 +1,331 @@
...
+        self.transmission_function.free = False
+        # Options
+        self.shirley_background.active = False
+        self.spin_orbit_splitting = False
+        self.spin_orbit_branching_ratio = 0.5
+        self.spin_orbit_splitting_energy = 0.61
+        self.isbackground = False
+        self.convolved = True
+
+    def function(self, x):
```
dnjohnstone, line 229: Doesn't this need a docstring?

hyperspy/_components/pes_voigt.py
```diff
@@ -0,0 +1,331 @@
...
+        x2 : float
+            Defines the right limit of the spectral range to use for the
+            estimation.
+
+        only_current : bool
+            If False estimates the parameters for the full dataset.
+
+        Returns
+        -------
+        bool
```
dnjohnstone: This seems like a very strange return for a method that is called estimate parameters - I understand why it is, but I think it should explain here.

hyperspy/_components/pes_voigt.py
```diff
@@ -0,0 +1,331 @@
+# -*- coding: utf-8 -*-
+# Copyright 2007-2016 The HyperSpy developers
```
dnjohnstone: probably needs an update more generally

hyperspy/_components/voigt.py
```diff
@@ -16,136 +16,110 @@
...
 sqrt2pi = math.sqrt(2 * math.pi)
+sigma2fwhm = 2 * math.sqrt(2 * math.log(2))
 
 
-def voigt(x, FWHM=1, gamma=1, center=0, scale=1):
-    """Voigt lineshape.
+class Voigt(Expression):
 
-    The voigt peak is the convolution of a Lorentz peak with a Gaussian peak.
+    r"""Voigt component.
```
dnjohnstone, line 34: stray r

dnjohnstone (commented):
This basically looks good to me except for some fairly superficial housekeeping and the conflict on updates to remove_background

hyperspy/_components/voigt.py
```diff
@@ -16,136 +16,110 @@
...
 
-    The formula used to calculate this is::
+    Symmetric peak shape based on the convolution of a Lorentzian and Normal 
+    (Gaussian) distribution:
+    
+        .. math::
```
ericpre: Indentation is not correct here and it makes the following formula not to render properly.

hyperspy/_components/pes_voigt.py
ericpre: This `r` is necessary to render the latex properly.

hyperspy/_components/pes_voigt.py
ericpre: Indeed, but since this is simply moving code around, this can be optional for PR.

hyperspy/_components/pes_voigt.py
ericpre: Same as above, since this is simply moving code around, this can be optional for PR.

hyperspy/_components/pes_voigt.py
ericpre: A quick search in the code shows that it is actually used: https://github.com/hyperspy/hyperspy/blob/ff2d9d38098831d0e2cde1239dd80bff056b9f67/hyperspy/signal_tools.py#L1089-L1095

As usual, documenting it would be good indeed.

hyperspy/_components/voigt.py
ericpre: Same as above, this `r` is necessary.

hyperspy/_components/voigt.py
```diff
@@ -171,58 +145,62 @@ def estimate_parameters(self, signal, E1, E2, only_current=False):
...
+        >>> data[:] = g.function(x).reshape((1, 1, 2000))
+        >>> s = hs.signals.Signal1D(data)
+        >>> s.axes_manager[-1].offset = -10
+        >>> s.axes_manager[-1].scale = 0.01
+        >>> g.estimate_parameters(s, -10, 10, False)
 
         """
         super(Voigt, self)._estimate_parameters(signal)
         axis = signal.axes_manager.signal_axes[0]
+        centre, height, sigma = _estimate_gaussian_parameters(signal, x1, x2,
```
ericpre, line 161: Only the `sigma` is estimated here which make it fairly inaccurate expect when the degree of gaussianity is high (in which case, it may be better to use a gaussian!).
If someone could find a more accurate way to estimate the `gamma` value, that would be great, I started to look at this in the literature, but I haven't successful so far.
Even if this is quite critical, we can leave to another PR.

hyperspy/signal_tools.py
```diff
@@ -1003,12 +1004,15 @@ def set_background_estimator(self):
...
-            self.bg_line_range = 'full'
+        elif self.background_type == 'Power Law':
+            self.background_estimator = components1d.PowerLaw()
+            self.bg_line_range = 'from_left_range'
         elif self.background_type == 'SkewNormal':
             self.background_estimator = components1d.SkewNormal()
             self.bg_line_range = 'full'
+        elif self.background_type == 'Voigt':
+            with ignore_warning(message="The API of the `Voigt` component"):
+                self.background_estimator = components1d.Voigt(legacy=False)
```
ericpre, line 1015:
```suggestion
                self.background_estimator = components1d.Voigt(legacy=False)
           self.bg_line_range = 'full'
```

hyperspy/_components/pes_voigt.py
jlaehne: OK, I'll replace it by a note that refers to the two respective components. Anyway, it is called `Parameters` in other components ;-)

hyperspy/_components/pes_voigt.py
jlaehne: Yes, but as I have not programmed the PES component, I'll leave it to someone more into the topic to add further explanations.

hyperspy/_components/pes_voigt.py
jlaehne: I'll add a note, if we agree on it, I can also add it to other components that are used by `remove_background`.

hyperspy/_components/pes_voigt.py
jlaehne: Yes, I can change it in the files used here, but it can be found across the project.

hyperspy/_components/voigt.py
jlaehne: good catch

hyperspy/_components/voigt.py
jlaehne: I agree, but I also have not found a good solution so far, thus I propose to merge this PR and leave it. You could open an issue if you deem it appropriate.

hyperspy/signal_tools.py
jlaehne: done, thanks

hyperspy/_components/pes_voigt.py
dnjohnstone: Ah - my bad, reviewing = learning = :)

PR #2272, opened 2019-11-05
Fix histogram complex

ericpre:
### Progress of the PR
- [x] Fix estimation of number of bins for the histogram in case of complex signal,
- [x] add "xth" syntax for vmin and vmax - I quite liked @francisco-dlp suggestion and it works nicely,
- [x] update docstring,
- [x] update user guide (visualisation of FFT),
- [x] add tests,
- [x] ready for review.

### Related PRs

- [ ] https://github.com/hyperspy/hyperspy_gui_ipywidgets/pull/25
- [ ] https://github.com/hyperspy/hyperspy_gui_traitsui/pull/26

The following example crashes:
```python
import hyperspy.api as hs
from hyperspy.signal_tools import ImageContrastEditor
s = hs.datasets.example_signals.object_hologram()

fft = s.fft(True)
fft.plot(True)
ceditor = ImageContrastEditor(fft._plot.signal_plot)
```


ericpre:
[Discussion on gitter on this topic](https://gitter.im/hyperspy/hyperspy?at=5d83d65a1998d5169567df23) with the non relevant part removed and without the images:

Joshua Taillon @jat255 Sep 19 20:26
has anyone played around with plotting the output of a digital micrograph FFT?
HyperSpy loads it as a complex2D signal
plotting it like so:

whereas DM shows:


Joshua Taillon @jat255 Sep 19 22:32
figured it out
DM displays the log of the modulus of complex values by default, with bilinear interpolation
there may be some value in making such a display the default for power spectra

[...]

Eric Prestat @ericpre Sep 20 13:12
@jat255: see hyperspy/hyperspy#1727 for a discussion on plotting the power spectrum by default
currently it does it by default when hyperspy can know it is a FFT (check if there is a FFT node in the metadata)
which works for FFT generated in hyperspy
for data imported from DM, it may be a matter of parsing the metadata correctly and it should work as you suggested

Joshua Taillon @jat255 Sep 20 16:35
@ericpre it works if I explicitly add s.metadata.Signal.FFT to the metadata structure and specify power_spectrum=True when plotting

Something like this however is much more pleasing:

and closer to what DM shows
plotted with
s.amplitude.plot(interpolation='bilinear', norm='log',
                 vmin=np.nanpercentile(s.amplitude.data, 66),
                 colorbar=None)

Thomas Aarholt @thomasaarholt Sep 20 20:09
That is very pleasing to the eye! Might be nice to add that to the documentation so that I can find that later :p

Eric R. Hoglund @erh3cq Sep 21 14:41
@jat255 please! I always code that type of thing in. It would be nice to have a default or function for FFTs and diffraction patterns. I think that DM some how samples and removes outliers then additionally removes the percentiles specified.

[...]

Eric Prestat @ericpre Sep 22 12:45
@jat255 and @erh3cq, I think it would easier to discuss this in an issue on github, because there are different things (including bug) going on here.
As you can imagine, finding a default suitable is not straightforward!
the contrast adjustment tool (press "h" on a image figure) has been improved significantly in 1.5 (hyperspy/hyperspy#2216) to help with this, but it is not working with plotting power_spectrum...

Eric Prestat @ericpre Sep 22 12:52
there is also the saturated_pixels default plotting value which can set in the preferences which should be useful in this case, but from a quick look, it seems that the value you want to adjust is the vmin only and not vmin and `vmax, which the saturated_pixels actually does.

Francisco de la Peña @francisco-dlp Sep 25 05:22
We could specify vmin and vmax in percentiles like this: vmin="30th".

[...]

Francisco de la Peña @francisco-dlp Sep 25 05:54
Regarding the complex image default plotting parameters, I think that this is something to discuss and implement (if it haven't been done already) in the external packages devoted to diffraction analysis such as pyxem and pixstem. HyperSpy cannot stick to TEM conventions as it aims to be more general, but external packages devoted to TEM analysis probably should.

ericpre:
I am wondering if it would be worth adding a slider for `vmin` and `vmax` on the GUIs to set the x-th percentile.

thomasaarholt:
LGTM

francisco-dlp (changes requested):
I think that there are 2 issues with the ``vmin(max)="<x>nth"`` implementation:

* The ``raise`` part is never executed because getting the syntax wrong raises an issue in the previous line. It would be nice to add a test to test that the right exception is raised.
* Currently it uses the current data to calculate ``vmin`` and ``vmax``, but those values are not updated when the data changes. In other words, it sets the contrast globally using local data. I think that most users would expect it to set the contrast locally instead. If you think that setting the contrast globally with the percentile syntax is useful, then something like ``"g30th"`` could do.

Some extra comments:
* It may be easier to implement the feature by modifying the ``saturated_pixels`` code.
* Wouldn't it be good to deprecated ``saturated_pixels``?
* For consistency, the feature could be implemented also for ``plot_images``

ericpre:
Yes, I agree it would be good to deprecate `saturated_pixels`, since `saturated=2*x` is equivalent to `vmin="xth"` and `vmax="xth"`. This would simplify the logic for us but also users...

ericpre:
If I haven't forgotten anything, this should be in much better shape than before. There are two PRs to the GUI toolkit too: hyperspy/hyperspy_gui_ipywidgets#25 and hyperspy/hyperspy_gui_traitsui#26.

hyperspy/drawing/utils.py
```diff
@@ -55,20 +59,30 @@ def contrast_stretching(data, saturated_pixels):
...
         data = np.ma.masked_less_equal(data, 0).compressed()
-    vmin = np.nanpercentile(data, saturated_pixels / 2.)
-    vmax = np.nanpercentile(data, 100 - saturated_pixels / 2.)
+
+    # If vmin, vmax are float or int, we keep the value, if not we calculate
+    # the precentile value
+    if not isinstance(vmin, (float, int)):
+        vmin = np.nanpercentile(data, _parse_value(vmin, 'vmin'))
+    if not isinstance(vmax, (float, int)):
+        vmax = np.nanpercentile(data, 100 - _parse_value(vmax, 'vmax'))
```
francisco-dlp: Wouldn't it be more intuitive to use `vmax="95th"` for the current equivalent of `vmax="5th"`. If yes, the widgets could use `IntRangeSlider` instead of `IntSlider`

hyperspy/defaults_parser.py
```diff
@@ -138,11 +138,16 @@ class GUIs(t.HasTraits):
 
 
 class PlotConfig(t.HasTraits):
-    saturated_pixels = t.CFloat(0.05,
```
francisco-dlp, line 141: I think that we shouldn't break the API for this release.

* I suggest to deprecate ``saturated_pixels`` instead for removal in 2.0.
* Change its default value to 0 (it will only affect new users with no ``hyperspyrc`` file.

I think that it was a poor idea to have this in preferences since there is no default value that is always good, so I think that it is better to remove it for good.

francisco-dlp (changes requested):
It looks very good now except for a couple of suggestions and 1 issue: the ``Reset`` button does not seem to work.

ericpre:
I don't think I will have time today and tomorrow to finish this PR for the 1.6 release, if nobody can take over, we can leave it for next release.

francisco-dlp:
I can take over, but, do you agree with my comments?

ericpre:
> I can take over, but, do you agree with my comments?

Yes, I agree.

> Wouldn't it be more intuitive to use vmax="95th" for the current equivalent of vmax="5th".

I wasn't sure about this and initially, I may have been influenced by the `saturated_pixel` approach but since this is not going to be used anymore, using `vmax="95th"` may be better.

francisco-dlp:
OK, thanks. I'll send you a PR in the afternoon.

hyperspy/misc/array_tools.py
```diff
@@ -389,15 +389,20 @@ def calculate_bins_histogram(data):
...
 
-    return max(bins_sturges, bins_fd)
+    if fd_bin_width:
+        bin_width = min(fd_bin_width, sturges_bin_width)
+    else:
+        # limited variance: fd_bin_width may be zero
+        bin_width = sturges_bin_width
+
+    # Cap at 250 to avoid bad surprised when getting very large number
+    return min(int(np.ceil(data.ptp() / bin_width)), 250)
```
tjof2, line 405: See #2456 - this function is removed in that PR.

ericpre:
The image comparison should be fixed now.

hyperspy/drawing/utils.py
ericpre: This is now done in https://github.com/hyperspy/hyperspy_gui_ipywidgets/pull/25.

hyperspy/misc/array_tools.py
tjof2: Indeed this function is now gone on Rnm

ericpre: Yes and this PR is for Rnp, so this should be sorted when merging Rnp into Rnm?

ericpre:
Shall we finish this PR before #2356?

francisco-dlp:
Certainly. It LGTM, although it would be good to increase a bit more the coverage.

PR #2284, opened 2019-11-21
Fix beta tilt

### LINKED ISSUE #2089: The take off angle function does not work for double tilt sample holder

ZanettaPM:
Just a small correction on the beta_tilt implementation to fix https://github.com/hyperspy/hyperspy/issues/2089

This PR allow to take into account float numbers and correct a misspelling

### DIFF
--- a/hyperspy/_signals/eds.py
+++ b/hyperspy/_signals/eds.py
@@ -716,10 +716,7 @@ class EDS_mixin:
         tilt_stage = mp.Stage.tilt_alpha
         azimuth_angle = mp.Detector.EDS.azimuth_angle
         elevation_angle = mp.Detector.EDS.elevation_angle
-        if 'beta_tilt' not in mp:
-            beta_tilt = 0.0
-        else:
-            beta_tilt = mp.Stage.tilt_beta
+        beta_tilt = mp.get_item('Stage.tilt_beta', 0.0)
 
         TOA = utils.eds.take_off_angle(tilt_stage, azimuth_angle,
                                        elevation_angle,beta_tilt)
--- a/hyperspy/_signals/eds.py
+++ b/hyperspy/_signals/eds.py
@@ -713,15 +713,17 @@ class EDS_mixin:
         elif self.metadata.Signal.signal_type == "EDS_TEM":
             mp = self.metadata.Acquisition_instrument.TEM
 
-        tilt_stage = mp.Stage.tilt_alpha
-        azimuth_angle = mp.Detector.EDS.azimuth_angle
-        elevation_angle = mp.Detector.EDS.elevation_angle
+        tilt_stage = mp.get_item('Stage.tilt_alpha', None)
+        azimuth_angle = mp.get_item('Detector.EDS.azimuth_angle', None)
+        elevation_angle = mp.get_item('Detector.EDS.elevation_angle', None)
         beta_tilt = mp.get_item('Stage.tilt_beta', 0.0)
 
-        TOA = utils.eds.take_off_angle(tilt_stage, azimuth_angle,
-                                       elevation_angle,beta_tilt)
-
-        return TOA
+        return utils.eds.take_off_angle(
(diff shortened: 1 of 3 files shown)


hyperspy/_signals/eds.py
```diff
@@ -706,7 +706,7 @@ def get_take_off_angle(self):
         tilt_stage = mp.Stage.tilt_alpha
         azimuth_angle = mp.Detector.EDS.azimuth_angle
         elevation_angle = mp.Detector.EDS.elevation_angle
-        if 'beta_tilt' not in mp:
+        if 'tilt_beta' not in mp.Stage:
             beta_tilt = 0.0
         else:
             beta_tilt = mp.Stage.tilt_beta
```
ericpre: This could be simplified to:
```python
beta_tilt = mp.get_item('Stage.tilt_beta', 0.0)
```

hyperspy/misc/eds/utils.py
```diff
@@ -270,8 +270,8 @@ def take_off_angle(tilt_stage,
     28.865971201155283
     """
 
-    alpha = math.radians(tilt_stage)
-    beta = -math.radians(beta_tilt)
+    alpha = math.radians(float(tilt_stage))
```
ericpre: Add a test for this bug?

hyperspy/_signals/eds.py
```diff
@@ -706,7 +706,7 @@ def get_take_off_angle(self):
         tilt_stage = mp.Stage.tilt_alpha
         azimuth_angle = mp.Detector.EDS.azimuth_angle
         elevation_angle = mp.Detector.EDS.elevation_angle
-        if 'beta_tilt' not in mp:
+        if 'tilt_beta' not in mp.Stage:
```
ericpre: It needs a test.

ericpre (commented):
Thanks for the bug fix.
It needs tests and I have suggested one change. Some commits should not be in this PR and some additional files are picked up. The easiest to deal with the unwanted commits would be to start a new branch from RELEASE_next_minor.

hyperspy/_signals/eds.py
```diff
@@ -706,7 +706,7 @@ def get_take_off_angle(self):
         tilt_stage = mp.Stage.tilt_alpha
         azimuth_angle = mp.Detector.EDS.azimuth_angle
         elevation_angle = mp.Detector.EDS.elevation_angle
```
ericpre: Catch the error when one of the two attributes is present and raise an error with a more user friendly message, for example, something along the line of:
The take off angle can not be calculated because  the XX angle is not defined in the metadata. Use `set_microscope_parameters` to set it.

ZanettaPM:
> Thanks for the bug fix.
> It needs tests and I have suggested one change. Some commits should not be in this PR and some additional files are picked up. The easiest to deal with the unwanted commits would be to start a new branch from RELEASE_next_minor.

Yes it is why I did, I don't really get why these commits still shows up. I'll try to fix that and add a test next week

ericpre:
> Yes it is why I did, I don't really get why these commits still shows up.

This branch was check out from `RELEASE_next_minor` about two years ago - see first commit of this PR: https://github.com/hyperspy/hyperspy/pull/2284/commits/8965d5f7f30ab80d17527c8cd39705578600dd0d. Make sure you check out from the last upsteam  `RELEASE_next_minor`  branch.

ericpre:
There are commits which aren't necessary in this PR and I think we should rebase the useful commits on the RnP? @tjof2, what do you think?

tjof2:
Agreed. Happy to take a look later this afternoon

ericpre:
Great thanks @tjof2!

tjof2:
I'm not quite sure why the `float()` was needed, and it was inconsistently applied anyway.

PR #2288, opened 2019-12-04
Add ``__getitem__`` method to ROIs

francisco-dlp:
### Description of the change

Add ``__getitem__`` method to the following ROIs:

* ``Point1DROI``
* ``Point2DROI``
* ``SpanROI``
* ``RectangularROI``

In this way, they can be used in place of tuples.

### Progress of the PR
- [x] add tests,
- [x] ready for review.

```python
>>> import hyperspy.api as hs
>>> import numpy as np
>>> im = hs.signals.Signal2D(np.random.random((10,30,30))
>>> roi = hs.roi.RectangularROI(left=2, right=10, top=0, bottom=5))
>>> tuple(roi)
(2.0, 10.0, 0.0, 5.0)
>>> im.align2D(roi=roi)
```

### DIFF
--- a/hyperspy/models/model1d.py
+++ b/hyperspy/models/model1d.py
@@ -895,7 +895,9 @@ class Model1D(BaseModel):
             If 'interactive' the signal range is selected using the span
              selector on the spectrum plot. The signal range can also
              be manually specified by passing a tuple of floats. If None
-             the current signal range is used.
+             the current signal range is used. Note that ROIs can be used
+             in place of a tuple.
+
         estimate_parameters : bool, default True
             If True will check if the component has an
             estimate_parameters function, and use it to estimate the
--- a/hyperspy/models/model1d.py
+++ b/hyperspy/models/model1d.py
@@ -29,7 +29,6 @@ from hyperspy.drawing.widgets import VerticalLineWidget, LabelWidget
 from hyperspy.events import EventSuppressor
 from hyperspy.signal_tools import SpanSelectorInSignal1D
 from hyperspy.ui_registry import add_gui_method, DISPLAY_DT, TOOLKIT_DT
-from hyperspy.misc.utils import signal_range_from_roi
 
 
 @add_gui_method(toolkey="hyperspy.Model1D.fit_component")
@@ -463,7 +462,7 @@ class Model1D(BaseModel):
 
         """
         try:
-            x1, x2 = signal_range_from_roi(x1)
+            x1, x2 = x1
         except TypeError:
             # It was not a ROI, we carry on
             pass
@@ -494,7 +493,7 @@ class Model1D(BaseModel):
 
         """
         try:
-            x1, x2 = signal_range_from_roi(x1)
+            x1, x2 = x1
         except TypeError:
             # It was not a ROI, we carry on
             pass
@@ -529,7 +528,7 @@ class Model1D(BaseModel):
 
         """
         try:
-            x1, x2 = signal_range_from_roi(x1)
+            x1, x2 = x1
         except TypeError:
             # It was not a ROI, we carry on
             pass
@@ -872,7 +871,6 @@ class Model1D(BaseModel):
             display=True,
             toolkit=None,
             **kwargs):
-        signal_range = signal_range_from_roi(signal_range)
         component = self._get_component(component)
         cf = ComponentFit(self, component, signal_range,
                           estimate_parameters, fit_independent,
--- a/hyperspy/roi.py
+++ b/hyperspy/roi.py
@@ -238,6 +238,9 @@ class BaseROI(t.HasTraits):
 
         return axes_out
 
+    def __getitem__(self, *args, **kwargs):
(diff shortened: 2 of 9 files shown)


ericpre:
This is very nice. A few comments/questions:
- Any reason not to do it for all ROIs?
- Since it makes `signal_range_from_roi` not necessary anymore, should it be deprecated?
- I guess a couple of docstring could be updated to mention ROI can be passed to argument like, `signal_range` or `roi`.
- Related to previous point, make this kind of argument (`signal_range`, `roi`, etc.) consistent across different method in which case we may do this in `Release_next_major` to avoid introducing more deprecation...

How far do you want to go in this PR?

francisco-dlp:
> - Any reason not to do it for all ROIs?

Just that it is not straightforwardly useful to get the ``(x, y, radious, innerradious)`` values of the ``CircleROI`` or the ``(x1, y1, x2, y2)`` coordinates of ``Line2DROI`` in tuple format. Do you think that it is worth adding it nevertheless for consistency?

> - Since it makes `signal_range_from_roi` not necessary anymore, should it be deprecated?

Thanks, done!

> - I guess a couple of docstring could be updated to mention ROI can be passed to argument like, `signal_range` or `roi`.

Done.

> Related to previous point, make this kind of argument (`signal_range`, `roi`, etc.) consistent across different method in which case we may do this in `Release_next_major` to avoid introducing more deprecation...

Yes, let's do that (where reasonable) in 2.0. There is also an inconsistency across HyperSpy in the order of ``"left", "right", "top", bottom"`` as different functions use different conventions. Something for 2.0 I think.

francisco-dlp:
After thinking about it I have added the feature for all ROIs, and I have taken a simpler, faster approach for the implementation.

hyperspy/models/model1d.py
```diff
@@ -463,7 +462,7 @@ def set_signal_range(self, x1=None, x2=None):
 
         """
         try:
-            x1, x2 = signal_range_from_roi(x1)
+            x1, x2 = x1
         except TypeError:
```
ericpre, line 466: Should the error below become a `ValueError` in case the length of the tuple doesn't match?

hyperspy/models/model1d.py
```diff
@@ -494,7 +493,7 @@ def remove_signal_range(self, x1=None, x2=None):
 
         """
         try:
-            x1, x2 = signal_range_from_roi(x1)
+            x1, x2 = x1
```
ericpre, line 496: Same as above.

hyperspy/models/model1d.py
```diff
@@ -529,7 +528,7 @@ def add_signal_range(self, x1=None, x2=None):
...
```
ericpre, line 531: Same as above.

hyperspy/roi.py
```diff
@@ -979,6 +1038,7 @@ def __init__(self, x1, y1, x2, y2, linewidth=0):
         super(Line2DROI, self).__init__()
         self.x1, self.y1, self.x2, self.y2 = x1, y1, x2, y2
         self.linewidth = linewidth
+        self._tuple = (self.x1, self.y1, self.x2, self.y2)
```
ericpre: `linewidth`?

ericpre (commented):
Looks good to me. I added a few other comments.

hyperspy/models/model1d.py
francisco-dlp: I am not sure about it: I think that if the length of the tuple doesn't match we want to raise the exception because the only valid inputs are a matching ROI or a number.

francisco-dlp:
Thanks for the review. It should be ready for another iteration.

hyperspy/roi.py
```diff
@@ -980,6 +1069,12 @@ def __init__(self, x1, y1, x2, y2, linewidth=0):
         self.x1, self.y1, self.x2, self.y2 = x1, y1, x2, y2
         self.linewidth = linewidth
 
+    def __getitem__(self, *args, **kwargs):
+        _tuple = (self.x1, self.y1, self.x2, self.y2, self.linewidth)
+        _tuple = (self.x1, self.y1, self.x2, self.y2)
```
ericpre: I guess this line should be removed?

francisco-dlp: Yes, thanks. It turns out that I forgot to add tests for Circle and Line2D. After adding them in the last commit I found 1 more bug of course...

hyperspy/models/model1d.py
ericpre: Yes, I agree. Actually, it looks to me that it is not possible to get a `TypeError` with/without this PR? If so, shall we removed it?

francisco-dlp: I think that it is fine as is: you can get a ``TypeError`` by passing a number as argument.

ericpre: Yes, indeed!

PR #2289, opened 2019-12-06
Add support for reading EMPAD files.

ericpre:
It needs smallest possible test file to be ready for review. If someone can provide some, this would be helpful. In the mean time, I open this PR, in case it can be of interested.

### Progress of the PR
- [x]  Add EMPAD reader,
- [x] add tests,
- [x] ready for review.


woozey:
Smallest I have is 66.6 MB, let me know if you want to have this one.

ericpre:
@woozey, instead of adding a test file, it is now creating the raw data before the test and only the xml file is original.
The data I have are not calibrated, so I am not sure if reading the calibration of the signal dimension is working correctly.
Could you please share some of your file? So I can checked if there are any difference in the xml file, etc. Are they calibrated (signal dimension)?

hyperspy/io_plugins/empad.py
```diff
@@ -0,0 +1,181 @@
...
+            'record-by':'image'}   
+    if om.has_item('root.count'):
+        # Stack of images
+        info.update({'series_count':int(om.root.scan_parameters.series_count)})
+    elif om.has_item('root.pix_x') and om.has_item('root.pix_y'):
+        # 2D x 2D
+        info.update({'scan_x':int(om.root.pix_x),
+                     'scan_y':int(om.root.pix_y)})
+    else:
+        raise IOError("Unsupported Empad file: the scan parameters can not "
```
tjof2: "cannot" rather than "can not" :)

hyperspy/io_plugins/empad.py
```diff
@@ -0,0 +1,181 @@
+# -*- coding: utf-8 -*-
+# Copyright 2007-2016 The HyperSpy developers
```
dnjohnstone: 2007-2020

dnjohnstone (approved):
This basically looks good to me - a stupid thing about dates, but otherwise ok and would be good to have in. 

Sorting the merge conflicts and then verifying that the tests are ok (current failures are due to the old tifffile issue) would make this ready to merge for me.

tjof2 (approved):

ericpre:
@tjof2, if you are happy with this PR, can you please merge it! ;)

francisco-dlp:
Why this doesn't have an entry in the User Guide?

ericpre:
It does! It was missing from this PR but it has been rectified in https://github.com/hyperspy/hyperspy/pull/2439.

francisco-dlp:
Great, thanks!

PR #2292, opened 2019-12-15
Enh pointers as rois

kociak:
This PR concerns the user interface and aims at facilitating the browsing of Signals
### Description of the change
Using ROIs instead of pointers as a standard for signals.
Making a signal plot with an arbitrary number of spectra
Allowing the generation of filtered maps by using SpanRois on the new signal plot
Speeding up the drawing of the spectra and filtered maps

### Progress of the PR
- [X] Modifying mpl_hse and Signal1DFigure to handle the new signal_plot and filtered maps
- [X] Generating filtered maps
- [] Map generation relies on a Hack of mpl_hse and should be properly implemented (probably creating a new subclass of mpl_hse)
- [] Fixing the bug introduced in Signal1DFigure when trying to use Signal1DFigure for other applications than plotting spectra (probably creating a new subclass of Signal1DFIgure)
-[] Handling Signals with navigation shape of 1
-[] Handling Signal2D
- [X] increase the update of spectra related to ROIs
- [X] increase the update of filtered maps related to SpanRois
- [] fix bugs related to spectra update
- [] speed updates are still hacks of the BlittedFigure, and should be properly implemented
- [] speed updates of the Rois and SpanRois widgets should be implemented 
- [ ] update docstring,
- [ ] update user guide,
- [ ] add tests,
- [ ] ready for review.

```python
>>> import hyperspy.api as hs
>>> import numpy as np
>>> #this is a small spectrum image to start with
>>> s=hs.signals.Signal1D(np.random.random((128,128,100)))
>>> #with no slices, this is quite fast
>>> toto.plot(number_of_rois=2,number_of_slices=0)
>>> #close the previous plot, then try with slices (remove the sharp sign)
>>> toto.plot(number_of_rois=2,number_of_slices=2)
>>> #not that fast. But the filtered maps are updated quite fast when playing with the slices, although the spanRoi don't
>>> #you can try with more rois, and/or more slices. Slices do slow down more than rois
>>> #I put an arbitrary number of maximum rois (7)
>>> #toto.plot(number_of_rois=7,number_of_slices=7)
```


kociak:
After a RL discussion with @francisco-dlp , I have cleaned up all the original code I had commented instead of deleting it.
I hope it is now much more clear, but in case it is not, just let me know!

kociak:
I have merged 1.7.0 dev into my PR branch.
The PR is ready for review. It is not entirely compatible with standard use of hyperspy. Indeed, I noticed I introduced some bugs in the spanROIs when using removebackground, but for this one I need the help of contributors that understand/wrote the related code (@ericpre ???). Also, if the PR passes, the plan is to improve the ergonomy (being able to add arbitrary ROI shapes and a minimal interface to add/remove ROIs, in particular).

ericpre (commented):
This is a feature which is long overdue in HyperSpy! I think that there is too much in this PR and we needs to start a new PR focusing on the following tasks first:
- Add a ROI as pointer
- Add support for accepting ROI to the `navigator` argument, to add a ROI automatically to the signal dimension and its `sum` (or any other function) would make the navigator. I would expect that it should a minor change, since [the navigator is already interactive](https://github.com/hyperspy/hyperspy/blob/512d129d4959ad62f602763f953e831b649ae3a8/hyperspy/signal.py#L2607-L2626).

Others features/improvements should be done in a separate PR, because as I understand it, I don't think it would be part of the `MPL_HyperExplorer` and others. These objects are already more than complicated enough!

*Note that https://github.com/hyperspy/hyperspy/pull/2615 fixes a bug with a ROI, which misled me in thinking that ROI has significantly more overhead than drawing widgets in my previous attempt at addressing this missing feature in #1668. I have learnt quite a bit since then!!

@kociak, what do you think? I am very keen to get this features added, so I am more than happy to review and help.

kociak:
This make perfect sense, thanks!  Re-reading my code after a year, it is clear that I am not using the right hyperspy tools. Also, I see indeed  #2615 fixing some issues I had that made me try to hack (without great success and probably reasons...) the `blitter` functions. Before starting a new, leaner version of the PR, I'll rewrite your proposal @ericpre (essentially to make sure I don't get confused with the object namings):

Considering first `Signal1D`'s, we want the hyperspy user to be able to have, on the navigator, a (navigation) ROI as a pointer. The "signal" window will then display the summation of all the signal in the ROI, i.e, a 1D array (a spectrum, say). At the same time, the content of the navigator, which usually is a (ìnteractive) `sum` over the whole signal dimension, should be more appropriately a summation over a given slice of in the signal direction. This slice would be defined with a signal ROI in the signal window, and in the case of Signal1D`'s, a `SpanROI`.
Btw, written like this, extension to `Signal2D`'s might be trivial and part of the new PR.

Also, two stupid questions:
- At the time being, there is a possibility to have _two_ pointers. Just to make it sure: we do want to keep this feature, right?
-  do I need to close this very PR before starting a new one?

In any case, I will start to simplify the code and focus on the adequation of my changes to hyperspy architecture, waiting for confirmation of the above points.

ericpre:
> Also, two stupid questions:
> - At the time being, there is a possibility to have _two_ pointers. Just to make it sure: we do want to keep this feature, right?

Yes, indeed and once we have something simple and clear in place, we can consider more than 2 pointers. If I remember correctly, you previously pointed out a recent matplotlib functionality which allow to add more vertical axis.

> -  do I need to close this very PR before starting a new one?

It doesn't matter really. The only thing which is useful is to have link between relevant PRs, which is very convenient when going through some of these discussion or previous work.

With changing the pointer, a lot of reference image will need to be updated and the [plot testing](https://hyperspy.readthedocs.io/en/latest/dev_guide/testing.html#plot-testing) are going to fail when running the plotting test with `--mpl`. We currently use a specific version of [``freetype`` and ``matplotlib`` ](https://github.com/hyperspy/hyperspy/blob/512d129d4959ad62f602763f953e831b649ae3a8/conda_environment_dev.yml#L6-L8) to make sure the test are reproducible but at some point, we need to update the reference images at a more recent version. If this is a bit annoying, I can help to sort this out, passing the plotting test without the ``--mpl`` option will be already a good start!

kociak:
Thanks. For sure I'll need help for the graphical tests (tests might be what will delay my next submission as I never done proper tests in my coding life, but I know this is super important, so I'll try my best) and I appreciate your offer. Let's got back to code now!

ericpre:
In this case, plotting is fairly well covered and it shouldn't be necessary to add much more test (only to adjust them). You can check when something is broken by running `pytest` in the `hyperspy/hyperspy/tests/drawing` to run the plotting test only.

ericpre:
In this case, plotting is fairly well covered and it shouldn't be necessary to add much more test (only to adjust them). You can check when something is broken by running `pytest` in the `hyperspy/hyperspy/tests/drawing` to run the plotting test only.

kociak:
OK. I have though about how to do it, and I think I will even split the PR in two. One for the ROI as pointer, and one for "slicing", ie, getting the navigation to display a slice of the signal, instead of the default sum. Actually, 1/ they serve two different purposes 2/they won't be implemented the same so 3/ they should be tested independently. I will spend some time on making the split nicely ...

ericpre:
Usually, it is good and useful to start a new PR after a first exploratory iteration!

ericpre:
I have merged #2615 to avoid conflicts with your upcoming PR and so that you can use the improvement it includes!

PR #2293, opened 2019-12-19
Fix fast background

askorikov:
### Description of the change
Fixed incorrect background subtraction using "fast" option due to the missing scaling for binned signals. Updated version of #2264 with a correct target branch and added tests.

### Progress of the PR
- [x] Change implemented,
- [x] Ready for review.

```python
>>> from hyperspy.components1d import PowerLaw
>>> from hyperspy.signals import Signal1D
>>> import numpy as np

# Generate a power law spectrum
>>> s = Signal1D(np.empty((100,)))
>>> axis = s.axes_manager.signal_axes[0]
>>> axis.scale = 0.02
>>> axis.offset = 1
>>> g1 = PowerLaw(50015.156, 1.2)
>>> s.data = g1.function(axis.axis)
>>> s.metadata.Signal.binned = True

# Subtract background using "fast" option
>>> s2 = s.remove_background(signal_range=[1.0, 2.0], background_type='PowerLaw', fast=True)

# Assert that the result is relatively close to zero
>>> assert(np.linalg.norm(s2.data) / np.linalg.norm(s.data) < 0.05)
```

### DIFF
--- a/hyperspy/_signals/signal1d.py
+++ b/hyperspy/_signals/signal1d.py
@@ -1044,8 +1044,10 @@ class Signal1D(BaseSignal, CommonSignal1D):
             only_current=False)
         if fast and not self._lazy:
             try:
-                axis = self.axes_manager.signal_axes[0].axis
-                result = self - background_estimator.function_nd(axis)
+                axis = self.axes_manager.signal_axes[0]
+                scale_factor = axis.scale if self.metadata.Signal.binned else 1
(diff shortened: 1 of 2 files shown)


askorikov:
Not sure why exactly the tests fail, but now on my computer the same errors also get generated while running tests for the current state of RELEASE_next_patch

ericpre:
I have seen the same failure on a different PR and this is most likely related to an updated library.

francisco-dlp:
LGTM, merging.

PR #2299, opened 2020-01-08
Add `x` keyword, ``crop`` and ``_slice_me`` to `FunctionalDataAxis` and fixes

francisco-dlp:
### Description of the change

This PR is based on @jlaehne 's https://github.com/ericpre/hyperspy/pull/17 and is inspired by the original approach to solve the same problem in https://github.com/ericpre/hyperspy/pull/11. Originally I intended to fix a couple of things in https://github.com/ericpre/hyperspy/pull/17 , but then I realized that the original approach taken by @jlaehne  in  https://github.com/ericpre/hyperspy/pull/11 (that I argued against in https://github.com/ericpre/hyperspy/pull/11#issuecomment-568157095 and led @jlaehne to implement https://github.com/ericpre/hyperspy/pull/17) was better (my apologies @jlaehne).

With this PR a ``FunctionalDataAxis`` can be constructed:
* With an array, ``x``, and an expression
* With ``size`` and ``expression``, in which case ``x`` becomes ``np.arange(size)`` and is updated if ``size`` changes.

This also includes:

* Fixes for ``_slice_me`` and ``crop``
* A method to convert a linear axis to a FDA, see example below.
* The expression parameters are now traits
*  Initializing the expression parameters is now mandatory

This is WIP. I'll work on the documentation and tests once we decide if this is the right way to implement ``FunctionalDataAxis``.

### Progress of the PR
- [ ] add tests,
- [ ] ready for review.

```python
>>> import hyperspy.api as hs
>>> s = hs.datasets.eelsdb(title="Magnesium")[0]
>>> s.axes_manager[0].convert_to_functional_data_axis(expression="1.2398E3 / x", units="nm", name="wavelength")
>>> s.isig[::-1].plot()
```
Figure_Magnesium_carbonate_Signal


francisco-dlp:
I have changed the type of the ``x`` parameter of ``FunctionalDataAxis`` from array to ``BaseDataAxis``. This has the advantage of using the same mechanism as ``LinearDataAxis`` for cropping and slicing when ``x`` is a ``LinearDataAxis``. Another advantage is that we could now use ``x`` to plot a secondary axes using the [new feature](https://matplotlib.org/gallery/subplots_axes_and_figures/secondary_axis.html) in matplotlib.

hyperspy/axes.py
```diff
@@ -688,66 +682,57 @@ def crop(self, start=None, end=None):
...
+                 size=t.Undefined,
                  **parameters):
         super().__init__(index_in_array, name, units, navigate)
-        self.size = size
-
-        self.parameters_list = [key for key in parameters.keys()]
+        if x is None:
+            if size is t.Undefined:
+                raise ValueError("Please provide either `x` or `size`.")
+            self.x = LinearDataAxis(scale=1, offset=0, size=size)
```
jlaehne: Defaulting `offset=1` would avoid `divide by zero` error when using a reciprocal function.

jlaehne (commented):
OK, seems to make more sense now. Slicing starts to work also for FDA, but as commented an `isig` operation modifies `x` permanently.

francisco-dlp:
Good catch @jlaehne, I'll fix it as soon as I have time this week.

Regarding ``offset=1``, I agree that that would prevent getting an exception, but exceptions are sometimes good because it get users thinking on what they are doing. After thinking on this from the users point of view, I think that the best approach is not to have a default axis. In this way users are obliged to know what they're doing and they won't get an exception unless they get it wrong. What do you think?

jlaehne:
Good point that we do not need to pamper the user too much.

Bye, the way - nice solution with `x = BaseDataAxis` and the potential for dual axis plotting.

jlaehne:
@francisco-dlp I lsot a bit track, did you fix the issue with `isig`?

hyperspy/exceptions.py
```diff
@@ -190,6 +190,15 @@ def __init__(self, navigation_size, expected_navigation_size):
             self.navigation_size, self.expected_navigation_size)
 
 
+class NonLinearAxisError(Exception):
+
+    def __init__(self):
+        self.error = ("Non linear axis are not supported.")
```
tjof2: Should be either "axis is" or "axes are"

jlaehne:
There is still a problem with `isig`. For this minimal example:
```
S = hs.signals.Signal1D(data=np.arange(10))
S.axes_manager[0].convert_to_functional_data_axis(expression = '1 / (x+1)')
S.isig[0.6:0.1].plot()
S.isig[0.6:0.1].plot()
```
Plot works the first time, but not the second time.
`S.isig[0.6:0.1].data.size` returns `8` no matter how often you call it.
However, `S.isig[0.6:0.1].axes_manager[0].axis.size` will decrease by `1` for every time you call it (It always looses the first value of the axis vector!).

francisco-dlp:
Thanks @jlaehne, I think that it should be fixed now.

Notice also that I've just changed every mention to *non-linear* axes to non-uniform axes, that is the most widespread denomination.

ericpre:
Should this PR go to the `non_uniform_axes` branch?

doc/user_guide/axes.rst
```diff
@@ -0,0 +1,484 @@
+Axes handling
+*************
+
+
+The navigation and signal dimensions
+------------------------------------
+
+In HyperSpy, the data is interpreted as a signal array and, therefore, the data
+axes are not equivalent. HyperSpy distinguishes between *signal* and
```
ericpre: I would remove the first sentence because it is not clear at all.

doc/user_guide/axes.rst
```diff
@@ -0,0 +1,484 @@
...
+*navigation* axes and most functions operate on the *signal* axes and
+iterate over the *navigation* axes. For example, an EELS spectrum image (i.e.
```
ericpre: Make the EELS spectrum image example more generic.

doc/user_guide/axes.rst
```diff
@@ -0,0 +1,484 @@
...
+in the Signal2D subclass.
+
+.. code-block:: python
+
+    >>> im = hs.signals.Signal2D(np.zeros((30, 10, 20)))
+    >>> im
+    <Signal2D, title: , dimensions: (30|20, 10)>
+
+
+Axes storage and ordering
```
ericpre: I would move this section to the end, because this is a written for a more advanced usage of the `DatAxis`.

doc/user_guide/axes.rst
```diff
@@ -0,0 +1,484 @@
...
+convention is `F order` (column-major), where it is the other way round: the
+first axis of an array is the fastest to iterate over. In both cases, the
+further an axis is from the `fast axis` the slower it is to iterate over this
+axis. In the book analogy, you could think about reading the first lines of
+all pages, then the second and so on.
+
+When data is acquired sequentially, it is usually stored in acquisition order.
+When a dataset is loaded, HyperSpy generally stores it in memory in the same
+order, which is good for the computer. However, HyperSpy will reorder and
+classify the axes to make it easier for humans. Let's imagine a single numpy
```
ericpre: I am not sure this is correct for all IO plugins? Mention the `optimize` argument of some functions?

doc/user_guide/axes.rst
```diff
@@ -0,0 +1,484 @@
...
+  used in plotting. The latter enables the :ref:`conversion of units
+  <quantity_and_converting_units>`.
+* ``navigate`` (bool) determines, whether it is a navigation axis.
+* ``size`` (int) gives the number of elements in an axis.
+* ``index`` (int) determines the "current position for a navigation axis and
+  ``value`` (float) returns the value at this position.
+* ``low_index`` (int) and ``high_index`` (int) are the first and last index.
+* ``low_value`` (int) and ``high_value`` (int) are the smallest and largest
+  value.
+* The ``axis`` (array) vector stores the values of the axis points. However,
```
ericpre: I think using the term `vector` will be confusing for many users, because they will think of it as an spatial/euclidean vector and not as a vector in the numpy sense. If we keep it, it needs to be explained.
In my opinion, we should mentioned it when introducing what the `axis` attribute is but this should be secondary, not be a defining element of `axis` and not necessary to understand `vector` in this context to understand the rest.

hyperspy/drawing/mpl_he.py
```diff
@@ -99,7 +99,7 @@ def plot_navigator(self,
             sl = signal1d.Signal1DLine()
             sl.data_function = self.navigator_data_function
             sl.set_line_properties(color='blue',
-                                   type='step')
+                                   type='step' if axis.is_uniform else 'line')
```
ericpre, line 102: Why changing 'step' to `line`?

hyperspy/signal.py
```diff
@@ -2086,9 +2087,17 @@ def get_dynamic_explorer_wrapper(*args, **kwargs):
                 return navigator()
 
         if not isinstance(navigator, BaseSignal) and navigator == "auto":
-            if (self.axes_manager.navigation_dimension == 1 and
+            if (self.axes_manager.navigation_dimension > 1 and
+                    np.any(np.array([not axis.is_uniform for axis in
```
ericpre, line 2090: This should be documented.

hyperspy/signal.py
```diff
@@ -3552,14 +3628,16 @@ def ifft(self, shift=None, return_real=True, **kwargs):
 
         Note
         ----
-        For further information see the documentation of
-        :py:func:`numpy.fft.ifftn`
+        Requires a linear axis. For further information see the documentation
```
ericpre: linear -> uniform

hyperspy/signal.py
```diff
@@ -3937,25 +4015,35 @@ def map(
         >>> sigmas = hs.signals.BaseSignal(np.linspace(2,5,10)).T
         >>> im.map(scipy.ndimage.gaussian_filter, sigma=sigmas)
 
+        Note
+        ----
+        Currently requires a linear axis.
```
ericpre: linear -> uniform

hyperspy/signal.py
```diff
@@ -3937,25 +4015,35 @@ def map(
...
             _logger.warning(
-                "The function you applied does not take into "
-                "account the difference of units and of scales in-between"
-                " axes.")
+                "At least one axis of the signal is non-uniform. Can your "
+                "`function` operate on non-uniform axes?")
+        else:
+            # Check if the signal axes have inhomogeneous scales and/or units and
+            # display in warning if yes.
+            scale = set()
```
ericpre: A bit of matter of taste, but I usually find list comprehension more readable:
```python
scale = set([axis.scale for axis in self.axes_manager.signal_axes])
```

hyperspy/axes.py
```diff
@@ -1343,24 +1743,33 @@ def _get_dimension_str(self):
     def __repr__(self):
         text = ('<Axes manager, axes: %s>\n' %
                 self._get_dimension_str())
-        ax_signature = "% 16s | %6g | %6s | %7.2g | %7.2g | %6s "
+        ax_signature_uniform = "% 16s | %6g | %6s | %7.2g | %7.2g | %6s "
+        ax_signature_non_uniform = "% 16s | %6g | %6s |  non-uniform axis  | %6s "
```
ericpre: There is one space too much with `non-uniform axis`: the columns are not aligned between uniform and non-uniform axis.

ericpre (commented):
I have lost track of this but i have tried to catch up and comments on this PR. Most of my comment are most likely not related to this PR, in which case, we come back to them in another PR.

I can do another review once this is ready.

jlaehne:
> Thanks @jlaehne, I think that it should be fixed now.
> 
> Notice also that I've just changed every mention to _non-linear_ axes to non-uniform axes, that is the most widespread denomination.

Looks good now.

hyperspy/axes.py
jlaehne: OK

hyperspy/signal.py
jlaehne: OK

hyperspy/signal.py
jlaehne: OK

doc/user_guide/axes.rst
jlaehne: using array instead of vector

doc/user_guide/axes.rst
jlaehne: OK

doc/user_guide/axes.rst
jlaehne: I think a specific example is helpful, added a list of other examples and also the 2D case.

doc/user_guide/axes.rst
jlaehne: OK

hyperspy/exceptions.py
jlaehne: OK

jlaehne:
I opened a PR to this branch that addresses part of the issues raised by @ericpre and the typo found by @tjof2. Also it corrects more mentions of the 'linear axis' to 'uniform axis'

jlaehne:
Looking again at the terminology, I find that both "non uniform" and "non linear" seem to be used. Google even has more hits for the latter, but I agree that "non uniform" sounds more intuitive, because we are basically talking about non uniform data spacing along the axis.

jlaehne:
Working on examples for the documentation, I realized that only the following works, which is a bit weird as you have to define an axis in the dictionary:
```
dict0 = {'expression': 'a / x + b', 'a': 100, 'b': 10, 'x': UniformDataAxis(size=10,offset=10,scale=0.1)}
s = hs.signals.Signal1D(np.ones(500), axes=[dict0])
```

Might make sense to add the functionality to define a nested dictionary (which might actually be needed for IO functionality?) so that the following would also work:
```
dict0 = {'expression': 'a / x + b', 'a': 100, 'b': 10, 'x': {'size': 10, 'offset': 10, 'scale': 0.1}}
s = hs.signals.Signal1D(np.ones(500), axes=[dict0])
```
Similarly, `get_axis_dictionary` currently returns: `... 'x': <Unnamed axis, size: 500>` and it might make sense to have it output the dictionary of that axis? (which you of course can already get if you specifically ask for the dictionary of the `x`-attribute: `s.axes_manager[0].x.get_axis_dictionary()`)

In a way that is linked to what #2345 is about, to be able to also initialize the signal using: 
```
axis = FunctionalDataAxis[expression='a / x + b', a=100, b=10, x=UniformDataAxis(size=10,offset=10,scale=0.1))
s = hs.signals.Signal1D(np.ones(500), axes=[axis])
```

francisco-dlp:
Regarding non-linear vs non-uniform, you are right @jlaehne. The nomenclature issue is that the axes are not linear or uniform. We probably should speak instead of *non-linear axis scale* and *non-uniform axis spacing*. I was inclined to use non-uniform because, in the multi-dimensional case, people speak about functions that operate on non-uniform grids e.g. [NFFT](https://en.wikipedia.org/wiki/Non-uniform_discrete_Fourier_transform). So, I was concerned that, by not using non-uniform, people where going to miss that part of the literature. However, now I don't know which one is better.

Concerning how to initialize `FunctionalDataAxis`, I think that you are right too, but I have to think a bit more about it.

jlaehne:
Well you can consider a logarithmic scale to be non-linear as well, even if your data is on a linear grid. Thinking about it, I am happy with non-uniform. It was just that the concept was running as non-linear for so long. Any oppinion from @ericpre?

I realized that we currently mix 'non-uniform' and 'non uniform'. We should probably stick to the former. Should I try to make that consistent across the files?

francisco-dlp:
I am starting to think that this is not at all ready to release even if we mention in the documentation that these features are in beta-state. Not only there are still a large number of functions that don't work on non-uniform axes, but also I think that the axes themselves need some more thought and work. My main current concern is that the current implementation assumes that the data represents variables sampled at discrete points. However, when `metadata.Signal.binned=True`, that no longer holds. For data on a linear scale is doesn't matter because we can always divide by the scale, but for non-linear scales we need to store more information in the axes. In particular, we must store the left and right position of the bins or one of those and their size. The fact that this is not yet implemented is very concerning because we know that, if we implement saving of signals with non-uniform axes now, we'll need to change the format again as soon as we implement the axes properly.

So, what about going back to the original plan of not including this in v1.6 and making a beta release of v1.7 as soon as we implement properly the feature for binned data? Both conda-forge and pypi support beta releases, so it shouldn't be a problem for external packages that require non-uniform axes to depend on a beta release.

ericpre:
> Well you can consider a logarithmic scale to be non-linear as well, even if your data is on a linear grid. Thinking about it, I am happy with non-uniform. It was just that the concept was running as non-linear for so long. Any oppinion from @ericpre?

I don't have any strong opinion on this, I would tend to agree that "non-uniform" may be better.

> 
> I realized that we currently mix 'non-uniform' and 'non uniform'. We should probably stick to the former. Should I try to make that consistent across the files?

Yes, indeed "non-unifrom axis" is more correct than "non uniform axis" and making the documentation more consistent is never a bad thing! :)


> So, what about going back to the original plan of not including this in v1.6 and making a beta release of v1.7 as soon as we implement properly the feature for binned data? Both conda-forge and pypi support beta releases, so it shouldn't be a problem for external packages that require non-uniform axes to depend on a beta release.

I agree on the binning issue and on the fact we should avoid making a format version and implement saving file from the beginning.

Even if the the non linear axis is not fully working at the moment, it doesn't break anything else as far as we know, so I am not convinced we need to do a separate beta release? Doing a normal single release would be more manageable and save us time but also user can use it more readily. The only reason we may need to do a beta release may be for 2.0 because they will be a lot of API changes and we may breaks a few things... so user will need to opt in to use the beta version and they will know they may be taking risk to have things broken!

jlaehne:
> Yes, indeed "non-uniform axis" is more correct than "non uniform axis" and making the documentation more consistent is never a bad thing! :)

Well, it were just a few instances in the end, I included the changes in a PR to this branch working on the FDA documentation.

> I agree on the binning issue and on the fact we should avoid making a format version and implement saving file from the beginning.

Ah, I did not have binned signals in mind as I am currently not using them. For the rest, I really had the impression that we are closing in and could use the momentum of the upcoming release to get it done. I have a few colleagues hesitant to adopt hyperspy as long as this functionality is not available and it pops up in the issues every now and then as well.

Do we really need to permanently store the bins in the axis or would it be enough to calculate the bins when needed by providing functions (e.g. `get_bins`) for `DataAxis` and `FunctionalDataAxis` classes, respectively, that calculate the bins from the axis (start value and size) when needed? The latter could simply be added later without changing the API or IO functions, while for the moment we raise exceptions when operating on binned signals. The advantage would also be that we do not have to add so much code to axis.py and dinstinguish in how binned and unbinned signals are saved. The metadata flag would remain the only difference. The functions where it is relevant would have to dinstinguish the two types, but they have to do so anyway.

francisco-dlp:
@ericpre, I agree with you, a beta release will be so rarely used that it is probably not worth the effort. Also, it is not reasonable to wait until all  the features work properly with non-uniform axes as this will take a long time. Also, by including it in a standard release we will boost the development of the non-uniform axes features, since users will be motivated to contribute their fixes once they discover that they are 1 feature away of getting their analysis done ;) On the other hand, in its current state, I think that too many things are broken and some important parts are half-cooked.

My main concern are "binned" signals, which are incidentally the most common kind of signals among HyperSpy users. Currently even `plot` is broken for them, since in their case it is misleading to plot e.g. counts vs energy when the bin size changes. We should plot counts / bin_size vs energy instead. This means that people cannot even use HyperSpy to load a dataset a have a look at it in a reliable way. Image plots suffer from the extra issue that currently HyperSpy plots them on an uniform grid. I think that we should usematplotlib's [NonUniformImage](https://matplotlib.org/3.2.1/gallery/images_contours_and_fields/image_nonuniform.html) or the better documented `pcolormesh` instead.

Hence, I will be comfortable releasing this when:

- [ ] Plot works properly for binned signals and images
- [ ] Binned versions of the axes are implemented
- [ ] Saving the axes in the hspy format is implemented
- [ ] Curve fitting on binned signals works (currently it doesn't)
- [ ] Most things that don't work on non-uniform axes are identified and an informative `NotImplementedError` is raised.

Low hanging remaining issues that we could consider fixing before releasing: 

- [ ] Implement `fft` and `ifft` using `nfft`. I like https://github.com/jakevdp/nfft
- [ ] `split`
- [ ] (some) ROIs:
    - [ ] `RectangularROI`
    - [ ] `SpanROI`
    - [ ] `Point2DROI`

The question is how to proceed. Should we releave v1.6 this week and include this in v1.7 or wait until the 4 main issues above (and any more that you may identify) are addressed to release v1.6?

ericpre:
Sounds like a very sensible plan to me.
Indeed, `DataAxis` needs to become binning aware and simply adding a `bin_width` attribute of the same shape as `axis` should be enough? But it also means that the generic functions using binning need to be adjusted accordingly.

francisco-dlp:
@jlaehne, for `FunctionalDataAxis` there is no need to store anything extra, but for `DataAxis` there is no way around it. Instead of creating a separate class for "binned" data axes, we could just add an `is_binned` attribute. Actually I think that we should deprecate `metadata.Signal.binned` since this is an attribute of the axes, not the signal. For example, if we transpose the signal and the navigation axes were not *binned* then the signal will no longer be "binned".

I agree with the argument of using the momentum of the release to get most of this done, but this inevitably means that we have to further delay v1.6 if you agree with the points of my previous post. We have to make sure that others agree.

ericpre:
I don't think I really answered the question... ;) Sort out binning signal for plotting, saving and fitting for v1.6 sounds sensible to me.

francisco-dlp:
OK, I'll point to this discussion in #2251 in case somebody argues against further delaying v1.6.

To proceed I suggest to

1. Merge this as-is in the `non-uniform-axes`
2. Open an issue to track the progress on the checklist above
3. Use the tracking issue to discuss how to best implement the features before starting work on it

francisco-dlp:
Tracking issue: #2398 

Do you agree to merge this into `non_uniform_axes` as-is?

ericpre:
Yes. Is there anything more you want to do in this PR?

jlaehne:
From my POV the issues that this PR was about were addressed, so please do merge with `non_uniform_axes`. And we can then launch new PRs. But it is important to have the new concept for FDA merged.

I will continue to contribute to the discussion in #2398.

PR #2300, opened 2020-01-11
Fix warning test suite

ericpre:
There is an issue with the warnings of the test suite on travis (see https://travis-ci.org/hyperspy/hyperspy/builds/634148599). Fixing the warning seems to be fix this error.

### Progress of the PR
- [x] raise `NotImplementedError` for bss of lazy signal,
- [x] replace the use of `np.alltrue` by `np.allclose`,
- [x] catch irrelevant warning,
- [x] ready for review.

### DIFF
--- a/hyperspy/learn/mva.py
+++ b/hyperspy/learn/mva.py
@@ -537,6 +537,9 @@ class MVA():
 
         """
         from hyperspy.signal import BaseSignal
+        if self._lazy:
+            raise NotImplementedError("Blind source separation is not "
+                                      "implemented for lazy signal.")
 
         lr = self.learning_results
 
--- a/hyperspy/learn/mva.py
+++ b/hyperspy/learn/mva.py
@@ -21,6 +21,7 @@ import types
 import logging
 
 import numpy as np
+import dask.array as da
 import matplotlib.pyplot as plt
 from matplotlib.ticker import MaxNLocator, FuncFormatter
 try:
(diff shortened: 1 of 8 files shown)


hyperspy/tests/mva/test_bss.py
```diff
@@ -122,12 +122,12 @@ def test_mask_diff_order_0_on_loadings(self):
...
         # This test, unlike most other tests, either passes or raises an error.
         # It is designed to test if the mask is correctly dilated inside the
-        # `blind_source_separation_method`. If the mask is not correctely
+        # `blind_source_separation_method`. If the mask is not correctly
         # dilated the nan in the loadings should raise an error.
         mask = self.s._get_navigation_signal(dtype="bool")
         mask.isig[5] = True
         self.s.learning_results.loadings[5, :] = np.nan
-        self.s.blind_source_separation(3, diff_order=1, mask=mask,
+        self.s.blind_source_separation(2, diff_order=1, mask=mask,
```
francisco-dlp: While this is true, it is very inconvenient not to be able to perform BSS on lazy signals. We could instead compute the factors used for BSS, perform BSS and store the results in dask arrays to make `get_bss_model` (the potentially memory intensive step) work lazily.

ericpre: Do you mean that the BSS does not need to be compute lazily, since the data have already been reduced? If so, it should have been already working, including `get_bss_model`!
It should be a bit better now and also better tested.

hyperspy/learn/mva.py
```diff
@@ -514,7 +521,7 @@ def blind_source_separation(self,
         comp_list : boolen numpy array
             choose the components to use by the boolean list. It permits
             to choose non contiguous components.
-        mask : bool numpy array or Signal instance.
+        mask : :py:class:`~hyperspy.signal.BaseSignal` (or subclass)
```
ericpre, line 524: Shall we do the same (allow only signal and not numpy array as mask) for decomposition?

francisco-dlp: Probably yes because it simplifies the code what protect us from some bugs. However, I would leave that for another PR.

PR #2304, opened 2020-01-16
hspy read support for non linear axis and more error catches

jlaehne:
### Description of the change
[moved here from ericpre#11, see there for initial discussion]
- Fixed `get_axis_dictionary` for FunctionalDataAxis so that nested dictionary is returned. Signal can also be initialized with such a nested dictionary.
- Adapted hspy-reader so that non-uniform `DataAxis` or `FunctionalDataAxis` are correctly read back into a signal object (by using `hdfgroup2dict`).
- Added `non_uniform_axis` to plugin characteristics, and added error message on write for file formats where this property is `False` (so far all but `hspy`)
- `create_axis()` now also accepts lists of axis objects and not only lists of dictionaries.

### Progress of the PR
- [X] Fixed FDA axis-dictionary,
- [X] Added signal initialization by passing axis object (as alternative to dict),
- [X] Enable reading of non uniform axes for hspy-format,
- [X] non_uniform_axis io-plugin characteristic,
- [X] Debug,
- [X] Redefine way axes is saved (add `_type`)
- [X] make `hyperspy.axes.create_axis` public
- [x] Add tests,
- [X] Update docstrings,
- [x] Update documentation,
- [x] ready for review.

### Minimal example of the functionality
```
from hyperspy.axes import FunctionalDataAxis, UniformDataAxis
data=np.arange(10)
ax0 = UniformDataAxis(size=data.size,offset=1)
expression = "a / x"
axis = FunctionalDataAxis(size=ax0.size, expression=expression,  a=1, x=ax0)
S = hs.signals.Signal1D(data, axes=(axis, ))
S.axes_manager[0].get_axis_dictionary()
S.save('test.hspy')
Srel = hs.load('test.hspy')
```


jlaehne:
moved here from ericpre#11, see there for discussion

jlaehne:
@francisco-dlp I'm a bit stuck on how to proceed on this one. Would I have to include the type of axis already in the axis dictionary, or would it be enough to add this during export?

jlaehne:
We still wanted to change the way the axes is stored (see https://github.com/ericpre/hyperspy/pull/11#issuecomment-568157095) and I am a bit stuck on where exactly to implement that as the read/write functions are very recursively defined. @ericpre, @francisco-dlp any suggestions, or otherwise you might be able to implement it much faster?

jlaehne:
Concerning the storage of axes. Currently, the hdf5 writer creates a subgroup for each axes and names them `axes-0, axes-1, ...`. The dictionary equivalent would be:
```
{'axes-0': {'name': ...}, 'axes-1': {'name': ...}, ...}
```
The type of axis is not saved, but automatically determined from the attributes, which are present. For HyperSpy that is OK, but to be clearer to other packages that want to read our file format it may not be a bad idea to include this information.

If we want to explicitly store the type of axis as suggested in https://github.com/ericpre/hyperspy/pull/11#issuecomment-568157095, I see several possibilities:

- Either we change the axis dictionary so that it includes the type in form of a nested dictionary: `{'FunctionalDataAxis' : {'name' : ...}}`. Though maybe the most elegant, it completely breaks the current API (not only concerning IO) and will need quite a lot of tweaks to keep it backwards compatible.
- When saving the signal, we add the type of axis as an extra subgroup so that we save e.g. 
`{'axes-0': {'FunctionalDataAxis' : {'name' : ...}}}` instead of `{'axes-0': {'name': ...}}`, but basically when reading in, we would drop that again, as `create_axis` automatically determines the type of axis. It could be potentially helpful for other readers that want to read our format.
-  Alternatively, we add an attribute `_type = self.__class__.__name__` to the axis object (and thus to the dictionary). It would the equivalent to `_id_name` for a Component. That way the type is at least contained in the saved object, but we avoid the unnecessary layer of dictionary nesting. It would have to be optional in the creation of an axis object to be backwards compatible, but if missing would be automatically set from the types of present attributes.

What do you think: @francisco-dlp

francisco-dlp:
I like the last option for its simplicity and consistency with how serialization is implemented for other HyperSpy objects. What do you think @ericpre?

jlaehne:
> I like the last option for its simplicity and consistency with how serialization is implemented for other HyperSpy objects. What do you think @ericpre?

OK, I implemented that, plus one of the suggestions from #2345

If you agree to these changes, I would adapt the docstrings and documentation and see if I should write more tests.

hyperspy/axes.py
```diff
@@ -598,12 +599,14 @@ class DataAxis(BaseDataAxis):
 
     def __init__(self,
                  index_in_array=None,
+                 _type=t.Undefined,
```
francisco-dlp: What about adding ``**kwargs`` instead? Then the constructor could check if the `_type` corresponds with the class and raise a ``ValueError`` if it doesn't.

ericpre: As I understand, `_type` is not used but it is here to allow creating axis from the output of `get_axis_dictionary()`. However, this doesn't work very well, because we need to know the class before hand and instead it is better to use the `_create_axis` function to create an axis. Would it be better to remove the `_type` of all the constructor and make the `_create_axis` function public and use this function when creating axis from the `get_axis_dictionary()`? It will need a deprecation warning.

Using the `**kwargs` approach may be a bit annoying with `FunctionalDataAxis` since it will need to be removed before passing it the `Expression` component.

hyperspy/axes.py
```diff
@@ -709,9 +713,13 @@ def __init__(self,
                 raise ValueError("Please provide either `x` or `size`.")
             self.x = UniformDataAxis(scale=1, offset=0, size=size)
         else:
-            self.x = x
-            self.size = self.x.size
+            if type(x) is dict:
```
ericpre: Prefer `if isinstance(x, dict):`?

hyperspy/axes.py
```diff
@@ -1345,7 +1355,10 @@ def create_axes(self, axes_list):
         if len(indices) == len(axes_list):
             axes_list.sort(key=lambda x: x['index_in_array'])
         for axis_dict in axes_list:
-            self._append_axis(**axis_dict)
+            if type(axis_dict) is dict:
```
ericpre: Similar as above, prefer `if isinstance(axis_dict, dict):`?

ericpre:
> I like the last option for its simplicity and consistency with how serialization is implemented for other HyperSpy objects. What do you think @ericpre?

Agree, adding the `_type` attribute and having it in the dictionary (from `get_axis_dictionary()`) is better.

hyperspy/axes.py
jlaehne: For hyperspy itself, and axis initialization in particular, `_type` is not really needed. `_create_axis` recognizes the desired type of axis from the attributes (expression -> FDA, axis -> DA, else UDA). At the moment, the public version is `create_axes`, which takes a list of dictionaries instead of a single one. So the introduction of `_type` is only to ensure that a saved `.hspy` file would contain that bit of information (and that it can be more easily accessed than via `__class__.__name__`).

Having it in the `**kwargs` allows for a consistency check between the automatically defined type and the given one.  However, when e.g. converting between axes types, it then has to be removed from the dictionary or explicitly set to the correct type. As it is right now, any given `_type` would simply be ignored and the automatically determined one is instead used when a new axis is created.

jlaehne:
I implemented `_type` via `**kwargs` and a check for consistency between `dict._type` and the axis being created. This also required the axis-conversion functions (e.g. `convert_to_uniform_axis`) to explicitly change the value in the dict.

Also fixed the tests failing due to `_type` being added to the dictionary. And added legacy saving (`legacy=True`) of signals with only uniform axes, so that they can be read by older versions of HyperSpy.

jlaehne:
Any thoughts/comments @ericpre, @francisco-dlp (also in relation to https://github.com/hyperspy/hyperspy/pull/2304#discussion_r433982173). If you agree on this implementation, I would update the docstrings/user guide.

hyperspy/axes.py
ericpre: We agree on the approach but I am not convinced that we should have the `_type` attribute since it is not used; it is necessary only when exporting an axis as a dictionary, so it should be defined only then.

Also `create_axes` is a method of the `AxesManager` class and I think it may be still convenient to create an axis independently of the `AxesManager` using a public method. I don't have any particular use case in mind, but creating axis from `DataAxis` is used in two places in hyperspy.

hyperspy/axes.py
```diff
@@ -853,7 +865,12 @@ def __init__(self,
         self.size = size
         self.update_axis()
         self._is_uniform = True
-        self._type = self.__class__.__name__
+        if '_type' in kwargs:
```
ericpre: Same as above.

hyperspy/axes.py
```diff
@@ -719,7 +724,14 @@ def __init__(self,
                 self.x = x
                 self.size = self.x.size
         self._expression = expression
-        self._type = self.__class__.__name__
+        if '_type' in parameters:
```
ericpre, line 726: It should not be necessary to repeat it if **kwargs is passed to `super().__init__(...)`.

hyperspy/axes.py
```diff
@@ -1017,12 +1034,14 @@ def convert_to_functional_data_axis(self, expression, units=None, name=None, **k
         d.update(kwargs)
         this_kwargs = self.get_axis_dictionary()
         self.__class__ = FunctionalDataAxis
+        d["_type"] = 'FunctionalDataAxis'
```
ericpre, line 1025: Same as above.

hyperspy/axes.py
```diff
@@ -1017,12 +1034,14 @@ def convert_to_functional_data_axis(self, expression, units=None, name=None, **k
...
         self.__init__(expression=expression, x=UniformDataAxis(**this_kwargs), **d)
         self.axes_manager = axes_manager
 
     def convert_to_non_uniform_axis(self):
         d = super().get_axis_dictionary()
         self.__class__ = DataAxis
+        d["_type"] = 'DataAxis'
```
ericpre, line 1032: Same as above.

hyperspy/axes.py
```diff
@@ -592,21 +592,27 @@ def convert_to_uniform_axis(self):
             scale_err = max(self.axis[1:] - self.axis[:-1]) - scale
             _logger.warning('The maximum scale error is {}.'.format(scale_err))
         self.__class__ = UniformDataAxis
+        d["_type"] = 'UniformDataAxis'
```
ericpre, line 599: I don't think this is necessary - but I haven't check, so I may be wrong! ;)

hyperspy/io_plugins/hspy.py
```diff
@@ -696,6 +702,16 @@ def write_signal(signal, group, **kwds):
 
     for axis in signal.axes_manager._axes:
         axis_dict = axis.get_axis_dictionary()
+        if 'legacy' in kwds:
```
ericpre: I am not convinced that this is worth supporting saving to an old format?

ericpre (commented):
A couple of suggestion to simplify the code.

hyperspy/io_plugins/hspy.py
jlaehne: Legacy support is something @francisco-dlp  had brought up.

hyperspy/axes.py
jlaehne: It is necessary only, if `_type` is an attribute, as otherwise it would keep the old `_type`. But if we put the `_type` in the export only, then it will be obsolete here and at a few other places, of course.

hyperspy/axes.py
jlaehne: Good point, otherwise it is just an alias for `__class__.__name__`.

hyperspy/axes.py
jlaehne: would be obsolete, if we throw out the attribute and use it only on export

hyperspy/axes.py
ericpre: > It is necessary only, if `_type` is an attribute, as otherwise it would keep the old `_type`.

It should get overwritten in the constructor, this is why I suspect it should not be necessary.

jlaehne: The constructor checks for consistency between the automatically chosen type and the passed parameter. Had been suggested by @francisco-dlp when proposing the **kwargs solution. During conversion relying on a dictionary this has to be explicitly set (or alternatively deleted). Otherwise the consstructor throws an error.

hyperspy/axes.py
jlaehne: good point

hyperspy/axes.py
jlaehne: No, it is not obsolete, as long as `_type` is part of the dictionary and we have the consistency check.

hyperspy/axes.py
jlaehne: done, is only in super() now

hyperspy/axes.py
jlaehne: done

hyperspy/axes.py
jlaehne: see below

jlaehne:
@ericpre, @francisco-dlp I removed `_type` from the attributes. It is still part of the exported dictionary and a consistency check is done if '_type' is part of the **kwargs passed to the constructor (i.e. when working with a loaded signal that might in the future not necessarily originate from hyperspy).

jlaehne:
@ericpre raised the question whether `_create_axis` should become a public instead of a private method: https://github.com/hyperspy/hyperspy/pull/2304/files/70b90d4e1c8cd7278874ca4538e7281c30eaeaaa#r441692447

I added this private method to the user guide, but it would make sense to convert it to a public one. But should we do that here or in a separate PR?

jlaehne:
The current failure is on `TestPowerLaw.test_exception_gradient_calculation`. I don't think it is related to this PR.

jlaehne:
@ericpre, @franciscodlp, could you have another look at this PR? i would like to have it merged into `non_uniform_axis` before starting on the other features such as binning.

hyperspy/io_plugins/hspy.py
ericpre: I think the following is clear enough and significantly better than introducing a new saving argument:
https://github.com/hyperspy/hyperspy/blob/60610f4c400ea648f9a95ab5e0dea3a0be670e74/hyperspy/io_plugins/hspy.py#L127-L132

doc/user_guide/axes.rst
```diff
@@ -380,12 +391,13 @@ Corresponding output of :py:class:`~.axes.AxesManager`:
 Defining a new axis
 -------------------
 
-An axis object can be created through the ``axes.create_axis()`` method, which
+An axis object can be created through the ``axes._create_axis()`` method, which
```
ericpre, line 394: If useful enough to be in the user guide, it should be a public method.

hyperspy/io_plugins/hspy.py
ericpre: I think supporting legacy saving is not necessary because the following is clear enough and significantly better than introducing a new saving argument:
https://github.com/hyperspy/hyperspy/blob/60610f4c400ea648f9a95ab5e0dea3a0be670e74/hyperspy/io_plugins/hspy.py#L127-L132

ericpre:
> @ericpre raised the question whether `_create_axis` should become a public instead of a private method: https://github.com/hyperspy/hyperspy/pull/2304/files/70b90d4e1c8cd7278874ca4538e7281c30eaeaaa#r441692447
> 
> I added this private method to the user guide, but it would make sense to convert it to a public one. But should we do that here or in a separate PR?

Doing it in this PR is fine, however, rather than `hyperspy.axes.DataAxis.create_axis`, a `hyperspy.axes.create_axis` public function may be good!

doc/user_guide/axes.rst
```diff
@@ -404,17 +417,61 @@ method:
...
+Note that there is currently no consistency check whether a signal object has
+the right number of axes of the right dimensions. Most functions will however
+fail if you pass a signal object where the axes do not match the data 
+dimensions and shape.
+
+**Add a single axis** to the `AxesManager` of a signal using ``_axes.append()``:
+
+.. code-block:: python
+    
+    >>> s.axes_manager._axes.append(evaxis)
```
ericpre: Only public API should be documented in the user guide.

hyperspy/io_plugins/hspy.py
jlaehne: OK, that makes sense; removed the "legacy" option and related tests.

doc/user_guide/axes.rst
jlaehne: OK, I guess it is enough that `axes_manager.create_axes` can append a list of axes dictionaries or axes objects, so we do not need to have an extra public function for single axes.

However, the `axes_manager.create_axes` is in fact rather an `append_axes` function and might be confused with the `axes.create_axis` function. On the other hand, I would refrain from renaming such a public function not to break any scripts.

jlaehne:
> Doing it in this PR is fine, however, rather than `hyperspy.axes.DataAxis.create_axis`, a `hyperspy.axes.create_axis` public function may be good!

Done. It was already in the uppermost level of the axes class and not a function of DataAxis.

doc/user_guide/axes.rst
```diff
@@ -404,17 +417,55 @@ method:
...
+Usually, the axes are directly added to a signal during :ref:`signal 
+initialization<signal_initialization>`. However, you may wish to add/remove
+axes from the `AxesManager` of a signal.
+
+Note that there is currently no consistency check whether a signal object has
+the right number of axes of the right dimensions. Most functions will however
+fail if you pass a signal object where the axes do not match the data 
+dimensions and shape.
+
+You can **add a set of axes** to the `AxesManager` by passing either a list of
```
ericpre: Any reason to write "add a set of axes" in bold? I find it a bit too much, since there are very few things in the user guide which are in bold. Maybe italic would be enough if you want to highlight this part.
```
You can *add a set of axes* to the :py:class:`~.axes.AxesManager` by passing either a list of
axes dictionaries to :py:meth:`~.axes.AxesManager.create_axes`:
```

doc/user_guide/axes.rst
```diff
@@ -404,17 +417,55 @@ method:
...
+or a list of axes objects:
+
+.. code-block:: python
+
+    >>> from hyperspy.axes import UniformDataAxis, DataAxis
+    >>> axis0 = UniformDataAxis(offset=300,scale=1,size=500)
+    >>> axis1 = DataAxis(axis=np.arange(12)**2)
+    >>> s.axes_manager.create_axes([axis0,axis1])
+
+**Remove an axis** from the `AxesManager` using ``remove()``, e.g. for the last axis:
```
ericpre: Same as above:
```
*Remove an axis* from the :py:class:`~.axes.AxesManager` using :py:meth:`~.axes.AxesManager.remove`,  e.g. for the last axis:
```

ericpre (approved):
Looks good to me, a couple of formatting tweaks for the user guide maybe?

ericpre:
> > Doing it in this PR is fine, however, rather than `hyperspy.axes.DataAxis.create_axis`, a `hyperspy.axes.create_axis` public function may be good!
> 
> Done. It was already in the uppermost level of the axes class and not a function of DataAxis.

Sorry, I misread it!

doc/user_guide/axes.rst
jlaehne: ok

doc/user_guide/axes.rst
jlaehne: ok

jlaehne:
@francisco-dlp, any more comments on this one?

jlaehne:
The current error is in `test_eels.py` and was inherited from upstream.

ericpre:
> The current error is in `test_eels.py` and was inherited from upstream.

Yes, indeed but these failing tests are working fine in other development branches, so it means that it needs to be fixed in the `non_uniform_axes` but agreed not necessarily in this PR.

@francisco-dlp, please let us know if you want to review this PR again. If not, we can merge it to make @jlaehne's life easier and save him having to deal with too many branches in parallel!

jlaehne:
@ericpre @francisco-dlp can this be merged?

PR #2306, opened 2020-01-17
Gallery examples

### LINKED ISSUE #1069: Archlinux installation instructions are old
As the title says, the archlinux installation instructions are old / deprecated (written for python2 as well). As it stands now, they should be at least removed from the user guide, preferably updated with a working set of steps.

### LINKED ISSUE #1952: Galleries of example
Although almost all functionalities of HyperSpy are documented in the user guide, it is usually not straightforward for users to find them in the documentation. We could try to address this issue by making a galleries of example, in the same this way is done by other libraries, for example:
- http://scikit-image.org/docs/stable/auto_examples
- https://matplotlib.org/gallery/index.html
- https://mne.tools/stable/auto_examples/index.html

This would be complementary to the user guide and the notebook demos.

## List of potential examples:
- import a spectrum from a `txt` file using `np.loadtxt` and calibrate the signal axis.
- export a spectrum using the `msa` format (similar to `csv` format)
- Make interactive figure from different signal/ROI or comparison signal/model (see https://gitter.im/hyperspy/hyperspy?at=6009c1ca004fab47415777df)
- https://github.com/hyperspy/hyperspy/issues/3127#issuecomment-1509669262
- https://github.com/hyperspy/hyperspy/issues/2335
- rotate data in navigation space (`s2 = s.T.map(rotate, angle=45, inplace=False).T` where `rotate` comes from `scipy.ndimage`)
- https://github.com/hyperspy/hyperspy/pull/3227

ericpre:
Add a gallery of examples, using `sphinx-galley`: https://sphinx-gallery.github.io/stable/index.html.

### Progress of the PR
- [x] Setup the sphinx-gallery,
- [x] tidy up old (unused?) sphinx extension,
- [x] recycle the examples of the folder `examples`,
- [x] add example import/export single spectrum, 
- [x] tweak toctree of the doc,
- [x] closes #1069,
- [x] ready for review.

It is not ready to merge, but it would be good to have some feedback on this because continuing.

To improve/do:
- [ ] generate automatically a preview image for examples having a plot
- [ ] clarify the purpose of the examples and how it relates with the rest of the documentation (see discussion below)
- [ ] add https://github.com/hyperspy/hyperspy/issues/2335#issuecomment-595365378
- [ ] add example on image conversion to `tif` and `jpg` to illustrate exporting for quick screening (with data rescaling) or raw data (see https://gitter.im/hyperspy/hyperspy?at=5e70bf06fbeef0035697fe5d)

Close #1952.


thomasaarholt:
Some tips for testers: 
Install sphinx-gallery (`conda install sphinx-gallery` or `pip install sphinx-gallery`)
`cd hyperspy/doc` and call `make html`, then navigate to `_build/index.html` and open that.

@ericpre, this is gorgeous! Should we prepend the examples filenames with `plot_`, or do you have something else in mind?

Also, wow, the examples are a bit old :p I don't think I've ever seen the `s.to_EELS()` method before!

dnjohnstone:
Shouldn't this kind of thing be in hyperspy-website or hyperspy-doc so that it doesn't have to be distributed with the main code and things can be kept lightweight? Or can we put it in some other repo?

ericpre:
This is part of the doc and it is built at the same time as the normal doc using `sphinx`. This is how it works and moving it to somewhere else would be a pain.
In any case, I think, it is sensible to include the doc in the source distribution similarly to the tests, etc.

The weight of the tarball comes from the reference tests file and there is not much to win with removing the doc from the source distribution.

It would be good to populate this with examples of simple tasks and having example of workflows in notebook?

examples/create_signal/README.rst
```diff
@@ -0,0 +1,4 @@
+Signal Creation and plotting
+============================
```
dnjohnstone, line 2: Creation --> creation for case consistency across examples

dnjohnstone (commented):
This basically looks good to me in the changes made - one very minor capital to lower case suggestion, but otherwise a definite positive step forward for this bit of the repo.

I think you're right about having it here and my previous comment about moving it out of the repo was wrong.

What did you particularly want feedback on to move it further forward? It's already kind of a big PR so might be good to have minimal further changes.

ericpre:
The main points of discussion would be:
- what examples would be suitable for this?
- is it worth having this (in comparison)? We already have the notebook demos repository, examples in the user guide and the docstrings...
- does it overlap with other documentation?

My motivation for this PR is to have something in place where we can put short examples, which will not fit elsewhere or will be not straightforward to find in the user guide. For example, import/exporting data, plotting two 1D signals on the same figure using the complex signal trick or the kind of things which are mentioned sometimes in the gitter or elsewhere.

Since it integrates fairly nicely in the documentation (just after the user guide), it should have a good visibility.

dnjohnstone:
Ok, my opinion is that this is a useful complement to other sections if we have a structure like:

1. Documentation/docstrings = usage of particular methods/functions (as it is already)
2. Notebooks = common processing workflows and tutorials related to workshops
3. Sphinx examples = FAQ code snippets

In other words, things should go here if they are a few line solution to a "How do I...?" question that we feel gets asked often.

Does that sound reasonable?

ericpre:
Agree, when I come back to this, I will try to clarify what is doing what in the documentation. It may be worth revisit the website at some point too.

ericpre:
This has been done in https://github.com/hyperspy/hyperspy/pull/3050.

PR #2309, opened 2020-01-23
Codecov

francisco-dlp:
Fix coverage report by enabling codecov.

We used to use coveralls, but it has been broken for several years (!), doesn't anybody miss it?


thomasaarholt:
Good idea! It looks like the 84% coverage listed on the hyperspy github page is just a relic! And here's I've been all content with the fact that we were 2% points ahead of numpy!

README.rst
```diff
@@ -8,8 +8,8 @@
 .. |AppVeyor| image:: https://ci.appveyor.com/api/projects/status/github/hyperspy/hyperspy?svg=true&branch=RELEASE_next_patch
 .. _AppVeyor: https://ci.appveyor.com/project/hyperspy/hyperspy/branch/RELEASE_next_patch
 
-.. |Coveralls| image:: https://coveralls.io/repos/github/hyperspy/hyperspy/badge.svg?branch=RELEASE_next_minor
-.. _Coveralls: https://coveralls.io/github/hyperspy/hyperspy?branch=RELEASE_next_minor
+.. |Codecov| image:: https://codecov.io/gh/hyperspy/hyperspy/branch/RELEASE_next_patch/graph/badge.svg
```
dnjohnstone, line 11: Which version of the code do we want the badges to point to? I think the main purpose of the badges is for it to look correct for the latest release when visiting the github page and on pypi. My approach would be for it to point to the master branch and for all releases to be made off of that. Anyway this is consistent with other badges here, so pragmatically I think it's ok.

dnjohnstone (commented):
Thanks @francisco-dlp I would merge this subject to resolution of my comment regarding which badge the link should point to.

Out of interest, what was the issue with coveralls? It seems to work fine for us in pyxem?

README.rst
dnjohnstone: Ignore me.... I forgot that this repo is structured unusually in terms of branch naming i.e. there is no master...

.travis.yml
```diff
@@ -85,3 +85,90 @@ deploy:
   on:
     tags: true
     condition: $MINIMAL_ENV = false
+
```
dnjohnstone: Are all these extra empty lines supposed to be here?

francisco-dlp: I don't know how they got in there. Fixed, thanks!

francisco-dlp:
Thanks for the review @dnjohnstone. There was nothing particularly wrong with coveralls, it was working for us until we broke it ourselves. We could have fixed it instead, but codecov seems to offer a few more possibilities these days e.g. an interesting browser extension https://docs.codecov.io/docs/browser-extension

dnjohnstone (approved):
Great, thanks! It's a please merge from me!

PR #2310, opened 2020-01-25
SAMFire 2D

dnjohnstone:
SAMFire currently only works for Model1D objects. This PR includes some stuff towards making it work for Model2D also.

This was something that we did get working with @to266 years ago, but since then lots of things changed, but also lots of things didn't. Anyway I tried to resolve merge conflicts correctly but there are certainly some things in this PR that should be split out in to other PRs or may no longer be relevant with Components moving to Expressions.

@pc494 you may be interested in this.


dnjohnstone:
In fact, I hope fixing this thing up could be a way to start a conversation regarding using similar ideas that inspired SAMFire to enhance some template matching thoughts @pc494 and I have been toying with in pyxem.

There is a general problem where:

(1) You can create a library of possible Signals that should match your measured Signals.
(2) You can define in a meaningful way the "distance" between the Signals in the library.
(3) You can define a correlation metric to determine which is the best matching signal.

The goal is to find the best match at each navigation position. The analogue to SAMFire is that rather than search through every library entry at every navigation position we can use the prior that it is likely neighbouring pixels will correspond to Signals that are nearby in the library.

We have effectively implemented an architecture for doing this without the "SAMFire" bit in pyxem.

I think that this is a candidate for upstreaming to hyperspy with appropriate generalization.

PR #2311, opened 2020-01-25
update to copyright 2020

dnjohnstone:
Just some minor houskeeping we've not kept on top of in the past few years.

### DIFF
--- a/doc/user_guide/bibliography.rst
+++ b/doc/user_guide/bibliography.rst
@@ -17,7 +17,7 @@ Bibliography
 
 :ref:`[Candes2011] <Candes2011>`
    E. Candes, X. Li, Y. Ma and J. Wright,
-   "Robust principal component analysis?" *J. ACM* 58(3) (2011): 1–37 
+   "Robust principal component analysis?" *J. ACM* 58(3) (2011): 1–37
    [`<https://arxiv.org/pdf/0912.3599.pdf>`_].
 
 .. _Chantler2005:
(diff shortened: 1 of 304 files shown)


ericpre:
I am not convinced there is a need to keep these copyright statement in each file. Looking at the major python projects, they put a docstring describing what is in the module instead, which is actually useful.

dnjohnstone:
> I am not convinced there is a need to keep these copyright statement in each file. Looking at the major python projects, they put a docstring describing what is in the module instead, which is actually useful.

I'm fine with getting rid of the copyright notices - but I think we should either get rid of them or have them standardized rather than the current situation of doing it a bit at random. Just needs a decision and a stated policy that is adhered to for development, historically it always seemed to be that we had this notice? 

Regarding module docstrings - certainly fine with that - the website suggests we follow the numpy docstring conventions (http://hyperspy.org/hyperspy-doc/current/dev_guide/writing_docs.html)  - if we actually did that then "Each module should have a docstring with at least a summary line. Other sections are optional, and should be used in the same order as for documenting functions when they are appropriate:" I think there are quite a few places where this standard is not kept in the code and we should improve that.

dnjohnstone:
@francisco-dlp what is the "official" hyperspy policy on these copyright statements?

francisco-dlp:
Those copyright statements are a requirement of the [GPL license](https://www.gnu.org/licenses/gpl-howto.en.html) I think. Thanks @dnjohnstone for updating them. We should probably think on automating this...

PR #2312, opened 2020-01-25
PEP8 style compliance

dnjohnstone:
The developer guide says that we follow PEP coding conventions: http://hyperspy.org/hyperspy-doc/current/dev_guide/coding_style.html

There are quite a few places where the code does not. I used autopep8 and a little script to improve the compliance of the code.


tjof2:
I use [black](https://black.readthedocs.io/en/stable/index.html) which is typically a bit stricter than using autopep8, but for consistency (" vs ', line lengths, etc) is IMHO unrivalled. Perhaps that's a discussion for another day...

dnjohnstone:
I'd be up for switching to black - I like strict consistency for this kind of thing.

We could do it in this PR, would do the following:

- [ ] update the documentation to say we use the Black code style
- [ ] Run the code through black

@francisco-dlp @ericpre - opinions?

hyperspy/api_nogui.py
```diff
@@ -3,17 +3,17 @@
 # Set the PyQt API to 2 to avoid incompatibilities between matplotlib
 # traitsui
 
+from hyperspy import docstrings
```
ericpre, line 6: Revert this changes: the logger needs to be imported and called before the rest.

hyperspy/samfire_utils/samfire_worker.py
```diff
@@ -27,8 +29,6 @@
 
 matplotlib.rcParams['backend'] = 'Agg'
```
ericpre, line 30: If the matplotlib backend is set before the import of hyperspy (and pyplot), I guess that there should be a reason. It not setting the matplotlib backend to `agg` should be removed.

ericpre (changes requested):
See comments on import.

francisco-dlp:
We've already passed the whole code through autopep8 once a while ago (just after the 1.0 release if I remember well). However, because not everybody uses ``autopep8`` and we don't check for compliance in the PRs, the situation today is not as good as it should be as this PR shows.

However, I think this is not the time to run the code through autopep8 again, because it makes merging the old PRs a nightmare. Since we are planning to merge as many PRs as possible before the split, I think that just before the split will be a good time to do this.

Once we go for it, I am also in favor of using black.

ericpre:
Black in combination with pre-commit would be good: https://black.readthedocs.io/en/stable/version_control_integration.html.

>However, I think this is not the time to run the code through autopep8 again, because it makes merging the old PRs a nightmare. Since we are planning to merge as many PRs as possible before the split, I think that just before the split will be a good time to do this.

I agree with postponing this for now to avoid creating additional merge conflicts.

dnjohnstone:
Given #2316 is now open and we will use black instead of autopep8 - I'm closing this.

PR #2314, opened 2020-01-25
Ragged mapping

dnjohnstone:
I've pulled out some ragged mapping bug fixes that were wrapped up in #1497

@ericpre wrote these - but I couldn't figure out how to separate them out and preserve authorship. Are you ok with these being included like this for a patch release?


tjof2:
I believe there's no test of ragged mapping without this being merged @dnjohnstone? #2391 nearly introduced a bug that would have broken it, I only spotted it by accident since there was no ragged test.

tjof2:
Since #1497 is nearly done @dnjohnstone can this be closed?

PR #2321, opened 2020-02-21
Fix #2206: "Centre='variables(/trials)' causes error"

### LINKED ISSUE #2206: Centre='variables(/trials)' causes error
When I apply centre='variables', the spectrum extracted from PCA_SI and SI in the same pixel is quite strange. Centre='trials' is the same.
```
SI.decomposition(normalize_poissonian_noise=True, centre='variables')
PCA_SI = SI.get_decomposition_model(8)
(SI + PCA_SI* 1j).plot()
```

Without **centre**
图片

With **centre='trials'**
图片

tjof2:
### Description of the change
Fixes #2206.

> @francisco-dlp commented on 12 Jun 2019
> Yes, `centre` and `normalize_poisson_noise` are not compatible. That should raise an error.

### Progress of the PR
- [x] Change implemented
- [x] update user guide

### Example

`s.decomposition(normalize_poissonian_noise=True, centre='variables')` now raises a ValueError:

```
normalize_poissonian_noise=True is only compatible with centre=None,
not centre='variables'.
```

### DIFF
--- a/doc/user_guide/mva.rst
+++ b/doc/user_guide/mva.rst
@@ -217,7 +217,9 @@ To perform Poissonian noise normalization:
      >>> s.decomposition(True)
 
 More details about the scaling procedure can be found in
-:ref:`[Keenan2004] <Keenan2004>`.
+:ref:`[Keenan2004] <Keenan2004>`. Note that  Poisson noise normalization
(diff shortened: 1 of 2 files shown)


ericpre:
Thanks @tjof2!

PR #2322, opened 2020-02-23
Fix import io plugins

ericpre:
### Description of the change
As mentioned by @thomasaarholt in https://gitter.im/hyperspy/hyperspy?at=5e4e922edafa3029f64113e9, the `try/expect` bloc may shadow an error. Since `h5py` and `pillow` are not optional dependency, there is no reason to catch and skip an `ImportError`. 

### Progress of the PR
- [x] Remove `try/expect` bloc for reader import `h5py` and `pillow`,
- [x] ready for review.

### DIFF
--- a/hyperspy/io_plugins/__init__.py
+++ b/hyperspy/io_plugins/__init__.py
@@ -21,14 +21,15 @@ import logging
 
 from hyperspy.io_plugins import (msa, digital_micrograph, fei, mrc, ripple,
                                  tiff, semper_unf, blockfile, dens, emd,
-                                 protochips, edax, bruker)
+                                 protochips, edax, bruker, hspy, image)
 
 
 io_plugins = [msa, digital_micrograph, fei, mrc, ripple, tiff, semper_unf,
-              blockfile, dens, emd, protochips, edax, bruker]
+              blockfile, dens, emd, protochips, edax, bruker, hspy, emd, image]
 
 
 _logger = logging.getLogger(__name__)
+
 
 try:
     from hyperspy.io_plugins import netcdf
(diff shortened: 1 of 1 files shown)


ericpre:
@thomasaarholt, to get a better idea of the issue, can you please try:

``` python
from hyperspy.io_plugins import hspy
from hyperspy.io_plugins import emd
```

and see if it raise an error.

thomasaarholt:
In `RELEASE_next_minor`, pyUSID is also part of that try block, see this [line](https://github.com/hyperspy/hyperspy/blob/49fa05c96ca0619615eeccc077f14ed51f1481f1/hyperspy/io_plugins/__init__.py#L46). Is it best to wait with this PR until _minor is merged back into _patch? (this was the cause of the warning that confused me)

ericpre:
@thomasaarholt, to be clear, the issue you mentioned in https://gitter.im/hyperspy/hyperspy?at=5e4e922edafa3029f64113e9 was caused by pyusid not being installed?
Indeed the warning was confusing and this is why it needs to be removed!

> In `RELEASE_next_minor`, pyUSID is also part of that try block, see this [line](https://github.com/hyperspy/hyperspy/blob/49fa05c96ca0619615eeccc077f14ed51f1481f1/hyperspy/io_plugins/__init__.py#L46). Is it best to wait with this PR until _minor is merged back into _patch? (this was the cause of the warning that confused me)

There should be a merge conflict in way or another.

thomasaarholt:
Yep, that is correct (pyusid not installed - which is logical since I pulled the latest Rnm). I was very confused initially until I looked up what could trigger the warning :)

PR #2327, opened 2020-02-29
Fix deprecation warning

ericpre:
### Progress of the PR
- [x] Specify `iterpath='serpentine'` in the test suite to avoid deprecation warning,
- [x] add numpy module to the Voigt components to avoid deprecation warning about `scipy.real`
- [x] ready for review.

### DIFF
--- a/hyperspy/tests/model/test_edsmodel.py
+++ b/hyperspy/tests/model/test_edsmodel.py
@@ -1,6 +1,5 @@
 
 import numpy as np
-import pytest
 
 from hyperspy.misc.test_utils import assert_warns
 from hyperspy.misc import utils
--- a/hyperspy/tests/model/test_edsmodel.py
+++ b/hyperspy/tests/model/test_edsmodel.py
@@ -248,9 +248,10 @@ class TestMaps:
 
     def test_lines_intensity(self):
         s = self.s
+
         m = s.create_model()
         # HyperSpy 2.0: remove setting iterpath='serpentine'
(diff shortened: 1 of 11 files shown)


hyperspy/tests/model/test_edsmodel.py
```diff
@@ -250,7 +250,8 @@ def setup_method(self, method):
     def test_lines_intensity(self):
         s = self.s
         m = s.create_model()
-        m.multifit()    # m.fit() is just too inaccurate
+        # HyperSpy 2.0: remove setting iterpath='serpentine'
+        m.multifit(iterpath='serpentine')    # m.fit() is just too inaccurate
```
tjof2: I know it's an old comment, but can we quantify what we mean by `# m.fit() is just too inaccurate`

ericpre: Indeed, this is a good point, that sounds odd to me and it would be good to do something about it!

ericpre: It look to me that this statement is not correct now and we can simply removed it. I added a test to confirm this but I think we could also remove this test comparing `fit` and `multifit`.
This comment was commited 4 years ago (https://github.com/hyperspy/hyperspy/commit/5b3b92f3ddf4788a08df86129dee1c7cdb881be8) and the commit message doesn't say what could the reason for this comment, so I don't see any reason to keep it, unless it is possible to demonstrate that it doesn't work.

tjof2: If the test is cheap to run, I'd vote to keep it.

PR #2328, opened 2020-02-29
Nexus io plugin

pquinn-dls:
### Description of the change

A nexus and general hdf reader
Allows user to - 
find datasets in a hdf  or nexus file
fine metadata in a hdf or nexus file
load nexus data as a signal  with all metadata
load a hdf dataset as a signal with metadata from file
save a nexus file with metadata
Nexus files also contain multiple datasets so we can search or load specific datasets from the file. 

As nexus and hdf is commonly used I also added a keyword to the hyperspy loader
"preferred_format=Nexus"  - if the file is hdf5 (h5 or related) and you wish to load it using the nexus format it will skip the hyperspy or usid hdf readers.

### Progress of the PR
- Code written and tests written
- [x] update docstrings,
- [x] ready for review.

```python

>>> import hyperspy.api as hs
>>> s1=hs.load("test.nxs")
>>> s2=hs.load("test.hdf",preferred_format="Nexus")

>>> s2=hs.load("file2.nxs",dset_search_keys=["Fe"])

>>> s2=hs.load("file2.nxs",dset_search_keys=["Fe"],meta_search_keys=["M1"])

>>> from hyperspy.io_plugins.nexus import get_datasets_in_file,get_metadata_in_file,file_reader
>>> datalist = get_datasets_in_file("file1.nxs")

>>> from hyperspy.io_plugins.nexus import get_datasets_in_file,get_metadata_in_file,file_reader
>>> dict=get_metadata_in_file("file1.nxs")

### CODE REVIEW

**Comment by dnjohnstone** [comment]
@pquinn-dls thanks for doing this - I'd had a chat with @chrisallen80 about using the nexus convention as a standard for file saving with pyxem, which I'm keen on and this would obviously save me doing some of that.

Before you go too much further I think it would be good if others who have worked more on the IO stuff e.g. @sem-geologist @ericpre  agree in principle that this should be a sort of special case of the .hdf5 reader - or would people prefer it to be a separate reader?

**Comment by pquinn-dls** [comment]
I should say this Nexus-lite
NXdata allows for  axes to be associated with a dataset and I look for NXdata sets and any hdf datasets not linked to a NXdata set. 
This is more or less what the Diamond GDA and DAWN software does as the implementations of nexus varied. 

Similarly when storing - I think storing NXdata is consistent with Hyperspy but I didn't want to
add any of the other NX objects.

**Comment by dnjohnstone** [comment]
For those unfamiliar with the nexus format - https://www.nexusformat.org/

**Comment by magnunor** [comment]
Semi-related: there was some work in LiberTEM in getting a STEM application definition into the nexus format.

But I'm not sure how far it progressed: https://github.com/LiberTEM/LiberTEM/issues/15.
@uellue or @sk1p might have some more information.

 Having such a definition will make it even more relevant including a nexus reader in HyperSpy.

**Comment by pquinn-dls** [comment]
Definitions really need a specific loader so that would require a further development to have nexus plugins for specific definitions or as with above pull request you can specify the preferred loader to use.

To explain the problem an example would be NXTomo for raw tomography data
The data is stored as  data[nFrames, xsize, ysize]
but depending on how the experiment was performed a frame might be  a background, flatfield or an actual projection so a second array image_key holds an index to indicate what type of frame each image is  projection = 0, flat field = 1, dark field = 2, invalid = 3
To create a useful hyperspy signal you would need to take the frames - split out projections, apply the dark and flat fields and return this corrected dataset or split the data into it's types but either way it can't be done in a generic way - it's application or definition specific

**Comment by francisco-dlp** [comment]
I think that somebody familiar with the NEXUS format would be a much better reviewer than myself for this PR—maybe @dnjohnstone ? Once a expert does a review I can have a look at the code if desired.

Some small comments:

* I think that from the user perspective it is not clear what ``preferred_format=Nexus`` does. What about simply ``format="Nexus"`` to force using a particular reader instead?
* Given a hdf5 file, wouldn't it be possible to detect what sort of reader is the most suitable?
* Regarding the tomo issue raised by @pquinn-dls, I think that for this community the second approach (returning all datasets) is preferred.  Several reader already return multiple signals inside a list, so that shouldn't be an issue.

**Comment by dnjohnstone** [comment]
> I think that somebody familiar with the NEXUS format would be a much better reviewer than myself for this PR—maybe @dnjohnstone ? Once a expert does a review I can have a look at the code if desired.

Well I'd like this to be in hyperspy so happy to help get it in. @pquinn-dls do you have some handy (i.e. not massive) example data that could be shared? I've only got limited stuff from Tiarnan that's a bit of a pain for testing...

**Comment by pquinn-dls** [comment]
@dnjohnstone 
There are a number of small example file from most facilities using nexus here (which also show how much people don't quite stick to using NXdata properly)
https://github.com/nexusformat/exampledata

As above I'm not sure if it needs more support for application definitions going forward (e.g. nexus_plugins in a folder to begin with to allow for specific loaders in future)

**Comment by pquinn-dls** [comment]
Tidied up the recursive search aspects of the code, all tests fixed (also needed USID changes as can also read hdf5 so preferred_format='USID' required)   Travis build error looks to be unrelated.

**Comment by dnjohnstone** [comment]
> Tidied up the recursive search aspects of the code, all tests fixed (also needed USID changes as can also read hdf5 so preferred_format='USID' required) Travis build error looks to be unrelated.

Yeah - I'm also having that problem with travis on mac for pyxem since yesterday afternoon - I have no idea where it's come from at this point... Anyway, not really important for this PR.

File: doc/user_guide/io.rst [hunk 1]
```diff
@@ -1030,6 +1032,162 @@ HDF5 file. All other keyword arguments will be passed to
...
+defines an agreed common layout that facilities can adhere to.
+ 
+Nexus metadata and data are stored in Hierarchical Data Format Files (HDF5) with
+a .nxs extension although standards HDF5 extensions are sometimes used.
+
+Files must use the ``.nxs``,``.hdf``,``.hdf5``,``.h5`` file extension 
+in order to use this io plugin.
+Using the ``.nxs`` extension will default to the Nexus loader 
+Using any other extension with Nexus reader will require the 
+format is specfied as Nexus in the loader
```
dnjohnstone: "Using any generic hdf5 extension with the Nexus reader will require the format to be specified as Nexus using the hdf5_format argument of the load function"

Related to this - please can we rename preferred_format --> hdf5_format to be more explicit?

Also a hs.load("my_data.hdf5", hdf_format="nexus")  code block example somewhere in this documentation would be helpful.

doc/user_guide/io.rst
```diff
@@ -1030,6 +1032,162 @@ HDF5 file. All other keyword arguments will be passed to
...
+    Out[3]:
+    (['/entry/xsp3_addetector', '/entry/xsp3_addetector_total'],
+     ['/entry/solstice_scan/keys/uniqueKeys', '/entry/solstice_scan/scan_shape'])    
+    
+
+Reading
+^^^^^^^
+Nexus files can contain multiple datasets within the same file.
+HyperSpy supports reading in one or more datasets however Nexus
+can contain a vary large number of datasets.
```
dnjohnstone: Left me asking so what? Is the point "...large number of datasets and so additional arguments allow the user to specify which data from the Nexus file is loaded."?

doc/user_guide/io.rst
```diff
@@ -1030,6 +1032,162 @@ HDF5 file. All other keyword arguments will be passed to
...
+
+.. code-block:: python
+
+    >>> hs.load("sample.nxs",nxdata_only=False)
+
+.. note::
+
+    Given that HDF5 files can accommodate very large datasets, setting 
+    ``lazy=True`` is strongly recommended if the contents of the HDF5 file 
+    are not known apriori. This prevents issues with regard to loading datasets
```
dnjohnstone: "_a priori_" in italics

doc/user_guide/io.rst
```diff
@@ -1030,6 +1032,162 @@ HDF5 file. All other keyword arguments will be passed to
...
+    If it is important that the files be closed after reading,
+    set ``lazy=False``.
+
+.. code-block:: python
+
+    >>> hs.load("sample.h5")
+    [<Signal2D, title: HAADF, dimensions: (|128, 128)>,
+    <Signal1D, title: EELS, dimensions: (|64, 64, 1024)>]
+
+We can load a specific datasets using the ``dset_search_path`` keyword argument.
```
dnjohnstone: "We can load specific..."

doc/user_guide/io.rst
```diff
@@ -1030,6 +1032,162 @@ HDF5 file. All other keyword arguments will be passed to
...
+.. code-block:: python
+
+    >>> # Loading a specific dataset
+    >>> hs.load("sample.h5", dset_search_path='EDS')
+
+Nexus files also support parameters or dimensions that have been varied 
+non-linearly.
+Since HyperSpy Signals expect linear variation of parameters / axes, such 
+non-linear information would be lost in the axes manager and replaced with
+indices
```
dnjohnstone: @ericpre @jlaehne @francisco-dlp - how can we sort #1927 ? I spoke to people trying to use hyperspy for Raman data, CL, PL data (and diffraction data) yesterday and this functionality is really needed for all these applications...

doc/user_guide/io.rst
```diff
@@ -1030,6 +1032,162 @@ HDF5 file. All other keyword arguments will be passed to
...
+Writing
+^^^^^^^
+Signals can be written to new Nexus files using the standard ``.save()`` 
+function.
+When Hyperspy creates a signal it creates and holds Hyperspy metadata 
+and original_metadata seperately.
+When we load a Nexus file the Nexus metadata is stored in 
+original_metadata. Conversely when we save to Nexus only the 
+original_metadata is currently stored to the Nexus file along with 
+the signal data and axes.
```
dnjohnstone: So s.save("output.nxs") loses any hyperspy metadata in this implementation? Is it a total nightmare to store that metadata in nexus? If it can be saved I think it would be preferable, but I appreciate that might not be viable right now.

doc/user_guide/io.rst
```diff
@@ -1030,6 +1032,162 @@ HDF5 file. All other keyword arguments will be passed to
...
+
+If you wish to save using the ``.hdf5``.``.h5`` or ``.hdf`` extensions
+you must specify the format as Nexus
+
+.. code-block:: python
+
+    >>> sig.save("output.h5",preferred_format="Nexus")
+
+Note that the model and other secondary data artifacts linked to the signal 
+are not written to the file but these can be implemented at a later stage.
```
dnjohnstone: I'd just remove the "but these can be implemented at a later stage" from the docs and, assuming this is merged without that, we can raise an issue to say that it should be implemented eventually.

hyperspy/io.py
```diff
@@ -52,6 +52,7 @@ def load(filenames=None,
          new_axis_name="stack_element",
          lazy=False,
          convert_units=False,
+         preferred_format=None,
```
dnjohnstone: as above I think "hdf_format" might be more explicit and therefore better? Is there any issue with that?

hyperspy/io.py
```diff
@@ -299,11 +303,18 @@ def load_single_file(filename, **kwds):
...
     i = 0
-    while extension.lower() not in io_plugins[i].file_extensions and \
-            i < len(io_plugins) - 1:
-        i += 1
+    
+    if preferred_format:
+        for index,plugin in enumerate(io_plugins):
+           if extension.lower() in plugin.file_extensions and \
+                preferred_format.lower() in plugin.format_name.lower():
+               i=index
```
dnjohnstone: ah - I suppose the issue with hdf5_format instead of preferred_format is that actually the user could specify any format and try to use that reader to load some file. Given that,  I think it might be best to go with just "format"?

hyperspy/io.py
dnjohnstone: Ok, having read more of the code I see that the user could specify any format and try to use that reader to load some file. Given that, I think it might be best to go with just "format"?

In any case - don't forget to add a description to the Parameters section of this docstring.

doc/user_guide/io.rst
dnjohnstone: Ok, having read more of the code I see that the user could specify any format and try to use that reader to load some file. Given that, I think it might be best to go with just "format"?

hyperspy/tests/io/test_nexus_hdf.py
```diff
@@ -0,0 +1,164 @@
+# -*- coding: utf-8 -*-
+# Copyright 2007-2016 The HyperSpy developers
```
dnjohnstone: 2007-2020

hyperspy/tests/io/test_nexus_hdf.py
```diff
@@ -0,0 +1,164 @@
...
+        s.original_metadata.set_item('test3',64.0)        
+        s.save(tmpfilepath)
+        l = load(tmpfilepath)
+        assert isinstance(l.original_metadata.attrs.test1, float)
+        assert isinstance(l.original_metadata.attrs.test1, float)
+        assert isinstance(l.original_metadata.attrs.test2, float)
+        assert l.original_metadata.attrs.test2 == 54.0
+
+#
+#    def test_general_metadata(self, tmpfilepath):
```
dnjohnstone: What's going on with this one?

dnjohnstone (commented):
Thanks @pquinn-dls - so far this seems to be doing what I think it should, but I've got to try some more files because of how generic this file type is... I think what you've written is also pretty generic though...

I think a pragmatic solution is to say that this is acceptable when we are sure it works for the nexus files obtained from I14, taking a "best effort" attitude to nexus files from other facilities until their users/staff scientists decide it's important to them.

It would be nice if we could save the hyperspy metadata (or at least some metadata in a dictionary or something) into the nexus file though - I think this would mean that @M0hsend and @ptim0626 could use that save function to convert .mib data recorded on the Merlin/Medipix at ePSIC to the nexus file format while adding in the appropriate metadata. Do you think that would be possible? I think it would be a nice immediate application of this!

Another general comment is that your docstrings need a once over - most of them don't describe parameters.

hyperspy/io.py
pquinn-dls: format is python function so better to use alternative name ?   io_format ?

dnjohnstone: Good point.... file_format?

hyperspy/tests/io/test_nexus_hdf.py
pquinn-dls: fixed

hyperspy/io.py
pquinn-dls: changed to file_format

hyperspy/io.py
pquinn-dls: changed to file_format - note this impact usid as well...

doc/user_guide/io.rst
pquinn-dls: Changed the documentation quite a bit. Took your comment on being pragmatic about what works for us so made changes to add saving hyperspy metadata and several signals based on workflows. Also clarified the differences in how metadata is structured to match the hdf and nexus formats - e.g. attributes with datasets

doc/user_guide/io.rst
pquinn-dls: Changed - problem is hyperspy returns a list of signals, I'd prefer a dict or dictionarybrowser .  We can have 20 or 30 signals or datasets and the order they're saved in might vary so if you want a specific signal you can't just use an index - you need to count which index it is or use a method to sort through the signal names.  Easier to only load the ones you want  - they are loaded in the order of the search keys

doc/user_guide/io.rst
```diff
@@ -1027,11 +1033,279 @@ HDF5 file. All other keyword arguments will be passed to
...
+    │   │   ├── original_metadata
+    │   │   │   ├── hyperspy_metadata
+    │   │   ├── learning_results
+ 
+
+To save multiple signals the file_writer method can be called directly.
+
+.. code-block:: python
+
+    >>> from hyperspy.io_plugins.nexus file_writer
```
ericpre: typo:
```python
from hyperspy.io_plugins.nexus import file_writer
```

doc/user_guide/io.rst
```diff
@@ -1027,11 +1033,279 @@ HDF5 file. All other keyword arguments will be passed to
...
+processing step when the data was collected.
+For example in experiment 1 Fe,Ca,P,Pb were collected in experiment 2
+Ca,P,K,Fe,Pb were collected. HyperSpy supports reading in one or more datasets and returns a list
+of signals but in this example case the indexing is different.
+To control which data or metadata is loaded and in what order
+some additional loading arguments are provided.
+
+Extra loading arguments
++++++++++++++++++++++++
+- ``dset_search_keys`` : ``str`` or ``list`` - Absolute path(s) or substring(s) used to find one or more datasets. (default is ``None`` all Nexus Datasets will be read)
```
ericpre: Would ``dataset_keys`` be more simple or more explicit? It is not clear to me what does the `substrings` refers to. How different it is from the path?

typo:
```python
``str`` or ``list`` of strings
```

doc/user_guide/io.rst
```diff
@@ -1027,11 +1033,279 @@ HDF5 file. All other keyword arguments will be passed to
...
+- ``meta_search_keys`` : ``str`` or ``list`` of strings - Absolute path(s) or substring(s) to use to find one or more datasets. (default is ``None`` all Nexus Metadata will be read)
```
ericpre: Similar as above: would ``metadata_keys`` be better?

hyperspy/misc/io/tools.py
```diff
@@ -76,6 +77,33 @@ def ensure_directory(path):
     if directory and not os.path.exists(directory):
         os.makedirs(directory)
 
+def find_filenames_matching_substrings(path,substrings,sortfiles=True):
```
ericpre: This is not used anywhere and not tested.

hyperspy/tests/io/test_nexus_hdf.py
```diff
@@ -0,0 +1,143 @@
...
+import numpy as np
+import pytest
+from hyperspy.io import load
+import hyperspy.api as hs
+from hyperspy.signal import BaseSignal
+
+
+dirpath = os.path.dirname(__file__)
+
+file1 = os.path.join(dirpath, 'nexus_files', 'file1.nxs')
```
ericpre: The test files are too large: it should be possible to reduce the size of the file to almost nothing by filling with zeros and saving it with compression.

doc/user_guide/io.rst
```diff
@@ -1027,11 +1033,279 @@ HDF5 file. All other keyword arguments will be passed to
...
+from a Nexus file. A mapping dictionary is the preferred method to map
+file metadata to hyperspy metdata.   
+
+Inspecting
+^^^^^^^^^^
+Looking in a Nexus file for specific metadatda is often useful - .e.g to find
+what position a specific stage was at: 
+
+    >>> from hyperspy.io_plugins.nexus import get_metadata_in_file
+    >>> results=get_metadata_in_file("sample.hdf5",search_keys=["stage1_z"])
```
ericpre: To be more precise: ``read_metadata_from_file`` instead of ``get_metadata_in_file``?
Typo: remove `results=` to get the output displayed?

ericpre (commented):
I have left a couple of comments after reading briefly.
Regarding the `hdf5` file format conflict, we should possibly find a more generic way that it is currently done in this PR but I guess this may be out of the scope of this PR. As @francisco-dlp already mentioned, if there is a way to recognise what a nexus file is then it would be useful. For example, we do this to make the difference between Berkeley and Velox emd files.

doc/user_guide/io.rst
pquinn-dls: dset was used in the usid plugin and docs so was trying to be consistent - but can change

pquinn-dls:
looking through the code for io.py "load_with_reader" exists and allows the reader as an argument but this only works for single files. 

Is it easier to - 
Change load_with_reader to load_single_file_with_reader()
Create a load_with_reader that allows a list of filenames 
Document load with reader in the user guide

This would be more generic I guess.

ericpre:
> looking through the code for io.py "load_with_reader" exists and allows the reader as an argument but this only works for single files.
> 
> Is it easier to -
> Change load_with_reader to load_single_file_with_reader()
> Create a load_with_reader that allows a list of filenames
> Document load with reader in the user guide
> 
> This would be more generic I guess.

Agree, we should be able to specify a `reader` argument to the `hs.load` and similar thing with a `writer` argument and the `save` function. Let's do this in a separate PR to keep it manageable.

doc/user_guide/io.rst
pquinn-dls: Changed to dataset_keys. Fixed typo

doc/user_guide/io.rst
pquinn-dls: Changed to dataset_keys and metadata_keys as suggested.

hyperspy/tests/io/test_nexus_hdf.py
pquinn-dls: Replaced with smaller files

hyperspy/misc/io/tools.py
pquinn-dls: removed

doc/user_guide/io.rst
pquinn-dls: Agreed and changed.

doc/user_guide/io.rst
pquinn-dls: Fixed - to save hyperspy_metadata and restore it from nexus

pquinn-dls:
@ericpre , @dnjohnstone   Think this is now complete. Tested against various files. Removed the "file_format" key so it only loads nxs files for now.  Shrunk the nxs test files. Can save and re-load hyperspy metadata to nxs.  Extensive tests. Re-did the documentation to reflect the changes.

jlaehne:
Just as information, non-uniform (non-linear) axis is progressing. The IO for `hspy` format is basically ready: #2304.

For the Nexus plugin, I would propose to merge when reviewed and then afterwards to start a new PR towards the non-uniform axis branch to include read/write for non-uniform axis. In the meantime, can you change the mention in the docstrings to "non-uniform axis" as we have decided for this term now.

hyperspy/io_plugins/nexus.py
```diff
@@ -0,0 +1,1251 @@
...
+    str
+        decoded version of input value
+
+    """
+    try:
+        text = value.decode(NX_ENCODING)
+    except UnicodeDecodeError:
+        if NX_ENCODING == 'utf-8':
+            text = value.decode('latin-1')
+        else:
```
ericpre: I don't follow the logic here and I suspect this could be simplify. What can be the string encoring of nexus file?

hyperspy/io_plugins/nexus.py
```diff
@@ -0,0 +1,1251 @@
...
+    ensures inter-compatibility across io_plugins)
+
+    Arrays of length 1 - return the single value stored     
+    
+    Large datasets are returned as dask arrays if lazy=True 
+    
+    Parameters
+    ----------
+    value : input read from hdf file (array,list,tuple,string,int,float)
+    lazy  : bool  {default: True}
```
ericpre: Even if this is simple function, maybe specify "only for h5py.Dataset with size >=2".

hyperspy/io_plugins/nexus.py
```diff
@@ -0,0 +1,1251 @@
...
+    try:
+        text = value.decode(NX_ENCODING)
+    except UnicodeDecodeError:
+        if NX_ENCODING == 'utf-8':
+            text = value.decode('latin-1')
+        else:
+            text = value.decode('utf-8')
+    return text.replace('\x00','').rstrip()
+
+def _parse_from_file(value,lazy=True):
```
ericpre: Usually, the default is to use `lazy=False` - usually, numpy raises an clear when there is not enough memory to load data and does it fairly quickly.

hyperspy/io_plugins/nexus.py
```diff
@@ -0,0 +1,1251 @@
...
+    * :py:meth:`~.io_plugins.nexus.read_metadata_from_file`
+                 
+   
+    """
+    # search for NXdata sets...
+
+    mapping  = kwds.get('mapping',{})   
+    original_metadata={}
+    learning = {}
+    fin = h5py.File(filename,"r+")
```
ericpre: Check that this is nexus file? For example, something along the line of: https://github.com/nexusformat/exampledata/blob/master/critique.py

hyperspy/io_plugins/nexus.py
```diff
@@ -0,0 +1,1251 @@
...
+        If true use lazy opening, if false read into memory
+    
+    Returns
+    -------
+    dask or numpy array
+        
+    """
+    data = group[dataset]
+    if data.dtype.type is np.string_ or data.dtype.type is np.object_:
+        return None
```
ericpre: Raise an error instead of return None?

hyperspy/io_plugins/nexus.py
```diff
@@ -0,0 +1,1251 @@
...
+    # strip out the small metadata if needed
+    if small_metadata_only:
+        stripped_metadata = _find_smalldata_in_dict(stripped_metadata)
+    if verbose:
+        pprint.pprint(stripped_metadata)
+    return stripped_metadata    
+
+
+def read_datasets_from_file(filename,search_keys=None,
+                          verbose=True):
```
ericpre, line 971: If may be convenient supporting a file handle too.

hyperspy/io_plugins/nexus.py
```diff
@@ -0,0 +1,1251 @@
...
+            for nxd in nexus_data_paths:
+                print(nxd)
+        else:
+            print("NXdata not found")
+        if hdf_dataset_paths:    
+            print("Unique HDF datasets found")
+            for hdfd in hdf_dataset_paths:
+                print(hdfd,fin[hdfd].shape)
+        else:
+            print("No HDF datasets not found or data is captured by NXdata")
```
ericpre: This could simplify by using the h5py [`visit`](https://h5py.readthedocs.io/en/latest/high/group.html#h5py.Group.visit) function with a nexus specific callable?

hyperspy/io_plugins/nexus.py
```diff
@@ -0,0 +1,1251 @@
...
+    #stripped_metadata = _find_search_keys_in_dict(metadata,search_keys=search_keys)
+    # strip out the small metadata if needed
+    if small_metadata_only:
+        stripped_metadata = _find_smalldata_in_dict(stripped_metadata)
+    if verbose:
+        pprint.pprint(stripped_metadata)
+    return stripped_metadata    
+
+
+def read_datasets_from_file(filename,search_keys=None,
```
ericpre: The name `read_datasets_from_file` is misleading - unlike `read_metadata_from_file` which actually read the metadata, this function doesn't read the datasets.

hyperspy/io_plugins/nexus.py
```diff
@@ -0,0 +1,1251 @@
...
+    ----------
+    data  -  numpy, dask array or h5py.Dataset
+    
+    Returns
+    -------
+    tuple 
+        Estimated chunk sizes for each axis
+    
+    """
+    chunks= h5py._hl.filters.guess_chunk(data.shape, None, np.dtype(data.dtype).itemsize)
```
ericpre: We should avoid using private API from h5py. It would be better to reuse the [`calculate_chunks` ](https://github.com/hyperspy/hyperspy/blob/42ad87fa6de3f947d75c969e4b50ccfab432ba25/hyperspy/io_plugins/emd.py#L64-L95) function and move it to a "hdf5_utils" or similar in `hypersy.misc.io`, which can be used by other reader. There is also a [`get_signal_chunks`](https://github.com/hyperspy/hyperspy/blob/42ad87fa6de3f947d75c969e4b50ccfab432ba25/hyperspy/io_plugins/hspy.py#L492-L535) in `hyperspy.io_plugins.hspy`.

hyperspy/io_plugins/nexus.py
```diff
@@ -0,0 +1,1251 @@
...
+        toreturn= _byte_to_string(toreturn)
+    if isinstance(toreturn, (np.int,np.float)):
+        toreturn= toreturn
+    if isinstance(toreturn, (np.ndarray)) and toreturn.dtype.char == "U": 
+        toreturn = toreturn.astype("S")
+    return toreturn
+
+
+def _parse_to_file(value):    
+    """ Check that value is compatible with hdf5 and if not convert to
```
ericpre: This could be moved to a "hdf5_utils" or similar in `hypersy.misc.io`.

doc/user_guide/io.rst
```diff
@@ -1036,6 +1039,303 @@ HDF5 file. All other keyword arguments will be passed to
...
+a .nxs extension although standards HDF5 extensions are sometimes used.
+Files must use the ``.nxs`` file extension in order to use this io plugin.
+Using the ``.nxs`` extension will default to the Nexus loader 
+
+As the Nexus format uses HDF5 and needs to read data and metadata structured
+in different ways the loader is written to quite flexible and can also be used 
+to inspect any hdf5 based file.  
+
+
+Differences with respect to hspy
```
ericpre, line 1099: I will put first the reading and writing sections since this one and the "inspecting" section is secondary in my opinion.

hyperspy/io_plugins/nexus.py
```diff
@@ -0,0 +1,1251 @@
...
+                    if "NX_class" in value["attrs"] and value["attrs"]["NX_class"]=="NXdata":
+                        continue
+                _write_nexus_attr(dictionary[key], group[key])
+                
+def _guess_signal_type(data):   
+    """Checks last 2 dimensions
+    An area detector will be square, > 255 pixels each side and
+    if rectangular ratio of sides is at most 4.
+    An EDS detector will generally by ndetectors*2048 or ndetectors*4096
+    the ratio test will be X10-1000
```
ericpre: The shape of the data is subjective to determine the type of data, is there a better way to find the signal type? I would expect the nexus format to have metadata/attributes from which we can infer what signal type it should be. If not, the default should be BaseSignal and optionally use this logic to guess the signal type automatically? It should be documented too.

hyperspy/misc/io/tools.py
```diff
@@ -19,6 +19,7 @@
 
 import os
 import logging
+import fnmatch
```
ericpre: Remove left over code from previous commits?

hyperspy/io_plugins/nexus.py
```diff
@@ -0,0 +1,1251 @@
...
+    Returns
+    -------
+    str
+        key prepended with 'fix_' if key is 
+        in a set of excluded of keywords
+    
+    """
+    exclusion_list=["keys"]
+    
+    if any([key.startswith(x) for x in exclusion_list]):
```
ericpre: since `exclusion_list` contains only one item, this should be simplified.

hyperspy/io_plugins/nexus.py
```diff
@@ -0,0 +1,1251 @@
...
+                tree["attrs"]={}
+            tree["attrs"][new_key] =  _parse_from_file(item,lazy=lazy)
+
+        for key, item in group.items():
+            if rootname != "":
+                rootkey = rootname +"/"+ key
+            else:
+                rootkey = "/" + key
+            new_key =_fix_exclusion_keys(key)
+            if type(item) is h5py._hl.dataset.Dataset:
```
ericpre: Better to use `isinstance(item, h5py.Dataset)` to avoid using private API.

hyperspy/io_plugins/nexus.py
```diff
@@ -0,0 +1,1251 @@
...
+                    tree[new_key]["value"] = _parse_from_file(item,lazy=lazy)
+                    for k,v in item.attrs.items():
+                        if "attrs" not in tree[new_key].keys():
+                            tree[new_key]["attrs"]={}
+                        tree[new_key]["attrs"][k] =  _parse_from_file(v,lazy=lazy)
+                else:
+                    # this is to support hyperspy where datasets are not saved
+                    # with attributes
+                    tree[new_key] = _parse_from_file(item,lazy=lazy)   
+            elif type(item) is h5py._hl.group.Group:
```
ericpre: Better to use `isinstance(item, h5py.Group)` to avoid using private API.

doc/user_guide/io.rst
```diff
@@ -1036,6 +1039,303 @@ HDF5 file. All other keyword arguments will be passed to
...
+and returns a list of signals but in this example case the indexing is different.
+To control which data or metadata is loaded and in what order
+some additional loading arguments are provided.
+
+Extra loading arguments
++++++++++++++++++++++++
+- ``dataset_keys`` : ``all``, ``hardlinks``, ``str`` or ``list`` of strings - Absolute path(s) or string(s) to search for in the path to find one or more datasets. (default is ``all`` all Nexus Datasets will be read)
+- ``metadata_keys`` : ``all``, ``str`` or ``list`` of strings - Absolute path(s) or string(s) to search for in the path to find metadata. (default is ``all`` all Nexus Metadata will be read)
+- ``nxdata_only`` : ``bool`` - Only convert NXdata formatted data to signals
+- ``small_metadata_only`` : ``bool`` - Only load items of size<2 into the metadata to avoid linking to large datasets
```
ericpre: Specify default value.

hyperspy/io_plugins/nexus.py
```diff
@@ -0,0 +1,1251 @@
...
+                else:
+                    tree[key]=value                    
+            elif isinstance(value,dict):
+                if key == "attrs":
+                    tree[key] = find_smallmeta_in_tree(value,sizefilter=False)
+                else:
+                    tree[key] = find_smallmeta_in_tree(value,sizefilter=True)                    
+        return tree
+    metadata_dict=find_smallmeta_in_tree(tree)
+    return metadata_dict
```
ericpre: Same as above: can it be simplified by using the h5py [`visit`](https://h5py.readthedocs.io/en/latest/high/group.html#h5py.Group.visit) function with a nexus specific callable?

hyperspy/io_plugins/nexus.py
```diff
@@ -0,0 +1,1251 @@
...
+                    p = p.setdefault(d,{})
+                p[mod_keys[-1]] = value
+            if isinstance(value,dict):
+                find_searchkeys_in_tree(value,rootkey)
+                    
+    if search_keys in ["all","hardlinks"]:
+        return tree
+    else:
+        find_searchkeys_in_tree(tree,rootname)
+    return metadata_dict
```
ericpre: Can it simplified by using the h5py [`visit`](https://h5py.readthedocs.io/en/latest/high/group.html#h5py.Group.visit) function with a nexus specific callable?

ericpre (changes requested):
A couple of comments - mostly to simplify or reuse existing code.
I have tried some of the dataset in https://github.com/nexusformat/exampledata/tree/master/DLS and it returns an error or an empty list.

hyperspy/io_plugins/nexus.py
pquinn-dls: The issue with visit is that it doesn't follow soft or external links within the file and many of our use cases for Nexus have linked files.
See  https://github.com/h5py/h5py/issues/671

ericpre: OK, it would be good to add the h5py github issue as reference!

hyperspy/io_plugins/nexus.py
pquinn-dls: Ok have removed.   Nexus does not actually store any information on how data should be plotted or presented - it stores the data and axes. I've added  attributes so that when a file is saved the navigation details are stored with the axis data.

doc/user_guide/io.rst
pquinn-dls: Fixed

hyperspy/io_plugins/nexus.py
pquinn-dls: Agreed and changed

hyperspy/io_plugins/nexus.py
pquinn-dls: Ok changed lazy=False as default throughout but added note in Nexus documentatio nto recommend using lazy.

hyperspy/io_plugins/nexus.py
pquinn-dls: Added line to docstring

hyperspy/io_plugins/nexus.py
pquinn-dls: Agreed and simplified

hyperspy/io_plugins/nexus.py
pquinn-dls: Thanks

The empty_list was due to the default setting - nxdata_only =True. This isn't generic enough so I've changed to nxdata_only = False default. 
I've found the error which is due to incorrect formatting in the file but I've put in a check if
its a str instead of a list.

I've tested against the following files from the nexus examples so think I've covered al variations.
DLS\i16\hdf5\538039.nxs
DLS\NXquadric\hdf5\sample_capillary.nxs
DLS\p45\hdf5\p45-316.nxs
DLS\p45\hdf5\p45-1168.nxs
DLS\p45\hdf5\p45-2194.nxs
DLS\reflections\hdf5\thaumatin_integrated.nxs
DLS\reflections\hdf5\thaumatin_integrated_multisample.nxs
Soleil\hdf5\file_1.nxsSoleil\hdf5\file_2.nxs
SwissFEL\hdf5\lyso009a_0087.JF07T32V01_master.h5
nxpdb\hdf5\4n8z.h5   - this hs string arrays only  returns empty
IPNS\LRMECS\hdf5\lrcs3701.nx5
hdf5\writer_1_3__niac2014.h5
hdf5\focus2007n001335.hdf
hdf5\NXtest.h5
APS\scan2nexus\hdf4\2iddf_0106.nexus
APS\scan2nexus\hdf4\mts_0347.nexus
APS\EPICSareaDetector\hdf5\AgBehenate_228.hdf5
APS\NXsas\hdf5\nexus-example.hdf5
DLS\i03_i04_NXmx\hdf5\Therm_6_2.nxs -   this previously failed in your test
DLS\p45\hdf5\p45-2194.nxs
DLS\p45\hdf5\p45-316.nxs

hyperspy/io_plugins/nexus.py
pquinn-dls: have changed to get_signal_chunks

hyperspy/io_plugins/nexus.py
pquinn-dls: It wasn't cleari if this would be of wider benefit without some general refactoring around hdf so I've left it as it is for now.

hyperspy/misc/io/tools.py
pquinn-dls: Fixed

doc/user_guide/io.rst
pquinn-dls: Ok have reordered to reading writing, inspection

hyperspy/io_plugins/nexus.py
pquinn-dls: Agreed - simplified to just look for key

hyperspy/io_plugins/nexus.py
pquinn-dls: Based on the comments around lazy etc. I've removed this particular filter to remove data >2 from metadata.
if a user wants to control the metadata they should use the metadata_keys to filter and it simplifies the code.

hyperspy/io_plugins/nexus.py
pquinn-dls: The none was to skip  over string arrays and only process numeric .I've moved the logic into the _find_data method to only list numeric datasets.

hyperspy/io_plugins/nexus.py
pquinn-dls: I've added to the docstring or did you want it in the biblio?

hyperspy/io_plugins/nexus.py
pquinn-dls: Changed to list_datasets_in_file

hyperspy/io_plugins/nexus.py
pquinn-dls: The readers works on any hdf or nexus file and I tend to have a mixture. Raw h5 which are then linked to within a nexus file so I omitted a strict check. As it's filtered by the nxs extension I didn't think it was needed.  From #2377 there is a need for more straightforward way to lazy load from a file.  If you can specify the reader in the load function you could generically load from hdf.

hyperspy/io_plugins/nexus.py
```diff
@@ -0,0 +1,1130 @@
...
+    This will save the signal in NXdata format in the file.
+    As the form of the metadata can vary and is not validated it will 
+    be stored as an NXcollection (an unvalidated collection)
+
+    Parameters
+    ----------
+    filename : str    
+        Path of the file to write
+    signals : signal or list of signals
+    store_original_metadata : bool , default : False
```
ericpre: There needs to be a description with a tabulation for sphinx to render the docstring correctly, something along the lines of will do:
```suggestion
    signals : signal or list of signals
        Signal(s) to be written
    small_metadata_only : bool , default : False
```

ericpre: Typo: `save` instead of `store`.

hyperspy/io_plugins/nexus.py
```diff
@@ -0,0 +1,1130 @@
...
+        for i,sig in enumerate(signals):
+            signal_name = sig.metadata.General.title if \
+            sig.metadata.General.title else 'unnamed__%d'% i
+            if "/" in signal_name:
+                signal_name = signal_name.replace("/", "_")
+            if signal_name.startswith("__"):
+                signal_name = signal_name[2:]
+            # store type of signal
+            if hasattr(sig,"_signal_type"):
+                sig.metadata.Signal.signal_type = sig._signal_type
```
ericpre: This is not necessary, because it should not be done here: setting `signal_type` is done when creating signal or by using the `set_signal_type` method. Anything else setting the `_signal_type` attribute is not supported (inappropriate used of private API) and will lead to bug or confusing behaviour.

hyperspy/io_plugins/nexus.py
```diff
@@ -1043,22 +922,17 @@ def read_metadata_from_file(filename,search_keys="all",
...
-    if small_metadata_only:
-        stripped_metadata = _find_smalldata_in_dict(stripped_metadata)
+    stripped_metadata = _load_metadata(fin,lazy=lazy,search_keys=search_keys)
     if verbose:
         pprint.pprint(stripped_metadata)
     return stripped_metadata    
 
 
-def read_datasets_from_file(filename,search_keys=None,
+def list_datasets_in_file(filename,dataset_keys=None,
```
ericpre: Add a test for this function.

hyperspy/io_plugins/nexus.py
ericpre: OK, let's leave as it is for now and when the io plugins move to a separate library, we should revisit this to share more code between plugins and be more consistent across io plugins.

hyperspy/io_plugins/nexus.py
ericpre: Yes, this is fine, thanks!

hyperspy/io_plugins/nexus.py
ericpre: Agree, we need to be able to specify a reader but also have a mechanism to pick a suitable reader for a given hdf5 file, in order to avoid relying on the extension. If you can add this check now, it would avoid us coming back to this later.
This will come at some point soon - at the latest when we split hyperspy.

> From #2377 there is a need for more straightforward way to lazy load from a file. If you can specify the reader in the load function you could generically load from hdf.

Not sure if I understand correctly: you are suggesting that specifying the nexus reader in `hs.load` will allow reading any hdf5 file (even if this is not a nexus file)?

ericpre (commented):
A few more comments (mostly typo or formatting). Can you merge with `RELEASE_next_minor` to get the coverage? Sorry there was an issue with the latest version of pytest-cov and this has been fixed now.

hyperspy/io_plugins/nexus.py
ericpre: Is there anything in the metadata, which can be used to determine if this should be a signal1d or signal2d?
Detector type, acquisition mode, etc.

ericpre:
From https://github.com/hyperspy/hyperspy/pull/2328#issuecomment-606709317
> To explain the problem an example would be NXTomo for raw tomography data
> The data is stored as data[nFrames, xsize, ysize]
> but depending on how the experiment was performed a frame might be a background, flatfield or an actual projection so a second array image_key holds an index to indicate what type of frame each image is projection = 0, flat field = 1, dark field = 2, invalid = 3
> To create a useful hyperspy signal you would need to take the frames - split out projections, apply the dark and flat fields and return this corrected dataset or split the data into it's types but either way it can't be done in a generic way - it's application or definition specific

Note for future improvement:
The case mentioned above could be done by post processing when loading the data. It is already possible to define functions which will be applied to the data at loading time. It doesn't support using multiple data set but it could be adapted:
https://github.com/hyperspy/hyperspy/blob/2e02941d35d7594a8ac8506b129d75db5171a5a1/hyperspy/io.py#L481-L483

Providing there is a NEXUS definition to determine what should be done, it would be a matter of having post processing recipes for specific definitions.

hyperspy/io_plugins/nexus.py
pquinn-dls: Ok will remove - I wanted to store the signal_type but the metadata signal_type set isn't set when creating a signal1d or signal2d  e.g. via hs.signals   or as_signal1D.  
signal_type only seems to be set for a signal outside the core signals - e.g. EDSSEMSpectrum 
so if I remove it will only store "" for signal 1D and 2D.

ericpre: Indeed, this is expect behaviour: `Signal1D` and `Signal2D` are generic and therefore their `signal_type` is not defined. When creating/changing signal class, this is a logic which looks at signal dimension to assign the correct class:
https://github.com/hyperspy/hyperspy/blob/2229ffde051af21d3adc7ae626a72bf769c84374/hyperspy/io.py#L404-L479

pquinn-dls:
Failing due to test_eelsdb_eels on macos

ericpre (approved):

PR #2330, opened 2020-03-03
Fix '"is" with a literal' warnings in Python 3.8

thomasaarholt:
### Description of the change
Python 3.8 is now warning when one is comparing an object to a literal with `is`. That is, `array.dtype is "float"` should really be `array.dtype == "float"`. This PR changes the code to reflect those warnings. 
`hyperspy.api` now imports without warning, and I've checked the warnings from running the tests for any further warnings of the same kind.

For what is meant by the term _literal_, see [this stackoverflow post](https://stackoverflow.com/questions/34189086/what-are-literals-in-python).

**Clarification with an example:**
The following will raise a warning:
```python
1 is None
<stdin>:1: SyntaxWarning: "is" with a literal. Did you mean "=="?
```
But the following will not:
```python
a = 1
a is None
```

This will also raise a warning:
```python
a = "mystring"
a is "abc"
<stdin>:1: SyntaxWarning: "is" with a literal. Did you mean "=="?
```

### Progress of the PR
- [x] ready for review.

### DIFF
--- a/hyperspy/tests/io/test_dm3.py
+++ b/hyperspy/tests/io/test_dm3.py
@@ -51,7 +51,7 @@ class TestImageObject():
         images = self._load_file(fname)
         image = images[0]
         # Should return None because the tags are missing
-        assert image._get_microscope_name(image.imdict.ImageTags) is None
+        assert image._get_microscope_name(image.imdict.ImageTags) == None
 
         fname = os.path.join(MY_PATH, "dm3_2D_data",
                              "test_diffraction_pattern.dm3")
@@ -67,8 +67,8 @@ class TestImageObject():
         assert self.imageobject._get_time("6:56:37 pm") == "18:56:37"
 
     def test_parse_string(self):
-        assert self.imageobject._parse_string("") is None
-        assert self.imageobject._parse_string("string") is "string"
+        assert self.imageobject._parse_string("") == None
+        assert self.imageobject._parse_string("string") == "string"
 
 
 def test_missing_tag():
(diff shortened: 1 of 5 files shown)


hyperspy/tests/io/test_dm3.py
```diff
@@ -67,8 +67,8 @@ def test_get_time(self):
         assert self.imageobject._get_time("6:56:37 pm") == "18:56:37"
 
     def test_parse_string(self):
-        assert self.imageobject._parse_string("") is None
-        assert self.imageobject._parse_string("string") is "string"
+        assert self.imageobject._parse_string("") == None
```
thomasaarholt: This one got caught up by accident, but I don't think it is any less clear like this.

ericpre: `_parse_string` is return a `None` type so the correct test is with `is`.

hyperspy/tests/io/test_dm3.py
```diff
@@ -51,7 +51,7 @@ def test_get_microscope_name(self):
         images = self._load_file(fname)
         image = images[0]
         # Should return None because the tags are missing
-        assert image._get_microscope_name(image.imdict.ImageTags) is None
+        assert image._get_microscope_name(image.imdict.ImageTags) == None
```
ericpre: Same as below.

ericpre (changes requested):

hyperspy/tests/io/test_dm3.py
thomasaarholt: Ah, yes. I searched to be sure - [PEP8 ](https://www.python.org/dev/peps/pep-0008/)is in agreement with you.

ericpre (approved):

Issue #1361, opened 2016-12-18
mini_batch_sparse_pca not working


thomasaarholt:
`s.decomposition(True, algorithm="mini_batch_sparse_pca", output_dimension=5)` doesn't work. Thought I'd try it to see what it would return for my EDX map, but it returned an empty array of loadings and factors.
I think we should really update the documentation to reflect the new changes to PCA. Perhaps with the addition of ORNMF we could do a proper update of at least the `algorithm` argument descriptor?

thomasaarholt:
~~I may have screwed up, working on the wrong branch. I'll reopen this if it still isn't working later.~~
It's real.

tjof2:
> but it returned an empty array of loadings and factors.

I think this is because you are using the default kwargs for [MiniBatchSparsePCA](https://scikit-learn.org/stable/modules/generated/sklearn.decomposition.MiniBatchSparsePCA.html). The same applies to the `'sparse_pca'` algorithm option too.

In particular, the `alpha` parameter. Try setting it very small to 1e-5 or so. For example, using the default values

```
>>> s.decomposition(algorithm="mini_batch_sparse_pca", output_dimension=5)
>>> loadings = s.get_decomposition_loadings().data
>>> factors = s.get_decomposition_factors().data
>>> print(loadings.mean())
>>> print(factors.mean())
>>> print(np.mean(factors == 0))

0.0
0.0
1.0 (i.e. all values are 0)
```

while

```
>>> s.decomposition(algorithm="mini_batch_sparse_pca", output_dimension=5, alpha=1e-5)
>>> loadings = s.get_decomposition_loadings().data
>>> factors = s.get_decomposition_factors().data
>>> print(loadings.mean())
>>> print(factors.mean())
>>> print(np.mean(factors == 0))

0.007826386225173738
0.18074411912790012
0.2896505376344086
```

> *alpha* : Sparsity controlling parameter. Higher values lead to sparser components.

A value of 1 is quite high, hence all the zeros

tjof2:
> *alpha* : Sparsity controlling parameter. Higher values lead to sparser components.
A value of 1 is quite high, hence all the zeros

tjof2:
Heuristically, we could set `alpha` to `1/sqrt(n_features)` if it  is not given explicitly, and pass through to sklearn.

tjof2:
Certainly from a documentation point-of-view, this is something to be addressed in #1159

tjof2:
[...]

Issue #1159, opened 2016-07-15
Improve decomposition documentation


thomasaarholt:
Currently the documentation for decomposition is very disorganised. The docstring currently looks like following. I think it could be useful to separate out a single general section, and a section/groups for each algorithm so that it's clearer what arguments one should be adjusting to get a correct result.

``` python
Signature: s1.decomposition(
    normalize_poissonian_noise=False, 
    algorithm='svd', 
    output_dimension=None, 
    centre=None, 
    auto_transpose=True, 
    navigation_mask=None, 
    signal_mask=None, 
    var_array=None, 
    var_func=None, 
    polyfit=None, 
    reproject=None, 
    return_info=False, 
    **kwargs)
Docstring:
Decomposition with a choice of algorithms

The results are stored in self.learning_results

Parameters
----------
normalize_poissonian_noise : bool
    If True, scale the SI to normalize Poissonian noise
algorithm : 'svd' | 'fast_svd' | 'mlpca' | 'fast_mlpca' | 'nmf' |
    'sparse_pca' | 'mini_batch_sparse_pca' | 'RPCA_GoDec' | 'ORPCA'
output_dimension : None or int
    number of components to keep/calculate
centre : None | 'variables' | 'trials'
    If None no centring is applied. If 'variable' the centring will be
    performed in the variable axis. If 'trials', the centring will be
    performed in the 'trials' axis. It only has effect when using the
    svd or fast_svd algorithms
auto_transpose : bool
    If True, automatically transposes the data to boost performance.
    Only has effect when using the svd of fast_svd algorithms.
navigation_mask : boolean numpy array
    The navigation locations marked as True are not used in the
    decompostion.
signal_mask : boolean numpy array
    The signal locations marked as True are not used in the
    decomposition.
var_array : numpy array
    Array of variance for the maximum likelihood PCA algorithm
var_func : function or numpy array
    If function, it will apply it to the dataset to obtain the
    var_array. Alternatively, it can a an array with the coefficients
    of a polynomial.
reproject : None | signal | navigation | both
    If not None, the results of the decomposition will be projected in
    the selected masked area.
return_info: bool, default False
    The result of the decomposition is stored internally. However, some algorithms generate some extra
    information that is not stored. If True (the default is False) return any extra information if available

Returns
-------
(X, E) : (numpy array, numpy array)
    If 'algorithm' == 'RPCA_GoDec' or 'ORPCA' and 'return_info' is True,
[...]

PR #2035, opened 2018-08-18
ENH: Add subspace tracking to ORNMF (consistent with ORPCA)

tjof2:
The ORPCA matrix factorization function supports a SGD method. Using this allows the subspace to be tracked if it changes over time, which is useful for e.g. foreground-background separation in videos.

This change copies the same subspace tracking code over to ONMF, which currently assumes that when `X = WH^T + E`, `W` is stationary for the whole dataset. If your `nsamples` dimension is time, this allows for tracking variations in `W`.

Also, the ONMF algorithm hasn't been accessible from `s.decomposition()` since it was introduced to HyperSpy - this was presumably an oversight.

PR #2336, opened 2020-03-16
Printing important decomposition information

tjof2:
### Description of the change
@thomasaarholt and I were discussing the idea of printing out information about `s.decomposition()`, after I discovered that the issue described in #1361 was likely caused by a hidden default argument passed to sklearn.

This PR introduces a `print_info=True/False` argument to `s.decomposition`, which prints out the values of the main arguments of the function such as `normalize_poissonian_noise`, and, if the chosen algorithm is an sklearn estimator, such as NMF, prints out `repr(estimator)` as well to highlight the (default) arguments sklearn is be using.

There are couple of bits open for review here

1. `polyfit=` is not documented in the docstring (see #1159)
2. Should the default value of `print_info` be True or False?
3. Should we log **ALL** the arguments? I've ignored the mask arguments, which can be ndarrays, for example.

### Progress of the PR
- [x] Change implemented
- [x] update docstring (if appropriate)
- [x] add tests
- [x] ready for review.

### Example
Default arguments/algorithm (no sklearn)
```python
>>> import numpy as np
>>> import hyperspy.api as hs
>>> s0 = hs.signals.Signal2D(np.abs(np.random.randn(16, 16, 512)))
>>> s0.decomposition(output_dimension=3)

Decomposition info:
  normalize_poissonian_noise=False
  algorithm=svd
  output_dimension=3
  centre=None
```
With an sklearn object
```python
>>> import numpy as np
>>> import hyperspy.api as hs
>>> s0 = hs.signals.Signal2D(np.abs(np.random.randn(16, 16, 512)))
>>> s0.decomposition(output_dimension=3, algorithm='nmf')

Decomposition info:
  normalize_poissonian_noise=False
  algorithm=nmf
  output_dimension=3
  centre=None
scikit-learn estimator:
NMF(alpha=0.0, beta_loss='frobenius', init=None, l1_ratio=0.0, max_iter=200,
    n_components=3, random_state=None, shuffle=False, solver='cd', tol=0.0001,
    verbose=0)
```

### DIFF
--- a/hyperspy/learn/mva.py
+++ b/hyperspy/learn/mva.py
@@ -109,6 +109,7 @@ class MVA():
                       polyfit=None,
                       reproject=None,
                       return_info=False,
+                      print_info=True,
                       **kwargs):
         """Decomposition with a choice of algorithms
 
@@ -116,18 +117,18 @@ class MVA():
 
         Parameters
         ----------
-        normalize_poissonian_noise : bool
+        normalize_poissonian_noise : bool, default False
             If True, scale the SI to normalize Poissonian noise
         algorithm : 'svd' | 'fast_svd' | 'mlpca' | 'fast_mlpca' | 'nmf' |
             'sparse_pca' | 'mini_batch_sparse_pca' | 'RPCA_GoDec' | 'ORPCA'
         output_dimension : None or int
             number of components to keep/calculate
         centre : None | 'variables' | 'trials'
-            If None no centring is applied. If 'variable' the centring will be
-            performed in the variable axis. If 'trials', the centring will be
-            performed in the 'trials' axis. It only has effect when using the
-            svd or fast_svd algorithms
-        auto_transpose : bool
+            If None (default), no centering is applied.
+            If 'variable', the centring will be performed in the variable axis.
+            If 'trials', the centring will be performed in the 'trials' axis.
+            It only has effect when using the svd or fast_svd algorithms
+        auto_transpose : bool, default True
             If True, automatically transposes the data to boost performance.
             Only has effect when using the svd of fast_svd algorithms.
         navigation_mask : boolean numpy array
@@ -142,6 +143,8 @@ class MVA():
             If function, it will apply it to the dataset to obtain the
             var_array. Alternatively, it can a an array with the coefficients
             of a polynomial.
+        polyfit : None | ???
+            TODO: currently undocumented
         reproject : None | signal | navigation | both
             If not None, the results of the decomposition will be projected in
             the selected masked area.
@@ -150,6 +153,10 @@ class MVA():
             some algorithms generate some extra information that is not
             stored. If True (the default is False) return any extra
             information if available
+        print_info : bool, default True
+            If True, print information about the decomposition being performed.
+            In the case of sklearn.decomposition objects, this includes the
+            values of all arguments of the chosen sklearn algorithm.
 
         Returns
         -------
@@ -166,6 +173,17 @@ class MVA():
 
         """
         to_return = None
+        to_print = [
+            "Decomposition info:",
+            "  normalize_poissonian_noise={}".format(normalize_poissonian_noise),
+            "  algorithm={}".format(algorithm),
+            "  output_dimension={}".format(output_dimension),
+            "  centre={}".format(centre),
+            "  auto_transpose={}".format(auto_transpose),
+            "  polyfit={}".format(polyfit),
+            "  reproject={}".format(reproject)
+        ]
+
         # Check if it is the wrong data type
         if self.data.dtype.char not in ['e', 'f', 'd']:  # If not float
             raise TypeError(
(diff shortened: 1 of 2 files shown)


thomasaarholt:
This is great! I'd like someone else to chime in on whether this should be on by default. An argument for is visibility of parameters, arguments against include clutter and possible confusion/"decomposition appears scary".

I think the combination of "our" arguments with the sklearn ones after looks very good.

Some parameters are still not covered, like the ones reported in the warning in this screenshot (method, etc.):
```python
import hyperspy.api as hs
import numpy as np
s = hs.signals.Signal1D(np.random.random((64, 64, 100)))
s.decomposition(algorithm='ORPCA', output_dimension=4)
```
image

tjof2:
@thomasaarholt ORPCA is an inconsistency, well spotted!

I think I'd like #2035 merged first before fixing it though - because it requires a little bit of work to e.g. print out the info in the same way as an sklearn estimators via __repr__.

> and possible confusion/"decomposition appears scary"

I'd be more worried that people were using decomposition without understanding why! But a valid point, clarity and brevity is super important to make this effective.

tjof2:
> and possible confusion/"decomposition appears scary"

I'd be more worried that people were using decomposition without understanding why! But a valid point, clarity and brevity is super important to make this effective.

tjof2:
@francisco-dlp  any view on whether this should be "on" or "off" by default?

francisco-dlp:
This is a really nice feature that will help many users discover the many parameters involved in ``decomposition``. Therefore +1 for having this on by default. However, I wonder if this should go to the ``logger`` instead of ``stdout`` as at present.  What do you think @ericpre?

tjof2:
I ~also think the logger is better, and~ agree with it being on by default as you write. I'll tidy this up tomorrow.

That said, this is really INFO logging level information, so 

1. there's then no need for a `print_info` argument, and 
2. if the default logging level is WARNING, the people who would benefit most from this (end-users) won't see this info.

thomasaarholt:
I agree with @tjof2. Since this is INFO, people won't see it. And if it is WARNING, people will worry and cite "errors", when this is nothing to worry about. My vote goes to `print_info=True` by default, since expert users / users running this in a loop will know to turn it off.

ericpre:
Putting the information to stdout or the logger can be important for downstream library or programme. Since the output to stdout can be switch off using `print_info=False`, this should not be an issue.
In this specific case, sending the information is `stdout` is appropriate to my opinion (since it can be disable) and I agree that the default should be `print_info=True`.

hyperspy/learn/mva.py
```diff
@@ -166,6 +173,17 @@ def decomposition(self,
...
         to_return = None
+        to_print = [
+            "Decomposition info:",
+            "  normalize_poissonian_noise={}".format(normalize_poissonian_noise),
+            "  algorithm={}".format(algorithm),
+            "  output_dimension={}".format(output_dimension),
+            "  centre={}".format(centre),
+            "  auto_transpose={}".format(auto_transpose),
+            "  polyfit={}".format(polyfit),
+            "  reproject={}".format(reproject)
```
ericpre: I think we could remove these to make it more succinct.

tjof2: I think auto-transpose is important. Agree on the others.

tjof2: That said, `auto_transpose` only affects SVD/fast SVD, so I'll remove it. I initially thought it affected the sklearn estimators but it doesn't.

tjof2:
Thanks for the comments @ericpre - I've removed some of the variables we print to focus on the important ones, and also added some tests to boost coverage.

Output now looks like this:
```
>>> import numpy as np
>>> import hyperspy.api as hs
>>> s0 = hs.signals.Signal2D(np.abs(np.random.randn(16, 16, 512)))
>>> s0.decomposition(output_dimension=3, algorithm='nmf')

Decomposition info:
  normalize_poissonian_noise=False
  algorithm=nmf
  output_dimension=3
  centre=None
scikit-learn estimator:
NMF(alpha=0.0, beta_loss='frobenius', init=None, l1_ratio=0.0, max_iter=200,
    n_components=3, random_state=None, shuffle=False, solver='cd', tol=0.0001,
    verbose=0)
```

I think this is complete now pending CI tests.

tjof2:
The failures in CI look to be unrelated - and are in `test_image.py`. Perhaps this is related to #2348?

```
hyperspy/tests/mva/test_decomposition.py ................................ [ 56%]
```

ericpre:
> The failures in CI look to be unrelated - and are in `test_image.py`. Perhaps this is related to #2348?
> 
> ```
> hyperspy/tests/mva/test_decomposition.py ................................ [ 56%]
> ```

Yes, the failures are unrelated to this PR.

ericpre:
Once the tests of you last commit are passed on appveyor, we can merged it. Appveyor is doing one build at a time, so it is a bit slow...

tjof2:
You're right @ericpre  it is quite slow! :D getting there though...

Success! :D

Issue #1626, opened 2017-05-17
Nbagg/notebook backend: Side-by-side plotting of navigator and signal?


thomasaarholt:
Now that nbagg/notebook backend is becoming usable, I'm finding that it is very awkward to embed a spectrum image in a notebook, as the navigator is on top, and the signal below. Most laptop screens are widescreen, not highscreen, although the latter is reasonably popular with desktops. It would be nice to have choice to have the two plot windows next to one another in the notebook.

Could we somehow support both horizontal and vertical layouts?

thomasaarholt:
(btw, I'm loving `s.plot(navigator="slider")` in nbagg)

thomasaarholt:
Sounds like this will be possible to implement with the introduction of  matplotlib 2.1.

thomasaarholt:
@francisco-dlp I think you mentioned the possibility of this with matplotlib 2.1. Could you remind us why you thought 2.1 might help?

PR #2338, opened 2020-03-17
Horizontal plotting with the ipympl backend

thomasaarholt:
### Description of the change
I've added support for plotting side-by-side horizontally using the `%matplotlib widget` backend. (`%matplotlib ipympl` is equivalent, but the former looks more user-friendly).

I've basically placed `fig.canvas` from the navigator and signal inside a ipywidgets `HBox` and `display`ed that.

I've also added two additional plotting styles:

-  'vertical' style, which differs from the regular plotting style by being centered in the browser. 
- 'hidden' style, which doesn't show (but does create) the plot, so that a user can grab the widget canvases themselves in order to create fun plot layouts.
```python
# displaying either of these will show the plot in a different cell
# and can be wrapped in other widgets
# hs.preferences.Plot.enable_ipympl_plotting = True
# hs.preferences.Plot.ipympl_plot_style = 'hidden'
s._plot.navigator_plot.figure.canvas
s._plot.signal_plot.figure.canvas
```

Note that it requires a recent version of ipympl, but not version 0.5.3! (0.5.6 was released today)

### Progress of the PR
- [x] basic support added
- [x] refinements
- [x] updated preferences
- [ ] update user guide
- [ ] add tests,
- [x] ready for review.

```python
%matplotlib widget
import hyperspy.api as hs
import numpy as np
s = hs.signals.Signal1D(np.random.random((64, 64, 256)))
hs.preferences.Plot.enable_widget_plotting = True
# hs.preferences.Plot.ipympl_plot_style = 'horizontal' | 'vertical' | 'hidden' # defaults to horizontal
s.plot()
```

Example on one of my spectrum images:
image

### DIFF
--- a/hyperspy/defaults_parser.py
+++ b/hyperspy/defaults_parser.py
@@ -142,11 +142,10 @@ class GUIs(t.HasTraits):
 class PlotConfig(t.HasTraits):
     enable_ipympl_plotting = t.CBool(
         False,
-        desc='Use the widget-based ipympl plotting instead of the classic plot. Allows for horizontal plotting. Requires ipywidget_gui.')
-    ipympl_plot_style = t.Str(
-        'horizontal',
-        label='ipympl plot style (horizontal", "vertical" or "dontshow")',
-        desc = 'Set the orientation of ipympl plotting or dont show at all for custom plotting')
+        label='Enable ipympl plotting: Use the widget-based ipympl plotting for horizontal plotting. Requires ipywidget_gui_hyperspy.')
+    ipympl_plot_style = t.Enum(
+        ['horizontal', 'vertical', 'hidden'],
+        label='ipympl plot style: (Use "hidden" for custom plot layouts)')
     saturated_pixels = t.CFloat(0.,
                                 label='Saturated pixels (deprecated)',
                                 desc='Warning: this is deprecated and will be removed in HyperSpy v2.0'
--- a/hyperspy/defaults_parser.py
+++ b/hyperspy/defaults_parser.py
@@ -141,7 +141,7 @@ class GUIs(t.HasTraits):
 
 class PlotConfig(t.HasTraits):
     enable_ipympl_plotting = t.CBool(
-        True,
+        False,
         desc='Use the widget-based ipympl plotting instead of the classic plot. Allows for horizontal plotting. Requires ipywidget_gui.')
     ipympl_plot_style = t.Str(
         'horizontal',
--- a/hyperspy/defaults_parser.py
+++ b/hyperspy/defaults_parser.py
@@ -140,12 +140,12 @@ class GUIs(t.HasTraits):
 
 
 class PlotConfig(t.HasTraits):
-    enable_ipympl_plotting = t.CBool(
+    enable_widget_plotting = t.CBool(
         False,
-        label='Enable ipympl plotting: Use the widget-based ipympl plotting for horizontal plotting. Requires ipywidget_gui_hyperspy.')
-    ipympl_plot_style = t.Enum(
-        ['horizontal', 'vertical', 'hidden'],
-        label='ipympl plot style: (Use "hidden" for custom plot layouts)')
+        label='Enable widget plotting: Use the ipympl-based widget plotting for horizontal plotting. Requires ipywidgets_gui_hyperspy.')
+    widget_plot_style = t.Enum(
+        ['horizontal', 'vertical'],
+        label='ipympl plot style:')
     saturated_pixels = t.CFloat(0.,
                                 label='Saturated pixels (deprecated)',
                                 desc='Warning: this is deprecated and will be removed in HyperSpy v2.0'
--- a/hyperspy/defaults_parser.py
+++ b/hyperspy/defaults_parser.py
@@ -140,6 +140,13 @@ class GUIs(t.HasTraits):
 
 
 class PlotConfig(t.HasTraits):
+    enable_ipympl_plotting = t.CBool(
+        True,
+        desc='Use the widget-based ipympl plotting instead of the classic plot. Allows for horizontal plotting. Requires ipywidget_gui.')
+    ipympl_plot_style = t.Str(
+        'horizontal',
+        label='ipympl plot style (horizontal", "vertical" or "dontshow")',
+        desc = 'Set the orientation of ipympl plotting or dont show at all for custom plotting')
     saturated_pixels = t.CFloat(0.,
                                 label='Saturated pixels (deprecated)',
                                 desc='Warning: this is deprecated and will be removed in HyperSpy v2.0'
--- a/hyperspy/defaults_parser.py
+++ b/hyperspy/defaults_parser.py
@@ -140,12 +140,9 @@ class GUIs(t.HasTraits):
 
 
 class PlotConfig(t.HasTraits):
-    enable_widget_plotting = t.CBool(
-        False,
-        label='Enable widget plotting: Use the ipympl-based widget plotting for horizontal plotting. Requires ipywidgets_gui_hyperspy.')
     widget_plot_style = t.Enum(
         ['horizontal', 'vertical'],
-        label='ipympl plot style:')
+        label='Widget plot style: (requires jupyter-matplotlib)')
     saturated_pixels = t.CFloat(0.,
                                 label='Saturated pixels (deprecated)',
                                 desc='Warning: this is deprecated and will be removed in HyperSpy v2.0'
--- a/hyperspy/drawing/mpl_he.py
+++ b/hyperspy/drawing/mpl_he.py
@@ -23,17 +23,8 @@ from traits.api import Undefined
 import matplotlib as mpl
 from ipywidgets.widgets import HBox
 from hyperspy.drawing import widgets, signal1d, image
-<<<<<<< HEAD
 from hyperspy.defaults_parser import preferences
-
-=======
-try:
-    # IPython is a dependency of ipympl
-    # display is only used during widget plotting with ipympl
-    from IPython.display import display
-except ImportError:
-    pass
->>>>>>> added basic support for horizontal plotting with the ipympl backend
+from IPython.display import display
 
 _logger = logging.getLogger(__name__)
 
--- a/hyperspy/drawing/mpl_he.py
+++ b/hyperspy/drawing/mpl_he.py
@@ -21,7 +21,6 @@ import logging
 
 from traits.api import Undefined
 import matplotlib as mpl
-from ipywidgets.widgets import HBox, VBox
 from hyperspy.drawing import widgets, signal1d, image
 from hyperspy.defaults_parser import preferences
 from IPython.display import display
@@ -50,7 +49,11 @@ class MPL_HyperExplorer(object):
         self.pointer = None
         self._pointer_nav_dim = None
 
+<<<<<<< HEAD
     def plot_signal(self, **kwargs):
+=======
+    def plot_signal(self, signal_widget=None):
+>>>>>>> 270d5a349 (changed to be able to take signal_widget and navigation_widget as arguments to s.plot)
         # This method should be implemented by the subclasses.
         # Doing nothing is good enough for signal_dimension==0 though.
         if self.axes_manager.signal_dimension == 0:
@@ -60,6 +63,7 @@ class MPL_HyperExplorer(object):
             for axis in self.axes_manager.signal_axes:
                 axis.offset = -axis.high_value / 2.
 
+<<<<<<< HEAD
     def plot_navigator(self, title=None, **kwargs):
         """
         Parameters
@@ -72,6 +76,22 @@ class MPL_HyperExplorer(object):
             :py:meth:`hyperspy.drawing.signal1d.Signal1DLine`.
 
         """
+=======
+    def plot_navigator(self,
+                       colorbar=True,
+                       scalebar=True,
+                       scalebar_color="white",
+                       axes_ticks=None,
+                       saturated_pixels=None,
+                       vmin=None,
+                       vmax=None,
+                       no_nans=False,
+                       centre_colormap="auto",
+                       title=None,
+                       min_aspect=0.1,
+                       navigator_widget=None,
+                       **kwds):
+>>>>>>> 270d5a349 (changed to be able to take signal_widget and navigation_widget as arguments to s.plot)
         if self.axes_manager.navigation_dimension == 0:
             return
         if self.navigator_data_function is None:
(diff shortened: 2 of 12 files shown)


thomasaarholt:
How should I handle the imports of `IPython.display.display` and `ipywidgets.widgets.HBox`? I see the latter failing because I have a "hard import" on ipywidgets. Should I wrap it in a try-except, like I did with `display`?

hyperspy/drawing/mpl_he.py
```diff
@@ -19,9 +19,15 @@
...
 from traits.api import Undefined
-
+import matplotlib as mpl
+from ipywidgets.widgets import HBox
 from hyperspy.drawing import widgets, signal1d, image
-
+try:
+    # IPython is a dependency of ipympl
+    # display is only used during widget plotting with ipympl
+    from IPython.display import display
```
ericpre: IPython is an implicit dependency of HyperSpy through ipyparallel and it is imported used elsewhere in HyperSpy, so it would be better to add ipython as dependency instead of having a try/expect block here.

hyperspy/drawing/mpl_he.py
```diff
@@ -175,6 +181,21 @@ def plot(self, **kwargs):
             self.plot_navigator(**kwargs.pop('navigator_kwds', {}))
         self.plot_signal(**kwargs)
 
+        if "ipympl" in mpl.get_backend():
```
ericpre: This should be used for the notebook backend too. It would need to be optional too.

hyperspy/drawing/utils.py
```diff
@@ -130,7 +130,15 @@ def create_figure(window_title=None,
     fig : plt.figure
 
     """
+    mpl_was_interactive = False
```
ericpre: This doesn't look a good approach to me. It should possible to include the matplotlib figure in a widget output and display this widget.

ericpre (commented):
It may be worth considering `jupyter-sidecar` too, even if at the moment, it needs an update to be compatible with jupyter lab 2.0.

hyperspy/drawing/mpl_he.py
thomasaarholt: This doesn't work for the notebook backend, afaik (by notebook backend, I mean `%matplotlib notebook`). The reason is that the with the notebook backend, the `figure.canvas` is not a widget. 

```python
%matplotlib notebook
# %matplotlib widget

from ipywidgets.widgets import Output

o = Output()
with o:
    plt.figure()
    plt.plot([1,2,3])
    
display(o)
```

hyperspy/drawing/utils.py
thomasaarholt: Yes, that makes much more sense.

francisco-dlp:
This is great!

ericpre:
@thomasaarholt have you looked at https://github.com/jupyter-widgets/jupyterlab-sidecar? It would be great to be able to use sidecar too!

```python
form hyperspy.api as hs
from sidecar import Sidecar
sc = Sidecar(title=s.metadata.General.title)
with sc:
    s.plot() 
```

thomasaarholt:
Yes! Last I looked at it, it didn't support Jupyterlab 2.0, but that is now supported.

How would we let the user choose? How about a preference setting `['Vertical', 'Horizontal', 'Jupyterlab']`, where it `try`s the user setting, and otherwise goes with 'Vertical'?

With that I'm thinking of what would happen if a user normally uses sidecar, but then opens a notebook in jupyter notebook instead of lab, where sidecar is not supported.

ericpre:
> How would we let the user choose? How about a preference setting `['Vertical', 'Horizontal', 'Jupyterlab']`, where it `try`s the user setting, and otherwise goes with 'Vertical'?

A preference setting along with an argument should of the `plot` sounds good to me.
 
> With that I'm thinking of what would happen if a user normally uses sidecar, but then opens a notebook in jupyter notebook instead of lab, where sidecar is not supported.

The sidecar will work only if the `widget` backend is used (which will be a good indication the jupyterlab is being used or not) and `sidecar` is installed.

francisco-dlp:
I have recently discovered the ``Create new view for output`` option of jupyter lab that you get in the context menu when right-clicking on any output. It seems to be similar to sidecar but more flexible since you can open as many as you like and you can customize the layout.

ericpre:
It seems that "creating new view for output" is only for the frontend and this is no python API for to this: https://github.com/jupyterlab/jupyterlab/issues/3692
https://github.com/jupyterlab/jupyterlab/issues/4021

thomasaarholt:
I've been chatting with the dev of ipympl, who is also contributing to sidecar.

Currently an issue with ipympl is that the plot size is static and doesn't scale with the sidecar size, but a "resizeable" flag should hopefully be coming soon.
https://github.com/matplotlib/ipympl/issues/207

Also, there's discussions on making sidecar support new tabs rather than some sort of "custom" output windows.
https://github.com/jupyter-widgets/jupyterlab-sidecar/issues/49

thomasaarholt:
I've added a bunch of updates to this, please reread top post and try it. Currently it defaults to False, but it should be safe to turn on by default with 1.6 as it will only impact users with ipympl installed, in the `notebook` and `lab`.

hyperspy/defaults_parser.py
```diff
@@ -139,6 +139,12 @@ class GUIs(t.HasTraits):
 
 
 class PlotConfig(t.HasTraits):
+    enable_ipympl_plotting = t.CBool(
```
ericpre: Why is it necessary? This is actually misleading, since selecting this option implies that the widget matplotlib backend will be used, which it doesn't do: the backend is set by matplotlib.

hyperspy/drawing/utils.py
```diff
@@ -27,6 +27,7 @@
 import warnings
 import numpy as np
 import logging
+import ipywidgets
```
ericpre: `ipywidgets` is not a hyperspy dependency and should not become one.

hyperspy/drawing/utils.py
```diff
@@ -131,7 +132,15 @@ def create_figure(window_title=None,
...
-    fig = plt.figure(**kwargs)
+    if "ipympl" in mpl.get_backend() and preferences.Plot.enable_ipympl_plotting:
+        # this output is currently not used, just used for hiding.
+        # instead we get at the figure canvas widgets themselves later
+        output_widget = ipywidgets.widgets.Output()
+        with output_widget:
+            fig = plt.figure(**kwargs)
+    else:
+        output_widget = None
+        fig = plt.figure(**kwargs)
```
ericpre: I don't see how this can be useful, can you elaborate?
If you want to put the figure in a specific widget, it would be more sensible to add an argument `output_widget` to this function and reuse this here. This may be a way to fix the issue with the histogram tool not being displayed when using jupyter lab.

hyperspy/drawing/mpl_he.py
```diff
@@ -179,6 +180,29 @@ def plot(self, **kwargs):
             self.plot_navigator(**kwargs.pop('navigator_kwds', {}))
         self.plot_signal(**kwargs)
 
+        if "ipympl" in mpl.get_backend() and preferences.Plot.enable_ipympl_plotting:
+            # Then we can use the widget backend for two figures horizontally
+            plot_style = preferences.Plot.ipympl_plot_style
+            if plot_style != "hidden":
```
ericpre: How is it expected to be used.

hyperspy/drawing/mpl_he.py
```diff
@@ -20,10 +20,11 @@
 import logging
 
 from traits.api import Undefined
-
+import matplotlib as mpl
+from ipywidgets.widgets import HBox, VBox
```
ericpre: Move this import to where it is used and put it in a try, except block to handle the case, where ipywidgets is not installed.

ericpre (changes requested):

hyperspy/defaults_parser.py
thomasaarholt: Sorry, I got mixed up with ipywidgets and ipympl there.
The purpose is to enable plotting using the ipywidgets (output) widget interface.
How about `enable_ipywidgets_plotting`? I think `enable_output_plotting` would be a bit strange.

hyperspy/drawing/utils.py
thomasaarholt: So, the principal reason for using `with output_widget:` here is to prevent the figure from being displayed immediately. But the Output widget itself is a bit useless because we already have a good handle on the "figure widget" through the following two objects, which I get at in `mpl_he.py`:

`s._plot.navigator_plot.figure.canvas`
`s._plot.signal_plot.figure.canvas`

I'm looking at how difficult it would be to add the output widget as an argument. I agree it would be neater to not just throw it away, but I'm not sure how useful it is - unless, as you suggest it may be used for the histogram tool. (but I'm not really familiar with that problem)

hyperspy/drawing/mpl_he.py
thomasaarholt: I assume you're referring to `plot_style == "hidden"`? 
I added it as an afterthought - it means that anyone making some sort of "custom plotting tool" where they maybe link several figures, could for instance link several signals through the interactive interfaces (ROI / Events), and then have the figures appear in a nice layout of their creation.

Of course, since it's a global setting, plotting some other signal would also be hidden, so this definitely isn't the best way to do this. I'm just experimenting. Perhaps something like `s.plot(navigator_output_widget=output_widget1, signal_output_widget = output_widget2)` is a better alternative?

hyperspy/defaults_parser.py
ericpre: Why should it be optional to use the output widget?

hyperspy/drawing/mpl_he.py
ericpre: See the `navigator_kwds` to arguments to the navigator plot: https://hyperspy.readthedocs.io/en/latest/user_guide/visualisation.html#customising-image-plot

thomasaarholt: I've got a much better implementation coming now.

hyperspy/drawing/utils.py
ericpre: > I agree it would be neater to not just throw it away, but I'm not sure how useful it is

This would allow to put the figure in an specified output widget -> enable "custom plot layout" you mentioned above using hyperspy public API.

hyperspy/drawing/mpl_he.py
ericpre: The following works fine:
```python
%matplotlib notebook

from ipywidgets.widgets import Output

o = Output()
display(o)
with o:
    plt.figure()
    plt.plot([1,2,3])
```

Can it be adapted to get it to work with the notebook backend?

ericpre:
xref #1626.

hyperspy/drawing/mpl_he.py
thomasaarholt: Maybe. Just to note, the following does not work:
```python
%matplotlib notebook

from ipywidgets.widgets import Output, HBox

o = Output()
o2 = Output()
with o:
    plt.figure()
    plt.plot([1,2,3])
with o2:
    plt.figure()
    plt.plot([3,2,1])

display(HBox([o, o2]))
```

thomasaarholt: Weird. It sorta does work, but only if I place the display statement before the `with` statements.

Additionally, the layout css does not have any effect:
```python
%matplotlib notebook

from ipywidgets.widgets import Output, HBox

o = Output()
o2 = Output()
margin = "auto 0px auto 0px"
o.layout.margin = margin
o2.layout.margin = margin

display(HBox([o, o2]))
with o:
    plt.figure()
    plt.plot([1,2,3])
with o2:
    plt.figure()
    plt.plot([3,2,1])
```

thomasaarholt:
I've made some changes now to specify widgets to s.plot().
I'll see if I can put those into `navigator_kwds `, and also turn the various `def plot...(self, widget)` widget arguments into a class attribute. It was a bit difficult to navigate the plotting code, so there may be a few superfluous inclusions of `widget=None` to some functions as well.

```python
%matplotlib widget
import numpy as np
import hyperspy.api as hs
from ipywidgets import Output, HBox

hs.preferences.Plot.enable_widget_plotting = True
hs.preferences.Plot.widget_plot_style = 'horizontal'

o1 = Output()
o2 = Output()

s1.plot(navigator_widget=o2, signal_widget=o1)

display(HBox([o1, o2]))
```

thomasaarholt:
@ericpre or @francisco-dlp I was wondering if you could help me with a specific weird behaviour here.
In[`mpl_he.py`](https://github.com/thomasaarholt/hyperspy/blob/widget_plot/hyperspy/drawing/mpl_he.py#L201) I set `navigator_kwds['widget'] = navigator_widget`, where `navigator_widget` is an Output widget.

On the first .plot() call, everything works fine, but on the second, the navigator plot appears in the output cell of the previous plot call.

If you look at the docstring in the notebook, you can see that after calling `s.plot()`, the default parameters of `navigator_kwds` now includes the previous Output widget. This explains the why a second navigator figure appears in the old output widget, but is itself strange behaviour.


I do not understand the mechanism, but I've been able to reproduce it by [creating a `SEEHERE` dict](https://github.com/thomasaarholt/hyperspy/blob/widget_plot/hyperspy/_signals/signal1d.py#L1582) that is an argument to `s.plot()`, and then [modifying the dict in the same area](https://github.com/thomasaarholt/hyperspy/blob/widget_plot/hyperspy/drawing/mpl_he.py#L204) that we create the output widgets in the other case.

I've highlighted the Output behaviour in red and the SEEHERE example in blue in the screenshots below.

Default values before calling s.plot()
image
Default values after calling s.plot()
image

To test the behaviour:
```python
%matplotlib widget
import hyperspy.api as hs
import numpy as np
s = hs.signals.Signal1D(np.random.random((4,4)))

s.plot() # inspect this before calling it - restart kernel if you have already

# Call this in a separate cell
s.plot() # Adds another navigator figure to the previous cell

# Call this, and it works
s.plot(navigator_kwds = {}) # this creates expected behaviour.
```

ericpre:
I don't know what is the question but I suspect the default behaviour of jupyter widget is not to be displayed as output of the current cell. That's why, I don't find it surprising that the previous output widget is used. They may be workaround this, but first I would figure out what should be the correct expected behaviour.

CSSFrancis:
@thomasaarholt where did this get left at? This is something that would be really helpful for visualizing 4-D STEM data and presenting.  I can pick it up if you would like me to.  It seems like this PR is most of the way there.

thomasaarholt:
By all means pick it up. It would probably be easier to make a new branch and copy over the relevant bits of code.

CSSFrancis:
superseded by #3140

PR #2341, opened 2020-03-24
Default values for ROIs

thomasaarholt:
### TLDR:
This lets one call `hs.roi.Line2DROI()` without specifying coordinates, defaulting to sensible values when called `interactive`ly on a signal.

### Description of the change
It is cumbersome to have to manually specify the coordinates of a ROI - especially when you just quickly want a ROI on an image to extract a certain part of it with `roi.interactive()`.

This PR is an attempt to allow `None` (or really, the `Undefined` `traits` type) to be an allowed value for ROIs so that they can be properly instantiated at a later time.

I have:
- chosen sensible values as default (the central half of an axis - 1/4 to 3/4). For this I introduced a new helper function `get_central_half_limits_of_axis(ax)` in `hyperspy.utils.axes` (new file) that returns indices of the central half of an axes using #2386.
- introduced two new methods:
    1. `roi.parameters`: property that returns the parameter names and values as a dict
    2. `roi._set_default_values()`: sets the appropriate values when using it interactively with un-set values.
- reduced code repetition by moving methods to parent classes where logical (`is_valid`, `__getitem__` and `__repr__`)
- updated the `__repr__` of ROIs to use f-strings and the `:G` formatter where possible.

### Progress of the PR
- [x] Working implementation for all ROIs
- [x] Need to change the \_\_repr\_\_ of rois to take into account the Undefined trait
- [x] add tests,
- [x] ready for review.

```python
%matplotlib widget
import hyperspy.api as hs

s2 = hs.datasets.artificial_data.get_atomic_resolution_tem_signal2d().isig[:31, :31]
s1 = s2.isig[0,:]
# 1D
pr = hs.roi.Point1DROI()
sr = hs.roi.SpanROI()

roi1 = [pr, sr]

# 2D
cr = hs.roi.CircleROI()
rr = hs.roi.RectangularROI()
lr = hs.roi.Line2DROI()
p2r = hs.roi.Point2DROI()

roi2 = [cr, rr, lr, p2r]

colors = ['blue', 'green', 'red', 'black']
s1.plot()
for i, roi in enumerate(roi1):
    roi.interactive(s1, color = colors[i])


s2.plot()
for i, roi in enumerate(roi2):
    roi.interactive(s2, color = colors[i])
    

```


ericpre:
@thomasaarholt, what is the status of this PR? The relative slicing that you added in #2386 could be used here to set the default value?

thomasaarholt:
Yes! The lack of any comments here made me think that this PR perhaps wasn't desirable, but then again the post is a block of text where I never asked for feedback.

I'm currently deeply involved with a different programming project, but this change should be small enough so I'll try to get it done soon.

ericpre:
> The lack of any comments here made me think that this PR perhaps wasn't desirable, but then again the post is a block of text where I never asked for feedback.

Indeed... actually, I commented on this PR in https://github.com/hyperspy/hyperspy/pull/2343#issuecomment-619249511.

thomasaarholt:
There is one "breaking" change in this: 
`rect = RectangularROI(...)` takes arguments as `left, top, right, bottom`, but when indexed afterwards as `rect[i]`, it would return it in a different order -> `left, right, top, bottom` for `i = (0,1,2,3)`. The n
As an example:
```python
>>> rect = hs.roi.RectangularROI(2, 2, 5, 5)
>>> rect
RectangularROI(left=2, top=2, right=5, bottom=5)
>>> rect[2]
2.0
```
Now it is indexed in the same order.
I'd like to interpret this part of this PR as a bugfix rather than a breaking change. I'm happy to submit it as a different PR if needed.

ericpre:
> I'd like to interpret this part of this PR as a bugfix rather than a breaking change. I'm happy to submit it as a different PR if needed.

Agreed and I think we need to make sure that we get it right... because it can be quite disruptive for users... Definitely better figuring this out in a separate PR.

thomasaarholt:
I've manually checked that `roi.gui()` works when the values are undefined.

hyperspy/roi.py
```diff
@@ -83,9 +83,35 @@ def __init__(self):
...
 
+    def __repr__(self):
+        attr = []
+        for name, value in self._get_attributes().items():
+            if value is t.Undefined:
+                attr.append(f"{name}={value}")
+            else:
+                # otherwise format value with the General specifer
+                attr.append(f"{name}={value:G}")
+        return f"{self.__class__.__name__}" + "(" + ", ".join(attr) + ")"
```
ericpre: To make it more readable:
```suggestion
        return f"{self.__class__.__name__}({', '.join(attr)})"
```

hyperspy/roi.py
```diff
@@ -83,9 +83,35 @@ def __init__(self):
...
+        if t.Undefined in self._get_attributes().values():
+            return True
+        else:
+            return False
+
+    def _raise_error_if_undefined_attrs(self):
+        "Raise an error i.e. when about to __call__ on a signal, with Undefined attrs."
+        if self._attributes_have_undefined():
+            raise AttributeError(
+        "Some ROI attributes have not yet been set. Set them before slicing a signal.")
```
ericpre: These two functions can be merged, there is no reason to separate them.

hyperspy/roi.py
```diff
@@ -333,6 +360,16 @@ def _apply_roi2widget(self, widget):
         """
         raise NotImplementedError()
 
+    def _set_default_attributes(self, signal):
+        """When the ROI is called interactively with Undefined attributes,
+        use these values instead.
+        """
+        raise NotImplementedError()
+
+    def _get_attributes(self):
```
ericpre: This method is defined for the second time.

hyperspy/roi.py
```diff
@@ -580,14 +625,23 @@ class Point1DROI(BasePointROI):
...
+        _tuple = tuple(self._get_attributes().values())
         return _tuple.__getitem__(*args, **kwargs)
 
+    def _set_default_attributes(self, signal):
+        ax0, *_ = self._parse_axes(None, signal.axes_manager)
+        # If roi attributes are undefined, use center of axes
+        if t.Undefined in self._get_attributes().values():
+            self.value = ax0._parse_value('rel0.5')
+
+    def _get_attributes(self):
```
ericpre: Would it be better to call this method `parameters` and return something similar as the `values`/`items` method of a dictionary would do? I am suggesting this, because this method is mainly used with `_get_attributes().values()`.
Also, I find that the name not very descriptive and actually misleading: as a private method, we will expect something following python jargon and therefore return all or any attributes.

hyperspy/roi.py
```diff
@@ -922,20 +991,64 @@ class CircleROI(BaseInteractiveROI):
...
+    def parameters(self):
+        return {
+            "cx":self.cx,
+            "cy":self.cy,
+            "r": self.r
+            }
+
+    @property
+    def parameters_optional(self):
+        return {"r_inner":self.r_inner}
```
ericpre: Are these necessary?

hyperspy/roi.py
```diff
@@ -922,20 +991,64 @@ class CircleROI(BaseInteractiveROI):
...
+            "cy":self.cy, 
+            "r": self.r,
+            "r_inner": self.r_inner
+            }
+
+    def _attributes_have_undefined(self):
+        # Redefine this method for CircleROI, since r_inner
+        # can is ok to set as undefined, but would normally 
+        # be caught in self._get_attributes(). It is the last entry.
+        if t.Undefined in tuple(self._get_attributes().values())[:-1]:
```
ericpre: Maybe better to use the line2D approach with linewidth and set the `r_inner` to 0 by default.

hyperspy/roi.py
```diff
@@ -580,14 +625,23 @@ class Point1DROI(BasePointROI):
...
         super(Point1DROI, self).__init__()
+        value = value if value is not None else t.Undefined
         self.value = value
 
     def __getitem__(self, *args, **kwargs):
-        _tuple = (self.value,)
+        _tuple = tuple(self._get_attributes().values())
         return _tuple.__getitem__(*args, **kwargs)
 
+    def _set_default_attributes(self, signal):
```
ericpre: `_set_defaults_values` would fit better, I find `attributes` too generic.

ericpre (changes requested):
See comments to make it more simple.

hyperspy/roi.py
thomasaarholt: No, they were part of a draft approach, which I forgot to remove.

thomasaarholt:
Thanks for the comments. I ended up keeping `_attributes_have_undefined` and `_raise_error_if_undefined_attrs`, since I ended up using the former a lot today (they are now called `_parameters_have_undefined` and `_raise_error_if_undefined_para`).

We now use `parameters` and `values`. The former is a dict, the latter is a tuple (and both are @property). I toyed with creating a `parameters` class, in similar vein to what we have with models, but decided against it. 

I also refactored some code (like `self.is_valid()`) into `BaseROI`. I think it fits correctly there. Will probably add some more tests when I see how coverage is looking.

hyperspy/roi.py
```diff
@@ -374,6 +416,9 @@ def interactive(self, signal, navigation_signal="same", out=None,
                     'fft_shift', False):
                 raise NotImplementedError('ROIs are not supported when data '
                                           'are shifted during plotting.')
+        # Undefined if roi initialised without specifying parameters
+        if self._parameters_have_undefined():
```
ericpre: This can equivalent to `t.Undefined in list(self)`, which has the advantage:
- it is explicit and clear what it is.
- use standard python syntax
- don't add another method to these class

One issue with the current approach is that its behaviour is not consistent and misleading: `CircleROI` and `Line2D` don't check all values while other does.

hyperspy/roi.py
```diff
@@ -83,17 +83,51 @@ def __init__(self):
...
+
     _ndim = 0
     ndim = property(lambda s: s._ndim)
 
+    @property
+    def parameters(self):
+        raise NotImplementedError()
+
+    @property
+    def values(self):
```
ericpre: Since the same can be done with `tuple(roi)` or `list(roi)`, I don't think this property is necessary. Same reason as in my other comment.

ericpre (changes requested):
I have provided more details on why we shouldn't add more method which doesn't add any functionality. The code for the ROIs is already quite long and we shall not make unnecessarily longer.

ericpre:
> Thanks for the comments. I ended up keeping `_attributes_have_undefined` and `_raise_error_if_undefined_attrs`, since I ended up using the former a lot today (they are now called `_parameters_have_undefined` and `_raise_error_if_undefined_para`).

As mentioned in the review, I am not convinced that adding these method is a good approach.

hyperspy/roi.py
thomasaarholt: I ended up removing them leaving
```python
        if t.Undefined in list(self):
            raise AttributeError(
        "Some ROI parameters have not yet been set. "
        "Set them before slicing a signal."
        )
```
in the three `__call__()` instances where `_raise_error` was used, and using `t.Undefined in list(self)` where `_attributes_have...` was used.

hyperspy/roi.py
thomasaarholt: Yep.

hyperspy/roi.py
thomasaarholt: Agreed

thomasaarholt:
@ericpre I think everything is now taken into account. Those were good suggestions, it's a lot cleaner now. It's been a while since I've thought about taking `list(self)` and what that means.

The failing tests are due to numpy 1.20, see #2635.

The change to setting CircleROI's `r_inner=0` means that unless the user sets it manually, it cannot be `t.Undefined`. Could you look over my last commit with a critical eye and see if you agree with the changes?

hyperspy/roi.py
```diff
@@ -580,17 +610,20 @@ class Point1DROI(BasePointROI):
...
         self.value = value
 
-    def __getitem__(self, *args, **kwargs):
-        _tuple = (self.value,)
-        return _tuple.__getitem__(*args, **kwargs)
-
+    def _set_default_values(self, signal):
+        ax0, *_ = self._parse_axes(None, signal.axes_manager)
+        # If roi parameters are undefined, use center of axes
+        if t.Undefined in list(self):
```
ericpre: This set the defaults value _only when_ some on the parameters are undefined. Would it be more simple to always set defaults value regardless of the current state? As I understand, this is supposed to be used only that way?.

Also, it may be good to add a single line docstring, outlining how this method is supposed to be used. A lot of the method doesn't have it, which obviously is not good at all, and we need to improve the situation...

hyperspy/roi.py
```diff
@@ -777,19 +808,37 @@ class RectangularROI(BaseInteractiveROI):
...
+        left, top, right, bottom  = (
+            para if para is not None 
+            else t.Undefined for para in (left, top, right, bottom))
         self._bounds_check = True   # Use reponsibly!
-        self.top, self.bottom, self.left, self.right = top, bottom, left, right
+        self.left, self.top, self.right, self.bottom = left, top, right, bottom
 
     def __getitem__(self, *args, **kwargs):
+        # Note: RectangularROI is currently indexed in a different way
+        # than it is initialised. This should be fixed properly in a PR.
```
ericpre, line 818: Thanks for adding this!

hyperspy/roi.py
```diff
@@ -700,19 +729,26 @@ class SpanROI(BaseInteractiveROI):
...
+        if t.Undefined in list(self):
+            self.left, self.right = get_central_half_limits_of_axis(ax0)
 
+    @property
+    def parameters(self):
+        return {"left":self.left, "right":self.right}
 
     def is_valid(self):
-        return (t.Undefined not in (self.left, self.right) and
+        return (not t.Undefined in list(self) and
```
ericpre: Sorry, I was misleading when writing `not t.Undefined in list(self) ` and I think it is more appropriate to use `tuple` instead of `list`: this is more standard and more consistent with the rest of the code.

hyperspy/roi.py
```diff
@@ -900,48 +949,55 @@ def _apply_roi2widget(self, widget):
...
+    cx, cy, r, r_inner = (t.CFloat(t.Undefined),) * 3 + (t.CFloat(0.),)
     _ndim = 2
 
-    def __init__(self, cx, cy, r, r_inner=None):
+    def __init__(self, cx=None, cy=None, r=None, r_inner=0):
         super(CircleROI, self).__init__()
+        cx, cy, r = (
+            para if para is not None 
+            else t.Undefined for para in (cx, cy, r))
+        r_inner = r_inner if r_inner is not None else 0
```
ericpre: `if r_inner is not None else 0` can be removed: by definition, `None` is not allowed, a `TraitError` will be raised if there is an attempt to set it to `None`.

hyperspy/tests/utils/test_roi.py
```diff
@@ -486,6 +486,37 @@ def test_line2droi_angle(self):
...
+
+    def test_undefined_call(self):
+        rect = RectangularROI(None, None, None, None)
+        with pytest.raises(AttributeError, match='not yet been set'):
+            rect(self.s_s)
+
+    def test_undefined_call_circle(self):
+        circ1 = CircleROI(None, None, None, None)
+        with pytest.raises(AttributeError, match='not yet been set'):
+            circ1(self.s_s)
```
ericpre: Do this test for all ROI, I would add a test which create the ROI then slice a signal to test `__call__`. This test is very fast and it is good to make sure that all ROI follows a well defined and consistent behaviour to avoid the current issue we have with `RectangularROI` for example!

hyperspy/utils/axes.py
```diff
@@ -0,0 +1,24 @@
...
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with HyperSpy.  If not, see <http://www.gnu.org/licenses/>.
+import hyperspy.axes
+
+__doc__ = hyperspy.axes.__doc__
+
+def get_central_half_limits_of_axis(ax):
```
ericpre: This create a cyclic import somewhere:
```python
import hyperspy.roi
```

hyperspy/tests/utils/test_roi.py
```diff
@@ -486,6 +486,37 @@ def test_line2droi_angle(self):
         with pytest.raises(ValueError):
             r.angle(axis='z')
 
+    def test_repr_None(self):
+        # Setting the args=None sets them as traits.Undefined, which didn't 
+        # have a string representation in the old %s style.
```
ericpre, line 489: Rewrite this to something along the line of:
```python
    for roi in [Point1DROI, Point2DROI, RectangularROI, SpanROI]:
        r = roi()
        for value in tuple(r):
            assert value == t.Undefined
        repr(r)
    for roi in [CircleROI, Line2DROI]:
        r = roi()
        for value in tuple(r)[:-2]:
            assert value == t.Undefined
        assert tuple(r)[-1] == 0
        repr(r)
```

hyperspy/tests/utils/test_roi.py
```diff
@@ -486,6 +486,37 @@ def test_line2droi_angle(self):
...
+        with pytest.raises(AttributeError, match='not yet been set'):
+            rect(self.s_s)
+
+    def test_undefined_call_circle(self):
+        circ1 = CircleROI(None, None, None, None)
+        with pytest.raises(AttributeError, match='not yet been set'):
+            circ1(self.s_s)
+        circ2 = CircleROI(5, 5, 1, None)
+        circ2(self.s_s) # r_inner as undefined should not raise error
+
```
ericpre, line 522: Testing the `add_widget` method would be good too.
It should be possible to put this in a for loop as in comment above.

hyperspy/roi.py
```diff
@@ -173,6 +190,11 @@ def __call__(self, signal, out=None, axes=None):
               space can fit the right number of axis, and use that if it
               fits. If not, it will try the signal space.
         """
+        if t.Undefined in list(self):
```
ericpre: Use `is_valid` here?

hyperspy/roi.py
```diff
@@ -173,6 +190,11 @@ def __call__(self, signal, out=None, axes=None):
...
+            raise AttributeError(
+        "Some ROI parameters have not yet been set. "
```
ericpre: Since this error message is used several times, I would define it in this module and reuse it. I am not sure this is a `AttributeError`, because the attribute exists and doesn't fail, however its value is not compatible, so I would think, this is more a `ValueError`.
https://docs.python.org/3/library/exceptions.html#AttributeError

ericpre (changes requested):
Thanks, I find this version much better! My reviews may have sound a bit picky and annoying but I really think, we should keep the ROI classes as tidy as possible.

I think the structure is good now and the suggestions I have added, are mainly about tests and other improvements.

hyperspy/utils/axes.py
thomasaarholt: Good catch. But I'm getting a headache just looking at it. I can easily solve it by just keeping it in `axes.py`, or by deleting `hyperspy/utils/roi.py` (it's got nothing in it?). But that feels like cheating.

Doesn't the existance of `import hyperspy.roi` in `hyperspy/utils/roi.py` mean that one cannot use it in `hyperspy/roi.py` without circular imports? That seems counterproductive?

ericpre: `hyperspy/utils/roi.py` defined what goes in the API through `hyperspy.api_nogui`, so I would leave it as it is for now!

Putting this function in `hyperspy.roi` should be good enough and possibly as a private function. Do you think it could used elsewhere?

hyperspy/roi.py
thomasaarholt: Are you suggesting that we "overwrite" the ROI parameters if they are set? I think there is value in being able to give a ROI non-default parameters (for instance, from a different previous ROI) and be able to "take a look" interactively to see where it sits - especially since we have no "non-interactive" ROI plot functionality!

hyperspy/roi.py
thomasaarholt: Gotcha. Also, I saw from the code you highlighted here that `A not in B` is much more readable that `not A in B`!

hyperspy/roi.py
ericpre: I am suggesting that the method does what its name implies.
> I think there is value in being able to give a ROI non-default parameters (for instance, from a different previous ROI) and be able to "take a look" interactively to see where it sits - especially since we have no "non-interactive" ROI plot functionality!

I don't understand how does it relate with the `_set_default_values` method here?

hyperspy/roi.py
thomasaarholt: Yep

hyperspy/tests/utils/test_roi.py
thomasaarholt: How about like this?
```python
    def test_widget_initialisation(self):
        self.s_s.plot()
        for roi in [Point1DROI, Point2DROI, RectangularROI, SpanROI, Line2DROI, CircleROI]:
            r = roi()
            r._set_default_values(self.s_s)
            r.add_widget(self.s_s)
```

hyperspy/tests/utils/test_roi.py
thomasaarholt: So far I have gone for the following, which is what I interpreted as what you meant, but I'm not sure how that is supposed to deal with the ordering issue with RectangularROI.

```python
    def test_undefined_call(self):
        for roi in [Point1DROI, Point2DROI, RectangularROI, SpanROI, Line2DROI, CircleROI]:
            r = roi()
            with pytest.raises(ValueError, match='not yet been set'):
                r(self.s_s)

    def test_default_values_call(self):
        for roi in [Point1DROI, Point2DROI, RectangularROI, SpanROI, Line2DROI, CircleROI]:
            r = roi()
            r._set_default_values(self.s_s)
            r(self.s_s)
```

hyperspy/roi.py
thomasaarholt: Nice

hyperspy/tests/utils/test_roi.py
thomasaarholt: Great. Used your example, but changed [::-2] to [::-1], as only the final parameter is != Undefined.

hyperspy/tests/utils/test_roi.py
ericpre: What needs to be tested here is that all ROI set the parameters after `_set_default_values`. If it doesn't fit well in a for loop, maybe it is better to do one ROI the one after another. I would keep your test as it is and add another test using interactive and checking that the values are correctly set by `_set_default_values`. These different tests will test slightly different supported usages.

ericpre:
Would it be possible to finish this PR while it is getting closed to be finished?

hyperspy/roi.py
thomasaarholt: I totally misunderstood your intent. Agreed and fixed.

thomasaarholt:
~~Should be good now.~~
~~Edit: Fixing now~~

Ok, please take a look

ericpre (approved):

ericpre:
Great, thanks!

PR #2343, opened 2020-03-24
Support for axis midpoint_value and midpoint_index

thomasaarholt:
### Description of the change
There have been several times I've wanted to be able to quickly get hold of the position between the `low_value` and `high_value` of an axis in the `axes_manager`. This PR adds that, and the index, in the form of `midpoint_value` and `midpoint_index`.

### Progress of the PR
- [x] add tests,
- [x] ready for review.

```python
data = np.random.random((20, 20))
s = hs.signals.Signal2D(data)
s.axes_manager[0].offset = 15
s.axes_manager[0].scale = 3
print("s.axes_manager[0].midpoint_value:", s.axes_manager[0].midpoint_value)
print("s.axes_manager[0].midpoint_index:", s.axes_manager[0].midpoint_index)

# s.axes_manager[0].midpoint_value: 43.5
# s.axes_manager[0].midpoint_index: 10
```

### DIFF
--- a/hyperspy/axes.py
+++ b/hyperspy/axes.py
@@ -212,9 +212,11 @@ class DataAxis(t.HasTraits, UnitConversion):
     size = t.CInt()
     low_value = t.Float()
     high_value = t.Float()
+    midpoint_value = t.Float()
     value = t.Range('low_value', 'high_value')
     low_index = t.Int(0)
     high_index = t.Int()
+    midpoint_index = t.Int()
     slice = t.Instance(slice)
     navigate = t.Bool(t.Undefined)
     index = t.Range('low_index', 'high_index')
@@ -473,6 +475,8 @@ class DataAxis(t.HasTraits, UnitConversion):
         if len(self.axis) != 0:
             self.low_value, self.high_value = (
                 self.axis.min(), self.axis.max())
+            self.midpoint_value = self.low_value + (self.high_value - self.low_value) / 2
+            self.midpoint_index = self.value2index(self.midpoint_value)
 
     def _update_slice(self, value):
         if value is False:
--- a/hyperspy/axes.py
+++ b/hyperspy/axes.py
@@ -212,7 +212,6 @@ class DataAxis(t.HasTraits, UnitConversion):
     size = t.CInt()
     low_value = t.Float()
     high_value = t.Float()
-    midpoint_value = t.Float()
     value = t.Range('low_value', 'high_value')
     low_index = t.Int(0)
     high_index = t.Int()
@@ -470,13 +469,17 @@ class DataAxis(t.HasTraits, UnitConversion):
     def update_index_bounds(self):
         self.high_index = self.size - 1
 
+    def fraction_value(self, fraction):
+        return self.low_value + (self.high_value - self.low_value) * fraction
+    
+    def fraction_index(self, fraction):
+        return self.value2index(self.fraction_value(fraction))
+
     def update_axis(self):
         self.axis = generate_axis(self.offset, self.scale, self.size)
         if len(self.axis) != 0:
             self.low_value, self.high_value = (
                 self.axis.min(), self.axis.max())
-            self.midpoint_value = self.low_value + (self.high_value - self.low_value) / 2
-            self.midpoint_index = self.value2index(self.midpoint_value)
 
     def _update_slice(self, value):
         if value is False:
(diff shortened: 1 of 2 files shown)


ericpre:
> There have been several times I've wanted to be able to quickly get hold of the position between the `low_value` and `high_value` of an axis in the `axes_manager`. 

Can you please provide examples where it would be useful?
Any reason why the middle would be more useful than the 1/3, 1/4 or anything else? Maybe a convenience function providing any fractional position along the axis would be more useful.

thomasaarholt:
> Can you please provide examples where it would be useful?

Most recently in #2341 for the default values of Point1DROI and Point2DROI and CircleROI, which triggered a memory of wanting this previously. But I can't remember what that was right now :/

> Any reason why the middle would be more useful than the 1/3, 1/4 or anything else? Maybe a convenience function providing any fractional position along the axis would be more useful.

Yes, this would be much more useful, good idea. It works nicely into the rest of the ROIs in the previous PR. `ax.fraction_value(fraction)` or `ax.value_at_fraction(fraction)`? Do we want it bounded so that `ax.low_value <= ax.fraction_value <= ax.high_value` is enforced?

ericpre:
> Yes, this would be much more useful, good idea. It works nicely into the rest of the ROIs in the previous PR. `ax.fraction_value(fraction)` or `ax.value_at_fraction(fraction)`?

`ax.value_at_fraction(fraction)` or `ax.value_at_proportion(fraction)` should both work, not sure what is best.

>  Do we want it bounded so that `ax.low_value <= ax.fraction_value <= ax.high_value` is enforced?

If the math are correct, I am not convinced it is necessary to checking this...

francisco-dlp:
I think that this calculation is so simple that it is easier to write it explicitly than finding this method and reading how to use it. This is true of other methods, and we should maybe consider streamlining hyperspy for v2.0.

dnjohnstone:
> I think that this calculation is so simple that it is easier to write it explicitly than finding this method and reading how to use it. This is true of other methods, and we should maybe consider streamlining hyperspy for v2.0.

Agree - I'd put this in the workflow rather than repo pile.

thomasaarholt:
That's fine. I'll close this. I'm toying with just turning it into a function rather than a method, but will only revisit this if I think it'll save on boilerplate code later on.

ericpre:
> I think that this calculation is so simple that it is easier to write it explicitly than finding this method and reading how to use it. This is true of other methods, and we should maybe consider streamlining hyperspy for v2.0.

Even if this is a fairly simple calculation, having a convenience method, such as `relative2value`, which would return the axis value corresponding to a relative coordinate (ranging 0 to 1) of the axis, would make the code more readable, for example to zoom in the centre of a power spectrum:
```python
import hyperspy.api as hs
s = hs.datasets.example_signals.object_hologram().fft(True)
s.plot(True)
ax = plt.gca()
```
 
```python
start = [(axis.axis[-1] - axis.axis[0]) / 4 + axis.axis[0] for
         axis in signal_axes]
end = [(axis.axis[-1] - axis.axis[0]) * 3 / 4 + axis.axis[0] for
       axis in signal_axes]
ax.set_xlim(start[0], end[0])
ax.set_xlim(start[1], end[1])
```
in comparison to 
```python
start = [axis.relative2value(1/4) for axis in signal_axes]
end = [axis.relative2value(3/4) for axis in signal_axes]
ax.set_xlim(start[0], end[0])
ax.set_ylim(start[1], end[1])
```

The idea of relative coordinates could also be useful in slicing, for example: `s.isig[:'r0.1'] ` or to initialise ROI.

francisco-dlp:
@ericpre, with you example the feature seems more appealing to me. The ``'r0.1'`` syntax is brilliant! I would only suggest to use ``'rel0.1'`` for better readability.

ericpre:
@thomasaarholt: what do you think about the `rel0.1` syntax? Would you be interested in implementing it?
This should be useful for #2341?

thomasaarholt:
Yes, I do like the idea and am tentatively happy to implement it. I had a quick look at what needs doing now, and [the code](https://github.com/hyperspy/hyperspy/blob/RELEASE_next_minor/hyperspy/misc/slicing.py#L196) is largely undocumented. I'll start by commenting the code a bit and see how far I get from there.

ericpre:
I would be surprised if you need to do much in [slicing](https://github.com/hyperspy/hyperspy/blob/RELEASE_next_minor/hyperspy/misc/slicing.py#L196) itself. The place where it most likely needs to go is in the `DataAxis` or `AxesManager` class.

https://github.com/hyperspy/hyperspy/pull/1730 may be useful as reference.

PR #2348, opened 2020-04-02
Fix CI

ericpre:
- [x] Temporary fix for png issue with pillow 7.1.0 (https://github.com/python-pillow/Pillow/issues/4509)
- [x] Fix matplotlib installation in the pip environment

The png issue should be fixed in pillow 7.1.1 (https://github.com/python-pillow/Pillow/pull/4512, https://github.com/python-pillow/Pillow/pull/4515)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -51,9 +51,6 @@ install:
   - if [[ $MINIMAL_ENV == false ]] ; then
       DEPS="${DEPS} ${DEPS_OPTIONAL}";
     fi
-  - if [[ $ENV == pip ]] ; then
-      DEPS="${DEPS} pillow==7.0";
-    fi
   - source ./ci/travis_install_with_$ENV.sh
   - pip install .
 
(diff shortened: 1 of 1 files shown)


ericpre:
The travis results does not show in the "check list"  at the moment. According to this issue (https://travis-ci.community/t/travis-ci-build-no-longer-showing-as-check-on-github-pull-request-page/6955/5), it may be necessary to reset grant access to Travis CI to fix it. I have just did for my account and it worked.

.travis.yml
```diff
@@ -51,6 +51,9 @@ install:
   - if [[ $MINIMAL_ENV == false ]] ; then
       DEPS="${DEPS} ${DEPS_OPTIONAL}";
     fi
+  - if [[ $ENV == pip ]] ; then
```
francisco-dlp: Could you add some comments here explaining why this is needed?

ericpre: This was to check that the pillow version was the issue. pillow 7.1.1 have been released a few hours ago, once it is available on pypi, I will remove this.

PR #2351, opened 2020-04-06
Phenom elid reader

djeedjay:
### Description of the change
This PR adds an io_plugin for reading Phenom desktop SEM .elid files.

### Progress of the PR
- [x] add tests,
- [x] ready for review.

```python
>>> import hyperspy.api as hs
>>> s = hs.load('PhenomEds.elid')
>>> s[0].plot()
```


hyperspy/io_plugins/phenom.py
```diff
@@ -0,0 +1,659 @@
...
+        eds_metadata['offset'] = offset
+        eds_metadata['dispersion'] = dispersion
+        has_variable_real_time = self._read_bool()
+        has_variable_live_time = self._read_bool()
+        data = np.empty([size, bins], dtype=np.uint32)
+        for i in range(size):
+            for bin in range(bins):
+                data[i, bin] = self._read_varuint32()
+        if has_variable_real_time:
+            real_time_values = [self._read_float64() for _ in range(size)]
```
ericpre: `real_time_values` are not used anywhere?

hyperspy/io_plugins/phenom.py
```diff
@@ -0,0 +1,659 @@
...
+        else:
+            real_time_values = [self._read_float64()] * size
+        if has_variable_live_time:
+            live_time_values = [self._read_float64() for _ in range(size)]
+        else:
+            live_time_values = [self._read_float64()] * size
```
ericpre: `live_time_values` are not used anywhere?

ericpre (changes requested):
Sorry for coming back late on this, our CI was not very fit recently but now it is in good shape! There are only a few details to sort out and then it should be good to merge:
- merge with upstream to update the CI and fix the merge conflict. Azure pipeline is going to fail to because the last version of tifffile package on conda-forge doesn't support it - installing `tifffile<=2019.7.26.2` does work, because of `tifffile` dependencies changes in version `2020.2.16` and this is not yet packaged on conda-forge. Since an appropriate error message is raised, so this is fine. `tifffile` package from pypi should be fine and this is covered on travis for the linux platform. To fix the failure on azure pipeline, it would be good to skip the phenom elid test when `imagecodecs` is not installed.
- the couple of images of the test images contains 0 only, is this normal? The EDS data are fine.
- there are a few variables not used - see comments below. If this is intentional, it would be good to leave a comment about it.

More details on the `tifffile` packaging issue on conda-forge:
https://github.com/conda-forge/tifffile-feedstock/issues/39
https://github.com/conda-forge/staged-recipes/pull/10331

gjdevos:
* test_phenom will now be skipped if imagecodes is not available.
* I replaced the original SEM images in a file by small empty images to reduce its size from 7 MB to just 100 kB. I suppose this is OK for the unit tests. If something would break in the images, the decoding of the tiff data object would fail.
* Thank you for spotting the missing live time and real time values in the metadata.

ericpre (approved):

ericpre:
Thanks! Indeed reducing the image from 7MB to 100 kB is very sensible! On build is failing because it could not connect to the EELS database, so this is fine.

For scientific communication (mostly for presentation), we put the affiliations of contributors but unfortunately, we don't have any better way to manually pull the information from the github contributor accounts. If you want to have your institution/company, you can update you github account (and we will get it from there) and provide this information here; up to you.

Another last thing slightly out of the scope of this PR: we have been discussing about making sure that live and real time are defined consistently in HyperSpy: should it be per pixel, map, frame, per detector is several, etc... According to this PR, it seems that live and real time _can be_ defined per pixel but not always. If defined per pixel, can you confirm this is the total time and not per frame? Any other relevant information is welcome!

djeedjay:
Thanks for merging my PR! Any idea about the next release timeframe?

I work for Thermo Fisher Scientific in the Phenom team. We like to support HyperSpy but we don't want this to be labeled as ThermoFisher software. That's why I chose to put it on my personal github account. You can document phenom file support as contributed by ThermoFisher, but I will not add the company name to my personal github account.

Ideally, you want real time and live time information for every spectrum. This means for every pixel in a line or grid scan. Depending on hardware, we can not always supply this for every pixel. Specifically, some of our systems support direct synchronisation between scan and EDS detector for every pixel position. This way, the scan and EDS detector can run continuously and we get the most efficient acquisition. Unfortunately, in this mode the EDS detector does not produce per pixel live time values.

francisco-dlp:
We have to resume discussion about the next release, but I think that most will agree that, given the high number of contributions that have made it into Rnm since the last release, v1.6 is long overdue, so chances are that we will agree to make a new release in the next few weeks. Does it sound good?

djeedjay:
The next few weeks is sooner than I expected, thank you!

PR #2352, opened 2020-04-06
Fix maximum-likelihood PCA method

tjof2:
### Description of the change

- Rewrite the MLPCA implementation in `hyperspy/learn/mlpca.py` to make it less like MATLAB and more like Python/Numpy
- Looking again at the reference implementation (DOI:[10.1016/S0003-2670(97)00270-5](https://doi.org/10.1016/S0003-2670(97)00270-5)), I think there's either a bug or a bit of the code missed that severely impacts performance. Since we're passing `output_dimension`, we only need `U = V[:output_dimension]` from each SVD, not the full matrix. Convergence seems much quicker as a result.
- I've made use of `np.linalg.multi_dot()`.
- The `kwargs` were never accessible from `s.decomposition()` - they are now, and I've renamed them to be clearer as `max_iter` and `tol`.
- We never explicitly test the MLPCA method, so I've added a quick unit test.

I intend to update the documentation properly as part of #1159, but I have improved the docstring here.

### Progress of the PR
- [x] add tests,
- [x] ready for review.

This is just an arbitrary low-rank matrix.
```python
import numpy as np

m, n, r = 100, 100, 3
rng = np.random.RandomState(101)
U = rng.uniform(0, 1, size=(m, r))
V = rng.uniform(0, 1, size=(n, r))
X = np.dot(U, V.T) + 0.01 * rng.randn(m, n)
varX = X.copy() / 2

# This is the old function on RELEASE_next_minor
%timeit x = mlpca_old(X, varX, p=3, fast=False, maxiter=100)
# 4.18 s ± 105 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)

# This is the new function on this branch
%timeit x = mlpca(X, varX, output_dimension=3, fast=False, max_iter=100)
# 181 ms ± 30.8 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
```

I should add that I checked it returned consistent results against the old function, again using an arbitrary low-rank matrix.


tjof2:
@francisco-dlp I think you're the best person to review this.

tjof2:
I should add that I checked it returned consistent results against the old function, again using an arbitrary low-rank matrix.

tjof2:
Tests failing on latest commit due to:
```
=========================== short test summary info ============================
FAILED hyperspy/tests/datasets/test_eelsdb.py::test_eelsdb_eels - requests.ex...
= 1 failed, 2733 passed, 146 skipped, 3 xfailed, 1 xpassed, 54 warnings in 642.06s (0:10:42) =
```

tjof2:
Codecov has some odd ideas about coverage here. Should be all good for review though, I updated the tests and made a few improvements.

tjof2:
Unrelated test failures (as always), and Codecov report seems a bit off. But ready for review - the latest change to remove `np.diag()` is good.

francisco-dlp:
This is great and it LGTM except for the test. I've send you a PR https://github.com/tjof2/hyperspy/pull/1 because I found it easier to modify it the code than explaining what the issue is.

francisco-dlp:
Oh, I forgot to mention a small detail: could you replace the few remaining ``np.dot`` with ``@``?

tjof2:
Thanks will do!

tjof2:
@francisco-dlp I changed the `np.dot()`s as requested and merged your change - this should be good to merge now assuming the only failing test is due to `mrcz` (cf. #2369).

tjof2:
Yeah, `TestPythonMrcz.test_Async[complex64]` failed _again_ :)

PR #2355, opened 2020-04-08
Add exponential component to the options for remove_background()

jasoncmyers:
### Description of the change
This modifies the exponential component to allow its use with remove_background().  This matches the "1st order log-polynomial" background option from Digital Micrograph, which is useful for fitting the low-loss regions of some EELS spectra. 
Also adds new unit tests for estimate_parameters() and function_nd().

### Progress of the PR
- [X] Change implemented,
- [X] update docstrings,
- [X] add tests,
- [X] ready for review.

```python
>>> import hyperspy.api as hs
>>> import numpy as np
>>> s = hs.signals.Signal1D(np.geomspace(1000, 10, num=100))
>>> s2 = s.remove_background(signal_range=(5., 20.), background_type='Exponential')
```


hyperspy/_components/exponential.py
```diff
@@ -57,3 +61,97 @@ def __init__(self, A=1., tau=1., module="numexpr", **kwargs):
...
+                t = -x2 / (log(geo_mean2) - log(A))
+
+                if s._lazy:
+                    A = A.map_blocks(np.nan_to_num)
+                    t = t.map_blocks(np.nan_to_num)
+                else:
+                    A = np.nan_to_num(A)
+                    t = np.nan_to_num(t)
+
+            except (FloatingPointError):
```
ericpre, line 136: When testing, I had the following warning:
```python
/home/eric/Dev/hyperspy/hyperspy/_components/exponential.py:120: RuntimeWarning: invalid value encountered in double_scalars
  geo_mean1 = exp(b1.sum(axis=-1)/b1.shape[-1])
/home/eric/Dev/hyperspy/hyperspy/_components/exponential.py:121: RuntimeWarning: invalid value encountered in double_scalars
  geo_mean2 = exp(b2.sum(axis=-1)/b2.shape[-1])
```
I think the `RuntimeWarning` could be added too here.

ericpre (changes requested):
Thanks @jasoncmyers, this is a very clean PR! Beside on comment on the capture of the warning, the docstring of the `remove_background` method needs to be updated to add  `Exponential` to the list of possible background type. Aslo I would add this background type to the  test: https://github.com/hyperspy/hyperspy/blob/deff3fb05a9ab9e7325199aab43861a229f81fd2/hyperspy/tests/signal/test_remove_background.py#L227.

hyperspy/_components/exponential.py
jasoncmyers: I've added RuntimeWarning to the `except` block, but I'm curious what test caused the warning.  Everything I saw while working on it was some variation of divide-by-zero.

hyperspy/_components/exponential.py
```diff
@@ -57,3 +61,97 @@ def __init__(self, A=1., tau=1., module="numexpr", **kwargs):
...
+
+                if s._lazy:
+                    A = A.map_blocks(np.nan_to_num)
+                    t = t.map_blocks(np.nan_to_num)
+                else:
+                    A = np.nan_to_num(A)
+                    t = np.nan_to_num(t)
+
+            except (RuntimeWarning, FloatingPointError):
+                _logger.warning('Exponential paramaters estimation failed with'
```
tjof2: "parameters"

hyperspy/_components/exponential.py
```diff
@@ -57,3 +61,97 @@ def __init__(self, A=1., tau=1., module="numexpr", **kwargs):
...
+            log = np.log
+
+        with np.errstate(divide='raise'):
+            try:
+                # use log and exp to compute geometric mean to avoid overflow
+                a1 = s.isig[i1:i_mid].data
+                b1 = log(a1)
+                a2 = s.isig[i_mid:i2].data
+                b2 = log(a2)
+                geo_mean1 = exp(b1.sum(axis=-1)/b1.shape[-1])
```
tjof2: How likely is it for `b1.sum(axis=-1)/b1.shape[-1]` > 709? Since this will still cause `RuntimeWarning: overflow encountered in exp`.

hyperspy/_components/exponential.py
ericpre: I was testing it interactively and it happens most likely because of the signal_range being too small. To reproduce:
```python
import hyperspy.api as hs

exponential = hs.model.components1D.Exponential()
exponential.A.value = 12500.
exponential.tau.value = 168.

signal = hs.signals.Signal1D(
    exponential.function(np.arange(100, 200, 0.02)))

# use a signal range too small
signal.remove_background(signal_range=(19.425, 19.426),
            background_type='Exponential',
            show_progressbar=None)
```

hyperspy/_components/exponential.py
ericpre: Indeed it may happen with very high number, but overall, it seems to be fairly unlikely.

To make it more simple, replace by:
```suggestion
                geo_mean1 = exp(b1.mean(axis=-1))
```
Same for geo_mean2

ericpre:
@jasoncmyers: I have pushed a commit to this branch to fix the test failure: the test suite hang on appveyor python 3.8 and running a garbage collection cycle did the trick. Make sure to pull this change before sorting out the last few details.

The failure on travis is fixed in #2361, so don't worry about this.

hyperspy/_components/exponential.py
jasoncmyers: Thanks, that was helpful!  I've changed it to raise an exception if the range is too small and then catch it and log an appropriate warning.

hyperspy/_components/exponential.py
jasoncmyers: Overflowing here would mean that the geometric mean of the signal range is greater than the float64 max value, so it shouldn't be possible without another overflow already occurring in at least one channel of the signal.  

As suggested, I've simplified it to use `mean()`.

ericpre (approved):

PR #2356, opened 2020-04-09
Add autoscale to plot

ericpre:
Add `autoscale` parameter to the `plot` method be able to reset automatically the displayed range or not when the data are changing or when navigating the signal.

### Progress of the PR
- [x] Add `autoscale` parameter to `plot` functions,
- [x] ~~Add `intensity_autoscale` and `axes_autoscale` parameter to `plot` methods to enable reset automatically or not the intensity and axes independently (and dot it consistently for Signal1D and Signal2D)~~
- [x] fix passing arguments to navigator through `navigator_kwds` dictionary,
- [x] update docstring (still need some improvement?),
- [x] improve docstring of the `plot` functions,
- [x] set default value explicitly for each `plot` functions (easier to visualise the default values interactively),
- [x] improve artificial dataset functions,
- [x] improve closing figure (fix a few bugs),
- [x] add more tests,
- [x] ready for review.

```python
import hyperspy.api as hs
s = hs.datasets.artificial_data.get_core_loss_eels_line_scan_signal(add_powerlaw=True)
s.plot(axes_autoscale=False)
ax = s._plot.signal_plot.ax
ax.set_xlim(500.0, 700.0)
ax.set_ylim(-10.0, 20.0)
```
Navigate the signal and the x axis will not changed anymore but the y axis will update automatically.


francisco-dlp:
All LGTM except for the changes to 1D autoscale. What's the point of adding ``"x"``? In hyperspy the x-axis cannot change on changing the navigation index. It could be useful to display the full axis after modifying the view interactively at a given pixel, but why not simply clicking on the figure's ``Reset original view`` button instead? Also, the way this is implemented breaks the current API because now ``True`` enables both axes autoscaling, while previously it affected only the ``y`` axis.

ericpre:
I wanted to make the `autoscale` parameter more consistent between `Signal1D` and `Signal2D` but actually it doesn't make sense. Agreed that if is not worth resetting automatically the signal axes  of the image because it can be done very easily using matplotlib.
Shall we ~remove~deprecate the `autoscale` parameter of `Signal2D` completely and reset the signal axes only when the signal axes are changed, as it is currently done with `Signal1D`?

francisco-dlp:
I think that ``autoscale`` is useful for ``Signal2D``  to compare absolute contrast between images.

ericpre:
@francisco-dlp I have changed it to make it more clear and explicit. If we are happy with this, then I will add more tests and finished it.

ericpre:
I have edited the first post of this PR to reflect the last changes.

ericpre:
The test failure in travis is due to the last release of scikit-image a couple of hours ago. This is fixed in #2393.

tjof2:
Does this just need the merge conflicts resolving?

ericpre:
@tjof2, this is finished from my side after I made it more explicit and clear following @francisco-dlp review.
Since this PR is introducing new arguments (`intensity_autoscale` and `axes_autoscale`), it would be good that we are happy with these changes and wait for @francisco-dlp's view on this.

hyperspy/_signals/signal1d.py
```diff
@@ -1546,6 +1548,30 @@ def estimating_function(spectrum,
 
     estimate_peak_width.__doc__ %= (SHOW_PROGRESSBAR_ARG, PARALLEL_ARG, MAX_WORKERS_ARG)
 
+    def plot(self,
+             navigator="auto",
+             plot_markers=True,
+             intensity_autoscale=True,
+             norm="auto",
+             axes_autoscale=False,
```
francisco-dlp: In the 1D case I would rename:

* ``axes_autoscale`` -> ``xaxis_autoscale``
* ``intensity_autoscale`` -> ``yaxis_autoscale``

francisco-dlp (changes requested):
Looks very good except for the small suggestion.

hyperspy/_signals/signal1d.py
ericpre: I still prefer the current syntax because it is consistent between 1D and 2D signal. More importantly, the `xaxis` and `yaxis` is consistent with matplotlib but not hyperspy: `x` and `y` in the context of axis in hyperspy can refer to navigation axis, which is then misleading!

francisco-dlp: The issue is that ``axes_autoscale`` is plural, while it just affects the x-axis, what I find confusing. As for ``intensity_autoscale``, it is not obvious that it refers to the ``y-axis``. I don't think that consistency with the 2D counterparts is required or even desired, since they actually act pretty differently. For examples ``axes_autoscale`` affects both axes in 2D but just the x axis in 1D. Furthermore, the docstring resorts to the x/y nomenclature to clarify which axes the keywords operate on...

ericpre: Good points, but I still think it would convenient to be consistent between 1D and 2D figure (actually not signal) because it can apply to the navigator too and be used through the `navigator_kwds` argument. These arguments will do different things in case of 1D and 2D but conceptually, there are equivalent.

Clearly, the docstring needs improvement! ;)

francisco-dlp: I think that that's one more argument for using different keywords because people tend to think that things with the same name do the same thing...

ericpre: > I think that that's one more argument for using different keywords because people tend to think that things with the same name do the same thing...

This doesn't mean that it is a good reason! Thinking about it a bit, I am convincing myself that it should be:

* `axes_autoscale` -> `axis_autoscale`
* `intensity_autoscale` stay the same

Because in hyperspy, we use the term `intensity` (e.g. in the `norm` argument) and here `axis` it refers to an `axis` of the `axes_manager`, and if there are 1, 2 or navigation or signal axes, it should not matter. 
The only occurrence of `xaxis` is in `xaxis_type` for plotting explained variance:
https://github.com/hyperspy/hyperspy/blob/54856fd9c9aefb62f7ea68f1d90431072151d060/hyperspy/learn/mva.py#L1294-L1307

francisco-dlp: I don't like ``axis_autoscale`` because it is not precise enough, which axis is it referring to?

I have another suggestion to try to break the impasse: use simply one keyword that can take `"x"`, `"y"` or `"xy"`. What do you think?

ericpre: > I don't like `axis_autoscale` because it is not precise enough, which axis is it referring to?

Well, it is referring to the relevant `axis` from the `axes_manager` (depending, if this is a navigator or signal Figure) against which the data are displayed.

> I have another suggestion to try to break the impasse: use simply one keyword that can take `"x"`, `"y"` or `"xy"`. What do you think?

I have been this route before and I find it confusing... I would actually prefer your first suggestion.

francisco-dlp: OK, I think that you'll like this small variation of my previous suggestion because it helps with consistency that is what you are (rightly) after: just one keyword that could be ``autoscale`` or ``axes_autoscale`` that takes `"x"`, `"y"` or `"xy"` (`"x"` and `"y"` not yet implemented for 2D. If we choose simply ``"autoscale"`` (my favourite), then we could remove the 2D ``intensity_autoscale`` keyword, since for 2D it could take `"x"`, `"y"`, `"z"` (for intensity) and any combination of those. I like this one, because it is minimalist, consistent and intuitive.

ericpre: Yes, this is good. Previously, I have considered `x`, `y`, or `xy` (but not `xyz`) and using `intensity_autoscale` separately, which was not great but having a combination of all of them in the same argument (`autoscale`) is nice - as you said minimalist and intuitive.

I can work on it this evening but if someone want to take over earlier, that's more than welcome! :)

francisco-dlp: After thinking on it, maybe it is not actually a good idea to include `"z"` in ``autoscale`` because it is redundant with (and less rich than) #2272's `"th"` syntax. So, maybe just ``axes_autoscale`` that behaves similarly (exactly the same is not possible) for 1D and 2D would do?

ericpre: This should be fine if we disable *z autoscaling* when `vmin` and/or `vmax` are provided?

francisco-dlp: We could indeed make `"z"` autoscaling a shortcut for ``vmin="0th", vmax="100th"`` but then, what happens if ``"z"`` is not in ``autoscale`` and ``vmin`` and ``vmax`` are not provided? I think that, when in doubt, simpler is better. Once ``saturated_pixels`` is gone, we can make by default ``vmin="0th", vmax="100th"``. In this case, disabling z-autoscaling simply involves setting a numeric value, what requires the user to know what he's doing (usually a good thing)

hyperspy/docstrings/plot.py
```diff
@@ -88,20 +88,65 @@
...
 
+
 PLOT2D_DOCSTRING = \
     """colorbar : bool, optional
             If true, a colorbar is plotted for non-RGB images.
+        autoscale : str
+            Set the axes to be reset to their maxima when the navigation
+            indices or data are changed. If 'x' or 'y' are in the string, the
+            extent of the image are reset automatically. If 'z' is in the
+            string, the displayed value of the image are reset automatically.
```
ericpre: Is this docstring clear enough?
There is a typo here: displayed value*s* but since there is a bit a queue on travis and azure pipeline, I will fix it after this is reviewed.

hyperspy/_signals/signal1d.py
ericpre: As it is currently implemented, if `vmin` and `vmin` are not provided and `'z'` is not in `autoscale`, the contrast will be set with for the first image to display. If the data change or the indices of the axes manager changes, the range of displayed values will not changed.
For hyperspy 2.0, when `saturated_pixels` is removed, then we can set the default values `vmin="0th"` and `vmin="100th"` and remove allowing setting `None` to `vmin`, `vmax` for simplification.

The alternative is to raise an error when `'z'` is not in `autoscale` and `vmin`/`vmax` are not provided, which is possibly what you mentioned earlier?

hyperspy/docstrings/plot.py
francisco-dlp: What about:

The string must contain any combination of the 'x', 'y' and 'z' characters. If 'x' or 'y' are in the string, the corresponding axis limits are set to cover the full range of the data at a given position. If 'z' is in the string when plotting images, the contrast will cover the full range of intensities of the image at each position.

ericpre: With changes regarding `z`:

The string must contain any combination of the 'x', 'y' and 'z' characters. If 'x' or 'y' are in the string, the corresponding axis limits are set to cover the full range of the data at a given position. If 'z' is in the string, the contrast of the image will be set automatically according to `vmin` and `vmax` when the data are changed or navigation indices are changed.

francisco-dlp: A small variation:

The string must contain any combination of the 'x', 'y' and 'z' characters. If 'x' or 'y' are in the string, the corresponding axis limits are set to cover the full range of the data at a given position. If 'z' is in the string, the contrast of the image will be set automatically according to `vmin` and `vmax` when the data or navigation indices change.

ericpre: Yes, it is good!

hyperspy/docstrings/plot.py
```diff
@@ -88,20 +88,65 @@
...
             The function used to normalize the data prior to plotting.
             Allowable strings are: ``'auto'``, ``'linear'``, ``'log'``.
             (default value is ``'auto'``).
             If ``'auto'``, intensity is plotted on a linear scale except when
             ``power_spectrum=True`` (only for complex signals).
+        autoscale : str
+            Set the axes to be reset to their maxima when the navigation
+            indices or data are changed. If 'x' is in the string, the
+            horizontal axis is reset automatically. If 'z' is in the string,
+            the vertical axis is reset automatically. Default is 'z'.
```
ericpre: To align with its signal2D counterpart:

The string must contain any combination of the 'x' and 'z'
characters. If 'x' or 'z' are in the string, the corresponding
axis limits are set to cover the full range of the data at a given
position and the axis limits will reset when the data or the
navigation indices are changed. Default is 'z'.

francisco-dlp: The docstring looks good, but the usage of 'z' seems counter-intuitive for me, couldn't that be 'y'? I know what's the rationale behind your choice, but it seems weird to name an axis 'z' when there is no y-axis...

ericpre: Indeed, and I like that autoscale='z' for plotting signal1D and signal2D does the same thing. Is there something different from `z` and `y` which could be used?

francisco-dlp: Well, maybe "i" for intensity. I find "z" more straightforward, but maybe that's just me. I think that it would be good to have some more opinions on this if possible. @hyperspy/developers, any opinion on this?

ericpre: I was also thinking about 'i' but I wasn't really convinced.

francisco-dlp: Yes, the issue with "i" is that a code that contains "yi" is not identified immediately by most people as some sort of axes configuration, while for "yz" it is easier to make a winning guess of what it means.

ericpre: What's about `v` for values? It fits well with `vmin` and `vmax`, which is used for images. Then it would:
- `xv` for signal2D
- `xyv` for signal2D

francisco-dlp: I would prefer ``i`` in that case, however my favourite remains ``z``. Let's try to break the gridlock, @tjof2, what do you think?

tjof2: Ha, I generally see `z` as meaning "depth", so would be bad for any future 3D plotting, while `v` is good for consistency with `vmin` and `vmax`.

francisco-dlp: Good point, I agree.

tjof2: For example say there was a theoretical `Signal3D` you could do `xyzv` as a string then as @ericpre says - e.g. scaling a 3D point cloud where each point is coloured according to a value.

ericpre:
It should be all good now!

francisco-dlp:
Looks good except for one small detail:

```python
s.plot(autoscale="Raise an error!")
```
should raise a ``ValueError``.

francisco-dlp:
Also this is missing documentation in the User Guide.

francisco-dlp:
And the docstring should mention that ``saturated_pixels``  is deprecated:

```rest
.. deprecated:: 1.6.0
          `saturated_pixels` will be removed in HyperSpy 2.0.0, it is replaced by
          `vmin`, `vmax` and `autoscale`.
```

francisco-dlp:
LGTM, would you like to fix the conflicts?

tjof2 (approved):

PR #2358, opened 2020-04-09
Performance improvements for estimate_shift_2D() and other FFTs

tjof2:
### Description of the change
As part of this [pyxem issue](https://github.com/pyxem/pyxem/issues/585), I noticed that `s.align2D()` spends a long time estimating the shifts between images. I've tracked this down to two bottlenecks:
- Switch `scipy.signal.medfilt` for `scipy.ndimage.median_filter`. 
The scipy documentation [reference for the old function](https://docs.scipy.org/doc/scipy/reference/generated/scipy.signal.medfilt.html) says itself that "The more general function scipy.ndimage.median_filter has a more efficient implementation of a median filter and therefore runs much faster.", so this seems like a sensible change.
- Changed how we calculate the optimal FFT size to be consistent with [`scipy.signal.fftconvolve`](https://docs.scipy.org/doc/scipy/reference/generated/scipy.signal.fftconvolve.html).

~On a related note, I see that in #2295, the code was switched from scipy FFT to numpy FFT. On my machine at least, the numpy FFT correlation implementation is 2x slower than scipy fftconvolve for the examples below.~ I know why this is now, it is fixed below by using `rfft()` where possible.

### Progress of the PR
- [x] Change implemented
- [x] Ready for review

```python
>>> s
<Signal2D, title: , dimensions: (100|200, 200)>

>>> %timeit s.estimate_shift2D(show_progressbar=False)
# Using sp.signal.medfilt
4.80 s ± 127 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
# Using sp.ndimage.median_filter
2.77 s ± 29.2 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)

>>> %timeit s.estimate_shift2D(show_progressbar=False, medfilter=False)
# Using old FFT size calculation
2.22 s ± 30 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
# Using new FFT size calculation
1.21 s ± 34.3 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
```
If we combine the two changes together:
```python
>>> %timeit s.estimate_shift2D(show_progressbar=False)
# Before
4.80 s ± 127 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
# After
1.92 s ± 40.7 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
```

@dnjohnstone I think this solves most of the issue discussed in pyxem for `center_direct_beam()`, aside from the threading already mentioned.

### DIFF
--- a/hyperspy/_signals/eels.py
+++ b/hyperspy/_signals/eels.py
@@ -30,8 +30,8 @@ from hyperspy.misc.elements import elements as elements_db
 import hyperspy.axes
 from hyperspy.defaults_parser import preferences
 from hyperspy.components1d import PowerLaw
-from hyperspy.misc.utils import (
-    isiterable, closest_power_of_two, underline)
+from hyperspy.misc.utils import isiterable, underline
+from hyperspy.misc.math_tools import optimal_fft_size
 from hyperspy.ui_registry import add_gui_method, DISPLAY_DT, TOOLKIT_DT
 from hyperspy.docstrings.signal1d import CROP_PARAMETER_DOC
 from hyperspy.docstrings.signal import SHOW_PROGRESSBAR_ARG, PARALLEL_ARG
@@ -630,9 +630,9 @@ class EELSSpectrum_mixin:
         tapped_channels = s.hanning_taper()
         # Conservative new size to solve the wrap-around problem
         size = zlp_size + self_size - 1
-        # Increase to the closest power of two to enhance the FFT
-        # performance
-        size = closest_power_of_two(size)
+        # Calculate optimal FFT padding for performance
+        complex_result = (zlp.data.dtype.kind == 'c' or s.data.dtype.kind == 'c')
+        size = optimal_fft_size(size, not complex_result)
 
         axis = self.axes_manager.signal_axes[0]
         if self._lazy or zlp._lazy:
@@ -1223,7 +1223,7 @@ class EELSSpectrum_mixin:
             # Use a size that is a power of two to speed up the fft and
             # make it double the closest upper value to workaround the
             # wrap-around problem.
-            esize = 2 * closest_power_of_two(axis.size)
+            esize = 2 * optimal_fft_size(axis.size)
             q = -2 * np.fft.fft(Im, esize,
                                 axis.index_in_array).imag / esize
 
--- a/hyperspy/_signals/eels.py
+++ b/hyperspy/_signals/eels.py
@@ -1222,7 +1222,7 @@ class EELSSpectrum_mixin:
             # Use an optimal FFT size to speed up the calculation, and
             # make it double the closest upper value to workaround the
             # wrap-around problem.
-            esize = 2 * optimal_fft_size(axis.size)
+            esize = optimal_fft_size(2 * axis.size)
             q = -2 * np.fft.fft(Im, esize,
                                 axis.index_in_array).imag / esize
 
--- a/hyperspy/_signals/eels.py
+++ b/hyperspy/_signals/eels.py
@@ -1222,7 +1222,7 @@ class EELSSpectrum_mixin:
             # Use an optimal FFT size to speed up the calculation, and
             # make it double the closest upper value to workaround the
             # wrap-around problem.
-            esize = optimal_fft_size(2 * axis.size)
+            esize = 2 * optimal_fft_size(axis.size)
             q = -2 * np.fft.fft(Im, esize,
                                 axis.index_in_array).imag / esize
 
--- a/hyperspy/_signals/eels.py
+++ b/hyperspy/_signals/eels.py
@@ -744,9 +744,8 @@ class EELSSpectrum_mixin:
         cl_size = self.axes_manager.signal_axes[0].size
         # Conservative new size to solve the wrap-around problem
         size = ll_size + cl_size - 1
-        # Increase to the closest multiple of two to enhance the FFT
-        # performance
-        size = int(2 ** np.ceil(np.log2(size)))
+        # Calculate the optimal FFT size
+        size = optimal_fft_size(size)
 
         axis = ll.axes_manager.signal_axes[0]
         if fwhm is None:
@@ -1220,7 +1219,7 @@ class EELSSpectrum_mixin:
             # Kramers Kronig Transform:
             # We calculate KKT(Im(-1/epsilon))=1+Re(1/epsilon) with FFT
             # Follows: D W Johnson 1975 J. Phys. A: Math. Gen. 8 490
-            # Use a size that is a power of two to speed up the fft and
+            # Use an optimal FFT size to speed up the calculation, and
             # make it double the closest upper value to workaround the
             # wrap-around problem.
             esize = 2 * optimal_fft_size(axis.size)
--- a/hyperspy/_signals/signal2d.py
+++ b/hyperspy/_signals/signal2d.py
@@ -26,18 +26,13 @@ from skimage.feature.register_translation import _upsampled_dft
 
 from hyperspy.defaults_parser import preferences
 from hyperspy.external.progressbar import progressbar
-from hyperspy.misc.math_tools import symmetrize, antisymmetrize
+from hyperspy.misc.math_tools import symmetrize, antisymmetrize, optimal_fft_size
 from hyperspy.signal import BaseSignal
 from hyperspy._signals.lazy import LazySignal
 from hyperspy._signals.common_signal2d import CommonSignal2D
 from hyperspy.docstrings.plot import (
     BASE_PLOT_DOCSTRING, PLOT2D_DOCSTRING, KWARGS_DOCSTRING)
 from hyperspy.docstrings.signal import SHOW_PROGRESSBAR_ARG, PARALLEL_ARG
-
-try:
-    from scipy.fft import next_fast_len
-except ImportError:
-    from scipy.fftpack import next_fast_len
 
 
 _logger = logging.getLogger(__name__)
@@ -107,7 +102,7 @@ def fft_correlation(in1, in2, normalize=False, real_only=False):
 
     # Calculate optimal FFT size
     complex_result = (in1.dtype.kind == 'c' or in2.dtype.kind == 'c')
-    fsize = [next_fast_len(a, not complex_result) for a in size]
+    fsize = [optimal_fft_size(a, not complex_result) for a in size]
 
     # For real-valued inputs, rfftn is ~2x faster than fftn
     if not complex_result and real_only:
--- a/hyperspy/_signals/signal2d.py
+++ b/hyperspy/_signals/signal2d.py
@@ -34,6 +34,11 @@ from hyperspy.docstrings.plot import (
     BASE_PLOT_DOCSTRING, PLOT2D_DOCSTRING, KWARGS_DOCSTRING)
 from hyperspy.docstrings.signal import SHOW_PROGRESSBAR_ARG, PARALLEL_ARG
 
+try:
+    from scipy.fft import next_fast_len
+except ImportError:
+    from scipy.fftpack import next_fast_len
+
 
 _logger = logging.getLogger(__name__)
 
@@ -80,7 +85,7 @@ def sobel_filter(im):
     return sob
 
 
-def fft_correlation(in1, in2, normalize=False):
+def fft_correlation(in1, in2, normalize=False, real_only=False):
     """Correlation of two N-dimensional arrays using FFT.
 
     Adapted from scipy's fftconvolve.
@@ -88,8 +93,12 @@ def fft_correlation(in1, in2, normalize=False):
     Parameters
     ----------
     in1, in2 : array
-    normalize: bool
-        If True performs phase correlation
+        Input arrays to convolve.
+    normalize: bool, default False
+        If True performs phase correlation.
+    real_only : bool, default False
+        If True, and in1 and in2 are real-valued inputs, uses
+        rfft instead of fft for approx. 2x speed-up.
 
     """
     s1 = np.array(in1.shape)
@@ -98,13 +107,22 @@ def fft_correlation(in1, in2, normalize=False):
 
     # Calculate optimal FFT size
     complex_result = (in1.dtype.kind == 'c' or in2.dtype.kind == 'c')
-    fsize = [sp.fft.next_fast_len(a, not complex_result) for a in size]
+    fsize = [next_fast_len(a, not complex_result) for a in size]
 
-    fprod = np.fft.fftn(in1, fsize)
-    fprod *= np.fft.fftn(in2, fsize).conjugate()
+    # For real-valued inputs, rfftn is ~2x faster than fftn
+    if not complex_result and real_only:
+        fft_f, ifft_f = np.fft.rfftn, np.fft.irfftn
+    else:
+        fft_f, ifft_f = np.fft.fftn, np.fft.ifftn
+
+    fprod = fft_f(in1, fsize)
+    fprod *= fft_f(in2, fsize).conjugate()
+
     if normalize is True:
         fprod = np.nan_to_num(fprod / np.absolute(fprod))
-    ret = np.fft.ifftn(fprod).real.copy()
+
+    ret = ifft_f(fprod).real.copy()
+
     return ret, fprod
 
 
@@ -198,8 +216,15 @@ def estimate_image_shift(ref, image, roi=None, sobel=True,
             im[:] = sp.ndimage.median_filter(im, size=3)
         if sobel is True:
             im[:] = sobel_filter(im)
+
+    # If sub-pixel alignment not being done, use faster real-valued fft
+    if sub_pixel_factor != 1:
+        real_only = False
+    else:
+        real_only = True
+
     phase_correlation, image_product = fft_correlation(
-        ref, image, normalize=normalize_corr)
+        ref, image, normalize=normalize_corr, real_only=real_only)
 
     # Estimate the shift by getting the coordinates of the maximum
     argmax = np.unravel_index(np.argmax(phase_correlation),
--- a/hyperspy/_signals/signal2d.py
+++ b/hyperspy/_signals/signal2d.py
@@ -213,10 +213,7 @@ def estimate_image_shift(ref, image, roi=None, sobel=True,
             im[:] = sobel_filter(im)
 
     # If sub-pixel alignment not being done, use faster real-valued fft
-    if sub_pixel_factor != 1:
-        real_only = False
-    else:
-        real_only = True
+    real_only = (sub_pixel_factor == 1)
 
     phase_correlation, image_product = fft_correlation(
         ref, image, normalize=normalize_corr, real_only=real_only)
--- a/hyperspy/_signals/signal2d.py
+++ b/hyperspy/_signals/signal2d.py
@@ -191,6 +191,10 @@ def estimate_image_shift(ref, image, roi=None, sobel=True,
         if hanning is True:
             im *= hanning2d(*im.shape)
         if medfilter is True:
+            # This is faster than sp.signal.med_filt,
+            # which was the previous implementation.
+            # The size is fixed at 3 to be consistent
+            # with the previous implementation.
             im[:] = sp.ndimage.median_filter(im, size=3)
         if sobel is True:
             im[:] = sobel_filter(im)
--- a/hyperspy/_signals/signal2d.py
+++ b/hyperspy/_signals/signal2d.py
@@ -95,8 +95,11 @@ def fft_correlation(in1, in2, normalize=False):
     s1 = np.array(in1.shape)
     s2 = np.array(in2.shape)
     size = s1 + s2 - 1
-    # Use 2**n-sized FFT
-    fsize = (2 ** np.ceil(np.log2(size))).astype("int")
+
+    # Calculate optimal FFT size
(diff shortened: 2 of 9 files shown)


tjof2:
I think that if the input data is real, we should use `np.fft.rfftn()` and that might be the speed-up to match scipy. I'll test and update tomorrow.

ericpre:
What are the numpy and scipy version on your machine? Do you have the numpy linked against the mkl libraries?

tjof2:
@ericpre I do - as I write I think the difference is due to scipy being "smarter" with the input than numpy, and that's all. I'll confirm tomorrow.

Either way, calculating the optimal size as I have here is the more obvious improvement according to my profiling. Will see what I can do tomorrow. 

Not sure why CI is failing btw.

Looking at https://github.com/scipy/scipy/blob/v1.4.1/scipy/signal/signaltools.py#L377 this is almost certainly the reason for the discrepancy - scipy is smarter with rfftn vs fftn for real data, while we rely only on fftn.

I'll test making this change tomorrow to see how it works.

tjof2:
@ericpre looking at https://github.com/scipy/scipy/blob/v1.4.1/scipy/signal/signaltools.py#L377 this is almost certainly the reason for the discrepancy - scipy is smarter with rfftn vs fftn for real data, while we rely only on fftn.

hyperspy/_signals/signal2d.py
```diff
@@ -95,8 +95,11 @@ def fft_correlation(in1, in2, normalize=False):
     s1 = np.array(in1.shape)
     s2 = np.array(in2.shape)
     size = s1 + s2 - 1
-    # Use 2**n-sized FFT
-    fsize = (2 ** np.ceil(np.log2(size))).astype("int")
+
+    # Calculate optimal FFT size
+    complex_result = (in1.dtype.kind == 'c' or in2.dtype.kind == 'c')
+    fsize = [sp.fft.next_fast_len(a, not complex_result) for a in size]
```
tjof2: May need to replace this with `scipy.fftpack.next_fast_len` for compatibility with scipy<1.4.0

hyperspy/_signals/signal2d.py
```diff
@@ -95,8 +95,11 @@ def fft_correlation(in1, in2, normalize=False):
     s1 = np.array(in1.shape)
     s2 = np.array(in2.shape)
     size = s1 + s2 - 1
-    # Use 2**n-sized FFT
-    fsize = (2 ** np.ceil(np.log2(size))).astype("int")
+
+    # Calculate optimal FFT size
+    complex_result = (in1.dtype.kind == 'c' or in2.dtype.kind == 'c')
```
tjof2, line 104: Investigate using rfftn for real-valued inputs.

ericpre:
> @ericpre looking at https://github.com/scipy/scipy/blob/v1.4.1/scipy/signal/signaltools.py#L377 this is almost certainly the reason for the discrepancy - scipy is smarter with rfftn vs fftn for real data, while we rely only on fftn.

Yes, this sounds sensible: this recent benchmark (https://github.com/project-gemmi/benchmarking-fft/) show the same thing - there is a consistent factor of 2 between real-to-complex and complex-to-complex.

tjof2:
Awesome thanks! I'll get it tidied up if it works then :-)

tjof2:
OK I've tested this now - it looks like we can use `rfft()` only when *not* doing sub-pixel alignment, otherwise the tests don't pass. So I've added a check for this to try and use the faster version if we can, when `sub_pixel_factor=1`.

```python
>>> %timeit s.estimate_shift2D(show_progressbar=False)
# Before
4.80 s ± 127 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
# After (using fftn())
1.92 s ± 40.7 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
# After (using rfftn())
1.44 s ± 40.2 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
```

tjof2:
OK I've tested this now, `rfft` is noticeably faster, but it looks like the tests only pass if we use `rfft()` when *not* doing sub-pixel alignment. 

So I've added a check for this to use the faster version if we can, when `sub_pixel_factor=1`.

```python
>>> %timeit s.estimate_shift2D(show_progressbar=False)
# Before
4.80 s ± 127 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
# After (using fftn())
1.92 s ± 40.7 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
# After (using rfftn())
1.44 s ± 40.2 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
```

Here is an example test error for sub-pixel alignment if we were to use `rfftn()`. As you can see, there are some differences in the shifts when using the faster function. Perhaps these differences are acceptable?

```python
tests/signal/test_2D_tools.py:76: AssertionError
______________ TestSubPixelAlign.test_estimate_subpix[True-stat] _______________

self = <hyperspy.tests.signal.test_2D_tools.TestSubPixelAlign object at 0x7f3be0035a10>
normalize_corr = True, reference = 'stat'

    @pytest.mark.parametrize(("normalize_corr", "reference"),
                             _generate_parameters())
    def test_estimate_subpix(self, normalize_corr, reference):
        s = self.signal
        shifts = s.estimate_shift2D(sub_pixel_factor=200,
                                    normalize_corr=normalize_corr)
        np.testing.assert_allclose(shifts, self.shifts, rtol=0.2, atol=0.2,
>                                  verbose=True)
E       AssertionError: 
E       Not equal to tolerance rtol=0.2, atol=0.2
E       
E       Mismatched elements: 2 / 20 (10%)
E       Max absolute difference: 0.875
E       Max relative difference: 0.41079812
E        x: array([[-0.   , -0.   ],
E              [ 4.07 ,  1.255],
E              [ 1.93 ,  3.255],...
E        y: array([[ 0.  ,  0.  ],
E              [ 4.3 ,  2.13],
E              [ 1.65,  3.58],...
```

To be fair, the [tests are created](https://github.com/hyperspy/hyperspy/blob/RELEASE_next_minor/hyperspy/tests/signal/test_2D_tools.py#L49) by applying shifts using `np.fft.fftn()`, so perhaps its no surprise that it matches fine when we undo it with `fftn()` but doesn't match so closely when we use `rfftn()`. Maybe `scipy.ndimage.interpolation.shift` in the tests would avoid any such bias.

@dnjohnstone any view on this behaviour and which is "better"?

ericpre:
> To be fair, the [tests are created](https://github.com/hyperspy/hyperspy/blob/RELEASE_next_minor/hyperspy/tests/signal/test_2D_tools.py#L49) by applying shifts using `np.fft.fftn()`, so perhaps its no surprise that it matches fine when we undo it with `fftn()` but doesn't match so closely when we use `rfftn()`.

Indeed, this is a fairly bias ground truth...

hyperspy/_signals/signal2d.py
```diff
@@ -34,6 +34,11 @@
     BASE_PLOT_DOCSTRING, PLOT2D_DOCSTRING, KWARGS_DOCSTRING)
 from hyperspy.docstrings.signal import SHOW_PROGRESSBAR_ARG, PARALLEL_ARG
 
+try:
+    from scipy.fft import next_fast_len
+except ImportError:
+    from scipy.fftpack import next_fast_len
```
tjof2: The functions aren't actually identical (see docs). Should adjust code for this. In particular, `scipy.fft.next_fast_len()` takes a second boolean argument for real-valued input, while the FFTPACK version does not.

tjof2:
I went through and looked at other places where we used the conservative power-of-two for FFT sizes, and replaced them with a wrapper function to `next_fast_len()` in scipy.

CI failures are unrelated - it's only broken on MacOS

hyperspy/_signals/eels.py
```diff
@@ -1220,10 +1219,10 @@ def kramers_kronig_analysis(self,
             # Kramers Kronig Transform:
             # We calculate KKT(Im(-1/epsilon))=1+Re(1/epsilon) with FFT
             # Follows: D W Johnson 1975 J. Phys. A: Math. Gen. 8 490
-            # Use a size that is a power of two to speed up the fft and
+            # Use an optimal FFT size to speed up the calculation, and
             # make it double the closest upper value to workaround the
             # wrap-around problem.
-            esize = 2 * closest_power_of_two(axis.size)
+            esize = 2 * optimal_fft_size(axis.size)
```
tjof2: Maybe this should be `optimal_fft_size(2 * axis.size)`

ericpre:
Would it make sense to also use rfft (when suitable) in https://github.com/hyperspy/hyperspy/blob/4be888574e5fc3df0725beb198ea7a3a79e89d0a/hyperspy/signal.py#L3412

On a similar topic, one thing I found not great is that the `ifft` method of `BaseSignal` returns the real part only while we should be able to choose between returning real or a both...

tjof2:
@ericpre I think you're probably right, but I'm not too sure about potential impact it could have generally - at least inside `estimate_shift_2D` I'm pretty confident on what the change does and that the tests cover the change in an acceptable manner, but I'm worried it might break things for other users in the general case.

ericpre:
Yes, good point, maybe add an option to choose one of the two?

tjof2:
@ericpre See latest commit.

tjof2:
> Yes, good point, maybe add an option to choose one of the two?

@ericpre I added an option to return either just the real part from the IFFT (default behaviour) vs. all.

ericpre:
Great thanks, this looks good to me. Is there any reason to revert the commit adding the option to use the `rfft` in `fft`?

tjof2:
@ericpre ~I'll unrevert it, then I think this is ready for review.~ Done.

tjof2:
CI failures are Mac - #2361

ericpre:
We have the same failure on windows python 3.8 as in #2355 which was "fixed" in https://github.com/hyperspy/hyperspy/pull/2355/commits/ae44538875af92dadb9bfb4eb7e3405ba5c6c721... and it makes appveyor hang on that build for 1h :/.

I guess the memory requirement have been increased in https://github.com/hyperspy/hyperspy/pull/2358/commits/995b4dd72082699b0c859bdf2c29f610b05a9c26 and the failing build, which was already borderline is now failing. To be more future proof, maybe we should have pytest fixture with a module scope, which will run a garbage collection instead letting python managing it on its own, which surprisingly doesn't seem to work....

tjof2:
How do we fix it here?

ericpre:
According to https://docs.pytest.org/en/latest/fixture.html#fixture-finalization-executing-teardown-code, I think that adding the following in `conftest.py` should run a garbarge collection at the end of each module:

```python
import gc

# run a gabarge collection at the end of each test module to avoid trouble in CI
@pytest.fixture(scope="module")
def garbage_collection():
    yield
    gc.collect()
```

tjof2:
Ok let's see if that runs

ericpre:
It doesn't seem to help... For the recrod, the hanging test is `test_samfire.py::TestSamfireWorker::test_main_result` and there is no error message.

Is it worth trying the same workaround as in #2355 and see it this helps? If so, we can merge this PR and try to sort out this issue in another PR.

tjof2:
@ericpre do you mean to `hyperspy/tests/signal/test_remove_background.py` as in https://github.com/hyperspy/hyperspy/pull/2355/commits/ae44538875af92dadb9bfb4eb7e3405ba5c6c721?

tjof2:
>Is it worth trying the same workaround as in #2355 and see it this helps? If so, we can merge this PR and try to sort out this issue in another PR.

Assuming you meant https://github.com/hyperspy/hyperspy/commit/ae44538875af92dadb9bfb4eb7e3405ba5c6c721, I've made that change and pushed it.

ericpre:
> @ericpre do you mean to `hyperspy/tests/signal/test_remove_background.py` as in [ae44538](https://github.com/hyperspy/hyperspy/commit/ae44538875af92dadb9bfb4eb7e3405ba5c6c721)?

This or reduce the memory usage of the tests (in case this PR increase a bit the memory usage which could triggers the issue with the rest of the test suite).

tjof2:
Ok I see what you mean - I also reverted https://github.com/hyperspy/hyperspy/commit/995b4dd72082699b0c859bdf2c29f610b05a9c26 to see if it helps.

tjof2:
@ericpre looks like it still hangs

ericpre:
Yes, it is annoying and difficult to debug... I have started a build the `RELEASE_next_minor` (https://ci.appveyor.com/project/ericpre/hyperspy/build/job/8goxcvrhoqsyeysv) to check that this PR is triggering the issue (yesterday it was the case, I think).

tjof2:
So it hangs somewhere around https://github.com/hyperspy/hyperspy/blob/RELEASE_next_minor/hyperspy/tests/samfire/test_samfire.py#L358 ? I don't understand why at all

ericpre:
`test_samfire.py` does use a lot of memory, from using the `pytest-memprof` plugin:

```
=========================================================================== memory consumption estimates ===========================================================================
/home/eric/Dev/hyperspy/hyperspy/tests/samfire::test_samfire.py::TestSamfireMain::()::test_multiprocessed                                                           - 1695.2 MB
/home/eric/Dev/hyperspy/hyperspy/tests/samfire::test_samfire.py::TestSamfireEmpty::()::test_setup                                                                   - 354.3 MB
/home/eric/Dev/hyperspy/hyperspy/tests/signal::test_1D_tools.py::TestFindPeaks1D::()::test_single_spectrum                                                          - 250.7 MB
/home/eric/Dev/hyperspy/hyperspy/tests/model::test_model.py::TestModel2D::()::test_fitting                                                                          - 223.2 MB
/home/eric/Dev/hyperspy/hyperspy/tests/drawing::test_plot_signal1d.py::TestPlotSpectra::()::test_plot_spectra[mosaic2]                                              - 161.3 MB
/home/eric/Dev/hyperspy/hyperspy/tests/drawing::test_plot_signal.py::test_plot_sig_nav[1-2-sig-complex_imag]                                                        - 149.6 MB
/home/eric/Dev/hyperspy/hyperspy/tests/signal::test_hologram_image.py::test_statistics[True-False-False-fourier]                                                    - 136.7 MB
/home/eric/Dev/hyperspy/hyperspy/tests/signal::test_apodization.py::test_apodization[True-hann-False]                                                               - 110.5 MB
/home/eric/Dev/hyperspy/hyperspy/tests/drawing::test_plot_signal.py::test_plot_sig_nav[0-2-sig-complex_real]                                                        - 103.3 MB
/home/eric/Dev/hyperspy/hyperspy/tests/signal::test_hologram_image.py::test_statistics[False-False-True-fourier]                                                    - 102.4 MB
/home/eric/Dev/hyperspy/hyperspy/tests/signal::test_2D_tools.py::TestSubPixelAlign::()::test_estimate_subpix[True-current]                                          - 96.2 MB
/home/eric/Dev/hyperspy/hyperspy/tests/drawing::test_plot_markers.py::test_iterate_markers                                                                          - 89.0 MB
/home/eric/Dev/hyperspy/hyperspy/tests/signal::test_eds_tem.py::Test_quantification::()::test_lazy_quant_zeta                                                       - 86.1 MB
/home/eric/Dev/hyperspy/hyperspy/tests/drawing::test_plot_signal_tools.py::test_plot_contrast_editor_norm[symlog]                                                   - 83.7 MB
/home/eric/Dev/hyperspy/hyperspy/tests/drawing::test_plot_signal.py::test_plot_sig_nav[2-2-nav-complex_imag]                                                        - 78.5 MB
/home/eric/Dev/hyperspy/hyperspy/tests/drawing::test_plot_signal.py::test_plot_sig_nav[1-2-sig-complex_real]                                                        - 78.3 MB
/home/eric/Dev/hyperspy/hyperspy/tests/drawing::test_plot_signal.py::test_plot_sig_nav[2-2-sig-complex_real]                                                        - 77.3 MB
/home/eric/Dev/hyperspy/hyperspy/tests/signal::test_hologram_image.py::test_statistics[True-True-True-statistical]                                                  - 75.7 MB
/home/eric/Dev/hyperspy/hyperspy/tests/drawing::test_figure.py::TestCloseFigure::()::test_close_figure[2-1-spectrum]                                                - 73.7 MB
/home/eric/Dev/hyperspy/hyperspy/tests/signal::test_1D_tools.py::TestFindPeaks1D::()::test_n_peaks                                                                  - 73.5 MB
/home/eric/Dev/hyperspy/hyperspy/tests/drawing::test_plot_markers.py::test_plot_line_markers                                                                        - 71.6 MB
/home/eric/Dev/hyperspy/hyperspy/tests/component::test_gaussian.py::test_estimate_parameters_binned[False-True-True]                                                - 68.1 MB

```

tjof2:
Ugh, ok, I'll work through it tomorrow. Any ideas how much I have to reduce memory usage by?

ericpre:
It seems that on my laptop, it is fluctuating quite a bit... from 0.5Go to 1.5Go, adding a garbage collection make it more stable.
I have pushed a branch to check appveyor and the situation is not as bad: https://ci.appveyor.com/project/ericpre/hyperspy/history but it should be possible to improve it.

tjof2:
Thanks! Let me know if you find anything we could add here, otherwise I'll take a look in depth tomorrow

tjof2:
@ericpre, you report `test_samfire.py::TestSamfireMain::()::test_multiprocessed` uses 1695.2 MB, I only see it using 210 MB on my machine. I'm using Python 3.7 - are you using 3.8? That might be it.

I see that it's also marked as `xfail`, but we allow it to pass?

I've pushed a change that more than halves the size of the test, still passes locally, let's see - I have a few other ideas too.

tjof2:
To summarise: I don't really know why my changes have affected the memory usage in 3.8, but here are the steps I've taken:

1. Made all `SamfirePool`s have `workers=1`
2. Clarified that `test_multiprocessed()` uses `multiprocessing` rather than `ipyparallel`
3. Added explicit `teardown_method()` to do garbage collection and try and delete the `SamfirePool`s where possible.
4. Shrunk the dataset size from `15, 15, 1024` to `15, 15, 400`.
5. Changed the dataset type to float32.
6. Investigated the reason why we xfail on  `test_multiprocessed()`. This one worries me a bit. If I run `pytest ./hyperspy/hyperspy/samfire/test*` it always passes, but if I do `pytest .` it fails half the time.

Point (1) seems to have helped it get past the first sticking point: [1.0.4485](https://ci.appveyor.com/project/hyperspy/hyperspy/builds/32191468/job/tcm2e2iwehof7xy6), but I'll let AppVeyor catch up and test the other changes I've made to see if it gets all the way through.

My worry is that is actually a  much deeper issue with parallel pools, since the correct way to use them is this, and we don't do that.
```
with multiprocessing.Pool() as pool:
    pool.map(...)
```
So that the context manager can properly collect and clean up. An interesting Python 3.8 issue where this is discussed is [here (39360)](https://bugs.python.org/issue39360).

ericpre:
> @ericpre, you report `test_samfire.py::TestSamfireMain::()::test_multiprocessed` uses 1695.2 MB, I only see it using 210 MB on my machine. I'm using Python 3.7 - are you using 3.8? That might be it.

It is fluctuating a lot and the one I posted here was possibly the first I ran (which was quite surprising) and it is one of the worse in term of memory usage.

tjof2:
Yeah I'm looking at it still - might have a fix.

tjof2:
@ericpre why does `from hyperspy.signals import BaseSignal` cause a 270 MB increase in memory on my machine?

tjof2:
FYI it looks like switching the dtype to float16 and improving how we close the `multiprocessing.Pool()` instances may have fixed AppVeyor - [see here](https://ci.appveyor.com/project/hyperspy/hyperspy/builds/32192927/job/dyv0935acxspf6sf).

tjof2:
This is now ready for review.

FWIW I believe the appveyor issues with samfire have nothing to do with the original changes in the PR to the FFTs.

I actually think there is something inherently wrong with `multiprocessing` as used in samfire and the test suite. If I install `pytest-random` and randomise the order of the tests, I get random pass/fail for the `test_multiprocessed`, and the estimated memory usage varies drastically with the test order as well.

hyperspy/_signals/eels.py
ericpre: I guess `optimal_fft_size(2 * axis.size)` should more sense and give slightly smaller number. I think you reverted a commit here when trying to debug the samfire, so I am not sure if you want to reintroduce this change or not.

ericpre (commented):
This looks very good to me! See comment about putting it back a previous change.
It maybe worth opening an issue about your observation on samfire to keep a record.

ericpre:
> @ericpre why does `from hyperspy.signals import BaseSignal` cause a 270 MB increase in memory on my machine?

As it is currently, BaseSignal will import almost everything, so this doesn't sound unexpected. However, if this is normal behaviour, this is a different question...

tjof2:
Thanks @ericpre  - I've made the change you requested so I think this is good. I'll open a separate issue(s) about samfire and the basesignal import.

ericpre:
Great thanks! To be on the safe side, I will merge once appveyor have gone through the backlog... but at last now it doesn't hang for 1h on a single build.

tjof2:
Sounds sensible, I'm happy with that!

> but at last now it doesn't hang for 1h on a single build.

Indeed, the last three commits that have run have worked fine.

tjof2:
Passed on all of the last 5 commits! Success :)

PR #2361, opened 2020-04-11
Fix travis

ericpre:
### Progress of the PR
- [x] Use wget instead of curl, which was not working since a couple of days,
- [x] install last version of tifffile from conda-forge instead of pypi to avoid mixing package source.
- [x] ready for review.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,7 +5,7 @@ env:
   global:
     - ENV=conda
     - MINIMAL_ENV=false
-    - DEPS="numpy ipython scipy matplotlib traits natsort requests tqdm sympy dill h5py python-dateutil ipyparallel dask scikit-image pint numexpr statsmodels sparse imageio pyyaml PTable tifffile==2020.2.16"
+    - DEPS="numpy ipython scipy matplotlib traits natsort requests tqdm sympy dill h5py python-dateutil ipyparallel dask scikit-image pint numexpr statsmodels sparse imageio pyyaml PTable tifffile>=2019.7.26"
     - DEPS_OPTIONAL="scikit-learn mrcz numba cython"
     - TEST_DEPS="pytest pytest-cov codecov pytest-mpl wheel"
     - MPLBACKEND="agg"
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,7 +5,7 @@ env:
   global:
     - ENV=conda
     - MINIMAL_ENV=false
-    - DEPS="numpy ipython scipy matplotlib traits natsort requests tqdm sympy dill h5py python-dateutil ipyparallel dask scikit-image pint numexpr statsmodels sparse imageio pyyaml PTable tifffile"
+    - DEPS="numpy ipython scipy matplotlib traits natsort requests tqdm sympy dill h5py python-dateutil ipyparallel dask scikit-image pint numexpr statsmodels sparse imageio pyyaml PTable tifffile==2020.2.16"
     - DEPS_OPTIONAL="scikit-learn mrcz numba cython"
     - TEST_DEPS="pytest pytest-cov codecov pytest-mpl wheel"
     - MPLBACKEND="agg"
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -62,7 +62,10 @@ install:
 
   # Install the dependencies of hyperspy.
   - ps: Add-AppveyorMessage "Installing conda packages..."
-  - if "%PLATFORM%"=="x64" set DEPS=%DEPS% "mrcz tifffile^>=2019.7.26"
+  # to avoid mising wheels and conda package, we pin tifffile to a supported
+  # version from conda-forge because the versions currently in defaults are too
+  # old which will make pip install a more recent version from pypi
+  - if "%PLATFORM%"=="x64" set DEPS=%DEPS% "mrcz tifffile==2019.7.26"
   - "conda install -y %DEPS% %TEST_DEPS%"
 
   - ps: Add-AppveyorMessage "Installing hyperspy..."
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -65,7 +65,7 @@ install:
   # to avoid mising wheels and conda package, we pin tifffile to a supported
   # version from conda-forge because the versions currently in defaults are too
   # old which will make pip install a more recent version from pypi
-  - if "%PLATFORM%"=="x64" set DEPS=%DEPS% "mrcz tifffile==2019.7.26"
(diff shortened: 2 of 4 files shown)


.travis.yml
```diff
@@ -5,7 +5,7 @@ env:
   global:
     - ENV=conda
     - MINIMAL_ENV=false
-    - DEPS="numpy ipython scipy matplotlib traits natsort requests tqdm sympy dill h5py python-dateutil ipyparallel dask scikit-image pint numexpr statsmodels sparse imageio pyyaml PTable tifffile"
+    - DEPS="numpy ipython scipy matplotlib traits natsort requests tqdm sympy dill h5py python-dateutil ipyparallel dask scikit-image pint numexpr statsmodels sparse imageio pyyaml PTable tifffile==2020.2.16"
```
dnjohnstone: Can we have >= for tiffile? I've had problems where some large facilities really don't want to install anything on their platforms as part of large python environments if it pins the version of anything that may cause incompatibility issues. If we don't pin it here then we'll know if we need to change it in future.

appveyor.yml
```diff
@@ -62,11 +62,11 @@ install:
 
   # Install the dependencies of hyperspy.
   - ps: Add-AppveyorMessage "Installing conda packages..."
-  - if "%PLATFORM%"=="x64" set DEPS=%DEPS% mrcz tifffile
+  - if "%PLATFORM%"=="x64" set DEPS=%DEPS% mrcz tifffile==2020.2.16
   - "conda install -y %DEPS% %TEST_DEPS%"
```
dnjohnstone, line 69: As above - on another matter - in pyxem we've removed appveyor completely by using choco to test on Windows using Travis, perhaps that would be good for hyperspy too?

dnjohnstone (commented):
A couple of comments - but certainly solves the immediate issue with mac on travis so good to get in as soon as...

.travis.yml
ericpre: This is pinned only for CI and I did this because I don't know a syntax which would work and be generic enough not to make the CI script more complicated for little reward. The idea is to make sure not to mixed pypi and conda package.

>  I've had problems where some large facilities really don't want to install anything on their platforms as part of large python environments if it pins the version of anything that may cause incompatibility issues.

It looks like defying the concept of python environment?! In any case, this PR doesn't change hyperspy installation requirement.

dnjohnstone: I know it doesn't change the hyperspy installation requirement and that's part of my issue. If the CI is pinned like this but the install requirement is not - then we won't know that a future update of tiffile has broken the build (because it won't break the pinned CI) until a user reports it. So I think that the CI and install requirement should at least be the same and for me preferably be >=

appveyor.yml
ericpre: Indeed, appveyor runs only one build at a time, which is quite slow... and travis also offers limited capacity... I have setup a CI repo last year to start using azure pipeline with the idea that these scripts would be easily reusable and we can used them from various repos. It works fairly well https://github.com/ericpre/CI, I have been using it to build the bundle https://github.com/ericpre/hyperspy-bundle/tree/test_release_cross_platform.

I think, fairly soon, we need to switch to azure pipeline and build with pip only on travis, so it would simplify the CI scripts.

.travis.yml
ericpre: There is no simple syntax to do this with the current setup and as mentioned below, I would prefer to switch to use azure pipeline to simplify everything and increase CI capacity.

francisco-dlp: Probably I am missing something, but, just in case, why not ">=2020.2.16" as suggested by @dnjohnstone ? That seems to be genuine ``conda`` syntax... Also I didn't get why we should not add the same constraint to ``setup.py``

ericpre: No I don't think that this syntax works with conda or we would need to change how CI is setup, which I don't want to do in this PR because I think we should most CI to azure pipeline. Currently, tifffile is installed from pypi because `defaults` is higher priority than `conda-forge` and `defaults` version is not recent enough, therefore it is updated by pip.
The idea here is to use this (temporary) workaround to avoid mixing wheels and conda packages.

francisco-dlp: I've just tested it and it works with ``conda``, or at least it does the right thing and doesn't through any errors:

```bash
conda install -c conda-forge "tifffile>=2020.2.16"
```

dnjohnstone: We've been using >= in orix for a while without any obvious issue.

https://github.com/pyxem/orix/blob/master/.travis.yml

ericpre: Great thanks, I may have tried with the first quotation marks in the wrong place! Giving a try now.

ericpre: Appveyor is not happy about it:
https://ci.appveyor.com/project/ericpre/hyperspy/builds/32194755/job/0a7ck2xagxtypcm7

```
if "%PLATFORM%"=="x64" set DEPS=%DEPS% "mrcz tifffile>=2019.7.26"
conda install -y %DEPS% %TEST_DEPS%
CondaValueError: invalid package specification: mrcz tifffile>=2019.7.26
Command exited with code 1

```


Do you have an example where it works on window?

ericpre:
@francisco-dlp: shall we rearrange CI to have something more efficient?

I would suggest to do the following:
- set up Azure Pipeline with mac/win/linux amtrix using conda
- keep pip only testing and building doc on travis/linux

This would simplify the CI scripts increase capacity using azure pipeline and make CI maintenance easier.

.travis.yml
francisco-dlp: I think that the issue is that in Windows those characters must be escaped in Windows, maybe something like this would do: ``^>=2019.7.26``?

ericpre: I thought that the double quotation mark would take of it. Giving a try now!

ericpre: humm, no luck...

```
if "%PLATFORM%"=="x64" set DEPS=%DEPS% "mrcz tifffile^>=2019.7.26"
conda install -y %DEPS% %TEST_DEPS%
CondaValueError: invalid package specification: mrcz tifffile^^^^>=2019.7.26
Command exited with code 1
```

francisco-dlp: I have not clue then, maybe a conda bug that affects only Windows. Anyway, to move on I suggest we pin for Windows and we keep ``>= `` in travis. Also adding some comments to explain why we've done all that will be helpful.

ericpre: Agree, this is done.

francisco-dlp:
@ericpre, about azure, it sounds like a good plan!

ericpre:
It may be more simple to close this one in favour of #2363.

ericpre:
Fixed in #2363.

PR #2368, opened 2020-04-20
Absolute thickness estimation by EELS log-ratio method

francisco-dlp:
### Description of the change

Implement absolute thickness estimation by the log-ratio method, including MFP estimation and angular correction. 

### Progress of the PR
- [x] add tests,
- [x] ready for review.


hyperspy/misc/eels/electron_inelastic_mean_free_path.py
```diff
@@ -0,0 +1,50 @@
...
+    
+    Notes:
+    ------
+    For details see Egerton, R. Electron Energy-Loss Spectroscopy in the Electron
+    Microscope.  Springer-Verlag, 2011.
+    """
+    theta_C = 20 # mrad
+    A = alpha ** 2 + beta ** 2 + 2 * _theta_E(density, beam_energy) ** 2 + np.abs(alpha ** 2 - beta ** 2) 
+    B = alpha ** 2 + beta ** 2 + 2 * theta_C ** 2 + np.abs(alpha ** 2 - beta ** 2)
+    return np.log(theta_C ** 2 / _theta_E(density, beam_energy) ** 2) / np.log(A * theta_C ** 2 / B / _theta_E(density, beam_energy) ** 2)
```
ericpre, line 122: Is this `np.log(theta_C ** 2 / _theta_E(density, beam_energy) ** 2` correct?

hyperspy/misc/eels/electron_inelastic_mean_free_path.py
```diff
@@ -0,0 +1,50 @@
...
+    density : float
+        Material density in g/cm**3
+    beam_energy : float
+        Electron beam energy in keV
+    alpha, beta : float
+        Convergence and collection angles in mrad.
+    
+    Notes:
+    ------
+    For details see Egerton, R. Electron Energy-Loss Spectroscopy in the Electron
```
ericpre: Maybe add Iakoubovskii et al. (http://onlinelibrary.wiley.com/doi/10.1002/jemt.20597)?

doc/user_guide/eels.rst
```diff
@@ -36,9 +36,17 @@ when saving in the hspy format (HyperSpy's HDF5 specification).
...
+.. versionadded:: 1.6
+    Option to compute the absolute thickness, including the angular corrections
+    and mean free path estimation.
+
 The :py:meth:`~._signals.eels.EELSSpectrum_mixin.estimate_thickness` can
 estimate the thickness from a low-loss EELS spectrum using the Log-Ratio
-method.
+method. If the beam energy, collection angle, convergence angle and sample
+density are known the absolute thickness is estimated, including the angular
+corrections, following the equations by Iakoubovskii et al. as found in
```
ericpre: I would add Iakoubovskii et al. (http://onlinelibrary.wiley.com/doi/10.1002/jemt.20597) too.

hyperspy/misc/eels/electron_inelastic_mean_free_path.py
francisco-dlp: I think so. The equations are not exactly as in the paper because I have decoupled the mean free path estimation from the angular correction. In this way the user can provide the iMFP independently.

hyperspy/misc/eels/electron_inelastic_mean_free_path.py
francisco-dlp: Done. Actually, in an earlier version I was citing their paper but I removed it when I added Egerton's relativistic correction not to mislead users 715d05cbc1cec2994fa232473e06e3ec5d765393 . However, thinking about it again following your comment, I've put it back and I have extended the UG documentation. Thanks!

hyperspy/misc/eels/electron_inelastic_mean_free_path.py
```diff
@@ -0,0 +1,62 @@
+import numpy as np
+
+def _F(beam_energy):
+    return (1 + beam_energy / 1022) / (1 + beam_energy / 511) ** 2
+
+def _theta_E(density, beam_energy):
+    gamma = 1 + beam_energy/511 # keV
+    return 5.5 * density ** 0.3 / (gamma * _F(beam_energy) * beam_energy)
+
+def iMFP(density, beam_energy):
```
ericpre: Shall we add this function and `iMFP_angular_correction` to the API? For example, it would be convenient to be able to do: 

```python
import hyperspy.api as hs
hs.eels.inelastic_mean_free_path(density, beam_energy)
```

hyperspy/misc/eels/electron_inelastic_mean_free_path.py
ericpre: OK, I will go through again (properly this time) with a piece of paper to confirm!

hyperspy/misc/eels/electron_inelastic_mean_free_path.py
francisco-dlp: I suggest we don't in order to stick to HyperSpy's data analysis mission. Once we split HyperSpy we can break the EELS features into 2 packages: a HyperSpy extension for EELS analysis and a EELS package independent from HyperSpy that could have functions like those ones. Does it sound good?

francisco-dlp:
I've changed this to WIP because, after testing it, I've found out that Ray's relativistic correction gives a poorer estimation than the original equation overall. I'll investigate this further and eventually remove the correction.

hyperspy/misc/eels/electron_inelastic_mean_free_path.py
ericpre: Checking that we are talking about the same equations: these are Eq. 5.5 and Eq 5.5(a) from "Egerton 3rd edition (2011) p. 297?
If so, I would expect the numerator to be:
```python
200 * 5.5 / ( 11 *  _theta_E(density, beam_energy) ** 2) )
```

I would suggest to reference with the equation number (and the page number as a convenience) in the docstring, since we are referencing a specific edition of Egerton's book.

hyperspy/misc/eels/electron_inelastic_mean_free_path.py
ericpre: I am fine either way.

hyperspy/misc/eels/electron_inelastic_mean_free_path.py
francisco-dlp: I've followed Iakoubovskii's equation instead, that is the same but looks different. I've now removed the citation to Egerton and pointed to the right equation in the paper to avoid confusion.

francisco-dlp:

The iMFP is now plain Iakoubovskii.

I've added an extra method to estimate the iMFP, `iMFP_TPP2M`. That's what I've used to decide if to keep or remove the gamma correction. (I've gone for removing it because the match with the gamma correction was worst for the 3 materials that I've tested it with.) However, I didn't add it to the API because it depends on some extra parameters and, therefore, it is less convenient the that Iakoubovskii's equation, albeit probably more accurate. Once we split hyperspy we can add it to the API as discussed above if that's OK.

PR #2363, opened 2020-04-16
Add azure pipelines and remove appveyor

ericpre:
The idea of the https://github.com/ericpre/ci-scripts repo is to share and centralise some of the CI configuration so that CI are easier to maintain and update.
Consumer of theses ci-scripts could use the upstream repo or their own fork. The later would give them more stability and also the possibility to update their fork very easily.
Longer term, we shall make the generation of release note, updating documentation etc. more automated.

If we are happy with this approach, the next steps are:
- set up azure pipeline for the hyperspy organisation (some instruction on https://github.com/ericpre/ci-scripts) 
- move https://github.com/ericpre/ci-scripts to https://github.com/hyperspy/ci-scripts

### Progress of the PR
- [x] remove appveyor,
- [x] most of the CI scripts are moved to a separate github repository (currently https://github.com/ericpre/ci-scripts),
- [x] add azure pipelines (linux, windows and macos in a conda environment),
- [x] update travis to build on pip only,
- [x] upload folder containing the output of the image comparison for matplotlib testing from azure to debug failure.
- [ ] update dev guide regarding accessing image comparison and setting up CI account.
- [x] ready for review.

For the azure builds of this branch: https://dev.azure.com/ericpre/hyperspy/_build?definitionId=10&_a=summary

PR #1852, opened 2018-03-03
Fix issue random failure async mrcz

ericpre:
Fix the asynchronous save/load cycle of the mrcz tests, which can happen occasionally in appveyor: add an additional sleep as recommended by @robbmcleod.
It has been tested over 6 appveyor build, which ran all fine while before it was failing every other build.

PR #2369, opened 2020-04-20
Doc test suite

ericpre:
## Progress of the PR
- [x] Explain in the dev guide how to access results of image comparison on azure pipeline,
- [x] skip a mrcz test on windows (mentioned in https://github.com/hyperspy/hyperspy/pull/2363#issuecomment-616623994) which seems to occur more often on azure,
- [x] backport fix of samfire failure (fixed in #2358 in RnM),
- [x] ready for review.


doc/dev_guide/testing.rst
```diff
@@ -112,23 +112,23 @@ If you need to add or change some plots, follow the workflow below:
     4. Use ``git add`` to put the new file in the git repository.
 
 When the plotting tests are failling, it is possible to download the figure
```
tjof2: Typo in failing

hyperspy/tests/io/test_mrcz.py
```diff
@@ -177,6 +178,8 @@ def test_MRC(self, dtype, compressor, clevel, lazy):
         print("MRCZ test ({}, {}, {}, lazy:{}) finished in {} s".format(
             dtype, compressor, clevel, lazy, perf_counter() - t_start))
 
+    @pytest.mark.xfail(sys.platform.startswith("win"),
```
tjof2: Why use xfail to skip instead of skip/skipif? You don't _expect_ it to fail, rather, it just often does.

tjof2: I guess, change the description to not say "skip" - since this test will in fact run, but not successfully.

ericpre: OK, in this case I was unsure which one to use and I changed it from `skipif` to `xfail`...

Maybe @robbmcleod would have an idea why this is happening significantly often (about 25%?)  since we moved from appveyor to azure pipeline, yesterday?

tjof2: @ericpre  is it the test that fails with this?

```
>       [version, versionlz, flags, typesize] = np.fromfile(filehandle, dtype='uint8', count=4)
E       ValueError: not enough values to unpack (expected 4, got 0)

C:\Miniconda\envs\test_env\lib\site-packages\mrcz\ioMRC.py:449: ValueError```

According to [`np.fromfile`](https://numpy.org/doc/stable/reference/generated/numpy.fromfile.html?highlight=fromfile#numpy.fromfile)

> Do not rely on the combination of tofile and fromfile for data storage, as the binary files generated are not platform independent. In particular, no byte-order or data-type information is saved. Data can be stored in the platform independent .npy format using save and load instead.

Comment (tjof2) [reply hyperspy/tests/io/test_mrcz.py]: Wait, what is this `do_async` test trying to achieve?

Comment (tjof2) [reply hyperspy/tests/io/test_mrcz.py]:
```python
if do_async:
    # Poll file on disk since we don't return the
    # concurrent.futures.Future
    t_stop = perf_counter() + MAX_ASYNC_TIME
    sleep(0.005)
    while(perf_counter() < t_stop):
        try:
            fh = open(mrcName, 'a')
            fh.close()
            break
        except IOError:
            sleep(0.001)
    print("Time to save file: {} s".format(
        perf_counter() - (t_stop - MAX_ASYNC_TIME)))
    sleep(0.005)
```

tjof2: This seems like a really odd "test" anyway, unless I'm missing something.

tjof2: @ericpre  I would wager that probably changing some of those `sleep()` periods will make it "work" better on azure...

ericpre: Yes, this sounds very sensible, there is already some precedent on this: https://github.com/hyperspy/hyperspy/pull/1852.

I increased a bit one `sleep` call and so far 2 build runs fine, I will run some build by restarting manually on my azure: https://dev.azure.com/ericpre/hyperspy/_build?definitionId=10&_a=summary and see it is good enough.

ericpre:
Out of 10 runs of the test suite on azure, there were no failure of the mrcz test, so this look promising!

hyperspy/tests/io/test_mrcz.py
robbmcleod: The asynchronous read/write tests are probably not required. I recall at the time you wanted 100 % test coverage, and `mrcz.asyncWriteMRC` doesn't return the future. If you wanted to do it properly without the `sleep` call, probably polling `psutil.Process().open_files()` would work. I don't recall if `psutil` is already a dependency or not.

tjof2:
I hope so! #2352 keeps failing due to this.

hyperspy/tests/io/test_mrcz.py
ericpre: Thanks @robbmcleod for the clarification and the pointer!

ericpre:
The number of failure on other branches indicates that the sleep workaround is working fine. @francisco-dlp, are you happy with this?

tjof2 (approved):

francisco-dlp:
I could suggest some language fixes, but I may just translate the document to Spanglish by doing so. @tjof2, what do you think?

@ericpre, would it be possible to keep the built documentation as an artifact for inspection?

ericpre:
> I could suggest some language fixes, but I may just translate the document to Spanglish by doing so. @tjof2, what do you think?

Usually, I trust your language suggestion, so please go ahead!

@ericpre, would it be possible to keep the built documentation as an artifact for inspection?

Indeed, this would be good, travis doesn't host any artifact itself, so the alternative are:
- it needs to put somewhere else and I can't think of anything convenient
- use azure pipeline
- we find something else.

Ideally we would like to have the html hosted somewhere to be easily displayed, without download, etc.. I will have a quick look at what can be done.

doc/dev_guide/testing.rst
```diff
@@ -83,6 +83,41 @@ Useful hints on testing:
...
+
+The test suite is run using continuous integration services provided by
+`Travis CI <https://travis-ci.org/github/hyperspy/hyperspy>`_ and
+`Azure Pipeline <https://dev.azure.com/franciscode-la-pena-manchon/hyperspy/_build>`_.
+The CI helper scripts are pulled from the
+`ci-scripts <https://github.com/hyperspy/ci-scripts>`_ repository.
+
+The testing matrix is as follow:
+
+- **Travis CI**: test all supported python version on Linux; all dependencies are
```
tjof2: "versions"

doc/dev_guide/testing.rst
```diff
@@ -83,6 +83,41 @@ Useful hints on testing:
...
+  pulled from `pypi <https://pypi.org>`_,
+- **Azure Pipeline**: test a range of python version on Linux, MacOS and Windows;
+  all dependencies are pulled from `anaconda cloud <https://anaconda.org/>`_
+  using the `Anaconda "defaults" <https://anaconda.org/anaconda>`_ and the
+  `"conda-forge" <https://anaconda.org/conda-forge>`_ channel (in this order of
+  priority)
+
+This testing matrix has been design to be simple and easy to maintain and also
```
tjof2: "designed"

doc/dev_guide/testing.rst
```diff
@@ -83,6 +83,41 @@ Useful hints on testing:
...
+to ensure that packages from pypi and Anaconda cloud are not mixed in order to
+avoid red herring failure of the test suite caused by application binary
```
tjof2: "failures"

doc/dev_guide/testing.rst
```diff
@@ -83,6 +83,41 @@ Useful hints on testing:
...
+interface (ABI) incompatibility between dependencies.
+
+The most recent version of packages will be available first on pypi and later
```
tjof2: "versions of packages", not version

doc/dev_guide/testing.rst
```diff
@@ -83,6 +83,41 @@ Useful hints on testing:
...
+on anaconda cloud. It means that if a recent release of a dependency breaks the
+test suite, it should happen first on travis - typical usual suspect would be
```
tjof2: "usual suspects would be" - delete typical

tjof2:
@ericpre  minor tweaks as requested by @francisco-dlp

ericpre:
One option could be:
https://www.netlify.com/blog/2016/07/20/introducing-deploy-previews-in-netlify/
https://www.netlify.com/legal/open-source-policy
(from https://github.com/zarr-developers/zarr-python/pull/516#issuecomment-555832034)

ericpre:
Another alternative would be to use readthedocs, but we need to request this feature to be enabled:
https://docs.readthedocs.io/en/stable/guides/autobuild-docs-for-pull-requests.html

francisco-dlp:
Netlify seems a bit overkill for this application. The RtDs option is perfect. I'll send them a message if you agree.

ericpre:
> Netlify seems a bit overkill for this application. The RtDs option is perfect. I'll send them a message if you agree.

Agreed! Since the release documentation is published on RtDs, this also have the advantage of streamlining the release process by avoiding any bad surprise with RtDs.

PR #2375, opened 2020-04-21
Support for Masking

### LINKED ISSUE #2208: Adding support for Masking
As Dask now supports masks I think it would be worth thinking about how to best add support for masks in Hyperspy.  There are some implementation in different programs within Hyperspy but wider support would simplify a lot of these cases. The question would be how do we want users to access the mask?  Would it be preferable to access it using signal.data.mask, through a defined set of signal functions or through adding a new property to the signal class which handles all of the interaction with the mask?

### LINKED ISSUE #2318: Masking off a data that falls outside a circle ROI
image

I just want to mask off the image that falls outside the ROI in the figure.

Is there any command to do this?

CSSFrancis:
### Description of the change
This pull request adds direct support for masked arrays in hyperpsy. It creates a hyperspy.ma module that mirrors the Dask.array.ma and numpy.array.ma modules.  

It also expands on those modules by allowing for masking of more advanced shapes using a `ma.mask_roi(signal, roi, axis) ` method.  It also uses the fancy slicing in hyperspy to mask using both isig and inav.

This should close #2208 and  close #2318

**Motivating this change**
- Add the ability to mask data but not delete the data
- Mask things like beam stops etc in 2 dimensional datasets
- Brings additional support for numpy operations

### Progress of the PR
- [x] Numpy.ma and Dask.array.ma methods implemented
    - [x] Add tests for Numpy.ma and Dask.array.ma
- [x] Add support for masking using roi
    - [x] Add tests for masking using roi
- [x] Chage save method to save and load mask in .hspy format
    - [x] Add tests for saving and loading masked arrays
- [ ] Update user guide with new mask section 
- [ ] ready for review.

### Example of uses
```python
>>> import hyperspy.api as hs
>>> import numpy as np
>>> s = hs.signals.Signal2D(np.ones((2,10,10)))
>>> hs.ma.asarray(s)
>>> type(s.data)
np.MaskedArray
>>> hs.ma.masked_equal(s, 1) # masked everywhere
>>> s.isig[0:4,0:3] = hs.ma.masked # masking a slice
>>> c = hs.roi.CircleROI(3,3,3)
>>> hs.ma.masked_roi(s, c, axes="signal") # chose the axes to apply the roi to
```


ericpre:
Thanks @CSSFrancis for this PR. I feel that this PR may not be a straighforward to get right because we will need to find a good syntax and we need to think carefully about its implementation. It is very likely that there will be a couple of iterations but this PR looks promising and it would be a very nice addition!

Without having thought much about it, here are a few things I am wondering:
- ROI are designed to be used interactively, how should it work with mask?
- if I remember correctly (@francisco-dlp please correct me if I am wrong), `nan` are used in some part of hyperspy to work lazily before dask started to support masked array. It may worth revisiting this and making things more consistent.
- masking is already use in various part of hyperspy, for example, in the machine learning bits and this is not always very consistent, so this would be good to address this issue in this PR.

CSSFrancis:
@ericpre Thanks for the feedback.  I figured that this might take a little while, because there are quite a few 'mask like' features in hyperspy.  In general, my opinion is that following the implementations from numpy and Dask is fairly straight forward, familiar for people who have used numpy and Dask for masking and easy to update as Dask and numpy release new iterations.

Where things start to get difficult is with changing the mask when you have a lazy signal.  Dask does a fairly good job of handling the predefined methods but it suffers because it can't mask slices efficiently and with things like masked_where require that the boolean mask be the same shape as the Dask Array.  

For the last case there might be potential for using a sparse array but I need to look into this more.

> Without having thought much about it, here are a few things I am wondering:
> 
> * ROI are designed to be used interactively, how should it work with mask?

I haven't used the ROI feature in hyperspy that much, but my thought was that ROI's are a rather flexible way to select some area and then you would harden that selection to create a mask. Similar to the way the __call__ function for ROI's slices the signal from that ROI. 

Maybe something like a .mask( ) function for the ROI would make more sense. 

It wouldn't be that hard to make a function like mask shape, so the user wouldn't have to go though the work of creating an ROI.  

> * if I remember correctly (@francisco-dlp please correct me if I am wrong), `nan` are used in some part of hyperspy to work lazily before dask started to support masked array. It may worth revisiting this and making things more consistent.

That makes sense.  I can see a use case for each nan and a mask so going though and making that consistent would be a addition to the code

> * masking is already use in various part of hyperspy, for example, in the machine learning bits and this is not always very consistent, so this would be good to address this issue in this PR.

That is definitely something I can look into. For the most part I think the masks in hyperspy are broken into two categories. 

- Full Frame Mask: Full boolean mask for masking the whole signal. 
-" Sparse" Mask: Masks that are consistent to just the signal axes or navigation axes

The first case is straight forward, the second is less straight forward and I can see why that might be a better representation in many cases.

francisco-dlp:
From a quick look to the code, it seems that this PR:

1. Wraps dask and numpy mask functions to make them work on both standard and lazy signals.
2. Adds support for saving masked arrays
3. Adds support for masking using ROIs

I don't think that point 1 is something that HyperSpy must address. numpy functions return dask arrays when operating on dask arrays. I ignore why this is not the case for masked arrays, but, in my opinion, that's something to fix in numpy/dask, not HyperSpy. Instead we could document how to use dask's and numpy's ``ma`` submodules with HyperSpy in the User Guide.

Adding support for saving masked arrays is very welcome, especially considering that HyperSpy already partially supports masked arrays and uses them to implement some ROIs.

Adding support for masking with ROIs is great. As @ericpre mentions, it is important to think on what's the best syntax for that. Adding a `mask` method to the ROIs sounds very good to me.

Instead of using ``ma.asarray`` function, we could consider adding ``mask`` and ``unmask`` methods to ``BaseSignal``.

The main question that we must address before continuing development in this branch is: what are we planning to use masked arrays for? It is clear that we cannot fully support masked arrays since many functions can't operate on masked arrays. Once we have a clear idea about when they are advantageous over ``nans`` and the navigation and signal masks that we currently support, it will be easier to take implementation decisions.

CSSFrancis:
@francisco-dlp Thanks for the input!  In regard to your comments, 

> I don't think that point 1 is something that HyperSpy must address. numpy functions return dask arrays when operating on dask arrays. I ignore why this is not the case for masked arrays, but, in my opinion, that's something to fix in numpy/dask, not HyperSpy. Instead we could document how to use dask's and numpy's `ma` submodules with HyperSpy in the User Guide.

That makes a fair bit of sense.  Originally I tried something like `np.ma.masked_below(s, 1 )` to work on a BaseSignal in hyperspy but that doesn't seem to work as expected.  I think it has something to do with the copy behavior when using the numpy mask method.  I think for the same reason Dask returns a numpy array when `np.ma` is used instead of `da.ma`. 

I think that there is some bug in numpy with regards to how they work with references and copying arrays in the ma module.

> Adding support for saving masked arrays is very welcome, especially considering that HyperSpy already partially supports masked arrays and uses them to implement some ROIs.
> 
> Adding support for masking with ROIs is great. As @ericpre mentions, it is important to think on what's the best syntax for that. Adding a `mask` method to the ROIs sounds very good to me.
> 
> Instead of using `ma.asarray` function, we could consider adding `mask` and `unmask` methods to `BaseSignal`.

> The main question that we must address before continuing development in this branch is: what are we planning to use masked arrays for? It is clear that we cannot fully support masked arrays since many functions can't operate on masked arrays. Once we have a clear idea about when they are advantageous over `nans` and the navigation and signal masks that we currently support, it will be easier to take implementation decisions.

I think the benefit of a mask over `nans` is that with a mask you aren't replacing and losing data rather you are excluding it. Just pulling from my experience with 4-D STEM of amorphous materials I can think of a couple of things..
- Masking the zero beam for easier visualization
- Masking a zero beam for determining the center of a diffraction pattern
- Masking crystalline diffraction spots in a partially amorphous sample
- Excluding over exposed diffraction patterns from some analysis

For something like masking a beam stop, however, I could see the benefit of something like using `nans` over a mask.

In relation to the signal and navigation masks used now I think a major benefit is saving the mask alongside the dataset.  This helps with the reproducibility of some analysis.  It also standardizes how masks are dealt with in hyperspy.  This is at the expense of some additional memory for a masked dataset but with a good lazy implementation this shouldn't be an issue.

CSSFrancis:
Looking into it a bit closer it seems like the problem is with the __array_wrap__ function being set twice, once with the masked_array class and once by hyperspy.

From the numpy documentation ["For ufuncs, it is hoped to eventually deprecate this method in favour of __array_ufunc__."](
https://numpy.org/doc/stable/reference/arrays.classes.html?highlight=__array_wrap#numpy.class.__array_wrap__) ~~I might look into  this a little bit further and see if this is fixed by implementing some of the newer numpy functions.~~

Edit: I think that this is still right there are just some issues with how precedence is handled

CSSFrancis:
@francisco-dlp After looking more into how the ma modules in Dask and Numpy work there appear to be a couple of well known bugs in the codes of each of them.  From what I understand both of these modules were created before numpy started to focus on a well implemented standard for Duck Typed and inherited arrays.

This might some something that Hyperspy suffers from as well.

As a duck-typed class things like `np.sum(signal1, signal2)` returns a signal and `np.sum(lazysignal1, lazysignal2)` returns a lazy signal but in the latter case the returned lazy signal's data is a numpy array rather than a Dask Array.  I think this was already a known issue in hyperspy but I can work on this over the weekend and see if there is now a solution given Numpy's new set of subclassing functions.

CSSFrancis:
Closing this as Stale

PR #2378, opened 2020-04-24
(Re)enable code coverage and fail on sphinx warnings

francisco-dlp:


francisco-dlp:
It took me a while to find what what's wrong but now it should be fine, so ready for review.

One of the issues was that the syntax to set 2 env variables in the same line was wrong since the variables separator must be a space instead of a semicolon. However, I've tried with the space and it didn't work either. As a workaround, now the default is to install the coverage requirements even for the builds that don't require it. This is a tiny drawback, so I give up trying to find the right syntax...

francisco-dlp:
With the latest commit the sphinx should fail on warnings.

francisco-dlp:
Although turning warnings into errors was working, I've reverted it because it wasn't very useful since an error was stopping the building process.

ericpre:
Yes, sorry about breaking codecov in #2363! 

We shall revisit "turning warning into errors" when the doc doesn't have warning in all branches (`RnM` should be in decent shape).

Is it normal that when following the link of codecov, we need to log in? Is there something to do about it?

francisco-dlp:
That's documented behavior and it doesn't seem that we can do something about it.

PR #2380, opened 2020-04-25
Add `attolight` io_plugin

jordiferrero:
## Description of the change
Adds the functionality to load hyperspectral cathodoluminescence data from any Attolight system. The file it loads is with extension `.bin`.
It has been tested and works, but it loads the signal as a `Signal1D` type. It should load a `CL_SEM_Spectrum` object instead (imported from the `lumispy` package.

- Creates `attolight` io_plugin file
- Adds the `bin` extension in the `io_plugins.py` file.

_This follows from the discussion in the the lumispy extension package (num 16) with @jlaehne, @francisco-dlp and @ericpre_

### Progress of the PR
It has been tested with the `lumispy` attolight example data file. It has an issue:
- [x] It does not import the correct `signal_type` lumispy class of `CL_SEM_Spectrum`. 

### Bugs to be fixed
The `assign_signal_subclass()` from `io.py` does not load the `CL_SEM_Spectrum` signal_type from the hyperspy extensions. 
This needs to be fixed.


thomasaarholt:
Hey! This could come in handy for me in the future (I don't have access to any attolight systems atm though)! I'm happy to help review this. 

Are you able to add a tiny (~100kB or smaller) .bin file to run tests with? If possible, add any possible annotations or similar features that attolight support to the file.

jordiferrero:
@thomasaarholt You can go to the [lumispy-demos](https://github.com/LumiSpy/lumispy-demos) repository and you can find a test `HYPCard.bin` file. 
For now, if you load it with the `lumispy` package it works (because the lumispy package is in a patched state before this PR goes forward). However, if you try to load the example `.bin` file using `hs.load()` without importing lumispy, then it does not load it as a `CLSEMSpectrum` object... but as a `Signal1D` object.

jlaehne:
The functionality would still need a test in hyperspy/tests/io/ and you can probably write one based on some other io format. It would probably be good to use a minimal input file also for the tests there that contains all necessary metadata, but has like only 2x2 pixels or so.

Note that there will be a feature freeze for v1.6 on May 13 (https://github.com/hyperspy/hyperspy/issues/2251#issuecomment-619799257), so maybe there is a chance to still push this one in.

francisco-dlp:
The reason why it doesn't load the data with the correct signal type is that HyperSpy doesn't know about 'CL_SEM_Spectrum'. This is something to fix in LumiSpy, not this PR.

```
>>> hs.print_known_signal_types() 
+-----------------------+---------------------------------------+--------------------+----------+
|      signal_type      |                aliases                |     class name     | package  |
+-----------------------+---------------------------------------+--------------------+----------+
|      CL_Spectrum      |        CL, cathodoluminescence        |     CLSpectrum     | lumispy  |
|      Luminescence     | Luminescence BaseSignal, luminescence |     CommonLumi     | lumispy  |
|   DielectricFunction  |          dielectric function          | DielectricFunction | hyperspy |
|        EDS_SEM        |                                       |   EDSSEMSpectrum   | hyperspy |
|        EDS_TEM        |                                       |   EDSTEMSpectrum   | hyperspy |
|          EELS         |                TEM EELS               |    EELSSpectrum    | hyperspy |
|        hologram       |                                       |   HologramImage    | hyperspy |
| Luminescence_Spectrum |          LuminescenceSpectrum         |    LumiSpectrum    | lumispy  |
+-----------------------+---------------------------------------+--------------------+----------+
```

jordiferrero:
Since all checks have passed (after adding tests and example dataset) and the special signal issue is not part of this PR, what next steps are required to merge in the `minor_release` branch?

@francisco-dlp Where in the lumispy package should I make the change (the structure is the same as hyperspy)? As far as I know, all signal classes are imported in the `__init__.py` file and stated in the `hyperspy_extension.yaml`.

ericpre:
Yesterday, when looking at this, I noticed the same as @francisco-dlp did, so I started to make a branch to fix the signal registration. If I may have time today to finish it and make a PR - there are more things.

It would make sense to first fix the signal registration in lumispy and hen check that this PR is working correctly before merging it.

ericpre:
https://github.com/LumiSpy/lumispy/pull/25

jordiferrero:
There is one last issue to address here and I am not sure where the problem comes from.
In theory, there are 2 scenarios that should work as follows:

1.  If only the `import hyperspy.api as hs` is imported, then running the `hs.load()` function on the attolight data should return a `Signal1D` object and the warning: `This file contains a signal provided by the lumispy Python package that is not currently installed. The signal will be loaded into a generic HyperSpy signal. Consider installing lumispy to load this dataset into its original signal class.`. **This works correctly and it is fully tested successfully.**

2.  If the `import lumispy as lum` is imported, then running the `lum.load()` function on the attolight data will call the hyperspy loading function, but now the lumispy signal classes should also be registered. This means that the data should return a `CLSEMSpectrum` object and no warning. **This is not working in my computer.** Moreover, I don't know if it is possible to run a test in hyperspy which relies on classes from another package...

Let me know if you have any ideas on how to proceed.

ericpre:
> There is one last issue to address here and I am not sure where the problem comes from.
> In theory, there are 2 scenarios that should work as follows:
> 
> 1. If only the `import hyperspy.api as hs` is imported, then running the `hs.load()` function on the attolight data should return a `Signal1D` object and the warning: `This file contains a signal provided by the lumispy Python package that is not currently installed. The signal will be loaded into a generic HyperSpy signal. Consider installing lumispy to load this dataset into its original signal class.`. **This works correctly and it is fully tested successfully.**
> 2. If the `import lumispy as lum` is imported, then running the `lum.load()` function on the attolight data will call the hyperspy loading function, but now the lumispy signal classes should also be registered. This means that the data should return a `CLSEMSpectrum` object and no warning. **This is not working in my computer.** Moreover, I don't know if it is possible to run a test in hyperspy which relies on classes from another package...
> 
> Let me know if you have any ideas on how to proceed.

Any reader needs to return a dictionary which can be parse to a hyperspy (or extension) signal. Most readers don't import hyperspy, and they do, this is to use utility functions. In this sense, I don't think there should be a warning about lumispy not being install. After the split, the io library will be "library agnostic" and will not depend on hyperspy but the reader should still provide suitable information to be used by hyperspy and others.

The recommended way is to do:
```python
>>> import hyperspy.api as hs
>>> s = hs.load(filename)
>>> s
<CLSEMSpectrum, title: , dimensions: (xx|yy)>
```

The tests here should make sure that the `signal_type` is set correctly but it should not checked which class the data have assign to since they are two different things which need to be tested separately. The class assignment need to be tested in lumispy (https://github.com/LumiSpy/lumispy/pull/25).

hyperspy/io_plugins/attolight.py
```diff
@@ -0,0 +1,351 @@
...
+writes = False
+# ----------------------
+
+# Attolight system specific parameters
+# ------------------------------------
+# Store the system specific parameters in a dictionary for
+# different Attolight SEMs around the world, with their identifier.
+
+attolight_systems = {
+    'cambridge_uk_attolight': {
```
ericpre: The characteristics of a particular system should not be hard-coded in the reader. What is the reasoning for it?

hyperspy/io_plugins/attolight.py
```diff
@@ -0,0 +1,351 @@
...
+                chamber_pressure_torr = float(line[line.find(':') + 1:-6])
+            if 'Real Magnification:' in line:
+                real_magnification = float(line[line.find(':') + 1:-3])
+
+    # Correct channels to the actual value, accounting for binning. Get
+    # channels on the detector used (if channels not defined, then assume
+    # its 1024)
+    try:
+        total_channels
+    except:
```
ericpre: Specify the error you expect to make sure you are not swallowing other bugs.

hyperspy/io_plugins/attolight.py
```diff
@@ -0,0 +1,351 @@
...
+
+def _get_metadata(filename, md_file_name, attolight_acquisition_system):
+    """Import the metadata from the MicroscopeStatus.txt file.
+    Returns binning, nx, ny, FOV, grating and central_wavelength.
+    Parameters
+    ----------
+    filename : str
+        The absolute folder path where the md_file_name exists.
+    """
+    path = os.path.join(filename, md_file_name)
```
ericpre: What's about something the line of:
```python
original_metadata = {}
with open(filename) as f:
    for line in f:
        try:
           key, value = line.split(":")
           original_metadata[key] = value
        except ValueError:
            # not enough values to unpack
            pass
```
Then you can use the pint library to parse the value and the units of the metadata you will be using.

hyperspy/io_plugins/attolight.py
```diff
@@ -0,0 +1,351 @@
...
+    # channels on the detector used (if channels not defined, then assume
+    # its 1024)
+    try:
+        total_channels
+    except:
+        total_channels = attolight_systems[attolight_acquisition_system]['channels']
+    channels = total_channels // binning
+
+    # Return metadata
+    return binning, nx, ny, FOV, grating, central_wavelength_nm, channels, amplification, readout_rate, \
```
ericpre: This is fairly error prone, it is significantly better to put these in a dictionary, see above.

hyperspy/io_plugins/attolight.py
```diff
@@ -0,0 +1,351 @@
...
+           exposure_time_ccd_s, dwell_time_scan_s, beam_acc_voltage_kv, gun_lens_amps, obj_lens_amps, aperture_um, \
+           chamber_pressure_torr, real_magnification
+
+
+def _store_metadata(dict_tree, hypcard_folder, md_file_name,
+                    attolight_acquisition_system):
+    """
+    Store metadata in the DictionaryTreeBrowser metadata. Stores
```
ericpre, line 129: No need to use the `DictionaryTreeBrowser` here and actually its usage in io plugins will be removed for the split of the io library. Instead, in this case, this is appropriate to use the `mapping` approach to map values of the `original_metadata` attribute of a signal - which contains all possible metadata - to the `metadata` attributes - which contains only standardise keys.

hyperspy/io_plugins/attolight.py
```diff
@@ -0,0 +1,351 @@
...
+        - MicroscopeStatus.txt to load all the metadata values
+        - Background file contained in the same folder
+        - SEM image (taken simultaneously) with the hyperspectral map.
+
+    Parameters
+    ----------
+    filename : str, None
+        The HYPCard.bin filepath for the file to be loaded, created by
+        the AttoLight software.
+        If None, a pop-up window will be loaded.
```
ericpre: Remove this line, since this is not correct - this is not what this function do.

hyperspy/io_plugins/attolight.py
```diff
@@ -0,0 +1,351 @@
...
+        the Cambridge Attolight SEM system.
+    lazy : bool
+        If True the file will be opened lazily, i.e. without actually reading
+        the data from the disk until required. Allows datasets much larger than
+        available memory to be loaded.
+
+    Returns
+    -------
+    s : Signal
+        A CLSEMSpectrum lumispy object containing the loaded data with the metadata,
```
ericpre: This function returns a dictionary.

hyperspy/io_plugins/attolight.py
```diff
@@ -0,0 +1,351 @@
...
+    NOTE
+    ----
+
+    The lumispy package must be installed for these files to run.
+    https://github.com/LumiSpy/lumispy
+
+    """
+
+    # Get folder name
+    hypcard_folder = os.path.split(os.path.abspath(filename))[0]
```
ericpre: use `os.path.dirname`.

hyperspy/io_plugins/attolight.py
```diff
@@ -0,0 +1,351 @@
...
+
+    # Get folder name (which is the experiment name)
+    name = os.path.basename(hypcard_folder)
+
+    if attolight_acquisition_system == 'cambridge_attolight' and len(name) > 37:
+        # CAUTION: Specifically delimeted by Attolight default naming system
+            name = name[:-37]
+
+    # Create metadata dictionary
+    meta = DictionaryTreeBrowser({
```
ericpre: Same as above, don't use DictionaryTreeBrowser.

hyperspy/io_plugins/attolight.py
```diff
@@ -0,0 +1,351 @@
...
+
+    # Get the calibrated scanning axis scale from the acquisition_systems
+    # dictionary
+    calax = cal_factor_x_axis / (FOV * nx)
+
+    size = data.shape[1]
+    name = ''
+    scale = calax * 1000
+    # changes micrometer to nm, value for the size of 1 pixel
+    units = '$nm$'
```
ericpre: Why using `"$"`?

hyperspy/io_plugins/attolight.py
```diff
@@ -0,0 +1,351 @@
...
+    exposure_time_ccd_s, dwell_time_scan_s, beam_acc_voltage_kv, gun_lens_amps, obj_lens_amps, aperture_um, \
+    chamber_pressure_torr, real_magnification = \
+        _get_metadata(hypcard_folder, metadata_file_name, attolight_acquisition_system)
+
+    # Add all parameters as metadata
+    _store_metadata(meta, hypcard_folder, metadata_file_name, attolight_acquisition_system)
+
+    # Load data
+    with open(filename, 'rb') as f:
+        data = np.fromfile(f, dtype=[('bar', '<i4')], count=channels * nx * ny)
```
ericpre, line 320: The data always have 3 dimensions?

hyperspy/io_plugins/attolight.py
```diff
@@ -0,0 +1,351 @@
...
+    ----------
+    filename : str, None
+        The HYPCard.bin filepath for the file to be loaded, created by
+        the AttoLight software.
+        If None, a pop-up window will be loaded.
+    attolight_acquisition_system : str
+        Specify which acquisition system the HYPCard was taken with, from the
+        attolight_systems dictionary file. By default, it assumes it is
+        the Cambridge Attolight SEM system.
+    lazy : bool
```
ericpre: Lazy support is not implemented.

ericpre (changes requested):
I have provided guidance on what needs to be changed. Please have a look at other reader to get an idea of the different patterns which can be used.

hyperspy/io_plugins/attolight.py
jordiferrero: Every system has its own calibration parameters, they need to be stored somewhere as they are needed for the loading of the file.
Other readers also have hard coded parameters... 
I don't know where else these could be stored...

hyperspy/io_plugins/attolight.py
jordiferrero: It's a `NameError`.

hyperspy/io_plugins/attolight.py
jordiferrero: Yes. Always 2 navigation dimensions and 1 signal dimension.

hyperspy/io_plugins/attolight.py
jordiferrero: True.

hyperspy/io_plugins/attolight.py
jordiferrero: I don't fully understand... What `mapping` function? What are the standardised keys?
What would this function return then instead of modifying the DictionartTreeBrowser?

hyperspy/io_plugins/attolight.py
jlaehne: Is there no way to record and save linescans or single spectra in the Attolight software?

hyperspy/io_plugins/attolight.py
jlaehne: Good question, there should at least be a way to also specify a local config file. If it is hard coded, new users cannot use the plugin before a PR is through and a new version out. Not sure what the best way would be to do so. However, if we want to provide a way to correct for grating/detector/system response in LumiSpy, we would also have to think of where to store such calibration files without shipping all of them with the package (except maybe some more general ones for common grating blazes or detector models).

francisco-dlp:
@ericpre, is this ready?

hyperspy/io_plugins/attolight.py
ericpre: This still doesn't right to me and needs clarification: either this format is poorly engineered or the PR in this approach is not correct - not expected to be used that way.
By not having other alternative than hard coding the calibration parameter in the reader, I understand that the only attolight software installed in Cambridge can read the calibration data generated in the system in Cambridge (because the calibration is hard coded in the software) and equivalent software installed at other facilities in the world will read the calibration wrongly assuming they will not be aware of the calibration of the Cambrigde system (and because they will use a different calibration parameter specific to their system)?!

Maybe someone from Attolight could advice on this?

> Other readers also have hard coded parameters...

This is not common practice at all to have the calibration parameters hard coded, what reader(s) are you referring to? I can't think of a reader where this is the case.

> I don't know where else these could be stored...

Usually, this information is stored in the file itself or in a separate file, as in the ripple or the empad format. If this is not possible, then, as a default, I would prefer not to calibrate at all with a warning maybe, because not having calibration is better than having a wrong calibration. It will strongly encourage the user to calibrate the data correctly, otherwise a wrong calibration would be misleading. If you can know on what system it has been acquired then you can use your lookup table to set the calibration and it there is not entry for a specific acquisition system, then it should fall back to default (no calibration).

hyperspy/io_plugins/attolight.py
ericpre: There should be no `DictionartTreeBrowser` in the reader and there is no reason for this.

> What `mapping` function? 

`mapping` is a dictionary, see for example https://github.com/hyperspy/hyperspy/blob/30d28d23f074bbccc92ff6a537addb22566d2d1d/hyperspy/io_plugins/digital_micrograph.py#L841

> What are the standardised keys?

https://hyperspy.readthedocs.io/en/latest/user_guide/metadata_structure.html#metadata-structure

ericpre:
It needs more work, in particular, there is one blocking issue with the calibration (https://github.com/hyperspy/hyperspy/pull/2380#discussion_r462866110).

francisco-dlp:
Yes, I fully agree that hard-wiring the calibration is not a good approach. I also agree in that ideally the calibration should be in the file (something to fix in Attolight's acquistion software?) or in a separate file that could be provided to the reader as an extra keyword. I'll change the milestone to v1.7.

jordiferrero:
Hi all,
Thank you for your feedback and I understand this may not be in the standards for a Hyperspy io plugin.
However, I have not had the time to correct for some of these main concerns (I am not sure if I have the skills for some of these corrections as I am learning on the way).

I have been thinking on where to have this Attolight io add-on and I believe that maybe returning it to the lumispy `io_plugin` section may be better. That would mean I would have time to further develop it to the high standards needed for Hyperspy without the need to PR and getting accepted. 

So I suggest I will move all the changes of this pull request to the Lumispy code and I will try to PR again in the future (hopefully once the io_plugin is itself a separate package). If you have major objections to doing so, please let me know :-)

francisco-dlp:
Although I understand that the process of getting a PR in may feel frustrating at times, I think that this is one of the key benefits of social coding: thanks to the peer-review the code quality tends to be a lot better than what anybody would do on his own and everybody learns something in the process. Hence, I suggest to persevere, not the least because I think that this is not very far from being ready. The blocking issue is not a difficult one to fix, the main difficulty is agreeing on a solution.

hyperspy/io_plugins/attolight.py
jlaehne: A solution could also be to have a default location (specific location in the file system, or same folder as data) where the calibration file should reside. Additionally, the location of the calibration file could be optionally passed as parameter. If no calibration file is found, a message/warning could be printed and it falls back to no calibration.

Then it would be good to add a 'calibrated=True/False' tag to the signal object. Is that compatible with the hyperspy conventions to add such a custom tag @ericpre, @franciscodlp?

Lumispy could then include a routine to separately do the calibration on an uncalibrated signal object (again by providing a calibration file through a default location or with the location passed to the function as parameter).

francisco-dlp: > A solution could also be to have a default location (specific location in the file system [...]) where the calibration file should reside

I see how this could be convenient. However, it is less transparent to the user than the other 2 options, so I think that you should consider carefully if this is worth it.

> [or same folder as data]

This seems to me the best option (after Attolight fixing the format to include the calibration).

> Additionally, the location of the calibration file could be optionally passed as parameter. If no calibration file is found, a message/warning could be printed and it falls back to no calibration.

Sounds excellent. Then, why not simply adding a ``calibration_file`` with a default to a file with a specific name to implement the previous option?

> Then it would be good to add a 'calibrated=True/False' tag to the signal object. Is that compatible with the hyperspy conventions to add such a custom tag @ericpre, @franciscodlp?

What about implementing this by setting ``calibration_file=None``

Do you (@ericpre, @jordiferrero) think that this could be included in v1.6 that we are planning to release today?

jordiferrero: I have now removed the hard coded values.
Instead, the user can now input a calibration file with the load function providing the 'attolight_calibration_file' parameter.
If no calibration file is provided, then the load function will not calibrate the navigation axis.
I am thinking on maybe removing the calibration for the navigation axis, as it is dependent on many parameters in the metadata... However, it is something that need to be done every time, so for now I am keeping it in the loading function.

I have also removed the calibration of the signal axis as it was not accurate.

hyperspy/io_plugins/attolight.py
jordiferrero: I have also removed the dependency on DictionaryTreeBrowser. Now the metadata is stored as a dictionary only, not as a DictionaryTree anymore. @ericpre is that correct now?

jordiferrero:
I have added the two main changes that we discussed:

1. No hard coding of microscope parameters in the loading function.
2. No use of DictionaryTreeBrowser dependency

Now the function is working with all tests passing (in Azure pipeline). 
However travis-ci is failing, and I don't know why? I think it has nothing to do with my code contribution...
@francisco-dlp @ericpre

LMSC-NTappy:
Dear @jordiferrero. I don't mean to interfere, but as I understood you are an attolight tool user. 

From what I see you wrote a plugin to load your files from the binary and you parse "microscopestatus.txt" to extract metadata. Since hyperspy 1.6, the .sur file format is supported, including original metadata parsing, so you should be able to load a cl map straight away.

jordiferrero:
Dear @LMSC-NTappy. Thank you for updating me on the `.sur` file import. That's very handy and kind of removes the need for this PR.

However, there are a few things missing:
1. I am not sure how to parse the metadata `MicroscopeStatus.txt`. The loading function won't just do it. What it the command/parameter to add?
2. How is the spectral axis (wavelength) calculated? It should be a non-linear array as far as I am concerned. If you know, could you maybe explain quickly how it is calculated?

Thank you for your "interference", it's very helpful.

LMSC-NTappy:
Dear @jordiferrero ,

regarding your questions: 

>     1. I am not sure how to parse the metadata `MicroscopeStatus.txt`. The loading function won't just do it. What it the command/parameter to add?
Not sure what you exactly mean. The content of microscopestatus.txt is already present in a CL .sur file. You can access it with `map.original_metadata` (the structure is slightly complicated because .sur files can contain several objects and channels/object, e.g. an profilometer surface (1 channels) + an RGBA image (3 channels))

```python
dat = hs.load('example.sur')
dict(dat.original_metadata.Object_0_Channel_0.Parsed.CCD)

{'Baseline_Clamp': 'Off',
 'Camera_Model': 'A.920',
 'Capture_Depth': 1024,
 'Exposure_Time': 0.5,
 'Exposure_Time_units': 's',
 'Horizontal_Binning': 1,
 'Pixel_Width': 0.026,
 'Pixel_Width_units': 'mm',
 'ReadMode': 'FVB',
 'Readout_Rate_horizontal_pixel_shift': '50Khz',
 'Signal_Amplification': 'x4',
 'Speed_vertical_pixel_shift': 1.7,
 'Temperature': -69,
 'Temperature_units': '°C',
 'Voltage_Amplitude_vertical_pixel_shift': 'normal'}
```

>     2. How is the spectral axis (wavelength) calculated? It should be a non-linear array as far as I am concerned. If you know, could you maybe explain quickly how it is calculated?
It is calculated from the grating equation you are probably used to, linearised to the closest straight line. If you want to correct for the error introduced, the grating equation parameters and deformation polynomials are saved in the metadata. 

Glad I can help! I hope this answers your questions

ericpre:
It sounds like the only thing missing is to mention in the user guide that the DigitalSurf MountainsMap file reader can open file generated on Attolight system!

@jordiferrero, should this PR be closed then?

jordiferrero:
@ericpre Yes, this PR should be closed. I'm happy for your contribution @LMSC-NTappy ! It's very helpful!

LMSC-NTappy:
No problem @jordiferrero, it's a pleasure. I'll keep an eye on LumiSpy and the implementation of CL-specific classes of signal1D as well.

Cheers

ericpre:
@jordiferrero, @LMSC-NTappy, can one of you two updates the user guide to clarify that the "sur" reader can read "attolight" files?

LMSC-NTappy:
@ericpre 
Done. See PR #2559

PR #2203, opened 2019-06-07
Kmeans cluster analysis

pquinn-dls:
Cluster analysis for Hyperspy. 
Implementation of k-means clustering using scikit-learn. 

I looked at the docs and structure of a previous pull-request for a fuzzy clustering method to try and make it possible to add other clustering methods if needed in future.

Methodology largely based around:
"Cluster analysis of soft X-ray spectromicroscopy data". - Ultramicroscopy 100 (2004) 35–57 
https://doi.org/10.1016/j.ultramic.2004.01.008

### Progress of the PR
- [x] Further updates to user guide
- [x] add tests
- [x] ready for review.

### Minimal example 
```python
>>> import hyperspy.api as hs
>>> import numpy as np
>>> s.decomposition()
>>> s.cluster(3)
>>> s.plot_cluster_results()

PR #2383, opened 2020-04-27
Improve machine learning API and documentation

### LINKED ISSUE #1000: s.decomposition() should raise an error on unsuccessful decomposition
As it currently stands, `s.decomposition()` will happily run on an image without having any effect. It would be nice if it raised an error if it didn't do anything.

### LINKED ISSUE #1159: Improve decomposition documentation

### LINKED ISSUE #1361: mini_batch_sparse_pca not working

### LINKED ISSUE #1678: parameter 'algorithm' from BSS mask the 'algorithm' argument from sklearn.fastica
In the blind_source_separation function they are two 'algorithm' parameters. One belongs to the main blind_source_separation function (change between different types of ICA, e.g. between fastica, jade, cubica etc.). The other one belongs to the sklearn.decomposition.fastica function and allows to change between 'parallel' and 'deflation' options in the FastICA. The two arguments  are masking each other.

### LINKED ISSUE #1721: EDSTEMSpectrum.decomposition missing standard arguments
Whilst EDSSEMSpectrum has decomposition arguments that I would expect, the TEM version does not. "algorithm" and "output_dimension" are missing.

### LINKED ISSUE #2112: Signal.decompositon tests are incomplete
Indeed we test a number of features of the method, but the method itself is untested. This leads to undetected bugs like #2109.

### LINKED ISSUE #2277: RPCA
Hey guys, 
I was trying to perform robust principal componente analysis to remove outliers from some espectra data, but the things are not going the way I expected. 
For example, I have 250 spectrums, so I've loaded them with s = hs.signals.Signal1D(file) (navigation = 250; signal = 512). Then, used the function s.decomposition() and, with the scree plot, i choosed the output dimension. Then, I performed the RPCA (both algorithms) and the denoising function (s.get_decomposition_model()) with the output_dimension I've found before. However, when I retrieve the 'denoised' data, it is exactly the same thing it was before.
I could not figure out what I am doing wrong since I've followed all the steps in the documentation (http://hyperspy.org/hyperspy-doc/current/user_guide/mva.html), at least I think I have.
By the way, i tried the spikes_removal_tool(), it worked pretty well, but it's too slow for the amount of data i have.
Can you help me with that ?
 Thanks !!

### LINKED ISSUE #2346: Improve test coverage for lazy, online decomposition algorithms
We have tests for ORPCA and ORNMF decomposition algorithms using a plain `np.ndarray` as the data input, but are currently lacking good tests for the *lazy* option for data loading.

tjof2:
### Description of the change

In summary, this is an overhaul of `s.decomposition()`, `s.blind_source_separation()` and related functions, docstrings, and User Guide documentation. It would be good to get some feedback from @dnjohnstone and @thomasaarholt based on the discussion in #1159.

Here are the major changes:

- Custom objects can now be passed as algorithm arguments. This means we no longer have to wrap each sklearn algorithm individually, and also means a user can pass an sklearn `Pipeline()` as their algorithm. For example:

```python
from sklearn.decomposition import IncrementalPCA, FastICA

s.decomposition(algorithm=IncrementalPCA(batch_size=3))
s.blind_source_separation(algorithm=FastICA(whiten=False, algorithm='deflation'))
```
- Singular value decomposition (SVD) has been rewritten to follow the sklearn approach for choosing a solver. So instead of saying `algorithm="fast_svd"` as before, you can now do `algorithm="svd", svd_solver="auto"` to automatically select the best algorithm. Passing `svd_solver="randomized"` preserves the old `"fast"` behaviour, and the deprecation warning reflects this.
- Decomposition of `LazySignals` has been tidied up to bring it into line with "standard" decomposition behaviour. The docstring and user guide has been updated, and explicit tests for lazy signals added.
- `orthomax()` has been moved to `hyperspy.learn.orthomax` and the performance optimized
- ORPCA and ORNMF algorithms have been improved such that they are consistent with each other, and verified that they work with `LazySignals`
- Inconsistent algorithm capitalization has been tidied up and deprecation warnings added (but the old methods will still work)
- BSS now has multiple options for whitening - the default is `"pca"` as before, but you can now do `"zca"` whitening as well as `None`
- Deprecated `centre="variables"/"trials"` for `centre="signal"/"navigation"` for more consistent naming
- The documentation for `mva.rst` has been largely rewritten
- Docstrings have been improved for everything in `hyperspy.learn.*`
- Tests have been written for everything in `hyperspy.learn.*`

### Issues affected by this PR

- Closes #1000 by checking that we raise an error on non-float data types, non-negative data for NMF and Poisson noise, and the `print_info=True` argument provides more information when `s.decomposition()` is run. Also include tests for other common decomposition errors & warnings.
- Closes #1159 by overhauling the "Machine learning" documentation section in the user guide, as well as the docstrings for `s.decomposition()`, `s.blind_source_separation()`, and many of the underlying algorithms including as `ORPCA()`, `ORNMF()`, `mlpca()` and `svd_pca()`.
- Closes #1361 by discussing the parameter & algorithm choices in the user guide, and providing links to the sklearn documentation for more information
- Closes #1678 by allowing the user to pass `sklearn.decomposition.FastICA(algorithm="deflation")` as the algorithm argument itself, such that `**kwargs` can be explicitly set outside of `s.decomposition()`. The user guide has been updated to include an example.
- Closes #1721 by tidying up the `EDSTEMSpectrum` docstring to explicitly refer the user to `MVA.decomposition()` and the user guide for more details
- Closes #2112 by improving `hyperspy.learn.*` test coverage considerably (was 74%, now 87%), including explicit tests for `s.decomposition(algorithm=...)`, and reducing time tests take
- Closes #2277 by discussing the parameter & algorithm choices in the user guide
- Closes #2346 by adding explicit tests for lazy decomposition algorithms

### Pull requests affected by this PR

Both of the pull requests below also touch `mva.py` and `mva.rst` in the User Guide, so there will be some conflict whether this is merged first or they are merged first.

- #2172 Multivariate curve resolution
- #2203 Kmeans cluster analysis

### Progress of the PR
- [x] Change implemented
- [x] Update docstrings
- [x] Update user guide
- [x] Add tests
- [x] Ready for review

### Open questions

The [Modular toolkit for Data Processing (MDP)](http://mdp-toolkit.sourceforge.net/) was last updated on 09.03.2016.

> Note that from this release MDP is in maintenance mode. 13 years after its first public release, MDP has reached full maturity and no new features are planned in the future.

The code needed to support MDP BSS algorithms is minimal (`try/except ImportError`) but perhaps we might want to remove it altogether?

The `orthomax()` function has no documentation except for that fact it was taken from `metpy`. @francisco-dlp, can you provide more insight into this? Also, I don't think the algorithm works for `gamma < 0` or `gamma > 1` - I got a lot of garbage output when testing it.


tjof2:
@ericpre your assistance with fixing this would be appreciated. Do I just follow the instructions here? https://hyperspy.readthedocs.io/en/latest/dev_guide/testing.html#plot-testing
```
hyperspy/tests/drawing/test_plot_mva.py ................FFFFFFF.         [ 15%]
```
FYI they fail because I introduced a `svd_flip()` function to flip the signs of U & V from an SVD, in the same way as sklearn. **If I comment that change out, the tests pass**. Should I update the test images?

From the sklearn docs, the purpose of  `svd_flip()`:
> Sign correction to ensure deterministic output from SVD.
>
> Adjusts the columns of u and the rows of v such that the loadings in the columns in u that are largest in absolute value are always positive.

One option would be to make `svd_flip` optional and test the behaviour of it on and off.

ericpre:
> @ericpre your assistance with fixing this would be appreciated. Do I just follow the instructions here? https://hyperspy.readthedocs.io/en/latest/dev_guide/testing.html#plot-testing
> 
> ```
> hyperspy/tests/drawing/test_plot_mva.py ................FFFFFFF.         [ 15%]
> ```

These instructions in the dev guide are slightly out of date: from #2369, the relevant part to fix the plotting test locally is:

> The plotting tests are tested on azure pipeline against a specific version of matplotlib generates the figure can make the tests fail.	matplotlib defined in ``conda_environment_dev.yml`` since small changes in the way matplotlib generates the figure can make the tests fail.

To check the output of the test failure, you can download the artifacts in the azure build:
https://dev.azure.com/franciscode-la-pena-manchon/hyperspy/_build/results?buildId=85&view=artifacts

This should be documented in #2369.

ericpre:
> FYI they fail because I introduced a svd_flip() function to flip the signs of U & V from an SVD, in the same way as sklearn. If I comment that change out, the tests pass. Should I update the test images?

Yes, this is fine to update the reference images.

tjof2:
> Yes, this is fine to update the reference images.

Thanks @ericpre I've done this. 

I'll fix the remaining broken tests tomorrow. I know what's wrong with all of them so it should be simple.

ericpre:
It is fairly rare a PR which closes so many issues, good job!

tjof2:
> It is fairly rare a PR which closes so many issues, good job!

Thanks @ericpre!

> I'll fix the remaining broken tests tomorrow. I know what's wrong with all of them so it should be simple.

This is done, all the CI tests pass, now ready for review. Changes are below:

1. `orthomax(..., gamma=2)` on Windows leads ultimately to the inversion of a singular matrix, but not on Linux/OSX. I've added a `try/except` to use `pinv` instead of `inv` in this case, and warn the user that this is what is happening. This ties in with my comment above about `orthomax` returning garbage for some values of gamma.
2. There were a few more tests that relied on sklearn, so failed when it wasn't present. `skipif()` has been added.

ericpre (commented):
This looks very good and obviously this is a significant improvement!
Being able to pass a sklearn algorithm or pipeline is very nice. I haven't look into the details of how the pyMCR algorithm are implemented and I am wondering if it would make #2172 by simply passing pyMCR algorithm? If so, we should possibly merge this PR before and solve the merge conflict in #2247 

One cosmetic maintenance detail: since we don't support python < 3.6, I would suggest to use f-string - it is nicer to read and it feel less like speaking German and its usual mapping exercise with the verb at the end of the sentence... ;)

There are a few warnings related to this PR when building the doc:

```
Returns
-------
/home/eric/Dev/hyperspy/hyperspy/learn/mva.py:docstring of hyperspy.learn.mva.MVA.decomposition:97: WARNING: duplicate citation Keenan2004, other instance in /home/eric/Dev/hyperspy/doc/api/hyperspy._signals.lazy.rst
WARNING: autodoc: failed to import module 'onmf' from module 'hyperspy.learn'; the following exception was raised:
No module named 'hyperspy.learn.onmf'
WARNING: autodoc: failed to import module 'orthomax' from module 'hyperspy.misc.machine_learning'; the following exception was raised:
No module named 'hyperspy.misc.machine_learning.orthomax'
WARNING: autodoc: failed to import module 'test_onmf' from module 'hyperspy.tests.mva'; the following exception was raised:
No module named 'hyperspy.tests.mva.test_onmf'
looking for now-outdated files... none found
pickling environment... done
checking consistency... /home/eric/Dev/hyperspy/doc/api/hyperspy.external.rst: WARNING: document isn't included in any toctree
/home/eric/Dev/hyperspy/doc/api/hyperspy.learn.onmf.rst: WARNING: document isn't included in any toctree
/home/eric/Dev/hyperspy/doc/api/hyperspy.misc.machine_learning.orthomax.rst: WARNING: document isn't included in any toctree
/home/eric/Dev/hyperspy/doc/api/hyperspy.tests.rst: WARNING: document isn't included in any toctree
done
preparing documents... done
writing output... [100%] user_guide/what_is_new                                                                                  
/home/eric/Dev/hyperspy/doc/user_guide/getting_started.rst:133: WARNING: Could not lex literal_block as "python". Highlighting skipped.
../CHANGES.rst:41: WARNING: undefined label: scree-plot (if the link has no caption the label must precede a section header)
../CHANGES.rst:345: WARNING: undefined label: scree-plot (if the link has no caption the label must precede a section header)
../CHANGES.rst:443: WARNING: undefined label: rpca-label (if the link has no caption the label must precede a section header)
../CHANGES.rst:807: WARNING: undefined label: scree-plot (if the link has no caption the label must precede a section header)
../CHANGES.rst:809: WARNING: undefined label: mva.get_results (if the link has no caption the label must precede a section header)
```

tjof2:
Thanks - I can replace with f-strings if preferred! I like them more too.

I'll take a look at the doc warnings.

I think you might be right about pyMCR, I certainly think that these changes make it easier to add new things like that.

> There are a few warnings related to this PR when building the doc:

I think the ones about `No module named 'hyperspy.tests.mva.test_onmf'` go away if you do `make clean` first? The others are fixed though.

tjof2:
Managed to put together some documentation on the `orthomax()` function. Test failure is unrelated.

ericpre (commented):
I am not sure, we should keep orthomax in hyperspy since it would have a better place somewhere else (typically https://github.com/scikit-learn/scikit-learn/pull/11064), particularly since this PR enable passing scikit-learn algorithm.

Along similar line, we could probably get ride of the `hyperspy..misc.machine_learning` package. It seems that the MILCA package is now on github: https://github.com/nordavinden/mikstur.

Another thing I am wondering: would it make sense to display the algorithm name in the default title of the figure when plotting loadings or factors and also in the default filename?

tjof2:
> I am not sure, we should keep orthomax in hyperspy since it would have a better place somewhere else (typically scikit-learn/scikit-learn#11064), particularly since this PR enable passing scikit-learn algorithm.

In which case `blind_source_separation()` requires **either** sklearn or MDP installed. Unlike `decomposition()` which can proceed without sklearn. I think we can keep it, but maybe with the "broken" code branch where `gamma < 0` or `gamma > 1` deleted.

tjof2:
I agree that the `hyperspy.misc.machine_learning` could be removed, or at least the remaining contents moved to `hyperspy.learn.*` so everything is in one place.

The `orthomax()` function is used in the pyMCR PR #2172 so we shouldn't remove it.

hyperspy/_signals/eds_tem.py
```diff
@@ -622,31 +623,9 @@ def decomposition(self,
...
-            Array of variance for the maximum likelihood PCA algorithm
-        var_func : function or numpy array
-            If function, it will apply it to the dataset to obtain the
-            var_array. Alternatively, it can a an array with the coefficients
-            of a polynomial.
-        polyfit :
-        reproject : None | signal | navigation | both
-            If not None, the results of the decomposition will be projected in
-            the selected masked area.
+        **kwargs
```
ericpre: It would be convenient for the user to have all the parameters described here without having to look at the docstring of `~.learn.mva.MVA.decomposition`. Shall we define the parameters docstring in `~.docstring.mva` and reuse it in `~.learn.mva.MVA.decomposition` and `here?

ericpre (approved):
This looks good to me and I think we should merge this PR before moving to #2172 and #2203. @francisco-dlp, do you want to have a look at this PR?

hyperspy/_signals/eds_tem.py
francisco-dlp: Let me some minutes to have a quick look, it's a good distraction from my current boring tasks...

doc/user_guide/mva.rst
```diff
@@ -13,249 +13,244 @@ decomposition algorithms, such as principal component analysis (PCA), or
...
-The behaviour of some machine learning operations can be customised
-:ref:`customised <configuring-hyperspy-label>` in the Machine Learning section
-Preferences.
+.. hint::
 
-.. Note::
+   HyperSpy will decompose a dataset, :math:`X`, into two new datasets:
+   one with the dimension of the signal space known as **factors** (:math:`A`),
+   and the other with the dimension of the navigation space known as **loadings**
+   (:math:`B`), such that :math:`X = A B^T`.
```
francisco-dlp, line 21: Is it worth mentioning that this is not an equality for several algorithms?

doc/user_guide/mva.rst
```diff
@@ -13,249 +13,244 @@ decomposition algorithms, such as principal component analysis (PCA), or
...
-The point at which the scree plot becomes linear (often referred to as
-the `elbow`) is generally judged to be a good estimation of the dimensionality
-of the data (or equivalently, the number of components that should be retained
-- see below). Components to the left of the elbow are considered part of the "signal",
-while components to the right are considered to be "noise", and thus do not explain
-any significant features of the data.
+One of the most popular uses of :py:meth:`~.learn.mva.MVA.decomposition`
+is data denoising. This is achieved by using a limited set of components
+to make a model of the original, omitting the less significant components that
+ideally contain only noise. This is also known as *dimensionality reduction*.
```
francisco-dlp: "of the original." -> "of the original dataset."?

I would remove "This is also know as...", because I think that that's most often used in the context of reducing the number of components before feeding them to a BSS algorithm, but I may be wrong.

hyperspy/_signals/eds_tem.py
tjof2: Agreed I wasn't sure of the best way forward

tjof2: This is the only place it's wrapped for example, because it sets different defaults for Poisson noise compared to the standard func

doc/user_guide/mva.rst
```diff
@@ -13,249 +13,244 @@ decomposition algorithms, such as principal component analysis (PCA), or
...
 
-.. figure::  images/screeplot2.png
-   :align:   center
-   :width:   500
+   >>> s = Signal1D(np.random.randn(10, 10, 200))
+   >>> s.decomposition()
 
-   PCA scree plot with number-based axis labeling and a threshold value
-   specified
+Several algorithms exist for performing PCA, and the default algorithm in
```
francisco-dlp: I wonder if, instead of presenting SVD as PCA and then explainig that that's not actually PCA,  we should have a section for SVD and then explain that SVD is wrongly called PCA in some communities such as EM.

In the section about PCA we could advice passing directly the sklearn PCA object instead of setting `centre=True` as those ones also normalize the variance as required if I remember well.

doc/user_guide/mva.rst
```diff
@@ -338,100 +324,338 @@ aid the physical interpretation of components for count data such as images,
...
+components using a process called Blind Source Separation (BSS). This largely
+depends on the particular application. For more information about blind source
 separation please see :ref:`[Hyvarinen2000] <Hyvarinen2000>`, and for an
 example application to EELS analysis, see :ref:`[Pena2010] <Pena2010>`.
 
-To perform BSS on the result of a decomposition, run the
-:py:meth:`~.learn.mva.MVA.blind_source_separation` method, e.g.:
+.. warning::
+
+   Currently, the BSS algorithms operate on the result of a previous
```
francisco-dlp: Maybe remove "Currently" since this is not going to change?

doc/user_guide/mva.rst
```diff
@@ -338,100 +324,338 @@ aid the physical interpretation of components for count data such as images,
...
+HyperSpy includes a number of plotting methods for visualizing the results
+of decomposition and blind source separation analyses. All the methods
+begin with ``plot_``.
+
+.. _mva.scree_plot:
+
+Scree plots
+-----------
+
+PCA will sort the components in the dataset in order of decreasing
```
francisco-dlp, line 584: Maybe it'll be good to clarify here that the scree plot is only available for SVD and PCA. Actually it is only when using PCA that what we plot a explained variance is actually explained variance, but that's maybe something for another pull request.

doc/user_guide/mva.rst
```diff
@@ -443,60 +667,66 @@ The decomposition and BSS results are internally stored as numpy arrays in the
...
-.. NOTE::
-  Data exported in this way cannot be easily  loaded into HyperSpy's
-  machine learning structure.
+.. warning::
+   Data exported in this way cannot be easily loaded into HyperSpy's
+   machine learning structure.
+
+.. _mva.preferences:
+
+Preferences
```
francisco-dlp: There is no such a section in preferences, to remove?

hyperspy/_signals/lazy.py
```diff
@@ -634,154 +635,193 @@ def _block_iterator(self,
...
+            warnings.warn(
+                "The argument `algorithm='ONMF'` has been deprecated and may "
+                "be removed in future. Please use `algorithm='ornmf'` instead.",
+                VisibleDeprecationWarning,
+            )
+            algorithm = "ornmf"
+
+        # Deprecate uppercase to favour lowercase (consistent
+        # with non-lazy decomposition)
+        if algorithm in ["PCA", "ORPCA", "ORNMF"]:
```
francisco-dlp, line 723: Actually, why not doing the exact opposite i.e. deprecate the lowercase alternatives in ``decomposition`` and explain that we'll remove them in 2.0? They are acronyms after all...

hyperspy/learn/mva.py
```diff
@@ -161,101 +168,211 @@ def decomposition(self,
...
         to_print = [
             "Decomposition info:",
-            "  normalize_poissonian_noise={}".format(normalize_poissonian_noise),
-            "  algorithm={}".format(algorithm),
-            "  output_dimension={}".format(output_dimension),
-            "  centre={}".format(centre),
+            f"  normalize_poissonian_noise={normalize_poissonian_noise}",
+            f"  algorithm={algorithm}",
+            f"  output_dimension={output_dimension}",
+            f"  centre={centre}",
```
francisco-dlp, line 332: What about deprecating `centre` and propose to use a proper PCA algorithm instead?

hyperspy/learn/mva.py
```diff
@@ -407,52 +453,86 @@ def decomposition(self,
...
                 loadings = W
-                factors = H
+                factors = H.T
+
+            elif is_sklearn_like:
+                if hasattr(estim, "fit_transform"):
+                    loadings = estim.fit_transform(data_)
+                elif hasattr(estim, "fit") and hasattr(estim, "transform"):
+                    estim.fit(data_)
+                    loadings = estim.transform(data_)
```
francisco-dlp, line 491: Maybe check also for the `explained_variance_` attribute?

hyperspy/_signals/eds_tem.py
tjof2: Last commit copies over the parameters to the docstring.

hyperspy/learn/mva.py
```diff
@@ -471,202 +550,234 @@ def decomposition(self,
...
-            decomposition. If False, performs it on the factors.
-        reverse_component_criterion : str
-            One of {'factors', 'loadings'}. Use either the factor or the 
-            loading to determine if the component needs to be reversed.
-        **kwargs : extra key word arguments
+            decomposition, otherwise, perform the BSS on the factors.
+        reverse_component_criterion : {"factors", "loadings"}, default "factors"
+            Use either the factors or the loadings to determine if the
+            component needs to be reversed.
+        whiten_method : {"pca", "zca", None}, default "pca"
```
francisco-dlp, line 693: Is ZCA any useful for BSS?

francisco-dlp: If the answer is yes and we keep the new option, we can consider using uppercase for acronyms as suggested above.

hyperspy/learn/mva.py
```diff
@@ -1334,36 +1598,45 @@ def normalize_poissonian_noise(self, navigation_mask=None,
                 signal_mask = slice(None)
             else:
                 signal_mask = ~signal_mask
-            # Rescale the data to gaussianize the poissonian noise
+
+            # Check non-negative
+            if dc[:, signal_mask][navigation_mask, :].min() <= 0.0:
```
francisco-dlp: I think that ``==0`` is fine. `<=` -> `<`?

francisco-dlp (changes requested):
I've left a number of minor comments. This is an amazing contribution!

hyperspy/learn/mva.py
tjof2: Done

hyperspy/_signals/lazy.py
tjof2: I can do that.

doc/user_guide/mva.rst
tjof2: Done

doc/user_guide/mva.rst
tjof2: Done.

hyperspy/learn/mva.py
tjof2: I think it can be, and for a trivial extra two lines since we do the SVD for whitening anyway, I thought I might as well add it. 

I guess there's this page explaining the differences: https://stats.stackexchange.com/questions/117427/what-is-the-difference-between-zca-whitening-and-pca-whitening

> For most other machine learning algorithms it should be absolutely irrelevant whether the data is whitened with PCA or ZCA.

doc/user_guide/mva.rst
tjof2: Done

doc/user_guide/mva.rst
tjof2: Done

hyperspy/learn/mva.py
tjof2: I check for it lower down in line 513.

hyperspy/learn/mva.py
tjof2: I know what you mean, but sklearn PCA only has the option to center along `axis=0`, it's not customisable, while here we can do `axis=0/axis=1`. I'd vote to leave it in for now (with the appropriate naming convention followed).

hyperspy/_signals/lazy.py
tjof2: Actually, @francisco-dlp, I chose lowercase because that's how sklearn typically takes all of its string arguments. I could also do `.lower()` and avoid the deprecation altogether.

hyperspy/learn/mva.py
francisco-dlp: The thing is that I think that we should favor a minimalist approach: if something makes no difference as far as we can tell, better not to implement it since it'll increase the maintenance burden. If you can provide an example where it is useful that's a different story...

tjof2: Ok sure - I'll take it out.

tjof2: No ignore that - see https://cbrnr.github.io/2018/12/17/whitening-pca-zca/ 

> A recent article by Montoya-Martínez et al. (2017) suggests that some ICA variants can be sensitive to the choice of the initial whitening procedure. Specifically, it can make a difference whether PCA or ZCA is used prior to performing Extended Infomax as implemented in EEGLAB.

hyperspy/learn/mva.py
francisco-dlp: You are right but I lean towards minimalism for this too: if somebody needs centring over the other axes he can simply transpose the signal.

This is of course entirely up to you. This PR is great and I won't hold it any longer for something that wasn't actually its target.

hyperspy/learn/mva.py
tjof2: If people can now provide their own BSS algorithms (e.g. they could pass their own implementation of Extended Infomax) then perhaps having flexibility with whitening is good.

hyperspy/_signals/lazy.py
tjof2: On reflection, I'll take your approach and go with uppercase. Working on it now.

francisco-dlp: I've been always uncomfortable with using lowercase for acronyms but I didn't address this earlier not to annoy the users with syntax changes. When @to266 implemented the lazy versions, he used uppercase and I thought that it was right even if I was aware that it was inconsistent with the standard `decomposition` syntax. Therefore, it is not surprising that I prefer to switch to uppercase everywhere regardless of what sklearn does. But then, maybe it's just me. @ericpre, what do you think?

hyperspy/learn/mva.py
francisco-dlp: Sounds good then.

hyperspy/learn/mva.py
tjof2: Ok @francisco-dlp  I'll leave it in for now I think

hyperspy/_signals/lazy.py
tjof2: Since we can actually deprecate "sklearn_pca" and "sklearn_fastica" etc in future as a result of this pull request (and get the user to pass their own objects), the lowercase no longer matters.

doc/user_guide/mva.rst
tjof2: I've added a note section for the first point, I'll leave the second.

doc/user_guide/mva.rst
tjof2: > as those ones also normalize the variance as required if I remember well.

No, sklearn just subtracts the mean. You can do a pipeline with `StandardScaler()->PCA()` of course.

tjof2: Done.

hyperspy/_signals/lazy.py
tjof2: I've left them as they are for now, and made all the other changes @francisco-dlp.

francisco-dlp (approved):

tjof2:
Perhaps merge this and then we can decide how to deprecate/handle uppercase vs lowercase arguments in a separate PR with a smaller diff? I'm already conscious it's a big change

francisco-dlp:
I fully agree.

PR #2386, opened 2020-04-29
Support for relative slicing

thomasaarholt:
### Description of the change
- Added relative indexing to DataAxis
- Quickly get the middle value or index by `DataAxis.get_value_from_relative_value(0.5)` or `DataAxis.get_index_from_relative_value(0.5)`
- Allows the syntax `s.isig['rel0.5':]` to slice in a relative manner, getting a fraction of the signal
- Alternative to #2343 

### Progress of the PR
- [x] add tests,
- [x] ready for review.

### DataAxis example
```python
>>> from hyperspy.axes import DataAxis
>>> ax = DataAxis(size=11, scale=0.1, offset=10, units='nm')
>>> ax.get_value_from_relative_value(0.5)
10.5
```
### Example of slicing a signal
```python
# gets central half of data (except last index :/)
>>> import hyperspy.api as hs
>>> import numpy as np
>>> s = hs.signals.Signal1D(np.arange(11))
>>> s.isig['rel0.25':'rel0.75'].data
array([2, 3, 4, 5, 6, 7])
```


thomasaarholt:
As soon as the tests are finished (and passed!), this is ready for review.

hyperspy/axes.py
```diff
@@ -343,8 +371,16 @@ def _get_index(self, value):
...
 
     def _parse_string_for_slice(self, value):
+        '''
+        Parses one of the start/stop/step parts of the slice
+        syntax. Returns integer if index, float if calibrated axis value.
+        '''
         if isinstance(value, str):
-            value = self._get_index_from_value_with_units(value)
+            # these are always returned as integers
+            if value[:3] == 'rel': # Slicing using relative syntax
```
ericpre: Matter of taste but I find `value.startswith('rel')` more readable.

hyperspy/axes.py
```diff
@@ -114,11 +114,39 @@ def _convert_compact_units(self, factor=0.25, inplace=True):
...
+        '''
+        value = self.low_value + relative_value * (self.high_value - self.low_value)
+        if nearest_axis_value:
+            value = self.index2value(self.value2index(value))
+        return value
+
+    def get_index_from_relative_value(self, relative_value):
+        "Get the index of a position in the axis by relative or fractional value"
+        value = self.get_value_from_relative_value(relative_value)
+        return self.value2index(value)
```
ericpre: IInstead to adding these two methods, wouldn't it more simple to allow `value2index` to take "string" value?

ericpre (commented):
A couple of comments.

hyperspy/axes.py
thomasaarholt: Totally agreed!

hyperspy/axes.py
thomasaarholt: Yes, and then for the first method, one would do `index2value(value2index(value)))`.
I'll take a look at doing the same for values with units.

ericpre: Yes, or alternatively, use the `axis` attribute of `DataAxis`: `ax.axis[ax.value2index(value)]`
But what would be the use case for the first method?

ericpre:
As currently discussed, with this PR the behaviour will be:
- `value2index`: convert a float or string to an index
- `index2value`: convert an index to a value (this method seems to support arrays)

Is it worth considering making both methods more "symmetric"?
- add support for array in `value2index`?
- make `index2value` returns a string or a float?

hyperspy/axes.py
thomasaarholt: The usecase was just general convenience, for instance for getting the central half (0.25 to 0.75) of an axes for plotting something with matplotlib on top of a signal.

thomasaarholt:
My two cents:

- Definitely add array support, that is very useful
- Making index2value return string or a float isn't high on my priorities list. I'd rather let someone interested do that.

hyperspy/axes.py
```diff
@@ -112,11 +112,38 @@ def _convert_compact_units(self, factor=0.25, inplace=True):
...
+        return self.value2index(self._get_value_from_value_with_units(value))
+
+    def _get_value_from_relative_string(self, relative_string):
+        "Get axes index by relative indexing using string like 'rel0.5'"
+        try:
+            assert relative_string.startswith('rel')
+            relative_value = float(relative_string[3:])
+        except (ValueError, AssertionError):
+           print(f"Expected relative units of kind 'rel0.2' but '{relative_string}' could not be interpreted this way.")
+           raise
```
ericpre: I am not sure this is really necessary, since this is a private method and it is called after `startswith('rel')` has been checked. To keep the code more readable, I would remove it. 
However, it would be good to check that what's follow `rel` is a number between 0 and 1.

hyperspy/axes.py
```diff
@@ -515,11 +552,33 @@ def value2index(self, value, rounding=round):
...
             If any value is out of the axis limits.
+            If the value is incorrectly formatted.
 
         """
         if value is None:
             return None
 
+        if isinstance(value, str):
+            if value == "":
+                raise ValueError("Cannot slice with an empty string")
```
ericpre: Is using `slice` always suitable wording?

hyperspy/axes.py
```diff
@@ -515,11 +552,33 @@ def value2index(self, value, rounding=round):
...
+                try:
+                    value = self._get_value_from_value_with_units(value)
+                except:
+                    raise ValueError(f"Expected calibrated value with units like '20nm' but '{value}' could not be interpreted this way.")
+            else:
+                try:
+                    assert value.startswith('rel')
+                    relative_value = float(value[3:])
+                except (AssertionError, ValueError):
+                    raise ValueError(f"Expected relative units of kind 'rel0.2' but '{value}' could not be interpreted this way.")
```
ericpre: Similar as above, I am not convinced this is necessary and it would be better to use `_get_value_from_relative_string` to avoid repetition.

hyperspy/axes.py
```diff
@@ -528,6 +587,20 @@ def value2index(self, value, rounding=round):
...
+                    try:
+                        value = self._get_value_from_value_with_units(value)
+                    except pint.errors.UndefinedUnitError:
+                        raise ValueError(f"Expected iterable of calibrated value with units like ['20nm'] but '{value}' could not be interpreted this way.")
+                else:
+                    try:
+                        assert value[0].startswith('rel')
+                        relative_value = np.char.replace(value, 'rel', '').astype(float)
+                    except (AssertionError, ValueError):
+                        raise ValueError(f"Expected array of relative units of kind ['rel0.2', 'rel0.3] but '{value}' could not be interpreted this way.")
```
ericpre: Similar as above, it should be possible to refactor to avoid repetition and checking if the input is suitable at the right place.

ericpre (changes requested):
There are a few repetition in the code and it needs to be refactored to check if the input is valid at the right place.

hyperspy/axes.py
```diff
@@ -535,6 +489,68 @@ def __deepcopy__(self, memo):
         cp = self.copy()
         return cp
 
+    def _parse_value_from_string(self, value):
+        test_value = value if type(value) == str else value[0] # else array-like, and test first entry
```
ericpre: This doesn't look right to me: this function is expecting a string, and in case this is not a string, why would you take the first item and not any other?

ericpre (changes requested):
There is a test failure in this PR. I would remove support for parsing array for the following reasons:
- it is arbitrarily parsing the first index
- it isn't done in the right place
- it isn't clear how it is supposed to be used and why it would be useful
- it isn't documented

ericpre:
Ping @thomasaarholt.

ericpre:
@thomasaarholt, any change to get this PR completed?

ericpre (approved):

PR #2387, opened 2020-04-30
Get EELS edges from a given energy window

ptim0626:
### Description of the change
A method to get EELS edges from a given energy window, similar to the one in EDS (get_xray_lines_near_energy). 

### Progress of the PR
- [x] add tests,
- [x] ready for review.

### Minimal example of the new feature
```python
>>> from hyperspy import signals
>>> s = signals.EELSSpectrum(np.ones(1024)) #dummy spectrum
>>> s.get_edges_near_energy(532)
['O_K']
>>>  s.get_edges_near_energy(532, width=10)
['O_K', 'Pd_M3', 'Sb_M5', 'Sb_M4']
```

### DIFF
--- a/doc/user_guide/eels.rst
+++ b/doc/user_guide/eels.rst
@@ -33,6 +33,21 @@ information is stored in the :py:attr:`~.signal.BaseSignal.metadata`
 attribute (see :ref:`metadata_structure`). This information is saved to file
 when saving in the hspy format (HyperSpy's HDF5 specification).
 
+A method :py:meth:`~._signals.eels.EELSSpectrum_mixin.get_edges_near_energy`
+can be helpful to identify possible elements in the sample.
+:py:meth:`~._signals.eels.EELSSpectrum_mixin.get_edges_near_energy` returns a
+list of edges arranged in the order closest to the specified energy within a
+window.
+
+.. code-block:: python
+
+    >>> s = hs.datasets.artificial_data.get_core_loss_eels_signal()
+    >>> s.get_edges_near_energy(532)
+    ['O_K']
+    >>> s.get_edges_near_energy(849, width=6)
+    ['La_M4', 'Fe_L1']
+
+
 Thickness estimation
 ^^^^^^^^^^^^^^^^^^^^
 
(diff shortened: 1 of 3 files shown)


hyperspy/_signals/eels.py
```diff
@@ -154,6 +154,50 @@ def generate_subshells(self, include_pre_edges=False):
                                 '%s_%s' % (element, shell))
                             e_shells.append(subshell)
 
+    def get_edges_near_energy(self, energy, width=1):
```
ericpre: 1 eV is fairly narrow, would 5 or 10 be a more sensible default width?

ericpre (commented):
Nice and clean PR! I have one comments, otherwise it looks good to me.

ericpre:
Github/Azure seems to have had hiccup and doesn't want to restart but since the last commit was only changes in the user guide, it is fine.

hyperspy/_signals/eels.py
ptim0626: Yes sure, will set the default to 10.

doc/user_guide/eels.rst
```diff
@@ -33,6 +33,21 @@ information is stored in the :py:attr:`~.signal.BaseSignal.metadata`
 attribute (see :ref:`metadata_structure`). This information is saved to file
 when saving in the hspy format (HyperSpy's HDF5 specification).
 
+A method :py:meth:`~._signals.eels.EELSSpectrum_mixin.get_edges_near_energy`
+can be helpful to identify possible elements in the sample.
+:py:meth:`~._signals.eels.EELSSpectrum_mixin.get_edges_near_energy` returns a
+list of edges arranged in the order closest to the specified energy within a
+window.
```
tjof2: Best to define the window here, e.g. that you can control it with the width, and it's measured in eV

ptim0626: Good idea, I will elaborate a bit.

tjof2 (approved):

ericpre (approved):

ericpre:
Thanks @ptim0626!

PR #2390, opened 2020-05-06
Relocate get_edges_near_energy()

ptim0626:
### Description of the change
This is actually just relocating the new get_edges_near_energy method (implemented in #2387) from EELS signal class to EELS utility function, as the operation of the method does not depend on any EELS signal. This will be much more convenient for users to use this function (no need to create any dummy signal).

No content is changed. Since I don't know how to reopen/restore the old PR, I decide to open a new one for this. Sorry for the clustering. 

### Progress of the PR
- [x] Change implemented,
- [x] add tests,
- [x] ready for review.

```python
>>> from hyperspy.misc.eels.tools import get_edges_near_energy
>>> get_edges_near_energy(532)
['O_K', 'Pd_M3', 'Sb_M5', 'Sb_M4']
```

### DIFF
--- a/doc/user_guide/eels.rst
+++ b/doc/user_guide/eels.rst
@@ -50,14 +50,6 @@ they are arranged in the order closest to 849 eV.
     >>> get_edges_near_energy(849, width=6)
     ['La_M4', 'Fe_L1']
 
-The same functionality is also available as a method in `EELS` signal.
-
-.. code-block:: python
-
-    >>> import hyperspy.api as hs
-    >>> s = hs.datasets.artificial_data.get_core_loss_eels_signal()
-    >>> s.get_edges_near_energy(532)
-    ['O_K', 'Pd_M3', 'Sb_M5', 'Sb_M4']
 
 Thickness estimation
 ^^^^^^^^^^^^^^^^^^^^
--- a/doc/user_guide/eels.rst
+++ b/doc/user_guide/eels.rst
@@ -50,6 +50,14 @@ they are arranged in the order closest to 849 eV.
     >>> get_edges_near_energy(849, width=6)
     ['La_M4', 'Fe_L1']
 
+The same functionality is also available as a method in `EELS` signal.
+
+.. code-block:: python
+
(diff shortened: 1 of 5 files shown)


ericpre:
Indeed, from a programming perspective, it makes sense to have this as a standalone function, but from the user perspective, it also convenient method to have for a EELS signal... 
Looking at its eds counterpart, it is defined as a standalone function and there is no convenience method for EDS signal.
I would be tempted to use both approaches: a standalone function and a convenience method. This may change anyway after the split. @francisco-dlp, what do you think?

ptim0626:
@ericpre Yes, it is always better to have more options to call the function from a user perspective. Perhaps I can also include the same method in the EELS signal class, which just call the standalone function?

francisco-dlp:
The function in `misc.eels.tools` looks good. However, I think that any function in the API or method in the signal should be more user friendly and complete. In particular, it could be similar to Gatan's [EELS.info](https://eels.info/atlas?major_edge_energy=300&name_symbol=&tabs%5Btabs__active_tab%5D=edit-tab-energy&reset=Reset#views-exposed-form-eels-atlas-atlas-page) one: instead of returning a list of edges, it could print a table with all the information needed to identify the edges.

What do you think?

ptim0626:
@francisco-dlp Yes I agree. The intention of this function/method is to emulate the one in EDS, but I could add an option to print out a text table for the method in the API/signal, including whether it is a major or minor edge. The information about major/minor edge is in the `elements_db`, but not whether the peak is sharp/broad or abrupt/delayed/very delayed, which are available in EELS.info (e.g. [https://eels.info/atlas/calcium](https://eels.info/atlas/calcium)). Do you think it is also helpful to include those sharp/broad/abrupt/delayed/very delayed information in the table?

francisco-dlp:
I think that users would find that information useful indeed. However, because a release is planned for this week, I suggest to leave this more ambitious changes for a different PR and merge  this one after reverting 76595dc  and 4cfe225 .

Issue #2320, opened 2020-02-20
estimate_peak_width(parallel=True) giving rise to nans


thomasaarholt:
The estimate_peak_width gives a random result of `nan`s when called with `parallel=True` (calling it again yields `nan`s in different places). When called with parallel=False it runs fine. 

[low-loss.zip](https://github.com/hyperspy/hyperspy/files/4229848/low-loss.zip), provided by @JMorzy.

```python
import hyperspy.api as hs
import numpy as np

s = hs.load("low-loss.hspy").inav[0, :]

print("Estimating ZLP width on spectrum image")
print(s.estimate_peak_width(parallel=True).data)

print()
print("Estimating ZLP width on each pixel at a time")
data = []
for i in s.axes_manager:
    data.append(s.inav[i].estimate_peak_width().data[0])
print(np.array(data))
```
![image](https://user-images.githubusercontent.com/2721423/74919748-594e6d80-53cb-11ea-940c-a8e75eb576bd.png)

I've traced the error to [line 1447 in signal1d.py](https://github.com/hyperspy/hyperspy/blob/49fa05c96ca0619615eeccc077f14ed51f1481f1/hyperspy/_signals/signal1d.py#L1447).

thomasaarholt:
The smaller I make the window, the fewer nans there are, until there are none appearing on a regular basis. `print(s.estimate_peak_width(window=2000).data)`. With window=None, the entire signal is used, otherwise the signal is of `window` length, centered around the ZLP.


Edit: ugh. The return value of `estimating_function` is:
```python
if len(roots) == 2:
    return np.array(roots)
else:
    return np.full((2,), np.nan)
```
This explains the existance of the nans, but doesn't explain why they only appear with `parallel=True`

tjof2:
@thomasaarholt  I don't see this using your example above - is it missing an argument to reproduce it? `hyperspy.__version__` returns `1.5.3.dev` for me, and if I do `s.estimate_peak_width(parallel=True).data` it works fine (no nans)

thomasaarholt:
The example should work as-is. You were right to explicitly state `parallel=True`, though. Modifying in case people have changed their preferences. I just reran it now to double check, and in two out of three cases it contains nans. I'm on `'1.6.dev'`.

tjof2:
@thomasaarholt  I still don't see this issue on `1.6.dev` - I ran your example 10 times and it worked each time (no nans)

[...]

PR #2391, opened 2020-05-07
Improve thread-based parallelism behaviour

### LINKED ISSUE #970: align2D
Has the documentation for align2D completely disappeared, or am I looking in completely the wrong places?! :)

If/once found, it might be helpful to have a short example useage. (The previous documentation was a little unclear).

tjof2:
### Description of the change
@dnjohnstone mentioned in https://github.com/pyxem/pyxem/issues/585#issuecomment-611554910 that

> Ah - yes that's exactly an issue that recently came up when we tried to use center_direct_beam() on the Diamond Light Source cluster... The massive thread spawning played horrifically with the job managers on the cluster and it led to this method taking many hours to complete rather than being accelerated by the additional computing resource.

This PR exposes the `max_workers` argument of `ThreadPoolExecutor` such that the user can directly control how many threads they launch. It also changes the default value (when `max_workers=None`) from `os.cpu_count()` to `min(32, os.cpu_count()` as per the [documentation here](https://docs.python.org/3/library/concurrent.futures.html#concurrent.futures.ThreadPoolExecutor) so that the risk of lots of threads being spawned on a large multicore machine is reduced.

I've added the kwarg to every instance of `s.map()` and `s._map_iterate()` and updated the docstrings.

I've also added a section in the docs to note that HyperSpy uses *thread-based parallelism* for `.map()`, unlike e.g. Samfire which uses *process-based parallelism*. @thomasaarholt noted in https://github.com/hyperspy/hyperspy/issues/2320#issuecomment-611433876 that sometimes threading offers no speed-up over serial, but it's a little involved to switch to process-based parallelism since it depends on identifying which functions release the GIL and out of the scope of this PR. I should add that for `align2D` and `estimate_peak_width`, I definitely do see a speed improvement over `parallel=False`.

Finally, I expanded the docs on `align2D()` a little, which closes #970.

### Progress of the PR
- [x] add tests,
- [x] ready for review.

```python
# Image alignment
>>> s.align2D(parallel=True, max_workers=4)

# Peak width estimation
>>> s.estimate_peak_width(parallel=True, max_workers=4)
```


dnjohnstone:
@mcduta @M0hsend linked to Ticket: DLS-22584

tjof2:
@dnjohnstone ultimately this new argument needs to be balanced with environment variables like `OMP_NUM_THREADS` and `MKL_NUM_THREADS`. As I wrote, the best way forward would be to identify which functions work best with threads/processes and then look at using joblib/dask instead. But that's a huge amount of work, and the level of control from this PR is better than currently available.

tjof2:
Fixed a very minor bug - now ready for review.

PR #2393, opened 2020-05-08
Fix import error with scikit-image 0.17.0

ericpre:
There is an API changes in the last release of scikit-image:
https://github.com/scikit-image/scikit-image/blob/master/doc/release/release_0.17.rst#api-changes

### Progress of the PR
- [x] make import of `_upsampled_dft` compatible with scikit-image 0.17
- [x] ready for review.

### DIFF
--- a/hyperspy/_signals/signal2d.py
+++ b/hyperspy/_signals/signal2d.py
@@ -22,7 +22,11 @@ import numpy.ma as ma
 import dask.array as da
 import scipy as sp
 import logging
-from skimage.feature.register_translation import _upsampled_dft
+try:
+    # For scikit-image >= 0.17.0
+    from skimage.registration._phase_cross_correlation import _upsampled_dft
+except ModuleNotFoundError:
+    from skimage.feature.register_translation import _upsampled_dft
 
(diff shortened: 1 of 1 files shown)


hyperspy/_signals/signal2d.py
```diff
@@ -22,7 +22,11 @@
 import dask.array as da
 import scipy as sp
 import logging
-from skimage.feature.register_translation import _upsampled_dft
+try:
+    # For scikit-image >= 0.17.0
+    from skimage.registration._phase_cross_correlation import _upsampled_dft
+except ModuleNotFoundError:
```
tjof2, line 28: Why not `ImportError`?

ericpre: This module doesn't exist in previous version.

tjof2: More that `ModuleNotFoundError` is a sub-class of `ImportError` but not that it really matters.

tjof2: Looks good to go!

tjof2 (approved):

PR #2307, opened 2020-01-19
Remove background return model

ericpre:
Close #2207
Require https://github.com/hyperspy/hyperspy_gui_ipywidgets/pull/26 and https://github.com/hyperspy/hyperspy_gui_traitsui/pull/27

### Progress of the PR
- [x] Return model when using `remove_background` 
- [x] Display chi2 in the UI,
- [x] Update fit when switching fast on/off,
- [x] fix connection on_close of Signal1DLine,
- [x] fix some deprecation warning,
- [x] fix issue with removing backgroung using polynomial component of signal with navigation dimension 0,
- [x] avoid reopening the plot when using it is already open,
- [x] Calculate chi2 when using "fast" (no full fit),
- [x] add tests,
- [x] ready for review.

### Minimal example of the bug fix or the new feature
```python
import hyperspy.api as hs
s = hs.datasets.artificial_data.get_core_loss_eels_signal(True)
m = s.remove_background(return_model=True)
m.print_current_values()
```
Note that this example can be useful to update the user guide.

PR #2262, opened 2019-10-08
Convert heaviside step function component to expression

jlaehne:
### Description of the change
Conversion of the heaviside step function component to an expression-based definition. As expression allows for only one `where`, this version follows the alternative definition without intermediate value at `x=n`. 

### Progress of the PR
- [X] Change implemented,
- [X] update docstring,
- [x] add tests,
- [x] ready for review.

### Minimal example of the bug fix or the new feature
```python
>>> import hyperspy.api as hs
>>> h1 = hs.model.components1D.HeavisideStep(A=3,n=2)
>>> x = np.arange(0,5,step=0.01)
>>> s = hs.signals.Signal1D(h1.function(x))
>>> s.axes_manager.signal_axes[0].axis = x
>>> s.plot()
>>> h1.print_current_values()
>>> # Your new feature...
```
Note that this example can be useful to update the user guide.

PR #2395, opened 2020-05-12
Implemented Arctan + EELSArctan as expressions

jlaehne:
### Description of the change
Implemented Arctan as Expression as proposed in https://github.com/hyperspy/hyperspy/issues/2251#issuecomment-529929541
* New component `EELSArctan` replaces `minimum_at_zero=True` option
* `Arctan` component will default to the former `minimum_at_zero=False`
* legacy feature supports old way of calling the functions (up to v2?) and raises warning if `Arctan` is called with `minimum_at_zero=True` 

### Progress of the PR
- [X] add tests,
- [X] ready for review.

```python
>>> import hyperspy.api as hs
>>> import numpy as np
>>> arctan1 = hs.model.components1D.Arctan(A=10., k=5, x0=1, minimum_at_zero=True)
>>> x = np.arange(arctan1.x0.value-5,arctan1.x0.value+5,step=0.01)
>>> s = hs.signals.Signal1D(arctan1.function(x))
>>> s.axes_manager.signal_axes[0].axis = x
>>> s.plot()
```

### DIFF
--- a/doc/user_guide/model.rst
+++ b/doc/user_guide/model.rst
@@ -103,6 +103,7 @@ The following general components are currently available for one-dimensional mod
 
 The following components developed with specific signal types in mind are currently available for one-dimensional models:
 
+* :py:class:`~._components.eels_arctan.EELSArctan`
 * :py:class:`~._components.eels_double_power_law.DoublePowerLaw`
 * :py:class:`~._components.eels_cl_edge.EELSCLEdge`
 * :py:class:`~._components.pes_core_line_shape.PESCoreLineShape`
--- a/hyperspy/_components/arctan.py
+++ b/hyperspy/_components/arctan.py
@@ -29,7 +29,8 @@ class Arctan(Expression):
 
     .. math::
     
-        f(x) = A \cdot \arctan\left[ k \left( x-x_0 \right)\right]
+        f(x) = A \cdot \arctan \left[ k \left( x-x_0 \right) \right]
+
 
     ============ =============
     Variable      Parameter 
@@ -39,11 +40,17 @@ class Arctan(Expression):
     :math:`x_0`   x0 
     ============ =============
 
+
     Parameters
     -----------
     A : float
+        Amplitude parameter. :math:`\lim_{x\to -\infty}f(x)=-A` and 
+        :math:`\lim_{x\to\infty}f(x)=A`
     k : float
+        Slope (steepness of the step). The larger :math:`k`, the sharper the 
+        step.
     x0 : float
+        Center parameter (position of zero crossing :math:`f(x_0)=0`).
 
     """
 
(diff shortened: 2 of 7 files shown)


jlaehne:
No idea, why I can't get rid of the test_bss errors in the checks.

ericpre:
This looks good to me, except for the bss test failures, which I don't understand neither where it comes from... `RELEASE_next_minor` is fine, merging it with this branch is not! @francisco-dlp, it seems the same thing happen in #2299, any idea what it is?

jlaehne:
But in #2299 the tests ran through now, I tried to copy `test_bss.py` from several other branches where it was fixed -- to no avail. That strategy worked in #2262 though. So I'm kind of puzzled.

tjof2:
Perhaps sklearn version now it's 0.23? Not sure really.

jlaehne:
It is really weird that it fails in some PRs and not in others - all of which are not related to bss. Maybe there is some other file than the test file that has changed? But I have merged 'RELEASE_next_minor' both into this one and into #2262.

tjof2:
Something may no longer be deterministic somewhere? @francisco-dlp

ericpre:
> It is really weird that it fails in some PRs and not in others - all of which are not related to bss. Maybe there is some other file than the test file that has changed? But I have merged 'RELEASE_next_minor' both into this one and into #2262.

Indeed, it is behaving very weirdly... typically, the issue doesn't occurs in https://github.com/hyperspy/hyperspy/pull/2307 and it has been kept up to date with `RELEASE_next_minor`!

> Perhaps sklearn version now it's 0.23? Not sure really.

sklearn is an optional dependencies and the "minimum requirement" build on travis without sklearn is failing too...


> Something may no longer be deterministic somewhere? @francisco-dlp

It consistently passes or fails... very puzzling...

ericpre:
I have rebased this branch (https://github.com/ericpre/hyperspy/commits/Arctan_Expression) and remove some of the last commits recently to check if the test suite pass and it still fails in the same way. All other branches pass fine...

tjof2:
@ericpre and @jlaehne  I think I know what the issue is - I can create a separate PR to fix the BSS tests to remove their dependency on this code change. See #2430 for the fix that will stop the tests failing here.

That said, the reason the tests fail is actually because this PR affects the values in the artificial datasets such as  `get_core_loss_eels_line_scan_signal()`, i.e. https://github.com/hyperspy/hyperspy/blob/bf7a3654d459e34ef5a3b0e01933589019b4a44f/hyperspy/datasets/artificial_data.py#L180

I would therefore check that this behaviour is intentional - clearly HyperSpy doesn't have a test to compare the artificial datasets (@francisco-dlp ?)

jlaehne:
Continued in  #2432

Issue #2398, opened 2020-05-19
Non-uniform axes progress tracking


francisco-dlp:
This issue tracks progress on the implementation of non-uniform axes.

Notice that, currently, the non-uniform axes features are merged into the `non_uniform_axes` branch.

If you intend to address any of the issues below, please discuss it here to avoid overlapping of efforts.

## Key issues to address before release

- [x] Plot works properly for binned signals and images (see #2728)
- [x] Binned versions of the axes are implemented
- [x] Saving the axes in the hspy format is implemented
- [x] Curve fitting on binned signals works (see #2728)
- [x] Most things that don't work on non-uniform axes are identified and an informative `NotImplementedError` is raised (see #2761 and https://github.com/hyperspy/hyperspy/issues/2398#issuecomment-853632828)
- [x] Expand the documentation on ``Signal.binned`` (here  ``DataAxis.is_binned``?) as it is a common source of confusion, see https://hyperspy--2399.org.readthedocs.build/en/2399/user_guide/signal.html#binned-and-unbinned-signals

## Issues that would be nice fixing before releasing, but should not hold a release: 

- [x] Use `pcolormesh` for image plots to allow for non uniform axes - (#3192)
- [ ] Implement `fft` and `ifft` using `nfft`. I like https://github.com/jakevdp/nfft
- [ ] `split` (partially works)
- [x] `stack` (also along non-uniform axes, done in #2728)
- [ ] `rebin` (#3410 is an alternative to rebinning non-uniform axes)
- [x] `remove_background` for "binned" signals. It doesn't work because the `estimate_parameters` methods assume linear binned axes.  (done in #2728)
- [ ] (some) ROIs:
    - [ ] `RectangularROI`
    - [ ] `SpanROI`
    - [ ] `Point2DROI`
- [x] Docstrings for all the axes classes (done in #2693)
- [x] Improve axes_manager documentation (based on points raised in #2345, at least partially implemented already e.g. in #2304)
- [x] Adapt get_axis_dictionary so that for FunctionalDataAxis it gives a nested dictionary. (done in #2304)
- [x] Add the possibility to use such a dictionary for FDA in the initialization of a signal. (done in #2304)
- [x] Introduce possibility to initialize both Signal and AxesManager by passing an axis object in place of a dictionary (would make it easier to copy an axis from one signal to another) (done in #2304)
- [x] `_create_axis` should become a public instead of a private method (done in #2304)
[...]

PR #2399, opened 2020-05-19
Non uniform axes

francisco-dlp:
See #2398 for a tracking issue. Checkout the hyperspy branch [`non_uniform_axes`](https://github.com/hyperspy/hyperspy/tree/non_uniform_axes) for testing.


tjof2:
Not sure why there are three LazySignal BSS tests that fail only on Python 3.7 on Travis...

```python
AttributeError: 'numpy.ndarray' object has no attribute 'compute'
```

Clearly a mix-up between numpy and dask somewhere

thomasaarholt:
How is this PR/branch doing? I'm looking for somewhere in the docs to place the "generalised iterpath" documentation from #2459, and noticed that this PR brings in a new `axes.rst` doc page. Am I better off waiting a few more weeks and using that file when this is merged, or should I crack on and find a place in `getting_started.rst/`signal.rst`?

Edit: Or should I rebase off this PR and make my changes based off this? I assume there'll be some merge conflicts with this anyway?

francisco-dlp:
Your last option sounds good.

jlaehne:
See #2398 for progress tracking.

hyperspy/_signals/signal1d.py
```diff
@@ -281,11 +281,13 @@ def _get_spikes_diagnosis_histogram_data(self, signal_mask=None,
                                              **kwargs):
         self._check_signal_dimension_equals_one()
         dc = self.data
```
jlaehne, line 276: This seems to be where the test failure comes from. @francisco-dlp do you have an idea, why this function differs from the one in RnM?

jlaehne (commented):
Trying to find the reason for the failure of `test_eels.py`. The function `_get_spikes_diagnosis_histogram_data` in `signal1d.py` differs from RnM. I guess that is the reason, but I do not understand yet why it was changed in this branch?

hyperspy/_signals/signal1d.py
jlaehne: `signal1d.py` was edited by @francisco-dlp 11 months ago: https://github.com/hyperspy/hyperspy/commit/8f696db649a4a9bfbb8ff76a178c297094f48662 (Fix spikes_removal_tool for non-linear axes)

The test failure in this branch was introduced by changes of @ericpre to `test_eels.py`: https://github.com/hyperspy/hyperspy/commit/bb236ef2f793805c37a7d2e2c749ff1b74982532 (Check zero loss in energy range and fix test)

The issue seems to be that the `self.signal.spikes_diagnosis(zero_loss_peak_mask_width=5.0)` in the test does not give the same results om RnM and nua branches.

ericpre: I suspect the derivative is not calculating correctly and this would be consistent with the failure of one of the bss test.

ericpre: The test failure was caused by the changes in the derivative which now uses `np.gradient` and this particular test was picking up this small difference. This is now fixed.

ericpre:
> Not sure why there are three LazySignal BSS tests that fail only on Python 3.7 on Travis...
> 
> ```python
> AttributeError: 'numpy.ndarray' object has no attribute 'compute'
> ```
> 
> Clearly a mix-up between numpy and dask somewhere

Similar as with https://github.com/hyperspy/hyperspy/pull/2399#discussion_r549014833, but even more subtle... this was caused by changes in the `derivative` method, which doesn't work for lazy signal *and* with numpy<1.17; more recent version works fine because of the array protocol and the dask function is picked up instead of the numpy one. To keep it simple, the test is now skipped, because numpy 1.16 *can be dropped* from January 2021 https://numpy.org/neps/nep-0029-deprecation_policy.html#drop-schedule.

doc/user_guide/axes.rst
```diff
@@ -0,0 +1,649 @@
...
+.. NOTE::
+
+    Not all features are implemented for non-uniform axes.
+
+
+Uniform data axis
+^^^^^^^^^^^^^^^^^
+
+The most common case is the :py:class:`~.axes.UniformDataAxis`. Here, the axis
+is defined by the ``offset`` and ``scale`` parameters, which determine the
```
thomasaarholt: I believe that the axis is also "defined by" its size, in addition to offset and scale.
I'd like to emphasize in the docs that the size is a required parameter when instantiating an axis object.

thomasaarholt:
Is anything missing to merge this?

ericpre:
> Is anything missing to merge this?

Yes, see https://github.com/hyperspy/hyperspy/issues/2398#issue-620896331

hyperspy/axes.py
```diff
@@ -668,6 +633,483 @@ def update_from(self, axis, attributes=["scale", "offset", "units"]):
...
+        return my_slice
+
+
+class UniformDataAxis(BaseDataAxis, UnitConversion):
+    def __init__(self,
+                 index_in_array=None,
+                 name=t.Undefined,
+                 units=t.Undefined,
+                 navigate=t.Undefined,
+                 size=1.,
```
thomasaarholt: The size parameter should be an int, not float.

jlaehne: see #2693

doc/user_guide/axes.rst
jlaehne: see #2693

hyperspy/io.py
```diff
@@ -782,10 +782,40 @@ def save(filename, signal, overwrite=None, **kwds):
     elif overwrite is False and is_file:
         write = False  # Don't write the file
     else:
```
jlaehne, line 792: When looking for missing coverage, I stumbled across the following segment. I do not really see, when the `else` statement would actually be called - except for when `overwrite` gets an invalid input, i.e. neither `bool` nor `None`. And then it partially repeats content from above.

From the history, this has been around fairly long @francisco-dlp , @ericpre

ericpre: This looks fine to me; it raises an error, when the input is not valid:
```python
import hyperspy.api as hs

s = hs.signals.Signal1D([0, 1, 2])
s.save('file.hspy', overwrite=0)
```
give the following error:
```python

  File "<ipython-input-1-f3574f48584c>", line 4, in <module>
    s.save('file.hspy', overwrite=0)

  File "/home/eric/Dev/hyperspy/hyperspy/signal.py", line 2840, in save

  File "/home/eric/Dev/hyperspy/hyperspy/io.py", line 776, in save
    raise ValueError(

ValueError: `overwrite` parameter can only be None, True or False.
```

jlaehne: But in what case is all the code following the `else:` statement executed then?

ericpre: Ok, I see what you meant now, I was looking at the wrong `else` block... sorry! It looks like a the `RELEASE_next_minor` branch haven't been merged correctly after https://github.com/hyperspy/hyperspy/pull/2537 but I can't find the exact commit to be sure.
I agree with you, this doesn't make sense!

jlaehne: Corrected in #2761

ericpre (approved):
As discussed in #2398, this should be good to merge!

jlaehne (approved):

ericpre:
@francisco-dlp, are you happy with getting this PR merged?

The test suite is failing because of the release of pillow 8.3: https://github.com/python-pillow/Pillow/issues/5571 and it is not related to this PR?

francisco-dlp:
Unfortunately I won't have time for a detailed review in the next few weeks. I have just had a quick look and tested some features and it looks great to me. There are of course things to polish and features to add, but I think that the current state is good enough to expose the feature to users.

PR #2403, opened 2020-05-26
Fix test broken with sympy >=1.6

francisco-dlp:
Fix broken warning with sympy v1.6.


francisco-dlp:
@ericpre, could you have a quick look at this so that we can fix the coverage report in #2203 among others?

PR #2405, opened 2020-05-27
Address various warnings in test suite

tjof2:
### Description of the change
There are a few warnings polluting the `pytest` report that are easily solved on `RELEASE_next_minor` - this PR address that, such that the remaining warnings are clearer for solving in future.

- Fix NMF convergence warnings and factor out data generation
- Fix H5py deprecation warning
- Fix Voigt API warnings
- Fix warnings from strings

I also looked at correctly raising warnings for parameter covariance estimation as per the approach taken in [`scipy.optimize.curve_fit()`](https://github.com/scipy/scipy/blob/v1.4.1/scipy/optimize/minpack.py#L511-L813) - rather than raising the unclear:
```
hyperspy/model.py:1134: RuntimeWarning: invalid value encountered in sqrt
    self.p_std = np.sqrt(np.diag(pcov))
```
when the parameter covariance is ill-defined, it now raises:
```
hyperspy/model.py:1162: OptimizeWarning: Covariance of the parameters could not be estimated
    warnings.warn("Covariance of the parameters could not be estimated", OptimizeWarning)
```
As it stands, this last point actually greatly _increases_ the number of warnings in the test suite. @ericpre / @francisco-dlp I would welcome a suggestion on how to proceed (for example, filtering out this warning if desired).

### Progress of the PR
- [x] Change implemented
- [x] ready for review

### DIFF
--- a/hyperspy/model.py
+++ b/hyperspy/model.py
@@ -29,7 +29,7 @@ import dill
 import scipy
 import scipy.odr as odr
 from scipy.optimize import (leastsq, least_squares,
-                            minimize, differential_evolution)
+                            minimize, differential_evolution, OptimizeWarning)
 from scipy.linalg import svd
 from contextlib import contextmanager
 
@@ -1115,16 +1115,16 @@ class BaseModel(list):
                     ls_b = self.free_parameters_boundaries
                     ls_b = ([a if a is not None else -np.inf for a, b in ls_b],
                             [b if b is not None else np.inf for a, b in ls_b])
-                    output = \
-                        least_squares(self._errfunc, self.p0[:],
-                                      args=args, bounds=ls_b, **kwargs)
+                    output = least_squares(self._errfunc, self.p0[:],
+                                           args=args, bounds=ls_b, **kwargs)
                     self.p0 = output.x
+                    ysize = len(output.fun)
+                    cost = 2 * output.cost  # res.cost is half sum of squares
 
                     # Do Moore-Penrose inverse, discarding zero singular values
                     # to get pcov (as per scipy.optimize.curve_fit())
                     _, s, VT = svd(output.jac, full_matrices=False)
-                    threshold = np.finfo(float).eps * \
-                        max(output.jac.shape) * s[0]
+                    threshold = np.finfo(float).eps * max(output.jac.shape) * s[0]
                     s = s[s > threshold]
                     VT = VT[:s.size]
                     pcov = np.dot(VT.T / s**2, VT)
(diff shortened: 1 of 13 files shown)


ericpre:
Most of them should be related to inappropriate fitting with signal not being suitable to the model?
- either the test needs to be defined in a better way,
- filter the warning if the results of the fit is not relevant?

tjof2:
Yes it seems like that's true. I can go through and look to see if the parameter standard deviation is used in the test, and if not, filter?

tjof2:
> Yes it seems like that's true. I can go through and look to see if the parameter standard deviation is used in the test, and if not, filter?

Rather than changing or going through 200+ tests, using `_logger.warning()` rather than `warnings.warn()` is simpler. Also, it will be printed multiple times by default, while `warnings.warn()` might only show once when in fact it occurs on every `.fit()` call. I will add a specific test for the behaviour though.

hyperspy/model.py
```diff
@@ -1133,19 +1132,37 @@ def fit(self, fitter="leastsq", method='ls', grad=False,
...
+                    pcov.fill(np.inf)
+                    warn_cov = True
+                elif (ysize > self.p0.size):
+                    pcov *= cost / (ysize - self.p0.size)
+                else:
+                    pcov.fill(np.inf)
+                    warn_cov = True
+
+                if warn_cov:
+                    _logger.warning("Covariance of the parameters could not be estimated")
```
ericpre: Even if this warning message is definitely better than the previous ones, would it be possible to be more specific for each of the cases above? I have no idea if this is possible/sensible or not.
Also what does it implies will not be clear for many users: is it worth adding that the error of the parameters will be `inf`? And that the model doesn't fit well the data?

tjof2: Yes I totally agree - I will be more specific with the warning message raised - doing that now.

tjof2: @ericpre  that's updated now.

tjof2: > Even if this warning message is definitely better than the previous ones, would it be possible to be more specific for each of the cases above? I have no idea if this is possible/sensible or not.

I've not created specific warnings since they all amount to the same thing - for example, `pcov.min() < 0` usually happens due to an ill-conditioned matrix inversion within the scipy MINPACK `leastsq` function leading to overflow, but this is effectively the same reason as `pcov = None` - the Hessian is singular, so can't be inverted.

tjof2:
Test failure on Travis is unrelated - only Python 3.7 failed in `TestUSID2HSbase.test_0_pos_n_spec`:
```bash
E       AssertionError: assert {'machine_id'..._28-08_53_02'} == {'machine_id':..._28-08_53_01'}
E         Differing items:
E         {'timestamp': '2020_05_28-08_53_02'} != {'timestamp': '2020_05_28-08_53_01'}
```
Otherwise, it's reduced the number of warnings from ~80 to ~30 on Travis.

ericpre:
This looks good to me. Do you also want to filter the `test_usid.py` warning? This warning comes from the pyUSID library and it should be fine to filter it.

> Test failure on Travis is unrelated - only Python 3.7 failed in `TestUSID2HSbase.test_0_pos_n_spec`:
> 
> ```shell
> E       AssertionError: assert {'machine_id'..._28-08_53_02'} == {'machine_id':..._28-08_53_01'}
> E         Differing items:
> E         {'timestamp': '2020_05_28-08_53_02'} != {'timestamp': '2020_05_28-08_53_01'}
> ```
> 
> Otherwise, it's reduced the number of warnings from ~80 to ~30 on Travis.

This error with the timestamp does occur once in a while... :/

tjof2:
Sure @ericpre - that's pushed - now below 10 warnings on travis and azure.

ericpre (approved):

ericpre:
Thanks, these warnings were annoying me for a while!!

PR #2410, opened 2020-06-08
.sur and .pro files support

LMSC-NTappy:
### Description of the change
Implements read support for a new type of files, .sur and .pro, for compatibility with the DigitalSurf Mountainsmap software. Notably used by attolight for SEM Cathodoluminescence spectral maps but also for some AFM data, spectra, multi-layer surfaces etc.
-supports metadata parsing (original_metadata as well),
-implemented as io_plugin (vastly adapted from the digitalmicrograph plugin)
-write supports not in the pipeline,

### Progress of the PR
- [x] add tests,
- [x] ready for review.

```python
>>> import hyperspy.api as hs
>>> s = hs.load('hyperspy/tests/io/sur_data/test_spectral_map.sur')
>>> s.plot()
```

Please, take note that this is my first PR ever so indulgence regarding coding style or incomplete documentation etc. is kindly requested. :)


ericpre (changes requested):
Thanks @LMSC-NTappy for this PR!
A couple of quick comments, before going into more details:
- test files needs to be smaller
- in the test suite, may be good to check for the dimension (signal, navigation)
- add entry in the user guide describing this format and what is supported by this reader
- add support for lazy loading?
- nice to see many comments in the reader!

Can you merge/rebase with upstream and solve the conflict, so that we can see if the test suite pass?

LMSC-NTappy:
Thanks @ericpre for the comments.

I addressed all your requested changes but the lazy loading support since it is a way bigger change than the rest and I haven't fully understood how it works yet.

I will try to support it in a further update

doc/user_guide/io.rst
```diff
@@ -1036,6 +1038,24 @@ HDF5 file. All other keyword arguments will be passed to
...
+This is a format developed by the digitalsurf company to handle various types of
+scientific measurements data such as profilometer,SEM,AFM,RGB(A) images, multilayer
+surfaces and profiles. Even though it is essentially a surfaces format, 1D signals
+are supported for spectra and spectral maps. Metadata parsing is supported, including
+user-customised metadata, as well as the loading of files containing multiple objects
+packed together.
+
+Lazy loading planned but not implemented.
+
+Write support not planned but would be considered if requested.
```
ericpre: I would say, that this is not necessary to mention this in the user guide, but this can stay, just that I am not really convinced that this is really relevant for users - the supported file format table already inform the user that lazy loading and writing is not supported.

hyperspy/io_plugins/sur.py
```diff
@@ -0,0 +1,1298 @@
...
+    """
+
+    ds = DigitalSurfHandler(filename)
+
+    ds._read_sur_file()
+
+    surdict = ds._build_sur_dict()
+
+    return [surdict,]
+    file_reader.__doc__ %= (OPTIMIZE_ARG.replace('False', 'True'))
```
ericpre: Since there is no `OPTIMIZE_ARG` Parameters in the docstring, this can be removed.

hyperspy/io_plugins/sur.py
```diff
@@ -0,0 +1,1298 @@
...
+        #Hence they will be made attributes.
+        #Object type
+        self._Object_type = "_UNKNOWN"
+
+        #Number of data objects in the file.
+        self._N_data_object = 1
+        self._N_data_channels = 1
+
+        #Boolean that sets to true once a .sur file was correctly initialised
+        self._initialized = True
```
ericpre: If this is not used anywhere, better to remove it.

hyperspy/io_plugins/sur.py
```diff
@@ -0,0 +1,1298 @@
...
+
+    ### Read methods
+    def _read_sur_file(self):
+        """Read the binary, possibly compressed, content of the surface
+        file. Surface files can be encoded as single or a succession
+        of objects. Accordingly the  """
```
ericpre: Sentences not completed?

hyperspy/io_plugins/sur.py
```diff
@@ -0,0 +1,1298 @@
...
+            unpacked_dict['_41_Minutes'],
+            unpacked_dict['_40_Seconds'])
+
+        #Metadata dictionary initialization
+        metadict = {
+            "General":{
+                "authors": unpacked_dict['_07_Operator_Name'],
+                "date":date_str,
+                "original_filename": os.path.split(self.filename)[1],
+                "time": time_str,
```
ericpre, line 933: Add the date and time only when not different from zero? Currently, it displays the following:
```
├── General
│   ├── authors = 
│   ├── date =    0- 0- 0
│   ├── original_filename = test_profile.pro
│   ├── time =  0: 0: 0
│   └── title = 
└── Signal
    ├── binned = False
    ├── quantity = CL Intensity a.u.
    └── signal_type = 
```

hyperspy/io_plugins/sur.py
```diff
@@ -0,0 +1,1298 @@
...
+                },
+            "Signal": {
+                "binned": False,
+                "quantity": quantity_str,
+                "signal_type": "",
```
ericpre, line 938: Is it possible to set a `signal_type`? This would assign the data to a specific class. See `hs.print_known_signal_types` for available `signal_type`.

hyperspy/io_plugins/sur.py
```diff
@@ -0,0 +1,1298 @@
...
+        self._initialized = True
+
+    ### Read methods
+    def _read_sur_file(self):
+        """Read the binary, possibly compressed, content of the surface
+        file. Surface files can be encoded as single or a succession
+        of objects. Accordingly the  """
+
+
+        if self.filename is not None:
```
ericpre: It should be possible to remove this `if`.

hyperspy/io_plugins/sur.py
```diff
@@ -0,0 +1,1298 @@
...
+    def _build_RGB_image(self,):
+        """Build an RGB image. The T axis is navigation and set from
+        P Size"""
+
+        #First object dictionary
+        hypdic = self._list_sur_file_content[0]
+
+        #Metadata are set from first dictionary
+        self.signal_dict['metadata'] = self._build_metadata(hypdic)
+        self.signal_dict['original_metadata'] = self._build_original_metadata()
```
ericpre: This sequence is repeated a couple of times, can it be moved to its own method?

hyperspy/tests/io/test_sur.py
```diff
@@ -0,0 +1,826 @@
...
+    md = s.metadata
+    assert md.Signal.quantity == 'Z'
+
+    omd = s.original_metadata
+    assert list(omd.as_dictionary().keys()) == ['Object_0_Channel_0',
+                                                'Object_0_Channel_1',
+                                                'Object_0_Channel_2']
+    assert list(omd.Object_0_Channel_0.as_dictionary().keys()) == ['Header']
+    assert list(omd.Object_0_Channel_0.Header.as_dictionary().keys()) == \
+        ['H01_Signature',
```
ericpre: Is it possible to make a list which can be used in different tests? This would be make the tests more readable.

hyperspy/tests/io/test_sur.py
```diff
@@ -0,0 +1,826 @@
...
+def test_load_spectra():
+    fname = os.path.join(MY_PATH, "sur_data",
+                         "test_spectra.pro")
+    s = load(fname)
+
+    assert s.data.shape == (65, 512)
+    assert s.data.dtype == dtype('float64')
+
+    md = s.metadata
+    assert md.Signal.quantity == 'CL Intensity a.u.'
```
ericpre: Put units in round backets? See http://hyperspy.org/hyperspy-doc/current/user_guide/metadata_structure.html#signal.

hyperspy/tests/io/test_sur.py
```diff
@@ -0,0 +1,826 @@
...
+         'H58_T_Axis_Name',
+         'H59_T_Step_Unit',
+         'H60_Comment',
+         ]
+
+def test_load_spectrum():
+    fname = os.path.join(MY_PATH, "sur_data",
+                         "test_spectrum.pro")
+    s = load(fname)
+    assert s.data.shape == (1,512)
```
ericpre: Can this shape be squeezed?

hyperspy/tests/io/test_sur.py
```diff
@@ -0,0 +1,826 @@
...
+         'H55_Z_Offset',
+         'H56_T_Spacing',
+         'H57_T_Offset',
+         'H58_T_Axis_Name',
+         'H59_T_Step_Unit',
+         'H60_Comment',
+         ]
+
+def test_load_surface():
+    fname = os.path.join(MY_PATH, "sur_data",
```
ericpre, line 370: Just to be sure, this is CL intensity map? If so, can you confirm that there is not metadata specifying the energy range? This information would be useful to be parsed.

ericpre (changes requested):
This is already in very good shape! I have comments for a couple of details.
It would be good to set the SEM metadata to the `metadata` (http://hyperspy.org/hyperspy-doc/current/user_guide/metadata_structure.html#sem) using the [`mapping` dictionary](https://github.com/hyperspy/hyperspy/blob/RELEASE_next_minor/hyperspy/io_plugins/digital_micrograph.py#L851-L875) which copy metadata from the `original_metadata` dictionary to the `metadata` when loading.

doc/user_guide/io.rst
LMSC-NTappy: I noted your comment and removed it

hyperspy/io_plugins/sur.py
LMSC-NTappy: Removed.

hyperspy/io_plugins/sur.py
LMSC-NTappy: Removed

hyperspy/tests/io/test_sur.py
LMSC-NTappy: this is done. It required a bit of work because in the file format 1D series of spectra and lone spectra are saved as a single type of object

hyperspy/tests/io/test_sur.py
LMSC-NTappy: It actually is not. The surface type is actually mostly used for AFM data. The only data that I could share without concerns are my CL surfaces summed along energy axis, in which case I can confirm you that there is no metadata saving the energy range.

hyperspy/io_plugins/sur.py
LMSC-NTappy: Indeed. I completed it. Thanks for noting!

hyperspy/io_plugins/sur.py
LMSC-NTappy: Done as well

hyperspy/io_plugins/sur.py
LMSC-NTappy: This is a bit difficult for the moment. The problem is that this data format does not make a difference between different spectra, e.g. Raman, CL, PL. This is also true for hyperspectral maps and surfaces (can be a CL image integrated on energy, loadings of a decomposed map etc... ).

Although this could be linked, in some cases, with lumispy types, I must admit that I would rather let users set the signal type themselves

hyperspy/io_plugins/sur.py
LMSC-NTappy: You are absolutely right. I removed it

hyperspy/io_plugins/sur.py
LMSC-NTappy: It is now its own method

hyperspy/tests/io/test_sur.py
LMSC-NTappy: You are perfectly right, I declared all these metadata keys as lists

hyperspy/io_plugins/sur.py
LMSC-NTappy:
```python
├── General
│   ├── authors = 
│   ├── date = 
│   ├── original_filename = test_profile.pro
│   ├── time = 
│   └── title = 
└── Signal
    ├── binned = False
    ├── quantity = CL Intensity (a.u.)
    └── signal_type = 
```

hyperspy/tests/io/test_sur.py
LMSC-NTappy: This is also implemented now. Brackets will always be added, unless the unit string is empty.

LMSC-NTappy:
> It would be good to set the SEM metadata to the `metadata` (http://hyperspy.org/hyperspy-doc/current/user_guide/metadata_structure.html#sem) using the [`mapping` dictionary](https://github.com/hyperspy/hyperspy/blob/RELEASE_next_minor/hyperspy/io_plugins/digital_micrograph.py#L851-L875) which copy metadata from the `original_metadata` dictionary to the `metadata` when loading.

Many thanks for the nice comment! I can do it, but as mentioned before this file format is (in percentage) mostly used for scanning probe microscopes, mechanical profilometers etc... All metadata keys parsed into `original_metadata.Object_0_Channel_0.Parsed` are instrument-manufacturer specific and cannot be assumed to follow any hierarchy or formatting

If I map anything from `original_metadata` into  `metadata.Acquisition_instrument` I have no other choice than doing it assuming that the data comes from one of attolight CL-SEM systems, which won't be generally true. 

Do you think it is a good idea to implement this nevertheless? If you think so I can do it

hyperspy/io_plugins/sur.py
ericpre: Sounds very sensible!

ericpre (approved):

ericpre:
> > It would be good to set the SEM metadata to the `metadata` (http://hyperspy.org/hyperspy-doc/current/user_guide/metadata_structure.html#sem) using the [`mapping` dictionary](https://github.com/hyperspy/hyperspy/blob/RELEASE_next_minor/hyperspy/io_plugins/digital_micrograph.py#L851-L875) which copy metadata from the `original_metadata` dictionary to the `metadata` when loading.
> 
> Many thanks for the nice comment! I can do it, but as mentioned before this file format is (in percentage) mostly used for scanning probe microscopes, mechanical profilometers etc... All metadata keys parsed into `original_metadata.Object_0_Channel_0.Parsed` are instrument-manufacturer specific and cannot be assumed to follow any hierarchy or formatting
> 
> If I map anything from `original_metadata` into `metadata.Acquisition_instrument` I have no other choice than doing it assuming that the data comes from one of attolight CL-SEM systems, which won't be generally true.
> 
> Do you think it is a good idea to implement this nevertheless? If you think so I can do it

This was a suggestion and this can be improved in the future if suitable. Thanks for the PR!

Issue #2409, opened 2020-06-07
Display edges near energy interactively


ericpre:
Following #2407, it would be nice to add an `interactive` option to the `print_edges_near_energy`, which would display the table in a widget (ipwidgets and/or traitsui) and change interactively by using a range widget or a vertical line.

tjof2:
#2418 was merged, does that solve it?

PR #2418, opened 2020-06-11
A method to print edge energy interactively

ptim0626:
This is related to issue #2409. Any suggestion is welcomed.

### Description of the change
An option is provided in `print_edges_near_energy()` to print the table interactively, or the method `interactive_edges_energy()` can be called directly to achieve this.

An option is also added in `print_edges_near_energy()` to show only the major edges if desired.

### Progress of the PR
- [x] Change implemented,
- [x] update docstring,
- [x] update user guide,
- [x] add tests,
- [x] ready for review.

iss2409_sc


francisco-dlp:
That's cool! Could you move the widgets code to https://github.com/hyperspy/hyperspy_gui_ipywidgets? HyperSpy should not depend on ipywidgets.

ericpre:
Nice indeed, for the interactivity part, an alternative may be something along the line what does the `calibrate` method:
Peek 2020-06-11 18-16

francisco-dlp:
@ericpre, that's ultracool :) In this way one can select an energy window an move it around: I think that there is no better in terms of usability but maybe somebody will surprise us.

francisco-dlp:
A small suggestion: instead of the interactive keyword, it could default to it if no arguments are passed to the function.

ericpre:
Having it for EDS signal would be very good too! :)

@ptim0626, please let us know if you need help to find your with the widget registration mechanism, etc.
The `calibrate` widget and method should be a good place to start: it is a bit convoluted and it involves [widgets in hyperspy_gui_ipywidgets](https://github.com/hyperspy/hyperspy_gui_ipywidgets/blob/e18aca582899bd9ba94cb74bf75da9c04bda5d52/hyperspy_gui_ipywidgets/tools.py#L67-L148) + [this](https://github.com/hyperspy/hyperspy_gui_ipywidgets/blob/e18aca582899bd9ba94cb74bf75da9c04bda5d52/hyperspy_gui_ipywidgets/hyperspy_extension.yaml#L33) + [this](https://github.com/hyperspy/hyperspy/blob/deff3fb05a9ab9e7325199aab43861a229f81fd2/hyperspy/hyperspy_extension.yaml#L246) and [this](https://github.com/hyperspy/hyperspy/blob/df2af64ae8e8e077e967ae2a9d76a3539beb47ed/hyperspy/_signals/signal1d.py#L854)! ;)



> A small suggestion: instead of the interactive keyword, it could default to it if no arguments are passed to the function.

Here is an example:
https://github.com/hyperspy/hyperspy/blob/df2af64ae8e8e077e967ae2a9d76a3539beb47ed/hyperspy/_signals/signal1d.py#L784-L800

francisco-dlp:
Another usability suggestion that is more involved: select an edge from the list and display vertical line markers at the position of all edges of that element in spectrum energy range—that's the usual way to identify edges when in doubt.

francisco-dlp:
A simpler variation (that may even be better usability-wise) would be to have an option (with the check button) to display markers at the position of all the edges in the *spectrum range* of all the elements that have edges in the selected *energy range*.

ptim0626:
@ericpre Thanks for pointing me to the ```calibrate()``` method, I shall have a good look at it and incorporate my method similarly.
@francisco-dlp I think it would be nice to have all those options for users, will see how far I can go (perhaps a bit too ambitious...)

ericpre:
> @ericpre Thanks for pointing me to the `calibrate()` method, I shall have a good look at it and incorporate my method similarly.
> @francisco-dlp I think it would be nice to have all those options for users, will see how far I can go (perhaps a bit too ambitious...)

Indeed, some of it can be done later in separate PR.
On the topic of displaying edge, there is this PR: https://github.com/hyperspy/hyperspy/pull/1527

ptim0626:
I have moved the ipywidget part to hyperspy/hyperspy_gui_ipywidgets#28. Now an energy range can be selected in the spectrum and it will display edges within the range, the way of the displayed edges can be changed, such as including only major edges. Upon selecting the edges, it will add markers to the signal to indicate their positions and an option is added so that edges of the same element in the energy axis range can be shown (useful for identifying edges). Here is a typical usage:
hyperspy_iss2409

On the matter of displaying edges, I do not include those in the EELS signal class and also not sure the status of #1527. My current edge label plotting tries to spread them out and does not explicitly check if any overlap occurred. If desired, I can refactor the relevant part to the EELS signal class and further improve the edge label plotting mechanism in a separate PR.

ericpre (commented):
This is already fairly nice. The code adding label on the figure should go in the EELS signal class, ideally most of it should go in a separate class so it can be also used for EDS signal but also other type of signal. Currently this PR doesn't support blitting, which doesn't matter with the notebook backend (since it doesn't support blitting) but for the qt backend, it can be significantly faster.

I think this could be simplify if following is possible:
- add tickboxes for each row in the table to display the label on the figure - I don't know if this is easily feasible but it will be easier to relate the selected/displayed edge and the corresponding information in the table
- when selecting a edge to plot, display all edges by default
- remove the "show_edge(s)" and "other edge(s)" selection widgets

The only advantage of the two selection widgets (in comparison to the tickbox approach) is that it is possible to select several at the same time but usually not many edges are relevant and we inspect these one by one, mainly because of the overlap issue.

There should be some speed issue somewhere because I would expect this to be faster.

If you want to keep this structure for this PR, then we can review it and the rest can be done in other PRs. Personally, I would prefer to have the label plotting done sorted first and then come back to this but this is up to you!

ptim0626:
@ericpre I will see what I can do for selecting edges, the reason I use those selection widget is that there seems to be no easy way to select entries in the table directly and I don't know how easy or difficult to align tickboxes to the rows of table accurately. I will see what I can do for this. And yes, I think should refactor the label plotting mechanism to a separate class, so both EELS and x-ray lines spectrum can utilise it.

ptim0626:
I implement the `plot` method in EELS as well, basically from the idea of #1527. The edge labelling mechanism is refactored into a new class. In term of scope I think this PR covers enough already. I emulate a 'clickable table' by assembling widgets into a grid and I add a button to update the table when it is told to do so it is smoother. I can add some tests and docs if this skeleton is ok.
hyperspy_iss2409_2

thomasaarholt:
Great! This looks really cool! I'll check it out this week if I get the chance!

francisco-dlp:
@ptim0626, before I forget it, could you update your name and institution in your GitHub profile. In this way they'll appear in the DOI that is automatically generated by zenodo using GitHub metadata.

hyperspy/misc/label_position.py
```diff
@@ -0,0 +1,220 @@
...
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with  HyperSpy.  If not, see <http://www.gnu.org/licenses/>.
+
+import copy
+import itertools
+
+import numpy as np
+from hyperspy.misc.eels.tools import get_edges_near_energy, get_info_from_edges
```
ericpre: Remove the `get_edges_near_energy` import, which is most likely a left over from previous version?

hyperspy/misc/label_position.py
```diff
@@ -0,0 +1,220 @@
...
+from hyperspy.drawing.marker import markers
+
+class SpectrumLabelPosition():
+    edge_label_style = {'ha' : 'center', 'va' : 'center',
```
ericpre: Add a short description of what it does and how it is expected to be used.

hyperspy/_signals/eels.py
```diff
@@ -1433,6 +1487,197 @@ def create_model(self, ll=None, auto_background=True, auto_add_edges=True,
...
+            The extra keyword arguments for plot()
+        """
+        
+        super().plot(**kwargs)
+
+        if include_edges is not False:
+            edges = self._get_edges_to_plot(include_edges, only_edges)
+            self.put_label_on_signal(edges)
+
+    def put_label_on_signal(self, edges, vertical_line_marker=None,
```
ericpre: Rename to `plot_edges_label` to be more consistent with other similar method (`plot_markers`, `_plot_xray_lines`)?

hyperspy/_signals/eels.py
```diff
@@ -1433,6 +1487,197 @@ def create_model(self, ll=None, auto_background=True, auto_add_edges=True,
...
+            vertical_line_marker, text_marker = slp.get_markers(edges)
+        if len(vertical_line_marker) != len(text_marker) or \
+            len(edges) != len(vertical_line_marker):
+            raise ValueError('The size of edges, vertical_line_marker and '
+                             'text_marker needs to be the same.')
+        
+        # add the markers to the signal and store them
+        self.add_marker(vertical_line_marker + text_marker)
+        info = dict(zip(edges, zip(vertical_line_marker, text_marker)))
+        self._edge_markers = {k:list(v) for k, v in info.items()}
```
ericpre: Is the `info` dictionary necessary? Alternatively, do the dictionary values need to be a list and not a tuple?
```suggestion
        self._edge_markers = {k:list(v) for k, v in zip(edges, zip(vertical_line_marker, text_marker))}
```

hyperspy/_signals/eels.py
```diff
@@ -1433,6 +1487,197 @@ def create_model(self, ll=None, auto_background=True, auto_add_edges=True,
...
+
+    def _get_edges_to_plot(self, include_edges, only_edges):
+        # get the dictionary of the edge to be shown
+        extra_element_edge_family = []
+        if include_edges is True:
+            try:
+                elements = self.metadata.Sample.elements
+            except AttributeError:
+                raise Warning("No elements defined. Add them with "
```
ericpre: I would use a `ValueError` instead of a warning, since it doesn't make sense to use `include_edges=True` if the elements are not defined.

hyperspy/_signals/eels.py
```diff
@@ -1433,6 +1487,197 @@ def create_model(self, ll=None, auto_background=True, auto_add_edges=True,
                           dictionary=dictionary)
         return model
 
+    def plot(self, include_edges=False, only_edges=('Major', 'Minor'),
```
ericpre: Rename `include_edges` to something more consistent with the rest of hyperspy? The corresponding argument for EDS signal is not a great example (`xray_lines`) but for model and more generally signal, we have `plot_components` and `plot_markers`, so I would suggest to use `plot_edges` or `plot_edges_label`. @hyperspy/core, any opinion?

For HyperSpy 2.0, it may be worth renaming `xray_lines` to something better.

hyperspy/signal_tools.py
```diff
@@ -238,7 +243,162 @@ def apply(self):
...
+            
+        self._clear_all_markers()
+            
+        edges_to_show = list(set(active).union(complementary))
+        vm, tm = self.slp.get_markers(edges_to_show)
+        self.signal.put_label_on_signal(edges_to_show, 
+                                        vertical_line_marker=vm, 
+                                        text_marker=tm)
+        self.active_edges = active 
+        self.active_complementary_edges = complementary
```
ericpre: Are these two lines necessary?

hyperspy/signal_tools.py
```diff
@@ -238,7 +243,162 @@ def apply(self):
...
+                        self.btns[pos].value = True
+
+    def _plot_labels(self, active=None, complementary=None):
+        # plot selected and/or complementary edges
+        if active is None:
+            active = self.active_edges
+        if complementary is None:
+            complementary = self.active_complementary_edges
+            
+        self._clear_all_markers()
```
ericpre: I am not sure removing all labels should be done here: this doesn't play well with the qt backend (labels are unnecessarily remove and re-added).

hyperspy/signal_tools.py
```diff
@@ -238,7 +243,162 @@ def apply(self):
...
+            description_all.append(description)
+        
+        self.energy_all = np.asarray(energy_all)
+        self.relevance_all = np.asarray(relevance_all)
+        self.description_all = np.asarray(description_all)
+
+    def update_table(self):
+        figure_changed = self.slp._check_signal_figure_changed()
+        if figure_changed:
+            self._clear_all_markers()
```
ericpre: When changing navigation index, drawing of the line is connected to the `axes_manager`: there should be a method which update the labels (their vertical position) when the nagivation indices change to take into account changes in intensity.
It may be worth looking at how some methods of the `BackgroundRemoval` tool are connected or how plotting x-rays lines works too.

hyperspy/misc/label_position.py
ericpre: This file should be not be signal specific and generic enough to support different type of signal: would it be possible to `get_info_from_edges`?

hyperspy/_signals/eels.py
```diff
@@ -1433,6 +1487,197 @@ def create_model(self, ll=None, auto_background=True, auto_add_edges=True,
...
+            if not line_markers:
+                self._edge_markers.pop(EELS_edge)
+
+    def remove_EELS_edges_markers(self, EELS_edges):
+        for EELS_edge in EELS_edges:
+            if EELS_edge in self._edge_markers:
+                line_markers = self._edge_markers[EELS_edge]
+                while line_markers:
+                    m = line_markers.pop()
+                    m.close()
```
ericpre: It may be worth considering using the argument `render_figure` to speed up removing marker when using matplotlib backend supporting blitting but let sort out first the `_clear_all_markers` bit.

ericpre (changes requested):
Nice, this is very good progress, considering this isn't a straightforward PR - the review takes time too! ;)
The structure is already in good shape but there are a couple of things to improve regarding adding/removing markers and update on changing navigation indices.

The custom table with the button is nice. The only drawback with this approach is that it seems that generating the table is slower than generating the html and I guess it may too slow to have the table updating interactively with moving the span selector. However, this is already good enough and this could be improve later (maybe doing it asynchronously or something similar).

francisco-dlp:
@ericpre and @ptim0626, what's the status of this? It would be great to include it in this week's release.

ptim0626:
@francisco-dlp sorry I was occupied by other things, will try my best to see if I can squeeze some time tomorrow to finish this off.

francisco-dlp:
Great, thanks!

hyperspy/misc/label_position.py
ptim0626: Yes it is

hyperspy/misc/label_position.py
ptim0626: Added

hyperspy/_signals/eels.py
ptim0626: Renamed to `plot_edges_label`

hyperspy/_signals/eels.py
ptim0626: Used `map` to change the zip's tuple to list in one go now. The values of the dictionary (which stores the markers) needs to be a list to allow popping and removing of the markers.

hyperspy/_signals/eels.py
ptim0626: Agreed.

hyperspy/_signals/eels.py
ptim0626: Renamed to `plot_edges`

hyperspy/signal_tools.py
ptim0626: Removed

hyperspy/signal_tools.py
ptim0626: Oh I haven't checked the qt backend! The marker's positions are now updated instead of removing them first and then adding them, this gives a smoother experience.

hyperspy/signal_tools.py
ptim0626: The event `axes_manager.events.indices_changed` is connected to a method `_on_nav_change` now so that whenever the indices of `axes_manager` changed, it gets the new figure properties and update the existing labels.

hyperspy/misc/label_position.py
ptim0626: Removed all dependency on `hyperspy.misc.eels.tools`

hyperspy/_signals/eels.py
ptim0626: `render_figure` is set `False` now whenever markers are added or closed.

francisco-dlp:
@ericpre, do you think that this one can make it into v1.6?

hyperspy/_signals/eels.py
```diff
@@ -1433,6 +1487,211 @@ def create_model(self, ll=None, auto_background=True, auto_add_edges=True,
...
+            vertical_line_marker, text_marker = slp.get_markers(edges)
+        if len(vertical_line_marker) != len(text_marker) or \
+            len(edges) != len(vertical_line_marker):
+            raise ValueError('The size of edges, vertical_line_marker and '
+                             'text_marker needs to be the same.')
+
+        # add the markers to the signal and store them
+        self.add_marker(vertical_line_marker + text_marker, render_figure=False)
+        added = dict(zip(edges, map(list, zip(vertical_line_marker, text_marker))))
+        self._edge_markers.update(added)
```
ericpre, line 1555: When changing indices, the labels doesn't update, while they do with the interactive tool. I guess the function drawning/updating the label should be connected to the `axes_manager` here.

An example to reproduce:
```python
s.add_elements(['Cr', 'Fe'])
s.plot(plot_edges=True)
```

hyperspy/signal_tools.py
```diff
@@ -244,6 +249,191 @@ def apply(self):
...
+        self.units = self.axis.units
+        self.active_edges = []
+        self.active_complementary_edges = []
+        self.slp = SpectrumLabelPosition(self.signal)
+        self.btns = []
+
+        self._get_edges_info_within_energy_axis()
+
+        self.signal.axes_manager.events.indices_changed.connect(self._on_nav_change,
+                                                                [])
```
ericpre, line 283: When connecting, we also need to connect the "disconnection" when closing. For example:

https://github.com/hyperspy/hyperspy/blob/3ec3018f59cee4c9db023408b09d51024fe53d42/hyperspy/signal_tools.py#L616-L621

ericpre (commented):
A few things need to be fixed but this is close to be finished. If this is possible to fix these tomorrow, it should be possible to include these in 1.6.

It would be good that the `SpectrumLabelPosition` class is also used for plotting X-rays line, but this could be done in another PR.

hyperspy/_signals/eels.py
ptim0626: Thanks for checking this! completely overlooked... now the labels will update both in static and interactive mode.

hyperspy/signal_tools.py
ptim0626: now it connects the 'disconnection' when `signal_plot` is closed

tjof2 (approved):

ericpre:
Thanks @ptim0626 for the quick update. I have added tests in https://github.com/ptim0626/hyperspy/pull/1. It would be good to increase the coverage of the `EdgesRange` (https://codecov.io/gh/hyperspy/hyperspy/compare/db4dad25bb47850cbeef00c886def2d058241e55...ef64a958f5aac7cbc518097fa2b5a0c0514b4ebf/src/hyperspy/signal_tools.py)

ericpre (approved):

PR #2420, opened 2020-06-14
Fix warnings always occurring with leastsq

### LINKED ISSUE #2419: Covariance warnings
With latest master, I'm getting a covariance warning with all leastsq model fittings:

> `WARNING:hyperspy.model:Covariance of the parameters could not be estimated. Estimated parameter standard deviations will be np.inf. This could indicate that the model is a poor description of the data.`

I get this error on all `m.fit()`, even when the fit fits the data exactly. Is this correct behaviour?
```python
import numpy as np
import hyperspy.api as hs
x = np.linspace(-5, 5, 1001)
data = g.function(x)
s = hs.signals.Signal1D(data)
s.axes_manager[-1].offset = -5
s.axes_manager[-1].scale = x[1] - x[0]

m = s.create_model()
G = hs.model.components1D.Gaussian()
m.append(G)
m.fit()
# WARNING:hyperspy.model:Covariance of the parameters could not be estimated. Estimated parameter standard deviations will be np.inf. This could indicate that the model is a poor description of the data.
(s - m.as_signal()).data.sum()
# results in 1e-15
```
With `fitter='mpfit'` I get reasonable values for the error, in `G.A.std`.

I believe this is a result of [a commit by @tjof2 17 days ago](https://github.com/hyperspy/hyperspy/commit/caf983fda14cf518a5853e8e6d51376841d7b6b3).

thomasaarholt:
Fixes a bug introduced with #2405, which prevented calculating std on parameters fitted with leastsq and raised a warning with basically all fitting done with leastsq.
Fixes #2419

Would normally go to _patch, but _minor hasn't been merged into it yet, so the code doesn't exist there yet.

### DIFF
--- a/hyperspy/model.py
+++ b/hyperspy/model.py
@@ -1146,7 +1146,7 @@ class BaseModel(list):
                     pcov = np.zeros((len(self.p0), len(self.p0)), dtype=float)
                     pcov.fill(np.inf)
                     warn_cov = True
-                elif pcov.min() < 0:
+                elif pcov.diagonal().min() < 0:
                     # Usually indicative of numerical overflow
                     # (covariance should be positive)
                     pcov.fill(np.inf)
--- a/hyperspy/tests/model/test_model.py
+++ b/hyperspy/tests/model/test_model.py
@@ -31,32 +31,6 @@ from hyperspy.exceptions import VisibleDeprecationWarning
 RTOL = 1E-6
 
 
-def test_leastsq_covar_est(caplog):
-    g = hs.model.components2D.Gaussian2D(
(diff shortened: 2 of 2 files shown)


thomasaarholt:
I'm introducing a test for this in the linear fitting PR I'm creating.

tjof2:
@thomasaarholt you can probably safely remove `test_leastsq_covar_est` which is failing, since that was introduced in #2405 to try and catch the error.

tjof2:
Codecov will probably complain, but it's a valid reason to remove it if the test wasn't correct in the first place! Thanks

thomasaarholt:
I'm really glad this was a bug 🤣 I was particularly worried as I'd been using the std errors from leastsq and mpfit as a comparison for the linear fitting I'm working on.

PR #2430, opened 2020-06-22
Use a simpler dataset for BSS testing

tjof2:
* #2395 encountered an unclear test failure, in that the BSS tests relied on the artificial dataset generation to test an unrelated piece of functionality (i.e. the ability to reverse components). This PR replaces the dataset with a random one that is easier to see how it works!
* Also correct Amari distance function and add test for Amari distance

### Progress of the PR
- [x] Change implemented (can be split into several points),
- [x] update docstring (if appropriate),
- [x] update user guide (if appropriate),
- [x] add tests,
- [x] ready for review.

PR #2431, opened 2020-06-22
Fix numpy1.19

ericpre:
Fix travis failure mentioned in #2430 due to the last version of numpy: https://github.com/numpy/numpy/releases/tag/v1.19.0

### Progress of the PR
- [x] fix axes_manager `next` (https://github.com/numpy/numpy/pull/15887),
- [x] fix `tostring` deprecation warning,
- [x] fix numpy ragged array deprecation warning,
- [x] fix complex change dtype deprecation warning,
- [x] ready for review.

### DIFF
--- a/hyperspy/axes.py
+++ b/hyperspy/axes.py
@@ -49,7 +49,7 @@ FACTOR_DOCSTRING = \
 class ndindex_nat(np.ndindex):
 
     def __next__(self):
-        return super(ndindex_nat, self).next()[::-1]
+        return super(ndindex_nat, self).__next__()[::-1]
 
 
 def generate_axis(offset, scale, size, offset_index=0):
--- a/hyperspy/_signals/complex_signal.py
+++ b/hyperspy/_signals/complex_signal.py
@@ -100,7 +100,7 @@ class ComplexSignal_mixin:
         # original plot
         self._plot_kwargs = {}
         if not np.issubdtype(self.data.dtype, np.complexfloating):
-            self.data = self.data.astype(np.complexfloating)
+            self.data = self.data.astype(np.complex128)
 
     def change_dtype(self, dtype):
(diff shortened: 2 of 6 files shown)


hyperspy/axes.py
```diff
@@ -49,7 +49,7 @@
 class ndindex_nat(np.ndindex):
 
     def __next__(self):
-        return super(ndindex_nat, self).next()[::-1]
+        return super(ndindex_nat, self).__next__()[::-1]
```
francisco-dlp, line 52: ``next(super(ndindex_nat, self))[::-1]``?

ericpre: This is the first thing I have tried and it doesn't work...

Here is a minimal example to reproduce:
```python
import hyperspy.api as hs

s = hs.signals.Signal1D(np.arange(10*10*10).reshape((10, 10, 10)))
[coordinate for coordinate in s.axes_manager._am_indices_generator()]
```
```python

  File "<ipython-input-3-181577425fbe>", line 1, in <module>
    [coordinate for coordinate in s.axes_manager._am_indices_generator()]

  File "<ipython-input-3-181577425fbe>", line 1, in <listcomp>
    [coordinate for coordinate in s.axes_manager._am_indices_generator()]

  File "/home/eric/Dev/hyperspy/hyperspy/axes.py", line 52, in __next__
    return next(super(ndindex_nat, self))[::-1]

TypeError: 'super' object is not an iterator
```

francisco-dlp: Interesting, it seems then that ``super`` is evaluated after ``next``.

ericpre:
I made this PR to the wrong branch (minor instead of patch), but it was based on the patch branch, so I manually push these commit to the patch branch!

PR #2432, opened 2020-06-22
Arctan expression

ericpre:
Tidy up #2395 (reset and rebase).

### DIFF
--- a/hyperspy/_components/arctan.py
+++ b/hyperspy/_components/arctan.py
@@ -16,9 +16,6 @@
 # You should have received a copy of the GNU General Public License
 # along with HyperSpy. If not, see <http://www.gnu.org/licenses/>.
 
-import math
-
-import numpy as np
 
 from hyperspy._components.expression import Expression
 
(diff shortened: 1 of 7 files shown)


hyperspy/_components/arctan.py
```diff
@@ -39,57 +40,35 @@ class Arctan(Component):
...
-        k = self.k.value
-        x0 = self.x0.value
-        return A * (x - x0) / (1 + (k * (x - x0)) ** 2)
-
-    def grad_x0(self, x):
-        A = self.A.value
-        k = self.k.value
-        x0 = self.x0.value
-        return -A * k / (1 + (k * (x - x0)) ** 2)
+        self.convolved = True
```
jlaehne: Just realized that `self.convolved` has changed from `False` to `True`. I overlooked that in the original PR.

hyperspy/_components/eels_arctan.py
```diff
@@ -0,0 +1,106 @@
...
+            k=k,
+            x0=x0,
+            position="x0",
+            module=module,
+            autodoc=False,
+            **kwargs,
+        )
+
+        self.isbackground = False
+        self.convolved = True
```
jlaehne, line 106: as above

jlaehne (approved):
Thanks for cleaning up the original PR ;-)

hyperspy/_components/arctan.py
tjof2: Is that perhaps why the BSS tests were failing - if that then changes the values of the artificial EELS datasets?

ericpre: I would actually suggest to remove it, so it falls back to the default - I don't see why this specific component shouldn't be convolved. @francisco-dlp?

> Is that perhaps why the BSS tests were failing - if that then changes the values of the artificial EELS datasets?

No, this is not the reason because the `function` method is used directly and it is not convolution aware (as it should be). However, `minimum_at_zero` was set to True and this may have been the reason?


For information on this, `model.convolved` is False by default and set to True when the `low_loss` is set
https://github.com/hyperspy/hyperspy/blob/226da1e5b88779b720e840d8d791f900f4111385/hyperspy/models/model1d.py#L317-L331

and in the `__call__` method, `component.convolved` is used only if `model.convolved=True`.

https://github.com/hyperspy/hyperspy/blob/226da1e5b88779b720e840d8d791f900f4111385/hyperspy/models/model1d.py#L405-L424

ericpre: @francisco-dlp: what would you recommend here: use the [default value of `convolved`](https://github.com/hyperspy/hyperspy/blob/fd8e637dc4b544c0b2d0892909007ddde2040c59/hyperspy/component.py#L755) or set it to `False`, as it was before this PR.

francisco-dlp: The ``convolved`` attribute (inherited from EELSLab) is there because in EELS you usually (there different opinions on this) convolve everything but the background components. That's why the default value is ``True``. With this in mind, I think that it is right to use the default value.

ericpre:
Removed setting attributes which were unnecessarily set. It should be all good now.

PR #2434, opened 2020-06-23
Correct warning message for signal types

### LINKED ISSUE #2433: Signal type warnings
#2425 (@tjof2), which fixes #1584, makes a lot of signal operations output the warning:
```python
WARNING:hyperspy.io:`signal_type=''` not understood. Setting signal type to `Signal1D`
```
It's easily testable by
```python
import hyperspy.api as hs
s = hs.signals.Signal1D([1,2,3])
m = s.create_model()
```
The test suite, in particular, outputs a large number of warnings, but it's also confusing the user.
Additionally, the signal type isn't actually set to anything afterwards, as far as I can see?

There are a couple of ways to resolve this:
- roll back the related changes and only give a warning if the user sets `s.set_signal('non_existing_signal_type')`, as originally intended by #1584 
- Automatically fill the signal type with the respective type (Signal1D, LazySignal1D, etc). Might need logic to convert the `signal_type` of `LazySignal1D.compute()` from LazySignal1D to Signal1D)

tjof2:
### Description of the change
Fixes #2433.

```python
>>> import hyperspy.api as hs
>>> s = hs.signals.Signal1D([1,2,3])
>>> m = s.create_model() # Prints no warning, as per example in #2433

>>> s.set_signal_type("EELS") # Prints no warning
>>> s.metadata
├── General
│   └── title = 
└── Signal
    ├── binned = True
    └── signal_type = EELS

>>> s.set_signal_type()
WARNING:hyperspy.io:`signal_type='None'` not understood. See `hs.print_known_signal_types()` for a list of known signal types.
>>> s.metadata
├── General
│   └── title = 
└── Signal
    ├── binned = False
    └── signal_type = None

>>> s.set_signal_type("LIGO")
WARNING:hyperspy.io:`signal_type='LIGO'` not understood. See `hs.print_known_signal_types()` for a list of known signal types.
>>> s.metadata
├── General
│   └── title = 
└── Signal
    ├── binned = False
    └── signal_type = LIGO
```
### Progress of the PR
- [x] add tests,
- [x] ready for review.

@thomasaarholt & @francisco-dlp  is this the correct behaviour now?

### DIFF
--- a/hyperspy/io.py
+++ b/hyperspy/io.py
@@ -402,7 +402,7 @@ def load_with_reader(filename, reader, signal_type=None, convert_units=False,
 
 
 def assign_signal_subclass(dtype, signal_dimension, signal_type="", lazy=False):
-    """Given record_by and signal_type, return the matching Signal subclass.
+    """Given dtype, signal_dimension and signal_type, return the matching Signal subclass.
 
     See `hs.print_known_signal_types()` for a list of known signal_types.
 
--- a/hyperspy/io.py
+++ b/hyperspy/io.py
@@ -431,7 +431,7 @@ def assign_signal_subclass(dtype, signal_dimension, signal_type="", lazy=False):
           'object' in dtype.name):
         dtype = 'real'
     else:
-        raise ValueError('Data type "{}" not understood!'.format(dtype.name))
+        raise ValueError(f'Data type "{dtype.name}" not understood!')
     if not isinstance(signal_dimension, int) or signal_dimension < 0:
         raise ValueError("signal_dimension must be a positive interger")
 
--- a/hyperspy/io.py
+++ b/hyperspy/io.py
@@ -401,18 +401,21 @@ def load_with_reader(filename, reader, signal_type=None, convert_units=False,
     return objects
 
 
-def assign_signal_subclass(dtype,
-                           signal_dimension,
-                           signal_type="",
-                           lazy=False):
-    """Given record_by and signal_type return the matching Signal subclass.
+def assign_signal_subclass(dtype, signal_dimension, signal_type="", lazy=False):
+    """Given record_by and signal_type, return the matching Signal subclass.
+
+    See `hs.print_known_signal_types()` for a list of known signal_types.
 
     Parameters
     ----------
     dtype : :class:`~.numpy.dtype`
-    signal_dimension: int
-    signal_type : {"EELS", "EDS", "EDS_SEM", "EDS_TEM", "DielectricFunction", "", str}
-    lazy: bool
+        Signal dtype
+    signal_dimension : int
+        Signal dimension
+    signal_type : str, default ""
+        Signal type. Optional. Will log a warning if it is unknown to HyperSpy.
+    lazy : bool, default False
+        If True, returns the matching LazySignal subclass.
 
     Returns
     -------
--- a/hyperspy/io.py
+++ b/hyperspy/io.py
@@ -441,6 +441,22 @@ def assign_signal_subclass(dtype,
                               if signal_type == value["signal_type"] or
                               "signal_type_aliases" in value and
                               signal_type in value["signal_type_aliases"]}
+
+    valid_signal_types = [v["signal_type"] for v in signals.values()]
+    valid_signal_aliases = [
+        v["signal_type_aliases"]
+        for v in signals.values()
+        if "signal_type_aliases" in v
+    ]
+    valid_signal_aliases = [i for j in valid_signal_aliases for i in j]
+    valid_signal_types.extend(valid_signal_aliases)
+
+    if signal_type not in set(valid_signal_types):
+        _logger.warning(
+            f"`signal_type='{signal_type}'` not understood. "
+            f"See `hs.print_known_signal_types()` for a list of known signal types."
+        )
+
     if dtype_dim_type_matches:
         # Perfect match found
         signal_dict = dtype_dim_type_matches
@@ -470,14 +486,7 @@ def assign_signal_subclass(dtype,
     for key, value in signal_dict.items():
         signal_class = getattr(importlib.import_module(value["module"]), key)
 
-        if value["signal_type"] == "":
-            _logger.warning(
-                f"`signal_type='{signal_type}'` not understood. "
-                f"Setting signal type to `{key}`"
-            )
-
         return signal_class
-
 
 def dict2signal(signal_dict, lazy=False):
     """Create a signal (or subclass) instance defined by a dictionary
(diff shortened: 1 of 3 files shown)


ericpre:
```python
>>> s.set_signal_type()
WARNING:hyperspy.io:`signal_type='None'` not understood. See `hs.print_known_signal_types()` for a list of known signal types.
>>> s.metadata
├── General
│   └── title = 
└── Signal
    ├── binned = False
    └── signal_type = None
```
To reset the signal type more explicit (set to *undefined* type), should we deprecate the default value `None` of and force using `s.set_signal_type("")`?

francisco-dlp:
+1

hyperspy/io.py
```diff
@@ -401,18 +401,21 @@ def load_with_reader(filename, reader, signal_type=None, convert_units=False,
     return objects
 
 
-def assign_signal_subclass(dtype,
-                           signal_dimension,
-                           signal_type="",
-                           lazy=False):
-    """Given record_by and signal_type return the matching Signal subclass.
+def assign_signal_subclass(dtype, signal_dimension, signal_type="", lazy=False):
+    """Given record_by and signal_type, return the matching Signal subclass.
```
francisco-dlp: Now that you are at it you could replace ``record_by`` with ``signal_dimension`` in the docstring.

tjof2: Done!

tjof2:
> To reset the signal type more explicit (set to undefined type), should we deprecate the default value None of and force using s.set_signal_type("")?

@ericpre this is done - let me know if it's not what you meant

hyperspy/io.py
```diff
@@ -441,6 +444,22 @@ def assign_signal_subclass(dtype,
                               if signal_type == value["signal_type"] or
                               "signal_type_aliases" in value and
                               signal_type in value["signal_type_aliases"]}
+
+    valid_signal_types = [v["signal_type"] for v in signals.values()]
```
ericpre, line 449: Use `set` to reduce the length?

hyperspy/io.py
```diff
@@ -441,6 +444,22 @@ def assign_signal_subclass(dtype,
...
+        if "signal_type_aliases" in v
+    ]
+    valid_signal_aliases = [i for j in valid_signal_aliases for i in j]
+    valid_signal_types.extend(valid_signal_aliases)
+
+    if signal_type not in set(valid_signal_types):
+        _logger.warning(
+            f"`signal_type='{signal_type}'` not understood. "
+            f"See `hs.print_known_signal_types()` for a list of known signal types."
+        )
```
ericpre: Since this is only necessary when `dtype_dim_type_matches` is empty, move it to the else block?

ericpre (commented):
A few comments, other than that look good to me.

hyperspy/io.py
tjof2: Not really necessary but its used lower down (`if not in set(...):`

hyperspy/io.py
tjof2: Agree I'll move it tomorrow

thomasaarholt:
Can we add on a hint for new developers, so that they are not discouraged?
`...hs.print_known_signal_types() for a list of known signal types and the developer guide for how to add more types.`

Other than that, looks good to me!

tjof2:
Thanks @thomasaarholt  - done!

hyperspy/signal.py
```diff
@@ -4721,9 +4721,9 @@ def set_signal_type(self, signal_type=None):
 
         HyperSpy ships with a minimal set of known signal types. External
         packages can register extra signal types. To print a list of
-        registered signal types in the current installation run this method
-        without arguments. Note that
-
+        registered signal types in the current installation, call
+        :py:meth:`hyperspy.utils.print_known_signal_types`, and see
+        the developer guide for details on how to add new signal_types.
```
ericpre, line 4726: I am not convinced that the developer guide is really useful here - the registration mechanism is already complicated enough for extension developer. Maybe it would be more useful to point to https://github.com/hyperspy/hyperspy-extensions-list

tjof2: @ericpre The developer guide has this lengthy section, which refers to the link you mention anyway. I don't think it's an issue.

https://hyperspy.readthedocs.io/en/latest/dev_guide/writing_extensions.html

ericpre: Yes, indeed and most users will not be interested in writing extension or learning about the registration mechanism! The extension list is a user facing document while the most of the content of this "writing extension" section of the developer guide is of little relevance for most users.
On this topic, it would be worth adding a column "signal type" to https://github.com/hyperspy/hyperspy-extensions-list, so it can be updated easily and independently from hyperspy documentation.

tjof2: @ericpre ok - I've updated the docstring to include a link to the repo - is that all?

hyperspy/io.py
ericpre: I was thinking from line 448; this is a detail but it makes the code slightly more readable and it avoid doing unnecessary list dictionary operation. What do you think?

francisco-dlp:
LGTM except for the warning message. As discussed above a ``signal_type`` may not be know to HyperSpy because the extension that provides it is not installed. @ericpre is working in https://github.com/hyperspy/hyperspy-extensions-list/pull/2 towards automatically generating the list of signals provided by extensions packages. I think that, once implemented, that'll be a very useful resource as the number of extensions grow. So +1 on adding a ling to https://github.com/hyperspy/hyperspy-extensions-list as suggested by @ericpre.

francisco-dlp:
I would suggest to leave this on hold until we merge  hyperspy/hyperspy-extensions-list#2

tjof2:
> I would suggest to leave this on hold until we merge hyperspy/hyperspy-extensions-list#2

@francisco-dlp I agree with the general idea of linking to the extensions list. However, as pointed out in #2433, right now on Rnm, even this throws a confusing and incorrect warning unless this is merged. 

```python
import hyperspy.api as hs
s = hs.signals.Signal1D([1,2,3])
m = s.create_model()
```

As it stands, the extensions list is already mentioned in the dcostring to `set_signal_type()`, so perhaps if I create a separate issue regarding the warning message, then this can be merged?

francisco-dlp:
For the sake of keeping things tidy I suggest to wait a bit for hyperspy/hyperspy-extensions-list#2. If there is no progress there before we release v1.6, we'll of course merge this regardless. I've just set the milestone to v1.6 not to forget it.

ericpre:
I will get hyperspy/hyperspy-extensions-list#2 sorted.

tjof2:
Is this good to merge now @francisco-dlp  following https://github.com/hyperspy/hyperspy-extensions-list/pull/3?

PR #2435, opened 2020-06-25
Fix bug in `signal.split`

### LINKED ISSUE #1298: Small bug in `signal.split`
Calling `split` with `number_of_parts` set but not `step_sizes` or `axis` raises a `ValueError` after 

``` python
elif number_of_parts is not 'auto' and step_sizes is not 'auto'
```

(line 2333) evaluates, because line 2319 sets `step_sizes`. Should be a quick fix, I'll do it in a bit, just putting the issue here now to remember.

tjof2:
### Description of the change
Closes #1298. Pretty certain `codecov` can be ignored here

### Progress of the PR
- [x] add tests,
- [x] ready for review.

### DIFF
--- a/hyperspy/signal.py
+++ b/hyperspy/signal.py
@@ -2577,6 +2577,10 @@ class BaseSignal(FancySlicing,
             A list of the split signals
 
         """
+        if number_of_parts != 'auto' and step_sizes != 'auto':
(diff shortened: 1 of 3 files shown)


thomasaarholt (approved):
LGTM

PR #2437, opened 2020-07-02
Docs: hspy compression saving times and chunking order

thomasaarholt:
### Description of the change
Saving hspy files can take a lot of time when using the default setting of compressing the data. (also, ref: #1739)
I think it makes sense to point this out, as for some data the compression gain is not huge, but the saving time takes much longer. In the random data example below (obvs not good for compression), hspy saving times take 15 times longer than without it.

I've also added a note about how chunking the hdf5 file is done according to numpy axes order, not hyperspy axes.

### Progress of the PR
- [x] ready for review.

### Example showing the increased saving times
```python
import numpy as np
import dask.array as da
import h5py
import hyperspy.api as hs

chunks = (100, 20, 1, 1, 300)

shape = (100, 100, 5, 5, 300)

# dask
x = da.random.random(shape, chunks=nav_chunks + sig_chunks)

with h5py.File('dask_example.hdf5', mode='w') as f:
    print('h5py from dask:')
    d = f.require_dataset('/data', shape=x.shape, dtype=x.dtype, chunks=chunks)
    %time da.store(x, d)
s = hs.signals.Signal1D(x)
print('Hyperspy from dask')
%time s.save('dask_example_compressed.hspy', chunks=chunks, overwrite=True) # chunks and compression
%time s.save('dask_example.hspy', chunks=chunks, overwrite=True, compression=None) # chunks no compression

# numpy

x = np.random.random(shape)

with h5py.File('numpy_example.hdf5', mode='w') as f:
    print('h5py from numpy:')
    d = f.require_dataset('/data', shape=x.shape, dtype=x.dtype, chunks=nav_chunks + sig_chunks)
    %time d[:] = x
    
s = hs.signals.Signal1D(x)
print('Hyperspy from numpy:')
%time s.save('numpy_example_compressed.hspy', chunks=chunks, overwrite=True) # chunks and compression
%time s.save('numpy_example.hspy', chunks=chunks, overwrite=True, compression=None) # chunks no compression
%time s.save('numpy_example_compressed.hspy', chunks=shape, overwrite=True, compression=None) # one chunk
```
**h5py from dask:**
Wall time: 563 ms
**Hyperspy from dask**
Wall time: 16.2 s
Wall time: 1.07 s
**h5py from numpy:**
Wall time: 466 ms
**Hyperspy from numpy:**
Wall time: 16.3 s
Wall time: 885 ms
Wall time: 2.12 s


tjof2:
Interesting. The default value of `compression='gzip'`, but, for example, if I do:
```python
import tables  # required so that h5py knows about blosc

# ...the rest of your code here...

%time s.save('data/dask_example_compressed.hspy', chunks=chunks, overwrite=True)
# Wall time: 14.7 s, file size 483M
%time s.save('data/dask_example_uncompressed.hspy', chunks=chunks, overwrite=True, compression=None)
# Wall time: 1.03 s, file size 573M
%time s.save('data/dask_example_lzf.hspy', chunks=chunks, overwrite=True, compression="lzf")
# Wall time: 3.04 s, file size 516M
%time s.save('data/dask_example_blosc.hspy', chunks=chunks, overwrite=True, compression=32001) # blosc
# Wall time: 1.62 s, file size 502M
```
The same results are seen for numpy - `blosc` is 10x faster than `gzip` and roughly equivalent to no compression in terms of writing speed.

BTW @thomasaarholt, your comparisons with `h5py from dask:`/`h5py from numpy:` aren't fair - you need to time the whole `with ...:` block. Doing so brings the timings into line with `s.save(compression=None/32001)` - I see around 1.1s or so.

We shouldn't consider changing the default compression I think, since:

> h5py (and HDF5 more broadly) support a number of different compression algorithms, e.g., GZIP, SZIP, or LZF (or even custom compression filters). However, only GZIP is built by default with HDF5, i.e., while data compressed with GZIP can be read on all platforms and installation of HDF5, other compressors may not be installed everywhere so that not all users may be able to access those files.

And to use `blosc` would require adding `pytables` to the list of dependencies. But it clearly makes a huge difference @francisco-dlp

doc/user_guide/io.rst
```diff
@@ -311,6 +311,10 @@ Extra saving arguments
 ^^^^^^^^^^^^^^^^^^^^^^^
 - `compression` : One of None, 'gzip', 'szip', 'lzf' (default is 'gzip').
 
+.. note::
+
+    Compression can signficantly slow down the saving speed by over an order of magnitude. If storage space
```
tjof2: Typo "significantly"

tjof2 (commented):
Probably worth adding a note as per my other comment, that only `compression=None/'gzip'` is guaranteed to be compatible across systems and HDF5 readers. And some discussion about `blosc` may be useful.

ericpre:
> And to use `blosc` would require adding `pytables` to the list of dependencies. But it clearly makes a huge difference @francisco-dlp

`blosc` is an [optional dependency](https://github.com/hyperspy/hyperspy/blob/5ae14d6e2ea7c58dbc4377f1711d9fc3544e33b0/setup.py#L76) and I don't think wheels are available on pypi, so installing it required compilers and so on. When using the conda package, `blosc` is available by default  through the [`mrcz` package](https://github.com/conda-forge/mrcz-feedstock/blob/c12e94e90130118f22d101e1e089c9a0e77d8501/recipe/meta.yaml#L27) which is a requirement of [`hyperspy-base`](https://github.com/conda-forge/hyperspy-feedstock/blob/2c695f94356472daae2704eaded4cc750f85e103/recipe/meta.yaml#L57).

It may be worth mentioning it in the documentation.

tjof2:
> We shouldn't consider changing the default compression I think

Other than possibly turning *off* compression by default, and instead documenting the options for compression - e.g. that only gzip is supported everywhere, and that blosc is much faster but requires pytables.

On this last point, it would be trivial to do a `try/except ImportError` to support it without adding a dependency? After all, pytables is included in the anaconda installer by default and is available on pypi without compilers, unlike blosc @ericpre - so it's likely that many users will already have it installed from conda and so could easily benefit.

Also, HDD storage is cheap, after all, whereas waiting an extra 14 seconds to only reduce the file size by 15% (which is the example above) is surely a worse user experience.

ericpre:
From a quick internet search, it seems that `h5py` doesn't support compressors out of the box (and therefore what I said about `blosc` above was irrelevant here) but there are several ways to support more compressors:
- use [hdf5-blosc](https://github.com/Blosc/hdf5-blosc); available on conda-forge but not on pypi
- use [pytables](https://github.com/PyTables/PyTables), which uses hdf5-blosc - available on pypi (wheels) and conda-forge
- use [hdf5plugin](https://github.com/silx-kit/), which uses hdf5-blosc - available on pypi (wheels) and conda-forge

It would be good to support more hdf5 compressors in hyperspy, which means we shall (optionally?) use one of the library above and improve the documentation accordingly.

The hdf5plugin seems to support more compressors than pytables, in terms of installation/availability they seem to be similar. According to https://github.com/silx-kit/hdf5plugin#installation, hdf5plugin supports openmp, but I don't know about pytables. It may be worth benchmarking pytables and hf5plugin...

tjof2:
hdf5plugin looks like it might be a nicer way to interface with the existing code. But since pytables is used by Pandas etc. maybe it's a better dependency to introduce. It also has the [`ptrepack`](https://www.pytables.org/usersguide/utilities.html#ptrepack) utility which is another option for fast writing - first write uncompressed, then use ptrepack to compress.

I don't think you need lots of compression options tbh - blosc is pretty much as good as it gets at the moment for speed with an acceptable compression ratio. Here's a useful discussion: http://www.pytables.org/usersguide/optimization.html#a-study-on-supported-compression-libraries

magnunor:
With regards to chunking sizes, Gary Paterson tested this for 4-dimensional data (see https://arxiv.org/pdf/1911.11560.pdf Fig 6 for more info). Essentially, chunk sizes of (16, 16, 16, 16) or (32, 32, 32, 32) seems to be a nice middle ground.

chunksize_hdf5

For this type of data, compression is very important: files can be easily be reduced from 8 GB to 1.6 GB. However, different data types compresses very differently, so finding a good chunk size which works for all data is really tricky.

ericpre:
Indeed, we shouldn't recommend particular chunks size, but it would be good to add relevant links from the dask array documentation to the user guide:
- https://docs.dask.org/en/latest/array-chunks.html
- https://docs.dask.org/en/latest/array-best-practices.html

@magnunor, do you have information regarding compressor and compressor provider relevant to the discussion?

from https://github.com/hyperspy/hyperspy/pull/2437#issue-443773436:
> I've also added a note about how chunking the hdf5 file is done according to numpy axes order, not hyperspy axes.

@thomasaarholt, is there supposed to be something about this in this PR? This would be a useful addition to the user guide.

magnunor:
@ericpre, for the example shown in the figure, level 4 GZIP compression from `h5py` was used

thomasaarholt:
@ericpre, I must not have pushed the changes associated with that, good catch. I'm currently on a week of Norway-holiday and don't have access to a pc. I will amend that when I'm back next week.

thomasaarholt:
Thanks for the feedback everyone, this is very good discussion. It will be useful to hear feedback from more "real-life" experiences as big data becomes more prevalent.

I've taken into account all feedback except the blosc discussion, clarified some concepts and fixed a rst/sphinx link.

Does anyone have anything more to add?

tjof2:
@thomasaarholt  do you want to convert the blosc part of this to a separate issue?

thomasaarholt:
I originally omitted it because I'm not familiar with blosc and didn't want to make a mistake, but if it's just about needing to import tables and use compression=32001 to use it for added speed and small file size, I can do that here. Is that correct? 

If we want more discussion on blosc, I suggest that you open a new issue on it. I don't fully appreciate the details of how we would support blosc and where it would be implemented.

tjof2:
>  if it's just about needing to import tables and use compression=32001 to use it for added speed and small file size, I can do that here. Is that correct?

I guess you can certainly mention that in the user guide as a "tip" for now, yes

doc/user_guide/io.rst
```diff
@@ -289,7 +288,17 @@ intensity<get_lines_intensity>`):
...
-By default, the data is saved in chunks that are optimised to contain at least one full signal. It is
+Data saved in the HDF5 format is typically divided into smaller chunks which can be loaded separately into memory, 
+allowing lazy loading. Chunk size can dramatically affect the speed of various HyperSpy algorithms, so chunk size is
+worth careful consideration when saving a signal. HyperSpy's default chunking sizes are probably not optimal
+for a given data analysis technique. For more comprehensible documentation on chunking,
+see the dask `array chunks
+<https://docs.dask.org/en/latest/array-chunks.html>`_ and `best practices
+<https://docs.dask.org/en/latest/array-best-practices.html>`_ docs. The chunks saved into HDF5 will
+match the dask array chunks in ``s.data.chunks`` when lazy loading.
+Chunk shape should follow the axes order of the numpy shape (``s.data.shape``), not the hyperspy shape.
```
ericpre: Detailed a bit more with an example, maybe? This is an usual source of confusion.

doc/user_guide/io.rst
```diff
@@ -309,7 +318,18 @@ passing ``chunks=True`` results in ``(7, 7, 256)`` chunks.
 
 Extra saving arguments
 ^^^^^^^^^^^^^^^^^^^^^^^
-- `compression` : One of None, 'gzip', 'szip', 'lzf' (default is 'gzip').
+`compression`:
+  One of ``None``, ``'gzip'``, ``'szip'``, ``'lzf'`` (default is ``'gzip'``). 
+  Optionally, if ``pytables`` is available by ``import tables``, the blosc compressor
```
ericpre: I think this is confusing because this is not an usual thing to do. It would be more simple to add `blosc` to the list, mentioned that `pytables` need to be installed to use it and raise an error if it is not installed, similarly as mentioned previously in https://github.com/hyperspy/hyperspy/pull/2437#issuecomment-653589024. Using `blosc` instead of `32001` will allow to change us compression filter provider in the future when necessary.

This should be a quick thing to do (along with adding a test checking that it this compressor works), so I would suggest to do it in this PR in order to avoid coming back to it.

doc/user_guide/io.rst
```diff
@@ -309,7 +318,18 @@ passing ``chunks=True`` results in ``(7, 7, 256)`` chunks.
...
+  can be used by ``compression=32001``. This can significantly improve the speed of saving with very
+  good compression.
+
+.. note::
+    Compression can significantly increase the saving speed. If storage space
```
ericpre: Instead of "storage space", I would use "file size" because it is not only about storage: bandwidth and transfer can also be a source of annoyange.

doc/user_guide/io.rst
```diff
@@ -289,7 +288,17 @@ intensity<get_lines_intensity>`):
...
+for a given data analysis technique. For more comprehensible documentation on chunking,
+see the dask `array chunks
+<https://docs.dask.org/en/latest/array-chunks.html>`_ and `best practices
+<https://docs.dask.org/en/latest/array-best-practices.html>`_ docs. The chunks saved into HDF5 will
+match the dask array chunks in ``s.data.chunks`` when lazy loading.
+Chunk shape should follow the axes order of the numpy shape (``s.data.shape``), not the hyperspy shape.
+
+By default, the data is saved in chunks that are optimised to contain at least one full signal shape. It is
 possible to customise the chunk shape using the ``chunks`` keyword. For example, to save the data with
 ``(20, 20, 256)`` chunks instead of the default ``(7, 7, 2048)`` chunks for this signal:
```
ericpre: I think it would be better to have this part in the [creating lazy signal section](https://hyperspy.readthedocs.io/en/latest/user_guide/big_data.html#creating-lazy-signals) and mentioned here with a link to it.

doc/user_guide/io.rst
```diff
@@ -309,7 +318,18 @@ passing ``chunks=True`` results in ``(7, 7, 256)`` chunks.
...
+  One of ``None``, ``'gzip'``, ``'szip'``, ``'lzf'`` (default is ``'gzip'``). 
+  Optionally, if ``pytables`` is available by ``import tables``, the blosc compressor
+  can be used by ``compression=32001``. This can significantly improve the speed of saving with very
+  good compression.
+
+.. note::
+    Compression can significantly increase the saving speed. If storage space 
+    is not an issue, it can be disabled by setting ``compression = None``.
+    Only ``compression = None/'gzip'`` is guaranteed to be compatible for reading with
+    other HDF5 readers.
```
ericpre: I would give the explanation rather than stating a fact: "the `gzip` compressors is supported on all platforms" with a link to https://h5py.readthedocs.io/en/stable/faq.html.

ericpre (changes requested):
A couple of suggested improvement.

ericpre:
What is the status of this PR?

hyperspy/tests/io/test_hdf5.py
```diff
@@ -651,6 +651,17 @@ def test_load_missing_y2_value(self):
             "test_marker_point_y2_data_deleted.hdf5"))
         assert len(s.metadata.Markers) == 5
 
+class TestSaveReadWithCompression():
+    def setup_method(self):
+        self.s = Signal1D(np.ones((3,3)))
+    
+    @pytest.mark.parametrize("compression", (None, "gzip", "szip", "lzf", "blosc", ))
+    def test_compression(self, compression):
```
ericpre: This test needs to be clean up after being executed

hyperspy/io_plugins/hspy.py
```diff
@@ -743,6 +777,10 @@ def file_writer(filename,
         expg = exps.create_group(group_name)
         if 'compression' not in kwds:
             kwds['compression'] = 'gzip'
+        elif kwds['compression'] == 'blosc':
```
ericpre: To avoid repeating this code twice, it could go in a function which will include the code of the existing `import_pytables_for_blosc` function.

doc/user_guide/conf.py
```diff
@@ -0,0 +1,299 @@
+# -*- coding: utf-8 -*-
```
ericpre: This file should be removed.

tjof2:
Merged in Rnp to get the CI and codecov working properly.

The test needs fixing though.

Looks like the new file `doc/user_guide/conf.py` shouldn't be in there though @thomasaarholt

ericpre (changes requested):
The test suite needs to be fixed.

hyperspy/tests/io/test_hdf5.py
thomasaarholt: Done.

hyperspy/io_plugins/hspy.py
thomasaarholt: I think my solution is what you wanted, please take a look.

thomasaarholt:
Right now one test is failing, `TestSaveReadWithCompression.test_compression[blosc]`, because `tables` isn't available. Do we want to add tables as a requirement in setup.py, or should I have the test expect an importerror?

doc/user_guide/io.rst
```diff
@@ -27,9 +27,15 @@ information.
 If the loaded file contains several datasets, the :py:func:`~.io.load`
 functions will return a list of the corresponding signal.
 
+<<<<<<< HEAD
```
ericpre: This merge conflict wasn't solved correctly.

ericpre (changes requested):

ericpre:
> Right now one test is failing, `TestSaveReadWithCompression.test_compression[blosc]`, because `tables` isn't available. Do we want to add tables as a requirement in setup.py, or should I have the test expect an importerror?

`pytables` needs to be added as an optional dependency, maybe with the `speed` entry in `extras_require`? In any case, the test suite should pass with pytables installed or not.

ericpre:
@thomasaarholt, do you want to try to finish this PR?

thomasaarholt:
I've rebased on latest _patch, and taken into account all your changes. Any further thoughts?

ericpre:
The test suite is failing on Azure pipeline.

doc/user_guide/big_data.rst
```diff
@@ -162,6 +162,33 @@ performed lazily. Important points are:
...
+    (10, 200, 300)
+    
+    >>> s = hs.signals.Signal1D(data)
+    >>> s # Note the reversed order of navigation dimensions
+    <Signal1D, title: , dimensions: (200, 10|300)>
+    
+    >>> s.save('chunked_signal.hspy', chunks=(10, 100, 300)) # Chunking first hyperspy dimension (second array dimension)
+    >>> s2 = hs.load('chunked_signal.hspy', lazy=True)
+    >>> s2.data.chunksize
+    (10, 100, 30)
```
ericpre:
```suggestion
    (10, 100, 300)
```

ericpre (changes requested):
This looks good to me, it just need the typo and the test suite to be fixed. Thanks!

doc/user_guide/big_data.rst
thomasaarholt: Good catch!
I've added a check to see if szip support is present (h5py docs say its dependent on the hdf5 install due to legal issues). Check the latest commit to see the changes.

thomasaarholt: I just saw that I could accept your code change directly here. That's a nice feature, but I'd already commited it :)

hyperspy/tests/io/test_hdf5.py
```diff
@@ -658,8 +658,9 @@ def setup_method(self):
     
     @pytest.mark.parametrize("compression", (None, "gzip", "szip", "lzf", "blosc", ))
     def test_compression(self, compression, tmp_path):
-        if compression == "szip" and sys.platform == "win32":
-            pytest.skip('szip not supported on windows')
+        if compression == "szip": # Support for szip is dependent on HDF5 install
```
ericpre: Why is it necessary to test the `szip` compression filters? This is testing the hdf5 install, not hyperspy... and we don't need to test this! I would simply remove using this compression filters from the list of parameters and remove the corresponding function in `hspy.py`.

hyperspy/io_plugins/hspy.py
```diff
@@ -743,6 +744,16 @@ def write_signal(signal, group, **kwds):
         for model in model_group.values():
             model.attrs['_signal'] = group.name
 
+def get_h5py_read_write_filters():
+    """Get tuple of lists of available read and write compression formats.
+    Some of these, like szip, are dependent on the HDF5 install due to 
+    licensing issues
+    """
+    from h5py._hl.filters import _gen_filter_tuples
```
ericpre: We should not rely on private API of h5py. And if this is used only in the test suite, I don't see why it would be in hyperspy public API.

doc/user_guide/io.rst
```diff
@@ -362,9 +363,23 @@ what, for large signal spaces usually leads to smaller chunks as ``guess_chunks`
...
 
+Choosing the correct chunk-size can significantly affect the speed of reading, writing and performance of many hyperspy algorithms.
+See the `chunking section <big_data.html#Chunking>`__ under `Working with big data <big_data.html>`__ for more information.
+
 Extra saving arguments
 ^^^^^^^^^^^^^^^^^^^^^^^
-- ``compression`` : One of None, 'gzip', 'szip', 'lzf' (default is 'gzip').
+`compression`:
+  One of ``None``, ``'gzip'``, ``'szip'``, ``'lzf'``, ``'blosc'`` (default is ``'gzip'``). 
+  The ``'blosc'`` compressor requires ``pytables`` to be installed. It can significantly improve the speed of saving with very
```
ericpre, line 373: Mention that `szip` is not available with all hdf5 installation?

ericpre (changes requested):
Except if I miss something, I don't see why we need to check if `hdf5` support `szip` compression filters!

thomasaarholt:
Right. I wasn't sure how many people rely on szip, and thought you'd ask me to *add* a check for it 😅 But your argument makes sense. I'll remove mention of szip in the code, and add a note about szip not always being available in the docs.

ericpre (approved):

ericpre:
Because of the issue with latest h5py on travis (now noticeable after #2569 was merged), the following test failure have been missed:
```python
_____________ TestSaveReadWithCompression.test_compression[blosc] ______________
[gw1] linux -- Python 3.7.1 /home/travis/virtualenv/python3.7.1/bin/python
self = <hyperspy.tests.io.test_hdf5.TestSaveReadWithCompression object at 0x7f82ca827cf8>
compression = 'blosc'
tmp_path = PosixPath('/tmp/pytest-of-travis/pytest-0/popen-gw1/test_compression_blosc_3')
    @pytest.mark.parametrize("compression", (None, "gzip", "lzf", "blosc", ))
    def test_compression(self, compression, tmp_path):
        if compression == 'blosc' and not pytables_is_installed():
            pytest.skip("pytables is optional and not installed")
    
>       self.s.save(tmp_path / 'test_compression.hspy', overwrite=True, compression=compression)
hyperspy/tests/io/test_hdf5.py:676: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
hyperspy/signal.py:2748: in save
hyperspy/io.py:756: in save
    writer.file_writer(str(filename), signal, **kwds)
hyperspy/io_plugins/hspy.py:810: in file_writer
    write_signal(signal, expg, **kwds)
hyperspy/io_plugins/hspy.py:731: in write_signal
    **kwds)
hyperspy/io_plugins/hspy.py:586: in overwrite_dataset
    dset = group.require_dataset(key, **these_kwds)
../../../virtualenv/python3.7.1/lib/python3.7/site-packages/h5py/_hl/group.py:212: in require_dataset
    return self.create_dataset(name, *(shape, dtype), **kwds)
../../../virtualenv/python3.7.1/lib/python3.7/site-packages/h5py/_hl/group.py:148: in create_dataset
    dsid = dataset.make_new_dset(group, shape, dtype, data, name, **kwds)
../../../virtualenv/python3.7.1/lib/python3.7/site-packages/h5py/_hl/dataset.py:110: in make_new_dset
    maxshape, scaleoffset, external, allow_unknown_filter)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
plist = <h5py.h5p.PropDCID object at 0x7f82cbb8fdb8>, shape = (3, 3)
dtype = dtype('float64'), chunks = (3, 3), compression = 'blosc'
compression_opts = None, shuffle = True, fletcher32 = None
maxshape = (None, None), scaleoffset = None, external = None
allow_unknown_filter = False
    def fill_dcpl(plist, shape, dtype, chunks, compression, compression_opts,
                  shuffle, fletcher32, maxshape, scaleoffset, external,
                  allow_unknown_filter=False):
        """ Generate a dataset creation property list.
    
        Undocumented and subject to change without warning.
        """
    
        if shape is None or shape == ():
            shapetype = 'Empty' if shape is None else 'Scalar'
            if any((chunks, compression, compression_opts, shuffle, fletcher32,
                    scaleoffset is not None)):
                raise TypeError(
                    f"{shapetype} datasets don't support chunk/filter options"
                )
            if maxshape and maxshape != ():
                raise TypeError(f"{shapetype} datasets cannot be extended")
            return h5p.create(h5p.DATASET_CREATE)
    
        def rq_tuple(tpl, name):
            """ Check if chunks/maxshape match dataset rank """
            if tpl in (None, True):
                return
            try:
                tpl = tuple(tpl)
            except TypeError:
                raise TypeError('"%s" argument must be None or a sequence object' % name)
            if len(tpl) != len(shape):
                raise ValueError('"%s" must have same rank as dataset shape' % name)
    
        rq_tuple(chunks, 'chunks')
        rq_tuple(maxshape, 'maxshape')
    
        if compression is not None:
            if isinstance(compression, FilterRefBase):
                compression_opts = compression.filter_options
                compression = compression.filter_id
    
            if compression not in encode and not isinstance(compression, int):
>               raise ValueError('Compression filter "%s" is unavailable' % compression)
E               ValueError: Compression filter "blosc" is unavailable
../../../virtualenv/python3.7.1/lib/python3.7/site-packages/h5py/_hl/filters.py:178: ValueError
```

PR #2441, opened 2020-07-11
Improve m.fit() functionality

### LINKED ISSUE #1202: least_squares does not take Jacobian
The new `"leastsq"` fitting can take bound constraints, using SciPy's [`least_squares` routine](http://docs.scipy.org/doc/scipy/reference/generated/scipy.optimize.least_squares.html). The Jacobian is not passed to this routine (instead it calculates it internally using finite differences) as I think the format of the required function is different to the other places it is used. 

"least_squares" takes `jac : {‘2-point’, ‘3-point’, ‘cs’, callable}, optional`

Relevant lines in model.py:

```
        if fitter == "leastsq":
                if bounded:
                    self.set_boundaries()
                    ls_b = self.free_parameters_boundaries
                    ls_b = ([a if a is not None else -np.inf for a, b in ls_b],
                            [b if b is not None else np.inf for a, b in ls_b])
                    output = \
                        least_squares(self._errfunc, self.p0[:],
                                      args=args, bounds=ls_b, **kwargs)
                    self.p0 = output.x

                    # Do Moore-Penrose inverse, discarding zero singular values
                    # to get pcov (as per scipy.optimize.curve_fit())
                    _, s, VT = svd(output.jac, full_matrices=False)
                    threshold = np.finfo(float).eps * \
                        max(output.jac.shape) * s[0]
                    s = s[s > threshold]
                    VT = VT[:s.size]
                    pcov = np.dot(VT.T / s**2, VT)

                elif bounded is False:
                    # This replicates the original "leastsq"
                    # behaviour in earlier versions of HyperSpy
                    # using the Levenberg-Marquardt algorithm
                    output = \
                        leastsq(self._errfunc, self.p0[:], Dfun=jacobian,
                                col_deriv=1, args=args, full_output=True,
                                **kwargs)
                    self.p0, pcov = output[0:2]

```

At least, I couldn't get it to work when I tried it a week or so ago. Any thoughts?

### LINKED ISSUE #1346: Different Evolution fitting bounded=True fails
Attempting `m.fit(fitter="Differential Evolution", bounded=True)` returns the following error.

I think it might be because only some of the components have bounds set to them, although I'm uncertain.

<ipython-input-123-a3120aa8700a> in <module>()
----> 1 m.fit(fitter="Differential Evolution", bounded=True)

/Users/thomas/Dropbox/0_Git/GitHub/hyperspy_aar/hyperspy/model.py in fit(self, fitter, method, grad, bounded, ext_bounding, update_plot, **kwargs)
   1168                                    b if b is not None else np.inf) for a, b in de_b))
   1169                     self.p0 = differential_evolution(tominimize, de_b,
-> 1170                                                      args=args, **kwargs).x
   1171 
   1172                 else:

/Users/thomas/anaconda/envs/sims/lib/python3.5/site-packages/scipy/optimize/_differentialevolution.py in differential_evolution(func, bounds, args, strategy, maxiter, popsize, tol, mutation, recombination, seed, callback, disp, polish, init)
    203                                          callback=callback,
    204                                          disp=disp,
--> 205                                          init=init)
    206     return solver.solve()
    207 

/Users/thomas/anaconda/envs/sims/lib/python3.5/site-packages/scipy/optimize/_differentialevolution.py in __init__(self, func, bounds, args, strategy, maxiter, popsize, tol, mutation, recombination, seed, maxfun, callback, disp, polish, init)
    357         if (np.size(self.limits, 0) != 2 or not
    358                 np.all(np.isfinite(self.limits))):
--> 359             raise ValueError('bounds should be a sequence containing '
    360                              'real valued (min, max) pairs for each value'
    361                              ' in x')

ValueError: bounds should be a sequence containing real valued (min, max) pairs for each value in x

### LINKED ISSUE #1388: `m.fit` does not fail graciously when analytical gradient is not available for a parameter
```python
>>> m.fit(fitter="L-BFGS-B", bounded=True, grad=True)
```
fails, where grad=False does not.

Please see error here: https://gist.github.com/thomasaarholt/d9506d0fc25657d3da3aab5aad50a735

### LINKED ISSUE #1627: Disable automatic weighting by inverse variance when fitting
Currently if signal variance is set, its inverse is used as weights when fitting, and there is no way to disable that.

### LINKED ISSUE #2276: EELSModel.fit bounded docstring information
In the current RELEASE_next_minor, it says in `EELSModel.fit`'s docstring, that `bounded=True` only works with `fitter="mpfit"`.

However, when I use it with `fitter="leastsq"`, bounding still seems to work fine.

Is this docstring information out of date?

https://github.com/hyperspy/hyperspy/blob/5cc1d15344e71ace89cca35cd1027502c976d979/hyperspy/models/eelsmodel.py#L315

tjof2:
### Description of the change
- Closes #1202 by supporting Jacobian for bounded least squares fitting.
- ~~As a result, can deprecate `fitter='mpfit'` in favour of `fitter='leastsq'` - since both are wrappers/derived from MINPACK (Fortran), but the `scipy.optimize.least_squares` route has more functionality, including for numerical gradient approximation.~~
- Add informative error messages for missing analytical gradients (closes #1388)
- Support numerical gradient approximation by finite differences where possible
- ~~Parameter error estimation available when `fitter='L-BFGS-B'`~~
- Added Huber loss function to complement least-squares and Poisson MLE
- Additional global optimization methods available if `scipy` version is >=1.2.0
- Clearer warning message for bounded global optimization (closes #1346)
- ~~Option to remove weighting by inverse variance (closes #1627)~~
- Update docstring for EELSModel fit() (closes #2276)
- Added `print_info=True/False` option to `.fit()`, similar to `.decomposition()`. ~~It is `False` by default since it will print *a lot* when using `m.multifit()`.~~ An example is shown below.
- Improved documentation for fitting algorithms.
- Improved tests for fitting algorithms.

### Progress of the PR
- [x] Changes implemented,
- [x] update docstring,
- [x] update user guide,
- [x] add tests,
- [x] ready for review.

### Minimal example
An example that _fails_ has been chosen deliberately to show the output of `print_info=True`:
```python
>>> m.fit(grad="2-point", fitter="L-BFGS-B", bounded=True, print_info=True)
Fit info:
  fitter=L-BFGS-B
  method=ls
  grad=2-point
  bounded=True
Fit result:
      fun: nan
 hess_inv: <6x6 LbfgsInvHessProduct with dtype=float64>
      jac: array([-2.27957549e-06,  8.37888231e+12, -3.67247360e+07, -1.31072000e+05,
        2.30918888e+04,  8.19200000e+03])
  message: b'ABNORMAL_TERMINATION_IN_LNSRCH'
     nfev: 14
      nit: 0
     njev: 2
   status: 2
  success: False
        x: array([4.72696350e+17, 3.84570141e+00, 1.00000000e-01, 1.00000000e+00,
       2.23000000e+03, 1.00000000e+00])
```


doc/user_guide/model.rst
```diff
@@ -14,8 +14,12 @@ dimensions i.e. spectra and images respectively. Most of the syntax is
 identical in either case. A one-dimensional model is created when a model
 is created for a :py:class:`~._signals.signal1d.Signal1D` whereas a two-
 dimensional model is created for a :py:class:`~._signals.signal2d.Signal2D`.
-At present plotting and gradient fitting methods tools for are not yet
-provided for the :py:class:`~.models.model2d.Model2D` class.
+
+.. note::
+
+    Plotting and analytic gradient-based fitting methods for are not yet
```
francisco-dlp: remove for?

doc/user_guide/model.rst
```diff
@@ -659,65 +659,66 @@ Fitting the model to the data
...
-                    methods changed to the `scipy.optimze.minimize()` notation.
+spectrum at a particular point in a spectrum-image), use
+:py:meth:`~.model.BaseModel.fit`. HyperSpy implements a number of
+different optimization approaches, each of which can have particular
+benefits and/or drawbacks depending on your specific application.
+A good approach to choosing an optimization approach is to ask yourself
+"Do you want to...":
+
+* Apply bounds to your model parameter values?
+* Use of gradient information to try and improve the fit result?
```
francisco-dlp: This sentence is not clear to me. What about: "are gradient based algorithms well suited for your model"?

doc/user_guide/model.rst
```diff
@@ -775,43 +791,80 @@ gaussian noise and proceed to fit as in the previous example.
...
+   >>> weights = np.ones_like(s.data)
+   >>> m.fit(weights=weights)
+
+.. warning::
+
+    When the attribute ``metadata.Signal.Noise_properties.variance``
+    is defined, the default behaviour is to perform a weighted least-squares
+    fit using the inverse variance.
+
+    In this scenario, to explicitly disable weighting, pass ``m.fit(weights=None)``.
```
francisco-dlp: Why would one want to disable weighting if the variance is defined? Also, wouldn't that be equivalent to remove variance from metadata?

doc/user_guide/model.rst
```diff
@@ -775,43 +791,80 @@ gaussian noise and proceed to fit as in the previous example.
...
+
+To avoid biased estimation in the case of data corrupted by Poisson noise,
+we can use Poisson maximum likelihood estimation (MLE) instead.
+This is an unbiased estimator for Poisson noise. To perform MLE, we must
+use a general, non-linear optimizer from the ref:`table above <optimizers-table>`,
+such as Nelder-Mead or L-BFGS-B:
 
 .. code-block:: python
 
    >>> m.fit(fitter="Nelder-Mead", method="ml")
```
francisco-dlp: Now that you are at it, you could maybe consider deprecating ``ml`` in favour of ``ml-poisson``.

doc/user_guide/model.rst
```diff
@@ -849,9 +902,12 @@ model:
     ...
     ...    # Return squared and summed weighted difference
     ...    return (difference**2).sum()
-    >>> m.fit(fitter='TNC', method='custom', min_function=my_custom_function)
+    >>> # We must use a general non-linear optimizer again
+    >>> m.fit(fitter='Nelder-Mead',
+    ...       method="custom",
```
francisco-dlp: Wouldn't it be possible to pass the function directly instead of ``method="custom"`` plus the function in ``min_function``? We could rename ``method``->``loss-function``.

hyperspy/models/eelsmodel.py
```diff
@@ -287,49 +286,125 @@ def resolve_fine_structure(
...
+
+                * "leastsq" performs least-squares optimization, and
+                  supports bounds on parameters.
+                * "odr" performs the optimization using the orthogonal
+                  distance regression (ODR) algorithm. It does not support
+                  bounds on parameters. See :py:mod:`scipy.odr` for more details.
+                * All of the available methods for :py:func:`scipy.optimize.minimize`
+                  can be used here:
+
+                    - "Powell"
```
francisco-dlp: I think that it is better not to list the methods here because in time this list will be incorrect.

hyperspy/models/model1d.py
```diff
@@ -554,12 +555,24 @@ def _jacobian(self, param, y, weights=None):
                         onlyfree=True)
                     if component.convolved:
                         for parameter in component.free_parameters:
+                            if not callable(parameter.grad):
+                                raise ValueError(
+                                    f"Analytical gradient not available for {parameter}. "
+                                    "Consider using a numerical approximation for the gradient "
+                                    "instead by calling `m.fit(grad='2-point')`."
```
francisco-dlp: by calling **e.g.** ...

And maybe calling ``m.fit(grad=False)`` instead?

hyperspy/models/model1d.py
```diff
@@ -554,12 +555,24 @@ def _jacobian(self, param, y, weights=None):
...
                                 parameter.grad(self.convolution_axis),
                                 self.low_loss(self.axes_manager),
                                 mode="valid")
                             if parameter._twins:
                                 for par in parameter._twins:
+                                    if not callable(par.grad):
+                                        raise ValueError(
+                                            f"Analytical gradient not available for {par}. "
+                                            "Consider using a numerical approximation for the gradient "
+                                            "instead by calling `m.fit(grad='2-point')`."
```
francisco-dlp: as above

hyperspy/models/model1d.py
```diff
@@ -568,9 +581,21 @@ def _jacobian(self, param, y, weights=None):
                             grad = np.vstack((grad, par_grad))
                     else:
                         for parameter in component.free_parameters:
+                            if not callable(parameter.grad):
+                                raise ValueError(
+                                    f"Analytical gradient not available for {parameter}. "
```
francisco-dlp: Since the error message is the same at multiple places, what about storing it in the global variable instead?

hyperspy/model.py
```diff
@@ -924,475 +956,698 @@ def red_chisq(self):
...
+                    # (f^k - f^{k+1})/max{|f^k|,|f^{k+1}|,1} <= ftol
+
+                    # Default value of ftol for L-BFGS-B
+                    ftol = 2.220446049250313e-9
+                    if "options" in kwargs:
+                        ftol = kwargs["options"].get("ftol", ftol)
+
+                    cost = self.fit_output.fun
+                    hess = np.diag(self.fit_output.hess_inv.todense())
+                    self.p_std = np.sqrt(max(1.0, np.abs(cost)) * ftol * hess)
```
francisco-dlp: This seems to consistently underestimate the error bounds. Besides, that doesn't compute the std, what may be misleading. Therefore I suggest to remove it.
Alternatives that can work for all optimizers would be to use the F-test like [lmfit does](https://lmfit.github.io/lmfit-py/confidence.html#method-used-for-calculating-confidence-intervals), bootstrapping or MLMC with emcee, again like [lmfit](https://lmfit.github.io/lmfit-py/fitting.html#label-emcee).

hyperspy/models/eelsmodel.py
francisco-dlp: as above

hyperspy/models/eelsmodel.py
```diff
@@ -287,49 +286,125 @@ def resolve_fine_structure(
...
+        grad=False,
+        bounded=False,
+        ext_bounding=False,
+        update_plot=False,
+        weights="inverse_variance",
+        print_info=False,
+        kind="std",
+        **kwargs,
+    ):
+        """Fits the model to the experimental data.
```
francisco-dlp, line 292: Instead of copying the docstring it would be better to put it in a variable like e.g. in https://github.com/hyperspy/hyperspy/blob/RELEASE_next_minor/hyperspy/signal.py#L2339

hyperspy/model.py
```diff
@@ -924,475 +956,698 @@ def red_chisq(self):
...
+
+        method : {"ls", "ml", "huber", "custom"}, default "ls"
+            The loss function to use for minimization.
+
+                * "ls" minimizes the least-squares loss function.
+                * "ml" minimizes the negative log-likelihood for Poisson-distributed
+                  data. Also known as Poisson maximum likelihood estimation
+                  (MLE). Not available if ``fitter`` is ``"leastsq"``
+                  or ``"odr"``.
+                * "huber" minimize the Huber loss function. Not available
```
francisco-dlp: The huber loss-function takes the important ``delta`` parameter. Shouldn't we offer a way to input the parameter? The usual way would be to add an extra keyword. An alternative could be passing e.g. ``"huber delta=10"``.

hyperspy/model.py
```diff
@@ -727,44 +755,48 @@ def _set_p0(self):
                                if parameter._number_of_elements == 1
                                else self.p0 + parameter.value)
 
-    def set_boundaries(self):
+    def set_boundaries(self, bounded=True):
```
francisco-dlp, line 763: The ``bounded`` keyword is not documented

francisco-dlp, line 763: Maybe we could consider making this method private? (With just a deprecation warning now of course.)

hyperspy/model.py
```diff
@@ -924,475 +956,698 @@ def red_chisq(self):
...
 
-    def fit(self, fitter="leastsq", method='ls', grad=False,
-            bounded=False, ext_bounding=False, update_plot=False,
-            **kwargs):
+    def _calculate_parameter_std(self, pcov, cost, ysize):
+        warn_cov = False
+
+        if pcov is None:  # Indeterminate covariance
+            p_var = np.zeros(len(self.p0), dtype=float)
+            p_var.fill(np.inf)
```
francisco-dlp: Why not filling with ``nan`` instead?

hyperspy/model.py
```diff
@@ -924,475 +956,698 @@ def red_chisq(self):
...
+            The optimization algorithm used to perform the fitting.
+
+                * "leastsq" performs least-squares optimization, and
+                  supports bounds on parameters.
+                * "odr" performs the optimization using the orthogonal
+                  distance regression (ODR) algorithm. It does not support
+                  bounds on parameters. See :py:mod:`scipy.odr` for more details.
+                * All of the available methods for :py:func:`scipy.optimize.minimize`
+                  can be used here:
+
```
francisco-dlp: As there is not reliable way to keep this list in sync with scipy, I suggest to remove it.

doc/user_guide/model.rst
```diff
@@ -659,65 +659,66 @@ Fitting the model to the data
...
+    | ``"odr"``                    |  No    | Yes       | Yes    | Only ``"ls"``  | local  |
+    +------------------------------+--------+-----------+--------+----------------+--------+
+    | ``"Nelder-Mead"``            |  No    | No        | No     | All            | local  |
+    +------------------------------+--------+-----------+--------+----------------+--------+
+    | ``"L-BFGS-B"``               |  Yes   | Yes       | Yes    | All            | local  |
+    +------------------------------+--------+-----------+--------+----------------+--------+
+    | ``"Differential Evolution"`` |  Yes   | No        | No     | All            | global |
+    +------------------------------+--------+-----------+--------+----------------+--------+
+
+\* ``mpfit`` is deprecated since HyperSpy 1.6. Use ``"leastsq"`` instead.
```
francisco-dlp: I remember testing them a while ago with a few datasets and finding that mpfit gave better results. However, my tests weren't very systematic. Have you tested the performance of mpfit vs leastsq with bounds?

doc/user_guide/model.rst
```diff
@@ -659,65 +659,66 @@ Fitting the model to the data
...
+spectrum at a particular point in a spectrum-image), use
+:py:meth:`~.model.BaseModel.fit`. HyperSpy implements a number of
+different optimization approaches, each of which can have particular
+benefits and/or drawbacks depending on your specific application.
+A good approach to choosing an optimization approach is to ask yourself
+"Do you want to...":
+
+* Apply bounds to your model parameter values?
+* Use of gradient information to try and improve the fit result?
+* Estimate the errors of the parameter values found by the fit?
```
francisco-dlp: errors -> std or confidence interval?

doc/user_guide/model.rst
```diff
@@ -659,65 +659,66 @@ Fitting the model to the data
...
+* Fit your data in the least-squares sense, or is your data corrupted
+  with Poisson noise?
```
francisco-dlp: when the data is corrupted by Poisson noise it is still possible to successfully use weighted least squares as long as the number of counts is not too low. This has the advantage that the std and the red-chisq can be computed.

doc/user_guide/model.rst
```diff
@@ -742,29 +743,44 @@ estimated standard deviation are stored in the following line attributes:
...
-estimated accurately. If the variance is not defined, the standard deviation of
-the parameters are still computed and stored in the
-:attr:`~.component.Parameter.std` attribute by setting variance equal 1.
-However, the value won't be correct unless an accurate value of the variance is
-defined in ``metadata.Signal.Noise_properties.variance``. See
-:ref:`signal.noise_properties` for more information.
+Optimization algorithms that take into account the gradient of the minimization function
+often out-perform so-called "derivative-free" optimization algorithms in terms of how rapidly
+they converge to a solution. HyperSpy can use analytical gradients for model-fitting, as well
+as numerical estimates of the gradient based on finite differences. Following the above example:
```
francisco-dlp: I would also mention here that:

* by default, the optimizers use analytical gradient and that using ``grad=True`` (as in your example) turns on analytical gradient if available for all components in the model.
* the advantages of analytical gradients are that i) they're more accurate ii) the optimization runs faster since there is no need to estimate the gradient

doc/user_guide/model.rst
```diff
@@ -742,29 +743,44 @@ estimated standard deviation are stored in the following line attributes:
...
+    the errors be estimated accurately.
+
+    If the variance is not defined, the standard deviation of
+    the parameters are still computed and stored in the
+    :attr:`~.component.Parameter.std` attribute by setting variance equal to 1.
+    However, the value won't be correct unless an accurate value of the variance is
+    provided. See :ref:`signal.noise_properties` for more information.
 
+Using gradient information
+^^^^^^^^^^^^^^^^^^^^^^^^^^
```
francisco-dlp: I would move this subsection under WLS below.

doc/user_guide/model.rst
```diff
@@ -775,43 +791,80 @@ gaussian noise and proceed to fit as in the previous example.
...
-   >>> m.fit()
+   ...     expected_value=hs.signals.Signal1D(np.arange(300))
+   ... )
+   >>> m.fit(weights="inverse_variance")
    >>> line.coefficients.value
    (1.0004224896604759, -0.46982916592391377)
    >>> line.coefficients.std
    (0.0055752036447948173, 0.46950832982673557)
 
+Instead of using the inverse variance, you can also provide your own
```
francisco-dlp: This is equivalent to setting the variance to the inverse of those weights. From The Zen of Python:

doc/user_guide/model.rst
```diff
@@ -775,43 +791,80 @@ gaussian noise and proceed to fit as in the previous example.
...
-Maximum likelihood optimisation
-^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
-We can use Poisson maximum likelihood estimation
-instead, which is an unbiased estimator for poissonian noise.
-To do so, we use a general optimizer called "Nelder-Mead".
+Maximum likelihood estimation
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+To avoid biased estimation in the case of data corrupted by Poisson noise,
```
francisco-dlp: I would add here "with very few counts" as it is only then when this becomes advantageous.

doc/user_guide/model.rst
```diff
@@ -775,43 +791,80 @@ gaussian noise and proceed to fit as in the previous example.
...
+
+    In this scenario, to explicitly disable weighting, pass ``m.fit(weights=None)``.
 
-Maximum likelihood optimisation
-^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
-We can use Poisson maximum likelihood estimation
-instead, which is an unbiased estimator for poissonian noise.
-To do so, we use a general optimizer called "Nelder-Mead".
+Maximum likelihood estimation
```
francisco-dlp: Poisson ML estimation?

doc/user_guide/model.rst
```diff
@@ -775,43 +791,80 @@ gaussian noise and proceed to fit as in the previous example.
...
 
-Because the noise is heterocedastic, the least squares optimizer estimation is
-biased. A more accurate result can be obtained by using weighted least squares
-instead that, although still biased for poissonian noise, is a good
-approximation in most cases.
+Because the noise is heteroscedastic, the least squares optimizer estimation is
+biased. A more accurate result can be obtained with weighted least squares,
+where the weights are proportional to the inverse of the noise variance.
+Although this is still biased for Poisson noise, it is a good approximation
+in most cases.
```
francisco-dlp: in most cases -> when there is a sufficient number of counts.

doc/user_guide/model.rst
francisco-dlp: In the end, I wonder if we should add the ``weights`` keyword at all...

doc/user_guide/model.rst
```diff
@@ -775,43 +791,80 @@ gaussian noise and proceed to fit as in the previous example.
...
    >>> s.estimate_poissonian_noise_variance(
-   ...     expected_value=hs.signals.Signal1D(np.arange(300)))
-   >>> m.fit()
+   ...     expected_value=hs.signals.Signal1D(np.arange(300))
+   ... )
+   >>> m.fit(weights="inverse_variance")
    >>> line.coefficients.value
    (1.0004224896604759, -0.46982916592391377)
    >>> line.coefficients.std
    (0.0055752036447948173, 0.46950832982673557)
```
francisco-dlp, line 808: It would be good to explain here that the variance can be set directly in metadata. Actually, probably it would be good to add a ``set_variance`` method that takes a hyperspy signal or np array and set the metadata entry for convinience.

It would be good to mention that the variance is only uses with ``method=="ls"``

doc/user_guide/model.rst
```diff
@@ -659,65 +659,66 @@ Fitting the model to the data
...
+    +------------------------------+--------+-----------+--------+----------------+--------+
+
+\* ``mpfit`` is deprecated since HyperSpy 1.6. Use ``"leastsq"`` instead.
+
+.. note::
+
+    **All** of the fitting algorithms available in :py:func:`scipy.optimize.minimize` are currently
+    supported by HyperSpy. The options listed above are generally sufficient for most applications,
+    but for more information on the other supported local and global optimization algorithms,
+    please see the `SciPy documentation <http://docs.scipy.org/doc/scipy/reference/optimize.html>`_.
```
francisco-dlp, line 706: What about adding a paragraph here to explain that HyperSpy supports several loss functions, including the custom one, and the pass to give details of each one. Maybe it'll be good to put all the subsections below (except for the gradient one)  under a Fitting methods or Fitting Loss Functions subsections for clarity.

doc/user_guide/model.rst
```diff
@@ -775,43 +791,80 @@ gaussian noise and proceed to fit as in the previous example.
...
-(``'ml'``) optimisation functions, a custom function can be passed to the
-model:
+   >>> # Gradient information is also supported
+   >>> m.fit(fitter="L-BFGS-B", method="huber", grad=True)
+
+
+Custom optimizations
+^^^^^^^^^^^^^^^^^^^^
+
+As well as the built-in ``ls``, ``ml`` and ``huber`` optimization functions,
```
francisco-dlp: It would be good to mention that currently analytical gradients are not supported for custom loss-functions.

doc/user_guide/model.rst
```diff
@@ -659,65 +659,66 @@ Fitting the model to the data
...
+    +==============================+========+===========+========+================+========+
+    | ``"leastsq"``                |  Yes   | Yes       | Yes    | Only ``"ls"``  | local  |
+    +------------------------------+--------+-----------+--------+----------------+--------+
+    | ``"mpfit"`` *                |  Yes   | Yes       | Yes    | Only ``"ls"``  | local  |
+    +------------------------------+--------+-----------+--------+----------------+--------+
+    | ``"odr"``                    |  No    | Yes       | Yes    | Only ``"ls"``  | local  |
+    +------------------------------+--------+-----------+--------+----------------+--------+
+    | ``"Nelder-Mead"``            |  No    | No        | No     | All            | local  |
+    +------------------------------+--------+-----------+--------+----------------+--------+
+    | ``"L-BFGS-B"``               |  Yes   | Yes       | Yes    | All            | local  |
```
francisco-dlp: If we decide to remove the conf limits estimation for L-BFGS-B as I suggest below, Errors should be back to ``False`` here.

hyperspy/model.py
```diff
@@ -924,475 +956,698 @@ def red_chisq(self):
...
+            _logger.info("Setting weights to 1/variance of signal noise")
+
+            # Note that we square this later in self._errfunc_sq()
+            weights = 1.0 / np.sqrt(variance)
+
+        return weights
+
+    def fit(
+        self,
+        fitter="leastsq",
```
francisco-dlp: What about deprecating ``fitter`` in favour of ``optimizer``?

hyperspy/model.py
```diff
@@ -924,475 +956,698 @@ def red_chisq(self):
...
+                  (MLE). Not available if ``fitter`` is ``"leastsq"``
+                  or ``"odr"``.
+                * "huber" minimize the Huber loss function. Not available
+                  if ``fitter`` is ``"leastsq"`` or ``"odr"``.
+                * "custom" allows passing your own minimisation function as a
+                  keyword argument ``min_function``, with an optional
+                  gradient argument ``min_function_grad``. Not available
+                  if ``fitter`` is ``"leastsq"`` or ``"odr"``.
+
+        grad : bool or ["2-point", "3-point", "cs"] or callable, default False
```
francisco-dlp: As it was, ``grad`` was a switch to use the analytical gradient. When ``False`` the numerical gradient was computed internally for the optimizers that use the gradient. It think that now it is not clear what happens when ``grad`` is ``False``.

hyperspy/model.py
```diff
@@ -727,44 +755,48 @@ def _set_p0(self):
...
-                        limits[1] = param.bmax
-                    if param._number_of_elements == 1:
-                        self.mpfit_parinfo.append(
-                            {'limited': limited,
-                             'limits': limits})
-                    else:
-                        self.mpfit_parinfo.extend((
-                            {'limited': limited,
-                             'limits': limits},) * param._number_of_elements)
+    def set_mpfit_parameters_info(self, bounded=True):
```
francisco-dlp, line 809: Maybe this one could be privatized too?

hyperspy/model.py
```diff
@@ -924,475 +956,698 @@ def red_chisq(self):
...
+                p_var.fill(np.inf)
+                warn_cov = True
+        else:
+            raise ValueError(f"pcov should be None or np.ndarray, got {type(pcov)}")
+
+        if warn_cov:
+            _logger.warning(
+                "Covariance of the parameters could not be estimated. "
+                "Estimated parameter standard deviations will be np.inf. "
+                "This could indicate that the model is a poor description "
```
francisco-dlp: Is that always true? If not, I suggest to remove it not to worry the users unnecessarily.

hyperspy/model.py
```diff
@@ -924,475 +956,698 @@ def red_chisq(self):
...
+
+        return weights
+
+    def fit(
+        self,
+        fitter="leastsq",
+        method="ls",
+        grad=False,
+        bounded=False,
+        ext_bounding=False,
```
francisco-dlp: That that you are at it, I think that we can deprecate ``ext_bounding``, this dates from the times when there was no optimizer with bound support available and is a (very) poor man's solution to the problem.

hyperspy/model.py
```diff
@@ -924,475 +956,698 @@ def red_chisq(self):
...
+                * "Dual Annealing" is a global optimization method.
+                  It does support bounds on parameters. See
+                  :py:func:`scipy.optimize.dual_annealing` for more
+                  details on available options. Requires ``scipy >= 1.2.0``.
+                * "SHGO" (simplicial homology global optimization" is a global
+                  optimization method. It does support bounds on parameters. See
+                  :py:func:`scipy.optimize.shgo` for more details on available
+                  options. Requires ``scipy >= 1.2.0``.
+
+        method : {"ls", "ml", "huber", "custom"}, default "ls"
```
francisco-dlp: As suggested above, what about ``method`` -> ``loss_function``?

hyperspy/model.py
```diff
@@ -924,475 +956,698 @@ def red_chisq(self):
...
+                If the attribute ``metada.Signal.Noise_properties.variance``
+                is defined as a ``Signal`` instance of the same
+                ``navigation_dimension`` as the signal, and ``method="ls"``,
+                then weighted least-squares fit is performed, using
+                the inverse of the noise variance as the weights.
+            If np.ndarray:
+                If the array has the same dimensions as the signal,
+                and ``method="ls"``, then weighted least squares
+                is performed using the array values as the weights.
+        print_info : bool, default False
```
francisco-dlp: Wouldn't it be better a ``return_info`` keyword that returns a ``OptimizeResult`` object with a nice ``__repr__`` method or even a subclass of it with html printing? In this way the user can more easily reuse that data.

For multifit this could return a similear ``OptimzeResult`` object with suitable HyperSpy signals as attributes for easy inspection of the result (the current print, as you point out, is quite useless for multi-dimensional fitting).

hyperspy/model.py
```diff
@@ -924,475 +956,698 @@ def red_chisq(self):
...
+            optimizer of your choice in :py:mod:`scipy.optimize`.
+
+        Returns
+        -------
+        None
+
+        Notes
+        -----
+        The chi-squared and reduced chi-squared statistics, and the
+        degrees of freedom, are computed automatically when fitting.
```
francisco-dlp: Maybe it'll be good to add here that they are computed only when fitting using least-squares.

hyperspy/model.py
```diff
@@ -924,475 +956,698 @@ def red_chisq(self):
...
 
         # Check for deprecated minimizers
-        optimizer_dict = {"fmin": "Nelder-Mead",
-                          "fmin_cg": "CG",
-                          "fmin_ncg": "Newton-CG",
-                          "fmin_bfgs": "BFGS",
-                          "fmin_l_bfgs_b": "L-BFGS-B",
-                          "fmin_tnc": "TNC",
-                          "fmin_powell": "Powell"}
+        optimizer_dict = {
```
francisco-dlp: Maybe ``deprecated_optimizer_dict`` for clarity?

hyperspy/model.py
```diff
@@ -924,475 +956,698 @@ def red_chisq(self):
...
+                    if grad is True:
+
+                        def _wrap_jac(*args, **kwargs):
+                            # Jacobian function computes derivatives along
+                            # columns, so we need the transpose instead
+                            return self._jacobian(*args, **kwargs).T
+
+                        jacobian = _wrap_jac
+
+                    elif grad is False or grad is None:
```
francisco-dlp: Then maybe make ``"2-point"`` the default and deprecate ``None`` and ``False``?

francisco-dlp (changes requested):
This is a much needed improvement of the fitting documentation and code. I have left quite a few comments, mainly with some extra suggestions now that you are at it.

tjof2:
Thanks @francisco-dlp for the detailed review - will work through it over the weekend!

tjof2:
Thanks @francisco-dlp for the detailed review - will work through it over the weekend!

doc/user_guide/model.rst
tjof2: Done

hyperspy/model.py
tjof2: Done

hyperspy/models/eelsmodel.py
tjof2: Agreed - done.

hyperspy/model.py
tjof2: Agreed - done.

hyperspy/model.py
tjof2: This is consistent with how `scipy.optimize.curve_fit` does it. I can replace it with `np.nan` if you prefer, but I think `np.inf` is fine.

hyperspy/model.py
tjof2: Agreed - will add.

tjof2: Done

hyperspy/models/eelsmodel.py
tjof2: Done

hyperspy/model.py
tjof2: Agreed - done

hyperspy/model.py
tjof2: Well, the use of "could" qualifies it, but I agree, it could be an unnecessary warning in that sense - removed.

hyperspy/model.py
tjof2: Sure - have done so.

hyperspy/model.py
tjof2: Yes good idea - done.

hyperspy/model.py
tjof2: I'll take a look at seeing if `self.fit_output` can be standardized into an `OptimizeResult` object for the other fitting methods.

hyperspy/model.py
tjof2: The alternative is to just use `hess_inv.todense()` as an estimate of the covariance matrix and use the same procedure as `leastsq` - rather than do `np.sqrt(max(1.0, np.abs(cost)) * ftol * hess)`. But that's not particularly clean etc. So I'll remove it as you suggest. 

Taking the lmfit approach (or another) is probably best left for a separate PR.

hyperspy/model.py
tjof2: Good catch - will resolve.

doc/user_guide/model.rst
tjof2: Yep!

doc/user_guide/model.rst
tjof2: Yep - done.

doc/user_guide/model.rst
tjof2: Hmmm, that'll probably be because `mpfit` uses the Levenberg–Marquardt algorithm with its own implementation of bounds, while `leastsq` with bounds actually uses the Trust Region Reflective algorithm (`method="trf"`).

There are a couple of options here, both of which could preserve `mpfit`.

1. Deprecate the options `["leastsq", "mpfit"]` with e.g. `["lm", "trf", "dogbox"]` with clearer documentation.
2. Don't do anything, leave `"mpfit"` as it is.

Option (1) is nice because the default fitter for `m.fit()` could then be `None`, and we could smartly suggest an appropriate method depending on the type of function, bounds, and gradients.

Option (2) is OK, but I think it would be better to deprecate `mpfit` for `lm` (Levenberg–Marquardt).

hyperspy/models/model1d.py
tjof2: Yep easy to resolve - done.

hyperspy/models/model1d.py
tjof2: Done

hyperspy/models/model1d.py
tjof2: Yep changed

doc/user_guide/model.rst
tjof2: Yep - changed.

doc/user_guide/model.rst
tjof2: > Why would one want to disable weighting if the variance is defined? Also, wouldn't that be equivalent to remove variance from metadata?

Yes - but for example, say you estimated the Poisson noise, and wanted to compare the results of weighted vs. unweighted fits.

I would say it is easier to just run `m.fit()` and `m.fit(weights=None)` to compare the results, rather than estimating the variance then unsetting it and (potentially) having to set it again.

doc/user_guide/model.rst
tjof2: Yep - std is correct here.

doc/user_guide/model.rst
tjof2: I've added a `set_noise_variance()` method for convenience.

> It would be good to mention that the variance is only uses with `method=="ls"`

Is this true? Currently it says:
```
    weights = None

    if weights == "inverse_variance":
        weights = self._convert_variance_to_weights()

    args = (self.signal()[np.where(self.channel_switches)], weights)
```
and then `args` is passed to the optimizer, be it `leastsq`, `odr`, `mpfit` or the general `minimize()`

hyperspy/model.py
tjof2: Sure!

doc/user_guide/model.rst
tjof2: Agreed.

doc/user_guide/model.rst
ericpre: Since mpfit is vendored and it is not supported since a while, at some point, it may break. If we don't have a good reason to keep mpfit, we should deprecate it.

tjof2: With a little bit of work it can probably support passing a custom minimizer (like we did for `s.decomposition()` and sklearn objects), so you could ultimately use `lmfit`, `scikit-optimize` or `openopt/copt` if desired. That could be option 3.

doc/user_guide/model.rst
francisco-dlp: I see, but the question is: why would you want to compare one with the other when one way is right and the other is not? Some may want to do for, say, educational purposes, but I don't think that it is worth adding a feature for that since that minority can always simply delete the variance metadata.

doc/user_guide/model.rst
francisco-dlp: In that case, I think that there is a bug in there: the weights should only be applied with method is LS to turn it into WLS. If we apply the weights when using Poisson-ML we get wrong results. Applying weights probably makes sense with the Huber-loss function though...

tjof2: I'll tidy it up!

doc/user_guide/model.rst
tjof2: Sure. I'll tidy it up

hyperspy/models/eelsmodel.py
tjof2: Done.

doc/user_guide/model.rst
francisco-dlp: That sounds excellent, but maybe something for a different PR? A possible problem (for which solutions exist) with that approach is that it won't work for parallel distributed fits since the library will need to be present in all the nodes.

Regarding mpfit, as mentioned above, in the past, I have obtained better results with it than with bounded ``scipy.optimize.least_squares``, so, unless we show that this not actually generally the case, I suggest that we keep it unless it gets broken beyond easy repair at some point. it seems that a maintained version of the code can be found in https://github.com/segasai/astrolibpy . However, it is not distributed in pypi as far as I can tell. Ours is also a maintained version of the code and so-far it has been easy to fix it every time that changes in numpy broke something. For the future (i.e. after The Split) If we want to continue offering mpfit, I suggest to contact @segasai to check if he would be happy to publish his code (that is the original code from which we forked our version a long ago) in pypi and conda-forge.

I like @tjof2's 1. option. Currently it seems that even for unbounded optimization the default bounding method is "trf", while it would be more efficient to use "lm". An ``auto`` mode could fix this, while leaving the user the option to choose a different method if required. If we keep mpfit, we could add ``mpfit`` to the list of bounds methods. Also, the optimizer in that case could be "LM" instead of "leastsq" as it is now, since "leastsq" is rather the loss-funtion and we could use all other optimizers to solve the same problem.

tjof2: Thanks @francisco-dlp  - I also like option 1. I'll work on that.

doc/user_guide/model.rst
tjof2: Sure.

hyperspy/model.py
tjof2: Agreed - done.

hyperspy/model.py
francisco-dlp: The things is that in hyperspy several functions (e.g. ``print_summary_statistics``, 2D plotting constrast adjustement...) know how to deal with ``nans`` but not with ``inf``s.

doc/user_guide/model.rst
tjof2: Done

hyperspy/model.py
tjof2: OK, I'll replace with nan in that case.

doc/user_guide/model.rst
ericpre: OK, maybe it would be worth trying to sort out the distribution of mpfit or even better integrate it in scipy!

hyperspy/model.py
ericpre: Would it be better to fix the functions which doesn't support `np.inf` to keep the consistency with scipy?

francisco-dlp: I don't know how to best do that. Actually, I don't know why ``curve_fit`` uses ``inf`` instead of ``nan`` for this, since a ``nan`` seems to me a better description of the situation. In hyperspy: if a dataset contains ``inf``s, I think that it is up to the user to decide how to best handle them, because the best approach depends on the situation.

hyperspy/model.py
tjof2: Yep.

doc/user_guide/model.rst
francisco-dlp: That's a good point: if the presumed advantage comes from a particular way of implementing the bounds, maybe that can be added to scipy's ``least_squares`` as an extra bounds method.

hyperspy/model.py
tjof2: I've got a few ideas for this - I'll test them out.

hyperspy/model.py
tjof2: I've added a `return_info` argument and it will always return an `OptimizeResult` object for all fitting methods.

francisco-dlp: What would be best, to store i. this info in ``fit_output`` as present ii. return it optionally, replacing ``print_info`` as I suggested above and deprecate ``fit_output``?

Actually, ``fit_output`` is currently not stored anywhere when saving the model, what I think is a bug. If we take option i, then we won't need ``print_info`` or ``return_info`` in ``multifit``, but maybe we could consider a ``store_info`` keyword to make it optional. Also we should consider storing this info when saving the model. In my opinion option 1 is better because it is consistent with the fact that we also store other fitting info, such as the chisq, as model attributes.

doc/user_guide/model.rst
tjof2: I think there is a (small) bug in the calculation of `m.p_std` for mpfit, it treats the `perror` variable from mpfit as being the estimated covariance matrix, when it isn't - its the estimated parameter standard deviations (prior to multiplying by any cost function). So if your assertion "mpfit gave better results" is based purely off the values in `p_std`, it might not necessarily be better. It's easy to fix though and may not necessarily be a problem.

tjof2:
Last commit was a bit of a mess - will update tomorrow

doc/user_guide/model.rst
tjof2: I've found a nicer way to keep mpfit rather than deprecating (for now)

hyperspy/model.py
tjof2: To be fair I think nan is probably better here after some though - it's not that the parameter estimate is very (i.e. infinitely) large, rather, it cannot be estimated accurately and therefore should not be relied upon.

hyperspy/model.py
tjof2: Ok, so let's store it and not return it. I think the print is a useful option still.

doc/user_guide/model.rst
tjof2: Just to be clear here - in removing the `weights` kwarg @francisco-dlp, are you happy for #1627 to just be closed as it is? Perhaps with some clearer documentation that once the variance is set, all fits are weighted until it is unset?

doc/user_guide/model.rst
tjof2: Yep - working on that

doc/user_guide/model.rst
francisco-dlp: I agree that it is a good idea to close #1627 by just documenting how to do it.

tjof2: Great thanks - I will do that.

tjof2:
I've addressed most of the comments now @francisco-dlp. 

Here are a few outstanding things which I personally think are better tackled in separate PRs to ensure they are done properly.

1. Storing and saving `model.fit_output`.
2. Concatenating/grouping/storing `model.fit_output` when calling `model.multifit()` - it could potentially get quite large for lots of pixels so we may only want to store some of the info
3. Using `grad="analytical"` by default if all the components implement it. This *changes* the default behaviour and hence the fitting results compared to current/previous HyperSpy versions (evidenced by a whole bunch of test failures elsewhere if I try it), whereas this PR only deprecates the old choice of `grad=True/False` without changing the end results for now. Doing it in a separate PR would be cleaner to make sure the change is correct, and it could be profiled to make sure e.g. it's not a performance regression.

Here, I've used `grad="auto"` to mean auto-differentiation where available. Perhaps a better keyword is `grad="fd"` for finite-difference, and preserve `"auto"` for item (3)? That ties in better with the kwarg `fd_scheme="2-point"` etc.

doc/user_guide/model.rst
tjof2: > Since mpfit is vendored and it is not supported since a while, at some point, it may break. If we don't have a good reason to keep mpfit, we should deprecate it

The code is actually much simpler than it looks @ericpre. It probably looks bad due to all the old IDL comments, old Fortran comments and old Python style (or lack of...). I've actually tidied it up a lot, I can easily push it in a separate PR. Since it's written in pure Python (unlike `scipy.optimize.least_squares` which wraps Fortran), it can be a little slower when optimizing, but the bound constraints are useful, so keeping it (for now at least) makes sense.

doc/user_guide/model.rst
```diff
@@ -659,65 +653,88 @@ Fitting the model to the data
...
-                    methods changed to the `scipy.optimze.minimize()` notation.
+spectrum at a particular point in a spectrum-image), use
+:py:meth:`~.model.BaseModel.fit`. HyperSpy implements a number of
+different optimization approaches, each of which can have particular
+benefits and/or drawbacks depending on your specific application.
+A good approach to choosing an optimization approach is to ask yourself
+the question "Do you want to...":
+
+* Apply bounds to your model parameter values?
+* Are gradient-based fitting algorithms appropriate for your model?
```
thomasaarholt: "Do you want to..." "are gradient-based fitting..." doesn't really work. 
Maybe: "Use gradient-based fitting algorithms?"

doc/user_guide/model.rst
```diff
@@ -659,65 +653,88 @@ Fitting the model to the data
...
+    | ``"SHGO"`` **                        |  Yes   | No        | No     | All            | global |
+    +--------------------------------------+--------+-----------+--------+----------------+--------+
+
+.. note::
+
+    \* **All** of the fitting algorithms available in :py:func:`scipy.optimize.minimize` are currently
+    supported by HyperSpy; however, only some of them support bounds and/or gradients. For more information,
+    please see the `SciPy documentation <http://docs.scipy.org/doc/scipy/reference/optimize.html>`_.
+
+    \*\* Requires ``scipy >= 1.2.0``.
```
thomasaarholt, line 707: Scipy 1.5.2 is out, and I notice that hyperspy [setup.py currently requires scipy 0.15](https://github.com/hyperspy/hyperspy/blob/fe7602e4ff9ae51a86c2a4c0d0545971ef972ae3/setup.py#L48) (released 2015). Maybe we can just set scipy 1.2 as required version? I notice conda installs scipy 1.5.0.

doc/user_guide/model.rst
```diff
@@ -659,65 +653,88 @@ Fitting the model to the data
...
+
+The default optimizer in HyperSpy is ``"lm"``, which stands for the `Levenberg-Marquardt
+algorithm <https://en.wikipedia.org/wiki/Levenberg%E2%80%93Marquardt_algorithm>`_. In
+earlier versions of HyperSpy (< 1.6) this was known as ``"leastsq"``.
+
+Loss functions
+^^^^^^^^^^^^^^
+
+HyperSpy supports a number of loss functions. The default is ``"ls"``,
+or the least-squares loss. For the vast majority of cases, this loss
```
thomasaarholt: Remove the "or" -> "The default is ``"ls"``, the least-squares loss."

thomasaarholt (commented):
Docs look great, only some minor things.

doc/user_guide/model.rst
tjof2: I have no objection, I was conservative in advancing it to 0.17.0 here. If @francisco-dlp and @ericpre have no objections/reasons why this is a bad idea, then we can advance it further

ericpre: Ubuntu 18.04 LTS have [scipy 0.19.1](https://packages.ubuntu.com/search?suite=default&section=all&arch=any&keywords=python3-scipy). We are not dealing dependencies requirement consistently but since there is no significant advantages to change it, I would keep it as broad as possible. Also, I suspect that scipy is one our dependencies which is the most difficult to build.

tjof2: I'll keep it at 0.17

doc/user_guide/model.rst
```diff
@@ -917,7 +1010,7 @@ component using mpfit and bounds on the ``centre`` parameter.
     >>> g1.centre.bmin = 7
     >>> g1.centre.bmax = 14
     >>> g1.centre.bounded = True
```
francisco-dlp: I think that we can remove this line, since the ``bounded`` attribute is now gone.

doc/user_guide/model.rst
```diff
@@ -775,43 +790,77 @@ gaussian noise and proceed to fit as in the previous example.
...
+        >>> line.coefficients.value
+        (1.0052331707848698, -1.0723588390873573)
+
+Poisson maximum likelihood estimation
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+To avoid biased estimation in the case of data corrupted by Poisson noise
+with very few counts, we can use Poisson maximum likelihood estimation (MLE) instead.
+This is an unbiased estimator for Poisson noise. To perform MLE, we must
+use a general, non-linear optimizer from the ref:`table above <optimizers-table>`,
```
francisco-dlp: this link to the table is broken.

doc/user_guide/model.rst
```diff
@@ -884,26 +935,68 @@ passed, using the following signature:
...
+Using gradient information
+^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+Optimization algorithms that take into account the gradient of
+the loss function will often out-perform so-called "derivative-free"
+optimization algorithms in terms of how rapidly they converge to a
+solution. HyperSpy can use analytical gradients for model-fitting,
+as well as numerical estimates of the gradient based on finite differences.
+
+If all the components in the model support analytical gradients,
```
francisco-dlp, line 955: I think that we need a ``versionadded`` directive here.

hyperspy/model.py
```diff
@@ -924,475 +1010,672 @@ def red_chisq(self):
...
+            # has been set, otherwise this returns None
+            weights = self._convert_variance_to_weights()
+
+            if weights is not None and loss_function == "ml-poisson":
                 raise ValueError(
-                    'method must be "ls", "ml" or "custom" but %s given' %
-                    method)
-            args = (self.signal()[np.where(self.channel_switches)],
-                    weights)
+                    "Weighted fitting is not supported for `loss_function='ml_poisson'`. "
```
francisco-dlp: I think that in this case, it is better to perform the operation without weights. Indeed a typical workflow is:

* Get close to the optimal parameters using WLS
* Refine with ML-Poisson.

It would be rather inconvenient having to delete the variance to perform these operations.

doc/user_guide/model.rst
```diff
@@ -928,17 +1021,60 @@ component using mpfit and bounds on the ``centre`` parameter.
              sigma |  True | 0.00999184 | 2.68064163 |       None |       None
             centre |  True | 9.99980788 | 2.68064070 |        7.0 |       14.0
 
+
+Optimization results
+^^^^^^^^^^^^^^^^^^^^
+
+After fitting the model, details about the optimization
+procedure, including whether it finished successfully,
+are stored in the attribute ``fit_output`` as a
```
francisco-dlp: In the future, we may add the feature of storing the multi-dimensional output instead of the last fit output in ``fit_output``. Planning for that, what would you think about replacing ``print_info`` with ``return_info``? I bet that it'll print the info all the same, with the added advantage that it makes it easier to explore the output of the latest fit.

hyperspy/docstrings/model.py
```diff
@@ -0,0 +1,99 @@
...
+                  of the Huber function is controlled by the ``huber_delta``
+                  keyword argument (the default value is 1.0).
+                * callable supports passing your own minimization function.
+
+        grad : {"auto", "analytical", callable, None}, default "auto"
+            Whether to use information about the gradient of the loss function
+            as part of the optimization. This parameter has no effect if
+            ``optimizer`` is a derivative-free or global optimization method.
+
+                * "auto" uses a finite difference scheme (if available) for numerical
```
francisco-dlp: I think that it'll be clearer to rename `"auto"` -> `"fd"`

hyperspy/docstrings/model.py
```diff
@@ -0,0 +1,99 @@
...
+# along with  HyperSpy.  If not, see <http://www.gnu.org/licenses/>.
+
+"""Common docstring snippets for model.
+
+"""
+
+FIT_PARAMETERS_ARG = """optimizer : str or None, default None
+            The optimization algorithm used to perform the fitting.
+
+                * "lm" performs least-squares optimization using the
```
francisco-dlp, line 26: Although I see the point of dividing each bounding approach into separate optimizers, I think that this can lead to conceptual confusion. Why not using ``"lm"`` for the optimizer and then use ``bounded`` to choose between ``"trf", "dogbox" and "mpfit"?

By the way, you've made mpfit the default optimizer for bounded-LM, is that because you've performed some tests and you've found it to be superior to the others?

francisco-dlp (changes requested):
This looks almost ready to me, except for the couple of things that I mention below.

doc/user_guide/model.rst
tjof2: Done

doc/user_guide/model.rst
tjof2: Done

doc/user_guide/model.rst
tjof2: Done

hyperspy/docstrings/model.py
tjof2: Done

hyperspy/docstrings/model.py
tjof2: But "trf" and "dogbox" aren't the Levenberg-Marquardt algorithm, so that doesn't make sense.

> mpfit the default optimizer for bounded-LM

It's the _only_ bounded Levenberg-Marquardt option.

"trf" and "dogbox" are alternative least-squares minimization algorithms, which do support bounds and (optionally) sparse Jacobians. But they are not the Levenberg-Marquardt algorithm (`optimizer="lm"`).

The approach taken here is the closest one for preserving backwards-compatibility/replicating previous behaviour.

hyperspy/model.py
tjof2: Ok, perhaps change this to a logger warning (i.e. that the noise variance is set but weighted fitting won't actually be done), and set `weights = None` to proceed then?

hyperspy/docstrings/model.py
francisco-dlp: You are right, scipy's API, where you get those options in ``method`` (what I (mis)understood as bounding_method), led me to conceptual misunderstanding, so clearly your approach is clearer :)

hyperspy/model.py
tjof2: Done.

francisco-dlp: For once I think that doing nothing is better: users don't like warnings and this one seems unnecessary to me: somebody using ML-Poisson should know what he's doing...

tjof2: OK will do!

doc/user_guide/model.rst
tjof2: I agree in principle. What about for consistency:

- `s.decomposition()` has `print_info=True/False` and `return_info=True/False` (implemented in #2336 and #2383 )
- `m.fit()` has `print_info=True/False` and `return_info=True/False` (implemented here?)

Note that `print_info` as it currently stands does also highlight a few of the key options you passed to `m.fit()`, so for example if in future we went with automatically choosing the `optimizer` or `grad`, you could print that out. While `m.fit_output` is simply an `OptimizeResult` object at the moment, with no additional details.

tjof2: I agree in principle. What about for consistency:

- `s.decomposition()` has `print_info=True/False` and `return_info=True/False` (implemented in #2336 and #2383 )
- `m.fit()` has `print_info=True/False` and `return_info=True/False` (implemented here?)

Note that `print_info` as it currently stands does also highlight a few of the key options you passed to `m.fit()`, so for example if in future we went with automatically choosing the `optimizer` or `grad`, you could print that out. While `m.fit_output` is simply an `OptimizeResult` object at the moment, with no additional details.

For example:
```
    Fit info:
      optimizer=L-BFGS-B
      loss_function=ls
      bounded=False
      grad="fd"
```

The printing also currently drops these two attributes of `OptimizeResult`
```
            # Drop these as they can be large (== size of data array)
            output_print.pop("fun", None)
            output_print.pop("jac", None)
```
Which if you have a large array, ends up printing a lot of output. But they can be returned of course.

tjof2: `print_info` could also be used to control the various `verbose/quiet` options of the different optimizers (in future)

tjof2: As it stands, I've just pushed a commit that implements `return_info`.

PR #2444, opened 2020-07-15
Fix offset signal EMPAD.

ericpre:
Fix offset value of the axis: the offset wasn't calibrated.


tjof2:
@ericpre is this done save for the failed test? Also perhaps worth recording here _why_ the offset needed fixing?

ericpre:
> @ericpre is this done save for the failed test? Also perhaps worth recording here _why_ the offset needed fixing?

Sorry, I don't understand! I have edited the first comment of this PR to mention what needed to be fixed.

tjof2:
Yes that makes sense!

PR #2445, opened 2020-07-18
Check signal specification conflicts

ericpre:
For https://github.com/hyperspy/hyperspy-extensions-list/pull/3, I think when we add an package to the extension list, it would be useful to check if this package introduce signal specifications conflicts. I will add tests and update the dev guide if we agree on the approach.

### Progress of the PR
- [x] If several signal class match are found, raise a warning and set to a generic signal since we don't have a reason to set it to one or another,
- [x] add function to check for signal specification conflicts,
- [ ] update dev guide,
- [ ] add tests,
- [x] ready for review.

### DIFF
--- a/hyperspy/io.py
+++ b/hyperspy/io.py
@@ -21,7 +21,7 @@ import glob
 import warnings
 import logging
 import importlib
-import copy
+import pprint
 
 import numpy as np
 from natsort import natsorted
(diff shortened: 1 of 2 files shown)


tjof2:
Any risk of this conflicting with #2434? e.g. multiple warning messages or one branch of code never being called?

ericpre:
Yes, these changes should be conflicting with #2434. I guess, we should complete #2434 and then solve the merge conflict in this PR.

tjof2:
That's waiting on https://github.com/hyperspy/hyperspy-extensions-list/pull/2 so hopefully there's no circular dependency!!

hyperspy/io.py
```diff
@@ -478,9 +478,8 @@ def _find_matching_subclass(dtype,
           'object' in dtype.name):
         dtype = 'real'
     else:
-        raise ValueError('Data type "{}" not understood!'.format(dtype.name))
-
-    if not isinstance(signal_dimension, int):
+        raise ValueError(f'Data type "{dtype.name}" not understood!')
+    if not isinstance(signal_dimension, int) or signal_dimension < 0:
         raise ValueError("signal_dimension must be an interger")
```
tjof2, line 483: Typo "integer". Also worth saying it must be >= 0?

PR #2446, opened 2020-07-18
Fix matplotlib 3.3.0

ericpre:
Fix issue when specifying legend in `hs.plot.plot_spectra`, which is broken with matplotlib 3.3.0 and this is why travis is hanging.

### Progress of the PR
- [x] fix legend,
- [x] fix deprecation warning,
- [x] ready for review.

### DIFF
--- a/hyperspy/drawing/_widgets/horizontal_line.py
+++ b/hyperspy/drawing/_widgets/horizontal_line.py
@@ -17,6 +17,7 @@
 # along with  HyperSpy.  If not, see <http://www.gnu.org/licenses/>.
 
 from hyperspy.drawing.widgets import Widget1DBase
+from hyperspy.drawing.utils import picker_kwargs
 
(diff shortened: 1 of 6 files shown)


tjof2 (approved):

francisco-dlp:
LGTM. Just a small question: why not requiring mpl >= 3.3.0 instead of adding logic to maintain compatibility with old versions?

ericpre:
I think we can't support only the last version of matplotlib, if other libraries are not supporting the last version, this would be problematic.

francisco-dlp:
I see how that can be a problem when packaging hyperspy for linux since most distributions' python packages tend to lag a bit behind.

tjof2:
Can we get this into Rnm too to fix Travis?

ericpre:
It is on RnM now.

PR #2449, opened 2020-07-23
Rename hyperspy/tests/mva/ to hyperspy/tests/learn/

tjof2:
### Description of the change
Purely for consistency, all the decomposition/BSS code lives in `hyperspy/learn/*`, so the tests are moved to be in `hyperspy/tests/learn/*`


ericpre:
I came across about it in the past and thought it would be good to rename it!

Issue #2448, opened 2020-07-22
incomplete format string in _get_element_and_line()


jonnyprouty:
[hyperspy/misc/eds/utils.py](https://github.com/hyperspy/hyperspy/blob/RELEASE_next_minor/hyperspy/misc/eds/utils.py) contains an error in _get_element_and_line(). Specifically, ValueError is passed an incomplete format string. The following code demonstrates the issue:
```
s = hs.signals.Signal1D(np.random.random((10,100)))
s.set_signal_type("EDS_SEM")
s.get_lines_intensity("Ca_Kb") # should raise KeyError

```
Returns the following:
```
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
<ipython-input-14-9a0d450cf112> in <module>
      1 s = hs.signals.Signal1D(np.random.random((10,100)))
      2 s.set_signal_type("EDS_SEM")
----> 3 s.get_lines_intensity("Ca_Kb") # should raise KeyError

~/.local/anaconda3/lib/python3.7/site-packages/hyperspy/_signals/eds.py in get_lines_intensity(self, xray_lines, integration_windows, background_windows, plot_result, only_one, only_lines, **kwargs)
    618         """
    619 
--> 620         xray_lines = self._parse_xray_lines(xray_lines, only_one, only_lines)
    621         if hasattr(integration_windows, '__iter__') is False:
    622             integration_windows = self.estimate_integration_windows(

~/.local/anaconda3/lib/python3.7/site-packages/hyperspy/_signals/eds.py in _parse_xray_lines(self, xray_lines, only_one, only_lines)
    523                                           only_lines=only_lines)
    524         xray_lines, xray_not_here = self._get_xray_lines_in_spectral_range(
--> 525             xray_lines)
    526         for xray in xray_not_here:
    527             warnings.warn("%s is not in the data energy range." % xray +

~/.local/anaconda3/lib/python3.7/site-packages/hyperspy/_signals/eds.py in _get_xray_lines_in_spectral_range(self, xray_lines)
    156         xray_lines_not_in_range = []
    157         for xray_line in xray_lines:
--> 158             line_energy = self._get_line_energy(xray_line)
    159             if low_value < line_energy < high_value:
    160                 xray_lines_in_range.append(xray_line)

~/.local/anaconda3/lib/python3.7/site-packages/hyperspy/_signals/eds.py in _get_line_energy(self, Xray_line, FWHM_MnKa)
[...]

PR #2450, opened 2020-07-24
type check for xray_lines in get_lines_intensity()

jonnyprouty:
Addresses issue #2448 where xray_lines argument type was not checked.

## Description of the change
- Change to get_lines_intensity()
Check type of xray_lines arguement and raise a TypeError if xray_lines is not None and not a list.
- Corrected typo in EDS test class Test_get_lines_intensity
- Added tests to Test_get_lines_intensity.test()
Function now checks type of xray_lines.
**Note** !!! Most arguments to get_lines_intensity() remain un-validated at runtime !!!
- Change to _unpack_data()
Fix syntax error in _unpack_data() where numerical comparison was performed incorrectly.
Bug turned up while testing get_lines_intensity() fix.

### Progress of the PR
- [x] add tests,
- [x] ready for review.

The following code shows that the new type check works as expected, but it is not a full test of get_lines_intensity().
```
import numpy as np
import hyperspy.api as hs

s = hs.signals.Signal1D(np.random.random((10,100)))
s.set_signal_type("EDS_SEM")

#should fail and raise KeyError
s.get_lines_intensity(xray_lines="Ca_Ka")
```

### DIFF
--- a/hyperspy/_signals/eds.py
+++ b/hyperspy/_signals/eds.py
@@ -20,6 +20,7 @@ import logging
 
 import numpy as np
 import warnings
+from collections.abc import Iterable
 from matplotlib import pyplot as plt
 
 from hyperspy import utils
@@ -550,7 +551,7 @@ class EDS_mixin:
 
         Parameters
         ----------
-        xray_lines: {None, list of string}
+        xray_lines: {None, Iterable* of strings}
             If None,
             if `metadata.Sample.elements.xray_lines` contains a
             list of lines use those.
@@ -560,6 +561,8 @@ class EDS_mixin:
             for the operation.
             Alternatively, provide an iterable containing
             a list of valid X-ray lines symbols.
+            * Note that while dictionaries and strings are iterable,
+            their use is ambiguous and specifically not allowed.
         integration_windows: Float or array
             If float, the width of the integration windows is the
             'integration_windows_width' times the calculated FWHM of the line.
@@ -616,6 +619,13 @@ class EDS_mixin:
         plot
 
         """
+        if xray_lines is not None and \
+            (not isinstance(xray_lines, Iterable) or \
+            isinstance(xray_lines, str) or isinstance(xray_lines, dict)):
+
+            raise TypeError(
+                "xray_lines must be a compatible iterable, but was "
+                "mistakenly provided as a %s." % type(xray_lines))
 
         xray_lines = self._parse_xray_lines(xray_lines, only_one, only_lines)
         if hasattr(integration_windows, '__iter__') is False:
--- a/hyperspy/_signals/eds.py
+++ b/hyperspy/_signals/eds.py
@@ -621,7 +621,7 @@ class EDS_mixin:
         """
         if xray_lines is not None and \
             (not isinstance(xray_lines, Iterable) or \
-            isinstance(xray_lines, str) or isinstance(xray_lines, dict)):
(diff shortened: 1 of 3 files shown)


jonnyprouty:
@ericpre, do you mind having a look at this draft pull request? Since it's my first pull request, I'd like to make sure everything is in order before I make a _real_ pull request. If you feel this is unnecessary, no worries, I can go ahead and mark this as ready for review. Otherwise, I'll address and suggestions you may have. Thanks in advance.

hyperspy/_signals/eds.py
```diff
@@ -617,6 +617,11 @@ def get_lines_intensity(self,
 
         """
 
+        if not (xray_lines is None or isinstance(xray_lines, list)):
```
ericpre: A further improvement is to support any compatible iterable type, such as numpy array and tuple, which could be done with:
```python
from collections.abc import Iterable

if isinstance(xray_lines, Iterable) and not isinstance(xray_lines, str):
   ...
```
since str object are also iterable. In this case, it would be good to update the docstring too `{None, list of string}` -> `{None, iterable of string}`.

ericpre (changes requested):
Thanks @jonnyprouty, the pull request looks like a good start! It suggested one improvement and it would be good to add a test checking that an error is raised when passing a string by using:
```python
   with pytest.raises(TypeError):
      # code raising error
```

hyperspy/_signals/eds.py
jonnyprouty: I incorporated your suggestion, with a small caveat. I'll add a comment to the new commit.

jonnyprouty:
>  it would be good to add a test checking that an error is raised when passing a string by using:
> 
> ```python
>    with pytest.raises(TypeError):
>       # code raising error
> ```

@ericpre , thanks for the suggestion. I ended up expanding the test for get_lines_intensity(). In the midst of writing the test it became clear that allowing the use of dictionaries would also complicate things a bit, so in addition to strings I have specifically disallowed dictionaries. The ambiguity in whether to use the key or the value as the xray_line of choice makes them a poor fit in my view.

hyperspy/_signals/eds.py
```diff
@@ -616,6 +619,13 @@ def get_lines_intensity(self,
         plot
 
         """
+        if xray_lines is not None and \
+            (not isinstance(xray_lines, Iterable) or \
+            isinstance(xray_lines, str) or isinstance(xray_lines, dict)):
```
ericpre: Simplify to:
```suggestion
            isinstance(xray_lines, (str, dict))):
```

ericpre (commented):
Yes, I agree about the dictionary. I have one last suggestion about making the if condition a bit more compact. Other than that, it looks good to me!

jonnyprouty:
> I have one last suggestion about making the if condition a bit more compact.

Ah, yes, that didn't occur to me, good catch!

> Other than that, it looks good to me!

Thanks very much for the help, I really do appreciate it.

ericpre (approved):

ericpre:
For each release, a DOI is generated by zenodo and the name in Github profile are automatically used for the contributors list. Do you want to update your name? If the name field is not provided, the github name is used instead.

jonnyprouty:
> Do you want to update your name? If the name field is not provided, the github name is used instead.

I just updated it, thanks for the heads up.

Issue #1480, opened 2017-02-22
Why is import hyperspy.api slow?


thomasaarholt:
Whilst I fully appreciate that part of the reason that hs is slow to import is that it hasn't been a priority to speed it up. It's fast enough. Still, considering that hyperspy only really needs a few dependencies (mentioned in #1438), and each of them imports almost instantaneously, I wonder why `hyperspy.api` takes longer. In fact, `import hyperspy.api` takes 3 times longer than `import numpy, scipy, matplotlib.pyplot, h5py, sklearn, IPython, traits, traitsui, ipyparallel`.

This isn't a big deal - I've just been restarting the kernel often and been wondering.
Perhaps some of the (optional) dependencies could be imported when necessary?

to266:
To me it seems like (yet another) traits thing (I use TK because pyqt5 is broken, traits again).

(when just importing traits/traitsui I doubt those libraries do anything, whereas in our case they check/verify the local preferences that may also be read from the disk. `slugify` is also infamous for slowing things down, although at some point it went through a serious training regime to speed it up).


```python
In [1]: %matplotlib tk

In [2]: %prun import hyperspy.api as hs
         1548372 function calls (1499419 primitive calls) in 1.506 seconds

   Ordered by: internal time

   ncalls  tottime  percall  cumtime  percall filename:lineno(function)
     1542    0.171    0.000    0.171    0.000 {built-in method marshal.loads}
3432/3349    0.118    0.000    0.469    0.000 {built-in method builtins.__build_class__}
   2052/1    0.075    0.000    1.509    1.509 {built-in method builtins.exec}
   134/87    0.038    0.000    0.106    0.001 {built-in method _imp.create_dynamic}
    21906    0.034    0.000    0.068    0.000 utils.py:143(slugify)
59936/53901    0.032    0.000    0.041    0.000 {built-in method builtins.getattr}
     2462    0.032    0.000    0.032    0.000 {method '_trait' of 'traits.ctraits.CHasTraits' objects}
      269    0.024    0.000    0.088    0.000 has_traits.py:472(__init__)
     6855    0.022    0.000    0.022    0.000 {built-in method posix.stat}
    37193    0.021    0.000    0.031    0.000 {built-in method builtins.hasattr}
148842/148816    0.021    0.000    0.021    0.000 {built-in method builtins.isinstance}
     2677    0.020    0.000    0.082    0.000 <frozen importlib._bootstrap_external>:1233(find_spec)
[...]

PR #2453, opened 2020-07-27
Fix duplicate and unused imports

tjof2:
### Description of the change
Follows on from #2451, removing unused and/or duplicate imports from `hyperspy/*`. These are found with the command: `flake8 hyperspy/ | egrep "F401|F811"`

Crucially, it also updates the `flake8` rules in `setup.cfg` to ignore any F401 warnings raised in files where the imports _appear_ unused, but are actually a part of the API, so that anyone who runs a similar command in future isn't caught out. There may well be unused imports in the ignored files, particularly in files like `api_nogui.py` (cf. #1480), but they ought to be handled more carefully.

### Progress of the PR
- [x] add tests,
- [x] ready for review.

### DIFF
--- a/hyperspy/drawing/_widgets/label.py
+++ b/hyperspy/drawing/_widgets/label.py
@@ -20,7 +20,6 @@ import numpy as np
 import matplotlib.transforms as transforms
 
 from hyperspy.drawing.widgets import Widget1DBase
-from hyperspy.drawing.utils import picker_kwargs
 
 
 class LabelWidget(Widget1DBase):
--- a/hyperspy/drawing/image.py
+++ b/hyperspy/drawing/image.py
@@ -27,7 +27,6 @@ from traits.api import Undefined
 import logging
 import inspect
 import copy
-from distutils.version import LooseVersion
 
 from hyperspy.drawing import widgets
 from hyperspy.drawing import utils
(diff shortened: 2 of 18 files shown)


tjof2:
Not sure why the Azure pipelines tests are failing - @ericpre?
```
ImportError while loading conftest '/home/vsts/work/1/s/hyperspy/conftest.py'.
hyperspy/conftest.py:20: in <module>
    from traits.etsconfig.api import ETSConfig
E   ModuleNotFoundError: No module named 'traits'
```

tjof2:
I don't think codecov is important here - the change is miniscule:
```
Name                                                       Stmts   Miss Branch BrPart     Cover
-----------------------------------------------------------------------------------------------
TOTAL before                                               28127   5674  12241   1720    76.78%
TOTAL after                                                28106   5673  12241   1720    76.77%
```
and the change in the  "Miss" column can be entirely attributed to `hyperspy/drawing/tiles.py`, which isn't covered by the tests in either case:
```
Name                                                                 Stmts   Miss Branch BrPart     Cover

---------------------------------------------------------------------------------------------------------
< hyperspy/drawing/tiles.py                                               50     50     12      0     0.00%
---
> hyperspy/drawing/tiles.py                                               49     49     12      0     0.00%
```

PR #2456, opened 2020-07-30
Better support for histogram bin estimation

### LINKED ISSUE #784: Memory leak
Just noticed, when running the following code

``` python
In [2]: ax1 = np.arange(-30, 100, 0.01)
In [3]: ax2 = np.arange(-40, 90, 0.01)
In [4]: s1 = hs.signals.EELSSpectrum(np.exp(-np.abs(ax1)))
In [5]: s2 = hs.signals.EELSSpectrum(np.exp(-np.abs(ax2)))
In [6]: s1 = hs.stack([s1]*5)
In [7]: s2 = hs.stack([s2]*5)
In [8]: s1.align_zero_loss_peak(also_align=[s2])
In [9]: s1.spikes_removal_tool()
```

and the clicking the button to plot histogram, there is a memory leak that nearly crashed my system.
Can anyone replicated that, or did I mess something up?

tjof2:
### Description of the change
- Increment required numpy version from `>=1.10.0` to `>=1.11.0` to support the use of [`np.histogram_bin_edges()`](https://numpy.org/doc/stable/reference/generated/numpy.histogram_bin_edges.html)
- Use `np.histogram_bin_edges()` where possible, allowing the removal of some of the astroML vendored code.
- Move the `hyperspy/external/astroML` code to `hyperspy/external/astropy` since the code can now be found in `astropy` instead.
- Improve tests for histogram bin estimation.
- Improved docstrings.
- Deprecated `bins="scotts"/"freedman"` in favour of `"scott"/"fd"` to be consistent with `numpy`.
- Closes #784 since the maximum number of bins defaults to 250, so the `MemoryError` can't happen by default. Added tests to verify this behaviour.

### Progress of the PR
- [x] Changes implemented,
- [x] update docstring,
- [x] update user guide,
- [x] add tests,
- [x] ready for review.

### Example usage
From the example in #784
```python
>>> import numpy as np
>>> import hyperspy.api as hs

>>> ax1 = np.exp(-np.abs(np.arange(-30, 100, 0.02)))
>>> ax2 = np.exp(-np.abs(np.arange(-40, 90, 0.02)))
>>> s1 = hs.signals.EELSSpectrum(ax1)
>>> s2 = hs.signals.EELSSpectrum(ax2)
>>> s1 = hs.stack([s1] * 5)
>>> s2 = hs.stack([s2] * 5)
>>> s1.align_zero_loss_peak(also_align=[s2])
>>> s1.get_histogram()
# WARNING:hyperspy.misc.hist_tools:Estimated number of bins using `bins='fd'` is too large (182966505). 
# Capping the number of bins at `max_num_bins=250`.
# Consider using an alternative method for calculating the bins such as `bins='scott'`, 
# or increasing the value of the `max_num_bins` keyword argument.
>>> s1.get_histogram(bins="scott", max_num_bins=2)
# WARNING:hyperspy.misc.hist_tools:Estimated number of bins using `bins='scott'` is too large (107). 
# Capping the number of bins at `max_num_bins=2`. 
# Consider using an alternative method for calculating the bins such as `bins='scott'`, 
# or increasing the value of the `max_num_bins` keyword argument.
```


ericpre:
I am wondering if some the dask histogram part (bin calculation) should also go in `dask.array` as a `histogram_bin_edges` function? See https://github.com/dask/dask/issues/2911.

francisco-dlp:
LGTM, though it would be nice to hit the missed line with a extra test.

Also, for the future it would be good to think on completing the dask support for the methods in numpy and contributing them to to dask.

tjof2:
> LGTM, though it would be nice to hit the missed line with a extra test.

@francisco-dlp I'll wait for the latest codecov report to finish and add the test if still required. Will fix the broken test images too.

> I am wondering if some the dask histogram part (bin calculation) s

Indeed, probably a good idea for future as noted already.

hyperspy/misc/array_tools.py
```diff
@@ -391,22 +391,12 @@ def calculate_bins_histogram(data, max_num_bins=250):
...
-    else:
-        # limited variance: fd_bin_width may be zero
-        bin_width = sturges_bin_width
+    # "auto" returns max('sturges', 'fd') which is what we want
+    n_bins = len(np.histogram_bin_edges(data, bins="auto"))
 
     # Cap at max_num_bins to avoid memory errors when
     # the number of bins is very large
-    return min(int(np.ceil(data.ptp() / bin_width)), max_num_bins)
+    return min(n_bins, max_num_bins)
```
francisco-dlp: Now that this function is just wrapping the numpy function, why not simply removing it?

tjof2: The wrapper is necessary for the same reason as #784 - the Freedman-Diaconis rule will potentially give huge numbers of bins, hence the cap.

tjof2: I'm not completely familiar with the reason why HyperSpy needs a separate `numba_histogram()` which is what this function actually feeds into - but to minimize the risk of breaking things, it seemed right to take this approach. 

The bulk of the changes in this PR actually affect `s.get_histogram()`, but since there is this separate function I added the same cap in.

francisco-dlp: Makes sense, but since the function is just used at one place, why not just copying the small code there?

tjof2: Ah OK yes I see what you mean! Will do that and verify the rest of the tests pass.

tjof2:
Removed the function as suggested @francisco-dlp, and fixed the plotting tests. This should be good to go. I think the codecov message about array_tools.py can be ignored - all that's happened is a function that was previously tested has been removed.

tjof2:
~I'll fix the merge conflict shortly @francisco-dlp~

Fixed - this is ready to go.

PR #2459, opened 2020-08-02
Generalised AxesManager iterpath 

thomasaarholt:
### Description of the change
I've generalised the `iterpath` property of the AxesManager so that instead of just `"serpentine"` or `"flyback"`, one can now give any iterable or iterator that generates indices of the correct size. This includes _generators_.

This affects the behaviour of `for index in s.axes_manager`, including `m.multifit(iterpath=iterpath)` so that one can specify the order of iteration or which specific positions should be iterated through.

- I've changed from the private property `axes_manager._iterpath` to the `@property` `axes_manager.iterpath`. I'd argue that this doesn't change the API to any significant degree (the multifit argument is unchanged). The new property approach includes checks to ensure that the iterpath is correctly entered.

- I've tidied up the logic around the `__next__` and `__iter__` methods, and removed the `_index` attribute since it isn't used anywhere and can't be used for anything useful.

- Note that since generators and iterators don't have a `len`, `m.multifit` gives a progressbar that is too long. I don't see any good way around that.

- ~~Slightly annoyingly, since I'm relying on np.ndindex and the serpentine generator (which is too complicated for me to reverse engineer), the custom iterpath must follow numpy indexing.~~ ~~This is the reason for the [::-1] in the `AxesManager.__next__` code. Hence, any custom iterpath must also give indices in reverse order (`(1,0,0)` becomes `(0,0,1)`).~~

- I've fixed the above by creating a custom modified `np.ndindex` to reverse its indices.

### Progress of the PR
- [x] update user guide pending merge of #2549
- [x] add tests,
- [ ] ready for review.

```python
import hyperspy.api as hs
import numpy as np

s = hs.signals.Signal1D(np.ones((3, 3, 10)))
ax = s.axes_manager
m = s.create_model()
G = hs.model.components1D.Gaussian()
m.append(G)

# Example 1
indices = np.array([(0,0), (0,1), (2,2)])
s.axes_manager.iterpath = indices
for i in s.axes_manager:
    print(i)

# Example 2
m.multifit(iterpath=indices)

# Example 3 (raises error)
s.axes_manager.iterpath = "test"
```


thomasaarholt:
This PR is ready, but the final commits (docs in the new `axes.rst`) lie in my branch based off `non-uniform axes` [here](https://github.com/thomasaarholt/hyperspy/tree/general_iterpath_non_uniform). I'm keeping these two separate until #2549 is merged in case there are significant changes, so that it's easy to see (if anything) has changed.

hyperspy/axes.py
```diff
@@ -649,6 +654,22 @@ def serpentine_iter(shape):
...
+    "Classic flyback scan pattern generator which yields indices in similar fashion to np.ndindex. Takes shape in hyperspy order, not numpy order."
+    shape = shape[::-1]
+    class ndindex_reversed(np.ndindex):
+        def __next__(self):
+            next(self._it)
+            return self._it.multi_index[::-1]
+
+    return ndindex_reversed(shape)
+
+iterpath_error = '''The iterpath scan pattern is set to "{}". \
```
ericpre: Move this to where it is used in the namespace tidy?

hyperspy/_signals/signal1d.py
```diff
@@ -275,6 +275,7 @@ def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
         if self.axes_manager.signal_dimension != 1:
             self.axes_manager.set_signal_dimension(1)
+        self.axes_manager.iterpath = self.axes_manager._iterpath
```
ericpre: Why is it necessary?

ericpre (changes requested):
The docstring of the `AxesManager` needs to be updated.

One suggestion I am not sure: now that `iterpath` is public, maybe the attribute name should be changed to avoid using an abbreviation and be more explicit. If so, may be worth considering changing the corresponding argument of `multifit`.

hyperspy/_signals/signal1d.py
thomasaarholt: Fixed.
The [`test_signal_iterator`](https://github.com/hyperspy/hyperspy/blob/e4c6994b61837a2c54c963d2b1afe3b86c1bf7c4/hyperspy/tests/signal/test_tools.py#L46) test was originally failing without those lines, but at a closer look I believe I can safely modify the test so it is no longer failing. It was calling `next(s).data[0]`, without calling the `__iter__()` method first, which wasn't initialising the `iterpath_generator`.

thomasaarholt:
Thanks for the comments @ericpre, the lines you commented about were annoying me, and I learnt a bit more about `__iter__` and `__next__` at the same time as fixing them.

I'd be happy to change the name - I agree that being clear is good here - although I think I can only suggest `iteration_path`. How does that sound?

Edit: Forgot to change AxesManager docstring. Doing so now.

hyperspy/axes.py
```diff
@@ -961,57 +977,56 @@ def _update_max_index(self):
...
+                raise ValueError(iterpath_error)
+        else:
+            # Pass a custom indices iterator
+            self._iterpath = path
+            try:
+                self._iterpath_generator = iter(self._iterpath)
+            except TypeError:
+                raise TypeError(
+                f"The iterpath '{self._iterpath}' is not a correct iterpath. "
+                "Ensure it is an iterable delivering incides with length equal "
```
ericpre: The error message doesn't match what it is checked and it would be good to check it.

hyperspy/axes.py
```diff
@@ -649,6 +654,16 @@ def serpentine_iter(shape):
         else:  # pragma: no cover
             break
 
+def flyback_iter(shape):
```
ericpre: Make this function private?

hyperspy/axes.py
```diff
@@ -773,12 +792,9 @@ def __init__(self, axes_list):
             # Default to Signal1D view if the view is not fully defined
             self.set_signal_dimension(len(axes_list))
 
+        self._iterpath = 'flyback'
```
ericpre: Should it be `self.iterpath = 'flyback'`?

hyperspy/axes.py
```diff
@@ -961,57 +977,56 @@ def _update_max_index(self):
...
+            else:
+                iterpath_error = (
+                    f'The iterpath scan pattern is set to "{path}". '
+                    'It must be either "serpentine" or "flyback", or an iterable '
+                    'of hyperspy indices, and is set either as multifit '
+                    '`iterpath` argument or `axes_manager.iterpath`')
+                raise ValueError(iterpath_error)
+        else:
+            # Pass a custom indices iterator
+            self._iterpath = path
```
ericpre: `self._iterpath` should not be set at this stage in case there is an exception below.

hyperspy/axes.py
```diff
@@ -679,6 +694,10 @@ class AxesManager(t.HasTraits):
         AttributeError when attempting to set its value.
     signal_axes, navigation_axes : list
         Contain the corresponding DataAxis objects
+    iterpath : string or iterable
+        Sets the order of iterating through the indices in the navigation dimension.
+        indices. Can be either "flyback" or "serpentine", or an iterable
+        of hyperspy indices.
```
ericpre: `hyperspy indices` -> `navigation indices`?

ericpre (changes requested):
> - Note that since generators and iterators don't have a `len`, `m.multifit` gives a progressbar that is too long. I don't see any good way around that.

When using an custom iterpath, such as an iterable, the length can be known from the length of `_iterpath`.

hyperspy/axes.py
thomasaarholt: Thanks, fixed.

hyperspy/axes.py
thomasaarholt: If I move it to the end of the AxesManager `__init__`, then yes :) Done. (Otherwise the `navigation_shape` which it uses, hasn't been created yet.)

hyperspy/axes.py
thomasaarholt: Fixed.

thomasaarholt:
I've fixed your comments. On the topic of changing the name of `iterpath`: @francisco-dlp you requested this feature in the first place 😄 any thoughts on whether we should use a more friendly name? Suggestions? (others welcome too)

hyperspy/axes.py
thomasaarholt: In the case of the `self._iterpath` being a generator, I don't think we can check if the length of the indices yielded are the correct size without using up the first generated value. I've added checks for the case where it is not a generator (but an array, list of tuples, etc).

thomasaarholt:
> 
> 
> > * Note that since generators and iterators don't have a `len`, `m.multifit` gives a progressbar that is too long. I don't see any good way around that.
> 
> When using an custom iterpath, such as an iterable, the length can be known from the length of `_iterpath`.

In the case of the `self._iterpath` being a generator, I don't think we can check this. We could offer the user the option to add their own length? I just realised, I have to check if the too-long progressbar (this isn't really a problem) can also be too-short (this is a problem), and possibly stop fitting if the iterpath is longer than the number of navigation size.

ericpre:
> > > * Note that since generators and iterators don't have a `len`, `m.multifit` gives a progressbar that is too long. I don't see any good way around that.
> > 
> > 
> > When using an custom iterpath, such as an iterable, the length can be known from the length of `_iterpath`.
> 
> In the case of the `self._iterpath` being a generator, I don't think we can check this. We could offer the user the option to add their own length? I just realised, I have to check if the too-long progressbar (this isn't really a problem) can also be too-short (this is a problem), and possibly stop fitting if the iterpath is longer than the number of navigation size.

Since not all iterables are generators and some of them have a length, it would be good using it when possible, for example in the case of:
https://github.com/thomasaarholt/hyperspy/blob/6d496e5bca2126adae57088f96c3723d78c01fa7/hyperspy/tests/axes/test_axes_manager.py#L405-L408

jlaehne:
> This PR is ready, but the final commits (docs in the new `axes.rst`) lie in my branch based off `non-uniform axes` [here](https://github.com/thomasaarholt/hyperspy/tree/general_iterpath_non_uniform). I'm keeping these two separate until #2549 is merged in case there are significant changes, so that it's easy to see (if anything) has changed.

The relevant changes to `axes.rst` should now be merged into `non_uniform_axes` through #2304 (#2549 was just based off this PR to include some changes made there -- I now rebased the branch to make that clear)

thomasaarholt:
I've rebased onto non_uniform_axes!

Once tests have finished running, I'd appreciate in particular if someone could take a look through the documentation I've added and see if it makes sense. The failing TestSpikesRemovalToolZLP::test_spikes_diagnosis is failing upstream too.

Reminder to self: There is a bit of text in the non_uniform_axes docs that exists both in the new axes.rst and getting_started.rst. I'll see if I can whip up a PR tomorrow evening.

hyperspy/axes.py
```diff
@@ -1427,57 +1445,103 @@ def _update_max_index(self):
...
+                    )
+            except TypeError:
+                # progressbar is shown, so user can monitor "iterations per second"
+                # but the length of the bar is unknown
+                maxval = None
+                warnings.warn(
+                    ("The AxesManager `iterpath` is missing the `__len__` method, so does not have a known length. "
+                    "The progressbar will only show run time and iterations per second, but no actual progress indicator."),
+                    UserWarning
+                    )
```
ericpre: If this is intended behaviour, this shouldn't a warning and adding it as info in the logger should be enough?

ericpre (commented):
I have one last comment, other than that, I think it should be good?

hyperspy/axes.py
thomasaarholt: Yes, I agree with that.

thomasaarholt:
Thanks @ericpre. Pushed that change now.

ericpre:
@thomasaarholt, is it finished from your side?

thomasaarholt:
Yes, this can be merged.

PR #2462, opened 2020-08-03
histogram bin estimation (merge Rnp into Rnm)

ericpre:
Making this PR to merge Rnp into Rnm, because there are an unexpected issue with the bin estimation used for histograms.

Also add one more test and the numba cache file to .gitignore.


hyperspy/signal_tools.py
```diff
@@ -710,9 +739,11 @@ def plot_histogram(self, max_num_bins=250):
         if self._vmin == self._vmax:
             return
         data = self._get_data()
-
-        # "auto" returns max('sturges', 'fd') which is what we want
-        n_bins = len(np.histogram_bin_edges(data, bins="auto"))
+        try:
+            # "auto" returns max('sturges', 'fd') which is what we want
+            n_bins = len(np.histogram_bin_edges(data, bins='auto'))
```
ericpre: This is now using `np.histogram_bin_edges` which cause a memory error with the following example:
```python
s = datasets.example_signals.object_hologram()
fft = s.fft(True)
fft.plot(True, vmin=None, vmax=None)
ceditor = ImageContrastEditor(fft._plot.signal_plot)
```

tjof2: Makes sense!

ericpre: I don't know this should be considered as a numpy bug since in this case numpy can't estimate the number of bins using the "fd". The old `calculate_bins_histogram` was able to calculate it:
https://github.com/hyperspy/hyperspy/blob/aad219d25386221ba405f634d8dc48bc525841d2/hyperspy/misc/array_tools.py#L375-L405

tjof2: The old function only calculated the bin width, whereas the numpy function calculates the width and then the bins themselves (which is presumably where the error arises).

Perhaps the old function could be restored, I don't have a particular view either way.

francisco-dlp: If that's the case, it seems wasteful to calculate the bins twice...

tjof2: Well it also depends how much of a speed-up the Numba histogram function actually gives in usual applications vs. Numpy histogram, because you could just use the numpy histogram function on its own. (I've read the blog post it's taken from)

Restoring the function is fine, though the addition of a kwarg for capping n_bins instead of a hard-coded 250 is preferable.

tjof2: > The old calculate_bins_histogram was able to calculate it

The old function will have also returned a massive number for the `fd` rule, it's only because it's capped _before_ the bins are calculated that it works, rather than afterwards as here

ericpre: I remove the use of `np.histogram_bin_edges` to avoid calculating twice.

hyperspy/signal_tools.py
```diff
@@ -710,12 +739,22 @@ def plot_histogram(self, max_num_bins=250):
...
+        iqr = np.subtract(*np.percentile(data, [75, 25]))
+        fd_bin_width = 2.0 * iqr * data.size ** (-1.0 / 3.0)
+
+        if fd_bin_width > 0:
+            bin_width = min(fd_bin_width, sturges_bin_width)
+        else:
+            # limited variance: fd_bin_width may be zero
+            bin_width = sturges_bin_width
+
+        self.bins = min(int(np.ceil(data.ptp() / bin_width)), 250)
```
tjof2: Replace `250` with `max_num_bins`

tjof2 (changes requested):

hyperspy/signal_tools.py
ericpre: Oh yes, I was too quick with copy and paste!

tjof2 (approved):

Issue #1438, opened 2017-02-09
Minimise requirements


francisco-dlp:
Currently HyperSpy depends on many libraries. However, it should be possible to reduce the list to just a few of them, with the rest being optional libraries to be declared in ``extra_require`` to unlock some features. It follows a list of the requirements that I think that we could make optional. Notice that #1437 makes it easy to install all dependencies ``pip install hyperspy[all]``:

* [ ] matplotlib
* [ ] h5py
* [ ] scikit-image
* [ ] ipython
* [ ] ipyparallel
* [ ] traitsui

What do you think?

magnunor:
How would the optional libraries be handled in the code? Do imports in a `try` statement?

francisco-dlp:
Something like that should do.

tjof2:
@francisco-dlp @ericpre perhaps it's worth looking again at tidying up at least the required versions either for or after 1.6?

For example, running `pipgrip --tree .` on Rnm highlights some apparent inconsistencies. 

- `numba` is not an optional dependency since `sparse` has relied on it since at least 0.4.0. Therefore the `jit_ifnumba` decorator can be removed in place of `from numba import jit` etc.
- `scikit-image>=0.15` requires `numpy>=1.15`, which is more recent than listed (`>=1.10`)
- `scikit-image>=0.15` requires `tifffile>=2019.7.26`, which is more recent than listed (`>=2018.10.18`)
- `scikit-image>=0.15` requires `scipy>=1.0.1`, which is more recent than listed (`>=0.17`)
- `tifffile[all]>=2018.10.18` requires `numpy>=1.15.1`, which is more recent than listed (`>=1.10`)
- `tifffile[all]>=2018.10.18` requires `matplotlib>=3.1`, which is more recent than listed (`>=2.2.3`)

Output of `pipgrip --tree .`:
```
. (1.6.dev0)
├── dask[array]!=2.0,>=0.18 (2.21.0)
│   ├── numpy>=1.13.0 (1.19.1)
│   ├── pyyaml (5.3.1)
│   └── toolz>=0.8.2 (0.10.0)
├── dill (0.3.2)
├── h5py>=2.3 (2.10.0)
│   ├── numpy>=1.7 (1.19.1)
│   └── six (1.15.0)
├── imageio (2.9.0)
│   ├── numpy (1.19.1)
│   └── pillow (7.2.0)
├── ipyparallel (6.3.0)
│   ├── decorator (4.4.2)
[...]

PR #2464, opened 2020-08-03
Release v1.6

francisco-dlp:
Update CHANGES for the release.

PRs with fixes and additions are more than welcome.

I'll direct the conda forge recipe here.


setup.py
```diff
@@ -45,9 +45,9 @@
 setup_path = os.path.dirname(__file__)
 
 
-install_req = ['scipy>=0.17',
+install_req = ['scipy>=1.4.0',
```
tjof2: In which case, there is code in `model.py` and docs in `model.rst` which refer to LooseVersion of scipy >=1.2.0 for `dual_annealing` and `shgo` as options for `model.fit()`. This logic can be tidied up.

francisco-dlp: Sounds good, would you like to take care of that in a PR to the main repository?

tjof2: Yep I'll do it later this morning

francisco-dlp: Great, thanks!

ericpre: scipy 1.4 has been released in December 2019, which I think is a bit too recent. Scipy is not straightforward to package/distribute, so I would suggest to avoid increasing this requirement. As for numpy, in my opinion, it would be sensible to be conservative and as a guideline to supports scipy version released in the 2 years for example (see https://numpy.org/neps/nep-0029-deprecation_policy.html)

francisco-dlp: The issue is that recent changes in ``decomposition`` require scipy v1.4.

ericpre: Which changes exactly? There have been a few PRs releted to decomposition, recently!

francisco-dlp: The [arpack test](https://github.com/hyperspy/hyperspy/blob/81c226ac4fc6a32e09022ce0ce2b29bb90308269/hyperspy/tests/learn/test_svd_pca.py#L82) fails for scipy < 1.4

tjof2: Note that 1.4 also brings `scipy.fft` over `fftpack`. You can do a `pytest.importorskip("scipy", minversion="1.4")` for that test.

tjof2: It can definitely go to scipy 1.0.1 because that's what scikit-image requires (cf. #1438).

I won't make a separate PR in this case given @ericpre's comment.

francisco-dlp: The thing is that, AFAIK, hyperspy is not available in any Linux distribution other than the ones with a rolling release policy such as Arch and [OpenSuse Tumbleweed](https://build.opensuse.org/package/show/home:andythe_great/python-hyperspy), for which requiring a recent scipy is not an issue. That's certainly because our dependencies, even when being conservative regarding scipy and numpy as currently, cannot be fulfilled in other distributions. Hopefully the situation will change with "The Split", but, until then, maybe it is not too bad to depend on a recent scipy?

ericpre: It seems that `svd_pca` will fail with `decomposition_output=None` and `svd_solver='arpack'.` but when using `decomposition`, the `output_dimension` is set (not None), so it should be fine to skip the failing test for this specific case?

CHANGES.rst
```diff
@@ -84,6 +84,33 @@ Enhancements
...
+    * "fmin_l_bfgs_b" -> "L-BFGS-B",
+    * "fmin_tnc" -> "TNC",
+    * "fmin_powell" -> "Powell",
+    * "mpfit" -> "lm" (in combination with `"bounded=True"`),
+    * "leastsq" -> "lm",
+    * Passing integer arguments to ``parallel`` to select the number of
+      workers is now deprecated. Use ``parallel=True, max_workers={value}``
+      instead.
+* The ``method`` keyword has been renamed to ``loss_function``.
+* The ``loss_funtion`` value ``"ml"`` has been renamed to ``ml-poisson``.
```
tjof2: Typo in loss function

setup.py
tjof2: No, the test for None is actually necessary, see here: https://github.com/hyperspy/hyperspy/blob/3ec3018f59cee4c9db023408b09d51024fe53d42/hyperspy/learn/svd_pca.py#L128-L131

CHANGES.rst
francisco-dlp: Fixed, thanks. Could you have a look at it again? I should be fairly complete now.

CHANGES.rst
```diff
@@ -1,11 +1,176 @@
...
+  is deprecated and will be removed.
+* Several syntax changes in the :py:meth:`~.learn.mva.decomposition` method:
+
+  * Several ``algorithm`` keyword values have been renamed as follows:
+
+    * ``"fast_svd"``: ``"svd"``,
+    * ``"fast_mlpca"``: ``"mlpca"``,
+    * ``"RPCA_GoDec"``: ``"rpca"``,
+    * ``"ORPCA"``: ``"orpca"``,
+    * ``"ORNMF"``: ``"ornmf"``,
```
tjof2: these need to be uppercase after #2467

tjof2 (approved):

CHANGES.rst
tjof2: Have done so

francisco-dlp: Thanks!

What do you think about renaming:

* "ml-poisson" -> "ML-Poisson" or "ML-poisson"?
* "lm" -> "LM"

to make it clearer that they're acronyms?

tjof2: - I think keeping `"lm"` lowercase is important to match `scipy.optimize.least_squares`.
- Happy with `"ML-poisson"`

francisco-dlp:
Test release at https://test.pypi.org/project/hyperspy/1.6.0/

francisco-dlp:
@ericpre, I think that this is v1.6, do you agree?

ericpre:
I am finishing checking that everything is fine with the GUIs - I am not 100% sure if a few more tweaks are necessary in hyperspy.

ericpre:
#2470

setup.py
ericpre: I still think that scipy 1.4 is too recent. I will make a PR to raise an error when using arpack svd solver and scipy < 1.4.

francisco-dlp: OK, we can go down to v1.1 that is 
already in Debian.

ericpre: #2471

francisco-dlp: OK, I'll set scipy==1.0.1 as suggested by @tjof2

ericpre (approved):

ericpre:
I have just realised that the wheels are not being build for python 3.7 on win and mac... We could do this on azure pipeline with these build running "on tag".

francisco-dlp:
Yes, I've just realized that myself too. However, I think that we can get them from azure's ariefacts isn't it?

ericpre:
yes but with the current setup, we are missing two wheels (python 3.7 for mac and windows):
https://dev.azure.com/franciscode-la-pena-manchon/hyperspy/_build/results?buildId=536&view=artifacts&type=publishedArtifacts

I will see if this is possible to run a pipeline on azure for the two missing wheel and without making a new tag...

francisco-dlp:
Hopefully the tag won't matter as long as  we  don't change the version in the package. So, it should be enough to add the extra builds to azure without changing anything just to get the file from there. I may be wrong of course.

Could you upload the wheels to pypi? I should help packaging other sort stuff now... The widgets are now released in pypi.

ericpre:
Yes, I will take the commit from which you made the tag to run the to run the pipeline and then I will upload the wheels on pypi!

francisco-dlp:
You could take the following one in which I add update the DOI, that's the one that I've used for the sdist.

ericpre:
ah yes and this is the one which is on pypi, so it is better to take that one!

PR #2465, opened 2020-08-04
Background removal: Spelling of component names and CLI aliases, documentation

jlaehne:
This PR is based on #2184 and resolves the original issue #2176.

The solution in #2176 differed/went beyond #2307 in three aspects:
1. 'Power law' would be the correct spelling in the GUI not 'Power Law'
2. The naming should be consistent also for the new components SkewNormal and SplitVoigt (here of course 'Split Voigt' and not 'Split voigt')
3. This implementation accepted the three aliases 'Power law', 'Power Law' and 'PowerLaw' in the CLI call of the `remove_background` function (same for the other two components).

With #2307 this is much cleaner than the original.

Also the tests and documentation have been updated to include missing components that have in the meantime been added to the background_removal method.

### Progress of the PR
- [X] add tests,
- [X] ready for review.

### DIFF
--- a/doc/user_guide/signal1d.rst
+++ b/doc/user_guide/signal1d.rst
@@ -56,11 +56,14 @@ Background removal
 The :py:meth:`~._signals.signal1d.Signal1D.remove_background` method provides
 background removal capabilities through both a CLI and a GUI. The GUI displays
 an interactive preview of the remainder after background subtraction. Currently,
-the following background types are supported: power law, offset, polynomial, 
-Gaussian, Lorentzian and skew normal. By default, the background parameters are
-estimated using analytical approximations (keyword argument ``fast=True``). For 
-better accuracy, but higher processing time, the parameters can be estimated 
-using curve fitting by setting ``fast=False``.
+the following background types are supported: Doniach, Exponential, Gaussian,
+Lorentzian, Polynomial, Power law (default), Offset, Skew normal, Split Voigt 
+and Voigt. By default, the background parameters are estimated using analytical
+approximations (keyword argument ``fast=True``). The fast option is not accurate
+for most background types - except Gaussian, Lorentzian, Offset and Power law -
+but it is useful to estimate the initial fitting parameters before performing a
+full fit. For better accuracy, but higher processing time, the parameters can
+be estimated using curve fitting by setting ``fast=False``.
 
 Example of usage:
 
--- a/doc/user_guide/signal1d.rst
+++ b/doc/user_guide/signal1d.rst
@@ -60,7 +60,7 @@ the following background types are supported: Doniach, Exponential, Gaussian,
 Lorentzian, Polynomial, Power law (default), Offset, Skew normal, Split Voigt 
 and Voigt. By default, the background parameters are estimated using analytical
 approximations (keyword argument ``fast=True``). The fast option is not accurate
-for most background types - except Gaussian, Lorentzian, Offset and Power law -
+for most background types - except Gaussian, Offset and Power law -
 but it is useful to estimate the initial fitting parameters before performing a
 full fit. For better accuracy, but higher processing time, the parameters can
 be estimated using curve fitting by setting ``fast=False``.
--- a/hyperspy/signal_tools.py
+++ b/hyperspy/signal_tools.py
@@ -1007,12 +1007,12 @@ class IntegrateArea(SpanSelectorInSignal1D):
 class BackgroundRemoval(SpanSelectorInSignal1D):
     background_type = t.Enum(
         'Doniach',
+        'Exponential',
         'Gaussian',
         'Lorentzian',
         'Offset',
         'Polynomial',
         'Power law',
-        'Exponential',
         'Skew normal',
         'Split Voigt',
         'Voigt',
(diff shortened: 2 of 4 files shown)


tjof2 (approved):

doc/user_guide/signal1d.rst
```diff
@@ -56,11 +56,14 @@ Background removal
...
-the following background types are supported: power law, offset, polynomial, 
-Gaussian, Lorentzian and skew normal. By default, the background parameters are
-estimated using analytical approximations (keyword argument ``fast=True``). For 
-better accuracy, but higher processing time, the parameters can be estimated 
-using curve fitting by setting ``fast=False``.
+the following background types are supported: Doniach, Exponential, Gaussian,
+Lorentzian, Polynomial, Power law (default), Offset, Skew normal, Split Voigt 
+and Voigt. By default, the background parameters are estimated using analytical
+approximations (keyword argument ``fast=True``). The fast option is not accurate
+for most background types - except Gaussian, Lorentzian, Offset and Power law -
```
ericpre: From what I remember, Lorentzian wasn't bad but not that accurate neither  Should it be removed from the exception list?

hyperspy/signal_tools.py
```diff
@@ -1052,6 +1052,12 @@ def __init__(self, signal, background_type='Power Law', polynomial_order=2,
             model = Model1D(signal)
         self.model = model
         self.polynomial_order = polynomial_order
+        if background_type in ['Power Law', 'PowerLaw']:
+            background_type = 'Power law'
+        if background_type in ['Skew Normal', 'SkewNormal']:
+            background_type = 'Skew normal'
+        if background_type in ['Split voigt', 'SplitVoigt']:
+            background_type = 'Split Voigt'
```
ericpre, line 1060: This should not be necessary anymore? Because of:
https://github.com/hyperspy/hyperspy/blob/3ec3018f59cee4c9db023408b09d51024fe53d42/hyperspy/signal_tools.py#L1240

hyperspy/tests/signal/test_remove_background.py
```diff
@@ -266,8 +266,10 @@ def compare_axes_manager_metadata(s0, s1):
 @pytest.mark.parametrize('show_progressbar', [True, False])
 @pytest.mark.parametrize('plot_remainder', [True, False])
 @pytest.mark.parametrize('background_type',
-                         ['Doniach', 'Gaussian', 'Lorentzian', 'Polynomial',
-                          'Power Law', 'Offset', 'SkewNormal', 'SplitVoigt',
+                         ['Doniach', 'Exponential', 'Gaussian', 'Lorentzian', 
+                          'Polynomial', 'Power law', 'Power Law', 'PowerLaw', 
+                          'Offset', 'Skew normal', 'Skew Normal', 'SkewNormal', 
+                          'Split Voigt', 'Split voigt', 'SplitVoigt',
```
ericpre: The number of combination for this test is getting unnecessarily large. Maybe the `background_type` parameter should be move to another test? The test suite is getting fairly slow, it is worth making sure that we don't add more test which are inefficient.

ericpre (commented):
Thanks @jlaehne for following up on this! I made some comments.

doc/user_guide/signal1d.rst
jlaehne: I'll test it again tonight.

hyperspy/signal_tools.py
jlaehne: I get an error message that it is not one of the known components when running e.g. `s.remove_background(background_type="PowerLaw")`:

```
TraitError: The 'background_type' trait of a BackgroundRemoval instance must be 'Doniach' or 'Gaussian' or 'Lorentzian' or 'Offset' or 'Polynomial' or 'Power Law' or 'Exponential' or 'SkewNormal' or 'SplitVoigt' or 'Voigt', but a value of 'PowerLaw' <class 'str'> was specified.
```

hyperspy/tests/signal/test_remove_background.py
jlaehne: I can move these variations to the specific tests for Power law, Skew normal and Split Voigt or create a dedicated tests only that these variations are accepted.

hyperspy/signal_tools.py
ericpre: Yes, my bad!

ericpre:
I made a PR https://github.com/jlaehne/hyperspy/pull/3 to improve the tests situation. @jlaehne, please have a look and merge it if you are happy with the changes!

doc/user_guide/signal1d.rst
jlaehne: I removed Lorentzian not to hold this up, but I will test again later.

ericpre:
Thanks @jlaehne!

ericpre (approved):

jlaehne:
After suppressing the call of the gui through an explicit specification of signal_range, the new test now runs through (I don't see the relation of the current partial fails to the introduced changes).

tjof2:
> I don't see the relation of the current partial fails to the introduced changes

The test fail on MacOS does indeed look unrelated.

The coverage decrease is due to some of the aliases here: https://github.com/hyperspy/hyperspy/blob/60ee7bcb67a8a34cd49ae552d8884e3515afb54d/hyperspy/signal_tools.py#L1055-L1060

This could be replaced by a dictionary mapping of aliases to avoid the if/else branches, then you wouldn't need to have so many combinations of tests to get the same coverage as mentioned here: https://github.com/hyperspy/hyperspy/pull/2465#discussion_r464942670

PR #2467, opened 2020-08-04
Uppercase acronyms

francisco-dlp:
Use uppercase for ``algorithm`` when the name of the algorithm is an acronym *and* (the algorithm is not likely to be removed in the near future *or* the algorithm name changed to lowercase in #2383).

### Progress of the PR
- [x] add tests,
- [x] ready for review.


tjof2 (approved):

tjof2:
Why not do `algorithm = algorithm.upper()` before subsequent comparison?

francisco-dlp:
I don't get what you mean, would you like to accept both lower and upper?

tjof2:
> I don't get what you mean, would you like to accept both lower and upper?

Is it a _breaking_ API change? i.e. could someone have old code that uses the lowercase names? Accepting lower and upper would mean their code still works?

Otherwise, ignore me!

francisco-dlp:
I think that it must be a breaking one, otherwise we'll get into bad habits ;)

tjof2:
> I think that it must be a breaking one, otherwise we'll get into bad habits ;)

Sure! Maybe https://github.com/hyperspy/hyperspy/blob/6ef773e19eaf6c19024f8dd580367927f7565a59/hyperspy/learn/mva.py#L537-L538 could return a list of valid string arguments in that case? Perhaps it could return the parameter list from the docstring:

```
raise ValueError(f"'algorithm={algorithm}' not recognised. Expected one of [list of string arguments, custom object]")
```

tjof2 (approved):

tjof2:
LGTM assuming tests pass.

francisco-dlp:
One more commit is on the way.

PR #2473, opened 2020-08-06
Build wheels

ericpre:
### Description of the change
- Build wheels for windows, mac and linux (manylinux1, manylinux2010),
- Use github action to run the build _on tag_ only,
- fix development version (`1.6.0.dev` to `1.6.1.dev0`),
- Fix cython `FutureWarning`.

When a tag is made or push to github, this workflow will run and do the following:
1. create a release on github
2. build the wheels according to [PEP517](https://www.python.org/dev/peps/pep-0517/)
3. Install the wheels and run the test suite
4. the wheels will be available as an artifact (see for example: https://github.com/ericpre/hyperspy/actions/runs/220801777)
5. if running from the upstream repository (`hyperspy/hyperspy`), the wheels will be uploaded to pypi. Obviously, this hasn't been tested yet but the github secret has been set in the repository and this should be ready to go for the next release. If run from a fork, the upload to pypi will be skip to avoid failure of the workflow (missing github secret only available from the `hyperspy/hyperspy` repository). Useful link: https://packaging.python.org/guides/publishing-package-distribution-releases-using-github-actions-ci-cd-workflows/

### DIFF
--- a/hyperspy/Release.py
+++ b/hyperspy/Release.py
@@ -25,7 +25,7 @@ name = 'hyperspy'
 # When running setup.py the ".dev" string will be replaced (if possible)
 # by the output of "git describe" if git is available or the git
 # hash if .git is present.
-version = "1.6.0.dev"
+version = "1.6.1.dev0"
 description = "Multidimensional data analysis toolbox"
 license = 'GPL v3'
 
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,6 +1,10 @@
-name: Build
+name: Release
 
-on: [push, pull_request]
+on:
+  push:
+    # Sequence of patterns matched against refs/tags
+    tags:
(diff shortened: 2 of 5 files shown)


tjof2 (approved):

ericpre:
I think this is in good shape but I will add running the test suite after installing the wheels.

ericpre:
Example of a build from my fork: https://github.com/ericpre/hyperspy/actions/runs/220801777

tjof2 (approved):

hyperspy/Release.py
```diff
@@ -25,7 +25,7 @@
 # When running setup.py the ".dev" string will be replaced (if possible)
 # by the output of "git describe" if git is available or the git
 # hash if .git is present.
-version = "1.6.0.dev"
+version = "1.6.1.dev0"
```
tjof2, line 28: Why the 0?

tjof2 (approved):

PR #2472, opened 2020-08-05
Tidy up long-running and duplicated tests

tjof2:
### Description of the change
- `test_cluster.py` had far too many parameter combinations, including needless duplications, to the extent that it currently constitutes ~15% of all the unit tests by count. This has been reduced considerably for now, it can probably go further.
- `test_hologram_image.py` had unnecessary parameter combinations of `parallel=True/False` which have been removed.
- `test_tools.py` was >1500 lines long and really hard to parse. It has been split up into a number of new files to make it easier to improve the coverage etc. in future. The file names are open for renaming, I chose them based on what seemed like a natural grouping.

This doesn't need to go in v1.6 @francisco-dlp  if you don't want it to! 

### Progress of the PR
- [x] Change implemented (can be split into several points),
- [x] update docstring (if appropriate),
- [x] update user guide (if appropriate),
- [x] add tests,
- [x] ready for review.

PR #2474, opened 2020-08-06
Additional test suite improvements

tjof2:
### Description of the change 

Continuation of #2472, improving the test suite performance and catching a few inconsistencies along the way. Directing it at `RELEASE_next_patch` so that the improvements can be merged into `RELEASE_next_minor` as well. This shaves off a further 90 seconds from the test suite runtime locally. #2472 already improved runs by around 2 minutes.

- Speed up cluster tests
- Speed up BSS tests
- Speed up hologram tests
- Speed up EDS model tests
- Add `pragma: no cover` and `pragma: no branch` to some obvious code branches that are never hit in testing
- Replace one warning with a `ValueError` since the code immediately fails after the warning anyway, plus added test to catch it
- Replace `assert np.allclose` with `np.testing.assert_allclose`, and `import numpy.testing as nt; nt.assert.*` to `import numpy as np; np.testing.assert.*`. This is purely for consistency across the test suite - for example, it makes it much easier in future to find/replace/update all the `np.testing.assert_*` functions in one `grep`/`sed` command.

### Progress of the PR
- [x] add tests,
- [x] ready for review.

### DIFF
--- a/hyperspy/_signals/eds_tem.py
+++ b/hyperspy/_signals/eds_tem.py
@@ -457,9 +457,11 @@ class EDSTEM_mixin:
                                                                 thickness)
                         mass_thickness.metadata.General.title = 'Mass thickness'
                     else:
-                        warnings.warn('Thickness is required for absorption '
-                        'correction with k-factor method. Results will contain '
-                        'no correction for absorption.')
+                        raise ValueError(
+                            'Thickness is required for absorption '
+                            'correction with k-factor method. Results will contain '
+                            'no correction for absorption.'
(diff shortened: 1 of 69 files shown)


tjof2:
Looking at the Python 3.7 test on Travis, it saves about 1 minute more:
```
- Rnm:          1169.84s (0:19:29)
- Rnp:           989.48s (0:16:29) (this is after #2472)
- This branch:   937.29s (0:15:37)
```

ericpre:
Very nice!

Issue #2475, opened 2020-08-06
histogram doesn't allow iterables as a bins argument


pc494:
Numpy's histogram functionality (https://numpy.org/doc/stable/reference/generated/numpy.histogram.html) offers three types for bins:

```
bins : int or sequence of scalars or str, optional

If bins is an int, it defines the number of equal-width bins in the given range (10, by default).
If bins is a sequence, it defines a monotonically increasing array of bin edges, including the rightmost edge,
allowing for non-uniform bin widths.
```

The wrapping in hist_tools isn't set up to handle arrays, and currently (I think) will also throw an error on an `int` as the `.get` method needs a hashable object.

```
 if isinstance(a, da.Array):
        return histogram_dask(a, bins=bins, max_num_bins=max_num_bins, **kwargs)

    _deprecated_bins = {"scotts": "scott", "freedman": "fd"}
    new_bins = _deprecated_bins.get(bins, None)
```
I think the wrap should handle all three cases and would be happy to write a fix.

tjof2:
@pc494 Can you add a test or two along with the fix? Evidently the existing ones didn't check for this and it was overlooked. Thanks for catching this!

Presumably it just needs the handling for anything that is `np.iterable` moved to above the `.get()`?

PR #2477, opened 2020-08-10
Fix for histogram (addressing #2475)

pc494:
### Description of the change
As described in issue (#2475):
- Numpy's histogram functionality offers three types for bins:
`int or sequence of scalars or str`
- The wrapping in hist_tools wasn't set up to handle arrays (or `int`)
- This PR fixes that issue
- The PR also adds a barebones test to prevent regressions

### Progress of the PR
- [ ] update docstring ~(if appropriate)~,
- [ ] update user guide ~(if appropriate)~,
- [x] add tests,
- [x] ready for review.

```python
>>> import hyperspy.api as hs
>>> import numpy as np
>>> s.get_histogram(bins=3)
```

### DIFF
--- a/hyperspy/Release.py
+++ b/hyperspy/Release.py
@@ -25,7 +25,7 @@ name = 'hyperspy'
 # When running setup.py the ".dev" string will be replaced (if possible)
 # by the output of "git describe" if git is available or the git
 # hash if .git is present.
-version = "1.6.0.dev"
+version = "1.7.0.dev"
 description = "Multidimensional data analysis toolbox"
 license = 'GPL v3'
 
--- a/hyperspy/misc/hist_tools.py
+++ b/hyperspy/misc/hist_tools.py
@@ -169,15 +169,16 @@ def histogram_dask(a, bins="fd", max_num_bins=250, **kwargs):
     if a.ndim != 1:
         a = a.flatten()
 
-    _deprecated_bins = {"scotts": "scott", "freedman": "fd"}
-    new_bins = _deprecated_bins.get(bins, None)
-    if new_bins:
-        warnings.warn(
+    if isinstance(bins,str):
+        _deprecated_bins = {"scotts": "scott", "freedman": "fd"}
+        new_bins = _deprecated_bins.get(bins, None)
+        if new_bins:
+            warnings.warn(
             f"`bins='{bins}'` has been deprecated and will be removed "
             f"in HyperSpy 2.0. Please use `bins='{new_bins}'` instead.",
             VisibleDeprecationWarning,
-        )
-        bins = new_bins
+            )
+            bins = new_bins
 
     _old_bins = bins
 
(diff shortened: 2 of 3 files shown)


tjof2:
Thanks @pc494! Please can you point it to (and probably base off) the `RELEASE_next_patch` branch instead?

tjof2:
Also while you're at it, please can you apply the same fix to the dask version?
https://github.com/pc494/hyperspy/blob/833793109edb3e7b51288eeeff96f05babb21368/hyperspy/misc/hist_tools.py#L172-L180

dnjohnstone:
Just to say - It would be great to have an approximate timeline for a 1.6.1 release when this is merged as this is currently breaking a couple of pyxem tests, please :)

tjof2:
See #2479 for the unrelated Travis test failures, but there's still a remaining issue with the docs build 

> Just to say - It would be great to have an approximate timeline for a 1.6.1 release when this is merged as this is currently breaking a couple of pyxem tests, please :)

cc. @francisco-dlp @ericpre

hyperspy/Release.py
```diff
@@ -25,7 +25,7 @@
 # When running setup.py the ".dev" string will be replaced (if possible)
 # by the output of "git describe" if git is available or the git
 # hash if .git is present.
-version = "1.6.0.dev"
+version = "1.7.0.dev"
```
dnjohnstone, line 28: Don't forget you're based on the wrong branch here. I.e. should be "1.6.1.dev"

pc494: Thanks, I had changed it on github but not rebased. Dealt with in 72703cb

tjof2: Pretty certain this file shouldn't be showing up at all in the "changed" list in this PR?

tjof2 (changes requested):
Looks great except `Release.py` still seems to be part of the diff.

pc494:
Still plugging away at getting rid of the stray commit

thomasaarholt:
If you're having some hard-to-detangle git trouble, it might be easier to just create a new branch from `hyperspy/RELEASE_next_patch`, manually add in the changes you've done here and then force push to this branch.

pc494:
> If you're having some hard-to-detangle git trouble, it might be easier to just create a new branch from `hyperspy/RELEASE_next_patch`, manually add in the changes you've done here and then force push to this branch.

Thanks, I went with this option in the end, the new PR is #2482

thomasaarholt:
Nice!
What you did is absolutely fine, but what you could have done is* is take the new branch you created for  #2482  and point it at this branch, with something like `git branch --set-upstream-to origin/fix-for-histogram` and then `git push --force` (force because the git commit history is different).

*after backing up this branch just in case - `git branch fix-for-histogram-backup`

pc494:
Good to know, thanks @tjof2!

PR #2479, opened 2020-08-10
Change rng.random to rng.random_sample

tjof2:
### Description of the change
Quite a few Travis tests failing on Rnm/Rnp - e.g. see https://travis-ci.org/github/hyperspy/hyperspy/jobs/716605308#L1706 as part of #2477, with the error below. I'm guessing this might be a numpy version issue but not looked into it.

```python
AttributeError: 'mtrand.RandomState' object has no attribute 'random'
``` 

This PR replaces it with `np.random.RandomState.random_sample`, which also draws from the half-open interval `[0.0, 1.0)`. 
Note that `np.random.RandomState.uniform` would also work, but that has the option of a different interval to `[0.0, 1.0)` so isn't necessary.

### Progress of the PR
- [x] ~update docstring (if appropriate),~
- [x] ~update user guide (if appropriate),~
- [x] ~add tests,~
- [x] ready for review.

### DIFF
--- a/hyperspy/tests/learn/test_bss.py
+++ b/hyperspy/tests/learn/test_bss.py
@@ -74,7 +74,7 @@ def test_bss_FastICA_object():
     """Tests that a simple sklearn pipeline is an acceptable algorithm."""
     rng = np.random.RandomState(123)
     S = rng.laplace(size=(3, 1000))
-    A = rng.random((3, 3))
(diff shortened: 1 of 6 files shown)


tjof2:
Travis failure is docs only, not touched here.

tjof2:
Will need to merge Rnp into Rnm as well to fix Travis everywhere.

PR #2485, opened 2020-08-16
Test against oldest version of matploltib, numpy and scipy

ericpre:
### Progress of the PR
- [x] add a build to travis to check the oldest supported version of matplotlib, numpy and scipy,
- [x] ready for review.

There are binaries incompatibilities issues for old version of numpy and the imagecodecs wheels on linux, which make me think we should possibly make imagecodecs an optional dependencies - at the moment imagecodecs is installed through `tifffile[all]`. This should not be an issue for more recent version and therefore it should be a fairly niche issue.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -37,7 +37,7 @@ install:
   # manually
   - pip install --upgrade pytest
   - if [ "$MINIMUM_VERSION" == true ]; then
-      pip install matplotlib==2.2.3 numpy==1.14 scipy==1.0.1 numba==0.40;
+      pip install matplotlib==2.2.3 numpy==1.15 scipy==1.0.1 numba==0.46 imagecodecs==2018.10.10;
       pip install -e .["${PIP_SELECTOR}"];
     else
       pip install --upgrade -e .["${PIP_SELECTOR}"];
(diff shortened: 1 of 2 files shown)


tjof2:
Is it worth documenting somewhere the rationale behind the different test environments? e.g. that Travis is pip, Azure is conda, 3.6 through to 3.8 is tested, codecov bot only runs if the  3.7 Linux Travis test passes, this "oldest version" test...

thomasaarholt:
@ericpre, this is a very good idea!
@tjof2 I'd like that! I just learnt now that Travis is testing pip! (how come it isn't in the azure suite? Just because it was already in Travis?)

ericpre:
There is already quite a bit in the dev guide on this topic: https://hyperspy.readthedocs.io/en/latest/dev_guide/testing.html#continuous-integration-ci

If we are happy with the approach, I will update the dev guide regarding this "oldest supported version" build.

tjof2:
The docs are good @ericpre but there's no rationale or explanation for e.g.

- Python 3.7 on Travis being the only place we test the minimum requirements for installation (no sklearn etc)
- how/why that test is different to the one you propose here in this PR
- codecov bot only works if the Python 3.7 Travis test passes. If it fails but all the others pass, there's no coverage report generated
- docs build is also done by readthedocs now

tjof2:
Let's tackle the docs separately, I'll take a look this week.

PR #2501, opened 2020-08-25
Fix y-limits when plotting several lines

### LINKED ISSUE #2481: Incorrect ylim on m.plot(True) when stepping through indices
When stepping through a multidimensional Signal1D, the `ylim` on the Signal plot resizes to the size of the final component in the model. I would expect it to keep track of the lowest value and the highest value of all the spectra that are plotted, and then set `plt.ylim(lowest, highest)`

In the example below, the horizontal red line is the signal data (`s`), the blue gaussian is the total model fit, and the green and red gaussians are the first and second components, respectively. In the second image, where I have stepped forward one index, the ylim currently scales to the red gaussian, hiding the rest of the plot.

```python
# %matplotlib widget
import hyperspy.api as hs
import numpy as np
from hyperspy.components1d import GaussianHF

data = np.ones((3, 200,))
s = hs.signals.Signal1D(data)

m = s.create_model()
G1 = GaussianHF(height=3, centre = 50, fwhm=20)
G2 = GaussianHF(height=2, centre = 50, fwhm=20)
m.extend([G1, G2])
m.plot(plot_components=True)
```

ericpre:
Fix #2481 and enhancements.

### Progress of the PR
- [x] Fix y limits when plotting several lines,
- [x] fix redundant drawing events when plotting several lines,
- [x] display remainder line on right axis when removing background,
- [ ] ~~update docstring (if appropriate),~~
- [ ] ~~update user guide (if appropriate),~~
- [x] add tests,
- [x] ready for review.

Example adapted from #2481
```python
import hyperspy.api as hs
import numpy as np

from hyperspy.components1d import GaussianHF
data = np.arange(5*4*200).reshape((5, 4, 200,))
s = hs.signals.Signal1D(data)

m = s.create_model()
G1 = GaussianHF(height=500, centre = 50, fwhm=20)
G2 = GaussianHF(height=100, centre = 50, fwhm=20)
m.extend([G1, G2])
m.plot(plot_components=True)
```
When changing the navigation index, all the lines are taken into account when calculating the y limits.

### Minimal example of the enhancement
```python
import hyperspy.api as hs
s = hs.datasets.artificial_data.get_core_loss_eels_line_scan_signal(True, True)
s.remove_background()
```
image


hyperspy/drawing/signal1d.py
```diff
@@ -364,15 +417,32 @@ def _auto_update_line(self, *args, **kwargs):
...
+               update_ylimits=False):
+        """Update the current spectrum figure
+
+        Parameters:
+        -----------
+        force_replot : bool
+            If True, close and open the figure. Default is False.
+        render_figure : bool
+            If True, render the figure. Useful to avoid firing matplotlib
+            drawing events too event. Default is True.
```
tjof2: "too often"

ericpre: Done, thanks!

tjof2 (approved):

ericpre:
I made this PR to RnM instead of RnP by mistake, so I have backport it to RnP by pushing these commit into RnP (the base was RnP).

tjof2:
Didn't catch the base there - assumed it was tagged incorrectly, sorry!

ericpre:
No problem, I do this mistake once in a while... 🤷‍♂️

PR #2502, opened 2020-08-26
Warn on mismatched axis calibrations when stacking

### LINKED ISSUE #1570: Bug (?) in load_data()

tjof2:
### Description of the change
Replaces #1572, closes #1570. 

When stacking a list of signals, will raise a warning if any of the signals have axis calibrations that don't match. The reference axis calibration is taken as signal with the "longest" axis, as per the sort function present in `hyperspy.misc.signal_tools.broadcast_signals()`.

### Progress of the PR
- [x] add tests,
- [x] ready for review.

```python
import hyperspy.api as hs
from hyperspy.misc.utils import stack

s1 = hs.signals.Signal1D([1.0, 2.0, 3.0])
s2 = hs.signals.Signal1D([1.0, 2.0, 3.0])

s1.axes_manager[0].scale = 1.0
s1.axes_manager[0].offset = 1.0
s1.axes_manager[0].units = "cm"

s2.axes_manager[0].scale = 2.0
s2.axes_manager[0].offset = 1.0 + 1e-5
s2.axes_manager[0].units = "mm"

s = stack([s1, s2], axis=0)
# WARNING:hyperspy.misc.signal_tools:Axis calibration mismatch detected along axis 0. The calibration of signal 0 along this axis will be applied to all signals after stacking.
# <Axes manager, axes: (|6)>
#             Name |   size |  index |  offset |   scale |  units 
# ================ | ====== | ====== | ======= | ======= | ====== 
# ---------------- | ------ | ------ | ------- | ------- | ------ 
#      <undefined> |      6 |        |       1 |       1 |     cm 

s = stack([s1, s2])
# WARNING:hyperspy.misc.signal_tools:Axis calibration mismatch detected along axis 0. The calibration of signal 0 along this axis will be applied to all signals after stacking.
# <Axes manager, axes: (2|3)>
#             Name |   size |  index |  offset |   scale |  units 
# ================ | ====== | ====== | ======= | ======= | ====== 
#    stack_element |      2 |      0 |       0 |       1 | <undefined> 
# ---------------- | ------ | ------ | ------- | ------- | ------ 
#      <undefined> |      3 |        |       1 |       1 |     cm 
```


hyperspy/misc/utils.py
```diff
@@ -907,102 +906,97 @@ def stack(signal_list, axis=None, new_axis_name='stack_element',
     from hyperspy.signals import BaseSignal
     import dask.array as da
     from numbers import Number
-    # TODO: remove next time
```
ericpre, line 910: I think this need to stay for now and it should be removed in v2.0 because it has been deprecated in v1.2 (see https://github.com/hyperspy/hyperspy/commit/28be16949b183585c8d4ff85d4d72762ce8c6a15 in #1219). The comment could be more explicit!

hyperspy/misc/signal_tools.py
```diff
@@ -56,11 +63,83 @@ def are_signals_aligned(*args, ignore_axis=None):
     return True
 
 
+def _check_axes_calibration(ax1, ax2):
+    """Check if the calibration of two axes matches.
```
ericpre: This equality operation could be useful for other operations and a better place for this function may be `hyperspy.axes` or an utils module?
A note for later, this would be to be adapted in the `non_uniform_axes`.

hyperspy/misc/utils.py
tjof2: Done.

hyperspy/misc/signal_tools.py
tjof2: I've moved it to `hyperspy/misc/axis_utils.py`. Had a bit of trouble with circular imports so haven't made it any more complicated.

PR #2503, opened 2020-08-26
Remove parallel argument from .as_signal

### LINKED ISSUE #1604: m.as_signal progress bar bug
When using `m.as_signal(parallel=True, show_progressbar=True)`, in an IPython terminal a large number of what looks like newlines are outputted:

```Python
import hyperspy.api as hs
import numpy as np

s = hs.signals.Signal1D(np.zeros((10, 10)))
m = s.create_model()
m.as_signal(parallel=True, show_progressbar=True)
```
Gives this:
parallel_true_show_progress_true

Using `m.as_signal(parallel=True, show_progressbar=False)` still tries to show the progress bar:
parallel_true_show_progress_false

Using `m.as_signal(parallel=False, show_progressbar=False)` and `m.as_signal(parallel=False, show_progressbar=True)` works fine.
parallel_false_show_progress_false
parallel_false_show_progress_true

### LINKED ISSUE #2443: `model.as_signal` is not working when using `component_list` and `parallel=True`
The following example raises an error with `parallel=True`:

```python
s = hs.signals.Signal1D(np.random.random((10,100)))
m = s.create_model()
l1 = hs.model.components1D.Lorentzian(centre=20)
m.append(l1)
s2 = m.as_signal(component_list=[l1], parallel=True)
```
Error:
```python

  File "<ipython-input-2-c9621a70c23c>", line 5, in <module>
    s2 = m.as_signal(component_list=[l1], parallel=True)

  File "/home/eric/Dev/hyperspy/hyperspy/model.py", line 541, in as_signal
    _ = next(_map)

  File "/opt/miniconda3/lib/python3.7/concurrent/futures/_base.py", line 598, in result_iterator
    yield fs.pop().result()

  File "/opt/miniconda3/lib/python3.7/concurrent/futures/_base.py", line 428, in result
    return self.__get_result()

  File "/opt/miniconda3/lib/python3.7/concurrent/futures/_base.py", line 384, in __get_result
    raise self._exception

  File "/opt/miniconda3/lib/python3.7/concurrent/futures/thread.py", line 57, in run
    result = self.fn(*self.args, **self.kwargs)

  File "/home/eric/Dev/hyperspy/hyperspy/model.py", line 538, in <lambda>
    show_progressbar=thing[2] + 1 if show_progressbar else False),

  File "/home/eric/Dev/hyperspy/hyperspy/model.py", line 563, in _as_signal_iter
    for x in component_list]

  File "/home/eric/Dev/hyperspy/hyperspy/model.py", line 563, in <listcomp>
    for x in component_list]

  File "/home/eric/Dev/hyperspy/hyperspy/model.py", line 336, in _get_component
    raise ValueError("The component is not in the model.")

ValueError: The component is not in the model.
```

### LINKED ISSUE #2506: Model.as_signal ignores channel_switches if parallel=True for single spectrum case
HyperSpy 1.6.

See also #2503 Remove parallel argument from as_signal.

```
import hyperspy as hs
import hyperspy.datasets
e = hs.datasets.artificial_data.get_core_loss_eels_signal(add_powerlaw=True)
s = e
m = hyperspy.models.model1d.Model1D(s)
c = hyperspy.signal_tools._get_background_estimator("powerlaw", 1)[0]
m.append(c)
c.estimate_parameters(s, 600, 650)
m.channel_switches[:] = False
m.channel_switches[250:300] = True
ms1 = m.as_signal(parallel=False, out_of_range_to_nan=False)
ms2 = m.as_signal(parallel=True, out_of_range_to_nan=False)
print(ms1.data)  # correct, only enabled channels calculated
print(ms2.data)  # incorrect, all data calculated
```

tjof2:
### Description of the change

As well as being the source of ~two~ three bugs (see linked issues), `m.as_signal(parallel=True)` does not actually offer any speed-up over `parallel=False`. In fact, it looks to be much slower:

```python
import hyperspy.api as hs
import numpy as np

def create_test_model():
    rng = np.random.RandomState(123)

    s = hs.signals.Signal1D(rng.uniform(size=(32,32,32,32)))
    m = s.create_model()
    l1 = hs.model.components1D.Lorentzian(centre=20)
    l2 = hs.model.components1D.Lorentzian(centre=10)
    m.append(l1)
    m.append(l2)
    
    return m

m1 = create_test_model()
%timeit m1.as_signal(parallel=True, show_progressbar=False)
# 19.9 s ± 1.44 s per loop (mean ± std. dev. of 7 runs, 1 loop each)

m2 = create_test_model()
%timeit m2.as_signal(parallel=False, show_progressbar=False)
# 8.98 s ± 93.3 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
```

This is because calling this function with `parallel=True` uses thread-based parallelism (`concurrent.futures.ThreadPoolExecutor`), but the code it runs is pure Python, so won't release the GIL. This PR removes `parallel=True` and `max_workers=True` arguments, such that they no longer have any effect (and will instead raise a deprecation warning). 

Linked issues:
- Closes #1604
- Closes #2443 
- See also #2506

### Progress of the PR
- [x] add tests,
- [x] ready for review.


tjof2:
I believe the codecov warning can be ignored, it's probably because a large number of lines was deleted.

ericpre:
Indeed threading in this case doesn't look right, since this operation is CPU bound... However, the speed of this method could be improved significantly by using the `function_nd` method of the components (when implemented) instead of iterating over the navigation axes.

The only reason to keep it would be to convert to multiprocessing for lazy signal (which is not supported anyway) and this is assuming that using `function_nd` can't be made to work lazily. Getting `function_nd` to work lazily may be easier anyway and possibly still faster.

tjof2:
@ericpre thanks for the comment, sounds like a good solution with function_nd.

 What's the best way to proceed then? Merge this and address separately? It certainly would be nice to remove the ThreadPoolExecutor because this is implemented here separately to `.map()` which was confusing at first.

ericpre:
I think it is fine to remove it in this PR and then using `function_nd` in a separate PR. If someone want to implement multi-processing, it is still possible to come back to this PR or through the git history. @francisco-dlp, what do you think?

cmeyer:
I've been looking through this code recently...

Are you sure the GIL is not released? I think if SymPy uses NumPy or NumExpr backends, the GIL is released.

An alternate explanation for parallel being slower is that there is overhead for SymPy (lambdify), so for small data sets, the overhead dominates the timing.

total time = number of threads x overhead + overlapped processing time

So decreasing the number of threads will speed things up if overhead is roughly the same magnitude as processing time.

But as the data set gets larger and processing time is much longer than overhead, threading becomes beneficial.

TL;DR: test with a variety of data sizes, i.e. 32x32 | 512 vs 512x512 | 2048 may have significant different responses to threading.

tjof2:
@cmeyer I'm not sure it matters since there is already quite a lot of other Python code included in each loop (for example, handling the progressbars even if `progressbar=False`). 

If it was just `parallel map(some numexpr function)` I'd certainly agree with you. But it's not, instead it's an implementation with a few other key parts. Also, if the `numpy` or `numexpr` function backend for SymPy is also multithreaded, it will likely interact poorly with this additional threading.

Anyway, as you suggested I tested a few further things with `show_progressbar=False` and `max_workers=4`. Here the dimensions are `(n, n|1024)`, and the value range for `n` is `[2**x for x in range(8)]`, so up to `128x128`.

i1

Repeating it with  `(n, n|2048)`:

i2

And lastly, repeating it with  `(n, n|64)`:

i3

Lastly, for something really large -  `(512, 512|2048)` -
```python
%timeit m1.as_signal(parallel=True, show_progressbar=False)
# 3min 39s ± 13.9 s per loop (mean ± std. dev. of 3 runs, 1 loop each)

%timeit m2.as_signal(parallel=False, show_progressbar=False)
# 1min 8s ± 582 ms per loop (mean ± std. dev. of 3 runs, 1 loop each)
```

cmeyer:
@tjof2 Those are great graphs that make the situation clear up to 120x120.

I'm not too concerned about this as it sounds like things are well in hand - but if you wanted to run another experiment, it would interesting to try out a smaller number of threads. We have a 12-core system and with hyper-threading, it says it has 24 processors. NumPy should clearly not utilize hyper-threading - so we've found that artificially reducing to 12 or even 4 results in better performance.

Also, your system may not have enough memory, but larger sizes up to 512x512 may show different results too.

I'm out for today, but I can run some tests on our machines in the next few days. Is the code to run the timing tests the same as the code at the top of this issue?

tjof2:
@cmeyer The results are for a 4-core CPU, with `max_workers=4` for 4 threads. I tried 2 and 3 and saw similar-ish results on my system. I've also tried it on a newer 6-core CPU with a few different options for `max_workers` and the same behaviour remains. 

> Also, your system may not have enough memory, but larger sizes up to 512x512 may show different results too.

See the last section of my previous comment that tests `(512, 512|2048)`.

Ultimately, unless someone is able to find an example where `parallel=True` is clearly faster than `parallel=False`, then I stand by the idea that it should be removed. Otherwise what do we say to users about when to use this option? Anyone blindly thinking it would speed things up would be very wrong based on the evidence thus far. There's no point using multiple threads to get similar performance to a single thread - surely one would need a significant speed-up to justify it.

Yes, you can use the code at the top here, I adapted it a little to produce the plots but it's the same principle. You can also do `pip install perfplot`, then:

```python
import hyperspy.api as hs
import numpy as np
import perfplot

def create_test_model(a=32, b=1024):
    rng = np.random.RandomState(123)

    s = hs.signals.Signal1D(rng.uniform(size=(a, a, b)))
    m = s.create_model()
    l1 = hs.model.components1D.Lorentzian(centre=20)
    l2 = hs.model.components1D.Lorentzian(centre=10)
    m.append(l1)
    m.append(l2)

    return m

perfplot.show(
    setup=lambda n: create_test_model(n, b=1024),
    kernels=[
        lambda m: m.as_signal(parallel=False, show_progressbar=False),
        lambda m: m.as_signal(parallel=True, show_progressbar=False, max_workers=4),
    ],
    labels=["parallel=False", "parallel=True"],
    n_range=[2**n for n in range(8)],
    xlabel="sqrt(nav dimension)",
    time_unit="s",
    logx=False,
    logy=True,
)
```
will produce: 

i1

ericpre:
@cmeyer, is there any benchmark you would like to add on this topic? If not, I will merge because I am convinced that the approach of this PR is correct and the speed improvement needs to be achieved by using `function_nd`.

cmeyer:
@ericpre Looks good - no objection from me. I'll continue to run benchmarks, but file separately or submit patches if needed.

PR #2510, opened 2020-08-30
Fix oldest versions build

ericpre:
The oldest version build introduced in #2485 wasn't working because pip was upgrading the libraries set at specific versions...

### Progress of the PR
- [x] Increase minimum version requirements for `numpy` and `scipy` to avoid pip upgrading them,
- [x] use the coming next-generation dependency resolver of pip (`--use-feature=2020-resolver`), which is be enabled from October 2020, see https://blog.python.org/2020/07/upgrade-pip-20-2-changes-20-3.html for all build except for the "oldest supported version" build,
- [x] set `imagecodecs` as optional dependencies (both have marginally used in hyperspy and come with compiled code which can caused binaries incompatibility),
- [x] remove `statsmodels` from dependencies and adapt a version of https://gist.github.com/agramfort/850437 for lowess smoothing, 
- [x] ready for review.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -21,7 +21,9 @@ matrix:
     env: PIP_SELECTOR="tests"
     python: 3.7
   - name: "Linux, 3.6, pip, mininum version"
-    env: MINIMUM_VERSION=true
+    env:
+      - MINIMUM_VERSION=true
+      - PIP_SELECTOR="tests"
     python: 3.6
   - name: "Linux, 3.7, pip, doc"
     env: PIP_SELECTOR="all, build-doc"
@@ -39,7 +41,7 @@ install:
   # Test against the oldest supported version of matplotlib, numpy and scipy
   # Use old version numba and imagecodecs to work around incompatibilities
   - if [ "$MINIMUM_VERSION" == true ]; then
-      pip install matplotlib==2.2.3 numpy==1.15 scipy==1.0.1 numba==0.40 imagecodecs==2018.10.10;
+      pip install matplotlib==2.2.3 numpy==1.15 scipy==1.0.1 numba==0.40;
       pip install -e .["${PIP_SELECTOR}"];
     else
       pip install --upgrade -e .["${PIP_SELECTOR}"];
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,10 +3,14 @@ language: python
 
 env:
   global:
+    # The default environment variable are defined here and overwritten
+    # in the matrix when necessary
     - ENV=pip
+    - PIP_ARGS="--upgrade --use-feature=2020-resolver -e"
     - PIP_SELECTOR="all, tests, coverage"
     - PYTEST_ARGS="--pyargs hyperspy"
-    - MINIMUM_VERSION=false
+    - OLDEST_SUPPORTED_VERSION=false
+    - MPLBACKEND="agg"
 
 matrix:
   include:
@@ -20,9 +24,10 @@ matrix:
   - name: "Linux, 3.7, pip, minimum requirement"
     env: PIP_SELECTOR="tests"
     python: 3.7
-  - name: "Linux, 3.6, pip, mininum version"
+  - name: "Linux, 3.6, pip, oldest supported versions"
     env:
-      - MINIMUM_VERSION=true
+      - OLDEST_SUPPORTED_VERSION=true
+      - PIP_ARGS="-e"
       - PIP_SELECTOR="tests"
     python: 3.6
   - name: "Linux, 3.7, pip, doc"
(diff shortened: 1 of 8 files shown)


tjof2:
Why not replace `lowess` in https://github.com/hyperspy/hyperspy/blob/5e8cdbb3e7a3aa6d76bcdd7b0ba45c2dd1345660/hyperspy/_signals/signal1d.py#L994 with https://gist.github.com/agramfort/850437 and remove the `statsmodels` dependency from HyperSpy altogether? It's BSD 3-clause.

ericpre:
> Why not replace `lowess` in
> 
> https://github.com/hyperspy/hyperspy/blob/5e8cdbb3e7a3aa6d76bcdd7b0ba45c2dd1345660/hyperspy/_signals/signal1d.py#L994
> 
> with https://gist.github.com/agramfort/850437 and remove the `statsmodels` dependency from HyperSpy altogether? It's BSD 3-clause.

Sounds very good idea to me!

ericpre:
I have added the lowess function from https://gist.github.com/agramfort/850437. The speed is a bit slower (maybe ~50% difference) but in some situation, it does a `LinAlgError: Singular matrix` error, for example:
```python
import hyperspy.api as hs
import numpy as np

m = 20
s = hs.signals.Signal1D(np.arange(m))
s.add_poissonian_noise()
s.smooth_lowess(smoothing_parameter=0.1, number_of_iterations=1, show_progressbar=False)
```

```python

  File "<ipython-input-1-bd00579c8f40>", line 7, in <module>
    s.smooth_lowess(smoothing_parameter=0.1, number_of_iterations=1, show_progressbar=False)

  File "/home/eric/Dev/hyperspy/hyperspy/_signals/signal1d.py", line 988, in smooth_lowess
    max_workers=max_workers)

  File "/home/eric/Dev/hyperspy/hyperspy/signal.py", line 4511, in map
    **kwargs)

  File "/home/eric/Dev/hyperspy/hyperspy/signal.py", line 4643, in _map_iterate
    range(res_data.size), executor.map(func, zip(*iterators))

  File "/opt/miniconda3/lib/python3.7/concurrent/futures/_base.py", line 598, in result_iterator
    yield fs.pop().result()

  File "/opt/miniconda3/lib/python3.7/concurrent/futures/_base.py", line 435, in result
    return self.__get_result()

  File "/opt/miniconda3/lib/python3.7/concurrent/futures/_base.py", line 384, in __get_result
    raise self._exception

  File "/opt/miniconda3/lib/python3.7/concurrent/futures/thread.py", line 57, in run
    result = self.fn(*self.args, **self.kwargs)

  File "/home/eric/Dev/hyperspy/hyperspy/misc/utils.py", line 1083, in func
    return function(dat, **these_kwargs)

  File "/home/eric/Dev/hyperspy/hyperspy/misc/lowess_smooth.py", line 48, in lowess
    beta = np.linalg.solve(A, b)

  File "<__array_function__ internals>", line 6, in solve

  File "/opt/miniconda3/lib/python3.7/site-packages/numpy/linalg/linalg.py", line 394, in solve
    r = gufunc(a, b, signature=signature, extobj=extobj)

  File "/opt/miniconda3/lib/python3.7/site-packages/numpy/linalg/linalg.py", line 88, in _raise_linalgerror_singular
    raise LinAlgError("Singular matrix")

LinAlgError: Singular matrix
```

tjof2:
@ericpre I'll take a look at improving it this afternoon

ericpre:
That would be nice, thanks!

tjof2:
@ericpre try the following in ` hyperspy/misc/lowess_smooth.py`:

```python
import numpy as np
from scipy.linalg import lstsq

def lowess(y, x, f=2.0 / 3.0, n_iter=3):
    """Lowess smoother (robust locally weighted regression).
        
    Fits a nonparametric regression curve to a scatterplot.
    
    Parameters
    ----------
    y, x : np.ndarrays
        The arrays x and y contain an equal number of elements; 
        each pair (x[i], y[i]) defines a data point in the 
        scatterplot.
    
    f : float
        The smoothing span. A larger value will result in a
        smoother curve. 
    n_iter : int
        The number of robustifying iteration. Thefunction will
        run faster with a smaller number of iterations.
        
    Returns
    -------
    yest : np.ndarray
        The estimated (smooth) values of y.
    
    """   
    n = len(x)
    r = int(np.ceil(f * n))
    h = [np.sort(np.abs(x - x[i]))[r] for i in range(n)]
    w = np.clip(np.abs((x[:, None] - x[None, :]) / h), 0.0, 1.0)
    w = (1 - w ** 3) ** 3
    yest = np.zeros(n)
    delta = np.ones(n)
    for _ in range(n_iter):
        for i in range(n):
            weights = delta * w[:, i]
            b = np.array([np.sum(weights * y), np.sum(weights * y * x)])
            A = np.array(
                [
                    [np.sum(weights), np.sum(weights * x)],
                    [np.sum(weights * x), np.sum(weights * x * x)],
                ]
            )
            
            beta = lstsq(A, b)[0]
            yest[i] = beta[0] + beta[1] * x[i]

        residuals = y - yest
        s = np.median(np.abs(residuals))
        delta = np.clip(residuals / (6.0 * s), -1.0, 1.0)
        delta = (1 - delta ** 2) ** 2

    return yest
```
Using your example:
```python
import hyperspy.api as hs

m = 20
s = hs.signals.Signal1D(np.arange(m))
s.add_poissonian_noise(random_state=123)

yest1 = lowess(y=s.data, x=s.axes_manager[-1].axis, f=0.1, n_iter=1)

from statsmodels.nonparametric.smoothers_lowess import lowess as lowess_sm

yest2 = lowess_sm(endog=s.data, exog=s.axes_manager[-1].axis, frac=0.1, it=1, is_sorted=True, return_sorted=False)

np.allclose(yest1, yest2)
# True
```

>  The speed is a bit slower (maybe ~50% difference) 

statsmodels uses a cythonized version, which will probably explain that.

ericpre:
Thanks @tjof2, this is working nicely! I added a numba jit and this is now running faster than the cython version from statsmodel. :) Also since the numba implementation is releasing the GIL, `map` parallelisation play well too!

tjof2 (approved):

Issue #2452, opened 2020-07-27
non-uniform axes branch mod.multifit() leaves the signal with navigation indices in an out-of-bound state


LMSC-NTappy:
The following code generates an error when mod.plot() is called:

```python
#generate a sample gaussian lineshape dataset
shape = (4,4,100)
dat = np.zeros(shape)
for i in range(shape[0]):
    for j in range(shape[1]):
        dat[i,j,:] = np.histogram(np.random.normal(50,2,int(1e5)),bins=100,range=(40,60))[0].astype('int32')
artificial_map = hs.signals.Signal1D(dat)

#Create gaussian model and fit
gauss = hs.model.components1D.Gaussian(centre=50,A=1e5,sigma=1)
mod = artificial_map.create_model()
mod.append(gauss)
mod.multifit()
```

Here is the error:

```python
mod.plot()
---------------------------------------------------------------------------
IndexError                                Traceback (most recent call last)
<ipython-input-6-97bc6e2748ea> in <module>
----> 1 mod.plot()

c:\users\nicolastappy\documents\git\hyperspy\hyperspy\models\model1d.py in plot(self, plot_components, **kwargs)
    658 
    659         # If new coordinates are assigned
--> 660         self.signal.plot(**kwargs)
    661         _plot = self.signal._plot
    662         l1 = _plot.signal_plot.ax_lines[0]

c:\users\nicolastappy\documents\git\hyperspy\hyperspy\signal.py in plot(self, navigator, axes_manager, plot_markers, **kwargs)
   2159                     " \"slider\", None, a Signal instance")
   2160 
-> 2161         self._plot.plot(**kwargs)
   2162         self.events.data_changed.connect(self.update_plot, [])
   2163         if self._plot.signal_plot:

c:\users\nicolastappy\documents\git\hyperspy\hyperspy\drawing\mpl_he.py in plot(self, **kwargs)
    174             pointer = self.assign_pointer()
    175             if pointer is not None:
--> 176                 self.pointer = pointer(self.axes_manager)
    177                 self.pointer.color = 'red'
    178                 self.pointer.connect_navigate()

c:\users\nicolastappy\documents\git\hyperspy\hyperspy\drawing\_widgets\rectangles.py in __init__(self, axes_manager, **kwargs)
     39 
     40     def __init__(self, axes_manager, **kwargs):
---> 41         super(SquareWidget, self).__init__(axes_manager, **kwargs)
     42 
     43     def _set_patch(self):

[...]

PR #2512, opened 2020-08-31
Fix for #2452 - index-dependent plot error in non-uniform-axis branch

jlaehne:
### Description of the change
Workaround for #2452. The `_get_step` function failed if index was on the last value of a navigation axis, because `index + 1` is then undefined. Instead now uses `index - 1` in this specific case.

### Progress of the PR
- [X] ready for review.
- [X] Add test


ericpre:
Can you please add tests checking the behaviour of this `_get_step` function?

jlaehne:
Added test

ericpre (approved):

PR #2513, opened 2020-08-31
fix gradients and use numpy function for heaviside expression

### LINKED ISSUE #2067: Error in heaviside function gradient?
While browsing through model components to help me figure out how to make my one of my own I stumbled upon something in the Heaviside function: 
https://github.com/hyperspy/hyperspy/blob/RELEASE_next_minor/hyperspy/_components/heaviside.py

I think that the function currently defined as grad_n is in reality grad_A:
since we fit A*H(x-n), A-gradient should be H(x-n)

and then I am unsure about grad_n, analytically I would say -A*delta(x-n) but this cannot be numerically evaluated...

Anyways, this is certainly not a big deal. 

Cheers

Nicolas

jlaehne:
### Description of the change
* Use `numpy.heaviside` instead of `numexpr.where` for definition of heaviside as in #2067
* Fixed gradients as in #2067

Should resolve  #2067.

### Progress of the PR
- [X] adapted tests,
- [X] ready for review.

```python
h1 = hs.model.components1D.HeavisideStep(A=3,n=2)
x = np.arange(0,5,step=0.2)
h1.function(x)
#array([0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 3., 3., 3., 3., 3., 3., 3., 3., 3., 3., 3., 3., 3., 3., 3.])
h1.A.grad(x)
#array([0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1])
```


hyperspy/_components/heaviside.py
```diff
@@ -29,10 +29,12 @@ class HeavisideStep(Expression):
     .. math::
 
         f(x) =
-        \\begin{cases}
-        0 & x<n\\\\
-        A & x>=n
-        \\end{cases}
+        \begin{cases}
```
ericpre, line 36: Is it worth documenting that this is using the `half-maximum convention` as defined by numpy?

ericpre (commented):
Looks good to me, see comments for the docstring.

ericpre:
Would it be possible to rebase on RnP and send it to RnP instead? If not, this is not a big deal.

hyperspy/_components/heaviside.py
jlaehne: Good point. I also added the table of variables we have for other components.

jlaehne:
I hope the rebase worked.

tjof2:
> I hope the rebase worked.

Doesn't look like it I'm afraid :( - you've included a number of files/changes on Rnm that shouldn't go into Rnp.

jlaehne:
OK, just wrote it into a new branch derived from RnP: #2514

Issue #2252, opened 2019-09-06
Closing the hyperspy-users google group


francisco-dlp:
We created the hyperspy-users Google group before we started using GitHub. Now, most of us use Gitter or GitHub issues for most of the things that we were using the Google group in the past. The only exception are release announcements. Should we close it? If yes, how do you suggest that we announce new releases? Twitter comes to my mind, but I am not an user. Thoughts?

ericpre:
The user list still have the advantage of being easier to search and browse. However, I am not convinced that it is as much used as it could be. There is significantly more discussion on gitter but most of it get lost over time, when there is no follow up with an issues of a PR on github.
Instead of losing this information, we should try to collect these example in galleries as discussed in #1952!

Regarding announcement of new releases on the user list, I don't know if it is useful to many people. From my experiment, I would say that people who want to be stay up to date, they will follow different things such as the gitter chat or check the website once in a while. 

I am not a Twitter user but I believe that a twitter account would be more useful than the user list and it would have less overlap between gitter and twitter.

thomasaarholt:
Is there any harm in leaving it open? We could simply encourage users to use the gitter chat, yet leave the email open for announcements. A twitter feed is also not a bad idea, though this is the sort of thing that would quickly fall into disuse.

ericpre:
The recent instance of an user asking the same question on the mailing list and on gitter indicates that this is not clear where is the best place to ask questions. From looking at the google group, it may appear that the question was left unanswered, which is not great and not true. Overall, it may appear that this is very little activity which is not representative.
I agree with closing the user group to avoid any confusion.

An alternative to closing this group would be to:
- keep the list for announcement only
- [stop members from posting](https://support.google.com/groups/answer/2646833?hl=en&ref_topic=2458761)
- make clear that this is for announcement only and recommend users to use gitter to ask questions and github for feature requests or bug reports.

But the google group would still appear as _not very active_.

jlaehne:
For announcements of new releases, people can just `watch (releases only)` the repo on github. So if that is the only use of the google group currently, it should be obsolete.

How about adding a line somewhere on the website that states "If you want to be notified about new releases, please watch the hyperspy repositorium on <link>GitHub</link>".

ericpre:
> For announcements of new releases, people can just `watch (releases only)` the repo on github. So if that is the only use of the google group currently, it should be obsolete.
> 
[...]

PR #2515, opened 2020-08-31
Remove reference to googlegroup

jlaehne:
Removes reference to google group based on the discussion at:
hyperspy/hyperspy#2252

[X] Remove googlegroup from `CONTRIBUTING`
[X] Add note on 'watching' GitHub releases to user guide

### DIFF
--- a/doc/user_guide/install.rst
+++ b/doc/user_guide/install.rst
@@ -14,6 +14,12 @@ see the :ref:`anaconda-install` section.
 To enable context-menu (right-click) shortcut in a chosen folder, use the
 `start_jupyter_cm <https://github.com/hyperspy/start_jupyter_cm>`_ library.
 
+.. note::
+
+    If you want to be notified about new releases, please *Watch (Releases only)* 
+    the `hyperspy repositorium on GitHub <https://github.com/hyperspy/hyperspy/>`_ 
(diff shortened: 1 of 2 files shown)


doc/user_guide/install.rst
```diff
@@ -14,6 +14,12 @@ see the :ref:`anaconda-install` section.
 To enable context-menu (right-click) shortcut in a chosen folder, use the
 `start_jupyter_cm <https://github.com/hyperspy/start_jupyter_cm>`_ library.
 
+.. note::
+
+    If you want to be notified about new releases, please *Watch (Releases only)* 
+    the `hyperspy repositorium on GitHub <https://github.com/hyperspy/hyperspy/>`_
```
tjof2: The standard term is "repository" I think.

jlaehne: sure ;-)

tjof2 (changes requested):
Needs to be based off RELEASE_next_patch I think @jlaehne

tjof2:
Same problem as before though I'm afraid, it includes commits that shouldn't be there.

jlaehne:
OK, figured it out should be fine now.

tjof2 (approved):

PR #2520, opened 2020-09-01
Support parallel test execution on CI

tjof2:
### Description of the change
Attempt to support `pytest-xdist` on Travis and Azure Pipelines. In both cases, it looks like there are 2 cores available per worker (do you know any more details @ericpre?), so parallelizing the test suite over at least 2 cores should be possible. This adds `pytest-xdist` to `extras_require` in `setup.py`. 

Users running tests locally won't see any difference running `pytest`, but they will now have the option to run it in parallel:

```bash
pytest -n 4 --dist loadfile
```

Note the use of `--dist loadfile`:

> Tests are grouped by their containing file. Groups are distributed to available workers as whole units. This guarantees that all tests in a file run in the same worker.

Also a few other changes to test suite:
- Add a `retry` decorator - this enables the functions in `hyperspy/misc/example_signals_loading.py` to be attempted a few times to avoid possible file locks when different test files try to access the same data. Also added the decorator to `test_empad.py` to avoid potential errors when removing files.
- Add `pragma: no cover` to a few `ImportError` branches
- Add tests for `tv_denoise.py`
- Split `test_hologram_image.py` for cleaner parallel testing
- Fix sklearn import in 2D peak-finding
- Replace `@jit(nopython=True, ...)` with `@njit(...)` everywhere, and add `pragma: no cover` since coverage can't measure jitted functions

### Progress of the PR
- [x] ~update docstring (if appropriate),~
- [x] add tests,
- [x] ready for review.

Improvements running locally:
```bash
# Without xdist:
pytest --mpl hyperspy/
# Time: 522.00s

# With xdist (2 cores)
pytest --mpl --dist loadfile -n 2 hyperspy/
# Time: 358.01s

# With xdist (4 cores)
pytest --mpl --dist loadfile -n 4 hyperspy/
# Time: 203.58s
```
There seems to be a decent speed-up on Travis and Azure Pipelines too, when compared against the latest Rnm/Rnp builds. It shaves at least 2 minutes off the Travis builds, and a similar amount off Azure Pipelines.


tjof2:
- Fixed the float bug on Windows in one of the tests added here.
- Will see if the new `retry` decorator helps to avoid file locks between different test files on Windows and MacOS causing failures.

thomasaarholt:
My only complaint is that I now do not have time to go and get a coffee when running the tests locally.

This is great! Runs nicely on my Windows pc. Number of tests passed is the same with and without `--dist loadfile -n 2`.

.travis.yml
```diff
@@ -8,7 +8,7 @@ env:
     - ENV=pip
     - PIP_ARGS="--upgrade --use-feature=2020-resolver -e"
     - PIP_SELECTOR="all, tests, coverage"
-    - PYTEST_ARGS="--pyargs hyperspy"
+    - PYTEST_ARGS="--pyargs hyperspy -n 2 --dist loadfile"
```
ericpre: Use `-n auto` instead?

tjof2:
I didn't try too hard to see if the other `--dist` options work any better - this was already a significant improvement and file-grouping seems logical given how many test files HyperSpy has already.

I suspect that as new tests are added or removed, more file lock errors may appear when multiple files in different workers try to access the same file at the same time. For now it seems like this PR catches them all. That said, the `retry` decorator adds a simple way to overcome those issues if and when they arise.

hyperspy/decorators.py
```diff
@@ -106,3 +111,55 @@ def wrapper(self, *args, **kwargs):
         else:
             cm(self, *args, **kwargs)
     return wrapper
+
+
+def retry(exceptions=Exception, max_attempts=3, delay=0):
```
ericpre: Wondering if the alternative of fixing the tests trying to access file at different time would more simple? It would save us on maintaining one more thing?!
In principle, there is no reason to use these files in particular, we can generate synthetic data.

thomasaarholt:
Is there a good reason to pass `-n 2` instead of `-n auto`? 
Just in case the workers get access to more cores in the future.

ericpre (commented):
This seems to be working fine, I added a couple of comments but it seems that this have been tried before and the conclusion was that it doesn't play well with lazy test: see https://github.com/hyperspy/hyperspy/pull/1413 and in particular https://github.com/francisco-dlp/hyperspy/pull/33.

ericpre:
How does it work for test running multi-threaded function - the one using `map(parallel=True)` typically?

tjof2:
@thomasaarholt Using `-n 2` instead of auto is because I recall some discussion on the internet about e.g. `os.cpu_count()` not being reliable on Travis. Thus it's just to avoid oversubscribing the VMs.

Also, as @ericpre raises, some of the tests include parallel map() calls. In general there probably is some contention between pytest and these but I don't think it's extreme as it's only in a few files relative to the size of the test suite overall. I've not investigated it in details but it clearly doesn't make the tests slower than before!

>  it seems that this have been tried before and the conclusion was that it doesn't play well with lazy test

I think the key is using `--dist loadfile` which is not the default behaviour. Your links didn't use that flag.

hyperspy/decorators.py
tjof2: Checking for a file lock and waiting would be an equivalent amount of work no? I don't mind either way, I just don't know how to implement any alternatives in a simple way.

.travis.yml
tjof2: See below

hyperspy/decorators.py
ericpre: I am thinking of fixing these tests by not reading these external files (EDS or hologram dataset), assuming that these issues arise from the `example_signals` dataset?

tjof2: Yes that's right @ericpre

tjof2: Though `test_empad.py` currently has a similar fix, the issue is actually different and it can be resolved in a different way.

ericpre: Or use https://github.com/box/flaky?

tjof2: Trying to avoid additional dependencies :-)

ericpre: Well, for testing, it should be fine, as long as this is documented...

tjof2: OK. I'll think this over and fix tomorrow morning

ericpre:
Since the results are a bit less readable than before (can't see easily from which test module fails), it may be worth trying to see if https://github.com/tonybaloney/pytest-azurepipelines makes the pytest outcome easier to read on azure pipeline.

tjof2:
It currently always prints a summary of failed/skipped/passed tests at the end though?

FWIW the alternative, pytest-sugar, looks terrible on Travis and I've not tried it on Azure. It works really nicely locally though

ericpre:
> It currently always prints a summary of failed/skipped/passed tests at the end though?
> 
> FWIW the alternative, pytest-sugar, looks terrible on Travis and I've not tried it on Azure. It works really nicely locally though

Yes, the summary is the same, but there are too many failure, it is a bit more difficult to read, while displaying on a new line for each test module provides an overview of the failure and help to figure out what is going on.

Indeed pytest-sugar is nice when using pytest-xdist.

tjof2:
Alright, will take a look tomorrow at all the comments, thanks all!

tjof2:
> Is there a good reason to pass `-n 2` instead of `-n auto`?

@thomasaarholt  and @ericpre  - see https://github.com/pytest-dev/pytest-xdist/issues/9. I'll leave it as 2.

I've added back `pytest-sugar` based on @ericpre's comment above about the output. 

It's not _perfect_ on Travis, it looks like this: 

a

That said, it doesn't seem to have any effect on Azure:

```
============================= test session starts ==============================
platform darwin -- Python 3.8.5, pytest-6.0.1, py-1.9.0, pluggy-0.13.1
Matplotlib: 3.1.3
Freetype: 2.9.1
rootdir: /Users/runner/work/1/s, configfile: setup.cfg
plugins: mpl-0.11, sugar-0.9.3, xdist-2.1.0, forked-1.3.0
gw0 I / gw1 I
gw0 [3826] / gw1 [3826]

........................................................................ [  1%]
........................................................................ [  3%]
........................................................................ [  5%]
```

francisco-dlp:
Isn't  pytest-dev/pytest-xdist#9 solved?

tjof2:
> Isn't pytest-dev/pytest-xdist#9 solved?

@francisco-dlp  Well it looks like the hard-coded stuff is still there. I can try a build with `-n auto` if preferred. See https://github.com/pytest-dev/pytest-xdist/blob/46cf204c64d7ec4f67b9a781baf272ea62530d08/src/xdist/plugin.py#L8-L35

tjof2:
Since `pytest-sugar` doesn't seem to work nicely on Travis/Azure, what about using `pytest-instafail` there instead? Yes it can produce a lot of output if there are a lot of failures, but at least the user won't have to wait until the end of the test suite to see what failed? Seems to work OK with `pytest-xdist` still.

With regards to the occasionally failing tests due to file locks, what about [`pytest-rerunfailures`](https://github.com/pytest-dev/pytest-rerunfailures)? That seems to be a good option that falls under the pytest-dev organisation on Github. Then it's just a case on Travis/Azure of adding the command line option `pytest --reruns 3` or something.

`pytest-xdist` apparently has a `--looponfail` option but that doesn't seem to actually do what we want.

> How does it work for test running multi-threaded function - the one using `map(parallel=True)` typically?

One option would be to add a `max_workers` argument to `hs.preferences.General` to go alongside the `parallel` argument. Then it can be set to a small number (e.g. 2) in `conftest.py` for the test suite, such that any conflict with the parallel tests is kept to a bare minimum (though not necessarily avoided completely).

Unscientific, but comparing against a recent Rnp CI build (where the runtimes can be quite variable it seems):

| CI Job                 |   Rnp [s] |   New [s] | Improvement   |
|:-----------------------|----------:|----------:|:--------------|
| Azure_Linux_Python38   |       653 |       384 | 1.70x         |
| Azure_Linux_Python37   |       724 |       376 | 1.93x         |
| Azure_Linux_Python36   |       662 |       400 | 1.66x         |
| Azure_MacOS_Python38   |       766 |       489 | 1.57x         |
| Azure_MacOS_Python36   |       879 |       591 | 1.49x         |
| Azure_Windows_Python38 |      1270 |       419 | 3.03x         |
| Azure_Windows_Python36 |      1232 |       417 | 2.95x         |
| Travis_Linux38         |       626 |       558 | 1.12x         |
| Travis_Linux37         |       970 |       776 | 1.25x         |
| Travis_Linux36         |       742 |       671 | 1.11x         |
| Travis_Linux37min      |       666 |       529 | 1.26x         |
| Travis_Linux36old      |       660 |       561 | 1.18x         |

ericpre:
Regarding failure caused by file lock, `pytest-rerunfailures` with a marker for each specific test loading dataset sounds a good solution to me! Considering the likelihood of this happening is not very high, it should not slow down the test suite significantly.

Even if the speed improvement on travis and azure pipeline are not much, one big advantage of this PR is that running the test suite locally with more core is much faster!

hyperspy/conftest.py
```diff
@@ -36,10 +36,10 @@
 hs.preferences.Plot.cmap_navigator = 'viridis'
 hs.preferences.Plot.cmap_signal = 'viridis'
 
-def pytest_configure(config):
-    config.addinivalue_line(
-        "markers", "parallel: a test that is itself parallel and should be run serially."
-    )
+# Set parallel to False by default, so only
+# those tests with parallel=True are run in parallel
+hs.preferences.General.parallel = False
```
ericpre, line 41: I am not clear at all with what the `parallel` marker does and I am wondering if this is still necessary?

tjof2:
Personally, I think you could avoid the markers altogether and just use the command-line argument with a max rerun of e.g. 3x? That way maintainers don't need to remember to include the marker. The tests only seem flaky on the CI anyway, I never encountered any of the errors locally after extensive runs.

That said, if the consensus is to use the marker, happy to include it instead.

Any views on `pytest-instafail`?

ericpre (commented):
Should `-n auto  --dist loadfile` be added to `[tool:pytest]/addopts` in `setup.cfg`? It would save us writing the parameter when running locally while in CI, it can be overwritten if necessary.

hyperspy/conftest.py
tjof2: This isn't a parallel marker - this sets the default behaviour of `.map(parallel=None)` and any other functions that use `.map()`. Setting it to `False` explicitly here means that any tests that might use parallel threads won't unless the test explicitly asks for it (`parallel=True`) to minimize conflicts with the parallelized test suite.

The parallel marker I think you're referring to has been removed entirely.

tjof2:
> Should -n auto --dist loadfile be added to [tool:pytest]/addopts in setup.cfg? It would save us writing the parameter when running locally while in CI, it can be overwritten if necessary.

I can do this! Assuming `-n auto` doesn't conflict with Travis :D Let's test it with the next commit here.

ericpre:
> Personally, I think you could avoid the markers altogether and just use the command-line argument with a max rerun of e.g. 3x? That way maintainers don't need to remember to include the marker. The tests only seem flaky on the CI anyway, I never encountered any of the errors locally after extensive runs.
> 
> That said, if the consensus is to use the marker, happy to include it instead.

Yes, I am fine with this approach too.
 
> Any views on `pytest-instafail`?

Personally, I like to have the whole test suite run, because it can help to figure out what cause of the failures.

tjof2:
> Personally, I like to have the whole test suite run, because it can help to figure out what cause of the failures.

It will run to completion, the name is slightly misleading. What it does is print the failure immediately rather than waiting until the end. It's similar to `pytest-sugar` in this regard, only it doesn't do all the other pretty-printing.

> pytest-instafail is a plugin for pytest that shows failures and errors instantly instead of waiting until the end of test session.

Adding `--maxfail=1` would make the test suite bail out immediately.

tjof2:
@ericpre 
Looks like using `-n auto` on Travis actually just uses one process - see https://travis-ci.org/github/hyperspy/hyperspy/jobs/723477078. Meanwhile, Azure will use 2 processes on Linux, 2 on Windows, and 4 on MacOS. That might explain why the TravisCI speed-up in the table above is less compared to Azure (if Travis was oversubscribed).

**tl;dr** It looks like using `-n auto` leads to no change in Travis run times, but the Azure times are significantly improved (as in the table above).

That said, using `-n auto` on my new workstation runs the full test suite in under 2 minutes so that's a small win!

.travis.yml
```diff
@@ -8,7 +8,7 @@ env:
     - ENV=pip
     - PIP_ARGS="--upgrade --use-feature=2020-resolver -e"
     - PIP_SELECTOR="all, tests, coverage"
-    - PYTEST_ARGS="--pyargs hyperspy"
+    - PYTEST_ARGS="--pyargs hyperspy --reruns 3"
```
ericpre: Add `-n 2` to overwrite the setting set in `setup.cfg`?

tjof2: Let's try it and see!

tjof2: Looks promising https://travis-ci.org/github/hyperspy/hyperspy/jobs/723489054

tjof2: Good suggestion!

tjof2:
I think this is nearly there, except for using `pytest-instafail` to output test failures immediately in the logs rather than when the test suite ends.

hyperspy/conftest.py
ericpre: Yes, the marker have been removed and this is why I am wondering whether it was really necessary or not?

ericpre:
> I think this is nearly there, except for using `pytest-instafail` to output test failures immediately in the logs rather than when the test suite ends.

I tried to find to find example on internet to an idea on how it looks but I haven't been successful so far, do you have one at hand?

Agreed this is in good shape, maybe update the dev guide to reflect the last change on test suite running in parallel by default versus setting it manually?

tjof2:
> I tried to find to find example on internet to an idea on how it looks but I haven't been successful so far, do you have one at hand?

Sure - I artificially added an `assert False` into a test halfway through the suite:

```python
============================= test session starts ==============================
platform linux -- Python 3.8.3, pytest-6.0.1, py-1.9.0, pluggy-0.13.1
Matplotlib: 3.2.2
Freetype: 2.10.2
plugins: cov-2.10.1, mpl-0.11, rerunfailures-9.1, forked-1.3.0, flake8-1.0.6, parallel-0.1.0, instafail-0.4.2, xdist-2.1.0
gw0 [3826] / gw1 [3826] / gw2 [3826] / gw3 [3826] / gw4 [3826] / gw5 [3826]
........................................................................ [  1%]
........................................................................ [  3%]
........................................................................ [  5%]
........................................................................ [  7%]
........................................................................ [  9%]
............s........................................................... [ 11%]
........................................................................ [ 13%]
........................................................................ [ 15%]
.................................................x...................... [ 16%]
......................................................................... [ 18%]
........................................................................ [ 20%]
........................................................................ [ 22%]
........................................................................ [ 24%]
........................................................................ [ 26%]
...................x.................sssss.............................. [ 28%]
........................................................................ [ 30%]
........................................................................ [ 31%]
........................................................................ [ 33%]
........................................................................ [ 35%]
........................................................................ [ 37%]
........................................................................ [ 39%]
........................................................................ [ 41%]
........................................................................ [ 43%]
........................................................................ [ 45%]
........................................................................ [ 47%]
........................................................................ [ 48%]
.....................................F
_______________________________ test_get_methods _______________________________
[gw2] linux -- Python 3.8.3 /home/tom/anaconda3/bin/python

    def test_get_methods():
        rng = np.random.RandomState(123)
        signal = signals.Signal1D(rng.random_sample(size=(11, 5, 7)))
        signal.decomposition()
        signal.cluster_analysis("signal", n_clusters=2)
        signal.unfold()
        cl = signal.get_cluster_labels(merged=True)
        np.testing.assert_array_equal(
            cl.data,
            (signal.learning_results.cluster_labels * np.arange(2)[:, np.newaxis]).sum(0),
        )
    
        cl = signal.get_cluster_labels(merged=False)
        np.testing.assert_array_equal(cl.data, signal.learning_results.cluster_labels)
    
        cl = signal.get_cluster_signals(signal="sum")
        np.testing.assert_array_equal(cl.data, signal.learning_results.cluster_sum_signals)
    
        cl = signal.get_cluster_signals(signal="centroid")
        np.testing.assert_array_equal(
            cl.data, signal.learning_results.cluster_centroid_signals
        )
        cl = signal.get_cluster_signals(signal="mean")
        np.testing.assert_array_equal(
            cl.data,
            signal.learning_results.cluster_sum_signals
            / signal.learning_results.cluster_labels.sum(1, keepdims=True),
        )
    
        cl = signal.get_cluster_distances()
        np.testing.assert_array_equal(cl.data, signal.learning_results.cluster_distances)
>       assert False
E       assert False

hyperspy/tests/learn/test_cluster.py:457: AssertionError
----------------------------- Captured stdout call -----------------------------
Decomposition info:
  normalize_poissonian_noise=False
  algorithm=SVD
  output_dimension=None
  centre=None
........................................................................ [ 51%]
........................................................................ [ 51%]
........................................................................ [ 53%]
........................................................................ [ 55%]
........................................................................ [ 57%]
<<< Test suite continues >>>
```

ericpre:
Thanks, shall we try and if we don't find it convenient, it would not be difficult to remove it!

tjof2:
Updated dev guide and added `pytest-instafail`.

doc/dev_guide/testing.rst
```diff
@@ -59,32 +59,71 @@ First ensure pytest and its plugins are installed by:
...
+   # To run on all the cores of your machine
+   $ pytest -n auto --dist loadfile
+
+   # To run on 2 cores
+   $ pytest -n 2 --dist loadfile
+
+The ``--dist loadfile`` argument will group tests by their containing file. The 
+groups are then distributed to available workers as whole units, thus guaranteeing 
+that all tests in a file run in the same worker. 
+
```
ericpre, line 104: Advice to install `pytest-sugar` to make the results to the test suite more readable when using `pytest-xdist`?

It is not convenient to add to the `tests` extra requirement because we use it to install on travis...

tjof2: Sure!

PR #2526, opened 2020-09-01
a

jat255:
### Requirements

### Description of the change
- Change A.
- Change B.
- etc.

### Progress of the PR
- [ ] add tests,
- [ ] ready for review.

```python
>>> import hyperspy.api as hs
>>> import numpy as np
>>> # Your new feature...
```


jat255:
I have no idea what happened here...

PR #2527, opened 2020-09-01
Update AUR information in install instructions

jat255:
Related to #1069, added some additional text to point to the package that will be consistently up to date with Rnm (since it installs from git).

### DIFF
--- a/doc/user_guide/install.rst
+++ b/doc/user_guide/install.rst
@@ -183,10 +183,19 @@ distributions and the installation via :ref:`Anaconda/Miniconda <anaconda-instal
 or :ref:`Pip <install-with-pip>` is recommended.
 
 However, packages of the latest HyperSpy release and the related
-*hyperpy-gui-ipywidgets/traitsui* packages are
-maintained for the rolling release distributions **Arch-Linux** (`Arch User Repository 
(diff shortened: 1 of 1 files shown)


tjof2 (approved):

ericpre:
Out of curiosity, how often is this package updated?
And is the version number correct? It should be something along the line of `1.7.dev0+gitX` where X is the number of commits since the last tag (or something similar).

jat255:
@ericpre you can see the details of the `PKGBUILD` (which defines how the package gets built on your local machine) [here](https://aur.archlinux.org/cgit/aur.git/tree/PKGBUILD?h=python-hyperspy-git).

The cool thing about the line `source=("git+https://github.com/hyperspy/hyperspy.git#branch=RELEASE_next_minor")` is that every time someone downloads/rebuilds the package, it updates to whatever the most recent commit to Rnm was. This "rebuilding" process is manually triggered, but many Arch users use something called an "AUR helper" that will go and check if there have been updates and notify you if the package needs to be rebuilt.

jat255:
Sorry, should add that at the time I built the package, it set the version to `pkgver=1.7.0.dev0.742a95fec`, pulling from `Release.py` plus the latest commit hash

ericpre:
Thanks for the explanation!

PR #2528, opened 2020-09-02
Replace adhoc code with sphinx-toggleprompt

francisco-dlp:
### Description of the change

Currently the code to remove the python prompt from the code relies on a couple of javascript files, [one of which](https://github.com/hyperspy/hyperspy/blob/RELEASE_next_minor/doc/_static/jquery.js), has just triggered a security alert from GitHub. This PR replaces that code with the sphinx extension sphinx-toggleprompt

I will also update the jquery.js file from all the project's documentation hosted in GH page, maybe breaking the feature to hide the prompts.

### Progress of the PR
- [x] ready for review.


francisco-dlp:
See https://github.com/hyperspy/hyperspy-doc/commit/fdec40ba65e51ca4ba1115aa587a95a1083295d8. After the upgrade, hiding the prompts still works.

ericpre (approved):

ericpre:
> See [hyperspy/hyperspy-doc@fdec40b](https://github.com/hyperspy/hyperspy-doc/commit/fdec40ba65e51ca4ba1115aa587a95a1083295d8). After the upgrade, hiding the prompts still works.

It was working fine on readthedoc (and travis wasn't complaining more than usual!), so it shouldn't be any different for the builds in the `hyperpsy/hyperspy-doc` repository or am I missing something?

francisco-dlp:
The problem is that the ``jquery.js`` file that we were shipping suffers from a vulnerability, although it was working fine. The builds in RtD should get updated tonight once this gets into Rnm. The stable version will require a new patch release—one more reason to release 1.6.1 asap.

ericpre:
I see, the doc build at https://hyperspy.org/documentation.html are fine but not the one on https://hyperspy.readthedocs.io/en/stable until there is a new release.

PR #2529, opened 2020-09-02
Fix for when .emi file has no AcquireDate

### LINKED ISSUE #2504: FEI Reader cannot handle .ser/.emi without AcquireDate in metadata
Line 482 of [`fei.py`](https://github.com/hyperspy/hyperspy/blob/RELEASE_next_minor/hyperspy/io_plugins/fei.py#L482) contains a reference to `ObjectInfo.AcquireDate` in the metadata structure of an .emi file (contained in the `objects` parameter). On one of our systems, when a user does a quantification on a spectrum and saves that as a unique file, TIA does not write an `AcquireDate`. This causes the file to fail loading, even though the .emi file is available.

jat255:
Resolves #2504 

As discussed in the issue, sometimes there is no `AcquireDate` defined in an `.emi` file, which causes HyperSpy to bomb out on loading that file, rather than handling gracefully. This PR allows for a graceful failure with a warning instead of an `AttributeError`.


jat255:
I also added a new test file that meets this condition to ensure the new code is covered.

PR #482, opened 2015-03-06
New det efficiency

pburdet:
Compute the detector efficiency from the layers. The efficiency is calculated by estimating the absorption of the different the layers in front of the detector.

Merge into #481 and #471

``` python

        >>> s = signals.EDSTEMSpectrum(np.ones(1024))
        >>> s.axes_manager.signal_axes[0].scale = 0.01
        >>> s.axes_manager.signal_axes[0].units = "keV"
        >>> s.detetector_efficiency_from_layers()
        <EDSTEMSpectrum, title: Detection efficiency, dimensions: (|1024)>

```
## TO DO
- [ ] doc

PR #2535, opened 2020-09-04
Detector efficiency (replaces #482)

tjof2:
### Description of the change
Replaces #482 and brings it up-to-date with Rnm. Ideally someone else should take this over, not my area of expertise. Looks like @sem-geologist  and @ZanettaPM commented on #482 and may be able to contribute - and #1835 is seemingly related.

> Compute the detector efficiency from the layers. The efficiency is calculated by estimating the absorption of the different the layers in front of the detector.

```python
        >>> s = signals.EDSTEMSpectrum(np.ones(1024))
        >>> s.axes_manager.signal_axes[0].scale = 0.01
        >>> s.axes_manager.signal_axes[0].units = "keV"
        >>> s.detector_efficiency_from_layers()
        <EDSTEMSpectrum, title: Detection efficiency, dimensions: (|1024)>
```

### Progress of the PR
- [x] add tests,
- [ ] ready for review.


pquinn-dls:
If interested in supporting XRF  signal it's not general enough. 
Ge detectors are still popular but this method only uses Si as the detector material.
In terms of units cm is standard in x-ray applications.  
Doesn't seem to deal with compounds - e.g. an air path.
I have code to do this so will suggest something. Many x-ray applications use xraylib to handle elemental properties and x-ray properties related to EDX/XRD.  Would there be support for moving across to this ?

ericpre:
I am not knowledgeable on the details of this topic and I will not be able to comment sensibly on this.
In term of distribution, xraylib should be added as an optional dependency, because xraylib is a C library and it isn't available on pypi. However, since it is packaged on conda-forge, it should be easy to install in an anaconda/miniconda distribution.

Issue #1828, opened 2018-02-08
Support Pathlib


thomasaarholt:
Python 3.4 introduced pathlib, a new way of handling paths in Python. [A really nice explanation is here.](https://medium.com/@ageitgey/python-3-quick-tip-the-easy-way-to-deal-with-file-paths-on-windows-mac-and-linux-11a072b58d5f)

Essentially, it unifies handling paths on windows, linux and mac, and makes life a bit easier.

It would be nice if I could feed a pathlib.Path to hs.load() directly. str(Path) returns the correctly formatted string.

francisco-dlp:
Supporting ``Path`` in ``hs.load`` sounds good and implementing it should be straightforward. Using it internally should lead to more readable code so :+1: from me.

tjof2:
Looks like this can be done in two stages - #2537 introduces it to most places in the core code (and supports it in `hs.load()` as originally requested here), but leaves the `io_plugins` untouched. They would need more work to support `pathlib.Path`, perhaps better left for the split into a separate package in #2174

tjof2:
#2537 is done, so it's only the io plugins that don't support it.

PR #2537, opened 2020-09-08
Initial support for pathlib

tjof2:
### Description of the change 
First pass at #1828 (supporting pathlib in HyperSpy). 

This supports `pathlib.Path` objects everywhere **except** `hyperspy/io_plugins/*`, because that would be considerably more work given the frequent uses of `os.path.splitext` and others. Also, if the `io_plugins` are going to be split in future, it's perhaps worth leaving changing them for now. Any `os.path` instances in the `hyperspy/tests/*` folders are also untouched here.

Instead, in this PR, `hs.load()` will accept `pathlib.Path` objects, including globs, but converts them to strings internally before passing them to the various file readers to maintain backwards compatibility.

A few other locations in the core code that use `os.path` have been changed to use pathlib instead.

### Progress of the PR
- [x] add tests,
- [x] ready for review.

```python
>>> import hyperspy.api as hs
>>> from pathlib import Path

>>> # Use pathlib.Path
>>> p = Path("/path/to/a/file.hspy")
>>> s = hs.load(p)

>>> # Use pathlib.Path.glob
>>> p = Path("/path/to/some/files/").glob("*.hspy")
>>> s = hs.load(p)
```

### DIFF
--- a/doc/user_guide/io.rst
+++ b/doc/user_guide/io.rst
@@ -28,10 +28,10 @@ If the loaded file contains several datasets, the :py:func:`~.io.load`
 functions will return a list of the corresponding signal.
 
 .. NOTE::
-    Note for python programmers: the data is stored in a numpy array
-    in the :py:attr:`~.signal.BaseSignal.data` attribute, but you will not
-    normally need to access it there.)
 
+    Note for Python programmers: the data is stored in a numpy array
+    in the :py:attr:`~.signal.BaseSignal.data` attribute, but you will not
+    normally need to access it there.
 
 HyperSpy will try to guess the most likely data type for the corresponding
 file. However, you can force it to read the data as a particular data type by
@@ -40,7 +40,7 @@ providing the ``signal`` keyword, which has to be one of: ``spectrum``,
 
 .. code-block:: python
 
-    >>> s = hs.load("filename", signal = "EELS")
+    >>> s = hs.load("filename", signal_type="EELS")
 
 Some file formats store some extra information about the data, which can be
 stored in "attributes". If HyperSpy manages to read some extra information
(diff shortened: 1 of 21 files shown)


tjof2:
Windows failures are an accidental removal of the `os` module, will fix.

doc/user_guide/io.rst
```diff
@@ -40,7 +40,7 @@ providing the ``signal`` keyword, which has to be one of: ``spectrum``,
 
 .. code-block:: python
 
-    >>> s = hs.load("filename", signal = "EELS")
+    >>> s = hs.load("filename", signal="EELS")
```
ericpre: This is not related to this PR, but it should be `signal_type` instead of `signal`, while, you are at it, may be worth fixing this too!

hyperspy/misc/config_dir.py
```diff
@@ -17,31 +17,28 @@
...
-if os.name == 'posix':
-    config_path = os.path.join(os.path.expanduser('~'), '.hyperspy')
-    os_name = 'posix'
-elif os.name in ['nt', 'dos']:
-    config_path = os.path.expanduser('~/.hyperspy')
-    os_name = 'windows'
+if os.name == "posix":
+    os_name = "posix"
+elif os.name in ["nt", "dos"]:
+    os_name = "windows"
```
ericpre: I would clean up this bit further (python2 left over?), I don't see why, we should define `os_name` in hyperspy. This is used only in place in hyperspy and we should use directly `os.name`.

tjof2:
Ready for review, codecov fixed.

ericpre (commented):
Looks good, two small suggestions!

hyperspy/misc/config_dir.py
tjof2: Done

doc/user_guide/io.rst
tjof2: Done

tjof2:
Took the opportunity to apply `pathlib.Path` to the few non-io-plugin tests. Not sure about codecov, I think it's fine to ignore.

PR #2539, opened 2020-09-08
Add warning on Windows and test to guard against #2320

### LINKED ISSUE #2320: estimate_peak_width(parallel=True) giving rise to nans

tjof2:
### Description of the change
Closes #2320.

There looks to be a bug in scipy where `scipy.interpolate.UnivariateSpline` is not thread-safe on Windows, so running `s.estimate_peak_width(parallel=True)` (which uses thread-based parallelism) will return garbage, as @thomasaarholt  points out in the issue.

Short of the issue being fixed in scipy, this logs a warning on Windows when `parallel=True` or `parallel=None`, and explicitly sets `parallel=False`.

### Progress of the PR
- [x] add tests,
- [x] ready for review.

### DIFF
--- a/hyperspy/_signals/signal1d.py
+++ b/hyperspy/_signals/signal1d.py
@@ -1467,9 +1467,6 @@ class Signal1D(BaseSignal, CommonSignal1D):
         background must be previously substracted.
         The estimation is performed by interpolation using cubic splines.
 
-        Warning: `parallel=True` is not currently supported on
-        Windows platforms.
-
         Parameters
         ----------
         factor : 0 < float < 1
@@ -1494,6 +1491,11 @@ class Signal1D(BaseSignal, CommonSignal1D):
         width or [width, left, right], depending on the value of
         `return_interval`.
 
+        Notes
+        -----
+        Parallel operation of this function is not supported
+        on Windows platforms.
+
         """
         if show_progressbar is None:
             show_progressbar = preferences.General.show_progressbar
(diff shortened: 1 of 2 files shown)


hyperspy/_signals/signal1d.py
```diff
@@ -1467,6 +1467,9 @@ def estimate_peak_width(
         background must be previously substracted.
         The estimation is performed by interpolation using cubic splines.
 
+        Warning: `parallel=True` is not currently supported on
```
ericpre: Use the `Raises` section of numpydoc instead and update/append the `PARALLEL_ARG` string mentioning the windows issue?

hyperspy/_signals/signal1d.py
```diff
@@ -1490,14 +1493,23 @@ def estimate_peak_width(
...
+        """
         if show_progressbar is None:
             show_progressbar = preferences.General.show_progressbar
         self._check_signal_dimension_equals_one()
         if not 0 < factor < 1:
             raise ValueError("factor must be between 0 and 1.")
 
+        from hyperspy.misc.config_dir import os_name
+
+        if parallel and os_name == "windows":  # pragma: no cover
```
ericpre: in case `parallel=None`, it will pick up the value of `preferences.General.parallel` and in this case, it would be better to set parallel=False.

tjof2: Good idea, will edit

ericpre (approved):

Issue #2495, opened 2020-08-24
pyUSID FutureWarnings and DeprecationWarnings


tjof2:
From the current tests:

```python
hyperspy/tests/io/test_usid.py: 16 warnings
DeprecationWarning: sidpy.Translator will soon be replaced by sidpy.Reader. Considerrestructuring code from Translator to Reader

hyperspy/tests/io/test_usid.py: 21 warnings
FutureWarning: pyUSID.io.dtype_utils.contains_integers has been moved to sidpy.base.num_utils.contains_integers. This copy in pyUSID willbe removed in future release. Please update your import statements

hyperspy/tests/io/test_usid.py: 17 warnings
FutureWarning: pyUSID.io.dtype_utils.validate_single_string_arg has been moved to sidpy.base.string_utils.validate_single_string_arg. This copy in pyUSID willbe removed in future release. Please update your import statements

hyperspy/tests/io/test_usid.py: 17 warnings
FutureWarning: pyUSID.io.dtype_utils.validate_string_args has been moved to sidpy.base.string_utils.validate_string_args. This copy in pyUSID willbe removed in future release. Please update your import statements

hyperspy/tests/io/test_usid.py: 17 warnings
UserWarning: pyUSID.io.hdf_utils.simple.write_ind_val_dsets no longer createsregion references for each dimension. Please use pyUSID.io.reg_ref.write_region_references to manually create region references

hyperspy/tests/io/test_usid.py: 19 warnings
FutureWarning: pyUSID.io.dtype_utils.lazy_load_array has been moved to sidpy.hdf.hdf_utils.lazy_load_array. This copy in pyUSID willbe removed in future release. Please update your import statements

hyperspy/tests/io/test_usid.py::TestUSID2HSbase::test_n_pos_0_spec
hyperspy/tests/io/test_usid.py::TestUSID2HSbase::test_0_pos_n_spec
hyperspy/tests/io/test_usid.py::TestUSID2HSbase::test_n_pos_m_spec
hyperspy/tests/io/test_usid.py::TestUSID2HSbase::test_lazy_load
hyperspy/tests/io/test_usid.py::TestUSID2HSdtype::test_complex
hyperspy/tests/io/test_usid.py::TestUSID2HSdtype::test_compound
hyperspy/tests/io/test_usid.py::TestUSID2HSdtype::test_non_linear_dimension
hyperspy/tests/io/test_usid.py::TestUSID2HSmultiDsets::test_pick_specific
[...]

PR #2543, opened 2020-09-09
Fix future and deprecation warning pyusid.

ericpre:
Fix pyUSID FutureWarnings and DeprecationWarnings #2495.


tjof2 (approved):

ericpre:
Not all warning are gone, but some of them should go away with https://github.com/pycroscopy/pyUSID/pull/54.

tjof2:
The main issue was the sidpy ones so that's fine I think

PR #2544, opened 2020-09-09
Add download badges and fix azure pipeline badge

ericpre:
I have just came across https://pepy.tech/project/hyperspy and interestingly, there is a non-negligible number of download of old version of hyperspy 1.5.1 and 1.5.2...

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -9,7 +9,7 @@
 |gitter|_ |DOI|_
 
 .. |Azure| image:: https://dev.azure.com/franciscode-la-pena-manchon/hyperspy/_apis/build/status/hyperspy.hyperspy?branchName=RELEASE_next_minor
-.. _Azure: https://dev.azure.com/franciscode-la-pena-manchon/hyperspy/_build/latest?definitionId=1?branchName=RELEASE_next_minor
+.. _Azure: https://dev.azure.com/franciscode-la-pena-manchon/hyperspy/_build?definitionId=1&_a=summary&view=branches
 
 .. |Travis| image:: https://api.travis-ci.org/hyperspy/hyperspy.png?branch=RELEASE_next_minor
 .. _Travis: https://travis-ci.org/hyperspy/hyperspy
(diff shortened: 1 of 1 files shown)


tjof2 (approved):

francisco-dlp:
I find the download badges confusing: the monthly one counts pypi downloads and the total one points to total downloads in conda-forge one...

tjof2:
Good point, I didn't spot that - agreed

ericpre:
What would be less confusing? To remove the total downloads from anaconda cloud?

francisco-dlp:
Actually, since we cannot count downloads reliably (for that we would need to aggregate pypi, conda, bundle, linux distros...), my personal preference it not to highlight this information. New potential users who are trying to judge the size of the hyperspy community before investing their time in learning how to use it can make up their minds by looking at  e.g. GH stars and used by.

That said, as @ericpre pointed out, monitoring pypi downloads can be useful to detect issues with the current version (it is indeed surprising that the 1.5.* remain popular, may it be because 1.6 require newer versions of some libraries?

tjof2:
Also recall that e.g. pyxem hasn't moved to 1.6.0 due to the histogram bug, that's waiting on the 1.6.1 release.

ericpre:
> Actually, since we cannot count downloads reliably (for that we would need to aggregate pypi, conda, bundle, linux distros...), my personal preference it not to highlight this information. New potential users who are trying to judge the size of the hyperspy community before investing their time in learning how to use it can make up their minds by looking at e.g. GH stars and used by.
> 
> That said, as @ericpre pointed out, monitoring pypi downloads can be useful to detect issues with the current version (it is indeed surprising that the 1.5.* remain popular, may it be because 1.6 require newer versions of some libraries?

Agreed, my thought process was similar and the is the reason why I added this link (https://pepy.tech/project/hyperspy) in the `README.rst` is to make it more accessible and "discoverable".
While I was at it, I also added the link to anaconda cloud but got side track and added the download from anaconda cloud too... and indeed, all these download numbers are misleading.

Shall we keep the https://pepy.tech/project/hyperspy badge and remove the anaconda download one?

francisco-dlp:
What about adding it to the devel guide, with a little explanation of the limitations and its usefulness for developers? Everything in README makes it into the pypi frontpage and the zotero DOI page, so I think that it is better to keep unreliable info aside.

PR #2545, opened 2020-09-10
Specify file reader for hs.load

### LINKED ISSUE #1478: hs.load: specifying file reader
Currently HyperSpy supports two different types of HDF5-"standards":
- Internal "HyperSpy" HDF5
- NCEM HDF5 (emd)

In the future HyperSpy will (probably) also support the FEI HDF5 (emd) file format.

Since the emd files are valid HDF5 files, they sometimes have the hdf5-extension. So when trying to load a NCEM EMD/HDF5-file with the .hdf5 extension it fails. This is ok, but it would be nice to be able to specify the file reader you want HyperSpy to use, and pass this to the `io.load_single_file` function.

This works:
```Python
import hyperspy.io
emd_reader = hyperspy.io.io_plugins[9]
s = hyperspy.io.load_with_reader("an_emd_file.hdf5", emd_reader)
```

tjof2:
### Description of the change
Closes #1478.

Introduces a `reader` kwarg to `hs.load()`, so that instead of inferring the file reader automatically from the file extension, the user can explicitly pass one themselves. Can either be a function or a string, for example:

```python
import hyperspy.api as hs
import hyperspy.io
emd_reader = hyperspy.io.io_plugins[9]

s = hs.load("an_emd_file.hdf5", reader=emd_reader)
s = hs.load("an_emd_file.hdf5", reader="emd")
```

### Progress of the PR
- [x] add tests,
- [x] ready for review.

### DIFF
--- a/hyperspy/io.py
+++ b/hyperspy/io.py
@@ -47,6 +47,42 @@ f_error_fmt = (
     "\t\tPath: %s")
 
 
+def _infer_file_reader(extension):
+    """Return a file reader from the plugins list based on the file extension.
+
+    If the extension is not found or understood, returns
+    the Python imaging library as the file reader.
+
+    Parameters
+    ----------
+    extension : str
+        File extension, without initial "." separator
+
+    Returns
+    -------
+    reader : func
+        The inferred file reader.
+
+    """
+    rdrs = [rdr for rdr in io_plugins if extension.lower() in rdr.file_extensions]
+
+    if not rdrs:
+        # Try to load it with the python imaging library
+        _logger.warning(
+            f"Unable to infer file type from extension '{extension}'. "
+            "Will attempt to load the file with the Python imaging library."
+        )
+
+        from hyperspy.io_plugins import image
+
+        reader = image
+    else:
+        # Just take the first match for now
+        reader = rdrs[0]
+
+    return reader
+
+
 def _escape_square_brackets(text):
     """Escapes pairs of square brackets in strings for glob.glob().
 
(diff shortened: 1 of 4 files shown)


hyperspy/io.py
```diff
@@ -47,6 +47,33 @@
...
+    i = 0
+
+    while (
+        extension.lower() not in io_plugins[i].file_extensions
+        and i < len(io_plugins) - 1
+    ):
+        i += 1
+
+    # Fall back to the python imaging library
+    return image if i == len(io_plugins) else io_plugins[i]
```
tjof2: Will update this to add a logger warning.

francisco-dlp:
This is brilliant. Some suggestions:

* It would be good to expose the IO plugins in the API so that something like this works:
  ```python
  hs.load(reader=hs.io.hspy)
  ```
* And we could take it one step further:
  ```python
  hs.load(reader=hs.io.HSPY(compression="lzf"))
  ```

In this way we could solve the current keyword clutter and difficulty of documenting all the options in the different readers.

Once we transition to RosettaSciIO it'll suffice to update the syntax e.g. as follows:

  ```python
  hs.load(reader=rio.HSPY(compression="lzf"))
  ```

tjof2:
Sounds good @francisco-dlp, will look at that proposal! Plus improve the exception handling a little further (see failed tests).

ericpre:
Another thing which would be useful is to check if the reader can the file, when a reader is not specified. In case of `hdf5` file, this should fairly well:
https://github.com/hyperspy/hyperspy/blob/912fc5776c0f8c604cb339332760b3cbd0058386/hyperspy/io_plugins/emd.py#L1624-L1676

And then, we could split the `emd` reader into `emd_velox` from the `emd_ncem`.

tjof2:
As it stands, this PR supports the basic idea of:

```python
import hyperspy.api as hs

# Infer from file extension
hs.load("file.hspy")

# Infer from `reader` str argument
hs.load("file.hspy", reader="hspy")

# Pass a reader explicitly method #1
import hyperspy.io
hspy_reader = hyperspy.io.io_plugins[8]
hs.load("file.hspy", reader=hspy_reader)

# Pass a reader explicitly method #2
from hyperspy.io_plugins import hspy
hs.load("file.hspy", reader=hspy)
```
In the last two options, passing anything with `hasattr(..., "file_reader")` will work. This is in preparation for the proposal @francisco-dlp has of `hs.load(reader=hs.io.HSPY(compression="lzf"))` - provided `HSPY` implements `file_reader`, it will work.

Right now, I think just documenting the above options in this PR is all that needs to be done (since it's pointed at Rnp).

Then a separate issue for @francisco-dlp's ideas, exposing the `io_plugins` to `hs.api`, and probably another issue for @ericpre's suggestion, would be a better way to proceed (tackling them separately).

tjof2:
Python 3.6 tests are failing on Azure because `importlib_resources` is missing.

hyperspy/io.py
```diff
@@ -47,6 +47,42 @@
...
+    reader : func
+        The inferred file reader.
+
+    """
+    rdrs = [rdr for rdr in io_plugins if extension.lower() in rdr.file_extensions]
+
+    if not rdrs:
+        # Try to load it with the python imaging library
+        _logger.warning(
+            f"Unable to infer file type from extension '{extension}'. "
```
ericpre, line 72: This warning will also be used when providing `reader='emd'` for example, which may not be the most appropriate but we can leave it as it is for now and improve it in a separate PR.

hyperspy/io.py
```diff
@@ -363,48 +399,68 @@ def load(filenames=None,
...
-            reader = image
-            return load_with_reader(filename, reader, **kwds)
-        except BaseException as e:
-            raise IOError(
-                "If the file format is supported, "
-                f"please report this error:\n{e}"
-            )
+    # File extension without "." separator
+    file_ext = os.path.splitext(filename)[1][1:]
+    reader = kwds.pop("reader", None)
```
ericpre, line 437: This keyword needs to be added to the docstring of the `load` function. A quick example in this there would be good too.

ericpre (changes requested):
This looks very good and would be useful for cases where a mixture of extensions is used for the same format. As mentioned by @pquinn-dls, this is relevant in case of the nexus file format, where `h5` or `nxs`seems to be used.

This PR simply needs some documentation improvement of the `load` function, see the corresponding comment. Maybe also add an example in the [nexus section](https://hyperspy.readthedocs.io/en/latest/user_guide/io.html#nexus) something along the line of:
```python
s = hs.load('file.h5', reader='nxs')
```

I triggered a azure pipeline re-run and it is happy now.

hyperspy/io.py
tjof2: Ok!

hyperspy/io.py
tjof2: Done!

ericpre (approved):

PR #2549, opened 2020-09-13
Non uniform axis binning #1

jlaehne:
### Description of the change
Handling of axis binning is the main issue in #2398 that needs to be implemented for the `non-uniform-axis` branch to be merged.

In https://github.com/hyperspy/hyperspy/pull/2299#issuecomment-630743193 it was proposed by @francisco-dlp that `is_binned` could become an attribute of the axis replacing `metadata.Signal.binned`. This PR would implement this change and , if we agree on this approach.

There are approx. 90 occurences of `metadata.Signal.binned`, though most of them are in the tests and as such easy to replace.

### Progress of the PR
- [X] Proposal of implementation
- [ ] Implement for all occurences
- [ ] Make non uniform axes binning aware
- [ ] add tests,
- [ ] ready for review.

```python
>>> import hyperspy.api as hs
>>> import numpy as np
>>> # Your new feature...
```


jlaehne:
@francisco-dlp , @ericpre should we go forward in this direction?

If the axes have the property `is_binned`, the bins can be calculated automatically - though for `DataAxis` it might make sense to do this on `_init` and in case of changes and explicitly store the array of bins.

ericpre:
We need to get the test failures sorted first in the `non_uniform_axes` branch, if not we will lost track of where the failure was introduced, etc.

ericpre:
> In [#2299 (comment)](https://github.com/hyperspy/hyperspy/pull/2299#issuecomment-630743193) it was proposed by @francisco-dlp that `is_binned` could become an attribute of the axis replacing `metadata.Signal.binned`. This PR would implement this change and , if we agree on this approach.
> 
> There are approx. 90 occurences of `metadata.Signal.binned`, though most of them are in the tests and as such easy to replace.

@jlaehne: shall this be done in a PR against `RELEASE_next_minor`, instead of the `non_uniform_axes`?

jlaehne:
Well, the full implementation must include all three new axes types - and the main motivation was making non_uniform_axes binning aware. That is  why I started it here. 

But if you think that it would be easier to first move the binning metadata to an axes property in RELEASE_next_minor, fix all affected tests and other occurences, and afterwards implement it for the non-uniform axes types, I could do it that way.

ericpre:
> But if you think that it would be easier to first move the binning metadata to an axes property in RELEASE_next_minor, fix all affected tests and other occurences, and afterwards implement it for the non-uniform axes types, I could do it that way.

I would prefer to do if that way, because the binning issue is independent of the `non_uniform_axes` branch and I think that everything that help keeping the `non_uniform_axes` branch as simple as possible is welcome! Once we have the binning issue fixed in `RELEASE_next_minor`, then we can merge it easily into the `non_uniform_axes` branch.

jlaehne:
Implemented for `RELEASE_next_minor` in  #2652. Once that one is merged, I can continue here.

jlaehne:
Replaced by #2652 and #2693

PR #2555, opened 2020-09-24
Set reading mode hdf5 file

ericpre:
Use `"r"` in favor of `"r+"` when possible because using `"r+"` change the "last modified" timestamp of the file and only reading a file shouldn't it.

### Progress of the PR
- [x] Use `mode="r"` in favor of `"r+"` in `hdf5.File` when possible,
- [x] fix wrong argument in user guide,
- [x] ready for review.


tjof2 (approved):

ptim0626:
Sorry to bring this up now, but I don't quite get why it is better to use '`r+'` for loading lazily, could it all just be `'r'`?

ericpre:
I think you are correct, I can't think of a reason to use `'r+'` when loading lazily.

ptim0626:
@ericpre in such case should we change back all to `r`? if not possible to revert I could do a simple PR on this.

PR #2558, opened 2020-10-07
Add visibility boolean for all markers, only used by text markers

hakonanes:
Signed-off-by: Håkon Wiik Ånes <hwaanes@gmail.com>

### Description of the change
I've added a `visible` parameter to the `MarkerBase.data` array, used by `Text` markers so that the call to `ax.draw_artist()` in `BlittedFigure._draw_animated()` for those markers with `visible=False` aren't drawn:
* Add `visible` parameter to `hyperspy.drawing._markers.text.Text.__init__()`, default to True, passed to `set_data()`. Docstring updated with a brief explanation of when it is useful.
* Add `visible` parameter to `hyperspy.drawing.marker.MarkerBase.set_data()`
* Update methods `Text._plot_marker()` and `Text.update()` to pass on the `visible` parameter to Matplotlib

Why? I'm implementing geometrical simulations of EBSD patterns, i.e. detector positions of Kikuchi bands and zone axes, in kikuchipy, and add these as markers to an EBSD signal as seen in the bottom GIF. Markers are added as iterables over the navigation shape. Not all bands/zone axes appear in all patterns, so setting `LineSegment` (bands) and `Point` (zone axes) coordinates to `np.nan` for some navigation indices ensures they are only plotted where they are present. However, setting `Text` (zone axes labels) coordinates to `np.nan` results in [`matplotlib.text.Text.draw()` writing a log warning to stdout](https://github.com/matplotlib/matplotlib/blob/master/lib/matplotlib/text.py#L677) for each marker... I therefore decided to set the `Text` coordinates of zone axes labels not present in a navigation point to (0, 0) and instead pass `visible=False` to `Text`, which [doesn't draw the marker](https://matplotlib.org/api/text_api.html#matplotlib.text.Annotation.draw), and thus solves my problem.

### Progress of the PR
- [x] Update docstring
- [x] Add to existing test `hyperspy.tests.drawing.test_plot_markers.test_iterate_markers()`.
- [x] Ready for review

Taken from the test `hyperspy.tests.drawing.test_plot_markers.test_iterate_markers()`:

```python
>>> import matplotlib
>>> matplotlib.rcParams["backend"] = "qt5agg"
>>> import numpy as np
>>> import scipy.misc
>>> from skimage.feature import peak_local_max
>>> from hyperspy.signal import BaseSignal
>>> from hyperspy.utils import markers
>>> ims = BaseSignal(scipy.misc.face()).as_signal2D([1, 2])
>>> index = np.array(
...     [peak_local_max(im.data, min_distance=100, num_peaks=4) for im in ims]
... )
# Make text markers "1" ("2") not visible in 0th (1st) image
>>> visible = [[True] * 3, [False, True, True], [True, None, True], [True] * 3]
>>> for i in range(4):
...     xs = index[:, i, 1]
...     ys = index[:, i, 0]
...     m = markers.text(
...         x=10 + xs, y=10 + ys, text=str(i), color='r', visible=visible[i], size=20
...     )
...     ims.add_marker(m, plot_marker=True, permanent=True)
>>> ims.plot()  # Observe that markers "1" ("2") are not visible in 0th (1st) image
>>> plt.show()  # Needed this in PyCharm for some reason today...
```

### Motivational GIF

Peek 07-10-2020 22-25

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -46,6 +46,7 @@ hyperspy/misc/etc/test_compilers.obj
 *nbc
 *nbi
 hyperspy/tests/io/edax_files.zip
+htmlcov/
 
 ### Code ###
 # Visual Studio Code - https://code.visualstudio.com/
--- a/hyperspy/drawing/_markers/text.py
+++ b/hyperspy/drawing/_markers/text.py
@@ -56,11 +56,11 @@ class Text(MarkerBase):
     >>> s.add_marker(m, permanent=True)
     """
 
-    def __init__(self, x, y, text, **kwargs):
+    def __init__(self, x, y, text, visible=None, **kwargs):
         MarkerBase.__init__(self)
         lp = {'color': 'black'}
         self.marker_properties = lp
-        self.set_data(x1=x, y1=y, text=text)
+        self.set_data(x1=x, y1=y, text=text, visible=visible)
         self.set_marker_properties(**kwargs)
         self.name = 'text'
 
@@ -81,8 +81,13 @@ class Text(MarkerBase):
         self.marker.set_position([self.get_data_position('x1'),
                                   self.get_data_position('y1')])
         self.marker.set_text(self.get_data_position('text'))
+        self.marker.set_visible(self.get_data_position('visible'))
 
     def _plot_marker(self):
         self.marker = self.ax.text(
-            self.get_data_position('x1'), self.get_data_position('y1'),
-            self.get_data_position('text'), **self.marker_properties)
+            x=self.get_data_position('x1'),
+            y=self.get_data_position('y1'),
+            s=self.get_data_position('text'),
+            visible=self.get_data_position('visible'),
+            **self.marker_properties,
+        )
--- a/hyperspy/drawing/_markers/text.py
+++ b/hyperspy/drawing/_markers/text.py
@@ -33,6 +33,10 @@ class Text(MarkerBase):
         The position of the text in y. see x arguments
     text : array or str
         The text. see x arguments
+    visible : bool, optional
+        Whether to plot the marker, default is True. This is useful when
+        adding a marker that changes with signal navigation index, but is
+        not visibile in all indices.
     kwargs :
         Keywords argument of axvline valid properties (i.e. recognized by
         mpl.plot).
@@ -56,7 +60,7 @@ class Text(MarkerBase):
     >>> s.add_marker(m, permanent=True)
     """
 
-    def __init__(self, x, y, text, visible=None, **kwargs):
+    def __init__(self, x, y, text, visible=True, **kwargs):
         MarkerBase.__init__(self)
         lp = {'color': 'black'}
         self.marker_properties = lp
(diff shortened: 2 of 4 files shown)


hakonanes:
Looking at the plot artifacts from Azure, it seems like this addition made text markers not visible in the failing tests... I'll look into it.

Edit: Turns out I set the default to `None`, which I ment to set to `True`. Let's see if that fixes it.

hakonanes:
Are there any HyperSpy developers who could consider this PR? I would be very happy to change something if necessary...

ericpre:
Maybe it would more simple to filter this specific logging warning, rather changing the signature of the test marker? As I understand, the behaviour is fine, this is simply the warning being displayed? The advantage of this approach is to keep the use of the text marker consistent with other markers.
@magnunor: is there better way to make marker invisible for specific navigation indices?

hakonanes:
Thanks for the answer, @ericpre!

> Maybe it would more simple to filter this specific logging warning, rather changing the signature of the test marker?

Yes, it would be ideal if I could just set a marker's coordinates to `np.nan` instead of having to set a `visible` parameter. There are a few different places this warning would have to be filtered: upon adding the marker to the signal, when the signal is plotted with the markers, and when the navigation indices change (navigating). If I remember correctly back to when I made this PR, at least these places:
* https://github.com/hyperspy/hyperspy/blob/RELEASE_next_minor/hyperspy/drawing/figure.py#L73
* https://github.com/hyperspy/hyperspy/blob/RELEASE_next_minor/hyperspy/drawing/marker.py#L171
* https://github.com/hyperspy/hyperspy/blob/RELEASE_next_minor/hyperspy/drawing/_markers/text.py#L85

The only workaround I found, and which we've opted for in kikuchipy, is for the user to set the Matplotlib log level to ERROR instead of WARNING before plotting zone axes labels on EBSD patterns. I consider this a temporary "solution".

I would be very happy to hear alternative solutions and try to make them work.

> As I understand, the behaviour is fine, this is simply the warning being displayed? The advantage of this approach is to keep the use of the text marker consistent with other markers.

Yes, apart from the warning being written hundreds/thousands of times to standard out when navigating a signal with text markers with coordinates set to np.nan, everything behaves as desired.

ericpre:
> I would be very happy to hear alternative solutions and try to make them work.

I would imagine that it would be possible to filter this specific warning.
https://docs.python.org/3.8/library/logging.html#logging.Logger.addFilter

But it would be good to hear @magnunor view on this.

ericpre:
Ping @magnunor.

ericpre:
@hakonanes, did you give a try at filtering the warnings as mentioned above?

hakonanes:
No, but I hope to get to it within the next week. Thanks for giving me a little nudge.

hakonanes:
Filtering the warning worked great, thanks @ericpre!

After placing the following in [the relevant kikuchipy file](https://github.com/pyxem/kikuchipy/blob/master/kikuchipy/simulations/geometrical_ebsd_simulation.py)

```python
import logging

class DisableMatplotlibWarningFilter(logging.Filter):
    def filter(self):
        message_to_disable = "posx and posy should be finite values"
        return not self.msg == message_to_disable

logging.getLogger("matplotlib.text").addFilter(DisableMatplotlibWarningFilter)
```

the warning is now not displayed.

PR #2561, opened 2020-10-18
Save image with scalebar

ericpre:
Add option to save image with scalebar. The scalebar is added as a matplotlib artist (using matplotlib-scalebar)

### Progress of the PR
- [x] save image (jpg, png) with scalebar,
- [x] update docstring,
- [x] update user guide,
- [x] add tests,
- [x] ready for review.

```python
import numpy as np
import hyperspy.api as hs

s = hs.signals.Signal2D(np.arange(100*100).reshape((100, 100)))
s.save('test.jpg', scalebar=True, scalebar_kwds={'location':'lower right'})

```


magnunor:
Why add an additional dependency, when the matplotlib already has a scalebar: `mpl_toolkits.axes_grid1.anchored_artists.AnchoredSizeBar`

It updates automatically when using the "Zoom" functionality in the interactive matplotlib viewer. For example:

```python
import numpy as np
from mpl_toolkits.axes_grid1.anchored_artists import AnchoredSizeBar

fig, ax = plt.subplots()
ax.imshow(np.ones((10, 10)))
scalebar = AnchoredSizeBar(ax.transData, 2, "2", 1)
ax.add_artist(scalebar)
fig.show()
```

ericpre:
The matplotlib sizebar doesn't play well with zooming because the scalebar goes out of the figure:

```python
import numpy as np
from mpl_toolkits.axes_grid1.anchored_artists import AnchoredSizeBar

fig, ax = plt.subplots()
ax.imshow(np.ones((100, 100)))
scalebar = AnchoredSizeBar(ax.transData, 20, "20", 1)
ax.add_artist(scalebar)

# change extent to simulate zoom
ax.images[0].set_extent((0, 10, 0, 10))
```
While matplotlib-scalebar adjust the scalebar automatically as it would be expected in this situtation:
```python
import numpy as np
from matplotlib_scalebar.scalebar import ScaleBar

fig, ax = plt.subplots()
ax.imshow(np.ones((100, 100)))
scalebar = ScaleBar(1) # 1 pixel = 1 meter
ax.add_artist(scalebar)

# change extent to simulate zoom
ax.images[0].set_extent((0, 10, 0, 10))
```

Beside this issue, the width of the scalebar is calculated automatically.

Since, this is a pure python library and it is added as optional dependency, I am not concerned about distribution. Longer term, once `matplotlib-scalebar` is packaged for the linux distribution, for which hyperspy is packaged, the scalebar widget could be simplified a lot by using the `matplotlib-scalebar`.

magnunor:
Good point! The size of the `AnchoredSizeBar` is rescaled automatically as a function of the zoom, but it will always show the same length. So if the extent is smaller than the sizebar it will be outside the figure area.

So `matplotlib-scalebar` will be better in this case, since the plotting is interactive.

jlaehne:
Can you merge RnM and fix the conflicts?

jlaehne:
I send you a PR to improve the consistency in the docstrings across the rest of drawings/utils.py. If you merge it in this branch, I would merge this one in RnM.

ericpre:
Thanks!

PR #2563, opened 2020-10-23
Fix default value autoscale

ericpre:
The default value of `ImagePlot.autoscale`, which is used ony for `BaseSignal` and this is why it affects `plot_decomposition_results` and `get_decomposition_loadings().plot()` but not calling `plot()` of a `Signal2D` instance.

### Progress of the PR
- [x] Fix default value,
- [x] Match signature of `MPL_HyperExplorer.plot_signals` with others `plot_signals`, 
- [x] ready for review.

### Minimal example of the bug
from https://gitter.im/hyperspy/hyperspy?at=5f92cfa0a7e77a0ff165ebef

```python
%matplotlib qt
import hyperspy.api as hs
import numpy as np

s = hs.signals.Signal1D(np.random.random((50, 50, 200)))
s.decomposition()

#not updating intensity/scale in loading maps, factors are updated; 'autoscale='v'' can not be used here
s.plot_decomposition_results()

l = s.get_decomposition_loadings()
f = s.get_decomposition_factors()

l.plot() #not updating intensity
l.plot(autoscale='v') #updating intensity
f.plot() #updating intensity
f.plot(autoscale='v') #updating intensity
```

### DIFF
--- a/hyperspy/drawing/image.py
+++ b/hyperspy/drawing/image.py
@@ -69,7 +69,7 @@ class ImagePlot(BlittedFigure):
 
         # Attribute matching the arguments of
         # `hyperspy._signal.signal2d.signal2D.plot`
-        self.autoscale = "z"
+        self.autoscale = "v"
         self.saturated_pixels = None
         self.norm = "auto"
         self.vmin = None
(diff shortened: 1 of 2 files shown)


thomasaarholt:
Thanks! I was getting annoyed by this one 👍 
LGTM.

Issue #2483, opened 2020-08-14
Bugs loading some Velox EMD files


din14970:
Of course just after I close my pull request on improving the Velox emd reader I find some datasets where the reader throws errors. The EMD files I am experiencing issues with can be downloaded here: https://owncloud.gwdg.de/index.php/s/D3oo5AEML7Zr9in

Datasets were collected in an image corrected Titan instrument with a Ceta camera. I'm using Hyperspy 1.6.0.

Dataset 2 has issues I believe because an FFT created in Velox may have been stored as an image dataset. The reader doesn't know how to handle this and I get this error:

```
---------------------------------------------------------------------------
OSError                                   Traceback (most recent call last)
<ipython-input-33-188d3ff6bef4> in <module>
----> 1 hs.load(folder+datasets[2])

/anaconda3/envs/temmeta/lib/python3.8/site-packages/hyperspy/io.py in load(filenames, signal_type, stack, stack_axis, new_axis_name, lazy, convert_units, escape_square_brackets, **kwds)
    337         else:
    338             # No stack, so simply we load all signals in all files separately
--> 339             objects = [load_single_file(filename, lazy=lazy,
    340                                         **kwds)
    341                        for filename in filenames]

/anaconda3/envs/temmeta/lib/python3.8/site-packages/hyperspy/io.py in <listcomp>(.0)
    337         else:
    338             # No stack, so simply we load all signals in all files separately
--> 339             objects = [load_single_file(filename, lazy=lazy,
    340                                         **kwds)
    341                        for filename in filenames]

/anaconda3/envs/temmeta/lib/python3.8/site-packages/hyperspy/io.py in load_single_file(filename, **kwds)
    381     else:
    382         reader = io_plugins[i]
--> 383         return load_with_reader(filename=filename, reader=reader, **kwds)
    384 
    385 

/anaconda3/envs/temmeta/lib/python3.8/site-packages/hyperspy/io.py in load_with_reader(filename, reader, signal_type, convert_units, **kwds)
    387                      **kwds):
    388     lazy = kwds.get('lazy', False)
--> 389     file_data_list = reader.file_reader(filename,
    390                                         **kwds)
    391     objects = []

/anaconda3/envs/temmeta/lib/python3.8/site-packages/hyperspy/io_plugins/emd.py in file_reader(filename, lazy, **kwds)
   1716             raise IOError("The file is not a supported EMD file.")
   1717     except Exception as e:
-> 1718         raise e
   1719     finally:
[...]

PR #2564, opened 2020-10-24
Bug fix 1 to Velox EMD reader: accept FFT

din14970:
### Description of the change
* In response to the issue #2483, I made a slight modification to the Velox EMD reader so that it now also reads in FFT's made in the program as ComplexSignal2D. Beforehand, `load` would simply fail on these files as this datatype is not expected. 
* I added a small emd file (modified experimental file) to the test suite containing only a dataset with this datatype and 12 pixels. I added a test that loads in this data and checks if a ComplexSignal2D is returned
* I also edited a small typo in a debug message
* I also fixed bug number 2 whereby the reader assumes a stack of STEM images, not TEM images. If one has a stack of TEM images, then `Frametime` is not in the metadata. If this key is not found, I look for `Exposuretime` in the detector metadata. An additional emd file was added to the tests with everything identical to an experimental dataset except the size of the "data".

Things I would still need some input on:
* I'm unsure which groups in the EMD file are necessary and which are not. Maybe I can cut down on the test emd file size even more.
* I read in the FFT data as-is, but the value seems limited as it seems the image needs to be doubled and centered. How to best handle this?
* I convert tuple array elements to complex64. I'm completely unsure how to handle this with `lazy=True`.

### Progress of the PR
- [x] add tests,
- [x] ready for review.

```python
>>> import hyperspy.api as hs
>>> hs.load("tests/io/emd_files/example_complex.emd")
```


hyperspy/io_plugins/emd.py
```diff
@@ -1100,8 +1100,13 @@ def _read_image(self, image_group, image_sub_group_key):
             # Change back to standard API once issue #977 is fixed.
             # Preallocate the numpy array and use read_direct method, which is
             # much faster in case of chunked data.
-            data = np.empty(h5data.shape)
+            data = np.empty(h5data.shape, h5data.dtype)
```
ericpre, line 1100: It seems that defining the dtype here make the `read_direct` call much slower for some dataset, such as the "17_ 20191205 1.3 Mx Imaging Ceta 1347.emd" file, you shared in #2483. This sounds very surprising and figuring out the reason may be complicated (issue https://github.com/h5py/h5py/issues/977 and co have been known since a while), so we could simply set the dtype only when loading complex data.

ericpre (changes requested):
This looks good except for the performance regression about `read_direct`.

> Things I would still need some input on:
>
> - I'm unsure which groups in the EMD file are necessary and which are not. Maybe I can cut down on the test emd file size even more.

The file is 141 KB is fine, but the 1.5 MB is a bit large, can you get a smaller one? The metadata shouldn't be that much, only the data itself should be large?

> - I read in the FFT data as-is, but the value seems limited as it seems the image needs to be doubled and centered. How to best handle this?

The 0, 0 is in the bottom left corner, which is one way to store it and fits well with the default in hyperspy of [`fft`](https://hyperspy.readthedocs.io/en/latest/user_guide/signal.html?highlight=fft#fast-fourier-transform-fft). Regarding the size of the fft, it is possible that the negative frequency are not saved, assuming that the data the fft was calculated from were real? Is so, I think this is fine to load as it is and document it accordingly.

> - I convert tuple array elements to complex64. I'm completely unsure how to handle this with lazy=True.

Lazy loading is fully supported at the moment for reading velox emd file and considering that fft should be fairly small data (single image), I am not convinced that this is stickily necessary? I would simply load it non-lazily even `lazy=True`. What is important is to read file containing fft without failure.

ericpre:
If you could rebase on the `next_patch_branch`, then it would be released as part of the 1.6.1 instead of 1.7.

hyperspy/io_plugins/emd.py
din14970: > but the 1.5 MB is a bit large

Will cut into this one and try to erase the larger file from commit history. I made the data 3x3x25 to keep FrameLookupTable the same as in the experimental file, but will cut the number of "frames".

> fft was calculated from were real? Is so, I think this is fine to load as it is and document it accordingly.

Ok, just with a comment in the code or add a warning to the user also? The data will not look like it does in Velox basically when plotted, which might confuse some users. Then again, I suppose most users will make their own fft's from the image data and not rely on the one from Velox.

> I would simply load it non-lazily even lazy=True

Fair point, I will implement it like that. Lazy reading works, but when I attempt plot or compute it fails because of the odd data type. I will add a logger warning that lazy doesn't work for FFT.

> It seems that defining the dtype here make the read_direct call much slower for some dataset

Wow you are right, the difference is really order of magnitude. Very odd. I will rework it so that it only happens if the dtype is special.

ericpre: > > fft was calculated from were real? Is so, I think this is fine to load as it is and document it accordingly.
> 
> Ok, just with a comment in the code or add a warning to the user also? The data will not look like it does in Velox basically when plotted, which might confuse some users. Then again, I suppose most users will make their own fft's from the image data and not rely on the one from Velox.

Maybe add a few sentences in the user guide (https://hyperspy.readthedocs.io/en/latest/user_guide/io.html#emd-velox) to explain that the FFT is not centered and only the positive frequency are saved in the file?
In any case, as you said, users would be better off calculating the FFT again.

din14970: done

ericpre:
Thanks @din14970!

ericpre (approved):

din14970:
Hope I can make more useful contributions in the future.

ericpre:
Sounds good, there are plenty of things to do in hyperspy and other libraries!

bryandesser:
There remains a small bug in reading Velox EMD files with FFTs. In `_read_image()`: `fft_dtype = [('realFloatHalfEven', '<f4'), ('imagFloatHalfEven', '<f4')]`; however, this is not always the case. If, for example, an FFT is taken of a region with odd number of pixels, then Velox saves the data as `fft_dtype = [('realFloatHalfOdd', '<f4'), ('imagFloatHalfOdd', '<f4')]`

This issue causes `hs.load()` to fail in the presence of an FFT with odd number of pixels.

din14970:
@bryandesser if you have such an example file could you make a PR to fix? Perhaps also move your comment to an issue.

bryandesser:
@din14970 I've created a PR [https://github.com/hyperspy/hyperspy/pull/3028#issue-1380192233](https://github.com/hyperspy/hyperspy/pull/3028#issue-1380192233). I have also moved the comment to an issue [https://github.com/hyperspy/hyperspy/issues/3029#issue-1380196506](https://github.com/hyperspy/hyperspy/issues/3029#issue-1380196506)

Issue #2566, opened 2020-10-29
Another Velox EMD bug: DPC signals


din14970:
The reader fails with DPC signals that are stored similarly but not identically to FFT's.
I should be able to work out a relatively quick fix.
For now I will also implement it also without support for lazy due to the weird data type.
Theoretically a person might collect a big stack of DPC images (maybe as some in-situ experiment?) so lazy support for the future might be nice.

din14970:
Solved in #2567, also the titles given to datasets was updated

PR #2567, opened 2020-10-29
added support for DPC

din14970:
### Description of the change
- Added support for reading in DPC signals stored in Velox EMD files, see issue #2566 

### Progress of the PR
- [x] add tests,
- [x] ready for review.


thomasaarholt:
You're being very efficient :D I don't need it for now, but could you check if EDS datasets that also collect DPC can be read as well?

din14970:
I don't have such a dataset to try, but I would expect that it should also work. The DPC signals are all just stored as images in the `Image` group, they should be mostly independent from the `SpectrumStream` group. Then again, the Velox files can throw interesting curve-balls. This would be another example though of where `lazy` reading of the DPC image stack might be desirable.

ericpre:
@din14970, they should be some metadata to parse when loading this type of data? This would be good to update the title to something more useful than DF4?

@thomasaarholt, as mentioned by @din14970, reading these image is independent of reading the EDS stream. Is there anything in particular you have in mind?

thomasaarholt:
If they're independent, then all is well - that was my only concern.

din14970:
@ericpre yes this is a big problem, because unfortunately all DPC related signals come from the detector that Velox identifies as DF4 (at least on our microscope) so they all get the same title in hyperspy. In the metadata that accompanies the dataset I could not find anything that provides more context to the dataset. Instead, one can parse the info in I think the `display` or `presentation` group, that stores the display title velox uses and links to the dataset path. However, in this group there are only entries for those datasets that are open as a subwindow in Velox, which is not all the datasets in the hdf5 file. I would have to dig deeper to see if it is storing some additional info elsewhere.

ericpre:
https://github.com/din14970/hyperspy/pull/1

din14970:
@ericpre The titles for DPC signals should now be fixed, it works at least on all the datasets I have. I added a minimal test emd file and a test which should cover most of the `_get_detector_name` function. Test file is still somewhat big (700 kb) because there are 8 datasets inside. All are only 2x2 but all the metadata add up.

hyperspy/tests/io/test_emd.py
```diff
@@ -633,3 +633,21 @@ def test_fei_no_frametime():
     assert isinstance(signal, Signal2D)
     assert signal.data.shape == (2, 3, 3)
     assert signal.axes_manager["Time"].scale == 0.8
+
+
+def test_fei_dpc_loading():
+    signal = load(os.path.join(my_path, 'emd_files', 'fei_example_dpc.emd'))
```
ericpre: Is it worth keeping this test file, now that the other file can cover this test?

hyperspy/tests/io/test_emd.py
```diff
@@ -633,3 +633,21 @@ def test_fei_no_frametime():
...
+    assert isinstance(signal, ComplexSignal2D)
+
+
+def test_fei_image_titles():
+    signals = load(os.path.join(my_path, 'emd_files', 'fei_example_titles.emd'))
+    assert signals[0].metadata.General.title == "B-D"
+    assert signals[1].metadata.General.title == "DPC"
```
ericpre, line 641: Check the signal class too?

hyperspy/io_plugins/emd.py
```diff
@@ -1102,6 +1099,17 @@ def _read_image(self, image_group, image_sub_group_key):
...
+        elif h5data.dtype == [('realFloat', '<f4'),
+                              ('imagFloat', '<f4')]:
+            _logger.debug("Found a DPC signal, loading as Complex2DSignal")
+            if self.lazy:
+                _logger.warning("Lazy not supported for DPC")
+            data = np.empty(h5data.shape, h5data.dtype)
+            h5data.read_direct(data)
+            data = data['realFloat'] + 1j * data['imagFloat']
+            # Set the axes in frame, y, x order
+            data = np.rollaxis(data, axis=2)
```
ericpre: Merge with the above:
```python
        # For Velox FFT or DPC data, dtype must be specified and lazy is not
        # supported due to special dtype. The data is loaded as-is; to get
        # a traditional view the negative half must be created and the data
        # must be re-centered
        dtype_name = h5data.dtype.names
        custom_dtype_name_list = [('realFloatHalfEven', 'imagFloatHalfEven'),
                                  ('realFloat', 'imagFloat')]
        if dtype_name in custom_dtype_name_list:
            _logger.debug("Found an FFT or DPC, loading as Complex2DSignal")
            if self.lazy:
                _logger.warning("Lazy not supported for FFT or DPC")
            data = np.empty(h5data.shape, h5data.dtype)
            h5data.read_direct(data)
            data = data[dtype_name[0]] + 1j * data[dtype_name[1]]
            # Set the axes in frame, y, x order
            data = np.rollaxis(data, axis=2)
```

ericpre (changes requested):
I have left a few comments. I have tried with the "17_ 20191205 1.3 Mx Imaging Ceta 1347.emd" file, you shared in #2483 and I get a `"Unrecognized_image_signal"` title. I don't know what is this dataset supposed to be (it seems to the same as the one in the same file?)

Other than that, this looks good to me.

hyperspy/tests/io/test_emd.py
din14970: Fair, I'll try to figure out a way to remove the file from the history.

hyperspy/tests/io/test_emd.py
din14970: done

hyperspy/io_plugins/emd.py
din14970: done

din14970:
@ericpre added some more conditions in the `_get_detector_name` function to catch FFT's and cross-correlation aligned datasets, but I suppose this is still not an exhaustive list and Thermo-Fisher can easily come up with additional `operations` that deserve different titles. All these will be caught with `Unrecognized_image_signal`. 

I still have to find the best way to remove the initial emd file from the history, unfortunately I added it in the same commit as the first commit in this PR.

ericpre (approved):
Great, thanks. Looks good to me!

din14970:
note to self: add binary files as separate commits. This was quite the hassle.

PR #2569, opened 2020-10-31
Support h5py 3.0

ericpre:
h5py 3.0 breaks a few io plugins because of change in string handling. Relevant links:
- https://docs.h5py.org/en/stable/whatsnew/3.0.html
- https://docs.h5py.org/en/3.0.0/strings.html

### Progress of the PR
- [x] Fix emd io plugin.
- [x] Fix hspy io plugin.
- [x] Fix nexus io plugin.
- [x] Use pytest `tmp_path` default fixture instead of a custom one.
- [x] ready for review.

### DIFF
--- a/hyperspy/io_plugins/emd.py
+++ b/hyperspy/io_plugins/emd.py
@@ -744,7 +744,7 @@ class EMD_NCEM:
                 value = value.decode()
             if key == 'units':
                 # Get all the units
-                units_list = re.findall("(\[.+?\])", value)
+                units_list = re.findall(r"(\[.+?\])", value)
                 units_list = [u[1:-1].replace("_", "") for u in units_list]
                 value = ' * '.join(units_list)
                 try:
--- a/hyperspy/io_plugins/emd.py
+++ b/hyperspy/io_plugins/emd.py
@@ -622,6 +622,18 @@ class EMD_NCEM:
         """
         return group.attrs.get('emd_group_type', False)
 
+    @staticmethod
+    def _read_dataset(dataset):
+        """Read dataset and use the h5py AsStrWrapper when the dataset is of
+        string type (h5py 3.0 and newer)
+        """
+        if (h5py.check_string_dtype(dataset.dtype) and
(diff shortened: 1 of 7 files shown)


hyperspy/io_plugins/hspy.py
```diff
@@ -544,6 +544,9 @@ def overwrite_dataset(group, data, key, signal_axes=None, chunks=None, **kwds):
             # Optimise the chunking to contain at least one signal per chunk
             chunks = get_signal_chunks(data.shape, data.dtype, signal_axes)
 
+    if np.issubdtype(data.dtype, np.dtype('U')):
+        # Saving numpy unicode type is not supported in h5py
+        data = data.astype(np.dtype('S'))
```
thomasaarholt, line 549: I notice that `'S'` zero-terminated bytes  are ["not recommended"](https://numpy.org/doc/stable/reference/arrays.dtypes.html) according to the numpy docs. Did you check to see if there are any alternatives here?

hyperspy/io_plugins/emd.py
```diff
@@ -850,6 +865,9 @@ def _write_signal_to_group(self, signal_group, signal):
         dataset = signal_group.require_group(title)
         data = signal.data.T
         maxshape = tuple(None for _ in data.shape)
+        if np.issubdtype(data.dtype, np.dtype('U')):
+            # Saving numpy unicode type is not supported in h5py
+            data = data.astype(np.dtype('S'))
```
thomasaarholt, line 870: Same here as below.

thomasaarholt (commented):
The `tmp_path` pytest fixture is really nice! I'll have to remember that. I've got one question about a numpy dtype in the review. This is mostly because I don't fully understand it, not because I have a strong opinion.

Otherwise, this looks good!

hyperspy/io_plugins/hspy.py
ericpre: Yes, I have read this too and I think this is fine because this is used only to write hdf5 dataset.

PR #2556, opened 2020-09-29
Bugfix for a channel_switches issue with 2d models fits

LMSC-NTappy:
### Description of the change
2dmodel.py is not taking into account the channel_switches when evaluating components on the axes, resulting in an error.
Doing what @ericpre is describing here was raising an error: 

>If I understand correctly, your aim is to able to use arbitrary mask to define the fitting ranges. `channel_switches` is actually the >mask, so you can simply do:
>```python
>m.channel_switches = your_mask
>```
>The only thing missing really is to document it in the user guide and in a couple of docstrings!
>
_Originally posted by @ericpre in https://github.com/hyperspy/hyperspy/issues/2454#issuecomment-683442074_

This PR fixes this. Example code below

### Progress of the PR
- [x] Change implemented (can be split into several points),
- [ ] update docstring (if appropriate),
- [ ] update user guide (if appropriate),
- [x] add tests,
- [x] ready for review.

### Minimal example of the bug fix or the new feature
```python
g = hs.model.components2D.Gaussian2D(A=1,
                                     centre_x=-5.0, 
                                     centre_y=-5.0, 
                                     sigma_x=1.0, 
                                     sigma_y=2.0
                                    )
scale = 0.1
x = np.arange(-10, 10, scale)
y = np.arange(-10, 10, scale)
X, Y = np.meshgrid(x, y)

im = hs.signals.Signal2D(g.function(X, Y))
im.axes_manager[0].scale = scale
im.axes_manager[0].offset = -10
im.axes_manager[1].scale = scale
im.axes_manager[1].offset = -10

m = im.create_model()
gt = hs.model.components2D.Gaussian2D()
m.append(gt)

mask = im<0.01
m.channel_switches = ~mask.data

m.fit()
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
<ipython-input-2-e19ac3d78e8f> in <module>
     30 m.channel_switches = ~mask.data
     31 
---> 32 m.fit()

~\Miniconda3\envs\hs\lib\site-packages\hyperspy\model.py in fit(self, optimizer, loss_function, grad, bounded, update_plot, print_info, return_info, fd_scheme, **kwargs)
   1396                         args=args,
   1397                         full_output=True,
-> 1398                         **kwargs,
   1399                     )
   1400 

[...]

PR #2570, opened 2020-11-03
Document setting signal range of Model2D using channel_switches

LMSC-NTappy:
### Description of the change
I realised the test on channel switches mask (2D signal ranges) I previously wrote was not actually applying the mask: #2556 . This is now fixed. I also updated doc to add instructions and an example (same as example below) on how to use `m.channel_switches`  for signal range selection wile fitting Signal2D objects.

### Progress of the PR
- [x] add tests,
- [x] ready for review.

```python
import hyperspy.api as hs
import numpy as np

#Creating a sample 2D gaussian dataset
g = hs.model.components2D.Gaussian2D(
    A=1, centre_x=-5.0, centre_y=-5.0, sigma_x=1.0, sigma_y=2.0,)
scale = 0.1
x = np.arange(-10, 10, scale)
y = np.arange(-10, 10, scale)
X, Y = np.meshgrid(x, y)

im = hs.signals.Signal2D(g.function(X, Y))
im.axes_manager[0].scale = scale
im.axes_manager[0].offset = -10
im.axes_manager[1].scale = scale
im.axes_manager[1].offset = -10

m = im.create_model() #Model initialisation
gt = hs.model.components2D.Gaussian2D()
m.append(gt)

sr = im._deepcopy_with_new_data( #signal range mask object
    np.zeros(im.axes_manager.signal_shape,dtype='bool'))
 sr.isig[-7.:-3.,-9.:-1.] = True

m.channel_switches = sr.data #setting channel switches
m.fit()
m.as_signal().plot()
```
image


doc/user_guide/model.rst
```diff
@@ -1183,6 +1183,41 @@ undesired spectral channels from the fitting process:
 * :py:meth:`~.models.model1d.Model1D.remove_signal_range`
 * :py:meth:`~.models.model1d.Model1D.reset_signal_range`
 
+In 2D models, those methods are not implemented (hyperspy 1.6.0). Instead, the
```
ericpre: No need to specify `hyperspy 1.6.0`. When (and if?) these methods will be implemented, the user guide will be update accordingly.

doc/user_guide/model.rst
```diff
@@ -1183,6 +1183,41 @@ undesired spectral channels from the fitting process:
...
+`m.channel_switches` attribute of a model can be set using boolean arrays of the
+same shape as the data's signal, where True means that the datapoint
+will be used in the fitting routine.
+
+A signal range can be set using axes units by creating another signal2D object:
```
ericpre: I find this sentence confusing, what's about:
```
The example below shows how this boolean array can be easily created from the signal and how the ``isig`` syntax can be used to define the signal range.
```

doc/user_guide/model.rst
```diff
@@ -1183,6 +1183,41 @@ undesired spectral channels from the fitting process:
...
+    >>> im.axes_manager[1].scale = scale
+    >>> im.axes_manager[1].offset = -10
+
+    >>> m = im.create_model() #Model initialisation
+    >>> gt = hs.model.components2D.Gaussian2D()
+    >>> m.append(gt)
+
+    >>> sr = im._deepcopy_with_new_data( #signal range mask object
+    ...   np.zeros(im.axes_manager.signal_shape,dtype='bool'))
+    >>> sr.isig[-7.:-3.,-9.:-1.] = True
```
ericpre: Avoid using private API and simplify to:
```suggestion
    >>> # Create a boolean signal of the same shape as the signal space of im
    >>> sr = (im.get_current_signal() == np.nan)
    >>> # Specify the signal range using the isig syntax
    >>> sr.isig[-7.:-3.,-9.:-1.] = True
```

Using `_deepcopy_with_new_data` will also copy the navigation axes in the axes manager and this is not what we want here: I don't think it would be working for the multidimensional data.

Use a variable name more explicit than `sr`? `signal_mask`?

hyperspy/tests/model/test_model2d.py
```diff
@@ -176,14 +176,15 @@ def test_channelswitches_mask():
...
     m = im.create_model()
     gt = hs.model.components2D.Gaussian2D(centre_x=-4.5,
                                           centre_y=-4.5,
                                           sigma_x=0.5,
                                           sigma_y=1.5)
     m.append(gt)
+    m.channel_switches = ~mask.data
     m.fit()
 
     np.testing.assert_allclose(gt.centre_x.value, -5.)
```
ericpre, line 200: Add a test to check that the mask was applied?

ericpre (changes requested):
I have suggested a couple of improvement.

ericpre (approved):

PR #2572, opened 2020-11-05
typo fix EMD reader

din14970:
### Description of the change
Super dumb but my last PR broke reading Velox EMD files for multiframe stem datasets due to a typo. Sorry!

### Progress of the PR
- [x] add tests,
- [x] ready for review.

### DIFF
--- a/hyperspy/io_plugins/emd.py
+++ b/hyperspy/io_plugins/emd.py
@@ -1140,7 +1140,7 @@ class FeiEMDReader(object):
             data = data[0, ...]
             i = 0
         else:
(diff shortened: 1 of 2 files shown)


ericpre:
Would it be possible to cover this with a test?

din14970:
@ericpre didn't want to add yet another test file, I just expanded one of the tests that tests loading of a dataset with a spectrum stream and an image stack. The option `load_SI_image_stack` seemed untested and this should catch this bit with the frame time. The `else` block where I test for exposure time should be covered by my test on the ceta camera stack dataset.

PR #2576, opened 2020-11-07
Plotting fixes

ericpre:
### Description of the change
- Don't set `imshow` interpolation default to allow using matplotlib default value or the `rcParams`. Since a while the matploltib default have been `nearest` and since matploltib 3.2, it is [`antialiased`](https://matplotlib.org/3.3.0/api/_as_gen/matplotlib.axes.Axes.imshow.html#matplotlib.axes.Axes.imshow), which is an improvement.
- Fix remove widget bug when a roi has been added to several signals.

### Progress of the PR
- [x] Change implemented,
- [x] add tests,
- [x] ready for review.

### Remove widget bug fix
```python
import numpy as np
import hyperspy.api as hs

s2 = s.deepcopy()
r = hs.roi.SpanROI(2, 4)
s.plot()
s2.plot()

s_roi = r.interactive(s)
s2_roi = r.interactive(s2)

r.remove_widget(s)
```

### DIFF
--- a/hyperspy/drawing/image.py
+++ b/hyperspy/drawing/image.py
@@ -557,8 +557,7 @@ class ImagePlot(BlittedFigure):
             else:
                 self.figure.canvas.draw_idle()
         else:  # no signal have been drawn yet
-            new_args = {'interpolation': 'nearest',
-                        'extent': self._extent,
+            new_args = {'extent': self._extent,
                         'aspect': self._aspect,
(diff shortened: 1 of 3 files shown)


thomasaarholt:
Nice! If people in the gitter chat start making noise about the figures looking different, then let's remember to point them towards the matplotlib antialiased default. I agree that it is an improvement!

PR #2578, opened 2020-11-08
Propagate the lazy argument when loading metadata

ptim0626:
This is a very simple bug fix for loading NeXus file lazily correctly. The `lazy` argument needs to be read by the function that load
metadata, as the items include the actual data and if the data is big, not saving it as dask array will cause memory issue.

### Description of the change
- pass the `lazy` argument correctly to the `_load_metadata` method in nexus file reader.

### Progress of the PR
- [x]  ready for review

Dataset in NeXus file exceeding the memory of the system can be loaded lazily.


ericpre (changes requested):
It would be good to cover this in the test suite - see https://codecov.io/gh/hyperspy/hyperspy/src/RELEASE_next_patch/hyperspy/io_plugins/nexus.py#L91.

PR #2579, opened 2020-11-09
Fix saving tiff file for signal with Å units

ericpre:
Fix saving as tiff for signal with `'Å'` units.

### Progress of the PR
- [x] Workaround ascii encoding error.
- [x] add tests,
- [x] ready for review.

```python
s = hs.signals.Signal2D(np.arange(200*200, dtype='float32').reshape((200, 200)))
for axis in s.axes_manager.signal_axes:
    axis.units = 'Å'
    axis.scale = 0.1
    axis.offset = 10

s.save('test.tif')
```
Error:
```python

  File "<ipython-input-4-6e3430f6830e>", line 1, in <module>
    s.save('test.tif')

  File "/home/eric/Dev/hyperspy/hyperspy/signal.py", line 2748, in save

  File "/home/eric/Dev/hyperspy/hyperspy/io.py", line 756, in save
    writer.file_writer(str(filename), signal, **kwds)

  File "/home/eric/Dev/hyperspy/hyperspy/io_plugins/tiff.py", line 112, in file_writer
    imwrite(filename,

  File "/opt/miniconda3/lib/python3.8/site-packages/tifffile/tifffile.py", line 787, in imwrite
    return tif.write(data, shape, dtype, **kwargs)

  File "/opt/miniconda3/lib/python3.8/site-packages/tifffile/tifffile.py", line 2125, in write
    addtag(*t)

  File "/opt/miniconda3/lib/python3.8/site-packages/tifffile/tifffile.py", line 1772, in addtag
    value = bytestr(value, 'ascii') + b'\0'

  File "/opt/miniconda3/lib/python3.8/site-packages/tifffile/tifffile.py", line 15622, in bytestr
    return s.encode(encoding) if isinstance(s, str) else s

UnicodeEncodeError: 'ascii' codec can't encode character '\xc5' in position 18: ordinal not in range(128)
```

### DIFF
--- a/hyperspy/io_plugins/tiff.py
+++ b/hyperspy/io_plugins/tiff.py
@@ -95,7 +95,7 @@ def file_writer(filename, signal, export_scale=True, extratags=[], **kwds):
             "because it is incompability with the 'ImageJ' tiff format")
     if export_scale:
         kwds.update(_get_tags_dict(signal, extratags=extratags))
-        _logger.debug("kwargs passed to tifffile.py imsave: {0}".format(kwds))
+        _logger.debug(f"kwargs passed to tifffile.py imsave: {kwds}")
 
         if 'metadata' not in kwds.keys():
             # Because we write the calibration to the ImageDescription tag
@@ -491,7 +491,9 @@ def _imagej_description(version='1.11a', **kwargs):
     for key, value in list(kwargs.items()):
         if value == 'µm':
             value = 'micron'
-        append.append('%s=%s' % (key.lower(), value))
+        if value == 'Å':
+            value = 'angstrom'
+        append.append(f'{key.lower()}={value}')
 
     return '\n'.join(result + append + [''])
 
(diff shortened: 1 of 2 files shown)


thomasaarholt:
The tif file your example creates raises a "corrupt tif" error when attempted opening in GMS. Any tricks you could suggest that would let it open there? 

As far as I can tell, the `"Å"` capability introduced here is only to let ImageJ open the files?

An alternative could be, if saving as a tif, to convert "Å" to "angstrom" (and back again when reading in with hyperspy). This file is openable (with correct units) in both GMS and ImageJ.

ericpre:
> The tif file your example creates raises a "corrupt tif" error when attempted opening in GMS. Any tricks you could suggest that would let it open there?

I have just checked again with GMS1 and GMS3 and it opens fine.

> An alternative could be, if saving as a tif, to convert "Å" to "angstrom" (and back again when reading in with hyperspy). This file is openable (with correct units) in both GMS and ImageJ.

This is exactly what this PR does!

thomasaarholt:
I honestly have no idea why it works now and didn't earlier. My bad!

thomasaarholt:
> This is exactly what this PR does!

I think I got tricked by the "_imagej_description" to think that this was just an imagej feature, enhanced by my experience from the error. I think I must have somehow corrupted the tif creation. No idea.

Ergo, this looks fine to me 😆

PR #2580, opened 2020-11-10
Fix absorption

### LINKED ISSUE #2575: Mass thickness unit in CL absorption correction in TEM-EDS quantification
It seems that the mass thickness returned by the CL_get_mass_thickness method in the _signals/eds_tem.py file has the wrong unit. That method should return the mass thickness in kg/m2. But I think it returns the mass thickness in g/cm2. On line 911 in eds_tem.py:
```python
elemental_mt = element_composition * thickness_map * 1E-9 * density
```
the elemental mass thickness is calculated as the product of composition (in %), thickness (in nm) and density (in g/cm3) and a factor of 1e-9 to get from % to fraction (1e-2) and from nm to cm (1e-7), resulting in a mass thickness in g/cm2.

ericpre:
Closes #2575.

### Progress of the PR
- [x] Fix mass thickness calculation for Cliff-Lorimer quantification
- [x] Fix unecessary progress bar when running absorption correction,
- [x] update tests,
- [x] ready for review.

### DIFF
--- a/hyperspy/_signals/eds_tem.py
+++ b/hyperspy/_signals/eds_tem.py
@@ -453,8 +453,10 @@ class EDSTEM_mixin:
                 if absorption_correction:
                     if thickness is not None:
                         mass_thickness = intensities[0].deepcopy()
-                        mass_thickness.data = self.CL_get_mass_thickness(composition.split(),
-                                                                thickness)
+                        mass_thickness.data = self.CL_get_mass_thickness(
+                            composition.split(),
+                            thickness
+                            )
                         mass_thickness.metadata.General.title = 'Mass thickness'
                     else:
                         raise ValueError(
@@ -473,16 +475,20 @@ class EDSTEM_mixin:
                 number_of_atoms = composition._deepcopy_with_new_data(results[1])
 
             if method == 'cross_section':
-                abs_corr_factor = utils_eds.get_abs_corr_cross_section(composition.split(),
-                                                       number_of_atoms.split(),
-                                                       toa,
-                                                       probe_area)
+                abs_corr_factor = utils_eds.get_abs_corr_cross_section(
+                    composition.split(),
+                    number_of_atoms.split(),
+                    toa,
+                    probe_area
+                    )
                 kwargs["absorption_correction"] = abs_corr_factor
             else:
                 if absorption_correction:
-                    abs_corr_factor = utils_eds.get_abs_corr_zeta(composition.split(),
-                                                       mass_thickness,
-                                                       toa)
+                    abs_corr_factor = utils_eds.get_abs_corr_zeta(
+                        composition.split(),
+                        mass_thickness,
+                        toa
+                        )
                     kwargs["absorption_correction"] = abs_corr_factor
 
             res_max = np.max((composition - comp_old).data)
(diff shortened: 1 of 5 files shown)


ericpre:
@petergraat and @k8macarthur, can you please review this PR?

hyperspy/_signals/eds_tem.py
```diff
@@ -886,21 +889,20 @@ def CL_get_mass_thickness(self, weight_percent, thickness):
```
petergraat, line 887: the two comment lines above this line (Input given by ... ... unit electric charge) are not relevant for this function.

hyperspy/_signals/eds_tem.py
```diff
@@ -886,21 +889,20 @@ def CL_get_mass_thickness(self, weight_percent, thickness):
 
         Parameters
         ----------
-
-        composition: stack of compositions as determined from an initial k_factor
+        weight_percent: stack of compositions as determined from an initial k_factor
             quantification.
             Probe current in nA
```
petergraat: This line (Probe current in nA) is not relevant for this function.

hyperspy/_signals/eds_tem.py
```diff
@@ -910,7 +912,10 @@ def CL_get_mass_thickness(self, weight_percent, thickness):
             [elements_db[element]['Physical_properties']['density (g/cm^3)']
                     for element in elements])
         for density, element_composition in zip(densities, weight_percent):
-            elemental_mt = element_composition * thickness_map * 1E-9 * density
+            # convert composition from % to fraction: factor of 1E-2
+            # convert thickness from nm to m: factor of 1E-9
+            # convert density from g/cm3 to kg/m2: factor of 1E3
+            elemental_mt = element_composition * thickness_map * density * 1E-8
```
petergraat, line 916: Yes, this should be OK now

petergraat (commented):
The calculation of elemental_mt is OK now.
I have added two remarks about comment lines that can be omitted.

hyperspy/_signals/eds_tem.py
k8macarthur: I think the units of composition is fractional and not percentage, therefore you're currently a factor of 100 out from what I think it should be.

ericpre: What make you think that this is fractional and not percentage?
According to the following, the result of the initial quantification is multiplied by 100 (in L449 below), so I would expect it to be percentage:
https://github.com/hyperspy/hyperspy/blob/c82ae3b1e01879bb0bb05dfc5685c00399f7465d/hyperspy/_signals/eds_tem.py#L446-L454

And `quantification_cliff_lorimer` returns weight fraction:
https://github.com/hyperspy/hyperspy/blob/c82ae3b1e01879bb0bb05dfc5685c00399f7465d/hyperspy/misc/eds/utils.py#L388-L410

k8macarthur: @ericpre sorry trying to multitask at the microscope this week and failing. You're right the composition at this point is in percent rather than fraction. I've done my best to comb through the rest of the code and I think it is all consistent apart from this one scaling error. I was looking in all the individual quantification functions and saw that they returned fractions and didn't check the umbrella function.

ericpre: No worries, this is easy to get lost in the conversion and this is why I pointed out where it came from exactly!

ericpre:
Thanks @petergraat and @k8macarthur for the review!

hyperspy/_signals/eds_tem.py
ericpre: Done!

PR #2581, opened 2020-11-12
Use mode=r for loading lazy data for hdf5 file as well

ptim0626:
### Description of the change
Related to #2555. It seems like there is no advantage of loading data lazily when the read mode for h5py is set to '`r+'`. This PR changes the mode to `'r'` for lazy data. This won't change the time stamp unnecessarily.

### Progress of the PR
- [x] ready for review.

Do not alter the time stamp when loading lazy data in hdf5 files.

### DIFF
--- a/hyperspy/io_plugins/emd.py
+++ b/hyperspy/io_plugins/emd.py
@@ -356,7 +356,7 @@ class EMD(object):
         """
         cls._log.debug('Calling load_from_emd')
         # Read in file:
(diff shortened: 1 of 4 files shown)


ericpre:
@ptim0626, is there anything else you would like to do in this PR?

ptim0626:
@ericpre  No, not on this PR. This one just focus on the read mode.

PR #2583, opened 2020-11-14
Use prettytable dependencies instead of ptable

ericpre:
Use prettytable dependencies instead of ptable because ptable is an unmaintained fork of prettytable. Both API are still compatible so any of the two can be used and it is better to use the one which is maintained.

### Progress of the PR
- [x] Use prettytable dependencies instead of ptable,
- [x] tidy up conda environment,
- [x] ready for review.

### DIFF
--- a/conda_environment.yml
+++ b/conda_environment.yml
@@ -3,7 +3,7 @@ channels:
 - defaults
 - conda-forge
 dependencies:
-- dask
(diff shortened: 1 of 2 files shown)


thomasaarholt:
Nice find and looks good. I realised I could edit the branch directly so I just changed a typo in a comment.

PR #2584, opened 2020-11-16
Place qt navigator+signal plots side by side

thomasaarholt:
### Description of the change
I've offset the navigator and signal plots by half their widths left and right, so that they appear next to one another. The small annoyance of having to always drag the plots left and right has somewhat prevented me from embracing the qt backend, even though it is much, much faster than nbagg.

I'm a little unsure how this performs on other platforms - I can only test on windows. I've tried testing with the tk backend, but that seems to be acting strangely on my testing environment without the change this PR brings.

Finally, whether or not the plots appear in front of the notebook appears to be a bit ambiguous. When testing it, sometimes the plots start appearing behind the browser. A browser refresh oddly enough seems to fix this. If anyone knows why this happens, I'd appreciate it.

### Progress of the PR
- [x] ready for review.

```python
# please try with different backends if you have them
%matplotlib qt
import hyperspy.api as hs
import numpy as np
s = hs.signals.Signal1D(np.random.random((10,10,10)))
s.plot()
```

### DIFF
--- a/hyperspy/drawing/mpl_he.py
+++ b/hyperspy/drawing/mpl_he.py
@@ -242,13 +242,15 @@ class MPL_HyperExplorer:
                     # Get coordinates of navigator plot
                     manager = self.navigator_plot.figure.canvas.manager
                     # Shift left
-                    left, top, width, height = manager.window.geometry().getRect()
-                    manager.window.setGeometry(left-width//2 ,top, width, height) # must take ints
+                    left1, top1, width1, height1 = manager.window.geometry().getRect()
+                    # setGeometry takes left, top, width, height
+                    # top = 0 results in hidden windows toolbar
+                    manager.window.setGeometry(0,30, width1, height1) 
 
                     # Get coordinates of signal plot
                     manager = self.signal_plot.figure.canvas.manager
                     # Shift right
-                    left, top, width, height = manager.window.geometry().getRect()
-                    manager.window.setGeometry(left+width//2 ,top, width, height) # must take ints
+                    left2, top2, width2, height2 = manager.window.geometry().getRect()
+                    manager.window.setGeometry(width1+1, 30, width2, height2) # +1 offset to left position for border
                 except:
                     pass
--- a/hyperspy/drawing/mpl_he.py
+++ b/hyperspy/drawing/mpl_he.py
@@ -23,7 +23,7 @@ from traits.api import Undefined
 
 from hyperspy.drawing import widgets, signal1d, image
 from hyperspy.defaults_parser import preferences
-
+import matplotlib.pyplot as plt
 
 _logger = logging.getLogger(__name__)
 
@@ -192,6 +192,7 @@ class MPL_HyperExplorer(object):
                 self.navigator_plot.events.closed.connect(
                     self.pointer.disconnect, [])
         self.plot_signal(**kwargs)
+        self.arrange_windows()
 
     def assign_pointer(self):
         if self.navigator_data_function is None:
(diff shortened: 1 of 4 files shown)


hyperspy/drawing/mpl_he.py
```diff
@@ -229,3 +230,25 @@ def close(self):
             if self.signal_plot:
                 self.signal_plot.close()
             self.close_navigator_plot()
+
+    def arrange_windows(self):
```
ericpre, line 234: This method should be private.

hyperspy/drawing/mpl_he.py
```diff
@@ -192,6 +192,7 @@ def plot(self, **kwargs):
                 self.navigator_plot.events.closed.connect(
                     self.pointer.disconnect, [])
         self.plot_signal(**kwargs)
+        self.arrange_windows()
```
ericpre, line 195: This should be called only when there is a `navigator_plot` figure:
```suggestion
        if self.navigator_plot is not None:
            self.arrange_windows()
```

hyperspy/drawing/mpl_he.py
```diff
@@ -229,3 +230,25 @@ def close(self):
...
+        """
+
+        if self.navigator_plot and self.signal_plot:
+            if "nbagg" not in plt.get_backend().lower(): # doesn't work for widget / nbagg backends
+                try: # Try incase some exotic backend raises an error
+                    # Get coordinates of navigator plot
+                    manager = self.navigator_plot.figure.canvas.manager
+                    # Shift left
+                    left, top, width, height = manager.window.geometry().getRect()
+                    manager.window.setGeometry(left-width//2 ,top, width, height) # must take ints
```
ericpre: Moving the window to `left-width//2` will not work when the window is already on the left hand side of the screen, which is the case my system (fedora) for the first figure... after this it goes to the right hand side of my current window...
It would be good to know what defined the initial position of the first window.

ericpre (changes requested):
This would be good indeed to improve the current situation!

> Finally, whether or not the plots appear in front of the notebook appears to be a bit ambiguous. When testing it, sometimes the plots start appearing behind the browser. A browser refresh oddly enough seems to fix this. If anyone knows why this happens, I'd appreciate it.

See https://github.com/matplotlib/matplotlib/issues/6378 but this may overwritten by the system (display server or window manager) setting.

hyperspy/drawing/mpl_he.py
thomasaarholt: @ericpre what backend are you using? `plt.get_backend()`.

Also, what other backends do you think people actively use?

It looks like any code to set the positions will be all backend-specific.

ericpre: This is with the qt backend. I am not sure if other backend works with the traitsui widgets and the qt backend is most likely the one which works best, so I would worry too much about other backend for now.

thomasaarholt: I've realised I can use pillow to get the screen resolution. Not sure how it plays with multiple monitors. I'll have a play with placing the figures as a function of screen size and come up with a draft.

ericpre: Using pillow to do this as a workaround doesn't sound good. Maybe, a simple solution would be to set the left position to zero for the navigator and offset the signal plot with the width of the navigator?

thomasaarholt:
@ericpre can you try now? 
I agree that we don't want to use pillow - it seems to be only a dependency of the IO, and eventually (😄) that will be split off from the main hyperspy repo.

ericpre:
To play with this, we need to understand how it works and after reading more about it (https://doc.qt.io/qt-5/application-windows.html#window-geometry and https://www.riverbankcomputing.com/static/Docs/PyQt5/api/qtwidgets/qwidget.html), it doesn't give me confidence that it is clear what it is doing. By the way, the method used here to set the position of the window are qt specific.

PR #2592, opened 2020-11-24
Move test suite to GitHub actions

ericpre:
### Progress of the PR
- [x] Move from travis to github actions,
- [x] update dev guide,
- [x] fix failure introduced in #2437 (https://github.com/hyperspy/hyperspy/pull/2437#issuecomment-732769088),
- [x] ready for review.


ericpre:
@thomasaarholt, I have removed the "blosc" compression as an option of the writing hspy file and instead added the relevant documentation in the user guide: this was not working consistently between conda and pip instal.

thomasaarholt:
Thanks for handling the blosc issue. I guess we can look at including it again later?

jlaehne:
LGTM

doc/user_guide/io.rst
```diff
@@ -369,18 +369,22 @@ See the `chunking section <big_data.html#Chunking>`__ under `Working with big da
 Extra saving arguments
 ^^^^^^^^^^^^^^^^^^^^^^^
 `compression`:
-  One of ``None``, ``'gzip'``, ``'szip'``, ``'lzf'``, ``'blosc'`` (default is ``'gzip'``). 
-  The ``'blosc'`` compressor requires ``pytables`` to be installed. It can significantly improve the speed of saving with very
-  good compression. ``'szip'`` may be unavailable as it depends on the HDF5 installation including it.
+  One of ``None``, ``'gzip'``, ``'szip'``, ``'lzf'`` (default is ``'gzip'``).
```
francisco-dlp: I find this part of the documentation a bit confusing. What about something on the following lines:

HyperSpy uses ``h5py`` for reading and writing HDF5 files and, therefore, it supports [all compression filters supported by h5py](https://docs.h5py.org/en/stable/high/dataset.html#dataset-compression). The default is ``'gzip'``. It is possible to enable other compression filters such as ``blosc`` by installing e.g. [hdf5plugin](https://github.com/silx-kit/hdf5plugin). However, be aware that loading those files will require installing the package providing the compression filter. If not available an error will be raised.

Compression can significantly increase the saving speed. If file size
    is not an issue, it can be disabled by setting ``compression = None``.
    Notice that only ``compression=None`` and ``compression='gzip'`` are available in all platforms, see the [h5py documentation](https://docs.h5py.org/en/stable/faq.html) for more details. Therefore, if you choose any other compression filter for saving a file, be aware that it may not be possible to load it in some platforms.

ericpre: Agreed, I have changed it.

Issue #2562, opened 2020-10-23
Clarify intensity and quantify in EELS model


ericpre:
from https://gitter.im/hyperspy/hyperspy?at=5f91eed3631a250ab286ce2f

> Weixin Song @winston-song Oct 22 21:42
>> I have a pretty simple question regarding the EELS quantification. I had a good look at the documentation but could not find an answer. If it is there, then please send me the link.  
The function: m.quantify() - “prints the intensity at the current locations of all the EELS ionisation edges in the model” according to the documentation. 
I have collected a STEM-EELS SI of 100x100 pixels and want to plot a quantitative map for a particular element. After all the standard pre-processing and model-fitting I run the command:
m["Cr_L3"].intensity.as_signal() to get the intensity of this edge to create a 2D map. 
A. My first question is regarding the definition of “intensity”, does it refer to the absolute quantification (including the cross-sections etc) or is it just one of the fitting parameters of the edge corresponding to the peak height? The BN example in the documentation gives a value of 0.045648 for the single spectrum quantification, what are the units for this value?
B. Is the m.quantify() function equivalent to m[“Edge”].intensity.as_signal()?
C. If they are not the same, how do I extract a 2D quantification map for a particular element. 
>>
>> Thanks,
>> Johan
>
> I also have the same question for a long time, I currently think the intensity value from quantify() is a number devided by the cross section, and different from edge.intensity, I hope any expert in this function can give more precise answer

> lukmuk @lukmuk 11:08
> @jwestraadt @winston-song I was also wondering about the same thing, but from my results it also seems that it is scaled by the cross-sections. When I normalize the m.quantify() values by their sum (i.e. print the values in at%), I get similar results as for a quantification in DigitalMicrograph (which includes the cross-sections).

lukmuk:
I can not speak for the other two commenters, but here are my thoughts on this issue:
When doing EELS quantification with the "integration window" method (Egerton chapter 4.5.1), the power-law background is substracted before the core-loss edge and then the raw EELS signal intensity is integrated in an energy window after the edge. This raw signal intensity is then weighted by the partial (Hartree-Slater) scattering cross-section for quantification, e.g., when comparing the concentration ratio between two elements (Egerton, eq. (4.66)). Therefore, the "intensity" in this context is not yet scaled by the scattering cross-section, i.e. just the raw EELS signal intensity.
[...]

PR #2593, opened 2020-11-25
Different outputs for EELS quantify()

lukmuk:
### Added different outputs in EELS quantify()
(Disclaimer: First pull request on Github, hopefully all works out)  
This change addresses issue #2562 by giving different possiblities for printed outputs for `quantify()` for EELS model-based quantification and adds some information in the user guide and docstring. The `quantify()` function now has a parameter 'output', which can be changed to manipulate the printed output:
- 'absolute': Same as now, i.e. the intensity values of the fitted core-loss edges. It's the default value, so it should not break any existing code. (?)
- 'normalized': Sums up all intensities and uses this for normalization of each intensity value in the model. This results in the elemental concentrations (if all elements are present in the model) and mimics the functionality of Digital Micrograph.
- None: No printed output. Useful, if only the newly returned dictionary with the intensity values is relevant for further analysis.

In addition, the elements[element][subshell] dictionary is returned, which can be used as an alternative to `m.component.element_subshell.intensity.value`.  
The documentation and user guide has been updated to accomodate for the changes. Hopefully this will clarify the meaning of the 'intensity' parameter, i.e. that it can be used directly to calculate elemental concentrations/ratios and does not have to be weightened by a (partial) scattering cross-section.  
The BN example from the EELS DB has 300 keV beam energy noted in its metadata, so I changed the value in the user guide: `s.set_microscope_parameters(beam_energy=300, convergence_angle=0.2, collection_angle=2.55)`. Else, running the example with 100 keV will yield skewed quantification results for BN (around 60/40 at% instead of ~50/50 at%).

### Progress of the PR
- [x] Change quantify() method.
- [x] update docstring,
- [x] update user guide,
- [x] ready for review.

### Minimal example of the new features
```python
import hyperspy.api as hs
#Follow BN EELS model fitting example from user guide 
s = hs.datasets.eelsdb(title="Hexagonal Boron Nitride", spectrum_type="coreloss")[0]
ll = hs.datasets.eelsdb(title="Hexagonal Boron Nitride", spectrum_type="lowloss")[0]
s.add_elements(('B', 'N'))
m = s.create_model(ll=ll)
m.enable_fine_structure()
m.smart_fit()
```
```python
#New quantify() output parameters
m.quantify() #same as default value 'absolute', i.e. current version of quantify()
m.quantify(output='absolute')
m.quantify(output='normalized') #Prints elemental concentrations
m.quantify(output=None) #No printed output
```
```python
#New quantify() returns dictionary
q = m.quantify(output=None)

#The dictionary can be used alternatively to the intensity component value of the model m
print('Intensity of B K edge: ', q['B']['K'])
print('Intensity of B K edge: ', m.components.B_K.intensity.value)
print('B/N ratio: ', q['B']['K']/q['N']['K'])
```
Regarding the beam energy of the example BN spectrum (300 keV instead of 100 keV):
```python
s = hs.datasets.eelsdb(title="Hexagonal Boron Nitride", spectrum_type="coreloss")[0]
s.metadata.Acquisition_instrument.TEM.beam_energy
```

### DIFF
--- a/doc/user_guide/eels.rst
+++ b/doc/user_guide/eels.rst
@@ -205,7 +205,7 @@ core-loss file
 
 .. code-block:: python
 
-    >>> s.set_microscope_parameters(beam_energy=100,
+    >>> s.set_microscope_parameters(beam_energy=300,
     ...                             convergence_angle=0.2,
     ...                             collection_angle=2.55)
 
@@ -293,7 +293,20 @@ Print the result of the fit
     B	0.045648
     N	0.048061
 
+See :py:class:`~._components.eels_cl_edge.EELSCLEdge` for more information about the intensity parameter.
+The intensity values can be used to calculate elemental concentrations and ratios:
 
+.. code-block:: python
+
+    >>> 0.045648/(0.045648+0.048061) #B concentration
+    0.4871250360157509
+    >>> 0.048061/(0.045648+0.048061) #N concentration
+    0.5128749639842491
+    >>> 0.045648/0.048061 #B/N ratio
+    0.9497929714321384
+
+This corresponds to 48.7% B and 51.3% N in the sample and a B/N ratio of 0.95. See :py:class:`~hyperspy.models.eelsmodel.EELSModel.quantify()` for more information.
+    
 Visualize the result
 
 .. code-block:: python
--- a/hyperspy/models/eelsmodel.py
+++ b/hyperspy/models/eelsmodel.py
@@ -597,13 +597,11 @@ class EELSModel(Model1D):
         print()
         if output == "absolute":
             print("Absolute quantification:")
-        elif output == "normalized":
-            print("Normalized elemental concentrations:")
-        if output == "absolute":
             print("Elem.\tIntensity")
         elif output == "normalized":
+            print("Normalized elemental concentrations:")
             print("Elem.\tConcentration")
-
+        
         for element in elements:
             if len(elements[element]) == 1:
                 for subshell in elements[element]:
--- a/hyperspy/models/eelsmodel.py
+++ b/hyperspy/models/eelsmodel.py
@@ -538,12 +538,51 @@ class EELSModel(Model1D):
         # Recover the channel_switches. Remove it or make it smarter.
         self.channel_switches = backup_channel_switches
 
-    def quantify(self):
-        """Prints the value of the intensity of all the independent
-        active EELS core loss edges defined in the model
 
+    def quantify(self, output = 'absolute'):
+        """Prints and returns the values of the intensity of all the independent
+        active EELS core loss edges defined in the model.
+        
+        Parameters
+        ----------
+        output : 'absolute', 'normalized', or None
+            If 'absolute', prints values of the intensity parameter of all EELSCLEdge components.\n
+            If 'normalized', prints intensity parameters normalized with respect to the sum of all intensities in the model. If all relevant elements are present in the analyzed spectrum this results in the elemental concentrations in atomic %.\n
+            If None, no output is printed.
+        
+        Returns
+        -------
+        dict
+            Python dictionary with element name, edge name and intensity value.
+        
+        Examples
+        -------
+        Load and fit the BN EELS example from the user guide:\n
+        >>> s = hs.datasets.eelsdb(title="Hexagonal Boron Nitride", spectrum_type="coreloss")[0]
+        >>> ll = hs.datasets.eelsdb(title="Hexagonal Boron Nitride", spectrum_type="lowloss")[0]
+        >>> s.add_elements(('B', 'N'))
+        >>> m = s.create_model(ll=ll)
+        >>> m.enable_fine_structure()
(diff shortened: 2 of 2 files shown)


doc/user_guide/eels.rst
```diff
@@ -205,7 +205,7 @@ core-loss file
 
 .. code-block:: python
 
-    >>> s.set_microscope_parameters(beam_energy=100,
+    >>> s.set_microscope_parameters(beam_energy=300,
```
ericpre, line 208: Actually, it seems that these information are already in the metadata! Remove it to simplify the user guide, since this is explained elsewhere?

hyperspy/models/eelsmodel.py
```diff
@@ -538,31 +538,92 @@ def _fit_edge(self, edgenumber, start_energy=None, **kwargs):
...
+                if output == "normalized":
+                    int_sum += edge.intensity.value
+
+        if output == None:
+            return elements
         print()
-        print("Absolute quantification:")
-        print("Elem.\tIntensity")
+        if output == "absolute":
+            print("Absolute quantification:")
```
ericpre, line 599: A suggestion for improvement: use prettytable to get a nicer display? For example:
https://github.com/hyperspy/hyperspy/blob/bb684ee33292e99e2bae2eb9ecd39f8506360164/hyperspy/_signals/eels.py#L242-L266

and 

https://github.com/hyperspy/hyperspy/blob/7ca78caac1fcae55d4be672f51632ba20504d418/hyperspy/utils/__init__.py#L71-L91

hyperspy/models/eelsmodel.py
```diff
@@ -538,31 +538,92 @@ def _fit_edge(self, edgenumber, start_energy=None, **kwargs):
         # Recover the channel_switches. Remove it or make it smarter.
         self.channel_switches = backup_channel_switches
 
-    def quantify(self):
-        """Prints the value of the intensity of all the independent
-        active EELS core loss edges defined in the model
+    def quantify(self, output="absolute"):
```
ericpre, line 541: I find the meaning the `output` not clear here. What's about having two argument: `normalise=True/False` and `return_output=True/False` (or `return_results`)?

ericpre (changes requested):
For a first PR, it looks fairly good! I have left some comments. It would be good to add a simple test to check there is not mistake in the calculation.
Also it would be good to merge with RnM to get the coverage - since the build measuring the coverage is now fixed.

doc/user_guide/eels.rst
lukmuk: I think `set_microscope_parameters` is in the user guide to show the function exits. I was just confused, that my quantification results were way off when running the whole user guide example for BN. As you suggested, we can remove it here as all all parameters are set in the EELSdb BN spectrum and just mention `set_microscope_parameters` in the text.

hyperspy/models/eelsmodel.py
lukmuk: Sounds good! I will look into that.

hyperspy/models/eelsmodel.py
lukmuk: Yeah, I think your suggestion here is better than my implementation. I tried to avoid too many parameters to not "clutter" the code. :-)

lukmuk:
@ericpre Thank you for the nice suggestions! I will work on them when I find time and update the PR.

hyperspy/models/eelsmodel.py
lukmuk: @ericpre I am back at work and will implement the changes shortly. However, to prevent a misunderstanding:
1) With `return_output` you mean to (optionally) return the dictionary --> This means that the printed output (in the prettytable) is always printed.
Or...
2) With `return_output` you mean to (optionally) print the output --> This will always return the dictionary but will omit the printed output.

I think you meant the first version as currently `quantify` will always print an output, but I just wanted to make sure.

Also, i found a `print_html` function for the prettytable in hyperspy/misc/utils.py which seems to generate the same output as simple `print`. Should I use the `print_html`?

```python
import prettytable
table = PrettyTable()
...
print(table)
#or
print(print_html(f_text=table.get_string,
                          f_html=table.get_html_string))
```

ericpre: I would say that both (displaying and returning results) should be optional. Indeed, currently `quantify` always display the results, but I guess this is a very old behaviour and it would an improvement to be able to disable it, for example to be able to use as part of other programme. Not to break the current behaviour, for now, we should keep the default to display the results.

> Also, i found a `print_html` function for the prettytable in hyperspy/misc/utils.py which seems to generate the same output as simple `print`. Should I use the `print_html`?

Yes, use `print_html`, because running the following in the notebook, doesn't display the html version of the table:
```python
from prettytable import PrettyTable
x = PrettyTable()
x.field_names = ["City name", "Area", "Population", "Annual Rainfall"]
x.add_row(["Adelaide", 1295, 1158259, 600.5])
print(x)
```

Issue #1812, opened 2018-01-23
Colored and overlayed elemental maps


jdsugar:
I am trying to maker overlaid and colored elemental maps from eds data. Much like the very simple operations that vendor software performs where I pick an element pick a color and then make a layered image that has all the elements in it. When I use get_lines_intensity, I can create greyscale maps, but I can’t figure out how to change their color and overlay multiple maps. And, when I try to plot them separately from the get_lines_intensity command I get an error.  Does anyone have any suggestions for how to do this? Thanks.

francisco-dlp:
There is a very old (maybe broken) function to overlay up to 3 images using the RGB channels, `plot_RGB_map`. You can import it as follows:

```python

from hyperspy.drawing.utils import plot_RGB_map
```

However, it is very limited. If you need something more powerful you could use directly matplotlib to overlay the images. See e.g. https://matplotlib.org/examples/pylab_examples/layer_images.html

The ``make_cmap``  from #1793 could help you with this.

I am sure that somebody else would have some more suggestions.

In  any case, it would be really nice to implement this feature is hyperspy. A possible syntax could be:

```python

hs.plot.plot_overlay_images((im1, "red", 0.3), (im2, "violet", 0.6))

```
where the 3rd argument is the weight or alpha parameter. Then one can imagine a ipywidgets GUI that enables changing the color and weight and nav index of the image interactively.

tjof2:
See also #19: 
> *@stefano-m commented on 21 Sep 2011*
> It would be nice to be able to create a composite "RGB" image from more than just three components
> E.g. see:
> http://dip4fish.blogspot.com/2011/05/making-mfish-image.html

Updated mfish link: https://dip4fish.blogspot.com/2014/03/combining-five-images-into-one-rgb.html

ericpre:
Done in #2599.

PR #2599, opened 2020-12-15
Plot overlay images

TomSlater:
### Description of the change
This PR implements a function to provide overlayed images of different colours (for example, for elemental maps). It aims to resolve issue #1812 .

Further improvements could be made by using ipywidgets to select colours and alpha values but this just adds initial functionality.

### Progress of the PR
- [ ] add entry to `CHANGES.rst` (if appropriate),
- [x] add tests,
- [x] ready for review.

```python
import hyperspy.api as hs
import numpy as np
from urllib.request import urlretrieve, urlopen
from zipfile import ZipFile
files = urlretrieve("https://www.dropbox.com/s/ecdlgwxjq04m5mx/HyperSpy_demos_EDS_TEM_files.zip?raw=1./HyperSpy_demos_EDX_TEM_files.zip")
with ZipFile("HyperSpy_demos_EDX_TEM_files.zip") as z:
    z.extractall()
si_EDS = hs.load("core_shell.hdf5")
im = si_EDS.get_lines_intensity()
hs.plot.plot_overlay_images(im, colors=['r','b'], alphas=[1.0,1.0], legend_list=['Fe','Pt'], scalebar=True, axes_decor='off')
```


TomSlater:
I've added a basic test for plotting but I can't seem to get pytest to create the plot when running with --mpl-generate-path (it is creating others but not the one I've added). Perhaps someone could help me with this?

ericpre:
I am wondering if it would make sense to incorporate it in `plot_images` instead of having a separate function?

> I've added a basic test for plotting but I can't seem to get pytest to create the plot when running with --mpl-generate-path (it is creating others but not the one I've added). Perhaps someone could help me with this?

Please check that your version of matplotlib and freetype match the one defined in `conda_environment_dev.yml` (https://hyperspy.readthedocs.io/en/latest/dev_guide/testing.html#plot-testing)

thomasaarholt:
> Please check that your version of matplotlib and freetype match the one defined in `conda_environment_dev.yml` (https://hyperspy.readthedocs.io/en/latest/dev_guide/testing.html#plot-testing)

It can be useful to create a testing evironment in the same way that @ericpre's nice github actions do it:

From the base hyperspy path, create and update a conda environment like this:

```bash
conda env create -f conda_environment.yml # install hyperspy dependencies
conda env update -f conda_environment_dev.yml # install hyperspy testing dependencies
conda activate test_env # test_env is the name set in the yml file
pip install -e . # install hyperspy as a develop install
```

@ericpre would it be a good idea to add the above to the docs? As an alternative to the first code block in [the "Running Tests" section](https://hyperspy.readthedocs.io/en/latest/dev_guide/testing.html#running-tests)?

TomSlater:
> I am wondering if it would make sense to incorporate it in `plot_images` instead of having a separate function?

I'm happy to incorporate it in to `plot_images` with an `overlay = True` argument or something similar. I don't have a preference really, I'd just followed Francisco's notes in #1812 when porting from my own function. Which do people think makes the most sense?

I'll have another go setting up a testing environment.

ericpre:
> > Please check that your version of matplotlib and freetype match the one defined in `conda_environment_dev.yml` (https://hyperspy.readthedocs.io/en/latest/dev_guide/testing.html#plot-testing)
> 
> It can be useful to create a testing evironment in the same way that @ericpre's nice github actions do it:
> 
> From the base hyperspy path, create and update a conda environment like this:
> 
> ```shell
> conda env create -f conda_environment.yml # install hyperspy dependencies
> conda env update -f conda_environment_dev.yml # install hyperspy testing dependencies
> conda activate test_env # test_env is the name set in the yml file
> pip install -e . # install hyperspy as a develop install
> ```
> 
> @ericpre would it be a good idea to add the above to the docs? As an alternative to the first code block in [the "Running Tests" section](https://hyperspy.readthedocs.io/en/latest/dev_guide/testing.html#running-tests)?

Maybe a link to https://hyperspy.readthedocs.io/en/latest/user_guide/install.html#installation-in-a-anaconda-minconda-distribution?

ericpre:
Here is an example:

image

A few suggestions:
- since there is no need for a colorbar, we can remove the white padding around the image, when `axes_decor=off`. This is convenient when saving the figure.
- add/remove a map when clicking on the legend, as it is currently done with `plot_spectra`? This should be possible by putting the for loop in a callback function and adapting the alpha accordingly. This would be used for a future widget adjusting the alpha.
- the fact that there are quite a bit of code which is the same or very similar to the `plot_images` function convinced me that it should be in the same function. The idea would be to switch overlay on/off simply by changing one parameter.
- the `legend_title` should be the same as the `label` parameter of `plot_images` and there should be possible to find the label automatically from the title. `plot_images` should be already doing some of it.

TomSlater:
> A few suggestions:
> 
> * since there is no need for a colorbar, we can remove the white padding around the image, when `axes_decor=off`. This is convenient when saving the figure.
> * add/remove a map when clicking on the legend, as it is currently done with `plot_spectra`? This should be possible by putting the for loop in a callback function and adapting the alpha accordingly. This would be used for a future widget adjusting the alpha.
> * the fact that there are quite a bit of code which is the same or very similar to the `plot_images` function convinced me that it should be in the same function. The idea would be to switch overlay on/off simply by changing one parameter.
> * the `legend_title` should be the same as the `label` parameter of `plot_images` and there should be possible to find the label automatically from the title. `plot_images` should be already doing some of it.

I think I've covered almost all of these now, thanks for suggestions @ericpre ! It's incorporated in to `plot_images`, uses `label` for the legend and can now add/remove maps by clicking on the legend. You can get an overlay by simply setting  `overlay=True` in the arguments of `plot_images`.

I'm not quite sure how to get rid of whitespace around the image, I think it may work in some cases but not all now. If anyone has any suggestions that would be very helpful.

ericpre:
One way to do it would be:

```python
import numpy as np

pixel_number = 500
data = np.arange(pixel_number**2).reshape(pixel_number, pixel_number)
shape = data.shape

dpi = 100
sampling_factor = 10 # adjust the width of single pixel
fig = plt.figure(figsize=[sampling_factor*v/dpi for v in shape], dpi=dpi)

ax = fig.add_axes([0, 0, 1, 1])
ax.axis('off')
ax.imshow(data, cmap='gray')

plt.savefig('test.png')
```

hyperspy/drawing/utils.py
```diff
@@ -488,8 +490,15 @@ def plot_images(images,
                 fig=None,
                 vmin=None,
                 vmax=None,
+                overlay=False,
+                overlay_title=None,
+                colors=[],
+                alphas=[],
```
ericpre: Set defaults value to `[1.0, 1.0, 1.0]`?

hyperspy/drawing/utils.py
```diff
@@ -488,8 +490,15 @@ def plot_images(images,
                 fig=None,
                 vmin=None,
                 vmax=None,
+                overlay=False,
+                overlay_title=None,
+                colors=[],
```
ericpre: set defaults colors to `['r', 'b', 'g']`?

hyperspy/drawing/utils.py
```diff
@@ -601,6 +610,29 @@ def plot_images(images,
...
+        If True, use the default title options.
+        If False, don't plot a title.
+    colors : list of char or hex string, optional
+	    `colors` should be a list compromising either characters or hex
+		strings, corresponding to colors acceptable to matplotlib. Details
+		can be found at https://matplotlib.org/2.0.2/api/colors_api.html.
+    alphas : list of floats, optional
+        `alphas` should be a list of floats corresponding to the alpha
+		value of each color.
+    legend: {None, list of str, 'auto'}, optional
```
ericpre: This is doing more or less the same as the `label` argument?

hyperspy/drawing/utils.py
```diff
@@ -878,135 +911,206 @@ def check_list_length(arg, arg_name):
...
+        
+        #Check if images all have same scale and therefore can be overlayed.
+        for im in images:
+            if (im.axes_manager[0].scale != 
+                images[0].axes_manager[0].scale):
+                raise ValueError("Images are not the same scale and so should"
+                                 "not be overlayed.")
+        
+        original_cmap = plt.rcParams['image.cmap']
+        plt.rcParams['image.cmap'] = 'gray'
```
ericpre: Why is it necessary?

hyperspy/drawing/utils.py
```diff
@@ -878,135 +911,206 @@ def check_list_length(arg, arg_name):
...
+        if colors == []:
+            for i in range(len(images)):
+                colors.append(list(BASE_COLORS)[i])
+        
+        #If no alphas are selected use 1.0
+        if alphas == []:
+            for i in range(len(images)):
+                alphas.append(1.0)
+    
+        ax.imshow(np.zeros_like(images[0].data))
```
ericpre: Why is it necessary?

hyperspy/drawing/utils.py
```diff
@@ -878,135 +911,206 @@ def check_list_length(arg, arg_name):
...
+        original_cmap = plt.rcParams['image.cmap']
+        plt.rcParams['image.cmap'] = 'gray'
+        import matplotlib.patches as mpatches
+        ax = f.add_subplot(1, 1, 1)
+        patches = []
+        
+        #If no colors are selected use BASE_COLORS
+        if colors == []:
+            for i in range(len(images)):
+                colors.append(list(BASE_COLORS)[i])
```
ericpre: Define it in the definition of the function: less code for us and more explicit for the user.

hyperspy/drawing/utils.py
```diff
@@ -878,135 +911,206 @@ def check_list_length(arg, arg_name):
...
+        for im in images:
+            if (im.axes_manager[0].scale != 
+                images[0].axes_manager[0].scale):
+                raise ValueError("Images are not the same scale and so should"
+                                 "not be overlayed.")
+        
+        original_cmap = plt.rcParams['image.cmap']
+        plt.rcParams['image.cmap'] = 'gray'
+        import matplotlib.patches as mpatches
+        ax = f.add_subplot(1, 1, 1)
```
ericpre: Use `ax = f.add_axes([0, 0, 1, 1])` with the figure size set the correct aspect ratio as mentioned in https://github.com/hyperspy/hyperspy/pull/2599#issuecomment-746963107

hyperspy/drawing/utils.py
```diff
@@ -878,135 +911,206 @@ def check_list_length(arg, arg_name):
...
+                patches.append(mpatches.Patch(color=colors[i],
+                                              label=legend_label))
+                
+        if legend is not None:
+            plt.legend(handles=patches, loc=legend_loc)
+            if legend_picking is True:
+                animate_legend(fig=f, ax=ax, plot_type='images')
+        
+        if axes_decor == 'off':
+            ax.margins=(0,0)
```
ericpre: Remove this once the padding is fixed?

hyperspy/drawing/utils.py
```diff
@@ -878,135 +911,206 @@ def check_list_length(arg, arg_name):
...
+        ax.imshow(np.zeros_like(images[0].data))
+        
+        #Loop through each image
+        for i, im in enumerate(images):
+            cols = ['k', colors[i]]
+            cmap=LinearSegmentedColormap.from_list('cmap'+str(i), cols)
+            my_cmap = cmap(np.arange(cmap.N))
+            my_cmap[:,-1] = np.linspace(0.3, 1, cmap.N)
+            my_cmap = ListedColormap(my_cmap)
+            ax.imshow(im.data, vmin=im.data.min(), vmax=im.data.max(),
```
ericpre: The `vmin` and `vmax` defined in the function should be used here.

hyperspy/drawing/utils.py
```diff
@@ -878,135 +911,206 @@ def check_list_length(arg, arg_name):
...
+    
+        ax.imshow(np.zeros_like(images[0].data))
+        
+        #Loop through each image
+        for i, im in enumerate(images):
+            cols = ['k', colors[i]]
+            cmap=LinearSegmentedColormap.from_list('cmap'+str(i), cols)
+            my_cmap = cmap(np.arange(cmap.N))
+            my_cmap[:,-1] = np.linspace(0.3, 1, cmap.N)
+            my_cmap = ListedColormap(my_cmap)
```
ericpre: Is this necessary? The following seems to be doing what we need?
```python
import numpy as np
from matplotlib.colors import LinearSegmentedColormap

# cmap = LinearSegmentedColormap.from_list("", ["black", "red"])
cmap = LinearSegmentedColormap.from_list("", ["white", "red"])

plt.figure()
plt.imshow(np.arange(100).reshape((10, 10)), cmap=cmap)
plt.colorbar()
```

ericpre (changes requested):
This is in good shape, I made suggestions to simplify the code.

hyperspy/drawing/utils.py
TomSlater: I was trying to have a default that allowed any number of images (same with alphas). Maybe this should be set to list(BASE_COLORS) as default? Is that descriptive enough for the user?

hyperspy/drawing/utils.py
TomSlater: I could set to a single value of 1.0? I could also take a single float value as well as a list, if the user wants all alphas to be the same.

hyperspy/drawing/utils.py
TomSlater: True, I'll remove this argument.

hyperspy/drawing/utils.py
TomSlater: This was to always set the background to black (rather than purple if the default cmap is viridis). I could add an argument to set cmap in case the user wants a different background colour. Or I could just remove and add a note on changing the default cmap if you want a different background colour.

hyperspy/drawing/utils.py
TomSlater: My code adds a transparency to the colormaps so that you can see the different images when overlayed. If you use the code as you've written above, you will only see one image at a time (although you could play with the alphas to show others).

hyperspy/drawing/utils.py
TomSlater: I think this is needed to set the background colour (due to the transparency of the other colormaps).

hyperspy/drawing/utils.py
TomSlater: Sorry, I should have said it adds a transparency gradient.

hyperspy/drawing/utils.py
ericpre: Ok, would it be better to use `ax.imshow(np.zeros_like(images[0].data), cmap='gray')` and avoid setting the matplotlib `rcParams`?

hyperspy/drawing/utils.py
```diff
@@ -848,7 +844,13 @@ def __check_single_colorbar(cbar):
     # Set overall figure size and define figure (if not pre-existing)
     if fig is None:
         k = max(plt.rcParams['figure.figsize']) / max(per_row, rows)
-        f = plt.figure(figsize=(tuple(k * i for i in (per_row, rows))))
+        if overlay:
+            shape = images[0].data.shape
+            dpi = 100
+            sampling_factor = 10
```
ericpre: I think it would be convenient to have the `sampling_factor` as a parameter, so that user can adjust the number of pixel of the image and save it from a script very easily. Not sure that `sampling_factor` is the most appropriate name?!

hyperspy/drawing/utils.py
```diff
@@ -925,32 +927,45 @@ def check_list_length(arg, arg_name):
...
         for i, im in enumerate(images):
+            
+            #Set vmin and vmax
+            centre = next(centre_colormaps)   # get next value for centreing
+            data = im.data
+            _vmin = vmin[idx] if isinstance(vmin, (tuple, list)) else vmin
+            _vmax = vmax[idx] if isinstance(vmax, (tuple, list)) else vmax
+            _vmin, _vmax = contrast_stretching(data, _vmin, _vmax)
+            if centre:
+                _vmin, _vmax = centre_colormap_values(_vmin, _vmax)
```
ericpre: It looks to me that we should ignore `centre` when using `overlay=True`. We should raise a warning when this option is set and ignored.

hyperspy/drawing/utils.py
ericpre: I am not convinced this is worth making the background color adjustable? See other comment below to avoid changing the `rcParams`.

hyperspy/drawing/utils.py
ericpre: Ah ok, I missed that important bit! Currently, it looks a bit cumbersome the way the cmap is defined, what's about the following:
```python
import numpy as np
from matplotlib.colors import LinearSegmentedColormap, to_rgba

def transparent_single_color_cmap(color):
    """ Return a single color matplotlib cmap with the transparency increasing
    linearly from 0 to 1."""
    return LinearSegmentedColormap.from_list("", [to_rgba(color, 0), to_rgba(color, 1)])

plt.figure()
data = np.arange(100).reshape((10, 10))
plt.imshow(data, cmap=transparent_single_color_cmap('r'))
plt.imshow(-data, cmap=transparent_single_color_cmap('b'))
```

hyperspy/drawing/utils.py
```diff
@@ -600,7 +608,27 @@ def plot_images(images,
         value. It must be in the range [0, 100]
         See :py:func:`numpy.percentile` for more explanation.
         If None, use the percentiles value set in the preferences.
-        If float of integer, keep this value as bounds.
+        If float or integer, keep this value as bounds.
+    overlay : boolean, optional
+        If True, overlays the images with different colors rather than plotting
+        each image as a subplot.
+    overlay_title: boolean, optional
```
ericpre: Can it be merged with the `suptitle` parameter?

ericpre (changes requested):
This is getting very nice! I have left a few minor comments. The `vmax` works well but not the `vmin` because of the transparency applied to the minimum value.

hyperspy/drawing/utils.py
```diff
@@ -488,8 +490,14 @@ def plot_images(images,
                 fig=None,
                 vmin=None,
                 vmax=None,
+                overlay=False,
+                colors='auto',
+                alphas=1.0,
+                legend_picking=True,
+                legend_loc='upper right',
+                pixsize_factor = 10,
```
ericpre: Set its default value to 1 and rename to `pixel_size_factor` to avoid using abbreviation?

hyperspy/drawing/utils.py
```diff
@@ -600,7 +608,28 @@ def plot_images(images,
...
+        `alphas` should be a single float value or a list of floats 
+        corresponding to the alpha value of each color.
+    legend_picking: bool, optional
+        If True (default), a spectrum can be toggled on and off by clicking on
+        the legended line.
+    legend_loc : {str, int}, optional
+        This parameter controls where the legend is placed on the figure;
+        see the pyplot.legend docstring for valid values
+    pixsize_factor : int, optional
+        Sets the size of the figure when plotting an overlay image. The higher
```
ericpre, line 632: A float can also be a valid input and it would be good to specify the default value. Mention that this parameter will be ignored if `fig` if provided.

ericpre (changes requested):
A few last details in comments.
What do you want to do for the `vmin` issue? The simplest would be to ignore it and add a warning to the logger that `vmin` is ignored.

ericpre:
The test failures should be fixed in https://github.com/hyperspy/hyperspy/pull/2603.

TomSlater:
> A few last details in comments.
> What do you want to do for the `vmin` issue? The simplest would be to ignore it and add a warning to the logger that `vmin` is ignored.

I think you're right. I've just ignored it. In most cases where this is used (elemental maps), it's probably not going to be an issue.

All else should be sorted now.

PR #2600, opened 2020-12-15
Correct typo in zeta_factor_quantification

k8macarthur:
Accidental use of the same name for two variables inside and outside a for loop in the zeta_factor quantification, meant the wrong value was take forward in the function. This has now been corrected here.


thomasaarholt:
@k8macarthur you'll need to adjust the tests as well since the function is now returning different results :)

k8macarthur:
I’m on it but had to stop for childcare. :)

k8macarthur:
So this should fix the error for the zeta factors but there's still a mismatch between cross sections and zeta factors now for absorption correction which is the last remaining test. I'll try and get to the bottom of this, this week if I can. Don't really want to merge until I've fixed all the things I've broken...

k8macarthur:
Ok this should be fixed now and ready for review. I'm not quite sure it became so messy when I tried to update from the main branch. I assume next_patch is the branch to submit this to?

I only change eds utils.py tem_eds.py and test_tem_eds.py.

ericpre:
The issue is that you have merged `RELEASE_next_minor` into this branch and since `RELEASE_next_minor` is ahead of `RELEASE_next_patch`, all the commits of `RELEASE_next_minor` which are in `RELEASE_next_patch` appear in this PR. The easiest would be to start an new branch and cherry pick your commits.

k8macarthur:
Resubmitted with correct branching history in #2602

PR #2602, opened 2020-12-17
Correct typo in zeta_factor_quantification : Take 2

k8macarthur:
Accidental use of the same name for two variables inside and outside a for loop in the zeta_factor quantification, meant the wrong value was take forward in the function. This has now been corrected here.

Tests have been updated and mismatch errors in notation between cross sections and zeta factors have been fixed.

I have also added the fix for plot_result=True errors.

### DIFF
--- a/hyperspy/misc/eds/utils.py
+++ b/hyperspy/misc/eds/utils.py
@@ -654,7 +654,7 @@ def get_abs_corr_cross_section(composition, number_of_atoms, take_off_angle,
     #calculate the total_mass in kg/m^2, or mass thickness.
     total_mass = np.zeros_like(number_of_atoms[0], dtype = 'float')
     for i, (weight) in enumerate(atomic_weights):
-        total_mass += (number_of_atoms[i] * weight / Av / 1E-3 / probe_area / 1E-18)
+        total_mass += (number_of_atoms[i] * weight / Av / 1E3 / probe_area / 1E-18)
     # determine mass absorption coefficients and convert from cm^2/g to m^2/kg.
     to_stack = material.mass_absorption_mixture(
         weight_percent=material.atomic_to_weight(composition)
--- a/hyperspy/misc/eds/utils.py
+++ b/hyperspy/misc/eds/utils.py
@@ -547,10 +547,10 @@ def quantification_zeta_factor(intensities,
 
     sumzi = np.zeros_like(intensities[0], dtype='float')
     composition = np.zeros_like(intensities, dtype='float')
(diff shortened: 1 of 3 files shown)


hyperspy/tests/signal/test_eds_tem.py
```diff
@@ -390,20 +385,19 @@ def test_method_error(self):
...
                                 absorption_correction=True)
-        np.testing.assert_allclose(res[0][0].data, np.array(
-            [[49.4889, 49.4889],
-             [49.4889, 49.4889]]), atol=1e-3)
+        np.testing.assert_allclose(res[0][1].data, np.ones((2, 2)) * 44.02534,
+                                   atol=1e-3)
         np.testing.assert_allclose(res2[0][0].data, res[0][0].data, atol=1e-3)
 
-    def test_quant_zeros(self):
+    def eros(self):
```
ericpre: Typo here? This would disable the test.

hyperspy/misc/eds/utils.py
```diff
@@ -652,25 +651,21 @@ def get_abs_corr_cross_section(composition, number_of_atoms, take_off_angle,
...
     to_stack = material.mass_absorption_mixture(
         weight_percent=material.atomic_to_weight(composition)
         )
     mac = stack(to_stack, show_progressbar=False) * 0.1
-
     acf = np.zeros_like(number_of_atoms)
-    constant = 1/(Av * math.sin(toa_rad) * probe_area * 1E-16)
-
-    #determine an absorption coeficcient per element per pixel.
+    csc_toa = 1/math.sin(toa_rad)
```
ericpre, line 664: This looks like a non-negligible change in the calculation, but it doesn't affect the test results. Is it expected?

k8macarthur: So this made the estimate sample thickness so small that there was no absorption. The zeta error made it so that there was only one loop in the zeta correction. Therefore it was only serendipity that they ended up matching. When I fixed the zeta error the corresponding one in the cross sections function appeared. No I agree it's non-negligible, just luck that they matched before this.

hyperspy/tests/signal/test_eds_tem.py
k8macarthur: Oops yes.

hyperspy/misc/eds/utils.py
k8macarthur: Basically the probe_area was being incorporate twice and it shouldn't have been.

ericpre (approved):
Thanks, looks good to me!

PR #2603, opened 2020-12-21
Fix navigation mask issue with quantification of single spectrum

ericpre:
One test added in https://github.com/hyperspy/hyperspy/pull/2602 failed on `RELEASE_next_minor` because when doing the quantification of a single spectrum, a navigation mask is computed automatically and clearly it shouldn't!

### Progress of the PR
- [x] Fix setting `navigation_mask` for the quantification of single spectrum
- [x] add tests,
- [x] ready for review.


ericpre:
Doc build fails because of latest sphinx release: https://github.com/sphinx-doc/sphinx/issues/8568

PR #2615, opened 2021-01-05
Fix drawing multiple widgets

ericpre:
### Description of the change
- Fix a bug with unnecessary drawing event, which makes the widgets "sticky" when picking it (only noticeable on matplotlib backend supporting blitting)
- Speed up adding widgets for matplotlib backend supporting blitting. Before this PR, there was an unnecessary full redraw event when adding a widget, now it draw only the added widget
- Add option not to render figure when removing a widget, so that it is possible to remove many widgets without rendering. This was slow for matplotlib backend supporting blitting because of similar reason as in the previous point
- Add `pick_tolerance` to plot preferences: at the moment, this is hardcoded and the tolerance is too small on 4K screen and this is a pain!

### Progress of the PR
- [x] Fix various unnecessary drawing event 
- [x] add tests,
- [x] ready for review.

```python
import numpy as np
import hyperspy.api as hs

size = 1024
s = hs.signals.Signal2D(np.arange(size**2).reshape([size]*2))
s.plot()

offset, width = 20, 25
number = 50
position_list = [offset*i for i in range(number)]

def add_roi(s, position_list):
    roi_list = []
    for pos in position_list:
        roi = hs.roi.RectangularROI(pos, pos, pos+width, pos+width)
        roi.add_widget(s)
        roi_list.append(roi)
    return roi_list

def remove_roi(s, roi_list, render_figure=False):
    for roi in roi_list:
        roi.remove_widget(s, render_figure=render_figure)

roi_list = add_roi(s, position_list)

remove_roi(s, roi_list, render_figure=True)

```

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -8,6 +8,7 @@ What's new
 RELEASE_next_patch (Unreleased)
 +++++++++++++++++++++++++++++++
 
+* Widgets plotting improvement and add `pick_tolerance` to plot preferences (`#2615 <https://github.com/hyperspy/hyperspy/pull/2615>`_)
 
 
 Changelog
--- a/doc/user_guide/interactive_operations_ROIs.rst
+++ b/doc/user_guide/interactive_operations_ROIs.rst
@@ -108,6 +108,14 @@ added before calling :py:meth:`~.roi.BaseInteractiveROI.interactive`.
   :align:   center
   :width:   500
 
+.. NOTE::
+
+    Depending on your screen and display settings, it can be difficult to `pick`
+    or manipulate widgets and you can try to change the pick tolerance in
+    the :ref:`HyperSpy plot preferences <configuring-hyperspy-label>`.
+    Typically, using a 4K resolution with a small scaling factor (<150 %), setting
+    the pick tolerance to 15 instead of 7.5 makes the widgets easier to manipulate.
+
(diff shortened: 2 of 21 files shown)


kociak (changes requested):
The modifications are working extremely nicely, and the speed improvement is so high that it is difficult to measure it!
I noticed that there is still a lag for the ROI to follow the mouse movement, but this seems to be another issue (it happens even for a single ROI).
The only modification I can suggest is to increase the handle size - increasing pick_tolerance alone is not enough in my configuration at least:
in class `class ResizersMixin`  i would init `self.resize_pixel_size = (10, 10)`instead of `self.resize_pixel_size = (5, 5)`

ericpre:
@kociak: this PR add a setting in the plot preferences to set the pick tolerance in https://github.com/hyperspy/hyperspy/pull/2615/commits/595afe4598d611ffa9b0ff987d2715c98126fc45. Can you try to set it with or though the GUI:
```python
hs.preferences.Plot.pick_tolerance = 10
```
and this change the handle size. If not, there is something wrong somewhere!

I have added a note in the user guide.

kociak:
Simply works, my bad! I think I can now accept the PR (let's see how I do that...)

kociak (approved):
Works perfectly

PR #2617, opened 2021-01-08
Improve lazy operations and Chunking

CSSFrancis:
### Description of the change
I have been looking into the map function on lazy datasets and have started to realize that it isn't very well optimized. Currently it creates a bunch of Dask delayed objects from the Dask Array.  It then either calculates each chunk many times or operates on only one signal at a time which is pretty slow. 

This is the line which slows things down considerably...

https://github.com/hyperspy/hyperspy/blob/512d129d4959ad62f602763f953e831b649ae3a8/hyperspy/_signals/lazy.py#L539

- I changed the function so that it uses the [map_chunks](https://docs.dask.org/en/latest/array-api.html#dask.array.map_blocks) function from dask 
- Constants are not longer passed as arrays of the same size as the input signal
- Mapped arrays have the same chunking structure as the input array

### Progress of the PR
- [x] Insert map_chunks
- [x] Map Chunks when using only constants
- [x] Map Chunks with Passed Signals
- [x] Allow for Ragged shapes
- [x] Allow for Changing signal shapes
- [x] add tests,
- [x] ready for review.

I still need to work on the ragged examples and changing signals.  A lot of this code is modified from @magnunor so any input he has might be useful.  I just figured that it might be worth putting in the effort to fix this now as opposed to later.

```python
import hyperspy.api as hs
import numpy as np
import time

test_data = hs.load("testdata.hspy", lazy=True) # (100,100,128,128) ~1 Gb

def multiply(data, data2, extra=1):
    return np.multiply(data,data2)

mult = hs.signals.Signal2D(np.ones((100,100,1,1)))*2

mapped = test_data.mapold(multiply,data2=mult, extra=2, ragged=False, inplace=False)
tic = time.time()
data = mapped.compute()
toc = time.time()
print("old time:", toc-tic)

mapped = test_data.mapnew(multiply,data2=mult, extra=2, ragged=False, inplace=False)
tic = time.time()
data = mapped.compute()
toc = time.time()
print("new time:", toc-tic)
```
```
old time: 375.81656885147095
new time:8.501422882080078
```


CSSFrancis:
Some of these tests are still failing because there is a little bit less flexibility in how this function works now.  This is mostly because the output signal size has to be explicit (or ragged) and the output data type has to be explicit as well.  There is a possibility to guess at each of these things, but as the map function is more of a back end type function it might be better to allow the user to set it themselves?

I also allowed for ragged signal to be returned as lazy signals.

ericpre:
Can you please rebase on `RELEASE_next_patch`, so that it will possibly be released more quickly? See https://hyperspy.readthedocs.io/en/latest/dev_guide/git.html#changing-base-branch to rebase.

On a slightly different topic, some files seems to be deleted by mistakes!

CSSFrancis:
Okay I'm still failing on a couple of tests, I can slog through them but I think that we need to be a little bit stricter with how we use the map function.  

There are a couple of things that came up again and again when writing this and made what should have been a pretty easy fix much more complicated than it needed to be. I think some of this is older legacy code, but some of it might be newer as well.  I was going to change the documentation to make things more explicit and if someone wants to comment to make sure that I understand how the map function (should) work in hyperspy that would be nice.

1 - In particular the _map_iterate function should be protected and only called by the map function.  This helps catch cases where people are passing improper arguments to the _map_iterate function.  The biggest error being made was passing the iterating_kwarg to the the map_function not as a BaseSignal but as an array which breaks the new code and is kind of a poor workaround for not really understanding the map function operation (3)

2 - Something else that I found frustrating was passing a BaseSignal in that doesn't have the same size as the navigation axes. I handled that case by just converting that to a numpy array if the navigation size is 0 and I threw an error if that doesn't work. Maybe this should be handled by the map function but it works well now so I wouldn't really touch it.

3 - This is related to the first two issues but in general the map function has two different operating principles.  If a BaseSignal is passed to the signal and has the same shape navigation signal, then it is iterated alongside the signal.  If a NumPy array is passed then the signal is assumed to be a constant and applied to every signal.

ericpre:
> There are a couple of things that came up again and again when writing this and made what should have been a pretty easy fix much more complicated than it needed to be.

Indeed, this may need a bit of work, but it we clearly understand what is the issue, then it shouldn't be that much work. It will pay off very easily in the long term instead of using workaround here and there. Thanks for taking a stab at it!

A few comments without looking too much into details.

> I have been looking into the map function on lazy datasets and have started to realize that it isn't very well optimized. Currently it creates a bunch of Dask delayed objects from the Dask Array. It then either calculates each chunk many times or operates on only one signal at a time which is pretty slow.

Just to make sure that I don't misunderstand the issue: this applies to a subset of use of the `map` method. When the `axis` keyword is provided and the `_map_all` method is used and in this is no room for improvement?

> There is a possibility to guess at each of these things, but as the map function is more of a back end type function it might be better to allow the user to set it themselves?

Indeed, this makes sense to allow the user to set the data dtype (optional), otherwise it should be inferred.

> 1 - In particular the _map_iterate function should be protected and only called by the map function. This helps catch cases where people are passing improper arguments to the _map_iterate function. The biggest error being made was passing the iterating_kwarg to the the map_function not as a BaseSignal but as an array which breaks the new code and is kind of a poor workaround for not really understanding the map function operation (3)

If I understand correctly, the only things which matters, is that the length of the arguments to iterate on is the same as for the navigation axis. It means that it could a BaseSignal, a numpy array or a list, etc, as long as the length is correct.

> 2 - Something else that I found frustrating was passing a BaseSignal in that doesn't have the same size as the navigation axes. I handled that case by just converting that to a numpy array if the navigation size is 0 and I threw an error if that doesn't work. Maybe this should be handled by the map function but it works well now so I wouldn't really touch it.

This sounds like the flow control needs to be improved!

> 3 - This is related to the first two issues but in general the map function has two different operating principles. If a BaseSignal is passed to the signal and has the same shape navigation signal, then it is iterated alongside the signal. If a NumPy array is passed then the signal is assumed to be a constant and applied to every signal.

To avoid confusion, what do you mean with _BaseSignal is passed to the signal_?

Your started to add documentation to the dev guide, which is good! In case, there are some comments or explanation, it would be very good to add them in the user guide too. For example, how to use these efficiently and highlight the fact that some operation will be more efficient than other - as in the situation where the `axis` parameter can be provided. I am mentioning this in case, you come across along these lines.
On a similar topic, adding examples about how chunks is done, how they are process and what is happening is not only useful in this PR to make sure we understand what the issue is but it would also help user how it works!

CSSFrancis:
> Just to make sure that I don't misunderstand the issue: this applies tTo avoid confusion, what do you mean with BaseSignal is passed to the signal?o a subset of use of the `map` method. When the `axis` keyword is provided and the `_map_all` method is used and in this is no room for improvement

Honestly I'm not really sure what the ``_map_all`` function does. It seems to just apply the function to the full dataset which is perfectly acceptable as long as the function properly handles lazy dataset.  For the most part I would hope that it passes a lot of the dask processing stuff to the function, and maybe even there shouldn't be a lazy implementation of the ``map_all`` method?  

> If I understand correctly, the only things which matters, is that the length of the arguments to iterate on is the same as for the navigation axis. It means that it could a BaseSignal, a numpy array or a list, etc, as long as the length is correct.

So for the ``map`` function this isn't correct.  The map function requires that if you want a signal to iterate alongside your function it must be a Basesignal with the same navigation axes.  This is the best way to do things (in my opinion) because it allows you to be very specific with how the datasets allign and reduces errors because of how numpy, matplotlib and hyperspy order arrays.
What that Basesignal is can be as flexible as NumPy and hyperspy allow... Which is really flexible. 

For the ``map_iterate`` function you can pass an ``iterating_kwargs`` variable which is the cause of a lot of headaches for me over the last couple of days. The old behavior was to take every argument and copy it so it was the same size as the dataset.  You can kind of see why this might be a problem, especially with a large dataset you are creating a set of non lazy arguments that might be just as big as the dataset! By copying everything, it was kind of a catch all fix, allowed you to pass in numpy arrays to ``iterating_kwargs`` didn't care about navigation axes etc.  In the end though I could see it being a huge problem because it is hard to be explicit about what should/ shouldn't iterate and how to align the navgiation axes between a numpy array and a hyperspy one.    

The solution is just to tighten the restrictions we place on the ``map`` function and bring it more into line with the functions intended functionality. I think it might have just drifted mostly as a result of more people using the ``_map_iterate`` function when they probably shouldn't be. 

> To avoid confusion, what do you mean with BaseSignal is passed to the signal?

Good point, I'm not great at including examples sometimes.  

```python

def multiply(data1, value):
    # this function multiplys some array by some value. 
    return np.multiply(data1,value

import hyperspy.api as hs
import numpy as np

s = hs.signals.Signal2D(np.ones((2,3,2,3))) # 4 D dataset
s2 = np.reshape(np.arange(6),(2,3))
s2_signal = hs.signals.BaseSignal(s2)# 2D dataset all in signalaxes
s2_navigation = s2_signal.T

s.map(multiply, value=s2_navigation) # This iterates s2 alongside s1 a
s.map(multiply, value=s2_signal) # This would try to multiply s by the array s2 at every nav position
s.map(multiply, value=s2) # This would try to multiply s by the array s2 at every nav position

# now if we wanted to break things a little bit and use ``_map_iterate`` (what I wouldn't recommend) we could do the following
#This currently works but won't in the new code...
s._map_iterate(multiply, iterating_kwargs = (('value'),(s2))) # This iterates s2 alongside s1
s._map_iterate(multiply, value=s2) #  This would try to multiply s by the array s2 at every nav position
```
It might not be clear from this example why the first one is better, but from a consistency standpoint it is just better to always be dealing with Signals rather than arrays and lists. There is less ambiguity about what is being applied where.  I can allow the second set of ``_map_iterate`` functions but I kind of question if forcing people to work with Signals might be the better option.  Maybe there is a case for some object that doesn't fit into BaseSignal, but I don't really know if that case exsists.

CSSFrancis:
@ericpre I think that this should be ready to go.  The oldest version of python is failing due a problem with rechunking the data with an older version of python and Dask.  I tried with python 3.6 and Dask 2.0 to recreate the error but I can't seem to recreate it.  Maybe bumping Dask to 0.20.1 or even to a more recent release might be useful as there has been a fair bit of development on that front.

Let me know if you need any explanation of any changes.

ericpre:
From the error, it looks to me that this may be due to numpy 1.16: https://numpy.org/devdocs/release/1.16.0-notes.html#numpy-functions-now-support-overrides-with-array-function.

Can you try to bump numpy to 1.17 (the `__array_function__` protocol is enable by default in 1.17).

ericpre:
This line needs to be changed for CI: https://github.com/hyperspy/hyperspy/blob/ecdae38fe9ffe20294600c87242d592240657260/.github/workflows/tests.yml#L27

ericpre:
Also needs dask > 2.1.0 to get `amax` to work and the `np.shape(output)` may be replaces by `output.shape` to avoid increasing dask requirement further.

doc/dev_guide/lazy_computations.rst
```diff
@@ -18,15 +18,25 @@ the two arrays are indeed almost identical, the most important differences are
    shape of the result depends on the values and cannot be inferred without
    execution. Hence, few operations can be run on ``res`` lazily, and it should
    be avoided if possible.
+ - **Computations in Dask are Lazy**: This means that operations only run when
+   the ``signal.compute()`` function is called. It also means that for things
+   like plotting sections of the signal will need to be computed.
```
francisco-dlp: This sentence is not clear to me

doc/dev_guide/lazy_computations.rst
```diff
@@ -18,15 +18,25 @@ the two arrays are indeed almost identical, the most important differences are
...
 will run the function on all pixels efficiently and put the results back in the
 correct order. ``func`` is not constrained by ``dask`` and can use whatever
 code (assignment, etc.) you wish.
 
+The ``map`` function is flexible and should be able to handle most operations that
+operate on some signal. If you add a ``BaseSignal`` with the same navigation size
+as the signal it will be iterated alongside the mapped signal otherwise a keyword
+argument is assumed to be constant and is applied to every signal.
+
+There are additional protected methods ``_map_iterate`` and ``_map_all`` which are
```
francisco-dlp: So then, maybe just let's not mention then?

hyperspy/_signals/lazy.py
```diff
@@ -525,83 +532,109 @@ def _map_iterate(self,
...
-        if ragged:
-            if inplace:
-                raise ValueError("In place computation is not compatible with "
-                                  "ragged array for lazy signal.")
-            # Shape of the signal dimension will change for the each nav.
-            # index, which means we can't predict the shape and the dtype needs
-            # to be python object to support numpy ragged array
-            sig_shape = ()
-            sig_dtype = np.dtype('O')
+        """This function has two different implementations
```
francisco-dlp: This sounds like a comment rather than a docstring... It may also be unnecessary since this is true of most methods in lazy.

hyperspy/misc/utils.py
```diff
@@ -1091,6 +1091,105 @@ def func(*args):
...
+                                            **iter_dict,
+                                            **kwargs)
+    try:
+        output_array = output_array.squeeze(-1)
+    except ValueError:
+        pass
+    return output_array
+
+
+def rechunk_signal(signal,
```
francisco-dlp: The structure of this function seems counter-intuitive to me:

* This is suppose to just rechunk the signal, however it seems to also return a signal for testing? 
* Why does it take and returns a ``parameter``? Shouldn't it be handled outside of the function?
So, why not splitting this function in 2, one to get the "test" signal and another one to rechunk, both of them without a ``key`` kwarg if possible?

hyperspy/_signals/lazy.py
```diff
@@ -525,83 +532,109 @@ def _map_iterate(self,
...
+                                        str(signal.axes_manager.navigation_shape) +
+                                        '> must be consistent with the size of the mapped signal <' +
+                                        str(self.axes_manager.navigation_shape) + '>')
+                elif signal.axes_manager.navigation_shape == ():
+                    if signal.axes_manager.signal_shape == (1,):
+                        kwargs[key] = signal.data[0]  # this really isn't an iterating signal.
+                    else:
+                        kwargs[key] = signal.data
+                else:
+                    rechunked_signal, test_kwarg = rechunk_signal(signal,
```
francisco-dlp: Shouldn't rechunking be an option (with default `True` for consistency with previous versions)? The thing is that rechunking takes time, so it is not always advantageous to rechunk a signal.

hyperspy/_signals/lazy.py
```diff
@@ -525,83 +532,109 @@ def _map_iterate(self,
...
-        res = map_result_construction(
-            self, inplace, res_data, ragged, sig_shape, lazy=not ragged)
-
-        return res
+            sig = self._deepcopy_with_new_data(mapped)
+            if ragged:
+                sig.axes_manager.remove(sig.axes_manager.signal_axes)
+                sig.__class__ = LazySignal
+                sig.__init__(**sig._to_dictionary(add_models=True))
+                #if len(sig.axes_manager._axes) == 0:
```
francisco-dlp: delete commented out lines?

hyperspy/_signals/signal1d.py
```diff
@@ -454,9 +454,8 @@ def shift1D(
                 post_array = da.full(tuple(post_shape),
                                      fill_value,
                                      chunks=tuple(post_chunks))
-
                 self.data = da.concatenate((pre_array, self.data, post_array),
-                                           axis=ind)
+                                           axis=ind).rechunk({ind:-1})
```
francisco-dlp, line 458: is rechunking strictly required? If not, I would suggest to add a keyword to control whether to automatically rechunk.

hyperspy/tests/signal/test_map_method.py
```diff
@@ -49,7 +49,7 @@ def test_constant_sigma(self, parallel):
     @pytest.mark.parametrize('parallel', [True, False])
     def test_constant_sigma_navdim0(self, parallel):
         s = self.im.inav[0]
-        s.map(gaussian_filter, sigma=1, parallel=parallel, ragged=self.ragged)
+        s.map(gaussian_filter, sigma=1, parallel=parallel, ragged=self.ragged, inplace= not self.ragged)
```
francisco-dlp: remove space after ``inplace= ``

francisco-dlp (changes requested):
Except for the few suggestions in the review, this looks very good to me!

magnunor:
Looking at this pull request now. Not exactly sure when I'll be finished, might be in an evening or two.

hyperspy/misc/utils.py
```diff
@@ -1091,6 +1091,105 @@ def func(*args):
 
     return func, iterators
 
+def process_function_blockwise(data,
+                               *args,
+                               function,
+                               nav_indexes=None,
+                               output_signal_size=None,
+                               iterating_kwargs=None,
```
magnunor: Is `iterating_kwargs` used in this function?

magnunor (commented):
Some notes I made while looking through the code in the pull request (included some original `map` functions which has not been touched by this pull request).

Still need to test the new functionality on some of my datasets.

----
## `_map_all`

There is one for lazy signals, and one for non-lazy signals.

Is used for:

- Signals with 1 signal dimensions, no variables which change as a function of navigation position
- Signals with several signal dimensions, which works across all the signal axes, and not with the parallel argument.

I'm guessing this function is useful in situations where the `function` can be applied to the whole dataset directly.

## `_map_iterate`

Is used for all the other axes and signal dimensions.

Separate functions for lazy and non-lazy signals.

### lazy signal `_map_iterate`

Has checks for `iterating_kwargs`, this should (probably?) be handled by raising an error if any input signals has different navigation dimensions than `self`.

- Has a function for automatically determining output size and output dtype. Nice!
- Has functionality for automatically finding `drop_axis` and `new_axis`. Nice!

## `ragged` parameter

Is there a situation where `ragged` does not result in `dtype=np.object`? If not, I'm guessing that parameter can used in `self.map` to set the dtype for the output signal.

----

Some bigger picture thoughts after reading through most of the `map` functionality in HyperSpy. These are (probably) out-of-scope for this pull request, since they are pretty extensive!

- Currently there are different implementations for lazy and non-lazy signals.
- This is a bit risky, as we can end up with `lazy_signal.map` and `not_lazy_signal.map` returning different results.
- These should probably be combined into some common functions, maybe by all using `process_function_blockwise`. Where the non-lazy signals are passed as dask arrays via `dask.array.from_array`.

doc/dev_guide/lazy_computations.rst
CSSFrancis: I'll included an example which should clear things up.  I was just trying to give a better explanation about what happens behind the scenes with Dask.  At first I think it is confusing why it seems like a function might run instantaneously and then something like plotting takes a long(er) time.

doc/dev_guide/lazy_computations.rst
CSSFrancis: Fair point. Going forward we can just focus on the map function and then call either function from there

hyperspy/_signals/lazy.py
CSSFrancis: So the reason for all of this rechunking is for the case of some signal iterating alongside some other signal.  Those two signals have to have the same chunk structure in the navigation dimension as each other or else the map blocks function doesn't work.

If that chunk strucuture is already there I don't think it really matters if you rechunk as it shouldn't really take any time

hyperspy/_signals/signal1d.py
CSSFrancis: If the code runs as it does right now then yes it is required because you are changing the size of the array and your chunk size no longer spans your signal array.  That should never be the case, but I should really add in a line of code to catch that case.

For this function I can move the padding inside the function (It probably should be anyways) and then you don't have to deal with the rechunking.

CSSFrancis:
> Some bigger picture thoughts after reading through most of the `map` functionality in HyperSpy. These are (probably) out-of-scope for this pull request, since they are pretty extensive!
> 
> * Currently there are different implementations for lazy and non-lazy signals.
> * This is a bit risky, as we can end up with `lazy_signal.map` and `not_lazy_signal.map` returning different results.
> * These should probably be combined into some common functions, maybe by all using `process_function_blockwise`. Where the non-lazy signals are passed as dask arrays via `dask.array.from_array`.

I'm liking the idea more and more of having dask do everything behind the scenes for both lazy and non lazy signals. The biggest change there would be the multiprocessing stuff for non lazy signals, but then again I have always found that to be less than optimized (admittedly I never tried very hard to get it to work properly. ) 

If dask handles the multiprocessing for lazy and non lazy signals then it would make things much more consistent (and multiprocessing that doens't work on lazy signals is admittedly fairly useless.  I will open up an issue tracker and maybe we can start a wish list of things for the map function.

doc/dev_guide/lazy_computations.rst
francisco-dlp: The sentence that troubles me is "It also means that for things like plotting sections of the signal will need to be computed."

hyperspy/misc/utils.py
francisco-dlp: Do you agree with this?

hyperspy/_signals/lazy.py
francisco-dlp: I see. The thing is that HyperSpy chunks always contain at least one full signal in that chunk. That is fine when the signal is small, but for large signals (e.g HR images) and suitable operations it may be better to use a different chunking style. Currently, that's not possible because of the automatic re-chunking inside map. Alternatively:

1. the user could disable chunking and take the responsibility of verifying that all signals are chunk-compatible
2.``map`` could offer the option to rechunk only the "keyword signals" to make them compatible with the main signal

hyperspy/_signals/signal1d.py
francisco-dlp: Sounds good.

hyperspy/misc/utils.py
CSSFrancis: Yes, I was not terribly happy with the way I was doing things either.  I think I have a better idea now, so let me see if I can get that to work

hyperspy/misc/utils.py
CSSFrancis: No I guess it isn't.  That was a left over from a previous iteration

hyperspy/misc/utils.py
CSSFrancis: To fix this I added in a method for re chunking lazy signals.  I think this is useful for other things as well as in this situation.  This also splits the function into two separate functions depending on if there needs to be a test function for trying to guess the output signal size and datatype

hyperspy/_signals/lazy.py
```diff
@@ -131,6 +133,33 @@ def compute(self, close_file=False, show_progressbar=None, **kwargs):
 
     compute.__doc__ %= SHOW_PROGRESSBAR_ARG
 
+    def rechunk(self,
+                nav_chunks="auto",
+                sig_chunks= -1,
+                threshold=None,
+                block_size_limit=None,
+                balance=False):
```
ericpre: `balance` can only be used for dask > 2.30, see

https://github.com/hyperspy/hyperspy/blob/165e407e777b503cdbc1b3d29dcf481ac15b7e4b/hyperspy/_signals/lazy.py#L1101-L1104

CSSFrancis: Thanks! I just replaced it with kwargs... probably should have done that originally.

doc/dev_guide/lazy_computations.rst
ericpre: Agree, this is better not to mention it here. If some information needs to be added regarding these functions, it would be more suitable as comment in the code or in the docstring.

hyperspy/_signals/lazy.py
```diff
@@ -521,87 +551,98 @@ def _map_all(self, function, inplace=True, **kwargs):
...
-
-        res = map_result_construction(
-            self, inplace, res_data, ragged, sig_shape, lazy=not ragged)
-
-        return res
+            sig = self._deepcopy_with_new_data(mapped)
+            if ragged:
+                sig.axes_manager.remove(sig.axes_manager.signal_axes)
+                sig.__class__ = LazySignal
+                sig.__init__(**sig._to_dictionary(add_models=True))
```
ericpre: To assign to the correct class, replace with:
```suggestion
                sig._lazy = True
                sig._assign_subclass()
```

hyperspy/_signals/lazy.py
```diff
@@ -521,87 +551,98 @@ def _map_all(self, function, inplace=True, **kwargs):
...
-            # No signal dimension for scalar
-            if np.isscalar(one_compute):
-                sig_shape = ()
-                sig_dtype = type(one_compute)
+        # unpacking keyword arguments
+        nav_indexes = self.axes_manager.navigation_indices_in_array
+        if ragged and inplace:
+            raise ValueError("Ragged and inplace are not compatible with a lazy signal")
+        chunk_span = np.equal(self.data.chunksize, self.data.shape)
+        chunk_span = [chunk_span[i] for i in self.axes_manager.signal_indices_in_array]
```
ericpre, line 581: Why is this line necessary?

hyperspy/_signals/lazy.py
```diff
@@ -521,87 +551,98 @@ def _map_all(self, function, inplace=True, **kwargs):
...
+        if not all(chunk_span):
+            print("The chunk size needs to span the full signal size, rechunking...")
```
ericpre: As a library, we wouldn't use `print`, because this is no way to control it. In this situation, the current practise is to add this message to `_logger.info`

hyperspy/_signals/lazy.py
```diff
@@ -131,6 +133,29 @@ def compute(self, close_file=False, show_progressbar=None, **kwargs):
...
 
+    def rechunk(self,
+                nav_chunks="auto",
+                sig_chunks=-1,
+                **kwargs):
+        """Rechunks the data using the same rechunking formula from Dask expect
+        that the navigation and signal chunks are defined seperately. Note, for most
+        functions sig_chunks should remain None so that it spans the entire signal.
+
+        Parameters:
```
ericpre, line 144: The docstring needs improvement:
- see https://hyperspy.readthedocs.io/en/latest/dev_guide/writing_docs.html#docstrings for formatting
- explain default value and what it does
- explain that `kwargs` are passed to the `dask.array.rechunk` - ideally with hyperlink to the dask documentation.

hyperspy/_signals/lazy.py
```diff
@@ -131,6 +133,29 @@ def compute(self, close_file=False, show_progressbar=None, **kwargs):
...
+        nav_chunks:
+            The navigation chunk size
+        sig_chunks:
+            The signal chunk size
+        """
+        if not isinstance(sig_chunks, tuple):
```
ericpre, line 156: Ideally, this function should be used in `_lazy_data`, so that it will be easier to maintain and ensure that the rechunking is performed in a consistent manner but we can leave it for another PR.

hyperspy/_signals/signal1d.py
ericpre: This still needs to be done?

hyperspy/_signals/signal1d.py
```diff
@@ -1435,7 +1438,10 @@ def find_peaks1D_ohaver(self, xdim=None,
                          parallel=parallel,
                          max_workers=max_workers,
                          inplace=False)
-        return peaks.data
+        if peaks._lazy:
```
ericpre: Any reason to force to compute here? Would it be better to leave it to the user (or the function where it is needed) to decide when it needs to be computed?

hyperspy/misc/utils.py
```diff
@@ -1091,6 +1091,104 @@ def func(*args):
...
+        The shape of the navigation axes
+    output_signal_shape: tuple
+        The shape of the output signal.  For a ragged signal
+        this is equal to 1.
+    dtype: np.dtype
+        The datatype for the output array. For preallocating.
+        For a ragged array this is np.object.
+    iterating_kwargs: list str
+        The keys for anything that is being mapped
+        alongside the data.
```
ericpre: The docstring needs updating: it doesn't match the definition.

hyperspy/misc/utils.py
```diff
@@ -1091,6 +1091,104 @@ def func(*args):
...
+        test_ind = (0,) * len(signal.axes_manager.navigation_axes)
+        if signal.axes_manager.signal_shape == ():
+            test_kwarg[key] = signal.inav[test_ind].data[0]
+        else:
+            test_kwarg[key] = signal.inav[test_ind].data
+    signal = signal.as_lazy()
+    new_chunks = nav_chunks + (*signal.axes_manager.signal_shape,)
+    signal.data = signal.data.rechunk(new_chunks)
+    rechunked_signal = (key, signal.data)
+    return rechunked_signal, test_kwarg
```
ericpre: This seems to be a left over from previous version. If so, it should be removed.

hyperspy/misc/utils.py
```diff
@@ -1091,6 +1091,104 @@ def func(*args):
...
+
+
+def guess_output_signal_size(test_signal,
+                             function,
+                             ragged,
+                             **kwargs):
+    """This function is for guessing the output signal shape and size.
```
ericpre, line 1230: Docstring needs to follow numpy style and describe arguments - including kwargs.

hyperspy/signal.py
```diff
@@ -4492,12 +4494,22 @@ def map(
...
 
         """
-        # Sepate ndkwargs
-        ndkwargs = ()
-        for key, value in list(kwargs.items()):
-            if isinstance(value, BaseSignal):
-                ndkwargs += ((key, value),)
-
+        if self.axes_manager.navigation_shape == () and self._lazy:
+            print("Converting signal to a non-lazy signal because there are no nav dimensions")
```
ericpre: As above, don't use `print` but use `_logger.warning` maybe?

hyperspy/signal.py
```diff
@@ -4492,12 +4494,22 @@ def map(
...
+        for key in ndkeys:
+            if kwargs[key].axes_manager.navigation_shape == self.axes_manager.navigation_shape:
+                ndkwargs[key] = kwargs.pop(key)
+            elif kwargs[key].axes_manager.navigation_shape == () or kwargs[key].axes_manager.navigation_shape == (1,):
+                kwargs[key] = np.squeeze(kwargs[key].data)  # this really isn't an iterating signal.
+            else:
+                raise ValueError('the size of the navigation_shape for some kwarg:' + key + ' <' +
+                                 str(kwargs[key].axes_manager.navigation_shape) +
+                                 '> must be consistent with the size of the mapped signal <' +
+                                 str(self.axes_manager.navigation_shape) + '>')
```
ericpre: Use f-string to be more readable:
```suggestion
                raise ValueError(f'The size of the navigation_shape for the kwarg {key} '
                                f'({kwargs[key].axes_manager.navigation_shape}) '
                                 'must be consistent with the size of the mapped signal '
                                f'({self.axes_manager.navigation_shape})')
```

hyperspy/signal.py
```diff
@@ -4542,11 +4554,19 @@ def map(
             res = self._map_all(function, inplace=inplace, **kwargs)
         else:
             # Iteration over coordinates.
-            res = self._map_iterate(function, iterating_kwargs=ndkwargs,
-                                    show_progressbar=show_progressbar,
-                                    parallel=parallel, max_workers=max_workers,
-                                    ragged=ragged, inplace=inplace,
-                                    **kwargs)
+            if self._lazy:
+                res = self._map_iterate(function, iterating_kwargs=ndkwargs,
```
ericpre: Instead of writing twice almost the same, would it be better to have `output_dtype` and `output_signal_size` added to `kwargs` in case of lazy signal and write the `self._map_iterate` only once?

hyperspy/_signals/lazy.py
```diff
@@ -521,87 +551,98 @@ def _map_all(self, function, inplace=True, **kwargs):
...
-                sig_dtype = type(one_compute)
+        # unpacking keyword arguments
+        nav_indexes = self.axes_manager.navigation_indices_in_array
+        if ragged and inplace:
+            raise ValueError("Ragged and inplace are not compatible with a lazy signal")
+        chunk_span = np.equal(self.data.chunksize, self.data.shape)
+        chunk_span = [chunk_span[i] for i in self.axes_manager.signal_indices_in_array]
+        if not all(chunk_span):
+            print("The chunk size needs to span the full signal size, rechunking...")
+            self.rechunk()
```
ericpre: As mentioned by @francisco-dlp  in https://github.com/hyperspy/hyperspy/pull/2617#discussion_r593019928, it would be good to make  rechunking optional.

ericpre (changes requested):
This is in good shape but there are still some comments to address!

hyperspy/_signals/lazy.py
CSSFrancis: The ragged inplace stuff isn't really necessary... That is just a holdover from previous code where it wasn't possible.  I can replace that.

The line below that is very important.  The way the function works you have to have a chunk that spans the signal dimension.  We are basically passing the chunk to the wrapping `process_dask_chunk` function which then applies the function to the data.  If you didn't have a chunk that spanned the signal dimension then things would break anytime you tried to do anything with a signal.

hyperspy/_signals/signal1d.py
CSSFrancis: It's just that the peaks are often used in other functions and this breaks them later on.  They were being computed before but kind of behind the scenes.  Honestly the peak finding stuff is a pain and keeps breaking.  There are lots of little cases where the behavior of the map function isn't really as advertised.  I can make this change but it will break a couple of tests and maybe some people's codes... If this is just a patch release I thought that we might want to hold off until at least the next minor release to make some changes there

hyperspy/_signals/lazy.py
CSSFrancis: I wanted to just fix the problems here rather than make any larger changes which is why I kept the ragged and in place message.  That is how the code used to operate so I didn't want to break anything that might depend on it in a patch.

hyperspy/_signals/signal1d.py
ericpre: > It's just that the peaks are often used in other functions and this breaks them later on. They were being computed before but kind of behind the scenes. Honestly the peak finding stuff is a pain and keeps breaking.

This is confusing: where was it computed before?

hyperspy/_signals/lazy.py
ericpre: OK, if I understand correctly, the current implementation of `process_dask_chunk` doesn't support chunking in the signal dimension and this could be an additional improvement which should be done in a separate PR since there are no reason why it shouldn't be possible. One for #2669.

hyperspy/_signals/lazy.py
ericpre: > Ideally, this function should be used in `_lazy_data`, so that it will be easier to maintain and ensure that the rechunking is performed in a consistent manner but we can leave it for another PR.

One for #2669.

hyperspy/_signals/lazy.py
ericpre: As per https://github.com/hyperspy/hyperspy/pull/2617/files#r598809213, one for #2669 too, then!

ericpre (changes requested):
This looks great to me, two things are left:
- https://github.com/hyperspy/hyperspy/pull/2617/files#r598858237
- increase coverage as highlighted by the codecov annotation (`hyperspy/_signals/signal1d.py#L475`, `hyperspy/_signals/lazy.py#L583` and `hyperspy/_signals/lazy.py#L649`)

magnunor:
Testing some of this functionality now, and the dask array in the signal is changed when running map. Example:

```python
import dask.array as da
import hyperspy.api as hs
def test_function(image):
    return image
dask_array = da.zeros((100, 110, 120, 130), chunks=(32, 32, 32, 32))
s = hs.signals.Signal2D(dask_array).as_lazy()
print(s.data.chunksize) # (32, 32, 32, 32)
s_out = s.map(function=test_function, inplace=False)
print(s.data.chunksize) # (32, 32, 120, 130)
```

I would expect the original dask array to be unchanged with `inplace=False`.

I'll probably have more feedback the next days, as I spend some time testing this.

magnunor:
There seems to be something strange going on with `iterating_kwargs`, the values seems to be somehow correlated to the chunk size of the input signal. Example:

```python
import numpy as np
import dask.array as da
import hyperspy.api as hs
def test_function(image, value):
    print(value) # 0, 1, 2, 3, ...., 31, 0, 1, ...., 31, 0, 1, ...
    return image + value
dask_array = da.zeros((100, 110, 120, 130), chunks=(32, 32, 32, 32))
s = hs.signals.Signal2D(dask_array).as_lazy()
iter_array, _ = np.meshgrid(range(110), range(100))
s_iter = hs.signals.BaseSignal(iter_array).T
s_out = s.map(function=test_function, value=s_iter, inplace=False)
```
Plotting `s_iter`:
iter_array_s_iter
Plotting `s_out` (showing the navigation window):
iter_array_s_out

hyperspy/signal.py
```diff
@@ -4492,12 +4494,22 @@ def map(
...
-        # Sepate ndkwargs
-        ndkwargs = ()
-        for key, value in list(kwargs.items()):
-            if isinstance(value, BaseSignal):
-                ndkwargs += ((key, value),)
-
+        if self.axes_manager.navigation_shape == () and self._lazy:
+            print("Converting signal to a non-lazy signal because there are no nav dimensions")
+            self.compute()
+        # Sepate ndkwargs depending on if they are BaseSignals.
```
magnunor, line 4505: It should be noted that this actually works with _all_ signals which inherits `BaseSignal`. Thus, the current implementation works fine with `Signal2D` and `Signal1D` as well.

hyperspy/_signals/lazy.py
```diff
@@ -521,87 +556,98 @@ def _map_all(self, function, inplace=True, **kwargs):
 
     def _map_iterate(self,
                      function,
-                     iterating_kwargs=(),
+                     iterating_kwargs={},
```
magnunor: Dicts (or lists or tuples) should not be used as default values for parameters like this, as it can lead to really strange bugs due to how python handles these things: https://docs.python-guide.org/writing/gotchas/

Change it to `iterating_kwargs=None` and further down:
```python
if iterating_kwargs is None:
    iterating_kwargs = None
```

ericpre: To clarify: the issue you are referring to is related to passing mutable object to functions and that mutable will be changed by function? The _pass by reference_ versus _pass by value_ story? `list` and `dict` are indeed mutable and will be indeed changed by the functions. In case of immutable (`tuple`, `str`, `int`, etc.) object, a copy will be made when it needs to be changed - because it is immutable.

The fact that `iterating_kwargs` is dict is what is concerning, because it this is a dictionary with immutable object as values, there is no issue. One possible concern is that `mutable` object are passed to the `map` method and modified by the `function` passed to `map`.

I don't see how setting `iterating_kwargs=None` will fix this issue and the only way to make sure mutable input object are not changed when calling `map` is to create a copy of the mutable objects in one way or another. If this is done by `map` or the responsibility of the user, I would say that we should leave this consideration for #2669, as for other improvements suggestion made in this PR!

magnunor: No, what I'm referring to is using mutable _default_ parameters, which can lead to some really, really strange results.

```python
def test_add_to_dict_bad(key, value, a={}):
    a[key] = value
    return a

def test_add_to_dict_correct(key, value, a=None):
    if a is None:
        a = {}
    a[key] = value
    return a

print(test_add_to_dict_bad("key", "value")) # {'key': 'value'}
print(test_add_to_dict_bad("key2", "value2")) # {'key': 'value', 'key2': 'value2'}

print(test_add_to_dict_correct("key", "value")) # {'key': 'value'}
print(test_add_to_dict_correct("key2", "value2")) # {'key2': 'value2'}
```

Essentially, if `iterating_kwargs={}` is used, the `iterating_kwargs` you used when you previously called the function will still be there. From https://docs.python-guide.org/writing/gotchas/#mutable-default-arguments:

> Python’s default arguments are evaluated once when the function is defined, not each time the function is called (like it is in say, Ruby). This means that if you use a mutable default argument and mutate it, you will and have mutated that object for all future calls to the function as well.

This is one of the common "gotchas" in python, so I prefer using the `parameter=None`, followed by `if parameter is None: ...` trick, for list, dicts, (even tuples).

hyperspy/misc/utils.py
```diff
@@ -1091,6 +1091,96 @@ def func(*args):
 
     return func, iterators
 
+def process_function_blockwise(data,
+                               *args,
+                               function,
+                               nav_indexes=None,
+                               output_signal_size=None,
+                               block_info=None,
+                               arg_keys=None,
```
ericpre, line 1167: `arg_keys` docstring is missing.

CSSFrancis:
@magnunor 
> Testing some of this functionality now, and the dask array in the signal is changed when running map. Example:
> 
> ```python
> import dask.array as da
> import hyperspy.api as hs
> def test_function(image):
>     return image
> dask_array = da.zeros((100, 110, 120, 130), chunks=(32, 32, 32, 32))
> s = hs.signals.Signal2D(dask_array).as_lazy()
> print(s.data.chunksize) # (32, 32, 32, 32)
> s_out = s.map(function=test_function, inplace=False)
> print(s.data.chunksize) # (32, 32, 120, 130)
> ```
> 
> I would expect the original dask array to be unchanged with `inplace=False`.
> 
> I'll probably have more feedback the next days, as I spend some time testing this.

I have fixed this but it involves making a copy of the signal when doing the rechunking so that the signal chunk spans the entire signal dimension.   Hopefully that works well enough and the copying doesn't require too much memory or time.

hyperspy/_signals/lazy.py
CSSFrancis: There are a couple of pretty big problems with chunking in the signal dimension.  For something simple like adding some value to every pixel in the dataset then there are no reasons why you couldn't chunk the signal dimension.  For something more complicated like finding a peak in a 2D signal or converting from carteisian coordinates to polar coordinates this is significantly more difficult.  For the second example we would need to use [dask.array.map_overlap](https://docs.dask.org/en/latest/array-overlap.html) which complicates things a fair amount.  

So while it is possible, and we might want to consider doing this, I can't think of too many signals for some dataset that are that large.  Even a 32 bit 4096 x 4096 image is only around 70 megabytes or so which is still on the small size for 1 chunk.  Even a computer with 4 Gb of ram could easily handle that.

The case where I am maybe more interested in is the case where you have a 4D dataset and you want to do something like find a peak in 4 dimensions so you do need to use the `dask.array.map_overlap` function and possibly apply a function to both the signal and navigation dimensions.  I think that should be a different map function entirely, which allows any chunking and operates on either the entire array or a subset of the dimensions.

magnunor: I agree with @CSSFrancis on this. Assuming the chunks span the whole signal dimensions greatly reduce the complexity, and potentially avoid a whole host of possible bad behaviors. The obvious ones are the ones he already mentioned, where you need the whole signal dimension (like a full diffraction pattern), to get the useful information out.

Most of the things people want to apply via `.map`, where the signal chunking can be kept, can be directly applied via `dask.array` functions, which typically retains the chunking.

hyperspy/_signals/lazy.py
ericpre: I see indeed, this is different, but this issue doesn't occur here, because no key is added to `iterating_kwargs`. Adapting your example above, this is equivalent to:
```python
def test_add_to_dict_not_that_bad(key, value, a={}):
    for key in a.keys():
        a[key] = value
    
    return a

print(test_add_to_dict_not_that_bad("key", "value")) # {'key': 'value'}
print(test_add_to_dict_not_that_bad("key2", "value2")) # {'key2': 'value2'}

print(test_add_to_dict_not_that_bad("key", "value", {'my_key':1})) # {'my_key1': 'value'}
print(test_add_to_dict_not_that_bad("key", "value", {'my_key2':2})) # {'my_key2': 'value'}
```

The `iterating_kwargs` is also created before [being passed as argument](https://github.com/CSSFrancis/hyperspy/blob/5defd64b77f818c9478c6f790f76d16918c8428e/hyperspy/signal.py#L4501-L4512) and users are not expected to call `_map_iterate` directly.

ericpre:
> This looks great to me, two things are left:
> 
> * https://github.com/hyperspy/hyperspy/pull/2617/files#r598858237
> * increase coverage as highlighted by the codecov annotation (`hyperspy/_signals/signal1d.py#L475`, `hyperspy/_signals/lazy.py#L583` and `hyperspy/_signals/lazy.py#L649`)

@CSSFrancis, any change to address the comments above. You can see the missing coverage at https://github.com/hyperspy/hyperspy/pull/2617/checks?check_run_id=2249986920 or in the PR diff https://github.com/hyperspy/hyperspy/pull/2617/files

ericpre:
The failure on azure pipeline is not related to this PR and should be sorted soon - see https://github.com/hyperspy/hyperspy/pull/2694#issue-607634045.

hyperspy/_signals/signal1d.py
CSSFrancis: I changed this so it isn't computed.  The error was just in how the find peaks method is using custom numpy dtypes to save values like the height, width etc.  I don't know if that is really the best way to be saving these values but that might be beside the point and something to look into later. With how ragged signals are saved with a numpy object dtype selecting values beased on dtpye from a dask array no longer works and you have to compute the lazy array before acessing the value.

CSSFrancis:
@ericpre I think that this should be good now.  Let me know if there are any more changes that need to be made.

ericpre (approved):
Thanks @CSSFrancis, this looks good to me.

The failure on azure pipeline is due to a the tifffile package being broken on anaconda defaults channels - see https://github.com/AnacondaRecipes/tifffile-feedstock/pull/2.

francisco-dlp (approved):
LGTM!

PR #2620, opened 2021-01-12
Read EDS frame jeol

ericpre:
As discussed in https://github.com/hyperspy/hyperspy/pull/2488.

### Progress of the PR
- [x] Add support for reading individual frame of EDS spectrum image
- [x] add tests,
- [x] ready for review.

```python
import os
filename = os.path.join(os.path.dirname(hs.__file__), 'tests', 'io',
                        'JEOL_files', 'rawdata.ASW')

s = hs.load(filename, rebin_energy=16)
eds_SI_sum_frames = s[-1]
assert eds_SI_sum_frames.data.shape == (512, 512, 256)

s = hs.load(filename, sum_frames=False, rebin_energy=16)
eds_SI = s[-1]
assert eds_SI.data.shape == (14, 512, 512, 256)
```

### DIFF
--- a/doc/user_guide/io.rst
+++ b/doc/user_guide/io.rst
@@ -1577,6 +1577,11 @@ Extra loading arguments
 
 - ``rebin_energy`` : Factor used to rebin the energy dimension. It must be a
   multiple of the number of channels, typically 4096. (default 1)
+- ``sum_frames`` : If False, each individual frame (sweep in JEOL software jargon)
+  is loaded. Be aware that loading each individual will use a lot of memory,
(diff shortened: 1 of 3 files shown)


ericpre:
@ialxn or @sempicor, can you please review this PR?

arnduveg:
Hi @ericpre , I looked at the changes and tested it on my machine. It seems to work for me. What do you need from me ?

ericpre:
@sempicor: if you think that the changes are good and that the approach makes sense to you then that's good. Nothing more is required!

Issue #2045, opened 2018-08-30
Bug: Memory issues using `map` with LazySignal


woozey:
I have tried to use `map()` applying a self-made function to a large dataset which was loaded lazily. In this example for simplicity I've replaced custom function by `np.mean()`:
```python
>>> h = hs.load(file_names[:2635], signal_type='hologram', lazy=True, stack=True)
>>> h
<LazyHologramImage, title: holo_time_series_01_Ltz_11_5K, dimensions: (2635|3710, 3838)>
>>> mean = h.map(np.mean,
            inplace=False,
            parallel=True,
            ragged=True)
>>> mean.compute()
>>> mean
<BaseSignal, title: holo_time_series_01_Ltz_11_5K, dimensions: (2635|)>
```
I was puzzled with `ragged` argument so I've tried both `ragged=True` and `ragged=False`, it doesn't play any role here... Next I tried plotting `mean.plot()` and it already took much longer than expected (though I haven't timed it properly). Next:
```python
>>> mean.save(dir_path+'_mean')
>>> os.path.getsize(dir_path+'_mean.hspy') / 2**30 # in GB
1.9926677970215678
```
2 GB for a one dimensional data set with 2635 'float32' is certainly to much... Also the performance suggests that it really stores 2 GB for the `mean` in memory.

Any ideas?

francisco-dlp:
Probably the issue is more the ``load`` function: when loading and stacking it keeps the metadata of all the files and stacks them too. If the ``original_metadata`` of all those files is big (can be the case in e.g. dm files), then that can explain the size of the file.

Regarding the computation time, ``map`` applied to ``np.mean`` should be exactly equivalent here to ``h.mean(axis=(1, 2))``. The time it takes in lazy mode depends on the chunks. As you are lazy-stacking the signals, it'll have to read them one-by-one for processing, and that can take longer than performing the same operation on a single optimized for the purpose hdf5 file.

francisco-dlp:
Probably it'll be a good idea to add the option not to stack the  ``original_metadata`` when loading and stacking. @woozey, could you confirm if that was actually your issue?

woozey:
[...]

PR #2623, opened 2021-01-19
Lazy DictionaryTreeBrowser

### LINKED ISSUE #368: Speed up DictionaryBrowser
As discussed in #363, `DictionaryBrowser` is currently the performance bottleneck for `Signal` creation and, therefore, for indexing.

One way to speed it up is to make it lazy, i.e. the DB machinery is not run until the `__getattribute__` method is executed.

### LINKED ISSUE #2536: Large (original_) metadata significantly slowing hyperspy methods
I experience rather significant slowdowns when I use hyperspy to either index or plot signals with large `original_metadata`. On my datasets times go from 280 ms to 30 sec when plotting.

I've tried creating a minimum working example, but the slowdowns are nowhere the same, and appear to be equal to the time (40 ms) needed to copy the array I'm adding to the metadata:

```python
s = hs.signals.Signal1D([[1,2,3], [2,3,4]])
%time s.inav[0] # Wall time: 1.97 ms
s.original_metadata.test = np.random.random((80, 500, 500))
%time s.inav[0] # Wall time: 45 ms
s.original_metadata.test2 = np.random.random((80, 500, 500))
%time s.inav[0] # Wall time: 87 ms
s.metadata.test3 = np.random.random((80, 500, 500))
%time s.inav[0] # Wall time: 143 ms
```
While I'm not particularly concerned with these times - should the above be copying the `original_metadata` with`inav`, rather than just referring to the original `original_metadata`? 

Here are some examples on a <40 | 500, 500> dataset nonrigidly corrected for drift with Smart Align. SA appears to save the previously processed datasets (raw data and rigidly corrected) in the original metadata.

[Link to datasets here](https://drive.google.com/drive/folders/1-_5Pw6ql6F_wU_hBDfBTQJ4A25nyaz_b?usp=sharing), with and without the processing done by Smart Align. I've included a large (~200 MB) and a small (~16MB) test case.

The problem is "taken care of" by calling `s.original_metadata = s.original_metadata.__class__()` to overwrite it with an empty metadata object, but it would be good to handle this better in the future. In particular, `s.plot()` is very slow.

Large example (~200 MB file)
```python
s = hs.load("HAADF.dm4")
%time s.inav[0] # Wall time: 8.67 s
s.original_metadata = s.original_metadata.__class__()
%time s.inav[0] # Wall time: 4 ms
```
```python
s = hs.load("Non-rigid Aligned HAADF.dm4")
%time s.plot() # Wall time: 32.4 s
s.original_metadata = s.original_metadata.__class__()
%time s.plot() # Wall time: 287 ms
```

In the attached small file, some metadata tags take *much* longer to display than others:
`r.original_metadata.ImageList.TagGroup0.ImageTags.Parent` is instant, and `r.original_metadata.ImageList.TagGroup0.ImageTags.SmartAlign.Tx_store` takes forever.

I'm not familiar with the DictionaryTreeBrowser, so I'm not sure what to suggest.

ericpre:
Make the `DictionaryTreeBrowser` lazy by default to reduce overhead when creating signal. The lazy attributes are processed when the first attribute is accessed.

Closes #368.

### Progress of the PR
- [x] Make `DictionaryTreeBrowser` lazy by default,
- [x] update docstring,
- [x] add tests,
- [x] ready for review.

```python
from hyperspy.misc.elements import elements

%timeit DictionaryTreeBrowser(elements) # 4.2 us
%timeit DictionaryTreeBrowser(elements, lazy=False) # 26 ms
```


hyperspy/misc/utils.py
```diff
@@ -264,22 +264,55 @@ class DictionaryTreeBrowser(object):
 
     """
 
-    def __init__(self, dictionary=None, double_lines=False):
+    def __init__(self, dictionary=None, double_lines=False, lazy=True):
+        """When creating a DictionaryTreeBrowser lazily, the dictionary is
+        added to the `_lazy_attributes` attribute. The first time a lazy
+        attribute is called or the DictionaryTreeBrowser is printed, the
+        DictionaryTreeBrowser process the lazy attributes with the
```
jlaehne: processes

jlaehne (commented):
Found just a minor typo.

francisco-dlp:
I am starting to wonder if we are not taking the DTB too far. It is really a shame that something that we added in the early days  just for convenience has ended up being the major performance bottleneck in HyperSpy. At the end of the day, the main issue is that we have to process the dictionary in order to transform it into a DTB, and that takes time. A different approach would be to create a subclass of ``Dict`` with the required features. That would eliminate the need of going back and forth from DTB to dict. This is the approach taken by [addict](https://github.com/mewwts/addict/blob/master/addict/addict.py).  We cannot simply replace DTB with addict, at least not until v2.0, because we would miss many features (invalid attr name browsing, hidden attributes, tree printing...). However, would subclassing ``addict`` to add the missing features be feasible / desirable?


By the way, there is  a [related thread in StackOverflow](https://stackoverflow.com/questions/4984647/accessing-dict-keys-like-an-attribute). The top answer suggest using a new 3.7 feature to implement something like this, dataclasses.

In the meantime, I agree that this PR is a good improvement of the situation, so I think that it would be worth to merge it in time for v1.6.2. Could you merge Rnp in? I'll try to find time to review it later this week.

ericpre:
Yes, I agree, it would be good to explore more standard alternative for 2.0.

Overall, the `DictionaryTreeBrowser` has very useful features and I don't think it was touch much in many years, so it is doing good job! This performance issue is noticeable when stacking many signals having very large `original_metadata` and I guess it didn't get fixed because it was not a big issue!

jlaehne:
Does this solve #2045? Would this PR be a chance to also add the requested option for making reading of `original_metadata` optional (with default `True`).

ericpre:
No it doesn't fix the memory issue mentioned of #2045, because the `original_metadata` are still copied, the "only" speed it up.
I agree that having the option not to read `original_metadata` would be good and doing the same for `hs.stack`, shall we do this in a separate PR?

jlaehne:
Yes, maybe it makes more sense to go for a separate PR.

@francisco-dlp still wanted to have a look at this one though.

hyperspy/misc/utils.py
ericpre: Thanks done!

PR #2624, opened 2021-01-20
Allow running the test suite without the pytest-mpl plugin

ericpre:
The `pytest-mpl` plugin is used to compare the plot output with reference images and if `pytest-mpl` is not installed, the test suite fails. Sometimes, it is convenient to run the test without `pytest-mpl` installed and this PR fixes it.

### Progress of the PR
- [x] Fix test suite failure when `pytest-mpl` is not installed.
- [x] ready for review.


jlaehne (approved):
LGTM

PR #2627, opened 2021-01-25
IO documentation

jlaehne:
### Description of the change
Some corrections to the IO documentation
* Mentions of PIL are amended by pillow, and hyperspy uses imageio and not directly PIL/pillow
* Replaced/removed non working links to format definitions (Image, Ripple, MSA)
* Further minor changes in the documentation.
* Pass on `**kwds` in image reader/writer and extend docstring.

### Progress of the PR
- [X] Add handling of **kwds,
- [x] Add tests,
- [x] ready for review.

### DIFF
--- a/hyperspy/io_plugins/image.py
+++ b/hyperspy/io_plugins/image.py
@@ -38,15 +38,16 @@ writes = [(2, 0), ]
 
 # TODO Extend it to support SI
 def file_writer(filename, signal, file_format='png', **kwds):
-    """Writes data to any format supported by PIL
+    """Writes data to any format supported by imageio (PIL/pillow).
 
-        Parameters
-        ----------
-        filename: str
-        signal: a Signal instance
-        file_format : str
-            The fileformat defined by its extension that is any one supported by
-            PIL.
+    Parameters
+    ----------
+    filename: str
+    signal: a Signal instance
+    file_format : str
+        The fileformat defined by its extension that is any one supported by
+        imageio (PIL/pillow), for a list see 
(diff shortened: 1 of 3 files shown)


hyperspy/io_plugins/image.py
```diff
@@ -38,15 +38,16 @@
...
-            The fileformat defined by its extension that is any one supported by
-            PIL.
+    Parameters
+    ----------
+    filename: str
+    signal: a Signal instance
+    file_format : str
+        The fileformat defined by its extension that is any one supported by
+        imageio (PIL/pillow), for a list see 
+        https://imageio.readthedocs.io/en/stable/formats.html.
```
ericpre: While you are at it, is it worth to add passing kwds to `imageio.imwrite` and update the docstring/user guide accordingly?

ericpre (commented):
Looks good, I have one suggestion for improvement.

hyperspy/io_plugins/image.py
jlaehne: I just stumbled across the broken links, but you are right it totally makes sense. I will pass on `**kwds` both for the reader and writer.

Any idea, why the `file_format` parameter is even there if it is not passed on? Maybe some legacy issue? Normally, the format will be determined automatically, but if wanted the `format` argument can be passed as keyword argument.

ericpre: Indeed, this is legacy and it is fine to remove it.

jlaehne: I already adapted that in the updated version.

Issue #2613, opened 2020-12-31
Should chunked data keep current chunk shape by default when saving?


thomasaarholt:
If a hyperspy signal already contains chunked dask array data, it will currently not automatically save the array with those chunk shapes unless specified in the `chunks` argument to `s.save()`. Should it? I think it would be the expected behaviour to keep the chunk shape.

It would be good to clarify this in the documentation, since people typically save large files.

**Example code:**
```python
import hyperspy.api as hs
import dask.array as da

data = da.random.random((10000, 3000))
data = data.rechunk((100, 3000))
s = hs.signals.Signal1D(data)
s = s.as_lazy()
s.data # chunks: (100, 3000)

s.save('test.hspy')
s2 = hs.load('test.hspy', lazy=True)
s2.data # chunks: (40, 3000)

s.save('test2.hspy', chunks = (100, 3000))
s3 = hs.load('test2.hspy', lazy=True)
s3.data # chunks: (100, 3000)
```

ericpre:
Indeed, this should be an easy fix, are you going to make a PR?

ericpre:
Fixed in #2629.

PR #2629, opened 2021-01-27
Use dask chunks when saving lazy signal

ericpre:
Use dask chunking when saving lazy signal instead of rechunking and leave the user decide what is the suitable chunking.

### Progress of the PR
- [x] Use dask chunking when saving lazy signal,
- [x] Fix incorrect chunksize when saving EMD NCEM file and specifying chunks,
- [x] Don't rechunk when not necessary, 
- [x] update docstring,
- [x] update user guide,
- [x] add tests,
- [x] ready for review.

```python
s = Signal2D(da.zeros((50, 100, 100))).as_lazy()
s.data = s.data.rechunk([50, 25, 25])
filename = 'test_chunking_saving_lazy.hspy'
s.save(filename)
s1 = load(filename, lazy=True)
assert s.data.chunks == s1.data.chunks
```

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -18,6 +18,7 @@ RELEASE_next_patch (Unreleased)
 * Fix disconnect event when closing navigator only plot (fixes `#996 <https://github.com/hyperspy/hyperspy/issues/996>`_), (`#2631 <https://github.com/hyperspy/hyperspy/pull/2631>`_)
 * Improve plotting navigator of lazy signals, add `navigator` setter to lazy signals (`#2631 <https://github.com/hyperspy/hyperspy/pull/2631>`_)
 * Use dask chunking when saving lazy signal instead of rechunking and leave the user to decide what is the suitable chunking (`#2629 <https://github.com/hyperspy/hyperspy/pull/2629>`_)
+* Fix incorrect chunksize when saving EMD NCEM file and specifying chunks (`#2629 <https://github.com/hyperspy/hyperspy/pull/2629>`_)
 
 
 Changelog
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -17,6 +17,7 @@ RELEASE_next_patch (Unreleased)
 * Add update instructions to user guide (`#2621 <https://github.com/hyperspy/hyperspy/pull/2621>`_)
 * Fix disconnect event when closing navigator only plot (fixes `#996 <https://github.com/hyperspy/hyperspy/issues/996>`_), (`#2631 <https://github.com/hyperspy/hyperspy/pull/2631>`_)
 * Improve plotting navigator of lazy signals, add `navigator` setter to lazy signals (`#2631 <https://github.com/hyperspy/hyperspy/pull/2631>`_)
+* Use dask chunking when saving lazy signal instead of rechunking and leave the user decide what is the suitable chunking (`#2629 <https://github.com/hyperspy/hyperspy/pull/2629>`_)
 
 
 Changelog
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -17,7 +17,7 @@ RELEASE_next_patch (Unreleased)
 * Add update instructions to user guide (`#2621 <https://github.com/hyperspy/hyperspy/pull/2621>`_)
 * Fix disconnect event when closing navigator only plot (fixes `#996 <https://github.com/hyperspy/hyperspy/issues/996>`_), (`#2631 <https://github.com/hyperspy/hyperspy/pull/2631>`_)
 * Improve plotting navigator of lazy signals, add `navigator` setter to lazy signals (`#2631 <https://github.com/hyperspy/hyperspy/pull/2631>`_)
-* Use dask chunking when saving lazy signal instead of rechunking and leave the user decide what is the suitable chunking (`#2629 <https://github.com/hyperspy/hyperspy/pull/2629>`_)
+* Use dask chunking when saving lazy signal instead of rechunking and leave the user to decide what is the suitable chunking (`#2629 <https://github.com/hyperspy/hyperspy/pull/2629>`_)
 
 
 Changelog
--- a/doc/user_guide/io.rst
+++ b/doc/user_guide/io.rst
@@ -359,8 +359,8 @@ Note that currently it is not possible to pass different customised chunk shapes
 arrays contained in a signal and its metadata. Therefore, the value of ``chunks`` provided on saving
 will be applied to all arrays contained in the signal.
 
-By passing ``True`` to ``chunks`` the chunk shape is guessed using ``h5py``'s ``guess_chunks`` function
-what, for large signal spaces usually leads to smaller chunks as ``guess_chunks`` does not impose the
+By passing ``True`` to ``chunks`` the chunk shape is guessed using ``h5py``'s ``guess_chunk`` function
+what, for large signal spaces usually leads to smaller chunks as ``guess_chunk`` does not impose the
 constrain of storing at least one signal per chunks. For example, for the signal in the example above
 passing ``chunks=True`` results in ``(7, 7, 256)`` chunks.
 
--- a/doc/user_guide/io.rst
+++ b/doc/user_guide/io.rst
@@ -345,8 +345,8 @@ This feature is particularly useful when using
 
 The hyperspy HDF5 format supports chunking the data into smaller pieces to make it possible to load only part
 of a dataset at a time. By default, the data is saved in chunks that are optimised to contain at least one
-full signal shape for lazy signal, while for lazy signal, the chunking of the dask is used. It is possible to
-customise the chunk shape using the ``chunks`` keyword. 
+full signal shape for non-lazy signal, while for lazy signal, the chunking of the dask is used. It is possible to
+customise the chunk shape using the ``chunks`` keyword.
 For example, to save the data with ``(20, 20, 256)`` chunks instead of the default ``(7, 7, 2048)`` chunks
 for this signal:
 
@@ -369,7 +369,7 @@ See the `chunking section <big_data.html#Chunking>`__ under `Working with big da
 
 Extra saving arguments
 ^^^^^^^^^^^^^^^^^^^^^^^
-- ``compression``: One of ``None``, ``'gzip'``, ``'szip'``, ``'lzf'`` (default is ``'gzip'``). 
+- ``compression``: One of ``None``, ``'gzip'``, ``'szip'``, ``'lzf'`` (default is ``'gzip'``).
   ``'szip'`` may be unavailable as it depends on the HDF5 installation including it.
 
 .. note::
@@ -555,7 +555,7 @@ the "r+" mode are incompatible).
 Images
 ------
 
-HyperSpy is able to read and write data too `all the image formats 
+HyperSpy is able to read and write data too `all the image formats
 <https://imageio.readthedocs.io/en/stable/formats.html>`_ supported by
 `imageio`, which used the Python Image Library  (PIL/pillow).
 This includes png, pdf, gif etc.
@@ -890,7 +890,7 @@ Extra loading arguments
   path. Relevant for emd file version >= 0.5 where groups can be named
   'group0000', 'group0001', etc.
 - ``chunks`` : None, True or tuple. Determine the chunking of the dataset to save.
-  See the ``chunks`` arguments of the ``hspy`` file format for more details. 
+  See the ``chunks`` arguments of the ``hspy`` file format for more details.
 
 
 For files containing several datasets, the `dataset_name` argument can be
@@ -938,14 +938,14 @@ the data size in memory.
     <EDSSEMSpectrum, title: EDS, dimensions: (179, 161|4096)>]
 
 .. note::
-    
+
     FFTs made in Velox are loaded in as-is as a HyperSpy ComplexSignal2D object.
     The FFT is not centered and only positive frequencies are stored in the file.
     Lazy reading of these datasets is not supported. Making FFTs with HyperSpy
     from the respective image datasets is recommended.
 
 .. note::
-    
+
     DPC data is loaded in as a HyperSpy ComplexSignal2D object. Lazy reading of these
     datasets is not supported.
 
--- a/doc/user_guide/io.rst
+++ b/doc/user_guide/io.rst
@@ -345,7 +345,8 @@ This feature is particularly useful when using
 
 The hyperspy HDF5 format supports chunking the data into smaller pieces to make it possible to load only part
 of a dataset at a time. By default, the data is saved in chunks that are optimised to contain at least one
-full signal shape. It is possible to customise the chunk shape using the ``chunks`` keyword. 
+full signal shape for lazy signal, while for lazy signal, the chunking of the dask is used. It is possible to
+customise the chunk shape using the ``chunks`` keyword. 
 For example, to save the data with ``(20, 20, 256)`` chunks instead of the default ``(7, 7, 2048)`` chunks
 for this signal:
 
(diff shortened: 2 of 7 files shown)


thomasaarholt:
Fixes ~#2629~ #2613.

Thanks Eric! I'll have a look through this tomorrow.

ericpre:
@thomasaarholt, are you still motivated to review this PR?

magnunor:
I can review it in a couple of days (after Wednesday next week).

thomasaarholt:
Heavy workload until weekend, will take a look though.

doc/user_guide/io.rst
```diff
@@ -345,7 +345,8 @@ This feature is particularly useful when using
 
 The hyperspy HDF5 format supports chunking the data into smaller pieces to make it possible to load only part
 of a dataset at a time. By default, the data is saved in chunks that are optimised to contain at least one
-full signal shape. It is possible to customise the chunk shape using the ``chunks`` keyword. 
+full signal shape for lazy signal, while for lazy signal, the chunking of the dask is used. It is possible to
```
magnunor: I think this sentence should be:

full signal shape for signal, while for lazy signal, the chunking of the dask array is used. It is possible to

hyperspy/tests/io/test_hdf5.py
```diff
@@ -707,3 +707,20 @@ def test_save_chunks_signal_metadata():
     with tempfile.TemporaryDirectory() as tmp:
         filename = os.path.join(tmp, 'test_save_chunks_signal_metadata.hspy')
     s.save(filename, chunks=(5, 5, 10))
+
+
+def test_chunking_saving_lazy():
```
magnunor, line 715: This test should also check for the case of `s.save("filename.hspy", chunks=(3, 3, 10, 10))`, ergo where the chunk size is specified.

CHANGES.rst
```diff
@@ -17,6 +17,7 @@ RELEASE_next_patch (Unreleased)
 * Add update instructions to user guide (`#2621 <https://github.com/hyperspy/hyperspy/pull/2621>`_)
 * Fix disconnect event when closing navigator only plot (fixes `#996 <https://github.com/hyperspy/hyperspy/issues/996>`_), (`#2631 <https://github.com/hyperspy/hyperspy/pull/2631>`_)
 * Improve plotting navigator of lazy signals, add `navigator` setter to lazy signals (`#2631 <https://github.com/hyperspy/hyperspy/pull/2631>`_)
+* Use dask chunking when saving lazy signal instead of rechunking and leave the user decide what is the suitable chunking (`#2629 <https://github.com/hyperspy/hyperspy/pull/2629>`_)
```
magnunor: the user decide what is the -> the user **to** decide what is the

magnunor (commented):
Very nice functionality, which is especially useful for saving big datasets.

The error message for specifying the wrong number of dimensions is a bit cryptic, for example:

```python
import dask.array as da
import hyperspy.api as hs
s = hs.signals.Signal2D(da.zeros((6, 6, 10, 10), chunks=(3, 3, 5, 5))).as_lazy()
s.save("test.hspy", chunks=(3, 5, 5))
```

Gives the error message:

```python
h5py/_debian_h5py_serial/_hl/filters.py in rq_tuple(tpl, name)
    128             raise TypeError('"%s" argument must be None or a sequence object' % name)
    129         if len(tpl) != len(shape):
--> 130             raise ValueError('"%s" must have same rank as dataset shape' % name)
    131 
    132     rq_tuple(chunks, 'chunks')

ValueError: "chunks" must have same rank as dataset shape
```

So it should be improved to be a bit clearer.

----------------------------------

Testing saving and loading with EMD, gives the wrong chunksize when loading the data:

```python
import numpy as np
import dask.array as da
import hyperspy.api as hs
chunks = (3, 3, 5, 5)
chunks_new = (4, 4, 6, 6)
s = hs.signals.Signal2D(da.zeros((6, 6, 10, 10), chunks=chunks)).as_lazy()
s.save("test_emd_chunks_specified.emd", overwrite=True, chunks=chunks_new)
s.close_file()
s_load = hs.load("test_emd_chunks_specified.emd", lazy=True)
print(s_load.data) # dask.array<transpose, shape=(6, 6, 10, 10), dtype=float64, chunksize=(6, 6, 4, 4), chunktype=numpy.ndarray>
# The chunksize should be (4, 4, 6, 6)
```

----------------------------------

After using inav, the chunking is not correct. This is probably due to `chunksize` being used somewhere where `chunks` should have been used.

```python
import dask.array as da
import hyperspy.api as hs
s = hs.signals.Signal2D(da.zeros((6, 6, 10, 10), chunks=(3, 3, 5, 5))).as_lazy()
s = s.inav[1:, 1:]
chunks_new = s.data.chunks
s.save("test_hspy_chunks_inav.hspy", overwrite=True)
s1 = hs.load("test_hspy_chunks_inav.hspy", lazy=True)
print(s.data.chunks)  # ((2, 3), (2, 3), (5, 5), (5, 5))
print(s1.data.chunks)  # ((2, 2, 1), (2, 2, 1), (5, 5), (5, 5))
```
----------------------------------
Possibly unrelated: with how EMD files are loaded. Essentially the signal/navigation information is lost. But I guess this is due to this not being supported in the EMD format?

```python
import numpy as np
import hyperspy.api as hs
s = hs.signals.Signal2D(np.zeros((2, 3, 5, 5)))
s.save("test_emd.emd", overwrite=True)
s_load = hs.load("test_emd.emd")
# s is <Signal2D, title: , dimensions: (3, 2|5, 5)>
# s_load is <BaseSignal, title: __unnamed__, dimensions: (|5, 5, 3, 2)>
```

ericpre:
> Very nice functionality, which is especially useful for saving big datasets.
> 
> The error message for specifying the wrong number of dimensions is a bit cryptic, for example:
> 
> ```python
> import dask.array as da
> import hyperspy.api as hs
> s = hs.signals.Signal2D(da.zeros((6, 6, 10, 10), chunks=(3, 3, 5, 5))).as_lazy()
> s.save("test.hspy", chunks=(3, 5, 5))
> ```
> 
> Gives the error message:
> 
> ```python
> h5py/_debian_h5py_serial/_hl/filters.py in rq_tuple(tpl, name)
>     128             raise TypeError('"%s" argument must be None or a sequence object' % name)
>     129         if len(tpl) != len(shape):
> --> 130             raise ValueError('"%s" must have same rank as dataset shape' % name)
>     131 
>     132     rq_tuple(chunks, 'chunks')
> 
> ValueError: "chunks" must have same rank as dataset shape
> ```
> 
> So it should be improved to be a bit clearer.

This is a h5py error, and I would say that we should leave h5py to deal with this error. Maybe open a ticket there, it you think that the error message is not clear?

> Testing saving and loading with EMD, gives the wrong chunksize when loading the data:
> 
> ```python
> import numpy as np
> import dask.array as da
> import hyperspy.api as hs
> chunks = (3, 3, 5, 5)
> chunks_new = (4, 4, 6, 6)
> s = hs.signals.Signal2D(da.zeros((6, 6, 10, 10), chunks=chunks)).as_lazy()
> s.save("test_emd_chunks_specified.emd", overwrite=True, chunks=chunks_new)
> s.close_file()
> s_load = hs.load("test_emd_chunks_specified.emd", lazy=True)
> print(s_load.data) # dask.array<transpose, shape=(6, 6, 10, 10), dtype=float64, chunksize=(6, 6, 4, 4), chunktype=numpy.ndarray>
> # The chunksize should be (4, 4, 6, 6)
> ```
> 
> After using inav, the chunking is not correct. This is probably due to `chunksize` being used somewhere where `chunks` should have been used.

I don't think this is related to this PR.
 
> ```python
> import dask.array as da
> import hyperspy.api as hs
> s = hs.signals.Signal2D(da.zeros((6, 6, 10, 10), chunks=(3, 3, 5, 5))).as_lazy()
> s = s.inav[1:, 1:]
> chunks_new = s.data.chunks
> s.save("test_hspy_chunks_inav.hspy", overwrite=True)
> s1 = hs.load("test_hspy_chunks_inav.hspy", lazy=True)
> print(s.data.chunks)  # ((2, 3), (2, 3), (5, 5), (5, 5))
> print(s1.data.chunks)  # ((2, 2, 1), (2, 2, 1), (5, 5), (5, 5))
> ```
> 
> Possibly unrelated: with how EMD files are loaded. Essentially the signal/navigation information is lost. But I guess this is due to this not being supported in the EMD format?
> 
> ```python
> import numpy as np
> import hyperspy.api as hs
> s = hs.signals.Signal2D(np.zeros((2, 3, 5, 5)))
> s.save("test_emd.emd", overwrite=True)
> s_load = hs.load("test_emd.emd")
> # s is <Signal2D, title: , dimensions: (3, 2|5, 5)>
> # s_load is <BaseSignal, title: __unnamed__, dimensions: (|5, 5, 3, 2)>
> ```

Indeed, I don't think this is possible.

hyperspy/tests/io/test_hdf5.py
ericpre: This was already covered in the non lazy case, but it doesn't hurt to add a test for this in the lazy case too!

magnunor:
> This is a h5py error, and I would say that we should leave h5py to deal with this error. Maybe open a ticket there, it you think that the error message is not clear?

Maybe, but I feel that having clear error messages relevant to the user. Possible something like this in:

```python
def overwrite_dataset(group, data, key, signal_axes=None, chunks=None, **kwds):
    if chunks is None:
        if isinstance(data, da.Array):
            # For lazy dataset, by default, we use the current dask chunking
            chunks = tuple([c[0] for c in data.chunks])
        else:
            # If signal_axes=None, use automatic h5py chunking, otherwise
            # optimise the chunking to contain at least one signal per chunk
            chunks = get_signal_chunks(data.shape, data.dtype, signal_axes)
    elif chunks is True:
        pass
    else:
        if len(chunks) != data.ndim:
            raise ValueError(
                "Chunk dimensions ({0}), must be the same as the "
                "dimensions of the signal ({1})".format(len(chunks), data.ndim)
            )
```
--------------------------------

>> After using inav, the chunking is not correct. This is probably due to chunksize being used somewhere where chunks should have been used.
>
> I don't think this is related to this PR.

Yeah, I can make a separate bug report about that.

-----------------------------

I don't think the EMD chunksize "flipping" was fixed, not sure if this is relevant for this pull request. The issue is essentially that the dask array chunks are "flipped", meaning the navigation and signal chunksize is switched:

```python
import numpy as np
import dask.array as da
import hyperspy.api as hs
chunks = (3, 3, 5, 5)
chunks_new = (4, 4, 6, 6)
s = hs.signals.Signal2D(da.zeros((6, 6, 10, 10), chunks=chunks)).as_lazy()
s.save("test_emd_chunks_specified.emd", overwrite=True, chunks=chunks_new)
s.close_file()
s_load = hs.load("test_emd_chunks_specified.emd", lazy=True)
print(s_load.data) # dask.array<transpose, shape=(6, 6, 10, 10), dtype=float64, chunksize=(6, 6, 4, 4), chunktype=numpy.ndarray>
# The chunksize should be (4, 4, 6, 6)
```
---------------------------

Other than that, it looks fine!

magnunor:
>>> After using inav, the chunking is not correct. This is probably due to chunksize being used somewhere where chunks should have been used.

>> I don't think this is related to this PR.

> Yeah, I can make a separate bug report about that.

Actually, I think this is a limitation with `h5py`. Essentially, it seems like the `chunks` parameter for `create_dataset` can only handle a tuple as input. Ergo, not the nested tuple we need for using `s.data.chunks`, which looks like `((2, 3), (2, 3), (5, 5), (5, 5))`. As opposed to `s.data.chunksize`: `(3, 3, 5, 5)`

ericpre:
> > This is a h5py error, and I would say that we should leave h5py to deal with this error. Maybe open a ticket there, it you think that the error message is not clear?
> 
> Maybe, but I feel that having clear error messages relevant to the user. Possible something like this in:

My concern is that this is a bad pattern: we shall not "patch" upstream library error messages and instead we shall raise error messages only where errors actually arise. The error message that you are suggesting more or less replaces "rank" with "signal dimension" and is misleading: usually, the signal dimension refer to the dimension of the signal space. 
This point is mentioned in the [user guide](https://hyperspy.readthedocs.io/en/latest/user_guide/big_data.html#chunking) and I added it to the docstring of the `save` method.

The EMD NCEM issue is instead a different issue and not related to specifying chunks when saving. In your example, the chunks size are correct saved in the emd file. It looks to me that your example above is related to the fact that the navigation and signal dimensions are not known when opening a EMD file.

hyperspy/tests/io/test_emd.py
```diff
@@ -274,7 +273,25 @@ def test_save_and_read(self, lazy):
...
+    s.save(filename)
+    s1 = load(filename, lazy=True)
+    assert s.data.chunks == s1.data.chunks
+
+    # with chunks=True, use h5py chunking
+    s.save(filename2, chunks=True)
+    s2 = load(filename2, lazy=True)
+    assert tuple([c[0] for c in s2.data.chunks]) == (13, 25, 13)
+    s1.close_file()
+    s2.close_file()
```
magnunor, line 295: There should be a "specify chunks" test for EMD as well.

magnunor (commented):
> My concern is that this is a bad pattern: we shall not "patch" upstream library error messages and instead we shall raise error messages only where errors actually arise. The error message that you are suggesting more or less replaces "rank" with "signal dimension" and is misleading: usually, the signal dimension refer to the dimension of the signal space.
This point is mentioned in the user guide and I added it to the docstring of the save method.

Yeah, I understand the reasoning and agree that is sub-optimal having to check for errors for "upstream" libraries. My main reason is making the error messages understandable for the average user. So that if they get an error, they're able to resolve it on their own, without having to google it.

However, this specific h5py error message isn't that bad, so I'm fine with not having a specific "HyperSpy" error for this.

-------

With regards to the EMD saving, I think there is an issue with the chunks being wrong when using `chunks` parameter in `s.save(..., chunks=(1, 1, 10, 10))`. For example, when the `chunks` is not used, the chunking in the HDF5 file is correct. When the `chunks` parameter is used, it is not correct.

Example of this:

```python
import numpy as np
import h5py
import dask.array as da
import hyperspy.api as hs
chunks = (1, 1, 10, 10)
s = hs.signals.Signal2D(da.zeros((3, 3, 10, 10), chunks=chunks)).as_lazy()
s.save("test_emd_chunks_not_specified.emd", overwrite=True)
s.close_file()

chunks = (1, 1, 10, 10)
s = hs.signals.Signal2D(da.zeros((3, 3, 10, 10), chunks=chunks)).as_lazy()
s.save("test_emd_chunks_specified.emd", overwrite=True, chunks=chunks)
s.close_file()

shape_specified = h5py.File("test_emd_chunks_specified.emd")["signals/__unnamed__/data"].shape
chunks_specified = h5py.File("test_emd_chunks_specified.emd")["signals/__unnamed__/data"].chunks

shape_not_specified = h5py.File("test_emd_chunks_not_specified.emd")["signals/__unnamed__/data"].shape
chunks_not_specified = h5py.File("test_emd_chunks_not_specified.emd")["signals/__unnamed__/data"].chunks

print("Chunks specified: shape {0}. chunks {1}".format(shape_specified, chunks_specified))
print("Chunks not specified: shape {0}. chunks {1}".format(shape_not_specified, chunks_not_specified))
```

Gives:

```python
Chunks specified: shape (10, 10, 3, 3). chunks (1, 1, 10, 10)
Chunks not specified: shape (10, 10, 3, 3). chunks (10, 10, 1, 1)
```

The question is if `chunks` in `s.data.rechunk(...)` should behave the same as the same as the `chunks` in `s.save(...)`.

ericpre:
> With regards to the EMD saving, I think there is an issue with the chunks being wrong when using `chunks` parameter in `s.save(..., chunks=(1, 1, 10, 10))`. For example, when the `chunks` is not used, the chunking in the HDF5 file is correct. When the `chunks` parameter is used, it is not correct.

Ok, I see, and this should be fixed now!

> The question is if `chunks` in `s.data.rechunk(...)` should behave the same as the same as the `chunks` in `s.save(...)`.

To be consistent, I went for the existing behaviour of the `hspy` plugin.

magnunor:
LGTM!

ericpre:
Thanks for the review!

Issue #1765, opened 2017-10-19
plotting with lazy loading


gwpaterson:
When plotting lazy signals, the navigation image appears to be generated in-core, resulting in possible out-of-memory errors. Since lazy signals are immutable, I think we should be able to test if the signal is lazy, calculate the navigation image out-of-core (chunked), and store it for future use of plotting that signal without having to worry about updating it later. 

Some of this is discussed here: http://hyperspy.org/hyperspy-doc/current/user_guide/big_data.html#navigator-plot

francisco-dlp:
I think that the navigator should be computed lazily, are you sure that this is not the case?

Storing the navigator sounds like an excellent idea. Actually we could add a method e.g.``BaseSignal.set_navigator`` for that so that the user can do it manually for standard signals, and it could be performed automatically for lazy signals as you propose.

gwpaterson:
Looking in a little more detail, you may be right about calculating nav images from native hs files loaded lazily. The problem may be due to the way chunking is passed from our own file format.

ericpre:
Done in #1765.

ericpre:
The issue discussed seems to be related to situations where the chunking is inappropriate and calling the lazy signal through [`__call__`](https://github.com/hyperspy/hyperspy/blob/c373169bb3e73e3b8bd1bf7e3bec5dff39776a66/hyperspy/signal.py#L2530) trigger a large computation and huge memory usage and it looks like it compute the whole signal - which is not the case but possible as computational equivalent...

I believe this is a minimal example to reproduce:
```python
import dask.array as da
import hyperspy.api as hs
import numpy as np

chunks = (32, 32, 32, 32)
data = da.random.random((512, 512, 256, 256), chunks=chunks)
s = hs.signals.Signal2D(data).as_lazy()
s._make_lazy(rechunk=True)
s1 = s.T
s1()
```

While removing the rechunking step works as expected:
```python
import dask.array as da
import hyperspy.api as hs
import numpy as np

chunks = (32, 32, 32, 32)
data = da.random.random((512, 512, 256, 256), chunks=chunks)
s = hs.signals.Signal2D(data).as_lazy()
s1 = s.T
s1()
```

The plotting have been improved in #1765, so I will close this issue and continue in #2637 in order to narrow down the issue.

PR #2631, opened 2021-01-29
Compute navigator for lazy signal

### LINKED ISSUE #996: Error when plotting output signal from s.integrate_in_range()
In the current master, running:

``` Python
s = hs.signals.Signal(np.arange(1000).reshape(100,10))
s_i = s.integrate_in_range()
# Select some area
s_i.plot()
```

Gives the error:

``` Python
<ipython-input-21-d328d94f2a72> in <module>()
----> 1 s_i.plot()

/home/magnunor/Documents/hyperspy/development/hyperspy/signal.py in plot(self, navigator, axes_manager, **kwargs)
   3342 
   3343         self._plot.plot(**kwargs)
-> 3344         self.events.data_changed.connect(self.update_plot, [])
   3345         if self._plot.signal_plot:
   3346             self._plot.signal_plot.events.closed.connect(

/home/magnunor/Documents/hyperspy/development/hyperspy/events.py in connect(self, function, kwargs)
    307         if function in self.connected:
    308             raise ValueError("Function %s already connected to %s." %
--> 309                              (function, self))
    310         if kwargs == 'auto':
    311             spec = inspect.signature(function)

ValueError: Function <bound method Signal.update_plot of <Signal, title: , dimensions: (10|)>> already connected to <hyperspy.events.Event: Event that triggers when the data has changed: {<bound method Signal.update_plot of <Signal, title: , dimensions: (10|)>>}>
```

The plot still appears, but the Error message is not so nice.

ericpre:
Closes #996 and #1765.

### Progress of the PR
- [x] Fix disconnect event when closing _navigator only_ plot (fixes #996),
- [x] the navigator can be set with the `navigator` attribute,
- [x] store the navigator in  `metadata._HyperSpy.navigator` so that it can be used after a save/load cycle,
- [x] Fix saving signals with metadata containing a signal _and_ specifying `chunks` when saving
- [x] add `compute_navigator` to lazy signals - see example,
- [x] tranpose the navigator when suitable and necessary,
- [x] add tests,
- [x] ready for review.

```python
import dask.array as da
import hyperspy.api as h

dim = 3
N = 10    
s = hs.signals.Signal1D(da.arange(N**dim).reshape([N]*dim)).as_lazy()
s.plot()
```

Specify the number of chunks when rechunking is suitable
```python
dim = 3
N = 50    
s = hs.signals.Signal1D(da.arange(N**dim).reshape([N]*dim)).as_lazy()
s.compute_navigator(chunks_number=5)
s.plot()
```


magnunor:
Nice! I'll give this a try in (probably) a day or two.

doc/user_guide/big_data.rst
```diff
@@ -153,6 +153,118 @@ offers the following online algorithms:
...
+
+The default signal navigator is the sum of the signal across all signal
+dimensions and all but 1 or 2 navigation dimensions. If the dataset is large,
+this can take a significant amount of time to perform with every plot.
+By default, a navigator is computed with minimally required approach to obtain
+a good signal-to-noise ratio image: the sum is taken on a single chunk of the
+signal space, in order to avoid to compute the navigator for the whole dataset.
+In the following example, the signal space is divided in 25 chunks (5 along on
+each axis), and therefore computing the navigation will only be perfomed over
+a small small sbset of the whole dataset by taking the sum on only 1 chunk out
```
magnunor: typo: "a small subset"

hyperspy/_signals/lazy.py
```diff
@@ -1013,6 +1014,100 @@ def post(a):
...
+                    self.compute_navigator()
+                navigator = self.navigator
+            elif nav_dim > 2:
+                navigator = 'slider'
+        super().plot(navigator=navigator, **kwargs)
+
+    def compute_navigator(self, index=None, chunks_number=None,
+                          show_progressbar=None):
+        """
+        Compute the navigator by taking the sum over a single chunk conatained
```
magnunor: contained

hyperspy/_signals/lazy.py
```diff
@@ -1013,6 +1014,100 @@ def post(a):
...
+
+        # Get the slice of the corresponding chunk
+        signal_size = len(signal_shape)
+        signal_chunks = tuple(chunks[i-signal_size] for i in range(signal_size))
+        _logger.info(f"Signal chunks: {signal_chunks}")
+        isig_slice = get_signal_chunk_slice(index, chunks)
+
+        _logger.info(f'Computing sum over signal dimension: {isig_slice}')
+        axes = [axis.index_in_array for axis in self.axes_manager.signal_axes]
+        navigator = self.isig[isig_slice].sum(axes)
```
magnunor, line 1110: The `sum(axes)` triggers a rechunk in `def _lazy_data(...)`. Is this intentional? If not, it can be disabled with `sum(axes, rechunk=False`).

hyperspy/_signals/lazy.py
```diff
@@ -1013,6 +1014,100 @@ def post(a):
...
+            # Determine the chunk size
+            signal_chunks = da.core.normalize_chunks(
+                [int(size / cn) for cn, size in zip(chunks_number, signal_shape)],
+                shape=signal_shape
+                )
+            # Needs to reverse the chunks list to match dask chunking order
+            signal_chunks = list(signal_chunks)[::-1]
+            navigation_chunks = ['auto'] * len(self.axes_manager.navigation_shape)
+            chunks = self.data.rechunk([*navigation_chunks, *signal_chunks],
+                                       balance=True).chunks
```
magnunor: The `balance` parameter is not available in the minimum version of dask specified in setup.py. So that needs to be bumped.

hyperspy/_signals/lazy.py
```diff
@@ -1013,6 +1014,100 @@ def post(a):
...
+            elif nav_dim > 2:
+                navigator = 'slider'
+        super().plot(navigator=navigator, **kwargs)
+
+    def compute_navigator(self, index=None, chunks_number=None,
+                          show_progressbar=None):
+        """
+        Compute the navigator by taking the sum over a single chunk conatained
+        the specified coordinate. Taking the sum over a single chunk is an
+        computationally efficient approach to compute the navigator. The data
```
magnunor, line 1041: a computationally efficient

doc/user_guide/big_data.rst
```diff
@@ -153,6 +153,118 @@ offers the following online algorithms:
...
+
+    >>> data = da.random.random((100, 100, 2000, 400), chunks=('auto', 'auto', 100, 100))
+    >>> s = hs.signals.Signal2D(data).as_lazy()
+    >>> s
+    <LazySignal2D, title: , dimensions: (100, 100|400, 2000)>
+    >>> s.compute_navigator(index=0, chunks_number=(2, 10))
+    >>> s.navigator.original_metadata
+    └── sum_from = [slice(0, 200, None), slice(0, 200, None)]
+
+An alternative is to calculate the navigator separetely and store it in the
```
magnunor: separately

hyperspy/misc/array_tools.py
```diff
@@ -400,3 +402,39 @@ def numba_histogram(data, bins, ranges):
...
+
+
+def get_signal_chunk_slice(index, chunks):
+    """
+    Convenience function returning the chunk slice in signal space containing
+    the specified index.
+
+    Parameters
+    ----------
+    index : int
```
magnunor: Should be tuple

magnunor (commented):
Very nice functionality! Some comments:

Should this be going to RELEASE_next_patch, or RELEASE_next_minor?

I'm not sure if the `chunk_number` parameter in the `compute_navigator` method is the most optimal way of doing this. Testing this on one of my experimental datasets, size (512, 512, 256, 256), chunking (32, 32, 32, 32):

```python
import hyperspy.api as hs
s0 = hs.load("test_experimental_data0.hspy", lazy=True)
s0.compute_navigator(index=(0, 0), chunks_number=7) # Chunk size (37, 37). Runtime: 10.7s

s1 = hs.load("test_experimental_data1.hspy", lazy=True)
s1_nav = s1.isig[:37, :37].sum((-1, -2), rechunk=False).T
s1_nav.compute() # Runtime: 10.7s

s2 = hs.load("test_experimental_data2.hspy", lazy=True)
s2.data = s2.data.rechunk((32, 32, 37, 37))
s2_nav = s2.isig[:37, :37].sum((-1, -2), rechunk=False).T
s2_nav.compute() # Runtime: 10.7s
```

So I'm not sure if the rechunking being done in `compute_navigator` improves the performance. I don't think the reduction in the number of chunks has that big effect. (Big caveat: this might be specific for how I chunk my datasets, and that I always use `rechunk=False` to avoid rechunking happening when running various functions.) 

Using `chunk_number` with an odd number is a clever way of getting the centre of the diffraction pattern, but this might be accomplished some other way. For example by allowing for multiple indicies to be included, or a range of number, or potentially slices. However, this is more advanced things which can be added later.

This breaks the functionality of passing a signal as a navigator. For example:

```python
s1 = hs.load("test_experimental_data1.hspy", lazy=True)
s1_nav = s1.isig[:37, :37].sum((-1, -2), rechunk=False).T
s1.plot(navigator=s1_nav)
```

The plotting ignores the `navigator` parameter, and calculates the navigator with `compute_navigator`.

-------------------------------------

Saving the navigator signal inside the metadata breaks the `chunks` parameter used in `s.save`. Example:

```python
import dask.array as da
import hyperspy.api as hs

data = da.random.random((20, 20, 50, 50), chunks=(10, 10, 25, 25))
s.compute_navigator()
s.save("test.hspy", chunks=(5, 5, 10, 10))
```

Gives the error: `ValueError: "chunks" must have same rank as dataset shape`

ericpre:
> Should this be going to RELEASE_next_patch, or RELEASE_next_minor?

Initially, I was not sure and I made this PR to the `RELEASE_next_minor` branch, but considering that this is fixing more things than expected, it grows more as a bug fix than new feature PR.
 
> I'm not sure if the `chunk_number` parameter in the `compute_navigator` method is the most optimal way of doing this. Testing this on one of my experimental datasets, size (512, 512, 256, 256), chunking (32, 32, 32, 32):
> 
> ```python
> import hyperspy.api as hs
> s0 = hs.load("test_experimental_data0.hspy", lazy=True)
> s0.compute_navigator(index=(0, 0), chunks_number=7) # Chunk size (37, 37). Runtime: 10.7s
> 
> s1 = hs.load("test_experimental_data1.hspy", lazy=True)
> s1_nav = s1.isig[:37, :37].sum((-1, -2), rechunk=False).T
> s1_nav.compute() # Runtime: 10.7s
> 
> s2 = hs.load("test_experimental_data2.hspy", lazy=True)
> s2.data = s2.data.rechunk((32, 32, 37, 37))
> s2_nav = s2.isig[:37, :37].sum((-1, -2), rechunk=False).T
> s2_nav.compute() # Runtime: 10.7s
> ```
> 
> So I'm not sure if the rechunking being done in `compute_navigator` improves the performance. I don't think the reduction in the number of chunks has that big effect. (Big caveat: this might be specific for how I chunk my datasets, and that I always use `rechunk=False` to avoid rechunking happening when running various functions.)

This doesn't aim at being better than the example you have shown. Actually, the fact that this PR does as well as for your example which I would imagine you would already have optimised (?) may be a good sign? I haven't check what are the different in this specific case, if you can provide a reproducible example, this would be nice!

> Using `chunk_number` with an odd number is a clever way of getting the centre of the diffraction pattern, but this might be accomplished some other way. For example by allowing for multiple indicies to be included, or a range of number, or potentially slices. However, this is more advanced things which can be added later.

Happy to consider any alternative, I think that the current approach is simple and should be robust against a range of chunking. The only thing that I am not happy with is the name of the argument: `chunk_number` but I couldn't think of a better name.

> This breaks the functionality of passing a signal as a navigator. For example:
> 
> ```python
> s1 = hs.load("test_experimental_data1.hspy", lazy=True)
> s1_nav = s1.isig[:37, :37].sum((-1, -2), rechunk=False).T
> s1.plot(navigator=s1_nav)
> ```
>
> The plotting ignores the `navigator` parameter, and calculates the navigator with `compute_navigator`.

humm, this is surprising because by looking at the [code](https://github.com/ericpre/hyperspy/blob/96395d1292d11d97ef2a62e76494049a1af51990/hyperspy/_signals/lazy.py#L1017-L1031), I don't see why it would be ignored.

 
> Saving the navigator signal inside the metadata breaks the `chunks` parameter used in `s.save`. Example:
> 
> ```python
> import dask.array as da
> import hyperspy.api as hs
> 
> data = da.random.random((20, 20, 50, 50), chunks=(10, 10, 25, 25))
> s = hs.signals.Signal2D(data).as_lazy()
> s.compute_navigator()
> s.save("test.hspy", chunks=(5, 5, 10, 10))
> ```
> 
> Gives the error: `ValueError: "chunks" must have same rank as dataset shape`

I would rephrase this as _highlighting_ a different bug :), because it look to me that the issue is with saving signal having a signal in the metadata and specifying the chunks?

magnunor:
> I would rephrase this as highlighting a different bug :), because it look to me that the issue is with saving signal having a signal in the metadata and specifying the chunks?

I agree! I actually found the same bug when implementing https://github.com/pyxem/pyxem/pull/655, the plan was to squash it at some point, but the pyxem release got priority.

> This doesn't aim at being better than the example you have shown. Actually, the fact that this PR does as well as for your example which I would imagine you would already have optimised (?) may be a good sign? I haven't check what are the different in this specific case, if you can provide a reproducible example, this would be nice!

It is a clever solution, but introduces potential extra complexity without improving performance. Since simply slicing with `isig` provides exactly the same performance to the `rechunking`. However, it is a useful and easy way of getting the center position of for example a diffraction pattern. So we might as well keep it.

hyperspy/_signals/lazy.py
ericpre: I will be inclined to stick to the default here: through `rechunk=True` in `sum`, it will defer the rechunking to `dask` and if this doesn't play well, then it means that either we are doing something or dask is doing something, in which case we should investigate and figure out. Unless, it can be made clear that this is a bad, I would suggest to stick to the defaults. What do you think?

ericpre:
> > This doesn't aim at being better than the example you have shown. Actually, the fact that this PR does as well as for your example which I would imagine you would already have optimised (?) may be a good sign? I haven't check what are the different in this specific case, if you can provide a reproducible example, this would be nice!
> 
> It is a clever solution, but introduces potential extra complexity without improving performance. Since simply slicing with `isig` provides exactly the same performance to the `rechunking`. However, it is a useful and easy way of getting the center position of for example a diffraction pattern. So we might as well keep it.

Indeed, if this is possible to re-use the `navigator` and save it, there is no point having this method public, users should use `isig` with the `rel` syntax if they want to always get the centre of the signal space. I was thinking of using the number of chunks because to easily pick a chunking which fits well with the current one, but I think this is a bit overkill just to get a navigator...

hyperspy/_signals/lazy.py
magnunor: I am not really sure to be honest. My experience after working with dask quite a bit, is that rechunking datasets which already has a reasonable chunk size just slows things down. Keeping the chunks "intact" from the file all the way through the processing is typically the best. However, this depends on what happens with the dask array after the rechunking: if it is computed right away (like in this case), it is typically fine.

So in this case, using the default (`rechunk=True`) should be ok.

magnunor:
LGTM!

PR #2633, opened 2021-01-30
Add functions  to search DictionaryTreeBrowser (e.g. metadata)

jlaehne:
### Description of the change
Add functions 
- to search if an item exists in DictionaryTreeBrowser (e.g. metadata), 
- to return its path,
- to return its value
The naming is derived from the `has_item()` and `get_item()` functions

Moved `test_dictionary_tree_browser.py` to `misc` folder.
Additionally, improved the test coverage of other functions in misc.utils.

### Progress of the PR
- [X] integrate into existing methods `has_item` and `get_item`,
- [x] add handling parts of a path (from the back, i.e. item + preceding nodes),
- [X] update docstring (including some improvements to related docstrings),
- [x] update user guide (corrected typos in rest of file),
- [X] add tests,
- [X] increase coverage of `misc.utils`,
- [x] add lazy support,
- [X] ready for review.

```python
tree = DictionaryTreeBrowser({  "Node1": {"leaf11": 11,  "Node11": {"leaf111": 111},   })
tree.has_nested_item('leaf111')
tree.get_nested_path('leaf111') 
tree.get_nested_item('leaf111') 
```


jlaehne:
How do I add labels to a PR @ericpre @thomasaarholt , do I need extra permissions?

ericpre:
This would be useful functionality to have but I am not sure if this is a good time to add more functionality to `DictionaryTreeBrowser`. As discussed in https://github.com/hyperspy/hyperspy/pull/2623#issuecomment-793424598, we should consider more standard alternative for hyperspy 2.0: the DTB have been very useful for more than 10 years now but it may be time time to switch to something better!

What do you think?

jlaehne:
I've seen https://github.com/hyperspy/hyperspy/pull/2623#issuecomment-793424598 - but it came only after implementing this one. Couldn't we integrate this functionality anyway at the moment. Should not be a problem to reproduce in the new framework once we move to a new metadata implementation. As far as I can see, neither https://github.com/mewwts/addict nor https://github.com/Infinidat/munch have search capabilities, so it would have to either be contributed there or again implemented on the HyperSpy side.

jlaehne:
Made this PR compatible with the new lazy DTB. Note that as tests were missing, also `export` was not working with lazy.

ericpre:
```python
>>> from hyperspy.misc.utils import DictionaryTreeBrowser
>>> tree = DictionaryTreeBrowser(
    {
        "Node1": {"leaf11": 11,
                  "Node11": {"leaf111": 111},
                  },
        "Node2": {"leaf21": 21,
                  "Node21": {"leaf211": 211},
                  },
    })

>>> tree.get_nested_item('Node1.Node11')

>>> tree.get_item('Node1.Node11')
    └── leaf111 = 111
```

In the case above, is it intended that `tree.get_nested_item('Node1.Node11')` returns `None`?

From the API perspective, would it make simple to add this functionality to `get_item` and `has_item` with a keyword argument? For example, it could something along the lines of `full_path=True` to keep the existing behaviour of `get_item` and for  `full_path=False`, it will return the same output as the `get_nested_item` of this PR.

jlaehne:
> In the case above, is it intended that `tree.get_nested_item('Node1.Node11')` returns `None`?
> 
Well, the idea I had in mind was to look for a specific key and therefore looking for a chain of keys does not work.  `tree.get_nested_item('Node11')` would work.

> From the API perspective, would it make simple to add this functionality to `get_item` and `has_item` with a keyword argument? For example, it could something along the lines of `full_path=True` to keep the existing behavior of `get_item` and for `full_path=False`, it will return the same output as the `get_nested_item` of this PR.
>
Would be a possibility. In fact, could even decide automatically based on the length of `item_path` which way to go. I just did not want to touch the existing function at first.

jlaehne:
> ```python
> >>> from hyperspy.misc.utils import DictionaryTreeBrowser
> >>> tree = DictionaryTreeBrowser(
>     {
>         "Node1": {"leaf11": 11,
>                   "Node11": {"leaf111": 111},
>                   },
>         "Node2": {"leaf21": 21,
>                   "Node21": {"leaf211": 211},
>                   },
>     })
> 
> >>> tree.get_nested_item('Node1.Node11')
> 
> >>> tree.get_item('Node1.Node11')
>     └── leaf111 = 111
> ```
> 
> In the case above, is it intended that `tree.get_nested_item('Node1.Node11')` returns `None`?

Remains to be done, but could be a useful feature as we are at it.

> From the API perspective, would it make simple to add this functionality to `get_item` and `has_item` with a keyword argument? For example, it could something along the lines of `full_path=True` to keep the existing behaviour of `get_item` and for `full_path=False`, it will return the same output as the `get_nested_item` of this PR.

Implemented like proposed here.

jlaehne:
> > In the case above, is it intended that `tree.get_nested_item('Node1.Node11')` returns `None`?
> 
> Remains to be done, but could be a useful feature as we are at it.

Partial paths (from the item-key backwards) work as well now.

@ericpre could you give this PR another review?

ericpre (approved):

PR #2616, opened 2021-01-06
Fix for numpy 1.20 and set numpy>=1.16

ericpre:
In line with [NEP 29](https://numpy.org/neps/nep-0029-deprecation_policy.html) Sympy have dropped support for the numpy<1.16 in their latest release (sympy 1.7.1) and the _oldest supported version_ CI build is failing accordingly. More libraries are going to drop support for numpy <1.16 soon and we can do the same and tidy up some code required to support numpy <1.16.

### Progress of the PR
- [x] Increase numpy minimum requirement to >=1.16,
- [x] Remove code required to support numpy <1.6
- [x] Fixes protochips reader for numpy 1.20 (`np.genfromtxt` changed behaviour https://numpy.org/devdocs/release/1.20.0-notes.html#numpy-genfromtxt-now-correctly-unpacks-structured-arrays)
- [x] run test suite against upstream dev (numpy, scipy, scikit-image, scikit-learn) using https://anaconda.org/scipy-wheels-nightly
- [x] update dev guide 
- [x] add entry to `CHANGES.rst` (if appropriate),
- [x] ready for review.

PR #2636, opened 2021-02-01
Lazily load modules to improve hyperspy import times

thomasaarholt:
### Speed up `import hyperspy.api as hs`
#1480 motivates us to reduce the time required to `import hyperspy.api as hs`.

This PR aims to use `lazyasd` to lazy import a number of modules so the imports are deferred to when they are actually needed, rather than all at once. I tested these by temporarily introducing a `raise ValueError` in the `__init__.py` files of the modules below, and working through the hyperspy code until it does not raise an error on `import hyperspy.api`.

This PR reduces `hyperspy.api` import time **from 4 sec to 1.7 sec** on my computer.

### Lazily import:
- [x] pint
- [x] sympy
- [x] skimage
- [x] dask
- [x] numba
- [x] hyperspy.io_plugins
- [x] sklearn
- [x] hyperspy-guis

Most packages were straightforward. I've detailed some of the more interesting cases here:
- numba:
The @numba.jit() decorator imports numba. Having had a [chat on the numba forums](https://numba.discourse.group/t/lazily-importing-jitted-functions-without-importing-numba/487), we arrived at a simple wrapping approach (see link).
- pint:
Calling `pint.UnitRegistry()` is *slow* and should be avoided! It takes about 120ms. To generally speed things up, I call it *lazily* once in `axes.py`, and import it from there to all other uses of the function (renamed the variable everywhere to `_ureg` - happy to choose a different name).
- GUIs:
`try/except ImportError` strangely doesn't catch the ModuleNotFoundError raised when `hyperspy_gui_***` isn't present. To get around this, I check with importlib to see if it is present first.
- io_plugins:
I use the same lazy import on the io_plugins, preventing a large chunk of the hyperspy code (io) from being called until needed.

### Progress of the PR
- [x] update dev user guide 
- [ ] add entry to `CHANGES.rst` (if appropriate),
- [x] add tests,
- [x] ready for review.

To test this in the notebook, first checkout the RELEASE_next_minor branch, and then in the notebook type:
```python
%%time
import hyperspy.api as hs
```
Then switch to this PR, restart the kernel and run the above again.

### Some further thoughts
- I can try the same for scipy, but that's a bigger job, and likely earning another 200-300 ms.
- `pkg_resources` is about 180 ms, and is a dependency of at least the hyperspy extension code and `prettytable`

### DIFF
--- a/conda_environment.yml
+++ b/conda_environment.yml
@@ -8,6 +8,7 @@ dependencies:
 - h5py
 - imageio
 - ipyparallel
+- lazyasd
 - matplotlib-base
 - mrcz
 - natsort
--- a/doc/dev_guide/speeding_up_code.rst
+++ b/doc/dev_guide/speeding_up_code.rst
@@ -67,3 +67,55 @@ will recythonize all Cython code and compile it.
 
 Cythonization and compilation will also take place during continous
 integration (CI).
+
+Delaying imports to keep hyperspy imports fast
+----------------------------------------------
+
+If we eagerly import all functionality used by hyperspy during 
+``import hyperspy.api``, the import will be very slow. This can be avoided
+by lazily importing modules so that they are only imported when (if!) they
+are actually used. We can use the `lazyasd module <https://github.com/xonsh/lazyasd>`_ 
+to lazily import ``numba`` in the following way:
+
+.. code-block:: python
+
+   import importlib
+   from lazyasd import lazyobject
+
+   @lazyobject
+   def numba():
+      return importlib.import_module('numba')
+
+The ``numba`` object will function exactly like after ``import numba``, but will 
+only be actually imported into the namespace when used (e.g. when using ``numba.jit``).
+
+This will be enough for most modules. For ``@numba.jit`` decorators however, another
+step is needed. A simple numba function like below will eagerly import numba with the
+jit decorator.
+
+.. code-block:: python
+
+   @numba.jit(nopython=True)
+   def add(N):
+      "Add 1 N times"
+      n = 0
+      for i in range(N):
+         n = n + 1
+      return n
+
+To fix this, we wrap the entire function in a ``lazyobject``, and rename the function:
+
+.. code-block:: python
+
+   @lazyobject
+   def add():
+      @numba.jit(nopython=True)
+      def _add(N):
+         "Add 1 N times"
+         n = 0
+         for i in range(N):
+            n = n + 1
+         return n
+      return _add
+
+The function can now be used as ``add(100)``, will only compile once, and lazily import numba.
--- a/hyperspy/_components/error_function.py
+++ b/hyperspy/_components/error_function.py
@@ -18,7 +18,7 @@
 
 from hyperspy._components.expression import Expression
 from distutils.version import LooseVersion
-import sympy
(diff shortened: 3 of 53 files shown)


ericpre:
Is it possible to make the `lazyasd` dependency optional, because it may not be available in most linux distribution?

Following #2616, we don't need to use to specify the dask function and we can use the numpy ones, which means that we can clean up many dask import.

jlaehne:
As I also was annoyed by slow loading after kernel restarts, so I tested this directly it on my arch-linux desktop system in the office:

```
# RELEASE next minor 1st import
CPU times: user 2.12 s, sys: 277 ms, total: 2.4 s
Wall time: 3.58 s
```

```
# RELEASE next minor after kernel or jupyter restart
CPU times: user 2.09 s, sys: 114 ms, total: 2.21 s
Wall time: 2.23 s
```

```
# pr2636 after kernel or jupyter restart
CPU times: user 2.07 s, sys: 156 ms, total: 2.23 s
Wall time: 2.25 s
```

So no real gain it seems!?

jlaehne:
As I also was annoyed by slow loading after kernel restarts, I tested this directly it on my arch-linux desktop system in the office:

Edit: Wait, `lazyasd` was not properly loaded - so there is some, but not a huge improvement (about 800 ms after PC restart, otherwise 400 ms):

```
# RELEASE next minor 1st import
CPU times: user 2.12 s, sys: 277 ms, total: 2.4 s
Wall time: 3.58 s
```

```
# RELEASE next minor after kernel or jupyter restart
CPU times: user 2.09 s, sys: 114 ms, total: 2.21 s
Wall time: 2.23 s
```

```
# pr2636 after PC restart
CPU times: user 1.79 s, sys: 313 ms, total: 2.1 s
Wall time: 2.87 s
```

```
# pr2636 after kernel or jupyter restart
CPU times: user 1.69 s, sys: 256 ms, total: 1.94 s
Wall time: 1.81 s
```

I will compare on my much older Laptop later.

ericpre:
Agree, I would be careful with quoting number because it can be misleading: it will depends on the platform and on what is installed on the system (typically in case of optional dependencies of hyperspy dependencies)

thomasaarholt:
Clarify how lazyasd wasn't properly loaded? 

My benchmarking PR would have been really useful now :/ but I got a bit scared of copying over repos. Looking over that PR again, I just realised that @ericpre meant transfer _ownership_ over the repo to hyperspy, and that's a bit easier than what I thought he meant (push the repo to hyperspy).

I'll take a look at that again.

jlaehne:
> Clarify how lazyasd wasn't properly loaded?

I was in the wrong environment where `lazyasd` was not available. So of course there could be no effect of this PR.

jlaehne:
> I will compare on my much older Laptop later.

Here comes the promised comparison - so on a not so good platform the gain is indeed quite prominent:

```
# RELEASE next minor
CPU times: user 4.02 s, sys: 223 ms, total: 4.24 s
Wall time: 4.61 s
```

```
# pr2636
CPU times: user 1.4 s, sys: 117 ms, total: 1.52 s
Wall time: 1.53 s
```

thomasaarholt:
Closing in favour of #2850

PR #2647, opened 2021-02-13
Changing regrex to dictionary look up

### LINKED ISSUE #2622: find_peaks() won't work for the laplacian/difference of gaussian methods
Running `find_peaks()` in a fairly normal way:

```
 dp_rb.find_peaks(method='laplacian_of_gaussian', 
                         min_sigma=0.7,
                         max_sigma=10,
                         num_sigma=30, 
                         threshold=0.046, 
                         overlap=0.5, 
                         log_scale=False,
                         exclude_border=True)

```
Throws an error:

```python 
TraitError: The 'method' trait of a PeaksFinder2D instance must be 'Local max' or 'Max' or
'Minmax' or 'Zaefferer' or 'Stat' or 'Laplacian of Gaussian' or 'Difference of Gaussian' 
or 'Template matching', but a value of 'Laplacian of gaussian' <class 'str'> was specified.
```

From elsewhere in the same stacktrace we have:

```python
~/miniconda3/envs/miniconda-1/lib/python3.8/site-packages/hyperspy/signal_tools.py in __init__(self, signal, method, peaks, **kwargs)
   1946             self.xc_template = kwargs['template']
   1947         if method is not None:
-> 1948             self.method = method.capitalize().replace('_', ' ')
```

from which the output will be `'Laplacian of gaussian'` (as capitalize acts on only the first position of a string) - I assume that the dictionary against which it's later checked is capitalized for GUI reasons.

Thoughts?

pc494:
### Description of the change
This PR attempts to fix #2622 by applying a dictionary lookup where previously a string regrex was used. I am requesting it be pulled in as a patch.

### Progress of the PR
- [x] add tests,
- [x] ready for review.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -20,6 +20,7 @@ RELEASE_next_patch (Unreleased)
 * Use 'dask_auto' when rechunk=True in ``change_dtype`` for lazy signal (`#2645 <https://github.com/hyperspy/hyperspy/pull/2645>`_)
 * Use dask chunking when saving lazy signal instead of rechunking and leave the user to decide what is the suitable chunking (`#2629 <https://github.com/hyperspy/hyperspy/pull/2629>`_)
 * Fix incorrect chunksize when saving EMD NCEM file and specifying chunks (`#2629 <https://github.com/hyperspy/hyperspy/pull/2629>`_)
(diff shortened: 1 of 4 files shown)


ericpre (commented):
Yes, it looks good to me, I made a PR to fix other bugs and add tests, which I noticed when reviewing this PR.

ericpre:
I have noticed some other GUIs glitches, which should be easy to fix, but we can leave this for another PR.

PR #2649, opened 2021-02-14
Non-uniform-axes - fix attributes in axes type conversion

jlaehne:
### Description of the change
I realized that the conversion function between different axes types (e.g. `convert_to_non_uniform_axis`) do not properly delete the attributes of the previous axes type. Only the dictionary is updated, but the attributes are still present.

1. uniform data axis -> functional or non-uniform data axis: `scale` and `offset` should be removed
2. functional data axis -> non-uniform data axis: `_expression`, `_function`, and `x` should be removed

Also, added missing tests for axes conversion functions.

### Progress of the PR
- [x] Change implemented for case 1,
- [x] Change implemented for case 2,
- [x] add tests,
- [x] ready for review.

```python
import hyperspy.api as hs
import numpy as np
s.axes_manager.signal_axes[0].convert_to_functional_data_axis(expression='x**2')
s.scale #should fail
s.axes_manager.signal_axes[0].convert_to_non_uniform_axis()
s._expression #should fail
s.x #should fail
```

### DIFF
--- a/hyperspy/axes.py
+++ b/hyperspy/axes.py
@@ -809,6 +809,7 @@ class FunctionalDataAxis(BaseDataAxis):
         self.__init__(**d, axis=self.axis)
         del self._expression
         del self._function
+        #del self.x
         self.remove_trait('x')
 
     def crop(self, start=None, end=None):
@@ -1086,11 +1087,11 @@ class UniformDataAxis(BaseDataAxis, UnitConversion):
         super()._init_trait_listeners()
         self.__class__ = DataAxis
         d["_type"] = 'DataAxis'
-        self.remove_trait('scale')
-        self.remove_trait('offset')
+        #self.remove_trait('scale')
+        #self.remove_trait('offset')
         self.__init__(**d, axis=self.axis)
-        #del self.scale
-        #del self.offset
+        del self.scale
+        del self.offset
 
 def _serpentine_iter(shape):
     '''Similar to np.ndindex, but yields indices
--- a/hyperspy/axes.py
+++ b/hyperspy/axes.py
@@ -723,7 +723,6 @@ class DataAxis(BaseDataAxis):
 
 
 class FunctionalDataAxis(BaseDataAxis):
-    x = t.Instance(BaseDataAxis)
     def __init__(self,
                  expression,
                  x=None,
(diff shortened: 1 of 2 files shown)


jlaehne:
For the attributes which are traits, i.e. `offset` and `scale` for `DataAxis` and `x` for `FunctionalDataAxis`, neither `del self.attribute` nor `self.remove_trait('attribute')` work to remove the attributes from the object. They are just set to their default values. Any suggestions on how to solve that @ericpre or @francisco-dlp ?

ericpre:
See https://github.com/jlaehne/hyperspy/pull/4 to fix the issue.

This behaviour is not documented but it is covered by tests: https://github.com/enthought/traits/issues/1423 and https://github.com/enthought/traits/blob/8ec7fb793278c4706fe34d9096baa56e106f7dc2/traits/tests/test_dynamic_trait_definition.py#L42-L54

jlaehne:
Thanks @ericpre ! I'll check again if anything should be added to the tests and then it should be ready.

jlaehne:
There is some functionality that could be covered by additional tests, but unrelated to this PR, so I will launch a separate PR for that. This one should then be ready.

hyperspy/axes.py
```diff
@@ -1071,17 +1075,23 @@ def convert_to_functional_data_axis(self, expression, units=None, name=None, **k
...
+        self.remove_trait('scale')
+        self.remove_trait('offset')
         self.__class__ = FunctionalDataAxis
         d["_type"] = 'FunctionalDataAxis'
         self.__init__(expression=expression, x=UniformDataAxis(**this_kwargs), **d)
         self.axes_manager = axes_manager
 
     def convert_to_non_uniform_axis(self):
         d = super().get_axis_dictionary()
+        super()._init_trait_listeners()
```
ericpre: Why is this necessary?

ericpre (commented):
I have left one question.

hyperspy/axes.py
jlaehne: Good point, must have remained from one of my tries to get the traits removal working.

jlaehne: (Now removed)

PR #2652, opened 2021-03-03
binned as axes parameter instead of metadata

jlaehne:
### Description of the change
In #2299 (comment) it was proposed by @francisco-dlp that `is_binned` should become an attribute of the axis replacing `metadata.Signal.binned` as it is more of an axis than a signal property. This PR implements this change as also discussed in #2549.

The implementation here is required to progress on the `non_uniform_axes` branch (#2399).

### Progress of the PR
- [X] Change implemented,
- [x] Add handler in file-reader for signals where binned is still part of the metadata
- [X] update docstring,
- [X] update user guide,
- [X] add entry to `CHANGES.rst,
- [X] update tests,
- [x] ready for review.

```python
import hyperspy.api as hs
import numpy as np
s.axes_manager[-1].is_binned = True
```


hyperspy/component.py
```diff
@@ -1121,7 +1121,7 @@ def set_parameters_not_free(self, parameter_name_list=None):
             _parameter.free = False
 
     def _estimate_parameters(self, signal):
-        self.binned = signal.metadata.Signal.binned
+        self.binned = signal.axes_manager[-1].is_binned
```
ericpre: I think, that this can be removed, since now `axis.is_binned` is used.

hyperspy/axes.py
```diff
@@ -210,6 +210,7 @@ class DataAxis(t.HasTraits, UnitConversion):
     high_index = t.Int()
     slice = t.Instance(slice)
     navigate = t.Bool(t.Undefined)
+    is_binnned = t.Bool(t.Undefined)
```
ericpre, line 213: Default to `False`?

ericpre (changes requested):
Is it possible to add a smaller test file?

This is going to break scripts which set the binning in the metadata. We shall raise a warning in [DTB](https://github.com/hyperspy/hyperspy/blob/80499ebec48ccc74ca4d6e32188dbeb14cca325a/hyperspy/misc/utils.py#L428) to warn the user that this usage is deprecated and also check that the `binned` is not defined in the metadata. If this is the case, I guess the metadata should take priority over the `is_binned` attribute of the axis because if it is in the metadata, it means that this has been added by the user and should be considered. Even this is deprecated usage, it is still expected to work.

hyperspy/axes.py
jlaehne: The default comes in the `init` on line 225. Shouldn't that be enough?

ericpre:
Shall we enforce consistency of the `is_binned` attribute for the two signal axes of a 2D signal? If so, we may need to make `_is_binned` private and let the user set it through a public method.

hyperspy/component.py
jlaehne: Yes, had to be adapted in all the tests using `estimate_parameters` though.

jlaehne:
> Is it possible to add a smaller test file?

Yes, I removed the `original_metadata` (had just resaved `example2_v2.2.hspy`, opened #2664 to reduce the size there as well. 800 kb of metadata are not needed for testing).
 
> This is going to break scripts which set the binning in the metadata. We shall raise a warning in [DTB](https://github.com/hyperspy/hyperspy/blob/80499ebec48ccc74ca4d6e32188dbeb14cca325a/hyperspy/misc/utils.py#L428) to warn the user that this usage is deprecated and also check that the `binned` is not defined in the metadata. If this is the case, I guess the metadata should take priority over the `is_binned` attribute of the axis because if it is in the metadata, it means that this has been added by the user and should be considered. Even this is deprecated usage, it is still expected to work.

Good point. I was still wondering how to best implement the backwards compability. In the file-reader, I chose the solution to warn, but to read the property into `axis.is_binned` for all signal axes and delete the metadata item. Unfortunately, the metadata-DTB is not aware of the `axes_manager`. Otherwise, one could do a similar approach for scripts that set the metadata -- every time the user tries to set the metadata, the axis attribute would be set instead. For the moment, I only see the option to add the warning and check both metadata and axis-attribute for every binning-aware function, which could eventually be deprecated with v2.0. Any better idea?

jlaehne:
> Shall we enforce consistency of the `is_binned` attribute for the two signal axes of a 2D signal? If so, we may need to make `_is_binned` private and let the user set it through a public method.

I have been pondering about that. An advantage, but also a limit in the current implementation - depending on the viewpoint. In the legacy reader, I decided to set `is_binned` for all signal axes to the value given in the metadata.

Most applications for 2DSignals (images type signals) will feature similar signal axes. However, that might not always be the case in the future. If I think of streak camera data (time resolved luminescence spectroscopy), which has two signal axes. One of them is time and one wavelength/energy. No reason to enforce consistency in something like binning for such a 2DSignal, I would say.

Also, is binning so far relevant for 2DSignals, or mostly used in 1DSignals like EELS and EDX?

hyperspy/axes.py
ericpre: Yes, this is working too. I suggested this, thinking that it would be more explicit, but actually, it can be misleading too, if in the future, it is not consistent with the default value defined in in `__init__`...

ericpre:
> Good point. I was still wondering how to best implement the backwards compability. In the file-reader, I chose the solution to warn, but to read the property into `axis.is_binned` for all signal axes and delete the metadata item. Unfortunately, the metadata-DTB is not aware of the `axes_manager`. Otherwise, one could do a similar approach for scripts that set the metadata -- every time the user tries to set the metadata, the axis attribute would be set instead. For the moment, I only see the option to add the warning and check both metadata and axis-attribute for every binning-aware function, which could eventually be deprecated with v2.0. Any better idea?

Yes, I agree, the only thing we can do is here:
- warn when the `metadata.Signal.binned` attribute is set through raising a warning in [`__setattr__`](https://github.com/hyperspy/hyperspy/blob/80499ebec48ccc74ca4d6e32188dbeb14cca325a/hyperspy/misc/utils.py#L428) method of the DTB.
- add a function which parse the binning information: it should first check if  `metadata.Signal.binned` exists and use that value and if not, it would fall back to the `DataAxis`, which would present in any case.

> I have been pondering about that. An advantage, but also a limit in the current implementation - depending on the viewpoint. In the legacy reader, I decided to set `is_binned` for all signal axes to the value given in the metadata.
> 
> Most applications for 2DSignals (images type signals) will feature similar signal axes. However, that might not always be the case in the future. If I think of streak camera data (time resolved luminescence spectroscopy), which has two signal axes. One of them is time and one wavelength/energy. No reason to enforce consistency in something like binning for such a 2DSignal, I would say.
> 
> Also, is binning so far relevant for 2DSignals, or mostly used in 1DSignals like EELS and EDX?

Yes, make sense! Maybe, we could add a note in the user guide, about the fact that both axis have independent `is_binned` value. On this topic, I don't think we should add a convenience method to the `AxesManager` to set it.

jlaehne:
backwards compatibility implemented and userguide updated.

hyperspy/signal.py
```diff
@@ -4773,6 +4776,7 @@ def __deepcopy__(self, memo):
         for oaxis, caxis in zip(self.axes_manager._axes,
                                 dc.axes_manager._axes):
             caxis.navigate = oaxis.navigate
+            caxis.is_binned = oaxis.is_binned
```
ericpre: This shouldn't be necessary, if this is not the case, there is a bug.

hyperspy/misc/utils.py
```diff
@@ -1226,3 +1232,19 @@ def is_hyperspy_signal(input_object):
...
     return isinstance(input_object,BaseSignal)
+
+
+def is_binned(signal, axis=-1):
+    """Backwards compatibility check utility for is_binned attribute.
+    
+    Can be removed in v2.0.
+    """
+    if signal.metadata.has_item('Signal.binned'):
+        warnings.warn('Use of the `binned` attribute in metadata is going to '
```
ericpre: I think we should remove this warning here, because when doing interactive operation we will get too many of them, for example:

```python
import hyperspy.api as hs
s.metadata.Signal.binned = True # warning well

s.remove_background(fast=True) # Far too many warning are displayed when moving the span selector
```

doc/user_guide/model.rst
```diff
@@ -28,8 +28,8 @@ Caveats
 -------
 
 * Before creating a model verify that the
-  :py:attr:`~.signal.BaseSignal.metadata.Signal.binned` metadata
-  attribute of the signal is set to the correct value because the resulting
+  :py:attr:`~.signal.BaseSignal.axes_manager.signal_axes[0].is_binned`attribute
```
ericpre: space missing:
```suggestion
  :py:attr:`~.signal.BaseSignal.axes_manager.signal_axes[0].is_binned` attribute
```

src/pip-delete-this-directory.txt
```diff
@@ -0,0 +1,5 @@
+This file is placed here by pip to indicate the source was put
```
ericpre: This file should be removed.

ericpre (commented):
I have added a few comments. Other than that, this looks good to me!

hyperspy/signal.py
jlaehne: OK, I thought it would have to be handled same as navigate.

hyperspy/misc/utils.py
jlaehne: An option could be to use a `DeprecationWarning` instead of `VisibleDeprecationWarning`, but I can also remove it altogether.

doc/user_guide/model.rst
jlaehne: nice catch ;-)

hyperspy/misc/utils.py
jlaehne: What do you prefer @ericpre?

ericpre: I think the warning in DTB is good enough and is acting at the right place and I would remove this warning. Since the binned information is set correctly when loading files, this is only when an user sets it in the metadata that it would this situation would occur, which case, a warning would already have been raised.

jlaehne: good point

ericpre (approved):

PR #2659, opened 2021-03-09
Improve load documentation

### LINKED ISSUE #2638: Improve load function example in user guide
There should be an example in the user guide showing that the `load` function returns a list of signal instead of a single signal, where several datasets are present in the file.

jlaehne:
### Description of the change
Improving, correcting, updating docstring and documentation of `hs.load()` starting from request in #2638

### Progress of the PR
- [X] ready for review.

### DIFF
--- a/hyperspy/io.py
+++ b/hyperspy/io.py
@@ -503,7 +503,8 @@ def load_with_reader(
 
 
 def assign_signal_subclass(dtype, signal_dimension, signal_type="", lazy=False):
-    """Given dtype, signal_dimension and signal_type, return the matching Signal subclass.
+    """Given dtype, signal_dimension and signal_type, return the matching
+    Signal subclass.
 
     See `hs.print_known_signal_types()` for a list of known signal_types,
     and the developer guide for details on how to add new signal_types.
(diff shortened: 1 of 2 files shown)


hyperspy/io.py
```diff
@@ -512,10 +514,10 @@ def assign_signal_subclass(dtype, signal_dimension, signal_type="", lazy=False):
...
-        Signal type. Optional. Will log a warning if it is unknown to HyperSpy.
-    lazy : bool, default False
-        If True, returns the matching LazySignal subclass.
+    signal_type : str, optional
+        Signal type. Default ''. Will log a warning if it is unknown to HyperSpy.
+    lazy : bool, optional
+        If True, returns the matching LazySignal subclass. Default is False.
 
     Returns
     -------
```
ericpre, line 524: Add `list of signal`?

jlaehne: Can `assign_signal_subclass` return a list? Do not see the point, because it is only called by `dict2signal` and also that one seems to operate on single signals only. But it does not return a `Signal or subclass`, but `Signal class or subclass`

ericpre: Sorry, I misread... I thought that this was still the docstring of the `load` function!

ericpre (approved):

PR #2663, opened 2021-03-10
Add python3.9

ericpre:
### Progress of the PR
- [x] Explicitly support python 3.9,
- [x] add python 3.9 to CI matrix,
- [x] bump minimum requirement of numpy to 1.17.1 and dask to 2.1.0 to be able to use the `__array__` protocol,
- [x] Add pre-release build to CI matrix to run the test suite against dependencies pre-released on pypi,
- [x] fix deprecation matplotlib 3.4.0rc1,
- [x] tidy up code (testing convenience functions) which is not necessary anymore,  
- [x] ready for review.


.github/workflows/tests.yml
```diff
@@ -38,13 +38,23 @@ jobs:
...
+          - os: ubuntu
+            PYTHON_VERSION: 3.8
+            PIP_SELECTOR: '[all, tests]'
+            LABEL: -dependencies_pre_release
+            PIP_ARGS: '--upgrade --pre'
+          - os: ubuntu
+            PYTHON_VERSION: 3.9
+            PIP_SELECTOR: '[all, tests]'
+            LABEL: -dependencies_pre_release
+            PIP_ARGS: '--upgrade --pre'
```
ericpre: Once numba 0.53 is released, this needs to be removed.

jlaehne: Why is it also in lines 51&52 then?

ericpre: I have moved removed these build to https://github.com/hyperspy/hyperspy-extensions-list/blob/ee1a24abc213c270c2e6c1f9bfcea694f0eeeab1/.github/workflows/tests.yml#L27-L38. Initially, to test python 3.9, I needed the pre-release version of numba from pypi and when doing this, I realised that now quite a few libraries are doing some pre-release which is useful to identify regression and report it so that it get sorted for the release.

jlaehne:
Coverage of `misc/test_utils.py` is decreasing, can that be avoided?

ericpre:
It looks to me that the decrease is due to the removal of many lines, which were previously covered: the absolute decreases a bit. There are a few lines not covered in this file, these are an [exception of the `assert_deep_almost_equal `](https://codecov.io/gh/hyperspy/hyperspy/src/add_python3.9/hyperspy/misc/test_utils.py). I am not even convinced that this is worth keeping this function, since the pytest output of failure when comparing nested structure is quite good nowadays! And it can be controlled with the verbosity.

jlaehne (approved):

jlaehne:
corrected some typos in the added documentation. Otherwise LGTM.

PR #2670, opened 2021-03-12
Cupy support

ericpre:
I pushed the `cupy_support` branch to `hyperspy/hyperspy` to give it visibility and so that others can make pull request to this branch. Also this PR should give an overview of what needs to be changed. I have used this branch to explore how hyperspy needs to be changed to work with cupy array. In short: not much, if we use the numpy >=1.20 and the experimental `like` keyword.

### Working (AFAIK!)
- [x] plotting,
- [x] `map` function,
- [x] ROI

### General consideration
- even for operation which are not particularly suitable for GPU computation, it is easy to get a 10x speed up.
- fitting with data on GPU works but doesn't use the GPU... This should be possible once `sympy` adds support for a [`cupy` backend](https://github.com/sympy/sympy/pull/20622).
- mask are not supported by `cupy` and where possible we should use `nan` instead - as this is done for example for the `CircleROI`.

### Relevant links:
https://numba.readthedocs.io/en/latest/cuda/cuda_array_interface.html
https://data-apis.github.io/array-api/latest
https://numpy.org/neps/nep-0035-array-creation-dispatch-with-array-function.html (see [numpy.zeros](https://numpy.org/doc/stable/reference/generated/numpy.zeros.html) docstring, for example)

### Progress of the PR:
- [x] change implemented,
- [x] add tests,
- [x] ready for review.

### DIFF
--- a/doc/user_guide/big_data.rst
+++ b/doc/user_guide/big_data.rst
@@ -264,6 +264,40 @@ instead:
     <LazySignal2D, title: , dimensions: (200, 200|512, 512)>
     >>> s.plot(navigator='slider')
 
+.. versionadded:: 1.7
+
+.. _big_data.gpu:
+
+GPU support
+-----------
+
+Lazy data processing using GPU requires setting the dask scheduler
+and explicitely transfering the data to the GPU.
+
+.. code-block:: python
+
+    >>> from dask_cuda import LocalCUDACluster
+    >>> from dask.distributed import Client
+    >>> import cupy as cp
+    >>> cluster = LocalCUDACluster()
+    >>> client = Client(cluster)
+    >>> # Create a dask array
+    >>> data = da.random.random(size=(20, 20, 100, 100))
+    >>> print(data)
+    ... dask.array<random_sample, shape=(20, 20, 100, 100), dtype=float64,
+    ... chunksize=(20, 20, 100, 100), chunktype=numpy.ndarray>
+    >>> # convert the dask chunks from numpy array to cupy array
+    >>> data = data.map_blocks(cp.asarray)
+    >>> print(data)
+    ... dask.array<random_sample, shape=(20, 20, 100, 100), dtype=float64,
+    ... chunksize=(20, 20, 100, 100), chunktype=cupy.ndarray>
+    >>> # Create the signal
+    >>> s = hs.signals.Signal2D(data).as_lazy()
+
+.. note::
+    See the dask blog on `Richardson Lucy (RL) deconvolution <https://blog.dask.org/2020/11/12/deconvolution>`_
+    for an example of lazy processing on GPU using dask and cupy
+
 
 .. _FitBigData-label:
 
--- a/doc/user_guide/big_data.rst
+++ b/doc/user_guide/big_data.rst
@@ -271,16 +271,27 @@ instead:
 GPU support
 -----------
 
-Lazy data processing using GPU requires setting the dask scheduler
-and explicitely transfering the data to the GPU.
+Lazy data processing on GPUs requires explicitly transferring the data to the
+GPU.
+
+On linux, it is recommended to use the
+`dask_cuda <https://docs.rapids.ai/api/dask-cuda/stable/index.html>`_ library 
+(not supported on windows) to manage the dask scheduler. As for CPU lazy
+processing, if the dask scheduler is not specified, the default scheduler
+will be used.
 
 .. code-block:: python
 
     >>> from dask_cuda import LocalCUDACluster
     >>> from dask.distributed import Client
-    >>> import cupy as cp
     >>> cluster = LocalCUDACluster()
     >>> client = Client(cluster)
+
+.. code-block:: python
+
+    >>> import hyperspy.api as hs
+    >>> import cupy as cp
+    >>> import dask.array as da
     >>> # Create a dask array
     >>> data = da.random.random(size=(20, 20, 100, 100))
     >>> print(data)
@@ -296,7 +307,7 @@ and explicitely transfering the data to the GPU.
 
 .. note::
     See the dask blog on `Richardson Lucy (RL) deconvolution <https://blog.dask.org/2020/11/12/deconvolution>`_
-    for an example of lazy processing on GPU using dask and cupy
+    for an example of lazy processing on GPUs using dask and cupy
 
 
 .. _FitBigData-label:
--- a/doc/user_guide/signal.rst
+++ b/doc/user_guide/signal.rst
@@ -1875,8 +1875,8 @@ GPU support
 GPU processing is supported thanks to the numpy dispatch mechanism of array functions
 - read `NEP-18 <https://numpy.org/neps/nep-0018-array-function-protocol.html>`_
 and `NEP-35 <https://numpy.org/neps/nep-0035-array-creation-dispatch-with-array-function.html>`_
-for more information. It means that most HyperSpy functions will work on GPU if the data is a
-:py:class:`cupy.ndarray` and that the required functions are
+for more information. It means that most HyperSpy functions will work on a GPU
+if the data is a :py:class:`cupy.ndarray` and that the required functions are
 implemented in ``cupy``.
 
 .. note::
@@ -1892,7 +1892,7 @@ implemented in ``cupy``.
     >>> type(s.data)
     ... cupy._core.core.ndarray
 
-Two convenience methods are avaiable to transfer data to or from the GPU:
+Two convenience methods are available to transfer data to or from the GPU:
 
 - :py:meth:`~.signal.BaseSignal.to_cpu`
 - :py:meth:`~.signal.BaseSignal.to_gpu`
--- a/doc/user_guide/signal.rst
+++ b/doc/user_guide/signal.rst
@@ -1886,15 +1886,16 @@ implemented in ``cupy``.
 .. code-block:: python
 
     >>> import cupy as cp
-    >>> # Create a cupy array (on GPU)
+    >>> # Create a cupy array (on GPU device)
     >>> data = cp.random.random(size=(20, 20, 100, 100))
     >>> s = hs.signals.Signal2D(data)
     >>> type(s.data)
     ... cupy._core.core.ndarray
 
-Two convenience methods are available to transfer data to or from the GPU:
+Two convenience methods are available to transfer data between the host and
+the (GPU) device memory:
 
-- :py:meth:`~.signal.BaseSignal.to_cpu`
-- :py:meth:`~.signal.BaseSignal.to_gpu`
+- :py:meth:`~.signal.BaseSignal.to_host`
+- :py:meth:`~.signal.BaseSignal.to_device`
 
 For lazy processing, see the :ref:`corresponding section<big_data.gpu>`.
--- a/doc/user_guide/signal.rst
+++ b/doc/user_guide/signal.rst
@@ -84,7 +84,7 @@ This also applies to the :py:attr:`~.signal.BaseSignal.metadata`.
     └── Signal
 	└── signal_type =
 
-Instead of using a list of *axes dictionaries* ``[dict0, dict1]`` during signal 
+Instead of using a list of *axes dictionaries* ``[dict0, dict1]`` during signal
 initialization, you can also pass a list of *axes objects*: ``[axis0, axis1]``.
 
 
@@ -292,7 +292,7 @@ For example, a numpy ragged array can be created as follow:
        >>> arr = np.array([[1, 2, 3], [1]], dtype=object)
        >>> arr
        array([list([1, 2, 3]), list([1])], dtype=object)
-       
+
 Note that the array shape is (2, ):
 
    .. code-block:: python
@@ -329,10 +329,10 @@ To create a hyperspy signal of a numpy ragged array:
 
        >>> s.axes_manager
        <Axes manager, axes: (2|ragged)>
-                   Name |   size |  index |  offset |   scale |  units 
-       ================ | ====== | ====== | ======= | ======= | ====== 
-            <undefined> |      2 |      0 |       0 |       1 | <undefined> 
-       ---------------- | ------ | ------ | ------- | ------- | ------ 
+                   Name |   size |  index |  offset |   scale |  units
+       ================ | ====== | ====== | ======= | ======= | ======
+            <undefined> |      2 |      0 |       0 |       1 | <undefined>
+       ---------------- | ------ | ------ | ------- | ------- | ------
             Ragged axis |               Variable length
 
 .. note::
@@ -345,7 +345,7 @@ To create a hyperspy signal of a numpy ragged array:
            array([[1, 2],
                   [1, 2]], dtype=object)
 
-       
+
     Unlike in the previous example, here the array is not ragged, because
     the length of the nested sequences are equal (2) and numpy will create
     an array of shape (2, 2) instead of (2, ) as in the previous example of
@@ -374,11 +374,11 @@ shape of the ragged array:
        >>> s = hs.signals.BaseSignal(arr)
        >>> s
        <BaseSignal, title: , dimensions: (|2)>
-       
+
        >>> s.ragged = True
        >>> s
-       <BaseSignal, title: , dimensions: (2|ragged)>       
-       
+       <BaseSignal, title: , dimensions: (2|ragged)>
+
 
 .. _signal.binned:
 
@@ -1211,7 +1211,7 @@ signal:
     uint8
 
     # By default `dtype=None`, the dtype is determined by the behaviour of
-    # numpy.sum, in this case, unsigned integer of the same precision as the 
+    # numpy.sum, in this case, unsigned integer of the same precision as the
     # platform interger
     >>> s4 = s.rebin(scale=(5, 2, 1))
     >>> print(s4.data.dtype)
@@ -1864,3 +1864,36 @@ and `y` direction, while the offset is determined by the ``offset`` parameter.
 The fulcrum of the linear ramp is at the origin and the slopes are given in
 units of the axis with the according scale taken into account. Both are
 available via the :py:class:`~.axes.AxesManager` of the signal.
+
+.. versionadded:: 1.7
+
+.. _gpu_processing:
+
+GPU support
+-----------
+
+GPU processing is supported thanks to the numpy dispatch mechanism of array functions
+- read `NEP-18 <https://numpy.org/neps/nep-0018-array-function-protocol.html>`_
+and `NEP-35 <https://numpy.org/neps/nep-0035-array-creation-dispatch-with-array-function.html>`_
+for more information. It means that most HyperSpy functions will work on GPU if the data is a
+:py:class:`cupy.ndarray` and that the required functions are
+implemented in ``cupy``.
+
+.. note::
+    GPU processing with hyperspy requires numpy 1.20 and dask 2021.3.0, to be
+    able to use NEP-18 and NEP-35.
+
+.. code-block:: python
+
+    >>> import cupy as cp
+    >>> # Create a cupy array (on GPU)
+    >>> data = cp.random.random(size=(20, 20, 100, 100))
+    >>> s = hs.signals.Signal2D(data)
+    >>> type(s.data)
+    ... cupy._core.core.ndarray
+
+Two convenience methods are avaiable to transfer data to or from the GPU:
+
+- :py:meth:`~.signal.BaseSignal.to_cpu`
+- :py:meth:`~.signal.BaseSignal.to_gpu`
+
--- a/doc/user_guide/signal.rst
+++ b/doc/user_guide/signal.rst
@@ -1876,11 +1876,11 @@ GPU processing is supported thanks to the numpy dispatch mechanism of array func
 - read `NEP-18 <https://numpy.org/neps/nep-0018-array-function-protocol.html>`_
 and `NEP-35 <https://numpy.org/neps/nep-0035-array-creation-dispatch-with-array-function.html>`_
 for more information. It means that most HyperSpy functions will work on a GPU
-if the data is a :py:class:`cupy.ndarray` and that the required functions are
+if the data is a :py:class:`cupy.ndarray` and the required functions are
 implemented in ``cupy``.
 
 .. note::
-    GPU processing with hyperspy requires numpy 1.20 and dask 2021.3.0, to be
+    GPU processing with hyperspy requires numpy>=1.20 and dask>=2021.3.0, to be
     able to use NEP-18 and NEP-35.
 
 .. code-block:: python
@@ -1897,3 +1897,4 @@ Two convenience methods are available to transfer data to or from the GPU:
 - :py:meth:`~.signal.BaseSignal.to_cpu`
 - :py:meth:`~.signal.BaseSignal.to_gpu`
 
+For lazy processing, see the :ref:`corresponding section<big_data.gpu>`.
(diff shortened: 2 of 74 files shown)


thomasaarholt:
Well, there's the motivation I need to push sympy/sympy#20622 across the finish line :)

din14970:
As mentioned in https://github.com/pyxem/pyxem/pull/743 and for future reference:

The `to_gpu` function copies the entire dataset to the GPU, even if you pass in a dask array. 

```
x = da.array([1, 2, 3])
g = cp.asarray(x)  # [1, 2, 3] now lives on GPU memory
```

From my understanding, `cp.as_array` will compute the result of the dask array and attempt to store it into GPU memory. For most datasets and GPUs I suppose that won't be a problem, but one should consider the option for lazy data that can not be loaded entirely into memory. A safer approach might be to let dask handle the copying, like: 

```
lazy_data_gpu = lazy_data.map_blocks(cp.asarray)
```

Then once the data can come back:

```
lazy_data = lazy_data_gpu.map_blocks(cp.asnumpy)
```

This they do indeed do in the blogpost you linked to: https://blog.dask.org/2020/11/12/deconvolution.

However, I'm not sure whether dask knows how to properly manage the GPU memory during `compute`. One might require setting up a client using e.g. `dask_cuda` to specify hardware details to avoid those issues.

hyperspy/signal.py
```diff
@@ -6096,6 +6132,15 @@ def _check_signal_mask(self, mask):
             raise ValueError("The shape of signal mask array must match "
                              "`signal_shape`.")
 
+    def to_gpu(self):
+        if not CUPY_INSTALLED:
+            raise BaseException('cupy is required.')
+        else:
+            self.data = cp.asarray(self.data)
```
ericpre, line 6492: As mentioned by @din14970 [below](https://github.com/hyperspy/hyperspy/pull/2670#issuecomment-803280165), we should define what to do with lazy signal.
I think that this convenient `to_gpu` method shouldn't be used for lazy signal and it should be changed to raise an error and suggest alternative.

ericpre:
I will rebase this branch regularly to avoid messing up with the history with merge commit.

ericpre:
> However, I'm not sure whether dask knows how to properly manage the GPU memory during `compute`. One might require setting up a client using e.g. `dask_cuda` to specify hardware details to avoid those issues.

If there are issues with dask not managing the GPU memory, this should be reported to dask. From the hyperspy perspective, it shouldn't matter if the operation are performed on numpy or cupy chunks as long as they are compatible with these arrays!

din14970:
> If there are issues with dask not managing the GPU memory, this should be reported to dask. 

As I said, I'm not sure. Experiments might include:

```python
huge_array = da.ones((5000, 5000, 5000))  #about 1 TB
cp.asarray(huge_array)  #should throw an out of memory error?

huge_array.map_blocks(cp.asarray)
huge_array.compute()  # should throw an out of memory error?

array_sum = da.sum(huge_array)  
array_sum.compute()  # what will happen here?
```

> From the hyperspy perspective, it shouldn't matter if the operation are performed on numpy or cupy chunks as long as they are compatible with these arrays!

yes, but lazy signals will happily accept most operations until you actually call `compute`. And I'm not sure the `compute` method of the hyperspy lazy signals has the same flexiblity as `dask.distributed`, which *might* be needed when dealing with cupy chunks. Maybe it's a not an issue at all, just something to look at.

thomasaarholt:
@ericpre I've resubmitted sympy/sympy#21155 now, but it isn't strictly necessary for CuPy support. `sp.lambdify` works with CuPy arrays as long as one has a recent numpy.
```python
import cupy as cp, sympy as sp

x = sp.symbols('x')
expr = 2*x**2 + 2*x + 3
f = sp.lambdify([x], expr, modules='numpy')

arr = cp.arange(3)
f(arr)
# array([ 3,  7, 15], dtype=int32)
```

Do you have any idea of curve-fitting optimizers that we can use on the GPU?

din14970:
@thomasaarholt 

> Do you have any idea of curve-fitting optimizers that we can use on the GPU?

I've also looked for this, but I have not found anything exactly like `scipy.optimize` for the GPU; would be interested to hear if there is something.
Maybe a very useless suggestion, but if the problem could be expressed in a "machine-learny" kind of way one could try [cuML](https://docs.rapids.ai/api/cuml/stable/api.html#) or PyTorch/TensorFlow. In the end it should all be solved by gradient descent right?

thomasaarholt:
@din14970 you're absolutely right about it being a good idea to look to the ml libraries. I wonder how well-suited GPUs are for iterative least squares problems though, since we often rely on the fit in one pixel as a starting point for the next. The biggest gains may be had by combining it with [SAMFIRE-like](http://hyperspy.org/hyperspy-doc/current/user_guide/model.html#smart-adaptive-multi-dimensional-fitting-samfire) approaches.

ericpre:
Without doing the iterative loop running on the gpu, there should be speed improvement (not huge) for large signal dimension or through using `function_nd` of the component. This is the same idea as the numexpr backend speed up fitting, the evaluation of the model is faster. However, there will be an overhead of transferring parameter values from/to the gpu during iteration and for large signal dimension, the time to evaluate the model should be larger than the overhead of transferring parameters values. The way the parameters are currently stored may need to change.

I came across https://github.com/gpufit/Gpufit, https://github.com/GooFit/GooFit previously (and https://github.com/mscipio/gpuKMfit just now) but I have no idea how easy it would be to use these in the model fitting and if they will work well.

ericpre:
https://github.com/tBuLi/symfit/pull/303

thomasaarholt:
I thought a bit more about how ML solutions fit/don't fit, writing it just to help my understanding/provide discussion.

- In ML, one typically trains a model with _loads_ of parameters to "fit" one big dataset. 
- In Hyperspy, we train a  model with few parameters on a much smaller dataset, but (in ML-terms) create many models with identical layers (components) but with different weights - one for each position in the navigator.

A lot of ML solutions include support for so-called "mini batches", which is a bit like having a very large signal dimension and not being able to fit it all into ram during fitting. Those won't be suitable for us (until we have some pretty crazy detectors or other type of data), since we can fit the signal dimension into ram. Instead, we probably want run several fitters in parallel, like how SAMFIRE does it.

Hopefully we can use the underlying GPU codes people develop for ML, and modify them appropriately for our problems.

@ericpre's point is good - for large Signals (especially Signal2Ds), we'll probably see a benefit. I think especially looking at `function_nd` and` component.map['values']` instead of looping over `parameter.value` for each pixel, and trying to find optimizers that can work in parallel will be great where the fit of the previous pixel isn't that useful for the next one.

din14970:
Just to report on a bit of playing I've done that partially answer my own questions

```
huge_array = da.ones((5000, 5000, 5000))  #about 1 TB
cp.asarray(huge_array)  #should throw an out of memory error?
```
Yes it does. `cp.asarray` will call `compute` under the hood and try to store the entire thing in the GPU memory. During compute Dask can discover there is no more GPU memory available. It does not know it a priori.

```
huge_array.map_blocks(cp.asarray)
huge_array.compute()  # should throw an out of memory error?
```
This is identical as the previous since `cp.asarray` calls compute under the hood. Again throws an error.

```
array_sum = da.sum(huge_array)  
array_sum.compute()  # what will happen here?
```
Whether this works or not depends on the way the compute graph is structured and the chunk size. Here is where things get really tricky. If at some point during compute dask tries to allocate more memory than is free on the GPU through cupy it can error out with `cupy.cuda.memory.OutOfMemoryError: Out of memory allocating`. This means that one should beware of GPU memory useage spikes that may occur due to improper chunking or execution.

thomasaarholt:
@din14970 nice writeup. For the last example, involving Sum, did you play with specifying chunk size? Try giving dask a chunk size of about 100MB, map_blocks it to cupy, and then sum and compute it. The approx 100MB chunk size is what both dask and hyperspy recommends, and I think (hope) that it should work fine for most signal operations we use in hyperspy.

din14970:
@thomasaarholt I didn't try this specifically for this Sum example because it turns out it's quite slow to transfer 1TB of data to the GPU and I didn't bother to wait it out, but I experimented with this in my more complicated workflow to index spot patterns with template matching. This pertains to my PR here: https://github.com/pyxem/pyxem/pull/743. Basically I can make the GPU run out of memory by performing the calculation over very large chunks. Interestingly, one test where I did not run out of GPU memory I observed that very large chunks, larger than typically recommended, significantly sped up the calculation (factor of 2-3). In my approach, the workflow is basically: data to GPU -> reduce -> send back reduced data, and it seems to work fine as long as enough VRAM is continuously freed before new stuff is added. Whether this happens or not depends on the structure of the execution graph, so I guess we just have to hope that Dask is smart enough in most reasonable circumstances.

thomasaarholt:
The 100MB recommendation is being very "safe". The fastest computation will be the one with the largest (sensibly structured) chunk size that will fit into (V)RAM. I reckon the 100MB suggestion is there to give regular users a working limit that let's them look at their data without tings crashing. Perhaps I'll write a doc PR to add that to dasks documentation. 

I too think that we have to rely on dask to do what is smart. Being too creative is likely to just make it difficult for us down the road.

thomasaarholt:
Another possibility is using [Jax](https://jax.readthedocs.io/en/latest/_autosummary/jax.scipy.optimize.minimize.html?highlight=minimize)! It seems to support GPU/cupy and has the scipy syntax. But Windows support is currently limited.

@din14970 see here for some dask discussion on an issue I created using your code above: https://github.com/dask/dask/issues/7687

din14970:
@thomasaarholt cool! thanks for raising that issue, there's some interesting links in there. We will still mostly have to deal with the issue of transferring data from host to device memory since we can't "generate" the experimental data on the GPU, unless perhaps one deals with simulated data (on that note it would be interesting to see for example abTEM interoperability with hyperspy). In any case, it may be worthwhile to dig a bit deeper into GPU memory management for large datasets.

din14970:
I thought I would raise awareness here to my issue here https://github.com/pyxem/pyxem/issues/778. It seems that when one uses `array.map_blocks(cupy.asarray)`, then some GPU processing, then `array.map_blocks(cupy.asnumpy)`, the single machine scheduler can get confused and schedule all the copying to GPU operations sequentially and running out of VRAM. A partial solution is including the "to GPU" and "from GPU" steps in the processing function, but this would be a waste if multiple GPU processing steps are desired. The `LocalCluster` approach does not seem to cause VRAM issues. It seems to me that the way forward is to rework the LazySignal's  [`compute` function](https://github.com/hyperspy/hyperspy/blob/8c74de87855a7e6cf53acfa85ca6b1e77367d619/hyperspy/_signals/lazy.py#L93) such that it relies on a `LocalCluster` from `dask.distributed` instead.

ericpre:
Thanks @din14970! This sounds like useful information and confirms that for now we should most likely not anything with lazy signal and gpu. Maybe for this PR, we should simply provide links to relevant documentation, example, blog post, etc?

thomasaarholt:
Are we currently able to test cupy as part of the CI?

Happy to test this on my 1080TI. Any example codes that are worth running?

ericpre:
> Are we currently able to test cupy as part of the CI?

I haven't came across any free CI service providing GPU.

> Happy to test this on my 1080TI. Any example codes that are worth running?

Any code using `map`, plotting, ROI would do.

doc/user_guide/big_data.rst
```diff
@@ -264,6 +264,40 @@ instead:
...
 
+.. versionadded:: 1.7
+
+.. _big_data.gpu:
+
+GPU support
+-----------
+
+Lazy data processing using GPU requires setting the dask scheduler
+and explicitely transfering the data to the GPU.
```
jlaehne: explicitly
transferring

doc/user_guide/signal.rst
```diff
@@ -1863,3 +1863,36 @@ and `y` direction, while the offset is determined by the ``offset`` parameter.
...
+.. code-block:: python
+
+    >>> import cupy as cp
+    >>> # Create a cupy array (on GPU)
+    >>> data = cp.random.random(size=(20, 20, 100, 100))
+    >>> s = hs.signals.Signal2D(data)
+    >>> type(s.data)
+    ... cupy._core.core.ndarray
+
+Two convenience methods are avaiable to transfer data to or from the GPU:
```
jlaehne: available

doc/user_guide/signal.rst
```diff
@@ -1863,3 +1863,36 @@ and `y` direction, while the offset is determined by the ``offset`` parameter.
...
+
+.. _gpu_processing:
+
+GPU support
+-----------
+
+GPU processing is supported thanks to the numpy dispatch mechanism of array functions
+- read `NEP-18 <https://numpy.org/neps/nep-0018-array-function-protocol.html>`_
+and `NEP-35 <https://numpy.org/neps/nep-0035-array-creation-dispatch-with-array-function.html>`_
+for more information. It means that most HyperSpy functions will work on GPU if the data is a
```
jlaehne: 'a' GPU

doc/user_guide/big_data.rst
```diff
@@ -264,6 +264,40 @@ instead:
...
+    >>> data = data.map_blocks(cp.asarray)
+    >>> print(data)
+    ... dask.array<random_sample, shape=(20, 20, 100, 100), dtype=float64,
+    ... chunksize=(20, 20, 100, 100), chunktype=cupy.ndarray>
+    >>> # Create the signal
+    >>> s = hs.signals.Signal2D(data).as_lazy()
+
+.. note::
+    See the dask blog on `Richardson Lucy (RL) deconvolution <https://blog.dask.org/2020/11/12/deconvolution>`_
+    for an example of lazy processing on GPU using dask and cupy
```
jlaehne: 'a' GPU

doc/user_guide/big_data.rst
```diff
@@ -264,6 +264,40 @@ instead:
...
     >>> s.plot(navigator='slider')
 
+.. versionadded:: 1.7
+
+.. _big_data.gpu:
+
+GPU support
+-----------
+
+Lazy data processing using GPU requires setting the dask scheduler
```
jlaehne: 'a' GPU

hyperspy/learn/mva.py
```diff
@@ -188,7 +194,7 @@ def decomposition(
                 number of components to extract is lower than 80% of the smallest
                 dimension of the data, then the more efficient "randomized"
                 method is enabled. Otherwise the exact full SVD is computed and
-                optionally truncated afterwards.
+                optionally truncated afterwards. For cupy array, only
```
jlaehne: sentence mot finished

hyperspy/learn/mva.py
```diff
@@ -232,6 +238,18 @@ def decomposition(
...
+            if algorithm == 'SVD':
+                if svd_solver == 'randomized':
+                    raise ValueError(
+                        "`svd_solver='randomized'` is not supported with "
+                        "cupy array."
+                        )
+                elif svd_solver == 'auto':
+                    svd_solver = 'full'
+            else:
+                raise TypeError("cupy array are not supported.")
```
jlaehne: arrays

hyperspy/learn/mva.py
```diff
@@ -2054,13 +2083,15 @@ def cluster_analysis(self,
             raise ImportError(
                 'sklearn is not installed. Nothing done')
 
+        if is_cupy_array(self.data):  # pragma: no cover
+            raise TypeError("cupy array are not supported.")
```
jlaehne: arrays

hyperspy/misc/utils.py
```diff
@@ -1509,3 +1512,89 @@ def is_binned(signal, axis=-1):
...
+
+    Returns
+    -------
+    bool
+        True if it is cupy array, False otherwise.
+
+    """
+    try:
+        import cupy as cp
+        return isinstance(array, cp.ndarray)
```
jlaehne, line 1569: coverage?

hyperspy/misc/utils.py
```diff
@@ -1509,3 +1512,89 @@ def is_binned(signal, axis=-1):
...
+        Array to determine whether numpy or cupy should be used
+
+    Returns
+    -------
+    array : numpy.ndarray
+
+    """
+    if is_cupy_array(array):
+        import cupy as cp
+        array = cp.asnumpy(array)
```
jlaehne, line 1598: coverage?

hyperspy/misc/utils.py
```diff
@@ -1509,3 +1512,89 @@ def is_binned(signal, axis=-1):
...
+    try:
+        import cupy as cp
+        return isinstance(array, cp.ndarray)
+    except ImportError:
+        return False
+
+
+def to_numpy(array):
+    """
+    Returns the array as an numpy array
```
jlaehne: a numpy array

hyperspy/roi.py
```diff
@@ -1073,8 +1074,15 @@ def __call__(self, signal, out=None, axes=None):
...
         ir = [slices[natax.index(axes[0])],
               slices[natax.index(axes[1])]]
+
         vx = axes[0].axis[ir[0]] - cx
         vy = axes[1].axis[ir[1]] - cy
+
+        # convert to cupy array when necessary
+        if is_cupy_array(signal.data):
+            import cupy as cp
+            vx, vy = cp.array(vx), cp.array(vy)
```
jlaehne, line 1085: coverage?

ericpre:
We can't cover the GPU code on CI because there are no GPUs. We could add `# prama: no cover` to please the coverage on CI but I think that it is more useful not to do this because by running the coverage locally (with GPU), we can assess the code coverage.

jlaehne:
Can you fix the conflict? I would be fine with merging.

jlaehne:
I think the only thing missing is a changelog entry (I added a progress section to the PR description)!

hyperspy/drawing/signal.py
```diff
@@ -92,7 +94,7 @@ def _plot_loading(loadings, idx, axes_manager, ax=None,
                       axes[0].high_value,
                       axes[1].high_value,
                       axes[1].low_value)
-        im = ax.imshow(loadings[idx].reshape(shape),
+        im = ax.imshow(to_numpy(loadings[idx].reshape(shape)),
```
din14970: don't you already call `to_numpy` on `loadings` above? Personally I would call `to_numpy` on all the relevant arguments at the beginning of these functions instead of inline so if something goes wrong there you error out early.

hyperspy/learn/mva.py
```diff
@@ -232,6 +240,18 @@ def decomposition(
...
+            if algorithm == 'SVD':
+                if svd_solver == 'randomized':
+                    raise ValueError(
+                        "`svd_solver='randomized'` is not supported with "
+                        "cupy array."
+                        )
+                elif svd_solver == 'auto':
+                    svd_solver = 'full'
+            else:
+                raise TypeError("cupy arrays are not supported.")
```
din14970: might repeat here `... with algorithm={algorithm}`, as they are supported for `SVD`

hyperspy/misc/utils.py
```diff
@@ -1543,3 +1546,89 @@ def is_binned(signal, axis=-1):
...
+
+    """
+    try:
+        import cupy as cp
+        return isinstance(array, cp.ndarray)
+    except ImportError:
+        return False
+
+
+def to_numpy(array):
```
din14970, line 1574: Just from the name and first line of the docstring one might expect this to work on any kind of array, also to compute dask arrays. I'm not sure whether there is a better alternative. As a general comment to these functions, I'm wondering whether the dask array case should be considered here or not.

hyperspy/signal.py
```diff
@@ -6414,6 +6458,58 @@ def _check_signal_mask(self, mask):
             raise ValueError("The shape of signal mask array must match "
                              "`signal_shape`.")
 
+    def to_gpu(self):
```
din14970: I'm wondering whether it might not be better to use CUDA terminology in these functions, e.g. `to_device` and `to_host`. Reasons being only CUDA enabled GPUs will work, and `to_cpu` is not entirely correct as a name with regards to what is happening.

din14970 (commented):
Sorry for very late reply, some minor considerations, more food for thought than anything.

jlaehne (approved):

hyperspy/signal.py
ericpre: > I'm wondering whether it might not be better to use CUDA terminology in these functions, e.g. `to_device` and `to_host`.

Yes, I agree, I went for `to_cpu` and `to_gpu`, because the meaning is straightforward for anyone - although this assume that the user understand that the data is in memory somewhere - but this is better to use the device/host terminology: it is simple to explain in the user guide and it will be consistent with other documentation.

> `to_cpu` is not entirely correct as a name with regards to what is happening.

I don't understand, why using `to_device` or `to_host` would be better?

hyperspy/misc/utils.py
ericpre: Good point, shall we raise an error in case of dask array?

din14970: Yes I think an error should be fine, might catch accidental incorrect use of to_numpy elsewhere.

hyperspy/signal.py
din14970: > I don't understand, why using to_device or to_host would be better?

The way I see it, `to_gpu` is "more correct" in the sense that a GPU is a device with both compute and memory, it is equivalent to the device. But under "host" I understand both RAM and CPU. And you really send the data from VRAM to RAM as final destination. Technically both `to_gpu` and `to_cpu` are sending data though the CPU though :smile:.

ericpre:
Thanks @jlaehne and @din14970 for the review. I rebased and the latest changes for @din14970's review comments are in [0030c52](https://github.com/hyperspy/hyperspy/pull/2670/commits/0030c5222b08b60fd9dfbfd1f19ff67bb3631bd5) and [8e0f38f](https://github.com/hyperspy/hyperspy/pull/2670/commits/8e0f38faac4260a47b4bd38e78b72229d7fcb2db).

hyperspy/drawing/utils.py
```diff
@@ -1003,8 +1013,8 @@ def transparent_single_color_cmap(color):
             for j, im in enumerate(ims):
                 ax = f.add_subplot(rows, per_row, idx + 1)
                 axes_list.append(ax)
-                data = to_numpy(im.data)
-                centre = next(centre_colormaps)   # get next value for centreing
+                centre = next(centre_colormaps) # get next value for centreing
```
jlaehne:
```suggestion
                centre = next(centre_colormaps) # get next value for centring
```

hyperspy/exceptions.py
```diff
@@ -205,3 +205,16 @@ class VisibleDeprecationWarning(UserWarning):
...
     """
     pass
+
+
+class LazyCupyConversion(Exception):
+
+    def __init__(self):
+        self.error = (
+            "Automatically converting data to cupy array is not supported "
+            "for lazy signal. Read the corresponding section in the user "
```
jlaehne:
```suggestion
            "for lazy signals. Read the corresponding section in the user "
```

hyperspy/signal.py
```diff
@@ -6458,11 +6458,11 @@ def _check_signal_mask(self, mask):
...
 
-    def to_gpu(self):
+    def to_device(self):
         """
-        Transfer data array to GPU device memory using cupy.asarray. Lazy
-        signal are not supported by this method, see user guide for information
-        on how to process data lazily using GPU.
+        Transfer data array from host to GPU device memory using cupy.asarray.
+        Lazy signal are not supported by this method, see user guide for
+        information on how to process data lazily using GPU.
```
jlaehne:
```suggestion
        Lazy signals are not supported by this method, see user guide for
        information on how to process data lazily using the GPU.
```

jlaehne (approved):

jlaehne:
After solving a merge conflict, two of the tests are now failing.

jlaehne:
Thanks for fixing.

PR #2665, opened 2021-03-10
Reduce size of hdf5 test file

jlaehne:
### Description of the change
Reduce original_metadata contained in `hdf5_files/example2_v2.2.hspy` to reduce filesize from > 800 kB to 184 kB.

### Progress of the PR
- [X] Change implemented (can be split into several points),
- [X] ready for review.

PR #2671, opened 2021-03-12
Reduce the size of hspy v2.2 test file

ericpre:
Alternative to #2665.

The `hspy` reference file was created using hyperspy 1.0.1 to make sure the metadata, etc. are correct. The file is now 26kB instead of 829kB.


thomasaarholt:
I truly hope we never had a 829 MB test file with hyperspy :laughing:

ericpre:
> I truly hope we never had a 829 MB test file with hyperspy

ups, sorry for the typo! :)

PR #2673, opened 2021-03-12
Fix finding dataset in group in EMD NCEM file

### LINKED ISSUE #2614: EMD (NCEM) io_plugin no longer imports mettadata
In previous versions loading a EMD (NCEM) file would load the images and metadata associated with the hdf5 file.  Now, either a dataset_path or stack_group needs to be specified for the data to load the signal and will not load the metadata regardless.
For example:
In the below image I am showing the emd file inspected in EMD viewer.  It is a single (4|4096,4096) image stack with plenty of metadata describing the scope, navigation axis, and signal axis.  Loading the signal without kwargs stacks two (|4) signals. Loading with `stack_group=False` loads two signals, one with he correct (|4,4096,4096) and some unkown (|4) with data that does not make and sense.  Loading with `dataset_path='/data/drift/data'` points to the correct data path.  None of the above load the metadata.
image
image

ericpre:
Fixes #2614.

### Progress of the PR
- [x] Fix bug with finding dataset path for (non-complicant?) file containing several dataset in a group,
- [x] ready for review.


jlaehne:
LGTM

jlaehne (approved):

PR #2676, opened 2021-03-16
fix squeeze, add tests and improve docs

### LINKED ISSUE #2121: Removing axes and data by index
I'm simulating some data, and the simulation software always returns a fixed number of axes - in this case five navigation axes and two signal axes.

Sometimes I don't iterate over one of the axes, resulting in a signal shape of (e.g) `(200, 5, 1, 1, 20|256, 256)`. In these cases I wish to remove the axes with size 1, as it makes it easier to handle later on. 

I don't want to use s.inav[], because I don't necessarily know the correct indices of the axes to be removed in the signal. Is there a way to get the index?

I've tried using s.axes_manager.remove(), but it only removes the axis, not also the underlying data, i.e.:

```python
>>> s = hs.signals.Signal2D(np.random.random((2,3,1,5,6,8,8)))
>>> s.axes_manager.remove(s.axes_manager[2])
>>> s.plot()
ValueError: axes don't match array
```

jlaehne:
### Description of the change
Closes #2121

The `BaseSignal.squeeze()` function missed the case where two axes with zero-dimension directly follow one on the other. Fixed by `reversed` iteration.
Improved Docstring, added to User Guide, added test.

### Progress of the PR
- [X] add entry to `CHANGES.rst` (if appropriate),
- [X] add tests,
- [X] ready for review.

```
>>> s = hs.signals.Signal2D(np.random.random((2,1,1,6,8,8)))
>>> s = s.squeeze()
>>> s
```

### DIFF
--- a/hyperspy/signal.py
+++ b/hyperspy/signal.py
@@ -2482,10 +2482,23 @@ class BaseSignal(FancySlicing,
     def squeeze(self):
         """Remove single-dimensional entries from the shape of an array
         and the axes. See :py:func:`numpy.squeeze` for more details.
+        
+        Returns
+        -------
+        s : signal
+            A new signal object with single-entry dimensions removed
+        
+        Examples
+        --------
+        >>> s = hs.signals.Signal2D(np.random.random((2,1,1,6,8,8)))
+        <Signal2D, title: , dimensions: (6, 1, 1, 2|8, 8)>
+        >>> s = s.squeeze()
+        >>> s
+        <Signal2D, title: , dimensions: (6, 2|8, 8)>
         """
         # We deepcopy everything but data
         self = self._deepcopy_with_new_data(self.data)
-        for axis in self.axes_manager._axes:
+        for axis in reversed(self.axes_manager._axes):
             if axis.size == 1:
                 self._remove_axis(axis.index_in_axes_manager)
         self.data = self.data.squeeze()
--- a/hyperspy/signal.py
+++ b/hyperspy/signal.py
@@ -2502,7 +2502,7 @@ class BaseSignal(FancySlicing,
             for axis in reversed(ax):
                 if axis.size == 1:
                     self._remove_axis(axis.index_in_axes_manager)
-            self.data = self.data.squeeze()
+        self.data = self.data.squeeze()
         return self
 
     def _to_dictionary(self, add_learning_results=True, add_models=False):
--- a/hyperspy/signal.py
+++ b/hyperspy/signal.py
@@ -2498,10 +2498,11 @@ class BaseSignal(FancySlicing,
         """
         # We deepcopy everything but data
         self = self._deepcopy_with_new_data(self.data)
-        for axis in reversed(self.axes_manager._axes):
-            if axis.size == 1:
-                self._remove_axis(axis.index_in_axes_manager)
-        self.data = self.data.squeeze()
+        for ax in (self.axes_manager.signal_axes, self.axes_manager.navigation_axes):
+            for axis in reversed(ax):
+                if axis.size == 1:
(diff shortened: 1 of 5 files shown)


hyperspy/signal.py
```diff
@@ -2482,10 +2482,23 @@ def get_title(signal, i=0):
...
+        >>> s = hs.signals.Signal2D(np.random.random((2,1,1,6,8,8)))
+        <Signal2D, title: , dimensions: (6, 1, 1, 2|8, 8)>
+        >>> s = s.squeeze()
+        >>> s
+        <Signal2D, title: , dimensions: (6, 2|8, 8)>
         """
         # We deepcopy everything but data
         self = self._deepcopy_with_new_data(self.data)
-        for axis in self.axes_manager._axes:
+        for axis in reversed(self.axes_manager._axes):
```
ericpre: This will not work all the time, for example:
```python
s = hs.signals.BaseSignal(np.random.random((2,1,1,6,8,8,1,1)))
s2 = s.transpose(signal_axes=[0, 1, 2])
s3 = s2.squeeze()
```
I suspect the navigation and signal axes may need to be iterated separately here.

ericpre (changes requested):
See comment.

hyperspy/signal.py
jlaehne: Good catch, a separate iteration over the two seems to do the job. I'll push a commit in a moment.

hyperspy/signal.py
```diff
@@ -2482,13 +2482,27 @@ def get_title(signal, i=0):
...
         self = self._deepcopy_with_new_data(self.data)
-        for axis in self.axes_manager._axes:
-            if axis.size == 1:
-                self._remove_axis(axis.index_in_axes_manager)
-        self.data = self.data.squeeze()
+        for ax in (self.axes_manager.signal_axes, self.axes_manager.navigation_axes):
+            for axis in reversed(ax):
+                if axis.size == 1:
+                    self._remove_axis(axis.index_in_axes_manager)
+            self.data = self.data.squeeze()
```
ericpre: Do numpy squeeze needs to be call inside the for loop?
```suggestion
        self.data = self.data.squeeze()
```

hyperspy/tests/signal/test_tools.py
```diff
@@ -51,6 +51,16 @@ def test_signal_iterator():
         for i, signal in enumerate(s):
             assert i == signal.data[0]
 
+def test_squeeze():
+    s = signals.Signal2D(np.random.random((2,1,1,6,7,8)))
+    assert s.squeeze().axes_manager.shape == (6, 2, 8, 7)
+    assert s.squeeze().data.shape == (2, 6, 7, 8)
+    s = signals.BaseSignal(np.random.random((2,1,1,6,8,7,1,1)))
```
ericpre: Wondering if this is worth adding another order of axis size which should give the same output, so that we can easily use `pytest.mark.parametrize` marker. For example, the following case should give the same output:
```python
s = hs.signals.BaseSignal(np.random.random((2, 1, 1, 6, 8, 7, 1, 1)))
....
s = hs.signals.BaseSignal(np.random.random((1, 1, 2, 6, 8, 1, 1, 7)))
```

ericpre (commented):
Yes, this is working fine now! Two small suggestions for improvement.

hyperspy/tests/signal/test_tools.py
jlaehne: Why not, I added three versions each for the two tested shapes

hyperspy/signal.py
jlaehne: done

ericpre (approved):

PR #2678, opened 2021-03-16
Ensure correct endian in numba_histogram

### LINKED ISSUE #2668: Histogram error with `>u2` dtype
The histogram function when plotting does not work with certain dtypes. Example of this:

```python
import dask.array as da
import hyperspy.api as hs
data = da.arange(0, 10000, dtype='>u2').reshape((100, 100))
s.plot()
# Press H
```

Gives the error message:
```python
File "hyperspy/hyperspy/signal_tools.py", line 964, in plot_histogram
    self.hist_data = self._get_histogram(data)
File "hyperspy/hyperspy/signal_tools.py", line 922, in _get_histogram
    return numba_histogram(data, bins=self.bins,

File "hyperspy/hyperspy/misc/array_tools.py", line 396:
def numba_histogram(data, bins, ranges):
    <source elided>
    # Adapted from https://iscinumpy.gitlab.io/post/histogram-speeds-in-python/
    hist = np.zeros((bins,), dtype=np.intp)
    ^

This error may have been caused by the following argument(s):
- argument 0: Unsupported array dtype: >u2
```

The `>u2` dtype sometimes comes from reading raw binary files from memory, via `numpy.memmap`. Ergo, the dtype is set directly when loading the file. While it is possible to change the dtype manually to something else, like `np.uint16` it would be nice to avoid this if possible.

The error seems to stem from the `numba_histogram` function utilizing numba, which does not support these types of dtypes (like `>u2`).

I'm not sure how best to fix this. One way could be to check if the data is this type of dtype, and if it is, convert it to `np.uint16`.

jlaehne:
### Description of the change
Makes sure that `numba_histogram` runs for big endian datatypes like `>u2` or `>f4` by converting data to little endian for execution.

Closes #2668

### Progress of the PR
- [X] Change implemented,
- [X] extend to other numba functions,
- [x] add tests,
- [X] ready for review.

```python
import numpy as np
from hyperspy.misc.array_tools import numba_histogram

# works
arr = np.arange(100, dtype='<u2')
numba_histogram(arr, 5, (0, 100))

# works
arr = np.arange(100, dtype='u2')
numba_histogram(arr, 5, (0, 100))

# works
arr = np.arange(100, dtype='>u2')
numba_histogram(arr, 5, (0, 100))
```

### DIFF
--- a/hyperspy/misc/array_tools.py
+++ b/hyperspy/misc/array_tools.py
@@ -313,6 +313,9 @@ def _linear_bin(dat, scale, crop=True):
         # Set up the result np.array to have a new axis[0] size for after
         # cropping.
         result = np.zeros((dim,) + dat.shape[1:])
+        # Make sure that native endian is used
+        if not dat.dtype.isnative:
+            dat = dat.astype(dat.dtype.type)
         # Carry out binning over axis[0]
         _linear_bin_loop(result=result, data=dat, scale=s)
         # Swap axis[0] back to the original axis location.
@@ -391,16 +394,16 @@ def numba_histogram(data, bins, ranges):
     hist : array
         The values of the histogram.
     """
-    if data.dtype.byteorder == '>':
-        return _numba_histogram(data.byteswap().newbyteorder(), bins, ranges)
-    else:
-        return _numba_histogram(data, bins, ranges)
+    # Make sure that native endian is used
+    if not data.dtype.isnative:
+        data = data.astype(data.dtype.type)
+    return _numba_histogram(data, bins, ranges)
 
 
 @njit(cache=True)
 def _numba_histogram(data, bins, ranges):
     """
-    Numba histogram computation requiring little endian datatype.
+    Numba histogram computation requiring native endian datatype.
     """
     # Adapted from https://iscinumpy.gitlab.io/post/histogram-speeds-in-python/
     hist = np.zeros((bins,), dtype=np.intp)
--- a/hyperspy/misc/array_tools.py
+++ b/hyperspy/misc/array_tools.py
@@ -375,7 +375,6 @@ def dict2sarray(dictionary, sarray=None, dtype=None):
     return sarray
 
 
-@njit(cache=True)
 def numba_histogram(data, bins, ranges):
     """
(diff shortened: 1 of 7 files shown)


jlaehne:
Question is, whether this affects any other functions that run numba? Namely

- `_linear_bin_loop` in `array_tools.py`
- `lowess` in `lowess_smooth.py`
- `_fast_mean` and `_fast_std` in `peakfinders2D.py`

Probably, the would need a similar wrapper.

hyperspy/misc/array_tools.py
```diff
@@ -392,6 +391,17 @@ def numba_histogram(data, bins, ranges):
     hist : array
         The values of the histogram.
     """
+    if data.dtype.byteorder == '>':
```
francisco-dlp: What about the following instead?

```python
    if data.dtype.byteorder == '>':
       data = data.byteswap().newbyteorder()
    return _numba_histogram(data, bins, ranges)
```
It is the same, but easier to maintain...

jlaehne: Sure, no problem. 

Should I add it for the other 4 numba-functions? I think we cannot rule out that someone calls them with the wrong endian. For the private functions I could just add it to the calling functions, only `lowess` would need a wrapper.

ericpre: The issue is that numba only support native endianess and this approach will not work on all platform. What's about:
```python
        # Numba only supported native dtype
        # https://github.com/numba/numba/issues/2243
        if not data.dtype.isnative:
            data = data.astype(data.dtype.type
```

jlaehne: Yes, that should do the job, indeed.

ericpre (approved):
Nice to take the opportunity to increase the coverage, thanks!

jlaehne:
> Nice to take the opportunity to increase the coverage, thanks!

Not much, but at least a few obvious ones.

In the end, I refrained from covering `_fast_mean` and `_fast_std` in `peakfinders2D.py`, because I could not trigger the error using the 'stat' peakfinder routine.

PR #2679, opened 2021-03-16
Fix navigation_mask decomposition

### LINKED ISSUE #2677: Bug in signal decomposition concerning navigation_mask
Please run:

```
%matplotlib qt
import hyperspy.api as hs
import hyperspy
import numpy as np
import matplotlib
from matplotlib import pyplot as plt

# some fake data
e = np.arange(0,2048)
spectra = [np.exp(-((e-p)/100) ** 2) for p in [200,600,1400]]

data = np.zeros((192,256,2048))
comp = [(0,90,4,1,1), (90,100,1,4,0.8), (100,110,1,4,1.0),(110,120,1,4,1.2),(120,192,1,1,4)]
for b,t,c0,c1,c2 in comp:
    data[b:t] = spectra[0]*c0+spectra[1]*c1+spectra[2]*c2
data = np.random.poisson(lam=100*data)*1.0
signal = hs.signals.Signal1D(data, signal_type='EDS_TEM')
signal.axes_manager[-1].name = 'E'
signal.axes_manager['E'].units = 'eV'
signal.axes_manager['E'].scale = 1.0e1
signal.axes_manager['E'].offset = 0

signal.plot()

#--------------------------------------------------
# do NMF decomposition
# define a mask
mask = np.ones((192,256)).astype(bool)
mask[90:120] = False

# one would expect to use navigation_mask=mask
# but this leads to "The shape of `mask` must match the shape of the `navigation_shape`."
# using navigation_mask=mask.T does not give that error, but the result makes no sense
# see the resulting loadings plot

signal.decomposition(normalize_poissonian_noise=True, 
                     navigation_mask=mask.T, 
                     algorithm='NMF', 
                     output_dimension=3, 
                     centre=None, 
                     max_iter=20,
                     return_info=False)    

loadings = signal.get_decomposition_loadings()
loadings.plot()
```

ericpre:
Fixes #2677. When passing a numpy array for the `navigation_mask` argument of `decomposition`, the array was missing a transpose operation. Passing hyperspy signal is working fine.

### Progress of the PR
- [x] Fix plotting boolean array (didn't play well with auto contrast).
- [x] add tests,
- [x] ready for review.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -37,6 +37,9 @@ RELEASE_next_patch (Unreleased)
 * Add support for python 3.9, fix deprecation warning with matplotlib 3.4.0 and bump minimum requirement to numpy 1.17.1 and dask 2.1.0. (`#2663 <https://github.com/hyperspy/hyperspy/pull/2663>`_)
 * Add integration test suite documentation in the developer guide. (`#2663 <https://github.com/hyperspy/hyperspy/pull/2663>`_)
 * Fix squeeze function for multiple zero-dimensional entries, improved docstring, added to user guide. (`#2676 <https://github.com/hyperspy/hyperspy/pull/2676>`_)
+* Use native endianess in numba jitted functions. (`#2678 <https://github.com/hyperspy/hyperspy/pull/2678>`_)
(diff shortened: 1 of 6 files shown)


jlaehne:
LGTM

jlaehne (approved):

PR #2681, opened 2021-03-17
Fix CL Quant Error

k8macarthur:
Previously the following code would produce different composition values in the EDX quantification.
```python
import hyperspy.api as hs
import numpy as np
def get_composition(intensities, xray_lines, kfactors, thickness):
    spc = hs.signals.EDSTEMSpectrum(np.array([0]))
    signals = []
    for index, intensity in enumerate(intensities):
        signal = hs.signals.BaseSignal(np.array([intensity]))
        signal.metadata.add_node('Sample')
        xray_line = xray_lines[index]
        element = xray_line[:-3]
        signal.metadata.Sample.elements = [element]
        signal.metadata.Sample.xray_lines = [xray_line]
        signals.append(signal)
        
    results = spc.quantification(intensities=signals, method='CL', factors=kfactors, absorption_correction=True, thickness=thickness)
    atomic_percents = [result.data[0] for result in results[0]]
    return atomic_percents

get_composition([1000, 180], ['Al_Ka', 'N_Ka'], [1.040, 2.547], 100)

get_composition([180, 1000], ['N_Ka', 'Al_Ka'], [2.547, 1.040], 100)
```

This pull request aims to fix that and add a unit test to check for this error in the future.

### DIFF
--- a/hyperspy/misc/eds/utils.py
+++ b/hyperspy/misc/eds/utils.py
@@ -440,7 +440,7 @@ def quantification_cliff_lorimer(intensities,
     else:
         index = np.where(intensities > min_intensity)[0]
         if absorption_correction is not None:
-            absorption_correction = absorption_correction.reshape(dim[0], dim2)
+            absorption_correction = absorption_correction
         else:
             # default to ones
             absorption_correction = np.ones_like(intens, dtype='float')
--- a/hyperspy/misc/eds/utils.py
+++ b/hyperspy/misc/eds/utils.py
@@ -440,7 +440,7 @@ def quantification_cliff_lorimer(intensities,
     else:
         index = np.where(intensities > min_intensity)[0]
         if absorption_correction is not None:
-            absorption_correction = absorption_correction
+            absorption_correction = absorption_correction.reshape(dim[0], dim2)
         else:
             # default to ones
             absorption_correction = np.ones_like(intens, dtype='float')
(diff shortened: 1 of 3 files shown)


hyperspy/misc/eds/utils.py
```diff
@@ -440,7 +440,7 @@ def quantification_cliff_lorimer(intensities,
     else:
         index = np.where(intensities > min_intensity)[0]
         if absorption_correction is not None:
-            absorption_correction = absorption_correction
+            absorption_correction = absorption_correction.reshape(dim[0], dim2)
```
ericpre: Why is this necessary? `dim2` is the other branch of the `if ... else` block but not in this one, so I guess this may not work.

This function should be rewritten to remove `if ... else` block and vectorise over the navigation dimension instead of doing a for loop. It should improve performance significantly but maybe it is better to leave it to another PR?

ericpre (changes requested):
Thanks for the quick fix! This looks good, except one comment. It would be good to add a test covering the line highlighted by the codecov annotation.

hyperspy/misc/eds/utils.py
k8macarthur: Sorry yeah, this doesn't need to be changed. I agree there's probably more vectorisation of this function that could be done but I'm not sure I'm good enough at that to do it.

ericpre (approved):

ericpre:
Corresponding discussion on https://gitter.im/hyperspy/hyperspy?at=60509e66e8267a46f21b103b

PR #2684, opened 2021-03-17
Fix image contrast tool

ericpre:
### Progress of the PR
- [x] Fix closing image contrast tool,
- [x] Fix setting vmin, vmax values when closing (previously, it will reset the value when changing navigation axis),
- [x] Fix blitting issue with matplotlib 3.4 (infinite recursive loop with the `draw_event`), related to https://github.com/matplotlib/matplotlib/pull/17480.
- [x] add tests,
- [x] ready for review.

```python
import numpy as np
import hyperspy.api as hs

N_nav, N_sig = 100, 128
nnav, nsig = 2, 2
data = np.random.random(N_nav**nnav*N_sig**nsig).reshape([N_nav]*nnav + [N_sig]*nsig)

s = hs.signals.Signal2D(data)
s.plot()
# press `h` key on a figure
```


jlaehne:
Can you fix the two codecov warnings that came up in the annotations?

Does this close any reported issues?

Otherwise, LGTM.

ericpre:
I don't think there is an issue open on this. I have increased the coverage of the range widget, the widgets will need some more tidying up and going further on these consideration will be a different story!!

jlaehne (approved):

PR #2686, opened 2021-03-20
Set snap option in roi.interactive

ericpre:
### Progress of the PR
- [x] Allow setting whether the ROI should snap or not in `roi.interactive`,
- [x] fix a bug when using `hs.interactive` with function returning `None`, 
- [x] improve user guide,
- [x] add tests,
- [x] ready for review.

```python
import hyperspy.api as hs
import numpy as np

im = hs.signals.Signal2D(np.arange(100).reshape(10, 10))
im.plot()

roi = hs.roi.CircleROI(cx=1.5, cy=3.2, r=1.2)
im_roi = roi.interactive(im, color="red", snap=False)

def print_centre(roi):
    print(roi.cx, roi.cy)

hs.interactive(print_centre, roi=roi, event=roi.events.changed,
               recompute_out_event=None)
```

### DIFF
--- a/hyperspy/tests/utils/test_roi.py
+++ b/hyperspy/tests/utils/test_roi.py
@@ -549,48 +549,70 @@ class TestInteractive:
         sr2 = r(s)
         np.testing.assert_array_equal(sr.data, sr2.data)
 
-    def test_interactive_snap_False(self):
+    @pytest.mark.parametrize('snap', [True, False, 'default'])
+    def test_interactive_snap(self, snap):
+        kwargs = {}
+        if snap != 'default':
+            kwargs['snap'] = snap
+        else:
+            # default is True
+            snap = True
         s = self.s
         r = RectangularROI(left=3, right=7, top=2, bottom=5)
         s.plot()
-        _ = r.interactive(s, snap=False)
-        r.x = 3.5
-        assert r.x == 3.5
+        _ = r.interactive(s, **kwargs)
         for w in r.widgets:
-            assert not w.snap_all
-            assert not w.snap_position
-            assert not w.snap_size
+            old_position = w.position
+            new_position = (3.25, 2.2)
+            w.position = new_position
+            assert w.position == old_position if snap else new_position
+            assert w.snap_all == snap
+            assert w.snap_position == snap
+            assert w.snap_size == snap
 
         p1 = Point1DROI(4)
-        _ = p1.interactive(s, snap=False)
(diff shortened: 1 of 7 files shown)


hyperspy/tests/utils/test_roi.py
```diff
@@ -548,3 +548,71 @@ def test_interactive(self):
         r.x += 5
         sr2 = r(s)
         np.testing.assert_array_equal(sr.data, sr2.data)
+
+    @pytest.mark.parametrize('snap', [True, False, 'default'])
+    def test_interactive_snap(self, snap):
```
jlaehne, line 553: What additional value does the 'default' option give here? It is changed to `True` in the test and thus does not differ from directly calling the function with `True`.

ericpre: This is to test that the default behaviour is `snap=True`, without setting it explicitly. I agree that this is not really necessary, because it is fairly obvious if the default value in the `interactive` or `add_widget` method is changed in the future! Shall I remove it?

jlaehne: Well, I misunderstood how the test works on first sight. You set `snap=True` in the `else` only for the test and not as part of the arguments. You can leave it as is.

jlaehne:
Looks good and the provided example works nicely.

Only thought I had is that if we go away from snapping to the edge of the bins, whether it would be a chance to directly implement the option discussed in https://github.com/hyperspy/hyperspy/pull/1491#issuecomment-332349731 - to choose whether to snap to center of bin (new variation) or edge of bin (default)? Then @magnunor could finish up #1491 with the correct functionality of how the DM ROIs are interpreted.

ericpre:
> Looks good and the provided example works nicely.
> 
> Only thought I had is that if we go away from snapping to the edge of the bins, whether it would be a chance to directly implement the option discussed in [#1491 (comment)](https://github.com/hyperspy/hyperspy/pull/1491#issuecomment-332349731) - to choose whether to snap to center of bin (new variation) or edge of bin (default)? Then @magnunor could finish up #1491 with the correct functionality of how the DM ROIs are interpreted.

#1491 uses markers and not widgets. The markers are navigation indices dependent and are not interactive; the issue mentioned in #1491 is related to the fact that the marker are positioned according to their values and are not able to be positioned to the edge of the bin and this PR will not help with this issue.

jlaehne (approved):

Issue #1568, opened 2017-04-05
s.get_lines_intensity() results in very large files


thomasaarholt:
I just saved two maps gotten from a EDS SI in HyperspyUI. The maps, Nb and Fe, are both extremely large (180 MB), despite being of shape (163, 153).
I think the issue is that the decomposition that I performed on the original EDS SI is kept in the files afterwards. It would be nice if this was removed on `s.get_lines_intensity`.

thomasaarholt:
Btw, the decomposition was performed via the console in normal hyperspy manner.

francisco-dlp:
This is probably due to what's stored in ``metadata``.

ericpre:
Fixed in #2691.

PR #2691, opened 2021-03-28
Add option not to stack original metadata in hs.stack

### LINKED ISSUE #1398: BUG: O(n) growth of original_metadata with stacking

### LINKED ISSUE #2045: Bug: Memory issues using `map` with LazySignal

### LINKED ISSUE #2536: Large (original_) metadata significantly slowing hyperspy methods

ericpre:
Fixes #2045, #2536, #1398 and #1568.
~The issue mentioned in #1398 is still not fixed...~

### Progress of the PR
- [x] Add option not to copy or load `original_metadata` in `hs.stack` or `hs.load`,
- [x] update docstring,
- [x] update user guide,
- [x] add tests,
- [x] ready for review.

An example to reproduce the issue in #1398
```python
import hyperspy.api as hs

s2 = hs.stack([s]*1000, copy_metadata=True) # if False or an integer, there is no slowdown when displaying the plot
s2.add_poissonian_noise()

lines = ['C_Ka', 'O_Ka', 'F_Ka', 'Na_Ka', 'Mg_Ka',
         'Al_Ka', 'Si_Ka', 'P_Ka', 'Zr_La', 'Nb_La',
         'K_Ka', 'Ca_Ka', 'Ti_Ka', 'Cr_Ka','Fe_Ka',
         'Ni_Ka', 'S_Ka', 'Cl_Ka', 'La_La', 'Ce_La',
         'Nd_La','Cu_Ka', 'Zn_Ka', 'Au_La', 'Th_Ma']
s2.add_lines(lines)
s2.plot(xray_lines=True)
```
In this case, the slowdown comes from the call of `_plot_xray_lines` and surprisingly, it depends on `original_metadata`:
```python
%time s2._plot_xray_lines(True) # ~12 s
```

### DIFF
--- a/doc/user_guide/io.rst
+++ b/doc/user_guide/io.rst
@@ -65,7 +65,7 @@ override this using the ``reader`` keyword:
     >>> s = hs.load("filename.some_extension", reader="hspy")
 
 Some file formats store some extra information about the data (metadata) and
-HyperSpy read most of them and store them in the
+HyperSpy reads most of them and stores them in the
 :py:attr:`~.signal.BaseSignal.original_metadata` attribute. Also, depending on
 the file format, a part of this information will be mapped by HyperSpy to the
 :py:attr:`~.signal.BaseSignal.metadata` attribute, where it can be used by
@@ -74,10 +74,10 @@ e.g. routines operating on the signal. See :ref:`metadata structure
 
 .. note::
 
-    Very large amount of metadata can slow down loading and processing, and
-    loading the :py:attr:`~.signal.BaseSignal.original_metadata` can be disable
+    Extensive metadata can slow down loading and processing, and
+    loading the :py:attr:`~.signal.BaseSignal.original_metadata` can be disabled
     using the ``load_original_metadata`` argument of the :py:func:`~.load`
-    function; in this case the :py:attr:`~.signal.BaseSignal.metadata` will
+    function; in this case, the :py:attr:`~.signal.BaseSignal.metadata` will
     still be populated.
 
 To print the content of the attributes simply use:
--- a/doc/user_guide/io.rst
+++ b/doc/user_guide/io.rst
@@ -64,14 +64,21 @@ override this using the ``reader`` keyword:
     # Load a .hspy file with an unknown extension
     >>> s = hs.load("filename.some_extension", reader="hspy")
 
-Some file formats store some extra information about the data (metadata), which
-can be stored in `attributes` of the signal object. If HyperSpy manages to read
-such metadata, it is stored in the
+Some file formats store some extra information about the data (metadata) and
+HyperSpy read most of them and store them in the
 :py:attr:`~.signal.BaseSignal.original_metadata` attribute. Also, depending on
 the file format, a part of this information will be mapped by HyperSpy to the
 :py:attr:`~.signal.BaseSignal.metadata` attribute, where it can be used by
 e.g. routines operating on the signal. See :ref:`metadata structure
 <metadata_structure>` for details.
+
+.. note::
+
+    Very large amount of metadata can slow down loading and processing, and
+    loading the :py:attr:`~.signal.BaseSignal.original_metadata` can be disable
+    using the ``load_original_metadata`` argument of the :py:func:`~.load`
+    function; in this case the :py:attr:`~.signal.BaseSignal.metadata` will
+    still be populated.
 
 To print the content of the attributes simply use:
 
@@ -144,7 +151,7 @@ Alternatively, regular expression type character classes can be used such as
 
     Wildcards are implemented using ``glob.glob()``, which treats ``*``, ``[``
     and ``]`` as special characters for pattern matching. If your filename or
-    path contains square brackets, you may want to set 
+    path contains square brackets, you may want to set
     ``escape_square_brackets=True``:
 
     .. code-block:: python
--- a/hyperspy/_signals/eds.py
+++ b/hyperspy/_signals/eds.py
@@ -1022,8 +1022,9 @@ class EDS_mixin:
             line_energy.append(self._get_line_energy(xray_line))
             relative_factor = elements_db[element][
                 'Atomic_properties']['Xray_lines'][line]['weight']
-            a_eng = self._get_line_energy(element + '_' + line[0] + 'a')
-            intensity.append(self.isig[a_eng].data * relative_factor)
+            a_eng = self._get_line_energy(f'{element}_{line[0]}a')
+            idx = self.axes_manager[-1].value2index(a_eng)
+            intensity.append(self.data[..., idx] * relative_factor)
         for i in range(len(line_energy)):
             line = markers.vertical_line_segment(
                 x=line_energy[i], y1=None, y2=intensity[i] * 0.8)
--- a/hyperspy/_signals/eds.py
+++ b/hyperspy/_signals/eds.py
@@ -1027,7 +1027,7 @@ class EDS_mixin:
             relative_factor = elements_db[element][
                 'Atomic_properties']['Xray_lines'][line]['weight']
             a_eng = self._get_line_energy(f'{element}_{line[0]}a')
-            idx = self.axes_manager[-1].value2index(a_eng)
+            idx = self.axes_manager.signal_axes[0].value2index(a_eng)
             intensity.append(self.data[..., idx] * relative_factor)
         for i in range(len(line_energy)):
             line = markers.vertical_line_segment(
--- a/hyperspy/_signals/eds.py
+++ b/hyperspy/_signals/eds.py
@@ -84,8 +84,8 @@ class EDS_mixin:
             else:
                 raise NotImplementedError(
                     "This method only works for EDS_TEM or EDS_SEM signals. "
-                    "You can use `set_signal_type(\"EDS_TEM\")` or"
-                    "`set_signal_type(\"EDS_SEM\")` to convert to one of these"
+                    "You can use `set_signal_type('EDS_TEM')` or"
+                    "`set_signal_type('EDS_SEM')` to convert to one of these"
                     "signal types.")
         line_energy = utils_eds._get_energy_xray_line(Xray_line)
         if units_name == 'eV':
@@ -99,11 +99,10 @@ class EDS_mixin:
                                                          line_energy)
         else:
             raise ValueError(
-                "%s is not a valid units for the energy axis. "
+                f"{units_name} is not a valid units for the energy axis. "
                 "Only `eV` and `keV` are supported. "
                 "If `s` is the variable containing this EDS spectrum:\n "
-                ">>> s.axes_manager.signal_axes[0].units = \'keV\' \n"
-                % units_name)
+                ">>> s.axes_manager.signal_axes[0].units = 'keV' \n")
         if FWHM_MnKa is None:
             return line_energy
         else:
@@ -287,7 +286,7 @@ class EDS_mixin:
                 elements_.add(element)
             else:
                 raise ValueError(
-                    "%s is not a valid chemical element symbol." % element)
+                    f"{element} is not a valid chemical element symbol.")
         self.metadata.set_item('Sample.elements', sorted(list(elements_)))
 
     def _get_xray_lines(self, xray_lines=None, only_one=None,
@@ -432,15 +431,15 @@ class EDS_mixin:
                     lines_len = len(xray_lines)
                     xray_lines.add(line)
                     if lines_len != len(xray_lines):
-                        _logger.info("%s line added," % line)
+                        _logger.info(f"{line} line added,")
                     else:
-                        _logger.info("%s line already in." % line)
+                        _logger.info(f"{line} line already in.")
                 else:
                     raise ValueError(
-                        "%s is not a valid line of %s." % (line, element))
+                        f"{line} is not a valid line of {element}.")
             else:
                 raise ValueError(
-                    "%s is not a valid symbol of an element." % element)
+                    f"{element} is not a valid symbol of an element.")
         xray_not_here = self._get_xray_lines_in_spectral_range(xray_lines)[1]
         for xray in xray_not_here:
             warnings.warn(f"{xray} is not in the data energy range.",
@@ -517,9 +516,8 @@ class EDS_mixin:
                 element_lines = [element_lines[select_this], ]
 
             if not element_lines:
-                _logger.info(
-                    ("There is no X-ray line for element %s " % element) +
-                    "in the data spectral range")
+                _logger.info("There is no X-ray line for element {element} "
+                             "in the data spectral range")
             else:
                 lines.extend(element_lines)
         lines.sort()
@@ -532,10 +530,9 @@ class EDS_mixin:
         xray_lines, xray_not_here = self._get_xray_lines_in_spectral_range(
             xray_lines)
         for xray in xray_not_here:
-            warnings.warn("%s is not in the data energy range." % xray +
-                          "You can remove it with" +
-                          "s.metadata.Sample.xray_lines.remove('%s')"
-                          % xray)
+            warnings.warn(f"{xray} is not in the data energy range. "
+                          "You can remove it with: "
+                          "`s.metadata.Sample.xray_lines.remove('{xray}')`")
         return xray_lines
 
     def get_lines_intensity(self,
@@ -632,7 +629,7 @@ class EDS_mixin:
 
             raise TypeError(
                 "xray_lines must be a compatible iterable, but was "
-                "mistakenly provided as a %s." % type(xray_lines))
+                "mistakenly provided as a {type(xray_lines)}.")
 
         xray_lines = self._parse_xray_lines(xray_lines, only_one, only_lines)
         if hasattr(integration_windows, '__iter__') is False:
@@ -668,19 +665,13 @@ class EDS_mixin:
                     (indexes[1] - indexes[0]) + (indexes[3] - indexes[2]))
                 img = img - (bck1 + bck2) * corr_factor
             img.metadata.General.title = (
-                'X-ray line intensity of %s: %s at %.2f %s' %
-                (self.metadata.General.title,
-                 Xray_line,
-                 line_energy,
-                 self.axes_manager.signal_axes[0].units,
-                 ))
+                f'X-ray line intensity of {self.metadata.General.title}: '
+                f'{Xray_line} at {line_energy:.2f} '
+                f'{self.axes_manager.signal_axes[0].units}')
             img.axes_manager.set_signal_dimension(0)
             if plot_result and img.axes_manager.navigation_size == 1:
-                print("%s at %s %s : Intensity = %.2f"
-                      % (Xray_line,
-                         line_energy,
-                         ax.units,
-                         img.data))
+                print(f"{Xray_line} at {line_energy} {ax.units} : "
+                      f"Intensity = {img.data:.2f}")
             img.metadata.set_item("Sample.elements", ([element]))
             img.metadata.set_item("Sample.xray_lines", ([Xray_line]))
             intensities.append(img)
@@ -947,15 +938,15 @@ class EDS_mixin:
     def _plot_xray_lines(self, xray_lines=False, only_lines=("a", "b"),
                          only_one=False, background_windows=None,
                          integration_windows=None, render_figure=True):
-        if (xray_lines is not False or\
-                background_windows is not None or\
-                integration_windows is not None:
+        if (xray_lines is not False or
+                background_windows is not None or
+                integration_windows is not None):
             if xray_lines is False:
                 xray_lines = True
             only_lines = utils_eds._parse_only_lines(only_lines)
             if xray_lines is True or xray_lines == 'from_elements':
-                if 'Sample.xray_lines' in self.metadata \
-                        and xray_lines != 'from_elements':
+                if ('Sample.xray_lines' in self.metadata and
+                        xray_lines != 'from_elements'):
                     xray_lines = self.metadata.Sample.xray_lines
                 elif 'Sample.elements' in self.metadata:
                     xray_lines = self._get_lines_from_elements(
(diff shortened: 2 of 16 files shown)


ericpre:
> In this case, the slowdown comes from the call of `_plot_xray_lines` and surprisingly, it depends on `original_metadata`:
> 
> ```python
> %time s2._plot_xray_lines(True) # ~12 s
> ```

The reason was the use of `isig`, which suffers from the issue with large `original_metadata`... It should be possible to mitigate this issue by making a shallow copy of the `original_metadata`.

francisco-dlp:
Another approach would be never copying ``original_metadata``. It actually doesn't make sense doing so: hyperspy doesn't use any information in-there after loading the file, so why keeping it around? For the same reason, we could make the ``load_original_metadata`` default ``False``. If an user insist in carrying ``original_metadata`` around, he can always copy it inside ``metadata``.

conda_environment.yml
```diff
@@ -3,7 +3,7 @@ channels:
 - defaults
 - conda-forge
 dependencies:
-- dask-core >2.0
+- dask-core >2.0,<2021.3.1
```
ericpre: It seems that Anaconda defaults is missing the `dask-core` 2021.3.1 packages and the latest package is picked up from conda-forge, which breaks dask...

ericpre:
> Another approach would be never copying `original_metadata`. It actually doesn't make sense doing so: hyperspy doesn't use any information in-there after loading the file, so why keeping it around? For the same reason, we could make the `load_original_metadata` default `False`. If an user insist in carrying `original_metadata` around, he can always copy it inside `metadata`.

Yes, indeed, we could change the defaults for the next major release, which would also speed up reading some files! The slowdown issue should be fine now, making a shadow copy seems to work well.

conda_environment.yml
ericpre: I have removed the pinning since the package is now available. It seems that it was a server issue.

ericpre:
@francisco-dlp, can you please review this PR? This should be the last one for 1.6.2!

hyperspy/misc/utils.py
```diff
@@ -947,14 +947,20 @@ def stack(signal_list, axis=None, new_axis_name="stack_element", lazy=None,
...
         until it finds a name that is not yet in use.
     lazy : {bool, None}
         Returns a LazySignal if True. If None, only returns lazy result if at
         least one is lazy.
+    copy_metadata : {bool, int}
+        If integer, this value define the index of the signal in ``signal_list``
+        from which the ``original_metadata`` are copied. If ``True``, the
+        ``original_metadata`` and ``metadata`` are stacked and saved in
+        ``original_metadata.stack_elements`` of the returned signal.
+        If False, the ``metadata`` and ``original_metadata`` are not copied.
```
francisco-dlp: This doesn't seem to be correct: the ``metadata`` is always copied (and never stacked) regardless of the value of ``copy_metadata``.

hyperspy/misc/utils.py
```diff
@@ -1051,14 +1057,24 @@ def stack(signal_list, axis=None, new_axis_name="stack_element", lazy=None,
...
+
         signal.get_dimensions_from_data()
-        signal.original_metadata.add_node("stack_elements")
 
-        for i, obj in enumerate(signal_list):
-            signal.original_metadata.stack_elements.add_node(f"element{i}")
-            node = signal.original_metadata.stack_elements[f"element{i}"]
-            node.original_metadata = obj.original_metadata.as_dictionary()
-            node.metadata = obj.metadata.as_dictionary()
+        if (not isinstance(copy_metadata, bool) and
```
francisco-dlp: What happens when it is not a ``bool`` nor an ``int``?

hyperspy/signal.py
```diff
@@ -2529,11 +2529,9 @@ def _to_dictionary(self, add_learning_results=True, add_models=False):
         """
         dic = {'data': self.data,
                'axes': self.axes_manager._get_axes_dicts(),
-               'metadata': self.metadata.deepcopy().as_dictionary(),
-               'original_metadata':
-               self.original_metadata.deepcopy().as_dictionary(),
-               'tmp_parameters':
-               self.tmp_parameters.deepcopy().as_dictionary(),
+               'metadata': self.metadata.as_dictionary(),
```
francisco-dlp: By not deepcopying  ``metadata``, any modification of the resulting dictionary also modifies ``metadata`` in the original signal. This is maybe OKish (``to_dictionary`` suggests that you get the same object in dictionary form), but it would be good to document this new behavior.

hyperspy/signal.py
```diff
@@ -2191,7 +2191,7 @@ def __init__(self, data, **kwds):
                 """, arguments=['obj'])
 
     def _create_metadata(self):
-        self.metadata = DictionaryTreeBrowser()
+        self.metadata = DictionaryTreeBrowser(lazy=False)
```
francisco-dlp: Could you add a comment here explaining why ``lazy`` must be set to ``False``?

hyperspy/_signals/eds.py
```diff
@@ -1022,8 +1022,9 @@ def add_xray_lines_markers(self, xray_lines):
             line_energy.append(self._get_line_energy(xray_line))
             relative_factor = elements_db[element][
                 'Atomic_properties']['Xray_lines'][line]['weight']
-            a_eng = self._get_line_energy(element + '_' + line[0] + 'a')
-            intensity.append(self.isig[a_eng].data * relative_factor)
+            a_eng = self._get_line_energy(f'{element}_{line[0]}a')
+            idx = self.axes_manager[-1].value2index(a_eng)
+            intensity.append(self.data[..., idx] * relative_factor)
```
francisco-dlp, line 1031: Although uncommon, it is possible to get a signal whose signal axis is not -1. So, maybe get the signal axis index to make this foolproof?

hyperspy/signal.py
```diff
@@ -4773,10 +4785,12 @@ def copy(self):
             self._plot = backup_plot
 
     def __deepcopy__(self, memo):
-        dc = type(self)(**self._to_dictionary())
+        dc = type(self)(**self._to_dictionary(add_original_metadata=False))
         if isinstance(dc.data, np.ndarray):
             dc.data = dc.data.copy()
 
+        dc.original_metadata = self.original_metadata.copy()
```
francisco-dlp: I find this counterintuitive: I think that users would expect that ``deepcopy`` deepcopies everything. That's documented in the ``deepcopy`` method below, but this may be frustrating for users who fail to notice the exception made with ``original_metadata``.

francisco-dlp (commented):
In addition to the comments above, I wonder if shallow copying ``original_metadata`` is the best approach to fix this since it may be counterintuitive and lead to hard to find bugs in user's scripts. An alternative comes to my mind:

1. Adding a ``load_original_metadata`` kwarg to ``load``.
2. Documenting the performance issue with ``original_metadata`` and explain that
    1. It is possible to not load the ``original_metadata``
    2. It can be deleted using something like
        ```python
        s0.original_metadata = type(s0.original_metadata)()
        ```
        That's a big ugly ... maybe addnig a ``reset()`` method to DTB would be better?

hyperspy/signal.py
ericpre: Yes, I thought the same: with deepcopy, we would expect a new copy, however, the [python documentation](https://docs.python.org/3/library/copy.html) implies that not everything is _always_ copied. A more convincing point is that `original_metadata` implies that these metadata are original and therefore not expected to be changed? Maybe the best would be to freeze it.

hyperspy/signal.py
ericpre: `DictionaryTreeBrowser.as_dictionary` creates a new dictionary and as far as I understand there deepcopy is not necessary here.

hyperspy/misc/utils.py
ericpre: I have added a test to check it and it seems fine to me: stacking the `metadata` and adding them in `original_metadata.stack_elements` is the same as before. Maybe the wording is not clear?

hyperspy/misc/utils.py
ericpre: Clarified the if/else logic.

hyperspy/signal.py
ericpre: I have removed it.

hyperspy/_signals/eds.py
ericpre: Done.

ericpre:
> In addition to the comments above, I wonder if shallow copying `original_metadata` is the best approach to fix this since it may be counterintuitive and lead to hard to find bugs in user's scripts. An alternative comes to my mind:
> 
> 1. Adding a `load_original_metadata` kwarg to `load`.
> 2. Documenting the performance issue with `original_metadata` and explain that
>    
>    1. It is possible to not load the `original_metadata`
>    2. It can be deleted using something like
>       ```python
>       s0.original_metadata = type(s0.original_metadata)()
>       ```
>       
>       
>       That's a big ugly ... maybe addnig a `reset()` method to DTB would be better?

I also considered adding an option not load `original_metadata` which would have the benefit of speeding up reading files with large metadata, for example, reading the dataset shared in https://github.com/hyperspy/hyperspy/issues/2536 is slow down because of the metadata. I would prefer not to do it in this PR.

I think that the right solution is to better define what is the purpose of the `original_metadata`. As mentioned [above](https://github.com/hyperspy/hyperspy/pull/2691#discussion_r610006241), it would make more sense to me not to edit the already existing `original_metadata` and enforce it with freezing, which could be implemented when moving away from DTB. This PR only shallow copy the `original_metadata`, and this is not an issue with the `metadata`. What do you think?

hyperspy/signal.py
francisco-dlp: See this example:

```python
>>> s = hs.signals.Signal1D([0])
>>> s.original_metadata.test = [0]
>>> s.original_metadata.as_dictionary()["test"].append(1)
>>> s.original_metadata
└── test = [0, 1]
```

hyperspy/signal.py
francisco-dlp: I am actually not very convinced about any of those 2 arguments. I still think that the expected behavior of ``deepcopy`` is deepcopying everything. As for freezing original metadata, the issue is that often the metadata stored using the microscope's software is wrong, so some may like to use HyperSpy to fix the metadata. By not deepcopying it, somebody could crop the data and amend the ``original_metadata`` in the cropped signal unaware that he's modifying the ``original_metadata`` of the original signal.

ericpre: Since this is too contentious, I have reverted these changes and updated the changelog and summary/title of this PR accordingly.

hyperspy/signal.py
ericpre: OK, yes, indeed, this should be fixed now and I added a test covering it.

ericpre:
This is ready for review, I have remove the shallow copy of `original_metadata` and I suspect that it would be appropriate to tackle this issue when splitting the IO part and define more precisely how `metadata` and `original_metadata` are expected to be used. The alternatives discussed in this PR seems to be suboptimal.

hyperspy/misc/utils.py
```diff
@@ -932,7 +935,7 @@ def closest_power_of_two(n):
 
 
 def stack(signal_list, axis=None, new_axis_name="stack_element", lazy=None,
-          show_progressbar=None, **kwargs):
+          copy_metadata=True, show_progressbar=None, **kwargs):
```
francisco-dlp: What about ``stack_original_metadata`` instead?

ericpre: And keep the same behaviour regarding `metadata`?

francisco-dlp: Sorry, I meant ``stack_metadata``

hyperspy/misc/utils.py
```diff
@@ -1051,14 +1055,25 @@ def stack(signal_list, axis=None, new_axis_name="stack_element", lazy=None,
...
+        if isinstance(copy_metadata, bool):
+            if copy_metadata:
+                signal.original_metadata.add_node('stack_elements')
+                for i, obj in enumerate(signal_list):
+                    signal.original_metadata.stack_elements.add_node(f'element{i}')
+                    node = signal.original_metadata.stack_elements[f'element{i}']
+                    node.original_metadata = obj.original_metadata.copy()
+                    node.metadata = obj.metadata.copy()
+            else:
+                signal.original_metadata = DictionaryTreeBrowser({})
```
francisco-dlp, line 1069: In this case ``metadata`` is lost.

hyperspy/misc/utils.py
ericpre: Done.

francisco-dlp:
I think that the long term solution to this issue is not loading the ``original_metadata`` in the first place. I agree that implementing it in the readers (i.e. truly not loading it at all) should be left for another PR. But what about now adding a ``load_metadata`` keyword to ``load`` not to populate ``Signal.original_metadata`` even if read by the reader? That keyword could be passed to get the infrastructure for the full implementation in-place.

Also, this is a subtle issue, so it may be worth to discuss it in the user guide.

ericpre:
> I think that the long term solution to this issue is not loading the `original_metadata` in the first place. I agree that implementing it in the readers (i.e. truly not loading it at all) should be left for another PR. But what about now adding a `load_metadata` keyword to `load` not to populate `Signal.original_metadata` even if read by the reader? That keyword could be passed to get the infrastructure for the full implementation in-place.
> 
> Also, this is a subtle issue, so it may be worth to discuss it in the user guide.

This should be all done now!

doc/user_guide/io.rst
```diff
@@ -64,15 +64,22 @@ override this using the ``reader`` keyword:
     # Load a .hspy file with an unknown extension
     >>> s = hs.load("filename.some_extension", reader="hspy")
 
-Some file formats store some extra information about the data (metadata), which
-can be stored in `attributes` of the signal object. If HyperSpy manages to read
-such metadata, it is stored in the
+Some file formats store some extra information about the data (metadata) and
+HyperSpy read most of them and store them in the
```
jlaehne: ... reads most of them and stores ...

doc/user_guide/io.rst
```diff
@@ -64,15 +64,22 @@ override this using the ``reader`` keyword:
...
 :py:attr:`~.signal.BaseSignal.original_metadata` attribute. Also, depending on
 the file format, a part of this information will be mapped by HyperSpy to the
 :py:attr:`~.signal.BaseSignal.metadata` attribute, where it can be used by
 e.g. routines operating on the signal. See :ref:`metadata structure
 <metadata_structure>` for details.
 
+.. note::
+
+    Very large amount of metadata can slow down loading and processing, and
```
jlaehne: Very large amount of -> Extensive

doc/user_guide/io.rst
```diff
@@ -64,15 +64,22 @@ override this using the ``reader`` keyword:
...
+    loading the :py:attr:`~.signal.BaseSignal.original_metadata` can be disable
```
jlaehne: disabled

doc/user_guide/io.rst
```diff
@@ -64,15 +64,22 @@ override this using the ``reader`` keyword:
...
+    using the ``load_original_metadata`` argument of the :py:func:`~.load`
+    function; in this case the :py:attr:`~.signal.BaseSignal.metadata` will
```
jlaehne: in this case,

hyperspy/io.py
```diff
@@ -188,6 +194,11 @@ def load(filenames=None,
         then square brackets are escaped before wildcard matching with
         ``glob.glob()``. If False, square brackets are used to represent
         character classes (e.g. ``[a-z]`` matches lowercase letters).
+    %s
+    %s Only used with ``stack=True``.
+    load_original_metadata : bool
+        If ``True``, all metadata will be added them to ``original_metadata``.
```
jlaehne: If ``True``, all metadata contained in the input file will be added to ``original_metadata``.

hyperspy/io.py
```diff
@@ -188,6 +194,11 @@ def load(filenames=None,
...
+        This doesn't change the metadata added to ``metadata``.
```
jlaehne: This does not affect the selected metadata, which is added to ``metadata``.

jlaehne (commented):
Some corrections to the documentation and docstrings. I have not found the time to look at the code in detail, but from the discussion with @francisco-dlp it seems that this PR as matured.

ericpre:
I will update the changes from @jlaehne this afternoon. @francisco-dlp, any chance you can have a look at the latest commits today?

francisco-dlp:
Yes, I'll have a look this afternoon.

ericpre:
All done!

hyperspy/io.py
```diff
@@ -482,35 +487,39 @@ def load_with_reader(
...
-            objects[-1].tmp_parameters.folder = folder
-            objects[-1].tmp_parameters.filename = filename
-            objects[-1].tmp_parameters.extension = extension.replace('.', '')
+            signal.tmp_parameters.folder = folder
+            signal.tmp_parameters.filename = filename
+            signal.tmp_parameters.extension = extension.replace('.', '')
             if convert_units:
-                objects[-1].axes_manager.convert_units()
+                signal.axes_manager.convert_units()
+            if not load_original_metadata:
```
francisco-dlp, line 513: It would have been better to modify ``dict2signal()`` in order to load ``original_metadata`` into a DTB. However, that's not trivial, so this is a good first step. Maybe we could open an issue about it not to forget it?

ericpre: Yes, I agree and there is already an open PR on this actually: https://github.com/hyperspy/hyperspy/pull/2092!

francisco-dlp: Good point!

PR #2693, opened 2021-03-31
Non uniform axis updates

jlaehne:
### Description of the change
Fixes for non_uniform_axes:
- update `check_axes_calibration` originally included in #2549 + tests, 
- https://github.com/hyperspy/hyperspy/pull/2399#discussion_r605010552 and https://github.com/hyperspy/hyperspy/pull/2399#discussion_r605036572
- Add docstrings for axes classes
- Increase coverage for axes.py

### Progress of the PR
- [x] add docstrings for axes classes,
- [x] increase coverage,
- [x] ready for review.


hyperspy/misc/axis_tools.py
```diff
@@ -20,7 +20,7 @@
 
 
 def check_axes_calibration(ax1, ax2):
-    """Check if the calibration of two DataAxis objects matches.
+    """Check if the calibration of two Axis objects matches.
```
ericpre, line 24: This function is used only once in hyperspy, would it be more simple to remove this function and instead use `np.allclose(ax1.axis, ax2.axis)`? This approach would be more transparent and would have the advantage of specifying the tolerance, when necessary.

ericpre (commented):
I suggested an alternative to the `check_axes_calibration` function.
While you are at it, could you increase a bit more the coverage, as pointed out by the github annotations. These should be simple to cover.

Maybe for another PR, it seems that some axis documentation in the user guide are redundant in `getting_started.rst and `axes.rst` as highlighted by the github annotations at the bottom of the "Files changes" tab.

hyperspy/misc/axis_tools.py
jlaehne: The tolerance could be added here, but it is true that as all three axes types do have an axis array as attribute, the simple comparison of these may be sufficient, but probably still checking units afterwards as well?

jlaehne: `np.allclose` fails on axes of different size that are compared, so I think having this function still makes sense, but I have simplified it extremely by comparing the axes vectors.

jlaehne:
> I suggested an alternative to the `check_axes_calibration` function.

I have made the function more slim, as discussed above.

> While you are at it, could you increase a bit more the coverage, as pointed out by the github annotations. These should be simple to cover.
 
There are somehow no uncovered lines mentioned any more.

> Maybe for another PR, it seems that some axis documentation in the user guide are redundant in `getting_started.rst and `axes.rst` as highlighted by the github annotations at the bottom of the "Files changes" tab.

That must have been some merge commit of RELEASE_next_minor that reintroduced these into `getting_started.rst`, originally these sections were completely moved. As the blame shows no recent changes to these sections, I have taken them out of that file again. We might have missed that without the annotations ;-)

hyperspy/misc/axis_tools.py
```diff
@@ -29,21 +29,19 @@ def check_axes_calibration(ax1, ax2):
...
-    if not np.allclose(ax1.offset, ax2.offset, atol=0, rtol=1e-7):
-        return False
-
-    if ax1.units != ax2.units:
-        return False
-
-    return True
+    if ax1.size == ax2.size:
+        if np.allclose(ax1.axis, ax2.axis, atol=0, rtol=rtol) and\
+            ax1.units == ax2.units:
```
ericpre: In case the units are the same but different string (`s` versus `second` for example), is it worth using pint to compare the units here?

ericpre (commented):
Looks good to me, I added a suggestion for the units comparison.

hyperspy/misc/axis_tools.py
jlaehne: Why not ;-)

jlaehne:
> Looks good to me, I added a suggestion for the units comparison.

I took up the suggestion and would propose to merge, if you do not object.

ericpre (approved):

Issue #1840, opened 2018-02-18
Direct processing with map and lazy signals


magnunor:
When working with very large datasets, it is nice to use lazy signals to avoid running out of memory.

However, using `map` to calculate the centre of mass takes a very long time compared to calculating the centre of mass directly from the dask array. Example of this:

```python
import time
import numpy as np
from scipy.ndimage import center_of_mass
import dask.array as da
import hyperspy.api as hs
from hyperspy._lazy_signals import LazySignal2D

def center_of_mass_frame(im):
    y, x = center_of_mass(im)
    return np.array([x, y])

data = da.random.random((128, 128, 256, 256), chunks=(1, 1, 256, 256))
s = LazySignal2D(data)

time0 = time.time()
s_com0 = s.map(
        function=center_of_mass_frame,
        ragged=False, inplace=False)
s_com0.compute()
s_com0 = s_com0.T
print(time.time() - time0)

###########
def calculate_directly(signal):
    size_x, size_y = signal.axes_manager.navigation_shape
    data = np.zeros((2, size_y, size_x))
    for ix in range(size_x):
        for iy in range(size_y):
            im_dask = signal.data[iy, ix]
            im = im_dask.compute()
            data[:, iy, ix] = center_of_mass_frame(im)
    s = hs.signals.Signal2D(data)
    return s

data = da.random.random((128, 128, 256, 256), chunks=(1, 1, 256, 256))
s = LazySignal2D(data)

time0 = time.time()
s_com1 = calculate_directly(s)
print(time.time() - time0)
```

The one using the `map` method uses about 530 seconds, while the one using the more "direct" function uses about 56 seconds.

I'm guessing the big difference is due to the "direct" method not having to pre-calculate the "history tree" (See the [HyperSpy docs](http://hyperspy.org/hyperspy-doc/current/user_guide/big_data.html#behind-the-scenes-technical-details) for more info).

Is this type of "direct" calculation already possible in HyperSpy? If not, would it be an idea to implement this? Since it could be very useful for a large range of data processing, especially where the size of the output data is significantly smaller than the original data. For example virtual apertures, centre of mass, ...

I've already implemented simple version of this in the [fpd_data_processing](https://gitlab.com/fast_pixelated_detectors/fpd_data_processing/blob/master/fpd_data_processing/lazy_tools.py#L52) library, which also takes into account the data chunking.

sk1p:
[...]

PR #2703, opened 2021-04-11
Improve non-lazy map method

magnunor:
**Note: this text was updated to reflect the current changes in this pull request.**

Recently, the lazy version of `s.map` was greatly improved (#2617). This was done by changing the `s._map_iterate` method for lazy signals, by utilizing `da.map_blocks`. This has several advantages, like i) much faster calculations, ii) the ability to get lazy outputs from `s.map` (i.e. not doing the calculation right away).

So currently, there are many different methods which can be called by `s.map`:
- Non-lazy signal:
	- `hyperspy/signal.py` : `_map_all` or `_map_iterate` or 
- Lazy signal:
	- `hyperspy/_signals/lazy.py` : `_map_all` or `_map_iterate` (the one which was improved in #2617)

-----------------
## Changes
This pull request changes:

- "Move" the lazy version of `_map_iterate` into (the non-lazy) `BaseSignal`
- Lazy signals will now inherit both `map` and `_map_iterate` from `BaseSignal`
- For lazy signals: remove `_map_all`
- Add `lazy_output` to `map`, so that non-lazy signals can get a lazy result from `map`
- Do various improvements to `_map_iterate`, to make it more robust

----------------
## Potentially breaking changes

As `s.map` is _very_ central to HyperSpy and related libraries, this needs to be tested carefully. For the other libraries this is done automatically via the new extension testing in HyperSpy.

- [x] Test with HyperSpy's unit tests
- [ ] Test with pyXem
- [x] Test with Atomap
- [x] Test with LumiSpy
- [x] Test with KikuchiPy
- Any other libraries?

----------------
## Benchmarking

One potential downside with these changes is worse performance when using `s.map`. Since removing the direct NumPy calculations, _might_ slow things down. Thus, this needs to be thoroughly checked.

----------------

## TODO

- [x] Finish implemenation
- [x] Fix failing unit test
- [x] Code style fixes
- [x] Remove methods/functions which are no longer used
- [x] Add docstring
- [x] Update user guide: `lazy_result` parameter
- [x] Update changelog

### DIFF
--- a/hyperspy/misc/utils.py
+++ b/hyperspy/misc/utils.py
@@ -1152,40 +1152,6 @@ def transpose(*args, signal_axes=None, navigation_axes=None, optimize=False):
                           optimize=optimize) for sig in args]
 
 
-def create_map_objects(function, nav_size, iterating_kwargs, **kwargs):
-    """To be used in _map_iterate of BaseSignal and LazySignal.
-
-    Moved to a separate method to reduce code duplication.
-    """
-    from hyperspy.signal import BaseSignal
-    from itertools import repeat
-
-    iterators = tuple(iterating_kwargs[key]._cycle_signal()
-                      if isinstance(iterating_kwargs[key], BaseSignal) else iterating_kwargs[key]
-                      for key in iterating_kwargs)
-    # make all kwargs iterating for simplicity:
-    iterating = tuple(key for key in iterating_kwargs)
-    for k, v in kwargs.items():
-        if k not in iterating:
-            iterating += k,
-            iterators += repeat(v, nav_size),
-
-    def figure_out_kwargs(data):
-        _kwargs = {k: v for k, v in zip(iterating, data[1:])}
-        for k in iterating_kwargs:
-            if (isinstance(iterating_kwargs[k], BaseSignal) and
-                isinstance(_kwargs[k], np.ndarray) and
-                    len(_kwargs[k]) == 1):
-                _kwargs[k] = _kwargs[k][0]
-        return data[0], _kwargs
-
-    def func(*args):
-        dat, these_kwargs = figure_out_kwargs(*args)
-        return function(dat, **these_kwargs)
-
-    return func, iterators
-
-
 def process_function_blockwise(data,
                                *args,
                                function,
@@ -1280,45 +1246,6 @@ def guess_output_signal_size(test_signal,
         output_dtype = output.dtype
         output_signal_size = output.shape
     return output_signal_size, output_dtype
-
-
-def map_result_construction(signal,
-                            inplace,
-                            result,
-                            ragged,
-                            sig_shape=None,
-                            lazy=False):
-    res = None
-    if inplace:
-        sig = signal
-    else:
-        res = sig = signal._deepcopy_with_new_data()
-
-    if ragged:
-        axes_dicts = signal.axes_manager._get_navigation_axes_dicts()
-        sig.axes_manager.__init__(axes_dicts)
-        sig.axes_manager._ragged = True
-        sig.data = result
-        sig._assign_subclass()
-    else:
-        if not sig._lazy and sig.data.shape == result.shape and np.can_cast(
-                result.dtype, sig.data.dtype):
-            sig.data[:] = result
-        else:
-            sig.data = result
-
-        # remove if too many axes
-        sig.axes_manager.remove(sig.axes_manager.signal_axes[len(sig_shape):])
-        # add additional required axes
-        for ind in range(
-                len(sig_shape) - sig.axes_manager.signal_dimension, 0, -1):
-            sig.axes_manager._append_axis(size=sig_shape[-ind], navigate=False)
-
-        sig.get_dimensions_from_data()
-        if not sig.axes_manager._axes:
-            add_scalar_axis(sig, lazy=lazy)
-
-    return res
 
 
 def multiply(iterable):
--- a/hyperspy/misc/utils.py
+++ b/hyperspy/misc/utils.py
@@ -1198,17 +1198,20 @@ def process_function_blockwise(data,
         # There aren't any BaseSignals for iterating
         for nav_index in np.ndindex(chunk_nav_shape):
             islice = np.s_[nav_index]
-            output_array[islice] = function(data[islice],
-                                            **kwargs)
+            output_array[islice] = function(data[islice], **kwargs)
     else:
         # There are BaseSignals which iterate alongside the data
         for index in np.ndindex(chunk_nav_shape):
             islice = np.s_[index]
+            iter_dict = {}
+            for key, a in zip(arg_keys, args):
+                arg_i = a[islice].squeeze()
+                # Some functions does not handle 0-dimension NumPy arrys
+                if arg_i.shape == ():
+                    arg_i = arg_i[()]
+                iter_dict[key] = arg_i
 
-            iter_dict = {key: a[islice].squeeze() for key, a in zip(arg_keys,args)}
-            output_array[islice] = function(data[islice],
-                                            **iter_dict,
-                                            **kwargs)
+            output_array[islice] = function(data[islice], **iter_dict, **kwargs)
     if not (chunk_nav_shape == output_array.shape):
         try:
             output_array = output_array.squeeze(-1)
@@ -1217,34 +1220,39 @@ def process_function_blockwise(data,
     return output_array
 
 
-def guess_output_signal_size(test_signal,
+def guess_output_signal_size(test_data,
                              function,
                              ragged,
                              **kwargs):
     """This function is for guessing the output signal shape and size.
-    It will attempt to apply the function to some test signal and then output
+    It will attempt to apply the function to some test data and then output
     the resulting signal shape and datatype.
 
     Parameters
     ----------
-    test_signal: BaseSignal
-        A test signal for the function to be applied to. A signal
-        with 0 navigation dimensions
-    function: function
+    test_data : NumPy array
+        Data from a test signal for the function to be applied to.
+        The data must be from a signal with 0 navigation dimensions.
+    function : function
         The function to be applied to the data
-    ragged: bool
+    ragged : bool
         If the data is ragged then the output signal size is () and the
         data type is 'object'
-    **kwargs: dict
+    **kwargs : dict
         Any other keyword arguments passed to the function.
     """
     if ragged:
         output_dtype = object
         output_signal_size = ()
     else:
-        output = function(test_signal, **kwargs)
-        output_dtype = output.dtype
-        output_signal_size = output.shape
+        output = function(test_data, **kwargs)
+        try:
+            output_dtype = output.dtype
+            output_signal_size = output.shape
+        except AttributeError:
+            output = np.asarray(output)
+            output_dtype = output.dtype
+            output_signal_size = output.shape
     return output_signal_size, output_dtype
 
 
--- a/hyperspy/signal.py
+++ b/hyperspy/signal.py
@@ -55,10 +55,12 @@ from hyperspy.drawing.marker import markers_metadata_dict_to_markers
 from hyperspy.misc.slicing import SpecialSlicers, FancySlicing
 from hyperspy.misc.utils import slugify
 from hyperspy.misc.utils import is_binned # remove in v2.0
+from hyperspy.misc.utils import process_function_blockwise, guess_output_signal_size
+from hyperspy.misc.utils import add_scalar_axis
 from hyperspy.docstrings.signal import (
     ONE_AXIS_PARAMETER, MANY_AXIS_PARAMETER, OUT_ARG, NAN_FUNC, OPTIMIZE_ARG,
     RECHUNK_ARG, SHOW_PROGRESSBAR_ARG, PARALLEL_ARG, MAX_WORKERS_ARG,
-    CLUSTER_SIGNALS_ARG, HISTOGRAM_BIN_ARGS, HISTOGRAM_MAX_BIN_ARGS)
+    CLUSTER_SIGNALS_ARG, HISTOGRAM_BIN_ARGS, HISTOGRAM_MAX_BIN_ARGS, LAZY_OUTPUT_ARG)
 from hyperspy.docstrings.plot import (BASE_PLOT_DOCSTRING, PLOT1D_DOCSTRING,
                                       BASE_PLOT_DOCSTRING_PARAMETERS,
                                       PLOT2D_KWARGS_DOCSTRING)
@@ -3659,7 +3661,6 @@ class BaseSignal(FancySlicing,
         if not self.axes_manager._axes and not self.ragged:
             # Create a "Scalar" axis because the axis is the last one left and
             # HyperSpy does not # support 0 dimensions
-            from hyperspy.misc.utils import add_scalar_axis
             add_scalar_axis(self)
 
     def _ma_workaround(self, s, function, axes, ar_axes, out):
@@ -4648,7 +4649,8 @@ class BaseSignal(FancySlicing,
         ragged=None,
         output_signal_size=None,
         output_dtype=None,
-        **kwargs
+        lazy_output=None,
+        **kwargs,
     ):
         """Apply a function to the signal data at all the navigation
         coordinates.
@@ -4669,19 +4671,37 @@ class BaseSignal(FancySlicing,
         ----------
 
         function : :std:term:`function`
-            Any function that can be applied to the signal.
-        %s
+            Any function that can be applied to the signal. This function should
+            not alter any mutable input arguments or input data. So do not do
+            operations which alter the input, without copying it first.
+            For example, instead of doing `image *= mask`, rather do
+            `image = image * mask`. Likewise, do not do `image[5, 5] = 10`
+            directly on the input data or arguments, but make a copy of it
+            first. For example via `image = copy.deepcopy(image)`.
         %s
         %s
         inplace : bool, default True
-            if ``True``, the data is replaced by the result. Otherwise
+            If ``True``, the data is replaced by the result. Otherwise
             a new Signal with the results is returned.
         ragged : None or bool, default None
             Indicates if the results for each navigation pixel are of identical
             shape (and/or numpy arrays to begin with). If ``None``,
-            the appropriate choice is made while processing. If True in case
-            of lazy signal, the signal will be compute at the end of the
-            mapping. Note: ``None`` is not allowed for Lazy signals!
+            the output signal will be ragged only if the original signal is ragged.
+        output_signal_size : None, tuple
+            Since the size and dtype of the signal dimension of the output
+            signal can be different from the input signal, this output signal
+            size must be calculated somehow. If both ``output_signal_size``
+            and ``output_dtype`` is ``None``, this is automatically determined.
+            However, if for some reason this is not working correctly, this
+            can be specified via ``output_signal_size`` and ``output_dtype``.
+            The most common reason for this failing is due to the signal size
+            being different for different navigation positions. If this is the
+            case, use ragged=True. None is default.
+        output_dtype : None, NumPy dtype
+            See docstring for output_signal_size for more information.
+            Default None.
+        %s
+        %s
         **kwargs : dict
             All extra keyword arguments are passed to the provided function
 
@@ -4712,36 +4732,60 @@ class BaseSignal(FancySlicing,
         parameter is variable:
 
         >>> im = hs.signals.Signal2D(np.random.random((10, 64, 64)))
-        >>> sigmas = hs.signals.BaseSignal(np.linspace(2,5,10)).T
+        >>> sigmas = hs.signals.BaseSignal(np.linspace(2, 5, 10)).T
         >>> im.map(scipy.ndimage.gaussian_filter, sigma=sigmas)
+
+        Rotate the two signal dimensions, with different amount as a function
+        of navigation index. Delay the calculation by getting the output
+        lazily. The calculation is then done using the compute method.
+
+        >>> from scipy.ndimage import rotate
+        >>> s = hs.signals.Signal2D(np.random.random((5, 4, 40, 40)))
+        >>> s_angle = hs.signals.BaseSignal(np.linspace(0, 90, 20).reshape(5, 4)).T
+        >>> s.map(rotate, angle=s_angle, reshape=False, lazy_output=True)
+        >>> s.compute()
+
+        Rotate the two signal dimensions, with different amount as a function
+        of navigation index. In addition, the output is returned as a new
+        signal, instead of replacing the old signal.
+
+        >>> s = hs.signals.Signal2D(np.random.random((5, 4, 40, 40)))
+        >>> s_angle = hs.signals.BaseSignal(np.linspace(0, 90, 20).reshape(5, 4)).T
+        >>> s_rot = s.map(rotate, angle=s_angle, reshape=False, inplace=False)
 
         Note
         ----
         Currently requires a uniform axis.
 
         """
-        if self.axes_manager.navigation_shape == () and self._lazy:
-            _logger.info("Converting signal to a non-lazy signal because there are no nav dimensions")
-            self.compute()
-
-        # Sepate ndkwargs depending on if they are BaseSignals.
+        if lazy_output is None:
+            lazy_output = self._lazy
+        if ragged is None:
+            ragged = self.ragged
+        # Separate ndkwargs depending on if they are BaseSignals.
+        self_nav_shape = self.axes_manager.navigation_shape
         ndkwargs = {}
         ndkeys = [key for key in kwargs if isinstance(kwargs[key], BaseSignal)]
         for key in ndkeys:
-            if kwargs[key].axes_manager.navigation_shape == self.axes_manager.navigation_shape:
+            nd_nav_shape = kwargs[key].axes_manager.navigation_shape
+            if nd_nav_shape == self_nav_shape:
                 ndkwargs[key] = kwargs.pop(key)
-            elif kwargs[key].axes_manager.navigation_shape == () or kwargs[key].axes_manager.navigation_shape == (1,):
-                kwargs[key] = np.squeeze(kwargs[key].data)  # this really isn't an iterating signal.
+            elif nd_nav_shape == () or nd_nav_shape == (1,):
+                # This really isn't an iterating signal.
+                kwargs[key] = np.squeeze(kwargs[key].data)
             else:
-                raise ValueError(f'The size of the navigation_shape for the kwarg {key} '
-                                 f'(<{kwargs[key].axes_manager.navigation_shape}> must be consistent'
-                                 f'with the size of the mapped signal '
-                                 f'<{self.axes_manager.navigation_shape}>')
+                raise ValueError(
+                    f"The size of the navigation_shape for the kwarg {key} "
+                    f"(<{nd_nav_shape}> must be consistent "
+                    f"with the size of the mapped signal "
+                    f"<{self_nav_shape}>"
+                )
         # TODO: Consider support for non-uniform signal axis
         if any([not ax.is_uniform for ax in self.axes_manager.signal_axes]):
             _logger.warning(
                 "At least one axis of the signal is non-uniform. Can your "
-                "`function` operate on non-uniform axes?")
+                "`function` operate on non-uniform axes?"
+            )
         else:
             # Check if the signal axes have inhomogeneous scales and/or units and
             # display in warning if yes.
@@ -4752,9 +4796,9 @@ class BaseSignal(FancySlicing,
                 units.add(self.axes_manager.signal_axes[i].units)
             if len(units) != 1 or len(scale) != 1:
                 _logger.warning(
-                    "The function you applied does not take into "
-                    "account the difference of units and of scales in-between"
-                    " axes.")
+                    "The function you applied does not take into account "
+                    "the difference of units and of scales in-between axes."
+                )
         # If the function has an axis argument and the signal dimension is 1,
         # we suppose that it can operate on the full array and we don't
         # iterate over the coordinates.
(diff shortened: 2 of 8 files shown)


magnunor:
One of the failing tests is essentially due to the test expecting a `BaseSignal` from a signal with one value, but `s._assign_subclass()` makes it into a Signal1D. To replicate this:

```python
s = hs.signals.BaseSignal(1) # <BaseSignal, title: , dimensions: (|1)>
s._assign_subclass() # <Signal1D, title: , dimensions: (|1)>
```
The question is then which signal type _should_ a signal with one value be? `BaseSignal` or `Signal1D`?

hakonanes:
> As s.map is very central to HyperSpy and related libraries, this needs to be tested carefully.
>    Test with KikuchiPy

Thanks for considering kikuchipy, @magnunor! We don't use the `Signal2D.map()` method in kikuchipy yet, we use dask directly for all electron backscatter diffraction (EBSD) pattern processing, so you don't have to worry about us.

ericpre:
> * Remove `signal._map_all` and `_signals.lazy._map_all`

Why do you want to remove `signal._map_all`?

In order to have an idea of how it will compare with the current implementation in the case fast `map` processing, what is the overhead of going from numpy array to dask array?

pc494:
> > * Remove `signal._map_all` and `_signals.lazy._map_all`
> 
> Why do you want to remove `signal._map_all`?
> 
> In order to have an idea of how it will compare with the current implementation in the case fast `map` processing, what is the overhead of going from numpy array to dask array?

Just to add this, my experience so far has suggested mapping numpy arrays throw much more interpretable stacktraces (vs dask). Might be worth bearing in mind

magnunor:
Reminder to myself to include fix for bug when last nav chunk is size 1: https://github.com/hyperspy/hyperspy/issues/2748

francisco-dlp:
@magnunor, since fixes should go  to Rnp, could you submit the fix for #2748 as a separate PR?

magnunor:
This pull request is basically being held up due to two of tests failing due to the issue of Signal1D/BaseSignal with exactly one value in the signal dimension (https://github.com/hyperspy/hyperspy/issues/2766)

Since those two tests check for something which are edge cases, I'm tempted to just change the failing test to something which is less of an edge case, so I can finish this rather big pull request. Especially since this pull request is also holding up several improvements in pyXem.

ericpre:
> > * Remove `signal._map_all` and `_signals.lazy._map_all`
> 
> Why do you want to remove `signal._map_all`?
> 
> In order to have an idea of how it will compare with the current implementation in the case fast `map` processing, what is the overhead of going from numpy array to dask array?

I think it would be useful to have an answer on these points, for example, I am not convinced that we should get ride of `_map_all`, which uses vectorization instead of iteration.

magnunor:
> I think it would be useful to have an answer on these points, for example, I am not convinced that we should get ride of _map_all, which uses vectorization instead of iteration.

The main reason for wanting to remove it, is to have one "data processing stream" with `map`. Having `_map_all` in addition to `_map_iterate` means there will be two separate pathways for the processing, which likely will result in different results in edge cases. Having one stream which is well-tested, is better than having two which are not so well-tested.
However, removing `_map_all` also has some downsides, the main being worse performance on smaller datasets.


----------
I did some benchmarking, comparing the old and new implementation.
For larger datasets, with navigation dimensions, there does not seem to be much difference between the old and new implementations.

However, for the image data (2-dimensions), there are some differences.

## Old implementation
non-lazy, 4096, time:  1.5359
non-lazy, 2048, time:  0.3805
non-lazy, 1024, time:  0.091
non-lazy, 512, time:  0.0234
non-lazy, 256, time:  0.0071
non-lazy, 128, time:  0.0028
non-lazy, 64, time:  0.0017

## New implementation
non-lazy, 4096, time:  1.6605
non-lazy, 2048, time:  0.3873
non-lazy, 1024, time:  0.0977
non-lazy, 512, time:  0.0258
non-lazy, 256, time:  0.0099
non-lazy, 128, time:  0.0059
non-lazy, 64, time:  0.0068

ericpre:
I am in -1 on removing the existing implementation:
- it will not work with cupy array

> Having `_map_all` in addition to `_map_iterate` means there will be two separate pathways for the processing, which likely will result in different results in edge cases. 
- which also means that if there is bug somewhere and using having both implementation will help spotting bugs. The improvement of the lazy implementation of `map` are recent and I would expect these to be less tested by users and more subject to bugs?!
 
> Having one stream which is well-tested, is better than having two which are not so well-tested.

- If this is the current situation, it would better to add the missing test in a separate PR so that when working on this PR, it is easier to notice what are the difference and addressed them where necessary.

Maybe what would will be good here is to add `_map_parallel` which would use the dask approach only when the `parallel` keyword is specified. Then it would be easy to test both implementation simply by using the `pytest.mark.parametrize`.

ericpre:
Another issue with using systematically dask array for in-memory processing is that function using multi-threading may not play as well. Is your benchmark [above](https://github.com/hyperspy/hyperspy/pull/2703#issuecomment-913208200) using multi-threading through the dask scheduler or the function itself? Is it fairly likely in the near future that numpy/scipy will more use multi-threading more widely - for example, scipy started to pythran, which would facilitate multi-threading.
Is it a bit speculative but IMOH it is worth investigating how it behaves.

magnunor:
I re-added `_map_all`, which will we used by `map` in cases where, 

- `parallel=False`
- `lazy_result=False`
- There are no iterating `kwargs`

Basically exactly how it worked before.

I also added some tests for checking that `parallel=False` and `parallel=True` gives the same results.

magnunor:
**One potential API breaking behavior**: the `max_workers` parameter. By replacing `multiprocessing` with `dask for `signal.map`, the `max_workers` parameter seems to become redundant. This is due to the  `dask.array.map_blocks` will using the default scheduler to do the processing. This scheduler (should?) handle the maximum amount of workers. As far as I know, this is typically done more "globally", with the scheduler being set when "setting up" the environment.

I'm not sure how best to handle this, but I think a better way of doing this is by setting a `dask` scheduler when importing hyperspy. Where this scheduler can be tweaked via HyperSpy's config.

With regards to `NumPy/SciPy` threading, I think this can be disabled if we don't want to use it in certain situations?

din14970:
> I'm not sure how best to handle this, but I think a better way of doing this is by setting a dask scheduler when importing hyperspy. 

I think hyperspy should move towards `dask.distributed`; I have a feeling the single machine schedulers will become depricated anyway in favor of the `LocalCluster`. This would also be necessary to fully support lazy cupy arrays, see https://github.com/hyperspy/hyperspy/pull/2670.

ericpre:
> **One potential API breaking behavior**: the `max_workers` parameter. By replacing `multiprocessing` with `dask for `signal.map`, the `max_workers`parameter seems to become redundant. This is due to the `dask.array.map_blocks` will using the default scheduler to do the processing. This scheduler (should?) handle the maximum amount of workers. As far as I know, this is typically done more "globally", with the scheduler being set when "setting up" the environment.
> 
> I'm not sure how best to handle this, but I think a better way of doing this is by setting a `dask` scheduler when importing hyperspy. Where this scheduler can be tweaked via HyperSpy's config.
> 
> With regards to `NumPy/SciPy` threading, I think this can be disabled if we don't want to use it in certain situations?

There are [different ways to set configuration of dask scheduler](https://docs.dask.org/en/latest/scheduling.html?highlight=multiprocessing#configuration) and here I guess it should be possible to keep this argument which would use the current configuation is `max_workers` is not specified, otherwise set the `num_workers` in the `compute` call.
It may be worth considering deprecating it in favour of recommending using `dask.config` context manager, in case users want to specify it. This would have the advantage to be transparent and clear for the users!

magnunor:
I re-added the `max_workers` arguments (much love to `git revert` :) ), by using `num_workers` in `compute`. For the next major release, I think deprecating `max_workers` with configuring a scheduler is a good idea.

Feature-wise, I think this pull request is complete. Still missing some documentation and maybe a couple of extra unit tests.

ericpre:
What are you planning to do for the failing pyxem tests?
The integration test workflow is currently passing because pyxem install overwrite the hyperspy version of this branch. A workaround to this issue would be to install hyperspy a second time in the github workflow but it would be better not to pin hyperspy in pyxem!

din14970:
I'm not really sure why the version of hyperspy was pinned in pyxem in the first place, pyxem tests seem to run fine with the latest version see https://github.com/pyxem/pyxem/pull/779 where I actually accidentally modified setup.py.

pc494:
> What are you planning to do for the failing pyxem tests? The integration test workflow is currently passing because pyxem install overwrite the hyperspy version of this branch. A workaround to this issue would be to install hyperspy a second time in the github workflow but it would be better not to pin hyperspy in pyxem!

pyxem needs to be better about this, I'm about to merge https://github.com/pyxem/pyxem/pull/779 and we will endeavour to have up to date support going forward.

magnunor:
> What are you planning to do for the failing pyxem tests?

I went through some of the failing `pyxem` tests, and at least some of them are due to `_map_iterate` being called directly. I don't really see a way of resolving them, without breaking the current version of `pyxem`. However, the code changes necessary to solve this in `pyxem` should be pretty easy.

So I think the easiest way is to do a `pyxem` release right after the next release of HyperSpy (which includes this pull request).

ericpre:
Some hyperspy tests are failing.

Here is a list of failing pyxem test with b90d7fc:
```
ERROR tests/signals/test_segments.py::TestVDFSegment::test_get_vdf_ncc_matrix[unique_vectors0]
ERROR tests/signals/test_segments.py::TestVDFSegment::test_correlate_segments[unique_vectors0-0.1-1-1]
ERROR tests/signals/test_segments.py::TestVDFSegment::test_correlate_segments[unique_vectors0-0.9-3-2]
ERROR tests/signals/test_segments.py::TestVDFSegment::test_correlate_segments_cropped[unique_vectors0]
ERROR tests/signals/test_segments.py::TestVDFSegment::test_correlate_segments_small_vector_threshold[unique_vectors0]
ERROR tests/signals/test_segments.py::TestVDFSegment::test_correlate_segments_bad_thresholds[unique_vectors0]
ERROR tests/signals/test_segments.py::TestVDFSegment::test_get_virtual_electron_diffraction[unique_vectors0]
ERROR tests/signals/test_segments.py::TestVDFSegment::test_get_virtual_electron_diffraction_no_intensities[unique_vectors0]
ERROR tests/signals/test_segments.py::TestVDFSegment::test_get_virtual_electron_diffraction_for_single_vectors[unique_vectors0]
FAILED tests/generators/test_indexation_generator.py::test_vector_indexation_generator_index_vectors
FAILED tests/generators/test_integration_generator.py::test_integration_generator_summation_method
FAILED tests/generators/test_subpixelrefinement_generator.py::Test_subpixelpeakfinders::test_assertioned_xc[diffraction_vectors0]
FAILED tests/generators/test_subpixelrefinement_generator.py::Test_subpixelpeakfinders::test_assertioned_com[diffraction_vectors0]
FAILED tests/generators/test_subpixelrefinement_generator.py::Test_subpixelpeakfinders::test_log[diffraction_vectors0]
FAILED tests/signals/test_diffraction_vectors.py::test_get_cartesian_coordinates[diffraction_vectors_map0]
FAILED tests/signals/test_diffraction_vectors.py::TestMagnitudes::test_get_magnitudes_map[diffraction_vectors_map0]
FAILED tests/signals/test_diffraction_vectors.py::TestMagnitudes::test_get_magnitude_histogram_map[diffraction_vectors_map0]
FAILED tests/signals/test_diffraction_vectors.py::TestFilterVectors::test_filter_magnitude_map_type[diffraction_vectors_map0]
FAILED tests/signals/test_diffraction_vectors.py::TestFilterVectors::test_filter_magnitude_map[diffraction_vectors_map0]
FAILED tests/signals/test_diffraction_vectors.py::TestFilterVectors::test_filter_detector_edge_map_type[diffraction_vectors_map0]
FAILED tests/signals/test_diffraction_vectors.py::TestFilterVectors::test_filter_detector_edge_map[diffraction_vectors_map0]
FAILED tests/signals/test_diffraction_vectors.py::TestDiffractingPixelsMap::test_get_dpm_values[diffraction_vectors_map0]
FAILED tests/signals/test_diffraction_vectors.py::TestDiffractingPixelsMap::test_get_dpm_type[diffraction_vectors_map0]
FAILED tests/signals/test_diffraction_vectors.py::TestDiffractingPixelsMap::test_get_dpm_title[diffraction_vectors_map0]
FAILED tests/signals/test_diffraction_vectors.py::TestDiffractingPixelsMap::test_get_dpm_in_range[diffraction_vectors_map0]
FAILED tests/signals/test_diffraction_vectors.py::TestDiffractingPixelsMap::test_get_dpm_binary[diffraction_vectors_map0]
FAILED tests/signals/test_electron_diffraction2d.py::TestSimpleMaps::test_apply_affine_transforms_paths
FAILED tests/signals/test_indexation_results.py::test_TemplateMatchingResults_plot_best_results_on_signal
FAILED tests/signals/test_pixelated_stem_class.py::TestDiffraction2DRadialAverage::test_simple
FAILED tests/signals/test_pixelated_stem_class.py::TestDiffraction2DRadialAverage::test_different_shape
FAILED tests/signals/test_pixelated_stem_class.py::TestDiffraction2DRadialAverage::test_nav_0
FAILED tests/signals/test_pixelated_stem_class.py::TestDiffraction2DRadialAverage::test_nav_1
FAILED tests/signals/test_pixelated_stem_class.py::TestDiffraction2DRadialAverage::test_big_value
FAILED tests/signals/test_pixelated_stem_class.py::TestDiffraction2DRadialAverage::test_correct_radius_simple
FAILED tests/signals/test_pixelated_stem_class.py::TestDiffraction2DRadialAverage::test_correct_radius_random
FAILED tests/signals/test_pixelated_stem_class.py::TestDiffraction2DRadialAverage::test_correct_disk_x_y_and_radius_random
FAILED tests/signals/test_pixelated_stem_class.py::TestDiffraction2DAngleSector::test_get_angle_sector_mask_radial_average1
FAILED tests/signals/test_pixelated_stem_class.py::TestAngularSliceRadialAverage::test_same_radius
FAILED tests/signals/test_pixelated_stem_class.py::TestAngularSliceRadialAverage::test_different_radius
FAILED tests/signals/test_pixelated_stem_class.py::TestAngularSliceRadialAverage::test_different_radius_not_square_image
FAILED tests/signals/test_pixelated_stem_class.py::TestAngularSliceRadialAverage::test_slice_overlap
FAILED tests/signals/test_polar_diffraction2d.py::TestPearsonCorrelation::test_pcorrelation_inplace
FAILED tests/signals/test_polar_diffraction2d.py::TestPearsonCorrelation::test_masking_pcorrelation[mask1]
FAILED tests/signals/test_segments.py::TestLearningSegment::test_separate_learning_segments[0.5-1-1-50-10-1-False-False]
FAILED tests/signals/test_virtual_dark_field_image.py::TestVDFImage::test_get_vdf_segments[unique_vectors0-1-1-20-5-1-False-0]
= 36 failed, 1201 passed, 225 skipped, 6 xfailed, 358 warnings, 9 errors in 135.37s (0:02:15) =
```

From a quick look at the pyxem code, [`_map-_iterate`](https://github.com/pyxem/pyxem/search?q=_map_iterate) is used in `Diffraction2D.radial_average` and `PolarDiffraction.get_pearson_correlation`. In other test failures, there are broadcasting or indexing issues.

magnunor:
> From a quick look at the pyxem code, _map-_iterate is used in Diffraction2D.radial_average and PolarDiffraction.get_pearson_correlation. In other test failures, there are broadcasting or indexing issues.

I started working on the pyxem functions which utilize `_map_iterate`, and started a discussion in pyxem's issue tracker: https://github.com/pyxem/pyxem/issues/759#issuecomment-943512074

magnunor:
This pull request is now updated to take the recent addition of `s.ragged` (https://github.com/hyperspy/hyperspy/pull/2842) in account.

The next step is to fix the functions in pyxem which were broken due to this pull request.

hyperspy/signal.py
```diff
@@ -4663,14 +4665,26 @@ def map(
         In general, only navigation axes (order, calibration, and number) are
         guaranteed to be preserved.
 
+        This method, ``map``, use two different methods to do the actual
+        processing: ``_map_all`` and ``_map_iterate``. ``_map_all`` inputs the
```
ericpre: I think that `_map_all` and `_map_iterate` are irrelevant to the users and shouldn't be mentioned here.

hyperspy/signal.py
```diff
@@ -4710,36 +4740,60 @@ def map(
...
+        >>> s.map(rotate, angle=s_angle, reshape=False, lazy_result=True)
+        >>> s.compute()
+
+        Rotate the two signal dimensions, with different amount as a function
+        of navigation index. In addition, the result is returned as a new
+        signal, instead of replacing the old signal.
+
+        >>> s = hs.signals.Signal2D(np.random.random((5, 4, 40, 40)))
+        >>> s_angle = hs.signals.BaseSignal(np.linspace(0, 90, 20).reshape(5, 4)).T
+        >>> s_rot = s.map(rotate, angle=s_angle, reshape=False, inplace=True)
```
```suggestion
        >>> s_rot = s.map(rotate, angle=s_angle, reshape=False, inplace=False)
```

hyperspy/signal.py
```diff
@@ -4641,12 +4642,13 @@ def map(
         function,
         show_progressbar=None,
         parallel=None,
-        max_workers=None,
         inplace=True,
         ragged=None,
         output_signal_size=None,
         output_dtype=None,
-        **kwargs
+        lazy_result=None,
```
ericpre: To be consistent with `output_dtype` and other, use `lazy_output` instead of `lazy_result`?

hyperspy/misc/utils.py
```diff
@@ -1228,21 +1194,22 @@ def process_function_blockwise(data,
     output_shape = chunk_nav_shape + tuple(output_signal_size)
     # Pre-allocating the output array
     output_array = np.empty(output_shape, dtype=dtype)
+    # Using deepcopy to avoid the function changing the original kwargs
```
ericpre: Why would the applied function changes the kwargs? If it does, then it is a bug in the applied function, not here, and I am not convinced that it is a concern of the `map`/`process_function_blockwise` function.
Deepcopying can have an expensive computing cost and it is 3 times here * the number of times process_function_blockwise is called, it can add very bad.

hyperspy/misc/utils.py
```diff
@@ -1276,51 +1243,15 @@ def guess_output_signal_size(test_signal,
         output_dtype = object
         output_signal_size = ()
     else:
-        output = function(test_signal, **kwargs)
+        # copy.deepcopy is used to avoid the function altering
+        # the original kwargs.
+        output = function(test_signal, **copy.deepcopy(kwargs))
+        output = np.asarray(output)
```
ericpre: Why is `np.asarray` necessary?

hyperspy/signal.py
```diff
@@ -4641,12 +4642,13 @@ def map(
...
+        max_workers=None,
```
ericpre: Move it back to its original place to avoid changing order of argument - which is not necessary here.

hyperspy/signal.py
```diff
@@ -4710,36 +4740,60 @@ def map(
...
         """
-        if self.axes_manager.navigation_shape == () and self._lazy:
-            _logger.info("Converting signal to a non-lazy signal because there are no nav dimensions")
-            self.compute()
-
-        # Sepate ndkwargs depending on if they are BaseSignals.
+        if lazy_result is None:
+            lazy_result = self._lazy
+        if ragged is None:
+            ragged = self.ragged
```
ericpre, line 4764: For `ragged=None`, this is not what the docstring describe?
The current behaviour is defined by:
https://github.com/hyperspy/hyperspy/blob/a2e787185743ec1dbae990d67715f2cf828879dc/hyperspy/signal.py#L4978-L4979

ericpre (commented):
Thanks @magnunor, I added some questions/comments. :)
There are 52 commits in this PR, which is a bit too many to keep a sensible git history. Is it possible to rewrite the git history?

hyperspy/signal.py
magnunor: I've been using `lazy_result` for this in `pyxem`, based on @francisco-dlp suggestion a while back (https://github.com/hyperspy/hyperspy/issues/1840#issuecomment-367296689).

So ideally it would be nice to keep that consistent in `s.map` as well.

hyperspy/misc/utils.py
magnunor: Since the output from `function` can be basically anything, for example a tuple, the easiest way to get the dtype and shape of the output object is to convert it into a numpy array. Especially since the output from `function` will be put into a numpy array anyway.

```python
import numpy as np
from scipy.ndimage import center_of_mass
output = center_of_mass(np.ones((10, 10))) # (4.5, 4.5)
```
However, there might be other, better, ways of doing this "numpy-conversion".

hyperspy/misc/utils.py
magnunor: The deepcopy is there to avoid some very strange and non-intuitive "bugs", where mutable objects are passed as arguments to the function used in `s.map`. For example a list or NumPy array. I added the deepcopy after some very strange behavior in a function.

A slightly contrived example of this:

```python
import numpy as np
import dask.array as da
import hyperspy.api as hs
def test_function(image, threshold):
    threshold[0] += 1
    mask = image > threshold[0]
    image *= mask
    return image
temp_data = np.arange(0, 20)
data = np.zeros((10, 10, 20))
for iy, ix in np.ndindex(data.shape[:-1]):
    data[iy, ix] = temp_data
data = da.from_array(data, chunks=(5, 5, 20))
s = hs.signals.Signal1D(data).as_lazy()
threshold = [10, 0, 0, 0]
s_out = s.map(test_function, threshold=threshold, inplace=False, lazy_result=False)
```

Gives the images. **With deepcopy**

031_deepcopy_arg

**Without deepcopy**.

031_not_deepcopy_arg


In the last image here, where the deepcopy was not used, a different mask is applied for each position in each chunk. Leading to the different results.

----------

I do agree that this will have some impact on the performance when using `s.map`. However, for most arguments which are passed, this should be very low. For example if the arguments are a single `int`, `float` or `bool`. If iterating signals are passed as parameters, this will have more of an impact. However, the deepcopy will be limited to the specific navigation index which is being processed.

However, there is at least one situation where the deepcopy will have a pretty severe performance hit: if a very large NumPy array is passed as a non-iterating argument. Here, the whole array will be copied for every navigation position. But I think this an edge case.

Some tests for different dataset sizes:

**Large dataset, 64 x 64 x 512 x 512**

```python
from time import time
import numpy as np
import dask.array as da
import hyperspy.api as hs
def testfunction(image, mask):
    return image * mask
nav_x, nav_y = 64, 64
data = np.random.randint(0, 100, size=(nav_y, nav_x, 512, 512))
s = hs.signals.Signal2D(data)
data_mask = np.zeros((nav_y, nav_x, 512, 512), dtype=bool)
mask = np.zeros((512, 512), dtype=bool)
mask[0:20, 0:10] = True
for iy, ix in np.ndindex(nav_y, nav_x):
    data_mask[iy, ix, :, :] = mask
s_mask = hs.signals.Signal2D(data_mask)
###################################
s_out = s.map(testfunction, mask=s_mask, lazy_result=True, inplace=False)
###################################
t0 = time()
s_out.compute()
print(time() - t0)
# With deepcopy: 2.16 s, 2.15 s, 2.29 s, 2.29 s
# Without deepcopy: 2.06 s, 2.19 s, 2.2 s, 2.05 s
```

So the code with deepcopy is about 5% slower here.

**Small dataset, 2048 x 2048**

```python
from time import time
import numpy as np
import hyperspy.api as hs
def testfunction(image, mask):
    return image * mask
data = np.random.randint(0, 100, size=(2048, 2048))
s = hs.signals.Signal2D(data)
data_mask = np.zeros((2048, 2048), dtype=np.float64)
###################################
s_out = s.map(testfunction, mask=data_mask, lazy_result=True, inplace=False)
###################################
t0 = time()
s_out.compute()
print(time() - t0)
# With deepcopy: 0.101, 0.101, 0.101
# Without deepcopy: 0.1009, 0.101, 0.1009
```

Here, the difference is less than 1%.

-------------------

## Summary

I think we need some way of avoiding changes in mutable object "jumping" between navigation pixels. I guess ideally the functions people use shouldn't be changing the input parameters, but I don't think we can assume that. Thus, I fear that people will run into very strange and subtle behavior when using `s.map`.

There might be some other way than using `deepcopy`, but I'm not sure how this can be done. The worse performance is indeed a negative, but I feel that it is small enough to be worth it.

I guess one way of mitigating this, could be to include a `run_safe` parameter, where the deepcopy is not performed. This would default to `run_safe=True`, where people who want to maximize the performance could use `run_safe=False`.

hyperspy/signal.py
magnunor: I rewrote the docstring text for the ragged parameter to reflect the changes.

The bigger potential issue is that this breaks the current behavior for `s.map` of non-lazy signals. In the current `RELEASE_next_minor` the `_map_iterate` automatically turns the signal into a ragged signal, if the returned signal dimensions has different shapes. However, it does this by effectively doubling the required memory in the process.

Lets assume an input navigation shape of the signal as `(128, 128)`. It does this by saving the results to a `np.empty((128, 128), dtype=object)`, then checking how the shapes looks like in each navigation position. If they all have the same size, the array is converted to a non-`object` array via:

```python
res_data = np.stack(res_data.ravel()).reshape(shape)
```

However, (at least on my computer) this doubles the memory requirement in this process. For big datasets, this can be an issue.

----------

This same functionality could easily be replicated in the new `s.map` in this pull request, but I think the "double memory need" is really bad. Essentially, to handle ragged results automatically we either have to find a way to "convert" an `object` NumPy array into a "standard" array (ideally without using a lot of memory, OR possibly check several navigation positions to see of the returned output is always the same.

This code tries to make a new "non-object" array from an `object` array, but doubles up the memory use.

```python
import numpy as np
data = np.empty((128, 128), dtype=object)
for iy, ix in np.ndindex(data.shape):
    data[iy, ix] = np.random.randint(0, 1024, size=(256, 256), dtype=np.uint16)
data_new = np.empty((128, 128, 256, 256), dtype=np.uint16)
for iy, ix in np.ndindex(data.shape):
    data_new[iy, ix] = data[iy, ix]
```

magnunor:
I cleaned up the commit history quite a bit, so it should be more readable now.

It should be ready for review, given that the tests pass.

--------------

Some outstanding issues

### Deepcopy input arguments

https://github.com/hyperspy/hyperspy/pull/2703#discussion_r762562064

### `lazy_result` parameter name

https://github.com/hyperspy/hyperspy/pull/2703#discussion_r762562064

### `ragged` signal behavior

https://github.com/hyperspy/hyperspy/pull/2703#discussion_r762562064

magnunor:
The failing azure tests seems to be unrelated to the changes in this pull request: `DeprecationWarning: distutils Version classes are deprecated. Use packaging.version instead.`

magnunor:
I just noticed that I forgot to update the changelog, and add information to the user guide. Doing this now.

magnunor:
I ran the extension tests and in addition the pyxem tests which has been failing previously, one of the tests in `lumispy` is also failing now:

```python
 self = <lumispy.tests.signals.test_luminescence_spectrum.TestLumiSpectrum testMethod=test_errors_raise>

    def test_errors_raise(self):
        s = LumiSpectrum(np.ones(50))
        for bkg, error in error_backgrounds:
            self.assertRaises(error, s.remove_background_from_file, bkg)
        # Test that a GUI is opened if s.remove_background_from_file is passed without a background
        # s.remove_background_from_file()
        # Test double background removal
        s.remove_background_from_file(backgrounds[0][0], inplace=True)
        self.assertRaises(
>           RecursionError, s.remove_background_from_file, backgrounds[0][0]
        )
E       AttributeError: 'LumiTransient' object has no attribute 'remove_background_from_file'
```

I see that the function itself uses `map`, but not really sure what causes this. https://github.com/LumiSpy/lumispy/blob/71949d2df196823d91672c40a41e64b7a024d473/lumispy/signals/luminescence_spectrum.py#L415

hyperspy/misc/utils.py
ericpre: Yes, there is no doubt that a deepcopy is required when a mutable object is changed by the function to be "mapped", and the question is where should this deepcopy takes place. In my opinion, this is a concern of the function the deepcopy should be done there! This is the reason I was saying that if the function changes one of its mutable argument, this is a bug of the function, not the `map` method.

I think that it simply needs documentation in the docstring, specify that the `function` shouldn't change mutable argument and instead if should make a copy.

hyperspy/signal.py
ericpre: I don't have strong view but I would lean on the side of prioritising consistency between arguments of the same function other consistency with other library. 
Another reason that I find `lazy_output` more appropriate is that we have the pattern of using the [`out` keyword](https://hyperspy.readthedocs.io/en/stable/user_guide/signal.html#reusing-a-signal-for-output) to specify where to put the output.

Both [`output` ](https://hyperspy.readthedocs.io/en/stable/search.html?q=output) and [`result`](https://hyperspy.readthedocs.io/en/stable/search.html?q=result) are used in the hyperspy user guide but their meaning is slightly different - output is more generic which in my opinion better fit the scope of `map`. With `result`, there is a sense of "finality", while "output" doesn't carry that meaning, it can an intermediate output (to be used by something else) or a final output.

hyperspy/misc/utils.py
ericpre: OK, I was questioning this because not all array objects will support `np.asarray`. Maybe something along the line of:
```suggestion
        try:
           output_dtype = output.dtype
        except AttributeError:
           output = np.asarray(output)
           output_dtype = output.dtype
```

hyperspy/signal.py
```diff
@@ -4672,16 +4674,29 @@ def map(
...
+            If ``True``, the data is replaced by the result. Otherwise
             a new Signal with the results is returned.
         ragged : None or bool, default None
             Indicates if the results for each navigation pixel are of identical
             shape (and/or numpy arrays to begin with). If ``None``,
-            the appropriate choice is made while processing. If True in case
-            of lazy signal, the signal will be compute at the end of the
-            mapping. Note: ``None`` is not allowed for Lazy signals!
+            the output signal will be the same as the input signal. Ergo, if
+            this signal is ragged, the output signal will also be ragged.
```
```suggestion
            the output signal will be ragged only if the original signal is ragged.
```

hyperspy/signal.py
ericpre: Yes, the docstring simply needed to be updated! I wasn't not a big fan of this `None` option.

hyperspy/misc/utils.py
magnunor: I agree that functions should not change their mutable arguments, and if they do that is a problem with that function. However, I worry that users will write their own custom made functions to be used with `map`, and since the "mutable object pitfall" is not well known among non-expert python programmers, I fear that users will get wrong results without even knowing it.

Also, without deepcopy, iterating signals which are passed to map can also be changed:

```python
import numpy as np
import hyperspy.api as hs
def test_function(image, iterating):
    iterating[0] = 2
    return image
s = hs.signals.Signal2D(np.zeros((1, 2, 20, 20)))
s_iter = hs.signals.Signal1D(np.zeros((1, 2, 2)))
s.map(test_function, iterating=s_iter)
s_iter
# array([[[2., 0.],
#         [2., 0.]]])
# All values here should be 0.
```

In addition, I'm not even sure if commonly used libraries such as scikit-image, scikit-learn, ... check for things like this? That the functions does not change its input arguments? I'm sure they try, but is this actually tested in the test suite?

------------

I think a compromise here could be an argument in `s.map`, `deepcopy_arguments`. Which by default is `True`, so that people who want to get the best performance can set it to `False`. I think most users would rather be sure that the results from `s.map` is correct, at a cost of a couple of percent of runtime.

hyperspy/signal.py
magnunor: ok, I changed the argument from `lazy_result` to `lazy_output`.

This might lead to changing the argument name in pyxem as well.

hyperspy/misc/utils.py
magnunor: Good point! Done!

magnunor:
@jlaehne, I'm not too familiar with LumiSpy, so wondering if you have any input on the failing test in https://github.com/hyperspy/hyperspy/pull/2703#issuecomment-998952474

hyperspy/misc/utils.py
magnunor: I added a `deepcopy_args` parameter to map, which toggles the deepcopying. It defaults to `True`.
I also added deepcopying for the input data as well, as I discovered that it is also affected by the same issue.

This also made it pretty easy to compare the performance for the deepcopy and non-deepcopy versions.

In short, the increase in runtime was about 0.86% to 1.9%, for the deepcopying. Ergo, doing the deepcopy makes the runtime about 1.5% longer.

-----------------
## Testing script

```python
from tqdm import trange
import time
import numpy as np
import hyperspy.api as hs

def mask_data(image, mask):
    image *= mask
    output = np.abs(np.fft.fftshift(np.fft.fft2(image)))
    return output[0, 0]

size = (10, 10, 500, 500)
s = hs.signals.Signal2D(np.random.random(size))
mask = np.zeros(size, dtype=bool)
mask[:, :, 200:, :] = True
s_mask = hs.signals.Signal2D(mask)

time_list_not_deepcopy = []
for i in trange(100):
    t0 = time.time()
    s_out = s.map(mask_data, mask=s_mask, inplace=False, deepcopy_args=False)
    time_list_not_deepcopy.append(time.time() - t0)
print("Signal size: {0}. Not deepcopy: {1}".format(s.axes_manager.shape, np.mean(time_list_not_deepcopy)))

time_list_deepcopy = []
for i in trange(100):
    t0 = time.time()
    s_out = s.map(mask_data, mask=s_mask, inplace=False, deepcopy_args=True)
    time_list_deepcopy.append(time.time() - t0)
print("Signal size: {0}. Deepcopy: {1}".format(s.axes_manager.shape, np.mean(time_list_deepcopy)))
```

----------------
## Results:

Signal size: (50, 50, 200, 200). Not deepcopy: 3.401603078842163
Signal size: (50, 50, 200, 200). Deepcopy: 3.4594222259521485
**1.7% slower**

Signal size: (2000, 2000). Not deepcopy: 0.5066886377334595
Signal size: (2000, 2000). Deepcopy: 0.5110295653343201
**0.85% slower**

Signal size: (4000, 4000). Not deepcopy: 2.219548602104187
Signal size: (4000, 4000). Deepcopy: 2.25101553440094
**1.4% slower**

Signal size: (10, 10, 500, 500). Not deepcopy: 0.9807310342788697
Signal size: (10, 10, 500, 500). Deepcopy: 0.9996297264099121
**1.9% slower**

ericpre: My opinion is that it simply needs to be explained to the docstring and it is not worth having an additional argument to the `map` method - the list of arguments is already long enough and I prefer to keep the API simple instead of adding argument to toggle workarounds.

When an user is changing arguments in the function itself, it sounds like an non-trivial usage of the `map` function and it would be fair to expect users to read the docstring correctly in this case or to be aware of this issue!

jlaehne:
> @jlaehne, I'm not too familiar with LumiSpy, so wondering if you have any input on the failing test in [#2703 (comment)](https://github.com/hyperspy/hyperspy/pull/2703#issuecomment-998952474)

I'm puzzled why he suddenly runs the test for a `LumiTransient` signal object. The LumiSpy signal hierarchy basically has two branches, `LumiSpectrum` and `LumiTransient` and the tested `remove_background_from_file` function is defined in the former branch, so it makes sense that `LumiTransient` objects do not have that function available. In `test_luminescence_spectrum.py` only `LumiSpectrum` objects are defined, so I do not quite understand why he suddenly runs the test for a different signal type as well.

magnunor:
@ericpre, sorry about harping on about this, but I'm still quite skeptical about not making copies of the input parameters. Mostly due to me encountering this exact issue, with the input parameters being changed, leading to very confusing results.

So writing a summary of the issue, so that other people can look through this, since it is a bit hard to follow it through the various comments "embedded" with the code.

------------------------------------------

# Background

A number of python objects are mutable, which means the original object can be altered by changing the content of that object. For example NumPy arrays:

```python
data0 = np.ones((2, 2))
data1 = data0
data1 *= 0
print(data0)
[[0., 0.],
 [0., 0.]]
```

One way of avoiding this, is by making a deepcopy:

```python
data0 = np.ones((2, 2))
import copy
data1 = copy.deepcopy(data0)
data1 *= 0
print(data0)
[[1., 1.],
 [1., 1.]]
```

# This pull request

In the context of this pull request, this extends to any function being used via `map`. Thus, if for example, we want to apply a mask on our signal, we can inadvertently change our original data. Even if we use `inplace=False`.

```python
def test_function(image, mask):
    image *= mask
    return image
s = hs.signals.Signal2D(np.ones((2, 2, 4, 4)))
mask = np.ones((4, 4), dtype=bool)
mask[:2, :2] = False
s_out = s.map(test_function, mask=mask, inplace=False)
print(s.data[0, 0]) # Print the original signal, which should be only 1.
[[0. 0. 1. 1.]
 [0. 0. 1. 1.]
 [1. 1. 1. 1.]
 [1. 1. 1. 1.]]
```
Here, the original NumPy array has been altered via the `s.map` function: it should only be `1.`s in the data, but the "top-left" corner of the original data has been altered.

## `lazy_output`

This gets even worse for lazy outputs, since this doesn't happen before `compute` is run.
```python
def test_function(image, mask):
    image *= mask
    return image
s = hs.signals.Signal2D(np.ones((2, 2, 4, 4)))
mask = np.ones((4, 4), dtype=bool)
mask[:2, :2] = False
s_out = s.map(test_function, mask=mask, inplace=False, lazy_output=True)
print(s.data[0, 0])
[[1. 1. 1. 1.]
 [1. 1. 1. 1.]
 [1. 1. 1. 1.]
 [1. 1. 1. 1.]]
s_out.compute()
print(s.data[0, 0])
[[0. 0. 1. 1.]
 [0. 0. 1. 1.]
 [1. 1. 1. 1.]
 [1. 1. 1. 1.]]
```

## Changing input arguments

This also extends to the input arguments itself.

```python
def test_function(image, mask):
    x, y = np.random.randint(2, 4, size=2)
    mask[x, y] = not mask[x, y]
    image *= mask
    return image
s = hs.signals.Signal2D(np.ones((2, 2, 4, 4)))
mask = np.ones((4, 4), dtype=bool)
mask[:2, :2] = False
s_out = s.map(test_function, mask=mask, inplace=False, deepcopy_args=False)
print(s.data)
[[[[0. 0. 1. 1.]
   [0. 0. 1. 1.]
   [1. 1. 1. 0.]
   [1. 1. 1. 1.]]

  [[0. 0. 1. 1.]
   [0. 0. 1. 1.]
   [1. 1. 1. 0.]
   [1. 1. 1. 0.]]]

 [[[0. 0. 1. 1.]
   [0. 0. 1. 1.]
   [1. 1. 1. 0.]
   [1. 1. 0. 0.]]

  [[0. 0. 1. 1.]
   [0. 0. 1. 1.]
   [1. 1. 0. 0.]
   [1. 1. 0. 0.]]]]
```

Here, the changes in `mask` is carried over from each navigation index to navigation index. This is unexpected behavior.

# The problem

I think it is fairly common to use the `*=` or `data[x, y] = ...` when working with NumPy arrays. Thus, I am a bit worried that the aforementioned issues highlighted in the examples: I fear that people will inadvertently be changing their data and arguments, without even knowing it. Potentially leading to very wrong results.

# The solution

There are two(?) ways of avoiding this:

1. Make sure every function you use does not change any of its inputs
2. Make a copy of any input sent to the function


## Solution 1

While most functions should not change their inputs, my worry is that already mentioned `*=` and `data[x, y] = ...` operations are fairly common, and I don't think this is typically tested for? In addition, this should also be tested for various edge cases with input parameters.

For example, if a user is utilizing a function from some other library which (maybe inadvertently) does change the input via `*=`, the user would have to make a "wrapper" function which does the copying.


## Solution 2

Make a `deepcopy` of both the input data, iterating kwargs and arguments. This avoids all the aforementioned issues, since all the operations will be done on a copy of the input data and input arguments. Note that this will only be for one position per chunk at a time. Thus, it will not require a great deal of memory, as the copy will disappear after the input function is finished running. Note that the deepcopying is not applied to the HyperSpy signals themselves, but to the NumPy arrays, and to any arguments.

The negative with this solution, is that it adds a little bit to the processing time. My testing (on a very weak laptop), gives about 1.5% increased processing time for some different cases (see https://github.com/hyperspy/hyperspy/pull/2703#discussion_r774167512).

This can be mitigated by having a parameter for `s.map`, which disables this deepcopy behaviour: currently in the pull request, this is done via a `deepcopy_args` parameter.

---------------

In my opinion, the small hit (1.5%) to performance in solution 2 is worth it, to be really sure you don't get any of the potentially strange problems mentioned earlier. Personally, I would be nervous about actually using `s.map` without copying the input parameters.

ericpre:
Shall we leave the deepcopy discussion for a separate PR?
This is an public API (and behaviour) changes, which would be better to discuss separately. There is already enough changes and discussions on other aspects of this PR.

magnunor:
> Shall we leave the deepcopy discussion for a separate PR?
This is an public API (and behaviour) changes, which would be better to discuss separately. There is already enough changes and discussions on other aspects of this PR.

Agreed! I'll remove the `deepcopy` behavior from this pull request, improve the docstring as you suggested, and make a new PR with my proposed deepcopy changes.

magnunor:
@ericpre, I removed the deepcopy functionality, and cleaned up the commits.

ericpre (approved):
Great, thanks!
The test failures are unrelated to hyperspy and fixes in #2866, so it is all good.

Issue #2698, opened 2021-04-09
crop_signal1d not updating plot


magnunor:
```python
import numpy as np
import hyperspy.api as hs
s = hs.signals.Signal1D(np.random.randint(0, 99, 1000))
s.crop_signal1D() # Select a region. Press Apply
```

The plot does not update, as seen in the image here:

![linux_crop_figure_not_updating](https://user-images.githubusercontent.com/1690979/114225537-b9f38d80-9972-11eb-85f0-2b1d018d26ab.jpg)

Furthermore, pressing "OK" in the dialog box gives this error:

```python
File "hyperspy/hyperspy/_signals/signal1d.py", line 1258, in crop_signal1D
  self.crop(axis=self.axes_manager.signal_axes[0].index_in_axes_manager,
File "hyperspy/hyperspy/signal.py", line 2865, in crop
  i1, i2 = axis._get_index(start), axis._get_index(end)
File "hyperspy/hyperspy/axes.py", line 340, in _get_index
  return self.value2index(value)
File "hyperspy/hyperspy/axes.py", line 544, in value2index
  raise ValueError("The value is out of the axis limits")
ValueError: The value is out of the axis limits
```

Replotting the signal fixes the plotting limits.

![linux_replot_fixes_it](https://user-images.githubusercontent.com/1690979/114225560-c2e45f00-9972-11eb-8beb-6500060dc9ae.jpg)

I see this in both the most recent version of RELEASE_next_patch, the conda-forge version and the newest HyperSpy bundle.

ericpre:
It looks to me that the error comes from the fact that the dialog box doesn't close and therefore the value selected are in the axis anymore.

There were discussing about deprecating method such crop (https://github.com/hyperspy/hyperspy/pull/2256) and the motivation were that it does the same thing as `isig` combined with a ROI but isig is more powerful and more reproducible. The initialisation of the ROI are still a bit annoying but it is now easier thanks to https://github.com/hyperspy/hyperspy/pull/2341.

PR #2704, opened 2021-04-12
Set requirement matplotlib <3.4

ericpre:
When checking #2698, I notice that #2684 didn't fully add support for matplotlib 3.4 and in the following example, the widget doesn't appear when it should:

```python
import numpy as np
import hyperspy.api as hs
from hyperspy.drawing._widgets.range import ModifiableSpanSelector

s = hs.signals.Signal1D(np.arange(100))
s.plot()
ax = s._plot.signal_plot.ax

span = ModifiableSpanSelector(ax, useblit=True)
```

This is not straightforward to fix and the simplest is to pin matplotlib for the moment...

### Progress of the PR
- [x] Set requirement matplotlib <3.4
- [x] ready for review.

### DIFF
--- a/hyperspy/drawing/_widgets/range.py
+++ b/hyperspy/drawing/_widgets/range.py
@@ -402,7 +402,7 @@ class ModifiableSpanSelector(SpanSelector):
     def update(self, *args):
         # Override the SpanSelector `update` method to blit properly all
         # artirts before we go to "modify mode" in `set_initial`.
-        self.set_visible(True)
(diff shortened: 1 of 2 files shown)


ericpre:
Using this approach will be annoying for an existing qt backend issue on latest mac os release, see https://github.com/matplotlib/matplotlib/pull/19334. The workaround is to set an environment variable.

ericpre:
Closed in favour of #2705.

PR #2709, opened 2021-04-14
Fix extension registration that was not working in some complex scenarions

francisco-dlp:
### Description of the change

In order to find extension hyperspy uses ``pkg_resources.iter_entry_points``. However, that can fail in certain cases. This PR replace it with ``importlib_metadata`` that is know to work more reliably.

``importlib_metadata`` is now part of the standard library. However, only Python 3.9 ships with the minimum version of the library that we need, so I have added it as a requirement.

This should fix extension discovery in some online services such as Kaggle notebooks.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -69,6 +69,8 @@ install_req = ['scipy>=1.1',
                'prettytable',
                'tifffile>=2018.10.18',
                'numba',
+                # included in stdlib since v3.8, but this required version requires Python 3.9
+               'importlib_metadata>=1.6.0',
                ]
 
 extras_require = {
--- a/hyperspy/extensions.py
+++ b/hyperspy/extensions.py
@@ -19,10 +19,10 @@
 import logging
 import copy
 import pkgutil
-import pkg_resources
 import yaml
 
 from pathlib import Path
+import importlib_metadata as metadata
 
 
 _logger = logging.getLogger(__name__)
@@ -37,8 +37,8 @@ EXTENSIONS["GUI"]["widgets"] = {}
 ALL_EXTENSIONS = copy.deepcopy(EXTENSIONS)
 
 _external_extensions = [
-    entry_point.module_name
-    for entry_point in pkg_resources.iter_entry_points('hyperspy.extensions')]
+    entry_point.module
+    for entry_point in metadata.entry_points()['hyperspy.extensions']]
 
 for _external_extension_mod in _external_extensions:
     _logger.info("Enabling extension %s" % _external_extension_mod)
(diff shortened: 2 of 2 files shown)


setup.py
```diff
@@ -69,6 +69,8 @@
                'prettytable',
                'tifffile>=2018.10.18',
                'numba',
+                # included in stdlib since v3.8, but this required version requires Python 3.9
+               'importlib_metadata>=1.6.0',
```
ericpre, line 73:
```suggestion
               'importlib_metadata>=1.6.0;python_version<"3.9.X"',
```
In case you know which minimum version of python include `importlib_metadata`, it may be good to specify it, to make it easier in the future to know when to remove it?

ericpre (approved):
I can't reproduce the error on Kaggle, import extension and hyperspy gui works fine for me.
In any case, `pkg_resources` is known to be slow down import time, so this is good to replace it!

I have left one comment.

setup.py
francisco-dlp: At the moment their API is a bit unstable, so I think that it is good to use the package version regardless of the Python version. In this way we can better manage the required version. In the future I imagine that the package version will be deprecated so  the deprecation warning should be enough to remind us this. Does it make sense?

francisco-dlp:
It's curious that you cannot reproduce the issue in Kaggle. I remember that at some point  the GUI was working as it was the notebook backend, but for me it is not longer the case--maybe they serve different versions of the notebook?

setup.py
ericpre: yes, fine, there is already a comment.

ericpre:
> It's curious that you cannot reproduce the issue in Kaggle. I remember that at some point the GUI was working as it was the notebook backend, but for me it is not longer the case--maybe they serve different versions of the notebook?

Could it be that the kernel(?) needs a restart after installing the libraries?

ericpre:
I confirm that with this PR, no restart of the kernel is necessary and otherwise a restart is necessary.

Interestingly, pip doesn't seem to honour the dependencies requirement and `importlib_metadata` stays at 1.4... when installing hyperspy from this branch, which then breaks importing hyperspy! After updating `importlib_metadata` to the latest version it works fine.

francisco-dlp:
Yes, that's it: Kaggle seems to use some trickery to import modules without restarting the kernel, but HyperSpy without this PR failts to register extensions without a kernel restart.

Regarding the version issue, that makes sense: if the required version is older than the one shipped with Python, it doesn't install it and probably uses Python's version instead. So, there was a bug there, the minimum required version turns out to be 3.6. Fixed in #2711.

ericpre:
> Interestingly, pip doesn't seem to honour the dependencies requirement and `importlib_metadata` stays at 1.4...

Actually, this is incorrect, the current version on kaggle is 3.4 instead of 1.4... and this explains why pip doesn't need to update it. It seems that the issue has been well characterised now!

PR #2711, opened 2021-04-15
Avoid importing extensions on registration

francisco-dlp:
HyperSpy currently uses ``pkgutil.get_loader`` to find the path of the extensions in order to register them. However that method imports the package just to retrieve the path. This PR uses ``import lib_metadata`` to retrieve the path and read the YAML file without importing the extension.

### DIFF
--- a/hyperspy/extensions.py
+++ b/hyperspy/extensions.py
@@ -18,7 +18,6 @@
 
 import logging
 import copy
-import pkgutil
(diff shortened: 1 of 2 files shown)


ericpre (approved):
Yes, this works well on kaggle: `importlib_metadata` is updated when installing from this PR.

PR #2713, opened 2021-04-20
Add drone CI to test arm64

ericpre:
I will suggest to run a cron job (daily or weekly) and not run the test suite on pull request or on branch.

### Progress of the PR
- [x] Run test suite on arm64 using drone CI
- [x] Fix failing test,
- [x] ready for review.

### DIFF
--- a/.drone.yml
+++ b/.drone.yml
@@ -1,0 +1,18 @@
+kind: pipeline
+name: test on arm64
+
+platform:
+  arch: arm64
+
+steps:
+- name: test
+  image: quay.io/condaforge/mambaforge:latest
+  environment:
+    MATPLOTLIB: agg
+  commands:
+    - lscpu
+    - mamba install hyperspy-base --only-deps
+    - mamba install scikit-learn pytest pytest-xdist
+    - pip install -e .
+    - pytest -n 4
+
--- a/README.rst
+++ b/README.rst
@@ -1,6 +1,6 @@
 .. -*- mode: rst -*-
 
-|Azure|_ |Github|_ |rtd|_ |Codecov|_
+|Azure|_ |Github|_ |Drone|_ |rtd|_ |Codecov|_
 
 |python_version|_ |pypi_version|_ |anaconda_cloud|_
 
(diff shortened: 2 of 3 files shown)


ericpre:
Example of successful build: https://cloud.drone.io/ericpre/hyperspy/15

francisco-dlp:
Why are you suggesting not to run them at PRs? Is it because they take twice as long as the others?

ericpre:
Yes, it is to avoid building queue and not to add burden to development until we know that this is working fine. Maybe, once we have more experience with drone CI, we can consider using it in the same way as with other CI.
Another point to justify that we don't need comprehensive testing on this architecture is that `arm64` is a linux system and that hyperspy doesn't have much code which depends on the type of processor. If there are failures, it will be most likely related to dependencies and a cron job should be enough to have an idea of the status of the libraries stack.

francisco-dlp:
Sounds good to me.

ericpre:
I have "activated" the `hyperspy/hyperspy` repository on drone and this is why, now drone appears in the "checks" of this PR.
The badge have been added to the readme in the last commit, so that we can spot it when it fails. Shall we set it to run daily for now at least?

.drone.yml
```diff
@@ -0,0 +1,18 @@
+kind: pipeline
+name: test on arm64
+
+platform:
+  arch: arm64
+
+steps:
+- name: test
+  image: quay.io/condaforge/mambaforge:latest
```
ericpre, line 9: By the way, it is very nice to use this docker image with mambaforge included! :)

francisco-dlp: Running daily sounds good.

Yes, mambaforge is a breezy of fresh air. This weekend I switched from a broken miniforge installation to mambaforge and I am not looking back, By the way, what broke my miniconda installation was uninstalling mamba_gator. I uninstalled it because it was using a lot of resources every time that I was opening jupyterlab.

ericpre: > By the way, what broke my miniconda installation was uninstalling mamba_gator. I uninstalled it because it was using a lot of resources every time that I was opening jupyterlab.

Indeed, this is now fixed in the release version - see https://github.com/mamba-org/gator/pull/133.

There is still a bug with uninstalling and reinstalling `gator`... which I think I was able to reproduce.

ericpre:
I have changed the settings on Drone CI:
- disable build on pull request
- enable daily cron job

ericpre:
@francisco-dlp, is it good to merge? Thanks!

jlaehne:
Why is `.drone.yml` a hidden file and the other pipelines not?
Do the pipelines have to be in the main folder or could they also be under workflows?
And why does `readthedocs.yaml` have a different ending then the other `yml` files?

ericpre:
> Why is `.drone.yml` a hidden file and the other pipelines not?
> Do the pipelines have to be in the main folder or could they also be under workflows?

It depends what the consumer of the files is expecting and we follow the standard specification for each service provider using these files.

> And why does `readthedocs.yaml` have a different ending then the other `yml` files?

As with `jpeg` versus `jpg`, both extensions work equally, as far as I am aware.

.drone.yml
ericpre: > There is still a bug with uninstalling and reinstalling `gator`... which I think I was able to reproduce.

Now is now [fixed](https://github.com/mamba-org/gator/pull/140) in gator 5.1.1.

PR #2725, opened 2021-04-29
Enhancement of NeXus file IO

ptim0626:
### Description of the change
The PR contains some improvement and bug fixes about the IO of NeXus file. 
- set correct chunk size for lazy dataset (previously it is chunked incorrectly)
- load NeXus dataset and HDF5 dataset consistently
- add option to skip loading array original metadata (this is to avoid duplicating the loading of some large arrays)
```
hs.load('nexus_file.nxs', skip_array_metadata=True)
```
- add option to load a specific dataset by specifying absolute path
```
hs.load('nexus_file.nxs', dataset_path='/entry/instrument/data')
```
- add option to skip certain original metadata when writing new file by specifying keys
```
s.save('output.nxs', skip_metadata_key=['sample', 'detector'])
```
Besides the new options, it also adds tests to cover previously uncovered parts. Any comment is welcomed.

### Progress of the PR
- [x] Changes implemented,
- [x] update docstring,
- [x] update user guide,
- [x] add tests,
- [x] ready for review.


doc/user_guide/io.rst
```diff
@@ -1361,8 +1361,10 @@ some additional loading arguments are provided.
 Extra loading arguments
 +++++++++++++++++++++++
 
-- ``dataset_keys``: ``None``, ``str`` or ``list`` of strings - Default is ``None`` . Absolute path(s) or string(s) to search for in the path to find one or more datasets.
+- ``dataset_keys``: ``None``, ``str`` or ``list`` of strings - Default is ``None`` . String(s) to search for in the path to find one or more datasets.
+- ``dataset_paths``: ``None``, ``str`` or ``list`` of strings - Default is ``None`` . Absolute path(s) to search for in the path to find one or more datasets.
```
ericpre: There is already a `dataset_path` used for other hdf5 file reader. Would it be worth using the same without `s`? The drawback is that this is not consistent with `dataset_keys` or `metadata_keys`.
Maybe, it would be good to define if we stick to plural or singular in this case... and deprecate/rename arguments accordingly. I don't know if one is better than the other!

doc/user_guide/io.rst
```diff
@@ -1391,16 +1393,16 @@ hdf datasets will be returned as signals:
...
-We can load a specific dataset using the ``dataset_keys`` keyword argument.
+We can load a specific dataset using the ``dataset_paths`` keyword argument.
 Setting it to the absolute path of the desired dataset will cause
 the single dataset to be loaded:
 
 .. code-block:: python
 
     >>> # Loading a specific dataset
-    >>> hs.load("sample.nxs", dataset_keys='/entry/experiment/EDS/data')
+    >>> hs.load("sample.nxs", dataset_paths="/entry/experiment/EDS/data")
```
ericpre: Maybe add an example to illustrate the difference between `dataset_keys` and `dataset_paths`?

ericpre (commented):
This looks good to me - see comments.

doc/user_guide/io.rst
ptim0626: Personally I would prefer to go with the plural as it implies that multiple paths can be searched for, but happy to go with singular for consistency with other readers if it is more suitable.

ericpre: Actually, after looking at the use of singular versus plural in argument name, it seems that `numpy` (and `hyperspy`) uses the `axis` argument (see for example `sum`), which can take one or several items, so I would suggest to use the singular.

ptim0626: I deprecated the plural arguments and now use singular for nexus file io.

ericpre (approved):

PR #2728, opened 2021-05-02
Fixes for binned non_uniform_axes

jlaehne:
Work on key issues from #2398

### Description of the change
- Fix plotting for binned spectrum images with non uniform axis obtained by `convert_to_non_uniform_axis` (`index_in_array` was lost)
- Fix curve fitting for binned spectrum images with non-uniform axis
- Handle occurences where 'scale' is used for binned signals also in the non-uniform case (e.g. in signals used for background removal)
- Stacking along non-uniform axes
- Raise warnings where still necessary, e.g. for EELS and EDX; add to docstrings
- Test for warnings
- Increase coverage of `axes.py` and other files affected by the non-uniform axis branch

### Progress of the PR
- [X] fix plotting,
- [X] fix curve fitting,
- [X] handle 'scale' for non-uniform, binned signals,
- [X] stacking for non-uniform axes,
- [X] warnings for EELS and EDS models,
- [X] check for additional functions that need to raise warnings,
- [X] increase coverage,
- [X] add tests,
- [X] ready for review.


jlaehne:
The failing test in `test_mrcz.py` is unrelated (fails also in other branches).

For the remaining 3 Codecov warnings I am not sure which test to amend to catch these lines.

hyperspy/_components/doniach.py
```diff
@@ -150,19 +150,25 @@ def estimate_parameters(self, signal, x1, x2, only_current=False):
...
             self.A.value = height * 1.3
-            if is_binned(signal) is True:
+            if is_binned(signal):
             # in v2 replace by
             #if axis.is_binned:
-                self.A.value /= axis.scale
+                if axis.is_uniform:
+                    self.A.value /= axis.scale
+                else:
+                    self.A.value /= np.gradient(axis.axis)[axis.value2index(centre)]
```
ericpre: This could simplify to:
```suggestion
                # use gradient to support non uniform axis
                self.A.value /= np.gradient(axis.axis)[axis.value2index(centre)]
```

hyperspy/_signals/signal1d.py
```diff
@@ -1098,9 +1098,15 @@ def _remove_background_cli(
...
                 axis = self.axes_manager.signal_axes[0]
-                scale_factor = axis.scale if is_binned(self) else 1
+                if is_binned(self):
                 # in v2 replace by
-                # scale_factor = axis.scale if axis.is_binned else 1
+                # if axis.is_binned:
+                    if axis.is_uniform:
+                        scale_factor = axis.scale
+                    else:
+                        scale_factor = np.gradient(axis.axis)
```
ericpre, line 1113: Similar as above, can be simplify to:
```suggestion
                    scale_factor = np.gradient(axis.axis)
```

hyperspy/tests/signals/test_remove_background.py
```diff
@@ -46,11 +46,14 @@ def setup_method(self, method):
         self.signal.axes_manager[0].scale = 0.01
 
     @pytest.mark.parametrize('binning', (True, False))
+    @pytest.mark.parametrize('uniform', (True, False))
```
ericpre: This is adding redundant test and we only need to cover a non-uniform when binning is True. We can either split test the non-uniform in a separate different test or find the suitable syntax for `parametrize` to reduce the number of combination. I think this is one, but I don't know remember how it is...

hyperspy/axes.py
```diff
@@ -971,12 +965,14 @@ def get_axis_dictionary(self):
...
     def convert_to_non_uniform_axis(self):
         d = super().get_axis_dictionary()
+        axes_manager = self.axes_manager
         d["_type"] = 'DataAxis'
         self.__class__ = DataAxis
         self.__init__(**d, axis=self.axis)
         del self._expression
         del self._function
         self.remove_trait('x')
+        self.axes_manager = axes_manager
```
ericpre, line 977: Am I understanding correctly, that this changes is fixing a bug? After calling `convert_to_non_uniform_axis`, the `axes_manager` attribute of the axis is a different object/has a different id? If so, is it worth checking adding a test checking the id before and after?

hyperspy/signal.py
```diff
@@ -4300,9 +4300,9 @@ def integrate1D(self, axis, out=None):
         (64,64)
 
         """
-        if is_binned(self, axis=axis) is False:
+        if (not is_binned(self, axis=axis)) or (not self.axes_manager[axis].is_uniform):
```
ericpre: An alternative would be to add support of non uniform axes in `integrate_simpson`?

hyperspy/tests/model/test_model.py
```diff
@@ -620,6 +620,25 @@ def test_print_current_values_component_list(self):
...
 
 
+class TestModelUniformBinned:
+    def setup_method(self, method):
+        self.m = hs.signals.Signal1D(np.arange(10)).create_model()
+        self.o = hs.model.components1D.Offset()
+        self.m.append(self.o)
+
+    @pytest.mark.parametrize("uniform", [True, False])
+    @pytest.mark.parametrize("binned", [True, False])
```
ericpre, line 630: I would remove `binned=False` because it is redundant.

hyperspy/_components/exponential.py
```diff
@@ -149,10 +146,13 @@ def estimate_parameters(self, signal, x1, x2, only_current=False):
...
 
-            if is_binned(signal) is True:
+            if is_binned(signal):
             # in v2 replace by
             #if axis.is_binned:
-                A /= axis.scale
+                if axis.is_uniform:
+                    A /= axis.scale
+                else:
+                    A /= np.gradient(axis.axis)[i_mid]
```
thomasaarholt: Just wondering about the use of `i_mid` here. Is that essentially a "best guess" for the bin width? Have we considered if there are other approaches here?

hyperspy/_signals/signal1d.py
```diff
@@ -1073,9 +1073,9 @@ def _remove_background_cli(
             zero_fill=False, show_progressbar=None, model=None,
             return_model=False):
         """ See :py:meth:`~hyperspy._signal1d.signal1D.remove_background`. """
-        if fast and not self.axes_manager.signal_axes[0].is_uniform:
-            raise NotImplementedError(
-                "This operation with `fast=True` is not implemented for non-uniform axes.")
+        #if fast and not self.axes_manager.signal_axes[0].is_uniform:
```
thomasaarholt: Just remove these now-commented lines?

hyperspy/misc/utils.py
```diff
@@ -1007,8 +1010,25 @@ def stack(signal_list, axis=None, new_axis_name="stack_element", lazy=None,
         newlist = broadcast_signals(*signal_list, ignore_axis=axis_input)
 
         if axis is not None:
-            step_sizes = [s.axes_manager[axis].size for s in newlist]
-            axis = newlist[0].axes_manager[axis]
+            step_sizes = [s.axes_manager[axis_input].size for s in newlist]
+            axis = newlist[0].axes_manager[axis_input]
+            if not axis.axes_manager[axis_input].is_uniform:
+                # stack axes if non-uniform
+                if hasattr(axis.axes_manager[axis_input],'_expression'):
```
thomasaarholt: Is it more explicit/clear to check if the type of the axis is FunctionalDataAxis?

thomasaarholt:
Aside from a few small comments, LGTM! Excellent choice of axis name and units in the tests - "lovely plumage" :)

hyperspy/_components/exponential.py
jlaehne: Yes, that was my idea of a "best guess" in the parameter estimator as we have no idea how the axis evolves. For `Gaussian` and the like I took the `centre` position, but `Exponential` and `Polynomial` do not have that, so I resorted to this guess. If anyone has a better idea, I'm happy to take that up, but for the estimator it should not get too complicated either I would say. An alternative could be `np.mean(np.gradient(...))` instead.

hyperspy/misc/utils.py
jlaehne: I could check the value of the `__class__` attribute to be explicit, but `_expression` is unique as well. But you are right that it might be clearer to others that way.

jlaehne:
> Aside from a few small comments, LGTM! Excellent choice of axis name and units in the tests - "lovely plumage" :)

Yes, we definitely need more Monty (in HyperSpy's) Python :)

hyperspy/misc/utils.py
```diff
@@ -1007,8 +1010,25 @@ def stack(signal_list, axis=None, new_axis_name="stack_element", lazy=None,
...
+                    axis.axes_manager[axis_input].convert_to_non_uniform_axis()
+                for i, _s in enumerate(signal_list):
+                    if i == 0:
+                        pass
+                    elif (axis.axis[0] < axis.axis[-1] and axis.axis[-1] < \
+                            _s.axes_manager[axis_input].axis[0]) or \
+                            (axis.axis[-1] < axis.axis[0] and \
+                            _s.axes_manager[axis_input].axis[-1] < axis.axis[0]):
+                        axis.axis = np.concatenate((axis.axis, _s.axes_manager[axis_input].axis))
+                        print(axis.__class__)
```
thomasaarholt: There's a print statement here - should that be removed?

hyperspy/misc/utils.py
thomasaarholt: I was thinking `if type(axis.axes_manager[axis_input]) is FunctionalDataAxis`.

I'm also a little confused on how the `axis` variable changes in this function - initially its an int / string, but then `newlist[0].axes_manager[axis_input]` gets returned into it - is `axis` now the DataAxis, and `axis.axes_manager[axis_input] == axis`?

If so, my previous if statement gets reduced to `if type(axis) is FunctionalDataAxis`.

thomasaarholt: Arguably, `if axis.is_uniform` could just be `if type(axis) is UniformDataAxis`, and we could do away with the `axis.is_uniform` and `axis._is_uniform` bits.

hyperspy/misc/utils.py
```diff
@@ -1007,8 +1010,25 @@ def stack(signal_list, axis=None, new_axis_name="stack_element", lazy=None,
         newlist = broadcast_signals(*signal_list, ignore_axis=axis_input)
 
         if axis is not None:
-            step_sizes = [s.axes_manager[axis].size for s in newlist]
-            axis = newlist[0].axes_manager[axis]
+            step_sizes = [s.axes_manager[axis_input].size for s in newlist]
+            axis = newlist[0].axes_manager[axis_input]
```
thomasaarholt: `newlist` is an old variable that should have been renamed. Would you mind renaming it to `broadcasted_sigs`?

hyperspy/_components/exponential.py
thomasaarholt: I think that's fine. But I suggest you add a few short comments explaining that this is a best guess, perhaps with your suggested alternative. I think you should do this (copy paste) on each component where you use such a guess. I'm just imagining someone trying to decipher this in the future where perhaps you are unavailable :)

ericpre (commented):
A first iteration of comments, I may have some more later.

hyperspy/axes.py
jlaehne: I added this, because otherwise axes were loosing the `is_binned` attribute - I guess I should add a test that checks that `is_binned`, `navigate` and the like remain the same.

hyperspy/_components/doniach.py
jlaehne: I can do so for all the occurences, just thought it might be computationally more easy to stay with the scale for the uniform axes.

hyperspy/axes.py
jlaehne: done

hyperspy/signal.py
jlaehne: Actually, `integrate_simpson` should work for non-uniform axes as it uses the axis vector, I'll change that here. `sum` does not make sense on a non-uniform axis! The only open question is non_uniform axes that are binned. Can we stick to `integrate_simpson` for these, probably not? Instead in `sum` we should probably define a non-uniform variation that multiplies the gradient of the axis with the intensity and then sums up the result. What do you think?

hyperspy/tests/model/test_model.py
jlaehne: `binned=False and uniform=False` is not checked elsewhere as far as I see

hyperspy/_components/doniach.py
ericpre: Fair point, it is in the order of 10s of us, which is still very small and `estimate_parameters` is not expected to be called many many times. Maybe an alternative to avoid repetition is to define a scaling factor before the `if only_current is True:`:
```python
factor = 1
if is_binned(signal):
   if axis.is_uniform:
      factor = axis.scale
  else:
      factor = np.gradient(axis.axis)[axis.value2index(centre)]

if only_current is True:
   ...
   self.A.value /= factor
```

Similarly as you did for the test.

hyperspy/tests/model/test_model.py
ericpre: Sorry, I should have said not necessary, because there is no code, which corresponds to the case `binned=False` and `uniform=False`! In any case, my comment can be dismissed because this test is quick - unlike `test_background_remove` where the number of added cases to test starts to be significant and the time to run this test is not negligible.

hyperspy/_signals/eds_tem.py
```diff
@@ -267,6 +267,10 @@ def get_calibration_from(self, ref, nb_pix=1):
         # Setup the axes_manager
         ax_m = self.axes_manager.signal_axes[0]
         ax_ref = ref.axes_manager.signal_axes[0]
+        for _axis in [ax_m, ax_ref]:
+            if not _axis.is_uniform:
+                raise NotImplementedError(
+                    "The function is not implemented for non-uniform axes.")
```
ericpre, line 278: Is it worth adding a `Raises` section in the docstring?

hyperspy/axes.py
ericpre: I agree that this is need to be set but I can't reproduce the issue you describe here:
```python
import hyperspy.api as hs
import numpy as np

s = hs.signals.Signal1D(np.arange(100*5).reshape(5, 100))

axis = s.axes_manager[1]
axis.is_binned = True
assert axis.axes_manager is s.axes_manager

axis.convert_to_non_uniform_axis()

assert axis.is_binned
assert not axis.navigate
assert axis.axes_manager is None
```

hyperspy/misc/utils.py
```diff
@@ -921,18 +920,21 @@ def stack(signal_list, axis=None, new_axis_name="stack_element", lazy=None,
...
-        have the same dimensions. Otherwise the
-        signals are stacked over the axis given by its integer index or
-        its name. The data must have the same shape, except in the dimension
-        corresponding to `axis`.
+        have the same dimensions. Otherwise the signals are stacked over the
+        axis given by its integer index or its name. The data must have the
+        same shape, except in the dimension corresponding to `axis`. If the
+        stacking axis of the first signal is uniform, it is extended up to the
+        new length; if it is non-uniform, the axes vectors of all signals are
+        concatenated along this direction; if it is a `FunctionalDataAxis`,
```
ericpre, line 936: In case of `FunctionDataAxis` could be treated similarly as the `UniformAxis` regarding the extension of the axis?

hyperspy/axes.py
ericpre: OK, I think that I understand what you meant, this relate to the bullet point in the description of this PR:
- Fix plotting for binned spectrum images with non uniform axis obtained by `convert_to_non_uniform_axis` (`index_in_array` was lost)

And the reason for this is because `axes_manager` is set `None` in `BaseDataAxis.__init__`.

If this is correct, the corresponding test needs to be updated.

hyperspy/misc/utils.py
```diff
@@ -1004,13 +1007,29 @@ def stack(signal_list, axis=None, new_axis_name="stack_element", lazy=None,
...
         # Matching axis calibration is checked here
-        newlist = broadcast_signals(*signal_list, ignore_axis=axis_input)
+        broadcasted_sigs = broadcast_signals(*signal_list, ignore_axis=axis_input)
 
         if axis is not None:
-            step_sizes = [s.axes_manager[axis].size for s in newlist]
-            axis = newlist[0].axes_manager[axis]
+            step_sizes = [s.axes_manager[axis_input].size for s in broadcasted_sigs]
+            axis = broadcasted_sigs[0].axes_manager[axis_input]
+            if not axis.axes_manager[axis_input].is_uniform:
```
ericpre:
```suggestion
            if not axis.is_uniform:
```

hyperspy/misc/utils.py
```diff
@@ -1004,13 +1007,29 @@ def stack(signal_list, axis=None, new_axis_name="stack_element", lazy=None,
...
+                # stack axes if non-uniform (convert to DataAxis if FunctionalDataAxis)
+                if type(axis) is FunctionalDataAxis:
+                    axis.axes_manager[axis_input].convert_to_non_uniform_axis()
+                for i, _s in enumerate(signal_list):
+                    if i == 0:
+                        pass
```
ericpre:
```suggestion
                for _s in signal_list[1:]:
```

hyperspy/misc/utils.py
```diff
@@ -1004,13 +1007,29 @@ def stack(signal_list, axis=None, new_axis_name="stack_element", lazy=None,
...
+                    elif (axis.axis[0] < axis.axis[-1] and axis.axis[-1] < \
```
ericpre: It is a bit difficult to read, maybe add a local variable `_axis = _s.axes_manager[axis_input]`

hyperspy/signal.py
ericpre: From https://github.com/hyperspy/hyperspy/pull/2303, it seems that the current compromise is to have warning. I would keep `sum` as computing the `sum`: this is the correct behaviour, even if users may mistaken it with `integrate`. The alternative, as discussed elsewhere would be to deprecate `sum` but anything else will be even confusing.

hyperspy/_components/exponential.py
ericpre: I would prefer to take the mean rather than a single point.

hyperspy/_signals/eds_tem.py
jlaehne: Would apply to quite a number of functions. I guess if it is not on the bucket list of things to solve soon (like split and rebin) it would make sense. On the other hand, there is always the danger that once we fix it, it is forgotten in the docstring.

hyperspy/_signals/signal1d.py
jlaehne: Similar as above. As here the single value scale is replaced by the whole vector gradient, I would leave it.

hyperspy/axes.py
jlaehne: Right, I mixed up the attributes. Tests updated to check for an axes associated to a signal.

hyperspy/misc/utils.py
jlaehne: Well not as simple. A `FunctionalDataAxis` can have either a `UniformDataAxis` or a `DataAxis` as `x`. So we could keep the `expression` of the first element and extend `x` either like one or the other axis type (by defining a private function that handles the case of `DataAxis` to be reused here).

hyperspy/signal.py
jlaehne: I should not touch `sum`, I agree. However, also for binned signals with uniform axes, `sum` is not a correct choice for an integral if `scale != 1`. So maybe we should add a function `integrate_rectangle`, which scales the sum by either `scale` for uniform axes or using `np.gradient` for non-uniform axes.

hyperspy/tests/model/test_model.py
jlaehne: It is `binned=False and uniform=True`, which is not checked elsewhere. Question is if we need to have every eventuality tested, but I guess it does not hurt here.

hyperspy/tests/signals/test_remove_background.py
jlaehne: @pytest.mark.parametrize('binning,uniform', [(True, False), (True, True), (False, True)])

hyperspy/signal.py
ericpre: Are you saying that `intergrate1D` will be incorrect for non-binned and non-uniform axis? In this case, maybe we should raise a `NotImplementedError`?

hyperspy/_signals/eds_tem.py
ericpre: Yes, it is not strictly necessary but I think that it would be good to add it here to have the docstring as good as possible. We strongly encourage users to read docstring, so they need to be of the highest standard as possible! ;)
Also, usually, when there is not someone pushing for it, it is very likely that these will not be implemented for a while and I would not worry too much about updating the docstring! There is a tickbox in the PR summary to remind us about it!

hyperspy/tests/model/test_model.py
ericpre: Yes, I agree here, this is not a big deal. There is a balance between what is manageable and necessary. Last year or so, we had to rewrite some tests which were overkilled and wasting time, so now I try to be more careful with tests, make sure there are quick and testing only what is necessary.

hyperspy/misc/utils.py
ericpre: OK, then an alternative to the current approach is to extend only in the case of `FunctionalDataAxis` with `x` is an `UniformDataAxis`.
I leave it to your judgement.

hyperspy/signal.py
jlaehne: I try to be more clear:
- For non-binned (even for non-uniform) axis, I do not see any reason that `integrate_simpson` is wrong, so I just took out the `or ...` in the above statement.
- For binned, uniform axis, it simply uses `sum` and the result is thus only equal to the integral if `scale == 1`. I would consider this a bug of the current implementation.
- For binned, non-uniform axis, it currently also would use `sum` and that makes no sense at all (but `sum` already raises a warning that we do not need to duplicate).

To ensure a correct behavior of `integrate1D` for binned signals (latter two cases), I would propose to implement and call an `integrate_rectangle` function (to replace the use of `sum`) that:
- for uniform axis, multiplies the sum by `axis.scale`
- for non-uniform axis, multiplies data by `np.gradient(axis)` along the specified axis before doing the summation

hyperspy/signal.py
```diff
@@ -4272,8 +4272,8 @@ def integrate1D(self, axis, out=None):
 
         The integration is performed using
         `Simpson's rule <https://en.wikipedia.org/wiki/Simpson%%27s_rule>`_ if
-        `axis.is_binned` is ``False`` and simple summation over the
-        given axis if ``True``.
+        `axis.is_binned` is ``False`` or `axis.is_uniform` is ``False`` and
+        simple summation over the given axis if both are ``True``.
```
ericpre: This change needs to be reverted since this is the case anymore!

hyperspy/signal.py
ericpre: As it is now looks good to me: uniform and non-uniform axis behave consistently. The `integrate1D` of binned signal is a separate discussion, which most likely needs more detailed explanation in the docstring/user guide - personally, I am not sure I fully understand the reasoning!

ericpre (approved):
Looks good to me. There is one docstring left to update and two comments I will leave it to you, if you want to take them into account or not.

hyperspy/signal.py
jlaehne: OK, I'll open a separate issue for discussion (#2735) and I added a point to the non-uniform axes tracker (#2398)

hyperspy/_signals/eds_tem.py
jlaehne: I added it to all relevant docstrings, where a NotImplementedError for non-uniform axes is raised.

hyperspy/misc/utils.py
jlaehne: I implemented a proper `stack` for both versions of FunctionalDataAxis.

jlaehne:
I took into account all the comments, thanks for all the feedback @ericpre and @thomasaarholt.

thomasaarholt (approved):

thomasaarholt:
Really nice @jlaehne!

Issue #2726, opened 2021-04-29
align_zero_loss_peak not working for EELS data with 2D navigation axis


ptim0626:
To reproduce:
```python
import numpy as np
import hyperspy.api as hs
s = hs.signals.EELSSpectrum(np.zeros((10, 20, 100)))
s.align_zero_loss_peak()
```
gives
```python
ValueError: The size of the navigation_shape for the kwarg shift (<(200,)> must be consistentwith the size of the mapped signal <(20, 10)>
```

The problem seems to be [here](https://github.com/hyperspy/hyperspy/blob/869d077b364489f2ddd56aafdec7cd806892e586/hyperspy/_signals/signal1d.py#L472), the `ravel` converts the shift array to 1D, which is causing problem. I may miss something, but I don't follow why `squeeze` (or just nothing) was changed to `ravel` ([here](https://github.com/hyperspy/hyperspy/commit/f078c8e23af72f14e4abf190abf4ef8a0dbaf6f7#)).

ericpre:
Do you want to make a PR to fix it? #2617 was a large PR, it is possible that we missed it...

ptim0626:
yes sure, just want to make sure if there's any particular reason for the change before reverting it. I will add a test for EELS data with 2D navigation axis to guard against this as well.

ericpre:
Done in #2729.

PR #2729, opened 2021-05-04
Handle data with 2D navigation axis in shift1D

ptim0626:
### Description of the change
This is a small fix related to #2726. Changing `ravel` back to `squeeze` to handle data with 2D navigation axis in shift1D.

### Progress of the PR
- [x] Change implemented,
- [x] add tests,
- [x] ready for review.

After the change, the following code will work:
```python
import numpy as np
import hyperspy.api as hs
s = hs.signals.EELSSpectrum(np.zeros((10, 20, 100)))
s.align_zero_loss_peak()
```


ericpre (approved):
Perfect, thanks!

PR #2731, opened 2021-05-04
shift1d broken?

AEljarrat:
### Description of the change
It seems the `hs.signals.Signal1D.shift1d` method is broken when using signals with navigation dimension greater than one. I have noticed while trying to align the zero-loss-peak on some EELS. It looks like there's an issue as `shift1d` looses the information of the navigation dimensions from the shift array through a `ravel` call. It seems the `map` computation used to calculate the actual shifts needs this information.

### Progress of the PR
- [ ] add tests,
- [x] ready for review.

```python
%matplotlib qt
import numpy as np
import hyperspy.api as hs

from itertools import product

navigation_shape = [11, 10, ]
signal_shape = [128,]

# generate a synthetic ZLP data set
zdata = np.random.random(navigation_shape + signal_shape)
ids = np.random.randint(signal_shape[0]*0.25, signal_shape[0]*0.75, size=navigation_shape)
for ii in product(*[range(ii) for ii in navigation_shape]):
    zdata[ii][ids[ii]] = 10. 
z = hs.signals.EELSSpectrum(zdata)
z.gaussian_filter(2.5)

# align: this is broken, and this PR fixes it
z.align_zero_loss_peak()
```

I understand if this PR is useless as I only contribute code from time to time. At least this can serve as bug report. 

Thanks for the great work with hyperspy !

Best regards,
Alberto.

### DIFF
--- a/hyperspy/_signals/signal1d.py
+++ b/hyperspy/_signals/signal1d.py
@@ -471,7 +471,7 @@ class Signal1D(BaseSignal, CommonSignal1D):
             axis.offset += minimum
             axis.size += axis.high_index - ihigh + 1 + ilow - axis.low_index
         if isinstance(shift_array, np.ndarray):
-            shift_array = BaseSignal(shift_array.ravel()).T
+            shift_array = BaseSignal(shift_array).T
 
         self.map(_shift1D,
                  shift=shift_array,


ericpre:
Done in #2729, which was created earlier today! Sorry about this regression, #2617 was a large PR changing improving the lazy `map`. The supported parameter of the non-lazy `map` has also changed to reduce inconsistency and ambiguity but we missed this regression...

AEljarrat:
I see, thank you @ericpre for taking care of everything!

Issue #2735, opened 2021-05-09
integrate1D of binned signals


jlaehne:
For binned signals, `integrate1D` currently uses `sum` to perform the integration. In principle, this is a simplified version of the rectangle rule (Riemann sum: https://en.wikipedia.org/wiki/Riemann_sum), where the signal intensity is assumed to be constant over the complete interval associated with an axis point.

However, for uniform axes, this simplification means that `scale == 1` is assumed. Otherwise the result of `sum` is not really the `integral`, but in fact `integral/scale`.

For non-uniform axes, using the `sum` function does not really make sense mathematically. To calculate an integral using the rectangle rule, the data would first need to be multiplied with `np.gradient(axis)`, before the sum is performed.

I would propose to implement a new function `integrate_rectangle` or `integrate_riemann` that basically calculates:
- for uniform axes: `sum(data)*scale`
- for non-uniform axes: `sum(data*np.gradient(axis))`

This function would be called by `integrate1D` when integrating over a binned axis.

thomasaarholt:
Am I correctly interpreting that the current `integrate1D` behaviour actually is a bug for `is_binned == False`, which would be fixed by multiplying by `scale`? (assuming uniform axis)

I think you make a good case and completely agree with your approach.

thomasaarholt:
~~Does this issue have any implications for the case when the signal/axis is binned?
Actually, should the second word in your issue be "unbinned"? :)~~

jlaehne:
No, I am talking about the behavior for `is_binned == True`: https://github.com/hyperspy/hyperspy/blob/ea89fe2a12609b7fe272e048a9bd83af8a51dfda/hyperspy/signal.py#L4225

For unbinned signals, it used `integrate_simpson`: https://github.com/hyperspy/hyperspy/blob/ea89fe2a12609b7fe272e048a9bd83af8a51dfda/hyperspy/signal.py#L4223
which is totally fine and should work both for uniform and non-uniform axes, as the axis vector is explicitly passed to the scipy routine doing the integration: https://github.com/hyperspy/hyperspy/blob/ea89fe2a12609b7fe272e048a9bd83af8a51dfda/hyperspy/signal.py#L4009

thomasaarholt:
Apologies, my brain expected the first line in `integrate1D`, `if is_binned(self, axis=axis)` to be "`if True`", rather than `if ... is False`.

Regardless, yep, agreed with what you're saying.

jlaehne:
[...]

PR #2736, opened 2021-05-11
Swap order of if statement in integrate1D

thomasaarholt:
I have flipped the if statement into a more readable manner.
Tiny change, but was a source of confusion for me in #2735.

### Progress of the PR
- [x] ready for review.

### DIFF
--- a/hyperspy/signal.py
+++ b/hyperspy/signal.py
@@ -4217,12 +4217,13 @@ class BaseSignal(FancySlicing,
         (64,64)
 
         """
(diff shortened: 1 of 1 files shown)


thomasaarholt:
This would have gone to Rnp, but the code I'm changing hasn't been pushed there yet.

jlaehne:
Ah, saw this too late. Duplicated in #2737 (with some additional amendments of the documentation there). But I guess does not hurt merging both.

thomasaarholt:
Nah, let's stick with yours :)

PR #2737, opened 2021-05-11
clarifications on integrate/sum for binned/non-uniform axes  

### LINKED ISSUE #2735: integrate1D of binned signals

jlaehne:
### Description of the change
Based on the discussion in #2735, I did some small clarifications to the docstrings and warnings concerning the inegration/summation of binned axes in view of the non-uniform axes implementation.

### Progress of the PR
- [X] Change implemented,
- [X] update docstring,
- [X] update user guide,
- [X] add tests,
- [X] ready for review.

### DIFF
--- a/doc/user_guide/signal.rst
+++ b/doc/user_guide/signal.rst
@@ -278,9 +278,6 @@ The following example shows how to transform between different subclasses.
        <ComplexSignal1D, title: , dimensions: (20, 10|100)>
 
 
-
-
-
 .. _signal.binned:
 
 Binned and unbinned signals
@@ -288,7 +285,9 @@ Binned and unbinned signals
 
 Signals that are a histogram of a probability density function (pdf) should
 have the ``is_binned`` attribute of the signal axis set to ``True``. The reason
-is that some methods operate differently on signals that are *binned*.
+is that some methods operate differently on signals that are *binned*. An
+example of *binned* signals are EDS spectra, where the multichannel analyzer
+integrates the signal counts in every channel (=bin).
 Note that for 2D signals each signal axis has an ``is_binned``
 attribute that can be set independently. For example, for the first signal
 axis: ``signal.axes_manager.signal_axes[0].is_binned``.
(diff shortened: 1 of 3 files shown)


ericpre (commented):
Could you also mentioned this in the user guide: https://hyperspy.readthedocs.io/en/stable/user_guide/signal.html#binned-and-unbinned-signals?

jlaehne:
Yes, we should definitely update the binned signal section of the user guide.

ericpre:
The user guide link above is the doc of the release version. I should have shared the one of this branch (https://hyperspy--2737.org.readthedocs.build/en/2737/user_guide/signal.html#binned-and-unbinned-signals) or the one for `RELEASE_next_minor`: https://hyperspy.readthedocs.io/en/latest/user_guide/signal.html#binned-and-unbinned-signals, which is very similar.

Maybe without going too much in details of the meaning of `binned` axis, maybe it is worth given the example of the `integrate1D` method, what is different and it's reasoning? Rewording the content of the docstring or what @francisco-dlp explained in https://github.com/hyperspy/hyperspy/issues/2735#issuecomment-836743647 should be good.

jlaehne:
OK, I updated the user guide to include EDS as an example of a binned signal and I added a small paragraph on the integration of binned signals. I would consider this PR as finished (closing #2735), but would keep the item in the todo list of #2398 open in case @francisco-dlp or someone else has another idea of how to further improve the documentation of binned signals.

ericpre (approved):
Thanks!

Issue #2739, opened 2021-05-13
nan and maps non supported by current implementation of scaling factor for binned non-uniform axes


LMSC-NTappy:
The new implementation of scaling factor for binned non-uniform #2728 axes created some issues. Eg in the gaussian implementation:

https://github.com/hyperspy/hyperspy/blob/d1bcd1c540c9a7489a5527f32b15b5680e12f55b/hyperspy/_components/gaussian.py#L165-L168

the evaluation of `value2index(centre)` at line 168 relies on the fact that `_estimate_gaussian_parameters ` at line 165 will output a non-NaN value for centre. 

```python
s = hs.signals.Signal1D(np.zeros(100))

nuax = FunctionalDataAxis(expression="x**2",x=s.axes_manager.signal_axes[0],name='NUAX')
snu = s.deepcopy()
snu.axes_manager.set_axis(nuax,index_in_axes_manager=0)

g = hs.model.components1D.Gaussian(A=1,centre=80,sigma=10)
g.estimate_parameters(snu,20.,900.)
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
<ipython-input-51-6b634f634055> in <module>
----> 1 g.estimate_parameters(snu,20.,900.)

c:\users\xxxxxxx\documents\git\hyperspy\hyperspy\_components\gaussian.py in estimate_parameters(self, signal, x1, x2, only_current)
    166                                                               only_current)
    167         scaling_factor = axis.scale if axis.is_uniform \
--> 168                          else np.gradient(axis.axis)[axis.value2index(centre)]
    169         if only_current is True:
    170             self.centre.value = centre

c:\users\xxxxxxx\documents\git\hyperspy\hyperspy\axes.py in value2index(self, value, rounding)
    567             return (np.abs(self.axis - value)).argmin()
    568         else:
--> 569             index = int(value)
    570             if self.size > index >= 0:
    571                 return index

ValueError: cannot convert float NaN to integer
```
This, regardless of whether the axis is binned or not

Additionnaly, this implementation also fails for signals with navigation dimensions > 0.  It is because `value2index(centre)` at line 168 relies on the fact that `_estimate_gaussian_parameters ` at line 165 will not output an array of values, which it will.

```python 
s = hs.signals.Signal1D(np.zeros((10,10,100)))

nuax = FunctionalDataAxis(expression="x**2",x=s.axes_manager.signal_axes[0],name='NUAX', bin)
snu = s.deepcopy()
[...]

PR #2740, opened 2021-05-14
Fixes binned nua 2

LMSC-NTappy:
### Description of the change
nan and arrays non supported by current implementation of scaling factor for binned non-uniform axes #2739.
This PR brings compatibility
- I started with @ericpre  branch that implements a `_get_scaling_factor(signal, axis, parameter)` function to obtain the scaling factors of binned signals for all cases, ie uniform or non-uniform axes. I made this function compatible with arrays and scalars
- I made `value2index `compatible with numpy arrays in `BaseDataAxis` (it was the case for `UniformDataAxis`)
- I made small changes to _get_scaling_factor so that it is also compatible with scalars and arrays.

Note that nan values still throw errors (they evaluate as out of bounds in value2index). I think that this is a better option than somehow forwarding it, it will make the parameters estimation fail in some weird cases, such as when it is applied to an identically zero signal range. Unfortunately this case can be encountered in practice but specific countermeasures can be implemented (in my case replacing the zero by a "noisy zero".

Now the question remains, what should we do with the `_get_scaling_factor` function? It can stay in components for the moment I guess but maybe it should be moved to misc axis tool? For the moment it takes a signal as an argument but the sole purpose of it is to check for "binnedness" of the signal, which is being moved to an axis property. I have not followed the latest developments of the binned property 

### Progress of the PR
- [x] add tests,
- [x] ready for review.

```python
#Map signal
s = hs.signals.Signal1D(np.arange(900).reshape(3,3,100))

#Creation of a non linear axis signal
binned = True
nuax = FunctionalDataAxis(expression="x**2",x=s.axes_manager.signal_axes[0],name='NUAX',is_binned=binned)
snu = s.deepcopy()
snu.axes_manager.set_axis(nuax,index_in_axes_manager=2)
snu.metadata.Signal.binned = binned

#Parameter estimation, which calls both index2value and _get_scaling_factor
g.estimate_parameters(snu,20.,900.)

g.A.map,g.centre.map
(array([[(  428.16814439, nan,  True), ( 2168.63105067, nan,  True),
         ( 4048.60105503, nan,  True)],
        [( 5807.35705211, nan,  True), ( 7565.93140807, nan,  True),
         ( 9324.43395088, nan,  True)],
        [(11082.90089812, nan,  True), (12841.34763715, nan,  True),
         (14599.78180928, nan,  True)]],
       dtype=[('values', '<f8'), ('std', '<f8'), ('is_set', '?')]),
 array([[(441.        , nan,  True), (344.43347639, nan,  True),
         (337.073903  , nan,  True)],
        [(334.36492891, nan,  True), (332.95678271, nan,  True),
         (332.09390126, nan,  True)],
        [(331.51094891, nan,  True), (331.09071877, nan,  True),
         (330.77342315, nan,  True)]],
       dtype=[('values', '<f8'), ('std', '<f8'), ('is_set', '?')]))
```


LMSC-NTappy:
only failing checks I got on my machine are runtime errors from qt in the peak finding interactive tests. I think unrelated

LMSC-NTappy:
I realised I did the same branching f*ckup as for #2741 so I will also rebase this off upstream/non_uniform_axes. 

I'll create a new PR tonight

hard truth is that I still don't know how to use versioning properly, sorry about that...

ericpre:
Well, it is fairly easy to make mistakes, when not used to git/github! Personally, I found that after I have settled with a git GUI (smartgit), which I am happy with, it made my life much easier.
Also, I find the PR preview on github (just before actually making the PR) quite useful to check my commits. Quite regularly, I use it to compare branches without make any PR!

PR #2741, opened 2021-05-14
Artificial data for non_uniform_axes

LMSC-NTappy:
### Description of the change
I implemented functions to generate two artificial datasets, a single spectrum and a 10x10 spectral map, reasonably similar to the luminescence of a semiconductor band-edge (it is actually just a gaussian). I implemented it following the style of the already existing artificial data series.

I think having such functions would help most testing endeavors in the development of non_uniform_axes (see #2740 and #2739)

### Progress of the PR
- [x] add tests,
- [x] ready for review.

```python
import hyperspy.api as hs
import  hyperspy.datasets.artificial_data as ad
s = ad.get_luminescence_map_nonuniform(uniform=False,add_baseline=False,add_noise=False)
s.plot()
```


hyperspy/datasets/artificial_data.py
```diff
@@ -356,3 +363,231 @@ def get_atomic_resolution_tem_signal2d():
...
+    data = data.reshape((10,10,1024))
+    sig = Signal1D(data,axes=[spax_y,spax_x,nm_axis])
+    #sig.metadata.General.title = 'Artificial Luminescence map'
+
+    #if not uniform, transformation into non-linear axis
+    if not uniform:
+        hc = 1239.84198 #nm/eV
+
+        #eV axis creation. Note how the slice is inverted to have low energy first
+        evax = FunctionalDataAxis(expression="a/x",
```
ericpre, line 575: Would it be simpler to use `convert_to_functional_data_axis`?

hyperspy/datasets/artificial_data.py
```diff
@@ -356,3 +363,231 @@ def get_atomic_resolution_tem_signal2d():
...
+    --------
+    get_low_loss_eels_signal,
+    get_core_loss_eels_signal,
+    get_low_loss_eels_line_scan_signal,
+    get_core_loss_eels_line_scan_signal,
+    get_core_loss_eels_model,
+    get_atomic_resolution_tem_signal2d,
+    get_luminescence_spectrum_nonuniform
+    """
+    from hyperspy.signals import Signal1D
```
ericpre, line 511: Unless strictly necessary, the import should be at the top of the file. Can you move them to the top?

hyperspy/datasets/artificial_data.py
```diff
@@ -356,3 +363,231 @@ def get_atomic_resolution_tem_signal2d():
 
     s = Signal2D(image)
     return s
+
+
+def get_luminescence_spectrum_nonuniform(uniform=False, add_baseline=False, add_noise=True,random_state=None):
```
ericpre, line 368: To reduce duplication of code, an alternative would be to adapt and re-use `get_luminescence_map_nonuniform` use to covert this function. Maybe make rename `get_luminescence_map_nonuniform` to `get_luminescence_signal` and add a `navigation_dimension` argument to `get_luminescence_signal` to cover the behaviour of both functions?

ericpre (commented):
Indeed, this would be useful! I have added a few suggestions for improvement.

Also, there are a couple of files, which should not be in this PR. The easiest may be to create a new branch starting from latest upstream and "cherry-pick" the relevant commits of this PR. Rebasing may not work well here.

hyperspy/datasets/artificial_data.py
LMSC-NTappy: I agree. Will do

hyperspy/datasets/artificial_data.py
LMSC-NTappy: OK I did this by imitation of the other artificial datasets function. Will move it to the top.

hyperspy/datasets/artificial_data.py
LMSC-NTappy: It was. I did

LMSC-NTappy:
> 
> 
> Indeed, this would be useful! I have added a few suggestions for improvement.
> 
> Also, there are a couple of files, which should not be in this PR. The easiest may be to create a new branch starting from latest upstream and "cherry-pick" the relevant commits of this PR. Rebasing may not work well here.

I have created another PR to fix this. See #2742

PR #2743, opened 2021-05-16
Fix estimate_parameters for binned and non uniform axis

### LINKED ISSUE #2739: nan and maps non supported by current implementation of scaling factor for binned non-uniform axes

LMSC-NTappy:
### Description of the change
Rebase of PR #2740 with the correct version of `non_uniform_axes`.
Fixes #2739.

Summary: this PR fixes incorrect behaviour in `BaseDataAxis.value2index()` which was not accepting arrays as input and was computing the axis gradient even for non-binned axes.

### Progress of the PR
- [x] add tests,
- [x] ready for review.

### DIFF
--- a/hyperspy/_components/exponential.py
+++ b/hyperspy/_components/exponential.py
@@ -52,7 +52,7 @@ class Exponential(Expression):
     """
 
     def __init__(self, A=1., tau=1., module="numexpr", **kwargs):
-        super(Exponential, self).__init__(
+        super().__init__(
             expression="A * exp(-x / tau)",
             name="Exponential",
             A=A,
@@ -86,7 +86,7 @@ class Exponential(Expression):
         bool
 
         """
-        super(Exponential, self)._estimate_parameters(signal)
+        super()._estimate_parameters(signal)
         axis = signal.axes_manager.signal_axes[0]
         i1, i2 = axis.value_range_to_indices(x1, x2)
         if i1 + 1 == i2:
--- a/hyperspy/axes.py
+++ b/hyperspy/axes.py
@@ -559,22 +559,33 @@ class BaseDataAxis(t.HasTraits):
         ------
         ValueError
             If any value is out of the axis limits.
-
+            If value is NaN
+            If value is an array containing at least 1 NaN value
+            If value is an array containing at least 1 out-of-bound value
         """
         if value is None:
             return None
-        if self.low_value <= value <= self.high_value:
-            return (np.abs(self.axis - value)).argmin()
+        #Should evaluate on both arrays and scalars. Raises error if there are
+        #nan values in array
+        if np.all((value >= self.low_value)*(value <= self.high_value)):
+            #initialise the index same dimension as input, force type to int
+            index = np.asarray(value).astype(int).copy()
+            #assign on flat, iterate on flat.
+            for i,v in enumerate(np.asarray(value).flat):
+                index.flat[i] = (np.abs(self.axis - v)).argmin()
+            #Squeezing to get a scalar out if scalar in. See squeeze doc
+            return np.squeeze(index)[()]
         else:
-            index = int(value)
-            if self.size > index >= 0:
-                return index
-            else:
-                raise ValueError(
-                    f'The value {value} is out of the limits '
-                    f'[{self.low_value:.3g}-{self.high_value:.3g}] of the '
-                    f'"{self._get_name()}" axis.'
-                    )
+            ### I commented out this because it seemed dangerous
+            # index = int(value)
+            # if self.size > index >= 0:
+            #     return index
+            # else:
+            raise ValueError(
+                f'The value {value} is out of the limits '
+                f'[{self.low_value:.3g}-{self.high_value:.3g}] of the '
+                f'"{self._get_name()}" axis.'
+                )
 
     def index2value(self, index):
         if isinstance(index, da.Array):
@@ -623,7 +634,6 @@ class BaseDataAxis(t.HasTraits):
                 i1 = self.value2index(v1)
             if v2 is not None and self.low_value < v2 <= self.high_value:
                 i2 = self.value2index(v2)
-
         return i1, i2
 
     def update_from(self, axis, attributes):
--- a/hyperspy/axes.py
+++ b/hyperspy/axes.py
@@ -26,13 +26,17 @@ import traits.api as t
 from traits.trait_errors import TraitError
 import pint
 from sympy.utilities.lambdify import lambdify
-
 from hyperspy.events import Events, Event
+from hyperspy.misc.array_tools import numba_closest_index_round, \
+                                        numba_closest_index_floor, \
+                                        numba_closest_index_ceil
+
 from hyperspy.misc.utils import isiterable, ordinal
 from hyperspy.misc.math_tools import isfloat
 from hyperspy.ui_registry import add_gui_method, get_gui
 from hyperspy.defaults_parser import preferences
 from hyperspy._components.expression import _parse_substitutions
+
 
 import warnings
 import inspect
@@ -563,6 +567,10 @@ class BaseDataAxis(t.HasTraits):
         """
         if value is None:
             return None
+        else:
+            value = np.asarray(value)
+
+
 
         #Should evaluate on both arrays and scalars. Raises error if there are
         #nan values in array
@@ -572,27 +580,28 @@ class BaseDataAxis(t.HasTraits):
             #and only evaluate self.axis - v input, v a scalar within value.
             if rounding is round:
                 #Use argmin(abs) which will return the closest value
-                rounding_index = lambda x: np.abs(x).argmin()
+                # rounding_index = lambda x: np.abs(x).argmin()
+                index = numba_closest_index_round(self.axis,value)
             elif rounding is math.ceil:
                 #Ceiling means finding index of the closest xi with xi - v >= 0
                 #we look for argmin of strictly non-negative part of self.axis-v.
                 #The trick is to replace strictly negative values with +np.inf
-                rounding_index = lambda x: np.where(x<0,+np.inf,x).argmin()
+                index = numba_closest_index_ceil(self.axis,value)
             elif rounding is math.floor:
                 #flooring means finding index of the closest xi with xi - v <= 0
                 #we look for armgax of strictly non-positive part of self.axis-v.
                 #The trick is to replace strictly positive values with -np.inf
-                rounding_index = lambda x: np.where(x>0,-np.inf,x).argmax()
+                index = numba_closest_index_ceil(self.axis,value)
             else:
                 raise ValueError(
-                    f'{rounding} is not a supported rounding function. Use '
+                    f'Non-supported rounding function. Use '
                     f'round, math.ceil or math.floor'
                     )
             #initialise the index same dimension as input, force type to int
-            index = np.empty_like(value,dtype=int)
+            # index = np.empty_like(value,dtype=int)
             #assign on flat, iterate on flat.
-            for i,v in enumerate(np.asarray(value).flat):
-                index.flat[i] = rounding_index(self.axis - v)
+            # for i,v in enumerate(value):
+                # index.flat[i] = rounding_index(self.axis - v)
             #Squeezing to get a scalar out if scalar in. See squeeze doc
             return np.squeeze(index)[()]
         else:
--- a/hyperspy/axes.py
+++ b/hyperspy/axes.py
@@ -565,22 +565,39 @@ class BaseDataAxis(t.HasTraits):
         """
         if value is None:
             return None
+
         #Should evaluate on both arrays and scalars. Raises error if there are
         #nan values in array
         if np.all((value >= self.low_value)*(value <= self.high_value)):
+            #Only if all values will evaluate correctly do we implement rounding
+            #function. Rounding functions will strictly operate on numpy arrays
+            #and only evaluate self.axis - v input, v a scalar within value.
+            if rounding is round:
+                #Use argmin(abs) which will return the closest value
+                rounding_index = lambda x: np.abs(x).argmin()
+            elif rounding is math.ceil:
+                #Ceiling means finding index of the closest xi with xi - v >= 0
+                #we look for argmin of strictly non-negative part of self.axis-v.
+                #The trick is to replace strictly negative values with +np.inf
+                rounding_index = lambda x: np.where(x<0,+np.inf,x).argmin()
+            elif rounding is math.floor:
+                #flooring means finding index of the closest xi with xi - v <= 0
+                #we look for armgax of strictly non-positive part of self.axis-v.
+                #The trick is to replace strictly positive values with -np.inf
+                rounding_index = lambda x: np.where(x>0,-np.inf,x).argmax()
+            else:
+                raise ValueError(
+                    f'{rounding} is not a supported rounding function. Use '
+                    f'round, math.ceil or math.floor'
+                    )
             #initialise the index same dimension as input, force type to int
             index = np.empty_like(value,dtype=int)
             #assign on flat, iterate on flat.
             for i,v in enumerate(np.asarray(value).flat):
-                index.flat[i] = (np.abs(self.axis - v)).argmin()
+                index.flat[i] = rounding_index(self.axis - v)
             #Squeezing to get a scalar out if scalar in. See squeeze doc
             return np.squeeze(index)[()]
         else:
-            ### I commented out this because it seemed dangerous
-            # index = int(value)
-            # if self.size > index >= 0:
-            #     return index
-            # else:
             raise ValueError(
                 f'The value {value} is out of the limits '
                 f'[{self.low_value:.3g}-{self.high_value:.3g}] of the '
@@ -1006,7 +1023,7 @@ class FunctionalDataAxis(BaseDataAxis):
     crop.__doc__ = DataAxis.crop.__doc__
 
     def _slice_me(self, slice_):
-        """Returns a slice to slice the 'x' vector of the corresponding 
+        """Returns a slice to slice the 'x' vector of the corresponding
         functional data axis and set the axis accordingly.
 
         Parameters
@@ -1652,7 +1669,7 @@ class AxesManager(t.HasTraits):
         axis: BaseDataAxis axis to replace the current axis with
 
         index_in_axes_manager: index of the axis in current signal to remplace
-            with axis passed in argument 
+            with axis passed in argument
 
         See also
         --------
--- a/hyperspy/axes.py
+++ b/hyperspy/axes.py
@@ -545,7 +545,7 @@ class BaseDataAxis(t.HasTraits):
         return value
 
     def value2index(self, value, rounding=round):
-        """Return the closest index to the given value if between the limit.
+        """Return the closest index to the given value if between the axis limits.
 
         Parameters
         ----------
@@ -558,10 +558,8 @@ class BaseDataAxis(t.HasTraits):
         Raises
         ------
         ValueError
-            If any value is out of the axis limits.
-            If value is NaN
-            If value is an array containing at least 1 NaN value
-            If value is an array containing at least 1 out-of-bound value
+            If value is out of bounds or contains out of bounds values (array).
+            If value is NaN or contains NaN values (array).
         """
         if value is None:
             return None
@@ -1173,10 +1171,12 @@ class UniformDataAxis(BaseDataAxis, UnitConversion):
         Raises
         ------
         ValueError
-            If any value is out of the axis limits.
-            If the value is incorrectly formatted.
-
+            If value is out of bounds or contains out of bounds values (array).
+            If value is NaN or contains NaN values (array).
+            If value is incorrectly formatted str or contains incorrectly
+                formatted str (array).
         """
+
         if value is None:
             return None
 
--- a/hyperspy/axes.py
+++ b/hyperspy/axes.py
@@ -569,7 +569,7 @@ class BaseDataAxis(t.HasTraits):
         #nan values in array
         if np.all((value >= self.low_value)*(value <= self.high_value)):
             #initialise the index same dimension as input, force type to int
-            index = np.asarray(value).astype(int).copy()
+            index = np.empty_like(value,dtype=int)
             #assign on flat, iterate on flat.
             for i,v in enumerate(np.asarray(value).flat):
                 index.flat[i] = (np.abs(self.axis - v)).argmin()
--- a/hyperspy/component.py
+++ b/hyperspy/component.py
@@ -26,7 +26,7 @@ from distutils.version import LooseVersion
 from pathlib import Path
 
 import hyperspy
-from hyperspy.misc.utils import slugify
+from hyperspy.misc.utils import slugify, is_binned
 from hyperspy.misc.io.tools import (incremental_filename,
                                     append2pathname,)
 from hyperspy.misc.export_dictionary import export_to_dictionary, \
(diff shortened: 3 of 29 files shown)


hyperspy/axes.py
```diff
@@ -559,22 +559,33 @@ def value2index(self, value, rounding=round):
...
         """
         if value is None:
             return None
-        if self.low_value <= value <= self.high_value:
-            return (np.abs(self.axis - value)).argmin()
+        #Should evaluate on both arrays and scalars. Raises error if there are
+        #nan values in array
+        if np.all((value >= self.low_value)*(value <= self.high_value)):
+            #initialise the index same dimension as input, force type to int
+            index = np.asarray(value).astype(int).copy()
```
ericpre:
```suggestion
            index = np.empty_like(value, dtype=int)
```

hyperspy/tests/axes/test_data_axis.py
```diff
@@ -314,6 +314,46 @@ def test_calibrate(self):
         with pytest.raises(TypeError, match="only for uniform axes"):
             self.axis.calibrate(value_tuple=(11,12), index_tuple=(0,5))
 
+    def test_functional_value2index(self):
```
ericpre, line 323: Because this test is for a method of `BaseDataAxis`, it would be better to move this test to the `TestBaseDataAxis` class.

hyperspy/component.py
```diff
@@ -1216,3 +1216,40 @@ def print_current_values(self, only_free=False, fancy=True):
...
+
+    """
+
+    if is_binned(signal):
+    # in v2 replace by
+    #if axis.is_binned:
+        if axis.is_uniform:
+            scaling_factor = axis.scale
+        else:
+            # if isinstance(parameter, (int, float)):
```
ericpre: Now that `value2index` takes array as input, this can be clean up.

hyperspy/axes.py
```diff
@@ -559,22 +559,33 @@ def value2index(self, value, rounding=round):
...
-                    f'The value {value} is out of the limits '
-                    f'[{self.low_value:.3g}-{self.high_value:.3g}] of the '
-                    f'"{self._get_name()}" axis.'
-                    )
+            ### I commented out this because it seemed dangerous
+            # index = int(value)
+            # if self.size > index >= 0:
+            #     return index
+            # else:
+            raise ValueError(
```
ericpre, line 608: Add a test for this case?

hyperspy/axes.py
```diff
@@ -559,22 +559,33 @@ def value2index(self, value, rounding=round):
...
-            if self.size > index >= 0:
-                return index
-            else:
-                raise ValueError(
-                    f'The value {value} is out of the limits '
-                    f'[{self.low_value:.3g}-{self.high_value:.3g}] of the '
-                    f'"{self._get_name()}" axis.'
-                    )
+            ### I commented out this because it seemed dangerous
+            # index = int(value)
```
ericpre: Yes, I agree and I ignore it because, I think that it may be better to discuss this in a separate PR. If you want to fix this in this PR, the `value2index` of `UniformDataAxis` needs to be changed consistently.

hyperspy/tests/axes/test_data_axis.py
```diff
@@ -314,6 +314,46 @@ def test_calibrate(self):
...
+        arval[1,1] = -0.3
+        with pytest.raises(ValueError):
+            self.axis.value2index(arval)
+        #One NaN in array in --> error out
+        arval[1,1] = np.nan
+        with pytest.raises(ValueError):
+            self.axis.value2index(arval)
+        #Single-value-array-in --> scalar out
+        arval = np.array([1.0])
+        assert type(self.axis.value2index(arval)) == np.int32
```
ericpre: This test is failing:
```python
        arval = np.array([1.0])
>       assert type(self.axis.value2index(arval)) == np.int32
E       AssertionError: assert <class 'numpy.int64'> == <class 'numpy.int32'>
E        +  where <class 'numpy.int64'> = type(1)
```

maybe use something along the line of:
```python
np.issubdtype(self.axis.value2index(arval), int)
```

ericpre (commented):
This looks good to me and I have left a couple of comments.
I will leave someone else to review/merge it since I influenced this PR!

hyperspy/tests/axes/test_data_axis.py
LMSC-NTappy: I thought too, but since BaseDataAxis can practically not be instantiated it felt awkward. It is, indeed, a method of BaseDataAxis but will never ever be called outside the subclasses. 

I can still move it there if you think that's better. I'll need to do something like 
```python
bax = BaseDataAxis()
bax.size = 100
etc...
```

hyperspy/component.py
LMSC-NTappy: Sure, I just forgot

hyperspy/axes.py
LMSC-NTappy: There are, it's the assert that cover nan input and out-of-bound values

hyperspy/axes.py
LMSC-NTappy: Ok, you are right it's way more elegant

hyperspy/tests/axes/test_data_axis.py
LMSC-NTappy: By the time i saw your comments I had already changed it to `np.isscalar` which I think works too but I'll change it to `np.issubdtype` if you think it's best

hyperspy/axes.py
LMSC-NTappy: This change already brings consistency with UniformDataAxis.

As implemented, FunctionalDataAxis.value2index allows for something very wrong to happen, ie it interpretates the value as an index if it can interpretate it as an int.

_Voyez plutot_

```python
>>> from hyperspy.axes import FunctionalDataAxis,UniformDataAxis
>>> uax = UniformDataAxis(size=101.0,offset=0.0,scale=0.01)
>>> uax.value2index(50)
<ipython-input-17-5cf2ab8da91f> in <module>
      1 uax = UniformDataAxis(size=101.0,offset=0.0,scale=0.01)
----> 2 uax.value2index(50)

c:\users\nicolastappy\documents\git\hyperspy\hyperspy\axes.py in value2index(self, value, rounding)
   1177                 return index
   1178             else:
-> 1179                 raise ValueError("The value is out of the axis limits")
   1180 
   1181     def update_axis(self):

ValueError: The value is out of the axis limits
```

Which is expected. Now
```python
>>> nuax = FunctionalDataAxis(expression="x**2",x=uax)
>>> nuax.value2index(69.69)
69
>>> nuax.value2index(99)
99
```
Which is absolutely wrong since `nuax` spans 0 to 1 like `uax`. This works with `uax` because double values are converted to indices before. I think this was placed here with the intent to let value2index fall back on `int==index`
like `isig[]` slices but in fact, it's simply wrong.

I am fine making another PR about this, but I'd rather have this fixed ASAP. If everyone is fine with that I'll just like to delete these line that are commented out

jlaehne:
Looks good on first sight, I'll have a closer look in the next days when I find some time.

hyperspy/axes.py
ericpre: Yes, indeed, sorry, I may have confused myself with the previous PR, when copying and pasting the review which I started in #2740!

hyperspy/tests/axes/test_data_axis.py
ericpre: Yes, `np.isscalar` is fine too!

hyperspy/axes.py
ericpre: I looked at it in more depth, and indeed this is no reason for this to be there. When looking at it quickly, I used the `UniformDataAxis.value2index` as a reference, but it is different, because the index is calculated before. I was expecting this to follow a logic but I looks to me that when this was implemented, this fell through the cracks!

I would suggest to remove it because the if block before that cover scalar and arrays. Would it be possible to remove `UniformDataAxis.value2index` and use `BaseDataAxis.value2index` instead?

hyperspy/tests/axes/test_data_axis.py
ericpre: I would say that it would still be more tidy to have this test in `TestBaseDataAxis` but at the same time, this is true that most test are in `TestDataAxis`!
I was suggesting this because it happen to me in the past to write tests which were already written somewhere else, just not where I expected them to be...

ericpre (approved):
This looks good to me, just need a tiny bit of tidying up. @jlaehne, I leave it to you. Thanks!

hyperspy/axes.py
LMSC-NTappy: I will do that then.

I will then also implement the "rounding" argument (which I actually forgot in the first place)

LMSC-NTappy: > I would suggest to remove it because the if block before that cover scalar and arrays. Would it be possible to remove `UniformDataAxis.value2index` and use `BaseDataAxis.value2index` instead?

This failed pretty badly. That requires to support: string input with units, and relative position in the axis. So  `UniformDataAxis.value2index` uses some of `UnitConversion` methods to parse string inputs. I think UnitConversion should be made compatible with FunctionalDataAxis in another PR. 

I forced push back my latest modifications

ericpre: Agree, let's finish this PR and #2747 and then it would be easier to remove `UniformDataAxis.value2index`.

hyperspy/_components/exponential.py
```diff
@@ -52,7 +52,7 @@ class Exponential(Expression):
     """
 
     def __init__(self, A=1., tau=1., module="numexpr", **kwargs):
-        super(Exponential, self).__init__(
+        super().__init__(
```
jlaehne, line 55: Why remove the attributes here and not for other components?

hyperspy/signal.py
```diff
@@ -2656,7 +2656,13 @@ def get_1D_sum_explorer_wrapper(*args, **kwargs):
             navigator = self
             # Sum over all but the first navigation axis.
             am = navigator.axes_manager
-            navigator = navigator.sum(am.signal_axes + am.navigation_axes[1:])
+            with warnings.catch_warnings():
+                warnings.filterwarnings("ignore", category=UserWarning,
+                                        module='hyperspy'
+                                        )
+                navigator = navigator.sum(
```
jlaehne, line 2663: It would make sense to use integrate1D here, which correctly decides on whether a `sum` can be made (binned signal) or an integration is needed (esp. non-uniform, unbinned signal)  -- in fact, it is on the todo-list in #2398. Then we could avoid catching the warning here.

hyperspy/signal.py
```diff
@@ -4317,7 +4323,11 @@ def integrate1D(self, axis, out=None):
         if is_binned(self, axis=axis):
         # in v2 replace by
         # self.axes_manager[axis].is_binned
-            return self.sum(axis=axis, out=out)
+            with warnings.catch_warnings():
```
jlaehne: Isn't this catch obsolete as this warning should be raised only if `is_binned = False`: https://github.com/hyperspy/hyperspy/blob/914bebd2944f3d63d9ffe02a52aa11d0f311e5a9/hyperspy/signal.py#L3700

hyperspy/tests/axes/test_data_axis.py
jlaehne: Couldn't we add a comment to `TestBaseDataAxis` to indicate that this is tested in `TestDataAxis` @ericpre ?

jlaehne: Actually, it is checked for all three axes types individually, so that could be mentioned in the `TestBaseDataAxis` class.

jlaehne, line 323: For the other two axes types, the different types of inputs for `value2index` and possible errors are tested in a number of individual functions, while here they are all in one function. Wouldn't it make sense to decide for one way for all three axes types? I'm fine with either.

jlaehne: Thinking about it, the single function is probably more clear. Would you want to adapt it for the other axes types?

jlaehne (commented):
Thanks a lot for the help with the nua branch! I left a few questions, but as said previously this looks good otherwise.

hyperspy/axes.py
jlaehne: Add to todo list in #2398?

hyperspy/_components/exponential.py
ericpre: This is old python 2 syntax (https://www.python.org/dev/peps/pep-3135, https://docs.python.org/3/library/functions.html#super), when I come across them, I trend to remove them...

hyperspy/tests/axes/test_data_axis.py
LMSC-NTappy: >Actually, it is checked for all three axes types individually, so that could be mentioned in the TestBaseDataAxis class.

Ok i will do it. 

I would suggest, in a later PR, to formalize the fact that BaseDataAxis is an abstract class, and provide guidelines for the implementation of different axes types.

hyperspy/axes.py
LMSC-NTappy: On a second thought, I think that keeping the possibility of overriding `BaseDataAxis.value2index` is still a good option. Coding universal axes methods in `BaseDataAxis` is smart, because it allows users to easily adapt hyperspy to their needs, while relying on basic functionality. 

However, I would argue that a few specific cases (`UniformDataAxis `being the most common one) will cover most of the experimental needs. For those, keeping override of `BaseDataAxis ` as a standard encourages the implementation of optimised code in dedicated axis classes 

Here is a picture of `UniformDataAxis.value2index()` vs `BasedataAxis.value2index()`, which, I think, is self-explanatory
 
image

hyperspy/_components/exponential.py
LMSC-NTappy: I can take care of removing it from other components then

hyperspy/axes.py
ericpre: Do you know which part is slow? If so, can it be fixed?

hyperspy/signal.py
ericpre: Is there a reason to use `integrate1D` in favour of `sum`?

Using `integrate_simpson` will not work for `len(am.signal_axes + am.navigation_axes[1:] > 1` where `am` is the `axes_manager`.

The following example will not work:
```python
import hyperspy.api as hs
import numpy as np

s = hs.signals.Signal2D(np.arange(5*10*15).reshape(5, 10, 15))
s.plot(navigator='spectrum')
```

LMSC-NTappy: The idea is to preserve the physical meaning of the signal I presume. But I agree that this is out of reach for a simple PR like this one.

The objective of this pull request is just to fix some faulty behaviour in `BaseDataAxis.value2index()` (I updated the description to make this clear) but I am fine with attacking the `Integrate1D` question in another PR

hyperspy/_components/exponential.py
LMSC-NTappy: ...Done

hyperspy/axes.py
LMSC-NTappy: It cannot unfortunately. `np.argmin(axis-value)` is the nominal strategy for finding a closest value using numpy, as far as I know.
Still it is O(N). In comparison, explicit inversion of the axis function as in `UniformDataAxis` is probably O(1) or something close. 

I don't see any way to improve the speed of this. This needs to be applicable in `DataAxis`, which is purely value-defined.

A solution can be found in `FunctionalDataAxis` though, provided the function can be inverted.

hyperspy/signal.py
jlaehne: Yes, that was the idea - but probably you're right to leave this one for later.

hyperspy/signal.py
jlaehne: Only open point remaining.

LMSC-NTappy: I think this is necessary until the v2 because `is_binned` will evaluate from the signal attribute until then. So `is_binned` could evaluate `True` and then `not ax.is_binned` could evaluate `True` as well, if you integrate on a non-uniform-non-binned axis of a signal that otherwise has `signal.metadata.Signal.binned = True` set due to another axis. 

See here:
https://github.com/hyperspy/hyperspy/blob/0c6da6ec9ca1d27b6bc33a3f5e4ad31339bcb45c/hyperspy/misc/utils.py#L1430-L1438

(I did not implement this change however )

hyperspy/tests/axes/test_data_axis.py
LMSC-NTappy: Should be good now

hyperspy/signal.py
ericpre: Thanks, your are both correct and this is now fixed in 2e757e9.

hyperspy/axes.py
ericpre: Assuming the for loop is the cause of the poor performance, could it be vectorised? If not, use numba?

LMSC-NTappy: I agree to try.

I will try to have numba'ed functions for rounding and checking if values are within limits. numba_closest_index_round, numba_closest_index_floor, numba_closest_index_ceil, numba_check_limits and let the BaseDataAxis.value2index take care of input formatting

hyperspy/axes.py
```diff
@@ -559,22 +559,50 @@ def value2index(self, value, rounding=round):
...
+            #initialise the index same dimension as input, force type to int
+            index = np.empty_like(value,dtype=int)
+            #assign on flat, iterate on flat.
+            for i,v in enumerate(np.asarray(value).flat):
+                index.flat[i] = rounding_index(self.axis - v)
+            #Squeezing to get a scalar out if scalar in. See squeeze doc
+            return np.squeeze(index)[()]
+        else:
+            raise ValueError(
+                f'The value {value} is out of the limits '
```
jlaehne, line 609: Can you add a test for this line (see codecov warning)?

hyperspy/axes.py
```diff
@@ -559,22 +559,50 @@ def value2index(self, value, rounding=round):
         ------
         ValueError
```
jlaehne, line 564: Can you harmonize the docstrings of the two value2index functions (also Parameters section and first line) - or define a common string that is used in both.

hyperspy/axes.py
LMSC-NTappy: Ok it kinda works with numba.

However the argument stays the same. I could make it ~10 times faster with numba, I can probably make `UniformDataAxis.value2index() `10times faster as well

image

hyperspy/axes.py
LMSC-NTappy: Looks weird to me. This is the case where a value is out of axis and is tested here:

https://github.com/LMSC-NTappy/hyperspy/blob/2e757e940dd11b0d913cb1c78348beb54c133f63/hyperspy/tests/axes/test_data_axis.py#L332-L336

https://github.com/LMSC-NTappy/hyperspy/blob/2e757e940dd11b0d913cb1c78348beb54c133f63/hyperspy/tests/axes/test_data_axis.py#L348-L355

hyperspy/axes.py
LMSC-NTappy: It should be pretty harmonised now

hyperspy/axes.py
jlaehne: It is definitely nice to have this optimized, but I agree to your above statement that it does not hurt to override certain methods if they can be implemented more efficiently.

hyperspy/axes.py
jlaehne: Weird, now even more warnings have appeared, but one is even for a comment

hyperspy/axes.py
jlaehne: @ecripre I would propose to finish this PR and we can further optimize `value2index` in later PRs, what do you think?

ericpre: > @ecripre I would propose to finish this PR and we can further optimize `value2index` in later PRs, what do you think?

Yes!

jlaehne:
OK, I merge this PR as the main objectives are more than covered. Further optimization that we discussed at several points can be contributed as new PRs. Thanks again!

LMSC-NTappy:
This was a nice ride. Thank you!

PR #2749, opened 2021-05-18
Add warning to user guide saying that non-uniform axes are in beta state

ericpre:

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -102,7 +102,7 @@ jobs:
     steps:
       - uses: actions/checkout@v2
 
-      - uses: ammaraskar/sphinx-action@master
+      - uses: ericpre/sphinx-action@latest_sphinx
         with:
 
           pre-build-command: "apt-get update -y && apt-get install -y ${{ env.BUILD_DEPS }} ${{ env.LATEX_DEPS }}  && pip install -e .[all,build-doc]"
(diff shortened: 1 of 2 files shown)


francisco-dlp:
What's the point of this warning?

ericpre:
The idea of warning users that non-uniform axis in beta state is to set users expectation that the implementation may not work as well as with uniform axis, i.e they may be things that we have missed. Maybe, it needs more explanation? For example mention that the API of non-uniform axis may break in a non major release.

francisco-dlp:
Maybe we can leave the decision about the status and the API policy once we merge non uniform axes into Rnm?

ericpre:
Yes, fine with me!

jlaehne (approved):

jlaehne:
I think we were pretty thorough in adding tests, so chances that something may break might not even be higher than with other functionalities of hyperspy not yet covered by tests. Also we raise errors or warnings for everything we have identified as not (yet) implemented and it is already mentioned in a `Note` that not all functions are implemented.

That said, I do not object to a warning, currently I would say that we are in `beta` state - but I expect some more advances before it is merged (though I think we are not so far from that any more). So +1 for the idea to wait until we merge nua into RnM to decide on the status.

jlaehne:
As the non_uniform_axes branch was merged into RnM, I propose rebasing+merging or just dropping this PR. I think we have added enough warnings, but I do not object to adding the beta note either.

ericpre:
The failure coming from changes in the development version of scipy and has been reported in https://github.com/scipy/scipy/issues/14589.

The motivation for warning user of the beta state of non-uniform axis is to be able to change the API if necessary.

jlaehne (approved):

PR #2750, opened 2021-05-19
Cleanup for the implementation of shift1D

LMSC-NTappy:
### Description of the change
- Removed some unnecessary operations performed in `signal1d._shift1D`
- `axis` was computed twice without being used inbetween
- I replaced the function by a more compact and understandable line of code
- Change invisible to the user
- 30% speed gain in _shift1D execution

### Progress of the PR
- [x] ready for review.

There is room for more improvement by removing (offset,scale,size) kwargs, which are actually not needed.

```python
import hyperspy.api as hs
from hyperspy._signals.signal1d import _shift1D

s = hs.datasets.artificial_data.get_low_loss_eels_signal()

#Before the Change
%timeit _shift1D(s.data,shift=4.0,\
         original_axis=s.axes_manager[0].axis,\
         offset=s.axes_manager[0].offset,\
         scale =s.axes_manager[0].scale,\
         size  =s.axes_manager[0].size)
260 µs ± 6.44 µs per loop (mean ± std. dev. of 7 runs, 1000 loops each)

#After the change
%timeit _shift1D(s.data,shift=4.0,\
         original_axis=s.axes_manager[0].axis,\
         offset=s.axes_manager[0].offset,\
         scale =s.axes_manager[0].scale,\
         size  =s.axes_manager[0].size)
181 µs ± 2.36 µs per loop (mean ± std. dev. of 7 runs, 10000 loops each)

```

### DIFF
--- a/hyperspy/_signals/signal1d.py
+++ b/hyperspy/_signals/signal1d.py
@@ -251,16 +251,14 @@ def _shift1D(data, **kwargs):
     size = kwargs.get('size', 2)
     if np.isnan(shift) or shift == 0:
         return data
-    axis = np.linspace(offset, offset + scale * (size - 1), size)
 
-    si = sp.interpolate.interp1d(original_axis,
-                                 data,
-                                 bounds_error=False,
-                                 fill_value=fill_value,
-                                 kind=kind)
-    offset = float(offset - shift)
-    axis = np.linspace(offset, offset + scale * (size - 1), size)
-    return si(axis)
+
+    #This is the interpolant function
+    si = interpolate.interp1d(original_axis, data, bounds_error=False,
+                              fill_value=fill_value, kind=kind)
+
+    #Evaluate interpolated data at shifted positions
+    return si(original_axis-shift)
 
 
 class Signal1D(BaseSignal, CommonSignal1D):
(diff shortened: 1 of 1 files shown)


thomasaarholt:
The change looks good! 

Did you check if there were any tests that covered the shift1d behaviour? If not, could you write one and confirm that the behaviour is the same before and after this PR?

LMSC-NTappy:
> Did you check if there were any tests that covered the shift1d behaviour? If not, could you write one and confirm that the behaviour is the same before and after this PR?

Indeed, there is this
https://github.com/hyperspy/hyperspy/blob/ea89fe2a12609b7fe272e048a9bd83af8a51dfda/hyperspy/tests/signals/test_1D_tools.py#L107-L117

which:
-generates a spectral image,
-shifts spectra by random amounts except for the first one that stays in place to serve as a reference
-estimate shifts of the shifted signal and verify that it is identical to the randomly chosen shift values.

I think this testing procedure is quite robust and the PR passes it.

thomasaarholt (approved):

hyperspy/_signals/signal1d.py
```diff
@@ -251,16 +251,14 @@ def _shift1D(data, **kwargs):
     size = kwargs.get('size', 2)
```
ericpre: `offset`, `scale` and `size` are not necessary anymore, remove them?

LMSC-NTappy: indeed! They are now removed

ericpre (approved):

PR #2754, opened 2021-05-25
Add fix for lazy s.map and nav chunks with size of 1

### LINKED ISSUE #2748: Mapping of lazy signals with chunksize 1 at the last axis of the navigation axis
Mapping of lazy signal seems not working if the chunk of the last axis of the navigation axes is 1 or (size mod chunk) is 1:

```python
import numpy as np
import dask.array as da
import hyperspy.api as hs

def f(p):
    return np.array([1,2,3])

s = hs.signals.Signal2D(da.random.random((10, 15, 32, 32))).as_lazy()
s.rechunk(nav_chunks=(1, 1))
ss = s.map(f, inplace=False, parallel=False, ragged=True)
ss.compute()
```
Same for `nav_chunks=(1, 2)` 
```python
s.rechunk(nav_chunks=(1, 2))
ss = s.map(f, inplace=False, parallel=False, ragged=True)
ss.compute()
```
It gives 
```python
```
when computing the dask array. This will print out all failed chunks at the last dimension of the navigation axis.
```python
for k in range(1,16):
    try:
        s.rechunk(nav_chunks=(1, k))
        ss = s.map(f, inplace=False, parallel=False, ragged=True)
        ss.compute()
    except IndexError:
        print(k)
```
gives
```python
1
2
7
14
```

Is this the expected behaviour?

magnunor:
Fix for https://github.com/hyperspy/hyperspy/issues/2748

(Also relevant for https://github.com/hyperspy/hyperspy/pull/2703)

### DIFF
--- a/hyperspy/misc/utils.py
+++ b/hyperspy/misc/utils.py
@@ -1229,10 +1229,11 @@ def process_function_blockwise(data,
             output_array[islice] = function(data[islice],
                                             **iter_dict,
                                             **kwargs)
-    try:
-        output_array = output_array.squeeze(-1)
-    except ValueError:
-        pass
+    if not (chunk_nav_shape == output_array.shape):
+        try:
+            output_array = output_array.squeeze(-1)
+        except ValueError:
+            pass
     return output_array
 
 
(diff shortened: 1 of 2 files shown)


magnunor:
Ready for review when/if the tests pass

magnunor:
The failing tests seems to be due to unrelated issues. For the macos, some compile error with gcc. And for the Windows_Python38 some image comparison?

So this is ready for review!

jlaehne:
LGTM.

The same tests are currently failing elsewhere too, probbaly need to be fixed in the actions setup.

francisco-dlp:
LGTM too.

Issue #2237, opened 2019-07-29
Enhancement requst: axes_manager should have a convenience method for easy calibration


jat255:
A colleague of mine came across a confusing issue when trying to change the units of a signal that did not have the axes already calibrated. 

Given the axes manager:

Navigation axis name | size | index | offset | scale | units
-- | -- | -- | -- | -- | --
phase | 10 | 0 | 0.0 | 20.0 | degrees

Signal axis name | size | offset | scale | units
-- | -- | -- | -- | --
x | 3956 | 0.0 | 1.0 |  
y | 2574 | 98.0 | 1.0

Running `s.axes_manager.convert_units(axes="x", units="um", same_units=True, factor=19.512195)` results in a `DimensionalityError: Cannot convert from 'dimensionless' (dimensionless) to 'micrometer' ([length])`.

While this error is descriptive, it would be nice if the code could "force" the units as given if there is no calibration currently defined. It would allow for slightly more concise code than:

```python
s.axes_manager[0].name = 'x'
s.axes_manager[0].scale = 19.512195
s.axes_manager[0].units = 'um'
```

What are others' thoughts?

ericpre:
I would say that not allowing to use `convert_units` for an axis which is not already calibrated is a good behaviour and the error message is fairly clear.
What would possibly be more suitable is a `calibrate` method and there is already [one](https://github.com/hyperspy/hyperspy/blob/d856f2af233d5d4d5aac096d51b834666135273b/hyperspy/axes.py#L553) for the `DataAxis` but it is not documented and it does not seem to do what you want here. So maybe it would worth thinking about extending this method and add a similar one to the `AxesManager`.

jat255:
I agree; I think having an easily used `calibrate` method would be a good idea, and would be more concise. Then if `convert_units` fails, it could point the user to `calibrate` first.

francisco-dlp:
@jat255, would you like to close this then?

jat255:
@francisco-dlp alright to leave it on the wish list? I think there's an opportunity for enhancement here, I just haven't gotten around to it yet.

francisco-dlp:
Sounds good, but in that case it would be good to change the title to something that describes better the improvement that you are proposing.

PR #2756, opened 2021-05-28
Add ``set()`` and ``get`` methods to ``AxesManager``'s ``signal_axes`` and ``navigation_axes``

francisco-dlp:
The goal is to ease the task of setting up the axes attributes, see below for an example. 

The syntax is inspired by (but different from) https://matplotlib.org/stable/users/whats_new.html#axes-spines-access-shortcuts

### Progress of the PR
- [x] add tests,
- [x] ready for review.

```python
>>> import hyperspy.api as hs
>>> import numpy as np
>>> s = hs.signals.Signal1D(np.random.random((10, 20, 30)))
>>> s.axes_manager.navigation_axes.set(name=("x", "y"), scale=1e-3, units="nm")
>>> s.axes_manager.signal_axes.set(name="Energy loss", offset=300, units="eV")
>>> s.axes_manager.navigation_axes.get("name", "offset")
{'name': ('x', 'y'), 'offset': (0.0, 0.0)}
>>> s.axes_manger
<Axes manager, axes: (20, 10|30)>
               x |     20 |      0 |       0 |   0.001 |     nm 
               y |     10 |      0 |       0 |   0.001 |     nm 
     Energy loss |     30 |        |   3e+02 |       1 |     eV
```

### DIFF
--- a/doc/user_guide/axes.rst
+++ b/doc/user_guide/axes.rst
@@ -107,6 +107,17 @@ attributes, e.g.:
     >>> s.axes_manager[0]
     <X axis, size: 20, index: 0>
 
+.. version_added:: 2.2
+    `set` method for `navigation_axes` and `signal_axes`.
+
+It is also possible to set multiple attributes of multiple axes at once, using the :meth:`~.misc.utils.TupleSA.set`
+of the `navigation_axes` and `signal_axes` attributes.
+For example:
+
+.. code-block:: python
+
+    >>> s.axes_manager.navigation_axes.set(name=("X", "Y"), offset=10, units="nm")
+
 
 Once the name of an axis has been defined it is possible to request it by its
 name e.g.:
@@ -118,6 +129,8 @@ name e.g.:
     >>> s.axes_manager["X"].scale = 0.2
     >>> s.axes_manager["X"].units = "nm"
     >>> s.axes_manager["X"].offset = 100
+
+The 
 
 
 It is also possible to set the axes properties using a GUI by calling the
--- a/doc/user_guide/axes.rst
+++ b/doc/user_guide/axes.rst
@@ -111,7 +111,7 @@ attributes, e.g.:
     :meth:`~.misc.utils.TupleSA.set` method for :attr:`~.axes.AxesManager.navigation_axes` and :attr:`~.axes.AxesManager.signal_axes`.
 
 It is also possible to set multiple attributes of multiple axes at once, using the :meth:`~.misc.utils.TupleSA.set`
-of the `navigation_axes` and `signal_axes` attributes.
+of the :attr:`~.axes.AxesManager.navigation_axes` and :attr:`~.axes.AxesManager.signal_axes` attributes.
 For example:
 
 .. code-block:: python
--- a/doc/user_guide/axes.rst
+++ b/doc/user_guide/axes.rst
@@ -107,7 +107,7 @@ attributes, e.g.:
     >>> s.axes_manager[0]
     <X axis, size: 20, index: 0>
 
-.. version_added:: 2.2
+.. versionadded:: 2.2
     `set` method for `navigation_axes` and `signal_axes`.
 
 It is also possible to set multiple attributes of multiple axes at once, using the :meth:`~.misc.utils.TupleSA.set`
--- a/doc/user_guide/axes.rst
+++ b/doc/user_guide/axes.rst
@@ -130,7 +130,6 @@ name e.g.:
     >>> s.axes_manager["X"].units = "nm"
     >>> s.axes_manager["X"].offset = 100
 
-The 
 
 
 It is also possible to set the axes properties using a GUI by calling the
--- a/doc/user_guide/axes.rst
+++ b/doc/user_guide/axes.rst
@@ -108,15 +108,18 @@ attributes, e.g.:
     <X axis, size: 20, index: 0>
 
 .. versionadded:: 2.2
-    :meth:`~.misc.utils.TupleSA.set` method for :attr:`~.axes.AxesManager.navigation_axes` and :attr:`~.axes.AxesManager.signal_axes`.
+    :meth:`~.misc.utils.TupleSA.set` and :meth:`~.misc.utils.TupleSA.get` methods for :attr:`~.axes.AxesManager.navigation_axes`
+    and :attr:`~.axes.AxesManager.signal_axes`.
 
 It is also possible to set multiple attributes of multiple axes at once, using the :meth:`~.misc.utils.TupleSA.set`
 of the :attr:`~.axes.AxesManager.navigation_axes` and :attr:`~.axes.AxesManager.signal_axes` attributes.
-For example:
+The :meth:`~.misc.utils.TupleSA.get` returns a dictionary of the attributes.  For example:
 
 .. code-block:: python
 
     >>> s.axes_manager.navigation_axes.set(name=("X", "Y"), offset=10, units="nm")
+    >>> s.axes_manager.navigation_axes.get("name", "offset", "units")
+    {"name" : ("X", "Y"), "offset" : (10, 10), "units" : ("nm", "nm")}
 
 
 Once the name of an axis has been defined it is possible to request it by its
--- a/doc/user_guide/axes.rst
+++ b/doc/user_guide/axes.rst
@@ -108,7 +108,7 @@ attributes, e.g.:
     <X axis, size: 20, index: 0>
 
 .. versionadded:: 2.2
-    `set` method for `navigation_axes` and `signal_axes`.
+    :meth:`~.misc.utils.TupleSA.set` method for :attr:`~.axes.AxesManager.navigation_axes` and :attr:`~.axes.AxesManager.signal_axes`.
 
 It is also possible to set multiple attributes of multiple axes at once, using the :meth:`~.misc.utils.TupleSA.set`
 of the `navigation_axes` and `signal_axes` attributes.
--- a/hyperspy/misc/utils.py
+++ b/hyperspy/misc/utils.py
@@ -1645,7 +1645,11 @@ class TupleSA(tuple):
     """A tuple that can set the attributes of its items
     """
     def __getitem__(self, *args, **kwargs):
-        return type(self)(super().__getitem__(*args, **kwargs))
+        item = super().__getitem__(*args, **kwargs)
+        try:
+            return type(self)(item)
+        except TypeError:
+            return item
 
     def __setattr__(self, name, value):
         no_name = [item
--- a/hyperspy/misc/utils.py
+++ b/hyperspy/misc/utils.py
@@ -1655,8 +1655,12 @@ class TupleSA(tuple):
             raise AttributeError(
                 f"'The items {no_name} have not attribute '{name}'")
         else:
-            for item in self:
-                setattr(item, name, value)
+            if isiterable(value) and not isinstance(value, str):
+                for item, value_ in zip(self, value):
+                    setattr(item, name, value_)
+            else:
+                for item in self:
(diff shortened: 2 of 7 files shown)


francisco-dlp:
This is WIP. I will complete it once I get feedback on the syntax and functionality. Some notes:
* Instead of ``__setattr__`` we could use a standard method e.g. ``set_attributes()``. That comes with the advantage of being more standard (the unusual usage of the ``__setattr__`` syntax may lead some to confusion) and enabling documentation through docstring. However, it is less convenient.
* This could also be useful for ``Model.components`` e.g.: 
   ```python
   m.components.select("GaussianHF").name = ("peak1", "peak2", "peak3")
   m.components.select(("peak1", "peak3")).fwhm = 2.6
   ```
  However, that would require an even less standard syntax: notice how the last line sets the ``fwhm.value`` attribute without specifying ``value``...

So, should convenience win over orthodoxy? If no, can you come up with an orthodox and equally convenient syntax?

ericpre:
This will be indeed very convenient and this would nicely replace the following usage (which I do quite regularly):
```python
for axis, name in zip(s.axes_manager.signal_axes, ("x", "y")):
   axis.name = name
```

If we want to keep it standard (not changing `__setattr__`), we can simply use the following syntax:
```python
s.axes_manager.navigation_axes[:].name = ("x", "y")
```

Regarding `m.components`, I agree that it would be very nice too. Maybe consider renaming `select` by `filter`. Not specifying `value` seems to be slightly off outside of the scope of this PR.

ericpre:
Since this PR will make more easy to set units and scale of the axis, I think that it would address https://github.com/hyperspy/hyperspy/issues/2237.

There is a related issue on the topic of indexing the `axes_manager`: https://github.com/hyperspy/hyperspy/issues/814

ericpre:
I am for overwriting `__setattr__` as currently implemented in this PR because of its convenience and that it outweigh strictly sticking to standards.

Combining this PR with https://github.com/hyperspy/hyperspy/pull/3385 would be a clear and very useful syntax to set the axis properties of all navigation or signals axes at once:

```python
s.axes_manager["sig"].scale = 0.1
```
I think that this is a no brainer, really!

@francisco-dlp, if you are still happy with this approach, I can finish it - it isn't much work.

francisco-dlp:
@CSSFrancis's syntax (see https://github.com/hyperspy/hyperspy/pull/3385#issuecomment-2185318534)  offers the advantages that it can be more easily documented in a docstring and is more compact when setting several attributes:

```python
s.axes_manager["nav"](scale=0.2, units="nm", offsets=0.2)
```

But the, why not the less magical:

```python
s.axes_manager["nav"].set_attributes(scale=0.2, units="nm", offsets=0.2)
```

What do you think?

ericpre:
I would be reluctant to have `s.axes_manager["nav"](scale=0.2, units="nm", offsets=0.2)`, because it would mean implementing `__call__` and I don't think that it fits well with what we would expect from `__call__`. A `set_attributes` method will be much better in that respect.

@CSSFrancis, @francisco-dlp, are you proposing this in addition to the implementation of `__setattr__` or in place?
I had a go at implementing the corresponding getter - it is slightly more complicated but still possible.

CSSFrancis:
Yea __call__ might be a step to far.

What about copying the syntax for the matplotlib collections object where the `set` function can be used to set each of the attributes. But you can also just use the dot notation for a simple attribute.

So something like:

```python

s.axes_manager.navigation_axes.set(scale=1, units="nm")

# and also for a single axis

s.axes_manager[0].set(scale=1, units="nm")

# or for the shorter version

s.axes_manger["nav"].set(scale=1, units="nm"]

# also should somethings like this be possible?

s.axes_manager[0:2].set(scale=1, units="nm")

s.axes_manager.navigation_axes.set(name=("x", "y"), units="nm", scale=(.1,1))


# or what about something like this be possible?

s.axes_manager[("x", "y")].scale=.1
```

In general we should think:

1. How should we be able to index axes: strings? slices? 
2. What object is returned. (Most likely a subclass of a Tuple)
3. What functions should the returned object have? Setting properties for a group sounds like a good start but maybe there are other functions we could add

ericpre:
+1 on `set`.
Adding `set`/`get` as matplotlib does to set artists attributes seems to be a very compromise:
- concise syntax (shorter than `set_attributes`), that allow to set and get several attributes
- have the advantage of `set_attributes` discussed above:
  - doesn't overwrite `__setattr__`
  - documentation to docstring

The `set`/`get` could also have optional argument that specify the type of container returned (for example, `dict` or `tuple`)? 

> 1. How should we be able to index axes: strings? slices?

This should be fairly covered, except that not everything is well documented

> 2. What object is returned. (Most likely a subclass of a Tuple)

Yes, the getter should most likely return tuple, or tuple of tuple, to cover the case when several attributes are requested

> * What functions should the returned object have? Setting properties for a group sounds like a good start but maybe there are other functions we could add

Not sure if I understand this one! :) The attributes will be returned in a container, which be a tuple, dict, etc. What do you mean with functions?

CSSFrancis:
We are going to need to make a new class to handle this: i.e. `TupleSA` in this PR.

The last point was what do we want this class to do besides just set attributes. For example:

```python
# current
s.axes_manager.signal_shape
# alternative 
s.axes_manager.signal_axes.shape
```

I'm not sure if we want to have the class do anything more than just set attributes but figured it should be at least discussed. I think there is an argument for both syntaxes which makes me lean towards having the class only set attributes and leave most functionality with the AxesManager class.

ericpre:
I think that it would be useful to have getter to mirror the behaviour of the setter while keeping this object behaving a tuple or a container.

Regarding `s.axes_manager.signal_shape`, I would say that this is good as it is currently and I don't see any benefit to move it to the `s.axes_manager.signal_axes` tuple/container - it doesn't need to know about its content, while the `AxesManager` needs to know about it.

francisco-dlp:
> also should somethings like this be possible?
> 
> s.axes_manager[0:2].set(scale=1, units="nm"

It could be implemented, but I can't think on an application for such a syntax. I think that using `signal_axes` and `navigation_axes` leads to clearer code.

If we all agree on this, I will refactor the code to add support for multiple keywords and replace `__setattr__` with `set`/`set_attributes`.

Personally, I prefer the more verbose `set_attributes` to `set`, because, again, it makes the code easier to read, but I am happy to go for `set` if there is more consensus around it.

CSSFrancis:
> > also should somethings like this be possible?
> > s.axes_manager[0:2].set(scale=1, units="nm"
> 
> It could be implemented, but I can't think on an application for such a syntax. I think that using `signal_axes` and `navigation_axes` leads to clearer code.

The use is for things like a 5D  STEM where you have a time axis and you might want to:

```python 
s.axes_manager[0:2].set(scale=1, units="nm")
s.axes_manager[2].set(scale=2, units="sec")
```
It just saves one line though :) 

> If we all agree on this, I will refactor the code to add support for multiple keywords and replace `__setattr__` with `set`/`set_attributes`.
> 
> Personally, I prefer the more verbose `set_attributes` to `set`, because, again, it makes the code easier to read, but I am happy to go for `set` if there is more consensus around it.

I'm usually for more verbose but in this case I think `set` is pretty well established.  From a purely asthetics standpoint `s.axes_manager.navigation_axes.set_attributes(scale=1.0, name=("x", "y", units="nm")` is getting fairly close to  88 characters. Also autocomplete is not super good when you are 3 levels down so less typing is nice. 

As much as I'm a fan of verbose naming in python I've sometimes wondered how much time something like `s.axes.nav.set(scale=1.0, units="nm")` would have saved me.

francisco-dlp:
Without adding slicing to `AxesManager` you could still do:

```python
s.axes_manager["nav"][0:2].set(scale=1, units="nm")
s.axes_manager["nav"][2].set(scale=2, units="sec")
```
The extra 7 characters make it clear that you are setting the attributes of the navigation axes, which might be worth it?

OK for `set` then.

ericpre:
I was also thinking that autocomplete rarely work well with slicing/indexing, hence the preference for something short.

> Without adding slicing to `AxesManager` you could still do:
> 
> ```python
> s.axes_manager["nav"][0:2].set(scale=1, units="nm")
> s.axes_manager["nav"][2].set(scale=2, units="sec")
> ```

This looks very good to me!

doc/user_guide/axes.rst
```diff
@@ -107,6 +107,17 @@ attributes, e.g.:
     >>> s.axes_manager[0]
     <X axis, size: 20, index: 0>
 
+.. versionadded:: 2.2
+    `set` method for `navigation_axes` and `signal_axes`.
```
ericpre:
```suggestion
    :meth:`~.misc.utils.TupleSA.set` method for :attr:`~.axes.AxesManager.navigation_axes` and :attr:`~.axes.AxesManager.signal_axes`.
```

doc/user_guide/axes.rst
```diff
@@ -107,6 +107,17 @@ attributes, e.g.:
...
+
+It is also possible to set multiple attributes of multiple axes at once, using the :meth:`~.misc.utils.TupleSA.set`
+of the `navigation_axes` and `signal_axes` attributes.
```
ericpre:
```suggestion
of the :attr:`~.axes.AxesManager.navigation_axes` and :attr:`~.axes.AxesManager.signal_axes` attributes.
```

doc/user_guide/axes.rst
```diff
@@ -119,6 +130,8 @@ name e.g.:
     >>> s.axes_manager["X"].units = "nm"
     >>> s.axes_manager["X"].offset = 100
 
+The
```
ericpre:
```suggestion
```

doc/user_guide/axes.rst
```diff
@@ -107,6 +107,17 @@ attributes, e.g.:
     >>> s.axes_manager[0]
     <X axis, size: 20, index: 0>
 
+.. versionadded:: 2.2
+    `set` method for `navigation_axes` and `signal_axes`.
+
+It is also possible to set multiple attributes of multiple axes at once, using the :meth:`~.misc.utils.TupleSA.set`
```
ericpre, line 114: `~.misc.utils.TupleSA.set` needs to be added to the API reference.

hyperspy/misc/utils.py
```diff
@@ -1639,3 +1639,43 @@ def display(obj):
...
+
+class TupleSA(tuple):
+    """A tuple that can set the attributes of its items
+    """
+    def __getitem__(self, *args, **kwargs):
+        item = super().__getitem__(*args, **kwargs)
+        try:
+            return type(self)(item)
+        except TypeError:
+            return item
```
ericpre, line 1653: Add a comment? Something along the line of:
```python
# for single item case, e.g. a[0]
# TupleSA requires an iterable item
```

ericpre (commented):
This looks good!

What's about adding a `get` method to mirror the `set` method. For example:
```python
>>> s.axes_manager.signal_axes.get("scale")
(1.0, 2.0)

>>> s.axes_manager.signal_axes.get("name", "scale")
(("x", "y"), (1.0, 2.0))
```

doc/user_guide/axes.rst
francisco-dlp: Done

hyperspy/misc/utils.py
francisco-dlp: Done

francisco-dlp:
Thanks @ericpre for the review and the suggestion to implement `get`. I have implemented a version that returns a dictionary instead of a list, so that it can be used in combination with `set` to copy attributes to other axes.

ericpre:
pre-commit.ci autofix

ericpre (approved):
I am looking forward to use it!

PR #2761, opened 2021-06-02
non-uniform updates (NotImplementedErrors, tests, spelling, saving-fix)

jlaehne:
### Description of the change
- Replace remaining mentions of 'non-linear' instead of 'non-uniform' axis
- Replace remaining misspellings of 'non uniform' instead of 'non-uniform' axis
- Add missing catches for non-uniform axis  few sentences and/or a bulleted list to describe and motivate the change:
- Implement `interpolate_in_between` for non-uniform axis with `delta` being `float`
- fix error in io-save-code introduced by some merge of RnM: https://github.com/hyperspy/hyperspy/pull/2399/files#r629224770

### Progress of the PR
- [x] Changes implemented (can be split into several points),
- [x] add tests,
- [x] ready for review.

### DIFF
--- a/hyperspy/io.py
+++ b/hyperspy/io.py
@@ -753,6 +753,7 @@ def save(filename, signal, overwrite=None, **kwds):
     # Check if the writer can write
     sd = signal.axes_manager.signal_dimension
     nd = signal.axes_manager.navigation_dimension
+    nua = signal.axes_manager.all_uniform
 
     if writer.writes is False:
         raise ValueError(
@@ -771,6 +772,14 @@ def save(filename, signal, overwrite=None, **kwds):
             f"Please try one of {strlist2enumeration(yes_we_can)}"
         )
 
+    if writer.non_uniform_axis is False and nua is False:
+        yes_we_can = [plugin.format_name for plugin in io_plugins
+                      if plugin.non_uniform_axis is True]
+        raise OSError("Writing to this format is not supported for "
+                      "non-uniform axes. Use one of the following "
+                      f"formats: {strlist2enumeration(yes_we_can)}"
+        )
+
     # Create the directory if it does not exist
     ensure_directory(filename.parent)
     is_file = filename.is_file()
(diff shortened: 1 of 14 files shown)


hyperspy/io.py
```diff
@@ -771,6 +772,14 @@ def save(filename, signal, overwrite=None, **kwds):
             f"Please try one of {strlist2enumeration(yes_we_can)}"
         )
 
+    if writer.non_uniform_axis is False and nua is False:
+        yes_we_can = [plugin.format_name for plugin in io_plugins
+                      if plugin.non_uniform_axis is True]
+        raise OSError("Writing to this format is not supported for "
```
ericpre: Would a `TypeError` be approapriate? See https://docs.python.org/3/library/exceptions.html#TypeError

hyperspy/io.py
```diff
@@ -771,6 +772,14 @@ def save(filename, signal, overwrite=None, **kwds):
             f"Please try one of {strlist2enumeration(yes_we_can)}"
         )
 
+    if writer.non_uniform_axis is False and nua is False:
+        yes_we_can = [plugin.format_name for plugin in io_plugins
```
ericpre: While you are at it: could you rename `yes_we_can` to `compatible_writers` or something similar, it mislead me during the review!

hyperspy/io.py
```diff
@@ -771,6 +772,14 @@ def save(filename, signal, overwrite=None, **kwds):
             f"Please try one of {strlist2enumeration(yes_we_can)}"
         )
 
+    if writer.non_uniform_axis is False and nua is False:
```
ericpre:
```suggestion
    if not writer.non_uniform_axis and not signal.axes_manager.all_uniform:
```
and remove `nua = signal.axes_manager.all_uniform` above to improve readability?

hyperspy/tests/signals/test_1D_tools.py
```diff
@@ -218,11 +218,26 @@ def test_single_spectrum(self):
         np.testing.assert_array_equal(s.data, np.arange(20))
         assert m.data_changed.called
 
+    def test_single_spectrum_nonuniform(self):
```
ericpre: Since the expected results are exactly the same, would it make sense to use `pytest.mark.parametrize('uniform', [True, False])` and add a `if` block in the test to convert to a non-uniform axis? Same for the other test?

ericpre (approved):
Looks good to me - I added suggestions for improvement to make it slightly more readable.

hyperspy/io.py
jlaehne: OR `IOError` like for  the "This file format does not support this data." exception just before? Would make sense to use the same kind of error for both I would say. What do you think?

hyperspy/io.py
jlaehne: Sure, though I found it funny ;-)

hyperspy/io.py
ericpre: Initially, I thought the same, but `IOError` is an alias for `OSError` and in this case, I prefer `TypeError` because we are trying to save a type of axes which is not supported, which I think is closer to the description in https://docs.python.org/3/library/exceptions.html#TypeError than https://docs.python.org/3/library/exceptions.html#OSError

> Changed in version 3.3: EnvironmentError, IOError, WindowsError, socket.error, select.error and mmap.error have been merged into OSError, and the constructor may return a subclass.

from https://docs.python.org/3/library/exceptions.html#OSError.filename2

hyperspy/io.py
ericpre: Indeed!

hyperspy/io.py
jlaehne: OK, but then I'll change the `IOError` above as well.

PR #2765, opened 2021-06-07
value2index: convert index to int

jlaehne:
### Description of the change
Follow up for #2743: Using `numba` for the `value2index` function can lead to a non-integer index that is not usable for slicing. Therefore, `int` is enforced and tested.

Fixes failing test from LumiSpy: https://github.com/LumiSpy/lumispy/pull/78#issuecomment-855796214

Along the way realized that rounding in `value2index` is not sufficiently tested for. Fixed bug for `rounding=math.floor`. However, also `rounding=round` was not working as it should: 0.5 was rounded down and not up.

### Progress of the PR
- [X] Fix rounding,
- [X] add tests for integer,
- [X] add tests for rounding,
- [X] ready for review.

### DIFF
--- a/hyperspy/axes.py
+++ b/hyperspy/axes.py
@@ -562,11 +562,16 @@ class BaseDataAxis(t.HasTraits):
         return value
 
     def value2index(self, value, rounding=round):
-        """Return the closest index to the given value if between the axis limits.
+        """Return the closest index/indices to the given value(s) if between the axis limits.
 
         Parameters
         ----------
         value : number or numpy array
+        rounding : function
+                Handling of values intermediate between two axis points:
+                If `rounding=round`, use round-half-away-from-zero strategy to find closest value.
+                If `rounding=math.floor`, round to the next lower value.
+                If `round=math.ceil`, round to the next higher value.
 
         Returns
         -------
@@ -1154,13 +1159,18 @@ class UniformDataAxis(BaseDataAxis, UnitConversion):
         return d
 
     def value2index(self, value, rounding=round):
-        """Return the closest index to the given value if between the axis limits.
+        """Return the closest index/indices to the given value(s) if between the axis limits.
 
         Parameters
         ----------
         value : number or string, or numpy array of number or string
                 if string, should either be a calibrated unit like "20nm"
                 or a relative slicing like "rel0.2".
+        rounding : function
+                Handling of values intermediate between two axis points:
+                If `rounding=round`, use python's standard round-half-to-even strategy to find closest value.
+                If `rounding=math.floor`, round to the next lower value.
+                If `round=math.ceil`, round to the next higher value.
 
         Returns
         -------
--- a/hyperspy/axes.py
+++ b/hyperspy/axes.py
@@ -594,17 +594,17 @@ class BaseDataAxis(t.HasTraits):
             if rounding is round:
                 #Use argmin(abs) which will return the closest value
                 # rounding_index = lambda x: np.abs(x).argmin()
-                index = numba_closest_index_round(self.axis,value)
+                index = numba_closest_index_round(self.axis,value).astype(int)
             elif rounding is math.ceil:
                 #Ceiling means finding index of the closest xi with xi - v >= 0
                 #we look for argmin of strictly non-negative part of self.axis-v.
                 #The trick is to replace strictly negative values with +np.inf
-                index = numba_closest_index_ceil(self.axis,value)
+                index = numba_closest_index_ceil(self.axis,value).astype(int)
             elif rounding is math.floor:
                 #flooring means finding index of the closest xi with xi - v <= 0
                 #we look for armgax of strictly non-positive part of self.axis-v.
                 #The trick is to replace strictly positive values with -np.inf
-                index = numba_closest_index_ceil(self.axis,value)
+                index = numba_closest_index_ceil(self.axis,value).astype(int)
             else:
                 raise ValueError(
                     f'Non-supported rounding function. Use '
--- a/hyperspy/axes.py
+++ b/hyperspy/axes.py
@@ -604,7 +604,7 @@ class BaseDataAxis(t.HasTraits):
                 #flooring means finding index of the closest xi with xi - v <= 0
                 #we look for armgax of strictly non-positive part of self.axis-v.
                 #The trick is to replace strictly positive values with -np.inf
-                index = numba_closest_index_ceil(self.axis,value).astype(int)
+                index = numba_closest_index_floor(self.axis,value).astype(int)
             else:
                 raise ValueError(
                     f'Non-supported rounding function. Use '
--- a/hyperspy/axes.py
+++ b/hyperspy/axes.py
@@ -583,8 +583,6 @@ class BaseDataAxis(t.HasTraits):
         else:
             value = np.asarray(value)
 
-
-
         #Should evaluate on both arrays and scalars. Raises error if there are
         #nan values in array
         if np.all((value >= self.low_value)*(value <= self.high_value)):
--- a/hyperspy/axes.py
+++ b/hyperspy/axes.py
@@ -32,6 +32,7 @@ from hyperspy.misc.array_tools import (
     numba_closest_index_floor,
     numba_closest_index_ceil,
     round_half_towards_zero,
+    round_half_away_from_zero,
 )
 from hyperspy.misc.utils import isiterable, ordinal
 from hyperspy.misc.math_tools import isfloat
@@ -1192,14 +1193,26 @@ class UniformDataAxis(BaseDataAxis, UnitConversion):
 
         value = self._parse_value(value)
 
-        if rounding is round:
-            rounding = round_half_towards_zero
-        elif rounding is math.ceil:
-            rounding = np.ceil
-        elif rounding is math.floor:
-            rounding = np.floor
+        multiplier = 1E12
+        index = 1 / multiplier * np.trunc(
+            (value - self.offset) / self.scale * multiplier
+            )
 
-        index = rounding((value - self.offset) / self.scale)
+        if rounding is round:
+            # When value are negative, we need to use half away from zero
+            # approach on the index, because the index is always positive
+            index = np.where(
+                value >= 0 if np.sign(self.scale) > 0 else value < 0,
+                round_half_towards_zero(index, decimals=0),
+                round_half_away_from_zero(index, decimals=0),
+                )
+        else:
+            if rounding is math.ceil:
+                rounding = np.ceil
+            elif rounding is math.floor:
+                rounding = np.floor
+
+            index = rounding(index)
 
         if isinstance(value, np.ndarray):
             index = index.astype(int)
--- a/hyperspy/axes.py
+++ b/hyperspy/axes.py
@@ -27,10 +27,12 @@ from traits.trait_errors import TraitError
 import pint
 from sympy.utilities.lambdify import lambdify
 from hyperspy.events import Events, Event
-from hyperspy.misc.array_tools import numba_closest_index_round, \
-                                        numba_closest_index_floor, \
-                                        numba_closest_index_ceil
-
+from hyperspy.misc.array_tools import (
+    numba_closest_index_round,
+    numba_closest_index_floor,
+    numba_closest_index_ceil,
+    round_half_towards_zero,
+)
 from hyperspy.misc.utils import isiterable, ordinal
 from hyperspy.misc.math_tools import isfloat
 from hyperspy.ui_registry import add_gui_method, get_gui
@@ -1190,13 +1192,12 @@ class UniformDataAxis(BaseDataAxis, UnitConversion):
 
         value = self._parse_value(value)
 
-        if isinstance(value, (np.ndarray, da.Array)):
-            if rounding is round:
-                rounding = np.round
-            elif rounding is math.ceil:
-                rounding = np.ceil
-            elif rounding is math.floor:
-                rounding = np.floor
+        if rounding is round:
+            rounding = round_half_towards_zero
+        elif rounding is math.ceil:
+            rounding = np.ceil
+        elif rounding is math.floor:
+            rounding = np.floor
 
         index = rounding((value - self.offset) / self.scale)
 
--- a/hyperspy/misc/array_tools.py
+++ b/hyperspy/misc/array_tools.py
@@ -452,7 +452,6 @@ def get_signal_chunk_slice(index, chunks):
                 return chunk_slice
     raise ValueError("Index out of signal range.")
 
-@njit(cache=True)
 def numba_closest_index_round(axis_array,value_array):
     """For each value in value_array, find the closest value in axis_array and
         return the result as a numpy array of the same shape as value_array.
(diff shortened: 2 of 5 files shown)


jlaehne:
The `test_value_changed_event` function revealed that actually rounding was not performed properly even prior to #2743.

@LMSC-NTappy could you review this PR and check the performance of the updated rounding function?

hyperspy/misc/array_tools.py
```diff
@@ -470,8 +470,8 @@ def numba_closest_index_round(axis_array,value_array):
     index_array = np.empty_like(value_array,dtype='uint')
     #assign on flat, iterate on flat.
     for i,v in enumerate(value_array.flat):
-        index_array.flat[i] = np.abs(axis_array - v).argmin()
-
+        vdiff_array = np.abs(axis_array - v)
+        index_array.flat[i] = np.flatnonzero(vdiff_array == np.min(vdiff_array))[-1]
```
LMSC-NTappy: Here the idea is to deal with the case where two identical values are found in `vdiff_array`  in which case you want to return the last one to ensure rounding above, I guess.

Unfortunately, while this might solve a failing test, this is not a solution for two reasons:
1. if the axis is sorted in decreasing order, this will still round below
2. the case where  `vdiff_array==np.min(vdiff_array)` matches more than 1 value does not typically happens. 

Here is an example:
```python
>>> import numpy as np
>>> ax = np.linspace(-10.0,-11.0,11)
>>> ax
array([-10. , -10.1, -10.2, -10.3, -10.4, -10.5, -10.6, -10.7, -10.8,
       -10.9, -11. ])
>>> vdiff_array = np.abs(ax+10.15)
>>> vdiff_array
array([0.15, 0.05, 0.05, 0.15, 0.25, 0.35, 0.45, 0.55, 0.65, 0.75, 0.85])
>>> vdiff_array==np.min(vdiff_array)
array([False, False,  True, False, False, False, False, False, False,
       False, False])
```
As you can see, this performs a rounding below (case 1) and we only have one single value matched in the equality comparison operator (case 2). Of course, the reason for this behaviour is float arithmetics. We have:

```python
>>> vdiff_array[1] 
0.05000000000000071
>>> vdiff_array[2]
0.049999999999998934
```

which, no matter how you look at it, will still round below in most strategies, I think the best way to deal with it would be to bias very slightly vdiff_array towards upper-rounding. I would advocate something like `np.abs(axis_array - v + MACHINEEPSILON)` and `MACHINEEPSILON` would be determined using some relative tolerance, eg `rtol=1e-10; MACHINEEPSILON = np.min(np.diff(axis))*rtol `.

I don't pretend this is the ideal solution. But it can shifts a commonly encountered problem (incorrect value2indexing exactly half the interval between axis points) to a much less commonly encountered one (incorrect value2indexing a value that is very close but not equal to exactly half the interval between axis points)

hyperspy/axes.py
```diff
@@ -594,17 +594,17 @@ def value2index(self, value, rounding=round):
...
                 #we look for argmin of strictly non-negative part of self.axis-v.
                 #The trick is to replace strictly negative values with +np.inf
-                index = numba_closest_index_ceil(self.axis,value)
+                index = numba_closest_index_ceil(self.axis,value).astype(int)
             elif rounding is math.floor:
                 #flooring means finding index of the closest xi with xi - v <= 0
                 #we look for armgax of strictly non-positive part of self.axis-v.
                 #The trick is to replace strictly positive values with -np.inf
-                index = numba_closest_index_ceil(self.axis,value)
+                index = numba_closest_index_floor(self.axis,value).astype(int)
```
LMSC-NTappy, line 613: Looks Good To Me.

However, this behaviour is weird because in the `numba_closest_index` functions, the returned array is initialized with `"uint"` dtype and argmax / argmin functions should not return anything else than `int` types. I wonder what causes this behaviour... 

Well, if it works like this then who am I to judge?

LMSC-NTappy:
@jlaehne I took a look, see comments above.

I'll check the execution speed, however I'd like to finish discussing this rounding behaviour issue.

Cheers

Nicolas

hyperspy/axes.py
ericpre: I agree, this is odd, there should be a bug somewhere and from a quick look, it should not return float type and casting explicitly to `int` after calling jitted function shouldn't be necessary...

jlaehne: I thought so as well.

However, for some reason a statement such as `s.isig[value2index(x)+1:]` would throw an error for `value2index(x)+1` not being an `int` that can be used as index. It is explicitly for when another integer is added/subtracted to the result of `value2index` It is the source of the failure of the tests in https://github.com/LumiSpy/lumispy/pull/78#issuecomment-855796214
As the failure appeared only recently and only for the `non_uniform_axes` branch it must be a result of rewriting the `value2index` function.

Adding the `.astype(int)` fixes this problem. (Note that the `value2index` function for `UniformDataAxis` also explicitly casts to `int`.)  Therefore, I propose to go for this "workaround".

hyperspy/misc/array_tools.py
jlaehne: Yes, the idea was to enforce rounding up for a `0.5` value.

I like your idea with the upwards bias, I had thought in a similar direction. I'll implement that to see.

Though, I now realize that numpy also does not necessarily round 0.5 upwards, but it rounds to the next even number: https://github.com/numpy/numpy/blob/e94ed84010c60961f82860d146681d3fd607de4e/numpy/core/fromnumeric.py#L2723-L2789 
(explicitly mentioning the issue with machine precision as well).

hyperspy/axes.py
LMSC-NTappy: >  Therefore, I propose to go for this "workaround".

I share this opinion

hyperspy/misc/array_tools.py
```diff
@@ -470,8 +470,9 @@ def numba_closest_index_round(axis_array,value_array):
     index_array = np.empty_like(value_array,dtype='uint')
     #assign on flat, iterate on flat.
     for i,v in enumerate(value_array.flat):
-        vdiff_array = np.abs(axis_array - v)
-        index_array.flat[i] = np.flatnonzero(vdiff_array == np.min(vdiff_array))[-1]
+        rtol=1e-10
+        machineepsilon = np.min(np.diff(axis_array))*rtol
+        index_array.flat[i] = np.abs(axis_array - v + machineepsilon).argmin()
```
LMSC-NTappy: Idk if that's why the tests fail, but probably you need to replace:  `machineepsilon = np.min(np.abs(np.diff(axis_array)))*rtol` 

Also, I'd suggest you put this outside for loop for efficiency. this does not depend on value and need to be computed only once

hyperspy/misc/array_tools.py
LMSC-NTappy: Yeah exactly. 

Actually I wonder why this rounding approach is used in hyperspy. When slicing with a float value, I would personally prefer knowing in advance whether the sliced axis will contain or not the value I input.

I would agree it's not a big deal, but in the end we spend some effort to preserve this functionality.

hyperspy/axes.py
```diff
@@ -594,17 +594,17 @@ def value2index(self, value, rounding=round):
...
                 #Use argmin(abs) which will return the closest value
                 # rounding_index = lambda x: np.abs(x).argmin()
-                index = numba_closest_index_round(self.axis,value)
+                index = numba_closest_index_round(self.axis,value).astype(int)
             elif rounding is math.ceil:
                 #Ceiling means finding index of the closest xi with xi - v >= 0
                 #we look for argmin of strictly non-negative part of self.axis-v.
                 #The trick is to replace strictly negative values with +np.inf
-                index = numba_closest_index_ceil(self.axis,value)
+                index = numba_closest_index_ceil(self.axis,value).astype(int)
```
ericpre, line 608: Try to add `# pragma: no cover` next to the definition of the jitted functions to avoid this warning, see https://github.com/hyperspy/hyperspy/pull/2520 for example.

hyperspy/misc/array_tools.py
jlaehne: It is python rounding which uses the round-to-even strategy to avoid accumulation of offset: https://realpython.com/python-rounding/#pythons-built-in-round-function

hyperspy/misc/array_tools.py
jlaehne: `macineepsilon` has to be of course added to `v` and not to `axis_array`. so `- machineepsilon`. I also differentiate now between positive and negative values to round away from zero (otherwise we would have rounded up `-0.5`).

hyperspy/misc/array_tools.py
jlaehne: To be consistent with the behaviour for non-uniform axes, I have used a machineepsilon here as well to have 'round-away-from-zero' instead of 'round-to-even'

jlaehne: Well, in the end I think it does not make sense to touch this function,because round is of course used elsewhere in the code as well and we should be consistent there.

LMSC-NTappy:
In any case, this is not noticeably worse than the previous implementation.

image

Once the rounding behavior is settled this can be integrated IMO

jlaehne:
Status is that for uniform axis `value2index` uses round-half-to-even and for non-uniform axis, round-half-away-from-zero.

Are we OK with that difference? Then I would  just add the difference to the docstring.

Otherwise, I would have to redo the non-uniform axis behavior again to also perform a round-half-to-even. Instead of the biasing away from zero that would mean that I would have to revert to checking whether `axis_array - v` has two values of same distance to zero and if so check which one of them would result in an even value.

hyperspy/misc/array_tools.py
LMSC-NTappy: > To be consistent with the behaviour for non-uniform axes, I have used a machineepsilon here as well to have 'round-away-from-zero' instead of 'round-to-even'

Not sure what behaviour you mean. You want a behaviour similar to numpy.rint? https://numpy.org/doc/stable/reference/generated/numpy.rint.html#numpy.rint 

In any case, I quickly checked out your branche and I think there still is an issue of uniformity between the Uniform and Non Uniform Axes

```python
>>>from hyperspy.axes import UniformDataAxis

>>>inax = [[-11.0,-10.9],[-10.9,-11.0],[+10.9,+11.0],[+11.0,+10.9]]
>>>inval = [-10.95,-10.95,10.95,10.95]

>>>for i,j in zip(inax,inval):
>>>    ax = UniformDataAxis(scale=i[1]-i[0],offset=i[0],size=len(i))
>>>    nua_idx = super(type(ax),ax).value2index(j,rounding=round)
>>>    unif_idx = ax.value2index(j,rounding=round)
>>>    print("Value2Index IN ax: {}, val: {} --> UAX: {}, NUA: {}".format(ax.axis,j,unif_idx,nua_idx))

Value2Index IN ax: [-11.  -10.9], val: -10.95 --> UAX: 1, NUA: 0
Value2Index IN ax: [-10.9 -11. ], val: -10.95 --> UAX: 0, NUA: 1
Value2Index IN ax: [10.9 11. ], val: 10.95 --> UAX: 0, NUA: 1
Value2Index IN ax: [11.  10.9], val: 10.95 --> UAX: 1, NUA: 0

```

I assume the behaviour you desire is the one from NUA?

LMSC-NTappy (approved):
Sorry I didn't see your reply.

Yes I am fine with that. I didn't see your comment soon enough. If you think it doesn't matter that both behave differently that's ok for me.

ericpre:
> Status is that for uniform axis `value2index` uses round-half-to-even and for non-uniform axis, round-half-away-from-zero.
> 
> Are we OK with that difference? Then I would just add the difference to the docstring.
> 
> Otherwise, I would have to redo the non-uniform axis behavior again to also perform a round-half-to-even. Instead of the biasing away from zero that would mean that I would have to revert to checking whether `axis_array - v` has two values of same distance to zero and if so check which one of them would result in an even value.

Humm, it is not great if `value2index` doesn't give the same results for different type of axis even if the axis are exactly the same... I don't think that the rounding really matters but it should be done consistently! It may be good to get sorted now, while we know what is the issue and what needs to be done.

jlaehne:
OK, I'll give it a try the next days.

jlaehne:
> Humm, it is not great if `value2index` doesn't give the same results for different type of axis even if the axis are exactly the same... I don't think that the rounding really matters but it should be done consistently! It may be good to get sorted now, while we know what is the issue and what needs to be done.

Indeed, `value2index` should behave the same for any type of axis.

However, the only way I could come up with to implement the `round-to-even` strategy for non uniform axes is by deactivating `numba` for that function. If I have to decide for example whether `10.1` or `10.2` has an even last digit, I can check the last character of `repr(x)`. As a float itself does not hold any information on how many digits are relevant, I have not found any pure numba way of checking that.

LMSC-NTappy:
> the only way I could come up with to implement the round-to-even strategy for non uniform axes is by deactivating numba for that function

 A consistent behavior is important but I prefer a solution that does not require sacrificing the efficiency of `BaseDataAxis.value2index` to deal with this specific case. Besides, `UniformDataAxis.value2index` does not perform the round-to-even strategy:

```python
>>>from hyperspy.axes import UniformDataAxis
>>>dd = UniformDataAxis(size=10, scale=0.1, offset=10);
>>>test_vals = dd.axis[:-1:]+0.05
>>>print(dd.axis)
[10.  10.1 10.2 10.3 10.4 10.5 10.6 10.7 10.8 10.9]
>>>print(test_vals)
[10.05 10.15 10.25 10.35 10.45 10.55 10.65 10.75 10.85]
>>>dd.value2index(test_vals,rounding=round)
array([ 1,  2,  2,  4,  5,  6,  7,  8,  9])
```

In the end, I am afraid we are chasing a chimaera. Because round is evaluated on the result of floating-point operations. In `UniformDataAxis.value2index`:
https://github.com/hyperspy/hyperspy/blob/2ddc359c5a6c0ff300507c383dfcc1de644ff11a/hyperspy/axes.py#L569

Which will, I think, hardly ever map to the exact midpoint between two integers indices.

I propose to implement the same biasing on `UniformDataAxis.value2index` and not care about the round-to-even policy. IEEE 754 recommends the round-to-even operation only to remove bias from addition of many rounded values. I think we can all agree that this absolutely won't be an issue here. Or am I missing something?

ericpre:
Yes, I agree that the method used is not very critical here and we should choice a method, which is efficient, simple and consistent between different type of axes. If it requires to slightly change the behaviour of `UniformDataAxis.value2index`, this sounds like a good compromise to me.

> I propose to implement the same biasing on `UniformDataAxis.value2index` and not care about the round-to-even policy

@LMSC-NTappy, what do you mean? Sorry if you are referencing to something above that I may have missed!

jlaehne:
OK, I checked only the case for 10.15 and 10.25 - and from that thought it uses the numpy routine - good that you checked the full array. So I'll just revert the last commit and then we should go for that solution instead of chasing further.

ericpre:
In https://github.com/jlaehne/hyperspy/pull/7, I have changed the rounding behaviour of both non-uniform and uniform axis to "round half towards zero". The different type of axis should behave consistently now!

jlaehne:
@ericpre, when I use the example of @LMSC-NTappy from above that checks a complete 10-value axis, I still get a disagreement:

```
>>> from hyperspy.axes import UniformDataAxis
>>> dd = UniformDataAxis(size=10, scale=0.1, offset=10);
>>> test_vals = dd.axis[:-1:]+0.05
>>> print(dd.axis)
[10.  10.1 10.2 10.3 10.4 10.5 10.6 10.7 10.8 10.9]
>>> print(test_vals)
[10.05 10.15 10.25 10.35 10.45 10.55 10.65 10.75 10.85]
>>> dd.value2index(test_vals,rounding=round)
array([1, 2, 2, 4, 5, 6, 7, 7, 9])
>>> super(type(dd),dd).value2index(test_vals,rounding=round)
array([0, 1, 2, 3, 4, 5, 6, 7, 8])
```

What was the motivation for using round-half-to-zero instead of -away-from-zero?

ericpre:
@jlaehne, @LMSC-NTappy: the inconsistency should between uniform and non-uniform axes should be fixed in 0659013. Could you please review it?

> What was the motivation for using round-half-to-zero instead of -away-from-zero?

I was thinking that it will help to avoid out of axis error.

jlaehne:
nice, LGTM! You covered it all by tests, but also the locally run test cases work consistently now.

jlaehne:
I think we can merge this one now.

LMSC-NTappy:
Looks good to me! 

I tested the speed and it's as before. In fact the speed of UniformDataAxis.value2index() has doubled!

Congrats, ready to merge IMO! 

image

ericpre:
Thanks! It was tedious but I think it ended up in a good place!

Regarding the choice of round-half-to-zero instead of round-half-away-from-zero, it should be very easy to change, if necessary.

Issue #2766, opened 2021-06-08
assign_subclass and signal with one value?


magnunor:
In https://github.com/hyperspy/hyperspy/pull/2703, one of the tests is failing due to how `s._assign_subclass` works.

Essentially, should a signal with one value be a `BaseSignal`, or a `Signal1D`? Currently, `_assign_subclass` turns a signal with one value into a `Signal1D`, but the test expects it to be a `BaseSignal`.

To replicate this:

```python
s = hs.signals.BaseSignal(1) # <BaseSignal, title: , dimensions: (|1)>
s._assign_subclass() # <Signal1D, title: , dimensions: (|1)>
```
The question is then which signal type _should_ a signal with one value be? `BaseSignal` or `Signal1D`?

jlaehne:
I would have thought that `BaseSignal` is just a parent class for all other signal classes, but is normally not used. `Signal1D` provides many additional routines, but then most of them don't make sense on a signal with one value only. So it seems to me to indeed be an edge case.

thomasaarholt:
If the array is a zero-dimensional array, then I'd say it should probably go to basesignal or a "Signal0D".

francisco-dlp:
HyperSpy does not support array scalars (i.e. 0-dimensional arrays) because scalar arrays, unlike ndarrays, are immutable. That's why HyperSpy converts any array scalar in a 1D array of shape `[1]`.

This is indeed a corner case. Considering that, as pointed out by @jlaehne, most ``Signal1D`` methods make no sense for 1-element arrays, I think that it'll be better to assign them to ``BaseSignal``.

ericpre:
Done in #2773.

PR #2773, opened 2021-06-18
_assign_subclass: signal_shape (1,) now becomes BaseSignal

magnunor:
As discussed in https://github.com/hyperspy/hyperspy/issues/2766, currently a signal with signal dimensions `(1,)` is assigned to be a `Signal1D`. This is not optimal, as most methods which apply to `Signal1D` probably won't work very well if such 1-value signals.

So this pull request changes how `_assign_subclass` works, by now making signals with signal dimension of `(1,)` `BaseSignal` instead of `Signal1D`.

### Progress of the PR
- [x] Change implemented
- [x] add tests
- [x] ready for review

### Minimal example
```python
import hyperspy.api as hs
import numpy as np
s = hs.signals.Signal1D(1) # <Signal1D, title: , dimensions: (|1)>
s._assign_subclass() # <BaseSignal, title: , dimensions: (|1)>
```


ericpre (commented):
The signal_dimension is not set correctly, for example:
```python
import hyperspy.api as hs

s = hs.signals.Signal1D(1)
s._assign_subclass()
print(s.__class__.__name__, 'with signal dimension:', s.axes_manager.signal_dimension)
```

`s` has a signal dimension of 1, while it should be zero.

magnunor:
When first making this pull request, I wasn't really sure which avenue to take. Either via changing how the `AxesManager` works, or by changing how `_assign_subclass` works. 

The original implementation changed how `_assign_subclass` worked, which I think is the less "risky" way of doing it. Since it much less changes the functionality. I.e. less odds of something breaking in subtle ways. However, the downside is that (some) base signals would still report having 1 signal dimension.

The new implementation, changing how the `AxesManager` object sets the `signal_dimensions`, has way larger effects. Possible unintended.

In fact, the "new" implementation should really go way deeper, since now it only changes how `signal_dimension` is handled. While ignoring `axes_manager.signal_axes`.

I'm not really sure which way to go with this, since I feel that disallowing `BaseSignal` with signal axes is potentially a lot of work.

ericpre:
Previous discussion on this topic: #643.

> The original implementation changed how `_assign_subclass` worked, which I think is the less "risky" way of doing it. Since it much less changes the functionality. I.e. less odds of something breaking in subtle ways. However, the downside is that (some) base signals would still report having 1 signal dimension.

In which case, things requiring `signal_dimension == 1` may be broken because the `signal_dimension` value will be incorrect.

> In fact, the "new" implementation should really go way deeper, since now it only changes how `signal_dimension` is handled. While ignoring `axes_manager.signal_axes`.

What do you mean with ignoring `axes_manager.signal_axes`?

> I'm not really sure which way to go with this, since I feel that disallowing `BaseSignal` with signal axes is potentially a lot of work.

I don't understand why `BaseSignal` with signal axes would be disallowed?

hyperspy/axes.py
```diff
@@ -1203,7 +1203,14 @@ def _update_attributes(self):
...
         self._getitem_tuple = tuple(getitem_tuple)
-        self.signal_dimension = len(self.signal_axes)
+        if len(self.signal_axes) == 1:
+            if self.signal_axes[0].size == 1:
+                signal_dimension = 0
+            else:
+                signal_dimension = len(self.signal_axes)
+        else:
+            signal_dimension = len(self.signal_axes)
+        self.signal_dimension = signal_dimension
```
jlaehne: Could be simplified to
```
        if len(self.signal_axes) == 1 and self.signal_axes[0].size == 1:
            self.signal_dimension = 0
        else:
            self.signal_dimension = len(self.signal_axes)
```

magnunor:
These signals will now return `signal_dimension=0`. However, I'm not sure how best to handle some of the functionality which breaks. For example:

```python
import numpy as np
import hyperspy.api as hs
s = hs.signals.Signal1D([1])
s1 = hs.signals.Signal1D([2])
hs.stack([s, s1])
```
Gives the error:

```python
hyperspy/misc/utils.py in stack(signal_list, axis, new_axis_name, lazy, stack_metadata, show_progressbar, **kwargs)
   1022     if len(signal_list) > 1:
   1023         # Matching axis calibration is checked here
-> 1024         newlist = broadcast_signals(*signal_list, ignore_axis=axis_input)
   1025 
   1026         if axis is not None:

hyperspy/misc/signal_tools.py in broadcast_signals(ignore_axis, *args)
    188                     data = np.broadcast_to(data, thisshape)
    189                 else:
--> 190                     data = da.broadcast_to(data, thisshape)
    191 
    192             ns = s._deepcopy_with_new_data(data)

dask/array/core.py in broadcast_to(x, shape, chunks)
   4349         new != old for new, old in zip(shape[ndim_new:], x.shape) if old != 1
   4350     ):
-> 4351         raise ValueError("cannot broadcast shape %s to shape %s" % (x.shape, shape))
   4352 
   4353     if chunks is None:

ValueError: cannot broadcast shape (1, 1) to shape (1,)
```

It also breaks the following test:

```python
import numpy as np
import hyperspy.api as hs
s32 = hs.signals.Signal1D(np.ones((2, 3)))
s12 = hs.signals.Signal1D(np.ones((2, 1)))
print(s32.axes_manager.signal_dimension) # 1 signal dim
print(s12.axes_manager.signal_dimension) # 0 signal dim
s32 + s12
```
Gives the error 

```python
002_test_script.py in <module>
      5 print(s32.axes_manager.signal_dimension)
      6 print(s12.axes_manager.signal_dimension)
----> 7 s32 + s12

hyperspy/signal.py in __add__(self, other)

hyperspy/signal.py in _binary_operator_ruler(self, other, op_name)
   2253                 else:
   2254                     # They are broadcastable but have different number of axes
-> 2255                     ns, no = broadcast_signals(self, other)
   2256                     sdata = ns.data
   2257                     odata = no.data

hyperspy/misc/signal_tools.py in broadcast_signals(ignore_axis, *args)
    186             if data.shape != thisshape:
    187                 if isinstance(data, np.ndarray):
--> 188                     data = np.broadcast_to(data, thisshape)
    189                 else:
    190                     data = da.broadcast_to(data, thisshape)

<__array_function__ internals> in broadcast_to(*args, **kwargs)

numpy/lib/stride_tricks.py in broadcast_to(array, shape, subok)
    178            [1, 2, 3]])
    179     """
--> 180     return _broadcast_to(array, shape, subok=subok, readonly=True)
    181 
    182 

numpy/lib/stride_tricks.py in _broadcast_to(array, shape, subok, readonly)
    121                          'negative')
    122     extras = []
--> 123     it = np.nditer(
    124         (array,), flags=['multi_index', 'refs_ok', 'zerosize_ok'] + extras,
    125         op_flags=['readonly'], itershape=shape, order='C')

ValueError: input operand has more dimensions than allowed by the axis remapping
```

Which happens in https://github.com/hyperspy/hyperspy/blob/RELEASE_next_minor/hyperspy/misc/signal_tools.py#L190

One solution could be to add `if` `else` in `hyperspy/misc/signal_tools/broadcast_signals` for this specific situation, ergo to check if the signal dimension is 0, then if the data size is 1, then put signal dimension to 1 inside the functions. But not sure if doing that is the best solution: that function is a bit difficult to follow, so I'm a bit hesitant in touching it.

ericpre:
@magnunor, I made a PR to fix the issues of this PR, please have a look.

magnunor:
@ericpre, that seems to have fixed the issue!

`Build doc` is failing, but I think that is unrelated.

Ready for review!

ericpre (approved):

PR #2775, opened 2021-06-23
added ValueError for unknown densities

### LINKED ISSUE #2771: Error : Minor issue with density_of_mixture
I encountered a minor issue using the `density_of_mixture` function from `hyperspy.misc.material`. My version of hyperspy is 1.6.2 and here is the code to reproduce the error : 

```
import hyperspy.misc.material as m
m.density_of_mixture([1.0],elements=["At])
```
It yields a `TypeError : unsupported operand type(s) for /: 'float' and 'numpy.str_'`

I could not reproduce this error for other elements (I did not test them all). I guess there is a small mistake in the tables, probably only for Astate.

adriente:
### Description of the change

Some entries of the elements database have unknown densities due to their very high radioactivity : Astate and Francium. When calling the function `_density_of_mixture` from `hyperspy.misc.material` with these elements it would raise a `TypeError` with a difficult to understand message.
With this PR, the function raises a `ValueError` with a more understandable error message.

### Progress of the PR
- [x] Change implemented 
- [x] ready for review.
- [x] make a test.
- [x] make a changelog entry.

```python
import hyperspy.misc.material as m
m._density_of_mixture([1.0],["At"])
```
#### New output
```
  File "/home/teurtrie/hyperspy/hyperspy/misc/material.py", line 224, in _density_of_mixture
    sum_densities[i] = weight / densities[i]
TypeError: unsupported operand type(s) for /: 'float' and 'numpy.str_'

  File "<stdin>", line 1, in <module>
  File "/home/teurtrie/hyperspy/hyperspy/misc/material.py", line 236, in _density_of_mixture
    raise ValueError(
ValueError: The density of one of the elements is unknown (Probably At or Fr).
```

### DIFF
--- a/hyperspy/misc/material.py
+++ b/hyperspy/misc/material.py
@@ -218,19 +218,23 @@ def _density_of_mixture(weight_percent,
         [elements_db[element]['Physical_properties']['density (g/cm^3)']
             for element in elements])
     sum_densities = np.zeros_like(weight_percent, dtype='float')
-    if mean == 'harmonic':
-        for i, weight in enumerate(weight_percent):
-            sum_densities[i] = weight / densities[i]
-        sum_densities = sum_densities.sum(axis=0)
-        density = np.sum(weight_percent, axis=0) / sum_densities
-        return np.where(sum_densities == 0.0, 0.0, density)
-    elif mean == 'weighted':
-        for i, weight in enumerate(weight_percent):
-            sum_densities[i] = weight * densities[i]
-        sum_densities = sum_densities.sum(axis=0)
-        sum_weight = np.sum(weight_percent, axis=0)
-        density = sum_densities / sum_weight
-        return np.where(sum_weight == 0.0, 0.0, density)
+    try : 
+        if mean == 'harmonic':
+            for i, weight in enumerate(weight_percent):
+                sum_densities[i] = weight / densities[i]
+            sum_densities = sum_densities.sum(axis=0)
+            density = np.sum(weight_percent, axis=0) / sum_densities
+            return np.where(sum_densities == 0.0, 0.0, density)
+        elif mean == 'weighted':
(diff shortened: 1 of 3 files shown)


hyperspy/misc/material.py
```diff
@@ -218,19 +218,23 @@ def _density_of_mixture(weight_percent,
...
+            return np.where(sum_densities == 0.0, 0.0, density)
+        elif mean == 'weighted':
+            for i, weight in enumerate(weight_percent):
+                sum_densities[i] = weight * densities[i]
+            sum_densities = sum_densities.sum(axis=0)
+            sum_weight = np.sum(weight_percent, axis=0)
+            density = sum_densities / sum_weight
+            return np.where(sum_weight == 0.0, 0.0, density)
+    except TypeError : 
+        raise ValueError(
```
ericpre, line 236: A test is needed to cover this case, you can simply add your example from #2775 in https://github.com/hyperspy/hyperspy/blob/2ddc359c5a6c0ff300507c383dfcc1de644ff11a/hyperspy/tests/utils/test_material.py#L58.

ericpre (commented):
This looks good, it needs a test and to make the PR to the `RELEASE_next_patch` branch because it is a bug fix.
To avoid pulling change from the `RELEASE_next_minor` branch, the base branch will need to changed, see https://hyperspy.readthedocs.io/en/stable/dev_guide/git.html#changing-base-branch for explanation. If you prefer, I can sort it out before merging.

ericpre (commented):
It needs a changelog entry  in the `upcoming_changes` folder too - see [`upcoming_changes/README.rst`](https://github.com/hyperspy/hyperspy/blob/RELEASE_next_minor/upcoming_changes/README.rst)!

ericpre (approved):
Thanks, looks good. A few commits are unnecessary and I will remove them before merging.

Issue #2774, opened 2021-06-22
BLO output writer not working for lazy signals


din14970:
It seems that the BLO writer does not work for lazy signals. The error seems to be in the line containing `tofile`. This is problematic for converting larger 4DSTEM/NBED datasets to BLO. I'll try to fix this. 
In addition, the writer is very particular about units and does not seem to accept angstroms.

hakonanes:
FYI, the writer was written by Vidar, @vidartf.

Yeah, I guess you have to write each pattern to memory in the loop
https://github.com/hyperspy/hyperspy/blob/2ddc359c5a6c0ff300507c383dfcc1de644ff11a/hyperspy/io_plugins/blockfile.py#L290-L293

ericpre:
Can you please provide the traceback, in #2777, I can't see what changes would fix saving lazy signal.

din14970:
The issue is that dask arrays don't have th`tofile` method that `np` has.

Full tracebacks for reference (on `RELEASE_next_patch` `bd1c12b74`):

When one does not run into unit conversion issues (e.g. set signal units to `nm`):
```
---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
<ipython-input-11-ebacc5399fa7> in <module>
----> 1 data_file.save("test.blo")

~/Documents/PythonProjects/FORKS/hyperspy/hyperspy/signal.py in save(self, filename, overwrite, extension, **kwds)
   2838         if extension is not None:
   2839             filename = filename.with_suffix(f".{extension}")
-> 2840         io.save(filename, self, overwrite=overwrite, **kwds)
   2841 
   2842     def _replot(self):

~/Documents/PythonProjects/FORKS/hyperspy/hyperspy/io.py in save(filename, signal, overwrite, **kwds)
    781         # Pass as a string for now, pathlib.Path not
    782         # properly supported in io_plugins
--> 783         writer.file_writer(str(filename), signal, **kwds)
    784 
    785         _logger.info(f'{filename} was created')

~/Documents/PythonProjects/FORKS/hyperspy/hyperspy/io_plugins/blockfile.py in file_writer(filename, signal, **kwds)
    274             endianess +
    275             'u1')
--> 276         vbf.tofile(f)
    277         # Zero pad until next data block
    278         if f.tell() > int(header['Data_offset_2']):

AttributeError: 'Array' object has no attribute 'tofile'
```

Dimension issue (signal unit is `1/nm`):

```
[...]

Issue #2639, opened 2021-02-05
Improved handling of navigator in non-lazy signals


magnunor:
With the improved handling of navigator generation for lazy signals in https://github.com/hyperspy/hyperspy/pull/2631, it would be nice to extend parts of this to non-lazy signals as well.

Currently, when `s.plot()` is used for non-lazy signals, the navigator signal is discarded after the plot is closed.

For very large datasets, the creation of the navigator signal can take a couple of seconds. So it would be nice if the navigator signal for non-lazy signals was stored, similarly as non-lazy signals.

CSSFrancis:
I've been meaning to implement this.

Curious on peoples thoughts of expanding this to changing the navigator to a set of virtual images. Something like:


```python
s.metadata._Hyperspy.navigators = [<Signal2D, title:VDF , dimensions: (|128, 128)>, <Signal2D, title:VBF , dimensions: (|128, 128)>, ...]
```

Then maybe you could cycle through them by pressing some key etc or plot a specific virtual image or use the new subfigure feature to plot "all" of the virtual images.

```python
s.plot(navigator = "VDF")
s.plot(navigator = "all")
```

ericpre:
This makes sense, for EDS or EELS, it is very common to have more than one signal acquired simultaneously.
Another option is to have the navigators display on separates figures / subfigures with the pointer added on all of them. Being able to cycle through them will be useful too to save on number of figure and will be more simple for adding navigation ROI.

PR #2777, opened 2021-06-23
Fix blo writing for lazy datasets + other improvements

din14970:
### Description of the change
As discussed in #2774 the BLO writer doesn't work for lazy datasets and throws errors when the diffraction patterns have weird units like 1/angstrom. This PR:

- Fixes the option to write lazy datasets
- Puts the conversion of units into a `try` block and warns the users if unit conversion was unsuccessful

In addition, since 4D-STEM data may be collected at dtypes > uint8, the default writer can easily cause overflow errors. I added an extra kwarg to the writer that allows the images to be rescaled to an appropriate range before conversion to 8-bit. A number of ways to do this are possible, the options I included are:

- `None` (default): current behavior, no rescaling and allow overflow. However, I added a warning in case the dtype is not uint8 that users may want to use the `intensity_scaling` argument in `save`.
- `'dtype'`: use the min and max value of the signal dtype range. Nonsensical for floats so throws an error.
- `'minmax'`: calculate the minimum and maximum of the dataset and map those onto 0 and 255 resp.
- `'crop'`: everything below 0 and above 255 is set to 0 and 255 resp.
- a 2-tuple of floats/ints: these intensity values are scaled linearly to 0 and 255

In addition, since the process in the loop can be quite slow for large datasets I added a tqdm progressbar.

Still needs more testing and possibly user guide additions, just posting this up here in case someone already wants to have a look.

### Progress of the PR
- [x] add tests,
- [x] ready for review.

```python
import hyperspy.api as hs
import numpy as np
s = hs.signals.Signal2D(np.arange(25).reshape(1, 1, 5, 5).astype(np.float64)*1000, lazy=True)
s.save("test.blo", intensity_scaling="minmax")
```

### DIFF
--- a/doc/user_guide/io.rst
+++ b/doc/user_guide/io.rst
@@ -813,6 +813,14 @@ Extra saving arguments
   - `'crop'`: everything below 0 and above 255 is set to 0 and 255 respectively
   - 2-tuple of `floats` or `ints`: the intensities between these values are
     scaled between 0-255, everything below is 0 and everything above is 255.
+- ``navigator_signal``: the BLO file also stores a virtual bright field (VBF) image which
+  behaves like a navigation signal in the ASTAR software. By default this is
+  set to `'navigator'`, which results in the default :py:attr:`navigator` signal to be used.
+  If this signal was not calculated before (e.g. by calling :py:meth:`~.signal.BaseSignal.plot`), it is 
+  calculated when :py:meth:`~.signal.BaseSignal.save` is called, which can be time consuming.
+  Alternatively, setting the argument to `None` will result in a correctly sized
+  zero array to be used. Finally, a custom ``Signal2D`` object can be passed,
+  but the shape must match the navigation dimensions.
 
 .. _dens-format:
 
--- a/doc/user_guide/io.rst
+++ b/doc/user_guide/io.rst
@@ -797,6 +797,23 @@ loading the file using a different mode. However, note that lazy loading
 does not support in-place writing (i.e lazy loading and the "r+" mode
 are incompatible).
 
+Extra saving arguments
+^^^^^^^^^^^^^^^^^^^^^^
+
+- ``intensity_scaling`` : in case the dataset that needs to be saved does not
+  have the `np.uint8` data type, casting to this datatype without intensity
+  rescaling results in overflow errors (default behavior). This option allows
+  you to perform linear intensity scaling of the images prior to saving the 
+  data. The options are:
+  
+  - `'dtype'`: the limits of the datatype of the dataset, e.g. 0-65535 for 
+    `np.uint16`, are mapped onto 0-255 respectively. Does not work for `float`
+    data types.
+  - `'minmax'`: the minimum and maximum in the dataset are mapped to 0-255.
+  - `'crop'`: everything below 0 and above 255 is set to 0 and 255 respectively
+  - 2-tuple of `floats` or `ints`: the intensities between these values are
+    scaled between 0-255, everything below is 0 and everything above is 255.
+
 .. _dens-format:
 
 DENS heater log
--- a/hyperspy/io_plugins/blockfile.py
+++ b/hyperspy/io_plugins/blockfile.py
@@ -117,7 +117,7 @@ def get_header_from_signal(signal, endianess='<'):
         axes_manager.convert_units('signal', 'cm')
     except Exception:
         warnings.warn("Could not convert between units, scales in the "
-                      "blo file may be incorrect")
+                      "blo file may be incorrect", UserWarning)
 
     if axes_manager.navigation_dimension == 2:
         NX, NY = axes_manager.navigation_shape
@@ -172,7 +172,8 @@ def file_reader(filename, endianess='<', mmap_mode=None,
     header = np.fromfile(f, dtype=get_header_dtype_list(endianess), count=1)
     if header['MAGIC'][0] not in magics:
         warnings.warn("Blockfile has unrecognized header signature. "
-                      "Will attempt to read, but correcteness not guaranteed!")
+                      "Will attempt to read, but correcteness not guaranteed!",
+                      UserWarning)
     header = sarray2dict(header)
     note = f.read(header['Data_offset_1'] - f.tell())
     # It seems it uses "\x00" for padding, so we remove it
@@ -289,7 +290,8 @@ def file_writer(filename, signal, **kwds):
         if signal.data.dtype != "u1":
             warnings.warn("Data does not have uint8 dtype: values outside the "
                           "range 0-255 may result in overflow. To avoid this "
-                          "use the 'intensity_scaling' keyword argument.")
+                          "use the 'intensity_scaling' keyword argument.",
+                          UserWarning)
     elif scale_strategy == "dtype":
         original_scale = dtype_limits(signal.data)
         if original_scale[1] == 1.:
--- a/hyperspy/io_plugins/blockfile.py
+++ b/hyperspy/io_plugins/blockfile.py
@@ -116,8 +116,7 @@ def get_header_from_signal(signal, endianess='<'):
         axes_manager.convert_units('navigation', 'nm')
         axes_manager.convert_units('signal', 'cm')
     except Exception:
-        warnings.warn("Could not convert between units, scales in the "
-                      "blo file may be incorrect", UserWarning)
+        warnings.warn("Could not convert between units, scales in the blo file may be incorrect", UserWarning)
 
     if axes_manager.navigation_dimension == 2:
         NX, NY = axes_manager.navigation_shape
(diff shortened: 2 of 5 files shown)


hyperspy/io_plugins/blockfile.py
```diff
@@ -255,7 +262,52 @@ def file_reader(filename, endianess='<', mmap_mode=None,
...
+    elif scale_strategy == "dtype":
+        original_scale = dtype_limits(signal.data)
+        if original_scale[1] == 1.:
+            raise ValueError("Signals with float dtype can not use 'dtype'")
+    elif scale_strategy == "minmax":
+        minimum = signal.data.min()
+        maximum = signal.data.max()
+        if signal._lazy:
+            minimum = minimum.compute()
+            maximum = maximum.compute()
```
ericpre: Group these computations in a single `dask.compute`: 
See https://docs.dask.org/en/latest/best-practices.html#avoid-calling-compute-repeatedly

hyperspy/io_plugins/blockfile.py
```diff
@@ -270,9 +322,13 @@ def file_writer(filename, signal, **kwds):
         # Write virtual bright field
         vbf = signal.mean(
             signal.axes_manager.signal_axes[
-                :2]).data.astype(
-            endianess +
-            'u1')
+                :2])
+        if signal._lazy:
+            vbf.compute()
```
ericpre: Since this can be a long computation, it may be worth adding an argument to select one of the following:
- use the `navigator` if it has been already computed
- pass a signal
- compute the mean, as it is currently the case
- do nothing, image full of zeros?

hyperspy/io_plugins/blockfile.py
```diff
@@ -287,7 +343,12 @@ def file_writer(filename, signal, **kwds):
                                         ('ID', endianess + 'u4')])
         dp_head['MAGIC'] = 0x55AA
         # Write by loop:
-        for img in signal._iterate_signal():
+        for img in tqdm(signal._iterate_signal(),
```
ericpre: Use https://github.com/hyperspy/hyperspy/blob/2ddc359c5a6c0ff300507c383dfcc1de644ff11a/hyperspy/external/progressbar.py#L32 and https://github.com/hyperspy/hyperspy/blob/2ddc359c5a6c0ff300507c383dfcc1de644ff11a/hyperspy/model.py#L470-L471

hyperspy/io_plugins/blockfile.py
```diff
@@ -287,7 +343,12 @@ def file_writer(filename, signal, **kwds):
...
+                        total=signal.data.shape[0] * signal.data.shape[1]):
+            # to ensure dask datasets, get a tofile method
             dp_head.tofile(f)
+            img = np.array(img)
+            if scale_strategy is not None:
+                img = rescale_intensity(img, in_range=original_scale, out_range=np.uint8)
```
ericpre: Instead of using a `for` loop, would it be better to use `s.unfold_navigation_space` or `s.data.reshape(new_shape)`?

hyperspy/io_plugins/blockfile.py
```diff
@@ -109,8 +112,11 @@ def get_header_from_signal(signal, endianess='<'):
     # The navigation and signal units are 'nm' and 'cm', respectively, so we
     # convert the units accordingly before saving the signal
     axes_manager = signal.axes_manager.deepcopy()
-    axes_manager.convert_units('navigation', 'nm')
-    axes_manager.convert_units('signal', 'cm')
+    try:
+        axes_manager.convert_units('navigation', 'nm')
+        axes_manager.convert_units('signal', 'cm')
+    except Exception:
+        warnings.warn("Could not convert between units, scales in the blo file may be incorrect", UserWarning)
```
ericpre: I guess the units in blockfile are "cm" for signal axes and "nm" for navigation axes and the ASTAR softwares will expect the scale to be exported in these units and it was raising an error, in order to comply with the format definition. Is it worth adding more details to the warning explaining this?

ericpre (commented):
Looks good, see suggestions for improvement.

hyperspy/io_plugins/blockfile.py
din14970: Correctly calibrated 4D-STEM datasets will of course have 1/length or angle units, so this error will pop up for most datasets processed with Hyperspy/Pyxem. Since the conversion to cm depends on a bunch of factors that are typically not stored in the metadata like real pixel size, it will be hard to make this conversion reliable for any dataset. So indeed I will expand on the warning.

hyperspy/io_plugins/blockfile.py
ericpre: I had a look at this and the performance when writing lazy signal and it is not very good, one order of magnitude worse than using hdf5 file - and there is faster than hdf5.
One issue is that it uses `LazySignal._iterate_signal`, which rechunk the data most likely unnecessarily. I think the best to fix this issue in a separate PR, which I would be happy to do!

din14970: A loop never looks great in python and indeed performance is terrible, which is why I added the progress bar. I was thinking in the direction of using [`dask.array.store`](https://docs.dask.org/en/latest/array-api.html?highlight=store#dask.array.store) on a `np.memmap` object, but haven't experimented yet to see if it would work.

hyperspy/io_plugins/blockfile.py
din14970: Since I got rid of the loop, I'm using something modeled after https://github.com/hyperspy/hyperspy/blob/2ddc359c5a6c0ff300507c383dfcc1de644ff11a/hyperspy/_signals/lazy.py#L121 with the dask progress bar. There is no more progress bar possible with non-lazy datasets.

hyperspy/io_plugins/blockfile.py
din14970: I implemented something similar (if `None` -> array of zeros, if `'navigator'` -> navigator for lazy dataset, mean for non-lazy, if a signal is passed, signal is used.) but it would be cleaner and benefit from a resolution of  #2639

hyperspy/io_plugins/blockfile.py
din14970: so this seems to work and is in my opinion quite elegant. Probably it's also the fastest it will go. With the loop I was counting about an hour to convert one 11 GB lazy dataset into another, now it takes 3 minutes including the intensity rescaling. This involved the extraction from a `.hspy` file and saving to a `.blo` file. From what I could tell I did not have any issues with undue memory pressure, even though I saw this issue pop up https://github.com/dask/dask/issues/5985 . Perhaps something to keep in mind if there are any memory related issues in the future.

doc/user_guide/io.rst
```diff
@@ -797,6 +797,23 @@ loading the file using a different mode. However, note that lazy loading
...
+  you to perform linear intensity scaling of the images prior to saving the 
+  data. The options are:
+  
+  - `'dtype'`: the limits of the datatype of the dataset, e.g. 0-65535 for 
+    `np.uint16`, are mapped onto 0-255 respectively. Does not work for `float`
+    data types.
+  - `'minmax'`: the minimum and maximum in the dataset are mapped to 0-255.
+  - `'crop'`: everything below 0 and above 255 is set to 0 and 255 respectively
+  - 2-tuple of `floats` or `ints`: the intensities between these values are
+    scaled between 0-255, everything below is 0 and everything above is 255.
```
ericpre, line 815: The new `vbf` argument needs to be added here too.

hyperspy/io_plugins/blockfile.py
```diff
@@ -126,168 +151,284 @@ def get_header_from_signal(signal, endianess='<'):
...
+        '<' (default) or '>' determining how the bits are written to the file
+    intensity_scaling : str or 2-Tuple of float/int
+        If the signal datatype is not uint8 this argument provides intensity
+        linear scaling strategies. If 'dtype', the entire dtype range is mapped
+        to 0-255, if 'minmax' the range between the minimum and maximum intensity is
+        mapped to 0-255, if 'crop' the range between 0-255 is conserved without
+        overflow, if a tuple of values the values between this range is mapped
+        to 0-255. If None (default) no rescaling is performed and overflow is
+        permitted.
+    vbf : str or Signal2D
```
ericpre: It is better to avoid abbreviation, would `navigator` be better?

ericpre (commented):
It needs two small tweaks before merging.

hyperspy/io_plugins/blockfile.py
din14970: navigator="navigator" seems a bit strange to me though. maybe `nav_image`, `navigator_image` or `virtual_image`. Not sure if I have any favorites.

din14970:
I'm still a bit puzzled by the two lines that are somehow no longer being covered by tests. I'm trying to resolve this.

hyperspy/io_plugins/blockfile.py
din14970: I went with `navigator_signal`

din14970:
should be ok now if you are happy with docs and renaming of parameter

ericpre (approved):

Issue #2780, opened 2021-06-25
Adding reader for TVIPS datastream files


din14970:
This issue serves mainly to notify about something I'm currently working on, and to have something to reference. [TVIPS](https://www.tvips.com/) is a very small German company that sells CMOS cameras that are fast and have good dynamic range. They seem to be decently popular in Germany. We have these cameras installed our JEOL instruments and use them for in-situ studies and various 4D-STEM experiments. This kind of data is collected in a quite inconvenient stream form, starting with a main header and each image frame preceded by a small frame header. In addition, files are capped at a certain size, so datasets continue by creating and saving to additional files (`file_000.tvips`, `file_001.tvips`, ...). There is very limited metadata included in the files, so users must typically reconstruct the shape of the data hypercube themselves. To deal with this data, I've previously written [a GUI based converter program](https://github.com/din14970/TVIPSconverter) to convert the data to blo, hspy or tiff. However, it is annoying to have to duplicate very large datasets before one can work with them, and besides the additional wasted disk space it can take 10-30 min for a conversion. I think at this point I have enough knowledge to implement a file reader directly in Hyperspy. I envision adding the following arguments; most of them are only relevant for 4D-STEM datasets:

* `scan_shape`: in case the dataset is a 4D STEM dataset and the original shape can not be automatically detected
* `first_frame`: index of the first frame to include in the dataset in case it can not be automatically detected
* `last_frame`: index of the last frame in case it can not be automatically detected
* `winding_scan`: `boolean`, whether the scan unit operated under flyback mode or "snake-scan" mode
* `hysteresis`: scan point offset of even scan rows to correct for miss-aligned "snake-scan" mode

By default, the data would just be loaded as an image stack unless `scan_shape` is defined.
The original implementation in my GUI converter relies on a loop over the files. I hope I can do something a bit smarter with `np.memmap`, even though the array data is non contiguous and possibly split over multiple files.

For `original_metadata` I was planning to only include the main header. However there is the possibility to record additional information like temperature in the frame metadata, so it might be handy to be able to also optionally load and return this information.

magnunor:
More file formats is always nice!

With regards to lazy loading + `scan_shape` combined with `np.memmap`, there is probably some clever things you can do with regards to the chunking to get optimal performance. For example making sure the chunks doesn't extend over several files, to avoid read amplification.

ericpre:
Done in #2781.

PR #2781, opened 2021-06-26
Add .tvips file reader plug-in

din14970:
### Description of the change
As discussed in #2780 this PR aims to add reading capabilities for .tvips files.

It needs more testing but at first sight it seems to work on most of the files I have.  It will probably take a bit longer to finish this as I will need to create dummy files for the tests - the real files I have are all huge.

### Progress of the PR
- [x] Change implemented
  - [x] `file_reader` function that parses the data and metadata
  - [x] binding to the `io.save` function
- [x] add tests,
- [x] ready for review.

```python
import hyperspy.api as hs
# loads as an image stack
hs.load("datafile.tvips", lazy=True)
# loads as a 4D STEM dataset
hs.load("datafile.tvips", lazy=True, scan_shape=(100, 100), scan_start_frame=20)
```


hyperspy/io_plugins/tvips.py
```diff
@@ -0,0 +1,347 @@
...
+from hyperspy.misc.array_tools import sarray2dict
+
+# Plugin characteristics
+# ----------------------
+format_name = 'tvips'
+description = ('Read support for TVIPS CMOS camera stream/movie files. Can be'
+               'used for in-situ movies or 4D-STEM datasets.')
+full_support = False
+# Recognised file extension
+file_extensions = ['blo', 'BLO']
```
harripj: Should this be ```'tvips'``` and ```'TVIPS'```?

harripj (changes requested):
Requesting clarity on file_extensions for the plugin.

hyperspy/io_plugins/tvips.py
din14970: yes, I haven't had time to look at this recently, I think at the beginning of october I can resume work on this

hyperspy/io_plugins/tvips.py
```diff
@@ -0,0 +1,347 @@
...
+        # Create the axis objects for each axis
+        dim = data_stack.ndim
+        axes = [
+            {
+                'size': data_stack.shape[i],
+                'index_in_array': i,
+                'name': names[i],
+                'scale': scales[i],
+                'offset': offsets[i],
+                'units': units[i], }
```
ericpre: Following up on discussion in https://github.com/din14970/TVIPSconverter/issues/15#issuecomment-916041983, in these dictionary, a value should be defined for the `'navigate'` key to set the navigation axes (`'navigate'=True` means that this axis is a navigation axes!).

hyperspy/io_plugins/tvips.py
```diff
@@ -130,27 +196,152 @@ def file_reader(filename,
...
+            scan_dim = int(np.sqrt(total_scan_frames))
+            if not np.allclose(scan_dim, np.sqrt(total_scan_frames)):
+                raise ValueError("Scan was not square, please supply a scan_shape and start_frame.")
+            scan_shape = (scan_dim, scan_dim)
+            # there may be discontinuities which must be filled up
+            indices = _guess_scan_index_grid(record_idxs, scan_start_frame, scan_stop_frame)
+        # scan shape and start are provided
+        else:
+            total_scan_frames = scan_shape[0] * scan_shape[1]
+            indices = np.arange(scan_start_frame, scan_start_frame+total_scan_frames).reshape(scan_shape[0], scan_shape[1])
```
harripj: ```scan_shape``` could be adapted to take any number of dimensions, for example to load the signal as 1d. Something like this would seem to work for any dimensionality:
```python
total_scan_frames = np.prod(scan_shape)
indices = np.arange(scan_start_frame, scan_start_frame+total_scan_frames).reshape(scan_shape)
```

harripj (changes requested):

hyperspy/io_plugins/tvips.py
harripj: This addition fixed things for me:
```'navigate': True if i < len(scan_shape) else False```

File: hyperspy/io_plugins/tvips.py [hunk 4]
```diff
@@ -0,0 +1,347 @@
...
+        dim = data_stack.ndim
+        axes = [
+            {
+                'size': data_stack.shape[i],
+                'index_in_array': i,
+                'name': names[i],
+                'scale': scales[i],
+                'offset': offsets[i],
+                'units': units[i], }
+            for i in range(dim)]
```
harripj, line 418: Following on from the previous [comment ](https://github.com/hyperspy/hyperspy/pull/2781/files/a1bee7caf10dcac17a9785eb38a715bf594e3dc2#r705321351), a similar addition here:
```'navigate': True if not i else False```

File: doc/user_guide/io.rst [hunk 5]
```diff
@@ -1905,6 +1907,70 @@ load the same file without extra arguments:
...
+  is ignored.
+- ``winding_scan_axis``: if the acquisition software collected data without
+  beam flyback but with a winding "snake" scan, then every second scan row
+  or column needs to be reversed to make sense of the data. This can be
+  indicated with values `"x"` or `"y"`, depending on whether winding happened
+  along the primary or secondary axis.
+- ``hysteresis``: if winding scan was active it is likely there is an overshoot
+  of a few pixels (2-5) every second scan row. This parameter shifts every
+  second row by the indicated number of scan points to align even and odd scan
+  rows.
```
ericpre: Specify default value for `winding_scan_axis` and `hysteresis`.

doc/user_guide/io.rst
```diff
@@ -1905,6 +1907,70 @@ load the same file without extra arguments:
...
+Extra saving arguments
+^^^^^^^^^^^^^^^^^^^^^^
+
+- ``max_file_size``: approximate maximum size of individual files in bytes. 
+  In this way a dataset can be split into multiple files. A file needs to be
+  at least the size of the main header in the first file plus one frame and its
+  frame header.
+- ``version``: TVIPS file format version, defaults to version 2. Only version
+  1 and 2 are currently supported.
+- ``frame_header_extra_bytes``: the number of bytes to pad the frame headers
```
ericpre, line 1977: Specify default value for `max_file_size` and `frame_header_extra_bytes`.

hyperspy/io_plugins/tvips.py
```diff
@@ -0,0 +1,588 @@
...
+# Recognised file extension
+file_extensions = ['tvips', 'TVIPS']
+default_extension = 0
+# Writing capabilities:
+writes = False
+non_uniform_axis = False
+
+
+_logger = logging.getLogger(__name__)
+_ureg = pint.UnitRegistry()
```
ericpre: Use `from hyperspy.api_nogui import _ureg` instead.

As per https://pint.readthedocs.io/en/stable/tutorial.html#using-pint-in-your-projects.

hyperspy/io_plugins/tvips.py
```diff
@@ -0,0 +1,588 @@
...
+        If it concerns a 4D-STEM dataset, the (..., scan_y, scan_x) dimension can
+        be provided. "auto" can also be selected, in which case the rotidx
+        information in the frame headers will be used to try to reconstruct
+        the scan. Additional navigation axes must be prepended.
+    scan_start_frame : int
+        First frame where the scan starts. If scan_shape = "auto" this is
+        ignored.
+    winding_scan_axis : str
+        "x" or "y" if the scan was performed with winding scan along an axis
+        as opposed to flyback scan.
```
ericpre: Specify what it does when it is `None`.

hyperspy/io_plugins/tvips.py
```diff
@@ -0,0 +1,588 @@
...
+                'navigate': True if i < len(scan_shape) else False,
+            }
+            for i in range(dim)]
+        if lazy:
+            chunking = {}
+            for ax_index in range(indices.ndim):
+                chunking[ax_index] = "auto"
+            for ax_index in range(indices.ndim, indices.ndim + 2):
+                chunking[ax_index] = None
+            data_stack = data_stack.rechunk(chunking)
```
ericpre: Is there any value in making rechunking optional (possibly with a default to `True`, if this is sensible) and having it as a parameter of the reader? If so, there is another one in the `else` branch below.

hyperspy/io_plugins/tvips.py
```diff
@@ -0,0 +1,588 @@
...
+            milliseconds = rotator * time_increment
+            timestamps = (timestamp + milliseconds / 1000).astype(int)
+            milliseconds = milliseconds % 1000
+            file_memmap["timestamp"] = timestamps
+            file_memmap["ms"] = milliseconds
+            file_memmap["rotidx"] = rotator + 1
+            data = signal.data[current_frame:current_frame + frames_saved]
+            if signal._lazy:
+                if show_progressbar is None:
+                    show_progressbar = preferences.General.show_progressbar
```
ericpre:
```suggestion
                show_progressbar = kwds.get(
                    'show_progressbar',
                    preferences.General.show_progressbar
                    )
```
And remove the line setting `show_progressbar` above

din14970:
@ericpre @harripj sorry for the long delay, I think this should be good to go

harripj:
@din14970 no need to be sorry, thank you for your work!

I will preface my comments here by acknowledging that I am by no means an expert in hyperspy, so please forgive me if I am making some basic mistakes.

Firstly this is an excellent addition to hyperspy and very useful for anybody working with data from TVIPS cameras and data.

I have tested this on some .tvips data that we have collected (not in the testing data uploaded to this PR). On my Mac the `lazy` loader works well and does not load the whole file to memory. We discussed [previously](https://github.com/din14970/TVIPSconverter/issues/15#issuecomment-915967600) that there may be an issue on Windows, but if I understood, this is actually related to a Dask issue and not this PR.

When I run the example, eg. `hs.load('datafile.tvips', lazy=True)`, the following warning is always raised for me: ```WARNING:hyperspy.io:`signal_type='imaging'` not understood. See `hs.print_known_signal_types()` for a list of installed signal types or https://github.com/hyperspy/hyperspy-extensions-list for the list of all hyperspy extensions providing signals.``` Perhaps there is an extension that I need to install to suppress this warning? I then installed `pyxem` but the warning persists. Is the `"imaging"` signal type related to another package?

If I try to load a file as 1d or 2d, all works as expected. If the `scan_shape` provided is too large for the file the following error is raised: `IndexError: Index is out of bounds for axis 0 with size 19143`, but perhaps this could be more informative. It would be nice if the number of frames in the dataset(s) can be determined beforehand to determine if the scan indices (shape + offset) fit and a clearer error is raised to the user.

To me this looks great, thanks very much for working on this! I will leave this to someone involved with the hyperspy project to provide some more feedback.

din14970:
@harripj Thanks for having another look. The goal would of course be to make it work on real TVIPS datasets, the test datasets were actually artificially created.

> We discussed https://github.com/din14970/TVIPSconverter/issues/15#issuecomment-915967600 that there may be an issue on Windows, but if I understood, this is actually related to a Dask issue and not this PR.

Yes I'm not sure how to deal with this, I also don't have a Windows machine on which I could test/debug. Perhaps if it's a real issue for you you could have a deeper look into the task graph on Windows to figure out what is going on?

> WARNING:hyperspy.io:`signal_type='imaging'` not understood

You actually caught a bug here, so thanks for that - I use `signal_type` in the metadata to distinguish between diffraction and imaging data, but it seems hyperspy uses this info for more than just information purposes. Will remove this.

>  It would be nice if the number of frames in the dataset(s) can be determined beforehand to determine if the scan indices (shape + offset) fit and a clearer error is raised to the user.

Good idea, will implement

harripj:
> The goal would of course be to make it work on real TVIPS datasets, the test datasets were actually artificially created.

For our experimental datasets it works as expected!

> Yes I'm not sure how to deal with this, I also don't have a Windows machine on which I could test/debug. Perhaps if it's a real issue for you you could have a deeper look into the task graph on Windows to figure out what is going on?

I can test this PR again on Windows next week and let you know. I think even if the problem exists it is likely Dask related and beyond the scope of this PR. I don't think that this potential issue should stop this PR from being merged.

Thanks again for implementing this @din14970!

hyperspy/io_plugins/tvips.py
```diff
@@ -424,17 +431,18 @@ def file_reader(filename,
...
         'General': {
             'original_filename': os.path.split(filename)[1],
             'date': date,
             'time': time,
             'time_zone': "UTC",
         },
-        "Signal": {
-            'signal_type': sigtype
-        },
+        # "Signal": {
```
ericpre: You can use any signal type listed in https://github.com/hyperspy/hyperspy-extensions-list. In this case, it could be `diffraction` or `electron_diffraction`. If I remember correctly, TVIPS camera can be used for X-rays, so if there is way to make sure that it was an electron diffraction, then it would be more accurate to use `electron_diffraction`. 

For imaging case, it should be left "empty" as there is no extension which implements a corresponding signal.

ericpre (commented):
This looks very good! I added a few suggestions.

It would be good to try to get the full coverage - see codecov warning in the "files changed" tab. When it is not possible to cover or not worth it, you can use `# pragma: no cover`.

ericpre:
> > Yes I'm not sure how to deal with this, I also don't have a Windows machine on which I could test/debug. Perhaps if it's a real issue for you you could have a deeper look into the task graph on Windows to figure out what is going on?
> 
> I can test this PR again on Windows next week and let you know. I think even if the problem exists it is likely Dask related and beyond the scope of this PR. I don't think that this potential issue should stop this PR from being merged.

That would be great! Let see how it goes and then think about it if we need to! ;)

doc/user_guide/io.rst
din14970: Done

doc/user_guide/io.rst
din14970: done

hyperspy/io_plugins/tvips.py
din14970: done

hyperspy/io_plugins/tvips.py
din14970: done

hyperspy/io_plugins/tvips.py
din14970: I now only add the `Signal` key if mode is 2 and set `sigtype` to `diffraction`. I don't know whether it's possible to distinguish between electrons or X-rays, but it should be relatively easy for users to change from diffraction to electrondiffraction if relevant.

hyperspy/io_plugins/tvips.py
din14970: done

hyperspy/io_plugins/tvips.py
din14970: I've made `rechunking` a kwarg. If it's "falsy" no rechunking occurs. I'm pretty sure this would mean each file corresponds to a chunk. If it's "auto" = default then optimal rechunking occurs along navigation dimensions as above. If it's anything else, it's passed directly to the `chunks` argument in `dask.array.rechunk`, in case the user wants to do a custom rechunking.

din14970:
@ericpre A couple lines are missing from coverage because I don't have files to test them on. I only ever found and collected data with header v2. It's in there because I found it in an older implementation for a reader. It is possible the format evolves further, which the exceptions should take care of. I've added the pragma lines as you suggest.

On a subject totally unrelated to this PR, it appears your affiliation has changed and your e-mail address is no longer valid. I've sent e-mails to your superstem account but I'm not sure if this one is valid either. If not, you probably haven't received any of the e-mail traffic related to the pyxem paper on which I invited you as co-author, so I'm informing you via this channel that the paper is now accepted in Ultramicroscopy. Could you send me an e-mail if you want to review the proofs once I receive them?

ericpre (approved):
Looks good to me. Before merging, it would be good to know the outcome of @harripj's testing on windows. Thanks!

harripj:
Just a follow up on this. On my Windows machine I seem to be experiencing the same behaviour as [before](https://github.com/din14970/TVIPSconverter/issues/15#issuecomment-915967600):

If I load one of our experimental datasets (size 9.9 GB) as ` data = hs.load('dset.tvips', lazy=True)` the file is not loaded to memory as expected. With `lazy=False` the peak memory usage is double the file size before settling back to ~file size:

image

Using the lazy `loader` again, if I just extract the center pixel from each frame as:
```python
center = data[:, 256, 256].compute()
```
The whole file is loaded to memory and persists:
image

Interestingly the memory presists even if I call `del center; del data`.

Dask version:  `2022.03.0`
NumPy version : `1.21.5`
Hyperspy version: `1.7.0.dev0`, ie. local install of this PR.

ericpre:
Thanks @harripj for checking it. Let's merge this PR and park it as a "known issue on windows"... Fixing it can be done in a separate PR.

PR #2785, opened 2021-06-28
Fix loading duplicated dataset for NeXus file loader

### LINKED ISSUE #2779: Nexus file save/load two signals return
When saving and loading a `nxs` style HDF5 file in HyperSpy, an extra signal is returned. Example:

```python
import numpy as np
import hyperspy.api as hs
s = hs.signals.Signal1D(np.random.random((20, 1000)))
s.save("test_nexus.nxs")
s_load = hs.load("test_nexus.nxs")
```

`s_load` is a list of two signals:
```python
[<Signal1D, title: unnamed__0, dimensions: (20|1000)>,
 <BaseSignal, title: entry1_unnamed__0_data, dimensions: (1000, 20|)>]
```

Is this intended behaviour?

ptim0626:
### Description of the change
This fixes #2779, which avoid loading duplicated dataset for the NeXus file loader when `nxdata_only=False`.

### Progress of the PR
- [x] Change implemented
- [x] updated tests,
- [x] ready for review.

```python
import hyperspy.api as hs
import numpy as np
s = hs.signals.Signal1D(np.random.random((20, 1000)))
s.save("test_duplicated_load.nxs")
s_load = hs.load("test_duplicated_load.nxs")
```

s_load is `<Signal1D, title: unnamed__0, dimensions: (20|1000)>`.

### DIFF
--- a/hyperspy/io_plugins/nexus.py
+++ b/hyperspy/io_plugins/nexus.py
@@ -311,6 +311,16 @@ def _extract_hdf_dataset(group, dataset, lazy=False):
     """
 
     data = group[dataset]
+
+    # exclude the dataset tagged by the signal attribute to avoid extracting
+    # duplicated dataset, which is already loaded when loading NeXus data
+    if 'signal' in data.parent.attrs.keys():
+        data_key = data.parent.attrs['signal']
+        if isinstance(data_key, bytes):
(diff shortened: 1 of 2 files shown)


ericpre (approved):
Thanks!

PR #2787, opened 2021-07-01
fix broken "developer guide" hyperlink

wstolp:


thomasaarholt:
Thanks! I think it is fine to point to "latest" for the dev guide, rather than stable.

ericpre:
> Thanks! I think it is fine to point to "latest" for the dev guide, rather than stable.

Agreed!

PR #2788, opened 2021-07-01
rebin real and imaginary parts separately in case of complex data array

wstolp:
Rebinning a complex2d signal gave me an error, something with numba that I didn't fully grasp. The problem is solved by getting `data.real` and `data.imag` and rebinning them separately in the BaseSignal class.

Let me know if I'm doing something wrong!

for example this would give me an error:
```python
a = np.ones((10,10), dtype='complex')
s = hs.signals.ComplexSignal2D(a)
pi = np.pi
s.rebin(scale=(pi, pi))
```

error:
```bash
TypingError: No implementation of function Function(<built-in function iadd>) found for signature:
 
iadd(array(float64, 2d, C), array(complex128, 1d, C))
 
There are 16 candidate implementations:
    - Of which 12 did not match due to:
    Overload of function 'iadd': File: <numerous>: Line N/A.
      With argument(s): '(array(float64, 2d, C), array(complex128, 1d, C))':
     No match.
    - Of which 2 did not match due to:
    Overload in function 'NumpyRulesInplaceArrayOperator.generic': File: numba/core/typing/npydecl.py: Line 213.
      With argument(s): '(array(float64, 2d, C), array(complex128, 1d, C))':
     Rejected as the implementation raised a specific error:
       AttributeError: 'NoneType' object has no attribute 'args'
  raised from /home/ugct.ugent.be/wstolp/Programs/HyperSpyTemp/py3.9-testenv/lib64/python3.9/site-packages/numba/core/typing/npydecl.py:224
    - Of which 2 did not match due to:
    Operator Overload in function 'iadd': File: unknown: Line unknown.
      With argument(s): '(array(float64, 2d, C), array(complex128, 1d, C))':
     No match for registered cases:
      * (int64, int64) -> int64
      * (int64, uint64) -> int64
      * (uint64, int64) -> int64
      * (uint64, uint64) -> uint64
      * (float32, float32) -> float32
      * (float64, float64) -> float64
      * (complex64, complex64) -> complex64
      * (complex128, complex128) -> complex128
```


wstolp:
Oh, I might add that there are no rebin tests for complex signals in the test folder.

ericpre:
Thanks for making the PR.
I would suggest an alternative, which should be more efficient. The error shows that numba complains about unsupported dtype signature in `During: typing of intrinsic-call at /home/eric/Dev/hyperspy/hyperspy/misc/array_tools.py (230)` (this part is missing [above](https://github.com/hyperspy/hyperspy/pull/2788#issue-682024978)) and the reason for this is that the `result` is of type `float` instead of `complex`. Therefore, specify the correct when creating the `result` array in https://github.com/hyperspy/hyperspy/blob/3a070641306cb727f6710897dfcb86b46dafb642/hyperspy/misc/array_tools.py#L315 should fix it:
`result = np.zeros((dim,) + dat.shape[1:], dtype=dat.dtype)`

It should be more efficient because instead of iterating over the real and imaginary part, numba will do the sum with complex type directly.

Can you also add a test, maybe simply add a complex type in the following test: https://github.com/hyperspy/hyperspy/blob/3a070641306cb727f6710897dfcb86b46dafb642/hyperspy/tests/signals/test_linear_rebin.py#L28

wstolp:
Thanks for your feedback, that's certainly a much better solution. I guess I make a completely new pull request now and close the current one?
For the testing, I'll do it when I have time for it.

ericpre:
Yes, in order not to include the commit of this PR, the easiest is to make a new PR, but there is the alternative to remove the last commit locally (reset to previous commit or use amend) and "force push" to this branch.

Regarding adding test, the good news is that it will be very easy to add to the existing test mentioned above:
https://github.com/hyperspy/hyperspy/blob/3a070641306cb727f6710897dfcb86b46dafb642/hyperspy/tests/signals/test_linear_rebin.py#L28

by adding `'c8'` to the list of parameter type to test, in other word: replace `@pytest.mark.parametrize('dtype', ['<u2', 'u2', '>u2', '<f4', 'f4', '>f4']) ` by `@pytest.mark.parametrize('dtype', ['<u2', 'u2', '>u2', '<f4', 'f4', '>f4', 'c8'])`. If you change the test without the fix and run this test, you will see that it fails with the same error message that you mentioned above.

PR #2789, opened 2021-07-05
fix complex rebin and add complex test

wstolp:
My second try to fix rebinning of complex signals, as suggested by @ericpre:

- I added a test for complex data types.
- I specified the data type during creation of the array that will receive the rebinned data.
- To pass the test I had to rearrange the endianness conversion (don't know what this is tbh).

See discussion in https://github.com/hyperspy/hyperspy/pull/2788.

### DIFF
--- a/hyperspy/misc/array_tools.py
+++ b/hyperspy/misc/array_tools.py
@@ -310,12 +310,12 @@ def _linear_bin(dat, scale, crop=True):
                 "Re-adjust your scale values or run code with "
                 "crop=False to avoid this error."
             )
-        # Set up the result np.array to have a new axis[0] size for after
-        # cropping.
-        result = np.zeros((dim,) + dat.shape[1:])
         # Make sure that native endian is used
         if not dat.dtype.isnative:
             dat = dat.astype(dat.dtype.type)
+        # Set up the result np.array to have a new axis[0] size for after
+        # cropping.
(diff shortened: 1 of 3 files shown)


ericpre (commented):
This looks good, can you please add an entry for the changelog as mentioned in the PR template - see https://github.com/hyperspy/hyperspy/blob/RELEASE_next_minor/upcoming_changes/README.rst for more details?

wstolp:
Sorry I did not read that template attentively and now it's gone. Maybe it would be good to put that info in the developer documentation as well.
Anyway I added an RST file. Would that be all? :)

ericpre (approved):
Thanks!

ericpre:
> Sorry I did not read that template attentively and now it's gone. Maybe it would be good to put that info in the developer documentation as well.

No worries! Yes, it would be good to see how this can be improved, but at the same time information should not be duplicated, because very often, one duplicate get outdated... Also as a new contributor, if you have any other comments on what is not clear, this is more than welcome!

> Anyway I added an RST file. Would that be all? :)

Yes, this is merged now! We use zenodo to reference hyperspy and the contributor list is pulled automatically from the github contributor profile. If you want to be credited with your full name, you need to update your github profile. Adding affiliation may be good too, because we use it sometimes for conference abstract, etc.

PR #2790, opened 2021-07-06
Handle polynomial with order >10 for estimate_parameters

ptim0626:
### Description of the change
This short PR fixes the issue for `estimate_parameters` in the `Polynomial` component if someone wants to use a polynomial with an order greater than 10. This allows the usage of, such as `remove_background` in non-interactive mode, with a polynomial order greater than 10, if one desires.

### Progress of the PR
- [x] Change implemented,
- [x] modify tests,
- [x] ready for review.

This works after the fix:
```python
import numpy as np
import hyperspy.api as hs
s = hs.signals.Signal1D(np.random.random((4,4,32)))
s.remove_background(signal_range=(None, None), background_type='Polynomial', polynomial_order=12)
```

### DIFF
--- a/hyperspy/_components/polynomial.py
+++ b/hyperspy/_components/polynomial.py
@@ -33,7 +33,7 @@ class Polynomial(Expression):
     """n-order polynomial component.
 
     Polynomial component consisting of order + 1 parameters.
-    The parameters are named "a" followed by the corresponding order, 
+    The parameters are named "a" followed by the corresponding order,
     i.e.
 
     .. math::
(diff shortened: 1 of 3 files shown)


ericpre (commented):
Looks good! Can you please add changelog entry in the `upcoming_changes` folder as explained in [`upcoming_changes/README.rst`](https://github.com/hyperspy/hyperspy/blob/RELEASE_next_minor/upcoming_changes/README.rst)

ptim0626:
Oh yes I thought it is only for new feature/bigger changes, done now.

ericpre (approved):

ericpre:
Changelog entries for bug fixes is also useful, for example, in the case of this PR, if someone want to know which minimum version of hyperspy is required to include your bug fix, they can figure out easily by reading the changelog!

PR #2795, opened 2021-07-08
Add context manager to switch iterpath

ericpre:
### Progress of the PR
- [x] Add context manager to set switch iterpath,
- [x] Remove `_iterate_signal` of lazy signal, its non-lazy counterpart works fine
- [x] simplify `_iterate_signal` and make it use `__call__` so that lazy signal would benefit from the speed improvement in #2568,
- [x] Fix warning when building user guide, 
- [x] add tests,
- [x] ready for review.

```python
import hyperspy.api as hs
import numpy as np

s = hs.signals.Signal1D(np.arange(2*3*4).reshape([3, 2, 4]))

with s.axes_manager.switch_iterpath('serpentine'):
    for indices in s.axes_manager:
        print(indices)
```

### DIFF
--- a/hyperspy/signal.py
+++ b/hyperspy/signal.py
@@ -3498,7 +3498,8 @@ class BaseSignal(FancySlicing,
 
     def _iterate_signal(self, iterpath=None):
         """Iterates over the signal data. It is faster than using the signal
-        iterator.
+        iterator, because it avoids making deepcopy of metadata and other
+        attributes.
 
         Parameters
         ----------
--- a/hyperspy/signal.py
+++ b/hyperspy/signal.py
@@ -21,6 +21,7 @@ import warnings
 import inspect
 from contextlib import contextmanager
 from datetime import datetime
+from itertools import product
 import logging
 from pint import UnitRegistry, UndefinedUnitError
 from pathlib import Path
@@ -117,7 +118,6 @@ class ModelManager(object):
 
     def _save(self, name, dictionary):
 
-        from itertools import product
         _abc = 'abcdefghijklmnopqrstuvwxyz'
 
         def get_letter(models):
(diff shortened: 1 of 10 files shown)


hyperspy/signal.py
```diff
@@ -3496,37 +3496,32 @@ def _make_sure_data_is_contiguous(self):
                          "for more information.".format(self))
             self.data = np.ascontiguousarray(self.data)
 
-    def _iterate_signal(self):
-        """Iterates over the signal data.
+    def _iterate_signal(self, iterpath=None):
+        """Iterates over the signal data. It is faster than using the signal
+        iterator.
```
thomasaarholt: What is "the signal iterator" that this is faster than? Assuming you mean `for si in s`?
How about `"""Iterates over the signal data. It is faster than iterating over a signal, since it returns the array directly and not a HyperSpy signal.`

thomasaarholt (approved):
Lots of nice things here!

LGTM!

hyperspy/signal.py
ericpre: Actually, it would be useful to explain why it is faster! I don't remember right now and I will need to look it up again.

thomasaarholt: Yes, that was my best guess :p

ericpre: > What is "the signal iterator" that this is faster than?

The `BaseSignal` class has an iterator implementation, i. e. `__iter__` and `__next__` are implemented, see https://docs.python.org/3/library/stdtypes.html#iterator-types for more explanation.

ericpre:
@thomasaarholt, I have improved the docstring, can you please have a look at it again?

thomasaarholt:
Yep, that looks good and is sensible! You can merge when you want :)

PR #2797, opened 2021-07-15
Lazy save improvement

ericpre:
### Progress of the PR
- [x] Add option to close file when saving lazy signal to hspy format,
- [x] add option not to write dataset, which can useful when saving large signal (hspy/zspy format only),
- [x] fix issue with latest `h5py` version (3.5)
- [x] update docstring,
- [x] update user guide,
- [x] add tests,
- [x] ready for review.

```python
import hyperspy.api as hs
import numpy as np

s = hs.signals.Signal1D(np.arange(1000).reshape(10, 100))

fname = 'test.hspy'
s.save(fname, overwrite=True)

s2 = hs.load(fname, lazy=True, mode='a')
s2.axes_manager[-1].scale = 0.1
# write everything, except the dataset itself
s2.save(fname, overwrite=True, write_dataset=False)

s3 = hs.load(fname)
print(s3.axes_manager[-1].scale)
```

### DIFF
--- a/doc/user_guide/io.rst
+++ b/doc/user_guide/io.rst
@@ -461,7 +461,7 @@ numbers as strings if any other strings are present:
 By default, a :py:class:`zarr.storage.NestedDirectoryStore` is used, but other
 zarr store can be used by providing a `zarr store <https://zarr.readthedocs.io/en/stable/api/storage.html>`_
 instead as argument to the :py:meth:`~.signal.BaseSignal.save` or the
-:py:func:`~.io.load` function. If a zspy file have been saved with a different
+:py:func:`~.io.load` function. If a zspy file has been saved with a different
 store, it would need to be loaded by passing a store of the same type:
 
 .. code-block:: python
@@ -471,7 +471,9 @@ store, it would need to be loaded by passing a store of the same type:
     >>> store = zarr.LMDBStore(filename)
     >>> signal.save(store) # saved to LMDB
 
-To load again this file
+To load this file again
+
+.. code-block:: python
 
     >>> import zarr
     >>> filename = 'test.zspy'
@@ -495,16 +497,16 @@ Extra saving arguments
 
         Lazy operations are often i-o bound, reading and writing the data creates a bottle neck in processes
         due to the slow read write speed of many hard disks. In these cases, compressing your data is often
-        beneficial to the speed of some operation. Compression speeds up the process as there is less to
+        beneficial to the speed of some operations. Compression speeds up the process as there is less to
         read/write with the trade off of slightly more computational work on the CPU.
 
 
-- ``chunks``: tuple of interger or None. Define the chunking used for saving
+- ``chunks``: tuple of integer or None. Define the chunking used for saving
   the dataset. If None, calculates chunks for the signal, with preferably at
   least one chunk per signal space.
 - ``close_file``: only relevant for zarr store (``ZipStore``, ``DBMStore``)
-  requiring store to flush data to disk. if ``False``, doesn't close the file
-  after writing. The file  should not be closed if the data need to be accessed
+  requiring store to flush data to disk. If ``False``, doesn't close the file
+  after writing. The file should not be closed if the data need to be accessed
   lazily after saving.
   Default is ``True``.
 - ``write_dataset``: if ``False``, doesn't write the dataset when writing the file.
@@ -1415,7 +1417,7 @@ Nexus metadata and data are stored in Hierarchical Data Format Files (HDF5) with
 a .nxs extension although standards HDF5 extensions are sometimes used.
 Files must use the ``.nxs`` file extension in order to use this io plugin.
 Using the ``.nxs`` extension will default to the Nexus loader. If your file has
-a HDF5 extension, you can also explicitly set the Nexus file reader:
+an HDF5 extension, you can also explicitly set the Nexus file reader:
 
 .. code-block:: python
 
--- a/doc/user_guide/io.rst
+++ b/doc/user_guide/io.rst
@@ -458,38 +458,59 @@ the inner-most structures are converted to numpy arrays when saved. This
 procedure homogenizes any types of the objects inside, most notably casting
 numbers as strings if any other strings are present:
 
+By default, a :py:class:`zarr.storage.NestedDirectoryStore` is used, but other
+zarr store can be used by providing a `zarr store <https://zarr.readthedocs.io/en/stable/api/storage.html>`_
+instead as argument to the :py:meth:`~.signal.BaseSignal.save` or the
+:py:func:`~.io.load` function. If a zspy file have been saved with a different
+store, it would need to be loaded by passing a store of the same type:
+
+.. code-block:: python
+
+    >>> import zarr
+    >>> filename = 'test.zspy'
+    >>> store = zarr.LMDBStore(filename)
+    >>> signal.save(store) # saved to LMDB
+
+To load again this file
+
+    >>> import zarr
+    >>> filename = 'test.zspy'
+    >>> store = zarr.LMDBStore(filename)
+    >>> s = hs.load(store) # load from LMDB
+
 Extra saving arguments
 ^^^^^^^^^^^^^^^^^^^^^^
 
-- ``compressor``: A `Numcodecs Codec <https://numcodecs.readthedocs.io/en/stable/index.html?>`_.
-   A compresssor can be passed to the save function to compress the data efficiently. The defualt
-   is to call a Blosc Compressor object.
+- ``compressor``: `Numcodecs codec <https://numcodecs.readthedocs.io/en/stable/index.html?>`_,
+  a compresssor can be passed to the save function to compress the data efficiently. The default
+  is to call a Blosc compressor object.
 
-.. code-block:: python
+    .. code-block:: python
 
-    >>> from numcodecs import Blosc
-    >>> compressor=Blosc(cname='zstd', clevel=1, shuffle=Blosc.SHUFFLE)
-    >>> s.save('test.zspy', compressor = compressor) # will save with Blosc compression
+        >>> from numcodecs import Blosc
+        >>> compressor=Blosc(cname='zstd', clevel=1, shuffle=Blosc.SHUFFLE) # Used by default
+        >>> s.save('test.zspy', compressor = compressor) # will save with Blosc compression
 
-.. note::
+    .. note::
 
-    Lazy operations are often i-o bound, reading and writing the data creates a bottle neck in processes
-    due to the slow read write speed of many hard disks. In these cases, compressing your data is often
-    beneficial to the speed of some operation. Compression speeds up the process as there is less to
-    read/write with the trade off of slightly more computational work on the CPU."
+        Lazy operations are often i-o bound, reading and writing the data creates a bottle neck in processes
+        due to the slow read write speed of many hard disks. In these cases, compressing your data is often
+        beneficial to the speed of some operation. Compression speeds up the process as there is less to
+        read/write with the trade off of slightly more computational work on the CPU.
 
-- ``write_to_storage``: The write to storage option allows you to pass the path to a directory (or database)
-   and write directly to the storage container.  This gives you access to the `different storage methods
-   <https://zarr.readthedocs.io/en/stable/api/storage.html>`_
-   available through zarr. Namely using a SQL, MongoDB or LMDB database.  Additional downloads may need
-   to be configured to use these features.
 
-.. code-block:: python
+- ``chunks``: tuple of interger or None. Define the chunking used for saving
+  the dataset. If None, calculates chunks for the signal, with preferably at
+  least one chunk per signal space.
+- ``close_file``: only relevant for zarr store (``ZipStore``, ``DBMStore``)
+  requiring store to flush data to disk. if ``False``, doesn't close the file
+  after writing. The file  should not be closed if the data need to be accessed
+  lazily after saving.
+  Default is ``True``.
+- ``write_dataset``: if ``False``, doesn't write the dataset when writing the file.
+  This can be useful to overwrite signal attributes only (for example ``axes_manager``)
+  without having to write the whole dataset, which can take time. Default is ``True``.
 
-    >>>  filename = 'test.zspy/'
-    >>>  os.mkdir('test.zspy')
-    >>>  store = zarr.LMDBStore(path=filename)
-    >>>  signal.save(store.path, write_to_storage=True) # saved to Lmdb
 
 .. _netcdf-format:
 
--- a/doc/user_guide/io.rst
+++ b/doc/user_guide/io.rst
@@ -425,7 +425,7 @@ Extra saving arguments
   the dataset. If None, calculates chunks for the signal, with preferably at
   least one chunk per signal space.
 - ``close_file``: if ``False``, doesn't close the file after writing. The file
-  should not be close if the data need to be accessed lazily after saving.
+  should not be closed if the data need to be accessed lazily after saving.
   Default is ``True``.
 - ``write_dataset``: if ``False``, doesn't write the dataset when writing the file.
   This can be useful to overwrite signal attributes only (for example ``axes_manager``)
@@ -676,7 +676,7 @@ Extra saving arguments
   scalebar. Useful to set formattiong, location, etc. of the scalebar. See the
   `matplotlib-scalebar <https://pypi.org/project/matplotlib-scalebar/>`_
   documentation for more information.
-- ``output_size`` : (int, tuple of length 2 or None, optional): the output size 
+- ``output_size`` : (int, tuple of length 2 or None, optional): the output size
   of the image in pixels:
 
   * if ``int``, defines the width of the image, the height is
@@ -1837,7 +1837,7 @@ Extra loading arguments
   acquired last frame, which typically occurs when the acquisition was
   interrupted. When loading incomplete data (``only_valid_data=False``),
   the missing data are filled with zeros. If ``sum_frames=True``, this argument
-  will be ignored to enforce consistent sum over the mapped area. 
+  will be ignored to enforce consistent sum over the mapped area.
   (default True).
 
 
--- a/doc/user_guide/io.rst
+++ b/doc/user_guide/io.rst
@@ -405,21 +405,31 @@ Extra saving arguments
 - ``compression``: One of ``None``, ``'gzip'``, ``'szip'``, ``'lzf'`` (default is ``'gzip'``).
   ``'szip'`` may be unavailable as it depends on the HDF5 installation including it.
 
-.. note::
+    .. note::
 
-    HyperSpy uses h5py for reading and writing HDF5 files and, therefore, it
-    supports all `compression filters supported by h5py <https://docs.h5py.org/en/stable/high/dataset.html#dataset-compression>`_.
-    The default is ``'gzip'``. It is possible to enable other compression filters
-    such as ``blosc`` by installing e.g. `hdf5plugin <https://github.com/silx-kit/hdf5plugin>`_.
-    However, be aware that loading those files will require installing the package
-    providing the compression filter. If not available an error will be raised.
+        HyperSpy uses h5py for reading and writing HDF5 files and, therefore, it
+        supports all `compression filters supported by h5py <https://docs.h5py.org/en/stable/high/dataset.html#dataset-compression>`_.
+        The default is ``'gzip'``. It is possible to enable other compression filters
+        such as ``blosc`` by installing e.g. `hdf5plugin <https://github.com/silx-kit/hdf5plugin>`_.
+        However, be aware that loading those files will require installing the package
+        providing the compression filter. If not available an error will be raised.
 
-    Compression can significantly increase the saving speed. If file size is not
-    an issue, it can be disabled by setting ``compression=None``. Notice that only
-    ``compression=None`` and ``compression='gzip'`` are available in all platforms,
-    see the `h5py documentation <https://docs.h5py.org/en/stable/faq.html#what-compression-processing-filters-are-supported>`_
-    for more details. Therefore, if you choose any other compression filter for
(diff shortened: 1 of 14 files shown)


doc/user_guide/io.rst
```diff
@@ -401,21 +401,31 @@ Extra saving arguments
...
+        ``compression=None`` and ``compression='gzip'`` are available in all platforms,
+        see the `h5py documentation <https://docs.h5py.org/en/stable/faq.html#what-compression-processing-filters-are-supported>`_
+        for more details. Therefore, if you choose any other compression filter for
+        saving a file, be aware that it may not be possible to load it in some platforms.
+
+- ``chunks``: tuple of interger or None. Define the chunking used for saving
+  the dataset. If None, calculates chunks for the signal, with preferably at
+  least one chunk per signal space.
+- ``close_file``: if ``False``, doesn't close the file after writing. The file
+  should not be close if the data need to be accessed lazily after saving.
```
jlaehne: closed

hyperspy/io.py
```diff
@@ -271,6 +271,17 @@ def load(filenames=None,
...
         default value will change to True in version 2.0.
+    chunks : tuple of integer or None
+        Only for hspy files. Define the chunking used for saving the dataset.
+        If None, calculates chunks for the signal, with preferably at least
+        one chunk per signal space.
+    close_file : bool, optional
+        Only for hspy files. Close the file after writing, default is True.
+    write_dataset : bool, optional
+        Only for hspy files. If True, write the dataset, otherwise, don't
+        write. Useful to save attributes without having the write the whole
```
jlaehne: to write (not the write)

hyperspy/io_plugins/hspy.py
```diff
@@ -769,38 +774,77 @@ def write_signal(signal, group, **kwds):
...
+        The signal to save.
+    chunks : tuple of integer or None
+        Define the chunking used for saving the dataset. If None, calculates
+        chunks for the signal, with preferably at least one chunk per signal
+        space.
+    close_file : bool, optional
+        Close the file after writing. Default is True.
+    write_dataset : bool, optional
+        If True, write the dataset, otherwise, don't write. Useful to save
+        attributes without having the write the whole dataset. Default is True.
```
jlaehne: to write

jlaehne (commented):
some typos

ericpre:
I will come back to this after #2825 is merged.

hyperspy/io_plugins/_hierarchical.py
```diff
@@ -573,59 +571,65 @@ def overwrite_dataset(cls, group, data, key, signal_axes=None,
                     # if the shape or dtype/etc do not match,
                     # we delete the old one and create new in the next loop run
                     del group[key]
-        if dset == data:
```
ericpre, line 576: This is never used and it breaks with latest version of `h5py` (3.5).

hyperspy/io_plugins/zspy.py
```diff
@@ -111,59 +112,44 @@ def _get_object_dset(group, data, key, chunks, **kwds):
...
     ----------
     filename: str
     signal: a BaseSignal instance
-    *args, optional
     **kwds, optional
     """
     if "compressor" not in kwds:
         from numcodecs import Blosc
         kwds["compressor"] = Blosc(cname='zstd', clevel=1)
-    if ("write_to_storage" in kwds and kwds["write_to_storage"]) or isinstance(filename, MutableMapping):
```
ericpre, line 150: This was actually not working as expected when opening a file which has been saved with a store different than the `NestedDirectoryStore` or `DirectoryStore`, because `zarr` is initialising the correct store based on the extension (https://github.com/zarr-developers/zarr-python/blob/523dbb85cbad5a5fa3b74a2ec81ec7ec63123b64/zarr/creation.py#L160-L178). For this reason, I remove it here.

hyperspy/io_plugins/zspy.py
```diff
@@ -111,59 +112,44 @@ def _get_object_dset(group, data, key, chunks, **kwds):
...
                 data = data.rechunk(dset.chunks)
-            path = group._store.dir_path() + "/" + dset.path
-            data.to_zarr(url=path,
-                         overwrite=True,
-                            **kwds)  # add in compression etc
-        elif data.dtype == np.dtype('O'):
-            group[key][:] = data[:]  # check lazy
+            da.store(data, dset)
         else:
-            path = group._store.dir_path() + "/" + dset.path
```
ericpre, line 125: Using `zarr` group private attribute is asking for trouble!
@CSSFrancis, in #2825, you mentioned that this was necessary in some cases? I have extended the test coverage and this seems to be working fine and I think that this use of the API is supported, so if this doesn't work, it could be a bug somewhere else.
Can you please check?

ericpre (commented):
The test suite is failing on python 3.6 and I think that it is not worth trying to get it to work because python 3.6 is not supported anymore and we should drop it in the next minor release.

ericpre:
@CSSFrancis, would you like to review this PR? Thanks!

hyperspy/io_plugins/_hierarchical.py
CSSFrancis: That is good to know.  I figured that was left over from previous version but it wasn't tested so I didn't know if there was a case where it was important.

CSSFrancis:
> @CSSFrancis, would you like to review this PR? Thanks!

Yea let me see if I can get to this today or tomorrow

hyperspy/io_plugins/zspy.py
```diff
@@ -111,59 +112,44 @@ def _get_object_dset(group, data, key, chunks, **kwds):
...
         if isinstance(data, da.Array):
             if data.chunks != dset.chunks:
                 data = data.rechunk(dset.chunks)
-            path = group._store.dir_path() + "/" + dset.path
-            data.to_zarr(url=path,
-                         overwrite=True,
-                            **kwds)  # add in compression etc
-        elif data.dtype == np.dtype('O'):
-            group[key][:] = data[:]  # check lazy
+            da.store(data, dset)
```
CSSFrancis:
```suggestion
            da.store(data, dset, lock=False)
```

hyperspy/io_plugins/zspy.py
CSSFrancis: So the issue arises when using a dask.distributed scheduler. The solution is to set lock = False.  I didn't realize this when I originally wrote the code.  It might be worth writing tests in the future to test dask.distrubted compatibility if we are planning to fully support it (Which I would recommend)

CSSFrancis: It seems like the pickling is intending to be thread safe so it throws an error if lock is set to True

hyperspy/io_plugins/zspy.py
CSSFrancis: Ahh, very true.  Removing the write to storage argument should be fine.

hyperspy/tests/io/test_zspy.py
```diff
@@ -35,20 +35,6 @@ def signal(self):
         s = Signal1D(data)
         return s
 
-    def test_save_N5_type(self, signal, tmp_path):
-        filename = tmp_path / 'testmodels.zspy'
-        store = zarr.N5Store(path=filename)
-        signal.save(store, write_to_storage=True)
-        signal2 = load(filename)
-        np.testing.assert_array_equal(signal2.data, signal.data)
-
```
CSSFrancis: This should not be deleted I think.  You can remove the `write_to_storage=True` but this test shows if passing a MutableMapping object works for saving the data.

CSSFrancis (changes requested):
@ericpre Looks like this cleans up a lot of the little things that I wasn't terribly confident about and reduces the code complexity a fair bit! Thanks for doing this. 

One note (as I said above) is that the `store` function doesn't play well with dask.distrubted unless lock=False.  It might be worth adding in some tests for compatibility there if the goal is to fully integrate with dask, but I am not sure how easy deploying something like that on a testing environment is. 

I assume that replacing `create_group` with `require_group` is preferable because if won't throw an error if the group already exists?

Other than that, one of the tests which tested passing a Mutable Mapping object as the file name was deleted and should be covered.

ericpre:
Thanks for the review!

> One note (as I said above) is that the store function doesn't play well with dask.distrubted unless lock=False. It might be worth adding in some tests for compatibility there if the goal is to fully integrate with dask, but I am not sure how easy deploying something like that on a testing environment is.

Yes, this is what dask is doing too: https://github.com/dask/dask/blob/62f96e7edb27cb94eca49d5ac2f5639fb8cca44a/dask/array/core.py#L3388

> I assume that replacing create_group with require_group is preferable because if won't throw an error if the group already exists?

Yes, this is correct, same for `h5py` and [`zarr`](https://zarr.readthedocs.io/en/stable/api/hierarchy.html?highlight=require_group#zarr.hierarchy.Group.require_group).

> Other than that, one of the tests which tested passing a Mutable Mapping object as the file name was deleted and should be covered.

As mentioned in https://github.com/hyperspy/hyperspy/pull/2797#discussion_r733619173, this was not working correctly, but I had a look at it again, one way to get it to work is to pass store of the same in the `load` function.

It should be all good now!

hyperspy/io_plugins/zspy.py
ericpre: > So the issue arises when using a dask.distributed scheduler. The solution is to set lock = False. I didn't realize this when I originally wrote the code. It might be worth writing tests in the future to test dask.distrubted compatibility if we are planning to fully support it (Which I would recommend)

Yes, it would be good to have some tests using `dask.distributed.Client`. I don't know how well it will play in CI and I had a quick look at how dask are testing it, and I didn't understood straightaway, so I would leave it for another PR! :)

hyperspy/io_plugins/hspy.py
```diff
@@ -193,45 +188,86 @@ def file_reader(
...
+        writer.write()
+    except BaseException:
+        raise
+    finally:
+        del smd.record_by
+
+    if close_file:
+        f.close()
+
+overwrite_dataset = HyperspyWriter.overwrite_dataset
```
ericpre, line 267: `overwrite_dataset` is actually part of the public API, so I put back here... it was breaking kikuchipy, as pointed out by https://github.com/hyperspy/hyperspy/runs/3973795422?check_suite_focus=true!

ericpre:
Running the integration test suite show that it was breaking the public API... see https://github.com/hyperspy/hyperspy/pull/2797#discussion_r734456427. This is fixed in caf6a01.

@CSSFrancis, does it look good to you?

CSSFrancis:
> @CSSFrancis, does it look good to you?

Yea with the changes it looks good! Thanks for cleaning this stuff up.

hyperspy/io_plugins/hspy.py
hakonanes: Very good catch, thanks a lot.

ericpre:
@jlaehne, do you want to do another review of this PR?

hyperspy/io.py
```diff
@@ -277,6 +288,17 @@ def load(filenames=None,
         acquisition stopped before the end: if True, load only the acquired
         data. If False, fill empty data with zeros. Default is False and this
         default value will change to True in version 2.0.
+    chunks : tuple of integer or None
+        Only for hspy files. Define the chunking used for saving the dataset.
```
jlaehne: In the user guide it says these extra arguments are only relevant for `zarr` format, here it says they are only relevant for `hspy` format. Also some details in the description differ between here and io.rst.

jlaehne:
> @jlaehne, do you want to do another review of this PR?

I sent a PR correcting some typos. There are still some codecov warnings (mostly about untested exceptions). Otherwise nothing that caught my eye.

hyperspy/io_plugins/_hierarchical.py
```diff
@@ -22,14 +22,62 @@
...
+
+def get_signal_chunks(shape, dtype, signal_axes=None, target_size=1e6):
+    """
+    Function that calculates chunks for the signal, preferably at least one
+    chunk per signal space.
+
+    Parameters
+    ----------
+    shape : tuple
+        the shape of the dataset to be sored / chunked
```
jlaehne: sorted or stored ?

hyperspy/io.py
ericpre: Yes, this docstring is for `save`, not `load`! :facepalm:

ericpre:
Thanks @jlaehne, I increased the coverage and fixed the docstring.

doc/user_guide/io.rst
```diff
@@ -405,21 +405,31 @@ Extra saving arguments
...
+        providing the compression filter. If not available an error will be raised.
+
+        Compression can significantly increase the saving speed. If file size is not
+        an issue, it can be disabled by setting ``compression=None``. Notice that only
+        ``compression=None`` and ``compression='gzip'`` are available in all platforms,
+        see the `h5py documentation <https://docs.h5py.org/en/stable/faq.html#what-compression-processing-filters-are-supported>`_
+        for more details. Therefore, if you choose any other compression filter for
+        saving a file, be aware that it may not be possible to load it in some platforms.
+
+- ``chunks``: tuple of interger or None. Define the chunking used for saving
```
magnunor: Typo: should be integer

hyperspy/io_plugins/_hierarchical.py
```diff
@@ -22,25 +23,82 @@
...
+    Function that calculates chunks for the signal, preferably at least one
+    chunk per signal space.
+
+    Parameters
+    ----------
+    shape : tuple
+        The shape of the dataset to be stored / chunked.
+    dtype : {dtype, string}
+        The numpy dtype of the data.
+    signal_axes: {None, iterable of ints}
```
magnunor: signal_axes : {None, iterable of ints}

hyperspy/io_plugins/_hierarchical.py
```diff
@@ -22,25 +23,82 @@
...
+        The axes defining "signal space" of the dataset. If None, the default
+        h5py chunking is performed.
+    target_size: int
```
magnunor: target_size : int

hyperspy/io_plugins/_hierarchical.py
```diff
@@ -405,39 +497,39 @@ def group2dict(self,
...
 class HierarchicalWriter:
-    """An object used to simplify and orgainize the process for
-    writing a Hierachical signal.  (.hspy format)
     """
-    def __init__(self,
-                 file,
-                 signal,
-                 group,
-                 **kwds):
+    An object used to simplify and orgainize the process for writing a
```
magnunor:
```suggestion
    An object used to simplify and organize the process for writing a
```

hyperspy/io_plugins/_hierarchical.py
```diff
@@ -405,39 +497,39 @@ def group2dict(self,
...
+    Hierachical signal, such as hspy/zspy format.
```
magnunor:
```suggestion
    Hierarchical signal, such as hspy/zspy format.
```

doc/user_guide/io.rst
```diff
@@ -448,38 +458,61 @@ the inner-most structures are converted to numpy arrays when saved. This
...
+    >>> s = hs.load(store) # load from LMDB
+
 Extra saving arguments
 ^^^^^^^^^^^^^^^^^^^^^^
 
-- ``compressor``: A `Numcodecs Codec <https://numcodecs.readthedocs.io/en/stable/index.html?>`_.
-   A compresssor can be passed to the save function to compress the data efficiently. The defualt
-   is to call a Blosc Compressor object.
+- ``compressor``: `Numcodecs codec <https://numcodecs.readthedocs.io/en/stable/index.html?>`_,
+  a compresssor can be passed to the save function to compress the data efficiently. The default
```
magnunor:
```suggestion
  a compressor can be passed to the save function to compress the data efficiently. The default
```

magnunor (commented):
I tested this on my own computer, and I'm getting an error:

```python
import dask.array as da
import hyperspy.api as hs

s = hs.signals.Signal1D(da.random.random((10, 20, 50))).as_lazy()
s.save("test_dataset.hspy")
```
Giving the error:
```python
hyperspy/io_plugins/_hierarchical.py in overwrite_dataset(cls, group, data, key, signal_axes, chunks, **kwds)
    574                     # we delete the old one and create new in the next loop run
    575                     del group[key]
--> 576         if dset == data:
    577             # just a reference to already created thing
    578             pass

ValueError: The truth value of an array with more than one element is ambiguous. Use a.any() or a.all()
```

I do have a bit of a strange mix of package versions, so it could be caused by that.

ericpre:
> I tested this on my own computer, and I'm getting an error:
> 
> ```python
> import dask.array as da
> import hyperspy.api as hs
> 
> s = hs.signals.Signal1D(da.random.random((10, 20, 50))).as_lazy()
> s.save("test_dataset.hspy")
> ```
> 
> Giving the error:
> 
> ```python
> hyperspy/io_plugins/_hierarchical.py in overwrite_dataset(cls, group, data, key, signal_axes, chunks, **kwds)
>     574                     # we delete the old one and create new in the next loop run
>     575                     del group[key]
> --> 576         if dset == data:
>     577             # just a reference to already created thing
>     578             pass
> 
> ValueError: The truth value of an array with more than one element is ambiguous. Use a.any() or a.all()
> ```
> 
> I do have a bit of a strange mix of package versions, so it could be caused by that.

Thanks @magnunor! It seems that you are on a wrong branch, the code at line 576 is different in this branch:
https://github.com/ericpre/hyperspy/blob/9c5bf6fa1550867eee8a586ad4a8d163d38f4f2a/hyperspy/io_plugins/_hierarchical.py#L576

magnunor:
> It seems that you are on a wrong branch, the code at line 576 is different in this branch:

Indeed! Seems like `git` had changed how it handled pulling branches from a remote, so this,

```
git checkout -b ericpre-lazy_save_improvement RELEASE_next_minor
git pull https://github.com/ericpre/hyperspy.git lazy_save_improvement
```
failed.

hyperspy/io_plugins/_hierarchical.py
```diff
@@ -22,25 +23,82 @@
...
     """A generic Reader class for reading data from hierarchical file types."""
-    def __init__(self,
-                 file):
-        """ Initalizes a general reader for hierarchical signals.
 
-        Parmeters
-        ---------
+    def __init__(self, file):
+        """
+        Initalizes a general reader for hierarchical signals.
```
magnunor:
```suggestion
        Initializes a general reader for hierarchical signals.
```

hyperspy/io_plugins/hspy.py
```diff
@@ -113,28 +116,22 @@ def __init__(self, file):
 
 
 class HyperspyWriter(HierarchicalWriter):
-    """An object used to simplify and orgainize the process for
+    """
+    An object used to simplify and orgainize the process for
```
magnunor:
```suggestion
    An object used to simplify and organize the process for
```

doc/user_guide/io.rst
```diff
@@ -405,21 +405,31 @@ Extra saving arguments
...
+    .. note::
+
+        HyperSpy uses h5py for reading and writing HDF5 files and, therefore, it
+        supports all `compression filters supported by h5py <https://docs.h5py.org/en/stable/high/dataset.html#dataset-compression>`_.
+        The default is ``'gzip'``. It is possible to enable other compression filters
+        such as ``blosc`` by installing e.g. `hdf5plugin <https://github.com/silx-kit/hdf5plugin>`_.
+        However, be aware that loading those files will require installing the package
+        providing the compression filter. If not available an error will be raised.
+
+        Compression can significantly increase the saving speed. If file size is not
```
magnunor, line 417: Is this sentence correct? Or at least ambiguous?

I'm guessing it should say that using compression can cause file saving and loading to be much slower.

magnunor (commented):
I got the previous issue sorted out, and tested the functionality a little bit. Some observations:

- `write_dataset=False` for updating both `hspy` and `zspy` files works fine, for both `axes_manager` and `metadata`
- `zspy` does not require loading the data with any specific `mode`. As opposed to `hspy`, which requires `mode="a"`
- The filename can be dropped when doing `s.save(overwrite=True, write_dataset=False)`
- Loading via `zspy` is much faster, both using `DirectoryStore` and `LMDBStore`. For a 512 x 512 x 256 x 256 dataset, with uint16, it took about 20 seconds. The same dataset took about 100 seconds with `hspy`

This is probably unrelated, but doing:

```python
s = hs.load("test_data.hspy", lazy=True)
s.compute()
```
used two times as much memory (64 GB), as opposed to `s = hs.load("test_data.hspy", lazy=False)`, which used 32 GB.

This is using the most recent version of `dask`, `2021.10.0`

doc/user_guide/io.rst
ericpre: Yes, this is correct in many cases when the IO time is balanced against with the CPU time - most of the time, CPU are fast enough and compressor are efficient enough.

ericpre:
> This is probably unrelated, but doing:
> 
> ```python
> s = hs.load("test_data.hspy", lazy=True)
> s.compute()
> ```
> 
> used two times as much memory (64 GB), as opposed to `s = hs.load("test_data.hspy", lazy=False)`, which used 32 GB.
> 
> This is using the most recent version of `dask`, `2021.10.0`

Yes, this is not related to this PR - at least I don't see how it can be and I would expect that the same happen with other branches. Do you observe the same memory usage with `zspy`?

ericpre:
Thanks @CSSFrancis, @jlaehne and @magnunor for the reviews, I will merge this PR to be able to rebase https://github.com/hyperspy/hyperspy/pull/2839 and fix CI, so that it is easier to work on #2842 and co.

magnunor:
> Yes, this is not related to this PR - at least I don't see how it can be and I would expect that the same happen with other branches. Do you observe the same memory usage with zspy?

Yes, this was also with `zspy`.

Issue #2784, opened 2021-06-28
Saving and Loading Large Datasets


CSSFrancis:
As more users are starting to work with larger datasets (specifically datasets which even one copy doesn't fit into RAM) and operating on clusters with multiple cores there appears to be a couple of fairly severe bottlenecks with regards to the `hdf5` file format.

# Problems with HDF5
Specifically the main issues with `h5py` (and hdf5) as best I can summarize:

- Compression of the data locks up the GIL/ operates on one Core (and is very slow)
    - Any h5py operation locks up the GIL 
- Saving when using a multiprocessing/distrubted dask scheduler throws an error https://github.com/dask/dask/issues/3074
    - There doesn't seem to be a perscribed solution to this other than using `zarr`
    - H5py parallel support is offered but it seems to be mostly an mpi implementation which seem tricky to set up correctly and isn't very optimized https://docs.h5py.org/en/latest/mpi.html 

All of these things severely limit things like parallel operations in dask where i/o is a large overhead. 

# Solutions

- Adding in the ability to save/ load large datasets in the [`zarr`](https://zarr.readthedocs.io/en/stable/)  file format rather than `hdf5`.
    - `zarr` follows much of the same structure as hdf5 but is designed for parallel operations and fast compression/ decompression of chunks to help speed up i/o operations. 
   - Maybe throw a warning if a user is saving a lazy object over 15 GB that `zarr` might work better than h5py

- Otherwise looking into the parallel h5py support might be rewarding but I haven't found a clear way to implement that yet.

I wouldn't suggest that `zarr` replace the `hdf5` format we are currently using, rather that for larger data type applications we give the ability to use a filetype that might better suit each users needs.

# Extra Information
- Comparison of `h5py` with `zarr` -->https://alimanfoo.github.io/2016/05/16/cpu-blues.html
- Talk on `hdf5` vs `zarr` https://www.youtube.com/watch?v=-l445lCPTts

Based on a conversation here: https://github.com/jacobjma/abTEM/issues/27

Just pinging some people who might have additional thoughts...
@magnunor @pc494 @din14970 @jacobjma @ericpre

din14970:
[...]

PR #2798, opened 2021-07-18
Zarr Reading and Writing (.zspy format)

CSSFrancis:
### Description of the change
This change allows for files to be written as zarr files and the read from zarr files.  This helps to provide a different option for #2784

### What `zarr` allows for:
- i-o operations with `dask-distributed` (not supported with hdf5 format)
- Support for `zarr` files
- Compression/decompression without locking up GIL
- Many different compression options 

### Progress of the PR
- [x] Added `zarr` i-o plugin
- [x] Add in tests
- [x] Added in a method for displaying lazy data (maybe should go in a separate PR)
- [x] Clean up code for `zarr` plugin
- [x] update user guide --> Add in guide for using dask-distributed and hyperspy
- [x] ready for review.

```python
import hyperspy.api as hs
import numpy as np
s.save("signal.zspy)

hs.load("signal.zspy)
```

### Additional Thoughts:
For the most part I just ended up copying the code for the `hspy` format and then replaced/ changed it to allow for loading from `zarr` in almost the same way.  There are a couple of places where bigger changes had to occur.  For the most part this should be okay, even if the code isn't the prettiest.

That being said It would be nice is someone more familiar with the `hspy` i-o plugin looked over the code to make sure I didn't delete too much of the legacy-type code.


CSSFrancis:
I tested just loading a random dataset.  Obviously the compression isn't very good with this dateset but there are some key take away.

- Saving the data is ***much much*** faster (maybe I'm not setting up the compression in hdf5 appropriately)
- Loading the data/ preforming an operation is about 10x faster 
- Ability to run in parallel is greatly improved.
-
```python
import hyperspy.api as hs
import numpy as np
import dask.array as da
from numcodecs import Blosc
s = da.random.random((1000,1000,100))
s = hs.signals.Signal2D(s).as_lazy()
display(s)
```
```python
%%time
#only once because it is very slow
s.save("test_hs_comp.hspy")
```
CPU times: user 23.6 s, sys: 1.22 s, total: 24.9 s
Wall time: 3min 1s
```python
%%timeit
s.save("test_hs_comp.zarr", compressor=Blosc(cname='zstd', clevel=1, shuffle=Blosc.SHUFFLE))
```
883 ms ± 77.8 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
```python
%%timeit
s = hs.load("test_hs_comp.hspy")
```
3.17 s ± 128 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
```python
%%timeit
s = hs.load("test_hs_comp.zarr")
```
644 ms ± 13 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
```python
hdf5_data = hs.load("test_hs_comp.hspy", lazy=True)
zarr_data = hs.load("test_hs_comp.zarr", lazy=True)
```
```python
%%timeit
hdf5_data.sum(axis=1).compute()
```
3.41 s ± 45.1 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
```python
%%timeit
zarr_data.sum(axis=1).compute()
```
216 ms ± 3.65 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)

CSSFrancis:
> Added in a method for displaying lazy data (maybe should go in a separate PR)

In reference to this point I just copied most of what dask does when displaying data and changed it so that it aligns more with the hyperspy signal axes/ navigation axes.

For example:
`display(signal)`
image

CSSFrancis:
@ericpre would you mind taking a look at this when you have the chance?

ericpre:
This looks great. The only thing that I am not comfortable with is to create a new specification format... and this is the reason why, in https://github.com/ericpre/hyperspy/tree/zarr_backend, I have used zarr directory store as a temporary storage - and also because it was quick to test! In any case, even if I would be reluctant to create a new specification, the advantages of the zarr format outweigh this issue.
I don't guarantee that I will have time to look at it very soon, but in the next few weeks, it should be fine!

magnunor:
Very nice! I'll give this a try with one of my experimental datasets, and see how it performs.

magnunor:
I skimmed through the code, and should `hyperspy/tests/signal/test_mask` be in the pull request?

magnunor:
## Some comments and questions

- Reading through the documenation for `zarr`, I see that the "folder" output is the standard way of saving the data. Are there other ways of saving the data, which puts it into a single file?
- Should `.zspy` have compression on by default? Users might expect this, since this is how `.hspy` works

## A little bit of benchmarking

Dataset: 256 x 256 x 256 x 256, magnetic STEM-DPC, 8.6 GB uncompressed

### Saving

- Saving, with the dataset already in memory:
-- `zspy`: 9.29 seconds. 1.1 GB output file
---  `s.save("000_test_data_dpc.zspy", chunks=(32, 32, 32, 32), compressor=Blosc(cname='zstd', clevel=1, shuffle=Blosc.SHUFFLE))`
-- `hspy`: 83.94 seconds. 1.0 GB output file
--- `s.save("000_test_data_dpc.hspy", chunks=(32, 32, 32, 32))`

### Loading

- Loading the files which were saved above
-- Loading zspy: 5.77 seconds
-- Loading hspy: 24.29 seconds

### Center of mass

- Loading the files which were saved above
-- Center of mass, zspy: 8.03 seconds
-- Center of mass, hspy: 47.0 seconds

So very nice improvement!

CSSFrancis:
@magnunor  Thanks for looking at this!

> I skimmed through the code, and should `hyperspy/tests/signal/test_mask` be in the pull request?

Yea I'm not sure how that got included.  Thanks for picking up on that.

> ## Some comments and questions
> * Reading through the documentation for `zarr`, I see that the "folder" output is the standard way of saving the data. Are there other ways of saving the data, which puts it into a single file?

Yea there are a couple but most of them leverage some type of database such as MongoDB, SQL, or LMDB.  This is mostly to support multi-threading/ multiprocessing.   They all have additional download requirements, require you to handle opening/closing which is why I took the default approach. There is also the option to save data as a zipped file, but that seems less than ideal for large data sets, especially when you can always just manually zip the dataset afterwards.

Support for the different storage types might be of interest as there is support for Distributed/cloud storage and some of the databases might have some additional optimization for i-o. I don't think that isn't a great way to do this easily for all of the options though as each of the storage types is slightly different.

> * Should `.zspy` have compression on by default? Users might expect this, since this is how `.hspy` works

Yea I think it should have compression on by default.  I've added that in now so you won't have to mess with Numcodecs. That should simplify things a little bit.

hyperspy/_signals/lazy.py
```diff
@@ -168,6 +170,83 @@ def rechunk(self,
                                                                   **kwargs)
                                                 )
 
+    def _repr_html_(self):
+        table = self._repr_html_table()
```
ericpre, line 174: This is nice, but I think that it should be done in a separate PR.
I suspect that it could be simplified using https://github.com/dask/dask/pull/8019.

hyperspy/_signals/lazy.py
```diff
@@ -268,6 +347,11 @@ def _get_navigation_chunk_size(self):
         nav_chunks = tuple([self.data.chunks[i] for i in sorted(nav_axes)])
         return nav_chunks
 
+    def _get_signal_chunk_size(self):
+        sig_axes = self.axes_manager.signal_indices_in_array
+        sig_chunks = tuple([self.data.chunks[i] for i in sorted(sig_axes)])
```
ericpre, line 352: is the sorted, necessary here? `axes_manager.signal_indices_in_array` could be correct?

conda_environment.yml
```diff
@@ -4,6 +4,7 @@ channels:
 dependencies:
 - dask-core >2.0
 - dill
+- fsspec
```
ericpre, line 7: Is it necessary to specify fsspec?

ericpre (commented):
It looks to me that it should be possible to reuse the `hspy` reader by adding a `backend` parameter, along with a dictionary mapping and a few if/else block where necessary. the `zspy` reader could import the functions from the `hspy` reader and specify the `zarr` backend as argument of these functions (`hdfgroup2dict`, `hdfgroup2signaldict`, etc.).
As you mentioned, there is very little difference between the `hspy` and `zspy` reader and from a quick comparison between the two, I couldn't see reason why it would not possible to easily reuse the `hspy`. The API of zarr is very similar to the `h5py` one and it would be a shame not to benefit from it here!

hyperspy/io_plugins/zspy.py
```diff
@@ -0,0 +1,575 @@
...
+    ----------
+    filename: str
+    signal: a BaseSignal instance
+    *args, optional
+    **kwds, optional
+    """
+    if "compressor" not in kwds:
+        from numcodecs import Blosc
+        kwds["compressor"] = Blosc(cname='zstd', clevel=1)
+    store = zarr.storage.NestedDirectoryStore(filename,)
```
ericpre, line 548: Add a keyword argument to allow the user to specify the storage class to use. Also maybe a dictionary mapping string to commonly use storage, for example, `directory`, `nested_directory`, `zip`?

ericpre:
> > ## Some comments and questions
> > 
> > * Reading through the documentation for `zarr`, I see that the "folder" output is the standard way of saving the data. Are there other ways of saving the data, which puts it into a single file?
> 
> Yea there are a couple but most of them leverage some type of database such as MongoDB, SQL, or LMDB. This is mostly to support multi-threading/ multiprocessing. They all have additional download requirements, require you to handle opening/closing which is why I took the default approach. There is also the option to save data as a zipped file, but that seems less than ideal for large data sets, especially when you can always just manually zip the dataset afterwards.

From what I remember with using the `ZipStore` is that it was not performing as well and according to the documentation 
ZipStore it supports multithreading but not multiprocessing: https://zarr.readthedocs.io/en/stable/api/storage.html#zarr.storage.ZipStore

conda_environment.yml
CSSFrancis: I don't think it is... but some of the tests were failing because it wasn't included.  I don't know enough about how the tests work but it seemed to help some of them.

hyperspy/_signals/lazy.py
CSSFrancis: Thanks for pointing this out!  I'll add this in as a different PR and maybe see if there are other places where better html representation can be added into hyperspy.

CSSFrancis:
@ericpre 
> It looks to me that it should be possible to reuse the `hspy` reader by adding a `backend` parameter, along with a dictionary mapping and a few if/else block where necessary. the `zspy` reader could import the functions from the `hspy` reader and specify the `zarr` backend as argument of these functions (`hdfgroup2dict`, `hdfgroup2signaldict`, etc.).
> As you mentioned, there is very little difference between the `hspy` and `zspy` reader and from a quick comparison between the two, I couldn't see reason why it would not possible to easily reuse the `hspy`. The API of zarr is very similar to the `h5py` one and it would be a shame not to benefit from it here!

Thanks for the insight! I must admit that in the beginning I didn't really have a great handle on the `hspy` reader which is why I was hesitant to touch it. I'll work on reducing the code duplication, hopefully that will reduce overhead in the future with additional changes to the `hspy` file format and make things quite a bit cleaner.

ericpre:
Yes, it makes sense and it seems that you have sorted what is necessary to get it to work so you have done the difficult part! In this situation, I would simply re-start with a new branch and use this one as a reference to know where the hspy reader need to be adapted to add a zarr backend. There will still be a separate zspy reader and this reader should be very simple because it will only use a few functions of the hspy reader with the appropriate parameters.

CSSFrancis:
> s, it makes sense and it seems that you have sorted what is necessary to get it to work so you have done the difficult part! In this situation, I would simply re-start with a new branch and use this one as a reference to know where the hspy reader need to be adapted to add a zarr backend. There will still be a separate zspy reader and this reader should be very simple because it will only use a few functions of the hspy reader with the appropriate parameters.

@ericpre  I've been looking into this and the easiest way to do this would probably to make a hyperspyReader class.  Then I could just extend the class and override the methods that need to be changed.  I can include a bunch of if statements in the hspy reader code, but I'm a little worried that might complicate a rather central piece of code in hyperspy. 

Let me know what you think, or if you know of a different way to handle this drop in

ericpre:
OK, I would expect that both approach will work but if you prefer to refactor to a class, then that's fine! 
For the functional approach, you could use a dictionary to make a mapping specific for each file format instead of using `if` block - some may not be replaceable.
For example:
```python
import h5py
import zarr

mapping_hdf5 = dict(format='hdf5', group=h5py.Group, dataset=h5py.Dataset)
mapping_zarr= dict(format='zarr', group=zarr.Group, dataset=zarr.Array)
```
and then use these mapping where necessary. There are recursive calls, and I am not sure if `functools.partial` can be used to preset the mapping and avoid having pass it along. If not, then this could be a reason why a class approach would be more readable and maintainable?!

ericpre:
When reviewing https://github.com/hyperspy/hyperspy/pull/2568 and trying to figure out what would be a good caching approach for our visualisation needs , I tried the zspy format of this breanch with some caching https://github.com/zarr-developers/zarr-python/pull/306 and this works well. We are hitting some dask overhead in the 10 ms range when using many chunks (using this [dataset size and chunking](https://github.com/hyperspy/hyperspy/pull/2568#issuecomment-784315898)). Using zarr slicing without dask is in the 0.1 ms, which is brilliant because the plotting is fast but `isig` and `inav` are also very fast, which means that the ROIs are very fast too!

CSSFrancis:
Code is reformatted here #2825.

PR #2803, opened 2021-07-29
Add 'de5' as a known extension of the emd format

TomSlater:
### Description of the change
Direct Electron have a new file format from their cameras that broadly matches Berkley's emd file format. You can currently use the emd reader to load these 'de5' files. This PR adds 'de5' as a known extension to the emd reader so that you don't have to specify the file reader.

### Progress of the PR
- [ ] add a changelog entry in the `upcoming_changes` folder (see [`upcoming_changes/README.rst`](https://github.com/hyperspy/hyperspy/blob/RELEASE_next_minor/upcoming_changes/README.rst)),
- [ ] add tests,
- [ ] ready for review.

```python
import hyperspy.api as hs
s = hs.load('example_file.de5')
```


ericpre (commented):
Would it be possible to know what version of emd file is this format using? Do you have a example of such file, which could be added to the test suite?

It would be good to rebase, to remove the first 4 commits.

TomSlater:
I don't have an example yet, but I'll generate a minimal one for testing as soon as I can.

jlaehne:
I guess this has to be rebased on and adapted to #2972?

ericpre:
@TomSlater, do you have a set of small example file to add to the test suite?

francisco-dlp:
One wonders what's the point of following a standard and then changing the file extension. Rather than adding support for the unstandard EMD file extension to HyperSpy, shouldn't users request to Direct Electron to use the standard extension?

TomSlater:
Sorry, I forgot about this.

I did take some data but I didn't take it with me when moving. The detector is due to be removed from ePSIC (or maybe already has been) so I'd suggest that there's now little drive to implement this.

I also agree with Francisco that it would make sense to request that Direct Electron use a .emd extension if their format meets the standards for EMD.

In short, if no one has any objections I will close this PR and suggest to Direct Electron that they allow users to save with a .emd extension.

jlaehne:
> One wonders what's the point of following a standard and then changing the file extension. Rather than adding support for the unstandard EMD file extension to HyperSpy, shouldn't users request to Direct Electron to use the standard extension?

Even if they would implement it ... many people would still be using the old software version and end up with a different extension, I guess?

ericpre:
> One wonders what's the point of following a standard and then changing the file extension. Rather than adding support for the unstandard EMD file extension to HyperSpy, shouldn't users request to Direct Electron to use the standard extension?

Yes, this is a different approach, which would simply need documentation, assuming that the following works:
```python
s = hs.load('example_file.de5', reader='emd')
```

However, https://emdatasets.com says that "EMD files typically have a “.emd” extension, though they may also have “.h5″ or other extensions.", which means that this complies with the `emd` file format specification.

francisco-dlp:
Well, if the format is so flexible regarding the file extension, then I think that it is fine to merge this. As for test files, if it is EMD compliant, then we don't need them, isn't it?

ericpre:
A test would still be good to check that to the `emd` plugins is used to read `de5` file, but this can be done by writing/reading the a file temporarily.

To get things to move along swiftly, I am volunteering myself! ;)

francisco-dlp:
Are you sure that that's necessary? The extension to reader matching is tested for every single file, so there is no reason why it wouldn't work for this one. Still, an extra test never hurts ;)

jlaehne:
Has been contributed to #2972 in https://github.com/francisco-dlp/hyperspy/pull/49

PR #2747, opened 2021-05-18
Fix parse value string nua and fix circular import

ericpre:
Merging #2386 into the `non_uniform_axes` wasn't done correctly.

Moving imports to the top of `artificial_data.py` introduces a circular import, for example:
```python
from hyperspy.signals import Signal1D
```
gives the following circular `ImportError`:
```python
Traceback (most recent call last):

  File "<ipython-input-1-d9a231321fff>", line 1, in <module>
    from hyperspy.signals import Signal1D

  File "/home/eric/Dev/hyperspy/hyperspy/signals.py", line 67, in <module>
    importlib.import_module(

  File "/opt/miniconda3/lib/python3.8/importlib/__init__.py", line 127, in import_module
    return _bootstrap._gcd_import(name[level:], package, level)

  File "/home/eric/Dev/hyperspy/hyperspy/_signals/signal2d.py", line 46, in <module>
    from hyperspy.utils.peakfinders2D import (

  File "/home/eric/Dev/hyperspy/hyperspy/utils/__init__.py", line 39, in <module>
    import hyperspy.datasets.example_signals

  File "/home/eric/Dev/hyperspy/hyperspy/datasets/__init__.py", line 21, in <module>
    from hyperspy.datasets import artificial_data

  File "/home/eric/Dev/hyperspy/hyperspy/datasets/artificial_data.py", line 10, in <module>
    from hyperspy.signals import Signal1D,Signal2D,EELSSpectrum

ImportError: cannot import name 'Signal2D' from partially initialized module 'hyperspy.signals' (most likely due to a circular import) (/home/eric/Dev/hyperspy/hyperspy/signals.py)
```

### Progress of the PR
- [x] Fix circular import and remove unnecessary import,
- [x] fix relative slicing, 
- [x] raise `ValueError` when converting units with non-uniform axis, 
- [x] add tests,
- [x] ready for review.

### Minimal example of the bug fix or the new feature
```python
import hyperspy.api as hs
s = hs.datasets.artificial_data.get_luminescence_signal(2)
s2 = s.isig[:'rel0.5']
```
Would give the following error message (fixed with this PR):
```python

  File "/home/eric/Dev/hyperspy/hyperspy/misc/slicing.py", line 193, in __getitem__
    return self.obj._slicer(slices, self.isNavigation, out=out)

  File "/home/eric/Dev/hyperspy/hyperspy/misc/slicing.py", line 276, in _slicer
    array_slices = self._get_array_slices(slices, isNavigation)

[...]

PR #2805, opened 2021-08-09
Add hs.datasets.example_signals back in to init

thomasaarholt:
### Description of the change
commit a737190 somehow got stopped added `datasets` into being added in hs.example_signals.
I don't think that was intentional, and this fixes that.

 
### Progress of the PR
- [x] ready for review.


ericpre:
This comes from https://github.com/hyperspy/hyperspy/pull/2747, which fixed a circular import. This issue is only present on `RELEASE_next_minor` since the `non-uniform_axes` branch have been merged into `RELEASE_next_minor`.

PR #2813, opened 2021-08-25
Correct EELS Curve Fitting example in documentation

gguzzina:
Minor changes in documentation:

- Fix HT value in EELS Curve Fitting example
- Make a note that metadata is automatically included when importing from eelsdb


ericpre (approved):
Thanks, this looks good!
I thought that the kV typo was already fixed but actually it is done in #2593, which is not finished.

PR #2825, opened 2021-09-16
Zarr Reading and Writing (.zspy format) - Reformatted

CSSFrancis:
This PR is a refactoring of #2798 and all further development should be done here.


ericpre:
Thanks @CSSFrancis, this looks good and very promising! I will not have time in the next two weeks to look at the details and I will do that beginning of October then.

doc/user_guide/io.rst
```diff
@@ -418,6 +420,64 @@ Extra saving arguments
...
+    >>> s.save('test.zspy') # will save in nested directory
+    >>> hs.load('test.zspy') # loads the directory
+
+
+When saving to ``zspy``, all supported objects in the signal's
+:py:attr:`~.signal.BaseSignal.metadata` is stored. This includes lists, tuples and signals.
+Please note that in order to increase saving efficiency and speed, if possible,
+the inner-most structures are converted to numpy arrays when saved. This
+procedure homogenizes any types of the objects inside, most notably casting
+numbers as strings if any other strings are present:
```
ericpre, line 447: Is it possible to add a link to the documentation?

doc/user_guide/big_data.rst
```diff
@@ -397,6 +397,14 @@ Other minor differences
   convenience, ``nansum``, ``nanmean`` and other ``nan*`` signal methods were
   added to mimic the workflow as closely as possible.
 
+Saving Big Data
+^^^^^^^^^^^^^^^^^
+
+* **Zspy Format** When saving big data the hdf5 format struggles to compress,
+save, load and operate on datasets > 50 GB.  A better option is to use the
+:ref:`zspy-format`.
```
ericpre: I would suggest to write in a less informal manner:
```suggestion
The most efficient format supported by HyperSpy to write data is the :ref:` zspy format <zspy-format>`, mainly because it supports writing currently from concurrently from multiple threads or processes.
```

doc/user_guide/io.rst
```diff
@@ -418,6 +420,64 @@ Extra saving arguments
     saving a file, be aware that it may not be possible to load it in some platforms.
 
 
+.. _zspy-format:
+
+ZSpy - HyperSpy's Zarr Specification
+------------------------------------
+
+This is the an additional format which guarantees that no
```
ericpre:
```suggestion
Similarly to the :ref:`hspy format <hspy-format>`, the zspy format guarantees that no
```

doc/user_guide/io.rst
```diff
@@ -418,6 +420,64 @@ Extra saving arguments
...
+information will be lost in the writing process and that supports saving data
+of arbitrary dimensions. It is based on the `Zarr project <https://zarr.readthedocs.io/en/stable/index.html>`_. Which exists as a drop in
+replacement for hdf5 with the intention to fix some of the speed and scaling
+issues with the hdf5 format. **If you are working with very large datasets lazily
+we recommend using .zspy for saving and loading your data**
```
ericpre:
```suggestion
issues with the hdf5 format and is therefore suitable for saving big data.
```
and add a link to the `Saving Big Data` paragraph.

hyperspy/io_plugins/hierarchical.py
```diff
@@ -0,0 +1,665 @@
+from distutils.version import LooseVersion
```
ericpre: `distutils` is deprecated, you can use `from packaging.version import Version` instead of `distutils.versionLoose.Version`

hyperspy/io_plugins/hierarchical.py
```diff
@@ -0,0 +1,665 @@
...
+    signal_axes: tuple
+        The indexes of the signal axes
+    chunks: tuple
+        The chunks for the dataset. If None then get_signal_chunks will be called
+    get_signal_chunks: func
+        A function to get the signal chunks for the dataset
+    get_object_dset:
+        A function to get the object dset for saving ragged arrays.
+    kwds:
+        Any additional keywords
```
ericpre: Describe that the kwds are used for: it makes the code much easier to understand.

hyperspy/io_plugins/hierarchical.py
```diff
@@ -0,0 +1,665 @@
...
+class HierarchicalWriter:
+    """An object used to simplify and orgainize the process for
+    writing a Hierachical signal.  (.hspy format)
+    """
+    def __init__(self,
+                 file,
+                 signal,
+                 expg,
+                 **kwds):
+        self.file = file
```
ericpre: It needs a docstring.

hyperspy/io_plugins/hierarchical.py
```diff
@@ -0,0 +1,665 @@
...
+        self.signal = signal
+        self.expg = expg
```
ericpre: `expg` is not explicity, should it be renamed to `group` or `target_group`, etc. ?

hyperspy/io_plugins/hierarchical.py
```diff
@@ -0,0 +1,665 @@
...
+        if tmp.dtype == np.dtype('O') or tmp.ndim != 1:
+            self.dict2group(dict(zip(
+                [str(i) for i in range(len(value))], value)),
+                group.create_group(_type + str(len(value)) + '_' + key),
+                **kwds)
+        elif tmp.dtype.type is np.unicode_:
+            if _type + key in group:
+                del group[_type + key]
+            print(self.unicode_kwds)
+            print(kwds)
```
ericpre: Left over, which needs to be removed!

hyperspy/io_plugins/nexus.py
```diff
@@ -48,16 +48,13 @@
 
 def _byte_to_string(value):
     """Decode a byte string.
-
```
ericpre, line 51: How come that these lines (here and below) are being removed? I don't see why they should be removed?

hyperspy/io.py
```diff
@@ -334,15 +334,15 @@ def load(filenames=None,
             filenames = _escape_square_brackets(filenames)
 
         filenames = natsorted([f for f in glob.glob(filenames)
-                               if os.path.isfile(f)])
+                               if os.path.isfile(f) or os.path.isdir(f)])
```
ericpre: I guess a directory should be accepted only when using zarr format?
```suggestion
                               if os.path.isfile(f) or (os.path.isdir(f) and os.path.splitext(f)[1] == '.zspy')])
```

If so, there are similar `if` statement which needs to be changed.

doc/user_guide/io.rst
```diff
@@ -418,6 +420,64 @@ Extra saving arguments
...
+    >>> compressor=Blosc(cname='zstd', clevel=1, shuffle=Blosc.SHUFFLE)
+    >>> s.save('test.zspy', compressor = compressor) # will save with Blosc compression
+
+.. note::
+
+    Compression can significantly increase the saving speed. If file size is not
+    an issue, it can be disabled by setting ``compressor=None``. In general we recommend
+    compressing your datasets as it can greatly reduce i-o overhead
+
+- ``write_to_storage``: The write to storage option allows you to pass the path to a directory (or database)
```
ericpre, line 469: A more alternative would be to define the `store` argument to pass a zarr store, for example, as defined by `store = zarr.ZipStore('data/example.zip', mode='w')`. If the store argument is specified, then the zspy can pick up the path from the store itself and it doesn't need to be defined explicitely.
An even more simple alternative is to accept _zarr store_ for the filename but I am not convinced it is good because it is a bit misleading.

upcoming_changes/2798.new.rst
```diff
@@ -0,0 +1,2 @@
+Add in `zspy` saving specification for saving large data.  Helps with saving and loading large datasets.
+Using `zarr` background
```
ericpre:
```suggestion
Add in `zspy` format: hspy specification with the zarr format. Particularly useful to speed up loading and saving large by using concurrency.

```

hyperspy/io_plugins/zspy.py
```diff
@@ -0,0 +1,237 @@
...
+                           exact=True,
+                           chunks=chunks))
+    dset = group.require_dataset(key,
+                                 data.shape,
+                                 object_codec=numcodecs.VLenArray(int),
+                                 **these_kwds)
+    return dset
+
+
+def _get_signal_chunks(shape, dtype, signal_axes=None):
```
ericpre: Why not reusing the the `hyperspy.io_plugins.hspy.get_signal_chunks` and move it to `hyperspy.io_plugins.hierarchical?

hyperspy/io_plugins/zspy.py
```diff
@@ -0,0 +1,237 @@
...
+description = \
+    'A default file format for HyperSpy based on the zarr standard'
+full_support = False
+# Recognised file extension
+file_extensions = ['zspy']
+default_extension = 0
+# Writing capabilities
+non_uniform_axis = True
+writes = True
+
```
ericpre, line 49: Should it have a reader version too here, which match the `hspy` one?

hyperspy/io_plugins/hierarchical.py
```diff
@@ -0,0 +1,665 @@
...
+
+import h5py
+
+import numpy as np
+import dask.array as da
+from traits.api import Undefined
+from hyperspy.misc.utils import ensure_unicode, multiply, get_object_package_info
+from hyperspy.axes import AxesManager
+
+version = "3.1"
```
ericpre: Rename to `hspy_reader_version` and import its value from `hyperspy.io_plugins.hspy` to avoid being out of sync?

hyperspy/tests/io/test_zspy.py
```diff
@@ -0,0 +1,706 @@
...
+    'TIME': '12:00',
+    'TITLE': 'NIO EELS OK SHELL',
+    'VERSION': '1.0',
+    'XLABEL': 'Energy',
+    'XPERCHAN': 3.1,
+    'XUNITS': 'eV',
+    'YLABEL': 'Counts',
+    'YUNITS': 'Intensity'}
+
+
```
ericpre: It seems that many of these tests are repeated from `test_hdf5`, would it make sense to use `pytest.mark.parametrize` to check with `hspy` and `zspy`, where possible? It could easily get out of sync if we keep the duplicated tests.

hyperspy/io_plugins/hierarchical.py
```diff
@@ -0,0 +1,665 @@
...
+        got_data = False
+        maxshape = tuple(None for _ in data.shape)
+        while not got_data:
+            try:
+                these_kwds = kwds.copy()
+                these_kwds.update(dict(shape=data.shape,
+                                       dtype=data.dtype,
+                                       exact=True,
+                                       maxshape=maxshape,
+                                       chunks=chunks,
```
ericpre: I am getting this two warning when saving file:
```python
/home/eric/Dev/others/zarr-python/zarr/creation.py:240: UserWarning: ignoring keyword argument 'maxshape'
  warn('ignoring keyword argument %r' % k)
/home/eric/Dev/others/zarr-python/zarr/creation.py:240: UserWarning: ignoring keyword argument 'shuffle'
  warn('ignoring keyword argument %r' % k)
```
To reproduce
```python
import dask.array as da
import hyperspy.api as hs

data = da.random.randint(0, 256, size=(256, 256, 256, 256), chunks=(32, 32, 32, 32))
s.save('test.zspy')
```

ericpre (commented):
This looks very good! See comments in the changes and:
- there is a empty file which needs to be removed: hyperspy/io_plugins/spy.py
- when overwriting an existing folder, it doesn't ask to overwrite, while other reader do
- it would be good to increase the coverage as highlighted by the github annotations but this can be done in a separate PR.

hyperspy/io_plugins/nexus.py
CSSFrancis: Oh, that was a mistake.  I think they just got auto formatted at some point.

doc/user_guide/io.rst
CSSFrancis: Yea I considered both of those possibilities but decided against them for kind of the reason you described above. 

For the passing of a store argument  your filename doesn't really mean anything as only the store argument will be passed. If I remember correctly this complicates things a little bit. 

The second one is pretty do-able and I think that I got it working at one point. I could be convinced to get that working again.  The question is if that really makes sense.

hyperspy/io_plugins/zspy.py
CSSFrancis: This is based on the how to most efficiently save using default signal chunking.  HDF5 tries to make chunks that are around 1 mb where zarr says that an ideal chuck size is 100 mb. The

hyperspy/io_plugins/zspy.py
CSSFrancis: Yea I just import it from the hspy file in line 28.

hyperspy/tests/io/test_zspy.py
CSSFrancis: Good idea.  I'll do this.  It should also clean up some of the hspy tests as well

hyperspy/io_plugins/hierarchical.py
CSSFrancis: It seems like the problem is with the require_dataset function.  The default was previously Shuffle=True and max_shape=data.shape.

I'm not sure why that is.  Taking away the two keyword arguments doesn't seem to do anything (at least from the point of view of the tests). You can still pass them in as key words which seems like the better default behavior.

CSSFrancis:
@ericpre 

> This looks very good! See comments in the changes and:
> 
> * there is a empty file which needs to be removed: hyperspy/io_plugins/spy.py

Done!

> * when overwriting an existing folder, it doesn't ask to overwrite, while other reader do

I changed this behavior so it should be working.  It was kind of an oversight...

> * it would be good to increase the coverage as highlighted by the github annotations but this can be done in a separate PR.

I've added in a little bit of coverage, but obviously there are some parts which are still lacking, mostly because the hspy format isn't at 100% coverage. I might get around to that in another PR, as that should probably be at 100%

The only thing I'm not super confident with this PR is removing the shuffle and maxshape.  I think that they are legacy from some older h5py requirement.  For the maxshape I can pretty confidently say that removing it is fine.  For the shuffle, I have just set the default to True, but at least now that can also be set using the kwds arguement.

doc/user_guide/io.rst
```diff
@@ -418,6 +420,64 @@ Extra saving arguments
...
+
+    >>> from numcodecs import Blosc
+    >>> compressor=Blosc(cname='zstd', clevel=1, shuffle=Blosc.SHUFFLE)
+    >>> s.save('test.zspy', compressor = compressor) # will save with Blosc compression
+
+.. note::
+
+    Compression can significantly increase the saving speed. If file size is not
+    an issue, it can be disabled by setting ``compressor=None``. In general we recommend
+    compressing your datasets as it can greatly reduce i-o overhead
```
ericpre: This is not clear what is meant here. I suspect that the idea is: compression allow reducing IO overhead, which can increase reading/writing when enough computation power is available, in other words, balancing compression time against IO time. This is particularly relevant for the zarr format because this is done concurrently by default, while for the hdf5 format, this is more complicated to do so.

hyperspy/io_plugins/hspy.py
```diff
@@ -45,7 +43,7 @@
 # Writing capabilities
 writes = True
 non_uniform_axis = True
-version = "3.1"
+version = version
```
ericpre, line 48: This can be removed, to make the "Plugin characteristics" of `hspy` and `zspy` symmetrical.

hyperspy/io_plugins/hierarchical.py
```diff
@@ -0,0 +1,685 @@
...
+    group: Zarr.Group or h5py.Group
+        The group to write the data to
+    data: Array-like
+        The data to be written
+    key: str
+        The key for the data
+    signal_axes: tuple
+        The indexes of the signal axes
+    chunks: tuple
+        The chunks for the dataset. If None then get_signal_chunks will be called
```
ericpre:
```suggestion
        The chunks for the dataset. If ``None`` and saving lazy signal, the chunks
        of the dask array will be used otherwise the chunks will determine by
        ``get_signal_chunks``.
```

hyperspy/io_plugins/hierarchical.py
```diff
@@ -0,0 +1,685 @@
...
+        """
+        self.file = file
+        self.signal = signal
+        self.group = group
+        self.Dataset = None
+        self.Group = None
+        self.unicode_kwds = None
+        self.ragged_kwds = None
+        self.kwds = kwds
+        self.overwrite_dataset=None
```
ericpre: As far as my python knowledge goes, this is not a standard pattern to overwrite a method in a child class: I will make a PR to this branch to refactor it.

hyperspy/io_plugins/zspy.py
```diff
@@ -0,0 +1,238 @@
...
+                 file,
+                 signal,
+                 expg, **kwargs):
+        super().__init__(file, signal, expg, **kwargs)
+        self.Dataset = Array
+        self.unicode_kwds = {"dtype": object, "object_codec": numcodecs.JSON()}
+        self.ragged_kwds = {"dtype": object,
+                            "object_codec": numcodecs.VLenArray(int),
+                            "exact":  True}
+        self.overwrite_dataset = overwrite_dataset
```
ericpre: As far as I know, this is not a standard pattern to implement class inheritance, I made https://github.com/CSSFrancis/hyperspy/pull/1 to simplify it.

hyperspy/io_plugins/zspy.py
ericpre: Would it be possible to make the target size a parameter and use only one function?

doc/user_guide/io.rst
ericpre: Going for option 2 should be good because this is a pattern used by the [`zarr.group`](https://zarr.readthedocs.io/en/stable/api/hierarchy.html#zarr.hierarchy.group), as it can take a path or a zarr store as first argument.

ericpre (commented):
I made some more comments and a PR to simplify the class inheritance: https://github.com/CSSFrancis/hyperspy/pull/1.

doc/user_guide/io.rst
CSSFrancis: This is mostly for large data sets which are stored on slow spinning disk harddrives.  Like you said, in my  personal case I am often i-o bound for lazy operations and have an excess of computational power.

Maybe this is better:

"Lazy operations are often i-o bound, reading and writing the data creates a bottle neck in processes due to the slow read write speed of many hard disks. In these cases, compressing your data is often beneficial to the speed of some operation.   Compression speeds up the process as there is less to read/write with the trade off of slightly more computational work on the CPU."

ericpre: Yes, this is clear and actually, I mentioned concurrently writing/reading, which is a bit misleading here because this is an "orthogonal" consideration.

hyperspy/io_plugins/zspy.py
CSSFrancis: I just looked through your simplification.  Thanks! That is what I was originally going for but got stuck because I wasn't really familiar with class methods. 

I think that is much more readable.

hyperspy/io_plugins/hierarchical.py
CSSFrancis: I'm not terribly certain what is "pythonic" in this case.  I think there is a time and a place for overriding methods in a child class.  That being said your refactoring is much cleaner and more readable (and originally what I wanted to do).

hyperspy/io_plugins/zspy.py
ericpre: Yes, the crux was to be able to reuse the `overwrite_dataset` in the nexus reader and I had to look it up! :)

CSSFrancis:
@ericpre 
In reference to https://github.com/hyperspy/hyperspy/pull/2825#discussion_r723609625 I think that the current get signal chunks function is very confusing.  If I am going to make a generic function I might as well just change it...

```python
    def _get_signal_chunks(shape, dtype, signal_axes, target_size=1024*1024):
        """Function that calculates chunks for the signal, preferably at least one
        chunk per signal space.

        Parameters
        ----------
        shape : tuple
            the shape of the dataset to be sored / chunked
        dtype : {dtype, string}
            the numpy dtype of the data
        signal_axes: {None, iterable of ints}
            the axes defining "signal space" of the dataset. If None, the default
            h5py chunking is performed.
        target_size: int
            The target number of bytes for one chunk
        """
        typesize = np.dtype(dtype).itemsize
        if signal_axes is None:
            return h5py._hl.filters.guess_chunk(shape, None, typesize)

        # largely based on the guess_chunk in h5py
        bytes_per_signal = multiply([shape[i] for i in signal_axes]) * typesize
        signals_per_chunk = np.floor_divide(target_size,bytes_per_signal)
        if signals_per_chunk < 2:
            # signal is larger than chunk max
            chunks = [s if i in signal_axes else 1 for i, s in enumerate(shape)]
            return tuple(chunks)
        else:
            # signal is smaller than chunk max
            navigation_axes = tuple(i for i in range(len(shape)) if i not in
                                    signal_axes)
            num_nav_axes = len(navigation_axes)
            sig_axes_chunk = np.floor(signals_per_chunk**(1/num_nav_axes))
            remainder = np.floor_divide(signals_per_chunk - (sig_axes_chunk**num_nav_axes),
                                        sig_axes_chunk)
            chunks = [s if i in signal_axes else sig_axes_chunk for i, s in enumerate(shape)]
            chunks[navigation_axes[0]] = chunks[navigation_axes[0]]+remainder
        return tuple(int(x) for x in chunks)
```

I think that this makes quite a bit more sense and gets rid of the while loop. It changes the behavior a little (I think) but the previous version was not actually creating 1 mb chunks but rather chunks that were 1024*1024* bit depth which I don't thinks makes much sense.

hyperspy/io.py
```diff
@@ -780,24 +786,30 @@ def save(filename, signal, overwrite=None, **kwds):
...
-        _logger.info(f'{filename} was created')
-        signal.tmp_parameters.set_item('folder', filename.parent)
-        signal.tmp_parameters.set_item('filename', filename.stem)
-        signal.tmp_parameters.set_item('extension', extension)
+        if not isinstance(filename, MutableMapping):
+            writer.file_writer(str(filename), signal, **kwds)
+            _logger.info(f'{filename} was created')
+            signal.tmp_parameters.set_item('folder', filename.parent)
+            signal.tmp_parameters.set_item('filename', filename.stem)
+            signal.tmp_parameters.set_item('extension', extension)
```
ericpre, line 813: It would be good to set them when saving using a store too?

hyperspy/io_plugins/zspy.py
```diff
@@ -0,0 +1,206 @@
...
+    ----------
+    filename: str
+    signal: a BaseSignal instance
+    *args, optional
+    **kwds, optional
+    """
+    if "compressor" not in kwds:
+        from numcodecs import Blosc
+        kwds["compressor"] = Blosc(cname='zstd', clevel=1)
+    if ("write_to_storage" in kwds and kwds["write_to_storage"]) or isinstance(filename, MutableMapping):
```
ericpre, line 150: Is the `"write_to_storage"` still necessary?

hyperspy/io_plugins/zspy.py
```diff
@@ -0,0 +1,206 @@
...
+        if isinstance(filename, MutableMapping):
+            store = filename.path
+        else:
+            store = filename
+    else:
+        store = zarr.storage.NestedDirectoryStore(filename,)
+    f = zarr.group(store=store, overwrite=True, cache_attrs=False)
```
ericpre: What is the reason to use `cache_attrs=False`? Add a comment to explain?

hyperspy/signal.py
```diff
@@ -2856,9 +2857,11 @@ def save(self, filename=None, overwrite=None, extension=None,
             else:
                 raise ValueError('File name not defined')
 
-        filename = Path(filename)
-        if extension is not None:
-            filename = filename.with_suffix(f".{extension}")
+        print(type(filename))
```
ericpre: Debugging left over.

hyperspy/io_plugins/zspy.py
```diff
@@ -0,0 +1,206 @@
...
+    print(f)
```
ericpre: Debugging left over.

hyperspy/io.py
```diff
@@ -780,24 +786,30 @@ def save(filename, signal, overwrite=None, **kwds):
...
+            write = True  # Write the file
+        elif overwrite is False and is_file:
+            write = False  # Don't write the file
+        else:
+            raise ValueError("`overwrite` parameter can only be None, True or "
+                             "False.")
     else:
-        raise ValueError("`overwrite` parameter can only be None, True or "
-                         "False.")
+        write=True
```
ericpre: Please add a comment to explain why `write = True`.

hyperspy/io_plugins/_hierarchical.py
```diff
@@ -460,9 +461,49 @@ def __init__(self,
         self.kwds = kwds
 
     @staticmethod
-    def _get_signal_chunks():
-        raise NotImplementedError(
-            "This method must be implemented by subclasses.")
+    def _get_signal_chunks(shape, dtype, signal_axes=None, target_size=1e6):
```
ericpre, line 464: There is also  https://github.com/hyperspy/hyperspy/blob/3517f8afbd6e9060d2ca8ae1f9fc10ccd7b8b799/hyperspy/_signals/lazy.py#L189-L263
which does pretty much the same thing. I guess, we need a standalone `guess_chunks` which should be used here and in `hyperspy._signals.lazy.LazySignal`.

Since this is indirectly related to this purpose of this PR, maybe it is better to do this in a separate PR. I would prefer to not change the behaviour in this PR and sort it out completely in a separate PR, with adding test, etc. There is already a lot in this PR! What do you think?

For reference, here is the auto-chunking from zarr:
https://github.com/zarr-developers/zarr-python/blob/be3d657484590adacaedf980a2951ddfa92b68b8/zarr/util.py#L60-L103

ericpre (commented):
As mentioned in [one of the comments](https://github.com/hyperspy/hyperspy/pull/2825#discussion_r724778881), I would be inclined not to change how the chunking is calculated and sort it out properly in a separate PR.
Some tests of this PR are failing and a few other minor comments need to considered. Other than that, it should be good to go!

hyperspy/io_plugins/zspy.py
CSSFrancis: Kind of... I allowed for a folder or a Mutable Mapping object to be passed in.  I guess I could just look to see if there is a file there that `zarr` can read and then write to that object if there is a file there.

hyperspy/io_plugins/zspy.py
CSSFrancis: That was just from some testing.  It doesn't actually matter here. Removed

hyperspy/io.py
CSSFrancis: I set them if the attribute is there.

hyperspy/io_plugins/_hierarchical.py
CSSFrancis: We can sort it out in a separate PR. Do you want me to roll back to the previous _get_signal_chunks.

I would argue against the `zarr` and the `h5py` guess chunk functions.  Something about using a while loop rather than directly calculating the chunk shape makes me itch and is very confusing. `zarr` does a little bit better job explaining what is happening, but I think it is still fairly unpythonic. 

Honestly, chunking in hyperspy is a little bit broken and could use some major work.

I can make a issue to track some of the issues that I have come across.

hyperspy/io_plugins/_hierarchical.py
```diff
@@ -0,0 +1,726 @@
...
+        if "file_format_version" in self.file.attrs:
+            version = self.file.attrs["file_format_version"]
+            if isinstance(version, bytes):
+                version = version.decode()
+            if isinstance(version, float):
+                version = str(round(version, 2))
+        elif "Experiments" in self.file:
+            # Chances are that this is a HSpy hdf5 file version 1.0
+            version = "1.0"
+        elif "Analysis" in self.f:
```
ericpre:
```suggestion
        elif "Analysis" in self.file:
```

hyperspy/io_plugins/zspy.py
```diff
@@ -0,0 +1,205 @@
...
+        elif data.dtype == np.dtype('O'):
+            group[key][:] = data[:]  # check lazy
+        else:
+            path = group._store.dir_path() + "/" + dset.path
+            dset = zarr.open_array(path,
+                                   mode="w",
+                                   shape=data.shape,
+                                    dtype=data.dtype,
+                                    chunks=chunks,
+                                    **kwds)
```
ericpre, line 131: It it necessary?

hyperspy/io_plugins/zspy.py
ericpre: I would suggest not to try to pass store, etc. because there are issues in the current implementation. We can revisit later.

ericpre (commented):
I would suggest to leave the chunking discussion for another PR. Same for passing zarr store instead of filename string, because the it doesn't fully work.

There are two other comments/questions.

hyperspy/io_plugins/zspy.py
CSSFrancis: At least on my machine it seems to be.  If you take it out then the array is never updated.  There are some things with how the synchronizer is handled that I am still rather confused with.  There is no `flush` method for the directory store but I think that reopening the file is equivalent to this.  

If I tried a bunch of different things but this is the one that worked in the end

ericpre: OK, let's leave it there for now and we can revisit after I rebase https://github.com/hyperspy/hyperspy/pull/2797 on this branch.

ericpre:
There are two tests failing:
```python
______________________ test_get_signal_chunks[1000000.0] ______________________
[gw1] win32 -- Python 3.8.10 c:\hostedtoolcache\windows\python\3.8.10\x64\python.exe

target_size = 1000000.0

    @pytest.mark.parametrize("target_size", (1e6,1e7))
    def test_get_signal_chunks(target_size):
        chunks = HierarchicalWriter._get_signal_chunks(shape=[15, 15, 256, 256],
                                                       dtype=int,
                                                       signal_axes=(2, 3),
                                                       target_size=target_size)
>       assert (np.prod(chunks)*8 < target_size)
E       assert (196608 * 8) < 1000000.0
E        +  where 196608 = <function prod at 0x0000017CB7857160>((3, 1, 256, 256))
E        +    where <function prod at 0x0000017CB7857160> = np.prod

hyperspy\tests\io\test_hdf5.py:857: AssertionError
RRRF
_____________________ test_get_signal_chunks[10000000.0] ______________________
[gw1] win32 -- Python 3.8.10 c:\hostedtoolcache\windows\python\3.8.10\x64\python.exe

target_size = 10000000.0

    @pytest.mark.parametrize("target_size", (1e6,1e7))
    def test_get_signal_chunks(target_size):
        chunks = HierarchicalWriter._get_signal_chunks(shape=[15, 15, 256, 256],
                                                       dtype=int,
                                                       signal_axes=(2, 3),
                                                       target_size=target_size)
>       assert (np.prod(chunks)*8 < target_size)
E       assert (2359296 * 8) < 10000000.0
E        +  where 2359296 = <function prod at 0x0000017CB7857160>((6, 6, 256, 256))
E        +    where <function prod at 0x0000017CB7857160> = np.prod
```

CSSFrancis:
@ericpre What a strange little bug... the windows operating systems running the tests are 32 bit so `int` is by default 32 bits where on the 64 bit operating systems `int` by default is 64 bits.

ericpre:
> @ericpre What a strange little bug... the windows operating systems running the tests are 32 bit so `int` is by default 32 bits where on the 64 bit operating systems `int` by default is 64 bits.

The machine on CI are 64 bits (`'win32'`, which must be the output of `sys.platform` is misleading here...`), and they must be another reason. Anyway, since we say that the chunking estimation will need a refactor in a follow-up PR, we can leave it for now.

In 2b236f5, I made zarr optional because there are no packages available on pypi or conda-forge for architecture other than x86_64/amd64.

ericpre:
Thanks for the good work on this, @CSSFrancis!

PR #2826, opened 2021-09-20
Update docs on installing hyperspy dev version

thomasaarholt:
### Description of the change
To make for an easier debugging experience, developers should (first) install the hyperspy dev install in a fresh environment.
I suggest specifying `python` on environment creation, since I've had trouble with creating totally empty environments in the past.

### Progress of the PR
- [x] ready for review.

### DIFF
--- a/doc/user_guide/install.rst
+++ b/doc/user_guide/install.rst
@@ -291,9 +291,11 @@ Installation in a Anaconda/Miniconda distribution
 Install the runtime and development dependencies requirements using conda:
 
 .. code-block:: bash
-
-    $ conda install hyperspy-base -c conda-forge --only-deps
-    $ conda install hyperspy-dev -c conda-forge
+ 
+    $ conda create -n hspy_dev python # create an empty environment with latest python
+    $ conda activate hspy_dev # activate environment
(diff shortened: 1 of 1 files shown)


francisco-dlp:
That's a good point. However, I am not sure if it is good to change the default instructions because:
* people who are familiar with environments know how to modify the code to set up an environment if that's what they want to do
* those who are not familiar with environments will find it harder to get started since they'll have to learn how to use and manage environments

What about adding a note explaining the advantages of using an environment for debugging and moving the updated instructions there?

thomasaarholt:
I mostly disagree 😆  But I've added an "optional" introduction with a link. What do you now think?

Note that we have already mention environments earlier in the document (under [Further information](http://hyperspy.org/hyperspy-doc/current/user_guide/install.html#further-information)), in the context of regular users, and we are here talking about developers.

I don't think we should include a "why should you use an environment" explanation in hyperspy docs - there are tons of good resources on the net, and I don't want to bloat our docs. The "Optional" bit in the ingress for this part should be enough to trigger interest.

francisco-dlp:
I meant an explanation about why environment may help debugging (as you explained above). But, on a second thought, I agree that even that belongs elsewhere.

Looks great now, thanks!

PR #2828, opened 2021-09-27
Impulse io

DENSmerijn:
### Description of the change
- Added Impulse csv file reader, Impulse is the new software for DENSsolutions In Situ systems
- Added Impulse reader test script and data
- Updated documentation
- Modified other csv file reader (PC) to include the 'protochips' file extension (required to specify which csv reader is used)

### Progress of the PR
- [x] add tests,
- [x] ready for review.

```python
import hyperspy.api as hs
hs.load('ImpulseLogfile.csv', reader='impulse')
```


doc/user_guide/io.rst
```diff
@@ -843,14 +845,28 @@ Extra saving arguments
 
 .. _dens-format:
 
-DENS heater log
+DENSsolutions formats
 ---------------
```
ericpre: Sphinx will raise warning about the title underline being too short.

doc/user_guide/io.rst
```diff
@@ -843,14 +845,28 @@ Extra saving arguments
...
+
+To read Impulse logfiles, use the reader argument to define the correct file reader:
+
+.. code-block:: python
+
+    >>> hs.load("filename.csv", reader="impulse")
+
+
+DENSsolutions DigiHeater logfile
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```
ericpre: Same as above: the title underline needs to match the length of the title.

hyperspy/io_plugins/impulse.py
```diff
@@ -0,0 +1,207 @@
...
+            "Signal": {
+                "signal_type": "",
+                "quantity": self._parse_quantity_units(quantity),
+            },
+        }
+
+    def _parse_quantity_units(self, quantity):
+        quantity_split = quantity.strip().split(" ")
+        if len(quantity_split) > 1:
+            if quantity_split[-1][0] == "(" and quantity_split[-1][-1] == ")":
```
ericpre: This two `if` statement could be merged.

hyperspy/io_plugins/impulse.py
```diff
@@ -0,0 +1,207 @@
...
+            if not self._is_impulse_csv_file():
+                raise IOError(invalid_file_error)
+            self._read_metadatafile()
+        self.logged_quantity_name_list = self.column_names[2:]
+
+    def _is_impulse_csv_file(self):
+        if "TimeStamp" in self.column_names and len(self.column_names) >= 3:
+            return True
+        else:
+            return False
```
ericpre:
```suggestion
        return "TimeStamp" in self.column_names and len(self.column_names) >= 3
```

Since this is short is used only once, I would put it directly where it is used: avoiding redirection is usually more readible.

hyperspy/io_plugins/impulse.py
```diff
@@ -0,0 +1,207 @@
...
+        data = np.genfromtxt(
+            self.filename,
+            delimiter=",",
+            dtype=None,
+            names=names,
+            skip_header=1,
+            encoding="latin1",
+        )
+        self._data_dictionary = dict()
+        for i, name, name_dtype in zip(range(len(names)), self.column_names, names):
```
ericpre:
```suggestion
        for i, (name, name_dtype) in enumerate(zip(self.column_names, names)):
```

hyperspy/io_plugins/impulse.py
```diff
@@ -0,0 +1,207 @@
...
+                                1
+                            ].strip()
+                    self.original_metadata["Notes"] = notes
+                    self.start_datetime = np.datetime64(
+                        dt.strptime(
+                            self.original_metadata["Experiment_date"]
+                            + self.original_metadata["Experiment_time"],
+                            "%d-%m-%Y%H:%M:%S",
+                        )
+                    )
```
ericpre: This is not used.

hyperspy/io_plugins/impulse.py
```diff
@@ -0,0 +1,207 @@
...
+        if len(quantity_split) > 1:
+            if quantity_split[-1][0] == "(" and quantity_split[-1][-1] == ")":
+                return quantity_split[-1]
+            else:
+                return "NA"
+        else:
+            return "NA"
+
+    def _get_metadata_time_axis(self):
+        return {"value": self.time_axis, "units": self.time_units}
```
ericpre: This is not used, better not to keep it.

hyperspy/io_plugins/impulse.py
```diff
@@ -0,0 +1,207 @@
...
+    csvs = []
+    for key in csv_file.logged_quantity_name_list:
+        try:
+            csvs.append(csv_file.get_dictionary(key))
+        except BaseException:
+            raise IOError(invalid_file_error)
+    return csvs
+
+
+class impulseCSV:
```
ericpre: Python class name follows CapWords convention.
```suggestion
class ImpulseCSV:
```

hyperspy/io_plugins/impulse.py
```diff
@@ -0,0 +1,207 @@
...
+    csv_file = impulseCSV(filename)
+    return _impulseCSV_log_reader(csv_file)
+
+
+def _impulseCSV_log_reader(csv_file):
+    csvs = []
+    for key in csv_file.logged_quantity_name_list:
+        try:
+            csvs.append(csv_file.get_dictionary(key))
+        except BaseException:
```
ericpre: I am not convinced, that it is necessary to catch the exception here, because at this point, we already checked that it is supposedly a valid impulse file. One major drawback of catching the error is that it is more inconvenient to find where the error come from because the information about the original error is lost: this can be particular inconvenient when trying to help an user.

hyperspy/io_plugins/impulse.py
```diff
@@ -0,0 +1,207 @@
...
+                "signal_type": "",
+                "quantity": self._parse_quantity_units(quantity),
+            },
+        }
+
+    def _parse_quantity_units(self, quantity):
+        quantity_split = quantity.strip().split(" ")
+        if len(quantity_split) > 1:
+            if quantity_split[-1][0] == "(" and quantity_split[-1][-1] == ")":
+                return quantity_split[-1]
```
ericpre: Is it expected that this is not covered by the tests?

hyperspy/io_plugins/impulse.py
```diff
@@ -0,0 +1,207 @@
...
+        return {
+            "General": {
+                "original_filename": os.path.split(self.filename)[1],
+                "title": "%s" % quantity,
+                "date": self.original_metadata["Experiment_date"],
+                "time": self.original_metadata["Experiment_time"],
+            },
+            "Signal": {
+                "signal_type": "",
+                "quantity": self._parse_quantity_units(quantity),
```
ericpre, line 87: Usually, when a metadata is not know, we don't populate it.

hyperspy/io_plugins/impulse.py
```diff
@@ -0,0 +1,207 @@
...
+
+    def _parse_quantity_units(self, quantity):
+        quantity_split = quantity.strip().split(" ")
+        if len(quantity_split) > 1:
+            if quantity_split[-1][0] == "(" and quantity_split[-1][-1] == ")":
+                return quantity_split[-1]
+            else:
+                return "NA"
+        else:
+            return "NA"
```
ericpre: `'NA'` can be ambiguous: better to return an empty string, or not to populate the metadata.

hyperspy/io_plugins/impulse.py
```diff
@@ -0,0 +1,207 @@
...
+                    self.start_datetime = np.datetime64(
+                        dt.strptime(
+                            self.original_metadata["Experiment_date"]
+                            + self.original_metadata["Experiment_time"],
+                            "%d-%m-%Y%H:%M:%S",
+                        )
+                    )
+
+            else:
+                raise IOError(no_metadata_file_error)
```
ericpre: Even if there is no metadata file or it is corrupted, it would still be useful to be able to open the file? Maybe simply add a warning?

hyperspy/io_plugins/impulse.py
```diff
@@ -0,0 +1,207 @@
...
+        else:
+            raise IOError(invalid_filenaming_error)
+
+    def _get_axes(self):
+        scale = np.diff(self.time_axis[1:-1]).mean()
+        units = "s"
+        offset = 0
+        self.time_units = "Seconds"
```
ericpre: Since `_get_metadata_time_axis` is not used, this should be removed too.

ericpre (changes requested):
This looks good, it needs minor improvement. I added comments/suggestions and push a [pull request](https://github.com/DENSsolutions/hyperspy/pull/1) to this branch.

It will need a changelog entry too.

DENSmerijn:
Hi ericpre, 

Thanks for reviewing the code, I believe all your suggested improvements were implemented now.

~ Merijn

ericpre (approved):
Yes, all good, thanks!

PR #2829, opened 2021-10-04
Fix trigger extension test suite workflow

ericpre:
The `test_extension.yml` is not triggered when adding a label to a PR.


ericpre:
This is working: [this event](https://github.com/hyperspy/hyperspy/pull/2703#event-5404583893) triggered this [workflow run](https://github.com/hyperspy/hyperspy/actions/runs/1304078411).

PR #2830, opened 2021-10-04
Protect axes manager attribute

ericpre:
Some attributes of the `AxesManager` can be changed, which can in turn "break" the `AxesManager`.

### Progress of the PR
- [x] Deprecate `AxesManager.set_signal_dimension` in favour of using `as_signal1D`, `as_signal2D` or `transpose` of the signal instance instead.
- [x] protect `AxesManager` attributes:
   - make `signal_dimension` and `navigation_dimension` properties
   - make `signal_axes` and `navigation_axes` properties
   - make `signal_size` and `navigation_size` properties
   - make `signal_shape` and `navigation_shape` properties
- [x] add tests,
- [x] ready for review.


jlaehne (approved):
There is a few codecov warnings that should be low hanging fruit.

Otherwise looks good to me.

jlaehne:
@ericpre can you fix the conflicts?

@thomasaarholt do you want to have a look at this PR? Didn't you bring up something in this direction a while ago?

ericpre:
I have rebased but it would suggest to wait for the changes with `map` are done in hyperspy and pyxem to consider this PR because the changes of this PR breaks pyxem if I remember correctly.

jlaehne:
> I have rebased but it would suggest to wait for the changes with `map` are done in hyperspy and pyxem to consider this PR because the changes of this PR breaks pyxem if I remember correctly.

pyxem seems about ready to consider this one for v1.7? @magnunor @pc494 @CSSFrancis, can anyone of you review this one in view of the possible problem with pyxem?

CSSFrancis:
@jlaehne I ran the test suite for `pyxem` using https://github.com/pyxem/pyxem/tree/hyperspy_release_next_minor. It seems like there aren't any more issue with the map function.  At least none of the tests are failing.   On that account this is probably ready for a review.

jlaehne:
> 

Did you also run it specifically for this PR to see if it would introduce any additional problems?

hyperspy/_signals/signal1d.py
```diff
@@ -1657,10 +1655,6 @@ def plot(self,
...
 class LazySignal1D(LazySignal, Signal1D):
 
-    """
-    """
     _lazy = True
 
-    def __init__(self, *args, **kwargs):
-        super().__init__(*args, **kwargs)
-        self.axes_manager.set_signal_dimension(1)
+    pass
```
magnunor: I think this `pass` can be removed? The `_lazy = True` should be enough.

hyperspy/_signals/signal2d.py
```diff
@@ -954,3 +951,5 @@ def find_peaks(self, method='local_max', interactive=True,
 class LazySignal2D(LazySignal, Signal2D):
 
     _lazy = True
+
+    pass
```
magnunor: I think this `pass` can be removed? The `_lazy = True` should be enough.

CSSFrancis:
> Did you also run it specifically for this PR to see if it would introduce any additional problems?

Yes I ran it for this PR.  It didn't seem to introduce any new problems.

magnunor:
I'm having a quick look at this PR as well, gimme a couple of minutes and I'll have some feedback.

magnunor (commented):
I think the changes in this pull request is good! Having attributes protected makes things more robust.

------------

I ran the pyxem tests using this PR branch, and a number of tests failed. Examples of this:

- `correlation.axes_manager.navigation_axes = self.axes_manager.navigation_axes`
- `sn.axes_manager.navigation_axes = self.axes_manager.navigation_axes`

I think this should be pretty easy to sort out.

-----------

One other minor thing:

- Should there be a change log entry?

------------

I tested the functionality, and should `s.axes_manager.signal_dimension` change the signal type?

```python
import numpy as np
import hyperspy.api as hs
s = hs.signals.Signal2D(np.zeros((2, 20, 30)))
print(s)
s.axes_manager.signal_dimension = 1
print(s)
```

Gives:
```python
<Signal2D, title: , dimensions: (2|30, 20)>
<Signal2D, title: , dimensions: (20, 2|30)>
```
Ergo, should the second line be `Signal1D`?

Maybe not, since `set_signal_dimension` works the same way (in `RELEASE_next_minor`):

```python
s = hs.signals.Signal2D(np.zeros((2, 20, 30)))
print(s)
s.axes_manager.set_signal_dimension(1)
print(s)
```

Gives:
```python
<Signal2D, title: , dimensions: (2|30, 20)>
<Signal2D, title: , dimensions: (20, 2|30)>
```

jlaehne:
Well, the integration tests still show 44 fails against the development branch of `pyxem`.

And also two fails against the development branch of `kikuchipy` (@hakonanes).

magnunor:
> Well, the integration tests still show 44 fails against the development branch of pyxem.

This is expected, due to the `map` fixes not being in the `master` branch, but rather in a separate branch at the moment. Specifically https://github.com/pyxem/pyxem/tree/hyperspy_release_next_minor

> correlation.axes_manager.navigation_axes = self.axes_manager.navigation_axes

Any tips on how to fix this, essentially how to "copy" navigation axes from one signal to another? I'm not too familiar with the internal workings of the `axes_manager`.

jlaehne:
> > correlation.axes_manager.navigation_axes = self.axes_manager.navigation_axes
> 
> Any tips on how to fix this, essentially how to "copy" navigation axes from one signal to another? I'm not too familiar with the internal workings of the `axes_manager`.

Use e.g. the `axes_manager.set_axis` function (https://github.com/LumiSpy/lumispy/blob/2258151a2405b74b0911835513686cd3983c67cf/lumispy/signals/luminescence_spectrum.py#L161) iterating over the navigation_axis dimensions?

The `axes_manager` also has `copy` and `deepcopy` functions.

ericpre:
See also: https://hyperspy.readthedocs.io/en/latest/user_guide/axes.html#adding-removing-axes-to-from-a-signal

ericpre:
> I tested the functionality, and should `s.axes_manager.signal_dimension` change the signal type?
> 
> ```python
> import numpy as np
> import hyperspy.api as hs
> s = hs.signals.Signal2D(np.zeros((2, 20, 30)))
> print(s)
> s.axes_manager.signal_dimension = 1
> print(s)
> ```
> 
> Gives:
> 
> ```python
> <Signal2D, title: , dimensions: (2|30, 20)>
> <Signal2D, title: , dimensions: (20, 2|30)>
> ```
> 
> Ergo, should the second line be `Signal1D`?
> 
> Maybe not, since `set_signal_dimension` works the same way (in `RELEASE_next_minor`):
> 
> ```python
> s = hs.signals.Signal2D(np.zeros((2, 20, 30)))
> print(s)
> s.axes_manager.set_signal_dimension(1)
> print(s)
> ```
> 
> Gives:
> 
> ```python
> <Signal2D, title: , dimensions: (2|30, 20)>
> <Signal2D, title: , dimensions: (20, 2|30)>
> ```

Indeed, this is not good and it isn't possible to update the signal type because the `AxesManager` doesn't have access to the signal object itself. There are two possibilities to solve this bug:
1. Add the signal object as a private attribute of `AxesManager`
2. don't allow setting the signal dimension through the `AxesManager`. Instead, `as_signal1D`, `as_signal2D` or `transpose` should be used for the public API.

I would be in favour of option 2.

ericpre:
I went for option 2, because I think it better to have only one way to doing a thing. I keep the history for the review but once we are happy with this PR, I will squash some commits and add the changelog entry.

hakonanes:
> And also two fails against the development branch of kikuchipy (@hakonanes).

Thanks for pinging me, @jlaehne. The failing tests are unrelated to this PR and HyperSpy (relate to dask and only affects the tests and not functionality).

hyperspy/tests/signals/test_ragged_signal.py
```diff
@@ -121,5 +121,9 @@ def test_conversion_signal():
     data = np.empty((2, 3, 4), dtype=object)
     data.fill(np.array([10, 20, 30, 40, 50]))
     s = hs.signals.BaseSignal(data, ragged=True)
+    
     with pytest.raises(ValueError):
-        s.axes_manager.signal_dimension = 1
+        s.axes_manager._set_signal_dimension(1)
+
+    with pytest.raises(RuntimeError):
+        s.as_signal1D(0)
```
jlaehne: newline at file end missing

hyperspy/signal.py
```diff
@@ -2073,9 +2073,9 @@ def _plot_x_results(factors, loadings, factors_navigator, loadings_navigator,
                     factors_dim, loadings_dim):
     factors.axes_manager._axes[0] = loadings.axes_manager._axes[0]
     if loadings.axes_manager.signal_dimension > 2:
-        loadings.axes_manager.signal_dimension = loadings_dim
+        loadings.axes_manager._set_signal_dimension(loadings_dim)
```
jlaehne, line 2076: a few codecov warnings

jlaehne (approved):

hyperspy/signal.py
ericpre: Done. For the other one, `IntegrateArea` and `integrate_in_range` are deprecated and I don't think it is worth it, particularly since this is testing a UI function, for which writing a test is a bit of a pain!

jlaehne (approved):

magnunor:
I tested `axes_manager.set_axis`, and there seems to be an "inversion" of the indices:

```python
import numpy as np
import hyperspy.api as hs

s0 = hs.signals.Signal1D(np.zeros((5, 10, 15)))
s0.axes_manager.navigation_axes[0].scale = 0.1
s0.axes_manager.navigation_axes[1].scale = 0.2
s1 = hs.signals.Signal1D(np.zeros((5, 10, 20)))

s1.axes_manager.set_axis(s0.axes_manager.navigation_axes[0], 0)
s1.axes_manager.set_axis(s0.axes_manager.navigation_axes[1], 1)
```

```python
print(s0.axes_manager)
<Axes manager, axes: (10, 5|15)>
     <undefined> |     10 |      7 |       0 |     0.1 | <undefined> 
     <undefined> |      5 |      3 |       0 |     0.2 | <undefined> 
     <undefined> |     15 |      0 |       0 |       1 | <undefined>

print(s1.axes_manager)
<Axes manager, axes: (5, 10|20)>
     <undefined> |      5 |      3 |       0 |     0.2 | <undefined> 
     <undefined> |     10 |      7 |       0 |     0.1 | <undefined> 
     <undefined> |     20 |      0 |       0 |       1 | <undefined>
```

---------------------

Ergo, to properly "copy" the navigation axes:

```python
s1.axes_manager.set_axis(s0.axes_manager.navigation_axes[0], 1)
s1.axes_manager.set_axis(s0.axes_manager.navigation_axes[1], 0)
```

ericpre:
@magnunor: I am getting more and more convinced that setting axis manually is an antipattern. I had a look at how it is used in pyxem and from what I understand, there are 3 main cases:
- when `map` is used it is simply not necessary to set the axes manually, because `map` is doing it.
- `set_signal_dimension`, which can be replace by `transpose`, etc, as explained in the deprecation warning of this PR
- *the `DiffractionVectors` case*, for which the two cases above doesn't work straightaway, but I suspect that setting can work to fix issue case by case without really addressing the issue inherent to the `DiffractionVectors` class.

On this topic, there are also the `transfer_navigation_axes_to_signal_axes` and co, for which, I suspect that we should find a better pattern than creating a signal and then set the axis. Maybe it would be better to create the axis dictionary (using `to_dictionary`, for example) and to pass the dictionaries when creating the signal.

ericpre:
@CSSFrancis, since you have dealing *quite a bit* with `map` & co, can you please comment on my [summary above](https://github.com/hyperspy/hyperspy/pull/2830#issuecomment-1097194139)? Does it sound sensible to you?

jlaehne:
> I tested `axes_manager.set_axis`, and there seems to be an "inversion" of the indices:

@ericpre did you check the question of inversion of indices?

ericpre:
> > I tested `axes_manager.set_axis`, and there seems to be an "inversion" of the indices:
> 
> @ericpre did you check the question of inversion of indices?

No, but this PR suggest to use alternatives to `set_axis`. ;) I am not sure if `set_axis` is a good pattern to follow. What is the use case for `set_axis`?

CSSFrancis:
@ericpre 
> @magnunor: I am getting more and more convinced that setting axis manually is an antipattern. I had a look at how it is used in pyxem and from what I understand, there are 3 main cases:
>
> * when `map` is used it is simply not necessary to set the axes manually, because `map` is doing it.
> * `set_signal_dimension`, which can be replace by `transpose`, etc, as explained in the deprecation warning of this PR
> * _the `DiffractionVectors` case_, for which the two cases above doesn't work straightaway, but I suspect that setting can work to fix issue case by case without really addressing the issue inherent to the `DiffractionVectors` class.

I agree with you on all of these points.  It would seem that directly modifying the axes_manager (except for things like the name, unit and offset) should probably be avoided, in that way we should act like `numpy` where things like `.shape` , `ndim` etc are properties of an array.  In a similar way the number of axes, and their sizes should be set by the signal and not the other way around. I think that this was the initial intent of this PR.

In the diffraction vectors case this is actually very helpful mostly because we already have a `ragged` data type for arrays with a `dtype=object` which we can then deal with differently.  I think I tried #2876 with this PR and it helps clean up some of the code simply because there are fewer ambiguous cases.  

The one thing that I might suggest as an extension of this would be to change `ragged` into a property which just checks if the dtype of the signal is object rather than something than manually setting the ragged attribute.  Maybe that is a bigger change than I am thinking so I can handle that in #2876 as well.

> On this topic, there are also the `transfer_navigation_axes_to_signal_axes` and co, for which, I suspect that we should find a better pattern than creating a signal and then set the axis. Maybe it would be better to create the axis dictionary (using `to_dictionary`, for example) and to pass the dictionaries when creating the signal.

I think that is a good idea and naturally extends the first point.  In general when constructing signals I avoid passing an `axis` parameter mostly because it doesn't do a great job checking to make sure your axes are the correct size etc.  It is kind of annoying that I can't just set something like the unit instead you need a size etc to construct the axis.  So I am already doing as you suggest and setting things after the fact. I would image most people are doing something similar so it would probably be a minor change that would have little impact to how people are operating currently.

> No, but this PR suggest to use alternatives to set_axis. ;) I am not sure if set_axis is a good pattern to follow. What is the use > case for set_axis?

I would say that there isn't really a case for `set_axis`. There is a case for something which copies an axes.  Something which copies over the name, unit, and offset which is what I think most people are trying to do with the `set_axis` function.

jlaehne:
> I am not sure if `set_axis` is a good pattern to follow. What is the use case for `set_axis`?

Well, we use it in LumiSpy, when replacing e.g. a wavelength axis by an energy axis: https://github.com/LumiSpy/lumispy/blob/afd9723aec9ea45888ba60c03441909edbc64e33/lumispy/signals/luminescence_spectrum.py#L161

But maybe then the point of @magnunor above should be a bug report separate from this PR?

ericpre:
@CSSFrancis: great, this is reassuring as this confirms what I understood! :)
I think, we need to leave the diffraction vectors situation out of this PR, because this is already complicated enough. What matters is that it is going in the right direction.

> > I am not sure if `set_axis` is a good pattern to follow. What is the use case for `set_axis`?
> 
> Well, we use it in LumiSpy, when replacing e.g. a wavelength axis by an energy axis: https://github.com/LumiSpy/lumispy/blob/afd9723aec9ea45888ba60c03441909edbc64e33/lumispy/signals/luminescence_spectrum.py#L161
> 
> But maybe then the point of @magnunor above should be a bug report separate from this PR?

@jlaehne: ok, this is indeed a case, which is not [covered above](https://github.com/hyperspy/hyperspy/pull/2830#issuecomment-1097194139), because the type of axis changes. I would also leave it for later.

jlaehne:
OK, with this I see all the relevant points as solved.

PR #2833, opened 2021-10-08
Fix url "typo" in workflow: Atomap extension testing

magnunor:


ericpre (commented):
Would it make sense to enable the step running the atomap test suite for dev version only? Currently, this workflow doesn't run the atomap test suite.

jlaehne (approved):

ericpre (approved):

PR #2835, opened 2021-10-11
Bugfix around partly swept frame

nem1234:
### Description of the change
- Bug fix for JEOL-EDS (io_plugins/jeol.py) pts file reader
-  [Bug] When the final frame was partly recorded (measurement was manually interrupted), the readcube function summing up upper swept area of the final (broken) frame and the counts in upper and lower-frame become unbalanced.
-  [Bug] When the final frame was partly recorded, the readcube_frames function hangs-up (overrun hyperdata array)
- Count the frame index to detect the partly swept frame
- function readcube: ignore final broken frame
- function readcube_frames: extend array for 1 more frame. the axes value keeps the number of completely swept frame

### Progress of the PR
- [ ] add tests,
- [ ] ready for review.

```python
import hyperspy.api as hs
s = hs.load("broken.pts")
```

### DIFF
--- a/doc/user_guide/io.rst
+++ b/doc/user_guide/io.rst
@@ -1762,6 +1762,9 @@ Extra loading arguments
   dataset, it can be integer or a tuple of length 2 to define ``x`` and ``y``
   separetely and it must be a mutiple of the size of the navigation dimension.
   (default 1).
+- ``only_valid_data`` : ignore incomplete (partly aquiered, interrupted) frame when reading pts file.
+  (default True).
+
 
 Example of loading data downsampled, and with energy range cropped with the
 original navigation dimension 512 x 512 and the EDS range 40 keV over 4096
--- a/doc/user_guide/io.rst
+++ b/doc/user_guide/io.rst
@@ -1762,7 +1762,11 @@ Extra loading arguments
   dataset, it can be integer or a tuple of length 2 to define ``x`` and ``y``
   separetely and it must be a mutiple of the size of the navigation dimension.
   (default 1).
-- ``only_valid_data`` : ignore incomplete (partly aquiered, interrupted) frame when reading pts file.
+- ``only_valid_data`` : for ``pts`` file only, ignore incomplete and partly
+  acquired last frame, which typically occurs when the acquisition was
+  interrupted. When loading incomplete data (``only_valid_data=False``),
+  the missing data are filled with zeros. If ``sum_frames=True``, this argument
+  will be ignored to enforce consistent sum over the mapped area. 
   (default True).
 
 
--- a/hyperspy/io_plugins/jeol.py
+++ b/hyperspy/io_plugins/jeol.py
@@ -309,13 +309,13 @@ def read_pts(filename, scale=None, rebin_energy=1, sum_frames=True,
         data, swept = datacube_reader(rawdata, data, sweep,
                                       rebin_energy, channel_number,
                                width_norm, height_norm, np.iinfo(SI_dtype).max)
-        if sum_frames and sweep == swept:
+        if not sum_frames and sweep == swept:
             data = data[:sweep]  # dispose additional frame
         if sweep < swept:
             if sum_frames:
-                _logger.warning("Broken frame %d was ignored" % (swept))
+                _logger.info("Broken frame %d was ignored" % (swept))
             else:
-                _logger.warning("Broken frame %d was read, but not counted" % (swept))
+                _logger.info("Broken frame %d was read, but not counted" % (swept))
 
                 
         hv = meas_data_header["MeasCond"]["AccKV"]
--- a/hyperspy/io_plugins/jeol.py
+++ b/hyperspy/io_plugins/jeol.py
@@ -309,6 +309,8 @@ def read_pts(filename, scale=None, rebin_energy=1, sum_frames=True,
         data, swept = datacube_reader(rawdata, data, sweep,
                                       rebin_energy, channel_number,
                                width_norm, height_norm, np.iinfo(SI_dtype).max)
+        if sum_frames and sweep == swept:
+            data = data[:sweep]  # dispose additional frame
         if sweep < swept:
             if sum_frames:
                 _logger.warning("Broken frame %d was ignored" % (swept))
--- a/hyperspy/io_plugins/jeol.py
+++ b/hyperspy/io_plugins/jeol.py
@@ -463,7 +463,7 @@ def readcube(rawdata, hypermap, sweep, only_valid_data, rebin_energy,
             y = int((value - 36864) / height_norm)
             if y < previous_y:
                 frame_idx += 1
-                if only_valid_data and frame_idx >= sweep:
+                if frame_idx >= sweep:
                     # skip incomplete_frame
                     break 
             previous_y = y
--- a/hyperspy/io_plugins/jeol.py
+++ b/hyperspy/io_plugins/jeol.py
@@ -185,8 +185,7 @@ def read_img(filename, scale=None, **kwargs):
 
 def read_pts(filename, scale=None, rebin_energy=1, sum_frames=True,
              SI_dtype=np.uint8, cutoff_at_kV=None, downsample=1,
-             read_incomplete_frame=False,
-             **kwargs):
+             only_valid_data=True, **kwargs):
     fd = open(filename, "br")
     file_magic = np.fromfile(fd, "<I", 1)[0]
 
@@ -297,7 +296,7 @@ def read_pts(filename, scale=None, rebin_energy=1, sum_frames=True,
         sweep = meas_data_header["Doc"]["Sweep"]
         if not sum_frames:
             data_shape.insert(0, sweep)
-            if read_incomplete_frame:
+            if not only_valid_data:
                 data_shape[0] += 1 # for partly swept frame
             axes.insert(0, {
                 "index_in_array": 0,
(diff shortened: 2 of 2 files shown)


nem1234:
I'm sorry,
the title of this PR should contain keyword of JEOL-EDS....

hyperspy/io_plugins/jeol.py
```diff
@@ -292,9 +292,9 @@ def check_multiple(factor, number, string):
         pixel_time = meas_data_header["Doc"]["DwellTime(msec)"]
 
         data_shape = [height, width, channel_number]
+        sweep = meas_data_header["Doc"]["Sweep"]
         if not sum_frames:
-            sweep = meas_data_header["Doc"]["Sweep"]
-            data_shape.insert(0, sweep)
+            data_shape.insert(0, sweep + 1)  # +1 for partly swept frame
```
ericpre: Since, the data of the incomplete frame is not kept, there is no need to have `sweep + 1`?

hyperspy/io_plugins/jeol.py
```diff
@@ -306,9 +306,16 @@ def check_multiple(factor, number, string):
...
         datacube_reader = readcube if sum_frames else readcube_frames
-        data = datacube_reader(rawdata, data, rebin_energy, channel_number,
+        data, swept = datacube_reader(rawdata, data, sweep,
+                                      rebin_energy, channel_number,
                                width_norm, height_norm, np.iinfo(SI_dtype).max)
+        if sweep < swept:
+            if sum_frames:
+                _logger.warning("Broken frame %d was ignored" % (swept))
+            else:
+                _logger.warning("Broken frame %d was read, but not counted" % (swept))
```
ericpre: Is it worth warning the user about skipping the incomplete last frame? Quite often, users can see a warning as something being wrong or broken.

An alternative would be to explicitly mention in the dev guide that when the last frame is incomplete, it is skipped. 

What happen for acquisition, where there is only a single frame and this frame is incomplete. What is the value of `sweep` in this case? If `sweep=0`, it means that no data will be loaded?

ericpre (commented):
Thanks @nem1234, I added some suggestions.
Would it be possible to add a very small file (smallest possible ideally) to make sure that no regression is introduced in the future?

hyperspy/io_plugins/jeol.py
nem1234: I sometimes need to check the broken frame, and this code keeps it in the bottom of data array.
On the other hand, the sweep value and axes data keeps the number of good frames. Is this incorrect?
In case of not sum_frames, the broken frame is kept in data array, and user can access through the data array. 
In case of sum_frames, just ignore the broken frame because summing up the broken frame is usually meaningless.

nem1234: If just dispose the broken frame, keep original line 297, and replace line 490-497
<pre>
                if frame_idx > sweep:
                    # frame_idx < sweep : normal operation
                    # frame_idx == sweep when sweeping was interrupted
                    # frame_idx > sweep: unknown error
                    raise ValueError("The frame number is too large")
</pre>
with 
 <pre>
               if frame_idx >= sweep:
                    break
</pre>

hyperspy/io_plugins/jeol.py
nem1234: OK,  I recognize that the log level should be lower.
(Or, make some comments into the metadata)
In my opinion, users should have a way to know what is happen during data loading. 

>What happen for acquisition, where there is only a single frame and this frame is incomplete. What is the value of sweep in this case? If sweep=0, it means that no data will be loaded?

In this code, I intended to make a "blank" image for sum_frames mode.
In case of "not sum_frames", this make additional partly swept image outside the data cube shown in axes values.

In the near future I'll prepare a small data for these cases. (Currently only quite large data is available in my hand).

nem1234:
Today I post a revised code.
Logging level for broken frame changed from "WARNING" to "INFO"
Remove additional frame from data cube when no broken frame is found

hyperspy/io_plugins/jeol.py
ericpre: > I sometimes need to check the broken frame, and this code keeps it in the bottom of data array.
On the other hand, the sweep value and axes data keeps the number of good frames. Is this incorrect?

Yes, you are correct! I muss have misread in my first pass and understood that the data of the incomplete frame was always skipped.

> In case of not sum_frames, the broken frame is kept in data array, and user can access through the data array.
In case of sum_frames, just ignore the broken frame because summing up the broken frame is usually meaningless.

Yes, this is a very sensible approach.

hyperspy/io_plugins/jeol.py
```diff
@@ -306,9 +306,18 @@ def check_multiple(factor, number, string):
 
         data = np.zeros(data_shape, dtype=SI_dtype)
         datacube_reader = readcube if sum_frames else readcube_frames
-        data = datacube_reader(rawdata, data, rebin_energy, channel_number,
+        data, swept = datacube_reader(rawdata, data, sweep,
+                                      rebin_energy, channel_number,
                                width_norm, height_norm, np.iinfo(SI_dtype).max)
+        if not sum_frames and sweep == swept:
+            data = data[:sweep]  # dispose additional frame
```
ericpre: Is the value `axes[0]['size']` correct for both cases: with and without incomplete last frame?

hyperspy/io_plugins/jeol.py
```diff
@@ -306,9 +306,18 @@ def check_multiple(factor, number, string):
...
+        if sweep < swept:
+            if sum_frames:
+                _logger.info("Broken frame %d was ignored" % (swept))
```
ericpre: Maybe to be more clear, something along the line of:
```suggestion
                _logger.info("The last frame (sweep) is incomplete because the acquisition stopped during this frame. The partially acquired frame is ignored. Use `sum_frames` to read all frames individually, including the last partially completed frame.))
```

hyperspy/io_plugins/jeol.py
```diff
@@ -306,9 +306,18 @@ def check_multiple(factor, number, string):
...
+            else:
+                _logger.info("Broken frame %d was read, but not counted" % (swept))
```
ericpre: In case of `sum_frames = False`, the data is kept, should it be removed, then?

ericpre (commented):
It may be worth considering adding an option to specify if the incomplete frame needs to be read or not?

hyperspy/io_plugins/jeol.py
nem1234: axes[0]['size'] always show the number of complete frames, and the incomplete last frame will be ignored when integrate along some axis. To access the incomplete frame, direct access to the data array is needed. 

> It may be worth considering adding an option to specify if the incomplete frame needs to be read or not?

By adding an option to read a broken frame, these two lines should be modified.

Is there a common option to load an incomplete frame?

hyperspy/io_plugins/jeol.py
nem1234: "The last frame (sweep) is incomplete because the acquisition stopped during this frame. The partially acquired frame is ignored. Use `sum_frames = False, read_incomplete_frame = True` to read all frames individually, including the last partially completed frame."

hyperspy/io_plugins/jeol.py
ericpre: > axes[0]['size'] always show the number of complete frames, and the incomplete last frame will be ignored when integrate along some axis. To access the incomplete frame, direct access to the data array is needed.
> 

The axis size needs to match the length of the corresponding axis of the numpy array.

> Is there a common option value to load an incomplete frame?

Actually, there is already one for the emi/ser reader which is fairly close: https://hyperspy.readthedocs.io/en/latest/user_guide/io.html#id5.

nem1234:
Patch was revised.
* New option: only_valid_data (default = True)
* Log messages are revised.
* Axes[0]['size'] and number of frames in data cube become consistent.

doc/user_guide/io.rst
```diff
@@ -1762,6 +1762,9 @@ Extra loading arguments
   dataset, it can be integer or a tuple of length 2 to define ``x`` and ``y``
   separetely and it must be a mutiple of the size of the navigation dimension.
   (default 1).
+- ``only_valid_data`` : ignore incomplete (partly aquiered, interrupted) frame when reading pts file.
```
ericpre:
```suggestion
- ``only_valid_data`` : for ``pts`` file only, ignore incomplete and partly acquired last frame, which typically occurs when the acquisition was interrupted. When loading incomplete data (``only_valid_data=False``), the missing data are filled with zeros. If ``sum_frames=True``, this argument will be ignored to enforce consistent sum over the mapped area. 
```

ericpre (approved):
This is very good, I suggested one improvement for the user guide.

As you mentioned in https://github.com/hyperspy/hyperspy/pull/2835#discussion_r726181362, you don't have a small dataset at hand but you will provide one soon. Normally, we wouldn't merge a PR without a test but if it makes your life easier to continue working on the lazy implementation, then I think, we should make an exception here!

hyperspy/io_plugins/jeol.py
```diff
@@ -67,7 +67,8 @@
 }
 
 
-def file_reader(filename, **kwds):
+def file_reader(filename, file_encoding = "iso8859-1", **kwds):
```
ericpre: Why this change?

ericpre (commented):
See question on encoding.

hyperspy/io_plugins/jeol.py
nem1234: I’m sorry, this is my mistake when merging my private code.
I will cancel it.

nem1234: It is another problem that is not in this PR.
So, I'll post again with another PR when it is ready.

Our JEOL EDS software (not a newest version) does not support utf-8 encoding.
It only affect when some comments or directory names are written in Japanese, and the code can be decoded as different character in utf-8 charset.
Basically they should be decoded as 'shift_jis' or 'iso8859-1', not 'utf-8'.
Of cause, newer version of JEOL EDS will support 'utf-8', but automatic charset selection sometimes makes error.

ericpre: Thanks for the explanation, this sounds sensible. As you said, it would be better to do this in a separate PR, as it usually allow to write changes more cleanly and would be easier for future reference, etc.

ericpre:
@nem1234, for information, in 9d8a62a, I squashed the last few commits to remove adding and remove encoding changes and clean the git history. After this, I "force-pushed" to the branch of this PR, which means this branch have been overwritten and that local and remote branches have diverged. If you were to continue working on this branch, git would most likely complain when trying to push changes from your local branch to the remote. In this case, you would need to check out this branch first and then add your changes.

Once CI is happy, I will merge.

PR #2837, opened 2021-10-20
Bug fix in _map_iterate method on lazy data with function that changes axes

SyHuang19:
### Description of the change
Fixed bug in `_map_iterate` method that gives incorrect output signal size when output axes of the function are different from the input signal. Previous version would feed `map_blocks` function in dask.array with incorrect new axis index.

### Progress of the PR
- [N/A ] update docstring (if appropriate),
- [N/A] update user guide (if appropriate),
- [x] add tests,
- [x] ready for review.

```python
import hyperspy.api as hs
import numpy as np
import dask.array as da
s = hs.signals.Signal2D(da.zeros((10,10,12,12))).as_lazy()
s.rechunk(nav_chunks=(5,5))
def test_axes(data):
    return np.zeros((5))
sl = s.map(test_axes,inplace=False)
```

Before the fix the output signal dimensions are (10, 10|10) instead of (10, 10|5).

### DIFF
--- a/hyperspy/tests/signals/test_map_method.py
+++ b/hyperspy/tests/signals/test_map_method.py
@@ -366,6 +366,7 @@ class TestLazyMap:
 
         s_out = self.s.map(function=f, inplace=False)
         assert s_out.data.shape[2:] == output_signal_size
+        assert s_out.axes_manager.signal_shape == output_signal_size[::-1]
 
 
 @pytest.mark.parametrize('ragged', [True, False, None])
--- a/hyperspy/tests/signals/test_map_method.py
+++ b/hyperspy/tests/signals/test_map_method.py
@@ -359,6 +359,14 @@ class TestLazyMap:
                                inplace=False)
         np.testing.assert_array_equal(s_out.mean(axis=(2, 3)).data, iter_array)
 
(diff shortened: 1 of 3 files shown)


hyperspy/tests/signals/test_map_method.py
```diff
@@ -359,6 +359,14 @@ def test_map_iterate_array(self):
                                inplace=False)
         np.testing.assert_array_equal(s_out.mean(axis=(2, 3)).data, iter_array)
 
+    @pytest.mark.parametrize("output_signal_size", [(3,), (3, 4), (3, 4, 5)])
+    def test_map_output_signal_size(self, output_signal_size):
+        def f(data):
+            return np.ones(output_signal_size)
+
+        s_out = self.s.map(function=f, inplace=False)
```
ericpre, line 367: Can you also check the `axes_manager` is correct?

ericpre (approved):
Thanks for the bug fix @SyHuang19, this looks good! I have left a suggestion for improvement.

As a note for ourselves, this will conflict with #2703.

hyperspy/tests/signals/test_map_method.py
SyHuang19: Thanks for the suggestion. `axes_manager` is added to test.

ericpre:
All good, thanks @SyHuang19!

PR #2838, opened 2021-10-21
io_plugins/jeol.py : Get scale information from pts file itself.

nem1234:
### Description of the change

- Read scale information from pts file it self.
- In case of pts file reading without asw file, scale is read from pts file header.
- Updating the test will be needed  

### Progress of the PR
- [ ] add tests,
- [ ] ready for review.

```python
import hyperspy.api as hs
data = hs.load("file.pts")
print(data.axes_manager)
```

### DIFF
--- a/hyperspy/tests/io/test_jeol.py
+++ b/hyperspy/tests/io/test_jeol.py
@@ -75,12 +75,12 @@ def test_load_project():
     # check scale (image)
     filename = TESTS_FILE_PATH / 'Sample' / '00_View000' / test_files[1]
     s1 = hs.load(filename)
-    np.testing.assert_allclose(s[0].axes_manager[0].scale,s1.axes_manager[0].scale)
+    np.testing.assert_allclose(s[0].axes_manager[0].scale, s1.axes_manager[0].scale)
     assert s[0].axes_manager[0].units == s1.axes_manager[0].units
     # check scale (pts)
     filename = TESTS_FILE_PATH / 'Sample' / '00_View000' / test_files[7]
     s2 = hs.load(filename)
-    np.testing.assert_allclose(s[6].axes_manager[0].scale,s2.axes_manager[0].scale)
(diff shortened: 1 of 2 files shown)


hyperspy/tests/io/test_jeol.py
```diff
@@ -72,6 +72,17 @@ def test_load_project():
     np.testing.assert_allclose(s[-1].axes_manager[2].scale, 0.00999866)
     assert s[-1].axes_manager[2].name == 'Energy'
 
+    # check scale (image)
+    filename = TESTS_FILE_PATH / 'Sample' / '00_View000' / test_files[1]
+    s1 = hs.load(filename)
+    np.testing.assert_allclose(s[0].axes_manager[0].scale,s1.axes_manager[0].scale)
+    assert s[0].axes_manager[0].units == s1.axes_manager[0].units
```
ericpre, line 79: Can you also check it against the actual value? For both, the units and scale

hyperspy/tests/io/test_jeol.py
```diff
@@ -81,11 +92,11 @@ def test_load_image():
...
     assert s.data.shape == (512, 512)
     assert s.axes_manager.signal_dimension == 2
-    assert s.axes_manager[0].units == 'px'
-    assert s.axes_manager[0].scale == 1
+#    assert s.axes_manager[0].units == 'px'
+#    assert s.axes_manager[0].scale == 1
     assert s.axes_manager[0].name == 'x'
-    assert s.axes_manager[1].units == 'px'
-    assert s.axes_manager[1].scale == 1
+#    assert s.axes_manager[1].units == 'px'
```
ericpre: Same as above, can you use the actual values?

ericpre (commented):
This looks good, the failure of the test suite on Github CI are related to the latest release of h5py and can be ignored - see https://hyperspy.org/hyperspy-doc/current/dev_guide/testing.html#continuous-integration-ci for explanation why it is currently failing on Github CI but not on Azure pipelines.

ericpre (approved):
Looks good!

PR #2841, opened 2021-10-24
jeol.py : Bugfix for broken sweep value. 

nem1234:
### Description of the change
Add support on asw reader to getmultiple files from pts file reader.
When the pts file reader will return a list of both eds spectrum image and STEM/SEM image (not yet supported) dictionaries,
they should be listed in the asw file reader.

### Progress of the PR
- [ ] add tests,
- [ ] ready for review.

not yet


nem1234:
Add additional test data, with different type of broken "Sweep" value
Add test code for the test data
Modify pts reader to support such broken data.
(scanning of PTS file before loading is required because it is difficult to recover after integrating incomplete frame)

ericpre:
@nem1234, I don't think we need this PR anymore since this has been done in #2846?

nem1234:
Thank you.
This was due to my misunderstanding of git branch.

PR #2843, opened 2021-10-24
Fix for h5py 3.5

ericpre:
h5py 3.5 breaks saving file because of a `h5py.dataset` and `numpy.ndarray` comparison which is actually never used!
See for example, this build https://github.com/hyperspy/hyperspy/actions/runs/1373766813 on github CI - azure pipeline is using conda-forge packages and h5py 3.5 is not yet available.

### Progress of the PR
- [x] Remove unused code, which breaks with h5py 3.5
- [n/a] add tests,
- [x] ready for review.

### DIFF
--- a/hyperspy/io_plugins/hspy.py
+++ b/hyperspy/io_plugins/hspy.py
@@ -595,9 +595,6 @@ def overwrite_dataset(group, data, key, signal_axes=None, chunks=None, **kwds):
             # if the shape or dtype/etc do not match,
             # we delete the old one and create new in the next loop run
             del group[key]
-    if dset == data:
-        # just a reference to already created thing
-        pass
     else:
         _logger.info(f"Chunks used for saving: {dset.chunks}")
         if isinstance(data, da.Array):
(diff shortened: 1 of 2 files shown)


hyperspy/io_plugins/hspy.py
```diff
@@ -595,9 +595,6 @@ def overwrite_dataset(group, data, key, signal_axes=None, chunks=None, **kwds):
             # if the shape or dtype/etc do not match,
             # we delete the old one and create new in the next loop run
             del group[key]
-    if dset == data:
-        # just a reference to already created thing
-        pass
     else:
```
jlaehne, line 606: This `else` statement seems orphaned now!?

hakonanes:
Great catch! I'm seeing this error in our [weekly check](https://github.com/pyxem/kikuchipy/runs/3993511535?check_suite_focus=true#step:6:427) of demanding documentation notebooks.

hakonanes (commented):
I've never written a while else statement before. It [seems like](https://www.pythontutorial.net/python-basics/python-while-else/) the else block will always be executed if the while condition becomes False. Is there a scenario where the while loop is terminated by a return or break, the execution continues, but the else block should be skipped? If not, would it be more readable if the else was just removed?

ericpre:
Yes, I agree, this is one of these where less is better! I have remove the else statement in 287868d.

hakonanes (approved):
Great!

jlaehne (approved):

hyperspy/io_plugins/hspy.py
jlaehne: OK, resolved by now - my comment was somehow not sent out earlier.

PR #2846, opened 2021-10-27
jeol.py:  Add lazy loading for spectrum image, and add loader for SEM/STEM image in pts file

nem1234:
### Description of the change
- Add file reader for SEM/STEM image in pts file
- Add lazy loading for spectrum image in pts file
- Add function to load selected frames in pts file
- Enable multpule image loading from pts file for asw file loader.
- Add test code and test images for incomplete frames.

### Progress of the PR
- [x] add tests,
- [x] ready for review.

```python
import hyperspy.api as hs
import numpy as np
s = hs.load( filename.pts, si_lazy=True, read_em_image=True, frame_list=[0,2,3])
```


hyperspy/io_plugins/jeol.py
```diff
@@ -94,13 +94,19 @@ def file_reader(filename, **kwds):
                                     if sub_ext in extension_to_reader_mapping.keys():
                                         reader_function = extension_to_reader_mapping[sub_ext]
                                         d = reader_function(file_path, scale, **kwds)
-                                        dictionary.append(d)
+                                        if isinstance(d, list):
```
nem1234, line 98: PTS file loader now returns spectrum image and optionally SEM/TEM image.
So, asw file reader need to accept list of images.

hyperspy/io_plugins/jeol.py
```diff
@@ -267,8 +319,45 @@ def check_multiple(factor, number, string):
             channel_number = int(
                 np.round((cutoff_at_kV - energy_offset) / energy_scale)
                 )
+        # pixel time in milliseconds
+        pixel_time = meas_data_header["Doc"]["DwellTime(msec)"]
 
-        axes = [
+        data_shape = [height, width, channel_number]
+        # Sweep value is not reliable
```
nem1234: Sweep value in the pts file header is not reliable.
In some case, sweep value shows the number of valid frames, and in another case, it shows sum of the numbers of valid and incomplete frame (may be version dependent.
So, pre-scanning the pts file is needed to check it.
(Original JEOL software also shows wrong number of frames, and operation with last incomplete frame seems unstable)

hyperspy/io_plugins/jeol.py
```diff
@@ -454,63 +543,242 @@ def parsejeol(fd):
...
                     raise ValueError("The range of the dtype is too small, "
                                      "use `SI_dtype` to set a dtype with "
                                      "higher range.")
-    return hypermap, frame_idx + 1
-
 
-@numba.njit(cache=True)
-def readcube_frames(rawdata, hypermap, sweep, only_valid_data, rebin_energy, channel_number,
-             width_norm, height_norm, max_value):  # pragma: no cover
+def readcube(rawdata, frame_ptr_list, frame_list,
```
nem1234: readcube and readcube_frames functions are merged into one function.
Inner loop "readframe_dense"(originally it is part of readcube) and "readframe_sparse"(for lazy loading) are optimized by numba.
Outer loop, readcube does not use numba.

hyperspy/io_plugins/jeol.py
```diff
@@ -454,63 +543,242 @@ def parsejeol(fd):
...
+    em_image : np.ndarray(width, height), dtype = np.uint16 or np.uint32
+    	np.ndarray to store decoded SEM/TEM image.
     """
-    frame_idx = 0
-    previous_y = 0
+
+    n_frames = len(frame_list)
+    EM_dtype = np.uint16
+    frame_step = 1
+    if sum_frames:
```
nem1234: In case of sum_frames,  spectrum image and SEM/STEM image are summing up to same frame.
To avoid overflow on integration of SEM/STEM image, data type of np.uint32 is selected for 16 frames and over.
(range of image intensity in each frame is 0-4095 (0-0xfff))

doc/user_guide/io.rst
```diff
@@ -1807,28 +1807,43 @@ files automatically.
...
-  (default True).
+- ``read_em_image`` : bool, default False.
+  for ``pts`` file only, If ``read_em_image=True``,
+  read SEM/STEM image from pts file if available. Spectrum Image and
+  SEM/STEM Image will be returned.
+- ``frame_list`` : list of integer or None, default None
+  for ``pts`` file only, frames in frame_list will be loaded.
+  for example, ``frame_list=[1,3]`` means second and forth frame is selected.
+  If ``None``, all frames are loaded.
+- ``si_lazy`` : bool, default False
```
ericpre: You prefer not to rely on the standard `lazy` argument to avoid loading images lazily too?

hyperspy/io_plugins/jeol.py
ericpre: Can you please add these comment in the code too? :)
For future reference, it is useful to the explanations supporting the observations! Also, when it is written in the code, it can save us having to dig on github using `git blame` with the hope of find the reasoning.

hyperspy/io_plugins/jeol.py
```diff
@@ -267,8 +319,45 @@ def check_multiple(factor, number, string):
...
+            sweep = len(frame_ptr_list) - 1
+        else:
+            sweep = len(frame_ptr_list) - 2
+        if not frame_list:
+            frame_list = range(sweep)
+        frame_list = list(frame_list)
+        frame_list2 = []
+        for frame_idx in frame_list:
+            if frame_idx < 0 or frame_idx >= sweep:
+                _logger.info(f"Ignoreing frame {frame_idx} : Selected frame is not found in pts data.")
```
ericpre:
```suggestion
                _logger.info(f"Ignoring frame {frame_idx} : Selected frame is not found in pts data.")
```

hyperspy/io_plugins/jeol.py
```diff
@@ -267,8 +319,45 @@ def check_multiple(factor, number, string):
...
                 np.round((cutoff_at_kV - energy_offset) / energy_scale)
                 )
+        # pixel time in milliseconds
+        pixel_time = meas_data_header["Doc"]["DwellTime(msec)"]
 
-        axes = [
+        data_shape = [height, width, channel_number]
+        # Sweep value is not reliable
+        # sweep = meas_data_header["Doc"]["Sweep"]
+        frame_ptr_list, last_valid, has_em_image = pts_prescan(rawdata, width, height)
```
ericpre: The drawback of this approach is to have to go through the dataset twice, which for large file is not insignificant. An alternative may be to try to use the unreliable `meas_data_header["Doc"]["Sweep"]` and if it works fine, then we continue otherwise, we fallback to the slower route of the "prescan" followed by the actual read.
I am not sure if I understand correctly the issue and therefore if this alternative is sensible... 

To make a sensible judgement call, it would good to have an idea if this issue is usual or it is only limited to some version of software, etc. In case, this issue isn't relevant for most systems, it would be shame to always double the reading time.

hyperspy/io_plugins/jeol.py
```diff
@@ -267,8 +319,45 @@ def check_multiple(factor, number, string):
...
+        if not frame_list:
+            frame_list = range(sweep)
+        frame_list = list(frame_list)
+        frame_list2 = []
+        for frame_idx in frame_list:
+            if frame_idx < 0 or frame_idx >= sweep:
+                _logger.info(f"Ignoreing frame {frame_idx} : Selected frame is not found in pts data.")
+                continue
+            frame_list2.append(frame_idx)
+        frame_list = np.array(frame_list2)
```
ericpre: It is a bit suggestive, I find the following more clear:
```suggestion
        if not frame_list:
            frame_list = np.arange(sweep)
        frame_list = np.asarray(frame_list)
        wrong_frames_idx = np.argwhere((frame_list<0) | (frame_list>=sweep))
        if len(wrong_frames_idx) > 0:
            wrong_frames = frame_list[wrong_frames_idx].flatten().tolist()
            _logger.info(f"Ignoring frame {wrong_frames} : Selected frame is not found in pts data.")
```

hyperspy/io_plugins/jeol.py
```diff
@@ -454,63 +543,244 @@ def parsejeol(fd):
...
+
+    Returns
+    -------
+    None
+
+    hypermap and em_image array will be modified
+    """
+    for value in rawdata:
+        dtype = value & 0xf000
+        dval = value & 0xfff
```
ericpre: Maybe as for `pts_prescan` use `value_type` and `value` since `dtype` is used widely in numpy? The use of `dtype` is a bit misleading for the reader.

hyperspy/io_plugins/jeol.py
```diff
@@ -186,7 +192,55 @@ def read_img(filename, scale=None, **kwargs):
...
+    downsample : int or [int, int]
+    	downsample along spacial axes to reduce memory size of spectrum image
+	value must be 2^n.
+	default: 1 (do not downsample)
+    only_valid_data : bool, default True
+    	read incomplete frame if only_valid_data == False
+        (usually interrupting mesurement makes incomplete frame)
+    read_em_image : bool, default False
+        read SEM/STEM image from pts file if read_em_image == True
+    frame_list : list of int [fr0, fr1, ...] or array [[fr0, dx0,dy0,dz0],...], default None
```
nem1234: add shifting of frames for express drift correction.
due to the memory usage, it is difficult to make drift correction after reading 4-dimensional data

hyperspy/io_plugins/jeol.py
ericpre: Yes, this sounds very sensible indeed! Can you please add this comment in the code itself for future reference: at first read, the `n_frames > 16:` seems arbitrary but it is not at all!

hyperspy/io_plugins/jeol.py
ericpre: What's about using two different arguments, for example, `frame_selection` and `drift_correction` instead of `frame_list` only? Conceptually, there are different thing and this would be easier to describe in the docstring and user guide.

Also it would be good to be able to do the following:

```python
s_list = hs.load('file.asw', select_type='images', sum_frames=False)
s_images = s_list[0] # select the correct signal
shifts = s_images.estimate_shift2D()
hs.load('file.asw', drift_correction=shifts, sum_frames=False)
```

If the `frame_ptr_list` is saved to `original_metadata`
```python
frame_ptr_list = s_images.original_metadata.frame_ptr_list
hs.load('file.asw', drift_correction=shifts, sum_frames=False, frame_start_index=frame_ptr_list)

File: hyperspy/io_plugins/jeol.py (hunk 7 above)
Comment (ericpre) [reply hyperspy/io_plugins/jeol.py]: Add `frame_ptr_list` to the `original_metadata` dictionary, so that it could be reused when reopening the file, for example to align frames?

File: hyperspy/io_plugins/jeol.py [hunk 11]
```diff
@@ -454,63 +558,258 @@ def parsejeol(fd):
 
 
 @numba.njit(cache=True)
-def readcube(rawdata, hypermap, sweep, only_valid_data, rebin_energy, 
-            channel_number, width_norm, height_norm, max_value):  # pragma: no cover
-    frame_idx = 0
-    previous_y = 0
+def pts_prescan(rawdata, width, height):   # pragma: no cover
```
ericpre: Is this function as slow as `readframe_dense`? If so, this is going to take a significant portion of the loading time (~50%?) and it may be worth trying to find an alternative way to know where the frames start.

ericpre (commented):
There is a lot of good work in this PR! I have left some questions/comments for discussion.

hyperspy/io_plugins/jeol.py
nem1234: Redesign pts reader

hyperspy/io_plugins/jeol.py
```diff
@@ -186,7 +192,54 @@ def read_img(filename, scale=None, **kwargs):
 
 def read_pts(filename, scale=None, rebin_energy=1, sum_frames=True,
              SI_dtype=np.uint8, cutoff_at_kV=None, downsample=1,
-             only_valid_data=True, **kwargs):
+             only_valid_data=True, read_em_image=False,
+             frame_list=[], frame_start_index=[], frame_shifts=[],
```
ericpre: It is usually better not to set a default value for argument, which are of mutable type - see https://docs.quantifiedcode.com/python-anti-patterns/correctness/mutable_default_value_as_argument.html.

A better pattern is set these values to `None` in the definition and inside the function, set the default value when it is `None`.

It is possibly not an issue here, but more a good practise to enforce!

hyperspy/io_plugins/jeol.py
```diff
@@ -359,18 +447,41 @@ def check_multiple(factor, number, string):
...
+                mode: {
+                    "beam_energy": hv,
+                    "magnification": meas_data_header["MeasCond"]["Mag"],
+                },
+            },
+            "General": {
+                "original_filename": os.path.basename(filename),
+                "date": datefile.date().isoformat(),
+                "time": datefile.time().isoformat(),
+                "title": "SEM/STEM Image"
```
ericpre: In case, there are several images, it is possible to retrieve some information from the header to differentiate these images and add it to the title?

hyperspy/io_plugins/jeol.py
```diff
@@ -453,64 +564,306 @@ def parsejeol(fd):
...
+            # accept last frame
+            if lazy:
+                data_list.append(frame_data)
+            frame_num += 1
+            target_frame_num += frame_step
+        else:
+            # incomplete data, not accepted
+            if sum_frames:
+                # cancel last frame
+                print("Removing data of last frame = ", frame_idx)
```
ericpre: We use the logger instead of calling `print` so that the user can control the level of verbosity.

hyperspy/io_plugins/jeol.py
```diff
@@ -453,64 +583,383 @@ def parsejeol(fd):
...
+        pv[0,:] = frame_count
+        pv[4,:] = 1
+        ptr += flen
+        frame_count += 1
+    if sum_frames:
+        data_shape = [height, width, channel_number]
+        ar_s = sparse.COO(v[1:4], v[4], shape=data_shape)
+    else:
+        data_shape = [frame_count, height, width, channel_number]
+        ar_s = sparse.COO(v[0:4], v[4], shape=data_shape)
```
ericpre: Use `DenseSliceCOO` from https://github.com/hyperspy/hyperspy/blob/b5815c01d303d6598484d97096622708b38e41e0/hyperspy/misc/io/fei_stream_readers.py#L26-L35

```suggestion
        data_shape = [height, width, channel_number]
        ar_s = DenseSliceCOO(v[1:4], v[4], shape=data_shape)
    else:
        data_shape = [frame_count, height, width, channel_number]
        ar_s = DenseSliceCOO(v[0:4], v[4], shape=data_shape)
```

To avoid the following error:
```python

  File "/home/eric/Dev/hyperspy/hyperspy/_signals/lazy.py", line 127, in compute
    self.data = data

  File "/home/eric/Dev/hyperspy/hyperspy/signal.py", line 2411, in data
    self._data = np.atleast_1d(np.asanyarray(value))

  File "/opt/miniforge/lib/python3.9/site-packages/numpy/core/_asarray.py", line 171, in asanyarray
    return array(a, dtype, copy=False, order=order, subok=True)

  File "/opt/miniforge/lib/python3.9/site-packages/sparse/_sparse_array.py", line 229, in __array__
    raise RuntimeError(

RuntimeError: Cannot convert a sparse array to dense automatically. To manually densify, use the todense method.
```

hyperspy/io_plugins/jeol.py
```diff
@@ -453,64 +583,383 @@ def parsejeol(fd):
...
+        if sum_frames:
+            # the first frame has integrated intensity
+            return hypermap[0,:height,:width], em_image[0,:height,:width], has_em_image, frame_num, frame_start_index, valid, max_shift
+        else:
+            return hypermap[:target_frame_num,:height,:width], em_image[:target_frame_num,:height,:width], has_em_image, frame_num, frame_start_index, valid, max_shift
+        
+    # for lazy loading
+    length = 0
+    for d in data_list:
+        length += len(d)
```
ericpre:
```suggestion
    length = np.sum([len(d) for d in data_list])
```

ericpre:
@nem1234, is it ready for review? This is already quite a lot in this PR and it would be good to finish it and get it merged.
It is easier to work with small self-contained PR (to review and discuss), rather than large PR, it you want to implement other feature, it would be better to do it in a separate PR.

nem1234:
This PR become ready for review.
I'm sorry it become large.

Performance for lazy loading (dask - COO array) is still not good.
I'll revise it in near future.

hyperspy/io_plugins/jeol.py
nem1234: added comments in python code

hyperspy/io_plugins/jeol.py
nem1234: frame_shifts option is introduced that is compatible to the result of estimate_shift2D

doc/user_guide/io.rst
nem1234: merge to standard "lazy" option.

nem1234:
due to my limited knowledge on git, recent codes was not pushed to here.
The performance of lazy loading is not good yet. 
I'll brush up it may be in the next PR.

hyperspy/io_plugins/jeol.py
```diff
@@ -186,7 +192,54 @@ def read_img(filename, scale=None, **kwargs):
...
+    read_em_image : bool, default False
+        read SEM/STEM image from pts file if read_em_image == True
+    frame_list : list of int [fr0, fr1, ...] or array [[fr0, dx0,dy0,dz0],...], default None
+    	list of frame numbers to be read (None for all data)
+    	if dx, dy, dz are specified, each images are shifted. dx, dy, dz are integers, in unit of pixel.
+	This is useful for express drift correction. (not suitable for a detailed analysis)
+    si_lazy : bool or tuple of chunksize, default False
+    	read spectrum image into sparse array if si_lazy == True
+    	SEM/STEM image is always read into dense array (np.ndarray)
+    	if tuple is specified, dask.array is generated using chunksize=si_lazy
```
ericpre: The si_lazy option has been removed:
```suggestion
```

hyperspy/io_plugins/jeol.py
```diff
@@ -186,7 +192,54 @@ def read_img(filename, scale=None, **kwargs):
...
+    lazy : bool, default False
+	set lazy flag not only spectrum image but also other data,
+	if lazy == True. This also set si_lazy = True.
+	Only the spectrum image data is read as a sparse array,
```
ericpre: Update since `si_lazy` has been removed and also mention sparse array - see other comments regarding lazy loading/sparse array.

doc/user_guide/io.rst
```diff
@@ -1840,28 +1807,46 @@ files automatically.
...
+  will be ignored to enforce consistent sum over the mapped area. 
+- ``read_em_image`` : bool, default False.
+  for ``pts`` file only, If ``read_em_image=True``,
+  read SEM/STEM image from pts file if available. In this case, both
+  spectrum Image and SEM/STEM Image will be returned as list.
+- ``frame_list`` : list of integer or None, default None
+  for ``pts`` file only, frames in frame_list will be loaded.
+  for example, ``frame_list=[1,3]`` means second and forth frame will be loaded.
+  If ``None``, all frames are loaded.
+* - ``frame_shifts`` : list of [[int, int]...] or [[int, int, int], ...] or None, default None
```
ericpre: Can you describe it with more details? For example, when reading it, it is obvious to me why it is possible to pass a "list of lists of length 2" (`[[int, int]...]`) but also a "list of lists of length 3" (`[[int, int, int], ...]`)?
Can you explain to what corresponds the first `int` and the second `int`, etc?

ericpre (commented):
Thanks @nem1234, this is in very good shape, I added some small comments/suggestions and there is a test failing on CI (see https://hyperspy.readthedocs.io/en/stable/dev_guide/testing.html#continuous-integration-ci if you want to know more about is meant with CI) because of memory usage:
```python
E           numpy.core._exceptions._ArrayMemoryError: Unable to allocate 15.0 GiB for an array with shape (15, 512, 512, 4096) and data type uint8
```
CI have limited memory, possibly about 2GB, and reduce memory usage of this specific test in one way or another should fix the test failure.

I will need to have another look later at some of the latest changes.

hyperspy/io_plugins/jeol.py
nem1234: fixed.

hyperspy/io_plugins/jeol.py
nem1234: fixed.

hyperspy/io_plugins/jeol.py
nem1234: Thank you.
Fixed as your suggestion

hyperspy/io_plugins/jeol.py
nem1234: fixed.

hyperspy/io_plugins/jeol.py
nem1234: This is a debugging code.
already removed.

hyperspy/io_plugins/jeol.py
nem1234: fixed.

doc/user_guide/io.rst
nem1234: fixed

hyperspy/io_plugins/jeol.py
nem1234: May be fixed.
new title should be   "S(T)EM Image extracted from xxxx.pts".

ericpre: Sorry, my comment wasn't very clear, I was referring to the type of detector, e.g. `BF`, `DF`, etc. If this information is available, it is useful to have it in the title, for example, when one want to do some processing on images from a specific detector, etc.
We should leave it for another PR! :)

hyperspy/tests/io/test_jeol.py
```diff
@@ -245,3 +272,141 @@ def test_shift_jis_encoding():
...
+                only_valid_data=False, sum_frames=True, lazy=False)
+    assert np.array_equal(s1[0], s2[0].data)
+    assert np.array_equal(s1[1], s2[1].data)
+
+def test_pts_frame_shift():
+    dir = TESTS_FILE_PATH2
+    dir2p = dir / 'Sample' / '00_Dummy-Data'
+    file = str(dir2p / test_files2[16])
+    #    dir = TESTS_FILE_PATH
+    #    file = str(dir / 'testpattern1.pts')
```
ericpre: Is it worth testing these files? If not, they should be removed? If they need to be removed, it would be good to remove them from the history to avoid unnecessarily increasing the size of the git repository. If is a bit annoying to do, I can do ti if necessary.

hyperspy/io_plugins/jeol.py
```diff
@@ -453,64 +594,394 @@ def parsejeol(fd):
...
+    class DenseSliceCOO(sparse.COO):
+        """Just like sparse.COO, but returning a dense array on indexing/slicing"""
+
+        def __getitem__(self, *args, **kwargs):
+            obj = super().__getitem__(*args, **kwargs)
+            try:
+                return obj.todense()
+            except AttributeError:
+                # Indexing, unlike slicing, returns directly the content
+                return obj
```
ericpre: Instead of copying it here, it is better to import it from `hyperspy/hyperspy/misc/io/fei_stream_readers` to avoid duplicating code.
Ideally, `DenseSliceCOO` should be moved somewhere else because it is generic enough to be used elsewhere than in the velox end reader but we can leave this for another time!

ericpre (commented):
I added some comments: some of them can/should be done in another PR but the one about the test file (comment test code about `testpattern1.pts`) should be sorted in this PR.

hyperspy/io_plugins/jeol.py
nem1234: I'm sorry, but I couldn't find the information of SEM/STEM detector/lens in the pts file header.
Also, the difference of SEM and TEM is not clear, and in the original code, the Accel Voltage is used to set the mode of SEM/TEM.

hyperspy/tests/io/test_jeol.py
nem1234: Ok, I agree to remove those files.
I'm sorry I'm not friendly with complicate GIT operation.
So, please do it on your side.

hyperspy/io_plugins/jeol.py
nem1234: fixed.

hyperspy/io_plugins/jeol.py
nem1234: The JEOL EDS software seems to made for SEM, and it communicate with TEM for a limited number of parameters like magnification and accel voltage. Some parameters in the header are filled with strange values, for example, the value of beam current. The image in JEOL EDS software always shows "BF"/"SEI" even if the ADF detector is used for STEM imaging.

ericpre: OK, thanks for the clarification!

ericpre:
For information I tidy up the history (one merge conflict in `io.rst` wasn't solved correctly and there were unnecessary test files which needed to be removed) by doing the following  
- create a back up by creating a new branch (`jeol_pts_emimage_backup`)
- reset to the head to the commit before the incorrect merge conflict
- cherry pick all commits (4f14a19 to eae6730) after the incorrect merge conflict (from `jeol_pts_emimage_backup` to `jeol_pts_emimage_backup`)
- solve merge conflict in f068f7d
- force push to this branch
- add changelog entry

ericpre (approved):

ericpre:
I have reduced the memory usage of some tests, other failures are not related to this PR.

ericpre:
Thanks @nem1234, this is a very nice contribution!

PR #2847, opened 2021-11-04
Fix hanning window calculation

tcpekin:
### Description of the change: BUGFIX
While I very well may be wrong, when taking a 3D array (EELS Spectrum Image) and applying a hanning window for fourier_ratio_deconvolution, the signals have a hanning window applied. This window should be applied to the signal axis, not the navigation axes, but the window size is is calculated using len(hs_object), which is just np.prod(nav_axes_size). This change should (hopefully) calculate the correct hanning window size for the signal axis, such that it matches the documentation. 

If I am wrong, please reject. I also did not know which branch to propose to, and I'm happy to change that if necessary.

### Progress of the PR
- [x ] Change implemented (can be split into several points),
- [n/a ] update docstring (if appropriate),
- [n/a ] update user guide (if appropriate),
- [n/a] add an changelog entry in the `upcoming_changes` folder (see [`upcoming_changes/README.rst`](https://github.com/hyperspy/hyperspy/blob/RELEASE_next_minor/upcoming_changes/README.rst)),
- [n/a] Check formatting changelog entry in the `readthedocs` doc build of this PR (link in github checks)
- [n/a ] add tests,
- [ x] ready for review.

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -1,5 +1,5 @@
 ### Requirements
-* Read the [developer guide](https://hyperspy.org/hyperspy-doc/current/dev_guide/index.html).
+* Read the [developer guide](https://hyperspy.org/hyperspy-doc/current/dev_guide/intro.html).
 * Base your pull request on the [correct branch](https://hyperspy.org/hyperspy-doc/current/dev_guide/git.html#semantic-versioning-and-hyperspy-main-branches).
 * Filling out the template; it helps the review process and it is useful to summarise the PR.
 * This template can be updated during the progression of the PR to summarise its status. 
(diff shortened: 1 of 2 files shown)


tcpekin:
I also have updated a broken link in the PR Guide.

ericpre:
Thanks for the PR! It look to me that there is no issue: given a hyperspy signal `s`, `s()` returns a numpy array of to the "current signal", i.e. the signal at the current index.

For example:
```python
import hyperspy.api as hs

print(s())

assert len(s()) == s.axes_manager.signal_axes[0].size
```
From reading, it seems that your changes are strictly equivalent! Does it make more sense with the example above?

tcpekin:
Whoops, sorry @ericpre, I think I missed the () on s when testing it myself. I can check for sure tomorrow, but I am 99% sure that is the case, sorry. I will close this pull request. I think the PR_TEMPLATE file still needs the link to be updated, but at this point I can't be sure. Thanks for the kind correction.

ericpre:
> I think the PR_TEMPLATE file still needs the link to be updated, but at this point I can't be sure. Thanks for the kind correction.

Yes, you are right, please go ahead with a pull request if you like! :)

Issue #1931, opened 2018-04-26
Feature request: 'axis.chunks' for Lazy signals


woozey:
It would be convenient to have an access to chunk size through `axes_manager` for each axis. At the moment one can get chunk size through `s.data.chunks`, which need some efforts to translate into navigation or signal axes notation.

The syntax can be as simple as that:
``` python
>>> s.axes_manager.signal_axes[0].chunks
((5,5,5,5), )
```

ericpre:
An alternative syntax has been implemented in #2855:

```python
>>> s.get_chunks_size(s.axes_manager.signal_axes[0])
((5,5,5,5), )
```

PR #2855, opened 2021-11-16
Lazy html representation

CSSFrancis:
### This adds in an html representation for a lazy signal.  I think this helps to visualize and understand higher dimensional data.  It separates chunks into the signal and navigation axes. 

- I think it was pretty hard to know how the chunking was being applied to the signal and navigation axes
- A lot of times I was using display(s.data) for lazy_signals to use the dask html representation. 

### Progress of the PR
- [x] update docstring
- [x] update user guide
- [x] add an changelog entry in the `upcoming_changes` folder (see [`upcoming_changes/README.rst`]
- [x] add tests (Not really sure how to test this?)
- [x] ready for review

```python
import hyperspy.api as hs
import dask.array as da
x = da.zeros((10,203,202,101))
s = hs.signals.Signal2D(x).as_lazy()
s.set_signal_type("diffraction")
s.metadata.General.title="Test"
display(s)
```
image

For anyone using lazy signals @ericpre @pc494 @magnunor @hakonanes I'm open to any suggestions  as to what you think might be useful to add to the html representation. 

**Other Things to Consider:**
- Adding in some information about the axes, potentially adding real units to the chunk images rather than pixels.
- Displaying the "fast" axes, or the axes you can operate on without rechunking.  
- Method of loading the data (Is the data compressed, memmapped, or computed)  This is related to the "cost" to rechunk the data. 

Anything else?


magnunor:
Good idea! Especially since the doing the "wrong" chunking in the wrong dimension can lead to some really slow processing, or the whole computer crashing due to lack of memory.

I'll try to chip in some review-time!

hakonanes:
This looks very nice and informative! I like this representation much more than checking `print(s.data)` and `print(s.data.chunksize)` all the time, which I'm currently doing.

I think the representation is dense enough, and shouldn't contain much more information. Perhaps the "fast" axes can be bold?

I'm not sure how many more new dependencies changing `dask[array]` to `dask` introduces, is it possible to only include the `widgets` and the `array` modules? Should perhaps this nice representation be an optional download with hyperspy?

CSSFrancis:
> I think the representation is dense enough, and shouldn't contain much more information. Perhaps the "fast" axes can be bold?

Good idea!  I can easily make that change. 

> I'm not sure how many more new dependencies changing `dask[array]` to `dask` introduces, is it possible to only include the `widgets` and the `array` modules? Should perhaps this nice representation be an optional download with hyperspy?

The widgets module was just added to help out with this kind of representations so I'm not sure you can install just dask.widgets. At least it isn't listed under packages 

https://github.com/dask/dask/blob/f5881891505b9a2ba2da195befb11ad7b4c7bb23/setup.py#L41-L50

I think that an optional download is probably the best case scenario for now.

hakonanes:
Regarding dask packages, I just did

```bash
conda create -n dask-test python=3.9
conda activate dask-test
pip install dask[array]
python -c "from dask.widgets import TEMPLATE_PATH"
```

which imported `TEMPLATE_PATH` successfully. Looking at [the init for dask.widgets](https://github.com/dask/dask/blob/main/dask/widgets/__init__.py) it seems like you'll get that module with `dask[diagnostics]`, and it seems like you'll get the diagnostics module with `dask[array]`. Based on this I think the current `dask[array]` suffices, or did you have a different experience?

ericpre:
xref https://github.com/hyperspy/hyperspy/issues/1931.

> - [ ]  add tests (Not really sure how to test this?)

Already testing that the code run is already good, even if the output is not checked. You could save the html code and compare to this, but it would mean that it is heavily dependent on the specific dask widgets implementation and therefore it is not good. 

> Other Things to Consider:
>
> - Adding in some information about the axes, potentially adding real units to the chunk images rather than pixels.
> - Displaying the "fast" axes, or the axes you can operate on without rechunking.
> - Method of loading the data (Is the data compressed, memmapped, or computed) This is related to the "cost" to rechunk the data.

In my opinion, this is good as it is in the image of your summary. If users want more information about the dask graph and the source of the task, they should use dask directly - there are some widgets to visualise it.

hyperspy/tests/signal/test_mask
```diff
@@ -0,0 +1,32 @@
+# -*- coding: utf-8 -*-
```
ericpre: It seems that this file shouldn't be part of this PR?

hyperspy/__init__.py
```diff
@@ -22,7 +22,14 @@
...
 from hyperspy import docstrings
-
+import os
+
+try:
+    from dask.widgets import TEMPLATE_PATHS
+    templates_path = os.path.join(os.path.dirname(os.path.abspath(__file__)), "widgets/templates")
+    TEMPLATE_PATHS.append(templates_path)
+except ModuleNotFoundError:
+    _logger.info("Dask widgets not loaded")
```
ericpre:
```suggestion
    _logger.info("Dask widgets not loaded (dask >=2021.11.1 is required)")
```

setup.py
```diff
@@ -89,6 +89,7 @@
     "speed": ["cython", "imagecodecs>=2020.1.31"],
     "usid": ["pyUSID>=0.0.7", "sidpy"],
     "scalebar": ["matplotlib-scalebar"],
+    "lazy-rep": ["dask>=2021.11.1"],
```
ericpre: I would simply remove it, because we don't want to install all dask packages and even if we specify `dask['array']`, I am not sure if `pip` will consider this minimum version. I remember that it wasn't able to install the correct version when a dependencies was specified several times. I haven't checked if this is still the case.

hyperspy/_signals/lazy.py
```diff
@@ -278,6 +325,11 @@ def _get_navigation_chunk_size(self):
         nav_chunks = tuple([self.data.chunks[i] for i in sorted(nav_axes)])
         return nav_chunks
 
+    def _get_signal_chunk_size(self):
```
ericpre: Maybe it would be worth to refactor `_get_signal_chunk_size` and `_get_navigation_chunk_size` to make it more generic and a public API to address #1931 at the same time?

For example, something along the line of:
```python
# the get the chunk size of the signal axes
s.get_chunk_size(s.axes_manager.signals_axes)

# the get the chunk size of the navigation axes
s.get_chunk_size(s.axes_manager.navigation_axes)

# to get the chunk size of any axis, for example axis 2 in the axes_manager
s.get_chunk_size(s.axes_manager[2])
```

ericpre (commented):
This looks good, I made some suggestions/comments.

Some alignments are a bit off, but this is not a big deal:
- the space about above "Title"
- "Navigation Axes` is misaligned with the chunks of the navigation space

image

Also, could you please rebase this branch to tidy up the history and remove some spurious files/commits?

doc/dev_guide/lazy_computations.rst
```diff
@@ -46,6 +46,20 @@ cases for lazy signals will have to be written. If a function operates on
...
+        import hyperspy.api as hs
+        x = np.zeros((20,20,10,10,10,))
+        x = hs.signals.Signal2D(x)
+        display(x)
+
+.. figure:: images/chunks.png
+
+This helps to visualize the chunk structure and identify axes where the chunk spans the entire
+axis (bolded axes).
+
```
ericpre: Should this go in the user guide?
Also it would be good to make the code of the example visible.

upcoming_changes/2855.new.rst
```diff
@@ -0,0 +1 @@
+Add :ref:`html representation<lazy._repr_html_>` for lazy signals
```
ericpre: The link is not working: `.. _lazy._repr_html_:` need to be added at the correct place in the documentation.

doc/dev_guide/lazy_computations.rst
```diff
@@ -46,6 +46,20 @@ cases for lazy signals will have to be written. If a function operates on
 arbitrary-sized arrays and the shape of the output can be known before calling,
 ``da.map_blocks`` and ``da.map_overlap`` are efficient and flexible.
 
+For more recent versions of dask (dask>2021.11) when using hyperspy in a jupyter
+notebook a helpful html representation is available.
+    .. code-block::python
```
ericpre: This example needs to use a lazy signal and it doesn't appear in the document, maybe because of the indentation?

hyperspy/_signals/lazy.py
```diff
@@ -163,6 +163,53 @@ def __init__(self, *args, **kwargs):
...
+
+        except ModuleNotFoundError:
+            return self
+
+    def _get_chunk_string(self):
+        nav_chunks = self.data.chunksize[:len(self.axes_manager.navigation_shape)][::-1]
+        string = "("
+        for chunks,axis in zip(nav_chunks,self.axes_manager.navigation_shape):
+            if chunks == axis:
+                string += "<b>"+str(chunks)+"</b>,"
```
ericpre, line 218: Can you add a test covering this line?

hyperspy/_signals/lazy.py
```diff
@@ -163,6 +163,53 @@ def __init__(self, *args, **kwargs):
...
+                string += str(chunks) +","
+        string= string.rstrip(",")
+        string +="|"
+
+        sig_chunks = self.data.chunksize[len(self.axes_manager.navigation_shape):][::-1]
+        for chunks,axis in zip(sig_chunks,self.axes_manager.signal_shape):
+            if chunks == axis:
+                string += "<b>"+str(chunks)+"</b>,"
+            else:
+                string += str(chunks) +","
```
ericpre: Same here, can you add a test covering this line?

hyperspy/_signals/lazy.py
```diff
@@ -350,10 +397,20 @@ def _get_dask_chunks(self, axis=None, dtype=None):
...
-        nav_chunks = tuple([self.data.chunks[i] for i in sorted(nav_axes)])
-        return nav_chunks
+    def get_chunk_size(self, axis=None):
+        """ Returns the chunk size for a set of given axis.
+        Parameters
+        ----------
+        axis %s
+        """
+        if axis is None:
+            axis = self.axes_manager.navigation_axes
```
ericpre: Can you add a test covering this line?

hyperspy/_signals/lazy.py
```diff
@@ -350,10 +397,20 @@ def _get_dask_chunks(self, axis=None, dtype=None):
...
+        axes = self.axes_manager[axis]
+        if not np.iterable(axes):
+            axes = (axes,)
```
ericpre, line 448: Can you add a test covering this line?

ericpre (commented):
This looks good.

This doesn't work when the array contain only one single chunk. It doesn't raise an error because it falls back to non html display, but it is confusing and it should be made consistent

Example to reproduce and raise the error by calling `_repr_html_` directly:

```python
import dask.array as da
import hyperspy.api as hs

s = hs.signals.BaseSignal(da.random.random((500, 1000))).as_lazy()

s._repr_html_()
```
which raises:
```python
NotImplementedError                       Traceback (most recent call last)
/tmp/ipykernel_80943/4231446018.py in <module>
----> 1 s._repr_html_()

~/Dev/hyperspy/hyperspy/_signals/lazy.py in _repr_html_(self)
    170             from dask.widgets import get_template
    171             from dask.utils import format_bytes
--> 172             nav_grid = svg(chunks=self.get_chunk_size(axis=self.axes_manager.navigation_axes),
    173                            size=config.get("array.svg.size", 160))
    174             sig_grid = svg(chunks=self.get_chunk_size(axis=self.axes_manager.signal_axes),

/opt/miniforge/lib/python3.9/site-packages/dask/array/svg.py in svg(chunks, size, **kwargs)
     27         raise NotImplementedError("Can't generate SVG with 0-length dimensions")
     28     if len(chunks) == 0:
---> 29         raise NotImplementedError("Can't generate SVG with 0 dimensions")
     30     if len(chunks) == 1:
     31         return svg_1d(chunks, size=size, **kwargs)

NotImplementedError: Can't generate SVG with 0 dimensions
```

doc/dev_guide/lazy_computations.rst
CSSFrancis: Yes, probably... I'll move it to the section dealing with big data

ericpre (approved):
I rebased to get the test suite to pass because of the ipython 8.0 issue and I tidy/fix minor issues with the doc and packaging.

PR #2858, opened 2021-11-22
Modify formula hydrogen k edge

daenjannis:
### Description of the change
There was an error in the calculation for the gdos of the hydrogenic k edge.
In the code of Sigma3k (https://sites.google.com/site/temsemeels/) you can see that it 
should be parameter akh instead of kh2 for the expression: 1-np.e(-2 * np.pi/kh2)


francisco-dlp:
It looks good, thanks. Could you add an entry to https://github.com/hyperspy/hyperspy/tree/RELEASE_next_patch/upcoming_changes ?

Since this is a bug fix, it would have been better to base your branch on `Release_next_patch`. Could you close this PR and send a new one to `Release_next_patch`?

PR #2863, opened 2021-12-17
bugfix: broken axis name when signal_type==""

nem1234:
### Description of the change
When signal_type=="" and unit == "nm", the axis name become broken.
(hit to   signal_type in ("CL"))

### Progress of the PR
- [x] add tests,
- [x] ready for review.

depends on dm4 file.

### DIFF
--- a/hyperspy/io_plugins/digital_micrograph.py
+++ b/hyperspy/io_plugins/digital_micrograph.py
@@ -537,15 +537,13 @@ class ImageObject(object):
         names = [t.Undefined] * len(self.shape)
         indices = list(range(len(self.shape)))
 
-        if self.signal_type == "":
-            pass
-        elif self.signal_type == "EELS":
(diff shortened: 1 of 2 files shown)


ericpre (commented):
Thanks for tracking down this bug!
Indeed, because python evaluate `"" in ('CL')` as `True`, which actually surprised me, it wrongly goes to this branch of the `if/elif` block.
I think, that a slightly better fix would be not to use this misleading syntax and replace `elif self.signal_type in ("CL")` by `elif self.signal_type == "CL"`, instead of adding `if` block to skip when `self.signal_type == ""`. What do you think?

nem1234:
('CL') seems seems not a tuple but a string. 
"", "C" and "L" matches to ("CL")
So, signal_type == "CL" is correct.

fixed.
I add a test code for this.

ericpre (approved):
Great, thanks!

PR #2865, opened 2021-12-21
Bugfixes, docstring improvements and add several unit tests for EELSModel class

magnunor:
This pull requests has several improvements for `EELSModel`:

- Add unit tests for several features which were previously not tested
- Fix bugs in 3 functions: `EELSModel.free_edges`, `EELSModel.enable_free_onset_energy` and `EELSModel.disable_free_onset_energy`
- Several docstring fixes, both typos, style fixes, and minor rewrites

### DIFF
--- a/hyperspy/models/eelsmodel.py
+++ b/hyperspy/models/eelsmodel.py
@@ -886,9 +886,13 @@ class EELSModel(Model1D):
             edges_list = [self._get_component(x) for x in edges_list]
         for edge in edges_list:
             if edge.isbackground is False:
-                edge.intensity.free = True
-                edge.onset_energy.free = True
-                edge.fine_structure_coeff.free = True
+                edge.set_parameters_free(
+                    parameter_name_list=[
+                        "intensity",
+                        "onset_energy",
+                        "fine_structure_coeff"
+                    ]
+                )
 
     def fix_fine_structure(self, edges_list=None):
         """Fixes the fine structure of the edges given in edges_list.
--- a/hyperspy/models/eelsmodel.py
+++ b/hyperspy/models/eelsmodel.py
@@ -794,7 +794,7 @@ class EELSModel(Model1D):
             edges_list = [self._get_component(x) for x in edges_list]
         for edge in edges_list:
             if edge.isbackground is False:
-                edge.free_onset_energy = True
+                edge.onset_energy.free = True
 
     def disable_free_onset_energy(self, edges_list=None):
         """Disable the automatic freeing of the onset_energy parameter during a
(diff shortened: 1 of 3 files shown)


ericpre (commented):
Can you be more specific in describing what does this PR fix? The git messages are generic and it would be good to put together the test and the fix in the same commit - it is useful when looking back at the git history to understand the reasoning of the changes.

magnunor:
@ericpre, the commit messages should be a bit more descriptive now, especially for the bug-fixes.

hyperspy/models/eelsmodel.py
```diff
@@ -794,7 +794,7 @@ def enable_free_onset_energy(self, edges_list=None):
             edges_list = [self._get_component(x) for x in edges_list]
         for edge in edges_list:
             if edge.isbackground is False:
-                edge.free_onset_energy = True
```
ericpre, line 796: Are you sure that `free_onset_energy` is disused? It is used in [`smart_fit`.](https://github.com/hyperspy/hyperspy/blob/1ffb3fab33e607045a37f30c1463350b72617e10/hyperspy/models/eelsmodel.py#L540-L545)

It should be possible to deprecate it and raise a warning when it is used.

hyperspy/models/eelsmodel.py
```diff
@@ -887,6 +887,8 @@ def free_edges(self, edges_list=None):
         for edge in edges_list:
             if edge.isbackground is False:
                 edge.intensity.free = True
```
ericpre: Use `set_parameters_free` instead?

ericpre (commented):
Thanks, I have left some questions/suggestions.

hyperspy/models/eelsmodel.py
magnunor: Initially it seemed like it was some leftover functionality, but now I'm not sure how best to handle it.

I'm guessing it is a bit confusing having both `free_onset_energy` and `onset_energy.free`. In addition, the `onset_energy.free` follows the "parameter" paradime, which is consistent with how the model fitting is done elsewhere?

I guess the `enable_free_onset_energy` function can set both `edge.free_onset_energy = True` and `edge.onset_energy.free = True`.

Then we can add a deprecation warning for `free_onset_energy`, and change `_fit_edge` from

```python
if edge.free_onset_energy is True:
    edge.onset_energy.free = True
    self.fit(**kwargs)
    edge.onset_energy.free = False
```

to 

```python
if edge.onset_energy.free is True:
    self.fit(**kwargs)
    edge.onset_energy.free = False
```

hyperspy/models/eelsmodel.py
magnunor: This would set the `effective_angle` parameter also to free, which I don't think would be the intention.

ericpre: Not if you specify the parameter list: https://hyperspy.readthedocs.io/en/latest/api/hyperspy.component.html#hyperspy.component.Component.set_parameters_free

PR #2868, opened 2022-01-17
Refactor of elements.py and test for the Hartree Slater cross sections.

### LINKED ISSUE #2817: Abimguous labeling of edges in elements.py and degenracy scaling with the 'factor' property
The P. Rez dataset of generalised oscillator strengths contains only the "odd" subshells, i.e. L1, N3, M5, etc.
This is, I suppose, due to the fact that the approximation used in the computation does not include any spin-orbit coupling. In this situation the missing edges are just shifted and scaled copies of the ones that are present.
This scaling accounts for the different degeneracy of the different levels (see Egerton, Appendix D).

For instance we have that, neglecting the shift in energy, L2 = 0.5*L3 or M4 = 2/3*M5.
This scaling is mostly handled by indicating for each edge which GOS file should actually be used, and including such scaling factors as 'factor' in the "elements.py" database.
See for instance the entry for the M4 edge of Europium:

```python
'M4': {'relevance': 'Major', 
     'threshold': 'Sharp peak',
     'edge': 'Delayed maximum',
     'factor': 0.6666666666666,
     'onset_energy (eV)': 1061.0,
     'filename': 'Eu.M5'},
```

These factors are in principle the same for all edges with the same 'name', and the edge-filename correspondence is trivial, so why store it in each entry with a lot of repetition?

It turns out that the database does not always distinguish between the subshells. A few elements, such as Na, Cl, or Al include only a combined L2,3 edge. The convention adopted, however, is to only label it L3 and scale it with a factor of 1.5 to account for the missing L2 edge:

```python
'L3': {'relevance': 'Major',
       # overlaps
       # with L2
       'threshold': 'Sharp peak',
       'edge': 'Delayed maximum',
       'factor': 1.5,
       'onset_energy (eV)': 31.0,
       'filename': 'Na.L3'},
```

The problem is that if we want to include more options for the generalised oscillator strengths it may be worth removing from the elements database the information which is specific for one single GOS dataset.

My proposal is:
- I change all the peaks such as the latter one to names like L2,3 to remove ambiguity
- I remove all the 'filename' and 'factor' field. This information can easily be included in a very small dictionary inside hartree_slater_gos.py with entries like {'L2':['L3', 0.5], 'L2,3':['L3', 1.5]} and the names can be generated on the fly in the loader function.
- The representation of the component remains unvaried, the information is just shifted out of elements.py

I can do this quite quickly, as soon as I get a go-ahead.
On the other hand, I won't bother if this kind of change in not welcome.

gguzzina:
### Description of the change
Just a code refactoring (closes #2817) 
The file `elements.py` contained a variety of information related specifically to the Hartree-Slater generalised oscillator strengths computed by Peter Rez and distributed by Gatan. In preparation to the inclusion of new GOS datasets, it was necessary to move the specialised information where it is relevant, in the file `hartree_slater_gos.py`.
A test has been added with the help of @magnunor . 

### Progress of the PR
~- [ ] add an changelog entry in the `upcoming_changes` folder (see [`upcoming_changes/README.rst`](https://github.com/hyperspy/hyperspy/blob/RELEASE_next_minor/upcoming_changes/README.rst)),~
~- [ ] Check formatting changelog entry in the `readthedocs` doc build of this PR (link in github checks)~
- [x] add tests,
- [x] ready for review.


gguzzina:
Considering that this is a change involving purely a refactor, without any "external" effect, I'm not sure whether a changelog entry is indicated.

Please let me know whether I should add one.

magnunor:
I had a quick look at the pull request, and it looks nice. But I wrote some parts of the code to generate and test the H-S edges, so someone else should maybe look at that part.

Note that these tests will not run on the continuous integration system, as the Rez H-S GOS data is not available there: `SKIPPED [1] hyperspy/tests/model/test_eelsmodel.py:235: Rez Hartree-Slater GOS data is not available, skipping`.

It will currently only run locally, if the person running the tests has the Rez H-S GOS files properly installed.

jlaehne (approved):
looks ok to me, but i have never used the eels functionality

ericpre (commented):
I rebase and added a release note. This looks good, however, there is one important issue: I think that the test should be removed for the following reasons:
- the `hartree_slater_rez_edge_archive.npz` contains the cross sections and I don't think we are allow to include them in the source code, as we would be *distributing* them.
- we don't need this keep this test in the hyperspy code, we just need to make sure that the cross section are the same before and after the PR and it only needs to be used in the PR (for testing and reviewing purposes).

Except if someone can justify that we are allow to distribute these cross-section, we need to remove the `hartree_slater_rez_edge_archive.npz` file, maybe the easiest is to cherry pick the relevant commits and make a new PR.

gguzzina:
The npz files does not contain the generalised oscillator strengths. It contains simulated spectra, computed for alpha=5, beta=10, HT=100. They cannot be used to reconstruct the original generalised oscillator strength dataset, and they are absolutely not replacement for it.

They are technically cross-sections, but in that sense every eels spectrum is just an energy differential cross section.

One could argue that f we're not allowed to distribute this, then we're also not allowed to distribute any plot we generate using them, at least not in PDF and SVG form, as it contains exactly the same data (though in a less friendly format).

gguzzina:
Sorry for the misclick 😅

ericpre:
> The npz files does not contain the generalised oscillator strengths. It contains simulated spectra, computed for alpha=5, beta=10, HT=100. They cannot be used to reconstruct the original generalised oscillator strength dataset, and they are absolutely not replacement for it.
> 
> They are technically cross-sections, but in that sense every eels spectrum is just an energy differential cross section.
> 
> One could argue that f we're not allowed to distribute this, then we're also not allowed to distribute any plot we generate using them, at least not in PDF and SVG form, as it contains exactly the same data (though in a less friendly format).

I agree and my concern is that we don't know what we can do! In doubt, I would prefer to be cautious and stay away from it because in my opinion, the benefits are limited:
- we can't run the test on CI anyway*
- if you add support for other GOS and these can be downloaded transparently for the user, the Hartree-Slater GOS will most likely be slowly phased out by users, if they are as good or better and work out of the box! :)

This is why, I think the advantage are not worth the risk and I would keep the test script and remove the npy file. We can run it manually only when some changes are made to this part of the code - which happen very very very rarely!

*When I was checking this PR, it seems that the GOS that I used from a recent GMS didn't work... I had to use the HS GOS of an old version to get it to work! I didn't try to figure out where the issue comes from, but potentially this indicate that our test it is broken anyway and that it went unnoticed, but since we are limited in what we can do (no versioning of the GOS, needs to copy the GOS or set the path correctly, etc.), I don't think that there is much more that we can do...

gguzzina:
> I agree and my concern is that we don't know what we can do! In doubt, I would prefer to be cautious and stay away from it because in my opinion, the benefits are limited:
> 

Fair enough. I will make a new PR/Branch/something.
 
> *When I was checking this PR, it seems that the GOS that I used from a recent GMS didn't work... I had to use the HS GOS of an old version to get it to work! I didn't try to figure out where the issue comes from, but potentially this indicate that our test it is broken anyway and that it went unnoticed, but since we are limited in what we can do (no versioning of the GOS, needs to copy the GOS or set the path correctly, etc.), I don't think that there is much more that we can do...

That's confusing, as the test does not access the GOS directly, but only simulates edges by generating an appropriate model. This means one of two things:

- Gatan changed their format in order to be incompatible with third party tools, which means hyperspy's loading of the GOS is now broken and fitting with HS GOS is impossible unless someone has an old copy.

- Gatan has changed the actual data in the GOS, and the change is significant enough that the test fails when it compares new generated edges with the old ones.

Out of curiosity, can you determine which one is true?

ericpre:
> That's confusing, as the test does not access the GOS directly, but only simulates edges by generating an appropriate model. This means one of two things:
> 
> * Gatan changed their format in order to be incompatible with third party tools, which means hyperspy's loading of the GOS is now broken and fitting with HS GOS is impossible unless someone has an old copy.
> * Gatan has changed the actual data in the GOS, and the change is significant enough that the test fails when it compares new generated edges with the old ones.
> 
> Out of curiosity, can you determine which one is true?

I suspect that this is the first one, but will need check and try to figure out from which version... I will report on this in a separate issue.

magnunor:
@ericpre, I removed any traces of the H-S cross sections, by dropping the commits I made then doing a force push. Now, only @gguzzina's commit (+ release notes) remains in this pull request.

Thus, `hartree_slater_rez_edge_archive.npz` is gone, and so is the script to generate it, and the unit test itself.

I tested on my own computer that the edges are the same for the old code, and the refactored code in this pull request.

magnunor:
Note: I have not tested this on the "new" H-S GOS, as I don't have access to those.

gguzzina:
Hi Magnus,
Thanks for taking care of that.

Seeing that it's getting harder to use the Rez data, wrapping up the new GOS is getting urgent.

jlaehne:
Could it make sense to rebase this to `release_next_patch`?

gguzzina:
@jlaehne I'm not sure what that would achieve.

The commits in the PR have no conflict with either `RELEASE_next_minor` nor `RELEASE_next_patch`, and can be merged in whichever branch is deemed more appropriate.

But more importantly these changes, as it is, offer no benefit to the users. The only point is to make it a little easier and cleaner to add support for new GOS datasets, and they serve no purpose until that is ready.

jlaehne:
Could be released with the next patch release, which should be coming soon. I was not sure about the urgency to have it available, but that is why I phrased it as a question.

ericpre:
There is something wrong with the git history!

gguzzina:
@magnunor has removed the commits with the problematic files, then force-pushed the new history to the branch. This was done to ensure no copy of those files would exist anywhere within the GIT history of hyperspy's GIT.

Or do you mean something else I haven't seen?

ericpre:
> Or do you mean something else I haven't seen?

Yes, the commits after [3bebf7a](https://github.com/hyperspy/hyperspy/pull/2868/commits/3bebf7a500380b50715fb7bf4eb889253bc15b58) shouldn't be here.

magnunor:
I think the reason for extra commits which @ericpre is talking about, is due to the change of base branch from `RELEASE_next_minor` (`Rnm`) to `RELEASE_next_patch` (`Rnp`). I'm guessing the branch contains all of the changes which were done in `Rnm`, and should *not* be merged into `Rnp`.

Since these changes does not need to be a part of the next `Rnp` release, I suggest to change the base branch back to `Rnm`. This to avoid the hassle of having to clean up the git history.

ericpre:
> I think the reason for extra commits which @ericpre is talking about, is due to the change of base branch from `RELEASE_next_minor` (`Rnm`) to `RELEASE_next_patch` (`Rnp`). I'm guessing the branch contains all of the changes which were done in `Rnm`, and should _not_ be merged into `Rnp`.

Currently, [RELEASE_next_minor is behind RELEASE_next_patch](https://github.com/hyperspy/hyperspy/branches) and these commits are in the RELEASE_next_patch but not the RELEASE_next_minor, so it must be something else wrong.

gguzzina:
I'd rebased the branch from `rnm` onto `rnp` locally, but forgot to push it.
Now I have, and the PR is again down to 2 commits.

ericpre (approved):

PR #2874, opened 2022-01-19
Add file reading metadata to Signals

### LINKED ISSUE #2873: Adding io_plugin and HyperSpy version to metadata?
For the sake of future interoperability and reproducibility, I think it would be a good idea to add a couple of pieces of information to the Signal metadata under General. In particular, my proposal is that we note which `io_plugin` was used, as well as the version of HyperSpy at the time the Signal was loaded. I would propose that the metadata go from:

```
└── General
    ├── original_filename = test-1.dm3
    └── title = test
```

to 

```
└── General
    ├── File
    │   ├── io_plugin = hyperspy.io_plugins.digital_micrograph
    |   ├── hyperspy_version = 1.6.5
    |   └── load_timestamp = 2022-01-19T10:08:59.798538
    ├── original_filename = test-1.dm3
    └── title = test
```

I won't start working on this until there's some discussion about. What does everyone think?

jat255:
### Description of the change
This PR adds three new metadata parameters under `signal.metadata.General.File` to document information about the system/software used to load data from one of the proprietary/native formats into HyperSpy. This is desirable for future reproducibility, as it will indicate exactly when, which HyperSpy version, and which import module was used to read the data.

Resolves #2873 

### Progress of the PR
- [x] add tests,
- [x] ready for review.

```python
>>> import hyperspy.api as hs
>>> s = hs.load('hyperspy/tests/io/msa_files/example1.msa')
>>> s.metadata
├── Acquisition_instrument
│   └── TEM
│       ├── Detector
│       │   └── EELS
│       │       ├── collection_angle = 3.4
│       │       ├── collection_angle_units = mR
│       │       ├── dwell_time = 100.0
│       │       └── dwell_time_units = ms
│       ├── beam_current = 12.345
│       ├── beam_current_units = nA
│       ├── beam_energy = 120.0
│       ├── beam_energy_units = kV
│       ├── convergence_angle = 1.5
│       └── convergence_angle_units = mR
├── General
│   ├── File
│   │   ├── hyperspy_version = 1.7.0.dev0
│   │   ├── io_plugin = hyperspy.io_plugins.msa
│   │   └── load_timestamp = 2022-01-19T12:57:21.222973-07:00
│   ├── date = 1991-10-01
│   ├── original_filename = example1.msa
│   ├── time = 12:00:00
│   └── title = NIO EELS OK SHELL
├── Sample
│   ├── thickness = 50.0
│   └── thickness_units = nm
└── Signal
    ├── quantity = Counts (Intensity)
    └── signal_type = EELS
```

### DIFF
--- a/doc/user_guide/metadata_structure.rst
+++ b/doc/user_guide/metadata_structure.rst
@@ -78,10 +78,18 @@ in the following sections of this chapter.
     │           ├── y (mm)
     │           └── z (mm)
     ├── General
-    |   |── FileReader
-    |   |   ├── hyperspy_version
-    |   |   ├── io_plugin
-    |   │   └── load_timestamp
+    |   |── FileIO
+    |   |   ├── 0
+    |   |   |   ├── operation
+    |   |   |   ├── hyperspy_version
+    |   |   |   ├── io_plugin
+    |   │   |   └── timestamp
+    |   |   ├── 1
+    |   |   |   ├── operation
+    |   |   |   ├── hyperspy_version
+    |   |   |   ├── io_plugin
+    |   │   |   └── timestamp
+    |   |   └── ...
     │   ├── authors
     │   ├── date
     │   ├── doi
@@ -161,33 +169,42 @@ notes
 
 .. _general-file-metadata:
 
-FileReader
-----------
+FileIO
+------
 
-Contains information about the software packages and versions used when the
-Signal was first created by reading the original data format (added in HyperSpy
-v1.7). In the event the signal is saved to HyperSpy's native format (``.hspy``)
-and reloaded, the metadata within the ``FileReader`` node will be persisted as
-a record of the software configuration when the conversion was made from
-the proprietary/original format to HyperSpy's format.
+Contains information about the software packages and versions used any time the
+Signal was created by reading the original data format (added in HyperSpy
+v1.7) or saved by one of HyperSpy's IO tools. If the signal is saved
+to HyperSpy's native format (``.hspy``), the metadata within the ``FileIO``
+node will be represent a history of the software configurations used when the
+conversion was made from the proprietary/original format to HyperSpy's
+format, as well as any time the signal was subsequently loaded from and saved
+to disk. Under the ``FileIO`` node will be one or more nodes named ``0``,
+``1``, ``2``, etc., each with the following structure:
+
+operation
+   type: Str
+
+   This value will be either ``"load"`` or ``"save"`` to indicate whether
+   this node represents a load from, or save to disk operation, respectively.
 
 hyperspy_version
     type: Str
 
     The version number of the HyperSpy software used to extract a Signal from
-    this data file.
+    this data file or save this Signal to disk
 
 io_plugin
     type: Str
 
     The specific input/output plugin used to originally extract this data file
-    into a HyperSpy Signal -- will be of the form
+    into a HyperSpy Signal or save it to disk -- will be of the form
     ``hyperspy.io_plugins.<plugin_name>``.
 
-load_timestamp
+timestamp
     type: Str
 
-    The timestamp of the computer running the data loading process (in a
+    The timestamp of the computer running the data loading/saving process (in a
     timezone-aware format). The timestamp will be in ISO 8601 format, as
     produced by the ``isoformat()`` method of the ``datetime`` class.
 
--- a/doc/user_guide/metadata_structure.rst
+++ b/doc/user_guide/metadata_structure.rst
@@ -78,6 +78,10 @@ in the following sections of this chapter.
     │           ├── y (mm)
     │           └── z (mm)
     ├── General
+    |   |── File
+    |   |   ├── hyperspy_version
+    |   |   ├── io_plugin
+    |   │   └── load_timestamp
     │   ├── authors
     │   ├── date
     │   ├── doi
@@ -154,6 +158,32 @@ notes
     type: Str
 
     Notes about the data.
+
+File
+----
+
+Contains information about the software packages and versions used when the
+Signal was created by reading the original data format (added in HyperSpy v1.7)
+
+hyperspy_version
+    type: Str
+
+    The version number of the HyperSpy software used to extract a Signal from
+    this data file.
+
+io_plugin
+    type: Str
+
+    The specific input/output plugin used to originally extract this data file
+    into a HyperSpy Signal -- will be of the form
+    ``hyperspy.io_plugins.<plugin_name>``.
+
+load_timestamp
+    type: Str
+
+    The timestamp of the computer running the data loading process (in a
+    timezone-aware format). The timestamp will be in ISO 8601 format, as
+    produced by the ``isoformat()`` method of the ``datetime`` class.
 
 Acquisition_instrument
 ======================
--- a/doc/user_guide/metadata_structure.rst
+++ b/doc/user_guide/metadata_structure.rst
@@ -174,8 +174,8 @@ FileIO
 
 Contains information about the software packages and versions used any time the
 Signal was created by reading the original data format (added in HyperSpy
-v1.7) or saved by one of HyperSpy's IO tools. If the signal is saved
-to HyperSpy's native format (``.hspy``), the metadata within the ``FileIO``
+v1.7) or saved by one of HyperSpy's IO tools. If the signal is saved to one
+of the ``hspy``, ``zspy`` or ``nxs`` formats, the metadata within the ``FileIO``
 node will represent a history of the software configurations used when the 
 conversion was made from the proprietary/original format to HyperSpy's
 format, as well as any time the signal was subsequently loaded from and saved
--- a/doc/user_guide/metadata_structure.rst
+++ b/doc/user_guide/metadata_structure.rst
@@ -78,7 +78,7 @@ in the following sections of this chapter.
     │           ├── y (mm)
     │           └── z (mm)
     ├── General
-    |   |── File
+    |   |── FileReader
     |   |   ├── hyperspy_version
     |   |   ├── io_plugin
     |   │   └── load_timestamp
--- a/doc/user_guide/metadata_structure.rst
+++ b/doc/user_guide/metadata_structure.rst
@@ -159,6 +159,8 @@ notes
 
     Notes about the data.
 
+.. _general-file-metadata:
+
 File
 ----
 
--- a/doc/user_guide/metadata_structure.rst
+++ b/doc/user_guide/metadata_structure.rst
@@ -176,7 +176,7 @@ Contains information about the software packages and versions used any time the
 Signal was created by reading the original data format (added in HyperSpy
 v1.7) or saved by one of HyperSpy's IO tools. If the signal is saved
 to HyperSpy's native format (``.hspy``), the metadata within the ``FileIO``
-node will be represent a history of the software configurations used when the
+node will represent a history of the software configurations used when the 
 conversion was made from the proprietary/original format to HyperSpy's
 format, as well as any time the signal was subsequently loaded from and saved
 to disk. Under the ``FileIO`` node will be one or more nodes named ``0``,
--- a/doc/user_guide/metadata_structure.rst
+++ b/doc/user_guide/metadata_structure.rst
@@ -161,11 +161,15 @@ notes
 
 .. _general-file-metadata:
 
-File
-----
+FileReader
+----------
 
 Contains information about the software packages and versions used when the
-Signal was created by reading the original data format (added in HyperSpy v1.7)
+Signal was first created by reading the original data format (added in HyperSpy
+v1.7). In the event the signal is saved to HyperSpy's native format (``.hspy``)
+and reloaded, the metadata within the ``FileReader`` node will be persisted as
+a record of the software configuration when the conversion was made from
+the proprietary/original format to HyperSpy's format.
 
 hyperspy_version
     type: Str
--- a/hyperspy/io.py
+++ b/hyperspy/io.py
@@ -869,9 +869,6 @@ def _add_file_load_save_metadata(operation, signal, io_plugin):
         'hyperspy_version': hs_version,
         'timestamp': datetime.now().astimezone().isoformat()
     }
-    if isinstance(signal.metadata, dict):
-        return _add_file_load_save_metadata_dict(operation, signal,
-                                                 io_plugin, mdata_dict)
 
     # get the largest integer key present under General.FileIO, returning 0
     # as default if none are present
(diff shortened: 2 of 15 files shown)


jat255:
Ok, I believe this is ready for review/detailed discussion

jat255:
There are some test failures, but they appear unrelated to any of my changes and seem to do with IPython's `display` not being defined properly

ericpre:
> There are some test failures, but they appear unrelated to any of my changes and seem to do with IPython's `display` not being defined properly

Yes, ipython 8.0.* has this bug and it fixed in the coming ipython 8.1. If you rebase on the `RELEASE_next_minor`, it will fix it - the ipython 8.0.* versions are excluded from dependencies, as a workaround.

jat255:
thanks, @ericpre. Just rebased onto latest Rnm

jat255:
Ok, tests have all passed, so I think this is ready to go

ericpre:
Without looking too much details, this looks good overall but it find it arbitrary that it saved this information only the first time a file is loaded!

An alternative would be to do this, each time,the file is loaded and saved. Would it be more useful and consistent? Something along the line of:
```python
├── General
│   ├── File
│   │   ├── 0_load
│   │   │   ├── hyperspy_version = 1.7.0.dev0
│   │   │   ├── io_plugin = hyperspy.io_plugins.msa
│   │   │   └── load_timestamp = 2022-01-19T12:57:21.222973-07:00
│   │   ├── 1_save
│   │   │   ├── hyperspy_version = 1.7.0.dev0
│   │   │   ├── io_plugin = hyperspy.io_plugins.hspy
│   │   │   └── load_timestamp = 2022-01-19T12:58:21.222973-07:00
```

jat255:
> find it arbitrary that it saved this information only the first time a file is loaded

I suppose my thought was that this information was being stored primarily for a record of when HyperSpy converted the data from the raw data into it's own data format (which only happens once). If we wrote this metadata every time a signal was loaded or saved, I think the metadata tree would quickly grow quite extensive, and I'm not sure there's a real benefit there.

ericpre:
Yes, maybe the list can get long in some scenarios which can be inconvenient to browse the metadata. I would imagine that it is always possible to find a use case, where it happens! :D However, I am not convinced this is an issue, for example, GMS (or some GMS plugins) do save large metadata (tens of MB) and now the metadata are loaded lazily to avoid slowing down processing.
For the inconvenience of browsing the metadata interactively, in the notebook, this shouldn't be an issue, because it is possible to collapse part of the tree and it is collapse by default.

Another alternative is to save this information when there is an actual file conversion, which can happen in the following cases:
- load raw data -> save as `hspy`
- load `hspy` with file format version X -> save `hspy` with file version Y: useful to track when the [file version changes](https://github.com/hyperspy/hyperspy/blob/db5743b294703cf0418d72846356c7f54db9ee95/hyperspy/io_plugins/hspy.py#L78-L98)
- load `hspy` -> save as `zspy`/`nexus`

Personally, I would prefer to keep the information for each load/save cycle, because it keeps a record on how many times, the file has possibly been changed and its implementation will be more simple than having to figure out if there is a file conversion.

ericpre:
@jat255, what do you think about the alternatives I mentioned above?

jat255:
@ericpre Apologies for the delay getting this one wrapped up. Hopefully it can still make it into 1.7. I've modified things to be closer to your suggestion, and the code will now create new metadata node for every load/save operation, like so:

```
├── FileIO
│   ├── 0
│   │   ├── hyperspy_version = 1.7.0.dev0
│   │   ├── io_plugin = hyperspy.io_plugins.msa
│   │   ├── operation = load
│   |   └── timestamp = '2022-03-24T12:10:03.0000-05:00'
│   ├── 1
│   │   ├── hyperspy_version = 1.7.0.dev0
│   │   ├── io_plugin = hyperspy.io_plugins.hspy
│   │   ├── operation = save
│   |   └── timestamp = '2022-03-24T12:11:03.0000-05:00'
│   └── 2
│       ├── hyperspy_version = 1.7.0.dev0
│       ├── io_plugin = hyperspy.io_plugins.hspy
│       ├── operation = load
│       └── timestamp = '2022-03-24T12:12:03.0000-05:00'
```

I agree this is a better and more thorough approach. I decided to make the nodes just `'0'`, `'1'`, etc. rather than `0_load`, since it makes it easier to process them without _a priori_ knowing which is a load and which is a save. This made more sense to me to be stored as a list, but I don't think our current metadata structure supports lists, or at least not interacting with them through `DictionaryTreeBrowser`, so I implemented it as a dictionary with numeric string keys (since plain numbers are not allowed in `DictionaryTreeBrowser`. Hopefully that's alright, but we can tweak if needed. 

All the I/O tests are passing locally for me, and I rebased on the latest Release_next_minor, so I think this should be ready for review.

doc/user_guide/metadata_structure.rst
```diff
@@ -155,6 +167,47 @@ notes
...
+.. _general-file-metadata:
+
+FileIO
+------
+
+Contains information about the software packages and versions used any time the
+Signal was created by reading the original data format (added in HyperSpy
+v1.7) or saved by one of HyperSpy's IO tools. If the signal is saved
+to HyperSpy's native format (``.hspy``), the metadata within the ``FileIO``
+node will be represent a history of the software configurations used when the
```
jlaehne:
```suggestion
node will represent a history of the software configurations used when the 
```

hyperspy/io.py
```diff
@@ -848,3 +852,55 @@ def save(filename, signal, overwrite=None, **kwds):
...
+    mdata_dict = {
+        'operation': 'load',
+        'io_plugin': reader.__loader__.name,
+        'hyperspy_version': hs_version,
+        'timestamp': datetime.now().astimezone().isoformat()
+    }
+    if 'General' not in signal_dict['metadata']:
+        # if we don't even have basic metadata, just return as-is (shouldn't
+        # happen, but this came up in certain tests)
+        return signal_dict
```
jlaehne: this 'odd case' is not tested for (codecov warning)

jat255: Locally, it was hit during the tests of the nexus IO plugin: https://github.com/hyperspy/hyperspy/blob/RELEASE_next_minor/hyperspy/tests/io/test_nexus_hdf.py#L335

I'm not familiar with the Nexus reader, so I don't know why it tries to store metadata directly as a `dict` rather than a `DictionaryTreeBrowser` (that seems inconsistent), and caused an error at this part of my changes. Regardless, I could add a specific test for it, I suppose.

jat255: Huh, now I see the same thing locally... not sure what that is. I'll just remove it if apparently it's not necessary.

hyperspy/io.py
```diff
@@ -848,3 +852,51 @@ def save(filename, signal, overwrite=None, **kwds):
...
+        }
+    else:
+        largest_index = \
+            max([int(i.replace('Number_', '')) for i in
+                 signal.metadata.General.FileIO.keys()])
+        signal.metadata.General.FileIO.add_dictionary({
+            f'{largest_index + 1}': mdata_dict
+        })
+
+    return signal
```
ericpre: These two functions could be merged by adding an argument, which specifies if the key is `'save'` or `'load'`.

hyperspy/io.py
```diff
@@ -848,3 +852,51 @@ def save(filename, signal, overwrite=None, **kwds):
...
+        # if we don't even have basic metadata as a DictionaryTreeBrowser,
+        # just return as-is (shouldn't happen, but this came up in tests of
+        # test_nexus_hdf.test_save_metadata_as_dict())
+        return signal
+    if 'FileIO' not in signal.metadata.General:
+        signal.metadata.General.add_dictionary = {
+            '0': mdata_dict
+        }
+    else:
+        largest_index = \
```
ericpre: An alternative would be to create it, if this is not there and the `if`/`else` block is not necessary, see for example:
```python

import hyperspy.api as hs

s = hs.signals.Signal1D([0, 1, 2])

index = 0
mdict = {'a': 'A', 'b': 'B'}

s.metadata.set_item(f'General.FileIO.{index}', mdict)

del s.metadata.General
s.metadata.set_item(f'General.FileIO.{index}', mdict)


```

hyperspy/io.py
```diff
@@ -848,3 +852,51 @@ def save(filename, signal, overwrite=None, **kwds):
                 signal.tmp_parameters.set_item('folder', file.parent)
                 signal.tmp_parameters.set_item('filename', file.stem)
                 signal.tmp_parameters.set_item('extension', extension)
+
+
+def add_file_load_metadata(signal_dict, reader):
```
ericpre: This method needs to be private.

ericpre (commented):
This looks good but the code could be simplified easily!

hyperspy/io.py
jat255: So one of the challenges here is that at file load time, the metadata is represented as a dictionary, while at file save time, it's a DictionaryTreeBrowser, which is why I had the two different implementations. I guess I could coerce to one format and back, but I wasn't totally sure if that was a safe operation

hyperspy/io.py
jat255: in this implementation, we'd still need an `if/else` to control what "number" should be used ("0", "1", "2", etc.), so I don't think it would simplify much, but I can change it if you'd like

hyperspy/io.py
ericpre: Not if you move `signal_dict = add_file_load_metadata(signal_dict, reader)` one line after - after it goes through `dict2signal`! :)

hyperspy/io.py
ericpre: Once you know the index to be used, you can use `s.metadata.set_item(f'General.FileIO.{index}', mdict)` regardless if `FileIO` already exist or not.

For example:
```python
import hyperspy.api as hs

s = hs.signals.Signal1D([0, 1, 2])
mdict = {'a': 'A', 'b': 'B'}

# do it 3 times 
for _ in range(2):
    largest_index = max([int(i.replace('Number_', '')) + 1 for i in
                     s.metadata.get_item('General.FileIO', {}).keys()] + [0, ])
    print('largest_index', largest_index)
    
    s.metadata.set_item(f'General.FileIO.{largest_index}', mdict)
```

jat255: Clever ;)

That's what I get for commenting without looking at the actual code...

jat255:
@ericpre @jlaehne does anyone know the purpose of this test from `test_nexus_hdf.py` (https://github.com/hyperspy/hyperspy/blob/RELEASE_next_minor/hyperspy/tests/io/test_nexus_hdf.py#L335-L350)?

```python
    def test_save_metadata_as_dict(self, tmp_path):
        s = self.s
        s.original_metadata.set_item("testarray1", ["a", 2, "b", 4, 5])
        s.original_metadata.set_item("testarray2", (1, 2, 3, 4, 5))
        s.original_metadata.set_item("testarray3", np.array([1, 2, 3, 4, 5]))
        s.original_metadata = s.original_metadata.as_dictionary()
        s.metadata = s.metadata.as_dictionary()
        fname = tmp_path / 'test.nxs'
        s.save(fname)
        lin = load(fname, nxdata_only=True)
        np.testing.assert_array_equal(lin.original_metadata.testarray1,
                                      np.array([b"a", b'2', b'b', b'4', b'5']))
        np.testing.assert_array_equal(lin.original_metadata.testarray2,
                                      np.array([1, 2, 3, 4, 5]))
        np.testing.assert_array_equal(lin.original_metadata.testarray3,
                                      np.array([1, 2, 3, 4, 5]))
```

It saves metadata directly as a dictionary rather than a DTB, which is causing an error when I try to use `signal.metadata.get_item('General.FileIO', {})`. I can build in an exception for it (like I did initially), but it seems weird that this one test breaks with the expected API for dealing with metadata...

jat255:
Also, something about the MRC data tests suddenly make them not pass on my local machine, and I'm not sure why.

jat255:
The coverage warning is for the part I added as a workaround for `test_save_metadata_as_dict()`. I can add more tests to get that covered, but I'd prefer to not have it at all...

ericpre:
> @ericpre @jlaehne does anyone know the purpose of this test from `test_nexus_hdf.py` (https://github.com/hyperspy/hyperspy/blob/RELEASE_next_minor/hyperspy/tests/io/test_nexus_hdf.py#L335-L350)?
> 
> ```python
>     def test_save_metadata_as_dict(self, tmp_path):
>         s = self.s
>         s.original_metadata.set_item("testarray1", ["a", 2, "b", 4, 5])
>         s.original_metadata.set_item("testarray2", (1, 2, 3, 4, 5))
>         s.original_metadata.set_item("testarray3", np.array([1, 2, 3, 4, 5]))
>         s.original_metadata = s.original_metadata.as_dictionary()
>         s.metadata = s.metadata.as_dictionary()
>         fname = tmp_path / 'test.nxs'
>         s.save(fname)
>         lin = load(fname, nxdata_only=True)
>         np.testing.assert_array_equal(lin.original_metadata.testarray1,
>                                       np.array([b"a", b'2', b'b', b'4', b'5']))
>         np.testing.assert_array_equal(lin.original_metadata.testarray2,
>                                       np.array([1, 2, 3, 4, 5]))
>         np.testing.assert_array_equal(lin.original_metadata.testarray3,
>                                       np.array([1, 2, 3, 4, 5]))
> ```
> 
> It saves metadata directly as a dictionary rather than a DTB, which is causing an error when I try to use `signal.metadata.get_item('General.FileIO', {})`. I can build in an exception for it (like I did initially), but it seems weird that this one test breaks with the expected API for dealing with metadata...

This comes from https://github.com/hyperspy/hyperspy/pull/2725/commits/0f7ff34dcebf6379d904c617d347fecd5b848ec3 and https://github.com/hyperspy/hyperspy/pull/2725/commits/ad66893f2d9aafc527637536d3db58230680ca2e, and it doesn't look right to me: `metadata` and `original_metadata` *have to* be a `DTB`! I will make a PR to fix it separately and then we can come back to this PR.

jat255:
> `metadata` and `original_metadata` _have to_ be a `DTB`!

That's what I was thinking! Glad I wasn't crazy trying to figure out why this part was inconsistent. I'll remove that workaround from my code, and then once you've done that other PR, this should hopefully be ready to merge in with just a rebase.

jlaehne:
Was closed automatically due to a 'fix' statement in #2913

ericpre:
@jat255, you should be good to go to remove the workaround discussed above.

jat255:
Ok @ericpre I removed that prior workaround, fixed one small issue with one of the tests for #2913, and rebased on the latest RELEASE_next_minor. Looks like all checks are passing, so should be good to go

PR #2842, opened 2021-10-24
Fix ragged array

ericpre:
The changes in #2773 break the pyxem test suite. In my opinion, using the use of ragged array is not clear enough and there is an inconsistency in slicing array. This has been been discussed and introduced in #1336, in particular starting from https://github.com/hyperspy/hyperspy/pull/1336#issuecomment-263876213.

Before this PR
```python
import hyperspy.api as hs
import numpy as np

data = np.empty((3, 4), dtype=object)
data.fill(np.array([10, 20]))
s = hs.signals.BaseSignal(data).T

print(s)
# <BaseSignal, title: , dimensions: (4, 3|)>

s2 = s.inav[0]
print(s2)
# <BaseSignal, title: , dimensions: (3|)>
assert s2.axes_manager.signal_dimension == 0
assert s2.axes_manager.signal_shape == ()
assert s2.axes_manager.navigation_shape == (3,)

s3 = s.inav[0, 0]
print(s3)
# <BaseSignal, title: , dimensions: (|1)>
assert s3.axes_manager.signal_dimension == 0
assert s3.axes_manager.signal_shape == (1,)
assert s3.axes_manager.navigation_shape == ()

np.testing.assert_allclose(s()[0], s2()[0])
np.testing.assert_allclose(s()[0], s3()[0]) # fail
```
With this PR
```python
data = np.empty((3, 4), dtype=object)
data.fill(np.array([10, 20]))
s = hs.signals.BaseSignal(data, ragged=True)

print(s)
# <BaseSignal, title: , dimensions: (4, 3|ragged)>

print(s.axes_manager)
# <Axes manager, axes: (4, 3|ragged)>
#             Name |   size |  index |  offset |   scale |  units 
# ================ | ====== | ====== | ======= | ======= | ====== 
#      <undefined> |      4 |      0 |       0 |       1 | <undefined> 
#      <undefined> |      3 |      0 |       0 |       1 | <undefined> 
# ---------------- | ------ | ------ | ------- | ------- | ------ 
#      Ragged axis |               Variable length

s2 = s.inav[0]
print(s2)
# <BaseSignal, title: , dimensions: (3|ragged)>
assert s2.axes_manager.signal_dimension == 0
assert s2.axes_manager.signal_shape == ()
assert s2.axes_manager.navigation_shape == (3,)

s3 = s.inav[0, 0]
print(s3)
# <BaseSignal, title: , dimensions: (|ragged)>
assert s3.axes_manager.signal_dimension == 0
[...]

PR #2877, opened 2022-02-04
Fixed ragged map with zero dim and Larger Iterating signal

CSSFrancis:
### Description of the change
This is kind of a small change to fix some bugs that I have come across with the map function

- Ragged signals always have a navigation dimension.  I think this makes sense as a numpy object array has to have some size rather than be some scalar.  
- Mapping now works when the iterating signal dimension is larger than the base signal

### Progress of the PR
- [ x] Change implemented (can be split into several points),
- [ x] add tests,
- [x ] ready for review.

### DIFF
--- a/hyperspy/signal.py
+++ b/hyperspy/signal.py
@@ -2641,11 +2641,6 @@ class BaseSignal(FancySlicing,
         axes = []
         for s in self.data.shape:
             axes.append({'size': int(s), })
-        # With ragged signal with navigation dimension 0 and signal dimension 0
-        # we return an empty list to avoid getting a navigation axis of size 1,
-        # which is incorrect, because it corresponds to the ragged dimension
-        if ragged and len(axes) == 1 and axes[0]['size'] == 1:
-            axes = []
         return axes
 
     def __call__(self, axes_manager=None, fft_shift=False):
@@ -4885,6 +4880,9 @@ class BaseSignal(FancySlicing,
             s_input = self.as_lazy()
         else:
             s_input = self
+        if ragged:
+            output_dtype = object
+            output_signal_size = ()
 
         # unpacking keyword arguments
         if iterating_kwargs is None:
@@ -4905,9 +4903,9 @@ class BaseSignal(FancySlicing,
         else:
             old_sig = s_input
         os_am = old_sig.axes_manager
-        autodetermine = (output_signal_size is None or output_dtype is None) # try to guess output dtype and sig size?
+        autodetermine = (output_signal_size is None or output_dtype is None)  # try to guess output dtype and sig size?
 
-        args, arg_keys = old_sig._get_iterating_kwargs(iterating_kwargs)
+        args, arg_keys, data = old_sig._get_iterating_kwargs(iterating_kwargs)
 
         if autodetermine:  # trying to guess the output d-type and size from one signal
             testing_kwargs = {}
@@ -4915,7 +4913,7 @@ class BaseSignal(FancySlicing,
                 test_ind = (0,) * len(os_am.navigation_axes)
                 testing_kwargs[key] = np.squeeze(args[ikey][test_ind]).compute()
             testing_kwargs = {**kwargs, **testing_kwargs}
-            test_data = np.array(old_sig.inav[(0,) * len(os_am.navigation_shape)].data.compute())
+            test_data = np.array(data[(0,) * len(os_am.navigation_shape)].compute())
             temp_output_signal_size, temp_output_dtype = guess_output_signal_size(
                 test_data=test_data,
                 function=function,
@@ -4927,11 +4925,11 @@ class BaseSignal(FancySlicing,
             if output_dtype is None:
                 output_dtype = temp_output_dtype
 
-        drop_axis, new_axis, axes_changed = self._get_drop_axis_new_axis(output_signal_size)
+        drop_axis, new_axis, axes_changed = self._get_drop_axis_new_axis(output_signal_size, data)
         chunks = tuple([old_sig.data.chunks[i] for i in sorted(nav_indexes)]) + output_signal_size
         mapped = da.map_blocks(
             process_function_blockwise,
-            old_sig.data,
+            data,
             *args,
             function=function,
             nav_indexes=nav_indexes,
@@ -4989,22 +4987,31 @@ class BaseSignal(FancySlicing,
                 sig.data = sig.data.compute(num_workers=max_workers)
         return sig
 
-    def _get_drop_axis_new_axis(self, output_signal_size):
-        am = self.axes_manager
-        if output_signal_size == self.axes_manager.signal_shape:
+    def _get_drop_axis_new_axis(self, output_signal_size, data=None):
+        if data is not None:
+            signal_shape = np.shape(data)
+            signal_shape = tuple(np.delete(signal_shape, self.axes_manager.navigation_indices_in_array))
+            nav_dim = self.axes_manager.navigation_dimension
+            sig_ind = tuple(range(nav_dim, len(signal_shape) + nav_dim))
+        else:
+            signal_shape = self.axes_manager.signal_shape
+            nav_dim = self.axes_manager.navigation_dimension
+            sig_ind = self.axes_manager.signal_indices_in_array
+
+
+        if output_signal_size == signal_shape:
             drop_axis = None
             new_axis = None
             axes_changed = False
         else:
             axes_changed = True
-            if len(output_signal_size) != len(am.signal_shape):
-                drop_axis = am.signal_indices_in_array
-                nav_dim = am.navigation_dimension
+            if len(output_signal_size) != len(signal_shape):
+                drop_axis = sig_ind
                 new_axis = tuple(range(nav_dim, len(output_signal_size) + nav_dim))
             else:
                 drop_axis = [it for (o, i, it) in zip(output_signal_size,
-                                                      am.signal_shape,
-                                                      am.signal_indices_in_array)
+                                                      signal_shape,
+                                                      sig_ind)
                              if o != i]
                 drop_axis = tuple(drop_axis)
                 new_axis = drop_axis
@@ -5012,6 +5019,9 @@ class BaseSignal(FancySlicing,
 
     def _get_iterating_kwargs(self, iterating_kwargs):
         signal_dim_shape = self.axes_manager.signal_shape
+        iterating_shapes = [iterating_kwargs[key].axes_manager.signal_shape for
+                            key in iterating_kwargs]
+        max_iter_len = max((len(s) for s in [signal_dim_shape,]+iterating_shapes))
         nav_chunks = self.get_chunk_size(self.axes_manager.navigation_axes)
         args, arg_keys = (), ()
         for key in iterating_kwargs:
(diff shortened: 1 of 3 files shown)


ericpre:
> * Ragged signals always have a navigation dimension.  I think this makes sense as a numpy object array has to have some size rather than be some scalar.

This is an important and not a trivial change, can you elaborate on this? 

> * Mapping now works when the iterating signal dimension is larger than the base signal

Can you give an example to illustrate what is the issue? This will help with considering other alternatives.

There are recent discussions on this topic:
- https://github.com/hyperspy/hyperspy/pull/2773
- https://github.com/hyperspy/hyperspy/pull/2842

and one important aspect is that we need to make sure that ragged and non-ragged signals behave consistently with map, slicing, etc. with respect to signal and navigation dimensions.

CSSFrancis:
Sorry I should have given some examples, I submitted this a little quickly last night.

> > * Ragged signals always have a navigation dimension.  I think this makes sense as a numpy object array has to have some size rather than be some scalar.
> 
> This is an important and not a trivial change, can you elaborate on this?

The issue is that ragged signals built using the constructor can possible have a signal_shape = () and navigation_shape = ().  This results in an error when you use the map function on a ragged array with one dimension. 

For example this block of code would fail(In this case the signal have a nav dimension of ( ) )

```python
import hyperspy.api as hs
import numpy as np
x = np.empty((1,), object)
x[0] = np.ones((2,3))
s = hs.signals.BaseSignal(x, ragged=True)
s.map(np.sum)
```

But this code block will not (In this case the signal have a nav dimension of 1)

```
import hyperspy.api as hs
import numpy as np
x = np.empty((1,), object)
x[0] = np.ones((2,3))
s = hs.signals.BaseSignal(x)
s.ragged=True
s.map(np.sum)
```
The reason this fails is that the rechunking function try to rechunk a fundamentally 1 dimensional array with hyperspy trying to tell that it is a zero dimensional array.

I should also note that there are some inconsistencies with numpy which had some small effects when trying to guess the output. 

```python
import hyperspy.api as hs
import numpy as np
x = np.empty((1,), object)
x[0] = np.ones((2,3))
s = hs.signals.BaseSignal(x)
s.ragged=True
s.inav[0]
# returns a ragged array
s.data[0]
# returns the object at index 0
```


> > * Mapping now works when the iterating signal dimension is larger than the base signal
> 
> Can you give an example to illustrate what is the issue? This will help with considering other alternatives.

Yea so the issue is this chunk of code doesn't run:

```python
def add_sum(image, add):
    out = np.sum(image) + np.sum(add)
    return out
import hyperspy.api as hs
import numpy as np
x = np.ones((10,20,2,3))
s = hs.signals.Signal2D(x)
s_add = hs.signals.BaseSignal(2 * np.ones((10, 20, 2, 2, 2))).transpose(3)


s_out = s.map(add_sum, inplace=False,  add=s_add)# This doesn't work
s_out = s_add.map(add_sum, inplace=False,  add=s)# This does work
```
The issue here is that the `map_blocks` function needs to have the blocks all the same length otherwise it takes the longest arg passed (not the signal passed).  There is a way around this using the `blockwise` function but that requires a little bit more setup and would require tearing apart much of the map function.  I tried that for a little bit but decided against it because it broke some other parts of the map function.

> There are recent discussions on this topic:
> 
> * [_assign_subclass: signal_shape (1,) now becomes BaseSignal #2773](https://github.com/hyperspy/hyperspy/pull/2773)
> * [Fix ragged array #2842](https://github.com/hyperspy/hyperspy/pull/2842)
> 
> and one important aspect is that we need to make sure that ragged and non-ragged signals behave consistently with map, slicing, etc. with respect to signal and navigation dimensions.

Ahh, I must have missed the recent stuff. I think that this is more consistent than before. At the very least the errors thrown in the above cases are very uninformative and potentially confusing.  There is a high likelyhood that few people have come across them, but I recently was trying to map a couple of ragged signals and came across both go these edge cases.

ericpre:
Thanks for the details, now I see where does it come from. I think, we need to define what is the expected and desired behaviour.

> For example this block of code would fail(In this case the signal have a nav dimension of ( ) )
> 
> ```python
> import hyperspy.api as hs
> import numpy as np
> x = np.empty((1,), object)
> x[0] = np.ones((2,3))
> s = hs.signals.BaseSignal(x, ragged=True)
> s.map(np.sum)
> ```
> 
> But this code block will not (In this case the signal have a nav dimension of 1)
> 
> ```
> import hyperspy.api as hs
> import numpy as np
> x = np.empty((1,), object)
> x[0] = np.ones((2,3))
> s = hs.signals.BaseSignal(x)
> s.ragged=True
> s.map(np.sum)
> ```

This looks to me like a bug: these two (slightly different) ways of creating a bug should give the same type of signals. What I would expect here is to have the same navigation shape regardless if the signal is ragged or not. In this case, the navigation dimension is zero so I would expect the navigation shape to be `()`.

> I should also note that there are some inconsistencies with numpy which had some small effects when trying to guess the output.
> 
> ```python
> import hyperspy.api as hs
> import numpy as np
> x = np.empty((1,), object)
> x[0] = np.ones((2,3))
> s = hs.signals.BaseSignal(x)
> s.ragged=True
> s.inav[0]
> # returns a ragged array
> s.data[0]
> # returns the object at index 0
> ```

Same bug as mentioned directly (in this comment) above, `s.inav[0]` shouldn't be possible in this case.

> > > * Mapping now works when the iterating signal dimension is larger than the base signal
> > 
> > 
> > Can you give an example to illustrate what is the issue? This will help with considering other alternatives.
> 
> Yea so the issue is this chunk of code doesn't run:
> 
> ```python
> def add_sum(image, add):
>     out = np.sum(image) + np.sum(add)
>     return out
> import hyperspy.api as hs
> import numpy as np
> x = np.ones((10,20,2,3))
> s = hs.signals.Signal2D(x)
> s_add = hs.signals.BaseSignal(2 * np.ones((10, 20, 2, 2, 2))).transpose(3)
> 
> 
> s_out = s.map(add_sum, inplace=False,  add=s_add)# This doesn't work
> s_out = s_add.map(add_sum, inplace=False,  add=s)# This does work
> ```
> 
> The issue here is that the `map_blocks` function needs to have the blocks all the same length otherwise it takes the longest arg passed (not the signal passed). There is a way around this using the `blockwise` function but that requires a little bit more setup and would require tearing apart much of the map function. I tried that for a little bit but decided against it because it broke some other parts of the map function.

It looks to me that `map` should be adjusted to special case ragged array. I guess @magnunor would have a similar issue to adjust pyxem code for the new implementation of `map` in https://github.com/pyxem/pyxem/pull/789.

CSSFrancis:
> This looks to me like a bug: these two (slightly different) ways of creating a bug should give the same type of signals. What I would expect here is to have the same navigation shape regardless if the signal is ragged or not. In this case, the navigation dimension is zero so I would expect the navigation shape to be `()`.

I guess that makes some sense, as long as the map function works on zero dimensional signals I think that it is not terribly important. I am still thinking that the array `np.empty((1,),dtype=object` should have a navigation dimension of 1 dimension.  It is pretty inconsistent with numpy if that is not that case

```python
print(np.empty((), dtype=object).shape) # ()
print(np.empty((1), dtype=object).shape) # (1,)
print(np.empty((2), dtype=object).shape) # (2,)
```
I doesn't seem like the second and third example should have different numbers of dimensions in hyperspy when in numpy they have the same number of dimensions. 

> > I should also note that there are some inconsistencies with numpy which had some small effects when trying to guess the output.
> > ```python
> > import hyperspy.api as hs
> > import numpy as np
> > x = np.empty((1,), object)
> > x[0] = np.ones((2,3))
> > s = hs.signals.BaseSignal(x)
> > s.ragged=True
> > s.inav[0]
> > # returns a ragged array
> > s.data[0]
> > # returns the object at index 0
> > ```
> 
> Same bug as mentioned directly (in this comment) above, `s.inav[0]` shouldn't be possible in this case.

I think it should be? Based on my comment above.   I would rather it actually return a non ragged signal kind of like how numpy handles it.  I think that would fix some of the weird edge cases you are talking about.

> It looks to me that `map` should be adjusted to special case ragged array. I guess @magnunor would have a similar issue to adjust pyxem code for the new implementation of `map` in [pyxem/pyxem#789](https://github.com/pyxem/pyxem/pull/789).

I think the issue is bigger than ragged signals as from my example you can see that it happens with non ragged signals.  Ragged signals are just the signals where it happens the most because they have no signal dimensions. 

Anyways, let me know what you think. I haven't really seen much of an error with ragged signals always having a navigation axis = 1 but obviously you have some more experience with this.

CSSFrancis:
I looked over this a little bit more.  It seems like there is still a bug with ragged signals mapped with a larger signal.

This test fails for example because there are extra dimensions added. 

```python
 def test_iter_kwarg_larger_shape_ragged(self):
        def return_img(image, add):
            return image
        x = np.empty((1,), dtype=object)
        x[0] = np.ones((4, 2))
        s = hs.signals.BaseSignal(x, ragged=True)

        s_add = hs.signals.BaseSignal(2 * np.ones((1, 201, 101))).transpose(2)
        s_out = s.map(return_img, inplace=False, add=s_add, ragged=True)
        np.testing.assert_array_equal(s_out.data[0], x[0])
```

I'll look this over a little bit more and see if I can find a way to fix this.

ericpre:
> I guess that makes some sense, as long as the map function works on zero dimensional signals I think that it is not terribly important. I am still thinking that the array `np.empty((1,),dtype=object` should have a navigation dimension of 1 dimension. It is pretty inconsistent with numpy if that is not that case
> 
> ```python
> print(np.empty((), dtype=object).shape) # ()
> print(np.empty((1), dtype=object).shape) # (1,)
> print(np.empty((2), dtype=object).shape) # (2,)
> ```
> 
> I doesn't seem like the second and third example should have different numbers of dimensions in hyperspy when in numpy they have the same number of dimensions.

Yes, this is exactly one think that I have to make more consistent as part of #2842. The rational is that the ragged dimension is in the signal dimension and this is communicated to the users through the `axes_manager`. The example you give above is actually not a ragged array: it simply has object dtype. Maybe a better example is:

```python
data = np.empty((2, ), dtype=object)
data[0] = np.arange(3) # data index 0 has shape (3, )
data[1] = np.arange(4) # data index 1 has shape (4, )
print(data.shape) # data shape is (2,) and the ragged dimension is not visible at all, while it is definitely there!
```
Considering this example, hyperspy is behaving consistently with numpy, as the ragged dimension is not visible. Hyperspy does explicitly mention it in the `axes_manager`.
Off the top of my head, a difference between hyperspy and numpy is that numpy will convert ragged array into normal array and hyperspy try to keep it as ragged array.

Sometimes it is not possible to find a solution, which provide a perfect consistency between numpy and hyperspy and all hyperspy features and we have to define a compromise somewhere. One of the concern that I have tried to address in #2842 is that slicing of ragged signal should behave consistently, as illustrated by the example in https://github.com/hyperspy/hyperspy/pull/2842#issue-1034496626. Then other things come up during the review.

I would suggest to separate the discussions, because otherwise it will difficult to follow for anyone (and also for our future self! ;)):
- is the behaviour of ragged signal satisfactory?
- is the map function working well on ragged signal?

On these two points, it seems that you reported two different bugs here (inconsistency when creating ragged signal for the first point).
Splitting the discussion should help to focus on the correct questions/issues while still keeping the wider context and perspective.

CSSFrancis:
> I would suggest to separate the discussions, because otherwise it will difficult to follow for anyone (and also for our future self! ;)):
> 
> * is the behaviour of ragged signal satisfactory?
> * is the map function working well on ragged signal?
> 
> On these two points, it seems that you reported two different bugs here (inconsistency when creating ragged signal for the first point). Splitting the discussion should help to focus on the correct questions/issues while still keeping the wider context and perspective.

That sounds like a good idea.  I kind of imagined this as a small fix, but obviously there is more to it that first meets the eye.

I will probably close this PR, move the relevant discussion and refactor the PR into two separate cases.

ericpre:
@CSSFrancis, I think this PR can be closed now, after #2878 and #2903 have been merged?

CSSFrancis:
Closing after #2903 #2878

PR #2883, opened 2022-02-16
Fix errors due to deprecation of scipy.ndimage.filters namespace

magnunor:
Currently there are a lot of errors being caused by the `scipy.ndimage.filters` namespace being deprecated: https://github.com/hyperspy/hyperspy/runs/5103855252?check_suite_focus=true#step:8:3954

This pull request fixes this.

-------------------------------
I checked this with the minimum version of `scipy` in `setup.py`: `1.1`, and it worked nicely! So no need to bump the scipy version requirement.

### DIFF
--- a/hyperspy/_signals/signal1d.py
+++ b/hyperspy/_signals/signal1d.py
@@ -26,7 +26,7 @@ import dask.array as da
 import scipy.interpolate
 import scipy as sp
 from scipy.signal import savgol_filter
-from scipy.ndimage.filters import gaussian_filter1d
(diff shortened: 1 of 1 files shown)


ericpre:
For reference: https://scipy.github.io/devdocs/release.1.8.0.html#clear-split-between-public-and-private-api

magnunor:
Looking at the automatic unit tests, there are a number of similar deprecated namespaces in `scipy`. I'll make a pull request fixing them as well.

PR #2887, opened 2022-02-16
Exclude ipython 8.0.* instead of 8.0

### LINKED ISSUE #2886: sklearn tests not running in automatic tests
There seems to be something wrong with scikit-learn in the automatic unit tests, with lots of tests being skipped. For example: https://github.com/hyperspy/hyperspy/runs/5217344984?check_suite_focus=true#step:8:1961

Skips around 70 tests due to it not finding `sklearn`.

However, it should be installed, based on the install command being `pip install --upgrade -e .'[all, tests]'`, and `setup.py` containing:

```python
extras_require = {
    # exclude scikit-learn==1.0 on macOS (wheels issue)
    # See https://github.com/scikit-learn/scikit-learn/pull/21227
    "learning": ['scikit-learn!=1.0;sys_platform=="darwin"'],
    ...
}
```

ericpre:
- Follow up of https://github.com/hyperspy/hyperspy/pull/2870, which is not enough since there is a ipython 8.0.1.
- Fix #2886.

### DIFF
--- a/conda_environment.yml
+++ b/conda_environment.yml
@@ -7,7 +7,7 @@ dependencies:
 - h5py
 - imageio
 - ipyparallel
-- ipython !=8.0
+- ipython !=8.0.*
 - matplotlib-base
 - mrcz
 - natsort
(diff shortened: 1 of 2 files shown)


magnunor:
Looks good to me!

PR #2590, opened 2020-11-23
Add handling for CL metadata in DM reader

jlaehne:
### Description of the change
Added import of various CL specific metadata fields to DM reader

### Progress of the PR
- [X] Change implemented (can be split into several points),
- [x] update docstring (if appropriate),
- [x] update user guide (if appropriate),
- [x] add tests,
- [x] ready for review.

PR #2894, opened 2022-02-24
Update CL metadata in dm-reader

jlaehne:
After fixing the metadata structure in LumiSpy, the mapping in the DM reader has to be updated. Implements the structure proposed in https://github.com/LumiSpy/lumispy/pull/109 (and should be merged only after tis PR).

Furthermore, an SEM as microscope is now properly detected and the metadata set accordingly.

Follow up on #2590, where this mapping was first introduced. AS #2590 is still unreleased, this redesign should be included in the v1.7 release.

### Progress of the PR
- [x] handle SEM as possible microscope instead of TEM,
- [x] update metadata according to https://github.com/LumiSpy/lumispy/pull/109,
- [x] add a changelog entry in the `upcoming_changes` folder (see [`upcoming_changes/README.rst`](https://github.com/hyperspy/hyperspy/blob/RELEASE_next_minor/upcoming_changes/README.rst)),
- [x] update tests,
- [x] ready for review.


hyperspy/io_plugins/digital_micrograph.py
```diff
@@ -1001,59 +1025,82 @@ def get_mapping(self):
                     "Acquisition_instrument.TEM.Detector.EDS.real_time",
                     None),
             })
-        elif self.signal_type == "CL":
+        elif self.signal_type == "CL" or ("Acquisition" in image_tags_dict.keys() and "Monarc_Spectrometer" in image_tags_dict["Acquisition"].keys()):
```
ericpre: Include this in https://github.com/hyperspy/hyperspy/blob/d42c03d5986c649d2121fa34cfe361e9506e5a86/hyperspy/io_plugins/digital_micrograph.py#L627-L638?

hyperspy/io_plugins/digital_micrograph.py
```diff
@@ -858,6 +858,25 @@ def _get_EELS_exposure_time(self, tags):
...
 
+    def _get_CL_detector_type(self, tags):
+        if "Acquisition_Mode" in tags and tags["Acquisition_Mode"] == "Parallel dispersive":
+            return "CCD"
+        elif "Acquisition_Mode" in tags and tags["Acquisition_Mode"] == "Serial dispersive":
+            return "PMT"
+        else:
+            _logger.info("CL detector type can't be read.")
+
+    def _get_CL_exposure_time(self, tags):
```
ericpre: Make it more generic and merge it with `_get_EELS_exposure_time`? https://github.com/hyperspy/hyperspy/blob/d42c03d5986c649d2121fa34cfe361e9506e5a86/hyperspy/io_plugins/digital_micrograph.py#L847-L859

ericpre (commented):
Looks good, I have left some suggestions for improvement.

jlaehne:
Thanks @ericpre. Two very good points that I took up.

Looking if there are easy ways to increase coverage, I realized that the functions `find_next_tag` and `find_next_data_tag` are not called anywhere and seem to be remnants from some early implementation. Any reason against removing them?

Also, having a file with complex data would do a big job for the coverage.

ericpre (approved):

ericpre:
> Looking if there are easy ways to increase coverage, I realized that the functions `find_next_tag` and `find_next_data_tag` are not called anywhere and seem to be remnants from some early implementation. Any reason against removing them?

I guess, this have been used for debugging purposes but this is not documented, so I agree that this is fine to tidy it up.

PR #2897, opened 2022-03-03
Update docstrings concerning area parameters

jlaehne:
### Description of the change
Clarified in the docstrings of `Gaussian`, `Lorentzian` and `GaussianHF` that `A` parameter is the area, in `Gaussian2D` the equivalent would be a volume.

### Progress of the PR
- [x] ready for review.

### DIFF
--- a/hyperspy/_components/gaussian.py
+++ b/hyperspy/_components/gaussian.py
@@ -70,7 +70,7 @@ class Gaussian(Expression):
     ============== ===========
     Variable        Parameter
     ============== ===========
-    :math:`A`       area
+    :math:`A`       A
     :math:`\sigma`  sigma
     :math:`x_0`     centre
     ============== ===========
--- a/hyperspy/_components/gaussian.py
+++ b/hyperspy/_components/gaussian.py
@@ -70,7 +70,7 @@ class Gaussian(Expression):
     ============== ===========
     Variable        Parameter
     ============== ===========
-    :math:`A`       A
+    :math:`A`       area
     :math:`\sigma`  sigma
     :math:`x_0`     centre
     ============== ===========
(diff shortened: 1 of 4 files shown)


hyperspy/_components/gaussian.py
```diff
@@ -70,7 +70,7 @@ class Gaussian(Expression):
     ============== ===========
     Variable        Parameter
     ============== ===========
-    :math:`A`       A
```
ericpre: As I understand it, this `A` is refering to the parameter name, as in the name of the hyperspy parameter object. I think, in this case, it makes sense to keep `A`, so that user can easily find the parameter names from the docstring.
The meaning of `A` is then explain below in the parameter docstring.

hyperspy/_components/gaussian2d.py
```diff
@@ -45,7 +45,7 @@ class Gaussian2D(Expression):
     Parameters
     ----------
     A : float
-        Amplitude (height of the peak scaled by :math:`2 \pi s_x s_y`).
+        Area (height of the peak scaled by :math:`2 \pi s_x s_y`).
```
ericpre: I am wondering if `area` is a good wording for a 2D gaussian?

hyperspy/_components/gaussianhf.py
```diff
@@ -65,8 +65,8 @@ class GaussianHF(Expression):
         Extra keyword arguments are passed to the ``Expression`` component.
 
 
-    The helper properties `sigma` and `A` are also defined for compatibility
-    with `Gaussian` component.
+    The helper properties `sigma` and `A` (Area) are also defined for
```
ericpre:
```suggestion
    The helper properties `sigma` and `A` (area) are also defined for
```

hyperspy/_components/gaussian2d.py
jlaehne: True, `volume` would be correct. Changed and added a statement that it is equivalent to the area in a 1D Gaussian.

ericpre (approved):

PR #2900, opened 2022-03-09
Make Bruker .bcf reader less restrictive if no element selection is found

### LINKED ISSUE #2881: Key error loading Bruker .bcf file
Screen Shot 2022-02-15 at 9 31 34 AM
Screen Shot 2022-02-15 at 9 17 07 AM
I'm trying to load a EDS map .bcf file using the most basic method (define path name and using hs.load(path), and get the error "Key error: 'Line'." Setting load_original_metadata to false does not resolve this. This file was recorded the same day as two others, one of which gives the same error but the third loads just fine. The files also open just fine in Bruker's ESPIRIT software.

hakonanes:
### Description of the change
* Fix #2881 (reported by @TMoon96) by making the Bruker .bcf reader less restrictive when no element selection is found.
* Other minor improvements of formatting of the documentation I read in preparation of this PR.

### Progress of the PR
- [n/a] add tests,
- [x] ready for review.


hakonanes:
The test failing in the ["Job MacOS_Python37" run on Azure](https://dev.azure.com/franciscode-la-pena-manchon/hyperspy/_build/results?buildId=2287&view=logs&j=8c9bcb2e-c86e-541d-bc4c-dffeaf2450e0&t=6146f9d3-ff17-5ef2-d748-f71dccf96a82&l=2392) is `TestSavingMetadataContainers.test_save_long_list()`:

https://github.com/hyperspy/hyperspy/blob/f3aaab500891b08b3520e39c0b7563d76435d4b9/hyperspy/tests/io/test_hdf5.py#L207-L215

Azure doesn't print the error message, just a one-liner stating that the test failed. Could it be that 1 s is too short a time for the test on MacOS with Python 3.7?

jlaehne (approved):

ericpre (commented):
Is it possible to have an answer to https://github.com/hyperspy/hyperspy/issues/2881#issuecomment-1041289726? And if @TMoon96 can reproduce the issue, it would be good to make a small test file.

ericpre:
> The test failing in the ["Job MacOS_Python37" run on Azure](https://dev.azure.com/franciscode-la-pena-manchon/hyperspy/_build/results?buildId=2287&view=logs&j=8c9bcb2e-c86e-541d-bc4c-dffeaf2450e0&t=6146f9d3-ff17-5ef2-d748-f71dccf96a82&l=2392) is `TestSavingMetadataContainers.test_save_long_list()`:
> 
> https://github.com/hyperspy/hyperspy/blob/f3aaab500891b08b3520e39c0b7563d76435d4b9/hyperspy/tests/io/test_hdf5.py#L207-L215
> 
> Azure doesn't print the error message, just a one-liner stating that the test failed. Could it be that 1 s is too short a time for the test on MacOS with Python 3.7?

Here is the error:
```_________ TestSavingMetadataContainers.test_save_long_list[test.zspy] __________
[gw2] darwin -- Python 3.7.12 /Users/runner/work/1/mambaforge/envs/test_env/bin/python

self = <hyperspy.tests.io.test_hdf5.TestSavingMetadataContainers object at 0x7f8547543050>
tmp_path = PosixPath('/private/var/folders/24/8k48jl6d249_n_qfxwsl6xvm0000gn/T/pytest-of-runner/pytest-0/popen-gw2/test_save_long_list_test_zspy_3')
file = 'test.zspy'

    @zspy_marker
    def test_save_long_list(self, tmp_path, file):
        s = self.s
        s.metadata.set_item('long_list', list(range(10000)))
        start = time.time()
        fname = tmp_path / file
        s.save(fname)
        end = time.time()
>       assert end - start < 1.0  # It should finish in less that 1 s.
E       assert (1646831101.232639 - 1646831098.0749178) < 1.0

hyperspy/tests/io/test_hdf5.py:215: AssertionError
```

hakonanes:
> Is it possible to have an answer to https://github.com/hyperspy/hyperspy/issues/2881#issuecomment-1041289726? And if @TMoon96 can reproduce the issue, it would be good to make a small test file.

The only solution I see is for @TMoon96 to provide one. Can wait a bit for that.

Thanks for providing the output.

ericpre:
@TMoon96, any chance that you can help with https://github.com/hyperspy/hyperspy/issues/2881#issuecomment-1041289726?

sem-geologist:
I would put this situation a bit differently.
The element selection is found, but information which Xray line is missing. Catching KeyError at same level as AttributeError is not good solution here. It would be good to see the file. I guess this situation is only for a limited atom number in some of Esprit OEM adaptations/versions, where is only single obvious choose of line (Li, Be, B, N, C, O, F, Ne?) it does not fill line and energy information. That could had slip past me when I wrote that code, as I had no such selection of low atomic number elements, or my version was filling line always. If it is the case of these low atomic number elements, it would be best to fill the line simply (add that dict node as `KA`). `energy` part of dictionary could be just skipped - it is not used in HyperSpy and I had not seen in Esprit that we could customize that (albeit on some WDS systems energy sometimes are let to be refined) it is not needed.

hakonanes:
> The element selection is found, but information which Xray line is missing.

Thank you for defining the issue more clearly. I'm happy to close this PR if another is opened which properly fixes the issue.

sem-geologist:
I had just checked on Esprit 1.9 and 2.1 Element tables, and I see that as for Li to Al (including) there is only single K line exposed (not even called `Ka`, but just `K`); Si is the first element which gives a choice for Ka or Kb. The least troublesome workaround would be to to change the lines:
```python
                self.elements[tmp_d['XmlClassName']] = {'line': tmp_d['Line'],
                                                        'energy': tmp_d['Energy']}
            except AttributeError
...
```
into using python dict .get method:
```python
                self.elements[tmp_d['XmlClassName']] = {'line': tmp_d.get('Line', 'Ka'),
                                                        'energy': tmp_d.get('Energy')}
            except AttributeError:
...
```
@hakonanes , You had already invested some time in this PR (and I see there is some typos fixes), please include the above suggested change and don't close this.

as for changelog note:
it should be then changed into something like this: "in case Bruker defined xml element node at  `SpectrumRegion` contains no information on specific selected X-ray line (if there is only single line available), suppose it is 'Ka' line."

hakonanes:
Thank you @sem-geologist, I've made the changes you suggested.

sem-geologist:
Looks good for me.

hyperspy/io_plugins/bruker.py
```diff
@@ -743,8 +743,8 @@ def _set_elements(self, root):
             for j in elements.findall(
                     "./ClassInstance[@Type='TRTSpectrumRegion']"):
                 tmp_d = dictionarize(j)
-                self.elements[tmp_d['XmlClassName']] = {'line': tmp_d['Line'],
-                                                        'energy': tmp_d['Energy']}
+                self.elements[tmp_d['XmlClassName']] = {'line': tmp_d.get('Line', 'Ka'),
```
ericpre, line 749: Add a comment to explain, why we fall back to `'Ka'` when there is no `'line'` key?

ericpre (commented):
Thanks @sem-geologist for figuring out the possible cause of the issue, could you please provide a small test file? @TMoon96 doesn't seem to reply and we get finish this PR without waiting for him to provide a test file or check that it fixes #2881.

hyperspy/io_plugins/bruker.py
hakonanes: Thanks for following this up. I added: "In case no information on the specific selected X-ray line is available, assume it is a 'Ka' line, reflecting the fact that element tables in Bruker Esprit (1.9 and 2.1) store a single K line for Li-Al"

sem-geologist:
@ericpre , I tried to get such files, but unfortunately our Esprit 1.9 and Esprit 2.1 fills those Line Tags regardless of atom number. Now it is clear why I supposed those XML `Line` nodes are always present here. I think the proposed fix is very logical to mitigate Error from thus far info we could get from @TMoon96 . Sorry, can't produce such crippled files on our machines.

ericpre (approved):

PR #2902, opened 2022-03-13
Fix test suite hanging

### LINKED ISSUE #2892: Test failing: oldest supported dependencies
Currently the test which checks the oldest supported versions of the dependencies is failing.

The error message is a bit difficult to understand: 
```python
 hyperspy/tests/utils/test_stack.py ................                      [ 99%]
Error: The operation was canceled.
```
so not sure what is causing it.

Might be related to the runtime being over 30 minutes? I vaguely remember the limit being 30 minutes.

ericpre:
This seems to be related to the collection test coverage. Closes #2892.

### Progress of the PR
- [x] Disable test coverage for this build
- [x] ready for review.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -97,7 +97,7 @@ extras_require = {
     # bug in pip: matplotib is ignored here because it is already present in
     # install_requires.
     "tests": ["pytest>=3.6", "pytest-mpl", "pytest-xdist", "pytest-rerunfailures", "pytest-instafail", "matplotlib>=3.1"],
-    "coverage":["pytest-cov", "codecov"],
+    "coverage":["pytest-cov"],
     # required to build the docs
     "build-doc": ["sphinx>=1.7", "sphinx_rtd_theme", "sphinx-toggleprompt", "sphinxcontrib-mermaid", "sphinxcontrib-towncrier"],
 }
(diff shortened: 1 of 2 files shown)


magnunor:
Guess this makes sense, as we really only need the coverage for one of the test-runs? Any reason for running the coverage for the oldest version of the dependencies, in addition to the most recent dependencies?

setup.py
```diff
@@ -97,7 +97,7 @@
     # bug in pip: matplotib is ignored here because it is already present in
     # install_requires.
     "tests": ["pytest>=3.6", "pytest-mpl", "pytest-xdist", "pytest-rerunfailures", "pytest-instafail", "matplotlib>=3.1"],
-    "coverage":["pytest-cov", "codecov"],
```
magnunor, line 100: I'm not too familiar with `codecov`, guessing `pytest-cov` supersedes it?

ericpre: https://github.com/codecov/codecov-python: as it says in the readme, it is deprecated.

ericpre:
> Guess this makes sense, as we really only need the coverage for one of the test-runs? Any reason for running the coverage for the oldest version of the dependencies, in addition to the most recent dependencies?

In principle, yes to cover as much of the code base as possible, typically, when there are branches in the code to support older versions of dependencies.

magnunor:
> In principle, yes to cover as much of the code base as possible, typically, when there are branches in the code to support older versions of dependencies.

Makes sense!

Coverage dropped a little bit, but that is expected given the changes.

Hopefully we can fix that issue at some point in the future.

Issue #2904, opened 2022-03-15
Ragged Arrays with Ragged Dim>2 don't Save


CSSFrancis:
This isn't exactly an issue that we can solve in hyperspy without some kind of hacky workarounds but something that I wanted to bring up here.  

I ran into this issue while working on creating a diffraction vectors class where both the zarr and hdf5 saving methods fail when trying to save a 2 dimensional array.  There was movement to solve this issue in zarr: https://github.com/zarr-developers/numcodecs/pull/200 but this largely seems to have stalled. 

### Example of the Failing Code
```python 
import numpy as np
import hyperspy.api as hs

x = np.empty(shape=(3), dtype=object)

for i in np.ndindex(x.shape):
    x[i] = np.random.random((4, 2))

s = hs.signals.BaseSignal(x).T

s.save("test.zspy")
s.save("test.hspy")
```

For a solution to this I might try to see what it takes to bring the `zarr` PR across the finish line if no one is working on it. I think that this is kind of an edge case, but supporting it in zarr might be significantly easier than trying to get it to work with hdf5. From my understanding there, the work around is to cast the data into a new datatype which seems complicated to get to work in our case.

ericpre:
Yes, indeed, the alternative approach of flattening the array when saving and un-flattening it when reading will be required to support the h5py format - see https://github.com/h5py/h5py/issues/876. Not the most elegant, but since it should be possible to change only the `HierarchicalReader` and `HierarchicalWriter` classes, it wouldn't be that bad neither! ;)

CSSFrancis:
Interesting work around, but not the worst thing as you said.  That is looking more promising to just make the changes to the `HierarchicalReader` and `HierarchicalWriter` classes because it would solve the problem for both cases.

I will start writing something up for fixing the issue.

CSSFrancis:
This is effectively closed with #2906.

PR #2906, opened 2022-03-18
BugFix: Basic way to save ragged arrays with ndim>1

CSSFrancis:
### This is a pretty basic way to fix the saving of signals with a ragged dim greater than 2

The motivation for this is with saving vector signals after some analysis.  Previously these signals weren't saving properly (see #2904)

**_The general procedure:_**
- Unwrapped ragged arrays with signal dim greater than two
- Save the original shape of the signal dim as a separate dataset 
- Load the unwrapped data and original shape
- Reshape the data using the original shape.

### Progress of the PR
- [x]  Change implemented (can be split into several points),
- [ ] add tests,
- [ ] ready for review.

```python
import hyperspy.api as hs
import numpy as np

x = np.empty(5,dtype=object)
for i in range(1,6):
    x[i-1]= np.array([list(range(i)),list(range(i))])

s = hs.signals.BaseSignal(x, ragged=True)

s.save("test.hspy")

s =hs.load("test.hspy")
```

@ericpre  Any suggestions you have here would be greatly appreciated.  Right now it works and is only a rather small change.  That being said it am not extremely happy with how I got it to work.

### DIFF
--- a/hyperspy/io_plugins/_hierarchical.py
+++ b/hyperspy/io_plugins/_hierarchical.py
@@ -234,6 +234,14 @@ class HierarchicalReader:
             exp["package_version"] = ""
 
         data = group['data']
+        try:
+            ragged_shape = group["ragged_shapes"]
+            new_data = np.empty(shape=data.shape, dtype=object)
+            for i in np.ndindex(data.shape):
+                new_data[i] = np.reshape(data[i], ragged_shape[i])
+            data = new_data
+        except KeyError:
+            pass
         if lazy:
             data = da.from_array(data, chunks=data.chunks)
             exp['attributes']['_lazy'] = True
@@ -626,7 +634,17 @@ class HierarchicalWriter:
                     del group[key]
 
         _logger.info(f"Chunks used for saving: {chunks}")
-        cls._store_data(data, dset, group, key, chunks)
+        if data.dtype == np.dtype('O'):
+            new_data = np.empty(shape=data.shape, dtype=object)
+            shapes = np.empty(shape=data.shape, dtype=object)
+            for i in np.ndindex(data.shape):
+                new_data[i] = np.ndarray.flatten(data[i])
+                shapes[i] = data[i].shape
+            shape_dset = cls._get_object_dset(group, shapes, "ragged_shapes", shapes.shape, **kwds)
+            cls._store_data(shapes, shape_dset, group, 'ragged_shapes', chunks=shapes.shape)
+            cls._store_data(new_data, dset, group, key, chunks)
+        else:
+            cls._store_data(data, dset, group, key, chunks)
 
     def write(self):
         self.write_signal(self.signal,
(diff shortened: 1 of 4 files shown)


hyperspy/io_plugins/_hierarchical.py
```diff
@@ -626,7 +634,17 @@ def overwrite_dataset(cls, group, data, key, signal_axes=None,
                     del group[key]
 
         _logger.info(f"Chunks used for saving: {chunks}")
-        cls._store_data(data, dset, group, key, chunks)
+        if data.dtype == np.dtype('O'):
```
ericpre, line 637: To enforce declaring signal as being ragged, I think we should use this approach only when the signal is ragged:
```suggestion
        if self.signal.ragged:
```

hyperspy/io_plugins/_hierarchical.py
```diff
@@ -626,7 +634,17 @@ def overwrite_dataset(cls, group, data, key, signal_axes=None,
...
+            new_data = np.empty(shape=data.shape, dtype=object)
+            shapes = np.empty(shape=data.shape, dtype=object)
+            for i in np.ndindex(data.shape):
+                new_data[i] = np.ndarray.flatten(data[i])
```
ericpre:
```suggestion
                new_data[i] = data[i].flatten()
```

ericpre (commented):
Thanks @CSSFrancis, this approach looks good to me.

ericpre:
I edited the example [above](https://github.com/hyperspy/hyperspy/pull/2906#issue-1173122868) to the recommend way of creating signals of ragged array as per https://hyperspy.readthedocs.io/en/latest/user_guide/signal.html#ragged-array.

hyperspy/io_plugins/_hierarchical.py
CSSFrancis: I don't know if that is the best behavior right now.  Currently the ragged attribute isn't firmly typed i.e. you can have a signal which has an object dtype but isn't ragged.  It probably shouldn't be the case going forward but for now I think that would cause some edge case signal to not save.

hyperspy/io_plugins/_hierarchical.py
CSSFrancis: Is there a reason for doing it this way instead?

ericpre: The method may be marginally faster than the function but the main reason is that it is shorter to write and read!

Actually, in this case, it may be better to use `ravel` to return a view and avoid making a copy.

hyperspy/io_plugins/_hierarchical.py
ericpre: Yes, this is the motivation to enforce the use of `ragged` signal when necessary and special case `ragged` processing only when the signal is explicitly set as ragged, otherwise this is speculative.

If I understand/remember correctly, the issue is with saving ragged array, not `object` type array? If so, I think that using `self.signal.ragged` is more appropriate.

PR #2862, opened 2021-12-13
Add support for JEOL sightX-tiff

nem1234:
### Description of the change
- fix read error when reading JEOL SightX tiff file (time string is different)
- Decode XML style information of images to original_metadata dictionary
- ToDo: decode metadata

### Progress of the PR
- [x] Change implemented (can be split into several points),
- [ ] update docstring (if appropriate),
- [ ] update user guide (if appropriate),
- [x] add an changelog entry in the `upcoming_changes` folder (see [`upcoming_changes/README.rst`](https://github.com/hyperspy/hyperspy/blob/RELEASE_next_minor/upcoming_changes/README.rst)),
- [x] Check formatting changelog entry in the `readthedocs` doc build of this PR (link in github checks)
- [x] add tests,
- [x] ready for review.

### Minimal example of the bug fix or the new feature
```python
import hyperspy.api as hs
import numpy as np
hs.load("sightx.tiff")
```

I will prepare samples of JEOL sightX tiff files.

PR #2908, opened 2022-03-22
Support for Hamamatsu Streak tiff metadata

LMSC-NTappy:
This PR concerns users of streak camera for time-resolved luminescence. Thus, if either @jlaehne, @jordiferrero or @ericpre have time to take a look, it would be great.

A significant part of the changes is a refactoring of the scale-parsing functions for .tiff files obtained from different SEM,TEM and optical microscope manufacturers. Previously, all manufacturer-specific operations where handled in a single function. I have moved them to individual functions and kept the main `_parse_scale_unit` function as some sort of interface.

Tests pass on my side.

Cheers,

Nicolas

### Description of the change
- Added axes and metadata parsing support for .tiff files generated by hamamatsu streak cameras
- Refactored the .tiff `_parse_scale_unit` function for clarity and ease of expanding support in the future
- Support initialization of non-uniform axes
- Issue warning in case of axis linearization error
- Support for a tiff-specific keyword argument in case of non-uniform axis parsing.

### Progress of the PR
- [x]  Updated Documentation (also for #2862, which was missing)
- [x] add tests,
- [x] ready for review.

```python
import hyperspy.api as hs
import hyperspy.tests.io
import os

dirname = os.path.join(os.path.dirname(hyperspy.tests.io.__file__),"tiff_files")
file = 'test_hamamatsu_streak_SCAN.tif'
#Get the example streak map
fname = os.path.join(dirname, file)

s = hs.load(fname) #Load as a linear axes but warn about linearisation
#Load as DataAxis
s = hs.load(fname, hamamatsu_streak_axis_type= 'data')
#Load as FunctionalDataAxis (3rd degree polynomial)
s = hs.load(fname, hamamatsu_streak_axis_type= 'functional')
#Load as UniformDataAxis without warning
s = hs.load(fname, hamamatsu_streak_axis_type= 'uniform')  
```


jlaehne:
A colleague has a matlab script to read the binary format for streak camera data (I think also Hamamatsu). Would be a good first IO-plugin for https://github.com/lumispy/lumispy . I would try to get a student to do the implementation, unless someone wants to give it a shot already.

hyperspy/io_plugins/tiff.py
```diff
@@ -451,9 +438,188 @@ def wave_len(ht):
...
+    if not 'Software' in op:
+        is_hamatif = False
+        return is_hamatif
+    else:
+        software = op['Software']
+        if not software.startswith('HPD-TA'):
+            is_hamatif = False
+
+    return is_hamatif
+def _get_hamamatsu_streak_description(tiff, op):
```
jlaehne, line 536: blank line before new `def` (several occurences)

hyperspy/io_plugins/tiff.py
```diff
@@ -451,9 +438,188 @@ def wave_len(ht):
...
+    for row in reader:
+        key = row[0].strip(" []")
+        key_dict = {}
+        for element in row[1:]:
+            spl = element.split('=')
+            if len(spl) == 2:
+                key_dict[spl[0]] = spl[1].strip('"')
+        dict_meta[key] = key_dict
+
+    import re
```
jlaehne: Imports in the middle of functions are quite unusual. Should at least be at the top of the method, but I think they are generally at the top of the files across HyperSpy,

jlaehne:
Looks good on first sight. A bit hard to review due to the restructuring (but I also hope the tests do their jobs as far as they exist).

There are quite a number of codecov warnings (see 'Files changed' tab). Could you try to increase coverage further, even though most of them are for the code that you moved around?

hyperspy/io_plugins/tiff.py
LMSC-NTappy: Ok, the idea was to run import only if the .tiff is recognized as hamamatsu but fair enough it doesn't matter much

hyperspy/io_plugins/tiff.py
LMSC-NTappy: corrected (also other style improvements)

LMSC-NTappy:
Thank you for reviewing, I have implemented your recommendations

> A colleague has a matlab script to read the binary format for streak camera data (I think also Hamamatsu). Would be a good first IO-plugin for https://github.com/lumispy/lumispy . I would try to get a student to do the implementation, unless someone wants to give it a shot already.

Always good to support more files I guess. On my side, working with .tiff files is enough so go ahead.

> Looks good on first sight. A bit hard to review due to the restructuring (but I also hope the tests do their jobs as far as they exist).


I understand, but in the long run I hope this helps further extending .tiff loading capabilities with a cleaner code.


> There are quite a number of codecov warnings (see 'Files changed' tab). Could you try to increase coverage further, even though most of them are for the code that you moved around?

The missing lines cover manufacturer-specific .tiff metadata (Zeiss, olympus etc), which I don't have access to. I could recreate it by altering another file with the tiff library I guess, but that would be for the sole purpose of making sure that the loader does not attempt to read metadata in very specific cases of invalid files (e.g, the manufacturer-expected field exists but it has not the correct value inside). I think we can trust the software to work as intended in those cases

LMSC-NTappy:
I have one more question though. 

In the hamamatsu case I cover, there is a linearization of the time axis being performed. The non-linearized axes are put in the `original_metadata` field for the sake of preserving information. 

In that case, is the policy to raise a warning upon loading the file?

jlaehne:
I could provide a Zeiss tiff file, if you would be willing to integrate that in the tests?

LMSC-NTappy:
I don't think there are any uncovered lines regarding zeiss .tiffs in the codecov right now.

The unhandled cases are: 
- conversion of the tiff unit when it was saved as centimeters and loaded calling force_read_resolution : https://github.com/LMSC-NTappy/hyperspy/blob/18daa09f0c225c2085dfcf631f5acb250c48c9d6/hyperspy/io_plugins/tiff.py#L305-L308
- Missing `'FEI_HELIOS`' key in an FEI_SEM metadata file: https://github.com/LMSC-NTappy/hyperspy/blob/18daa09f0c225c2085dfcf631f5acb250c48c9d6/hyperspy/io_plugins/tiff.py#L322-L326
- Missing `'PixelSizeX'` key in a tvips file: https://github.com/LMSC-NTappy/hyperspy/blob/18daa09f0c225c2085dfcf631f5acb250c48c9d6/hyperspy/io_plugins/tiff.py#L373-L376
- Jeol_sightX file for unit set in cm and not inches: https://github.com/LMSC-NTappy/hyperspy/blob/18daa09f0c225c2085dfcf631f5acb250c48c9d6/hyperspy/io_plugins/tiff.py#L430-L434
- Jeol_sightX file and camera length set in cm and not mm: https://github.com/LMSC-NTappy/hyperspy/blob/18daa09f0c225c2085dfcf631f5acb250c48c9d6/hyperspy/io_plugins/tiff.py#L452-L453

For the unit conversion cases, I think we can trust that if it works correctly from inches/mm, it will also work for cm since it's just the factor that changes.

For my hamamatsu metadata, I don't test that the `_is_streak_hamamatsu` function correctly returns False if the `'Artist'` or `'Software'` keys of metadata do exist but contain values incompatible with a streak map. I can write test for it if you want but tbh I really don't think it's necessary

hyperspy/models/model1d.py
```diff
@@ -436,7 +436,7 @@ def __call__(self, non_convolved=False, onlyactive=False,
             if self.signal.axes_manager[-1].is_uniform:
                 to_return *= self.signal.axes_manager[-1].scale
             else:
-                to_return *= np.gradient(self.signal.axes_manager[-1].axis)
+                to_return *= np.gradient(self.signal.axes_manager[-1].axis[self.channel_switches])
```
ericpre: This is unrelated to this PR, can you please do this change in a separate PR?

hyperspy/io_plugins/tiff.py
```diff
@@ -254,6 +252,9 @@ def _read_serie(tiff, serie, filename, force_read_resolution=False,
     else:
         dc = _load_data(*data_args, memmap=memmap, **kwds)
 
+    if _is_streak_hamamatsu(op):
+        op.update({'ImageDescription_Parsed': _get_hamamatsu_streak_description(tiff, op)})
```
ericpre:
```suggestion
        op.update({'ImageDescriptionParsed': _get_hamamatsu_streak_description(tiff, op)})
```
Just to avoid mixing CamelCase and snake_case in the same name! ;)

ericpre (commented):
> I have one more question though.
>
> In the hamamatsu case I cover, there is a linearization of the time axis being performed. The non-linearized axes are put in the original_metadata field for the sake of preserving information.
>
> In that case, is the policy to raise a warning upon loading the file?

This sounds like a very good opportunities to use a non-uniform axis?! :)

> For my hamamatsu metadata, I don't test that the _is_streak_hamamatsu function correctly returns False if the 'Artist' or 'Software' keys of metadata do exist but contain values incompatible with a streak map. I can write test for it if you want but tbh I really don't think it's necessary

If you think that it is not worth it adding a test, then please add `  # pragma: no cover`. In this case, this can be sorted by writing a simple test:  which could consist in opening a Hamamatsu file with tifffile, edit the tag, save the file and try to re-open it with hyperspy. Because it should be a simple test, I would be in favour of adding it.

Is it possible to have smaller test file? 600KB is not a big deal, but smaller would be better (10s of KB).

hyperspy/models/model1d.py
LMSC-NTappy: Oups, sorry my bad

LMSC-NTappy:
> This sounds like a very good opportunities to use a non-uniform axis?! :)

Indeed! I tend to forget that non-uniform axes will be released together with this patch :). I'll head this way but then I will need to pass **kwds to `_parse_scale_unit` so that the axis can be enforced linear. I am sure otherwise there will be some plot issues if one axis is linear and the other one is not.

> Is it possible to have smaller test file? 600KB is not a big deal, but smaller would be better (10s of KB).

I know, but I just cannot get anything smaller than that... I don't know how to crop/rebin/reduce their size without messing up the metadata. That's also why `_is_streak_hamamatsu` is so strict. If the file is altered everything seem to go south. You have a hint on how to do it maybe?

LMSC-NTappy:
> If you think that it is not worth it adding a test, then please add # pragma: no cover

I just have added the tests in the end. Wasn't such a big deal. Let's see if coverage improves

LMSC-NTappy:
Ok apparently the failing test are due to a deprecation in the tifffile library. 

I will fix that, but probably by removing the saving of .tiff altogether and manually passing the modified dictionaries to `_is_streak_hamamatsu`.  Otherwise it will get a bit messy

ericpre:
> > This sounds like a very good opportunities to use a non-uniform axis?! :)
> 
> Indeed! I tend to forget that non-uniform axes will be released together with this patch :). I'll head this way but then I will need to pass **kwds to `_parse_scale_unit` so that the axis can be enforced linear. I am sure otherwise there will be some plot issues if one axis is linear and the other one is not.

Yes, since not all feature are supported with non-uniform axes, it is possible that plot doesn't work out of the box. An alternative is to convert manually after reading the data using `convert_to_uniform_axis`.

> > Is it possible to have smaller test file? 600KB is not a big deal, but smaller would be better (10s of KB).
> 
> I know, but I just cannot get anything smaller than that... I don't know how to crop/rebin/reduce their size without messing up the metadata. That's also why `_is_streak_hamamatsu` is so strict. If the file is altered everything seem to go south. You have a hint on how to do it maybe?

Ok, if this is not possible to acquire a smaller file, then this is not worth it.
If it uses compression, it may be worth trying acquiring an image with all values identical, as this is something that should compress well - obviously, this is assuming that the size of the data comes from the data itself and not the metadata!

LMSC-NTappy:
> Yes, since not all feature are supported with non-uniform axes, it is possible that plot doesn't work out of the box. An alternative is to convert manually after reading the data using `convert_to_uniform_axis`.

I will give it a shot nevertheless, but there a bit more of refactoring involved so it will take a bit of time. 


> Ok,` if this is not possible to acquire a smaller file, then this is not worth it.
If it uses compression, it may be worth trying acquiring an image with all values identical, as this is something that should compress well - obviously, this is assuming that the size of the data comes from the data itself and not the metadata!

That's a very good point. Saving the file in compressed form reduces to 35 kB but so far I didn't have any luck in preserving the metadata. I will check next week with the tool administrator, who also has access to the software, if he can help me with this

Cheers

Nico

LMSC-NTappy:
Ok, now implemented with support for reading non-uniform axes. Behaviour is as follow

```python
s = hs.load('example_hamamatsu.tiff') #Load as a linear axes but warn about linearisation
#Load as DataAxis
s = hs.load('example_hamamatsu.tiff',hamamatsu_streak_axis_type= 'data')
#Load as FunctionalDataAxis (3rd degree polynomial)
s = hs.load('example_hamamatsu.tiff',hamamatsu_streak_axis_type= 'functional')
#Load as UniformDataAxis without warning
s = hs.load('example_hamamatsu.tiff',hamamatsu_streak_axis_type= 'uniform')  
```
After discussing with @jordiferrero we concluded that applying linearisation by default is ok if the user is warned about it.

This will very soon be ready for review, but I'll need to update the docs and change kwds terminology

The merge conflict I have tried to solve numerous time but so far it has escaped my attempts (and understanding)

LMSC-NTappy:
Voilà, now I am ready for a final review!

jordiferrero:
> Ok, now implemented with support for reading non-uniform axes. Behaviour is as follow
> 
> ```python
> s = hs.load('example_hamamatsu.tiff') #Load as a linear axes but warn about linearisation
> #Load as DataAxis
> s = hs.load('example_hamamatsu.tiff',hamamatsu_streak_axis_type= 'data')
> #Load as FunctionalDataAxis (3rd degree polynomial)
> s = hs.load('example_hamamatsu.tiff',hamamatsu_streak_axis_type= 'functional')
> #Load as UniformDataAxis without warning
> s = hs.load('example_hamamatsu.tiff',hamamatsu_streak_axis_type= 'uniform')  
> ```
> 
> After discussing with @jordiferrero we concluded that applying linearisation by default is ok if the user is warned about it.
> 
> This will very soon be ready for review, but I'll need to update the docs and change kwds terminology
> 
> The merge conflict I have tried to solve numerous time but so far it has escaped my attempts (and understanding)

Thanks @LMSC-NTappy ! That's lots of work in this PR.
This PR (with the implementation of `non_linear_axes` made us realise that most loader functions assume linear axis. While this is has always been the standard in hyperspy until now, it means the user is always modifying the raw data upon `hs.load`.
Does this mean that from now onwards, the `hs.load` should always be able to have the non-linear axis argument as done in the example above? Do you think there may be value for letting the user choose if they want the original data axes, a functional one or a linearised one? If you think this should be a point of discussion I can open an Issue :-)

doc/user_guide/io.rst
```diff
@@ -772,17 +772,46 @@ Currently HyperSpy has limited support for reading and saving the TIFF tags.
...
+    linearisation error. Initialise a UniformDataAxis.
+  * if ``data``, the raw time axis parsed from the metadata is used. Initialise
+    a DataAxis.
+  * if ``functional``, the best-fit 3rd-order polynomial axis is used, avoiding
+    linearisation error. Initialise a FunctionalDataAxis.
+
+By default, ``uniform`` is used but a warning of the linearisation error is issued.
+Explicitly passing ``hamamatsu_streak_axis_type='uniform'`` suppresses the warning.
+In all cases, the original axis values are stored in the ``original_metadata`` of the
+signal object.
```
jordiferrero, line 798: I suppose this is visible in the docstrings of the `hs.load` function? Or is it hidden inside other sub-loading functions?

jordiferrero (approved):

doc/user_guide/io.rst
LMSC-NTappy: >  Or is it hidden inside other sub-loading functions?

Exactly this. The load docstring is encyclopedic already, so I think new information would be lost inside. By default, a warning is issued specifying that information directly in the Command Line so  people won't miss it

hyperspy/io_plugins/tiff.py
```diff
@@ -178,41 +208,16 @@ def _read_serie(tiff, serie, filename, force_read_resolution=False,
...
-    axes = [{'size': size,
-             'name': str(name),
-             'scale': scale,
-             'offset': offset,
-             'units': unit,
-             }
-            for size, name, scale, offset, unit in zip(shape, names,
-                                                       scales, offsets,
-                                                       units)]
+        axes = _build_axes_dictionaries(shape, names, scales=[1.0] * len(names),
```
ericpre: Maybe it is more simple to set `None` as default value for `names`, `scale`, `offset` and `units` and let `_build_axes_dictionaries` figure out what should be the shape and default values.

hyperspy/io_plugins/tiff.py
```diff
@@ -275,185 +289,483 @@ def _load_data(serie, is_rgb, sl=None, memmap=None, **kwds):
...
+        del op['FEI_SFEG']
+    scales['x'] = float(op['fei_metadata']['Scan']['PixelWidth'])
+    scales['y'] = float(op['fei_metadata']['Scan']['PixelHeight'])
+    units.update({'x': 'm', 'y': 'm'})
+
+    # Finally, axes descriptors can be additionally parsed from digital micrograph or imagej-style files
+    if _is_digital_micrograph(op):
+        scales, offsets, units = _add_axes_digital_micrograph(op, scales, offsets, units)
+    if _is_imagej(tiff):
+        scales, offsets, units = _add_axes_imagej(tiff, op, scales, offsets, units)
```
ericpre: Is this necessary? If the file is from FEI, we should use the calibration from the FEI metadata?

hyperspy/io_plugins/tiff.py
```diff
@@ -275,185 +289,483 @@ def _load_data(serie, is_rgb, sl=None, memmap=None, **kwds):
...
+    # can be absent and thus is not used here.
+    scale_in_m = op['CZ_SEM'][''][3] * 1024 / tiff.pages[0].shape[1]
+    scales.update({'x': scale_in_m, 'y': scale_in_m})
+    units.update({'x': 'm', 'y': 'm'})
+
+    # Finally, axes descriptors can be additionally parsed from digital micrograph or imagej-style files
+    if _is_digital_micrograph(op):
+        scales, offsets, units = _add_axes_digital_micrograph(op, scales, offsets, units)
+    if _is_imagej(tiff):
+        scales, offsets, units = _add_axes_imagej(tiff, op, scales, offsets, units)
```
ericpre: Same as above

hyperspy/io_plugins/tiff.py
```diff
@@ -275,185 +289,483 @@ def _load_data(serie, is_rgb, sl=None, memmap=None, **kwds):
...
+        _logger.debug("getting TVIPS scale from XYResolution")
+        scales['x'], scales['y'] = _get_scales_from_x_y_resolution(
+            op, factor=1e-2)
         units.update({'x': 'm', 'y': 'm'})
 
+    # Finally, axes descriptors can be additionally parsed from digital micrograph or imagej-style files
+    if _is_digital_micrograph(op):
+        scales, offsets, units = _add_axes_digital_micrograph(op, scales, offsets, units)
+    if _is_imagej(tiff):
+        scales, offsets, units = _add_axes_imagej(tiff, op, scales, offsets, units)
```
ericpre: Same as above.

hyperspy/io_plugins/tiff.py
```diff
@@ -275,185 +289,483 @@ def _load_data(serie, is_rgb, sl=None, memmap=None, **kwds):
...
+    op['Olympus_SIS_metadata'] = sis_metadata
+    scales['x'] = round(float(sis_metadata['pixelsizex']), 15)
+    scales['y'] = round(float(sis_metadata['pixelsizey']), 15)
+    units.update({'x': 'm', 'y': 'm'})
+
+    # Finally, axes descriptors can be additionally parsed from digital micrograph or imagej-style files
+    if _is_digital_micrograph(op):
+        scales, offsets, units = _add_axes_digital_micrograph(op, scales, offsets, units)
+    if _is_imagej(tiff):
+        scales, offsets, units = _add_axes_imagej(tiff, op, scales, offsets, units)
```
ericpre: Same as above.

hyperspy/io_plugins/tiff.py
```diff
@@ -275,185 +289,483 @@ def _load_data(serie, is_rgb, sl=None, memmap=None, **kwds):
...
+        scales['x'], scales['y'] = _get_scales_from_x_y_resolution(op, factor=scale)
         units = {"x": "1 / nm", "y": "1 / nm", "z": t.Undefined}
-    return scales, units, offsets, intensity_axis
 
-def _get_scales_from_x_y_resolution(op, factor=1):
+    # Finally, axes descriptors can be additionally parsed from digital micrograph or imagej-style files
+    if _is_digital_micrograph(op):
+        scales, offsets, units = _add_axes_digital_micrograph(op, scales, offsets, units)
+    if _is_imagej(tiff):
+        scales, offsets, units = _add_axes_imagej(tiff, op, scales, offsets, units)
```
ericpre: Same as above

hyperspy/io_plugins/tiff.py
```diff
@@ -275,185 +289,483 @@ def _load_data(serie, is_rgb, sl=None, memmap=None, **kwds):
...
+        axes = _axes_hamamatsu_streak(tiff, op, shape, names, **kwds)
+        return axes
+    # Axes are otherwise set to defaults
+    else:
+        scales, offsets, units = _axes_defaults()
+        # Axes descriptors can be additionally parsed from digital micrograph or imagej-style files
+        if _is_digital_micrograph(op):
+            scales, offsets, units = _add_axes_digital_micrograph(op, scales, offsets, units)
+        if _is_imagej(tiff):
+            scales, offsets, units = _add_axes_imagej(tiff, op, scales, offsets, units)
```
ericpre, line 737: Move these above at the same level than others (fei, zeiss, etc.)?

ericpre (commented):
This looks good. Can you try to compress the two files? You can take example on how to do it from #2926.
See comments about the calibration.

hyperspy/io_plugins/tiff.py
LMSC-NTappy: It seems to me that the tiff loader/writer plugin was written to support digital micrograph or imagej metadata on top of other file formats. 

I think the usecase would be loading an FEI file with FEI axes calibrations, but with some calibration aspects, e.g. one axis or a noise model (see _test_read_unit_from_dm in the test_tiff.py for eyample), calibrated using dm3 or imagej-style metadata.

The problem is that not all the metadata parsers are supporting this, so I found it difficult to address this in a single level of code...

hyperspy/io_plugins/tiff.py
LMSC-NTappy: Will try to do it

hyperspy/io_plugins/tiff.py
ericpre: > It seems to me that the tiff loader/writer plugin was written to support digital micrograph or imagej metadata on top of other file formats.
>
> I think the usecase would be loading an FEI file with FEI axes calibrations, but with some calibration aspects, e.g. one axis or a noise model (see _test_read_unit_from_dm in the test_tiff.py for eyample), calibrated using dm3 or imagej-style metadata.

I am not sure, it may simply have been changed over time in an inconsistent manner...
 
> The problem is that not all the metadata parsers are supporting this, so I found it difficult to address this in a single level of code...

This is why, I would remove it.

ericpre:
@LMSC-NTappy, when do you think you will have time to finish it?

LMSC-NTappy:
> @LMSC-NTappy, when do you think you will have time to finish it?

Earliest I can do is to free some time on tuesday to wrap it up. Otherwise when I am back from a conference on may the 2nd

jlaehne:
Thanks Eric for finishing this one up!

Issue #2898, opened 2022-03-06
Bruker M4 Tornado X-ray fluorescence BCF file import works incorrectly


mstatkus:
BCF files are handled by `bruker.py` module, developed for SEM/TEM Bruker data. X-Ray fluorescence (XRF) mapping instrument Bruker M4 Tornado uses the same overall BCF file structure, but with many differences in metadata structure. Accelerating voltage data, required for processing of SEM files, is not provided in XRF BCF file header.

(I'm not sure whether it's a bug report or feature request, because there's no 'official' support of XRF BCFs.)

When `bruker.py` processes XRF BCF, `HyperHeader._set_microscope()` method is called in `HyperHeader. __init__()` (line 599) , voltage metadata is not found and `self.hv` is set to default value 0.0 (line 625).

`self.hv = self.sem_metadata.get('HV', 0.0)  # in kV`

Next, `HyperHeader.estimate_map_channels()` method is called by `BCF_reader.parse_hypermap` method (line 957); with `self.hv` set to zero, it results in incorrect cropping of energy axis (lines 768-772).

```
        bruker_hv_range = self.spectra_data[index].amplification / 1000
        if self.hv >= bruker_hv_range:
            return self.spectra_data[index].data.shape[0]
        else:
            return self.spectra_data[index].energy_to_channel(self.hv)
```

Example code, test BCF file is attached as .zip ([M4-xrf-test.zip](https://github.com/hyperspy/hyperspy/files/8193214/M4-xrf-test.zip) - sorry, it is about 25 Mb, I wasn't able to produce smaller one ).

```
import hyperspy.api as hs

hmap_full_load = hs.load('M4-xrf-test.bcf')
print (hmap_full_load[4].data.shape) 
# energy axis has 96 channels - incorrect
```

A simple workaround is to call file import with `cutoff_at_kV=40.` (or whatever the upper limit of energy axis is):

```
hmap_cutoff_load = hs.load('M4-xrf-test.bcf',cutoff_at_kV=40.)
print (hmap_cutoff_load[4].data.shape) 
# energy axis has 4095 channels - correct
```

However, for correct processing of X-ray lines, it is necessary to set 'fake' beam energy value after loading BCF file. 
```
hm = hmap_cutoff_load[4]
im = hm.get_lines_intensity()
# several lines are above limit of 50/2 kEv - errors

hm.set_microscope_parameters(beam_energy = 80)
# setting fake beam energy to 2*spectra energy range
[...]

PR #2910, opened 2022-03-25
FIX-ENH update cutoff_at_kV behavior for bruker.py reader

sem-geologist:
### Description of the change
Changes how bruker.py reader reacts to cutoff_at_kV:
* `None` as value does no more any cutoff
* introduced possible value `"zealous"` which cuts at last nonzero value
* previous behavior with None can be activated with value `"auto"`, with updated fallback if hv=0

This partly fixes the #2898

### Progress of the PR
- [x] update docstring
- [x] update user guide,
- [x] add/modify tests,
- [x] ready for review.

### DIFF
--- a/doc/user_guide/io.rst
+++ b/doc/user_guide/io.rst
@@ -1047,8 +1047,16 @@ Extra loading arguments
   The underlying method of downsampling is unchangeable: sum. Differently than
   ``block_reduce`` from skimage.measure it is memory efficient (does not creates
   intermediate arrays, works inplace).
-- ``cutoff_at_kV`` : if set (can be int or float >= 0) can be used either to crop
-  or enlarge energy (or channels) range at max values (default None).
+- ``cutoff_at_kV`` : if set (can be None, int, float (kV), one of 'auto' or
+  'lowest_constraint') can be used either to crop or enlarge energy (or number of
+  channels) range at max values. It can be used to conserve memory or enlarge
+  the range if needed to mach the size of other file. Default value is None
+  (which does not influence size). Numerical values should be in kV.
+  "auto" truncates to the last non zero channel (should not be used for stacks,
+  as low beam current EDS can have different last non zero channel per slice).
+  "lowest_constraint" truncates channels to SEM/TEM acceleration voltage or 
+  energy at last channel, depending which is smaller
+  (should be used only if X-Rays were generated using electron beam).
 
 Example of loading reduced (downsampled, and with energy range cropped)
 "spectrum only" data from bcf (original shape: 80keV EDS range (4096 channels),
--- a/doc/user_guide/io.rst
+++ b/doc/user_guide/io.rst
@@ -1060,34 +1060,35 @@ Extra loading arguments
 
 Example of loading reduced (downsampled, and with energy range cropped)
 "spectrum only" data from bcf (original shape: 80keV EDS range (4096 channels),
-100x75 pixels):
+100x75 pixels; SEM acceleration voltage: 20kV):
 
 .. code-block:: python
 
     >>> hs.load("sample80kv.bcf", select_type='spectrum', downsample=2, cutoff_at_kV=10)
     <EDSSEMSpectrum, title: EDX, dimensions: (50, 38|595)>
 
-load the same file without extra arguments:
+load the same file to limiting array size to SEM acceleration voltage:
 
 .. code-block:: python
 
-    >>> hs.load("sample80kv.bcf")
+    >>> hs.load("sample80kv.bcf", cutoff_at_kV='lowest_constraint')
     [<Signal2D, title: BSE, dimensions: (|100, 75)>,
     <Signal2D, title: SE, dimensions: (|100, 75)>,
-    <EDSSEMSpectrum, title: EDX, dimensions: (100, 75|1095)>]
+    <EDSSEMSpectrum, title: EDX, dimensions: (100, 75|1024)>]
 
 The loaded array energy dimension can by forced to be larger than the data
 recorded by setting the 'cutoff_at_kV' kwarg to higher value:
 
 .. code-block:: python
 
-    >>> hs.load("sample80kv.bcf", cutoff_at_kV=80)
+    >>> hs.load("sample80kv.bcf", cutoff_at_kV=60)
     [<Signal2D, title: BSE, dimensions: (|100, 75)>,
     <Signal2D, title: SE, dimensions: (|100, 75)>,
-    <EDSSEMSpectrum, title: EDX, dimensions: (100, 75|4096)>]
+    <EDSSEMSpectrum, title: EDX, dimensions: (100, 75|3072)>]
 
-Note that setting downsample to >1 currently locks out using SEM imagery
-as navigator in the plotting.
+loading without setting cutoff_at_kV value would return data with all 4096
+channels. Note that setting downsample to >1 currently locks out using SEM
+images for navigation in the plotting.
 
 .. _spx-format:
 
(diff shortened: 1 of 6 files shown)


sem-geologist:
looks that fixing the behaviour of cutoff_at_kV makes it behaves similarly to jeol reader where None does not do any hideous auto cutting, I think I should also now address and change docstring of common reader. I see it also has downsampling. I think there could be separate PR to look through all of readers and find same keywords to update.  I think this PR is complete.

hyperspy/io_plugins/bruker.py
```diff
@@ -938,9 +942,12 @@ def parse_hypermap(self, index=None,
...
             is 1.
-        cutoff_at_kV : None or float
-            Value in keV to truncate the array at. Helps reducing size of
-            array. Default is None.
+        cutoff_at_kV : None, float, int or str
+            Value or method to truncate the array at energy in kV. Helps reducing size of
+            array. Default is None (does not truncate). Numerical value is in kV.
+            "auto" truncates to the last non zero channel (should not be used with slices).
+            "lowest_constraint" truncates to instruments acceleration voltage
+            (should be used only if xrays were generated using electron bombardment).
```
ericpre: What happen if `"lowest_constraint"` is used with XRF data? Should it raise an error?

hyperspy/io.py
```diff
@@ -236,10 +236,13 @@ def load(filenames=None,
...
     cutoff_at_kV : None, int, float, optional
-        For Bruker bcf files, if set to numerical (default is None),
-        bcf is parsed into array with depth cutoff at coresponding given energy.
+        For Bruker bcf files and Jeol, if set to numerical (default is None),
+        hypermap is parsed into array with depth cutoff at set energy value.
         This allows to conserve the memory by cutting-off unused spectral
         tails, or force enlargement of the spectra size.
+        Bruker bcf reader accepts additional values: ``auto`` which cuts off
+        after last nonzero channel, ``lowest_constraint`` which cuts at 
+        electron beam current or last channel, dependent from which is lower.
```
ericpre: The description is [io.rst](https://github.com/hyperspy/hyperspy/pull/2910/files#diff-8f0479394ca2e6249c3c98028fce52b98b42e9aedecca1afbb2d350ca321c574) is less ambiguous and more comprehensive, use it to improve this one here and the one in `bruker.py`?

ericpre (commented):
Looks good, I have left some questions/suggestions.

hyperspy/io_plugins/bruker.py
sem-geologist: I hate errors, users hate errors. Errors should be thrown as last option if other does not work. I think in case lowest constraint is used it should check for hv, and if it is ==0, or if `.hv` instance is not here then fallback to channel number. That would prevent uXRF people of shooting self to the foot.

I also remind myself another scenario where `lowest_constraint` would not be good to use: "slicing" depth of the xray excitation with increasing/decreasing HV. I should also add that warning to the docstrings.

sem-geologist: or actually, `lowest_constraint` is not clever and redundant. I should replace it with simple `em_hv` (electron microscope high voltage) - simple to remember,  short to type, clear usage case. if no hv is present - then fallback to None and throw warnning.

ericpre: > I hate errors, users hate errors. Errors should be thrown as last option if other does not work. I think in case lowest constraint is used it should check for hv, and if it is ==0, or if `.hv` instance is not here then fallback to channel number. That would prevent uXRF people of shooting self to the foot.

Well, raising error has the advantage of being transparent with users and avoid behaviour that users may find surprising - it is always possible to find someone who be surprised. :D

sem-geologist: Ok, I remade `auto` to `last_non-zero`; and `lowest_constraint` to `em_hv` (electron microscope HV); which I think is more self explanatory. Now I need to adapt rest of files to that.

sem-geologist:
@ericpre , what do you think about the changes made? ("auto" to "last_non-zero"; "lowest_constraint" to "em_hv")

doc/user_guide/io.rst
```diff
@@ -1047,39 +1047,50 @@ Extra loading arguments
...
-  or enlarge energy (or channels) range at max values (default None).
+- ``cutoff_at_kV`` : if set (can be None, int, float (kV), one of 'last_non-zero'
+  or 'em_hv') can be used either to crop or enlarge energy (or number of
+  channels) range at max values. It can be used to conserve memory or enlarge
+  the range if needed to mach the size of other file. Default value is None
+  (which does not influence size). Numerical values should be in kV.
+  "last_non-zero" truncates to the last non zero channel (this option
+  should not be used for stacks, as low beam current EDS can have different
+  last non zero channel per slice). "em_hv" truncates channels to SEM/TEM
+  acceleration voltage or energy at last channel, depending which is smaller.
```
ericpre: For `"em_hv"`, we would expect (from the name ) to consider only the acceleration voltage and not the last non zero channel. Maybe it would be good to have an `"auto"` option:
-  `"em_hv"` cut at the acceleration voltage
- `"auto"` cut at the minimum of acceleration voltage and last non zero energy channel

sem-geologist: last channel != last non-zero channel

sem-geologist: to last nonzero there is other option `last_non-zero`

sem-geologist: `auto` is really not good as it is not clearvwhat to expect

ericpre: Yes, sorry, I misread. Why do you need to consider the energy at last channel?

sem-geologist: case1: hv 25kV; channels scales up to 20kV; last non-zero channel coresponds to 20kV; `em_hv` cuts at 20kV; `last_non-zero` same.
case2: hv 15kV; channels scales up to 20kV; last non-zero channel (very light material) coresponds to 14.65kV; em_hv cuts at 15kV; `last_non-zero` at 14.65kV.

sem-geologist: in case of detailed single large hypermap last_non-zero will be the most optimal option.

sem-geologist: and with esprit 2 64bit edition limit of 1024x768 raster size is lifted.

sem-geologist: also `last_non-zero` is safe auto option for xrf; where `em_hv` not.

ericpre: What I find inconsistent is that in case1 with `em_hv`, you end up with a value different than the acceleration voltage instead of extending the range of energy as it would do if you specify a value. If the energy range extension wouldn't be implemented when passing float value, I wouldn't find it "inconsistent". This is why I suggested to add the "auto" which will automatically choose the minimum between the two values.

This is not a big deal, but I think that adding the "auto" would be slightly more clearer.

sem-geologist: so probably I should switch `em_hv` back to `least_constraint` or `auto`. I forgot that I also included checking for no or 0.0 hv and in case of uXRF that would fall back to full channel count. So I guess then `auto` is really ok...

sem-geologist: now I am tempted to rename also `last_non-zero` into `zealous` (like Gimp's zealous crop)

ericpre:
Thanks, this looks good to me!
Would it be possible to clean the git history by squashing some of the commits? For example, in this PR, there are some iterative changes and for our future self, typically when browsing the git history, it is convenient not to have commits which do/undo changes.

sem-geologist:
squashed last 12. Is it enough? I was unfortunately mobile and I don't know how on github I could edit few files and commit a bunch. Thus such an amount of commits...

PR #2916, opened 2022-03-31
Fix eelsdb

ericpre:
This PR fix reading some spectra from EELS. For example, the following is broken:
```python
import hyperspy.api as hs

hs.datasets.eelsdb(spectrum_type='coreloss', element='Cu', edge='K', formula='Cu4O3')
hs.datasets.eelsdb(title="Porous cobalt coating with He-filled nanopores")
hs.datasets.eelsdb(title="Zero-loss c-FEG Hitachi Disp 0.214 eV")
```

### Progress of the PR
- [x] Add progress bar when downloading spectra from eelsdb,
- [x] add workaround for malformatted msa file from eelsdb,
- [x] don't warn when the time is missing in msa file,
- [x] don't try to download txt file from the eelsdb - I suggested to https://eelsdb.eu/about/ to convert these file to msa, as this would be the proper fix 
- [x] add tests,
- [x] ready for review.

Now, all spectra (saved as msa file) can be downloaded without error
```python
import hyperspy.api as hs

# get all spectra from the eelsdb database
list_spectra = hs.datasets.eelsdb(show_progressbar=True)
```

### DIFF
--- a/hyperspy/io_plugins/msa.py
+++ b/hyperspy/io_plugins/msa.py
@@ -159,8 +159,8 @@ def parse_msa_string(string, filename=None):
     --------
     file_data_list: list
         The list containts a dictionary that contains the parsed
-        information. It can be used to create a `:class:BaseSignal`
-        using `:func:hyperspy.io.dict2signal`.
+        information. It can be used to create a :py:class:`~.signal.BaseSignal`
+        using :py:func:`~.io.dict2signal`.
 
     """
     if not hasattr(string, "readlines"):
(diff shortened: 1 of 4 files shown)


hyperspy/io_plugins/msa.py
```diff
@@ -211,18 +211,26 @@ def parse_msa_string(string, filename=None):
...
-                        value.replace(' ', ''))
-                except BaseException:
-                    _logger.exception(
-                        "The %s keyword value, %s could not be converted to "
-                        "the right type", parameter, value)
+                error = f"The {parameter} keyword value, {value} could \
+                    not be converted to the right type."
+                if 'e' in value.lower():
+                    # Normally the offending mispelling is a space in the 
+                    # scientic notation, e.g. 2.0 E-06
```
jlaehne:
```suggestion
                    # Normally, the offending misspelling is a space in the
                    # scientific notation, e.g. 2.0 E-06
```

hyperspy/io_plugins/msa.py
```diff
@@ -211,18 +211,26 @@ def parse_msa_string(string, filename=None):
...
+                    try:
+                        parameters[parameter] = type_(value.replace(' ', ''))
+                    except BaseException:
```
jlaehne: Two exceptions not covered, add `#pragma: no cover`?

jlaehne (approved):

PR #2942, opened 2022-05-18
Fix for FEI tif files with no "Scan" key in metadata

### LINKED ISSUE #2941: FEI navcam tifs cannot be opened
#### Describe the bug
TIF-format images saved by FEI FIB/SEMs cannot be opened

#### To Reproduce

Try to open the attached .tif file:

```python
hs.load('51722 Nav-Cam.tif')

WARNING:matplotlib.font_manager:Matplotlib is building the font cache; this may take a moment.
ERROR:hyperspy.io:If this file format is supported, please report this error to the HyperSpy developers.
  File "/home/jat/git_repos/hyperspy/hyperspy/io_plugins/tiff.py", line 222, in _read_serie
    axes = _parse_scale_unit(tiff, page, op, shape, force_read_resolution, names, **kwds)
  File "/home/jat/git_repos/hyperspy/hyperspy/io_plugins/tiff.py", line 713, in _parse_scale_unit
    axes = _axes_fei(tiff, op, shape, names)
  File "/home/jat/git_repos/hyperspy/hyperspy/io_plugins/tiff.py", line 351, in _axes_fei
    scales['x'] = float(op['fei_metadata']['Scan']['PixelWidth'])
KeyError: 'Scan'
```

#### Expected behavior

Metadata should be ignored if not present, rather than causing a crash.

#### Python environement:
 - HyperSpy version: `RELEASE_next_minor`
 - Python version: Any

#### Additional context

Basically, the issue is that there is no `[Scan]` section for Navcam optical images, so it attempts to use the FEI reader, but it fails because of no `[Scan]` key in the metadata dictionary.

[navcam test.tar.gz](https://github.com/hyperspy/hyperspy/files/8719488/navcam.test.tar.gz)
(this image is blacked out, but the metadata is intact)

jat255:
Resolves #2941 

### Progress of the PR
- [x] add tests,
- [x] ready for review.

This is a small change to fix a bug I found (#2941) related to the expectation of certain metadata in all FEI-saved TIFF files. The images saved by the navigation camera do not have a `"Scan"` section, so HyperSpy would error out with a `KeyError`. This change makes that section a `try`/`except`; if it fails, the scale is instead calculated from the horizontal/vertical field widths and the number of pixels in the image. In the event that fails, it returns the image without any scale information. Tests have been added for each case.


hyperspy/io_plugins/tiff.py
```diff
@@ -899,7 +923,7 @@ def _parse_string(value):
 
 mapping_fei = {
     'fei_metadata.Beam.HV':
-        ("Acquisition_instrument.SEM.beam_energy", lambda x: float(x) * 1e-3),
+        ("Acquisition_instrument.SEM.beam_energy", _parse_beam_energy_FEI),
```
francisco-dlp, line 926: Why did you replace the lambda function with a standard function? The lambda function has the advantage that one does not need to scroll to the function to check what it does.

francisco-dlp (commented):
LGTM except for the minor comment in the code.

hyperspy/io_plugins/tiff.py
jat255: This was causing another error if that key doesn't exist (which it won't, for the navigation image). In that case, `x` will be `None` or `''`, and it causes a `TypeError`. It's safer (and more future-proof) to wrap each of those conversions in a `try`/`except`.

francisco-dlp (approved):

Issue #2946, opened 2022-05-24
BaseSignal.map() does not show progressbar


hakonanes:
Am I correct in expecting a progressbar in HyperSpy v1.7 for this operation?

```python
>>> import numpy as np
>>> import scipy.ndimage
>>> import hyperspy.api as hs
>>> s = hs.signals.Signal2D(np.random.random((10, 64, 64)))
>>> s.map(scipy.ndimage.gaussian_filter, show_progressbar=True, inplace=True, sigma=2.5)
```

I'm not getting one. It seems like this call

https://github.com/hyperspy/hyperspy/blob/5f15a5497d27718adda17a7e89b79fff88207e69/hyperspy/signal.py#L5042

should be wrapped in a `dask.diagnostics.ProgressBar()` context if `show_progressbar=True`.

ericpre:
Done in #2947.

PR #2947, opened 2022-05-24
Respect `show_progressbar` parameter in `BaseSignal.map()`

hakonanes:
### Requirements
Addresses #2946.

### Description of the change
`BaseSignal.map(show_progressbar=True)` doesn't show a progressbar. This PR adds it to `BaseSignal._map_iterate()`.

### Progress of the PR
- [n/a] add tests,
- [x] ready for review.

Setup

```python
>>> import numpy as np
>>> import scipy.ndimage
>>> import hyperspy.api as hs
>>> s = hs.signals.Signal2D(np.random.random((10, 64, 64)))
```

Before this PR

```python
>>> s.map(scipy.ndimage.gaussian_filter, show_progressbar=True, inplace=True, sigma=2.5)
```

After

```python
>>> s.map(scipy.ndimage.gaussian_filter, show_progressbar=True, inplace=True, sigma=2.5)
[########################################] | 100% Completed |  0.1s
```

### DIFF
--- a/hyperspy/signal.py
+++ b/hyperspy/signal.py
@@ -30,6 +30,7 @@ from pathlib import Path
 import warnings
 
 import dask.array as da
+from dask.diagnostics import ProgressBar
 from matplotlib import pyplot as plt
 import numpy as np
 from pint import UndefinedUnitError
@@ -5029,6 +5030,10 @@ class BaseSignal(FancySlicing,
 
         data_stored = False
 
+        if show_progressbar:
+            pbar = ProgressBar()
+            pbar.register()
+
         if inplace:
             if (
                 not self._lazy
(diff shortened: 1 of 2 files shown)


hakonanes:
I find the new `BaseSignal.map()` method very nice, so thank you to those who improved it (via implementation and review)! Especially the use of `dask.array.store()`, which we use ourselves in `kikuchipy`.

I've decided to replace part of the EBSD pattern iteration code using `dask` in `kikuchipy` with `BaseSignal.map()`, but we need the progressbar to maintain the existing feedback to users during pattern processing. Is it possible to make a patch release in the near future so that the mentioned work can proceed?

hakonanes:
I haven't taken the preferences into account here. What is the best way to do this? I was thinking within `BaseSignal.map()`:

```python
if show_progressbar is None:
```

and pass this to `self._map_iterate(show_progressbar=show_progressbar)`.

ericpre (approved):

hakonanes:
Thank you for the quick review, @ericpre.

PR #2952, opened 2022-06-03
Fix triggering event range widget

### LINKED ISSUE #2950: SpanROI broken?
#### Bug description
When interactively slicing a signal using `SpanROI`, the result does not seem to update. This includes the resulting signal and the ROI coordinates. 

#### Code to reproduce
Here are some steps to reproduce the behavior. First, run the following code:

```python
import hyperspy.api as hs
import numpy as np

s = hs.signals.Signal1D(np.random.rand(1024))
s.plot()
roi = hs.roi.SpanROI()
s_span = roi.interactive(s)
s_span.plot()
```
Move the widget in the signal plot. No updates???

#### Expected behavior
The coordinates and the result should update.

#### Python environment:
I've tested this in:
 - HyperSpy version 1.7.1.dev0 (RELEASE_next_minor branch)
 - Python version 3.10.2
And also:
 - HyperSpy version 1.7.0 (conda-forge)
 - Python version 3.10.4

#### Additional context
It's not the first time I think I've found a bug and it turns out I did something wrong! I've tried to troubleshoot this one as best as possible, but I only have reduced time. Thank you in advance for trying this on your machine and letting me know how it went!

ericpre:
Fix #2950 and some errors when plotting or updating a `Signal1D` plot with an empty slice of data.

### Progress of the PR
- [x] Trigger event when widget is changed (910e854)
- [x] avoid error with calculating limits in autoscaling code of `Signal1D` which occurs when plotting an empty slice of data
- [x] improve coverage 
- [x] add tests,
- [x] ready for review.

### DIFF
--- a/hyperspy/drawing/signal1d.py
+++ b/hyperspy/drawing/signal1d.py
@@ -177,13 +177,20 @@ class Signal1DFigure(BlittedFigure):
         self.ax.set_title(self.title)
         x_axis_upper_lims = []
         x_axis_lower_lims = []
+
         for line in self.ax_lines:
             line.plot(data_function_kwargs=data_function_kwargs, **kwargs)
-            x_axis_lower_lims.append(line.axis.axis[0])
-            x_axis_upper_lims.append(line.axis.axis[-1])
+            if len(line.axis.axis) > 1:
+                x_axis_lower_lims.append(line.axis.axis[0])
+                x_axis_upper_lims.append(line.axis.axis[-1])
+
         for marker in self.ax_markers:
             marker.plot(render_figure=False)
-        plt.xlim(np.min(x_axis_lower_lims), np.max(x_axis_upper_lims))
+
+        plt.xlim(min(x_axis_lower_lims, default=None),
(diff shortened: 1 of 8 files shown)


hyperspy/drawing/signal1d.py
```diff
@@ -484,28 +491,32 @@ def update(self, force_replot=False, render_figure=True,
...
+                if self._plot_imag:
+                    # Add real plot
+                    yreal = self._get_data(real_part=True)
+                    clipped_yreal = yreal[y1:y2]
+                    with ignore_warning(category=RuntimeWarning):
+                        # In case of "All-NaN slices"
+                        y_min = min(y_min, np.nanmin(clipped_yreal))
+                        y_max = max(y_max, np.nanmin(clipped_yreal))
+
             if y_min == y_max:
```
francisco-dlp, line 524: With the addition of the conditional execution of code above, it can happen that the execution reaches this line without having defined ``y_min`` and ``y_max``.

francisco-dlp (approved):

PR #2958, opened 2022-06-15
Add plot display range

### LINKED ISSUE #2899: Hyperspy read 0s for EMD EDX spectrums
Hi, I tried to load the EDX spectrums in the format of emd with Hyperspy. It reads the ADF images and the spectrums. ADF images look fine, but EDX spectrums are mostly zeros. I opened the same dataset with Velox and h5py and they seemed to be fine.

ericpre:
### Description of the change
Improve setting x limits when plotting EDS signal, which can happen in cases, as the one described in #2899. Since this is not the first time that this question is asked and this is simple issue, it would be good to make it more convenient.

### Progress of the PR
- [x] Display EDS spectrum from 0.1 keV,
- [x] add tests,
- [x] ready for review.

From the example given in #2899
```python
import hyperspy.api as hs

# For Signal1D
s = hs.load('2030 8.00 Mx 0001.emd')[-1]
s.sum().plot()

# For Signal2D
ima = hs.datasets.example_signals.object_hologram()
ima.plot(display_range=(150, 400, 200, 450))
```
Before this PR:
image

With this PR:
image


francisco-dlp:
I am not sure if it is a good idea to adjust the x-limit to 0.1 keV automatically—this hides a feature of the data from the user, what can cause other issues. Wouldn't it be better to simply document this particularity of EDX datasets and suggest to crop the data using ``s.isig[0.1:]`` as a cure?

francisco-dlp:
Also, this looks more like an "enhancement" than a bug-fix to me. Do you agree @ericpre?

hyperspy/_signals/eds.py
```diff
@@ -926,9 +927,15 @@ def plot(self,
         set_elements, add_elements, estimate_integration_windows,
         get_lines_intensity, estimate_background_windows
         """
+        if display_range is None:
+            axis = self.axes_manager[-1].axis
+            if len(axis) > 1 and axis[0] < 0.1:
```
francisco-dlp, line 932: This assumes that the units are keV.

By the way, is this documented?

ericpre:
> I am not sure if it is a good idea to adjust the x-limit to 0.1 keV automatically—this hides a feature of the data from the user, what can cause other issues. Wouldn't it be better to simply document this particularity of EDX datasets and suggest to crop the data using `s.isig[0.1:]` as a cure?

I was torn on this for a while and was thinking the same, however, I end up convincing myself that there is a use case to have this:
- the issue with high intensity of zero energy peak in EDS spectrum is mentioned once in a while by users. As you said, this can be work around by using `isig`, but it come at the cost of making a copy of the data, which is not always an option. Assuming that 99% of the user aren't interested in the zero energy peak, it is useful to have it as default. The other 1% would be advanced EDS user and spend effort looking at the data can afford changing the default setting to their preference.
- for `Signall2D`, I find it useful to be able to specify the range of displayed data without having to make a copy. An alternative to this would be use matplotlib differently, but there is currently a bug with the scalebar and it is more verbose.

Overall, because it is generic and the logic is simple (~I can't think of any issue that it could cause and I don't expect any because it is simple logic~, EDIT: clearly this statement didn't hold very long!! :D See https://github.com/hyperspy/hyperspy/pull/2958#discussion_r898827390), I came to the conclusion that this is a good compromise. The EDS plotting uses specific default values, which is not documented correctly.

> Also, this looks more like an "enhancement" than a bug-fix to me. Do you agree @ericpre?

Yes, it is mostly an enhancement, on the edge being a bug! ;)

hyperspy/_signals/eds.py
ericpre: Yes, you are right!

francisco-dlp:
``isig`` doesn't make a copy of the data, just a view, so it shouldn't be a problem to use it/

The issue I see is that the offending channel is also an issue for decomposition and curve fitting. By taking it out of the view, we hide the issue, making it harder for the users to understand what's wrong in their analysis.

PR #2964, opened 2022-06-17
Fix masked norm poisson

francisco-dlp:
### Description of the change

Fixes two `decomposition` bugs:

* When `normalize_poissonian_noise` is `True` and `signal_mask` is not `None`, the poisson noise normalization is actually not applied (!).
* For signals with signals dimensions higher than one, masking the signal dimension was raising an error.

### Progress of the PR
- [x] add tests,
- [x] ready for review.

One can verify that executing the code below with `True` -> `False` gives the same result with the unfixed code, but not with this fix.

```python
import hyperspy.api as hs
import numpy as np

mask = s._get_signal_signal(dtype="bool")
nav_mask = s._get_navigation_signal(dtype="bool")

mask.data[:1] = True
nav_mask.data[:1] = True

s.decomposition(True, navigation_mask=nav_mask.T, signal_mask=mask, )
```

### DIFF
--- a/upcoming_changes/2964.bugfix.rst
+++ b/upcoming_changes/2964.bugfix.rst
@@ -1,3 +1,3 @@
 Fix two bugs in :py:meth:`~.learn.mva.MVA.decomposition`:
-* The poisson noise normalization was not applied when giving a `signal_mask`
-* An error was raised when applying a `signal_mask` on a signal with signal dimension larger than 1.
+ * The poisson noise normalization was not applied when giving a `signal_mask`
+ * An error was raised when applying a ``signal_mask`` on a signal with signal dimension larger than 1.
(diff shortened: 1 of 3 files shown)


upcoming_changes/2964.bugfix.rst
```diff
@@ -0,0 +1,3 @@
+Fix two bugs in :py:meth:`~.learn.mva.MVA.decomposition`:
+* The poisson noise normalization was not applied when giving a `signal_mask`
+* An error was raised when applying a `signal_mask` on a signal with signal dimension larger than 1.
```
ericpre: Indentation needed to get the list to work?
```suggestion
 * The poisson noise normalization was not applied when giving a `signal_mask`
 * An error was raised when applying a ``signal_mask`` on a signal with signal dimension larger than 1.
```

Issue #2966, opened 2022-06-18
Fix errors reported by lgtm.com


jlaehne:
[![Language grade: Python](https://img.shields.io/lgtm/grade/python/g/hyperspy/hyperspy.svg?logo=lgtm&logoWidth=18)](https://lgtm.com/projects/g/hyperspy/hyperspy/context:python) [![Total alerts](https://img.shields.io/lgtm/alerts/g/hyperspy/hyperspy.svg?logo=lgtm&logoWidth=18)](https://lgtm.com/projects/g/hyperspy/hyperspy/alerts/)

The code checker `lgtm` has a number of errors, warnings and recommendations for HyperSpy. We should use that to consolidate our code:

https://lgtm.com/projects/g/hyperspy/hyperspy

jlaehne:
lgtm.com definitely includes some false-positives, but also identifies some real issues or chances for better code consistency.

For LumiSpy, I have now activated PR commenting by lgtm - takes a fairly long time (even though the project is still small), but can help to find some code inconsistencies that might be missed by manual review.

jlaehne:
I reported three types of false positives to LGTM that affect us:

- undefined exported names imported by function: https://github.com/github/codeql/issues/9642
- unused import to build API: https://github.com/github/codeql/issues/9651
- unused import in `try` statement: https://github.com/github/codeql/issues/9652

PR #2967, opened 2022-06-18
some lgtm code consolidation

jlaehne:
I started to fix some of the issues identified by [lgtm](https://lgtm.com/projects/g/hyperspy/hyperspy), see also #2966 

## Progress of the PR:

- [x] changes implemented
- [x] ready for review

### DIFF
--- a/hyperspy/_signals/lazy.py
+++ b/hyperspy/_signals/lazy.py
@@ -1056,8 +1056,6 @@ class LazySignal(BaseSignal):
                     if signal_mask is None
                     else to_array(signal_mask, chunks=sig_chunks)
                 )
-                ndim = self.axes_manager.navigation_dimension
-                sdim = self.axes_manager.signal_dimension
                 bH, aG = da.compute(
                     data.sum(axis=tuple(range(ndim))),
                     data.sum(axis=tuple(range(ndim, ndim + sdim))),
--- a/hyperspy/_signals/lazy.py
+++ b/hyperspy/_signals/lazy.py
@@ -24,6 +24,7 @@ import warnings
 import numpy as np
 import dask.array as da
 import dask
+from dask.diagnostics import ProgressBar
 from itertools import product
 from packaging.version import Version
 
@@ -50,9 +51,10 @@ lazyerror = NotImplementedError('This method is not available in lazy signals')
 
 
 try:
+    from dask.widgets import TEMPLATE_PATHS
     templates_path = os.path.join(os.path.dirname(os.path.abspath(__file__)),
                                   "..", "misc", "dask_widgets")
-    dask.widgets.TEMPLATE_PATHS.append(templates_path)
+    TEMPLATE_PATHS.append(templates_path)
 except ModuleNotFoundError:
     _logger.info("Dask widgets not loaded (dask >=2021.11.1 is required)")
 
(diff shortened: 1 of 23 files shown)


jlaehne:
I won't find more time to continue for the moment

ericpre:
ok, let's put it in draft for now, then.

ericpre (commented):
It would be good to add the corresponding documentation to the dev guide to help contributors finding relevant documentation on lgtm and briefly explain what are the comments used by lgtm, etc.

jlaehne:
Added some more corrections based on the checks by lgtm.com

Fixes a number of real issues and does some streamlining, such as removing duplicate imports.

I'm loosing the overview which issues I have addressed and which not.

Also, I think if adding more changes, it will get hard to review.

Therefore, I submit this for review and propose to continue in a separate PR once lgtm.com has reevaluated the HyperSpy.

hyperspy/_components/eels_cl_edge.py
```diff
@@ -211,7 +211,7 @@ def _calculate_effective_angle(self):
                 self.GOS.onset_energy,
                 self.convergence_angle,
                 self.collection_angle)
-        except BaseException:
```
ericpre: If my understanding is correct, the difference between the two is that `BaseException` will catch slightly more exception (`SystemExit`, `KeyboardInterrupt`, `GeneratorExit`) than `Exception` (from https://docs.python.org/3/library/exceptions.html#exception-hierarchy), but I don't know the reason to favour the use of one of the two.
@francisco-dlp, if I remember correctly, I have seen you using `BaseException`, do you have a view on this?

hyperspy/_signals/lazy.py
```diff
@@ -23,7 +23,8 @@
 
 import numpy as np
 import dask.array as da
-import dask
+import dask.threaded as dt
+from dask import __version__
```
ericpre, line 27: Is it good? This will add the `__version__` attribute to this module with its value set to the dask version!

hyperspy/signal.py
```diff
@@ -39,12 +39,12 @@
 import traits.api as t
 from tlz import concat
 
-import hyperspy
+import hyperspy.io as hypio
+import hyperspy.misc.array_tools as hypat
```
ericpre, line 43: This is subjective but I am not sure if this make the code better in a sense, as in more readable.

ericpre (commented):
This is useful but my impression is that we need to be careful before enforcing a specific convention.
These conventions need to be documented in the dev guide, so that we can use it as reference and it would avoid us having to figure out what are the conventions to follow - and their justification.
Is it expected that lgtm analysis takes so long? It seems to be in the 30-45 min range?

Regarding making the history more readable and the change/review easier to follow, would it useful to have commit message more explicit, as for example: "lgtm improvement: change BaseException to Exception", etc. and one commit per type of changes?

hyperspy/_signals/lazy.py
jlaehne: It avoids importing the whole of dask, just to have the version available. Any better idea?

hyperspy/signal.py
jlaehne: The file imports a dozen different modules from hyperspy independently, but then the whole of hyperspy again just for these two submodules.

hyperspy/_components/eels_cl_edge.py
jlaehne: According to lgtm, that is the point. `BaseException` is very general: https://lgtm.com/rules/6780080/

jlaehne:
> This is useful but my impression is that we need to be careful before enforcing a specific convention. These conventions need to be documented in the dev guide, so that we can use it as reference and it would avoid us having to figure out what are the conventions to follow - and their justification.

Many issues like unreachable code or duplicate assignments of the same variable are not necessarily about conventions but show potential problems with the code. Sometimes, they also help to find a different issue with the code. Others like unused imports are not critical, but save overhead.
But indeed, some other things are more about conventions and there are also some false-positives, like the one I reported here: https://github.com/github/codeql/issues/9642#issuecomment-1162774212
So we might have to deactivate some types of reports.

LGTM-PR-reports just list newly introduced or resolved issues. I think that would be very helpful.

> Is it expected that lgtm analysis takes so long? It seems to be in the 30-45 min range?

I don't know. It was just an observation.

> Regarding making the history more readable and the change/review easier to follow, would it useful to have commit message more explicit, as for example: "lgtm improvement: change BaseException to Exception", etc. and one commit per type of changes?

I started out going by the order reported on the site and did not want to end up with too many commits, but when I realized that it could get messy I did indeed start grouping things like with the BaseException, also because that is a disputable convention. Too sort the other ones in grouped commits may be not so straightforward now.

jlaehne:
replaced by #2977

PR #2968, opened 2022-06-21
Add type-hinting to commonly-used methods

thomasaarholt:
### Description of the change
1. Added "eager" imports under the `TYPE_CHECKING` filter (which is only `True` during e.g. pylance/mypy type checking) so that autocomplete works for the lazy modules imported through `hyperspy.api` and `hyperspy.api_nogui`.
2. Added type-hints for several commonly-used methods. Particularly useful when coding in VSCode, since the pylance engine understands e.g.:

```python
from hyperspy._signals.eels import EELSSpectrum

s = EELSSpectrum([1, 2, 3])
ax = s.axes_manager[0]
# ax.high<TAB> autocompletes to high_value

m = s.create_model()
a = m[0]
# a.na<TAB> autocompletes to name
```
Using the VSCode pylance Type Checking Mode setting "basic" is very useful, as it provides in-editor messages showing mypy errors. It also shows a lot of possible improvements, for instance where we are referencing child-class methods in parent-classes. For instance MVATools referring to entries of `axes_manager` even though this isn't an attribute of the 
MVATools class.

### Progress of the PR
- [ ] add tests,
- [ ] ready for review.

### DIFF
--- a/hyperspy/__init__.py
+++ b/hyperspy/__init__.py
@@ -18,6 +18,7 @@
 
 import importlib
 import logging
+from typing import TYPE_CHECKING
 from hyperspy import docstrings
 
 _logger = logging.getLogger(__name__)
@@ -40,6 +41,9 @@ More details in the :mod:`~hyperspy.api` docstring.
 
 """ % docstrings.START_HSPY
 
+if TYPE_CHECKING:
+    from hyperspy import api
+    from hyperspy.Release import version as __version__
 
 __all__ = [
     "api",
@@ -59,7 +63,7 @@ _import_mapping = {
 def __getattr__(name):
     if name in __all__:
         if name in _import_mapping.keys():
-            import_path = 'hyperspy' + _import_mapping.get(name)
+            import_path = 'hyperspy' + _import_mapping[name]
             return getattr(importlib.import_module(import_path), name)
         else:  # pragma: no cover
             # We can't get this block covered in the test suite because it is
--- a/hyperspy/_signals/eds_sem.py
+++ b/hyperspy/_signals/eds_sem.py
@@ -22,6 +22,7 @@ import traits.api as t
 
 from hyperspy._signals.eds import (EDSSpectrum, LazyEDSSpectrum)
 from hyperspy.defaults_parser import preferences
+from hyperspy.models.edstemmodel import EDSTEMModel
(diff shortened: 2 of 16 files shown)


thomasaarholt:
@ericpre might be interested in this issue too, regarding VSCode/pylance/pyright not supporting lazy loaded modules: https://github.com/microsoft/pyright/issues/3595

thomasaarholt:
There are a few places where we dynamically add entries to `__all__`, like in hyperspy/signals.py and `hyperspy/components1d.py` (and 2d). VSCode won't be able to autocomplete in those cases, but I think we just have to live with that. These don't work in jupyter notebook either (before running the code):
```python
import hyperspy.components1d.<TAB> # yields no suggestions in jupyterlab or vscode
import hyperspy.signals.<TAB> # same
```

ericpre:
> There are a few places where we dynamically add entries to `__all__`, like in hyperspy/signals.py and `hyperspy/components1d.py` (and 2d). VSCode won't be able to autocomplete in those cases, but I think we just have to live with that. These don't work in jupyter notebook either (before running the code):
> 
> ```python
> import hyperspy.components1d.<TAB> # yields no suggestions in jupyterlab or vscode
> import hyperspy.signals.<TAB> # same
> ```

This works fine with jupyter, a simple way to test it is to run the binder: [Live demos (Binder)](https://mybinder.org/v2/gh/hyperspy/hyperspy-demos/main)

thomasaarholt:
Odd. It works in the binder, but I just tried making a new conda env with hyperspy 1.7.1 and jupyter notebook, and there
`import hyperspy.comp<TAB>` suggests `component`/`components1d`/`components2d`, but `import hyperspy.components1d.<TAB>` doesn't do anything.

ericpre:
> Odd. It works in the binder, but I just tried making a new conda env with hyperspy 1.7.1 and jupyter notebook, and there `import hyperspy.comp<TAB>` suggests `component`/`components1d`/`components2d`, but `import hyperspy.components1d.<TAB>` doesn't do anything.

I have just tried on binder and locally and it works, did you wait enough (2-3s) that the module load after pressing `<TAB>`?

ericpre:
Should this PR be closed in favour of #2986?

PR #2971, opened 2022-06-29
LazySignal.compute will now pass kwargs to dask.array.Array.compute

magnunor:
Currently, `kwargs` are not passed from `s.compute` to the dask `compute` function which is done on the data.

There are several useful parameters like `scheduler='processes'` and `num_workers=4`, which is useful when working with large datasets.

So this pull request simply passes `kwargs` in `s.compute` to `s.data.compute()`. It also improves the docstring for `compute`, and adds a test.

### Progress of the PR
- [x] Change implemented
- [x] update docstring
- [x] add test
- [x] ready for review.

```python
import dask.array as da
data = da.zeros((100, 100, 100), chunks=(10, 20, 20))
s.compute(scheduler="processes", num_workers=2)
```

### DIFF
--- a/hyperspy/_signals/lazy.py
+++ b/hyperspy/_signals/lazy.py
@@ -243,14 +243,45 @@ class LazySignal(BaseSignal):
         Parameters
         ----------
         close_file : bool, default False
-            If True, attemp to close the file associated with the dask
+            If True, attempt to close the file associated with the dask
             array data if any. Note that closing the file will make all other
             associated lazy signals inoperative.
         %s
+        **kwargs : dict
+            Any other keyword arguments for :py:func:`dask.array.Array.compute`.
+            For example `scheduler` or `num_workers`.
 
         Returns
         -------
         None
+
+        Notes
+        -----
+        For alternative ways to set the compute settings see
+        https://docs.dask.org/en/stable/scheduling.html#configuration
+
+        Examples
+        --------
+        >>> import dask.array as da
+        >>> data = da.zeros((100, 100, 100), chunks=(10, 20, 20))
+        >>> s = hs.signals.Signal2D(data).as_lazy()
+
+        With default parameters
+
+        >>> s1 = s.deepcopy()
+        >>> s1.compute()
+
+        Using 2 workers, which can reduce the memory usage (depending on
+        the data and your computer hardware). Note that `num_workers` only
+        work for the 'threads' and 'processes' `scheduler`.
+
+        >>> s2 = s.deepcopy()
+        >>> s2.compute(num_workers=2)
+
+        Using a single threaded scheduler, which is useful for debugging
+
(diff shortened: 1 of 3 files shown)


hyperspy/_signals/lazy.py
```diff
@@ -241,19 +241,38 @@ def compute(self, close_file=False, show_progressbar=None, **kwargs):
...
+
+        With default parameters
+
+        >>> s1 = s.deepcopy()
+        >>> s1.compute()
+
+        Using processes scheduler and 2 workers
+
+        >>> s2 = s.deepcopy()
+        >>> s2.compute(scheduler="processes", num_workers=2)
```
ericpre: While you are at it, it is worth adding a link to https://docs.dask.org/en/stable/scheduling.html#configuration in a `notes` or `see also` section to inform users of alternative way to set compute settings?

hyperspy/_signals/lazy.py
```diff
@@ -241,19 +241,38 @@ def compute(self, close_file=False, show_progressbar=None, **kwargs):
             array data if any. Note that closing the file will make all other
             associated lazy signals inoperative.
         %s
+        **kwargs : dict
+            Any other keyword arguments for :py:func:`dask.array.Array.compute`.
+            For example scheduler or num_workers.
```
jlaehne:
```suggestion
            For example `scheduler` or `num_workers`.
```

jlaehne (approved):

hyperspy/_signals/lazy.py
magnunor: Done!

magnunor:
Sorry about the slow response! I rebased this onto the newest `RELEASE_next_minor`, and fixed the docstring improvements suggest by @jlaehne

Or should this rather be sent to `RELEASE_next_major`?

hyperspy/_signals/lazy.py
```diff
@@ -236,23 +236,42 @@ def compute(self, close_file=False, show_progressbar=None, **kwargs):
...
+
         """
         if "progressbar" in kwargs:
             warnings.warn(
                 "The `progressbar` keyword is deprecated and will be removed "
                 "in HyperSpy 2.0. Use `show_progressbar` instead.",
                 VisibleDeprecationWarning,
             )
-            show_progressbar = kwargs["progressbar"]
+            show_progressbar = kwargs.pop("progressbar")
```
ericpre: In the `RELEASE_next_major`, it can be removed.

ericpre (commented):
Yes, I think it will more simple if it goes directly to `RELEASE_next_major`.

hyperspy/_signals/lazy.py
magnunor: Done! Added it to `Notes`

magnunor:
All done!

Issue #2970, opened 2022-06-22
Should `hs.load` always return a list?


ericpre:
One common inconvenience/confusing that beginners experience with hyperspy is issue like https://github.com/hyperspy/hyperspy/issues/2959, because `hs.load` can return a list of signals or a signal... this is very simple to explain but this could be avoided if `hs.load` would always return a list, even when the file to load contains only a single dataset.

For example, we change the API to have a syntax very similar to what matplotlib does with `plot`

```python
import matplotlib.pyplot as plt

fig, ax = plt.subplots()
lines = ax.plot([0, 1, 2])

print(lines)
# lines is a list
# [<matplotlib.lines.Line2D at 0x252af81fa90>]

line, =  ax.plot([3, 4, 5])
print(line)
# line is not a list
# <matplotlib.lines.Line2D at 0x252afa14640>
```

jlaehne:
I'm not sure if would always return a list. In most cases I load only single signals and do not want to necessarily add a list indexation to access the signal.

I think your idea of an improved documentation is good ... and people need to learn reading docs and know what is in their datasets ;-)

PR #2975, opened 2022-07-18
Improve documentation `hs.load` returns list

ericpre:
Add a note in the user guide to explain that when a file contains several dataset, `hs.load` will return a list of signals instead of a single signal and therefore that the list indexation needs to be used.
https://github.com/hyperspy/hyperspy/issues/2970

### Progress of the PR
- [x] improve user guide,
- [x] add tests,
- [x] ready for review.

### DIFF
--- a/doc/user_guide/getting_started.rst
+++ b/doc/user_guide/getting_started.rst
@@ -164,7 +164,14 @@ Once HyperSpy is running, to load from a supported file format (see
 
 .. code-block:: python
 
-    >>> s = hs.load("filename")
+    >>> s = hs.load("filename.ext")
+
+The used IO-plugin will be inferred from the file extension. If you want to force
+the use of a specific IO-plugin, you can provide the ``reader`` attribute:
+
+.. code-block:: python
+
+    >>> s = hs.load("spam.ham", reader="hspy")
 
 .. note::
 
--- a/doc/user_guide/getting_started.rst
+++ b/doc/user_guide/getting_started.rst
@@ -168,8 +168,8 @@ Once HyperSpy is running, to load from a supported file format (see
 
 .. note::
 
-   When the file contains several datasets, the :py:func:`~.io.load`  function
-   will return a list of hyperspy signal, instead of a single hyperspy signal. 
+   When the file contains several datasets, the :py:func:`~.io.load` function
+   will return a list of HyperSpy signals, instead of a single HyperSpy signal. 
    Each signal can then be accessed using list indexation.
 
    .. code-block:: python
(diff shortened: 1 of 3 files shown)


doc/user_guide/getting_started.rst
```diff
@@ -166,6 +166,28 @@ Once HyperSpy is running, to load from a supported file format (see
 
     >>> s = hs.load("filename")
 
+.. note::
+
+   When the file contains several datasets, the :py:func:`~.io.load`  function
```
jlaehne:
```suggestion
   When the file contains several datasets, the :py:func:`~.io.load` function
```

doc/user_guide/getting_started.rst
```diff
@@ -166,6 +166,28 @@ Once HyperSpy is running, to load from a supported file format (see
...
+   will return a list of hyperspy signal, instead of a single hyperspy signal.
```
jlaehne:
```suggestion
   will return a list of HyperSpy signals, instead of a single HyperSpy signal. 
```

upcoming_changes/2975.doc.rst
```diff
@@ -0,0 +1 @@
+Add a note in the user guide to explain that when a file contains several dataset, :py:func:`~.io.load` returns a list of signals instead of a single signal and that list indexation can be used to access single signal.
```
jlaehne:
```suggestion
Add a note in the user guide to explain that when a file contains several datasets, :py:func:`~.io.load` returns a list of signals instead of a single signal and that list indexation can be used to access a single signal.
```

doc/user_guide/getting_started.rst
```diff
@@ -166,6 +166,28 @@ Once HyperSpy is running, to load from a supported file format (see
...
+
+   .. code-block:: python
+
+      >>> s = hs.load("spameggsandham.hspy")
+      >>> s
+      [<Signal1D, title: spam, dimensions: (32,32|1024)>,
+       <Signal1D, title: eggs, dimensions: (32,32|1024)>,
+       <Signal1D, title: ham, dimensions: (32,32|1024)>]
+
+   Using indexation to assess the first signal (index 0):
```
jlaehne:
```suggestion
   Using indexation to access the first signal (index 0):
```

jlaehne:
Is mentioned in the io section, but I guess it does not hurt to mention it here already.

I would also add here that the file type is inferred from the ending, but that the `reader` can be specified to force a certain format.

ericpre:
> I would also add here that the file type is inferred from the ending, but that the `reader` can be specified to force a certain format.

I agree that this need revisiting particularly after the split, where we should be able to simplify/consolidate this part of the documentation. As it was asked recently about it on gitter, I thought that I would do it right now before we forget about it!

jlaehne:
> > I would also add here that the file type is inferred from the ending, but that the `reader` can be specified to force a certain format.
> 
> I agree that this need revisiting particularly after the split, where we should be able to simplify/consolidate this part of the documentation. As it was asked recently about it on gitter, I thought that I would do it right now before we forget about it!

I just did another PR to this one (https://github.com/ericpre/hyperspy/pull/21), before forgetting about the `reader` statement. Still, it is true that we need to also mention RosettaSciIO here at some point.

PR #2977, opened 2022-07-20
Revised lgtm code consolidation

jlaehne:
### Description of the change
Replaces #2967 leaving different types of changes to different commits for easier review. Addresses comments raised in #2967 and leaves out disputable changes. IO-related changes are in a separate PR: https://github.com/francisco-dlp/hyperspy/pull/56.

Several changes do correct errors present in the current code.
Otherwise, duplicate imports, unnecessary statements are removed, ...

These changes should not need a specific documentation in the developer guide, which could be added in a separate PR. Once this PR is merged and HyperSpy reevaluated by `LGTM <https://lgtm.com/projects/g/hyperspy/hyperspy>`_, further issues could be addressed.

### Progress of the PR
- [x] ready for review.

### DIFF
--- a/hyperspy/_components/expression.py
+++ b/hyperspy/_components/expression.py
@@ -18,6 +18,7 @@
 
 from functools import wraps
 import numpy as np
+import sympy
 from sympy.utilities.lambdify import lambdify
 
 import warnings
@@ -236,7 +237,6 @@ class Expression(Component):
         possible.
         Useful to recompile the function and gradient with a different module.
         """
-        import sympy
         try:  # Expression is just a constant
             float(self._str_expression)
         except ValueError:
--- a/hyperspy/_components/expression.py
+++ b/hyperspy/_components/expression.py
@@ -19,7 +19,6 @@
 from functools import wraps
 import numpy as np
 import sympy
-from sympy.utilities.lambdify import lambdify
 
 import warnings
 
@@ -287,7 +286,7 @@ class Expression(Component):
         parameters.sort(key=lambda x: x.name)  # to have a reliable order
         # Create compiled function
         variables = [x, y] if self._is2D else [x]
-        self._f = lambdify(variables + parameters, eval_expr,
+        self._f = sympy.utilities.lambdify(variables + parameters, eval_expr,
                            modules=module, dummify=False)
 
         if self._is2D:
@@ -307,7 +306,7 @@ class Expression(Component):
                 for p in parameters:
                     grad_expr = sympy.diff(eval_expr, p)
                     name = self._rename_pars.get(p.name, p.name)
-                    f_grad = lambdify(variables + parameters,
+                    f_grad = sympy.utilities.lambdify(variables + parameters,
                                       grad_expr.evalf(),
                                       modules=module,
                                       dummify=False)
@@ -397,7 +396,7 @@ class Expression(Component):
             element_names = \
                 set([str(p) for p in element.free_symbols]) - set(variables)
             free_pseudo_components[para.name] = {
-                'function': lambdify(variables + tuple(element_names),
+                'function': sympy.utilities.lambdify(variables + tuple(element_names),
                                      element, modules=self._module),
                 'parameters': [getattr(self, self._rename_pars.get(e, e))
                                for e in element_names]
@@ -408,7 +407,7 @@ class Expression(Component):
             set(variables)
 
         fixed_pseudo_components = {
-            'function': lambdify(variables + tuple(element_names),
+            'function': sympy.utilities.lambdify(variables + tuple(element_names),
                                  remaining_elements, modules=self._module),
             'parameters': [getattr(self, self._rename_pars.get(e, e))
                            for e in element_names]
--- a/hyperspy/_components/expression.py
+++ b/hyperspy/_components/expression.py
@@ -18,7 +18,6 @@
 
 from functools import wraps
 import numpy as np
-import sympy
 from sympy.utilities.lambdify import lambdify
 
 import warnings
--- a/hyperspy/_signals/lazy.py
+++ b/hyperspy/_signals/lazy.py
@@ -23,7 +23,8 @@ import warnings
 
 import numpy as np
 import dask.array as da
-import dask
+import dask.threaded as dt
+import dask.__version__ as daskversion
 from dask.diagnostics import ProgressBar
 from itertools import product
 from packaging.version import Version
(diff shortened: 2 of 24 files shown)


jlaehne:
Note that LGTM will be discontinued at the end of the year in favor of GitHub Code Scanning: https://github.blog/2022-08-15-the-next-step-for-lgtm-com-github-code-scanning/

So it makes no sense to update anything in the user guide accordingly.

If we activate automatic code scanning for PRs now, there should be a PR from the LGTM team in October to transfer it to a GitHub Code Scanning Action. Otherwise, we can of course set up GitHub code scanning directly at some point.

hyperspy/_signals/lazy.py
```diff
@@ -23,7 +23,8 @@
 
 import numpy as np
 import dask.array as da
-import dask
+import dask.threaded as dt
+from dask import __version__ as daskversion
```
ericpre: I always find this subjective, is there any reason/recommendation/guidance to do one or another? If we enforce one approach over another, we need to have documentation and ideally with a good reason. :)

ericpre:
As it is possible to setup "Github Code Scanning" right now, would it be sensible to do that?
- https://github.blog/2020-09-30-code-scanning-is-now-available
- https://github.blog/2020-09-30-code-scanning-is-now-available

Would it be worth doing this on `RELEASE_next_major` as this will be one of the next release and it could save us to sort out potential merge conflict?

jlaehne:
How about setting it up for RosettaSciIO for the moment, test it there for a few weeks and then implement it here in HyperSpy?

In the meantime, could you review the changes in this PR?

ericpre:
For the review, it would be good to use the analysis tool to understand how it should be fixed. Did you use https://lgtm.com/projects/g/hyperspy/hyperspy to find what needs to be changed?

hyperspy/_signals/lazy.py
```diff
@@ -23,7 +23,8 @@
 
 import numpy as np
 import dask.array as da
-import dask
+import dask.threaded as dt
```
ericpre: Why is it better? Alias `dt` is also used for `datetime` which can make the code less readable, but this is subjective...

hyperspy/misc/ipython_tools.py
```diff
@@ -30,16 +30,17 @@ def get_ipython():
     """
     if is_it_running_from_ipython is False:
         return None
-    import IPython
-    ipy_version = Version(IPython.__version__)
+    from IPython import __version__ as ipythonversion
```
ericpre: Need to discuss - similar to `dask.__version__` import.

hyperspy/signal.py
```diff
@@ -2743,9 +2743,9 @@ def plot(self, navigator="auto", axes_manager=None, plot_markers=True,
             self._plot.close()
         if 'power_spectrum' in kwargs:
             if not np.issubdtype(self.data.dtype, np.complexfloating):
+                del kwargs['power_spectrum']
```
ericpre: This can be removed? As it raises an error, it shouldn't continue and there is no need to delete it.

hyperspy/signal.py
```diff
@@ -39,12 +39,13 @@
 import traits.api as t
 from tlz import concat
 
-import hyperspy
 from hyperspy.axes import AxesManager
 from hyperspy.api_nogui import _ureg
+from hyperspy.misc.array_tools import rebin as hsrebin
```
ericpre: I am not sure about the reason for these changes: the commit message says "remove import of whole hyperspy needed for only two lines" but the import are lazy, so they shouldn't be any performance concerns.

Personally, I find using non-standard aliases inconvenient when trying to know where a function comes from, when reading the code, for example and I find reading `hyperspy.misc.array_tools.rebin` more explicit/clear than `hsrebin`, particularly when the code is at line ~3000 and the import at the beginning of the file.

hyperspy/_components/expression.py
```diff
@@ -18,7 +18,6 @@
 
 from functools import wraps
 import numpy as np
-import sympy
```
jlaehne, line 21: wrong import removed, fixed in https://github.com/hyperspy/hyperspy/pull/2977/commits/c6861fbf54c4d1457f8085a40773ab41a7ab383f

jlaehne:
> For the review, it would be good to use the analysis tool to understand how it should be fixed. Did you use https://lgtm.com/projects/g/hyperspy/hyperspy to find what needs to be changed?

Yes, and I tried to clarify in the commit messages what is being fixed, so that if you review commit by commit, it should be fairly clear what is being done.

ericpre (commented):
This looks good except for the import changes, which needs discussion or more justification - why would one be better than another?

> > For the review, it would be good to use the analysis tool to understand how it should be fixed. Did you use https://lgtm.com/projects/g/hyperspy/hyperspy to find what needs to be changed?
> 
> Yes, and I tried to clarify in the commit messages what is being fixed, so that if you review commit by commit, it should be fairly clear what is being done.

The commit says very briefly what it is about but they don't provide context or reasoning on the changes, this is why I am asking for the lgtm analysis to try to understand what is actually the issue and how it should be fixed. For some cases, this is obvious but not always, typically the import changes! :)

hyperspy/_signals/lazy.py
jlaehne: LGTM recommends not to combine `import` and `import from` for the same module for maintainability: https://lgtm.com/rules/1818040193/

Maybe it would be better to do it the other way round and remove the `import from` instead of converting the other imports to additional `import from` statements. If you agree, I would change it that way round.

hyperspy/misc/ipython_tools.py
jlaehne: Here, `IPython` was imported only for the `version`, because all other uses have additional `import from` statements. Again, it would maybe make sense and remove all the `import from` statements by adding the module name to the calls.

hyperspy/signal.py
jlaehne: I guess you are right, before it was simply a line of code that would never be executed, so deleting it should not change anything.

hyperspy/signal.py
jlaehne: Again, everything else is imported by `import from` (and it might not always be clear from which module of hyperspy they come from), except for these two functions. However, `save` and `rebin` can not be imported without giving them new names.

Referencing any hyperspy function with their full path, will not make the code more readable either, as it would blow up the code by many long function names.

I could now:
a) just leave it as it was originally
b) give the functions more instructive new names, such as `hs_array_rebin`

jlaehne: Here I renamed the imports with more instructive names. Everywhere else, I opted for the more general import (except where `dask.array` was extensively used).

ericpre (approved):

PR #2972, opened 2022-07-05
Split IO #2

francisco-dlp:
### Description of the change

Split HyperSpy's IO plugins into a separate package RosettaSciIO. See #1978 for the related discussion.

This supersedes #2174

### Progress of the PR
- [ ] Change implemented (can be split into several points)
    - [x] Move all plugins to a separate folder, with its own setup.py, that must be installed separately.
    - [x] Move each plugin to a folder of its own
    - [x] Move all IO tests to rsciio
    - [x] Enable lazy loading of the plugins by moving the specifications to yamls files.
    - [x] Make rosettasciio fully independent of HyperSpy
        - [x] Don't use ``DictionaryTreeBrowser``. DTB was replaced by python-box
        - [x] Move ``misc.io`` code to ``rsciio``
        - [x] Move exceptions to ``rsciio``. 
        - [x] Move remaining imports to ``rsciio``Most of it is done, except for:
            - [x] ``rgb_tools``
            - [x]  ``emd`` relies on hyperspy for reading
        - [x] Make `_hierarchical` fully independent of HyperSpy
        - [x] Pass dictionary instead of signal to ``file_writer``s
            - [x] msa
            - [x] tvips
            - [x] semper
            - [x] blockfile
            - [x] emd
            - [x] panta rhei
            - [x] nexus
            - [x] image (https://github.com/francisco-dlp/hyperspy/pull/51)
            - [x] mrcz (https://github.com/francisco-dlp/hyperspy/pull/51)
            - [x] ripple (https://github.com/francisco-dlp/hyperspy/pull/51)
            - [x] nexus (https://github.com/francisco-dlp/hyperspy/pull/51)
            - [x] tiff (https://github.com/francisco-dlp/hyperspy/pull/51)
            - [x] usid (https://github.com/francisco-dlp/hyperspy/pull/51)
            - [x] zspy
            - [x] hspy
    - [x] Move User Guide IO chapter to rsciio
    - [x] Remove dependency on traits (basically for t.Undefined)
    - [ ]  Make tests independent of hyperspy (not strictly required for release)
- [ ] update docstring (if appropriate),
- [x] update user guide (if appropriate),
- [x] add an changelog entry in the `upcoming_changes` folder (see [`upcoming_changes/README.rst`](https://github.com/hyperspy/hyperspy/blob/RELEASE_next_minor/upcoming_changes/README.rst)) -- see #2979, 
- [ ] Check formatting changelog entry in the `readthedocs` doc build of this PR (link in github checks)
[...]

PR #2896, opened 2022-03-02
Add io_plugin for CEOS Panta Rhei *.prz file format

gguzzina:
*You can delete this section after you read it.*

### Description of the change
This patch implements an IO plugin for the PRZ files saved by CEOS PantaRhei, the software used to control CEOS spectrometers.
The data format itself is merely a NPZ file with a different extension. 
The file contains a numpy array with the data, and a dictionary that contains all metadata.
Most of the code handles the metadata conversion, due some different conventions between PantaRhei and hyperspy.

This patch implements both loading from and saving to PRZ files.

### Progress of the PR
- [x] Change implemented
- [x] update docstring
- [x] add an changelog
- [x] update user guide (if appropriate),
- [x] Check formatting changelog entry in the `readthedocs` doc build of this PR (link in github checks)
- [x] add tests,
- [x] ready for review.

### Minimal example of the bug fix or the new feature
```python
import hyperspy.api as hs
s = hs.load('dataset.prz')
s.align_zero_loss_peak() # Random command, merely an example
s.save('dataset_aligned.prz')
```

PR #2979, opened 2022-07-23
Merge RELEASE_next_minor into RELEASE_next_major

jlaehne:
- Merge current `RELEASE_next_minor` into `RELEASE_next_major`.
- Add missing changelog entry for #2972.

### DIFF
--- a/upcoming_changes/2975.doc.rst
+++ b/upcoming_changes/2975.doc.rst
@@ -1,1 +1,1 @@
-Add a note in the user guide to explain that when a file contains several dataset, :py:func:`~.io.load` returns a list of signals instead of a single signal and that list indexation can be used to access single signal.
+Add a note in the user guide to explain that when a file contains several datasets, :py:func:`~.io.load` returns a list of signals instead of a single signal and that list indexation can be used to access a single signal.
--- a/upcoming_changes/2975.doc.rst
+++ b/upcoming_changes/2975.doc.rst
@@ -1,0 +1,1 @@
+Add a note in the user guide to explain that when a file contains several dataset, :py:func:`~.io.load` returns a list of signals instead of a single signal and that list indexation can be used to access single signal.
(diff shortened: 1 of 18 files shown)


upcoming_changes/2975.doc.rst
```diff
@@ -0,0 +1 @@
+Add a note in the user guide to explain that when a file contains several datasets, :py:func:`~.io.load` returns a list of signals instead of a single signal and that list indexation can be used to access a single signal.
```
ericpre, line 1: Shall this changelog note be moved to `RosettaSciIO`?

jlaehne: But that part of the documentation was not moved, as far as I understand it. It is in the getting started section of the hypserspy userguide. One could modify this changelog to say that it concerns `getting_started`.

ericpre: Yes, sorry, it was not clear and I have put this comment in the wrong place! I was confused with the changelog entry for #2896, which I have noticed when checking the doc build on readthedocs. Actually, this is not related to this PR because it is already on the `RELEASE_next_major` branch. Sorry for the noise!

upcoming_changes/2972.maintenance.rst
```diff
@@ -0,0 +1 @@
+RosettaSciIO was split out of the `HyperSpy repository <https://github.com/hyperspy/hyperspy>`_ on July 23, 2022. The IO-plugins and related functions so far developed in HyperSpy were moved to the `RosettaSciIO repository <https://github.com/hyperspy/rosettasciio>`_.
```
ericpre: Would it be better to classify this note as API change?

jlaehne: I was indeed hesitating, which category to choose:
- `maintenance` is defined as concerning packaging, which I would say this one does.
- And the actual usage of the HyperSpy-API for the user does not change (they just have to install an additional package), which is why I hesitated to categorize it as `api`.

That said about my line of thought, I am open to categorize it differently.

ericpre: I suspect that API which is not in `hyperspy.api`, for example, `from hyperspy.io_plugins.msa import file_reader` is being used and in that sense it breaks the API.
Even if this changes will be transparent to the large majority of users, I think that this is useful to flag it at API changes to be on the safe side!

In any case, the 2.0 release will most likely not be as automated as with other and the changelog and the wording of the changelog will be tweaked before the release!

jlaehne: Fair argument, I changed it from `maintenance` to `api`

ericpre (approved):

PR #2981, opened 2022-07-24
Fix for dask 2022.7.1

ericpre:
Dask 2022.7.1 fixes an inconsistency between dask and numpy in the behaviour of `squeeze` in https://github.com/dask/dask/pull/9250, but it has a side effect in the `map` code, which is caught in the test suite.
@CSSFrancis, can you please have a look at this PR, since you are the main contributor for the latest change to this part of the code?

### Progress of the PR
- [x] Fix function to guess output size when using `map`,
- [x] Fix typo in tests name, which means that some were not running, 
- [n/a] add tests,
- [x] ready for review.

This was the failing test:

```python
import dask.array as da
import hyperspy.api as hs
import numpy as np


dask_array = da.zeros((10, 20, 100, 100), chunks=(5, 10, 50, 50))
s = hs.signals.Signal2D(dask_array).as_lazy()


def apply_func(data, f):
    return f(data, data)

s = s.inav[0:2, 0:2]
iter_add = hs.signals.BaseSignal([[np.add, np.add],
                                  [np.add, np.add]]).T
out = s.map(apply_func, f=iter_add, inplace=False)
np.testing.assert_array_equal(out.data, s.data)
```

### DIFF
--- a/hyperspy/signal.py
+++ b/hyperspy/signal.py
@@ -4994,7 +4994,9 @@ class BaseSignal(FancySlicing,
             testing_kwargs = {}
             for ikey, key in enumerate(arg_keys):
                 test_ind = (0,) * len(os_am.navigation_axes)
-                testing_kwargs[key] = np.squeeze(args[ikey][test_ind]).compute()
+                # For discussion on if squeeze is necessary, see
+                # https://github.com/hyperspy/hyperspy/pull/2981
+                testing_kwargs[key] = np.squeeze(args[ikey][test_ind].compute())[()]
             testing_kwargs = {**kwargs, **testing_kwargs}
             test_data = np.array(
                 old_sig.inav[(0,) * len(os_am.navigation_shape)].data.compute()
--- a/upcoming_changes/2981.bugfix.rst
+++ b/upcoming_changes/2981.bugfix.rst
@@ -1,0 +1,1 @@
+Fix error which occurs when guessing output size in the :py:meth:`~.signal.BaseSignal.map` function and using dask newer than 2022.7.1 
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -69,6 +69,11 @@ steps:
     conda list
   displayName: Install package
 
+- bash: |
+    source activate $ENV_NAME
+    conda clean --all -y
+  displayName: Clean conda cache
+
 # Note we must use `-n 2` argument for pytest-xdist due to
 # https://github.com/pytest-dev/pytest-xdist/issues/9.
 - bash: |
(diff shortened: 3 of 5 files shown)


jlaehne (approved):

upcoming_changes/2981.bugfix.rst
```diff
@@ -0,0 +1 @@
+Fix error which occurs when guessing output size in the :py:meth:`~.signal.BaseSignal.map` function and using using dask newer than 2022.7.1
```
jlaehne:
```suggestion
Fix error which occurs when guessing output size in the :py:meth:`~.signal.BaseSignal.map` function and using dask newer than 2022.7.1 
```

hyperspy/signal.py
```diff
@@ -4994,7 +4994,7 @@ def _map_iterate(
             testing_kwargs = {}
             for ikey, key in enumerate(arg_keys):
                 test_ind = (0,) * len(os_am.navigation_axes)
-                testing_kwargs[key] = np.squeeze(args[ikey][test_ind]).compute()
+                testing_kwargs[key] = np.squeeze(args[ikey][test_ind].compute())[()]
```
CSSFrancis, line 4999: I think that this is good but we should think about adding in a couple more tests if that is that case. Currently if I remove the `np.squeeze`  and `[()]` all of the tests still pass. 

I'll do some more testing later today but it is possible that the squeeze function isn't necessary.  I think older code used `inav` instead of operating directly on the data which meant that it failed when there was ragged signal passed and `np.squeeze` wasn't called.

CSSFrancis (commented):
@ericpre I think this looks good but I'll do some more testing later today.  I will try to remove the `np.squeeze` function entirely as I think it could cause some issues if there is an array with a dimension of size 1 which is necessary such as possibly in the case of some matrix multiplication. 

I'll see if I can't add tests for that case as well.

CSSFrancis:
Sorry it took me so long to get around to this. It took longer that anticipated for me to understand why some tests were failing.

The current reason for the squeeze function is in the case of trying to pass a signal which is something like <4,4|1> as an iterating signal where the 1 dimension is just an argument to be passed.  This happens with the Holography code a couple of times and fails if you remove the squeeze functions. (Note: the signal probably should be <4,4|>)

The previous reason for the `squeeze` function was because map blocks required the same dimension when iterating. That is no longer the case.

I am not sure if this is the best practice to keep these `squeeze` functions, another thing to consider would be:

- Removing the squeeze functions
- Adding in a warning when arrays of shape (1,) are passed that the array is being passed (ie. [False] rather than False)

This lets us retain the ability to pass in an array like [False] (even if there are limited use cases.)

I am also okay with keeping it as it stands right now and adding an inline comment to the effect of what I said above.

ericpre:
Thanks @CSSFrancis, I have left it as it is and added a reference to this PR in the code.

PR #2986, opened 2022-08-12
Add type hinting on user-facing methods

thomasaarholt:
### Description of the change
- Added type hinting on a few user-facing methods to make the experience better when using the pylance engine in VSCode.
- So far only added basesignal hints to `.mean()` and `.sum()`. See note below about `out`.

I am using `TYPE_CHECKING` to solve circular import errors. See [here](https://mypy.readthedocs.io/en/stable/runtime_troubles.html#import-cycles) for more info.

### Progress of the PR
- [ ] add tests,
- [x] ready for review.

Here is an example of how pylance recognizes the type of `m`, even before running the code.
image

### Notes
I intentionally return `BaseSignal` from `hs.load`. Strictly, it should be `Union[BaseSignal, List[BaseSignal]]`, since the function can return a list of signals. But since I'm not going for full mypy coverage and instead going for "help out the user", I think it is more friendly to stick to just the former.

I also return `BaseSignal` from methods that do have an `out` parameter, when it strictly should be `Union[None, BaseSignal]`. Again because of convenience: if you're using `out` then you won't be returning into an object so you don't care that there is a return type. When you _are_ returning you don't want the output type to possibly be `None` (this raises warnings about `None` not having the e.g. `.plot()` methods.

These could be fixed by using [function overloading](https://mypy.readthedocs.io/en/stable/more_types.html#function-overloading), but there would be a huge amount of code bloat since many methods having `out` as an argument.

### DIFF
--- a/hyperspy/_signals/eds_sem.py
+++ b/hyperspy/_signals/eds_sem.py
@@ -17,6 +17,7 @@
 # along with HyperSpy. If not, see <https://www.gnu.org/licenses/#GPL>.
 
 import logging
+from typing import TYPE_CHECKING
 
 import traits.api as t
 
@@ -25,6 +26,8 @@ from hyperspy.defaults_parser import preferences
 from hyperspy.ui_registry import add_gui_method, DISPLAY_DT, TOOLKIT_DT
 from hyperspy.signal import BaseSetMetadataItems
 
+if TYPE_CHECKING:
+    from hyperspy.models.edssemmodel import EDSSEMModel
 
 _logger = logging.getLogger(__name__)
 
@@ -278,7 +281,7 @@ class EDSSEMSpectrum(EDSSpectrum):
             return False
 
     def create_model(self, auto_background=True, auto_add_lines=True,
-                     *args, **kwargs):
+                     *args, **kwargs) -> "EDSSEMModel":
         """Create a model for the current SEM EDS data.
 
         Parameters
--- a/hyperspy/_signals/eds_tem.py
+++ b/hyperspy/_signals/eds_tem.py
@@ -17,6 +17,7 @@
 # along with HyperSpy. If not, see <https://www.gnu.org/licenses/#GPL>.
 
 
+from typing import TYPE_CHECKING
 import warnings
 import logging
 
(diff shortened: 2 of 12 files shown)


thomasaarholt:
I intentionally return `BaseSignal` from `hs.load`. Strictly, it should be `Union[BaseSignal, List[BaseSignal]]`, since the function can return a list of signals. But since I'm not going for full mypy coverage and instead going for "help out the user", I think it is more friendly to stick to just the former.

I also return `BaseSignal` from methods that do have an `out` parameter, when it strictly should be `Union[None, BaseSignal]`. Again because of convenience: if you're using `out` then you won't be returning into an object so you don't care that there is a return type. When you _are_ returning you don't want the output type to possibly be `None` (this raises warnings about `None` not having the e.g. `.plot()` methods.

These could be fixed by using [function overloading](https://mypy.readthedocs.io/en/stable/more_types.html#function-overloading), but there would be a huge amount of code bloat since many methods having `out` as an argument.

ericpre:
> I intentionally return `BaseSignal` from `hs.load`. Strictly, it should be `Union[BaseSignal, List[BaseSignal]]`, since the function can return a list of signals. But since I'm not going for full mypy coverage and instead going for "help out the user", I think it is more friendly to stick to just the former.
> 
> I also return `BaseSignal` from methods that do have an `out` parameter, when it strictly should be `Union[None, BaseSignal]`. Again because of convenience: if you're using `out` then you won't be returning into an object so you don't care that there is a return type. When you _are_ returning you don't want the output type to possibly be `None` (this raises warnings about `None` not having the e.g. `.plot()` methods.

If I understand correctly, you are saying that you prefer to providing type hints which are **sometimes** in favour of **always** correct at the cost of having a syntax slightly more complicated to read? If so, without any doubt, I prefer the "always correct" approach, because the other one is misleading! ;)

Would it be possible to add a test on CI to check that the imports in the `TYPE_CHECKING` if block are working, just to check for typos? I am concerned that we wouldn't notice typos and the typo hinting may fail silently.

hyperspy/_signals/eds_sem.py
```diff
@@ -25,6 +26,8 @@
 from hyperspy.ui_registry import add_gui_method, DISPLAY_DT, TOOLKIT_DT
 from hyperspy.signal import BaseSetMetadataItems
 
+if TYPE_CHECKING:
+    from hyperspy.models.edssemmodel import EDSSEMModel
```
ericpre, line 30: Is it possible to test these imports on CI?

ericpre (commented):
It needs to be rebase to fix the failure of the test suite. Ideally, it would be better to rebase on `RELEASE_next_major` as this will not be a minor release before release 2.0 - we need to start using rosettasciio ASAP.

PR #2988, opened 2022-08-21
Follow up of #2972 and tidy up left up rosettasciio

ericpre:
Follow up of #2972 to use https://github.com/hyperspy/rosettasciio.

### Progress of the PR
- [x] Remove rosettasciio folder,
- [x] update CI workflow accordingly, 
- [x] update dependency requirement, 
- [n/a] add an changelog entry in the `upcoming_changes` folder (see [`upcoming_changes/README.rst`](https://github.com/hyperspy/hyperspy/blob/RELEASE_next_minor/upcoming_changes/README.rst)),
- [n/a] Check formatting changelog entry in the `readthedocs` doc build of this PR (link in github checks)
- [x] add tests,
- [x] ready for review.


conda_environment.yml
```diff
@@ -6,18 +6,17 @@ dependencies:
...
 - fsspec
-- h5py
-- imageio
+- importlib-metadata
 - ipyparallel
 - ipython !=8.0.*
 - jinja2
 - matplotlib-base
-- matplotlib-scalebar
 - mrcz
```
jlaehne: Isn't this an IO dependency as well?

hyperspy/io.py
```diff
@@ -446,7 +446,11 @@ def load(filenames=None,
         )
 
     if not filenames:
-        raise ValueError('No file(s) provided to reader.')
+        if pattern:
+            # in case, the file doesn't exist
+            raise ValueError(f'No filename matches the pattern "{pattern}"')
+        else:
+            raise ValueError('No file(s) provided to reader.')
```
jlaehne: Add coverage?

hyperspy/tests/test_io.py
```diff
@@ -189,120 +188,110 @@ def test_file_not_found_error():
...
+            'operation': 'load',
+            'hyperspy_version': hs_version
+        },
+        '1': {
+            'io_plugin': 'rsciio.hspy.api',
+            'operation': 'save',
+            'hyperspy_version': hs_version
+        },
+        '2': {
+            'io_plugin': 'rsciio.hspy.api',
```
jlaehne, line 281: This now duplicates the item `'0'` above. Is it just to make sure that the saved file can be reread?

ericpre: They have a time difference, but in the test it is removed to simplify the test.

hyperspy/io.py
ericpre: Actually, this branch is not necessary because this case is covered in lines 419-420 above.

ericpre (commented):
Should be all done!

jlaehne (approved):

PR #2992, opened 2022-08-24
bugfix: initializing metadata breaks its default blank dicts when loading DigitalMicrograph files.

nem1234:
### Description of the change
- Bugfix: When reading second file, the metadata is not correctly initialized. Everytime loading dm4 images, the initializer overwrites its default blank dictionary.

### Progress of the PR
- [ ] add tests,
- [x] ready for review.

```python
a = hs.load(dm4 file with rectangle markers)
b = hs.load(dm4 file without rectangle markers)
then the rectangle marker on a is also shown on b

------------------------------------------
## pseudo code to explain how it works.
# current situation
def test0(m = {}):
    print("test0: Initial m =", m)
    m['abc'] = 10
    return m

# fixed situation
def test1(m = None):
    if m is None:
        m = {}
    print("Test1: Initial m =", m)
    m['abc'] = 10
    return m

a = test0()
a['def'] = 'x'
b = test0()
print(a, b, a==b)
# {} in the function argument is overwritten on first call 

c = test1()
c['def'] = 'x'
d = test1()
print(c, d, c==d)
```

### DIFF
--- a/hyperspy/io_plugins/digital_micrograph.py
+++ b/hyperspy/io_plugins/digital_micrograph.py
@@ -107,10 +107,12 @@ class DigitalMicrographReader(object):
         else:
             self.endian = 'big'
 
-    def parse_tags(self, ntags, group_name='root', group_dict={}):
+    def parse_tags(self, ntags, group_name='root', group_dict=None):
         """Parse the DM file into a dictionary.
 
         """
+        if group_dict is None:
+            group_dict = {}
         unnammed_data_tags = 0
         unnammed_group_tags = 0
         for tag in range(ntags):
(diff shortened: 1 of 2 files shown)


jlaehne:
The affected files are being moved to https://github.com/hyperspy/rosettasciio with `RELEASE_next_major`. The development of the files has already been moved out of HyperSpy. Patches contributed to HyperSpy would get lost. If the patch can wait, it should be resubmitted there. Otherwise, it would have to be duplicated there once it is ready to be merged.

jlaehne:
Actually, the exact same fix is already part of https://github.com/hyperspy/rosettasciio/pull/23

So, I am OK with simply merging it into `RELEASE_next_patch`, in case another patch release is made before the next major release.

ericpre:
I think that it worth doing a patch release, as there are already a few bugs fixed in the `RELEASE_next_patch` branch.

ericpre (approved):

jlaehne (approved):

PR #2994, opened 2022-08-24
Fix dm4/tiff reader, initializing metadata overwrites its default val…

nem1234:
### Description of the change
- This is the same as PR: #2992 but for RELEASE_next_major branch
- fixed: Initial metadata of dm3/dm4 files is overwritten when calling dm loader.
  
### Progress of the PR
- [ ] add tests,
- [ ] ready for review.

\## pseudo code to explain how it works.
\# current situation
def test0(m = {}):
    print("test0: Initial m =", m)
    m['abc'] = 10
    return m

\# fixed situation
def test1(m = None):
    if m is None:
        m = {}
    print("Test1: Initial m =", m)
    m['abc'] = 10
    return m

a = test0()
a['def'] = 'x'
b = test0()
print(a, b, a==b)
\# {} in the function argument is overwritten on first call 

c = test1()
c['def'] = 'x'
d = test1()
print(c, d, c==d)

### DIFF
--- a/rosettasciio/rsciio/digital_micrograph/api.py
+++ b/rosettasciio/rsciio/digital_micrograph/api.py
@@ -94,10 +94,12 @@ class DigitalMicrographReader(object):
         else:
             self.endian = 'big'
 
-    def parse_tags(self, ntags, group_name='root', group_dict={}):
+    def parse_tags(self, ntags, group_name='root', group_dict=None):
         """Parse the DM file into a dictionary.
 
         """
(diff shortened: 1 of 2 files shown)


jlaehne:
Thanks, but the affected files have been moved out of `RELEASE_next_major` and will be deleted in https://github.com/hyperspy/hyperspy/pull/2988

The fixes are already contained in https://github.com/hyperspy/rosettasciio/pull/23 and I will therefore close this PR.

Issue #2996, opened 2022-08-25
Release 1.7.x and 2.0.0


ericpre:
## [v2.0.0](https://github.com/hyperspy/hyperspy/milestone/31)
As the IO have been split - see https://github.com/hyperspy/hyperspy/pull/2972 and https://github.com/hyperspy/rosettasciio - we should get 2.0 released ASAP.

As this is major release, it is a good opportunity to change some API
- [x] Split spectroscopy (EDS and EELS) #3202
- [x] Split holography #3201 
- [x] Add intersphinx links to HoloSpy/EleXSpy ones available (e.g. for mentions introduced in #3216)
- [x] Clear all deprecations
  - [x] hyperspy/hyperspy#2998
  - [x] hyperspy/hyperspy#2999
  - [x] hyperspy/hyperspy#3000
  - [x] hyperspy/hyperspy#3007
  - [x] hyperspy/hyperspy#3014
  - [x] hyperspy/hyperspy#3017
  - [x] hyperspy/hyperspy#3037 of various functions and arguments
  - [x] hyperspy/hyperspy#3063
  - [x] Check for leftovers of the deprecation (especially in the documentation)
    - [x] Correct documentation and examples on `optimizer` and `loss_function` parameters in user guide (model-fitting page) and docstring of `m.fit()`
    - [x] fix marker naming in docstring examples (done in hyperspy/hyperspy#3063)
- [ ] go through some API changes worth doing
    - [x] #3223
    - [ ] #3170 
    - [ ]  for example https://github.com/hyperspy/hyperspy/issues/1227
    - [x] #3197 rename `crop_image` to `crop_signal2D` to be consistent with `crop_signal1D`?
    - [ ] reorganize metadata: https://github.com/hyperspy/hyperspy/issues/2095
    - [ ] review usage of `mask` versus `navigation_mask`, `signal_mask` and if consistency can be improved - see for example https://github.com/hyperspy/hyperspy/issues/3053
- [x] Clarify what is public/private, and improve API documentation?
      - See for example: https://github.com/hyperspy/hyperspy/pull/3014#discussion_r973575789
- [ ] Improve Plotting of Markers/ peakfinding:
    - [x] hyperspy/hyperspy#3148 
    - [x] hyperspy/hyperspy#3142
    - [ ] #3196 
    - [x] hyperspy/hyperspy#3175 
    - [x] #3140 
    - [ ] #3199 
- [x] Dependency Management:
[...]

PR #2998, opened 2022-08-26
Remove deprecation - Axes, IO, Model

ericpre:
First round of deprecation removal.

### Progress of the PR
- [x] `axes.py`,
- [x] `io.py`,
- [x] `model.py`, 
- [x] update tests,
- [x] ready for review.


jlaehne (approved):

CHANGES.rst
```diff
@@ -6,6 +6,60 @@ https://hyperspy.readthedocs.io/en/latest/user_guide/changes.html
 
 .. towncrier-draft-entries:: |release| [UNRELEASED]
 
+
+API Removal
+===========
```
jlaehne, line 11: Shouldn't this be a heading of one level lower than the new release?
(Also applies to the other deprecation PRs)

jlaehne, line 11:
```suggestion
-----------
```

CHANGES.rst
```diff
@@ -6,6 +6,60 @@ https://hyperspy.readthedocs.io/en/latest/user_guide/changes.html
...
+
+Following deprecation cycle, the following methods and functions have been removed:
```
jlaehne:
```suggestion
Following the deprecation cycle, the following methods and functions have been removed:
```

CHANGES.rst
ericpre: It doesn't play well with towncrier, so I would suggest to leave it as it is and we edit it during when we do the release.

jlaehne: Yep, just realized that as well.

CHANGES.rst
```diff
@@ -6,6 +6,63 @@ https://hyperspy.readthedocs.io/en/latest/user_guide/changes.html
 
 .. towncrier-draft-entries:: |release| [UNRELEASED]
 
+
+API Removal
+===========
+
+As the HyperSpy API evolves, APIs are occasionally reorganized or upgraded.
+When APIs evolve, the old API is deprecated and eventually removed in a major
+release. The functions and methods removed in HyperSpy 2.0 is listed below along
```
jlaehne:
```suggestion
release. The functions and methods removed in HyperSpy 2.0 are listed below along
```

jlaehne:
I have seen in the RosettaSciIO deprecation PR that not always the changed terminology was properly documented in the user guide. So the documentation should be searched again for all deprecated names in the three deprecation PRs to make sure that there are no remnants of old terminology.

ericpre:
Yes, this should be done after we remove all deprecation?

jlaehne:
> Yes, this should be done after we remove all deprecation?

It could be done in each of the three branches for the respective keywords and functions by searching a local copy (as I did in https://github.com/hyperspy/rosettasciio/pull/30). Otherwise, we should add a checkpoint to the todo list in https://github.com/hyperspy/hyperspy/issues/2996

ericpre:
As there are many deprecations, I made several PRs, so it is not too much pain to review, but the drawback of this approach is that, it is more difficult to find some left over which is why I suggested to do it later.

jlaehne:
OK, I added to the todo list in https://github.com/hyperspy/hyperspy/issues/2996

ericpre:
Sorry, I missed the point that you were referring to the user guide!

jlaehne (approved):

jlaehne:
I added a PR with some changes to the documentation.

jlaehne (approved):

PR #3008, opened 2022-09-05
Fix removing horizontal and vertical line widget

### LINKED ISSUE #2965: Matplotlib regression in widget removal
#### Describe the bug
Commit https://github.com/hyperspy/hyperspy/commit/02745991c338de35448be8ff98ff66c1e7577d51 modified widget removal code, and it appears to not work as intended. I had some code prior that would take the navigator image for a linescan plot, which would remove the navigation pointer. That's no longer working as of v1.7.

#### To Reproduce

```python
import hyperspy.api as hs
import numpy as np

sp = hs.datasets.example_signals.EDS_TEM_Spectrum()
s = hs.stack([sp * i for i in np.arange(.1, 1, .3)], new_axis_name='x')
s.plot()
s._plot.pointer.set_on(False)       # remove pointer
f = s._plot.navigator_plot.figure
f.savefig('/tmp/output.png')      # notice that pointer is in output (in 1.6.5, this works)
```

##### v1.7 output (with `matplotlib` v3.5.2)

output

##### v1.6.5 output (with `matplotlib` v3.4.3)

output

#### Python environment:
 - HyperSpy version: 1.7.0
 - Python version: 3.7.12
 - Matplotlib version: 3.5.2

#### Additional context

This appears to be because of the changes at https://github.com/hyperspy/hyperspy/blob/RELEASE_next_minor/hyperspy/drawing/widget.py#L134-L138

`p.remove()` appears to not do anything by my investigation. Also, even though I'm on v3.5.2 of `matplotlib`, the `container.remove(p)` command still works. I presume this was added do to some sort of deprecation I cannot find...

ericpre:
Fix #2965.

### Progress of the PR
- [x] Fix removing line widgets from figure: the lines were added twice to the `matplotlib` lines containter (`ax.lines`)
- [x] add tests,
- [x] ready for review.

```python
import hyperspy.api as hs
import numpy as np

s = hs.signals.Signal2D(np.arange(10*25*25).reshape(10, 25, 25))
s.plot()

# Remove pointer
s._plot.pointer.close(render_figure=True)
```

### DIFF
--- a/hyperspy/drawing/_widgets/horizontal_line.py
+++ b/hyperspy/drawing/_widgets/horizontal_line.py
@@ -31,14 +31,19 @@ class HorizontalLineWidget(Widget1DBase):
             self.patch[0].set_ydata(self._pos[0])
             self.draw_patch()
 
+    def _add_patch_to(self, ax):
+        """Create and add the matplotlib patches to 'ax'"""
+        self.blit = hasattr(ax, 'hspy_fig') and ax.figure.canvas.supports_blit
+        self._set_patch()
+        for p in self.patch:
+            p.set_animated(self.blit)
+
     def _set_patch(self):
         ax = self.ax
         kwargs = picker_kwargs(preferences.Plot.pick_tolerance)
-        self._patch = [ax.axhline(
-            self._pos[0],
-            color=self.color,
-            alpha=self.alpha,
-            **kwargs)]
+        self._patch = [
+            ax.axhline(self._pos[0], color=self.color, alpha=self.alpha, **kwargs)
+            ]
 
     def _onmousemove(self, event):
         """on mouse motion draw the cursor if picked"""
--- a/hyperspy/drawing/_widgets/vertical_line.py
+++ b/hyperspy/drawing/_widgets/vertical_line.py
@@ -31,13 +31,19 @@ class VerticalLineWidget(Widget1DBase):
             self.patch[0].set_xdata(self._pos[0])
             self.draw_patch()
 
+    def _add_patch_to(self, ax):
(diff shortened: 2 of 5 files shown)


ericpre:
@jat255, as it fixes https://github.com/hyperspy/hyperspy/issues/2965, can you please review this PR?

ericpre:
Anyone to review this PR, so that we can release 1.7.2? :)

jlaehne:
Couldn't sort out locally yet, which branch of `hyperspy_gui_traitsui` to use to check the behavior.

Otherwise, lgtm.

ericpre:
This one is not related to `hyperspy_gui_traitsui`. The issue is described in https://github.com/hyperspy/hyperspy/issues/2965 but in short, the issue is that the widgets (line) stay on the figure when removing the pointer from the navigator, while it should be removed too. An example to reproduce:
```python
import hyperspy.api as hs
import numpy as np

s = hs.signals.Signal1D(np.arange(10*50).reshape(10, 50))
s.plot()

# Remove pointer
s._plot.pointer.close()
```

This doesn't happen for pointer with square widget (instead of line) and the explanation is that when adding a the line widget, there was a duplicated reference to the line and matplotlib would remove only one (as expected). In this PR, this is fixed by changing way the line widget is added to the figure to avoid the duplicated reference!

jlaehne:
> Couldn't sort out locally yet, which branch of `hyperspy_gui_traitsui` to use to check the behavior.

Sorry, fetched the wrong PR to try it locally. Now I can reproduce that it is doing what it is supposed to do.

PR #3012, opened 2022-09-13
Axes Refactor

CSSFrancis:
### This is the first in a series of changes necessary to implement Vector Signals

This simplified the Axes class into three Classes which extend the BaseAxis class

- BaseAxis:
    - Only has a name and unit property
    - Defines how an axis exists in some AxisManager 
- DataAxis
    - Has some axis property
    - Has a `to_numpy_index` function
    - Has a `__get_item__` function which slices the axis
    - Can be increasing, decreasing or unordered (unordered accounts for arbitrary axis labeling)
- FunctionalAxis (Extends DataAxis)
    - Has some function which operates on the _axis property to define self.axis
    - Can be increasing, decreasing or unordered. 
- UniformAxis (Extends DataAxis)
    -  self.axis defined by the function y = scale*range(self.size)+offset 

Other notable changes include many of the previously set traits are replaced with functional properties.  This might mean that we might not have to depend on the traits package.  This might be preferable as I feel the traits implementation sometimes gets confusing. 

Some notable examples for the Data Axis:
```python
@property
def axis(self)
    return self._axis

@property
def high_value(self):
    if self.is_ordered is True:
        return self.axis[-1]
    elif self.is_ordered is False:
        return self.axis[0]
    else:
        raise ValueError("The axis is not ordered so there is no high/low value")

@property
def low_value(self):
    if self.is_ordered is True:
        return self.axis[0]
    elif self.is_ordered is False:
        return self.axis[-1]
    else:
         raise ValueError("The axis is not ordered so there is no high/low value")

@property
def low_index(self):
    return 0

@property
def high_index(self):
    return self.size - 1

@property
def size(self):
    return len(self._axis)
```

### Progress of the PR

- [x] Refactor Classes
- [ ] Clean up traits and inheritance
- [ ] Redo tests for slicing
    - [x] Test float slicing
    - [x] Test relative slicing
    - [x] Test negative slicing
    - [x] Test slicing with column labels
    - [x] Test slicing using a bool array
    - [x] Test slicing with array of indexes
- [ ]  Redo tests for changing axes class
- [ ] Redo tests for axes_manager 
- [ ] ready for review.

All of the following slices should work on both axes (and the arrays they represent)
```python

# slicing an unordered Data Axis
print(ax1.axis) # ["x", "y", "z"](NEW)
ax1["x"] # slice a DataAxis using an column label(NEW)
ax1[[True, False, True]] # slice a DataAxis using a bool array(NEW)
ax1[(0,2)] # slice a DataAxis using a tupleof indexes (NEW)

ax1[0] == ax1["x"] # True

# slicing an ordered Data Axis (same functionality as FunctionalDataAxis)
print(ax2.axis) # [1,2,3,5,7,10,12,7]
ax2[ax2.axis>4] # slice a DataAxis using a bool array(NEW)
ax2["-12.":]# slice a DataAxis using an negative float(NEW)
ax2["rel0.5":"rel0.9"] # relative slicing
ax2[2.0:12.] # real unit slicing

# slicing an ordered UniformAxis 
print(ax3.axis) # [4,6,8,10,12,14,16]
ax3[:.:,4.0] # slicing with a float step
ax3[[True, True, False, True, True, False, False]] # slice a DataAxis using a bool array(NEW) --> Returns DataAxis
ax3[(1,4,5,3)] # slice a DataAxis using a index array(NEW) --> Returns DataAxis

# Note that slicing some signal should work the same as slicing some array:

s.isig["x"] --> calls s.axes_manager.signal_axes[0]["x"]

```

I have submitted this PR as a draft so that any suggestions can implemented.


CSSFrancis:
I will note that some of these changes drastically shorten some of the code which is a result of removing lots of duplicate code.

More simplification is possible if Traits.UI is removed from the BaseAxis class (I haven't looked into the AxesManager but it could probably be removed from there too)

I don't believe that it is completely necessary and it significantly increases the complexity of the Axes objects.  I feel as this might be something of a leftover in the code base which we could probably remove in favor of the `@property` decorator. At the core every axis is either defined as an axis or a size/function.  It is pretty easy to define every other property as some function of those properties.  Some of these are more computationally costly than others but most are extremely cheap.

ericpre:
> More simplification is possible if Traits.UI is removed from the BaseAxis class (I haven't looked into the AxesManager but it could probably be removed from there too)
> 
> I don't believe that it is completely necessary and it significantly increases the complexity of the Axes objects. I feel as this might be something of a leftover in the code base which we could probably remove in favor of the `@property` decorator. At the core every axis is either defined as an axis or a size/function. It is pretty easy to define every other property as some function of those properties. Some of these are more computationally costly than others but most are extremely cheap.

These are necessary for the GUI widgets (traitsui and ipywidgets)! They may be room for some streamlining but I don't think you can remove that much code! ;)

CSSFrancis:
> These are necessary for the GUI widgets (traitsui and ipywidgets)! They may be room for some streamlining but I don't think you can remove that much code! ;)

Ahh! I see that now!  I'm still pretty happy with the fair bit of code reduction in this PR as I took out about a third of the `BaseAxis` and it's children's code. Most of it is due to removing lots of very similar functions like `_slice_me`, `value2index` and splitting how each slice is handled into separate functions.  I haven't touched the AxesManager code yet.

A couple of questions that came up while working:

- Should the Axes and AxesManager classes be defined in separate files?  The file is rather large which makes it hard to find things sometimes. 
- Should we throw and error when someone uses and index larger than the array? Currently one is thrown
    - For example `a[2:10] # a.size=5`  should this throw a value error or return `a[2:]`
    - For a numpy array there are no problems with indexes larger than the max index and I think we should follow numpy as closely as possible otherwise.
    - Right now it seems arbitrary that if `a.high_value=10.4` then `a[0:10.4]` doesn't throw and error but `a[0:10.4001]` does.  I would rather a warning come up than it automatically fail.

ericpre:
> A couple of questions that came up while working:
> 
> * Should the Axes and AxesManager classes be defined in separate files?  The file is rather large which makes it hard to find things sometimes.

Yes, but not now please, because it suspect that it would be more difficult to visualise the diff for the review.

> * Should we throw and error when someone uses and index larger than the array? Currently one is thrown
>   
>   * For example `a[2:10] # a.size=5`  should this throw a value error or return `a[2:]`
>   * For a numpy array there are no problems with indexes larger than the max index and I think we should follow numpy as closely as possible otherwise.
>   * Right now it seems arbitrary that if `a.high_value=10.4` then `a[0:10.4]` doesn't throw and error but `a[0:10.4001]` does.  I would rather a warning come up than it automatically fail.

Yes, I think that makes sense.

CSSFrancis:
Closed in favor of #3031

PR #3016, opened 2022-09-16
Fix SpanSelector for matplotlib 3.6

ericpre:
We have caught before the release of matplotlib 3.6 (happened today) with the test suite of https://github.com/hyperspy/hyperspy-extensions-list but because other test suites were failing for a while, it wasn't noticeable.

The issue is that we are using a vendored version of `matplotlib.widgets.SpanSelector` only for matplotlib<3.6 but the import "dispatch" was incorrect.

### Progress of the PR
- [x] Fix matplotlib `SpanSelector` import,
- [n/a] add tests,
- [x] ready for review.

### DIFF
--- a/hyperspy/drawing/_widgets/range.py
+++ b/hyperspy/drawing/_widgets/range.py
@@ -18,13 +18,18 @@
 
 import inspect
 import logging
+from packaging.version import Version
 
+import matplotlib
 import numpy as np
 
 from hyperspy.drawing.widget import ResizableDraggableWidgetBase
 from hyperspy.defaults_parser import preferences
 
-from hyperspy.external.matplotlib.widgets import SpanSelector
+if Version(matplotlib.__version__) >= Version('3.6.0'):
+    from matplotlib.widgets import SpanSelector
+else:
+    from hyperspy.external.matplotlib.widgets import SpanSelector
 
 _logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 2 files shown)


hyperspy/drawing/_widgets/range.py
```diff
@@ -18,13 +18,18 @@
...
 import numpy as np
 
 from hyperspy.drawing.widget import ResizableDraggableWidgetBase
 from hyperspy.defaults_parser import preferences
 
-from hyperspy.external.matplotlib.widgets import SpanSelector
+if Version(matplotlib.__version__) >= Version('3.6.0'):
+    from matplotlib.widgets import SpanSelector
+else:
+    from hyperspy.external.matplotlib.widgets import SpanSelector
```
jlaehne, line 32: Would it make sense to add a comment in `external/matplotlib/widgets.py` that once the minimum supported version of Matplotlib should be set to `3.6.0`, the function `SpanSelector` could be deleted? But I guess it is clear enough from this call?

ericpre: There is already this comment: https://github.com/hyperspy/hyperspy/blob/cf75ac8440dc146e678a985128a62d92a65dd574/hyperspy/external/matplotlib/widgets.py#L1-L2

jlaehne (approved):

Issue #2993, opened 2022-08-24
GUI masking tools


sivborg:
Hello!

I have been working on some features connected to HyperSpy, and so I was wondering whether they would be relevant in this project. If not, I will implement them somewhere else.

#### Describe the functionality you would like to see.

These tools allow GUI masking of features on-the-fly, which can remove or isolate specific features in the data. 

#### Describe the context
This is to be used on any signal where a mask might be necessary. Use cases include `hyperspy.model.BaseModel.multifit`, to mask out features that are not to be fitted.

These are currently implemented as wrappers around the different Selector classes from `matplotlib.widget`, and from that they leverage the continued maintenance of `matplotlib`.

#### Additional information
Here is a short animation of creating a mask.
![Animation](https://user-images.githubusercontent.com/106336292/182832240-2ed22a7d-1721-4390-861f-e90c996a5f51.gif)

So I am asking if this is a relevant fit for this project. If so, where in the software would be an appropriate location?

ericpre:
This is very relevant - I started to look at it a while ago but didn't have time to finish... It shouldn't be particularly difficult, but it needs to be done correctly, as in it should integrate with other hyperspy ROI and widgets. The matplotlib widgets have been improved quite a bit in recent release and this should help.

If you can make a PR, then we can discuss about it. Please use the `Release_next_major` branch as a base, as it is very unlikely that they will be a minor release.

sivborg:
Thank you for the prompt response! I will take some time to refactor and look into integration before opening the PR, where we will take it further.

sivborg:
The first steps of this proposal are currently implemented in #3030

ericpre:
Usually, we keep the issue open until it is merged.

PR #3030, opened 2022-09-22
PolygonROI

### LINKED ISSUE #3353: ROI selection
Hello everyone, 
I just started using hyperspy and I noticed that I cannot select customized ROI. I need to select triangles for ROI and I was wondering if there is anyway of doing it with hyperspy? 
Thanks

sivborg:
This is part of the implementation of #2993 , where the aim is to integrate GUI masking into HyperSpy from matplotlib widgets.

As a step forward, I have extended the ROI functionality to a two-dimensional polygonal region-of-interest. Since this step is independent of the GUI, and to keep the PRs in more manageable scopes, I decided to merge this functionality in first.

Moreover, this feature reuses a lot of code from `CircleROI` as the principle is very similar.

Since the masking tools are targeted for `RELEASE_next_major`, this PR is as well. However, this would also be suitable for a minor release, so I can close this PR and move the functionality to target `RELEASE_next_minor` if preferred.

While this PR has some masking functionality, I do not want it to interfere with #2375, so let me know if I should 

### Description of the change
- Added a `PolygonROI` region-of-interest class that can be used to mask areas as defined by a polygon
- Added to it a `boolean_mask` function that can retrieve the ROI as a boolean numpy array, which can be passed on to methods requiring a mask.

### Progress of the PR
- [x] add tests,
- [x] ready for review.

```python
import hyperspy.api as hs
import numpy as np
s = hs.signals.Signal1D(np.ones((50, 60, 4)))
polygon_roi = hs.roi.PolygonROI([(20, 20), (36, 20), (26, 36)])
masked = polygon_roi(s)
```

### Technical details.

The implementation is very similar to CircleROI, but uses a mask from rasterizing a polygon rather than a circle. 

As a polygon is not directly available as a widget in HyperSpy, I have omitted the interactive functionality for now. As the functionality is added to the GUI, `PolygonROI` can be extended to support this.

The algorithm used to rasterize the polygon into an array has a few limitations:
- A self-overlapping polygon can have its overlapping areas masked out, which can easily be unintended. I have tried to make this clear in the docstrings.
- The algorithm is implemented with Python built-ins, meaning it can be slow to run it repeatedly. One way to remedy this is to run the algorithm every time the polygon is updated, then cache the resulting array as a member. I'm not sure whether this efficiency is worth it, however.

Compared to other ROIs, this one does not have a fixed amount of parameters as a polygon can have as many points as necessary. This changes some of the conventions by displaying the parameters. Currenly, this is how the class behaves:
```python
r = hs.roi.PolygonROI([(10, 20),(20, 50),(30, 30)])
print(r.parameters)
print(tuple(r))
print(repr(r))
```
```
{'points': [(10, 20), (20, 50), (30, 30)]}
((10, 20), (20, 50), (30, 30))
PolygonROI(points=[(10, 20), (20, 50), (30, 30)])
```


hyperspy/roi.py
```diff
@@ -1468,6 +1468,270 @@ def __call__(self, signal, out=None, axes=None, order=0):
...
+
+    points = []
+    _ndim = 2
+
+    def __init__(self, points=None):
+        super(PolygonROI, self).__init__()
+        if points:
+            self.set_vertices(points)
+        else:
+            points = []
```
ericpre:
```suggestion
        super().__init__()
        if points:
            self.set_vertices(points)
```

hyperspy/roi.py
```diff
@@ -1468,6 +1468,270 @@ def __call__(self, signal, out=None, axes=None, order=0):
...
 
+class PolygonROI(BaseROI):
+    """Selects a polygonal region in a 2D space. The coordinates of the
+    polygon vertices are given as a list of tuples (x, y) in the `points`
+    attribute. An edge runs from the final to the first point in the list.
+    If the polygon self overlaps, the overlapping areas may be considered
+    outside of the polygon and masked away.
+    """
+
+    points = []
```
ericpre: Use `vertices` instead of `points`?

hyperspy/roi.py
```diff
@@ -1468,6 +1468,270 @@ def __call__(self, signal, out=None, axes=None, order=0):
...
+        mask = np.broadcast_to(mask, tiles)
+        # roi.data = np.ma.masked_array(roi.data, mask, hard_mask=True)
+        roi.data = np.where(mask, np.nan, roi.data)
+        if out is None:
+            return roi
+        else:
+            out.events.data_changed.trigger(out)
+
+    def _rasterized_mask(
+        self, xy_max=None, xy_min=(0, 0), x_scale=1, y_scale=1, inverted=False
```
ericpre: Since there is a very simple syntax to invert bool array (`~`), is the `inverted` argument necessary?

ericpre (commented):
From a quick look, this seems to be on the right track. I suspect that the most difficult part is to generate the mask, which is already done here. Do you wan to add the matplotlib polygon widget in this PR? I think that it shouldn't be too much code.

> Since the masking tools are targeted for RELEASE_next_major, this PR is as well. However, this would also be suitable for a minor release, so I can close this PR and move the functionality to target RELEASE_next_minor if preferred.

I think that it would be unlikely that they will be a minor release between the major release.

> While this PR has some masking functionality, I do not want it to interfere with https://github.com/hyperspy/hyperspy/pull/2375, so let me know if I should

As I understand, even if both PR are dealing with mask, their focus are different.

> The implementation is very similar to CircleROI, but uses a mask from rasterizing a polygon rather than a circle.

I guess that it would be good to make a base class for mask-ROI classes, as I would imagine that in the future, they could be more mask-ROI classes?

hyperspy/roi.py
sivborg: I am planning to have an `inverted` argument for the `__call__` function which masks out the interior of the polygon, rather than the outside. This would be tricky to do otherwise. So I put them there for consistency, but maybe it is a bit cluttered, actually, so I think I'll remove it.

hyperspy/roi.py
sivborg: At first I thought `points` would be a bit more consistent with the other ROI classes, but `vertices` is more accurate in this context, so I think I will change it.

sivborg:
Thanks for the look!
> Do you wan to add the matplotlib polygon widget in this PR? I think that it shouldn't be too much code.

I think I can draft the implementation of this pretty soon and decide after the scope of it is clear.
>I guess that it would be good to make a base class for mask-ROI classes, as I would imagine that in the future, they could be more mask-ROI classes?

Yes, there is a bit of shared code that could be useful to put in its own class. It might have to involve multiple inheritance, as `CircleROI` inherits from the interactive ROI while `PolygonROI` does not. So I think I will examine it a bit further, first.

sivborg:
Though I had some difficulties, I have finished a working prototype for integrating `matplitlob.widgets` selectors into HyperSpy, which here is done in the `PolygonWidget` class. The code requirements were actually not very demanding, so I believe I can fit it into this PR.

The interface for the MPL widgets are quite different from the existing HyperSpy ones, so I added a base class `MPLWidgetBase` that implements some common operations, so we can expand it to other widgets in `matplotlib.widgets` if desired. However, it might be a good idea to create a common base class between `WidgetBase` and `MPLWidgetBase`. I will look a bit into it.

sivborg:
Hello, I currently have a working implementation of the PolygonROI and accompanying PolygonWidget. I ended up spending quite some more time developing this, as I found it useful the option of having multiple polygons as a ROI, which is now implemented. Let me know what you think!

Here is a minimal example of how it works, with `get_dummy_signal` as defined at the end of this comment:
```
import hyperspy.api as hs
import hyperspy.drawing.widgets as widgets

s = get_dummy_signal()
s.plot()

roi = hs.roi.PolygonROI()
roi.add_widget(s, axes=[2,3]) # Leave out `axes` argument for navigation space

s_roi = roi(s, axes=[2,3])
s_roi.plot()
```

"Escape" resets the currently constructed polygon, "shift"+click to move it, grab the vertex handles to move the vertices. Click outside the finished polygon to start another one.
Click inside a finished polygon to select it, giving it a thick red outline. Then press "Delete" to remove the selected one. This can not be done while constructing another one.

You can also get a boolean mask from the ROI, which is useful for interfacing with other libraries:
```
mask = roi.boolean_mask(s.axes_manager,axes=[2,3])
```

I have tested that it works with the same interface/behaviour as the other ROIs, in particular CircleROI which is quite similar. I have checked and found it seems to work with `.interactive` and lazy signals, although not together as much. It works with IPython too, I've found. Also with different scale factors on x- and y-axis.
However, it does not currently have a `.gui()` implementation.

Here is an example of using it interactively with a histogram:
```
s = get_dummy_signal()
s.plot()

roi = hs.roi.PolygonROI([[10, 20], [14, 13], [1, 10]]) # You can supply initial polygons as lists of vertices.
roiint = roi.interactive(s,axes=[2,3])

roihist = hs.interactive(roiint.get_histogram,
                        event=roi.events.changed,
                        bins=150,
                        recompute_out_event=None
)
roihist.plot()
```

The public interface of `matplotlib.widgets.PolygonSelector`, which this functionality is based on, does not expose much of its behavior. So the implementation is a combination of basic `matplotlib` together with `PolygonSelector`.

Hope this will be of use, and thank you for any feedback! There are perhaps some things still that I would like to go over once more. Plus some tests.

(Here is my example signal)
```
def get_dummy_signal():
    s = np.indices((20,20)).T[np.newaxis][np.zeros((20,20),dtype=int)]
    s = s - np.indices((20,20)).T[:,:,np.newaxis,np.newaxis,:]
    s = 800 - np.sum(s**2,axis=-1)
    return hs.signals.Signal2D(s)
```

ericpre:
This looks very promising! :) It is quite a bit to review but I will try to do in the coming weeks.

sivborg:
I have reviewed some of the functionality recently, and have explored one particular issue with the plotting of a lazy signal after using `PolygonROI`.
The image for the navigator in the plot is in a lazy signal created by averaging the chunks containing the centre point of the signal. The centre point is the `index` argument in https://github.com/sivborg/hyperspy/blob/1720d0b1aeaf8822940a738092d58af4aac69241/hyperspy/_signals/lazy.py#L1221
However, in a signal cropped by `PolygonROI` this centre point can sometimes be cropped out, especially when using a region of multiple polygons. The result is that the navigator becomes blank due to adding NaN values. This also happens if the centre point is close to an edge, leading to some of the values in the chunk becoming NaN.
One way to remedy this is by choosing something else than the centre for `index` when finding the centre point being NaN in one navigator position (or all). Perhaps using metadata that marks that the signal has been cropped by `PolygonROI` to activate this case. However, the question is whether it is worth it to change the behaviour of the lazy signal for this. I should remark too that using CircleROI can also result in this, so it is not unique behavior.

CSSFrancis:
@sivborg would this not be fixed by just having the compute navigator function call `nansum`? Or are all of the values nan in the middle chunk?

Using the center chunk of the signal for computing the navigator is nice but does introduce some potential inconsistencies/ potential strange behavior.

sivborg:
@CSSFrancis Using `nansum` in the `compute_navigator` function is a possible solution, yes. But this may also happen in a non-lazy signal, so it might be necessary to add it there as well. 

Though in my use cases, the centre chunk is often uniformly nan, as I choose multiple polygon at large distances, leaving the middle as NaN. So another way to choose default `index` could be useful.

I have made a quick working solution, although I am not sure it is worth it to change so much of the existing functionality, potentially introducing some bugs, to fix a visual problem. A user could also fix this themselves by replacing an NaN with zeroes before plotting.

hyperspy/drawing/widget.py
```diff
@@ -1020,3 +1020,52 @@ def _add_patch_to(self, ax):
             self._set_resizers(True, ax)
         if hasattr(super(ResizersMixin, self), '_add_patch_to'):
             super(ResizersMixin, self)._add_patch_to(ax)
+
+
+class MPLWidgetBase(WidgetBase):
```
ericpre: This class should also be a parent class of `RangeWidget`, which uses `matplotlib.widgets.SpanSelector`.

Would it make more sense that it inherit from `ResizableDraggableWidgetBase` instead of `WidgetBase`?

hyperspy/drawing/_widgets/polygon.py
```diff
@@ -0,0 +1,304 @@
...
+    polygons, which can then be used as a region-of-interest.
+    The active polygon can be moved by shift-clicking.
+    A polygon vertex can be moved by clicking its handle. If incomplete,
+    it is also necessary to press control.
+    The active polygon can be deleted by pressing escape.
+    To delete other polygons, click inside it until it has a red outline,
+    then press `Delete`.
+    """
+
+    def __init__(self, axes_manager, mpl_ax=None, polygons=None, **kwargs):
```
ericpre: To keep the API consistent across widgets, we should avoid adding other argument, for example, here: `mpl_ax` and `polygons`.

If there is a need for these arguments, it would be better to implement it in a separate PR for all widgets.

hyperspy/drawing/_widgets/polygon.py
```diff
@@ -0,0 +1,304 @@
...
+                max(all_y_coordinates),
+            )
+            tup = (
+                (bounding_box[0] + bounding_box[1]) / 2,
+                (bounding_box[2] + bounding_box[3]) / 2,
+            )
+
+            self.position = tup
+
+    def _click_event(self, event):
```
ericpre: This should be all handled in matplotlib? If some changes are required in matplotlib, then we should make a PR to matplotlib and make a copy of the updated `matplotlib.widgets.PolygonSelector` in hyperspy (see https://github.com/hyperspy/hyperspy/blob/RELEASE_next_minor/hyperspy/external/matplotlib/widgets.py).

ericpre (commented):
This look already fairly good and sorry for coming back on this only now.
The API needs to be made consistent with the API of existing widgets - there are some comments in my review on that.
One thing that I find surprising in the current implementation of `PolygonROI` is that it consist of several polygons and once there are fixed, they can't be changed interactively... I think that this is a significant inconsistency with the existing behaviour of other ROIs. If the aim of this PR is to be able to select/mask areas with several polygons, then this should be done by adding a different functionality which will take care of combining multiple ROIs, but special casing a type of ROIs.

The rest should be small details!

sivborg:
Thank you for taking a look through! Regarding not being capable of editing existing polygons, an earlier commit actually had that possibility! However it used some "private" features in the `matplotlib.widgets.PolygonSelector` class, so I removed it to avoid any future problems if the `matplotlib.widgets.PolygonSelector)` implementation changes. However, by clicking inside an existing polygon, so it is highlighted in red, then pressing `delete`, you can remove and then redraw existing polygons. It is a bit clunky, but more safe wrt. the `matplotlib` features. Perhaps a PR in `matplotlib` could be useful? However, if you think we should restrict to a single polygon then that is also a possibility.

ericpre:
Why do you need to access private API? In the following example, you can edit the polygon:

```python
from matplotlib.widgets import PolygonSelector

fig, ax = plt.subplots()

selector2 = PolygonSelector(ax, lambda *args: None)
```

sivborg:
It is mostly a problem in the case where you have multiple polygons. It is possible to attach multiple `PolygonSelector`, however using several in one plot can make things very slow and do not work together smoothly, as far as I recall. `PolygonROI` can change to only implement one polygon at a time, though that would limit its use cases.

ericpre:
> It is mostly a problem in the case where you have multiple polygons. It is possible to attach multiple `PolygonSelector`, however using several in one plot can make things very slow and do not work together smoothly, as far as I recall. `PolygonROI` can change to only implement one polygon at a time, though that would limit its use cases.

This is an important point which is not consistent with other ROIs and I don't think that it is a good idea to introduce a different behaviour for this ROI. There is currently no pattern to handle multiple ROI in hyperspy and this is true that in some situations (ROI overlapping) multiple ROI doesn't play together. However, this is issue should be handle separately from the `PolygonROI` that you are adding here. I don't expect the multiple issue will be very difficult to address but it should be done in a separate PR and for all ROIs or least check that the implementation play well with other ROIs.

sivborg:
I am working on reimplementing an older commit where there was only one polygon in the ROI, for consistency. I am a bit busy currently but hope to have it done early next month!

ericpre:

ericpre:
@sivborg, do you need help with this PR?

sivborg:
Thank you for the consideration, @ericpre! However there is not too much to be done, so I can handle this myself once some more time opens up, which should not be too long. But thanks again!

sivborg:
Hello, I have some time now to finish up this PR after a busy spring and some LiberTEM development, so I have been implementing the final details. Another functionality I've worked a while on to add is a way to combine multiple `PolygonROI`, with the following syntax:
```python
roi = hs.roi.PolygonROI(((5,0),(120,0), (0,120)))
roi2 = hs.roi.PolygonROI(((150,180),(30,130), (180,0)))
roi3 = hs.roi.PolygonROI(((220,160),(160,160),(160,220), (220,220)))

s = hs.signals.Signal2D(np.zeros((256,256)))
mask = roi.boolean_mask(s.axes_manager, rois_to_combine=[roi2, roi3])
```
Which combines the defined ROIs into the following mask, as plotted with `imshow`:
Figure_1
These ROIs can also be defined by the widget. After this, it should be possible to crop the signal with this mask, though I am working on that.
The other ROIs do not have this combined ROIs as a functionality, as far as I've seen, but I think it is justified to add it to this one as it gives a lot more freedom to choose features - e.g. several diffraction spots in a TEM diffraction image. Moreover, it is a bit hidden from the interface so it should cause less confusion between ROIs.
Let me know if you have some thoughts on this!

ericpre:
> Hello, I have some time now to finish up this PR after a busy spring and some LiberTEM development, so I have been implementing the final details. Another functionality I've worked a while on to add is a way to combine multiple `PolygonROI`, with the following syntax:
> 
> ```python
> roi = hs.roi.PolygonROI(((5,0),(120,0), (0,120)))
> roi2 = hs.roi.PolygonROI(((150,180),(30,130), (180,0)))
> roi3 = hs.roi.PolygonROI(((220,160),(160,160),(160,220), (220,220)))
> 
> s = hs.signals.Signal2D(np.zeros((256,256)))
> mask = roi.boolean_mask(s.axes_manager, rois_to_combine=[roi2, roi3])
> ```

Great, this looks to be on the right track! :)

In principle, it would be good to be able to create mask from any ROI, would it make sense to use the following API (even if this PR only implement for one ROI):
```python
mask = hs.roi.create_mask_from_roi(s, rois=[roi1, roi2, roi3])
```

hyperspy/drawing/widget.py
sivborg: There is a possibility for `MPLWidgetBase` to inherit from `ResizableDraggableWidgetBase `. However, then `PolygonWidget` will need to implement resizing and dragging, which can be possible, but not obvious. For example, how can the `PolygonWidget` be resized? So I can do it if it seems worth it

And I looked a bit into `RangeWidget` inheriting from it, but it seems to implement a lot from `ResizableDraggableWidgetBase`. So the question is how to do that without losing a bit of functionality. Maybe multiple inheritance?

hyperspy/drawing/_widgets/polygon.py
sivborg: Yes indeed, these should be removed here! They were added to help development, but should be removed now that there are more conforming ways of doing this.

hyperspy/drawing/_widgets/polygon.py
sivborg: Indeed, this function is removed from a newer commit.

sivborg:
I have now implemented slicing with combined `PolygonROI`s! This works as follows:
```python
print(s) # <Signal2D, title: , dimensions: (256, 256|256, 256)>
roi = hs.roi.PolygonROI(((5,0),(120,0), (0,120)))
roi2 = hs.roi.PolygonROI(((150,180),(30,130), (180,0)))
roi3 = hs.roi.PolygonROI(((220,160),(160,160),(160,220), (220,220)))
sliced = roi.combine(s, rois=[roi2, roi3], axes=(2,3))
```
This did not take too many changes, as it only depended on combining different masks from each polygon.
However, there are some issues with plotting and the navigator, which I will elaborate on in a bit


> Great, this looks to be on the right track! :)
> 
> In principle, it would be good to be able to create mask from any ROI, would it make sense to use the following API (even if this PR only implement for one ROI):
> 
> ```python
> mask = hs.roi.create_mask_from_roi(s, rois=[roi1, roi2, roi3])
> ```

Thank you! And yes this seems to be a promising idea :) I quickly created implementation for several `PolygonROI`s, as it would only be a wrapper around the combination method in the class. The syntax is as follows:
```python
mask = hs.roi.create_mask_from_rois(rois = [roi, roi2, roi3], axes_manager=s.axes_manager, axes=(2,3))
```
The reason it needs `axes_manager` is so it can apply the scaling from the signal and get the correct limits. Although the default should work well in manycases, which assume the x- and y- scaling are 1.

I have also added a function for combining and slicing:
```python
sliced = hs.roi.combine_rois(s, rois=[roi, roi2, roi3], axes=(2,3))
```
Also a wrapper around the corresponding function in `PolygonROI`: `combine`.

sivborg:
Now the functionality of this PR seems to be at the place where I want it. So what remains for me is to take another close look, then this PR should be ready to review! What may need most scrutiny would perhaps be `PolygonWidget` and how it fits into the existing HyperSpy widget structure.

### Some final considerations

#### Navigator issues

Since the ROI masks out areas by replacing them with NaN, the navigator in `plot` may not display anything as its calculations also result in NaN. This is especially bad for lazy signals, as it might try to create the navigator image from a chunk where all values will be NaN. To remedy this, some changes in the navigator may be necessary, although that feels a bit out of the scope of this PR. As a current fix for non-lazy signals, one can replace all NaNs with a number:
```python
s.data = np.nan_to_num(s.data, copy=False)
```

#### Efficiency
`PolygonROI`, like `CircleROI`, copies the entire region of data in its bounding box. Therefore, this can require lots of RAM when using a large non-lazy signal and a large ROI. This seems to be inevitable considering it needs to mask out areas with NaN. In any case, it is preferable to use other ROIs like `RectangularROI` where that is sufficient.

#### Other known inaccuracies

Sometimes slicing with the ROI will have a row or column of NaN values at the edges which should have been sliced out. This is  due to making it more robust to not accidentally slice out parts of the ROI that are at the edges themselves. There is probably a good way to make sure this doesn't happen, but since the performance impact is that it only adds two more rows/columns I have deprioritised this issue.

sivborg:
@ericpre The PR should now be ready to review! It is mostly the `PolygonWidget` class I am unsure of, so would be great with a look there. Some code to get started:
```python
%matplotlib qt
import hyperspy.api as hs
import numpy as np
s = hs.signals.Signal2D(np.random.random((50,50, 20, 30)))
s.plot()
roi = hs.roi.PolygonROI()
wdg = roi.add_widget(s,axes=(2,3), color="blue")
```
```python
scrop = roi(s,axes=(2,3))
scrop.data = np.nan_to_num(scrop.data)
scrop.plot()
```
Otherwise please take a look at previous comments and examples!

I might also add some example files for how to use the new feature as well.

upcoming_changes/3030.new.rst
```diff
@@ -0,0 +1,4 @@
+Added :class:`~.api.roi.PolygonROI` and related functionality:
+- Added the ROI :class:`~.api.roi.PolygonROI` to create ROIs of arbitrary shape.
+- Added the widget :class:`~.api.widgets.PolygonWidget` to do this through the GUI.
+- Added functions :func:`~.api.widgets.combine_rois` and :func:`~.api.widgets.create_mask_from_rois` to use multiple ROIs to slice signals and create masks, respectively. Currently only for several :class:`~.api.roi.PolygonROI`.
```
ericpre: The formatting needs to be fixed:
image

Most likely something along the lines of:
```suggestion
Added :class:`~.api.roi.PolygonROI` and related functionality:

- Added the ROI :class:`~.api.roi.PolygonROI` to create ROIs of arbitrary shape.
- Added functions :func:`~.api.roi.combine_rois` and :func:`~.api.roi.create_mask_from_rois` to use multiple ROIs to slice signals and create masks, respectively. Currently only :class:`~.api.roi.PolygonROI` is supported.
```

hyperspy/drawing/_widgets/polygon.py
```diff
@@ -0,0 +1,115 @@
...
+class PolygonWidget(MPLWidgetBase):
+    """PolygonWidget is a widget for drawing an arbitrary
+    polygon, which can then be used as a region-of-interest.
+    The polygon can be moved by pressing 'shift' and clicking.
+    A polygon vertex can be moved by clicking its handle. If incomplete,
+    it is also necessary to press 'ctrl'.
+    The polygon can be deleted by pressing 'esc'.
+    """
+
+    def __init__(self, axes_manager, **kwargs):
```
ericpre, line 36: Can you add a test to cover the code of this class?

hyperspy/drawing/_widgets/polygon.py
```diff
@@ -0,0 +1,115 @@
...
+        """
+        Parameters
+        ----------
+        axes_manager : hyperspy.axes.AxesManager
+            The axes over which the `PolygonWidget` will interact.
+        """
+
+        super().__init__(axes_manager, **kwargs)
+
+        self.set_on(False)
```
ericpre: Why calling `set_on` here?

hyperspy/drawing/widget.py
ericpre: I mentioned `RangeWidget` because it is based on a matplotlib widgets and from reading the docstring, I would except that all widgets using matplotlib widgets should inherit from this class. Conceptually, it makes sense to have such a class to share functionality that wraps matplotlib widgets into hyperspy widgets, but it doesn't seem to work well and they could be two reasons for this:
1. this current implementation of the widgets using matplotlib widgets (`RangeWidget` and `PolygonWidget`) are not good/consistent and it needs to be refactored.
2. This class is actually not necessary - currently, it is only used by `PolygonWidget`.

In the interest of keeping the code consistent and maintainable, one of the two options above should be taken. I am slightly in favour of the first option, because it would allow to defer more of the widgets functionalities to matplotlib and migrated the `RectangleWidget` to use matplotlib widgets.

ericpre (commented):
In the interest of maintainability, `RangeWidget` and `PolygonWidget` should be implemented in a similar way because they both uses matplotlib widgets - the `set_on` and `_add_patch_to`/`set_mpl_ax` methods should be used in a similar fashion and this is not in this PR.

Tests needs to be added for the code that is not currently not covered, see annotations on github.

hyperspy/drawing/_widgets/polygon.py
sivborg: I have added some! Though the non-GUI interface of the class is a bit limited, so not many tests included.

hyperspy/drawing/_widgets/polygon.py
sivborg: If this is not called, the `is_on` property will be `True` before attaching to a plot. When a window is closed with `WidgetBase.close,` this gets set to `False`, so I thought having it `False` before attaching would make more sense. Or is the opposite a more desirable behaviour?

sivborg:
Yes, I agree a refactoring might be necessary. It is hard to make `RangeWidget` and `PolygonWidget` have common ancestry due to the former inheriting from other base classes with added functionaliy like `ResizableDraggableWidgetBase`. There might be a solution with multiple inheritance, but that also does not seem very attractive. I can take a look this week at the issues, but after that I will not have time until August, so if anyone else would meanwhile like to look at the refactoring that would be welcome.

sivborg:
I have now refactored and removed the `MPLWidgetBase` class. So it should be ready to merge in this state, but I can also take some time later to look into refactoring this and `RangeWidget` to simplify code and maintenance. Or it might be work better suited for another PR. I will be unavailable to look at this for a bit, but if anyone else are interested in doing so they are welcome to!

sivborg:
I have now rebased and looked over the code again, and I think most things should have been addressed. So I think this is ready for another review.

sivborg:
Seems that there are some tests failing due to an older MPL version. The version in the test environments is `3.5.1`, however the full functionality of `PolygonSelector`, which `PolygonWidget` depends on, is only available from `3.6.0`. I can, however, see if I can make it work on older versions as well, then do a version check.

sivborg:
The PR should be feature complete at this point! I will likely go over again soon and update documentation, make a few new tests and perhaps make an example with this, but the code in the actual classes themselves should be ready for review.

hyperspy/drawing/_widgets/polygon.py
```diff
@@ -0,0 +1,168 @@
...
+            List of  `(x, y)` tuples of the vertices of the polygon to add.
+        """
+
+        if self.ax is not None and self.is_on:
+            if len(vertices) > 2:
+
+                if Version(matplotlib.__version__) >= Version("3.6"):
+                    self._widget.verts = list(vertices)
+                else:
+                    # Accessing some private members to be able to change the `matplotlib`
```
ericpre: I thought that we could keep support for older version of matplotlib, but as mentioned in #3450, it is more simpler to use matplotlib>=3.6. Except if there are objection to do that in #3450, it should be possible to remove this branch - maybe best to wait that #3450 is completed.

hyperspy/roi.py
```diff
@@ -1511,6 +1514,509 @@ def __call__(self, signal, out=None, axes=None, order=0):
...
+            intersections.sort()
+
+            for i in range(1, len(intersections), 2):
+                raster_start = round(intersections[i - 1] / x_scale) - min_index_x
+                raster_end = round(intersections[i] / x_scale) - min_index_x
+                mask[row, raster_start : raster_end + 1] = True
+
+        return mask
+
+    def boolean_mask(
```
ericpre: Should it be made private to simplify the API and favour the use `hs.roi.create_mask_from_rois`?

hyperspy/roi.py
```diff
@@ -1511,6 +1514,509 @@ def __call__(self, signal, out=None, axes=None, order=0):
...
+        if not self.is_valid():
+            self.vertices = old_vertices
+            raise ValueError(
+                f"`vertices` is not an empty list or a list of fully defined two-dimensional \
+                    points with at least three entries:\n{vertices}"
+            )
+        if self.widgets:
+            self._update_widgets()
+
+    def _apply_roi(self, signal, inverted=False, out=None, axes=None, other_rois=None):
```
ericpre: I think that this need a small restructure, because API `PolygonROI._apply_roi` shouldn't have dependency/awareness of other rois. I guess that you done it that way here as a convenience when combining ROI.
I suspect that the share functionality should move in a separate (private?) function which would then be used by `combine_roi` and `PolygonROI.__call__`.

hyperspy/drawing/_widgets/polygon.py
```diff
@@ -0,0 +1,168 @@
...
+        ----------
+        axes_manager : hyperspy.axes.AxesManager
+            The axes over which the `PolygonWidget` will interact.
+        """
+
+        super().__init__(axes_manager, **kwargs)
+
+        if self.axes_manager is not None:
+            if self.axes_manager.navigation_dimension > 1:
+                self.axes = self.axes_manager.navigation_axes[0:2]
```
ericpre, line 49: Please add some tests that is not currently covered by the test suite, use the codecov yellow annotation on github to find out what is not covered.

hyperspy/drawing/_widgets/polygon.py
```diff
@@ -0,0 +1,168 @@
...
+                    self._widget._xs.append(self._widget._xs[0])
+                    self._widget._ys.append(self._widget._ys[0])
+
+                    self._widget._selection_completed = True
+                    self._widget.set_visible(True)
+                    self._widget._draw_polygon()
+                self._onselect(vertices)
+            self.ax.figure.canvas.draw_idle()
+
+    def get_vertices(self):
```
ericpre, line 136: Wondering if this should be a property/setter called `vertices`, to have a similar interface to other ROIs, for example, `SpanROI.left`, etc. The idea would be to get/set the parameters of ROI in a similar by using the name of the parameter - here `vertices`.

ericpre (commented):
@sivborg, it is getting there but not quite ready - see comments. I will need to have another look into more details.

It would be good to add an example in the galleries (something similar to https://hyperspy.org/hyperspy-doc/current/auto_examples/region_of_interest/ROI_navigator.html#sphx-glr-auto-examples-region-of-interest-roi-navigator-py) and add the link to this example in the release note.

The following doesn't add the widget to the plot while it should!
```python
import hyperspy.api as hs
import numpy as np

s = hs.signals.Signal1D(np.ones((50, 60, 4)))
roi = hs.roi.PolygonROI([(20, 20), (36, 20), (26, 36)])

s.plot()
s_roi = roi.interactive(s)
```

To fix CI, I rebased this branch and push some commits (which I removed since) to tried to fix some issues but actually it got more involved and opted to do these fixes in a separate PR (#3450). Once #3450 is merged, we would need to rebase this branch.

hyperspy/drawing/_widgets/polygon.py
ericpre: This branch now includes #3450 and the minimum supported version of matplotlib is 3.6.

sivborg: Yes, this simpifies quite a bit! So I have now removed the backwards compatibility.

hyperspy/roi.py
sivborg: The reason for including the other ROIs here is that it's more efficient to rasterize all of them together and then slice the signal once. But some other ways of doing this could be:
- In `combine_roi`, slice the signal once for each `PolygonROI` and stitch them together. Should be fairly straightforward but maybe a bit inefficient.
- Let `_apply_roi` take in a boolean mask it can combine with the polygon, which can be created with `hs.roi.create_mask_from_rois`. Then it only needs to slice once without needing the other ROIs directly. Would then also combine with other types of ROIs.
I think the latter would be best due to efficiency and also it would be easier to coordinate the parameters, such as x- and y-limits for cropping the finished signal. I will take a look a bit later on.

hyperspy/drawing/_widgets/polygon.py
sivborg: Yes, this is a good idea and should be implemented for consistency. So I have done just that!

sivborg:
@ericpre Thank you for the review. Hopefully we can finish up the features soon as I think we are very close as well.
> It would be good to add an example in the galleries (something similar to https://hyperspy.org/hyperspy-doc/current/auto_examples/region_of_interest/ROI_navigator.html#sphx-glr-auto-examples-region-of-interest-roi-navigator-py) and add the link to this example in the release note.

Yes, I agree with that. I will make a short example after I have ironed out some kinks in the code.

> The following doesn't add the widget to the plot while it should!
> 
> ```python
> import hyperspy.api as hs
> import numpy as np
> 
> s = hs.signals.Signal1D(np.ones((50, 60, 4)))
> roi = hs.roi.PolygonROI([(20, 20), (36, 20), (26, 36)])
> 
> s.plot()
> s_roi = roi.interactive(s)
> ```
Indeed! There was a bug in checking for the blit support, but should have been fixed now.

ericpre:
In the interest of consistency with other ROIs, would it be possible to get the following to work by setting some default value?

```python
import hyperspy.api as hs
import numpy as np

s = hs.signals.Signal1D(np.ones((50, 60, 4)))
roi = hs.roi.PolygonROI()

s.plot()
s_roi = roi.interactive(s)
```

sivborg:
> In the interest of consistency with other ROIs, would it be possible to get the following to work by setting some default value?
> 
> ```python
> import hyperspy.api as hs
> import numpy as np
> 
> s = hs.signals.Signal1D(np.ones((50, 60, 4)))
> roi = hs.roi.PolygonROI()
> 
> s.plot()
> s_roi = roi.interactive(s)
> ```

Yes, I think should be very accomplishable using similar code to the `CircleROI` default implementation. Then it should be possible to press escape to reset the polygon and start constructing a new one.
What I hope will not be a problem is the user not figuring out to press escape to make their own shape. They might even think that the shape they are presented with is the only one possible, thus missing out on the flexibility. The point of this ROI is its flexibility, so, as an exception, might it be relevant to actually not give a default implementation? Instead put the user straight into the polygon construction?

hyperspy/drawing/_widgets/polygon.py
sivborg: I have added a lot more tests now. Some edge cases are still not covered, however, but the main code is covered more thoroughly.

hyperspy/roi.py
sivborg: Yes, I agree this is a good thing to do. I simplified the function to `hs.roi.mask_from_rois` as well.

hyperspy/roi.py
sivborg: I changed it so it takes in the vertices of the other `PolygonROI`s instead of them directly. I found this to be both the most efficient and simple in terms of parameters. In any case, combining ROIs is now only done from the two non-member functions `combine_rois` and `mask_from_rois`, where you supply the ROIs directly.

ericpre:
pre-commit.ci autofix

ericpre:
I expected that it would be possible to add vertices manually after creating the polygon, in which case, the number of vertices could be changed once a polygon is completed, but [this isn't to be the case](https://matplotlib.org/stable/api/widgets_api.html#matplotlib.widgets.PolygonSelector) - vertices can only be removed, not added. This is a shame and this is something that would be useful in matplotlib.

Can you add example in the gallery? I will try to review in the coming days.

sivborg:
> I expected that it would be possible to add vertices manually after creating the polygon, in which case, the number of vertices could be changed once a polygon is completed, but [this isn't to be the case](https://matplotlib.org/stable/api/widgets_api.html#matplotlib.widgets.PolygonSelector) - vertices can only be removed, not added. This is a shame and this is something that would be useful in matplotlib.

Yes, that is true. Although it might not be too difficult to add at some point in the future.

> Can you add example in the gallery? I will try to review in the coming days.

I have added an example to use PolygonROI in the examples folder. Although this example is only on signal axes, so perhaps it would be interesting to expand it to show how it can be used in 4D, by slicing either the signal or navigation space.

ericpre (commented):
I push a few commits to:
- fix and improve documentation - readthedocs is still failing which is surprising
- start to add some tests - more are needed to increase the coverage in the `roi.py` module
- improve some code.

When using multiple selectors, they all moved together and this behaviour should be fixed in https://github.com/matplotlib/matplotlib/pull/29027.

What is left is to increase the test coverage as highlighted by the codecov warning in the "files changes" tab.

sivborg:
I have now added some further test coverage, with only a couple `NotImplementedError`s to remain. Now, however, the tests seem to fail due to some deprecated functionality in the `prettytable` package.
I think also there might have snuck a couple bugs into the polygon building, so I will see if I can get a handle on them.

sivborg:
I should have now fixed the bug in the polygon construction. It was created since the update of the ROI was moved from the `PolygonSelector`'s `onselect` callback to the Matplotlib `_onmove` callback. The former is only called when the polygon is finished constructing, while the latter is called continuously. So the way to fix it was to check more thoroughly if the new vertices are valid and not set them until the polygon is constructed.

ericpre (approved):

ericpre:
Thank you @sivborg, it wasn't an easy task to implement this type of ROI!

If there are things that need tweaking, this can be done in follow up PRs.

sivborg:
And thank you for the feedback and help, @ericpre !
While a challenge, it came with both valuable experiences and newfound skills :) No doubt I can use this competence in Git and software development in the future!
And I do think there are some tweaks to be had - but we will see what is most immediate.
Until next time!

Issue #2949, opened 2022-05-27
Benchmarking multifit


thomasaarholt:
I did a bit of benchmarking today using [pyinstrument](https://github.com/joerick/pyinstrument) following a possible performance regression on 1.6.5 vs 1.7.0 mentioned on [gitter](https://gitter.im/hyperspy/hyperspy?at=628e39a4da83520ac34af88a). The speed difference turned out to be due to the user, but I still found something interesting:

This [benchmark example](https://github.com/hyperspy/hyperspy/files/8789146/bench.py.zip) takes almost 2x the time on 1.7.0 as for 1.6.5.

Until we change the api in 2.0, [`is_binned`](https://github.com/hyperspy/hyperspy/blob/5f15a5497d27718adda17a7e89b79fff88207e69/hyperspy/misc/utils.py#L1541) checks whether the binning is set in the signal metadata or as a `axes.is_binned` parameter. The lookup in the metadata appears to consume ~25% of the time spent in the following benchmark multifit example. Removing the lookup to just the 2.0 api 

The multifit call of the benchmark takes:
- 84 sec on 1.6.5
- 160 sec on 1.7.0
- 104 sec if I use the new 2.0 api for `is_binned`

[profile log on 1.7.0](https://gist.github.com/thomasaarholt/1b2109aa0420e57193bffcf15567ea01)
[profile log on 1.6.5](https://gist.github.com/thomasaarholt/4122d2c11f4e9264e531cf48bea0be80)
[profile log on 1.7.0 with 2.0 api only (removing the if)](https://gist.github.com/thomasaarholt/c2ffe616de5f3c9c328c8ff3b9774cc7)

I just wanted to point this out here, since 25% can be a significant amount of time if one is doing very long fits.

**One other regression:**
16 of the 160 sec, or ~10% of time is spent on [`is_cupy_array`](https://github.com/hyperspy/hyperspy/blob/5f15a5497d27718adda17a7e89b79fff88207e69/hyperspy/misc/utils.py#L1552).

ericpre:
Good find, using caching should fix the performance regression! There may be room for other improvement.

I didn't know about `pyinstrument` and the output is readable... unlike other profiling tool I have used (or tried to!) in the past.

PR #3031, opened 2022-09-23
Handling Axis attributes with @property decorator

### LINKED ISSUE #3121: Protect `axis` attribute for uniform data axes
#### Describe the bug
Currently, the `axis` attribute of uniform data axes can be changed (after initialization of the axis), though it should be defined by `offset` and `scale` only. Instead, it should be a protected attribute that can be changed only if `offset` and `scale` are changed, but not directly by user input. An example is that users not understanding the difference between `DataAxis` and `UniformDataAxis`, might actually change an axis to a non-uniform one just by providing the `axis` array without converting the axes type, which can lead to problems in a later analysis.

Would be solved by #3031

#### To Reproduce
```
S=hs.signals.Signal1D(np.arange(100))
S.axes_manager[0].axis = np.ones(100)
```

`S.axes_manager` and `S.axes_manager[0].axis` are now in contradiction to each other.

#### Expected behavior
Throw an error when trying to manually change the `axis` vector for a uniform or functional data axis.

#### Python environement:
 - HyperSpy version: 1.7.4
 - Python version: 3.11

CSSFrancis:
### Description of the change
This set of changes supersedes #3012 and #2876 in the hope of creating small incremental changes to how Axes and slicing occurs before fully supporting vector signals.  @ericpre sorry about the confusion, but the changes to how the slicing works are rather larger than anticipated so I wanted to do this slow and incrementally.  

This change better defines what are the key attributes of an axes and focuses on only allowing changes to those attributes. 

- BaseDataAxis: The base class it has only a name and unit as well as attributes related to interfacing with the AxesManager.
- DataAxis: The most flexible of the Axes.  It has an axes attribute which is a 1 dimensional array. This can be increasing order, decreasing order or arbitrarily ordered.  (Note the last case does not support any kind of indexing yet)
- FunctionalDataAxis: Has an axis property that is calculated from x and the given function.
- UniformDataAxis: Has an axis property calculated from the size, offset and scale.
- AxesManager: Properties in the axes manager are directly pulled from the axes in the list.

These changes help to stop a lot of small bugs within the axes code.  Things like being able to set the size of a DataAxis don't make sense when the size is directly related to the axis.  Similarly with the UniformDataAxis, directly setting the Axis should throw an error. 

This also adds in some limited support for unordered (or unorderable axes) which are necessary for vector applications.  

### Progress of the PR
- [x] add tests,
- [x] ready for review.

```python
import hyperspy.api as hs

ax = hs.axes.DataAxis(["a", "b", "c"]) # No longer throws an error
ax.size = 10 # Throws AttributeError

ax = hs.axes.UniformDataAxis(size=10, scale=3, offset=1) 
ax.size = 10
ax.axis = [10,12,13,14] # Throws AttributeError
```

### DIFF
--- a/hyperspy/_signals/lazy.py
+++ b/hyperspy/_signals/lazy.py
@@ -333,7 +333,13 @@ class LazySignal(BaseSignal):
         dcshape = dc.shape
         for _axis in self.axes_manager._axes:
             if _axis.index_in_array < len(dcshape):
-                _axis.size = int(dcshape[_axis.index_in_array])
+                if _axis.size != int(dcshape[_axis.index_in_array]):
+                    if _axis.is_uniform:
+                        _axis.size = int(dcshape[_axis.index_in_array])
+                    else:
+                        _logger.warning(f"Axis {_axis} converted to a UniformDataAxis")
+                        _axis.convert_to_uniform_axis()
+                        _axis.size = int(dcshape[_axis.index_in_array])
 
         if axis is not None:
             need_axes = self.axes_manager[axis]
--- a/hyperspy/axes.py
+++ b/hyperspy/axes.py
@@ -1045,15 +1045,6 @@ class FunctionalDataAxis(DataAxis):
     def size(self):
         return self.x.size
 
-    def update_axis(self):
-        kwargs = {}
-        for kwarg in self.parameters_list:
-            kwargs[kwarg] = getattr(self, kwarg)
-        self.axis = self._function(x=self.x.axis, **kwargs)
-        # Set not valid values to np.nan
-        self.axis[np.logical_not(np.isfinite(self.axis))] = np.nan
-        self.size = len(self.axis)
-
     def update_from(self, axis, attributes=None):
         """Copy values of specified axes fields from the passed AxesManager.
 
@@ -1078,33 +1069,12 @@ class FunctionalDataAxis(DataAxis):
         raise TypeError("This function works only for uniform axes.")
 
     def get_axis_dictionary(self):
-        d = super().get_axis_dictionary()
-        d["expression"] = self._expression
-        d.update(
-            {
-                "size": _parse_axis_attribute(self.size),
-            }
-        )
-        d.update(
-            {
-                "x": self.x.get_axis_dictionary(),
-            }
-        )
-        for kwarg in self.parameters_list:
-            d[kwarg] = getattr(self, kwarg)
+        d = super(DataAxis, self).get_axis_dictionary()
+        d['expression'] = self._expression
+        d.update({'size': _parse_axis_attribute(self.size), })
+        d.update({'x': self.x.get_axis_dictionary(), })
+        d.update(self.parameters)
         return d
-
-    def convert_to_non_uniform_axis(self):
-        """Convert to a non-uniform axis."""
-        d = super().get_axis_dictionary()
-        axes_manager = self.axes_manager
-        d["_type"] = "DataAxis"
-        self.__class__ = DataAxis
-        self.__init__(**d, axis=self.axis)
-        del self._expression
-        del self._function
-        self.remove_trait("x")
-        self.axes_manager = axes_manager
 
     def crop(self, start=None, end=None):
         """Crop the axis in place.
@@ -1128,12 +1098,8 @@ class FunctionalDataAxis(DataAxis):
         calibration, the `start`/`end` tuple also has to be in descending
         order.
         """
-
         slice_ = self._get_array_slices(slice(start, end))
         self.x.crop(start=slice_.start, end=slice_.stop)
-        self.size = self.x.size
-        self.update_axis()
-
     crop.__doc__ = DataAxis.crop.__doc__
 
     def _slice_me(self, slice_):
@@ -1151,11 +1117,34 @@ class FunctionalDataAxis(DataAxis):
         """
         my_slice = self._get_array_slices(slice_)
         self.x._slice_me(my_slice)
-        self.update_axis()
         return my_slice
 
+    def convert_to_uniform_axis(self):
+        """Convert to an uniform axis."""
+        scale = (self.high_value - self.low_value) / (self.size-1)
+        offset = self.low_value
+        d = self.get_axis_dictionary()
+        if len(self.axis) > 1:
+            scale_err = max(self.axis[1:] - self.axis[:-1]) - scale
+            _logger.warning('The maximum scale error is {}.'.format(scale_err))
+        d["_type"] = "UniformDataAxis"
 
-class UniformDataAxis(BaseDataAxis, UnitConversion):
+        if self.axes_manager is not None:
+            self.axes_manager[self] = UniformDataAxis(**d, scale=scale, offset=offset)
+        else:
+            return UniformDataAxis(**d, scale=scale, offset=offset)
+
+    def convert_to_non_uniform_axis(self):
+        """Convert to a non-uniform axis."""
+        d = super().get_axis_dictionary()
+        d["_type"] = 'DataAxis'
+        if self.axes_manager is not None:
+            self.axes_manager[self] = DataAxis(**d)
+        else:
+            return DataAxis(**d)
+
+
+class UniformDataAxis(DataAxis, UnitConversion):
     """DataAxis class for a uniform axis defined through a ``scale``, an
     ``offset`` and a ``size``.
 
--- a/hyperspy/axes.py
+++ b/hyperspy/axes.py
@@ -926,7 +926,8 @@ class DataAxis(BaseDataAxis):
                                    scale=scale,
                                    offset=offset)
 
-class FunctionalDataAxis(BaseDataAxis):
+
+class FunctionalDataAxis(DataAxis):
     """DataAxis class for a non-uniform axis defined through an ``expression``.
 
     A `FunctionalDataAxis` is defined based on an ``expression`` that is
@@ -942,12 +943,25 @@ class FunctionalDataAxis(BaseDataAxis):
     ``size``. However, it can also be initialized with custom ``offset`` and
     ``scale`` values. Alternatively, it can be a non-uniform `DataAxis`.
 
+    A `FunctionalDataAxis` has its size set by the underlying  `BaseDataAxis` ``x``.
+    Other values are calculated from the ``axis`` attribute.
+
     Parameters
     ----------
     expression: str
         SymPy mathematical expression defining the axis.
     x : BaseDataAxis
         Defines x-values at which `expression` is evaluated.
+
+     Attributes
+    ----------
+    parameters: dict
+        A dictionary representing the additional varibles and their values. This
+        is changed using a.parameters["b"]=10.  This will automatically update the
+        axis value
+    axis: ndarray
+        The caculated axis as determined by the expression evaulated at every point in `x`
+
 
     Examples
     --------
@@ -974,61 +988,62 @@ class FunctionalDataAxis(BaseDataAxis):
     'a': 100,
     'b': 10}
     """
-
-    def __init__(
-        self,
-        expression,
-        x=None,
-        index_in_array=None,
-        name=None,
-        units=None,
-        navigate=False,
-        size=1,
-        is_binned=False,
-        **parameters,
-    ):
-        super().__init__(
+    x = t.Instance(DataAxis)
+    parameters = t.Dict()
+    axis = t.Property(t.Array,
+                      depends_on="x.axis, parameters",)
+    def __init__(self,
+                 expression,
+                 x=None,
+                 index_in_array=None,
+                 name=None,
+                 units=None,
+                 navigate=False,
+                 size=1,
+                 is_binned=False,
+                 **parameters):
+        super(DataAxis, self).__init__(
             index_in_array=index_in_array,
             name=name,
             units=units,
             navigate=navigate,
             is_binned=is_binned,
-            **parameters,
-        )
-        # These trait needs to added dynamically to be removed when necessary
-        self.add_trait("x", t.Instance(BaseDataAxis))
+            **parameters)
         if x is None:
-            if size is t.Undefined:
-                raise ValueError("Please provide either `x` or `size`.")
             self.x = UniformDataAxis(scale=1, offset=0, size=size)
         else:
             if isinstance(x, dict):
                 self.x = create_axis(**x)
             else:
                 self.x = x
-                self.size = self.x.size
         self._expression = expression
         if "_type" in parameters:
             del parameters["_type"]
         # Compile function
         expr = _parse_substitutions(self._expression)
         variables = ["x"]
-        expr_parameters = [
-            symbol for symbol in expr.free_symbols if symbol.name not in variables
-        ]
+        expr_parameters = [symbol for symbol in expr.free_symbols
+                           if symbol.name not in variables]
+
         if set(parameters) != set([parameter.name for parameter in expr_parameters]):
             raise ValueError(
                 "The values of the following expression parameters "
                 f"must be given as keywords: {set(expr_parameters) - set(parameters)}"
             )
         self._function = lambdify(
-            variables + expr_parameters, expr.evalf(), dummify=False
-        )
-        for parameter in parameters.keys():
-            self.add_trait(parameter, t.CFloat(parameters[parameter]))
-        self.parameters_list = list(parameters.keys())
-        self.update_axis()
-        self.on_trait_change(self.update_axis, self.parameters_list)
+            variables + expr_parameters, expr.evalf(), dummify=False)
+        self.parameters = parameters
+
+    @t.cached_property
+    def _get_axis(self):
+        new_axis = self._function(x=self.x.axis, **self.parameters)
+        # Set not valid values to np.nan
+        new_axis[np.logical_not(np.isfinite(new_axis))] = np.nan
+        return new_axis
+
+    @property
+    def size(self):
+        return self.x.size
 
     def update_axis(self):
         kwargs = {}
--- a/hyperspy/axes.py
+++ b/hyperspy/axes.py
@@ -1543,6 +1543,7 @@ class AxesManager(t.HasTraits):
     _step = t.Int(1)
 
     def __init__(self, axes_list):
+        self._ragged =False
         super().__init__()
         self.events = Events()
         self.events.indices_changed = Event(
@@ -1578,7 +1579,6 @@ class AxesManager(t.HasTraits):
             self.remove(self._axes)
         self.create_axes(axes_list)
 
-        self._update_attributes()
         self._update_trait_handlers()
         self.iterpath = "serpentine"
         self._ragged = False
@@ -1588,16 +1588,35 @@ class AxesManager(t.HasTraits):
         return self._ragged
 
     def _update_trait_handlers(self, remove=False):
-        things = {
-            self._on_index_changed: "_axes.index",
-            self._on_slice_changed: "_axes.slice",
-            self._on_size_changed: "_axes.size",
-            self._on_scale_changed: "_axes.scale",
-            self._on_offset_changed: "_axes.offset",
-        }
+        self.on_trait_change(self._on_index_changed,
+                             name='_axes._index', remove=remove)
+        self.on_trait_change(self._on_axes_changed,
+                             name='_axes._size', remove=remove)
+        self.on_trait_change(self._on_axes_changed,
+                             name='_axes._scale', remove=remove)
+        self.on_trait_change(self._on_axes_changed,
+                             name='_axes._offset', remove=remove)
 
-        for k, v in things.items():
-            self.on_trait_change(k, name=v, remove=remove)
+    def __setitem__(self, y, value):
+        """x.__getitem__(y) <==> x[y]
+        """
+        if isinstance(y, str) or not np.iterable(y):
+            ax = self._axes_getter(y)
+            self.axes[self.axes.index(ax)] = value
+            return
+        else:
+            axes = [self._axes_getter(ax) for ax in y]
+            indices = [self.axes.index(ax) for ax in axes]
+            for ind, v in zip(axes, indices):
+                self.axes[ind] = v
+
+    @property
+    def axes(self):
+        return self._axes
+
+    @axes.setter
+    def axes(self, value):
+        self._axes = value
 
     def _get_positive_index(self, axis):
         if axis < 0:
@@ -1811,12 +1830,14 @@ class AxesManager(t.HasTraits):
         """
         self._axes[index_in_axes_manager] = axis
 
-    def _update_max_index(self):
-        self._max_index = 1
+    @property
+    def _max_index(self):
+        _max_index = 1
         for i in self.navigation_shape:
-            self._max_index *= i
-        if self._max_index != 0:
-            self._max_index -= 1
+            _max_index *= i
+        if _max_index != 0:
+            _max_index -= 1
+        return _max_index
 
     @property
     def iterpath(self):
@@ -1963,28 +1984,11 @@ class AxesManager(t.HasTraits):
 
     def _append_axis(self, **kwargs):
         axis = create_axis(**kwargs)
-        axis.axes_manager = self
         self._axes.append(axis)
 
-    def _on_index_changed(self):
-        self._update_attributes()
-        self.events.indices_changed.trigger(obj=self)
-
-    def _on_slice_changed(self):
-        self._update_attributes()
-
-    def _on_size_changed(self):
-        self._update_attributes()
-        self.events.any_axis_changed.trigger(obj=self)
-
-    def _on_scale_changed(self):
-        self.events.any_axis_changed.trigger(obj=self)
-
-    def _on_offset_changed(self):
-        self.events.any_axis_changed.trigger(obj=self)
-
-    def convert_units(self, axes=None, units=None, same_units=True, factor=0.25):
-        """Convert the scale and the units of the selected axes. If the unit
+    def convert_units(self, axes=None, units=None, same_units=True,
+                      factor=0.25):
+        """ Convert the scale and the units of the selected axes. If the unit
         of measure is not supported by the pint library, the scale and units
         are not changed.
 
@@ -2119,87 +2123,83 @@ class AxesManager(t.HasTraits):
         if changes:
             self.events.any_axis_changed.trigger(obj=self)
 
-    def _update_attributes(self):
-        getitem_tuple = []
-        values = []
-        signal_axes = ()
-        navigation_axes = ()
+    @observe("_axes")
+    @observe(trait("_axes", notify=False).list_items())
+    def add_axis_manger(self, event):
+        """
+        The first case is triggered by someone setting the axes value.
+        Something like `am.axes=[axis1, axis2]`
+
+        The second observation is triggers when something inplace occurs.
+        I.e. `am.axes[1]= axis1`
+        """
         for axis in self._axes:
-            # Until we find a better place, take property of the axes
-            # here to avoid difficult to debug bugs.
             axis.axes_manager = self
-            if axis.slice is None:
-                getitem_tuple += (axis.index,)
-                values.append(axis.value)
-                navigation_axes += (axis,)
-            else:
-                getitem_tuple += (axis.slice,)
-                signal_axes += (axis,)
-        if not signal_axes and navigation_axes:
-            getitem_tuple[-1] = slice(axis.index, axis.index + 1)
 
-        self._signal_axes = signal_axes[::-1]
-        self._navigation_axes = navigation_axes[::-1]
-        self._getitem_tuple = tuple(getitem_tuple)
+    def _on_index_changed(self):
+        self.events.indices_changed.trigger(obj=self)
 
-        if len(self.signal_axes) == 1 and self.signal_axes[0].size == 1:
-            self._signal_dimension = 0
-        else:
-            self._signal_dimension = len(self.signal_axes)
-        self._navigation_dimension = len(self.navigation_axes)
+    def _on_index_changed(self):
+        self.events.indices_changed.trigger(obj=self)
 
-        self._signal_size = np.prod(self.signal_shape) if self.signal_shape else 0
-        self._navigation_size = (
-            np.prod(self.navigation_shape) if self.navigation_shape else 0
-        )
-
-        self._update_max_index()
+    def _on_index_changed(self):
+        self.events.indices_changed.trigger(obj=self)
 
     @property
     def signal_axes(self):
         """The signal axes as a tuple."""
-        return self._signal_axes
+        return tuple([a for a in self._axes if not a.navigate][::-1])
 
     @property
     def navigation_axes(self):
         """The navigation axes as a tuple."""
-        return self._navigation_axes
+        return tuple([a for a in self._axes if a.navigate][::-1])
 
     @property
     def signal_shape(self):
         """The shape of the signal space."""
-        return tuple([axis.size for axis in self._signal_axes])
+        return tuple([axis.size if hasattr(axis, "size")
+                       else None for axis in self.signal_axes])
 
     @property
     def navigation_shape(self):
         """The shape of the navigation space."""
         if self.navigation_dimension != 0:
-            return tuple([axis.size for axis in self._navigation_axes])
+            return tuple([axis.size for axis in self.navigation_axes])
         else:
             return ()
 
     @property
     def signal_size(self):
         """The size of the signal space."""
-        return self._signal_size
+        if self.signal_shape == ():
+            return 0
+        else:
+            return np.prod(self.signal_shape)
 
     @property
     def navigation_size(self):
         """The size of the navigation space."""
-        return self._navigation_size
+        if self.signal_shape == ():
+            return 0
+        else:
+            return np.prod(self.signal_shape)
 
     @property
     def navigation_dimension(self):
         """The dimension of the navigation space."""
-        return self._navigation_dimension
+        return len(self.navigation_axes)
 
     @property
     def signal_dimension(self):
         """The dimension of the signal space."""
-        return self._signal_dimension
+        if len(self.signal_axes) == 1 and self.signal_axes[0].size == 1:
+            return 0
+        else:
+            return len(self.signal_axes)
 
     def _set_signal_dimension(self, value):
-        if len(self._axes) == 0 or self._signal_dimension == value:
+        if len(self._axes) == 0 or self.signal_dimension == value:
             # Nothing to be done
             return
         elif self.ragged and value > 0:
--- a/hyperspy/axes.py
+++ b/hyperspy/axes.py
@@ -887,7 +887,6 @@ class DataAxis(BaseDataAxis):
 
         slice_ = self._get_array_slices(slice(start, end))
         self.axis = self.axis[slice_]
-        self.size = len(self.axis)
 
     def convert_to_uniform_axis(self):
         """Convert to an uniform axis."""
@@ -1188,7 +1187,7 @@ class UniformDataAxis(DataAxis, UnitConversion):
             )
         self.scale = scale
         self.offset = offset
-        self._size = size
+        self.size = size
         self._is_uniform = True
 
     @t.cached_property
--- a/hyperspy/axes.py
+++ b/hyperspy/axes.py
@@ -539,24 +539,8 @@ class DataAxis(BaseDataAxis):
         """
         my_slice = self._get_array_slices(slice_)
         self.axis = self.axis[my_slice]
-        self.update_axis()
         return my_slice
 
-    def update_axis(self):
-        """Set the value of an axis. The axis values need to be ordered.
-
-        Raises
-        ------
-        ValueError
-            If the axis values are not ordered.
-
-        """
-        if len(self.axis) > 1:
-            if isinstance(self.axis, list):
-                self.axis = np.asarray(self.axis)
-            if self._is_increasing_order is None:
-                raise ValueError("The non-uniform axis needs to be ordered.")
-        self.size = len(self.axis)
 
     def get_axis_dictionary(self):
         d = super().get_axis_dictionary()
@@ -1246,7 +1230,7 @@ class UniformDataAxis(DataAxis, UnitConversion):
         return my_slice
 
     def get_axis_dictionary(self):
-        d = super(DataAxis).get_axis_dictionary()
+        d = super(DataAxis, self).get_axis_dictionary()
         d.update({'size': self.size,
                   'scale': self.scale,
                   'offset': self.offset})
@@ -2139,11 +2123,8 @@ class AxesManager(t.HasTraits):
     def _on_index_changed(self):
         self.events.indices_changed.trigger(obj=self)
 
-    def _on_index_changed(self):
-        self.events.indices_changed.trigger(obj=self)
-
-    def _on_index_changed(self):
-        self.events.indices_changed.trigger(obj=self)
+    def _on_axes_changed(self):
+        self.events.any_axis_changed.trigger(obj=self)
 
     @property
     def signal_axes(self):
--- a/hyperspy/axes.py
+++ b/hyperspy/axes.py
@@ -1045,7 +1045,7 @@ class FunctionalDataAxis(DataAxis):
 
         """
         if attributes is None:
-            attributes = self.parameters_list + ["_expression", "x"]
+            attributes = ["parameters", ]
         return super().update_from(axis, attributes)
 
     def calibrate(self, *args, **kwargs):
@@ -1590,7 +1590,7 @@ class AxesManager(t.HasTraits):
         else:
             axes = [self._axes_getter(ax) for ax in y]
             indices = [self.axes.index(ax) for ax in axes]
-            for ind, v in zip(axes, indices):
+            for ind, v in zip(indices, value):
                 self.axes[ind] = v
 
     @property
--- a/hyperspy/axes.py
+++ b/hyperspy/axes.py
@@ -455,16 +455,37 @@ class BaseDataAxis(t.HasTraits):
 
 
 class DataAxis(BaseDataAxis):
-    """DataAxis class for a non-uniform axis defined through an ``axis`` array.
+    """DataAxis class for any non-uniform axis defined through an ``axis`` array.
 
-    The most flexible type of axis, where the axis points are directly given by
-    an array named ``axis``. As this can be any array, the property
-    ``is_uniform`` is automatically set to ``False``.
+       The most flexible type of axis, where the axis points are directly given by
+       an array named ``axis``. These points can be an abitrary array of integers/floats,
+       an array of strings or even an array of objects.
+
+       All other properties are defined from the ``axis`` array. Notable setting the ``size``
+       of a DataAxis object will result in an error.
+
+       The array can be indexed using direct references to objects in the axis array,
+       integer indexes or if `_is_increasing_order` is not None, float values.
+
+       As this can be any array, the property
+       ``is_uniform`` is automatically set to ``False``.
 
     Parameters
     ----------
     axis : numpy array or list
         The array defining the axis points.
+
+    Attributes
+    ----------
+    size
+    high_index
+    low_index
+    high_value
+    low_value
+    _is_increasing_order
+    index : int
+        The current index for the axis. This is used for interactive plotting and
+        model fitting
 
     Examples
     --------
@@ -480,6 +501,8 @@ class DataAxis(BaseDataAxis):
     'is_binned': False,
     'axis': array([  0,   1,   4,   9,  16,  25,  36,  49,  64,  81, 100])}
     """
+    axis = t.Array()  # Axis added as trait to track changes to axis
+    _index = t.Int(0)
 
     def __init__(
         self,
@@ -540,6 +563,53 @@ class DataAxis(BaseDataAxis):
         d.update({"axis": self.axis})
         return d
 
+    @property
+    def low_value(self):
+        if self._is_increasing_order is True:
+            return self.axis[0]
+        elif self._is_increasing_order is False:
+            return self.axis[-1]
+        else:
+            raise NotImplementedError("The axis is not ordered so there is no high/low value")
+    @property
+    def value(self):
+        return self.axis[self.index]
+
+    @value.setter
+    def value(self, val):
+        self.index = self.value2index(val)
+
+    @property
+    def high_value(self):
+        if self._is_increasing_order is True:
+            return self.axis[-1]
+        elif self._is_increasing_order is False:
+            return self.axis[0]
+        else:
+            raise NotImplementedError("The axis is not ordered so there is no high/low value")
+
+    @property
+    def index(self):
+        return self._index
+
+    @index.setter
+    def index(self, value):
+        if self._index != value:
+            if self.low_index <= value <= self.high_index:
+                self._index = value
+                self.events.index_changed.trigger(obj=self, index=self.index)
+
+    @property
+    def low_index(self):
+        return 0
+
+    @property
+    def high_index(self):
+        return self.size - 1
+
+    @property
+    def size(self):
+        return len(self.axis)
     def calibrate(self, *args, **kwargs):
         raise TypeError("This function works only for uniform axes.")
 
--- a/hyperspy/axes.py
+++ b/hyperspy/axes.py
@@ -30,6 +30,8 @@ import pint
 import traits.api as t
 from sympy.utilities.lambdify import lambdify
 from traits.trait_errors import TraitError
+from traits.api import observe
+from traits.observation.api import trait
 
 from hyperspy._components.expression import _parse_substitutions
 from hyperspy.api_nogui import _ureg
@@ -239,13 +241,20 @@ class UnitConversion:
     @units.setter
     def units(self, s):
         if s == "":
-            self._units = t.Undefined
+            s = t.Undefined
         self._units = s
 
 
 @add_gui_method(toolkey="hyperspy.DataAxis")
 class BaseDataAxis(t.HasTraits):
     """Parent class defining common attributes for all DataAxis classes.
+
+    The Base Data class has name and units attributes. Both of these can be freely set/ mutated
+     as long as the values are strings.
+     This axis cannot be plotted as it has no `.axis` attribute or size.
+     The axes_manager property should only be set by an `AxesManager` class to avoid
+     confusion. This value is automatically set when any Axis is appended or added to
+     an `AxesManager.axes` list attribute.
 
     Parameters
     ----------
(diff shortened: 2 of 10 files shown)


CSSFrancis:
The next PR which handles the slicing is found here https://github.com/CSSFrancis/hyperspy/tree/FancySlicingRefactor if that helps understand some of the motivations for these changes.

hyperspy/signal.py
```diff
@@ -3092,7 +3092,12 @@ def get_dimensions_from_data(self):
         """
         dc = self.data
         for axis in self.axes_manager._axes:
-            axis.size = int(dc.shape[axis.index_in_array])
+            new_size = int(dc.shape[axis.index_in_array])
+            if axis.size != new_size:  # Do nothing if already set
+                if not axis.is_uniform:  # Axis must be uniform to set the size
+                    axis.convert_to_uniform_axis()
```
ericpre, line 3326: I don't know if it would it be better to raise an error and say that this function doesn't support non-uniform axis? Maybe it would be good to revisit the purpose of this method - I find the name misleading...

hyperspy/axes.py
```diff
@@ -2010,86 +2002,85 @@ def update_axes_attributes_from(self, axes,
             self.events.any_axis_changed.trigger(obj=self)
 
     def _update_attributes(self):
```
ericpre: Is it still necessary?

hyperspy/signal.py
```diff
@@ -3221,7 +3225,7 @@ def rollaxis(self, axis, to_axis, optimize=False):
             s.axes_manager._axes,
             index=axis,
             to_index=to_index)
-        s.axes_manager._update_attributes()
+        #s.axes_manager._update_attributes()
```
ericpre: Left over?

hyperspy/signal.py
```diff
@@ -6343,16 +6346,16 @@ def transpose(self, signal_axes=None,
 
         # reconfigure the axes of the axesmanager:
         ram = res.axes_manager
-        ram._update_trait_handlers(remove=True)
+        #ram._update_trait_handlers(remove=True)
```
ericpre: Left over?

hyperspy/signal.py
```diff
@@ -6343,16 +6346,16 @@ def transpose(self, signal_axes=None,
...
         # _axes are ordered in array order
         ram._axes = [ram._axes[i] for i in array_order]
         for i, ax in enumerate(ram._axes):
             if i < len(navigation_axes):
                 ax.navigate = True
             else:
                 ax.navigate = False
-        ram._update_attributes()
-        ram._update_trait_handlers(remove=False)
+        #ram._update_attributes()
```
ericpre: Left over?

hyperspy/axes.py
```diff
@@ -741,102 +805,18 @@ def _is_increasing_order(self):
...
+        try:
+            steps = self.axis[1:] - self.axis[:-1]
+            if np.all(steps > 0):
+                return True
+            elif np.all(steps < 0):
+                return False
+            else:
+                # the axis is not ordered
+                return None
+        except:
```
ericpre: Specify exception type? Add a comment too?

hyperspy/axes.py
```diff
@@ -974,17 +953,20 @@ def __init__(self,
         for parameter in parameters.keys():
             self.add_trait(parameter, t.CFloat(parameters[parameter]))
         self.parameters_list = list(parameters.keys())
-        self.update_axis()
-        self.on_trait_change(self.update_axis, self.parameters_list)
 
-    def update_axis(self):
+    @property
+    def axis(self):
```
ericpre: It may be worth caching it in case the function calculation is slow.

hyperspy/axes.py
```diff
@@ -409,6 +357,203 @@ def _get_index(self, value):
...
+                 axis=[1],
+                 **kwargs):
+        super().__init__(
+            index_in_array=index_in_array,
+            name=name,
+            units=units,
+            navigate=navigate,
+            is_binned=is_binned,
+            **kwargs)
+        self.add_trait("_axis", t.Array)
```
ericpre: Why does it need to be a trait?

hyperspy/axes.py
```diff
@@ -269,20 +269,11 @@ class BaseDataAxis(t.HasTraits):
...
-    low_value = t.Float()
-    high_value = t.Float()
-    value = t.Range('low_value', 'high_value')
-    low_index = t.Int(0)
-    high_index = t.Int()
-    slice = t.Instance(slice)
+
     navigate = t.Bool(False)
     is_binned = t.Bool(False)
-    index = t.Range('low_index', 'high_index')
```
ericpre: `index` or `_index` still need to be a trait to be synchronise with the GUI widgets.

hyperspy/axes.py
```diff
@@ -2010,86 +2002,85 @@ def update_axes_attributes_from(self, axes,
...
-        self._signal_size = (np.prod(self.signal_shape)
-                             if self.signal_shape else 0)
-        self._navigation_size = (np.prod(self.navigation_shape)
-                                 if self.navigation_shape else 0)
+    def _on_scale_changed(self):
+        self.events.any_axis_changed.trigger(obj=self)
 
-        self._update_max_index()
+    def _on_offset_changed(self):
+        self.events.any_axis_changed.trigger(obj=self)
```
ericpre: Remove `_on_size_change`, `_on_scale_changed` and `_on_offset_changed` and use a single callback, named `_on_any_axis_change` for example?

ericpre (commented):
Thank you @CSSFrancis, this is very good!
There are a few details, which are worth tweaking further, but this is significant simplification of the axes implementation.
I guess this is one of these things which grew over time with more functionalities and ended up being a bit of a mess!

hyperspy/axes.py
CSSFrancis: I have the index as a trait for the DataAxis and not for the BaseDataAxis.  Is that okay?  My idea is that the BaseDataAxis would only ever be used if you are trying to define something like a ragged dimension.  In which case the index doesn't have a physical sense.  I can set it as <undefined> if that works better

hyperspy/axes.py
CSSFrancis: This could be potentially useful but I am a little hesitant about caching the axis as I think then we start going down the rabbit hole of adding a bunch of extra traits and listeners just to make sure things work correctly and even there it could get complicated and isn't guaranteed .  I would rather that some function which repeatedly calls the axis property use something like `ax1 = FunctionalDataAxis.axis` and then use that instance repeatedly. 

I can add properties which directly calculate the high and low values. Rather than getting the values from the axis which might help some. 

Either way it might be worth thinking about a case where caching speeds up things significantly. Even for a size of 1,000,000 values it only takes about 10 ms to calculate something like the sin or log of each value. I think in this case the trade off of speed for simplicity might be worth it.

hyperspy/axes.py
CSSFrancis: It is only necessary because the axes_manager isn't set when an axes is initialized so sometimes an axis can be added to and axis manager but not have `axis_manager` set.  The idea is that whenever the `_axes` variable is changed just make sure that is set. 

Let me see if there is a different way to set this which is easier to understand.

hyperspy/signal.py
CSSFrancis: I'm not a huge fan of this method either but I think that currently it is necessary if you mapped a function on a non-uniform-axis or crop something.  Previously it was kind of a nightmare with Non-Uniform axes as it would return an axis with a different size than the len of the axis.

It is usually called after the ~~`map`~~ `_deepcopy_with_new_data`  function to handle functions which change the size of the axis or when something is stacked/ data is changed.  In general the handling of matching data shape/dimensions and axis shape/dimensions should be handled better.

This might be a larger discussion than just this PR and something that I was meaning to bring up.

The core discussion is how does hyperspy handle the setting of the data attribute in relation to the axes.  Is the axes_manager immediately updated, sizes changed and axes dropped? How does it handle non uniform axes? Are they sliced, repeated, etc. or are they just changed into UniformAxes with offset 0 and scale 1. 

Maybe an `update_axes` function is the best way to go. With the option of passing a list of axes dictionaries might be the best idea and then just have each function call `update_axes` in each situation.

hyperspy/axes.py
ericpre: Sorry, it would have been less confusing if my comment was in a better place. What I meant is that the `index` or `_index` needs to be a `traiits` type, e.g. an instance of `traits.api.Range`.

ericpre (commented):
We should take this opportunity to add comments describing the expected use of properties and the reasoning for the relationship between the different axis and why different attributes/properties are defined in this specific way, etc.

hyperspy/axes.py
CSSFrancis: I kept this just because I didn't want anyone trying to set the _axis with anything other than an array as that would complicate the fancy slicing quite a bit.

hyperspy/axes.py
CSSFrancis: Ahh, that makes more sense.  That requires that the low and high indexes be traits as well. Or at least to my understanding that is the case.  I currently have it so that the index is set with a `@index.setter` function which won't update outside of the index high and low.  I'm not terribly familiar with how else this might interact with the API.

hyperspy/axes.py
CSSFrancis: This is now set with the `@observe('_axes') decorator which should be more intuitive and less likely for bugs to occur.  I will note that this assumes that some axes object is only a member of one AxesManager instance.  To change that would require some extra work whch might be out of this scope presently.

hyperspy/axes.py
CSSFrancis: Sounds good! This fails when you have an axis list like ["a","b","c", "d"] so some other value which cannot be subtracted

hyperspy/axes.py
CSSFrancis: I should note that the `self.events.any_axis_changed.trigger(obj=self)` should probably trigger when the `_axis` is changed.  I am not entirely sure what in the drawing is linked to that event.  That does argue for keeping the `_axis` attribute as a trait.

hyperspy/axes.py
CSSFrancis: Okay so I think the way to do this if we want to go down this path would be to keep the `_axis` attribute for all axes.

Then we could define a function like this:

```python
@observe('scale', postinit=True)
@observe('offset', postinit=True)
@observe('size', postinit=True)
def update_axis(event):
    self._axis  = np.arange(self.size)*self.scale+self.offset
```

Then for the FunctionalDataAxis we could create a function

```python
@observe(trait("x._axis", notify=False).list_items(), postinit=True)
@observe(self.parameters_list, postinit=True)
def update_axis(event):
    kwargs = {}
    for kwarg in self.parameters_list:
        kwargs[kwarg] = getattr(self, kwarg)
    new_axis = self._function(x=self.x.axis, **kwargs)
    # Set not valid values to np.nan
    new_axis[np.logical_not(np.isfinite(new_axis))] = np.nan
    self._axis = new_axis
```

hyperspy/signal.py
CSSFrancis: Part of the issue is the repeated use of this line of code `self._deepcopy_with_new_data(None)`

This state initializes a signal with no data but retains the axes.

A better behavior might be to formalize how we use the `_deepcopy_with_new_data` function. Specifically how axes are handled. I can make an issue to track this.

hyperspy/axes.py
CSSFrancis: Part of the reason I don't really like this is the fact that even though the `._axis` property is protected you could still directly modify the values for a UniformDataAxis and FunctionalDataAxis

hyperspy/axes.py
ericpre: > I kept this just because I didn't want anyone trying to set the _axis with anything other than an array as that would complicate the fancy slicing quite a bit.

Okay, maybe worth adding a short comment.

> I should note that the `self.events.any_axis_changed.trigger(obj=self)` should probably trigger when the `_axis` is changed. I am not entirely sure what in the drawing is linked to that event. That does argue for keeping the `_axis` attribute as a trait.

Yes.

hyperspy/axes.py
ericpre: > Either way it might be worth thinking about a case where caching speeds up things significantly. Even for a size of 1,000,000 values it only takes about 10 ms to calculate something like the sin or log of each value. I think in this case the trade off of speed for simplicity might be worth it.

The calculation itself may not take very long but when it is called repetitively, it can slow down other operations. An example of this is fitting (#2949) but in this specific case, I haven't check if it introduces a performance regression!
Better to leave it as it is for now and come back to it later.

> Then we could define a function like this:
> 
> ```python
> @observe('scale', postinit=True)
> @observe('offset', postinit=True)
> @observe('size', postinit=True)
> def update_axis(event):
>     self._axis  = np.arange(self.size)*self.scale+self.offset
> ```

Using `traits.api.observe` is good, as the old one is deprecated. It seems that it comes advantages, that may be beneficial for us.
I added an item to the [2.0 release list](https://github.com/hyperspy/hyperspy/issues/2996#issue-1350647756) to migrate the new Traits API in other place of the code.

> Part of the reason I don't really like this is the fact that even though the `._axis` property is protected you could still directly modify the values for a UniformDataAxis and FunctionalDataAxis

I wouldn't worry about it, because it is a common pattern in python.

hyperspy/signal.py
ericpre: > This might be a larger discussion than just this PR and something that I was meaning to bring up.

Yes, agreed!

hyperspy/axes.py
```diff
@@ -256,6 +258,15 @@ def units(self, s):
...
 
+    The Base Data class has name and units attributes. Both of these can be freely set/ mutated
+    as long as the values are strings.
+
+    This axis cannot be plotted as it has no `.axis` attribute or size.
+
+    The axes_manager property should only be set by an `AxesManager` class to avoid
+    confusion. This value is automatically set when any Axis is appended or added to
+    an `AxesManager.axes` list attribute.
+
```
ericpre, line 258: Add docstring for property and attributes as per https://numpydoc.readthedocs.io/en/latest/format.html#class-docstring.

The HyperSpy API documentation is a bit all over the place at the moment, but I have started to look at it and this should be sorted for the 2.0 release. Mentioning this here, because it would be good to document as much as possible the API docstring for the axes in this PR.

hyperspy/axes.py
```diff
@@ -2001,18 +2011,16 @@ def update_axes_attributes_from(self, axes,
         if changes:
             self.events.any_axis_changed.trigger(obj=self)
 
-    def _update_attributes(self):
+    @observe("_axes")
+    @observe(trait("_axes", notify=False).list_items())
+    def add_axis_manger(self, event):
```
ericpre, line 2137: Make this method private and fix the typo?

hyperspy/axes.py
```diff
@@ -2001,18 +2011,16 @@ def update_axes_attributes_from(self, axes,
         if changes:
             self.events.any_axis_changed.trigger(obj=self)
 
-    def _update_attributes(self):
+    @observe("_axes")
+    @observe(trait("_axes", notify=False).list_items())
```
ericpre, line 2136: Just to make sure that I am following correctly, `@observe(trait("_axes", notify=False).list_items())` is to call the `add_axis_manger` method when the attribute, which are of `traits` instances, are changed? If so, I don't see why we need it here? The `axes_manager` attribute of the axis needs to be set  only when the axis is added to a `AxesManager` (currently covered by `@observe("_axes")`)?

hyperspy/axes.py
ericpre: > Ahh, that makes more sense. That requires that the low and high indexes be traits as well. Or at least to my understanding that is the case.

Yes, this is correct.

> I currently have it so that the index is set with a `@index.setter` function which won't update outside of the index high and low. I'm not terribly familiar with how else this might interact with the API.

This is a details, which will be to sorted in this PR and we can come back to it toward the end of this PR when other things are settled.

hyperspy/axes.py
```diff
@@ -2009,87 +2029,78 @@ def update_axes_attributes_from(self, axes,
         if changes:
             self.events.any_axis_changed.trigger(obj=self)
 
-    def _update_attributes(self):
-        getitem_tuple = []
-        values = []
-        signal_axes = ()
-        navigation_axes = ()
+    @observe("_axes")
+    @observe(trait("_axes", notify=False).list_items())
```
ericpre, line 2136: Does it require to bump `traits` minimum requirement to 6.1?

ericpre (commented):
Sorry, I have done a review but forgot to submit it... I updated my comments with the latest changes.

hyperspy/axes.py
CSSFrancis: So the first case is triggered by someone setting the axes value. Something like `am.axes=[axis1, axis2]`

Because lists are mutable in place though that observation isn't triggered by a inplace operation. For that we need the second observation which triggers when something inplace occurs. Ie. `am.axes[1]= axis1`


See https://docs.enthought.com/traits/traits_user_manual/notification.html#notification-handler for more details.

hyperspy/axes.py
CSSFrancis: Yes we should update that dependency

CSSFrancis:
> Sorry, I have done a review but forgot to submit it... I updated my comments with the latest changes.

All good :) thanks for the review. To be honest I did the same thing the first time around when I replied...

hyperspy/signal.py
CSSFrancis: see #3041

hyperspy/axes.py
CSSFrancis: After looking through the documentation for Traits I think it should be fairly easy to cache the property.  I hadn't looked into the additional capabilities for caching.


```python
from traits.api import cached_property, Property

axis =  Property(observe=['scale", "size", "offset"])


@cached_property
def _get_axis(self):
    return np.arange(self.size)*self.scale+self.offset

```

https://docs.enthought.com/traits/traits_user_manual/advanced.html#caching-a-property-value

CSSFrancis: Or maybe it isn't quite this easy... Nothing with the traits package has been quite as intuitive as I first thought....

CSSFrancis: @ericpre So there are a few problems with this.  Mostly that inplace editing of traits is not supported.  We have been operating by adding traits to each object individually rather than adding traits to the Class as a whole which is not well supported for . The problem is that when we change the `__class__` variable and call `__init__` we skip over the `__new__` function where the listeners are set for some class. 

Unfortunately I think that means that in order to cache axes we would have copy axes when converting from one type to the next. 

That means that `s.axes_manager[0].convert_to_nonuniform_axis()` --> `s.axes_manager[0] = s.axes_manager[0].convert_to_nonuniform_axis()`

It isn't a huge change but would allow us to define the Axes classes like this.

```python
from traits.api import (cached_property, HasTraits, Str,Float, Int,observe, Property,Array, Undefined)

class BaseDataAxis(HasTraits):
    name=Str(Undefined)
    units=Str(Undefined)  

class DataAxis(BaseDataAxis):
    axis = Array()

class FunctionalDataAxis(DataAxis):
    x = Instance(BaseDatatAxis)
    axis = Property(observe=["x.axis"] # overwrites axis in DataAxis

    @cached_property
    def _get_axis(self):
        print("calculating axis")
        return self.func(x.axis, self.parameters)

class UniformDataAxis(DataAxis):
    scale = Float(1)
    offset = Float(0)
    size = Int(1)
    axis = Property(observe=["scale", "offset", "size"] # overwrites axis in DataAxis

    @cached_property
    def _get_axis(self):
        print("calculating axis")
        return np.arange(self.size)*self.scale+self.offset
```

CSSFrancis:
@ericpre This should be in a little bit better of shape now. I added in the ability to cache the axis value for both the `FunctionalDataAxis` and the `UniformDataAxis`. Let me know what still needs to be done to get this merged.  There is some documentation lacking but for the User Guide I was intending to do that all at once.  Once this is merged I will submit the PR for fancy slicing and we can go from there. 

The new changes improve the self documentation a little bit as well and are more in line with how the Traits package should be used.  The downside of this is that there is less no way to properly handle (at least to my knowledge) the inplace conversion from one axis to another.  This type of operation doens't properly set up the listeners for the different Axes types.  Additionally adding traits to the FunctionalDataAxis is not well supported.  Different parameters will have to be adjusted using `ax.parameters["power"]=3`.

For converting some axis we need to do:

`s.axes_manager[0] = s.axes_manager[0] .convert_to_non_uniform_axis()`

Not a huge api change but this can get a little clunky at times. 

It might be better to add something along the lines of :

`a.axes_manager.convert(axis=0, "non_uniform"]`

hyperspy/axes.py
CSSFrancis: So this runs into some complications as t.Range needs to be set with a value and not a Property initially.

We could do this:

```python
class DataAxis:
    axis = t.Array()  # Axis added as trait to track changes to axis
    high_index = t.Int(1)
    low_index = 0
    _index = t.Range(low_index, high_index)

    @observe("axis")
    def update_high_index(self, event):
        self.high_index = self.size-1
```

Or keep it the way it is.  I don't think it is too confusing right now but either works!

hyperspy/axes.py
```diff
@@ -688,46 +805,18 @@ def value_range_to_indices(self, v1, v2):
...
-        axes_manager = self.axes_manager
-        del d["axis"]
         if len(self.axis) > 1:
             scale_err = max(self.axis[1:] - self.axis[:-1]) - scale
             _logger.warning('The maximum scale error is {}.'.format(scale_err))
         d["_type"] = 'UniformDataAxis'
-        self.__class__ = UniformDataAxis
-        self.__init__(**d, size=self.size, scale=scale, offset=self.low_value)
-        self.axes_manager = axes_manager
+        new_axis = UniformDataAxis(**d, size=size, scale=scale, offset=offset)
```
ericpre: It is convenient to convert inplace and it would be good to avoid returning a different object.

hyperspy/axes.py
ericpre: An alternative would be not to use `@cached_property` and to do the cache manually. I can have a look at it.

hyperspy/axes.py
CSSFrancis: I can see if there is a way to do it. I think it means that we have manage how the changes to different attributes are handled which is a bit more involved.

CSSFrancis: I think I'll become an expert at the Traits package by the end of this :)

CSSFrancis: @ericpre Well it only took me a couple of days but I think the solution was a lot easier than I was making it. How do you feel about this?


```python

        """Convert to a non-uniform axis."""
        d["_type"] = 'DataAxis'
        new_axis = DataAxis(**d)
        if self.axes_manager is not None:
            self.axes_manager[self] = new_axis
        else:
            return new_axis
```

francisco-dlp:
Nice job @CSSFrancis ! This is a much needed refreshment of the axes code.

What do you mean with "unordered (or unorderable axes)"? Is it `hs.axes.DataAxis(["a", "b", "c"])`? It it is what it seems to me, I wonder if it wouldn't be better to implement this feature as simply axis labels. For example, an axis could have an extra optional attribute, `labels`, that could be a dictionary, containing labels. For example:

```python
velocity_axis = hs.axes.UniformDataAxis(name="velocity", units="m/s", size=3)
velocity_axis.labels = {0: "v_x", 1: "v_y", 2: "v_z"}
```
The advantage of using an extra attribute is that `.axis` remains a numeric array. Therefore, all code assuming this doesn't break. Also this enables adding labels that are not uniformly distributed over the axis (using `DataAxis` or `NonUniformDataAxis`.

The advantage of using a `dict` is that it enables adding labels to a subset of the axis points e.g.:

```python
temp = hs.axes.UniformDataAxis(name="temperature", units="degC", size=100)
temp.labels = {0: "Water freezing point", 100: "Water boiling point"}
```

CSSFrancis:
> Nice job @CSSFrancis ! This is a much needed refreshment of the axes code.

@francisco-dlp Thanks for the response.  This was exactly the kind of idea I was hoping this PR would generate. 
> 
> What do you mean with "unordered (or unorderable axes)"? Is it `hs.axes.DataAxis(["a", "b", "c"])`? It it is what it seems to me, I wonder if it wouldn't be better to implement this feature as simply axis labels. For example, an axis could have an extra optional attribute, `labels`, that could be a dictionary, containing labels. For example:
> 
Yes I would say that there are both unordered axes, something like `hs.axes.DataAxis(axis=[1,3,2,5,4,7]`) and unorderable axes which would be something like `hs.axes.DataAxis(axis=["a","b","c","d","e"])` or even something like   

```python
hs.axes.DataAxis(axis=[hs.axes.LabeledColumn(name="x", units="nm", scale=.1),
                       hs.axes.LabeledColumn(name="y", units="nm", scale=0.1),
                       hs.axes.LabeledColumn(name="kx", units="nm^-1", scale=0.01),
                       hs.axes.LabeledColumn(name="ky", units="nm^-1", scale=0.01)])
```

There are a couple of points for allowing each of these types but I do like the idea of adding a labels dictionary attribute.

### Let's start with the unordered axis. Which is numeric array but not necessarily increasing or decreasing. 

- I think that hyperspy should try to support as much numpy as possible. This includes something like:
```python
data = np.ones((10,10,10))
s = hs.signals.Signal2D(data)
s.axes_manager.signal_axes[1].scale=.1
data[:, :, [1,2,5,4]] # returns the 1st, 2nd, 5th and 4th indexes 
data.shape # (10, 10, 4)
s.isig[:. [1,2,5,4]] # shape == (10,10,4)
s.isig[:,[.1,.2,.5,.4]] # equivalent to above
``` 
This requires the support of unordered axes.  I don't think that this is too hard of an ask. There is already a fair bit of missing support for the DataAxis and functions which assume that axes have a scale or offset attribute, this is just an additional extension of that.  In this case any time an unordered axis is asked to do something that it can't it will throw an error. This might be something we can be more explicit about but I don't think this is necessarily a bad thing.

### For the unorderable axis this is an array of strings or some other object.

I think in this case I go back and forth. You might be right about adding in a labels attribute.  I like the concept of not having labels for each index and I think has a fair bit of potential for indexing using custom labels. The hard part is managing both parts of slicing; Slicing with a label and slicing the labels object. I will also say that the opposite syntax makes more sense to me. i.e. `velocity_axis.labels = {0: "v_x", 1: "v_y", 2: "v_z"}` --> `velocity_axis.labels = {"v_x":0, "v_y":1, "v_z":2}`

1: Slicing with a label: `velocity_signal.isig["v_x"]`
- In order to implement this would could add in a simple method:
```python
def label2index(label, label_dict):
    if label in label_dict:
       return label_dict[label]
    else:
        return label
```
- There is very low overhead for this, the implementation is trivial, and it would even allow things like `temp.labels["liquid_water_range"]=slice(0.,100.)`

2: Slicing the label dictionary:
- This is slightly more difficult and could be computationally difficult if you have many different labels. Ideally we want our labels to support at least indexes and real values (and possibly slices/arrays/ROI's?)   

```python
def slice_labels(self, slice, labels):
    new_indexes = np.arrange(self.size)[slice]
    for key, value in labels.items:
        if isinstance(value, float):
            value = self.value2index(float)
        if not value in new_indexes:
           labels.pop(key)
        elif value in new_indexes and isinstance(value, ind):
           labels[key]=np.where(value==new_indexes)[0][0]
    return labels
```

The case where we might have a long list if labels is after something like slicing with an n-dimensinal boolean array. In that case any of the sliced dimensions are flattened and the resultant axis should be an index of the values for each of the axes. Lets consider a 4 dimensional array sliced with a 4 dimensional boolean array. This results in a 1-Dimensional array of values.  If we call that one axis the index_axis then `index_axis.values= {(1,1,2,1):0,  (1,1,2,2):1, ..., (x,y,kx, ky):n}`  This way we maintain the axis information in the form of labeling each axis.


> The advantage of using a `dict` is that it enables adding labels to a subset of the axis points e.g.:
> 
> ```python
> temp = hs.axes.UniformDataAxis(name="temperature", units="degC", size=100)
> temp.labels = {0: "Water freezing point", 100: "Water boiling point"}
> ```

I feel like this could be extremely useful in many cases!  I think the general concept would be that labels would be an easy way to annotate and manipulate data without using ROI's.

francisco-dlp:
> I think that hyperspy should try to support as much numpy as possible.

Not sure about it: there are things that you can do with abstract arrays that you shouldn't do with the sort of data that HyperSpy deals with. HyperSpy is about data with axes that can be separated into "signal" and "navigation" axes. These simple constraints are HyperSpy's raison d'etre. They enable things like the `map` and `plot` methods, and, at the same time, limit the (e.g. numpy) features that it can support.

Technically we could support:

```python
s.isig[:. [1,2,5,4]] # shape == (10,10,4)
s.isig[:,[.1,.2,.5,.4]] # equivalent to above
```
but what for? And how to implement it in a way that makes sense?

> I will also say that the opposite syntax makes more sense to me. i.e. `velocity_axis.labels = {0: "v_x", 1: "v_y", 2: "v_z"}` --> `velocity_axis.labels = {"v_x":0, "v_y":1, "v_z":2}`

I see your point. 

>  it would even allow things like `temp.labels["liquid_water_range"]=slice(0.,100.)`

Nice idea. Maybe this should operate at the level of the `Signal`, because we are talking about ROIs and ROIs are classes that operate on `Signals`. Currently we can save markers but not ROIs. I can see the appeal of storing ROIs in a way that they are easy to use. What about:

```python
> s["liquid_water_range"] # Get's the right ROI from `s.metadata` and returnts `roi(s)`.
```

> I feel like this could be extremely useful in many cases! I think the general concept would be that labels would be an easy way to annotate and manipulate data without using ROI's.

We have to be careful not to implement several ways of performing the same thing. Your comment has made me realize that the `labels` dictionary is somehow an alternative to `Point1D` ROIs.

CSSFrancis:
> > I think that hyperspy should try to support as much numpy as possible.
> 
> Not sure about it: there are things that you can do with abstract arrays that you shouldn't do with the sort of data that HyperSpy deals with. HyperSpy is about data with axes that can be separated into "signal" and "navigation" axes. These simple constraints are HyperSpy's raison d'etre. They enable things like the `map` and `plot` methods, and, at the same time, limit the (e.g. numpy) features that it can support.

> Technically we could support:
> 
> ```python
> s.isig[:. [1,2,5,4]] # shape == (10,10,4)
> s.isig[:,[.1,.2,.5,.4]] # equivalent to above
> ```
> 
> but what for? And how to implement it in a way that makes sense?

I agree that a lot of what makes hyperspy work is the constraints, that being said I think that the above code __does__ fall under those constraints and is easily implemented using the existing infastructure.  (simply changing from a UniformDataAxis --> DataAxis and slicing or keeping an axis a DataAxis and slicing works). From a hyperspy standpoint supporting slicing with multi-indexes or boolean arrays I would like to think of as very much on the brand of manipulating data but saving the underlying metadata and information about the axis.

I use these kinds of functions many times to mask some data, take the sum of some area, compare two data points, getting the intensity at a set of points on the data set.  Additionally, in the case of having a list of vectors this type of slicing is very necessary to down select from a large list. 

Unfortunately, this often requires me to operate on the `.data` attribute. Which turns into creating a new signal rather than dealing with the hassle of editing the data attribute and changing the axes myself.  

I think that my workflow is about the worst thing we can have our users doing as it undermines the whole persistence of metadata, axes etc.  That is why I think supporting as much of numpy as reasonable is important as it stops much of this behavior. 

The implementation is all in #3055 if you are interested in taking a look.  The diff might be large because it includes this PR but I submitted it as a draft because I think it helps make sense of some of these changes. 

> > I will also say that the opposite syntax makes more sense to me. i.e. `velocity_axis.labels = {0: "v_x", 1: "v_y", 2: "v_z"}` --> `velocity_axis.labels = {"v_x":0, "v_y":1, "v_z":2}`
> 
> I see your point.
> 
> > it would even allow things like `temp.labels["liquid_water_range"]=slice(0.,100.)`
> 
> Nice idea. Maybe this should operate at the level of the `Signal`, because we are talking about ROIs and ROIs are classes that operate on `Signals`. Currently we can save markers but not ROIs. I can see the appeal of storing ROIs in a way that they are easy to use. What about:
>
> ```python
> > s["liquid_water_range"] # Get's the right ROI from `s.metadata` and returnts `roi(s)`.
> ```
> 
>  > I feel like this could be extremely useful in many cases! I think the general concept would be that labels would be an easy way to annotate and manipulate data without using ROI's.
> 
> We have to be careful not to implement several ways of performing the same thing. Your comment has made me realize that the `labels` dictionary is somehow an alternative to `Point1D` ROIs.

I think that the distinction __is__ that the labels attribute is associated with some axis rather than with some signal where as an ROI is related or even independent of a signal. The Point1D is just another way of directly indexing the data but we support both cases. I would rather see this as a way to annotate some axis or provide a secondary view.  Maybe another way to think about this is in terms of how we support conversions between units. This is just an extension of that to a custom mapping. 

Ultimately, I just think something that worked like this:

```
inten = s.isig["Cu_K", "Ni_Ka", "Si_Ka"]
print(s.axes_manager.signal_axes[0].axis) # [8.040, 7.477, 1.74]
inten.plot(labels=True) # Add in labels to graph
```

Is very intuitive and uses the nice new capabilities that hyperspy has added with the non-uniform axis pretty effectively.

CSSFrancis:
@francisco-dlp The relevant change is here for the uniform axis.

```python

if isinstance(_slice, (np.ndarray, list)):
    new_axis = self.convert_to_non_uniform_axis()
    return new_axis._slice_me(_slice)

```

In the slicing code then we just treat arrays and lists differently than tuples. https://github.com/numpy/numpy/pull/9686 This is the relevant change to numpy whose syntax I really think we should follow as that is what many of our users understand.  

I really don't think we want to diverge too much from this also because then we lose a lot of the functionality of a Signal being a duck array. (If numpy/ scipy ever figures that out 😃) 

Currently something like `s.isig[[1,2]]` works but has  a different behavior from `numpy` as it gets cast to a tuple. That seems like a recipe for confusion that is only going to get worse.

CSSFrancis:
@francisco-dlp did we ever fall either way on this?  I can move some of the code downstream to pyxem, but I do want to get as much into hyperspy as I can.  Let me know if there is any part of this that might need some more discussion/ information.

francisco-dlp:
Hi @CSSFrancis. Sorry to be slow to reply. I have some further thoughts on this and #3055, but I am too busy at the moment. I'll come back to you in the next few days.

CSSFrancis:
@francisco-dlp No worries!  Whenever you get around to it is fine, thanks for taking your time to review this.

francisco-dlp:
Thinking about it, I backtrack from my suggestion to use a `dict` to store axis labels. As you mention, that would provide an alternative way to annotate data, but I don't think that that's a good idea. To annotate data we already have markers. Therefore, according to the Zen of Python, we shouldn't provide an alternative. Beside, flexible axes labels cannot offer the level of customization that markers do. So, what about simply storing labels in a numpy `array`? This is still different from your current implementation that stores the labels in the axis array. I think that it is better to store them in their own array so that i) `axis` items remain numbers (no need to consider a different case) ii) slicing simply slices the labels array too if present (simpler implementation) iii) the placement of the axis labels in a plot can we adjusted by adjusting the axis values.

CSSFrancis:
@francisco-dlp Thanks for the comments. 

> Thinking about it, I backtrack from my suggestion to use a `dict` to store axis labels. As you mention, that would provide an alternative way to annotate data, but I don't think that that's a good idea. To annotate data we already have markers. Therefore, according to the Zen of Python, we shouldn't provide an alternative. 

I think that makes sense.  Maybe this does bring up a point that our idea of ROI's and markers are pretty similar in some cases.  Maybe we can start to think of markers as a collection of ROI's or even as a baseSignal with a collection of ROI's?  I sometimes think that the makers code can be a little obtuse, especially when you want different markers as a function of navigation position.  Maybe an alternative is to make a `MarkerSignal1D`  or (`ROISignal1D`) which can either have no navigation dimensions and be applied to the full signal or the same number of navigation dimensions which means it iterates alongside the signal when plotted. 
I can maybe move this discussion elsewhere. 

>Beside, flexible axes labels cannot offer the level of customization that markers do. So, what about simply storing labels in a numpy `array`? This is still different from your current implementation that stores the labels in the axis array. I think that it is better to store them in their own array so that i) `axis` items remain numbers (no need to consider a different case) ii) slicing simply slices the labels array too if present (simpler implementation) iii) the placement of the axis labels in a plot can we adjusted by adjusting the axis values.

I think this is a pretty good compromise and I should be able to implement this fairly easily. That being said I am slightly hesitant about if someone adjusts the `axis` value it is difficult to track the changes with the `label` and that could cause confusion. 

For example how can we handle `s.axes_manager[0].axis = np.array([1,2,4,5,6])`.  We would probably have to drop the labels and then throw a warning. That's not the worst thing but no exactly ideal.

hakonanes:
Pitching in as an extension developer on some points raised by @CSSFrancis. (Feel free to move this comment to another issue/discussion.)

> Maybe we can start to think of markers as a collection of ROI's or even as a baseSignal with a collection of ROI's?

*kikuchipy* uses *HyperSpy's* markers actively to draw Kikuchi bands on patterns (see this figures in [this tutorial](https://kikuchipy.org/en/stable/tutorials/geometrical_ebsd_simulations.html)). They are used as visual aid to confirm a correct indexing of electron backscatter diffraction (EBSD) patterns or a particular position of a pattern feature. I don't think this use fits the description of them as ROIs, at least not as mentioned here?

> especially when you want different markers as a function of navigation position

The [implementation](https://github.com/pyxem/kikuchipy/blob/develop/kikuchipy/draw/markers.py) in *kikuchipy* uses the same set of bands for all nav positions, but sets those not visible in a certain pattern to NaN. Each band's line coordinates vary per nav position. The current implementation of markers in *HyperSpy* fits this use perfectly. However, I have plans to make the marker coordinates calculated on-the-fly (lazily using Dask). Might be you've thought along these lines yourself, @CSSFrancis.

CSSFrancis:
> Pitching in as an extension developer on some points raised by @CSSFrancis. (Feel free to move this comment to another issue/discussion.)
> 
> 
> 
> > Maybe we can start to think of markers as a collection of ROI's or even as a baseSignal with a collection of ROI's?
> 
> 
> 
> *kikuchipy* uses *HyperSpy's* markers actively to draw Kikuchi bands on patterns (see this figures in [this tutorial](https://kikuchipy.org/en/stable/tutorials/geometrical_ebsd_simulations.html)). They are used as visual aid to confirm a correct indexing of electron backscatter diffraction (EBSD) patterns or a particular position of a pattern feature. I don't think this use fits the description of them as ROIs, at least not as mentioned here?

I see your point there. I was just commenting on some of the similarities between same markers and ROIs. There might even be some cases like creating a virtual dark field image from a set of ROI's where the line starts to get a little blurred. In that case we want the slicing capabilities of a ROI and the labeling and plotting capabilities of a marker. It's hard to say that something like a arrow or a text marker should be an ROI.

The way things are is good but this conversation sparked that thought. 



> > especially when you want different markers as a function of navigation position
> 
> 
> 
> The [implementation](https://github.com/pyxem/kikuchipy/blob/develop/kikuchipy/draw/markers.py) in *kikuchipy* uses the same set of bands for all nav positions, but sets those not visible in a certain pattern to NaN. Each band's line coordinates vary per nav position. The current implementation of markers in *HyperSpy* fits this use perfectly. However, I have plans to make the marker coordinates calculated on-the-fly (lazily using Dask). Might be you've thought along these lines yourself, @CSSFrancis.

Sorry that was kind of a throwaway line and I should have explained better. I also realize that this should probably be moved to a separate issue. 

Maybe this isn't as straight forward as I thought :) I always considered having to create as many markers as you needed for the signal with the most markers and setting some markers to nan a bug and not a feature. In your case I guess I could go either way. I will say that the way we are currently doing it is slow.

When I want to make a set of markers that change with every navigation dimension I have to create a set of markers that is as  long as the longest set of vectors(which can be very long) and add a bunch of np.nan markers. Once you add a bunch of these markers to a signal this slows down the plotting a lot. This is because each marker is plotted as a separate artist, so having 30+ artists really slows down Matplotlib and makes navigating a pain.

A better way would be to be able to define all the markers as a ragged array of vectors. Then just add one marker to the signal with the data equal to that ragged array.

But to that point when does this ragged array of vectors defining some set of markers start to look suspiciously like a ragged signal. We can even think of that set of makers as having a defined navigation dimension which can now be sliced alongside the signal or think about returning a set of markers from the `map` function. The idea of lazy markers is much easier to consider as well in this case as you don't need to know how many sets of markers you need.

hakonanes:
Thank you for elaborating. I agree with all your points made, and I'd say your experience with markers describes mine well. The creation of markers (geometrical simulations) in *kikuchipy* is currently using too much memory, so they cannot be added to a large dataset (> a couple GB) anyway. If we ever fix that, I hope to find time to look more into using ragged arrays as you say.

CSSFrancis:
> Thinking about it, I backtrack from my suggestion to use a `dict` to store axis labels. As you mention, that would provide an alternative way to annotate data, but I don't think that that's a good idea. To annotate data we already have markers. Therefore, according to the Zen of Python, we shouldn't provide an alternative. Beside, flexible axes labels cannot offer the level of customization that markers do. So, what about simply storing labels in a numpy `array`? This is still different from your current implementation that stores the labels in the axis array. I think that it is better to store them in their own array so that i) `axis` items remain numbers (no need to consider a different case) ii) slicing simply slices the labels array too if present (simpler implementation) iii) the placement of the axis labels in a plot can we adjusted by adjusting the axis values.

Sorry this has taken me a little while to get around to.  I've been struggling with this idea for a little bit and have gone back and forth on it.  I have decided that I don't really like the idea of having two arrays to slice.  This introduces some ambiguity such as what happens if the values are switched? Do the labels stay inplace with the index or switch with the value.  There are also instances where the axis is changed outside of the `_slice_me` function. That becomes a little more difficult to track and the labels might become different sizes to the axis.I am not sure that this implementation does it justice

I can remove the code related to column labels to get this and #3055 merged.  Maybe this is something that we can come back to in the future.  This also puts on hold some changes to the VectorSignal but I can look into different ways to accomplish that.

CSSFrancis:
@francisco-dlp I've come back to this a couple of times and it appears that there isn't an easy way to define the `is_array_increasing` function in a consistent way. 

The problem is that there is no way of tracking if the underlying data from an array is changed.  Previous the checking occurred after some change to the axis but this is not sufficient in many cases. 

The problem is some bit of code like this:

```python 
from hyperspy.axes import DataAxis
d = DataAxis(size=5)
d.axis[4]=-4
```

Because `numpy` operates directly on the underlying memory and not on the object this won't ever trigger a change or throw an error when the value is changed.  Even if we change the axis value directly:

```python
from hyperspy.axes import DataAxis
d = DataAxis(size=5)
d.axis = [1,2,-3,4,5]
```

Even though this throws an error it does it after the fact (after the underlying data is already changed). So 

Perhaps we can keep the ability to have unordered axes with less functionality in `hyperspy` much like the `DataAxis` doesn't support things like rebinning etc.

Can you maybe explain more of why you think unordered axes are detrimental to hyperspy?

jlaehne:
> These changes help to stop a lot of small bugs within the axes code. Things like being able to set the size of a DataAxis don't make sense when the size is directly related to the axis. Similarly with the UniformDataAxis, directly setting the Axis should throw an error.

Just came across someone again who wanted to change the `axis` array of a `UniformDataAxis` instead of converting to `DataAxis` first. We should really make that attribute a property (like done for others in #2830). I added a issue that should be solved by this PR: #3121.

CSSFrancis:
@jlaehne Good to keep track of these things.  I need to look into the PR again but I'd love to have your opinion on some of the proposed changes as you guy are the main user of the nonuniform axis.

The main point where this PR got stuck was allowing the axis array to accept;

1- Non increasing/decreasing values 
2- Other objects (i.e. strings, or other classes)

I think @francisco-dlp had some hesitations regarding having not strictly increasing or decreasing axis values. 

Regarding 1:
There isn't really a way to monitor something like someone changing the value of a numpy array as numpy views the raw memory.  There isn't a great way to be proactive about someone setting the value of a non uniform axis that isn't strictly increasing or decreasing.  That is to say throwing an error after `s.axes_manager[0].axis[4]=5` if the axis array isn't strictly increasing doesn't work.  There might be an eventual use case for having an axis that isn't strictly increasing or decreasing as well.

Regaurding2: 
I think there needs to be a better defined way to do this than our current way!


For example:

- Peak finding in 2 dimensions returns a ragged array with no information about the axes which the peaks come from.
- Peak finding in 1 dimensions returns a ragged array with `dtype = [('position', float), ('height', float),('width', float)])`
- In pyxem returning shifts has no information about the axis that created them
- Markers in hyperspy are have their information set as a custom `dtype`

I don't love using the numpy dtype specification to label columns because it tends to block some of the functionality of numpy arrays but it is also very important and useful to keep this information.  I wrote some of this up in #3078

jlaehne:
> 1- Non increasing/decreasing values

The use cases I deal with use non-uniform axes spacing, but still with continuously increasing/descreasing order. Sure it would be great to support more general cases, but for the moment I do not see an urgent need. However, as far as I know, it is true that we do not enforce an increasing/decreasing order when the axis is provided through an `axis` vector and that it could entail problems if someone does provide an unordered array there. In case we decide not to support the more general case for the moment, it would be good to check and enforce that.

> 2- Other objects (i.e. strings, or other classes)

I recently stumbled across the hard to work with output of peak finding as well. Indeed, returning a signal where one axis is ['position', 'height', 'width'] could be a solution. Another solution for his specific case would be to handle the output similar to the parameters of a fit-component that you can access as arrays individually - and convert to a signal as well (but one by one).

CSSFrancis:
This seems like it was pretty out of date so I had to pull some of the changes into a new branch and rebase on that. 

If someone wants to give this another quick review we can see if it is worth merging this in.

ericpre:
@CSSFrancis, I have the feeling that this will be quite a bit of work to finish this PR and most likely too much for the 2.0 release, what do you think? Are there API changes that would make a case to include this PR in the 2.0 release?

CSSFrancis:
> @CSSFrancis, I have the feeling that this will be quite a bit of work to finish this PR and most likely too much for the 2.0 release, what do you think? Are there API changes that would make a case to include this PR in the 2.0 release?

I don't think this should hold anything up.  There aren't really any breaking changes only certain things (which shouldn't have been possible) like setting the axis of a UniformData object no longer are. 

#3055 might have some slight api changes specifically related to slicing using lists, vs tuples vs arrays similar to how numpy changed their slicing slightly.  That being said I think the current implantation doesn't do exactly what it says it does. My intention is to pull out the:

- Slicing with a boolean array
- Slicing with a tuple of points 
- Slicing with an array of positions 

As much as I would like some of this stuff in 2.0.0 let's not hold it up any longer.  It might be needed for the pyxem 1.0.0 release but maybe we can circle back to this for a 2.1.0 release

CSSFrancis:
Circling back to this just so this doesn't get dropped.  I think that this should be in pretty good shape, but we will need to handle the case where you might have a non-increasing axis and it is used for a model.


```python
if self.axis._is_increasing_order is None:
    raise ValueError("The Axis used for fitting a model must be increasing or decreasing") 
```

ericpre:

jlaehne:
I think this is one of the next things we should get going on. I found a few more points concerning the `navigate` attribute, where warnings/errors are missing when changing the attribute (see #3293). Though possibly, we should not try to do too much in one PR.

CSSFrancis:
@jlaehne even though I'm usually in the several small PRs camp this is about the minimum size to get the logic correct for the Axes and AxesManager classes.


I'll come back to it this week to take another look.

hyperspy/_signals/lazy.py
```diff
@@ -335,7 +335,13 @@ def _get_dask_chunks(self, axis=None, dtype=None):
...
         for _axis in self.axes_manager._axes:
             if _axis.index_in_array < len(dcshape):
-                _axis.size = int(dcshape[_axis.index_in_array])
+                if _axis.size != int(dcshape[_axis.index_in_array]):
+                    if _axis.is_uniform:
+                        _axis.size = int(dcshape[_axis.index_in_array])
+                    else:
+                        _logger.warning(f"Axis {_axis} converted to a UniformDataAxis")
+                        _axis.convert_to_uniform_axis()
+                        _axis.size = int(dcshape[_axis.index_in_array])
```
ericpre, line 342: This function is about getting chunks and it is surprise that to do that it needs to edit the axis! Could it be a workaround for the data shape and the axis are out of sync?

hyperspy/axes.py
```diff
@@ -400,6 +367,240 @@ def _get_index(self, value):
...
+        try:
+            steps = self.axis[1:] - self.axis[:-1]
+            if np.all(steps > 0):
+                return True
+            elif np.all(steps < 0):
+                return False
+            else:
+                # the axis is not ordered
+                return None
+        except np.core._exceptions._UFuncNoLoopError:  # subtracting array of objects/strings will fail
```
ericpre: Avoid using numpy private API: `np.core._exceptions._UFuncNoLoopError` is a subclass of `TypeError`:
```suggestion
        except TypeError:  # subtracting array of objects/strings will fail
```

hyperspy/axes.py
ericpre: The problem with returning a new object is that the events connected to the old one will be lost.

hyperspy/_signals/lazy.py
CSSFrancis: The problem here is related to the fact that the axis and the array shape can be out of sync.  This causes some issues if they are out of sync and try to chunk the data.

hyperspy/axes.py
CSSFrancis: @ericpre Good point! I'll make the substitution

CSSFrancis:
> The problem with returning a new object is that the events connected to the old one will be lost.

@ericpre I'll make sure that this isn't the case...  

Seems like the things we need to make sure are handled correctly

1. Cases where non ordered nonUniform axes don't work (i.e. models) should raise errors.
2. Cases where the data.shape != the axes_manager axes shapes. Should raise errors or... more ideally be automatically fixed
3. Events and object propagation.  Axes should maintain events when changing types.  Axes in axes managers should be adjusted inplace

Any other things to check?

hyperspy/_signals/lazy.py
ericpre: Use `BaseSignal.get_dimensions_from_data` instead?

hyperspy/axes.py
```diff
@@ -933,25 +969,20 @@ def __init__(self,
                  size=1,
                  is_binned=False,
                  **parameters):
-        super().__init__(
+        super(DataAxis, self).__init__(
```
ericpre, line 1002: Is it necessary to add `DataAxis, self` in the `super` call here, is it not old python 2.x syntax?

hyperspy/axes.py
```diff
@@ -2009,87 +2060,85 @@ def update_axes_attributes_from(self, axes,
         if changes:
             self.events.any_axis_changed.trigger(obj=self)
 
-    def _update_attributes(self):
-        getitem_tuple = []
-        values = []
-        signal_axes = ()
-        navigation_axes = ()
+    @observe("_axes")
+    @observe(trait("_axes", notify=False).list_items())
```
ericpre, line 2136:
```suggestion
    # These two `observe` decorators are necessary to make sure that an item of the list is changed in place
    @observe("_axes")
    @observe(trait("_axes", notify=False).list_items())
```

ericpre (commented):
Regarding the changes with `convert_to_uniform_axis`, before `remove_trait`/`add_trait` were used to handle trait dynamically but why is it not possible anymore?

ericpre:
> Any other things to check?

Nothing that I can think of but even if something is broken after this PR, this is a good time to do it!

ericpre:
@CSSFrancis, are you happy if I take over this PR? I suspect you may be busy with others things! 😃

CSSFrancis:
@ericpre yea I kind of am trying to take a step back to finish up some pyxem stuff so if you pick this up I'd be very happy! 

Let me know if you help with anything! And I can make some time to look over this again.

PR #3034, opened 2022-09-24
Fix test extension workflow

ericpre:
Update to the `test_extension` workflow to install RosettaSciIO until it is released.


ericpre:
Superseded by #3038.

PR #3045, opened 2022-10-09
Simplify API import remove markers aliases

ericpre:
### Progress of the PR
- [x] Clean up samfire module imports,
- [x] clean up datasets module imports,
- [x] Remove `markers` module from `api`, as it is redundant with `api.plot.markers`,
- [x] Remove aliases of markers in favour of using class name.
- [x] update tests,
- [x] ready for review.


ericpre:
Should the markers name also be changed to use the class name?

```python
import hyperspy.api as hs

m = hs.markers.HorizontalLine(0)
print(m.name)
# 'horizontal_line`
```

francisco-dlp:
Code-wise it LGTM, but I wonder—wouldn't it be better to leave `markers` in `hs.plot` and remove them from `hs.marker`?

ericpre:
I have a slight preference for `api.markers` over `api.plot.markers` because of the following reason:
- markers are primarily used in plot but they can also used without being visualised on a figure
- `api.plot` contains contains convenience functions
- `api.markers` is shorter

These 3 reasons are weak but may be enough to go for `api.markers` over `api.plot.markers`?

francisco-dlp:
Personally, because, unlike ROIs, markers are only useful for plotting, I think it is more intuitive to place them under `plot`. I have always used them from `hs.plot.markers` and I was not aware that you could also find them in `hs.markers`. Notice that `hs.model` also contains "submenus".

Confusingly enough,  the User Guide documents both [`hs.plot.markers`](https://hyperspy.org/hyperspy-doc/current/user_guide/visualisation.html#markers) and [`hs.markers`](https://hyperspy.org/hyperspy-doc/current/user_guide/visualisation.html#permanent-markers).

ericpre:
Okay, I updated the changes to remove `api.markers` and keep `api.plot.markers`.

PR #3048, opened 2022-10-13
[bruker.py] Minimalist support of XRF data

vasole:
Related to issue #2898 

Following that issue, this PR tries to provide a reasonable value to self.hv to support XRF data

<del>I only have a huge XRF map (1273 x 1662 x 4096 uint16) and unless I down-sample when reading, the data size is wrongly returned as uint8 in the function `estimate_map_depth` </del>


vasole:
Well, the map seems to be displayed correctly. Forget about the comment concerning `estimate_max_depth`

vasole:
I think this PR arrives too late. I close it.

PR #3052, opened 2022-10-26
Fix pint import with pint>=0.20

taramatsci:
### Requirements
* Read the [developer guide](https://hyperspy.org/hyperspy-doc/current/dev_guide/intro.html).
* Base your pull request on the [correct branch](https://hyperspy.org/hyperspy-doc/current/dev_guide/git.html#semantic-versioning-and-hyperspy-main-branches).

### Description of the change
pint module has no longer the attribute unit, which resulted in breaking of the brc code. It can be directly invoked using Unit. 

### Progress of the PR
- [ ] add tests,
- [x] ready for review.

```python
import hyperspy.api as hs
s = hs.load(filename.bcf)
#resulted in an error
# Your new feature...
Corrected hyperspy/misc/axis_tools.py
```
Note that this example can be useful for updating the user guide.

### DIFF
--- a/hyperspy/misc/axis_tools.py
+++ b/hyperspy/misc/axis_tools.py
@@ -17,8 +17,8 @@
 
 
 import numpy as np
-from pint import Unit
-
+#from pint import Unit
+from hyperspy.api_nogui import _ureg  
 
 def check_axes_calibration(ax1, ax2, rtol=1e-7):
     """Check if the calibration of two Axis objects matches.
@@ -43,12 +43,12 @@ def check_axes_calibration(ax1, ax2, rtol=1e-7):
     """
     if ax1.size == ax2.size:
         try:
-            unit1 = Unit(ax1.units)
+            unit1 = _ureg.Unit(ax1.units)
         except:
             unit1 = ax1.units
         try:
             unit2 = ax2.units
-            unit2 = Unit(ax2.units)
+            unit2 = _ureg.Unit(ax2.units)
         except:
             pass
         if np.allclose(ax1.axis, ax2.axis, atol=0, rtol=rtol) and unit1 == unit2:
--- a/hyperspy/misc/axis_tools.py
+++ b/hyperspy/misc/axis_tools.py
@@ -17,8 +17,8 @@
 
 
 import numpy as np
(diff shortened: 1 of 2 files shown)


hakonanes:
Thank you for reporting this @Quantumstud.

This is breaking packages depending on HyperSpy as well. Restricting `Pint` to `<= 0.19` is a temporary fix. Asking the maintainers, would it be possible to make a 1.7.3 patch release?

hyperspy/misc/axis_tools.py
```diff
@@ -17,7 +17,7 @@
 
 
 import numpy as np
-from pint.unit import Unit
+from pint import Unit
```
ericpre: To be consistent and potentially avoid future surprise, it would be better to use `Unit` from the unit registry instance used in hyperspy. It means that `_ureg` needs to be imported here (`from hyperspy.api_nogui import _ureg`) and then `_ureg.Unit` needs to be used.

ericpre: To be consistent and potentially avoid future surprise, it would be better to use `Unit` from the unit registry instance used in hyperspy. It means that `_ureg` needs to be imported here (`from hyperspy.api_nogui import _ureg`) and then `_ureg.Unit` needs to be used.

ericpre (commented):
Thank for the bug fix.
As this is a bug fix, it needs to be based on the `RELEASE_next_patch` branch - this is what is meant with "Base your pull request on the [correct branch](https://hyperspy.org/hyperspy-doc/current/dev_guide/git.html#semantic-versioning-and-hyperspy-main-branches)" ! :)

I suggested an alternative bug fix. It would need a changelog entry too.

ericpre:
> This is breaking packages depending on HyperSpy as well. Restricting `Pint` to `<= 0.19` is a temporary fix. Asking the maintainers, would it be possible to make a 1.7.3 patch release?

Yes, agreed!

ericpre (approved):
Thanks @Quantumstud for the PR!
I rebased and added the changelog entry.

taramatsci:
Thank you Eric! Next time if I find a bug I will make it in RELEASE_next_patch.

PR #3055, opened 2022-10-27
Fancy Slicing Refactor

### LINKED ISSUE #3051: Support slicing with Boolean Arrays
#### Add in a way to operate on the entirety of the array without directly accessing the Data attribute
This is a part of #2876 and a natural extension of #3011. Specifically I would like to support something like:

```python
from skimge import disk

s[s.isig["x"]<5] # returns only the rows with values of `x` greater than 5. 

mask = np.zeros(20,20), dtype=bool)
mask[disk(5,5,4)]=1
in_mask = s.isig[mask]

s[s[: ,:,  "x"]<5] # same as above but not using `isig`

s[s>10]=0 # masking above some value
```

Alternatively, we would consider something like using `i` rather than directly adding support for `__get_item__` and `__set_item__` to a Basesignal.  ie ` s.i[s.isig["x"]<5]`.

As part of a more consistent application of masks for slicing data this helps to bring us more inline with `numpy` and other packages slicing capabilities.

#### Describe the context
The context of this is that currently we don't handle these kind of reduction applications very well.  With the addition of non-linear axes and their re-implementation in #3031 there is lots of infrastructure for handling these kinds of transformations from the axes side of things by converting axes from uniform to non uniform axes. 

#### Additional information/ Implementation:

In order to implement this we would just have to add some additional code to the `FancySlicer._slicer` class and the axes_manager class.  We would have to either flatten the data or return a ragged signal to handle this as well. 

I would suggest that when slicing with `isig` and `inav` we return a flattened array and when slicing the entire array we convert the last n dimensions to being ragged where `n = (s.dim - mask.dim)+1`

```python 

s[s.isig["x"]<5] # returns ragged signal

mask = np.zeros(20,20), dtype=bool)
mask[disk(5,5,4)]=1
in_mask = s.isig[mask] # returns non-ragged signal with last two dim flattened: signal_axes -->1 non uniform axis of tuples. 

in_mask = s.inav[mask] # returns non-ragged signal with first two dim flattened: navigation_axes --> 1 non uniform axis of tuples. 

s[s[: ,:,  "x"]<5]# returns ragged signal ragged dim ==2
s[s.inav[0]<5]# returns ragged signal ragged dim ==3 (for a 4 dim dataset)


```

CSSFrancis:
### Description of the change
This is the next in a series of PR to add additional support for custom Axes, additional slicing features and (eventually) support for Vector Signals. For previous changes see #3031. This also closes #3051.

I think that this is a particularly exciting set of changes and something that I've been meaning to get around to for some time! Particularly, I think that this opens many new avenues for analysis and data representation. 

Specific Changes:
## Changes to `DataAxis:
- Support for direct indexing using some value in a `DataAxis.axis` array. This is a "Basic" Vector type operation. 
```python
s = signals.Signal2D(np.arange(24).reshape(6, 4))
s.axes_manager[0].convert_to_non_uniform_axis()
s.axes_manager[0].axis = ["a", "b", "c", "d"]
s.isig[0] == s.isig["a"]
```

## Changes to masking using 1 dimensional mask or multi-index:
- 1 dimensional masks are now usable for slicing some dimensions
- Sliced axis is cast to a `DataAxis` if it is a `FunctionalDataAxis` or `UniformDataAxis`

```python
s = signals.Signal2D(np.arange(24).reshape(6, 4))
s.axes_manager[0].convert_to_non_uniform_axis()
s.axes_manager[0].axis = ["a", "b", "c", "d"]
s.isig[0,2] == s.isig["a","c"] == s.isig[[True, False, True, False]] 
```
## Changes to masking using n-dimensional mask:
- Allows for using an n-dimensional mask for slicing a signal. Specifically this can be used for isig and inav
- Axes are flattened and the indexes are merged and returned as a string "[index1 index2 ... indexn]" (note that this could be changed in the future to a custom class which is more friendly to indexing...)

## Added i  a `SpecialSignalSlicer` object for slicing the entire signal.
- This allows for slicing the entire signal in one go.  Note that this is primarily used in the case of slicing some full array.
- Axis is handled identical to above. Indexes are merged. 

```python
s = signals.Signal2D(np.arange(6**3).reshape(6, 6, 6))
new_sig = s.i[self.signal < 5] # returns a list of values above 5
```

## Changes to slicing:
- No longer supporting arbitrary Signal/Navigation axes. Axes must be ordered Navigation Axes --> Signal Axes. 
- This is left over from previous implementations (I believe)
- __Changes how tuple, lists and arrays are considered.  This is a pretty big change but is inline with numpy so it should happen sooner than later. It is also necessary for slicing with multi-indexes. __

### Progress of the PR
- [x] Add i `SpecialSignalSlicer` 
- [x] Add support for signal dimensional bool array
- [x] Add support for multidimensional bool array
- [x] Add tests for multidimensional signal bool array
- [ ] Reformatted handling slicing in the `axes` class
- [ ] Catch misordering of signal and navigation axes.
- [ ] add tests,
- [ ] ready for review.

Note that this PR should be re-based on #3031. Which should result in a smaller diff. I have just posted this as I think that it might help to explain and motivate some of the previous changes.

### DIFF
--- a/doc/user_guide/slicing.rst
+++ b/doc/user_guide/slicing.rst
@@ -1,0 +1,64 @@
+Slicing Tools
+**************
+
+Hyperspy attempts to copy the `numpy fancy indexing`_ approach in most cases but separates the
+handling of the signal and navigation axes.
+
+In addition to the numpy indexing/slicing hyperspy also supports indexing or slicing using real units
+as long as the axis is ordered, meaning the axis is constantly increasing/decreasing.
+For axes which are not ordered, slicing with real units is ambiguous but indexing with
+real unit values is still possible.
+
+In the case of slicing with a bool array, or an array of indexes a DataAxis is always returned.
+
+For example:
+
+.. code-block:: python
+
+    >>> s = hs.signals.Signal2D(np.random.rand((10,10,20,10)))
+    >>> print(type(s.axes_manager[2]))
+    <class 'hyperspy.axes.UniformDataAxis'>
+    >>> sliced_s = s.isig[[0,3,6,5,4]]
+    >>> print(type(sliced_s.axes_manager[2]))
+    <class 'hyperspy.axes.DataAxis'>
+    >>>print(sliced_s.axes_manager[2].axis)
+    [0. 3. 6. 5. 4.]
+
+    >>> bool_ind = np.zeros(10, dtype=bool)
+    >>> bool_ind[0,3,6,5,4] =True
+    >>>sliced_s = s.isig[bool_ind]
+    >>> print(type(sliced_s.axes_manager[2]))
+    <class 'hyperspy.axes.DataAxis'>
+    >>>print(sliced_s.axes_manager[2].axis)
+    [0. 3. 4. 5. 6.]
+    ...
+
+What is happening here is that once the data is no longer uniform it will convert to a DataAxis which can
+handle any arrangement of values. In this way no information is lost when slicing or indexing the data. In the
+top case you can see the resulting axis is unordered while the bottom case the axis is ordered.
+
+Due to the flexibility of the DataAxis we can also have labeled axes. This is useful for defining vectors,
+or other quantities within the framework of a signal.
+
+We can use this quite effectively in the case of trying to return multiple values.
+For example lets say we have a set of 2-D images and we want to find the center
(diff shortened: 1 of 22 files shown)


ericpre:

PR #3060, opened 2022-11-03
Fix a bug in .sur files for 1D series of spectra

LMSC-NTappy:
### Description of the change
Fixing a bug in the loading of "spectra series" data in the .sur and .pro format.  The data array was loaded as the transposed of what it needed to be, and this created an error at plot time (probably in other instances as well). This is a type of data that knows very little use so it went unnoticed until now.

Below is an example of code that used to raise an error and does not anymore. 

Many lines in the file are just changed because of pep8 re-formatting of the file.

### Progress of the PR
- [x] add changelog entry
- [x] update tests
- [x] Ready for review

```python
import hyperspy.api as hs
import hyperspy.tests.io
import pathlib
sigpath = pathlib.Path(hyperspy.tests.io.__file__).parent.joinpath('sur_data/test_spectra.pro')
sig = hs.load(sigpath)
sig.plot()
#raise ValueError
```

### DIFF
--- a/hyperspy/io_plugins/sur.py
+++ b/hyperspy/io_plugins/sur.py
@@ -16,19 +16,19 @@
 # along with  HyperSpy.  If not, see <https://www.gnu.org/licenses/#GPL>.
 
 # Plugin to read the mountainsmap surface format (sur)
-#Current state can bring support to the surface format if the file is an
-#attolight hyperspectral map, but cannot bring write nor support for other
-#mountainsmap files (.pro etc.). I need to write some tests, check whether the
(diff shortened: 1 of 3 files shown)


jlaehne:
Can you contribute the same PR to the rosettasciio repository, otherwise the fix will be lost in the 2.0 release? As there the code is reformatted using black, it also should be easier to review your changes and we could then also merge your fix here.

LMSC-NTappy:
Hi @jlaehne. This is now done in hyperspy/rosettasciio#66

jlaehne (approved):

PR #3062, opened 2022-11-10
Fix `plot_spectra` legend for "cascade" style

francisco-dlp:
### Description of the change

`plot_spectra` does not display the legend when `ax` is given and the style is `"cascade"`. This fixes it.

### Progress of the PR
- [x] add tests,
- [ ] ready for review.

The following doesn't display the legend without this fix:
```python
s = hs.signals.Signal1D((np.arange(100).reshape((5, 20))))
f = plt.figure()
ax = f.add_subplot(111)
hs.plot.plot_spectra(s, legend="auto", ax=ax, style="cascade")
```

### DIFF
--- a/hyperspy/drawing/utils.py
+++ b/hyperspy/drawing/utils.py
@@ -1481,7 +1481,7 @@ def plot_spectra(
         _make_cascade_subplot(spectra, ax, color, linestyle, padding=padding,
                               drawstyle=drawstyle)
         if legend is not None:
-            plt.legend(legend, loc=legend_loc)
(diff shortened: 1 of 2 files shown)


jlaehne (approved):

francisco-dlp:
While this fixes the issue, I don't understand why the issue was not caught by the test. Any idea @ericpre?

francisco-dlp:
I found the issue: when both `fig` and `ax` are given (as in the tests), there is no issue. I've fixed the tests to catch the bug.

ericpre (approved):

PR #3063, opened 2022-11-13
Remove deprecated components and legacy options

jlaehne:
### Description of the change
Remove the deprecated polynomial component and legacy options for arctan and Voigt components.

### Progress of the PR
- [x] add tests,
- [x] ready for review.


jlaehne:
As you wrote the new polynomial component, @thomasaarholt, your feedback on this PR would be valuable.

Do I understand it right that we still need the `convert_to_polynomial` function to properly load files saved with a model containing the old polynomial style? Then we need loading such a file, because so far this function was tested by creating an old polynomial, which is not possible anymore.

Similarly, we might need functions that check if a `arctan` or `Voigt` component that are loaded are in fact `EELSArctan` or `PESVoigt` components that now need to be converted?

ericpre:
> Similarly, we might need functions that check if a `arctan` or `Voigt` component that are loaded are in fact `EELSArctan` or `PESVoigt` components that now need to be converted?

Yes, I think so and we should add tests with model created from previous hyperspy version and make sure that we can read them correctly. I started to look at this last year but went a bit off track! ;) I think that the approach of this PR is correct and I added the test that I made previously in this branch: https://github.com/ericpre/hyperspy/tree/deprecation-components. It is easy to make a mistake between the different version and API changes, so we will need to be careful with test correctness, etc. With the test added in https://github.com/ericpre/hyperspy/tree/deprecation-components, there are errors when loading model saved with hyperspy 1.4 and 1.5 because the component are not converted when loading the model.

jlaehne:
> With the test added in https://github.com/ericpre/hyperspy/tree/deprecation-components, there are errors when loading model saved with hyperspy 1.4 and 1.5 because the component are not converted when loading the model.

I added conversion of legacy `Arctan` and `Voigt` when reloading the components and activated the tests for older file versions. Thanks for creating those files @ericpre

thomasaarholt (approved):
Great job! This all looks good to me!

And sorry that I haven't participated before!

doc/user_guide/model/fitting_model.rst
```diff
@@ -93,7 +93,7 @@ noise to it:
 
 .. code-block:: python
 
-    >>> s = hs.signals.Signal1D(np.arange(100, 300))
+    >>> s = hs.signals.Signal1D(np.arange(100, 300, dtype='float32')))
```
ericpre, line 96: Why is this change necessary?

hyperspy/_components/polynomial.py
ericpre, line 1: Should the following be removed from this file?
```python
        if "legacy" in kwargs:
            del kwargs["legacy"]
```

hyperspy/tests/component/test_arctan.py
```diff
@@ -32,26 +32,10 @@ def test_function():
     np.testing.assert_allclose(g.function(1), 0)
     np.testing.assert_allclose(g.function(1e4), 10*np.pi/2,1e-4)
 
-# Legacy tests
-def test_function_legacyF():
-    g = Arctan(minimum_at_zero=False)
+def test_function_delattr():
+    g = Arctan(minimum_at_zero=True, legacy=True)
```
ericpre:
```suggestion
    g = Arctan(minimum_at_zero=True)
```

hyperspy/tests/component/test_components.py
```diff
@@ -383,11 +298,14 @@ def test_3d_signal(self):
...
         s = self.m.as_signal()
         s = hs.stack([s]*2)
-        p = hs.model.components1D.Polynomial(order=2, legacy=False)
+        p = hs.model.components1D.Polynomial(order=2)
         p.estimate_parameters(s, None, None, only_current=False)
         axis = s.axes_manager.signal_axes[0]
         np.testing.assert_allclose(p.function_nd(axis.axis), s.data)
 
+    def test_function_legacy(self):
+        p = hs.model.components1D.Polynomial(order=2,legacy=True)
```
ericpre: Similar as for the `test_voigt.py` below, I am not sure this is necessary.

hyperspy/tests/component/test_voigt.py
```diff
@@ -39,20 +39,30 @@ def test_function():
...
 
 
+def test_function_legacy():
+    g = Voigt(legacy=True)
+    g.area.value = 5
+    g.sigma.value = 0.5
+    g.gamma.value = 0.2
+    g.centre.value = 1
+    np.testing.assert_allclose(g.function(0), 0.78853024)
+    np.testing.assert_allclose(g.function(1), 2.97832092)
```
ericpre: Now that we know that the behaviour is correct, I think that we should remove this test as we are testing for future regression.

hyperspy/_components/eels_arctan.py
ericpre, line 1: Remove the following?
```python
        # Not to break scripts once we remove the legacy Arctan
        if "minimum_at_zero" in kwargs:
            del kwargs["minimum_at_zero"]
```

hyperspy/_components/arctan.py
```diff
@@ -53,6 +53,8 @@ class Arctan(Expression):
 
     def __init__(self, A=1., k=1., x0=1., module=["numpy", "scipy"], **kwargs):
         # Not to break scripts once we remove the legacy Arctan
+        if "legacy" in kwargs:
+            del kwargs["legacy"]
         if "minimum_at_zero" in kwargs:
             del kwargs["minimum_at_zero"]
```
ericpre:
```suggestion
```

CHANGES.rst
```diff
@@ -24,6 +24,13 @@ Axes
 - :py:meth:`AxesManager.set_signal_dimension` is removed, use :py:meth:`~.signal.BaseSignal.as_signal1D`,
   :py:meth:`~.signal.BaseSignal.as_signal2D` or :py:meth:`~.signal.BaseSignal.transpose` of the signal instance instead.
 
+Components
+----------
+
+- The :py:class:`hyperspy._components.polynomial_deprecated.Polynomial` component is removed.
```
ericpre: From the user perspective this may be misleading: it can be read as the `Polynomial` component has been removed, it is its API, which has changed. 

Something along the line of:
```suggestion
- The API of the :py:class:`~.api.model.components1D.Polynomial` has changed after it has been deprecated since hyperspy 1.5. The old API had a single parameters ``coefficients`` and it has been replace with ``a0``, ``a1``, etc.
```

CHANGES.rst
```diff
@@ -24,6 +24,13 @@ Axes
...
+- The `legacy` option for :py:class:`hyperspy._components.arctan.Arctan` is removed, :py:class:`hyperspy._components.eels_arctan.EELSArctan` is now an independent component.
+- The `legacy` option for :py:class:`hyperspy._components.voigt.Voigt` is removed, :py:class:`hyperspy._components.pes_voigt.PESVoigt` is now an independent component.
```
ericpre:
```suggestion
- The `legacy` option (introduced in HyperSpy 1.6) for :class:`~.api.model.components1D.Arctan` is removed, use :class:`hyperspy._components.eels_arctan.EELSArctan` to use the old API.
- The `legacy` option (introduced in HyperSpy 1.6) for :class:`~.api.model.components1D.Voigt` is removed, use :class:`~.api.model.components1D.PESVoigt` to use the old API.
```

hyperspy/model.py
```diff
@@ -107,6 +107,16 @@ def _twinned_parameter(parameter):
 
 
 def reconstruct_component(comp_dictionary, **init_args):
+    # Restoring of Voigt and Arctan components saved with Hyperspy <v1.6
```
ericpre, line 110: I am wondering, if this "conversion" should be logged at the `INFO` level to avoid doing things silently?

hyperspy/_components/polynomial.py
```diff
@@ -146,13 +145,11 @@ def convert_to_polynomial(poly_dict):
     Convert the dictionary from the old to the new polynomial definition
```
ericpre, line 142: Similar as below: log this "conversion" in the logger at the INFO level to avoid doing things silently?

ericpre (commented):
This looks very good. I suggested not to test anymore against legacy code, because we know that we are assessing now that everything works as expected and we only need to keep testing against future regression.

@thomasaarholt, @jlaehne, did you checked that there is no mistake in the file and the test that I provided? This kind of thing is error prone and it would be good that it is checked by different pair of eyes!

doc/user_guide/model/fitting_model.rst
jlaehne: Otherwise, the data array was initialized as `'int'` and the following lines of code failed on my notebook.

hyperspy/_components/eels_arctan.py
jlaehne: This is required for reading of old files, which contain this parameter now not available in the component (comment updated)

hyperspy/_components/arctan.py
jlaehne: as above, `'legacy'` removed, but `'mimimum_at_zero'` necessary for file reading

hyperspy/_components/polynomial.py
jlaehne: is already logged by the `convert_to_polynomial` function

jlaehne:
> This looks very good. I suggested not to test anymore against legacy code, because we know that we are assessing now that everything works as expected and we only need to keep testing against future regression.

Thanks for the comments @ericpre , most are implemented. 

> @thomasaarholt, @jlaehne, did you checked that there is no mistake in the file and the test that I provided? This kind of thing is error prone and it would be good that it is checked by different pair of eyes!

I did not check all details for the test files, but I read the dictionaries using rsciio and had a look at the components there during the process of properly setting up the code to decide on what component to create out of the old file versions and for checking that the `legacy` parameter is not saved there and how to handle the `minimum_at_zero`. From what I have seen there, the files do make sense.

hyperspy/_components/polynomial.py
ericpre: Yes, indeed! 🤦‍♂️ :sweat_smile:

ericpre (approved):
Thank you @jlaehne, this looks all good!

PR #3067, opened 2022-11-23
Pass only positive numbers to closest_nice_number

edwinsupple:
### Description of the change
closest_nice_number takes a log and so assumes the input is positive. If the x axis scale value is negative (which is the case for some Talos .ser files), self.xmax-self.xmin is negative. I took the absolute value in scalebar.py here so that closest_nice_number does not return a domain error.

### Progress of the PR
- [ ] add tests,
- [ ] ready for review.

This would previously return a domain error:
```python
import hyperspy.api as hs
import numpy as np
dp = hs.load(filenames='...')

if dp.axes_manager[0].scale > 0:
    dp.axes_manager[0].scale*=-1
dp.plot()
```

### DIFF
--- a/hyperspy/drawing/_widgets/scalebar.py
+++ b/hyperspy/drawing/_widgets/scalebar.py
@@ -18,7 +18,7 @@
 
 
 from hyperspy.misc.math_tools import closest_nice_number
-
+import numpy as np
 
 class ScaleBar(object):
 
--- a/hyperspy/drawing/_widgets/scalebar.py
+++ b/hyperspy/drawing/_widgets/scalebar.py
@@ -95,8 +95,8 @@ class ScaleBar(object):
 
     def calculate_size(self, max_size_ratio=0.25):
         ps = self.pixel_size if self.pixel_size is not None else 1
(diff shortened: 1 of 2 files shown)


ericpre:
Can you provide a small test file to add to the test suite?

This fix will also need to be included in https://github.com/hyperspy/rosettasciio.

edwinsupple:
I added a unit test for my change. `calculate_size()` is not well tested otherwise.

I couldn't find the changed function anywhere in rosettasciio. If you can point me towards what needs changing happy to take another look otherwise I am going to leave it alone.

ericpre (approved):

ericpre:
Thank you! I am surprised that this bug was still present after so many year!

ericpre:
> I couldn't find the changed function anywhere in rosettasciio. If you can point me towards what needs changing happy to take another look otherwise I am going to leave it alone.

Sorry, my comment was incorrect: I misled myself thinking that a fix was required in the ser reader... in which case, the changes would have been necessary in rosettasciio too.

PR #3070, opened 2022-12-01
Bugfix: Do not raise AttributeError when loading markers

nem1234:
### Description of the change
- Do not raise AttributeError when loading markers
### Progress of the PR
- [ ] add tests,
- [ ] ready for review.

need Markers from rosettasciio

### DIFF
--- a/hyperspy/io.py
+++ b/hyperspy/io.py
@@ -33,7 +33,6 @@ from rsciio.utils.tools import overwrite as overwrite_method
 from rsciio import IO_PLUGINS
 
 from hyperspy import __version__ as hs_version
-from hyperspy.drawing.marker import markers_metadata_dict_to_markers
 from hyperspy.exceptions import VisibleDeprecationWarning
 from hyperspy.misc.utils import strlist2enumeration, get_object_package_info
 from hyperspy.misc.utils import stack as stack_method
@@ -812,12 +811,6 @@ def dict2signal(signal_dict, lazy=False):
                     value = function(value)
                 if value is not None:
                     signal.metadata.set_item(mpattr, value)
-    if "metadata" in signal_dict and "Markers" in mp:
-        markers_dict = markers_metadata_dict_to_markers(
-            mp['Markers'],
-            axes_manager=signal.axes_manager)
-        del signal.metadata.Markers
-        signal.metadata.Markers = markers_dict
     return signal
 
 
--- a/upcoming_changes/3070.bugfix.rst
+++ b/upcoming_changes/3070.bugfix.rst
@@ -1,0 +1,1 @@
+Bugfix deleting uninitialized marker_dict when loading files with markers


ericpre:
It seems that this code path is actually never used (anymore?) because loading markers from `hspy` file is done in the `assign_signal_subclass`. I don't see any reason to keep two different code path to set permanent markers to signal and we should keep only one. We have two option to sort this out:
- temporary use you suggested fix, add a tests to cover this specific behaviour and sort it out in follow up PR.
- simplify the situation by keeping only one of the two code path

I would suggest to take the second approach as this would sort it out correctly at first go. I suspect that we may need to keep the `assign_signal_subclass` approach and remove the other [code block](https://github.com/hyperspy/hyperspy/blob/a68e1f0789831598636743a7afe8d740415c1b38/hyperspy/io.py#LL757-LL763), in which we may just need to change where/how the markers are passed in the dictionary.
The cause of confusion here is that the supported items of the dictionary used in `dict2signal` is explicitly defined.

nem1234:
Thank you.
Unfortunately I don't have enough time next week.
I'll check it may be in ten days.

ericpre:
@nem1234, I don't know what has happen with the history of this branch but it would be good to tidy it up by starting from the latest `RELEASE_next_major` branch and cherry-pick the relevant commits. :)
I am happy to do it and then force-push to your branch or you can do it yourself. Please let me know what would you prefer!

nem1234:
Cleaned up codes to adopt with the RELEASE_next_major branch.
Current problem is creating test data that is small enough.
Real data is too large to commit.

ericpre:
My comment above was incorrect, we do need this [code block](https://github.com/hyperspy/hyperspy/blob/a68e1f0789831598636743a7afe8d740415c1b38/hyperspy/io.py#LL757-LL763).

Otherwise, the following test doesn't work because the markers aren't converted to the expected format in the `metadata` and are still in the dictionary form as they have been saved!

```python
import hyperspy.api as hs
import numpy as np

m = hs.plot.markers.Point(x=5, y=5)
s.add_marker(m, permanent=True)

fname = "test.hspy"

s.save(fname)
s2 = hs.load(fname)
s2.plot()
```

Sorry for the noise, your original fix was the right one!
For this PR, a test as the one above is enough to check that the markers are working for a save/load cycle using the `hspy`, there is no need to add a file to the history, as it good enough to create it during the execution of the test.
For loading other markers (digitalmicrograph or JEOL) in https://github.com/hyperspy/rosettasciio/pull/65, small files will be necessary to check that the marker are parsed correctly from the raw data, as writing these files is not supported.

nem1234:
Thank you.
I tried with RELEASE_next_major branch, and in this branch, and I recognized that this PR is not needed in RELEASE_next_major branch because signal.metadata.Markers is always initialized when mp['Markers'] exists.
I retract this PR.

ericpre:
OK, I think that it would still be good adding a [test mentioned above](https://github.com/hyperspy/hyperspy/pull/3070#issuecomment-1400865902) (with some assertion on the content of `metadata.Markers`.) to cover the behaviour, which would have been broken with the current version of this PR. Also, is `del signal.metadata.Markers` necessary here? If not, I would remove it while we are at it.

nem1234:
I don't think "del signal.metadata.Markers" is needed here.
But I don't know the behaviour of Python's garbage collection.

Also I can't realize why ' "metadata" in signal_dict ' is checked in this block.
The value of signal_dict is not used in this block.

ericpre:
> I don't think "del signal.metadata.Markers" is needed here.

Assigning the value of an attribute in python without deleting it is a usual pattern in python. Looking briefly at the [`DictionaryTreeBrowser`](https://github.com/hyperspy/hyperspy/blob/ccbc65579d18b1cf51a7a92c9e33990531e60606/hyperspy/misc/utils.py#L189), I can't a reason why it would be any different here.

> Also I can't realize why ' "metadata" in signal_dict ' is checked in this block.

The code is [not clearly written](https://github.com/hyperspy/hyperspy/blob/b7b6f6d369599d026720ce22f18b56c3f25056b3/hyperspy/io.py#L743-L821): I suspect this is in case the key `metadata` is not present in `signal_dict` (not sure if this is possible/allowed, as until now, this was all internal code and no specific API was enforced but this may be changing with rosettasciio) and therefore `mp` will then not be defined, as it is defined in a previous `if` block... A more explicit syntax would be:
```python
# beginning of dict2signal function
mp = signal_dict.get('metadata')
...
# in place of "if "metadata" in signal_dict:"
if mp is not None: 
   ...

# in place of "if "metadata" in signal_dict and "Markers" in mp:"
if mp is not None and "Markers" in mp:
   ...
```

nem1234:
Fix codes like above comment.

nem1234:
I'm sorry I misunderstood the reopen behaviour of github
I can't reopen this, so I put new PR #3091

Issue #3075, opened 2022-12-09
Plotting many Markers is slow


CSSFrancis:
#### Describe the bug
I've found that plotting a lot of markers can be very slow. The problem seems to be a bug where a linear increase in the number of artists results in an exponential increase in the plotting time for some signal.  The result is a sluggish interface.  This is part of the reason that things like the find peaks interface can appear sluggish, especially when there are many different peaks identified.

I discussed this a little bit in #3031 

For example when I run the code:
```python
num_points =  400
points = np.random.randint(0, 100, (num_points, 2,10,10))

%%timeit
s = hs.signals.Signal2D(np.random.random((10,10,100,100)))
markers = [ hs.plot.markers.Point(x = p[0],y= p[1],color="blue",) for p in points]  
for m in markers:
    s.add_marker(m,plot_marker=False, permanent=True, render_figure=False)
s.plot(vmin=0, vmax=1)
```

It takes around around .2ms to plot with 10 markers.  Most of this time is related to setting up the signal.  For 200 markers that time has increased to 1.86 seconds.  Most of which is related to the `plot` function, more specifically the underlying `_plot_signal` function.

![download](https://user-images.githubusercontent.com/41125831/206768648-9271d48e-cae1-4040-a7e7-d197440db76e.png)

The issue is related to the repeated calls to the `plt.scatter` function.  The issue is similar when adding many patches to the plot as is the case with adding arrows.  For the most part it is less probable that someone is adding 100's of arrows to a plot. As opposed to adding 100's of points or lines. 

#### Expected behavior
Markers should behave in a way similar to ragged signals in hyperspy.  Not only does this streamline the progression from finding a feature --> creating an marker --> visualization but it greatly speeds up visualization by reducing the number of markers.

It also reduces the dependency that each navigation position needs to have the same number of markers.  A fact that as we see above is an extreme determent to plotting speed.  

For the point marker the change in the code is relatively small and unbreaking.

If we change this line of code:

https://github.com/hyperspy/hyperspy/blob/7f6b448b91da35b71774a52860c95ba69deeb41c/hyperspy/drawing/_markers/point.py#L89-L90

To:

```
self.marker.set_offsets(np.squeeze(np.transpose([self.get_data_position('x1'),self.get_data_position('y1')])))
```

The plotting time doesn't change with the number of markers and we can also use this syntax to plot markers. 

```python
x= np.empty((10,10), dtype=object)
y= np.empty((10,10), dtype=object)

[...]

PR #3076, opened 2022-12-09
Vectorized Markers

CSSFrancis:
### Description of the change
Change described by #3075 

- Allows ragged point Markers initialized
- Allowed ragged line segment markers initialized. 
- Speeds up plotting many markers by ~100 times for plots with 100 or more active markers. 
- Allows for ragged initialization of points/ markers.

### Progress of the PR
- [x] add tests,
- [x] ready for review.

```python
x= np.empty((10,10), dtype=object)
y= np.empty((10,10), dtype=object)

points = np.random.randint(0, 100, (2,10,10, 500))

for i in np.ndindex(10,10):
    x_ind = (0,)+i
    y_ind = (1,)+i
    x[i]=points[x_ind]
    y[i]=points[y_ind]


s = hs.signals.Signal2D(np.random.random((10,10,100,100)))
markers =hs.plot.markers.Point(x =x,y=y,color="blue",)   # only adds one ragged marker object to the dataset. 
s.add_marker(markers,plot_marker=False, permanent=True, render_figure=False)
s.plot(vmin=0, vmax=1)
```


CSSFrancis:
@ericpre or @francisco-dlp What are your initial thoughts on the markers class being rewritten as a subclass of a signal.  Specifically a ragged signal?

This PR fixes some of the issues with slowness related plotting but there is still a pretty major missing functionality in that markers can't be lazy/ plotted lazily.

For large 4-D STEM datasets this is fairly crippling to the Strain mapping/ diffraction spot finding code as you can't use any of the interactive features without the system grinding to a halt and it reduces your ability to do any iterative discovery.  

Making markers a subclass of a Signal is I think a fairly elegant solution and gets ride of some of the major current issues.

1 - It allows for the ability to return a marker signal from a map function
2 - Markers are much more easily edited in similar form as signals (exclude markers near edges etc.)
3 - The navigation dimensions are consistent with the navigation dimensions of the signal (no reversing dimensions)
4 - Markers can be lazy and plotted alongside signals using the same lazy operations (caching etc.)

We can keep some of the old syntax for initializing markers but also have the ability to directly set the `.data` attribute as well when initializing. 

I feel like I've been circling around a couple of these points and wonder if it would be better for me to layout a full plan in a HSEP?  My hope was to bring together all of these parts over the winter break when I have some more time to devote to the change.

ericpre:
This is great that you are tackling this issue - indeed the markers are unusable when there are many...

> @ericpre or @francisco-dlp What are your initial thoughts on the markers class being rewritten as a subclass of a signal. Specifically a ragged signal?

Conceptually, markers shouldn't be signals, because their purpose is for plotting and there are not designed to do processing on them, for example, what would be the meaning of adding line of markers. That being said, they could be benefit of re-using *some of* the `BaseSignal` class to make handling of marker easier - maybe, slicing/indexing?

> Making markers a subclass of a Signal is I think a fairly elegant solution and gets ride of some of the major current issues.
> 
> 1 - It allows for the ability to return a marker signal from a map function

Conceptually, there are two different things. For example, in case of peak indexing: the signal contains the actual peak position and the markers are the representation of this signal. I would keep this distinction.

> 2 - Markers are much more easily edited in similar form as signals (exclude markers near edges etc.)
> 3 - The navigation dimensions are consistent with the navigation dimensions of the signal (no reversing dimensions)
> 4 - Markers can be lazy and plotted alongside signals using the same lazy operations (caching etc.)

I don't think that I follow this point. What is the advantage of having lazy markers? There are most of the time not large?

CSSFrancis:
@ericpre I realized I never got back to you.  I finally got back around to this problem and have some thoughts etc. 

> Conceptually, markers shouldn't be signals, because their purpose is for plotting and there are not designed to do processing on them, for example, what would be the meaning of adding line of markers. That being said, they could be benefit of re-using _some of_ the `BaseSignal` class to make handling of marker easier - maybe, slicing/indexing?

Slicing and indexing is a good thing to add.  It would be nice to say slice the navigation axes of a signal and retain the markers. As far as functions to markers go there are cases where things like shifting a marker or a bunch of markers might be very useful.
 
> Conceptually, there are two different things. For example, in case of peak indexing: the signal contains the actual peak position and the markers are the representation of this signal. I would keep this distinction.

There is a distinction here but it is very small.  There is no reason that the signal can't have the actual peak position.  In many cases due to how hyperspy prioritizes real units it is better to have the real units. This is something that it would be better to keep track of rather than to assume that the user is keeping track of.  In `pyxem` it is a real pain trying to separate real units and pixel units.... 

> I don't think that I follow this point. What is the advantage of having lazy markers? There are most of the time not large?

The major benefit of lazy makers isn't the size in memory rather the processing time. When you are working with lazy signals you don't want to compute things until you have to.  

The main workflow this is important for is something like the workflow below where you want to have a set of lazy peaks created from a method which runs lazily. While the same thing could be accomplished by slicing the data you lose a lot of what make lazy processing nice such as the ability to test small parts and then use the same code to process the whole dataset.  For plotting this also allows you to just move to a different section of the dataset and see if the same function works well. 

If you have a 100 Gb dataset it takes around 10 minutes to process so it is very hard to iterate quickly unless you have a fully lazy workflow.   This is part of the reason also why having markers as Signals would be nice as you could use `s.set_signal_type("marker")` to convert something to a marker. 

```
peaks = s.find_peaks()
print(peaks) # Lazy BaseSignal
lazy_marker = to_marker(peaks) # Lazy Marker
s.add_marker(lazy_marker)
s.plot()
```

CSSFrancis:
@ericpre So this is a bit more broken than I first anticipated and has been a point of pain and fusteration. 

I've been trying to make this consistent but ultimately it seems like there isn't a great way to handle all of the different markers.  The big problem is that incorporating text in any reasonable fashion means that things don't work well.

I think we should strip out the Text marker and have that be a separate class otherwise it forces the other markers to take lots of performance hits and stops them from being completely lazy etc. 

Then we could completely break the marker functionality and make it better but I don't know if that is the right thing to do.

CSSFrancis:
I think the best way to handle this would be to have the data variable either directly set or through kwargs.

For example:

```python

class MarkerBase(object):
...

    def set_data(self, data=None, **kwargs):
        if data is not None: # Case 1
            self.data = data
        elif any([kwargs[k].dtype == object for k in kwargs]): # Case 2
            shapes = np.array([np.array(kwargs[k]).shape for k in kwargs])
            data = np.empty(shape=shapes[0],
                                        dtype=object)
            for ind in np.ndindex[shapes[0]]:
                data[ind] = np.stack([kwargs[k] for k in kwargs],axis=1)  # Make into block array
            self.data = data
        else:
             self.data = np.stack([kwargs[k] for k in kwargs],axis=1)  # Make into block array
```

This does stop you from being able to define an iterating marker like this

`Point(x= range(3), y=range(3))` but we could add something like: `Point(x= range(3), y=range(3), iterating=True)`

Either way we would have to have the text marker defined through some other MarkerBase class.

CSSFrancis:
Closed in Favor of #3148

PR #3079, opened 2023-01-05
Fix test suite: add pooch as test dependency - required to use scipy.dataset

ericpre:
### Progress of the PR
- [x] Add pooch as test dependency - required to use scipy.datasets in latest scipy (1.10),
- [x] Fix warning when plotting 1D navigator with non-uniform axis.
- [x] Update tests to pass with latest reference dataset from scipy.datasets 
- [n/a] add tests,
- [x] ready for review.

### DIFF
--- a/hyperspy/signal.py
+++ b/hyperspy/signal.py
@@ -2784,6 +2784,13 @@ class BaseSignal(FancySlicing,
         elif self.tmp_parameters.has_item('filename'):
             self._plot.signal_title = self.tmp_parameters.filename
 
+        def sum_wrapper(s, axis):
+            with warnings.catch_warnings():
+                warnings.filterwarnings(
+                    "ignore", category=UserWarning, module='hyperspy'
+                    )
+                return s.sum(axis)
+
         def get_static_explorer_wrapper(*args, **kwargs):
             if np.issubdtype(navigator.data.dtype, np.complexfloating):
                 return abs(navigator(as_numpy=True))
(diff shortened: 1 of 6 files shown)


hyperspy/signal.py
```diff
@@ -2784,23 +2784,25 @@ def plot(self, navigator="auto", axes_manager=None, plot_markers=True,
         elif self.tmp_parameters.has_item('filename'):
             self._plot.signal_title = self.tmp_parameters.filename
 
+        def sum_wrapper(s, axis):
+            with warnings.catch_warnings():
+                warnings.filterwarnings(
+                    "ignore", category=UserWarning, module='hyperspy'
+                    )
+                return s.sum(axis)
```
CSSFrancis, line 2792: Is there any use for a generic warning suppressing function?  We could create a function that wraps some other function or be called explicitly. 

```python 

def suppress_warnings(func):
    def wrap(*args, **kwargs):
        with warnings.catch_warnings():
                warnings.filterwarnings(
                    "ignore", category=UserWarning, module='hyperspy'
                    )
                 return func(*args, **kwargs)
```

Maybe that is overkill though for a easy onetime fix.

CSSFrancis (commented):
This looks good for me.  Most of the changes are pretty small bug fixes.  I just had one small comment about the warning suppression and maybe thinking about making it more general.

hyperspy/signal.py
ericpre: Yes, this is a good idea, but  since this is used only in one place, I would prefer to keep it as it is. It is rare that we have such a situation: a call function in hyperspy which raise a warning by hyperspy!

CSSFrancis: Sounds good to me!

Issue #2815, opened 2021-08-26
How to include new generalised oscillator strength dataset


gguzzina:
We (at CEOS GmbH, together with the team of Prof. Kohl at the University of Münster) are working on computing a new database of Generalised Oscillator Strengths.
As this is long standing nuisance for people who want to perform quantitative EELS analysis, we think including support for it in hyperspy is a good idea.

The results we are getting for the moment are quite encouraging, but the dataset is not yet ready to be released. Still, I think it's a good time to gather technical ideas on how to do this.

My current proposal is to include data for all edges, appropriately downsampled, in a HDF5 archive.
When compared with the existing dataset, they cover a much higher "field of view" in both energy and momentum, and have better sampling on top of that, so it is a much larger dataset.
Nevertheless, at ~40MB it's not obscenely big by modern standards.

The HDF5 archive would then be distributed, probably by uploading it on Zenodo.

I would then write a new GOS class, probably a subclass of HartreeSlaterGOS, to use the new data, and also include patches to the hyperspy configuration dialog to let users select the file path.

Ideas and suggestions are very welcome. Does anyone have different preferences for the format of the dataset?
The discussion on the preferred way to distribute the data is ongoing with the team of Prof. Kohl, but ideas are welcome in that area as well.

gguzzina:
A little sneak peek :-) 
![25_Mn_L_0 005mrad](https://user-images.githubusercontent.com/63253/130943780-dbf42ed9-4fac-4ddb-a2c5-467a904d6054.png)

ericpre:
This sounds brilliant, I am sure many people will be looking forward this, me being the one!

> As this is long standing nuisance for people who want to perform quantitative EELS analysis, we think including support for it in hyperspy is a good idea.

Ineed, this is definitively a long standing issue!

A few thoughts about the technicalities:
- zenodo sounds very good
- I don't know if there is any standard specification for storing GOS dataset. If there isn't one, hdf5 with your own specifications would fit very well for the need here
- in hyperspy, I suspect that we should start using [pooch](https://www.fatiando.org/pooch) for downloading "large" test files and the implementation of this GOS component should use it. Since a few years, pooch seems to be well established in the python scientific community: it is used by scikit-image and co, which means that there is a community to support and it should be around for a while!
- we will need to keep track of the version of the dataset in the model component and being able to specify the version of dataset, when creating the model component to ensure reproducibility of the analysis.
[...]

PR #3082, opened 2023-01-13
Generalised oscillator strengths in GOSH format

gguzzina:
As discussed in #2815 a new dataset of Generalised Oscillator Strengths is now available at https://zenodo.org/record/6599071
Using this file finally removes the reliance on the older dataset formerly available through DM

Changes:
- add support for GOSH files (documented at https://gitlab.com/gguzzina/gosh) allowing to use the new dataset, and 
- The new GOS are implemented in a new file `gosh_gos.py`
- They can be selected by hand by using `create_model(GOS='gosh')`
- The new format is default, when available. `EELSSpectrum.create_model()` will try, in order, to use GOSH, then Hartree-Slater, then Hydrogenic.
- The path for the database file is set in `preferences.EELS.eels_gosh_database_path`

I based this on RELEASE_next_minor because the changes are very contained, and should not break backwards compatibility.

### Progress of the PR
- [x] update docstring,
- [x] update user guide,
- [x] add an changelog entry in the `upcoming_changes` folder,
- [x] add tests,
- [x] ready for review.

```python
import hyperspy.api as hs
import numpy as np
hl = hs.datasets.eelsdb(title="Hexagonal Boron Nitride", spectrum_type="coreloss")[0]
hl.add_elements(('B', 'N'))
hl.crop_signal1D(150.0, 610.0)

m = hl.create_model(GOS='gosh')
m.fit()
m.plot()
```


gguzzina:
@ericpre no pooch for now. That will have to wait.

hyperspy/misc/eels/hdf5_gos.py
```diff
@@ -0,0 +1,188 @@
...
+
+
+_logger = logging.getLogger(__name__)
+
+R = constants.value("Rydberg constant times hc in eV")
+a0 = constants.value("Bohr radius")
+
+class Hdf5GOS(GOSBase):
+
+    """Read Hartree-Slater Generalized Oscillator Strenght parametrized
```
ericpre: Typo?

hyperspy/misc/eels/hdf5_gos.py
```diff
@@ -0,0 +1,188 @@
...
+from hyperspy.misc.export_dictionary import (
+    export_to_dictionary, load_from_dictionary)
+
+
+_logger = logging.getLogger(__name__)
+
+R = constants.value("Rydberg constant times hc in eV")
+a0 = constants.value("Bohr radius")
+
+class Hdf5GOS(GOSBase):
```
ericpre:
```suggestion
class DFTGOS(GOSBase):
```
?

hyperspy/misc/eels/hdf5_gos.py
```diff
@@ -0,0 +1,188 @@
...
+        The tabulated energy axis
+    qaxis : array
+        The tabulated qaxis
+    energy_onset: float
+        The energy onset for the given element subshell as obtained
+        from iternal tables.
+
+    """
+
+    _name = 'Hartree-Slater'
```
ericpre: It needs updating.

hyperspy/_components/eels_cl_edge.py
```diff
@@ -55,8 +63,9 @@ class EELSCLEdge(Component):
         Usually a string, for example, 'Ti_L3' for the GOS of the titanium L3
         subshell. If a dictionary is passed, it is assumed that Hartree Slater
         GOS was exported using `GOS.as_dictionary`, and will be reconstructed.
-    GOS : {'hydrogenic', 'Hartree-Slater', None}
-        The GOS to use. If None it will use the Hartree-Slater GOS if
+    GOS : {'hydrogenic', 'GOS5', 'Hartree-Slater', None}
```
ericpre:
```suggestion
    GOS : {'hydrogenic', 'DFT', 'Hartree-Slater', None}
```

ericpre (commented):
This looks good, it simply need some test in https://github.com/hyperspy/hyperspy/blob/RELEASE_next_minor/hyperspy/tests/model/test_eelsmodel.py#L61!
To download the file, you can use something along the line of https://github.com/hyperspy/hyperspy/blob/f3e28eb4694ccdc46e38a841348d3658b920cb1f/hyperspy/tests/io/test_edax.py#L15-L40:

I suggested to use `DFT` instead of `GOS5` but there may be better name.

> @ericpre no pooch for now. That will have to wait.

No worries, we can do this in a follow up PR!

hyperspy/misc/eels/hdf5_gos.py
gguzzina: Well, of course I started by copying the HS file. I clearly did not look carefully enough for instances of the name.

gguzzina:
> I suggested to use `DFT` instead of `GOS5` but there may be better name.

I had actually started with that, but in the end I'm implementing compatibility for a file format, with no assumption on how the underlying data has been computed. While ours obviously have been computed by DFT, others may differ.

I don't particularly like GOS5 though. Suggestions are welcome.

francisco-dlp:
Glad to see this moving forward. Regarding the name, what about GOSH?

gguzzina:
GOSH is clever!
I think I could go with that, at the cost of changing it also in the Zenodo description, and the specifications repo.
Well, I'm computing a new version of the database anyway to include high loss edges.

ericpre:
> > I suggested to use `DFT` instead of `GOS5` but there may be better name.
> 
> I'd actually started with that, but in the end I'm implementing compatibility for a file format, with no assumption on how the underlying data has been computed. While ours obviously have been computed by DFT, others may differ.

Indeed in this case, there is two approaches you can take:
1. Have a base class (`Hdf5GOS`) supporting the format for the GOS and a child class `GOSH` or `GOS_DFT`, which is specific to this type of GOS and will know the path to the correct file. If one day, someone add another type of GOS, then another child class of `Hdf5GOS` can be created. In the [original issue](https://github.com/hyperspy/hyperspy/issues/2815#issuecomment-1165740767), @zezhong-zhang mentioned that they would interested in using it too, so that it may well happen sooner than later! ;)
2. Don't plan for hypothetical future addition and keep it as it is but with a name describing what it is instead of its format to be consistent with the rest of the code, e.g. `HartreeSlaterGOS` and `HydrogenicGOS`

> I don't particularly like GOS5 though. Suggestions are welcome.

To be fair, you are creating these GOS, so it is your call! The reason I suggested `DFT` is to be consistent with `Hydrogenic` and `Hartree-Slater`.

zezhong-zhang:
> 1. Have a base class (`Hdf5GOS`) supporting the format for the GOS and a child class `GOSH` or `GOS_DFT`, which is specific to this type of GOS and will know the path to the correct file. If one day, someone add another type of GOS, then another child class of `Hdf5GOS` can be created. In the [original issue](https://github.com/hyperspy/hyperspy/issues/2815#issuecomment-1165740767), @zezhong-zhang mentioned that they would interested in using it too, so that it may well happen sooner than later! ;)

@ericpre Yes, we can take the `DiracGOS` child class or `GOS_Dirac` to be consistent. The calling should be something like 

```python
m = hl.create_model(GOS='Dirac')
```

Our database will be released maybe in Feb. Will update the link in the community & pull request to get it supported. 

Good work  @gguzzina,  glad to see the integration with hyperspy!

gguzzina:
I'll go with some form of option 1 then, it seems the cleanest option for the moment.

I have a few ideas on how to make it all tidy again, in the unlikely event that GOS datasets were to suddenly proliferate.

gguzzina:
@ericpre @francisco-dlp 

On second thought, I don't think there should not be a class for every GOS available. I think there should be one class, supporting the file format, and the user should have an easy way to use any dataset they want. After all the entire point of making a standard file format is to have interoperability without changing the underlying software.

This should include the possibility to supply a file path, or to have one of these datasets downloaded for them using e.g. pooch.
Then all public ones can be included in a list of downloadable files that ships with hyperspy, and adding a new one is as hard as giving a file path.

Thoughts?

edit: I will however be using GOSH as the name for the file format.

ericpre:
Yes, this sounds good!

ericpre (approved):
I didn't check the math in the `integrateq` method but the rest looks good to me!

gguzzina:
> I didn't check the math in the `integrateq` method but the rest looks good to me!

Well, that's a 100% copy-paste from hartree_slater_gos.py

gguzzina:
The last commit corrects an error on my side, in the loading code. A normalisation factor was applied twice, but this would only become visible when the factor was different from 1.

I also published a new version of the dataset, including more edges. I will update the tests accordingly.

ericpre:
> > I didn't check the math in the `integrateq` method but the rest looks good to me!
> 
> Well, that's a 100% copy-paste from hartree_slater_gos.py

Ok, since the only method which is different is `readgosarray`/`readgosfile`, would it make sense to make another base class which contains all common methods and the child classes would only contain the method to read the GOS data?

gguzzina:
> Ok, since the only method which is different is `readgosarray`/`readgosfile`, would it make sense to make another base class which contains all common methods and the child classes would only contain the method to read the GOS data?

If we are doing that, we might as well make a few minor changes to HydrogenicGOS and merge that one in as well.

gguzzina:
I moved the integrateq method, present in both hartree_slater_gos and gosh_gos, into base_gos, the base class for all GOS.

hydrogenic_gos is just different enough that it's not trivial to include it without significant restructuring, and after two weeks of not doing it I just have to admit that I just don't have the time right now.

All tests still pass, but I'll conduct a few more tests by hand to ensure everything still works as before.

ericpre:
I tidied up a few things in this PR and got carried away:
- using pooch to download the data
- improving the structure of the class
- improving the coverage
- extending the `binding_energy` in the `elements` database to support the additional edges available in https://zenodo.org/record/7645765
- adding support for using different GOS from gosh file, see https://hyperspy--3082.org.readthedocs.build/en/3082/user_guide/eels.html#generalised-oscillator-strengths

@gguzzina, do you want to review my changes?

gguzzina:
@ericpre I'm happy to review your changes. Is there any urgency to this?
Right now I'm in over my head, but I also don't want to keep you guys waiting.

ericpre:
No there is no urgency. Do you have an approximate idea of when you would be able to review it?

gguzzina:
I'll try to find a little bit of time this week, otherwise week 23.

gguzzina:
I've gone through the changes, and it looks good for me!

ericpre:
Thanks @gguzzina, it is very nice to see this implemented!

PR #3091, opened 2023-01-26
Check existence of signal.metadata.Markers before deleting it (was PR #3070)

nem1234:
### Description of the change
- Do not raise AttributeError when loading markers
### Progress of the PR
- [ ] add tests,
- [ ] ready for review.

This error does not shown in RELEASE_next_major branch, but potentially raised when broken file reader is loaded.


upcoming_changes/3070.bugfix.rst
```diff
@@ -0,0 +1,3 @@
+Clean up codes around signal.metadata
```
ericpre:
```suggestion
Improve syntax in the `io` module.
```

hyperspy/io.py
```diff
@@ -812,12 +812,13 @@ def dict2signal(signal_dict, lazy=False):
                     value = function(value)
                 if value is not None:
                     signal.metadata.set_item(mpattr, value)
-    if "metadata" in signal_dict and "Markers" in mp:
-        markers_dict = markers_metadata_dict_to_markers(
+    if mp is not None and "Markers" in mp:
+        if "Markers" in signal.metadata:
+            del signal.metadata.Markers
```
ericpre:
```suggestion
```
What's about removing it completely? It is very very rare that we use `del` before an assignment in python and I don't see why it is necessary here. Looking at the history using git blame, it doesn't seem that there is a reason.

ericpre (commented):
I made some comment and it would be good to add a test along the line of:

```python
import hyperspy.api as hs
import numpy as np

m = hs.plot.markers.Point(x=5, y=5)
s.add_marker(m, permanent=True)

fname = "test.hspy"

s.save(fname)
s2 = hs.load(fname)
assert s.metadata.Markers == ...
s2.plot()
```

ericpre (approved):

PR #3093, opened 2023-01-28
Update metadata convention

### LINKED ISSUE #2095: Discussion: Hyperspy metadata SEM vs TEM
It is probably not an issue to most of the users, but I find the current hyperspy metadata structure to be a bit clumsy. My main issue is with SEM and TEM leafs. These are like singletons. It makes the tree unnecessary more complicated and adds additional level of depth. I have not seen even a single case, where acquisition instrument would have both SEM and TEM (and EELS and EDS under both of these). I think simplifying the tree structure would simplify parsing of files (particularly Bruker files where are no information about what kind of instrument the EDX detector is mounted on). SEM or TEM leaf could be refactored into simple attribute. i.e. `Acquisition_instrument.microscope_type`. The children of SEM/TEM could be moved one level up in the tree. Would this be possible in next_major version?

jlaehne:
### Description of the change
Adapt the metadata structure according to the discussion in #2095

### Progress of the PR
- [X] update metadata documentation, 
- [ ] Update occurences where metadata is accessed in HyperSpy,
- [ ] Create PR in RosettaSciIO to reflect the changes in the metadata structure,
- [ ] Check that tests are running on all ends,
- [ ] ready for review.

### DIFF
--- a/doc/reference/metadata.rst
+++ b/doc/reference/metadata.rst
@@ -20,63 +20,46 @@ in the following sections of this chapter.
 ::
 
     ├── Acquisition_instrument
-    │   ├── SEM
-    │   │   ├── Detector
-    │   │   │   ├── detector_type
-    │   │   │   └── EDS
-    │   │   │       ├── azimuth_angle (º)
-    │   │   │       ├── elevation_angle (º)
-    │   │   │       ├── energy_resolution_MnKa (eV)
-    │   │   │       ├── live_time (s)
-    │   │   │       └── real_time (s)
+    │   ├── Source
+    |   |   ├── source_type
+    |   |   (SEM/TEM)
     │   │   ├── beam_current (nA)
     │   │   ├── beam_energy (keV)
     │   │   ├── probe_area (nm²)
-    │   │   ├── convergence_angle (mrad)
     │   │   ├── magnification
     │   │   ├── microscope
-    │   │   ├── Stage
-    │   │   │   ├── rotation (º)
-    │   │   │   ├── tilt_alpha (º)
-    │   │   │   ├── tilt_beta (º)
-    │   │   │   ├── x (mm)
-    │   │   │   ├── y (mm)
-    │   │   │   └── z (mm)
     │   │   └── working_distance (mm)
-    │   └── TEM
-    │       ├── Detector
-    │       │   ├── EDS
-    │       │   │   ├── azimuth_angle (º)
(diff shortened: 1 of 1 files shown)


jlaehne:
In https://github.com/hyperspy/rosettasciio/issues/89#issuecomment-1433273678 it was suggested to actually migrate to the [NeXus metadata convention for electron microscopy](https://manual.nexusformat.org/classes/contributed_definitions/NXem.html), developed in the meantime, instead of improving our custom-scheme.

I'm copying in @sem-geologist @jordiferrero @CSSFrancis, as they were involved in the discussion in #2095.

jlaehne:
The main difference between the current proposal and the Nexus definition seems to be to me that the information I proposed to put in `source` actually moves another level up and directly resides in `instrument`, but some of the parameters we currently have there then go into the subnodes `EBEAM_COLUMN` and `OPTICAL_SYSTEM_EM`.

A rough overview is given in the following, though in the details the structure would need to be more thorougly revised.

Current proposal, which tried to be more generic concerning type of experiment (e.g. allowing also a laser as source without using a different definition):
```
├── Acquisition_instrument
│   ├── Source
│   │   ├── source_type
│   │   ├── beam_current (nA)
│   │   ├── beam_energy (keV)
│   │   ├── ...
│   │   ├── Stage
│   │   │   ├── rotation (º)
│   │   │   ├── ...
│   │   └── working_distance (mm)
│   ├── Detector
│   │   ├── ...
```

NeXus scheme (selected components), which is again more EM specific:
```
├── em_lab - NXinstrument
│   ├── EBAM_COLUMN - NXebeam_column
│   │   ├── probe (e.g. electrons)
│   │   ├── voltage (kV)
│   │   └── ...
│   ├── OPTICAL_SYSTEM_EM - NXoptical_system_em
│   │   ├── working_distance (mm)
│   │   └── ...
│   ├── Stage - NXStage
│   │   ├── rotation (º)
│   │   └── ...
│   ├── Detector - NXdetector
│   │   ├── ...
```

ericpre:
I am wondering if we should keep the metadata, as they are for HyperSpy 2.0 but plan for HyperSpy 3.0. Using the nexus definition for electron microscopy seems to be a very good option but it would take some time to do correctly. We need to get HyperSpy 2.0 released sooner than later and HyperSpy 3.0 could be focused solely on metadata changes.

ericpre:

PR #3099, opened 2023-02-08
update docstring of calibrate_energy_axis() in edsmodel

### LINKED ISSUE #3098: Improve `m.calibrate_energy_axis()` for EDS models
#### Describe the bug
Estimating the energy resolution with the method  `m.calibrate_energy_axis(calibrate='resolution')` can yield confusing results. The method takes the argument `xray_lines`, documented as: `The Xray lines. If 'all_alpha', fit all using all alpha lines`. The calibration method does two things. First it fits the width of all the specified lines, eventually all alpha lines. Secondly it calculates the FWHM of Mn_Ka, and this step has improvement possibilities.

The estimation of the Mn_Ka is done with the energy and FWHM of a reference peak in the users spectrum [Goldstein et al. (2003). "Scanning Electron Microscopy & X-ray Microanalysis", Plenum, third edition, p 315.].

1. [Documentation change] Calibration is done with `_set_energy_resolution()`, which calls on the function `_get_sigma()`. I would prefer if the documentation in `_get_sigma()`, more specifically the "Notes" in the docstring, to be a part of the method the user actually activates. The docstring in `m.calibrate_energy_axis(calibrate='resolution')` does not say how the calibration is done, which I think is nice to know. I can make a pull request for documentation change, but I thought I should ask if this change would be welcome.
2. [Code change] The function `_set_energy_resolution()` receives the argument `xray_lines`, but it really only uses `xray_lines[0]`. If the method is used with `xray_lines=all_alpha', the reference peak will just be the first alphabetical peak, and if the user specifies a list of peaks the reference will only be the first peak. 
      - Why does this matter? I have a SEM EDS spectrum of GaAs at 30, 15 and 10 kV with the following results:
           - 30 kV
                - `xray_lines='all_alpha'`, FWHM(Mn_Ka) = 139.9 eV
                - `xray_lines=['As_Ka']`, FWHM(Mn_Ka) = 139.8 eV
                - `xray_lines=['As_La']`, FWHM(Mn_Ka) = 131.2 eV
                - `xray_lines=['Ga_Ka']`, FWHM(Mn_Ka) = 133.6 eV
                - `xray_lines=['Ga_La']`, FWHM(Mn_Ka) = 130 eV
           - 15 kV, `xray_lines='all_alpha'`, FWHM(Mn_Ka) = 156.0 eV (It used the poorly defined peak As_Ka)
           - 10 kV, `xray_lines='all_alpha'`, FWHM(Mn_Ka) =129.7 eV
  
As the example shows, it matters what peak is used as reference peak in the function `_get_sigma()`.

All referenced HyperSpy code are in *hyperspy/hyperspy/models/edsmodel.py* 

#### To Reproduce
```
path = ''  # your value
file = ''  # your value
s = hs.load(path + '/' + file, signal_type='EDS_TEM')  # have to pretent it's a TEM signal
s.add_elements(['As', 'Ga'])
s.isig[zero_peak_end_index:]

m = s.create_model()
m.fit()
m.fit_background()

m.calibrate_energy_axis(calibrate='resolution')
print(s.metadata.Acquisition_instrument.TEM.Detector.EDS.energy_resolution_MnKa)
>>> 139.8

m.calibrate_energy_axis(calibrate='resolution', xray_lines=['Ga_Ka'])  # Given that you have Ga_Ka
print(s.metadata.Acquisition_instrument.TEM.Detector.EDS.energy_resolution_MnKa)
>>> 133.5
```

#### Expected behavior

One solution to the problem is to nudge the user harder to pick a line as argument for `xray_lines`. Another solution is to look at all the provided lines (either the list from the user or all alpha lines) and pick the line with highest total counts. I think at least for the `'all_alpha'` option the line should not just be the alphabetically first line. I am not sure if picking the line with highest total counts is most correct. Maybe some sort of average could be used, but then a threshold needs to be specified for well and poorly defined lines.

I can and will start working on this, but I want input on what sounds like the better solution here.

#### Python environement:
 - HyperSpy version: 1.7.3
 - Python version: 3.8.15

brynjarmorka:
### Description of the change
Addition to the docstring of `calibrate_energy_axis()` in `EDSmodel`

Issue: Improve m.calibrate_energy_axis() for EDS models- closes #3098

Is the update of the docstring clear enough? Or is it now too long?

### Progress of the PR
- [ ] ready for review.

### DIFF
--- a/hyperspy/models/edsmodel.py
+++ b/hyperspy/models/edsmodel.py
@@ -594,19 +594,25 @@ class EDSModel(Model1D):
                               **kwargs):
         """
         Calibrate the resolution, the scale or the offset of the energy axis
-        by fitting.
+        by fitting. The first Xray line in `xray_lines` is used as reference
+        peak, and should be well defined.
 
         Parameters
         ----------
         calibrate: 'resolution' or 'scale' or 'offset'
-            If 'resolution', fits the width of Gaussians place at all x-ray
+            If 'resolution', fits the width of Gaussians at provided Xray
             lines. The width is given by a model of the detector resolution,
             obtained by extrapolating the `energy_resolution_MnKa` in `metadata`.
-            This method will update the value of `energy_resolution_MnKa`.
+            Then, estimates and updates the energy resolution of the spectrum 
+            with the Fiori and Newbury (1978) formula, using the first Xray
+            line in `xray_lines` as reference peak. See `_get_sigma()` for
+            formula and literature reference.
             If 'scale', calibrate the scale of the energy axis.
             If 'offset', calibrate the offset of the energy axis.
         xray_lines: list of str or 'all_alpha'
-            The Xray lines. If 'all_alpha', fit all using all alpha lines
+            The Xray lines. If 'all_alpha', fit all using all alpha lines. 
+            The first Xray line in the list is used as reference peak, 
+            which is the alphabetically first line when 'all_alpha' is used.
         **kwargs : extra key word arguments
             All extra key word arguments are passed to fit or
             multifit, depending on the value of kind.
--- a/hyperspy/models/edsmodel.py
+++ b/hyperspy/models/edsmodel.py
@@ -190,7 +190,7 @@ class EDSModel(Model1D):
                 in self.xray_lines if xray_line.active]
 
     def add_family_lines(self, xray_lines='from_elements'):
-        """Create the Xray-lines instances and configure them appropiately
+        """Create the X-ray lines instances and configure them appropiately
 
         If a X-ray line is given, all the the lines of the familiy is added.
         For instance if Zn Ka is given, Zn Kb is added too. The main lines
@@ -333,7 +333,7 @@ class EDSModel(Model1D):
             component.active = False
 
     def _make_position_adjuster(self, component, fix_it, show_label):
-        # Override to ensure formatting of labels of xray lines
+        # Override to ensure formatting of labels of X-ray lines
         super(EDSModel, self)._make_position_adjuster(
             component, fix_it, show_label)
         if show_label and component in (self.xray_lines + self.family_lines):
@@ -407,7 +407,7 @@ class EDSModel(Model1D):
         Parameters
         ----------
         xray_lines: list of str or 'all_alpha'
-            The Xray lines. If 'all_alpha', fit all using all alpha lines
+            The X-ray lines. If 'all_alpha', fit all using all alpha lines
         """
         if xray_lines == 'all_alpha':
             xray_lines = [compo.name for compo in self.xray_lines]
@@ -434,7 +434,7 @@ class EDSModel(Model1D):
         Parameters
         ----------
         xray_lines: list of str or 'all_alpha'
-            The Xray lines. If 'all_alpha', fit all using all alpha lines
+            The X-ray lines. If 'all_alpha', fit all using all alpha lines
         """
         if xray_lines == 'all_alpha':
             xray_lines = [compo.name for compo in self.xray_lines]
@@ -468,7 +468,7 @@ class EDSModel(Model1D):
         Parameters
         ----------
         xray_lines: list of str or 'all_alpha'
-            The Xray lines. If 'all_alpha', fit all using all alpha lines
+            The X-ray lines. If 'all_alpha', fit all using all alpha lines
         """
         if xray_lines == 'all_alpha':
             xray_lines = [compo.name for compo in self.xray_lines]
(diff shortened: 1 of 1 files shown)


sivborg:
For anyone looking through this, the failing tests have nothing to do with this PR, which is a documentation update. 
Rather, they are due to the function `scipy.misc.ascent` being deprecated and moved to `scipy.datasets.ascent` in SciPy v1.10.0, giving a `DeprecationWarning`. This function is used in a couple of the tests, making them give errors.

hyperspy/models/edsmodel.py
```diff
@@ -595,19 +595,25 @@ def calibrate_energy_axis(self,
                               **kwargs):
         """
         Calibrate the resolution, the scale or the offset of the energy axis
-        by fitting.
+        by fitting. The first Xray line in `xray_lines` is used as reference
```
ericpre:
```suggestion
        by fitting. The first X-ray line in `xray_lines` is used as reference
```

hyperspy/models/edsmodel.py
```diff
@@ -595,19 +595,25 @@ def calibrate_energy_axis(self,
...
+        peak, and should be well defined.
```
ericpre: Can you define what's "well defined" mean? ;)

hyperspy/models/edsmodel.py
```diff
@@ -595,19 +595,25 @@ def calibrate_energy_axis(self,
...
 
         Parameters
         ----------
         calibrate: 'resolution' or 'scale' or 'offset'
-            If 'resolution', fits the width of Gaussians place at all x-ray
+            If 'resolution', fits the width of Gaussians at provided Xray
```
ericpre:
```suggestion
            If 'resolution', fits the width of Gaussians at provided X-ray
```

hyperspy/models/edsmodel.py
```diff
@@ -595,19 +595,25 @@ def calibrate_energy_axis(self,
...
             lines. The width is given by a model of the detector resolution,
             obtained by extrapolating the `energy_resolution_MnKa` in `metadata`.
-            This method will update the value of `energy_resolution_MnKa`.
+            Then, estimates and updates the energy resolution of the spectrum 
+            with the Fiori and Newbury (1978) formula, using the first Xray
```
ericpre:
```suggestion
            with the Fiori and Newbury (1978) formula, using the first X-ray
```

hyperspy/models/edsmodel.py
```diff
@@ -595,19 +595,25 @@ def calibrate_energy_axis(self,
...
+            line in `xray_lines` as reference peak. See `_get_sigma()` for
```
ericpre: Functions starting with an underscore are private and shouldn't be used in the documentation. It would be better to move the docstring of `_get_sigma` to the user guide or in this docstring, or both.

ericpre (commented):
Thanks @brynjarmorka for the pull request, please see some comments below.

hyperspy/models/edsmodel.py
brynjarmorka: What is the HS convention for writing 'X-ray lines'? In the docstrings in `edsmodel.py` file it is written as 'Xray lines' 15 times and once as 'Xray-lines', but never as 'X-ray lines'. Should I change all instances in the whole file to 'X-ray lines'?

I chose 'Xray lines' in my suggestion as it was the most used.

hyperspy/models/edsmodel.py
brynjarmorka: I do not have a definition, no. 

However, maybe the Fiori and Newbury equation in `_get_sigma()` could be used in "reverse" to determine if the peak is well defined? That is, if a (major) peak estimates a poor energy resolution, the peak is not well defined. Just a fun thought:)

hyperspy/models/edsmodel.py
brynjarmorka: Ok, I can remove the mention of the private function, and add both the reference and formula in "Notes" in the docstring.

brynjarmorka:
While doing some testing I found a small bug in the `'scale'` calibration too, where the input in 'xray_lines' was not really used. The old code sorts the alpha X-ray lines in the model by energy, and then selects the line with highest energy to calibrate the scale by finding its index. If the index is e.g. `10`, but the user just specifies one line, the function fails.

Also added a sentence in the docstring.

jlaehne:
@brynjarmorka could you rebase this on RELEASE_next_major as then the tests should be running through and we could include it in the v2.0 release.

ericpre:
Sorry, this one missed the `exspy` boat - didn't noticed that it was updated recently. Now it will need to be moved to [`exspy`](https://github.com/hyperspy/exspy). This [issue](https://github.com/hyperspy/exspy/issues/9) is opened not to forget about it.

PR #3101, opened 2023-02-14
check array type before explicit to_numpy conversion

Attolight-NTappy:
### Description of the change
I run a `cProfile `test of `multifit `and realized that in a simple case (single gaussian peak fit) close to 30% of the time is spent calling `hyperspy.misc.utils.to_numpy`. See the pink block in the screenshot below.

perfshot_b4

By explicitly checking if the data is of type `np.ndarray` in `signal.__call__` before calling `to_numpy` execution time goes down from 3.44 to 2.37 seconds (31% improvement) and tests pass. See screenshot below

perfshot_after

Let me know if I am missing on something there but if not I think I can identify other performance improvements along these lines in the multifit procedure. 

### Progress of the PR
- [x] add tests (No tests needed)
- [x] ready for review.

Performance profiling code. The generated `perf_multifit.prof` can be then visualised with `snakeviz`

```python
from hyperspy.datasets.artificial_data import get_luminescence_signal
import cProfile
import pstats

# import matplotlib.pyplot as plt
with cProfile.Profile() as pr:
    for i in range(10):
        s = get_luminescence_signal(navigation_dimension=2, uniform=True)
        m = s.create_model()
        g = Gaussian(A = 10000, centre = 300.0)
        m.append(g)
        m.multifit(iterpath='serpentine')

stats = pstats.Stats(pr)
stats.dump_stats(filename='perf_multifit.prof')
```

### DIFF
--- a/hyperspy/signal.py
+++ b/hyperspy/signal.py
@@ -2716,7 +2716,7 @@ class BaseSignal(FancySlicing,
             value = self._get_cache_dask_chunk(indices)
         else:
             value = self.data.__getitem__(indices)
-        if as_numpy:
+        if as_numpy and not isinstance(value,np.ndarray):
             value = to_numpy(value)
         value = np.atleast_1d(value)
         if fft_shift:


ericpre:
Thanks @Attolight-NTappy for the PR, I suggested a more generic alternative in https://github.com/hyperspy/hyperspy/pull/3109.

Attolight-NTappy:
Fine with me. I close

PR #3102, opened 2023-02-17
Fix matplotlib 3.7 and scikit-learn 1.4 deprecations

ericpre:
### Progress of the PR
- [x] Fix matplotlib 3.7 deprecation
        - [x] `register_cmap`
        - [x] `draw_all`  
        - [x] marker and widget line `set_data`,
        - [x] `button_press_event` callback,
        - [x] `figure.axes` now includes colorbars, which causes an error when double clicking on a axis of `hs.plot.plot_images`
- [x] fix scikit-learn 1.4 deprecation (set "n_init" default in test suite),
- [x] fix some tests in `test_plot_mva.py` 
- [x] add tests,
- [x] ready for review.


CSSFrancis:
@ericpre Thanks for doing this!  I looked into why the arrow keys were no longer working the other day, but never got around to fixing this.  I can review this if you want to ping me when you are done (that might be right now?)

ericpre:
> @ericpre Thanks for doing this! I looked into why the arrow keys were no longer working the other day, but never got around to fixing this. I can review this if you want to ping me when you are done (that might be right now?)

Regarding the arrow keys, there is a change in behaviour in the major branch, which was motivated by a conflict on the default matplotlib shortcuts: you need to use the `CRTL` key in addition to the arrow - https://hyperspy.org/hyperspy-doc/dev/changes.html#data-visualization.

Regarding this PR, I still need to cover some cases, but it should be close to be ready.

ericpre:
@CSSFrancis, this is ready to review.

CSSFrancis (approved):
This looks good to me.  A couple of general comments.

Is there a plan to drop support for 3.7 as well as 3.6. Python 3.7 was released 55 months ago and is about to go out of security support in 4 months.

Once 3.7 is dropped is the idea to remove the code related to supporting multiple versions of `matplotlib` with the next minor release?   We might want to make a note of what code should be removed with the dropping of 3.7 support.

ericpre:
As dropping python 3.7 doesn't simplify this PR, I didn't consider it. In the past we drop the support of a python version when it was beneficial, typically when adding support for a new python version to reduce number of build in the CI matrix, but otherwise, if there is no reason/benefit, we kept it as long as possible.

When support for matplotlib < 3.5/3.6 is dropped, there will be simplification in the code but this is too early.

The two limiting factors for support of dependencies version are:
- numba: usually the upper python version supported - at the time of writing, this python 3.10.
- google colab: lowest supported version for python (3.8)  matplotlib (3.2.2), scipy (1.7.3), scikit-learn (1.0.2)

CSSFrancis:
> As dropping python 3.7 doesn't simplify this PR, I didn't consider it. In the past we drop the support of a python version when it was beneficial, typically when adding support for a new python version to reduce number of build in the CI matrix, but otherwise, if there is no reason/benefit, we kept it as long as possible.

I see now, I misread the changes to the test.ymal file and assumed that matplot lib dropping support for 3.7 motivated some of the changes.

> When support for matplotlib < 3.5/3.6 is dropped, there will be simplification in the code but this is too early.

Sounds good!  


> The two limiting factors for support of dependencies version are:
> 
>     * numba: usually the upper python version supported - at the time of writing, this python 3.10.
> 
>     * google colab: lowest supported version for python (3.8)  matplotlib (3.2.2), scipy (1.7.3), scikit-learn (1.0.2)

This is good to know!

ericpre:
@CSSFrancis, do you want to merge it then? :)

CSSFrancis:
> @CSSFrancis, do you want to merge it then? :)

Sorry, still getting used to the ability to merge things :)

PR #3103, opened 2023-02-25
Fix `dask.array.random` 2023.2.1

ericpre:
Dask 2023.2.1 refactor the code to generate random numbers and now follows the syntax introduced in numpy 1.17 - see https://github.com/hyperspy/hyperspy/pull/3049 and https://numpy.org/doc/stable/reference/random/new-or-different.html#new-or-different for more details.

### Progress of the PR
- [x] Deprecate use of `numpy.random.RandomState` and start to use `np.random.default_rng` in test suite instead,,
- [x] Bump scipy minimum requirement to 1.4.0 to support `np.random.default_rng`,
- [x] update test depending on `numpy.random.RandomState`,
- [x] add tests,
- [x] ready for review.


ericpre:
@CSSFrancis, @jlaehne, @francisco-dlp, can one of you review this PR?

CSSFrancis:
I have a bit of free time and can do it now!

hyperspy/tests/signals/test_eels.py
```diff
@@ -268,7 +267,7 @@ def test_spikes_diagnosis(self):
         np.testing.assert_allclose(hist_data.data, expected_data)
 
         hist_data2 = self.signal._spikes_diagnosis(bins=25)
-        expected_data2 = np.array([285, 11, 13, 0, 0, 1, 12,  0])
+        expected_data2 = np.array([286, 10, 13, 0, 0, 1, 12,  0])
```
CSSFrancis, line 270: Will this test fail if you have an older version of numpy installed which uses the older random state model? If that is the case it might be worthwhile to add a comment. 

Why are the first two values shifted by one if the seed is the same as before?

hyperspy/misc/math_tools.py
```diff
@@ -174,47 +177,73 @@ def optimal_fft_size(target, real=False):
...
             np.random or dask.array.random
         If int:
-            Return a new RandomState instance seeded with ``seed``.
-        If np.random.RandomState:
-            Return it.
-        If dask.array.random.RandomState:
+            Return a new random state instance seeded with ``seed``.
+        If np.random.RandomState, np.random.Generator,
+        dask.array.random.RandomState or
             Return it.
```
CSSFrancis:
```suggestion
            Return seed.
```

hyperspy/misc/math_tools.py
```diff
@@ -174,47 +177,73 @@ def optimal_fft_size(target, real=False):
...
+            Return the random state singleton used by
             np.random or dask.array.random
         If int:
-            Return a new RandomState instance seeded with ``seed``.
-        If np.random.RandomState:
-            Return it.
-        If dask.array.random.RandomState:
+            Return a new random state instance seeded with ``seed``.
+        If np.random.RandomState, np.random.Generator,
+        dask.array.random.RandomState or
```
CSSFrancis:
```suggestion
        dask.array.random.RandomState:
```

hyperspy/misc/math_tools.py
```diff
@@ -174,47 +177,73 @@ def optimal_fft_size(target, real=False):
...
+            try:
+                return da.random.default_rng(seed)
+            except AttributeError:
+                return da.random.RandomState(seed)
+        else:
+            return np.random.default_rng(seed)
+
+    if isinstance(seed, (np.random.RandomState, da.random.RandomState)):
+        warnings.warn(
+            "Support for RandomState have been deprecated and will be removed "
```
CSSFrancis:
```suggestion
            "Support for RandomState generators have been deprecated and will be removed "
```

CSSFrancis (commented):
This seems like all pretty straight forward changes.  The only thing that I thought might be worth doing extra is adding a comment where you change a couple of the tests incase someone is testing with an older version of numpy and those tests start to fail.   It might give some idea of why that is the case.

hyperspy/tests/signals/test_eels.py
ericpre: This isn't possible because this is using the new numpy generators. It is still possible for the user to use legacy generators by specifying the seed and this would generate different results, but the default is using new generators.
In case of older version of dask, some tests are skipped for keep it simple.

CSSFrancis: Sounds good!

PR #3109, opened 2023-03-12
Speed up `to_numpy` to fix fitting slowdown

ericpre:
Slightly more generic alternative to https://github.com/hyperspy/hyperspy/pull/3101. This is happening when calling the `to_numpy` function many times, typically when fitting.

### Progress of the PR
- [x] check if array is of a `numpy.ndarray`  instance first to avoid going through the slower branch of the functions
- [x] add tests,
- [x] ready for review.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -47,6 +47,7 @@ hyperspy/misc/etc/test_compilers.obj
 *nbi
 hyperspy/tests/io/edax_files.zip
 .python-version
+*__pycache__
 
 ### Code ###
 # Visual Studio Code - https://code.visualstudio.com/
--- a/hyperspy/misc/utils.py
+++ b/hyperspy/misc/utils.py
@@ -1586,17 +1586,21 @@ def to_numpy(array):
 
     Raises
     ------
-    ValueError
-        If the provided array is a dask array
-
+    TypeError
+        When the input array type is not supported.
     """
-    if isinstance(array, da.Array):
-        raise LazyCupyConversion
-    if is_cupy_array(array):
+    if isinstance(array, np.ndarray):
+        return array
+    elif isinstance(array, da.Array):
+        raise TypeError(
+            "Implicit conversion of dask array to numpy array is not "
+            "supported, conversion needs to be done explicitely."
+            )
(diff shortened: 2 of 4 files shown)


jlaehne (approved):

jlaehne:
@Attolight-NTappy could you run your benchmark on this PR?

Attolight-NTappy:
Not as good but quite similar speedup.

I think the difference is marginal now.

image

jlaehne:
> Not as good but quite similar speedup.
> 
> I think the difference is marginal now.
> 
> image

Well, still 10% difference - were the two runs under exactly the same conditions? But the question is indeed whether one can catch it with such accuracy. One would possibly have to compare a longer fitting run on some dataset/model that takes a minute or so to fit.

ericpre:
If reproducible, this is true, that 10% for a large fitting is not negligible, it may be worth using both fixes (one generic, one specific), this one and #3101 with a comment why we need this redundancy.

jlaehne:
Sp add the line from #3101 here as well?

Attolight-NTappy:
Sorry for the confusion, I went too fast in my first message. The execution time from "to_numpy" function went down from 950 ms to 1 ms so we do win a lot there. In #3101 this last ms (which I assume is just the cumulated in-and-out time of the function) is avoided by not entering the function at all. However, the rest of the 10% exec. time is purely contextual so no need to worry there.

Imo this PR can be merged without further modification

ericpre:
Thanks @Attolight-NTappy for coming back on this. I looked more closely at the `BaseSignal.__call__` method using the following script and this is good enough for now.

```python
import hyperspy.api as hs
import numpy as np
from hyperspy.misc.utils import to_numpy

s = hs.datasets.artificial_data.get_core_loss_eels_signal()

%timeit isinstance(s.data, np.ndarray)
%timeit to_numpy(s.data)
%timeit s()
```

It should be possible to improve the time spent on `AxesManager.__next__` but this would for another time.

LMSC-NTappy:
No problem!

I agree with you, we can think of further improvements. I might take a look in the following weeks if I have time.

Cheers

Nico

PR #3120, opened 2023-03-30
Metadata mapping and signal assignment from .sur Signal format

Attolight-NTappy:
### Description of the change
This PR expands metadata parsing and signal assignments from .sur files. If an extension implementing "CL" as a signal type is found, the data is instantiated in this format and metadata are parsed, according to this specification:
https://docs.lumispy.org/en/latest/user_guide/metadata_structure.html

Implemented upon @jlaehne request

### Progress of the PR
- [x] add tests,
- [x] ready for review.

```python
import hyperspy.api as hs
import numpy as np
import pathlib
import hyperspy.tests.io as iotests

testmap = pathlib.Path(iotests.__file__).parent.joinpath('sur_data/test_spectral_map.sur')
sig = hs.load(testmap)

sig.medata
```

### DIFF
--- a/hyperspy/io_plugins/sur.py
+++ b/hyperspy/io_plugins/sur.py
@@ -72,7 +72,7 @@ class DigitalSurfHandler(object):
     Attributes
     ----------
     filename, signal_dict, _work_dict, _list_sur_file_content, _Object_type,
-    _N_data_object, _N_data_channels, _initialized
+    _N_data_object, _N_data_channels,
 
     Methods
     -------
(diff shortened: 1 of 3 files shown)


jlaehne:
Not sure, if we still need this here in the HyperSpy repository. Ideally, there will not be a minor release any more before HyperSpy 2.0.

Attolight-NTappy:
For me that's fine if you think it can be closed without merge

It depends on how you schedule release of 2.0.

jlaehne:
I think we are proceeding well with RosettaSciIO and there should be no more minor release before v2.0

PR #3126, opened 2023-04-12
is_binned optimisation during fit

Attolight-NTappy:
### Description of the change
The multifit optimisation story goes on.  Currently, up to 40% of the execution time is occupied by the evaluation of the `is_binned` axis property.

image

As shown in the screenshot above, this is a problem because it is repeatedly called during evaluation of the model function. I propose to evaluate once before starting `multifit`, temporarily store the result in an instance attribute and reset its value after the multifit is performed. This approach means to be as minimally intrusive as possible to the fitting functions, especially considering that they are so to say _on life support_ until 2.0.

After the changes, here are the performances

image

Execution goes down from 3.89 to 2.38 s

### Progress of the PR
- [ ] ready for review.

Test code has been updated to add shifts and amplitude variations to the gaussian. This way, the multifit is more representative of real data. Performances results can be visualised with snakeviz

```python
from hyperspy.datasets.artificial_data import get_luminescence_signal

import numpy as np
import cProfile
import pstats

# creating a luminescence signal
s = get_luminescence_signal(navigation_dimension=3, uniform=True)

#Adding a systematic and random shift
shift = (np.sin(np.arange(1000).reshape((10, 10, 10)) / 10 * 2 * np.pi) + 1.0) * 50
shift = shift + 10 * np.random.random(1000).reshape(10, 10, 10)
s.shift1D(shift)

#Adding a random intensity variation
s.data[:] = s.data * (2 + np.random.random(1000).reshape(10, 10, 10, 1))

#Create a model
m = s.create_model()
g = Gaussian(A=10000, centre=400.0)
m.append(g)

#Fitting with profiling
with cProfile.Profile() as pr:
    m.multifit(iterpath='serpentine')

#Dumping stats
stats = pstats.Stats(pr)
stats.dump_stats(filename='perf_multifit.prof')
```

### DIFF
--- a/hyperspy/model.py
+++ b/hyperspy/model.py
@@ -281,10 +281,16 @@ class BaseModel(list):
                 The Model that the event belongs to
             """, arguments=['obj'])
 
+        self.binned = None
+
     def __hash__(self):
         # This is needed to simulate a hashable object so that PySide does not
         # raise an exception when using windows.connect
         return id(self)
+
+    def __call__(self, non_convolved=False, onlyactive=False, component_list=None, binned=None):
+        """Evaluate the model numerically. Implementation requested in all sub-classes"""
+        raise NotImplementedError
 
     def store(self, name=None):
         """Stores current model in the original signal
@@ -942,7 +948,7 @@ class BaseModel(list):
     def _model_function(self, param):
         self.p0 = param
         self._fetch_values_from_p0()
-        to_return = self.__call__(non_convolved=False, onlyactive=True)
+        to_return = self.__call__(non_convolved=False, onlyactive=True, binned=self.binned)
         return to_return
 
     @property
@@ -1221,7 +1227,7 @@ class BaseModel(list):
 
     def _calculate_chisq(self):
         variance = self._get_variance()
-        d = self(onlyactive=True).ravel() - self.signal(as_numpy=True)[
+        d = self(onlyactive=True, binned=self.binned).ravel() - self.signal(as_numpy=True)[
             np.where(self.channel_switches)]
         d *= d / (1. * variance)  # d = difference^2 / variance.
         self.chisq.data[self.signal.axes_manager.indices[::-1]] = d.sum()
@@ -1902,6 +1908,8 @@ class BaseModel(list):
         maxval = self.axes_manager._get_iterpath_size(masked_elements)
         show_progressbar = show_progressbar and (maxval != 0)
 
+        self.binned = is_binned(self.signal)
+
         if linear_fitting:
             # Check that all non-free parameters don't change accross
             # the navigation dimension. If this is the case, we can fit the
@@ -1993,6 +2001,7 @@ class BaseModel(list):
                     para.map['std'] = para.std
                     para.map['is_set'] = True
 
+                self.binned = None
                 return
 
         i = 0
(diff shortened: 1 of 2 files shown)


hyperspy/model.py
```diff
@@ -281,11 +281,17 @@ def __init__(self):
                 The Model that the event belongs to
             """, arguments=['obj'])
 
+        self.binned = None
```
ericpre: Does it need to be public? Maybe add a comment to explain that it is used internally, how it is used and why?

ericpre:
Can you please compare with the `RELEASE_next_major` branch?

Attolight-NTappy:
> Can you please compare with the RELEASE_next_major branch?

Here it is. Indeed, no longer affected by `is_binned`. Overall performances is however slower.

image

ericpre:
Thank you @Attolight-NTappy, I was asking the situation on `RELEASE_next_major` branch because the `is_binned` function is removed in that branch, however, accessing the `is_binned` parameter from the axes still add a non-negligible penalty:
https://github.com/hyperspy/hyperspy/blob/a3cf2188fdad9a1bdc7032b143052e57462005f7/hyperspy/models/model1d.py#L435-L436

Also the binning penalty is most likely worse with non-uniform axis:
https://github.com/hyperspy/hyperspy/blob/a3cf2188fdad9a1bdc7032b143052e57462005f7/hyperspy/models/model1d.py#L438-L442

Maybe it would be also worth caching the normalisation factor related to binning when fitting?

hyperspy/model.py
```diff
@@ -1993,6 +2001,7 @@ def multifit(
                     para.map['std'] = para.std
                     para.map['is_set'] = True
 
+                self.binned = None
```
ericpre: Add a comment why does it need to be reset?

hyperspy/model.py
```diff
@@ -2031,6 +2040,8 @@ def multifit(
             _logger.info(f"Deleting temporary file: {autosave_fn}.npz")
             os.remove(autosave_fn + ".npz")
 
+        self.binned = None
```
ericpre: Add a comment as above?

Attolight-NTappy:
> however, accessing the is_binned parameter from the axes still add a non-negligible penalty

Indeed, I hadn't spotted it. I think similar improvement can be ported to 2.0. I can take care of it if you want. It is also the case for axes manager `__getitem__`. I also think that accessing the `traits.Property` have an access time 

> Also the binning penalty is most likely worse with non-uniform axis:
> Maybe it would be also worth caching the normalisation factor related to binning when fitting?

Probably yes. To be honest I have mixed feelings about binned signals altogether and long term I would be happy if we could move away from this paradigm. However, I understand that this is not me to decide :). Do you have an example of binned signal fitting that I could use as benchmark?

hyperspy/model.py
Attolight-NTappy: Now changed to private and comment added

hyperspy/model.py
Attolight-NTappy: Added a comment

hyperspy/model.py
Attolight-NTappy: Same comment added

ericpre:
Can you also rebase on `RELEASE_next_patch`?

Binned signal are necessary to correctly integrate the number of counts in case of EDS signal for example. I would expect that luminescence signal should be binned signal too.

Attolight-NTappy:
> Binned signal are necessary to correctly integrate the number of counts in case of EDS signal for example. I would expect that luminescence signal should be binned signal too.

OK, I see. We never work with binned signals in luminescence. Maybe we can discuss this in a dedicated thread ?

> Can you also rebase on RELEASE_next_patch?

Will do

Attolight-NTappy:
Rebased on RELEASE_next_patch in  #3130

Attolight-NTappy:
Closed as redundant with #3130

PR #3140, opened 2023-04-28
Horizontal Plotting using widgets

CSSFrancis:
### Overview
This is an extension to #2338.  It fixes the bug related to passing arguments to super that aren't in the original function.  By having those arguments implicitly handled by `**kwargs` it seems to fix the problem. This is built on the work previously done by @thomasaarholt 

## To Do:

- [x] Rebased on ReleaseNextMinor
- [x] Fix duplicating navigation axis
- [x] Add examples to the Documentation
- [x] Add tests?  I need to look into how to test widgets

This also allows more custom widget views which would be useful for things like plotting multiple signals. 

  ## Example:

```python

%matplotlib widget
import hyperspy.api as hs
import numpy as np

s = hs.signals.Signal2D(np.random.random((4,4,2,2)))
s.plot(plot_style="horizontal")
```
image

```python
%matplotlib widget
import hyperspy.api as hs
import numpy as np
from ipywidgets import Output, HBox, VBox, Layout
o = Output()
o2 = Output()
o3 = Output()
o4 = Output()
margin = "auto 0px auto 0px"
o.layout.margin = margin
o2.layout.margin = margin
o3.layout.margin = margin


s = hs.signals.Signal2D(np.random.random((4,4,2,2)))
s.plot(widget=o2, navigator_kwds={"widget":o})

s.isig[0,1].T.plot(widget=o3)

VBox([HBox([o, o3]), o2],layout=Layout(width='100%', display='flex' ,
align_items='center'))
```
image

### DIFF
--- a/doc/user_guide/visualisation.rst
+++ b/doc/user_guide/visualisation.rst
@@ -283,6 +283,12 @@ The same example with the feature disabled:
 
 .. _plot.customize_navigator:
 
+.. versionadded:: 2.0.0
+   ``plot_style`` keyword argument to allow for "horizontal" or "vertical" plotting when using the `ipympl` or
+`widget` backends. A default value can also be set using
+:ref:`HyperSpy plot preferences <configuring-hyperspy-label>`.
+
+
 Customizing the "navigator"
 ===========================
 
--- a/doc/user_guide/visualisation.rst
+++ b/doc/user_guide/visualisation.rst
@@ -284,8 +284,8 @@ The same example with the feature disabled:
 .. _plot.customize_navigator:
 
 .. versionadded:: 2.0.0
-   ``plot_style`` keyword argument to allow for "horizontal" or "vertical" plotting when using the
-   `ipympl` or `widget` backends. A default value can also be set using
+   ``plot_style`` keyword argument to allow for "horizontal" or "vertical" alignment of subplots (e.g. navigator 
+   and signal) when using the `ipympl` or `widget` backends. A default value can also be set using the
    :ref:`HyperSpy plot preferences <configuring-hyperspy-label>`.
 
 
--- a/doc/user_guide/visualisation.rst
+++ b/doc/user_guide/visualisation.rst
@@ -284,9 +284,9 @@ The same example with the feature disabled:
 .. _plot.customize_navigator:
 
 .. versionadded:: 2.0.0
-   ``plot_style`` keyword argument to allow for "horizontal" or "vertical" plotting when using the `ipympl` or
-`widget` backends. A default value can also be set using
-:ref:`HyperSpy plot preferences <configuring-hyperspy-label>`.
+   ``plot_style`` keyword argument to allow for "horizontal" or "vertical" plotting when using the
+   `ipympl` or `widget` backends. A default value can also be set using
+   :ref:`HyperSpy plot preferences <configuring-hyperspy-label>`.
 
 
 Customizing the "navigator"
--- a/hyperspy/defaults_parser.py
+++ b/hyperspy/defaults_parser.py
@@ -138,6 +138,9 @@ class PlotConfig(t.HasTraits):
     # Don't use t.Enum to list all possible matplotlib colormap to
     # avoid importing matplotlib and building the list of colormap
     # when importing hyperpsy
+    widget_plot_style = t.Enum(
+        ['horizontal', 'vertical'],
+        label='Widget plot style: (only with ipympl)')
     cmap_navigator = t.Str('gray',
                            label='Color map navigator',
                            desc='Set the default color map for the navigator.',
--- a/hyperspy/drawing/mpl_he.py
+++ b/hyperspy/drawing/mpl_he.py
@@ -187,34 +187,39 @@ class MPL_HyperExplorer(object):
                           "plotting backends are used.")
         plot_style = kwargs.pop("plot_style", None)
         # matplotlib plotting backend
+        def plot_sig_and_nav(plot_style):
+            if self.pointer is None:
+                pointer = self.assign_pointer()
+                if pointer is not None:
+                    self.pointer = pointer(self.axes_manager)
+                    self.pointer.color = 'red'
+                    self.pointer.connect_navigate()
+                self.plot_navigator(**kwargs.pop('navigator_kwds', {}))
+                if pointer is not None:
+                    self.navigator_plot.events.closed.connect(
+                        self.pointer.disconnect, [])
+            self.plot_signal(**kwargs)
+            if "ipympl" in backend:
+                if plot_style not in ["vertical", "horizontal", None]:
+                    raise ValueError("plot_style must be one of ['vertical', 'horizontal', None]")
+                if plot_style is None:
+                    plot_style = preferences.Plot.widget_plot_style
+                # If widgets do not already exist, we will `display` them at the end
+                from ipywidgets.widgets import HBox, VBox
+                from IPython.display import display
+                if not self.navigator_plot:
+                    display(self.signal_plot.figure.canvas)
+                elif plot_style == "horizontal":
+                    display(HBox([self.navigator_plot.figure.canvas,self.signal_plot.figure.canvas]))
+                else: # plot_style == "vertical":
+                    display(VBox([self.navigator_plot.figure.canvas, self.signal_plot.figure.canvas]))
         if "ipympl" in backend:
             import matplotlib.pyplot as plt
-            plt.ioff()  # plot ioff to hold plotting
-        if self.pointer is None:
-            pointer = self.assign_pointer()
-            if pointer is not None:
-                self.pointer = pointer(self.axes_manager)
-                self.pointer.color = 'red'
-                self.pointer.connect_navigate()
-            self.plot_navigator(**kwargs.pop('navigator_kwds', {}))
-            if pointer is not None:
-                self.navigator_plot.events.closed.connect(
-                    self.pointer.disconnect, [])
-        self.plot_signal(**kwargs)
-        if "ipympl" in backend:
-            if plot_style not in ["vertical", "horizontal", None]:
-                raise ValueError("plot_style must be one of ['vertical', 'horizontal', None]")
-            if plot_style is None:
-                plot_style = preferences.Plot.widget_plot_style
-            # If widgets do not already exist, we will `display` them at the end
-            from ipywidgets.widgets import HBox, VBox
-            from IPython.display import display
-            if not self.navigator_plot:
-                display(self.signal_plot.figure.canvas)
-            elif plot_style == "horizontal":
-                display(HBox([self.navigator_plot.figure.canvas,self.signal_plot.figure.canvas]))
-            else: # plot_style == "vertical":
-                display(VBox([self.navigator_plot.figure.canvas, self.signal_plot.figure.canvas]))
+            with plt.ioff():
+                plot_sig_and_nav(plot_style)
+        else:
+            plot_sig_and_nav(plot_style)
+
 
 
     def assign_pointer(self):
--- a/hyperspy/drawing/mpl_he.py
+++ b/hyperspy/drawing/mpl_he.py
@@ -20,10 +20,11 @@ from functools import partial
 import logging
 
 from traits.api import Undefined
+import matplotlib as mpl
 
 from hyperspy.drawing import widgets, signal1d, image
 from hyperspy.defaults_parser import preferences
-
+import warnings
 
 _logger = logging.getLogger(__name__)
 
@@ -180,6 +181,15 @@ class MPL_HyperExplorer(object):
         for key in ['power_spectrum', 'fft_shift']:
             if key in kwargs:
                 self.signal_data_function_kwargs[key] = kwargs.pop(key)
+        backend = mpl.get_backend()
+        if not "ipympl" in backend and "plot_style" in kwargs:
+            warnings.warn("The `plot_style` keyword is only used when the `ipympl` or `widget`"
+                          "plotting backends are used.")
+        plot_style = kwargs.pop("plot_style", None)
+        # matplotlib plotting backend
+        if "ipympl" in backend:
+            import matplotlib.pyplot as plt
+            plt.ioff()  # plot ioff to hold plotting
         if self.pointer is None:
             pointer = self.assign_pointer()
             if pointer is not None:
@@ -191,6 +201,21 @@ class MPL_HyperExplorer(object):
                 self.navigator_plot.events.closed.connect(
                     self.pointer.disconnect, [])
         self.plot_signal(**kwargs)
+        if "ipympl" in backend:
+            if plot_style not in ["vertical", "horizontal", None]:
+                raise ValueError("plot_style must be one of ['vertical', 'horizontal', None]")
+            if plot_style is None:
+                plot_style = preferences.Plot.widget_plot_style
+            # If widgets do not already exist, we will `display` them at the end
+            from ipywidgets.widgets import HBox, VBox
+            from IPython.display import display
+            if not self.navigator_plot:
+                display(self.signal_plot.figure.canvas)
+            elif plot_style == "horizontal":
+                display(HBox([self.navigator_plot.figure.canvas,self.signal_plot.figure.canvas]))
+            else: # plot_style == "vertical":
+                display(VBox([self.navigator_plot.figure.canvas, self.signal_plot.figure.canvas]))
+
 
     def assign_pointer(self):
         if self.navigator_data_function is None:
--- a/hyperspy/tests/drawing/test_horizontal_plotting.py
+++ b/hyperspy/tests/drawing/test_horizontal_plotting.py
@@ -1,0 +1,58 @@
+# Copyright 2007-2022 The HyperSpy developers
+#
+# This file is part of HyperSpy.
+#
+# HyperSpy is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# HyperSpy is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with HyperSpy. If not, see <https://www.gnu.org/licenses/#GPL>.
+import pytest
+
+import hyperspy.api as hs
+import numpy as np
+import matplotlib
+
+ipympl = pytest.importorskip("ipympl")
+ipywidgets = pytest.importorskip("ipywidgets")
+
+class TestIPYMPL:
+    def test_horizontal(self, capsys):
+        matplotlib.use("module://ipympl.backend_nbagg")
+        s = hs.signals.Signal2D(np.random.random((4, 4, 2, 2)))
+        s.plot(plot_style="horizontal")
+        captured = capsys.readouterr()
+        assert("HBox(children=(Canvas(" in captured.out)
+
+    def test_vertical(self,capsys):
+        matplotlib.use("module://ipympl.backend_nbagg")
+        s = hs.signals.Signal2D(np.random.random((4, 4, 2, 2)))
+        s.plot(plot_style="vertical")
+        captured = capsys.readouterr()
+
+        assert("VBox(children=(Canvas(" in captured.out)
+
+    def test_only_signal(self,capsys):
+        matplotlib.use("module://ipympl.backend_nbagg")
+        s = hs.signals.Signal2D(np.random.random(( 2, 2)))
+        s.plot()
+        captured = capsys.readouterr()
+        assert("Canvas(toolbar=Toolbar(" in captured.out)
+
+    def test_warnings(self,):
+        with pytest.warns(UserWarning):
+            s = hs.signals.Signal2D(np.random.random(( 4, 2, 2)))
+            s.plot(plot_style="vertical")
+
+    def test_misspelling(self, ):
+        matplotlib.use("module://ipympl.backend_nbagg")
+        with pytest.raises(ValueError):
+            s = hs.signals.Signal2D(np.random.random((4, 2, 2)))
+            s.plot(plot_style="Vert")
--- a/setup.py
+++ b/setup.py
@@ -68,7 +68,7 @@ install_req = [
 
 extras_require = {
     "learning": ["scikit-learn>=1.0.1"],
-    "gui-jupyter": ["hyperspy_gui_ipywidgets>=1.1.0"],
+    "gui-jupyter": ["hyperspy_gui_ipywidgets>=1.1.0", "ipympl"],
     # UPDATE BEFORE RELEASE
     "gui-traitsui": ["hyperspy_gui_traitsui @ git+https://github.com/hyperspy/hyperspy_gui_traitsui#egg=hyperspy_gui_traitsui"],
     #"gui-traitsui": ["hyperspy_gui_traitsui>=1.1.0"],
--- a/upcoming_changes/3140.new.rst
+++ b/upcoming_changes/3140.new.rst
@@ -1,0 +1,3 @@
+Changes to :meth:`~.api.signals.BaseSignal.plot`:
+- Allow Horizontal plotting using the `ipympl` or `widget` backends
+- Add in gui preferences to set a default to horizontal or vertical plotting


CSSFrancis:
Another note is that a preference can be set with `hs.preferences.Plot.widget_plot_style = 'horizontal'`

CSSFrancis:
Another useful example might be something like this where an interactive sum of some area in the sample is created and plotted beside the signal and the navigator.

```
%matplotlib widget
import hyperspy.api as hs
import numpy as np
from ipywidgets import Output, HBox,VBox
roi = hs.roi.CircleROI(cx=0.,cy=0, r_inner=0, r=1)
s = hs.signals.Signal2D(np.random.random((4,4,16,16)))
o = Output()
o2 = Output()
o3 = Output()
s.plot(widget=o2, navigator_kwds={"widget":o})
sliced_signal = roi.interactive(s, axes=s.axes_manager.signal_axes)

out = s.nansum( axis=(2,3)).T
out.metadata.General.title = "Integrated intensity"

interactive(sliced_signal.nansum,
            axis=sliced_signal.axes_manager.signal_axes,
            event=roi.events.changed,
            recompute_out_event=None,
            out=out,)
out.plot(widget=o3)

HBox([o,o2,o3])
```


image

CSSFrancis:
For some reason it also appears that it fails with python 3.8


Edit:  It appears that the newest version of ipython was released earlier today and doesn't support python 3.8 and lower.

CSSFrancis:
@ericpre  Any ideas of how to test something like this?

ericpre:
Add a notebook, run it and check some of the output?

CSSFrancis:
> Add a notebook, run it and check some of the output?

That was what I was thinking but it does come with some weird challenges so I was curious what you would suggest. That being said testing jupyter notebooks is __much much__ better than it was a couple of years ago. 

The [reviewNB/treon](https://github.com/ReviewNB/treon) app has some testing which tests output for some notebook which might be a good way to start testing.  I would recommend adding the [review-notebook-app](https://app.reviewnb.com/) for visualizing diffs as well.

It might be a good idea to test some other notebook functionality and widgets just to make sure there that the widgets and such are working properly.

ericpre:
In principle, functionality depending on `ipywidgets` needs to go in the `hyperspy_gui_ipywidgets` library and therefore there is no  requirement to test notebook in `hyperspy`, as in using some tool that the output each cell, etc. However, for plotting, this may not work as well and it may be ok to keep it here as long as it doesn't add any dependency.

With checking some of the output, I meant: putting a few assertion in the same notebook to check that the output contain matplotlib canvas, check the navigation and signal plot are in the correct output, for example by checking the data in the matplotlib figure, etc. I suspect that https://github.com/computationalmodelling/nbval will be good because we aren't interested in the content of the notebook itself and actually it would be better to avoid depending on this, as we don't want the test to fail because of small changes in ipympl or matplotlib, etc.

CSSFrancis:
@ericpre Kind of is sounding like there is no great solution... What about a test like this?


```python
import hyperspy.api as hs
import numpy as np
import matplotlib
from contextlib import redirect_stdout
import io

def test_horizontal():
    matplotlib.use("module://ipympl.backend_nbagg")
    f = io.StringIO()
    with redirect_stdout(f):
        s = hs.signals.Signal2D(np.random.random((4, 4, 2, 2)))
        s.plot(plot_style="vertical")

    assert("HBox(children=(Output(layout=Layout(margin='auto 0px auto 0px')),"
           " Output(layout=Layout(margin='auto 0px auto 0px')" in f.getvalue())

def test_vertical():
    matplotlib.use("module://ipympl.backend_nbagg")
    f = io.StringIO()
    with redirect_stdout(f):
        s = hs.signals.Signal2D(np.random.random((4, 4, 2, 2)))
        s.plot(plot_style="vertical")

    assert("VBox(children=(Output(layout=Layout(margin='0px auto 0px auto')),"
           " Output(layout=Layout(margin='0px auto 0px auto')" in f.getvalue())
```

Kind of a weird sudo test.  It will test that the code runs without error but it don't think it actually tests that anything is plotted or that the output is properly captured.  It also tests to make sure that the properly plot style is handled. 

That way we can just add ipympl as an extra dependency for testing only.

CSSFrancis:
@ericpre Any idea why are Azure-pipeline tests are failing?

ericpre:
There is a failure with one of the image comparison for the plotting test, which seem to be a regression from this PR: the output can be download at https://dev.azure.com/franciscode-la-pena-manchon/hyperspy/_build/results?buildId=3154&view=artifacts&pathAsName=false&type=publishedArtifacts

CSSFrancis:
@ericpre So I looked at the two images and there doesn't appear to be an `image_diff` published for the failing tests.  

These tests should be covered by the github actions as well, correct? And they are passing there.  I'm just trying to wrap my head around what the problem could be.  

The error thrown is:

```python

    def compare_image_to_baseline(self, item, fig, result_dir, summary=None):
        """
        Compare a test image to a baseline image.
        """
        from matplotlib.image import imread
        from matplotlib.testing.compare import compare_images
    
        if summary is None:
            summary = {}
    
        compare = get_compare(item)
        tolerance = compare.kwargs.get('tolerance', 2)
        savefig_kwargs = compare.kwargs.get('savefig_kwargs', {})
    
        baseline_image_ref = self.obtain_baseline_image(item, result_dir)
    
        test_image = (result_dir / "result.png").absolute()
>       fig.savefig(str(test_image), **savefig_kwargs)
E       AttributeError: 'NoneType' object has no attribute 'savefig'
```

But it seems like the figure is saved to the output file.

ericpre:
Image comparison only runs on azure pipeline and this is why it passes on github CI. The two images have different aspect ratio, which may be why there is no diff?

CSSFrancis:
Yea it does appear to be failing due to a mismatch in size. Apparently the test's don't actually run in that case so no diff is created.

hyperspy/tests/drawing/test_horizontal_plotting.py
```diff
@@ -0,0 +1,70 @@
...
+import pytest
+
+import hyperspy.api as hs
+import numpy as np
+import matplotlib
+
+ipympl = pytest.importorskip("ipympl")
+ipywidgets = pytest.importorskip("ipywidgets")
+from ipywidgets import Output
+from IPython.utils.capture import capture_output
```
ericpre: Can you use the pytest `capsys` fixture instead as this is a more generic pattern which can be used elsewhere (e.g. https://github.com/hyperspy/hyperspy/pull/3145/files#diff-ff280472a5fbc9ecab1b2116f11df993c1bcc5b6ffd5d5211a34b9614e3913f8), without depending on ipython?
See https://docs.pytest.org/en/7.1.x/how-to/capture-stdout-stderr.html#accessing-captured-output-from-a-test-function

hyperspy/defaults_parser.py
```diff
@@ -138,6 +138,9 @@ class PlotConfig(t.HasTraits):
     # Don't use t.Enum to list all possible matplotlib colormap to
     # avoid importing matplotlib and building the list of colormap
     # when importing hyperpsy
+    widget_plot_style = t.Enum(
+        ['horizontal', 'vertical'],
+        label='Widget plot style: (requires jupyter-matplotlib)')
```
ericpre:
```suggestion
    widget_plot_layout = t.Enum(
        ['horizontal', 'vertical'],
        label='Widget plot style (only with ipympl)')
```

hyperspy/drawing/image.py
```diff
@@ -324,10 +325,15 @@ def create_axis(self):
             self.ax.axis('off')
 
     def plot(self, data_function_kwargs={}, **kwargs):
+        """Plot the figure
+
+        Arguments:
```
ericpre: Use numpydoc style.

hyperspy/drawing/mpl_he.py
```diff
@@ -180,17 +180,82 @@ def plot(self, **kwargs):
...
-                self.navigator_plot.events.closed.connect(
-                    self.pointer.disconnect, [])
-        self.plot_signal(**kwargs)
+
+        navigator_kwds = kwargs.pop('navigator_kwds', {})
+
+        # widgets can be specified to contain the figure
+        navigator_widget = navigator_kwds.get('widget', None)
+        signal_widget = kwargs.get('widget', None)
+
```
ericpre: In case `navigator_widget` or `signal_widget` are defined but the `ipympl` is not used, it should raise a warning to avoid being ignore silently.

hyperspy/drawing/utils.py
```diff
@@ -155,13 +155,22 @@ def create_figure(window_title=None,
...
+    widget : ipywidgets output widget to place the figure in
 
     Returns
     -------
     fig : plt.figure
 
     """
-    fig = plt.figure(**kwargs)
+    if "autoscale" in kwargs:
+        raise ValueError
```
ericpre: Is it a left over from rebase?

hyperspy/drawing/utils.py
```diff
@@ -155,13 +155,22 @@ def create_figure(window_title=None,
...
+    if widget:
+        from IPython.display import display
```
ericpre: Note for making IPython an optional dependencies: https://github.com/hyperspy/hyperspy/pull/3145.

hyperspy/drawing/mpl_hie.py
```diff
@@ -31,6 +31,10 @@ def plot_signal(self, **kwargs):
             The kwargs are passed to plot method of the image figure.
 
         """
+
+        if self.signal_plot is not None:
+            self.signal_plot.plot(**kwargs)
+            return
```
ericpre: Is it some left over from the rebase?

CSSFrancis:
@ericpre This is good now I think.  Or at least worth a pass through. 

I can clean up the commit history a little bit but that requires editing some of @thomasaarholt's work so I would rather not just for the sake of completeness. I rebased this onto RELEASE_next_major but it would be nice to have this earlier rather than later. 

Specifically there are a couple of presentations coming up in the next couple of months where this functionality would be very helpful to present and visualize data.

hyperspy/drawing/mpl_he.py
```diff
@@ -180,17 +180,82 @@ def plot(self, **kwargs):
...
+                if pointer is not None:
+                    self.pointer = pointer(self.axes_manager)
+                    self.pointer.color = 'red'
+                    self.pointer.connect_navigate()
+                self.plot_navigator(**navigator_kwds)
+                if pointer is not None:
+                    self.navigator_plot.events.closed.connect(
+                        self.pointer.disconnect, [])
+            self.plot_signal(**kwargs)
+        if "ipympl" in backend:
```
ericpre: This can go in the previous `if` block and remove the `backend` variable, simply use `"ipympl" in matplotlib.get_backend()` to simplify code.

hyperspy/drawing/mpl_he.py
```diff
@@ -180,17 +180,82 @@ def plot(self, **kwargs):
...
+
+        navigator_kwds = kwargs.pop('navigator_kwds', {})
+
+        # widgets can be specified to contain the figure
+        navigator_widget = navigator_kwds.get('widget', None)
+        signal_widget = kwargs.get('widget', None)
+
+        backend = mpl.get_backend() # matplotlib plotting backend
+        if "ipympl" in backend:
+            with plt.ioff():
```
ericpre: Why is it necessary to use `plt.ioff()`? Add a comment?

hyperspy/drawing/figure.py
```diff
@@ -48,13 +48,16 @@ def create_figure(self, **kwargs):
 
         Parameters
         ----------
+        widget: ipywidgets output widget to place the figure in
         **kwargs
             All keyword arguments are passed to ``plt.figure``.
 
         """
+        widget = kwargs.pop('widget', None)
```
ericpre: Remove this line and use `create_figure(self, widget=None, **kwargs)` instead, so that the docstring is correct?

hyperspy/_signals/signal1d.py
```diff
@@ -1578,7 +1578,6 @@ def plot(self,
              autoscale='v',
              norm="auto",
              axes_manager=None,
-             navigator_kwds={},
```
ericpre: To be consistent, it should also be removed in `EDSSignal.plot`? Alternative, improve/bug fixes of setting/passing argument to plot functions could be improved in a separate PR, there is room for more!
#3147

hyperspy/drawing/mpl_he.py
```diff
@@ -180,17 +180,82 @@ def plot(self, **kwargs):
...
-            self.plot_navigator(**kwargs.pop('navigator_kwds', {}))
-            if pointer is not None:
-                self.navigator_plot.events.closed.connect(
-                    self.pointer.disconnect, [])
-        self.plot_signal(**kwargs)
+
+        navigator_kwds = kwargs.pop('navigator_kwds', {})
+
+        # widgets can be specified to contain the figure
+        navigator_widget = navigator_kwds.get('widget', None)
```
ericpre: This needs to be documented in the docstring.

hyperspy/drawing/mpl_he.py
```diff
@@ -180,17 +180,82 @@ def plot(self, **kwargs):
...
+        signal_widget = kwargs.get('widget', None)
+
+        backend = mpl.get_backend() # matplotlib plotting backend
+        if "ipympl" in backend:
+            with plt.ioff():
+                from ipywidgets.widgets import HBox, VBox, Output
+                plot_style = kwargs.pop("plot_style", None)
```
ericpre: This needs to be documented in the docstring.

hyperspy/drawing/mpl_he.py
```diff
@@ -180,17 +180,82 @@ def plot(self, **kwargs):
...
+                if plot_style is None:
+                    plot_style = preferences.Plot.widget_plot_style
+                # If widgets do not already exist, we will `display` them at the end
+                display_nav_widget_now = not navigator_widget
```
ericpre: Is there a reason to defer the creating of the widget? If not, I would be better to make the flow more linear and avoid redirection.

hyperspy/drawing/mpl_he.py
```diff
@@ -180,17 +180,82 @@ def plot(self, **kwargs):
...
+                    pointer = self.assign_pointer()
+                    if pointer is not None:
+                        self.pointer = pointer(self.axes_manager)
+                        self.pointer.color = 'red'
+                        self.pointer.connect_navigate()
+                    self.plot_navigator(**navigator_kwds)
+                    if pointer is not None:
+                        self.navigator_plot.events.closed.connect(
+                            self.pointer.disconnect, [])
+                self.plot_signal(**kwargs)
```
ericpre: I guess it should be possible to avoid this code duplication (here and the `else` block below)?

setup.py
```diff
@@ -80,6 +80,7 @@
         "pytest-instafail",
         # optional dependency of scipy to use scipy.datasets
         "pooch",
+        "ipympl",
```
ericpre: What's about putting it in `gui-jupyter` instead? Now `ipympl` is mature enough, works well with notebook and jupyterlab and will need to be used in the [notebook 7.0](https://jupyter-notebook.readthedocs.io/en/latest/migrate_to_notebook7.html) for which, the `nbaag` backend will not work anymore.

hyperspy/drawing/utils.py
```diff
@@ -30,9 +30,9 @@
 
 import matplotlib as mpl
 from mpl_toolkits.axes_grid1 import make_axes_locatable
-from matplotlib.backend_bases import key_press_handler
 from matplotlib.colors import LinearSegmentedColormap, BASE_COLORS, to_rgba
 import matplotlib.pyplot as plt
+import matplotlib.pyplot as plt
```
ericpre: Tidy up this import.

hyperspy/roi.py
```diff
@@ -409,7 +409,7 @@ def interactive(self, signal, navigation_signal="same", out=None,
         Returns
         -------
         :py:class:`~hyperspy.signal.BaseSignal` or one of its subclass
-            Signal updated with the current ROI selection 
+            Signal updated with the current ROI selection
```
ericpre: This is the kind of changes, which is not great for history but if this is pain to remove, we can keep it.

ericpre (commented):
Thank you @CSSFrancis for reviving this work. The approach is good but it needs tidying up!

hyperspy/tests/drawing/test_horizontal_plotting.py
CSSFrancis: Huh cool, didn't know about that.  That makes things a fair bit easier.

hyperspy/drawing/utils.py
CSSFrancis: Yea you were a little too quick for me :) I removed that now.

hyperspy/drawing/mpl_he.py
CSSFrancis: It is necessary based on https://matplotlib.org/ipympl/examples/full-example.html#interactions-with-other-widgets-and-layouting ~~but I don't seem to be getting an error if I remove it.  I'll remove it for now but will have to look into it further~~   I've cleaned this up a little bit

hyperspy/_signals/signal1d.py
CSSFrancis: There is! Right now the `plot` function with arguments defined in subclasses is [dangerous](https://fuhm.net/super-harmful/) This is part of the reason this PR first ran into problems from my understanding.

hyperspy/drawing/mpl_he.py
CSSFrancis: I'm not exactly sure what you are asking here?  Are you asking about setting the preferences and having the plot style? Or something related to calling creating the Output widget? or something else?

hyperspy/roi.py
CSSFrancis: Not sure if it was the best way to do it but I created a new commit undoing the change and then squashed the two commits... Seems like it worked.

CSSFrancis:
@ericpre Thanks for your comments!  This PR is starting to look a lot better, excited to start using it :)

CSSFrancis:
@ericpre it appears that when you plot using this method the button press events aren't going through.  You can still use your cursor to move the selector around but none of the key strokes are being captured.  Do you have any idea what that might be?

My best guess is that the plot isn't actually being selected  because it is wrapped in another widget which results in the button presses not being properly forwarded.

CSSFrancis:
Okay this is vastly over-complicated mostly because I was a little too stuck on the idea of using the `Output` class more realistically this could be accomplished by just calling `plt.ioff` and then taking the two canvases and putting them in a ipywidget. 

This also seems to fix the issue where the keys aren't registering.

CSSFrancis:
Honestly sorry this took so long to reach a much cleaner solution, but pretty happy with how things ended up. 

I think we start with this and come back to see if we need something like capturing the output.  That wouldn't be too hard, all we would have to do is return `s._plot.signal_plot.figure.canvas` and `s._plot.navigator_plot.figure.canvas`

jlaehne:
@ericpre is this PR good to merge? I'd like to use it at the sfmu in two weeks :-)

CSSFrancis:
@jlaehne It should be good to go now.  I added in some documentation/changelog which were removed when I simplified things quite a bit.

jlaehne:
Just checked it out locally. Works like a charm.

Only thing I noticed is that if I am too far zoomed in (to have large font sizes in a presentation), plots are starting to overlap and reach outside of the screen on the right (without a horizontal scroll bar) and then I don't have access to the resize feature in the lower right corner of the plot.

ericpre:
> t if I am too far zoomed in (to have large font sizes in a presentation), plots are starting to overlap and reach outside of the screen on the right (without a horizontal scroll bar) and then I don't have access to the resize feature in the lower right corner of the plot.

I wanted to have another look but I don't have time at the moment, and don't wait for my review - I wanted to revisit the initial approach where it is possible to pass widget as argument but this can be done in another PR.

CSSFrancis:
@jlaehne I can change things so the plots will stack rather than overlap if you make your window smaller. I didn't do that initially because I figured people might get frustrated if they have a small window and try to plot horizontally and get stacked plots.

Another thing to consider is that hyperspy can set a default dpi different to matplotlib if we want smaller figures. That might be a good preference to add. @hakonanes  has a good method for setting that value globally which works for presentations fairly well.

thomasaarholt:
I think I would recommend merging this now, and then considering changing that in a later PR. Just to get this in :)

jlaehne:
> @jlaehne I can change things so the plots will stack rather than overlap if you make your window smaller. I didn't do that initially because I figured people might get frustrated if they have a small window and try to plot horizontally and get stacked plots.

Just something I noticed, not fully decided which of the two is the best solution, but I guess we could still activate stacking in a separate PR or make it an option?

> Another thing to consider is that hyperspy can set a default dpi different to matplotlib if we want smaller figures. That might be a good preference to add. @hakonanes has a good method for setting that value globally which works for presentations fairly well.

That could indeed be an alternative to look in to optimize display for presentations. But in any case this horizontal plotting is really a big step forward for demos. I'll merge as soon as last test has rerun

doc/user_guide/visualisation.rst
```diff
@@ -283,6 +283,12 @@ The same example with the feature disabled:
 
 .. _plot.customize_navigator:
 
+.. versionadded:: 2.0.0
+   ``plot_style`` keyword argument to allow for "horizontal" or "vertical" plotting when using the `ipympl` or
+`widget` backends. A default value can also be set using
+:ref:`HyperSpy plot preferences <configuring-hyperspy-label>`.
```
jlaehne:
```suggestion
.. versionadded:: 2.0.0
   ``plot_style`` keyword argument to allow for "horizontal" or "vertical" plotting when using the
   `ipympl` or `widget` backends. A default value can also be set using
   :ref:`HyperSpy plot preferences <configuring-hyperspy-label>`.
```

doc/user_guide/visualisation.rst
```diff
@@ -283,6 +283,12 @@ The same example with the feature disabled:
 
 .. _plot.customize_navigator:
 
+.. versionadded:: 2.0.0
+   ``plot_style`` keyword argument to allow for "horizontal" or "vertical" plotting when using the
+   `ipympl` or `widget` backends. A default value can also be set using
+   :ref:`HyperSpy plot preferences <configuring-hyperspy-label>`.
```
jlaehne:
```suggestion
   ``plot_style`` keyword argument to allow for "horizontal" or "vertical" alignment of subplots (e.g. navigator 
   and signal) when using the `ipympl` or `widget` backends. A default value can also be set using the
   :ref:`HyperSpy plot preferences <configuring-hyperspy-label>`.
```

hakonanes:
> @hakonanes has a good method for setting that value globally which works for presentations fairly well.

Window size changes when we set the dots per inch globally, e.g. `plt.rcParams["figure.dpi"] = 300`. Increasing the DPI increases the window size, so we can fine-tune this for different screens. This is at least the behavior in my Ubuntu 22.04 with Matplotlib 3.6.

jlaehne:
@CSSFrancis, the horizontal plotting is not working any more with the newest bundle or a current conda environment using all the latest packages. Was there some upstream deprecation?

CSSFrancis:
@jlaehne thats quite weird. It should be tested. Is there any error?

CSSFrancis:
@jlaehne ipympl just returns a ipywidget. All we are doing is making an h box and then plotting them inside of it.

I actually think that the vertical plot is just a v box as well so if that works both should...

CSSFrancis:
@jlaehne can you maybe provide more context. 

1. Can you plot using the ipympl backend for example using `matplotlib`?
2. Can you print out all of the current versions?
3. Have you tried different web browsers? What web browser are you using/ is it updated?
4. What version of `hatch-nodejs-version` is being used?  I think they now bundle a version of node js using hatch but that was historically a __huge__ headache.

It's really hard for me to debug this as I can't seem to replicate it 😬 


I'll install the bundle on a new computer today and see if I can replicate it.

jlaehne:
I run Firefox 129.0.2 (but had the same issue on chromium 128.0.6613.113)
- Hyperspy 2.1.1
- matplotlib 3.9.2 
- jupyter_core 5.7.2
- jupyter_client 8.6.2
- jupyterlab_widgets 3.0.13
- ipympl 0.9.4
- nodejs 22.6.0

It used to work some while ago, but stopped working with some update. It simply plots the two plots above each as if the parameter was not set.

jlaehne:
Actually, I just downgraded to `ipympl=0.9.3` and it works. So the `ipympl=0.9.4` release broke this functionality.

Edit: It also affected most people at the eBEAM tutorials that had installed the bundle on various machines.

CSSFrancis:
@jlaehne That is extremely frustrating.  I'm sorry for that, ipympl should not be breaking things, expecially on a "bug patch" release....

Did the vertical plotting work using ipympl?

CSSFrancis:
@ericpre The major change is that they changed from a `setup.py` file to the `pyproject.toml`.   @jlaehne did downgrading change any other dependancies?

thomasaarholt:
Could it be that the [new version requires jupyter lab 4](https://github.com/matplotlib/ipympl/compare/0.9.3...0.9.4#diff-50c86b7ed8ac2cf95bd48334961bf0530cdc77b5a56f852c5c61b89d735fd711)? And that is what breaks things?

I'd try pip export (inspecting the installed packages) comparing two venvs with only that version difference. Currently on mobile.

ericpre:
@jlaehne, can you try #3437?

It may be related to a release of ipython, because previously the matplotlib backend name were encoded in ipython. See https://github.com/ipython/ipython/pull/14371. In the process, the name of the backend must have changed...

CSSFrancis:
@ericpre that would explain the wierd behavior :) Didn't think about that as a possibility.

PR #3142, opened 2023-05-04
Find peaks2d refactor

CSSFrancis:
### Description of the change
This changes a couple of things about find peaks2d

1. Interactive is False by default
2. Lazy signals return lazy peaks
3. `get_intensity` kwarg added to get the intensity for some peaks
4. Signal axes preserved in metadata. 

### Progress of the PR
- [x] add tests,
- [x] ready for review.

```python
import hyperspy.api as hs
import numpy as np
s = hs.signals.Signal2D(np.random.random((10,10,10,10)))
pks = s.find_peaks(threshold_abs=100, get_intensity=True)
```

### DIFF
--- a/hyperspy/_signals/signal2d.py
+++ b/hyperspy/_signals/signal2d.py
@@ -1039,7 +1039,8 @@ class Signal2D(BaseSignal, CommonSignal2D):
                              parallel=parallel, inplace=False, ragged=True,
                              max_workers=max_workers, **kwargs)
             peaks.metadata.add_node("Peaks") # add information about the signal Axes
-            peaks.metadata.Peaks.signal_axes = self.axes_manager.signal_axes
+        from copy import deepcopy
+        peaks.metadata.Peaks.signal_axes = tuple([deepcopy(s) for s in self.axes_manager.signal_axes])
         return peaks
 
     find_peaks.__doc__ %= (SHOW_PROGRESSBAR_ARG, PARALLEL_ARG, MAX_WORKERS_ARG,
--- a/hyperspy/_signals/signal2d.py
+++ b/hyperspy/_signals/signal2d.py
@@ -1030,10 +1030,6 @@ class Signal2D(BaseSignal, CommonSignal2D):
             peaks = self.map(method_func, show_progressbar=show_progressbar,
                              parallel=parallel, inplace=False, ragged=True,
                              max_workers=max_workers, **kwargs)
-            if peaks._lazy:
-                peaks.compute()
-
-
 
         return peaks
 
--- a/hyperspy/_signals/signal2d.py
+++ b/hyperspy/_signals/signal2d.py
@@ -969,9 +969,10 @@ class Signal2D(BaseSignal, CommonSignal2D):
             If True, the method parameter can be adjusted interactively.
             If False, the results will be returned.
         current_index : bool
-            if True, the computation will be performed for the current index.
+            If True, the computation will be performed for the current index.
         get_intensity : bool
-            if True, the value of the peak will be returned as a separate column.
+            If True, the intensity of the peak will be returned as an additional column,
+            the last one.
         %s
         %s
         %s
--- a/hyperspy/_signals/signal2d.py
+++ b/hyperspy/_signals/signal2d.py
@@ -22,6 +22,7 @@ import numpy.ma as ma
 import dask.array as da
 import logging
 import warnings
+from copy import deepcopy
 
 from functools import partial
 
@@ -1040,8 +1041,7 @@ class Signal2D(BaseSignal, CommonSignal2D):
                              parallel=parallel, inplace=False, ragged=True,
                              max_workers=max_workers, **kwargs)
             peaks.metadata.add_node("Peaks") # add information about the signal Axes
-        from copy import deepcopy
-        peaks.metadata.Peaks.signal_axes = tuple([deepcopy(s) for s in self.axes_manager.signal_axes])
+            peaks.metadata.Peaks.signal_axes = tuple([deepcopy(s) for s in self.axes_manager.signal_axes])
         return peaks
 
     find_peaks.__doc__ %= (SHOW_PROGRESSBAR_ARG, PARALLEL_ARG, MAX_WORKERS_ARG,
--- a/hyperspy/_signals/signal2d.py
+++ b/hyperspy/_signals/signal2d.py
@@ -23,6 +23,8 @@ import dask.array as da
 import logging
 import warnings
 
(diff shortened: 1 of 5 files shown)


CSSFrancis:
@ericpre This should be good to go. Let me know if you think that this is a reasonable change.

hyperspy/tests/signals/test_find_peaks2D.py
```diff
@@ -140,7 +147,13 @@ def test_find_peaks(self, method, dataset_name, parallel):
         else:
             shape = dataset.axes_manager.navigation_shape[::-1]
         assert peaks.data.shape == shape
-        assert peaks.data[0].shape[-1] == 2
+        first_ind = (0,)*len(shape)
+        if get_intensity:
+            peaks.data[first_ind]
+            assert peaks.data[first_ind].shape[-1] == 3
```
ericpre, line 153: Can you also check the size of the data? With `get_intensity=True`, they should be an additional column?

hyperspy/utils/peakfinders2D.py
```diff
@@ -27,6 +27,16 @@
 
 NO_PEAKS = np.array([[np.nan, np.nan]])
 
+def _get_intensity(X, f, **kwargs):
```
ericpre:
```suggestion
def _get_peak_position_and_intensity(X, f, **kwargs):
```
to improve readability, use a name that describe slightly better what it does.

Can you add a docstring?

hyperspy/_signals/signal2d.py
```diff
@@ -966,6 +970,8 @@ def find_peaks(self, method='local_max', interactive=True,
             If False, the results will be returned.
         current_index : bool
             if True, the computation will be performed for the current index.
+        get_intensity : bool
+            if True, the value of the peak will be returned as a separate column.
```
ericpre:
```suggestion
            if True, the intensity of the peak will be returned as an additional column, the last one.
```

hyperspy/_signals/signal2d.py
```diff
@@ -1030,11 +1038,8 @@ def find_peaks(self, method='local_max', interactive=True,
             peaks = self.map(method_func, show_progressbar=show_progressbar,
                              parallel=parallel, inplace=False, ragged=True,
                              max_workers=max_workers, **kwargs)
-            if peaks._lazy:
-                peaks.compute()
-
-
-
+            peaks.metadata.add_node("Peaks") # add information about the signal Axes
+            peaks.metadata.Peaks.signal_axes = self.axes_manager.signal_axes
```
ericpre: Can you please add documentation and tests?

hyperspy/utils/peakfinders2D.py
```diff
@@ -27,6 +27,16 @@
...
+    peaks = f(X, **kwargs)
+
+    if np.any(np.isnan(peaks)): #handle no peaks
+        return np.array([[np.nan, np.nan,np.nan]])
+    else:
+        maxima = tuple([tuple(np.round(peaks[:, d]).astype(int)) for d in range(2)])
```
ericpre: Is it not possible to use `np.round` on the whole array?

ericpre (commented):
For the change of behaviour with the `interactive` argument, it would be good to keep consistent with other methods, such as `remove_backgroud`, `spikes_removal_tool`, `calibrate`, the smoothing method. 
Usually, the default is to be interactive when no parameter is given, but it may be difficult here to use this approach considering that various peak finding method can be found. `spikes_removal_tool` and `calibrate` has `interactive=True` by default. To avoid introducing a third default, we would use one of the two current default.

Can you also add a change log entry for new `get_intensity` argument?

hyperspy/tests/signals/test_find_peaks2D.py
CSSFrancis: That is what this is doing,  It is checking to see that there are 3 columns instead of two. Or do you mean something else

CSSFrancis:
@ericpre I changed the default back to interactive.  That might be something that we should make a larger discussion for.

My opinion is that the default shouldn't depend on an optional dependency but could be convinced otherwise.

hyperspy/utils/peakfinders2D.py
```diff
@@ -27,6 +27,37 @@
...
+
+    """
+    peaks = f(X, **kwargs)
+
+    if np.any(np.isnan(peaks)): #handle no peaks
+        return np.array([[np.nan, np.nan,np.nan]])
+    else:
+        rounded_peaks = np.round(peaks).astype(int)
+        maxima = tuple([tuple(rounded_peaks[:,d]) for d in range(2)])
+        intensity = X[maxima]
```
ericpre: I find easier the following easier to follow/read rather using list comprehension and creating tuple of tuples in combination with numpy advanced indexing:
```suggestion
        peaks_indices = np.round(peaks).astype(int)
        intensity = X[peaks_indices[:, 0], peaks_indices[:, 1]]

```

hyperspy/tests/signals/test_find_peaks2D.py
ericpre: Sorry, this is what I meant but misread the test and therefore not what I intended to mean! 🤦‍♂️ (I should have finished my morning coffee before finishing this review! 😅 
Anyway, could we check some of the actual value (position, intensity) in addition to the shape? This would be useful to check that there is no issue with ordering, indexing, etc.

hyperspy/_signals/signal2d.py
```diff
@@ -1030,11 +1040,8 @@ def find_peaks(self, method='local_max', interactive=True,
             peaks = self.map(method_func, show_progressbar=show_progressbar,
                              parallel=parallel, inplace=False, ragged=True,
                              max_workers=max_workers, **kwargs)
-            if peaks._lazy:
-                peaks.compute()
-
-
-
+            peaks.metadata.add_node("Peaks") # add information about the signal Axes
+            peaks.metadata.Peaks.signal_axes = tuple([deepcopy(s) for s in self.axes_manager.signal_axes])
```
ericpre: Simplify to:
```suggestion
            peaks.metadata.Peaks.signal_axes = deepcopy(self.axes_manager.signal_axes)
```

ericpre (commented):
See some comments in review.

I pushed a commit to edit the API changes and change log entry.

hyperspy/tests/signals/test_find_peaks2D.py
CSSFrancis: All good :) Honestly some of these changes I probably should have caught myself

PR #3150, opened 2023-05-12
print out line name in missing line warning

w-markus:
### Description of the change
changed the warning message about a missing xray-line in an EDS model to include the name of the line

### Progress of the PR
- [x] update docstring (not necessary, IMHO),
- [x] update user guide (not necessary, IMHO),

### DIFF
--- a/hyperspy/model.py
+++ b/hyperspy/model.py
@@ -57,11 +57,11 @@ from hyperspy.misc.export_dictionary import (
     load_from_dictionary,
     parse_flag_string,
     reconstruct_object
-    )
+)
 from hyperspy.misc.model_tools import (
     current_model_values,
     _calculate_covariance,
-    )
+)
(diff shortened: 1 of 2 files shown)


ericpre:
Thank you for the pull request. It may be more better to rebase on the `RELEASE_next_major` branch as we are currently working toward releasing 2.0.
There are two test failures, which seems to be related to the changes of this PR.
One details on the code changes: it is usually good practice to avoid mixing style changes with actual code changes as it makes the git history more difficult to read - quite often it useful to look at the history to figure regression and understand reasoning of code changes.

ericpre:
@w-markus, now that this code has moved in [exspy](https://github.com/hyperspy/exspy), this PR will need to move there too. I have opened this [issue](https://github.com/hyperspy/exspy/issues/8) so that it doesn't get lost!

Issue #3107, opened 2023-03-09
Lazy Optimization/ Guidelines and Development Guide


CSSFrancis:
I wanted to start this thread because I think there is quite a bit of misunderstanding on how hyperspy handles `lazy` operations and how to optimize performance.  Sometimes I think that they way that I use hyperspy is a little different than the average user and even among some of the developers. 

To start with most of my thoughts on this subject is heavily influenced by Matthew Rocklin who started dask.  You can read it straight from the horses mouth [here](https://www.coiled.io/blog/prioritizing-pragmatic-performance-for-dask). 

## Rules for Developing

1. Rechunk in a function as little as possible!

This is something that `pyxem` got in huge trouble with in the past.  Rechunking is costly and when you have a signal that is larger than the available RAM it will cause big problems.  When I first started using pyxem it worked great until I hit larger and larger datasets and then everything started to slow down and break. `Rechunk=True` is still a default in many hyperspy functions and I think this should probably not be the case (see 2 below)

2. [Slow is Smooth and Smooth is Fast](https://www.coiled.io/blog/reducing-dask-memory-usage):

This is the new mantra of `dask` and I love it.  `Dask` is not the fastest scheduler but it ___is___ the most reliable scheduler.  In general we want functions that always work and work out of the box.  We don't want things failing (see above) and speed is secondary.  A computation that does run is infinity-X times faster than one that doesn't run.  The reality is that hyperspy/pyxem is running sub-optimally on most peoples computers.  On a modest cluster `hyperspy` can run on 100's of cores ___streaming data___.  [This](https://www.coiled.io/blog/prioritizing-pragmatic-performance-for-dask) article by the founder of `dask` explains our situation pretty perfectly.  (This is also your reminder to update `dask` to the newest version)

3. Put the `dask` optimization onus on the user:

Let the user configure their `dask` computing the way they want.  This means no default schedulers, no setting the number of workers in a function, and calling `compute` as little as possible inside of a function. In some cases it is preferable to call `persist` rather than `compute`.  Some processes work better using threads vs processes.  Some people have special hardware that works better for certain computations etc. Let that be something that users can 

4. Promote lazy literacy:

This goes with 3 but I think that `dask` and by extension `hyperspy` has a "it is too easy problem".  You can use hyperspy for years and not understand `dask`.  You can use lazy signals and make super convoluted task graphs but never take the time to understand them.  That becomes a problem when now you have a 1 Tb dataset and you have to be much more conscious of everything.

[...]

PR #3166, opened 2023-05-30
Change rechunk function parameter from default True, to default False

magnunor:
This pull request changes the default value for the `rechunk` parameter in all(?) functions from `True` to `False`. This to improve processing performance for large lazy datasets, as changing the chunking "structure" of large datasets can really be detrimental to performance.

For more information about this, see https://github.com/hyperspy/hyperspy/issues/3107

### Description of the change

- For all function parameters: change `rechunk=True`, to `rechunk=False`

### Progress of the PR
- [x] update docstring,
- [x] update user guide,
- [x] add an changelog entry in the `upcoming_changes` folder
- [x] add tests,
- [x] ready for review.

Make some test data
```python
import dask.array as da
import hyperspy.api as hs

data = da.arange(256*256*256*256).reshape(256, 256, 256, 256)
s.save("testdata.zspy", chunks=(32, 32, 32 ,32))
```
Then run this for processing:
```python
s = hs.load("testdata.zspy", lazy=True)
s_sum_rechunk = s.sum(rechunk=True)
s_sum_not_rechunk = s.sum(rechunk=False)

s_sum_rechunk.compute() # 12.41 s
s_sum_not_rechunk.compute() # 3.7 s
```
I'm not really sure if this is the best example, as it might be a bit hardware dependent.

-----

There is one function which automatically does rechunking internally in the function, [integrate_simpson](https://github.com/hyperspy/hyperspy/blob/RELEASE_next_major/hyperspy/_signals/lazy.py#L706). This one should probably be changed as well, but since it didn't involve the `rechunk` parameter in its function call, I didn't change it in this pull request.

----------

@CSSFrancis

### DIFF
--- a/doc/user_guide/big_data.rst
+++ b/doc/user_guide/big_data.rst
@@ -374,12 +374,12 @@ method can be used:
     >>> s.get_chunk_size(0) # The first navigation axis
     ((200,),)
 
-.. versionadded:: 1.3.2
+.. versionadded:: 2.0.0
 
-By default, HyperSpy tries to optimize the chunking for most operations. However,
-it is sometimes possible to manually set a more optimal chunking manually. Therefore,
-many operations take a ``rechunk`` or ``optimize`` keyword argument to disable
-automatic rechunking.
+Starting in version 2.0.0 HyperSpy does not automatically rechunk datasets as
+this can lead to reduced performance. The ``rechunk`` or ``optimize`` keyword argument
+can be set to ``True`` to let HyperSpy automatically change the chunking which
+could potentially speed up operations.
 
 .. versionadded:: 1.7.0
 
--- a/hyperspy/_signals/eds.py
+++ b/hyperspy/_signals/eds.py
@@ -167,10 +167,10 @@ class EDSSpectrum(Signal1D):
                 xray_lines_not_in_range.append(xray_line)
         return xray_lines_in_range, xray_lines_not_in_range
 
-    def sum(self, axis=None, out=None):
+    def sum(self, axis=None, out=None, rechunk=False):
         if axis is None:
             axis = self.axes_manager.navigation_axes
-        s = super().sum(axis=axis, out=out)
+        s = super().sum(axis=axis, out=out, rechunk=rechunk)
         s = out or s
 
         # Update live time by the change in navigation axes dimensions
--- a/hyperspy/_signals/lazy.py
+++ b/hyperspy/_signals/lazy.py
@@ -703,11 +703,11 @@ class LazySignal(BaseSignal):
 
     diff.__doc__ = BaseSignal.diff.__doc__
 
-    def integrate_simpson(self, axis, out=None):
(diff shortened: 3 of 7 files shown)


CSSFrancis:
@magnunor This is great!  `Dask` is getting smarter and smarter but it hasn't quite figured out how to minimize the transfer between different workers and still makes some less than intelligent rechunking in certain cases.  

It would be a good idea to keep an eye out because sometimes `dask` will drop something like [2022.11.0](https://www.coiled.io/blog/reducing-dask-memory-usage) which is truly wonderful.

doc/user_guide/big_data.rst
```diff
@@ -376,10 +376,10 @@ method can be used:
...
 .. versionadded:: 1.3.2
 
-By default, HyperSpy tries to optimize the chunking for most operations. However,
-it is sometimes possible to manually set a more optimal chunking manually. Therefore,
-many operations take a ``rechunk`` or ``optimize`` keyword argument to disable
-automatic rechunking.
+By default, HyperSpy does not change the chunking, as this can lead to reduced
+performance for large datasets. However, many operations take a ``rechunk`` or
+``optimize`` keyword argument which can be set to ``True`` to let HyperSpy
+change the chunking.
```
CSSFrancis: I think that this is confusing because the version added is `.. versionadded:: 1.3.2`

Maybe add in a new block 

```suggestion
.. versionadded:: 2.0.0

Starting in version 2.0.0 HyperSpy does not automatically rechunk datasets as this leads for reduced performance.  The ``rechunk`` or ``optimize`` keyword argument can be set to ``True`` to change the chunking which could potentially speed up operations.
```

CSSFrancis (commented):
This looks good to me.  One point is that in the guide it might be nice to add somewhere that the default changed with the 2.0.0 version.  Otherwise this is good to merge.

magnunor:
>  I think that this is confusing because the version added is .. versionadded:: 1.3.2
>
> Maybe add in a new block

Done!

I'll be away the whole of July, so if this PR is blocking the 2.0.0 release, feel free to improve and/or merge it. Maintainers should have access to change it.

doc/user_guide/big_data.rst
ericpre: Would it make sense to keep both?

ericpre:
> There is one function which automatically does rechunking internally in the function, [integrate_simpson](https://github.com/hyperspy/hyperspy/blob/RELEASE_next_major/hyperspy/_signals/lazy.py#L706). This one should probably be changed as well, but since it didn't involve the rechunk parameter in its function call, I didn't change it in this pull request.

I added the `rechunk` argument to `integrate_simpson` to enforce consistency.

ericpre (approved):

CSSFrancis (approved):
This looks good to me if we want to merge the changes.  It might be work rebasing the the branch before merging.

PR #3173, opened 2023-06-19
Pretty logging

CSSFrancis:
### Description of the change
This changes the default behavior of the logging module to print to the stdout rather than the stderror.  This gives much more flexibility and (more importantly) allows us to set to color of the outputs so that they __no longer__ look like errors in python. 

### Progress of the PR
- [ ] add tests,
- [ ] ready for review.

Screenshot from 2023-06-19 14-53-48

Screenshot from 2023-06-19 14-55-09

Similarly things like WARNINGS are now printed in yellow:

Screenshot from 2023-06-19 15-05-54

### DIFF
--- a/hyperspy/logger.py
+++ b/hyperspy/logger.py
@@ -70,17 +70,26 @@ def set_log_level(level):
     logger.setLevel(level)
 
 class ColoredFormatter(logging.Formatter):
-    """Colored log formatter."""
+    """Colored log formatter.
+
+    This class is used to format the log output. The colors can be changed
+    by changing the ANSI escape codes in the class variables.
+
+    This is a modified version of both this
+    https://github.com/herzog0/best_python_logger
+    and this https://stackoverflow.com/questions/384076/how-can-i-color-python-logging-output/56944256#56944256
+    """
     grey = "\x1b[38;20m"
     yellow = "\x1b[33;20m"
     red = "\x1b[31;20m"
     bold_red = "\x1b[31;1m"
     reset = "\x1b[0m"
+    green = "\x1b[1;32m"
     format = "%(levelname)s | Hyperspy | %(message)s (%(name)s:%(lineno)d)"
 
     FORMATS = {
         logging.DEBUG: grey + format + reset,
-        logging.INFO: grey + format + reset,
+        logging.INFO: green + format + reset,
         logging.WARNING: yellow + format + reset,
         logging.ERROR: red + format + reset,
         logging.CRITICAL: bold_red + format + reset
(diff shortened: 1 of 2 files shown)


CSSFrancis:
@ericpre I think you already cleaned up quite a bit of the common warnings in `hyperspy` so it is less likely that users will see them often.  

This is just another way to handle the logging to get rid of the annoying red output.  I am not sure how this might work on a windows computer.... I have one I can test on tomorrow.

hyperspy/logger.py
```diff
@@ -62,6 +66,44 @@ def set_log_level(level):
...
+
+    The logging output will also be redirected from the standard error file to
+    the standard output file.
+    """
+    formatter = ColoredFormatter()
+    handler = logging.StreamHandler(sys.stdout)
+    handler.setFormatter(formatter)
+
+    _logger = logging.getLogger(*args)
+    _logger.handlers[:] = []
```
ericpre:
```suggestion
    # Remove existing handler
    while len(_logger.handlers):
        _logger.removeHandler(_logger.handlers[0])
```

hyperspy/logger.py
```diff
@@ -62,6 +66,44 @@ def set_log_level(level):
...
+    grey = "\x1b[38;20m"
+    yellow = "\x1b[33;20m"
+    red = "\x1b[31;20m"
+    bold_red = "\x1b[31;1m"
+    reset = "\x1b[0m"
+    format = "%(levelname)s | Hyperspy | %(message)s (%(name)s:%(lineno)d)"
+
+    FORMATS = {
+        logging.DEBUG: grey + format + reset,
+        logging.INFO: grey + format + reset,
```
ericpre: What's about using green?

hyperspy/logger.py
```diff
@@ -62,6 +66,44 @@ def set_log_level(level):
...
     """
-    import logging
-    logging.basicConfig()  # Does nothing if already configured
-    logging.getLogger('hyperspy').setLevel(level)
+    logger = initialize_logger('hyperspy')
+    logger.setLevel(level)
+
+class ColoredFormatter(logging.Formatter):
+    """Colored log formatter."""
+    grey = "\x1b[38;20m"
```
ericpre, line 82: For future reference, where does these color code come from?

ericpre (approved):
This is indeed much better. In the qtconsole, it still display in red but we could leave it for now.

Do you want to add a changelog entry?

ericpre:
@CSSFrancis, this one should be easy to finish, can you please answer the comments so that could we get it out of the way? :)

hyperspy/logger.py
CSSFrancis: The color codes are taken from here. 
https://stackoverflow.com/questions/384076/how-can-i-color-python-logging-output

hyperspy/logger.py
CSSFrancis: Done!

PR #3174, opened 2023-06-19
Add deprecation decorators

CSSFrancis:
### Description of the change
This adds 2 helper function into hyperspy to improve the deprecation handling and have it more inline with downstream packages (kikchipy/ pyxem).

A decorator can be added to a function to track if it is deprecated or if an argument is deprecated with a potential replacement argument.  If there is a replacement argument that value will be passed in as the replacement argument and (for better or worse) the function will not fail.

I found that this streamlines the deprecation process and helps to improve consistency. 

### Progress of the PR
- [x] add tests,
- [x] ready for review.

```python
@deprecated(since=2.0, alternative="new_function", removal=2.1)
def deprecated_function():
    pass

@deprecated_argument(name="a", since="2.0", removal="2.1", alternative="b")
def function(self, b):
    return b

```

This is mostly copied from @hakonanes and kikchipy with some __slight__ changes

### DIFF
--- a/doc/dev_guide/coding_style.rst
+++ b/doc/dev_guide/coding_style.rst
@@ -62,4 +62,4 @@ def this_property(y):
 This will update the docstring, and print a visible deprecation warning telling the user to user the
 alternative function or argument.
 
-These deprecation wrappers are copied from work by Håkon Wiik Ånes and kikchipy.
+These deprecation wrappers inspired by those in `kikuchipy`
--- a/doc/dev_guide/coding_style.rst
+++ b/doc/dev_guide/coding_style.rst
@@ -38,7 +38,7 @@ adding a ``post-commit`` file to ``.git/hook/`` with the following content:
 
 Deprecations
 ============
-Hyperspy follows `semetic versioning <https://semver.org>`_ where changes follow such that:
+HyperSpy follows `semantic versioning <https://semver.org>`_ where changes follow such that:
 
 1. MAJOR version when you make incompatible API changes
 2. MINOR version when you add functionality in a backward compatible manner
--- a/doc/dev_guide/coding_style.rst
+++ b/doc/dev_guide/coding_style.rst
@@ -49,7 +49,7 @@ Deprecation warnings are raised whenever possible and feasible for functions/met
 so that users get a heads-up one (minor) release before something is removed or changes, with a possible
 alternative to be used.
 
-The decorator should be placed right above the object signature to be deprecated:
+A deprecation decorator should be placed right above the object signature to be deprecated:
 
 @deprecated(since=1.7.0, removal=2.0.0, alternative="bar")
 def foo(self, n):
--- a/doc/dev_guide/coding_style.rst
+++ b/doc/dev_guide/coding_style.rst
@@ -35,3 +35,31 @@ adding a ``post-commit`` file to ``.git/hook/`` with the following content:
     GIT_COMMITTER_NAME="black" GIT_COMMITTER_EMAIL="black@email.com" git
     commit --author="black <black@email.com>" -m "Automatic style
     corrections courtesy of black"
+
+Deprecations
+============
+Hyperspy follows `semetic versioning <https://semver.org>`_ where changes follow such that:
+
+1. MAJOR version when you make incompatible API changes
+2. MINOR version when you add functionality in a backward compatible manner
+3. PATCH version when you make backward compatible bug fixes
+
+This means that as little, ideally no, functionality should break between minor releases.
+Deprecation warnings are raised whenever possible and feasible for functions/methods/properties/arguments,
+so that users get a heads-up one (minor) release before something is removed or changes, with a possible
+alternative to be used.
+
+The decorator should be placed right above the object signature to be deprecated:
+
+@deprecated(since=1.7.0, removal=2.0.0, alternative="bar")
+def foo(self, n):
+    return n + 1
+
+@deprecated_argument(since=1.7.0, removal=2.0.0,name="x", alternative="y")
+def this_property(y):
+    return y
+
+This will update the docstring, and print a visible deprecation warning telling the user to user the
+alternative function or argument.
+
+These deprecation wrappers are copied from work by Håkon Wiik Ånes and kikchipy.
--- a/hyperspy/decorators.py
+++ b/hyperspy/decorators.py
@@ -115,7 +115,8 @@ def interactive_range_selector(cm):
 class deprecated:
     """Decorator to mark deprecated functions with an informative
     warning.
-    Adapted from
+
+    Inspired by
     `scikit-image
     <https://github.com/scikit-image/scikit-image/blob/main/skimage/_shared/utils.py>`_
     and `matplotlib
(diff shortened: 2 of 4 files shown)


ericpre:
@CSSFrancis, is the intended use internal only or is it something that would be expected to be used by other libraries? If the later, I am wondering, if this is something that would be suitable for https://scientific-python.org? If the former, it would only need documentation in the developer guide.

CSSFrancis:
I think internal use only. My opinion is that we want to be able to change and adjust it as we go. 

It might potentially be good to have a standard in the entire scipy ecosystem for deprecating code but everyone copying code isn't a bad thing as it does give a little more freedom.

I'll update the developer guide.

hyperspy/decorators.py
```diff
@@ -106,3 +110,123 @@ def wrapper(self, *args, **kwargs):
...
+
+
+class deprecated:
+    """Decorator to mark deprecated functions with an informative
+    warning.
+    Adapted from
+    `scikit-image
+    <https://github.com/scikit-image/scikit-image/blob/main/skimage/_shared/utils.py>`_
+    and `matplotlib
+    <https://github.com/matplotlib/matplotlib/blob/master/lib/matplotlib/_api/deprecation.py>`_.
```
ericpre, line 123: When copying code from somewhere else, we should make sure to comply with the license of the original code...

hakonanes: Do you mean the license of scikit-image and Matplotlib, or kikuchipy? I wrote this part in kikuchipy, and I give @CSSFrancis permission to include this piece of code and permission to any contributor to HyperSpy to modify it without including the license of kikuchipy in HyperSpy.

The [deprecation tools in kikuchipy](https://github.com/pyxem/kikuchipy/blob/develop/kikuchipy/_util/_deprecated.py) were not copied from [scikit-image](https://github.com/scikit-image/scikit-image/blob/e37e48d5dc137fceb548d28efdbe2d9a4d4dd316/skimage/_shared/utils.py#L357) or [Matplotlib](https://github.com/matplotlib/matplotlib/blob/main/lib/matplotlib/_api/deprecation.py), rather, I looked to their decorators for inspiration. I realize that the word "adapted" is less descriptive and something like "inspired" is better.

CSSFrancis: I think that we should be good then?  I also think that the scikit-image and Matplotlib licenses are very permissive so we shouldn't have any issues there...

doc/dev_guide/coding_style.rst
```diff
@@ -35,3 +35,31 @@ adding a ``post-commit`` file to ``.git/hook/`` with the following content:
     GIT_COMMITTER_NAME="black" GIT_COMMITTER_EMAIL="black@email.com" git
     commit --author="black <black@email.com>" -m "Automatic style
     corrections courtesy of black"
+
+Deprecations
+============
+Hyperspy follows `semetic versioning <https://semver.org>`_ where changes follow such that:
```
hakonanes:
```suggestion
HyperSpy follows `semantic versioning <https://semver.org>`_ where changes follow such that:
```

doc/dev_guide/coding_style.rst
```diff
@@ -35,3 +35,31 @@ adding a ``post-commit`` file to ``.git/hook/`` with the following content:
...
+1. MAJOR version when you make incompatible API changes
+2. MINOR version when you add functionality in a backward compatible manner
+3. PATCH version when you make backward compatible bug fixes
+
+This means that as little, ideally no, functionality should break between minor releases.
+Deprecation warnings are raised whenever possible and feasible for functions/methods/properties/arguments,
+so that users get a heads-up one (minor) release before something is removed or changes, with a possible
+alternative to be used.
+
+The decorator should be placed right above the object signature to be deprecated:
```
hakonanes:
```suggestion
A deprecation decorator should be placed right above the object signature to be deprecated:
```

doc/dev_guide/coding_style.rst
```diff
@@ -35,3 +35,31 @@ adding a ``post-commit`` file to ``.git/hook/`` with the following content:
...
+    return n + 1
+
+@deprecated_argument(since=1.7.0, removal=2.0.0,name="x", alternative="y")
+def this_property(y):
+    return y
+
+This will update the docstring, and print a visible deprecation warning telling the user to user the
+alternative function or argument.
+
+These deprecation wrappers are copied from work by Håkon Wiik Ånes and kikchipy.
```
hakonanes: While this note is nice, I suggest to rephrase to something like "The deprecation wrappers are inspired by those in kikuchipy". Or, I'm OK with also just removing it: so much of kikuchipy is inspired by how things are handled in HyperSpy, without exhaustive references back to HyperSpy (although there are a few), so I'm just happy that some functionality "flows" the other way.

hyperspy/decorators.py
```diff
@@ -106,3 +110,123 @@ def wrapper(self, *args, **kwargs):
         else:
             cm(self, *args, **kwargs)
     return wrapper
+
+
+class deprecated:
+    """Decorator to mark deprecated functions with an informative
+    warning.
+    Adapted from
```
hakonanes: Blank line between first sentence and remaining description.

hyperspy/decorators.py
```diff
@@ -106,3 +110,123 @@ def wrapper(self, *args, **kwargs):
...
+
+    def __init__(
+        self,
+        since: Union[str, int, float],
+        alternative: Optional[str] = None,
+        alternative_is_function: bool = True,
+        removal: Union[str, int, float, None] = None,
+    ):
+        """Visible deprecation warning.
+        Parameters
```
hakonanes, line 135: Blank line above parameter list

hyperspy/decorators.py
```diff
@@ -106,3 +110,123 @@ def wrapper(self, *args, **kwargs):
...
+        )
+        wrapped.__doc__ = new_doc
+
+        return wrapped
+
+
+class deprecated_argument:
+    """Decorator to remove an argument from a function or method's
+    signature.
+    Adapted from `scikit-image
```
hakonanes, line 200: Blank line above here.

hyperspy/tests/test_decorators.py
```diff
@@ -0,0 +1,134 @@
...
+
+        my_foo = Foo()
+
+        # Does not warn
+        with warnings.catch_warnings():
+            warnings.simplefilter("error")
+            assert my_foo.bar_arg(b=1) == {"b": 1}
+
+        # Warns
+        with pytest.warns(np.VisibleDeprecationWarning) as record2:
```
hakonanes: There's no `record1`, I suggest to rename to `record1` and `record2`.

hakonanes (commented):
This looks clean to me. I've pointed out a few typos and some style inconsistencies.

doc/dev_guide/coding_style.rst
CSSFrancis: Sounds good :) I'll just leave that they came from kikuchipy.

I do think they are quite a good way to handle deprecations and it might be good to promote them for wider usage.  If someone takes them from hyperspy it would be good to have the credit make it's way back to you!

hakonanes: Thanks, that is more than enough (:

hakonanes (approved):

PR #3148, opened 2023-05-08
Marker Collection

CSSFrancis:
### Description of the change
Rather than change how markers are created I decided to create a bit of a higher level class. 

This does duplicate some functionality but in this case we are copying the upstream behavior with  `matplotlib` so I feel that it should be acceptable. 

The idea here is to speed up plotting of many markers using [collections](https://matplotlib.org/stable/api/collections_api.html)
Rather than implement every collection individually I opted to pass the keyword arguments to the class. 

### Progress of the PR
- [x] Generic `MarkerCollection` class created
- [x] `MarkerCollection` Class plotting
- [x] Add `from_signal` method
- [x] Add methods for Converting MarkerDict --> Collection
   - [x] Support for Ellipse
   - [x] Support for Horizontal Line
   - [x] Support for Point
   - [x] Support for Line Segment
   - [x] Support for Vertical Line
   - [x] Support for Rectangle
- [x] Remove Old Marker Classes
- [x] Refactor TextMarker Class more inline with `MarkerCollection`
- [x] Add lazy plotting & caching
- [x] Add rotation to `Texts` marker
- [x] Add example to show how to interact with marker, for example, remove a marker.
- [x] update docstring (if appropriate),
- [x] update user guide (if appropriate),
- [x] add an changelog entry in the `upcoming_changes` folder (see [`upcoming_changes/README.rst`](https://github.com/hyperspy/hyperspy/blob/RELEASE_next_minor/upcoming_changes/README.rst)),
- [x] Check formatting changelog entry in the `readthedocs` doc build of this PR (link in github checks)
- [x] add tests,
- [x] ready for review.

### Minimal example of the bug fix or the new feature
```python
from matplotlib.collections import LineCollection, CircleCollection

data = np.empty((3, 4), dtype=object)
for i in np.ndindex(data.shape):
    data[i] = np.random.random((10, 2, 2))*3
s = Signal2D(np.random.random((3, 4, 5, 6)))
markers = MarkerCollection(LineCollection, segments=data, lw=2)
```
test

```python
s = Signal2D(np.random.random((3, 4, 5, 6)))
markers = MarkerCollection(CircleCollection,offsets=np.random.rand(10, 2)*5,sizes=(50,),linewidths=(2,),facecolors="black",)
s.axes_manager.signal_axes[0].scale=1
s.axes_manager.signal_axes[1].scale=1
s.plot()
s.add_marker(markers, plot_on_signal=True)
```
test

PR #3175, opened 2023-06-19
NewFeature: Added jump click to plotting

CSSFrancis:
### Description of the change
This is just a quick change to the plotting to allow for jumping.   I think we might want to make it possible to change which key is used with the jump.

I can make this a little more robust and we could potentially add it to all of the widgets but the main problem is if you have multiple widgets I don't know how it should work. Same with handing two singles overlaid in 1D.

### Progress of the PR
- [x] add tests,
- [x] ready for review.

```python
%matplotlib ipympl
import hyperspy.api as hs
import dask.array as da
s = hs.signals.Signal2D(np.random.random((100,100,100,100))).as_lazy()
s.plot()
```

https://github.com/hyperspy/hyperspy/assets/41125831/3e9918e6-b63e-46da-8775-4b1288ac3889

### DIFF
--- a/hyperspy/drawing/_widgets/rectangles.py
+++ b/hyperspy/drawing/_widgets/rectangles.py
@@ -40,6 +40,11 @@ class SquareWidget(Widget2DBase):
     def __init__(self, axes_manager, **kwargs):
         super(SquareWidget, self).__init__(axes_manager, **kwargs)
 
+    def connect(self, ax):
+        super(SquareWidget, self).connect(ax)
+        canvas = ax.figure.canvas
+        canvas.mpl_connect('button_press_event', self._onjumpclick)
+
     def _set_patch(self):
         """Sets the patch to a matplotlib Rectangle with the correct geometry.
         The geometry is defined by _get_patch_xy, and get_size_in_axes.
@@ -54,6 +59,11 @@ class SquareWidget(Widget2DBase):
             alpha=self.alpha,
             picker=True,)]
         super(SquareWidget, self)._set_patch()
+
+    def _onjumpclick(self, event):
+        # Callback for MPL pick event
+        if event.key=="shift":
+            self.position = (event.xdata, event.ydata)
 
     def _onmousemove(self, event):
         """on mouse motion move the patch if picked"""
--- a/hyperspy/drawing/_widgets/rectangles.py
+++ b/hyperspy/drawing/_widgets/rectangles.py
@@ -70,12 +70,10 @@ class SquareWidget(Widget2DBase):
         relevant events, and updates the patch position.
         """
         if self._navigating:
-            with (self.axes_manager.events.indices_changed.suppress_callback(
-                  self._on_navigate) or
-                  self.axes_manager.events.indices_changed.suppress_callback(
-                  self._onjumpclick)):
+            with self.axes_manager.events.indices_changed.suppress():
                 for i in range(len(self.axes)):
                     self.axes[i].value = self._pos[i]
+            self.axes_manager.events.indices_changed.trigger(obj=self.axes_manager)
         self.events.moved.trigger(self)
         self.events.changed.trigger(self)
         self._update_patch_position()
--- a/hyperspy/drawing/_widgets/rectangles.py
+++ b/hyperspy/drawing/_widgets/rectangles.py
@@ -40,10 +40,6 @@ class SquareWidget(Widget2DBase):
     def __init__(self, axes_manager, **kwargs):
         super(SquareWidget, self).__init__(axes_manager, **kwargs)
 
-    def connect(self, ax):
-        super(SquareWidget, self).connect(ax)
-        canvas = ax.figure.canvas
-
     def _set_patch(self):
         """Sets the patch to a matplotlib Rectangle with the correct geometry.
         The geometry is defined by _get_patch_xy, and get_size_in_axes.
--- a/hyperspy/drawing/_widgets/rectangles.py
+++ b/hyperspy/drawing/_widgets/rectangles.py
@@ -65,6 +65,18 @@ class SquareWidget(Widget2DBase):
         if event.key=="shift" and event.inaxes:
             self.position = (event.xdata, event.ydata)
 
+    def _pos_changed(self):
+        """Call when the position of the widget has changed. It triggers the
+        relevant events, and updates the patch position.
+        """
+        if self._navigating:
+            with self.axes_manager.events.indices_changed.suppress():
+                for i in range(len(self.axes)):
+                    self.axes[i].value = self._pos[i]
+        self.events.moved.trigger(self)
+        self.events.changed.trigger(self)
+        self._update_patch_position()
+
     def _onmousemove(self, event):
         """on mouse motion move the patch if picked"""
         if self.picked is True and event.inaxes:
--- a/hyperspy/drawing/_widgets/rectangles.py
+++ b/hyperspy/drawing/_widgets/rectangles.py
@@ -43,7 +43,6 @@ class SquareWidget(Widget2DBase):
     def connect(self, ax):
         super(SquareWidget, self).connect(ax)
         canvas = ax.figure.canvas
-        canvas.mpl_connect('button_press_event', self._onjumpclick)
 
     def _set_patch(self):
         """Sets the patch to a matplotlib Rectangle with the correct geometry.
@@ -61,22 +60,8 @@ class SquareWidget(Widget2DBase):
         super(SquareWidget, self)._set_patch()
 
     def _onjumpclick(self, event):
-        # Callback for MPL pick event
-        if event.key == "shift" and event.inaxes:
+        if event.key == "shift" and event.inaxes and self.is_pointer:
             self.position = (event.xdata, event.ydata)
-
-    def _pos_changed(self):
-        """Call when the position of the widget has changed. It triggers the
-        relevant events, and updates the patch position.
-        """
-        if self._navigating:
-            with self.axes_manager.events.indices_changed.suppress():
-                for i in range(len(self.axes)):
-                    self.axes[i].value = self._pos[i]
-            self.axes_manager.events.indices_changed.trigger(obj=self.axes_manager)
-        self.events.moved.trigger(self)
-        self.events.changed.trigger(self)
-        self._update_patch_position()
 
     def _onmousemove(self, event):
         """on mouse motion move the patch if picked"""
--- a/hyperspy/drawing/_widgets/rectangles.py
+++ b/hyperspy/drawing/_widgets/rectangles.py
@@ -62,7 +62,7 @@ class SquareWidget(Widget2DBase):
 
     def _onjumpclick(self, event):
         # Callback for MPL pick event
-        if event.key=="shift" and event.inaxes:
+        if event.key == "shift" and event.inaxes:
             self.position = (event.xdata, event.ydata)
 
     def _pos_changed(self):
@@ -70,7 +70,10 @@ class SquareWidget(Widget2DBase):
         relevant events, and updates the patch position.
         """
         if self._navigating:
-            with self.axes_manager.events.indices_changed.suppress():
+            with (self.axes_manager.events.indices_changed.suppress_callback(
+                  self._on_navigate) or
+                  self.axes_manager.events.indices_changed.suppress_callback(
+                  self._onjumpclick)):
                 for i in range(len(self.axes)):
                     self.axes[i].value = self._pos[i]
         self.events.moved.trigger(self)
--- a/hyperspy/drawing/_widgets/rectangles.py
+++ b/hyperspy/drawing/_widgets/rectangles.py
@@ -62,7 +62,7 @@ class SquareWidget(Widget2DBase):
 
     def _onjumpclick(self, event):
         # Callback for MPL pick event
-        if event.key=="shift":
+        if event.key=="shift" and event.inaxes:
             self.position = (event.xdata, event.ydata)
 
     def _onmousemove(self, event):
--- a/hyperspy/drawing/widget.py
+++ b/hyperspy/drawing/widget.py
@@ -323,6 +323,7 @@ class DraggableWidgetBase(WidgetBase):
 
     def __init__(self, axes_manager, **kwargs):
         super(DraggableWidgetBase, self).__init__(axes_manager, **kwargs)
+        self.is_pointer=False
         self.events.moved = Event(doc="""
             Event that triggers when the widget was moved.
 
@@ -384,10 +385,10 @@ class DraggableWidgetBase(WidgetBase):
         relevant events, and updates the patch position.
         """
         if self._navigating:
-            with self.axes_manager.events.indices_changed.suppress_callback(
-                    self._on_navigate):
+            with self.axes_manager.events.indices_changed.suppress():
                 for i in range(len(self.axes)):
                     self.axes[i].value = self._pos[i]
+            self.axes_manager.events.indices_changed.trigger(obj=self.axes_manager)
         self.events.moved.trigger(self)
         self.events.changed.trigger(self)
         self._update_patch_position()
@@ -457,6 +458,14 @@ class DraggableWidgetBase(WidgetBase):
         self.cids.append(canvas.mpl_connect('pick_event', self.onpick))
         self.cids.append(canvas.mpl_connect(
             'button_release_event', self.button_release))
+        canvas.mpl_connect('button_press_event', self._onjumpclick)
+
+    def _onjumpclick(self, event):
+        """This method must be provided by subclasses"""
+        pass
+        if event.key == "shift" and event.inaxes:
+            self.position = (event.xdata, event.ydata)
+
 
     def _on_navigate(self, axes_manager):
         if axes_manager is self.axes_manager:
--- a/hyperspy/drawing/widget.py
+++ b/hyperspy/drawing/widget.py
@@ -384,7 +384,8 @@ class DraggableWidgetBase(WidgetBase):
         relevant events, and updates the patch position.
         """
         if self._navigating:
-            with self.axes_manager.events.indices_changed.suppress():
+            with self.axes_manager.events.indices_changed.suppress_callback(
+                    self._on_navigate):
                 for i in range(len(self.axes)):
                     self.axes[i].value = self._pos[i]
         self.events.moved.trigger(self)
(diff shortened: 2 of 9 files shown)


magnunor:
This is very nice! I tested this a little bit, some comments:

- Currently `shift` + left **and** `shift` + right click "jumps" the navigation position. Should this be the default behavior?
- There is something strange going on: it seems like it is showing another position before jumping to the one I'm clicking. See example below: the "extra" position is the amorphous diffraction pattern. It might be visible here, as this is a lazy dataset being loaded from the hard drive.


output

- `Shift` + clicking outside the canvas give an error:
```python
hyperspy/hyperspy/drawing/_widgets/rectangles.py", line 66, in _onjumpclick
    self.position = (event.xdata, event.ydata)
    ^^^^^^^^^^^^^
hyperspy/hyperspy/drawing/widget.py", line 429, in <lambda>
    lambda s, v: s._set_position(v))
                 ^^^^^^^^^^^^^^^^^^
hyperspy/hyperspy/drawing/widget.py", line 423, in _set_position
    position = self._validate_pos(position)
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
hyperspy/hyperspy/drawing/widget.py", line 405, in _validate_pos
    pos = np.maximum(pos, [ax.low_value for ax in self.axes])
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
TypeError: '>=' not supported between instances of 'NoneType' and 'float'
```

magnunor:
> There is something strange going on: it seems like it is showing another position before jumping to the one I'm clicking. See example below: the "extra" position is the amorphous diffraction pattern. It might be visible here, as this is a lazy dataset being loaded from the hard drive.

This has some performance implications, especially for large lazy datasets, since it means that (potentially) two chunks are loaded into memory for each "jump".

I think this is due to this part of the code:

```python
        if np.ndim(value) == 0 and len(self.axes) == 1:
            self.position = [self.axes[0].index2value(value)]
        elif len(self.axes) != len(value):
            raise ValueError()
        else:
            p = []
            for i in range(len(self.axes)):
                p.append(self.axes[i].index2value(value[i]))
            self.position = p

    indices = property(lambda s: s._get_indices(),
                       lambda s, v: s._set_indices(v))
```

This part of the code: https://github.com/hyperspy/hyperspy/blob/d0a266acbf11655fd3c9296441773e1c1abbc820/hyperspy/drawing/widget.py#L363

CSSFrancis:
> This is very nice! I tested this a little bit, some comments:

Thanks for testing this! Always good to see where things might go wrong on a different computer

> * Currently `shift` + left **and** `shift` + right click "jumps" the navigation position. Should this be the default behavior?

I think yes?  Unless we want to make the click and drag marker behavior only work with the "left-click".  It wouldn't be too hard to do that. 

> * There is something strange going on: it seems like it is showing another position before jumping to the one I'm clicking. See example below: the "extra" position is the amorphous diffraction pattern. It might be visible here, as this is a lazy dataset being loaded from the hard drive.
>
> output

This is weird... I'm having a hard time replicating this as well.  My best guess here is that your computer is sending a click first to the edge of the canvas and then to the point you actually selected.  I've tested this and at least for my computer I don't think it is happening but I'm not sure.
 
> * `Shift` + clicking outside the canvas give an error:
> 
> ```python
> hyperspy/hyperspy/drawing/_widgets/rectangles.py", line 66, in _onjumpclick
>     self.position = (event.xdata, event.ydata)
>     ^^^^^^^^^^^^^
> hyperspy/hyperspy/drawing/widget.py", line 429, in <lambda>
>     lambda s, v: s._set_position(v))
>                  ^^^^^^^^^^^^^^^^^^
> hyperspy/hyperspy/drawing/widget.py", line 423, in _set_position
>     position = self._validate_pos(position)
>                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
> hyperspy/hyperspy/drawing/widget.py", line 405, in _validate_pos
>     pos = np.maximum(pos, [ax.low_value for ax in self.axes])
>           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
> TypeError: '>=' not supported between instances of 'NoneType' and 'float'
> ```

This is another thing that I'm not getting an error with but regardless it should be fixed by adding the `event.inaxes` line for some more strict checking.

CSSFrancis:
> > There is something strange going on: it seems like it is showing another position before jumping to the one I'm clicking. See example below: the "extra" position is the amorphous diffraction pattern. It might be visible here, as this is a lazy dataset being loaded from the hard drive.
> 
> This has some performance implications, especially for large lazy datasets, since it means that (potentially) two chunks are loaded into memory for each "jump".

Yea I think that this is a potential issue but I can't seem to replicate it.  My best guess is that it is a bug related to an older `matplotlib` version and has been fixed?  

> I think this is due to this part of the code:
> 
> ```python
>         if np.ndim(value) == 0 and len(self.axes) == 1:
>             self.position = [self.axes[0].index2value(value)]
>         elif len(self.axes) != len(value):
>             raise ValueError()
>         else:
>             p = []
>             for i in range(len(self.axes)):
>                 p.append(self.axes[i].index2value(value[i]))
>             self.position = p
> 
>     indices = property(lambda s: s._get_indices(),
>                        lambda s, v: s._set_indices(v))
> ```
> 
> This part of the code:
> 
> https://github.com/hyperspy/hyperspy/blob/d0a266acbf11655fd3c9296441773e1c1abbc820/hyperspy/drawing/widget.py#L363

Possibly... I can try it on a different computer/ operating system and see.  I wrote some tests so we __should__ be able to test this.

hakonanes:
Thank you for quickly writing up this PR on discontinuous navigation after our offline discussion, @CSSFrancis!

Plotting geometrical simulations on top of EBSD patterns using markers is helpful when evaluating indexing results. We use this approach extensively in kikuchipy ([see this tutorial](https://kikuchipy.org/en/stable/tutorials/hough_indexing.html#Validate-indexing-results)). My issue with continuous navigation is that all markers are created and added to a signal before navigating. This operation is memory-intensive in kikuchipy. I hope discontinuous navigation can enable lazy markers and computation of one marker at a time.

I tried this PR out locally, and it works nicely! However, I'm seeing the same behavior as @manunor:

> it seems like it is showing another position before jumping to the one I'm clicking

magnunor:
> Yea I think that this is a potential issue but I can't seem to replicate it. My best guess is that it is a bug related to an older matplotlib version and has been fixed?

This is with `matplotlib` version `3.7.1`, which is the most recent one. I think the problem is somewhere in HyperSpy. Adding a `print` for debugging:

```python
    def _onjumpclick(self, event):
        # Callback for MPL pick event
        if event.key=="shift" and event.inaxes:
            print(event.xdata, event.ydata)
            self.position = (event.xdata, event.ydata)
```
Running this, it only outputs one event per click.

magnunor:
My guess is that this is only really visible when loading a large dataset from the hard drive, maybe via a HDF5 (`.hspy`) file.

hakonanes:
> this is only really visible when loading a large dataset from the hard drive, maybe via a HDF5 (.hspy) file.

By 'this', do you mean the extra navigation positing showing? I see this behavior when running @CSSFrancis's example in the top comment, using a Dask array. With Matplotlib v3.7.1.

magnunor:
Another "bug": shift + clicking on the colorbar also works:

output

magnunor:
> By 'this', do you mean the extra navigation positing showing? I see this behavior when running @CSSFrancis's example in the top comment, using a Dask array. With Matplotlib v3.7.1.

Yes. I'm running Ubuntu. Could it be a Ubuntu specific issue?

hakonanes:
I'm also running Ubuntu.

CSSFrancis:
> > By 'this', do you mean the extra navigation positing showing? I see this behavior when running @CSSFrancis's example in the top comment, using a Dask array. With Matplotlib v3.7.1.
> 
> Yes. I'm running Ubuntu. Could it be a Ubuntu specific issue?

I'm running on Ubuntu 20.04.6 as well with Matplotlib v3.7.1 as well and haven't been able to replicate it.  I tried on my M1 macbook as well and couldn't replicate it there either.

CSSFrancis:
> > Yea I think that this is a potential issue but I can't seem to replicate it. My best guess is that it is a bug related to an older matplotlib version and has been fixed?
> 
> This is with `matplotlib` version `3.7.1`, which is the most recent one. I think the problem is somewhere in HyperSpy. Adding a `print` for debugging:
> 
> ```python
>     def _onjumpclick(self, event):
>         # Callback for MPL pick event
>         if event.key=="shift" and event.inaxes:
>             print(event.xdata, event.ydata)
>             self.position = (event.xdata, event.ydata)
> ```
> 
> Running this, it only outputs one event per click.

This is the part that I am stuck on... the position is changed which should call the `_pos_changed` function https://github.com/hyperspy/hyperspy/blob/70ca0762a332a7239a4c1eb75001062ab6c7495d/hyperspy/drawing/widget.py#L382-L393C13

You could try seeing if this function is getting called twice.

It might also be that the `self.axes_manager.events.indices_changed.suppress_callback` function isn't working properly so the `self.axes[i].value = self._pos[i]` value is not working properly.

CSSFrancis:
> It might also be that the `self.axes_manager.events.indices_changed.suppress_callback` function isn't working properly so the `self.axes[i].value = self._pos[i]` value is not working properly.

That is exactly what is happening!  You can kind of see it on the video where one of the indexes switches first and the second one switches a little bit after.

CSSFrancis:
Okay now try it out... It should be fixed now.  Only some of the events where suppressed causing the `update` function to still trigger.  @magnunor good catch! 

It was happening on my computer but apparently it was flashing too fast :)

@ericpre do you have any idea if changing from `suppress_callback(self._on_navigate)` -> `supress()` will have any larger effects?

CSSFrancis:
> Thank you for quickly writing up this PR on discontinuous navigation after our offline discussion, @CSSFrancis!
> 
> Plotting geometrical simulations on top of EBSD patterns using markers is helpful when evaluating indexing results. We use this approach extensively in kikuchipy ([see this tutorial](https://kikuchipy.org/en/stable/tutorials/hough_indexing.html#Validate-indexing-results)). My issue with continuous navigation is that all markers are created and added to a signal before navigating. This operation is memory-intensive in kikuchipy. I hope discontinuous navigation can enable lazy markers and computation of one marker at a time.

@hakonanes Lazy markers is something that I think would be a great thing to add!  Most of the groundwork in #3148 is there (I just need to allow a `MarkerCollection` to accept dask arrays) and I kind of wrote that with the concept of lazy markers in mind.

There are some potential drawbacks of lazy plotting markers (mostly that you can't operate in parallel to calculate them so it could be slow to plot them/ move from one chunk to the next).  There are potential ways around that so maybe we start by adding the feature and go from there.

magnunor:
Seems like there is something strange going on now with the suppression, now it doesn't update the signal:
 
file

CSSFrancis:
> Seems like there is something strange going on now with the suppression, now it doesn't update the signal:
> 
> file

Okay maybe I have found the bug but introduced some new ones :)  I'm going to have to look into the suppress code and figure out exactly which function to suppress rather than blanket suppressing all of them. It might just be the new `jumpclick` function.

sivborg:
Hello, I am also quite fond of this idea, so I can help out with some testing!
I have tried out the implementation on my platform. I use HyperSpy through a Linux server running JupyterHub through TLJH, with `%matplotlib widgets` used when plotting. I am currently running into the same issues as @magnunor above, but I will try again when there is an update :)

sivborg:
This fix makes it work as intended on my set-up!

magnunor:
That fixed the issue with the position not changing (https://github.com/hyperspy/hyperspy/pull/3175#issuecomment-1599278955), but the "double jump" is still there (the one in the animation in this comment: https://github.com/hyperspy/hyperspy/pull/3175#issuecomment-1598489774)

CSSFrancis:
Hmmm let me write some more tests and we can go from there.

CSSFrancis:
I added two tests to make sure that only one `index_changed` event was occurring as well as a test to make sure that the pick- drag event displays all of the points in-between (only once). 

I think that tests the two bugs that we were seeing here.

magnunor:
That did indeed fixed the issue! Now it works nicely, and there is one "jump" per click.

CSSFrancis:
> That did indeed fixed the issue! Now it works nicely, and there is one "jump" per click.

Thanks for all of your testing and finding some of the bugs I missed :). Probably should have just written the event tests first.  There also should probably be more mock event tests in general so this is probably a good start. 

@ericpre when you have a chance, can you look at how I handled the event suppression before this is merged?

magnunor:
> Thanks for all of your testing and finding some of the bugs I missed :). Probably should have just written the event tests first. There also should probably be more mock event tests in general so this is probably a good start.

Yeah, I find that for some features or changes that is a nice workflow: adding the tests first, then implementing the change.

hyperspy/drawing/_widgets/rectangles.py
```diff
@@ -55,6 +60,24 @@ def _set_patch(self):
...
+
+    def _pos_changed(self):
+        """Call when the position of the widget has changed. It triggers the
+        relevant events, and updates the patch position.
+        """
+        if self._navigating:
+            with self.axes_manager.events.indices_changed.suppress():
+                for i in range(len(self.axes)):
+                    self.axes[i].value = self._pos[i]
+            self.axes_manager.events.indices_changed.trigger(obj=self.axes_manager)
```
ericpre: Instead to making a `SquareWidget._pos_changed` the code should be changed in `DraggableWidgetBase._pos_changed`?

hyperspy/drawing/_widgets/rectangles.py
```diff
@@ -40,6 +40,11 @@ class SquareWidget(Widget2DBase):
     def __init__(self, axes_manager, **kwargs):
         super(SquareWidget, self).__init__(axes_manager, **kwargs)
 
+    def connect(self, ax):
```
ericpre: These changes should go in the `DraggableWidgetBase.connect`

hyperspy/drawing/_widgets/rectangles.py
```diff
@@ -55,6 +60,24 @@ def _set_patch(self):
             picker=True,)]
         super(SquareWidget, self)._set_patch()
 
+    def _onjumpclick(self, event):
```
ericpre, line 58: I guess, this could be move to a more generic class, such as `Widget2DBase`? For 1D widget, it may be necessary to move in the child class and update accordingly.

ericpre (commented):
This is useful., however, this is working only the rectangle widget. It shouldn't too difficult to make it work for all widgets.

CSSFrancis:
> This is useful., however, this is working only the rectangle widget. It shouldn't too difficult to make it work for all widgets.

Yea I think that it would be good.  I can add it for the `VerticalLine` class as well.  Working for all widgets might be a little more complicated when you have multiple widgets on one plot.  I'll look into it and see what I can do.

ericpre:
Indeed, the several widgets situation needs to be handle correctly, currently, when adding several square widgets, the issue will occur. The widgets have a `selected` attribute, which could be used to know if the widget needs to be moved or not: move the widget only when `selected is True` and there should be only one widget selected a time. However, I wouldn't be surprised if it is a bit buggy and an alternative would be to keep this functionality private and enable it only when used as pointer and come back to this in a separate PR.

CSSFrancis:
@ericpre I added in the ability to use the jump-click method with the Horizontal and Vertical plots and refactored things based on your suggestions.  I also added some tests for testing the dragging a Line Marker

upcoming_changes/3175.new.rst
```diff
@@ -0,0 +1,2 @@
+Added in the `_onjumpclick` method to pointer widgets which allows the user to quickly go to some position
```
ericpre: As this is primiraly user facing, it would be to word as:
```suggestion
Add functionality to select navigation position using ``shift`` + click in the navigator.
```

hyperspy/drawing/widget.py
```diff
@@ -457,6 +458,14 @@ def connect(self, ax):
         self.cids.append(canvas.mpl_connect('pick_event', self.onpick))
         self.cids.append(canvas.mpl_connect(
             'button_release_event', self.button_release))
+        canvas.mpl_connect('button_press_event', self._onjumpclick)
+
+    def _onjumpclick(self, event):
+        """This method must be provided by subclasses"""
+        pass
+        if event.key == "shift" and event.inaxes:
+            self.position = (event.xdata, event.ydata)
```
ericpre:
```suggestion
        if event.key == "shift" and event.inaxes:
            self.position = (event.xdata, event.ydata)
```

hyperspy/drawing/_widgets/rectangles.py
```diff
@@ -40,6 +40,10 @@ class SquareWidget(Widget2DBase):
     def __init__(self, axes_manager, **kwargs):
         super(SquareWidget, self).__init__(axes_manager, **kwargs)
 
+    def connect(self, ax):
+        super(SquareWidget, self).connect(ax)
+        canvas = ax.figure.canvas
```
ericpre: Some leftover, which needs to be removed!
```suggestion
    def connect(self, ax):
        super(SquareWidget, self).connect(ax)
        canvas = ax.figure.canvas
```

ericpre (approved):
That's brilliant! Just need some clean-up.

ericpre:
It seems that the very first event with `shift` key press is ignored if no previous matplotlib event was triggered. It may well come from matplotlib side: for example, when the figure is resized before trying triggering a matplotlib event, then it works as expected (first event not ignored) and this is suspicious!

CSSFrancis:
> It seems that the very first event with `shift` key press is ignored if no previous matplotlib event was triggered. It may well come from matplotlib side: for example, when the figure is resized before trying triggering a matplotlib event, then it works as expected (first event not ignored) and this is suspicious!

I think this is just because the figure needs to be active/ selected?  The same thing applies to using the arrow keys where you need to have clicked on the figure first.  It's kind of frustrating but I'm not really sure how you would fix that?

I'd have to dig into the `event.inaxes` but I suspect that the first click doesn't trigger event.inaxes.

ericpre:
> I think this is just because the figure needs to be active/ selected? The same thing applies to using the arrow keys where you need to have clicked on the figure first. It's kind of frustrating but I'm not really sure how you would fix that?

Yes, I thought so too but if this was the case, it would do it too for the click and drag of the pointer, which works fine at first attempt. For some reason, `shift` is ignore until the window is selected while other event are triggered fine... This is suing `qt` and `widget` backends on windows, I haven't tested other platform.
I can't think of a reason why it would be related to this PR, so this is good as it is!

Issue #2997, opened 2022-08-25
_Map_iterate is always Parallel


CSSFrancis:
#### _Map_iterate is always Parallel
Not sure if this is actually a bug or something that we should keep as the default.  Regardless it should be better documented. 

As every array which is passed to _map_iterate is cast to a dask array it is always run in parallel. The `parallel` kwarg isn't really used and the `num_workers` kwarg is only used when the data is stored in place or the data is computed on the spot for a non lazy output. 

#### Potential Solution
I think the best solution is to rewrite the documentation and remove the `parallel`  kwarg and maybe throw and warning when num_workers is passed to a lazy signal?

For the lazy case these arguments should be passed to the `s.compute` following #2971 or set when a scheduler is initialized. 

Just for reference this is how I usually handle setting my `dask` backend. 
```python
# using threaded scheduler
s.compute(num_workers=20)
```

```python
# set up a local cluster using distributed scheduler
from dask.distributed import Client
client = Client() 

s.compute(num_workers=20)
```

```python
# creating a remote distributed scheduler on a cluster using a SLURM Manager
from dask_jobqueue import SLURMCluster

cluster = SLURMCluster(cores=40,
                       memory ='120GB',
                       walltime="12:00:00")
from dask.distributed import Client
cluster.scale(1)
client = Client(cluster)

s.compute() #uses all available workers

```

jlaehne:
Close as #3198 merged?

CSSFrancis:
Yea

Issue #3152, opened 2023-05-12
Map function runs serial when forced to parallel


wehlgrundspitze:
#### Describe the bug
Despite indicating to the map function to perform the operation in parallel, it is executed in series and takes the time expected for serial execution.

#### To Reproduce
Minimum working example for parallel mapping as written in the hyperspy documentation:

```
import os
import time

import hyperspy.api as hs
import numpy as np

def slow_func(data):
    print("Func exec")
    time.sleep(1.)

    return data + 1

print(os.cpu_count())

s = hs.signals.Signal1D(np.arange(40).reshape((20, 2)))
print(s)

s.map(slow_func, parallel=True, max_workers=4)
```

#### Expected behavior
4 workers should run in parallel and the map function should finish in 5s. Instead it takes the 20s expected for serial operation.

#### Python environement:
 - HyperSpy version: 1.7.5
 - Python version: 3.10.6

#### Additional context

Since the parallel operation is performed internally via the dask framework, I went through the basic dask tutorial at https://tutorial.dask.org/02_array.html. The parallel operations within this tutorial were performed correctly with shorter execution time.

I found this behaivior on two different machines with Ubuntu 22.04. The phenomenon occurred while running the code in a Jupyter Nottebook as well es execute it with the c-python interpreter.

wehlgrundspitze:
Just for completeness: The print statement about the number of processors shows correctly 6 respectively 8 on the two machines.

CSSFrancis:
Admittedly this bug results from how we use dask to run things in parallel now. You need at least 4 chunks to run on 4 cores so we should be creating a dask array with 4 chunks.

If you make a bigger array it should work. And if you make a lazy array you can define the number of chunks correctly.

CSSFrancis:
For example this should work

```python
import os
import time

import hyperspy.api as hs
import numpy as np


def slow_func(data):
    print("Func exec")
    time.sleep(1.)

    return data + 1

print(os.cpu_count())

s = hs.signals.Signal1D(np.arange(40).reshape((20, 2)))
s=s.as_lazy()
s.rechunk(5)
print(s)

s.map(slow_func, parallel=True, max_workers=4, lazy_output=False)
```

CSSFrancis:
The issue is here:

Mostly just the first line where the signal is cast to a lazy signal.

[...]

PR #3179, opened 2023-06-22
Remove _map_all in favor of dask.compute(scheduler='single-threaded')

magnunor:
Currently there are two separate "tracks" for `map`:

1. `_map_iterate`
2. `_map_all`

-----
### `_map_iterate`

[Link to the code](https://github.com/hyperspy/hyperspy/blob/RELEASE_next_major/hyperspy/signal.py#L4967)

Utilizes `dask.array.map_blocks`, so it works with really big datasets, can return lazy output and can run in parallel. One disadvantage currently, is that can be difficult to debug as the error messages are typically nested within the error messages from the dask scheduler.

### `_map_all`

[Link to the code](https://github.com/hyperspy/hyperspy/blob/RELEASE_next_major/hyperspy/signal.py#L4948)

Runs by passing the whole `s.data` numpy array to the function given to `map`. However, it only works on a subset of `map`'s parameters, as it can for example not handle lazy outputs.

------
## The problem

Having these separate "tracks" can be a bit confusing, as the `map` function can potentially work very differently depending on the parameters given to `map`. 

-----
##  The solution?

This pull request removes `_map_all` and passes the `parallel` parameter to `_map_iterate`. If `parallel=True` (and `lazy_output=False`), it runs with whatever scheduler is the standard. If `parallel=False` (and `lazy_output=False`), it uses the `scheduler='single-threaded'`.

This `'single-threaded'` scheduler is well-suited for debugging, according to the dask documentation: https://docs.dask.org/en/stable/scheduler-overview.html#debugging-the-schedulers

Alternatively, we can keep `_map_all`, and just pass keep the changes to `parallel=False` + `_map_iterate`.

### Progress of the PR
- [x] Change implemented
- [ ] Add several if/else warnings for incompatible parameters
- [ ] Code cleanup
- [ ] update docstring
- [ ] update user guide
- [ ] add tests,
- [ ] ready for review.

### Example

```python
import numpy as np
from scipy.ndimage import rotate
import hyperspy.api as hs
s = hs.signals.Signal2D(np.random.random((10, 10, 20, 20)))
s1 = s.map(afunction, angle=30, reshape=True, inplace=False, parallel=False, lazy_output=False)
```

### Notes

I think getting the `map` function working optimally is pretty tricky, as it tries to abstract away a great deal of complexity. Making something which works smoothly for small datasets where parallelization and lazy processing is pointless, simultaneously as it should work for extremely large datasets where lazy processing is required and parallelization is essential. But having the same function which operates on each navigation position in the same way is very nice.

This addresses https://github.com/hyperspy/hyperspy/issues/2997
Should potentially be in for the 2.0.0 release? https://github.com/hyperspy/hyperspy/issues/2996

Of interest to @CSSFrancis.

### DIFF
--- a/hyperspy/signal.py
+++ b/hyperspy/signal.py
@@ -4731,7 +4731,7 @@ class BaseSignal(FancySlicing,
         self,
         function,
         show_progressbar=None,
-        parallel=None,
+        parallel=True,
         max_workers=None,
         inplace=True,
         ragged=None,
@@ -4912,57 +4912,29 @@ class BaseSignal(FancySlicing,
         # If the function has an `axes` or `axis` argument
         # we suppose that it can operate on the full array and we don't
         # iterate over the coordinates.
-        if not ndkwargs and not lazy_output and (self.axes_manager.signal_dimension == 1 and
-                             "axis" in fargs):
-            kwargs['axis'] = self.axes_manager.signal_axes[-1].index_in_array
-
-            result = self._map_all(function, inplace=inplace, **kwargs)
-        elif not ndkwargs and not lazy_output and "axes" in fargs and not parallel:
-            kwargs['axes'] = tuple([axis.index_in_array for axis in
-                                    self.axes_manager.signal_axes])
-            result = self._map_all(function, inplace=inplace, **kwargs)
-        else:
-            if show_progressbar is None:
-                from hyperspy.defaults_parser import preferences
-                show_progressbar = preferences.General.show_progressbar
-            # Iteration over coordinates.
-            result = self._map_iterate(
-                function,
-                iterating_kwargs=ndkwargs, # function argument(s) (iterating)
-                show_progressbar=show_progressbar,
-                ragged=ragged,
-                inplace=inplace,
-                lazy_output=lazy_output,
-                max_workers=max_workers,
-                output_dtype=output_dtype,
-                output_signal_size=output_signal_size,
-                **kwargs, # function argument(s) (non-iterating)
-            )
+        if show_progressbar is None:
+            from hyperspy.defaults_parser import preferences
+            show_progressbar = preferences.General.show_progressbar
+        # Iteration over coordinates.
+        result = self._map_iterate(
+            function,
+            iterating_kwargs=ndkwargs, # function argument(s) (iterating)
+            show_progressbar=show_progressbar,
+            ragged=ragged,
+            inplace=inplace,
+            parallel=parallel,
+            lazy_output=lazy_output,
+            max_workers=max_workers,
+            output_dtype=output_dtype,
+            output_signal_size=output_signal_size,
+            **kwargs, # function argument(s) (non-iterating)
+        )
         if not inplace:
             return result
         else:
             self.events.data_changed.trigger(obj=self)
 
     map.__doc__ %= (SHOW_PROGRESSBAR_ARG, PARALLEL_ARG, LAZY_OUTPUT_ARG, MAX_WORKERS_ARG)
-
-    def _map_all(self, function, inplace=True, **kwargs):
-        """
-        The function has to have either 'axis' or 'axes' keyword argument,
-        and hence support operating on the full dataset efficiently.
-        """
-        newdata = function(self.data, **kwargs)
-        if inplace:
-            self.data = newdata
-            self._lazy = False
-            self._assign_subclass()
-            self.get_dimensions_from_data()
-            return None
-        else:
-            sig = self._deepcopy_with_new_data(newdata)
-            sig._lazy = False
-            sig._assign_subclass()
-            sig.get_dimensions_from_data()
-            return sig
 
     def _map_iterate(
         self,
@@ -4971,6 +4943,7 @@ class BaseSignal(FancySlicing,
         show_progressbar=None,
         ragged=False,
         inplace=True,
+        parallel=True,
         output_signal_size=None,
         output_dtype=None,
         lazy_output=None,
@@ -5064,6 +5037,11 @@ class BaseSignal(FancySlicing,
             pbar = ProgressBar()
             pbar.register()
 
+        if not parallel:
+            scheduler = 'single-threaded'
+        else:
+            scheduler = None
+
         if inplace:
             if (
                 not self._lazy
@@ -5079,6 +5057,7 @@ class BaseSignal(FancySlicing,
                     self.data,
                     dtype=mapped.dtype,
                     compute=True,
+                    scheduler=scheduler,
                     num_workers=max_workers,
                 )
                 data_stored = True
@@ -5108,7 +5087,7 @@ class BaseSignal(FancySlicing,
         sig._assign_subclass()
 
         if not lazy_output and not data_stored:
-            sig.data = sig.data.compute(num_workers=max_workers)
+            sig.data = sig.data.compute(num_workers=max_workers, scheduler=scheduler)
 
         if show_progressbar:
             pbar.unregister()


CSSFrancis:
@magnunor Let me know if you want me to review any of this at any point (even if it isn't done yet).  Thank you for tackling this! 

A couple of thoughts.  From my understanding the `_map_all` function is supposed to act similar to how `duck_typing` in `numpy` is supposed to work.  This __kind of__ works and is a source of frustration for me because it would be __so__ cool if it works and scipy, dask, cupy and numpy all buy in.

For example if I call np.sum(x) where x is a `dask` array it will automatically call `da.sum` rather than `np.sum`.  This only really works for `numpy` ufuncs and from a `hyperspy` persepective ideally we should be able to do the same thing with `hyperspy` signals.  

That this you should be able to call `np.std(signal)` and it should just call `BaseSignal.std` or if `BaseSignal.std` isn't defined it just returns np.std(signal.data) and then casts it back to a hyperspy signal.  We kind of have this functionality but it should be better tested. (For example `np.std(signal)` no longer works because the `BaseSignal.std` function doesn't have a `dtype` kwarg.

What I want is the ability to say call `scipy.ndimage.filter(signal)` and have it return a signal.  Maybe something like this will never happen but [NEP22](https://numpy.org/neps/nep-0022-ndarray-duck-typing-overview.html) and [NEP18](https://numpy.org/neps/nep-0018-array-function-protocol.html) are both worth reading and I think after 3 years they are finally close to something.

I'll make a separate PR to make hyperspy more compliant, specifically we need a `signal.dtype` and `signal.shape` and tests to see what numpy functions work.  This should cover dask and numpy but not things like scipy etc.

This is a long way of saying you can probably remove the `_map_all` function but it should be replaced by better `duck_typing` in hyperspy.


Additionally we could maybe better handle functions that operate directly on the data.  An `apply()` function which applies some function and wraps it as a hyperspy signal, maintaining the axes if it is the same shape or making new ones if it doesn't might be useful. Something like below.

```python

def apply(self, func, axes_map, axis ,inplace=True, **kwargs): 
    """ Apply some function to signal.data and return a signal (or multiple signals).  For operations where
         where signal.data.shape == func(signal.data.shape) the axes are passed to the new signal.  For operations
         where signal.data.shape != func(signal.data.shape) the `axes_map` kwarg is used to replace (or remove) axes.

     Example
     ---------
     >>> from scipy.ndimage import gaussian_filter
     >>>  s.apply(gaussian_filter) 
    """
    new_data = func(self.data)
    ... # create signal and map axes etc.

```

Just a thought

CSSFrancis:
It might also be good to think about how to handle #3152 as well.  That just involves making many small chunks for when you are trying to operate on a dataset that is smaller than 100 mb. 

There is also the [apply_gufunc](https://docs.dask.org/en/stable/generated/dask.array.gufunc.apply_gufunc.html) function in dask now which I have considered replacing `blockwise` with.  It allows multiple outputs and might be a little easier to use. 

I got like 90% of the way through replacing it and then realized that multiple outputs in `dask` is really inefficient as you basically duplicate the task graph for every additional array you are computing.  For code reduction though it might be interesting, even if the time for every operation to complete is multiplied by number of outputs.  There is also probably a way to merge the task graphs to reduce the effect but I'm not sure. 

Sorry these are just a bunch of thoughts I've been having that never really got anywhere so I figured I would just throw them out and see if anything sticks.

magnunor:
> I got like 90% of the way through replacing it and then realized that multiple outputs in dask is really inefficient as you basically duplicate the task graph for every additional array you are computing. For code reduction though it might be interesting, even if the time for every operation to complete is multiplied by number of outputs. There is also probably a way to merge the task graphs to reduce the effect but I'm not sure.

Not sure if this is the same thing, but I tested using `da.compute` as it can accept multiple dask arrays as input and computes them.

For situations where the file reading is the bottleneck (`.hspy`) it seems to improve performance by quite a bit.

Example:

```python
import dask.array as da
from dask.diagnostics import ProgressBar
import hyperspy.api as hs

s = hs.load("test.hspy", lazy=True) # 8.6 GB, chunking (32, 32, 32, 32)

s_bs = s.get_direct_beam_position("blur", sigma=5)

circ_roi0 = hs.roi.CircleROI(cx=128, cy=128, r_inner=10, r=20)
circ_roi1 = hs.roi.CircleROI(cx=128, cy=128, r_inner=50, r=80)
circ_roi2 = hs.roi.CircleROI(cx=128, cy=128, r_inner=100, r=110)
s_adf0 = s.get_integrated_intensity(circ_roi0)
s_adf1 = s.get_integrated_intensity(circ_roi1)
s_adf2 = s.get_integrated_intensity(circ_roi2)
```

```python
with ProgressBar():
    data_output = da.compute((s_bs.data, s_adf0.data, s_adf1.data, s_adf2.data))
```

Takes about 70 seconds.

```python
s_adf0.compute()
s_adf1.compute()
s_adf2.compute()
s_bs.compute()
```

Takes 3.61 + 31.26 + 53.11 + 52.48 = 140.46 seconds.

-----

The same procedure with a `.zspy`-file with chunking `(32, 32, 32, 32)` there is a small improvement:

- `da.compute(...)`: 55 seconds
- `signal.compute()`: 60 seconds

CSSFrancis:
> > I got like 90% of the way through replacing it and then realized that multiple outputs in dask is really inefficient as you basically duplicate the task graph for every additional array you are computing. For code reduction though it might be interesting, even if the time for every operation to complete is multiplied by number of outputs. There is also probably a way to merge the task graphs to reduce the effect but I'm not sure.
> 
> Not sure if this is the same thing, but I tested using `da.compute` as it can accept multiple dask arrays as input and computes them.

Oh very cool! I guess `dask` is very good at optimizing the task graph before it computes anything so this make sense!  

I might pick this up again then just because there are a couple of instances where multiple outputs would be nice.  The main ones are things like returning some value and the error for each point as well but who knows what else people might be interested in.

Removing the `_map_all` function also allows me to use the `axis` kwarg which helps a fair but for making things more simple. 

```python
def get_strain(x):
    # Function to get strain and error
    return strain, err

strain, err = s.map(get_strain, inplace=False)

# slow compute
strain.compute()
err.compute()
# fast compute()
da.compute(strain,err)
```

CSSFrancis:
@magnunor what do you think about changing to using `da.gufunc` rather than blockwise?

I think this is motivation to make the change, and it seems more inline with what we are interested in. 


Just for example it would remove the need for the `process_chunkwise` function

```python
import dask.array as da
import numpy as np
a = da.ones((10,10,10,10), chunks=(2,2,10,10))
def sum_ax(d):
    return np.sum(d, axis=(0,1))
summed = da.apply_gufunc(sum_ax, "(i,j)->()" ,a, axes=[(2,3)])
```

And you could do things like this if for example you have your data chunked in the navigation dimension:

```python
import dask.array as da
import numpy as np
a = da.ones((10,10,10,10), chunks=(2,2,10,10))
def sum_ax(d):
    return np.sum(d, axis=(0,1))
summed = da.apply_gufunc(sum_ax, "(i,j)->()" ,a, axes=[(0,1)])
```

I wouldn't really recommend that for lazy signals  unless you have lots of checks on chunks but there are also instances where people might want to do that for non-lazy signals.  

It still require that the "core dimensions" are one chunk similar to how the signal axes are one chunk but this gives some flexibility to the map function similar to the `_map_all` function without the extra need to write custom vectorized functions.

magnunor:
Closing this, as I think adding the `parallel` parameter to `_map_iterate` can be done in a separate pull request: https://github.com/hyperspy/hyperspy/pull/3188

@CSSFrancis I haven't really looked that closely at `da.gufunc`, so not sure really.

PR #3186, opened 2023-06-27
plot residual

HanHsuanWu:
### Description of the change

As suggested by @thomasaarholt on gitter, I added a plot_residual argument to model1d.plot. 
This argument will plot a residual signal (Signal - Model) to the signal plot.
Currently, it has a bug when the size of signal and model are not the same. I am not sure what's the best way to do this so any help is appreciated.

### Progress of the PR
- [ ] add tests,
- [ ] ready for review.

```python
import hyperspy.api as hs
import numpy as np
s = hs.signals.Signal1D(np.tile(np.arctan(np.arange(-500, 500)), (10, 1)),dtype='float64')
s.add_gaussian_noise(std=0.1)
m = s.create_model()
line  = hs.model.components1D.Polynomial(order=2)
m.append(line)
m.fit()
m.plot(plot_residual=True)
```

### DIFF
--- a/hyperspy/models/model1d.py
+++ b/hyperspy/models/model1d.py
@@ -417,10 +417,8 @@ class Model1D(BaseModel):
                 component for component in component_list if component.active]
 
         if self.convolved is False or non_convolved is True:
-            if not ignore_channel_switches:
-                axis = self.axis.axis[self.channel_switches]
-            else:
-                axis = self.axis.axis
+            slice_ = slice(None) if ignore_channel_switches else self.channel_switches
+            axis = self.axis.axis[slice_]
             sum_ = np.zeros(len(axis))
             for component in component_list:
                 sum_ += component.function(axis)
@@ -717,24 +715,12 @@ class Model1D(BaseModel):
             s = ns
         return s
     
-    def _residual_for_plot(self, axes_manager):
-        """From an model1D object, the original signal is subtracted 
+    def _residual_for_plot(self,**kwargs):
+        """From an model1D object, the original signal is subtracted
         by the model signal then returns the residual
         """
 
-        old_axes_manager = None
-        if axes_manager is not self.axes_manager:
-            old_axes_manager = self.axes_manager
-            self.axes_manager = axes_manager
-            self.fetch_stored_values()
-
-        #Residual = Signal - model
-        s = (self.signal.__call__() - self.__call__(ignore_channel_switches = True))
-        if old_axes_manager is not None:
-            self.axes_manager = old_axes_manager
-            self.fetch_stored_values()
-
-        return s
+        return self.signal.__call__() - self.__call__(ignore_channel_switches=True)
     
     def plot(self, plot_components=False,plot_residual=False, **kwargs):
         """Plot the current spectrum to the screen and a map with a
--- a/hyperspy/models/model1d.py
+++ b/hyperspy/models/model1d.py
@@ -258,6 +258,7 @@ class Model1D(BaseModel):
         self._plot_components = False
         self._suspend_update = False
         self._model_line = None
+        self._residual_line = None
         self.axis = self.axes_manager.signal_axes[0]
         self.axes_manager.events.indices_changed.connect(
             self._on_navigating, [])
@@ -377,7 +378,8 @@ class Model1D(BaseModel):
     remove.__doc__ = BaseModel.remove.__doc__
 
     def __call__(self, non_convolved=False, onlyactive=False,
-                 component_list=None, binned=None):
+                 component_list=None, binned=None, 
+                 ignore_channel_switches = False):
         """Returns the corresponding model for the current coordinates
 
         Parameters
@@ -393,6 +395,9 @@ class Model1D(BaseModel):
         binned : bool or None
             Specify whether the binned attribute of the signal axes needs to be
             taken into account.
+        ignore_channel_switches: bool
+            If true, the entire signal axis are returned 
+            without checking channel_switches.
 
         cursor: 1 or 2
 
@@ -412,7 +417,10 @@ class Model1D(BaseModel):
                 component for component in component_list if component.active]
 
         if self.convolved is False or non_convolved is True:
-            axis = self.axis.axis[self.channel_switches]
+            if not ignore_channel_switches:
+                axis = self.axis.axis[self.channel_switches]
+            else:
+                axis = self.axis.axis
             sum_ = np.zeros(len(axis))
             for component in component_list:
                 sum_ += component.function(axis)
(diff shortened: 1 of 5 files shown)


hyperspy/models/model1d.py
```diff
@@ -710,15 +710,36 @@ def _model2plot(self, axes_manager, out_of_range2nans=True):
             ns[np.where(self.channel_switches)] = s
             s = ns
         return s
-
-    def plot(self, plot_components=False, **kwargs):
+    
+    def _residual2plot(self, axes_manager, out_of_range2nans=True):
+        old_axes_manager = None
+        if axes_manager is not self.axes_manager:
```
thomasaarholt: Could you document what happens here?

hyperspy/models/model1d.py
```diff
@@ -710,15 +710,36 @@ def _model2plot(self, axes_manager, out_of_range2nans=True):
             ns[np.where(self.channel_switches)] = s
             s = ns
         return s
-
-    def plot(self, plot_components=False, **kwargs):
+    
+    def _residual2plot(self, axes_manager, out_of_range2nans=True):
```
thomasaarholt: Hyperspy does use the `foo2bar` syntax here and there, but I think its generally considered better to use `foo_to_bar`. So here I'd suggest `_residual_to_plot`, or perhaps better `_residual_for_plot` and `out_of_range_to_nan`.

hyperspy/models/model1d.py
```diff
@@ -710,15 +710,36 @@ def _model2plot(self, axes_manager, out_of_range2nans=True):
...
+            old_axes_manager = self.axes_manager
+            self.axes_manager = axes_manager
+            self.fetch_stored_values()
+        #Residual = Signal - model
+        s = (self.signal.__call__() -
+            self.__call__(non_convolved=False, onlyactive=True))
+        if old_axes_manager is not None:
+            self.axes_manager = old_axes_manager
+            self.fetch_stored_values()
+        if out_of_range2nans is True:
```
thomasaarholt: Since there currently is no way to change the value of `out_of_range2nans` from a calling function, I suggest you just remove the `if` and always use the nan-behaviour.

If you do keep the `if`, then you can remove `is True`, since `if out_of_range2nans` will evaluate to `if True` or `if False`.

hyperspy/models/model1d.py
```diff
@@ -742,6 +763,21 @@ def plot(self, plot_components=False, **kwargs):
         self._model_line = l2
         self._plot = self.signal._plot
         self._connect_parameters2update_plot(self)
+        
+        #Optional to plot the residual of (Signal - Model)
+        if plot_residual is True:
```
thomasaarholt: `if plot_residual is True:` -> `if plot_residual:`

thomasaarholt (commented):
Thanks for the PR! I've added some general linting comments below as well as a few suggestions.

hyperspy/models/model1d.py
HanHsuanWu: I added some comment for what _residual_for_plot does.

HanHsuanWu:
Hey @thomasaarholt thanks for reviewing.
I have addressed your comments and ended up deleting the out_of_range2nans part.
I added a new argument for model1d.__call__() to ignore the channel_switches, so that it always plots the residual for the whole spectrum. We can add another argument to turn this on or off if needed. What do you think?

hyperspy/models/model1d.py
HanHsuanWu: I have added documentation for _residual_for_plot.

hyperspy/models/model1d.py
```diff
@@ -710,15 +718,36 @@ def _model2plot(self, axes_manager, out_of_range2nans=True):
...
+        if axes_manager is not self.axes_manager:
+            old_axes_manager = self.axes_manager
+            self.axes_manager = axes_manager
+            self.fetch_stored_values()
+
+        #Residual = Signal - model
+        s = (self.signal.__call__() - self.__call__(ignore_channel_switches = True))
+        if old_axes_manager is not None:
+            self.axes_manager = old_axes_manager
+            self.fetch_stored_values()
```
ericpre: This can be simplify to:

```suggestion
    def _residual_for_plot(self):
        """From an model1D object, the original signal is subtracted
        by the model signal then returns the residual
        """

        return self.signal.__call__() - self.__call__(ignore_channel_switches=True)

```
Github doesn't allow me to make a suggestion over a deleted line and `return s` below obviously also need to be removed in the above suggestion.

hyperspy/models/model1d.py
```diff
@@ -413,7 +418,10 @@ def __call__(self, non_convolved=False, onlyactive=False,
                 component for component in component_list if component.active]
 
         if self.convolved is False or non_convolved is True:
-            axis = self.axis.axis[self.channel_switches]
+            if not ignore_channel_switches:
+                axis = self.axis.axis[self.channel_switches]
+            else:
+                axis = self.axis.axis
```
ericpre: A more compact version:
```suggestion
            slice_ = slice(None) if ignore_channel_switches else self.channel_switches
            axis = self.axis.axis[slice_]
```

ericpre (commented):
Thank you @HanHsuanWu, this works well and is a long overdue!

I have suggested some simplification.

This would need a changelog entry.

hyperspy/models/model1d.py
HanHsuanWu: I added a keyword argument to _residual_for_plot to prevent the error message attached below.
def _residual_for_plot(self,**kwargs):
image

HanHsuanWu:
> Thank you @HanHsuanWu, this works well and is a long overdue!
> 
> I have suggested some simplification.
> 
> This would need a changelog entry.

What is a changelog entry and where do I add this?

ericpre:
See https://github.com/hyperspy/hyperspy/blob/RELEASE_next_minor/upcoming_changes/README.rst as per checklist of the first comment of the PR.
It is a note to describe the changes of this PR and will be added automatically to the list of changes (also called changelog) for each release.

HanHsuanWu:
> See https://github.com/hyperspy/hyperspy/blob/RELEASE_next_minor/upcoming_changes/README.rst as per checklist of the first comment of the PR. It is a note to describe the changes of this PR and will be added automatically to the list of changes (also called changelog) for each release.

Thanks Eric. I have added the changelog. Got a bit confused before.

HanHsuanWu:
@ericpre any further comments for this pull request?

ericpre:
Sorry, I don't have time at the minute, I will come back to it very soon.

ericpre (approved):
Thanks @HanHsuanWu, this looks good.
In 18a019a8c7db0ce36768d997b49af14a8c548588 I added a test and an example. I have also rebased on the `RELEASE_next_major` branch as the next release will be the 2.0.

PR #3199, opened 2023-07-20
N d navigator

CSSFrancis:
### Description of the change
This allows navigators with dimension >2 to be plot in a more reasonable (and interactive) way.

- Navigators will update when they have dimensions>2 
- Lazy signals will plot an image navigator when dimensions >2

### Progress of the PR
- [x] add tests,
- [x] ready for review.

```python
%matplotlib ipympl
import hyperspy.api as hs
import numpy as np
s = hs.signals.Signal2D(np.random.rand(10,10,10,10,10,10)).as_lazy()
s.plot()
```

https://github.com/hyperspy/hyperspy/assets/41125831/e07e1046-8a87-44b8-8bbe-1b2f999215e7

### DIFF
--- a/hyperspy/signal.py
+++ b/hyperspy/signal.py
@@ -3061,7 +3061,11 @@ class BaseSignal(
             nav_ind = am.indices[2:]  # image at first 2 nav indices
             slices = [slice(None)] * len(am.navigation_axes)
             slices[2:] = nav_ind
-            ind = navigator.isig.__getitem__(slices=slices)  # Get the value from the nav reverse because hyperspy
+            if len(navigator.axes_manager.signal_axes) > 0:
+                new_nav = navigator.transpose(signal_axes=len(am.navigation_axes)) # roll axes to signal axes
+            else:
+                new_nav=navigator
+            ind = new_nav.isig.__getitem__(slices=slices)  # Get the value from the nav reverse because hyperspy
             return np.nan_to_num(to_numpy(ind.data)).squeeze()
 
         def get_dynamic_explorer_wrapper(*args, **kwargs):
--- a/hyperspy/signal.py
+++ b/hyperspy/signal.py
@@ -3061,12 +3061,9 @@ class BaseSignal(
             nav_ind = am.indices[2:]  # image at first 2 nav indices
             slices = [slice(None)] * len(am.navigation_axes)
             slices[2:] = nav_ind
-            if len(navigator.axes_manager.signal_axes) > 0:
-                new_nav = navigator.transpose(
-                    signal_axes=len(am.navigation_axes)
-                )  # roll axes to signal axes
-            else:
-                new_nav = navigator
+            new_nav = navigator.transpose(
+                signal_axes=len(am.navigation_axes)
+            )  # roll axes to signal axes
             ind = new_nav.isig.__getitem__(
                 slices=slices
             )  # Get the value from the nav reverse because hyperspy
--- a/hyperspy/signal.py
+++ b/hyperspy/signal.py
@@ -3056,6 +3056,14 @@ class BaseSignal(
             navigator = sum_wrapper(self, am.signal_axes + am.navigation_axes[1:])
             return np.nan_to_num(to_numpy(navigator.data)).squeeze()
 
+        def get_dynamic_image_explorer(*args, **kwargs):
+            am = self.axes_manager
+            nav_ind = am.indices[2:]  # image at first 2 nav indices
+            slices = [slice(None)] * len(am.navigation_axes)
+            slices[2:] = nav_ind
+            ind = navigator.isig.__getitem__(slices=slices)  # Get the value from the nav reverse because hyperspy
+            return np.nan_to_num(to_numpy(ind.data)).squeeze()
+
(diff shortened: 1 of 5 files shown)


CSSFrancis:
@SyHuang19 You might find this interesting.

hyperspy/signal.py
```diff
@@ -2835,6 +2835,18 @@ def get_1D_sum_explorer_wrapper(*args, **kwargs):
...
+        def get_dynamic_image_explorer(*args, **kwargs):
+            am = self.axes_manager
+            nav_ind = am.indices[2:]  # image at first 2 nav indices
+            slices = [slice(None)] * len(am.navigation_axes)
+            slices[2:] = nav_ind
+            if len(navigator.axes_manager.signal_axes) > 0:
+                new_nav = navigator.transpose(signal_axes=len(am.navigation_axes)) # roll axes to signal axes
+            else:
+                new_nav=navigator
+            ind = new_nav.isig.__getitem__(slices=slices)  # Get the value from the nav reverse because hyperspy
```
jlaehne: comment incomplete?

CSSFrancis: More of just a terrible comment :) It took me a while to figure out what I was even trying to say so I might just delete it.

I just struggled to figure out if the index should be reversed or not because we are using the hyperspy index an not the array index.

jlaehne:
Not really working with so many dimensions, but sounds sensible :-)

hyperspy/signal.py
```diff
@@ -2835,6 +2835,18 @@ def get_1D_sum_explorer_wrapper(*args, **kwargs):
...
 
+        def get_dynamic_image_explorer(*args, **kwargs):
+            am = self.axes_manager
+            nav_ind = am.indices[2:]  # image at first 2 nav indices
+            slices = [slice(None)] * len(am.navigation_axes)
+            slices[2:] = nav_ind
+            if len(navigator.axes_manager.signal_axes) > 0:
+                new_nav = navigator.transpose(signal_axes=len(am.navigation_axes)) # roll axes to signal axes
+            else:
+                new_nav=navigator
```
ericpre: Add test for this code branch.

ericpre:

CSSFrancis:
> Re-opening because this has been closed automatically by mistake!

@ericpre Thank you!  Sorry this is why I shouldn't make code changes early in the morning :/

CSSFrancis:
pre-commit.ci autofix

CSSFrancis:
@jlaehne or @ericpre do you want to take another look at this?  I think this is a good start, it works well in the 5 dimensional case and somewhat well in the 6 dimensional case.   We could add a secondary navigator image but that should be a different PR.  It would also require more of a change.

hyperspy/signal.py
```diff
@@ -3056,15 +3056,18 @@ def get_1D_sum_explorer_wrapper(*args, **kwargs):
...
+            am = self.axes_manager
+            nav_ind = am.indices[2:]  # image at first 2 nav indices
+            slices = [slice(None)] * len(am.navigation_axes)
+            slices[2:] = nav_ind
+            new_nav = navigator.transpose(
+                signal_axes=len(am.navigation_axes)
+            )  # roll axes to signal axes
+            ind = new_nav.isig.__getitem__(
+                slices=slices
+            )  # Get the value from the nav reverse because hyperspy
```
ericpre, line 3069: I suspect that it should be possible to simplify this into:
```suggestion
            ind = navigator.inav.__getitem__(
                slices=self.axes_manager.indices[2:]
            )
```

However, this doesn't seem to play well with lazy signal, because the navigation shape is different, which shouldn't be.

hyperspy/signal.py
```diff
@@ -3124,21 +3127,20 @@ def is_shape_compatible(navigation_shape, shape):
...
                     navigator = navigator.T
-                    self._plot.navigator_data_function = get_static_explorer_wrapper
-                # Dynamic navigator
-                elif (
-                    axes_manager.navigation_shape
-                    == navigator.axes_manager.signal_shape
-                    + navigator.axes_manager.navigation_shape
-                ):
-                    self._plot.navigator_data_function = get_dynamic_explorer_wrapper
+                    if len(axes_manager.navigation_shape) > 2:
```
ericpre, line 3140: This branch doesn't seem to be covered?

ericpre (commented):
It works well, but I think we should fix the mismatch in navigation shape between lazy and non-lazy signals.  This would simplify a bit plotting code by removing some workaround.
If you are busy with other things, we get this PR merge and revisit later.

CSSFrancis:
> It works well, but I think we should fix the mismatch in navigation shape between lazy and non-lazy signals.  This would simplify a bit plotting code by removing some workaround.
> 
> If you are busy with other things, we get this PR merge and revisit later.


Hmm let me look through this and see what I can do. I'm not sure how much time I want to devote to this at the moment. I think that fixing this entirely might be a larger undertaking than I should take up at the moment...

ericpre:
No worries, this can be done in another PR, maybe try to cover the all new code.

ericpre (approved):

PR #3201, opened 2023-07-21
Split holography

CSSFrancis:
### Description of the change
Just starting this PR to track splitting the holography off. This should be fairly easy so we are going to use this as a warm up to splitting the EELS and EDS.

### Progress of the PR
- [x] Moved HologramImage to  separate folder
- [x] Add Hyperspy extension.yml
- [x] Move Documentation to separate docs
- [ ] Add setup.py file
- [ ] Fix testing so all new testing works
- [ ]  Add changelog
- [ ] Finish Split


ericpre:
For reference, I [wrote down](https://github.com/hyperspy/hyperspy/pull/2972#issuecomment-1193101728) the command that I used when splitting IO part using `git filter-repo`. The non-obvious things to keep the full history was to add **_all_** folders, including the ones that have been renamed!

CSSFrancis:
> For reference, I [wrote down](https://github.com/hyperspy/hyperspy/pull/2972#issuecomment-1193101728) the command that I used when splitting IO part using `git filter-repo`. The non-obvious things to keep the full history was to add **_all_** folders, including the ones that have been renamed!

@ericpre Thanks!  I was mostly going off of that thread for most of the changes.

I think I (mostly) have the right procedure.  This change should be much easier than the I-O code (and the EELS/EDS) mostly because the extension framework is already there and there aren't really any parts of the hyperspy testing code that are dependent on the holography code.

ericpre:
Sounds good.
I would not worry to get too much about getting everything (doc, installation) sorted at this stage, but the important part is not to forget files required in the new library in order to keep the history.

CSSFrancis:
> Sounds good.
> I would not worry to get too much about getting everything (doc, installation) sorted at this stage, but the important part is not to forget files required in the new library in order to keep the history.

Yea I think I've got most of it.  I searched for holography in the repo and just made sure I moved everything there.  That being said a second (or third) set of eyes is always helpful.

This one was pretty easy to do, the EELS and EDS will be a bit more of a pain based on the couple of times I've tried it but I think I can do that as well. 

The Installation should work (mostly) but the docs will need some triage later.

ericpre (approved):

ericpre:
@CSSFrancis, do you want to do the following steps?
1. merge this PR
2. run `git filter-repo`
3. create the `holospy` repository

For convenience, the relevant github documentation: https://docs.github.com/en/get-started/using-git/splitting-a-subfolder-out-into-a-new-repository

Then we are happy with everything, we can make another PR to clean up the `holospy` folder in this repository.

